PR #5219, opened 2026-04-21
[rocprofiler-sdk] Queue Interposition: shadow write pointer alternative to queue interception

bwelton:
## Summary

- SDK-level write-pointer virtualization to intercept HSA queue operations without modifying ROCR-Runtime
- Introduces `QueueState` struct and registry for per-queue state tracking (virtual wptr, ring buffer, doorbell)
- Wraps 12 `hsa_queue_*_write_index_*` and 4 `hsa_signal_store_*` functions via `CoreApiTable` interposition installed after `update_table` (chains through the tracing functors)
- Deferred-WDID model: the real `write_dispatch_id` only advances at doorbell time after `process_doorbell_impl` runs the WriteInterceptor chain
- Tracing-only: the gate in `registration.cpp` forces the legacy `hsa_amd_queue_intercept_create` path whenever a context registers `dispatch_counter_collection`, `dispatch_thread_trace`, or `pc_sampler`
- Default-on via `ROCPROFILER_QUEUE_INTERPOSITION` env var when none of the above are registered
- New `signal_pool.{hpp,cpp}` (refactored out of `queue.cpp`) and `internal_threading::TaskGroup` extensions (sized pool, fire-and-forget `async()`, `async_only` wait/join)

## Key files
- `source/lib/rocprofiler-sdk/hsa/queue_interposition.hpp` (~250 lines) — public API
- `source/lib/rocprofiler-sdk/hsa/queue_interposition.cpp` (~1160 lines) — implementation
- `source/lib/rocprofiler-sdk/tests/queue_interposition.cpp` (~354 lines) — 13 unit tests
- `source/lib/rocprofiler-sdk/hsa/signal_pool.{hpp,cpp}` — common signal pool
- `source/lib/rocprofiler-sdk/hsa/queue_controller.cpp` — integration (forks legacy vs interposition at queue create)
- `source/lib/rocprofiler-sdk/registration.cpp` — env-var gating + post-`update_table` install
- `source/lib/rocprofiler-sdk/internal_threading.{hpp,cpp}` — TaskGroup pool sizing + async()

## Test plan
- [x] 13 `unit.queue_interposition.*` tests pass (no HSA/HIP dependency)
- [x] Integration tests: `hip-graph-bubbles` (stresses doorbell drain with large CLR batches; this surfaced the doorbell-value-vs-virtual_wptr bug fixed in 944e25b) and `openmp-target` (exercises the non-HIP queue creation path)
[...]

PR #7420, opened 2026-06-17
[DO NOT MERGE] exp LKG: extsem patch + pre-#5219 develop (expect PASS)

jungx098:
## DO NOT MERGE — bisect control (with patch)

Tests `2a108a17` = #6632 extsem patch + develop **before** #5219 (queue interposition). Base is its real parent `5256d41b` so the merge ref IS this exact tree (no merge-up to current develop, no conflicts). Expect no_bubbles + transpose-sampling PASS. Re-run ~5x for bimodal. Delete after.


sputhala-amd:

PR #7421, opened 2026-06-17
[DO NOT MERGE] exp BAD: extsem patch + post-#5219 develop (expect FAIL)

jungx098:
## DO NOT MERGE — bisect isolation (with patch)

Tests `6ae3da33` = #6632 extsem patch + develop **after** #5219. Base is its real parent `2a108a17` (the LKG), so the diff is exactly the 4 develop commits dominated by #5219 (f642705661 'Queue Interposition: shadow write pointer'), with the patch held constant in base and head. Merge ref IS this tree (no conflicts). Expect no_bubbles + transpose-sampling FAIL. Re-run ~5x for bimodal. Delete after.


sputhala-amd:

sputhala-amd:

PR #7424, opened 2026-06-18
[rocprofiler-compute] Add profile data boundary architecture design doc

jamessiddeley-amd:
## Motivation

Profile data storage knowledge (CSV vs ROCPD, on-disk layout, file names) is currently spread thinly across profiling, analysis, and utility code so every new storage format touches many unrelated modules. This adds a design doc proposing a single profile data boundary with clear reader/writer interfaces, ahead of upcoming work for default rocpd profiling outputs, and other file formats.

## Technical Details

Adds `docs/design/clean_architecture.md` describing:
- A reader/writer contract (`ProfileDataReader` / `ProfileDataWriter`) in `src/interface/`, with format selection via `get_reader` / `get_writer` so no other module knows the format or disk location
- CSV and ROCPD implementations behind the boundary, with rocpd normalization owned by `rocpd_data.py`
- Mermaid diagrams for the current state, the Phase 1 target, and follow-up phases (backend orchestration, Profiler Hub / future formats)
- A data-ownership table separating profile source data, the `pmc_perf.csv` compatibility view, and derived analysis artifacts

The design is validated against the in-flight Profiler Hub work #7284: its native-tool rocpd write + set-based SQL merge fits inside interface/rocpd_data.py behind the existing writer, with no changes to profiling or analyze callers.

Docs-only; no code changes

## JIRA ID

N/A

## Test Plan

N/A

## Test Result

N/A

## Submission Checklist


projects/rocprofiler-compute/docs/design/clean_architecture.md
```diff
@@ -0,0 +1,241 @@
...
+
+- `ProfileDataWriter.finalize_pass(context)` records a completed profiling pass
+  in the selected storage format.
+- `ProfileDataReader.has_profile_data(workload_dir)` reports whether readable
+  profile data exists.
+- `ProfileDataReader.read_pmc_frame(workload_dir)` returns the PMC DataFrame used
+  by analysis.
+- `ProfileDataReader.materialize_pmc_perf(workload_dir, output_path)` produces
+  `pmc_perf.csv` only when a compatibility file is needed.
+
```
vedithal-amd: **Accuracy against current `rocprofiler-compute-develop`**

A few claims have drifted from the current branch and should be reconciled so the design reflects today's code:

- **rocpd is already the default** in both profile (`utils_profile.py:749`) and analyze join (`analysis_base.py:462`). The "Follow-up: Default rocpd" section reads as future work, but it has already landed. Please reframe it as "consolidate the already-default rocpd path" rather than "make rocpd default."
- **`pmc_perf.csv` is analyze-produced, not profile output.** The join writes it (`analysis_base.py:476`) and `create_df_pmc` reads it back (`file_io.py:316`). The "today" diagram visually groups it under profile, which is misleading. Note also that the csv intermediate is already on a deprecation path, so the design should align with its removal rather than preserve it.
- **`materialize_pmc_perf(...)` "only when a compatibility file is needed" is a silent behavior change.** Today pmc_perf is always written in analyze. Making it on-demand needs an explicit justification; otherwise it should stay produced by default through the boundary.

projects/rocprofiler-compute/docs/design/clean_architecture.md
```diff
@@ -0,0 +1,241 @@
...
+This is the only place that maps a configured format to an implementation. Both
+functions live in `interface/profile_data.py` next to the contracts they return.
+
+### Modules
+
+| File | Responsibility |
+| --- | --- |
+| `interface/profile_data.py` | Reader/writer protocols, options dataclasses, `get_reader` / `get_writer` selection, and shared post-load formatting (`prepare_pmc_frame`). |
+| `interface/csv_data.py` | CSV reader and writer implementation. |
+| `interface/rocpd_data.py` | ROCPD reader and writer implementation, including the rocpd row normalization (`process_rocpd_csv`). |
```
vedithal-amd: **Layering: separate backends, orchestrator, and format interface as three distinct layers**

The proposal collapses three responsibilities into one `interface/` boundary. They are distinct, and the design only works if they are split cleanly. Please define them explicitly:

- **Format interface** (`csv_data.py`, `rocpd_data.py`, future `parquet_data.py`): the only layer that knows on-disk layout and conversion for a given format. Knows a format, knows nothing about how profiling ran.
- **Backend** (`rocprofv3`, `rocprofiler_sdk`): owns execution per profiler. Command construction, env setup, attach/detach, subprocess, pass execution. Knows how to *run* a profiler, knows nothing about storage format.
- **Orchestrator** (what the doc currently calls `profile_data.py`): drives the flow, selects backend and format via `get_reader`/`get_writer`, calls `prepare_pmc_frame`. Format-agnostic and backend-agnostic; it wires the other two together.

Concrete consequences the current draft gets wrong:

- **`profile_data.py` is the orchestrator, not "the interface."** Relabel the boundary in the diagram. Only `csv_data`/`rocpd_data` are true adapters.
- **A single `csv_data.py` (reader + writer) leaks pandas into the profile path.** Profile is stdlib-only today (`utils_profile.py`, `utils_profile_csv.py`, `rocpd_data.py` import no pandas); analyze uses pandas. To avoid leaking the dependency, the CSV format interface should itself be split: a profile-side CSV writer interface that is stdlib-only, separate from an analysis-side CSV reader interface that uses pandas. Same format, two interfaces, partitioned by the dependency constraint of their caller.
- **`finalize_pass(context)` may not cover the post-profiling phase.** Application replay plus the roofline benchmark runs after passes complete. Confirm finalize_pass (a backend concern) can express this, or call it out as a boundary.

**Reminder (out of scope here):** PC-sampling analyze reads JSON produced by profile (`analysis_base.py:122`). That is another format the same three-way split should eventually serve, likely a JSON format-interface (and the on-disk shape may change). Worth a note that it belongs behind this boundary later.

projects/rocprofiler-compute/docs/design/clean_architecture.md
```diff
@@ -0,0 +1,241 @@
...
+The design is working when
+
+- Profile mode does not decide the profile storage layout.
+- Analyze mode can ask for a PMC DataFrame without knowing whether the source is
+  CSV, ROCPD, Profiler Hub, or parquet.
+- `pmc_perf.csv` compatibility is produced through the reader boundary.
+- Storage-format conditionals are isolated to `get_reader` / `get_writer` and the
+  implementations.
+- Adding a new profile source primarily means adding a new implementation and
+  extending selection in one place.
```
vedithal-amd: **Scope, phasing, and honesty about the refactor**

This is a major refactor and the plan should treat it as one:

- **Define explicit, ordered phases.** A single mega-PR is unreviewable. Each phase should be a self-contained logical unit that is independently CI-green and, critically, produces **no analyze output drift** (characterization/golden-output check before and after). State this as an acceptance gate per phase.
- **Be honest that this is a large refactor that must not block in-flight features.** A series of small, fast-to-review PRs is the mechanism that makes that possible; say so.
- **Tighten the document.** Goal, Follow-ups, and Success Criteria repeat the same points; consolidate.

projects/rocprofiler-compute/docs/design/clean_architecture.md
jamessiddeley-amd: 1. I will make the title more clear but to clarify, I didn't mean rocpd as the default output format (which has been implemented), I meant retaining the rocpd .db by default which is an incoming change that we have discussed offline at length

2. _The "today" diagram visually groups it under profile, which is misleading_. This isn't true, the diagram shows pmc perf in the "on disk" group with analyze generating and reading pmc_perf.csv, not profile. On deprecation alignment, that's already called out in Known Gaps: _"ROCPD analysis currently goes through a CSV-shaped intermediate; reading the retained ROCPD source directly is a later improvement... fixed when followed up with the default rocpd PR."_

3. 
This is a fair point, and I'll add clarification in the doc for my justification. By materialize_pmc_perf(...) "only when a compatibility file is needed," I don't mean we stop writing pmc_perf.csv, we still always make it during analyze, so it stays exactly as available as today for users, developers, and tests (nothing in the current behavior or test suite changes). The .csv file still appears on every analyze run.

What changes is just internal as in analysis no longer treats `pmc_perf.csv` as the data path it reads back. Instead it consumes the in memory frame from `read_pmc_frame()`, and `pmc_perf.csv` becomes a one-way export of that frame (written, never read back). 

If you think about it, pmc perf is sort of a bottlneck in analyze. we are spending a lot of process and time generating a csv file where analysis can only consume data via pmc_perf.csv, then any output format reader is forced to also produce a CSV just so analyze can read it. With `read_pmc_frame`, any profile output (rocpd, parquet, whatever) returns a frame directly and never touches CSV. With the current implementation of pmc perf, we are basically quietly making CSV a mandatory output of all formats forever. Today analysis can only consume data via pmc_perf.csv, so every storage format is forced to round-trip through a pmc_perf-shaped csv before analyze can touch it. For example rocpd is .db -> results_*.csv -> pmc_perf.csv -> frame. Why not just skip pmc perf materialization in runtime and go straight to frame?

So we keep the compatibility file by default while removing the read-back bottleneck from analyze.

projects/rocprofiler-compute/docs/design/profile_interface_architecture.md
```diff
@@ -0,0 +1,476 @@
+# Profile Interface Architecture
```
xuchen-amd, line 1: I was thinking about Stas comment on using the term "interface". While "interface" may not be the most accurate term to use for the currently proposed work, but I think this proposal has the potential to grow into a real interface that could be used by users. I see a lot of value in direct-interaction with profile mode output with the power of AI. And this potential interface would promote the interaction in a frictionless way.

cc @svolkov-amd

projects/rocprofiler-compute/docs/design/clean_architecture.md
jamessiddeley-amd: Good points, see updated doc for addressed comments like pandas in the profile path

projects/rocprofiler-compute/docs/design/profile_interface_architecture.md
jamessiddeley-amd: _I see a lot of value in direct-interaction with profile mode output with the power of AI. And this potential interface would promote the interaction in a frictionless way._
Agreed

svolkov-amd (approved):

PR #7425, opened 2026-06-18
fix(rocr): async event loop busy-spin from position-keyed event age

amd-arosa:
fix: async event loop busy-spin from position-keyed event age

## Motivation

using mirage + rocjitsu, noticed millions of ~3.1M WAIT_EVENTS ioctls with one event permanently reporting age=1 while its real age climbed past 14000

the reason why it's requesting so many times is rocr Runtime::AsyncEventsLoop is mismanaging the age, and then resetting to default of 1.

this causes performance issues.

## Technical Details

I noticed that error
```
[rj log CP] IOCTL pid=1 WAIT_EVENTS
[rj log CP] WAIT_EVENTS: pid=1 num=3 timeout=4294967294 wait_all=0 ev[0]=1(age=9567) ev[1]=6(age=10061) ev[2]=11(age=1)
[rj log CP] WAIT_COMPLETE: pid=1 ev=1 age=9567 poll_count=0 is_poll=false
[rj log CP] WAIT_COMPLETE: pid=1 ev=6 age=10061 poll_count=0 is_poll=false
[rj log CP] WAIT_COMPLETE: pid=1 ev=11 age=9538 poll_count=0 is_poll=false
``` 

thousands of times in a row.
 
it instantly completes because 11 was already not 1.

 ```
[rj log CP] CREATE_EVENT: event_id=11 type=0 auto_reset=true gpu_id=38144
 ```

I confirmed that his happened on real hardware using BPF.

slot | event | # distinct baselines sent | range
-- | -- | -- | --
ev[0] | 1 | 14,091 | 1 → 14,097
ev[1] | 6 | 13,885 | 0 → 14,614
ev[2] | 11 | 2 | {0, 1}

 
The spin exists because the baseline is 1 while the real age is 9538, making 9538 != 1 permanently true. Rate mismatch determines how often it spins; it does not create the spin
 
the sim told ROCR the right answer 4,015 times and ROCR sent 1 every time.
 
age_/hsa_events_ are indexed by the compacted unique_evts counter, but PushBack/CopyIndex/PopBack only maintain signal_/cond_/value_/handler_/arg_ 
When the signal list reorders (swap-remove) or grows, the event occupying a compacted slot changes, the guard hsa_events[unique_evts] != hsa_event fires, and the baseline is reset to the literal 1.
The PR keys the age by HsaEvent* so a live event keeps its real baseline across reordering.
 

## JIRA ID

#7425

## Test Plan

1. ci
2. rocjitsu

## Test Result

worked with rocjitsu.
in the ci, Unit_hipMemPrefetchAsync_v2_HostNuma_HostNumaCurrent failed.
```
 The following tests FAILED:
      971 - Unit_hipMemPrefetchAsync_v2_HostNuma_HostNumaCurrent (Failed)

  figure out why it's not working and fix.

  Unit_hipMemPrefetchAsync_v2_HostNuma_HostNumaCurrent
    With Host Numa Current
  -------------------------------------------------------------------------------
  /__w/rocm-systems/rocm-systems/projects/hip-tests/catch/unit/memory/hipMemPrefetchAsync_v2.cc:233
  ...............................................................................
  /__w/rocm-systems/rocm-systems/projects/hip-tests/catch/unit/memory/hipMemPrefetchAsync_v2.cc:265: FAILED:
    REQUIRE( status == cur_node )
  with expansion:
    0 == 1
```
so [pin the thread to the cpu to make Unit_hipMemPrefetchAsync_v2_HostNuma_HostNumaCurrent deterministic](https://github.com/ROCm/rocm-systems/pull/7425/commits/fd8a9dde4fc28a88915bb537ace613f9adefde83)

## Submission Checklist


atgutier (approved):
This looks like a potential power virus issue. The tests are failing but the current semantics for interruptible signals look like they are behaving more like memory signals.

Also may need a tweak.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
```diff
@@ -1761,8 +1766,20 @@ void Runtime::AsyncEventsLoop(void* _eventsInfo) {
...
-        event_age[unique_evts] = runtime_singleton_->KfdVersion().supports_event_age ? 1 : 0;
+      // Restore this event's last-known KFD age, keyed by the event itself
+      // rather than by the compacted slot index. The async list is reordered in
+      // place on handler removal (CopyIndex/PopBack) and grows on registration,
+      // so the event occupying a given slot changes frequently. Re-priming a
+      // live event's baseline to 1 on such a slot change makes KFD report it
+      // satisfied immediately (event_age != 1) and busy-spins this loop.
+      const uint64_t default_age =
+          runtime_singleton_->KfdVersion().supports_event_age ? 1 : 0;
+      if (init_age) {
```
atgutier: I think checking this can cause an unnecessary wake up too. Why do we need to check init age and initialize to 1? Shouldn't we just find the age in the map, if found set it to the age and only fallback to default if not found?

atgutier (changes requested):

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
amd-arosa: removed init_age

atgutier (approved):
LGTM. I'd like @dayatsin-amd to also sign-off before you merge.

dayatsin-amd (approved):

projects/rocr-runtime/runtime/hsa-runtime/core/inc/runtime.h
```diff
@@ -653,6 +653,13 @@ class Runtime {
     std::vector<HsaEvent*> hsa_events_; //!< A list of HSA events for KFD wait
     std::vector<uint64_t> age_;         //!< The age list for KFD wait
     std::vector<void*> arg_;
+    //! Last-known KFD event_age, keyed by the HSA event. The hsa_events_/age_
+    //! arrays above form a compacted, per-iteration view whose slot indices do
+    //! not track a given event across reordering of the async list (handler
+    //! removal swap-removes entries, registration appends them). This map keeps
+    //! each event's baseline so the KFD wait stays edge-triggered instead of
```
cfreeamd, line 663: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** `age_by_event_` entries are removed only when a handler returns `false` (via `event_age_map.erase`). For permanently-registered signals whose handlers always return `true`, entries accumulate. In practice the map size is bounded by the number of currently-registered signals (entries are only written for events currently in the wait list, so the signal must be alive at write time), but a comment noting this invariant would save future readers from re-deriving it.

cfreeamd (commented):
🤖 AI-assisted pre-review (for @cfreeamd) — `runtime.h` and `runtime.cpp` changes look correct. The age-by-event map approach cleanly solves the busy-spin without touching the async list's slot-indexed structure. Inline comment above is optional polish only.

cfreeamd (approved):
lgtm. Consider adding comment mentioned.

mangupta (approved):

amd-arosa:
#8208 broke the build.
the error we see in [this pr](https://github.com/ROCm/rocm-systems/actions/runs/28980175213/job/86157309990?pr=7425) is same [as the failure introduced in 8208](https://github.com/ROCm/rocm-systems/actions/runs/28891156809/job/86173012366)

amd-arosa:
#7803 probably broke `WSL / ROCR Runtime WSL Test`,
```
undefined reference to `rocr::log_flags'
undefined reference to `rocr::log_printf(char const*, int, char const*, ...)'
```
but that's not a required test. so informed author.

PR #7431, opened 2026-06-18
[hipFile] Install relocatable unit-test assets for packaged CI

derobins:
## Motivation

The hipFile tests need to be installable and runnable for TheRock CI.

## Technical Details

[hipFile] Install relocatable unit-test assets for packaged CI

Add cmake/AISInstallTests.cmake and cmake/AISGenerateInstallCTest.cmake,
which install the unit-test binaries (hipfile_tests, and internal_tests when
the AMD detail is built) to share/hipfile/test/ and emit a relocatable
CTestTestfile.cmake that invokes them by relative path under the "unit" label.

gtest_discover_tests() generates ctest fragments that reference the test
executables by absolute build-tree paths, so they cannot be shipped in an
install/artifact tree and run elsewhere. Generating our own ctest file with
relative paths lets the unit suite run from an installed tree via
`ctest -L unit`, which is what TheRock's fetch-artifact CI flow needs.

The CTestTestfile.cmake is generated at install time (not configure time),
because the generator enumerates each binary's GoogleTest cases with
--gtest_list_tests and emits one ctest entry per case. Per-case process
isolation is required: some internal tests leak state across cases and only
pass when each runs in its own process -- exactly the behavior
gtest_discover_tests provides. A single add_test per binary instead produced
68 false failures in internal_tests.

Enumeration runs the build-tree binaries (via $<TARGET_FILE:...>, expanded at
generate time -- install(CODE) supports generator expressions), because under
CPack component packaging the test component is staged in isolation, without
the runtime component's libraries, so a staged binary cannot be launched to
list its cases. The emitted entries still invoke the installed binary by
relative path; only the case-listing launch uses the build-tree path. The
relocatable file is placed with a single DESTDIR-aware file(INSTALL ... RENAME)
so staged packaging/CI installs land it correctly.

Ship the test assets in their own "tests" CPack component, so they install to a
separate hipfile-tests package instead of the runtime/devel packages, matching
the convention of sibling rocm-systems projects (e.g. rccl). AISInstallTests is
included before AISInstall so the component is registered before
rocm_create_package() consumes it.

Force the fetched GoogleTest to build static (BUILD_SHARED_LIBS=OFF in
AISUseGTest.cmake) even though hipFile defaults BUILD_SHARED_LIBS=ON. Otherwise
a CI image without a system GTest fetches and builds it shared, adding
libgtest.so/libgmock.so as runtime dependencies that the test package neither
ships nor can locate -- breaking install-time enumeration and `ctest -L unit`
from the installed tree. With static linkage, libhipfile.so is the only
project-private runtime dependency.

Set INSTALL_RPATH "$ORIGIN;$ORIGIN/../../../lib" with --enable-new-dtags on
the installed binaries so they resolve libhipfile.so from the packaged tree's
lib/ directory at run time. The RUNPATH remains overridable via
LD_LIBRARY_PATH, matching the rest of ROCm.

Scope is unit tests only: the "system" tests require a HIP-capable GPU and
the "stress" tests are gdb-wrapped concurrency testers, both of which are
intentionally excluded from the packaged suite and continue to run from the
build tree / on dedicated runners.

## JIRA ID

ROCM-1261
AIHIPFILE-27

## Test Plan

Manual confirmation that installed tests can be run from ctest

## Test Result

Tests pass

## Submission Checklist


projects/hipfile/cmake/AISInstallTests.cmake
```diff
@@ -0,0 +1,126 @@
...
+#   The CTestTestfile.cmake is generated at INSTALL time (not configure time)
+#   by AISGenerateInstallCTest.cmake, which enumerates each binary's GoogleTest
+#   cases and emits one ctest entry per case. Per-case isolation matters: some
+#   internal tests leak state across cases and only pass when each runs in its
+#   own process -- exactly the behavior gtest_discover_tests provides.
+#
+# Scope: UNIT tests only. The "system" tests require a HIP-capable GPU and the
+# "stress" tests are gdb-wrapped concurrency testers; both are intentionally
+# excluded from the packaged unit suite. They continue to run from the build
+# tree and on dedicated runners.
```
john00003, line 26: Will system tests live in TheRock at any point? I assume that even if so, that would be in a follow-up PR.

derobins: Definitely out of scope for this PR. We're just taking one step at a time. TheRock doesn't run any of our CI yet, so this isn't going to be exercised in the way it needs to yet.

gaoikawa (approved):

riley-dixon (commented):
We should probably add a step to install and run the test binaries in CI (for both unit & system/stress if covered by this PR).

If this intentionally omits system/stress tests, then we probably don't need to upload the test package as an artifact at this time. It's a lot easier for us to debug this here than after it gets to TheRock.

riley-dixon:
Our CTest binaries likes to mock us when we create the test package for some reason. No idea why. But my commit adds re-running the unit tests via the installable package.
Its a bit of a dirty PR considering we run the units tests twice, but eh I think its fine for now. We can clean it up later when we support the system/stress tests in this package.

riley-dixon (approved):

riley-dixon:
AIHIPFILE-227

PR #7434, opened 2026-06-18
[profiler-hub] Examples of writer for Schema V3 with integration test(using Pytest)

anujshuk-amd:
## Motivation

This pull request adds example programs and build support for the `profiler-hub` component, enabling end-to-end testing and demonstration of the public writer  API against the `schema_v3` ROCm profiler data layout. The examples are designed to be built both as part of the main project and standalone, and are integrated with the pytest-driven integration suite.

## Technical Details

1. Added schema_v3 writer examples (one record each):
   - memory_alloc_writer, memory_copy_writer, kernel_dispatch_writer,
     pmc_event_writer, region_writer
   - Each is a writer-only launcher: it takes a database path, registers the
     minimal dependencies, writes one record, flushes to disk, and prints
     db_path=<path>. It does not read back or validate anything.

2. Added pytest integration coverage under tests/integration/
   - conftest.py locates and runs an example binary against a Python-provided
     SQLite path, checks the exit code, and confirms the launcher echoed back the
     same db_path (failing on duplicate output keys).
   - profiler_hub_db.py opens the generated database and reads the schema_v3
     rows into a flat dictionary.
   - Each test_*.py validates the persisted fields against a Python-owned
     EXPECTED dict (memory alloc, memory copy, kernel dispatch, PMC event,
     region).
   - Fails (does not skip) if an expected binary is missing.
   - Per-test timeout via the timeout marker, overridable at runtime with
     PHUB_CI_TIMEOUT.

3. Build/CTest integration
   - Examples are built together with the tests (gated on
     PROFILER_HUB_BUILD_TESTS).
   - generate_ctest.py registers each pytest module with CTest, so the suite
     also runs via ctest.
   - Tests are running in CI/CD

## JIRA ID

JIRA ID : ROCPDSNA-82

## Test Plan

Passed existing tests and newly written tests.

## Test Result

Unit and pytest passed on local system and running in CI.

## Submission Checklist

### DIFF
--- a/profilers/profiler-hub/examples/schema_v3/CMakeLists.txt
+++ b/profilers/profiler-hub/examples/schema_v3/CMakeLists.txt
@@ -1,0 +1,40 @@
+# Copyright (c) Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+set(PROFILER_HUB_SCHEMA_V3_EXAMPLES
+    memory_alloc_writer_reader
+    kernel_dispatch_writer_reader
+)
+
+foreach(_example IN LISTS PROFILER_HUB_SCHEMA_V3_EXAMPLES)
+    set(_target profiler-hub_${_example})
+
+    add_executable(${_target} ${_example}.cpp)
+
+    target_link_libraries(${_target} PRIVATE ${PROFILER_HUB_EXAMPLE_LIBRARY})
+    target_compile_features(${_target} PRIVATE cxx_std_17)
+
+    # Scratch directory for the rocpd database created during the round-trip.
+    target_compile_definitions(
+        ${_target}
+        PRIVATE PHUB_INTEGRATION_TMP_DIR="${CMAKE_CURRENT_BINARY_DIR}"
+    )
+
+    # Keep the executables under the profiler-hub build tree at
+    # <build>/bin/examples.
+    set_target_properties(
+        ${_target}
+        PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${PROFILER_HUB_EXAMPLE_OUTPUT_DIR}"
+    )
+
+    # Locate the shared library relative to the binary
+    set_target_properties(${_target} PROPERTIES BUILD_RPATH "$ORIGIN/../../lib")
+
+    if(PROFILER_HUB_INSTALL_EXAMPLES)
+        install(
+            TARGETS ${_target}
+            DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/profiler-hub/examples
+            COMPONENT profiler-hub-examples
+        )
+    endif()
+endforeach()
--- a/profilers/profiler-hub/examples/schema_v3/CMakeLists.txt
+++ b/profilers/profiler-hub/examples/schema_v3/CMakeLists.txt
@@ -2,10 +2,11 @@
 # SPDX-License-Identifier: MIT
 
 set(PROFILER_HUB_SCHEMA_V3_EXAMPLES
-    memory_alloc_writer_reader
-    kernel_dispatch_writer_reader
-    memory_copy_writer_reader
-    region_writer_reader
+    memory_alloc_writer
+    kernel_dispatch_writer
+    memory_copy_writer
+    pmc_event_writer
+    region_writer
 )
 set(PROFILER_HUB_SCHEMA_V3_EXAMPLE_TARGETS)
 
@@ -17,12 +18,6 @@ foreach(_example IN LISTS PROFILER_HUB_SCHEMA_V3_EXAMPLES)
 
     target_link_libraries(${_target} PRIVATE ${PROFILER_HUB_EXAMPLE_LIBRARY})
     target_compile_features(${_target} PRIVATE cxx_std_17)
-
-    # Scratch directory for the rocpd database created during the round-trip.
-    target_compile_definitions(
-        ${_target}
-        PRIVATE PHUB_INTEGRATION_TMP_DIR="${CMAKE_CURRENT_BINARY_DIR}"
-    )
 
     # Keep the executables under the profiler-hub build tree at
     # <build>/bin/examples.
--- a/profilers/profiler-hub/examples/schema_v3/CMakeLists.txt
+++ b/profilers/profiler-hub/examples/schema_v3/CMakeLists.txt
@@ -4,6 +4,8 @@
 set(PROFILER_HUB_SCHEMA_V3_EXAMPLES
     memory_alloc_writer_reader
     kernel_dispatch_writer_reader
+    memory_copy_writer_reader
+    region_writer_reader
 )
 set(PROFILER_HUB_SCHEMA_V3_EXAMPLE_TARGETS)
 
--- a/profilers/profiler-hub/examples/schema_v3/CMakeLists.txt
+++ b/profilers/profiler-hub/examples/schema_v3/CMakeLists.txt
@@ -5,9 +5,11 @@ set(PROFILER_HUB_SCHEMA_V3_EXAMPLES
     memory_alloc_writer_reader
     kernel_dispatch_writer_reader
 )
+set(PROFILER_HUB_SCHEMA_V3_EXAMPLE_TARGETS)
 
 foreach(_example IN LISTS PROFILER_HUB_SCHEMA_V3_EXAMPLES)
     set(_target profiler-hub_${_example})
+    list(APPEND PROFILER_HUB_SCHEMA_V3_EXAMPLE_TARGETS ${_target})
 
     add_executable(${_target} ${_example}.cpp)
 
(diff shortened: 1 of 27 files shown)


profilers/profiler-hub/tests/integration/conftest.py
```diff
@@ -0,0 +1,85 @@
...
+  3. a fallback repo-wide search for the same binary.
+"""
+
+from __future__ import annotations
+
+import os
+import subprocess
+from pathlib import Path
+
+import pytest
```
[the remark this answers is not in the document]
anujshuk-amd: considered.

profilers/profiler-hub/examples/schema_v3/kernel_dispatch_writer_reader.cpp
```diff
@@ -0,0 +1,372 @@
...
+
+    std::optional<rt::kernel_dispatch_data_t> detail;
+    for(const auto& event : reader.get_events())
+    {
+        if(event.unique_identifier.type != rt::event_type_t::kernel_dispatch) continue;
+        detail = reader.get_kernel_dispatch_details(event);
+        if(detail.has_value()) break;
+    }
+
+    const auto& d = *detail;
```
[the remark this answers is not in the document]
anujshuk-amd: Not needed now. Since not reading in cpp.

profilers/profiler-hub/examples/schema_v3/kernel_dispatch_writer.cpp
```diff
@@ -0,0 +1,372 @@
...
+#include <profiler-hub/writer.hpp>
+#include <profiler-hub/writer_types.hpp>
+
+#include <cstdio>
+#include <exception>
+#include <iostream>
+#include <memory>
+#include <optional>
+#include <string>
+
```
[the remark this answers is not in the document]
anujshuk-amd: NA

profilers/profiler-hub/examples/schema_v3/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+
+    target_link_libraries(${_target} PRIVATE ${PROFILER_HUB_EXAMPLE_LIBRARY})
+    target_compile_features(${_target} PRIVATE cxx_std_17)
+
+    # Scratch directory for the rocpd database created during the round-trip.
+    target_compile_definitions(
+        ${_target}
+        PRIVATE PHUB_INTEGRATION_TMP_DIR="${CMAKE_CURRENT_BINARY_DIR}"
+    )
+
```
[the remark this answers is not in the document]
anujshuk-amd: compile-time path was removed.

profilers/profiler-hub/examples/schema_v3/memory_alloc_writer_reader.cpp
```diff
@@ -0,0 +1,223 @@
...
+    env.node_id    = 1;
+    env.process_id = 1000;
+    return env;
+}
+
+const wt::node_info_t         g_node  = make_node();
+const wt::process_info_t      g_proc  = make_process();
+const wt::event_data_t        g_event = make_event();
+const wt::memory_alloc_data_t g_alloc = make_alloc();
+const wt::trace_environment_t g_env   = make_env();
```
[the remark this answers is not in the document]
anujshuk-amd: NA. Old comment

profilers/profiler-hub/examples/schema_v3/memory_alloc_writer_reader.cpp
```diff
@@ -0,0 +1,223 @@
...
+}
+
+}  // namespace
+
+int
+main()
+{
+    remove_db();
+
+    bool found = false;
```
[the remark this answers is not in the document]
anujshuk-amd: NA

profilers/profiler-hub/tests/integration/profiler_hub_db.py
```diff
@@ -0,0 +1,551 @@
+# Copyright (c) Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""SQLite helpers for validating profiler-hub writer integration fixtures."""
+
+from __future__ import annotations
```
avansick-amd, line 6: This is a lot of framework to confirm a single row lands: ~550 lines of Python re-transcribing the schema, plus a hand-maintained expected-row dict per test. I'd suggest we do something like either add more Python tests to broaden coverage, or use a simple canned SQL query against an expected return value — same field-for-field check, a fraction of the code. Suggestion, not a blocker. (Longer-term, separate idea: a schema version stamp would let tests assert only the fields a given version guarantees — not raised with the team yet.)

profilers/profiler-hub/tests/integration/profiler_hub_db.py
```diff
@@ -0,0 +1,551 @@
...
+        return {
+            "event.stack_id": row["stack_id"],
+            "event.parent_stack_id": row["parent_stack_id"],
+            "event.correlation_id": row["correlation_id"],
+            "event.event_category": row["category"],
+            "event.extdata": row["extdata"],
+        }
+
+    def _read_common_metadata_info(self) -> dict[str, Any]:
+        node = self.row(f"select * from {self.table('rocpd_info_node')} where id = 1")
```
avansick-amd: literal '1'... can we pass this as a parameter from somewhere?

profilers/profiler-hub/tests/integration/profiler_hub_db.py
```diff
@@ -0,0 +1,551 @@
...
+        process = self.row(
+            f"""
+            select P.*, N.id as node_id
+            from {self.table('rocpd_info_process')} P
+            join {self.table('rocpd_info_node')} N on P.nid = N.id
+            where P.pid = ?
+            """,
+            (1000,),
```
avansick-amd: As above...

profilers/profiler-hub/tests/integration/profiler_hub_db.py
```diff
@@ -0,0 +1,551 @@
...
+        )
+        thread = self.row(
+            f"""
+            select T.*, P.pid as process_id
+            from {self.table('rocpd_info_thread')} T
+            join {self.table('rocpd_info_process')} P on T.pid = P.id
+            where T.tid = ?
+            """,
+            (100,),
```
avansick-amd: As above...

profilers/profiler-hub/tests/integration/test_kernel_dispatch.py
```diff
@@ -0,0 +1,113 @@
...
+
+from __future__ import annotations
+
+import pytest
+
+from profiler_hub_db import ProfilerHubDb
+
+pytestmark = pytest.mark.timeout(120)
+
+EXPECTED = {
```
avansick-amd, line 19: Can we get the expected values out of a text file or similar?  When the db is created, where do we get the values from on that end? can we share that here?

profilers/profiler-hub/source/CMakeLists.txt
```diff
@@ -18,7 +18,7 @@ target_sources(profiler-hub-objects PRIVATE ${profiler-hub_sources})
 # ------------------------------------------------------------ #
 target_include_directories(
     profiler-hub-objects
-    PRIVATE ${CMAKE_CURRENT_LIST_DIR}/common
+    PRIVATE ${CMAKE_CURRENT_LIST_DIR} ${CMAKE_CURRENT_LIST_DIR}/common
```
avansick-amd: Is this right?

avansick-amd (approved):
The python framework seems like a lot if we're sticking to happy path single row db validation.  We can look at expanding the test suite over time though.

profilers/profiler-hub/tests/integration/profiler_hub_db.py
anujshuk-amd: Adressed.

profilers/profiler-hub/tests/integration/profiler_hub_db.py
anujshuk-amd: Considered

profilers/profiler-hub/tests/integration/profiler_hub_db.py
anujshuk-amd: Considered

profilers/profiler-hub/tests/integration/profiler_hub_db.py
anujshuk-amd: Noted for future.

profilers/profiler-hub/tests/integration/test_kernel_dispatch.py
anujshuk-amd: Added comment in the file.

profilers/profiler-hub/source/CMakeLists.txt
anujshuk-amd: corrected

anujshuk-amd:
> The python framework seems like a lot if we're sticking to happy path single row db validation. We can look at expanding the test suite over time though.

That’s fair. The intent here is to keep the C++ examples writer-only and move validation into Python so we can inspect the persisted DB directly.I’m fine keeping it scoped and lightweight for now. But yes definetly we can expand.

PR #7436, opened 2026-06-18
Multi-arch CI : Limit failing occupancy tests to arch it supports.

jaydeeppatel1111:
## Motivation

This tests are develop to run on gfx1250. On multi-arch CI, TU gets compiled if gfx1250 is included in arch list however runtime needs to limit to arch with cluster support.

Multiarch CI https://github.com/ROCm/TheRock/actions/runs/27729666118/job/82064867736?pr=5946 has failures due to this.

## Technical Details

This tests are develop to run on gfx1250. On multi-arch CI, TU gets compiled if gfx1250 is included in arch list however runtime needs to limit to arch with cluster support.

## JIRA ID

Multi arch CI issue.

## Test Plan

Ran on gfx942 with multiarchs

## Test Result

Skipped on gfx942 with multiarchs bins

## Submission Checklist

### DIFF
--- a/projects/clr/rocclr/device/rocm/rocdevice.cpp
+++ b/projects/clr/rocclr/device/rocm/rocdevice.cpp
@@ -1755,6 +1755,13 @@ bool Device::populateOCLDeviceConstants() {
   if (HSA_STATUS_SUCCESS != hsaStatus && HSA_STATUS_ERROR_INVALID_ARGUMENT != hsaStatus)
     LogError("HSA_AMD_AGENT_INFO_CLUSTER_MAX_SIZE query failed");
 
+  // A cluster of size 1 is a regular single-block launch, not multi-block cluster support.
+  // Devices without multi-block cluster support (e.g. pre-gfx12.5) report a max size of 1, so
+  // normalize it to 0 to mean "no cluster support" consistently across all consumers.
+  if (info_.clusterMaxSize_ <= 1) {
+    info_.clusterMaxSize_ = 0;
+  }
+
   info_.gpuDirectRdmaWithHipVmmSupported_ =
       info_.virtualMemoryManagement_ && info_.dmabufSupported_;
 
--- a/projects/clr/rocclr/device/rocm/rocdevice.cpp
+++ b/projects/clr/rocclr/device/rocm/rocdevice.cpp
@@ -1755,13 +1755,6 @@ bool Device::populateOCLDeviceConstants() {
   if (HSA_STATUS_SUCCESS != hsaStatus && HSA_STATUS_ERROR_INVALID_ARGUMENT != hsaStatus)
     LogError("HSA_AMD_AGENT_INFO_CLUSTER_MAX_SIZE query failed");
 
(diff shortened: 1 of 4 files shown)


projects/clr/rocclr/device/rocm/rocdevice.cpp
```diff
@@ -1755,6 +1755,13 @@ bool Device::populateOCLDeviceConstants() {
   if (HSA_STATUS_SUCCESS != hsaStatus && HSA_STATUS_ERROR_INVALID_ARGUMENT != hsaStatus)
     LogError("HSA_AMD_AGENT_INFO_CLUSTER_MAX_SIZE query failed");
 
+  // A cluster of size 1 is a regular single-block launch, not multi-block cluster support.
+  // Devices without multi-block cluster support (e.g. pre-gfx12.5) report a max size of 1, so
+  // normalize it to 0 to mean "no cluster support" consistently across all consumers.
+  if (info_.clusterMaxSize_ <= 1) {
```
g-h-c: I don't think we should do this, because it would gate cluster launches with size with dimensions (1,1,1) which should be legal on all GPUs, even on the ones don't have multi-block cluster support.

I think to unblock this we should simply set deviceProps.clusterLaunch to 0 in all devices except gfx1250 where it would be 1.

jaydeeppatel1111: Hi @g-h-c, This is the same "> **1**" approach we discussed on over Teams initially. It gives exactly the behavior we agreed before - clusterLaunch = 0 on all devices except the gfx1250-class, where it's 1.

ROCR sets cluster_max_dim_ per-ASIC: gfx12.5+ (gfx1250/1251) report num_cu_per_se (> 1), every other ASIC keeps the default {1,1,1} → 1. So > 1 is true only on the multi-block-capable parts and 0 elsewhere, and it leaves clusterMaxSize_ untouched so trivial {1,1,1} clusters stay legal everywhere.

I have updated PR with above change, Can you please review?

g-h-c: Yes, that looks correct. 

Looking at ROCr code `cluster_max_dim_` stays as {1, 1, 1} and only if `workgroup_clusters_supported_` is true then it will be set with a different value.

g-h-c (approved):

mangupta (approved):

rahulc-gh:
Linux Failure test - is disabled as part of - https://github.com/ROCm/rocm-systems/pull/7454
Windows runners there is a queue and @agunashe  helped to verify locally observed below 3 failures and confirmed they are un-related to this change 

2209 - Unit_hipModuleLoadData_Positive_Basic (Failed)
        2239 - Unit_hipModuleLoadFatBinary_PosiiveTsts (Failed)
        2241 - Unit_hipExtModuleLaunchKernel_UniformWorkGroup (Failed)
        
 @kjayapra-amd  also confirmed windows tests can be override , so good to override and merge

PR #6909, opened 2026-06-08
[rocprofiler-systems] replace phantom ROCPROFSYS_USE_TRACE with ROCPROFSYS_TRACE

marantic-amd:
## Motivation

ROCPROFSYS_USE_TRACE was never a registered setting, so the env var and both config.cpp call sites had no effect and the perfetto category was silently disabled at runtime. Point them at the canonical ROCPROFSYS_TRACE and correct the docs.

## Technical Details

Documentation and code is updated so that it now references the correct variable, which is ROCPROFSYS_TRACE, for handling Perfetto output format.

## JIRA ID

AIPROFSYST-487

## Test Plan

- Run existing tests
- Add test that will make sure that the undocumented variable is not referenced anywhere

## Test Result

- Clean build pass without any regression
- All tests pass

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #7441, opened 2026-06-18
[DO NOT MERGE] bisect: LKG + 6909 (isolate which commit fails the patch)

jungx098:
## DO NOT MERGE — per-commit bisect of the #7420→#7421 window

Base = `exp-lkg-2a108a17` (LKG = #6632 extsem patch + develop pre-#5219, PASSES). Head = LKG + **one** cherry-picked develop commit: #6909 (env-var rename) — expect PASS (benign).

Diff is exactly that one commit (parent-as-base trick → mergeable, CI runs, merge ref = patch + that commit). If only the #5219 branch fails and the other three pass, #5219 is the isolated culprit with the patch present. Bimodal — needs repeats. Delete after.


sputhala-amd:

PR #7352, opened 2026-06-16
feature(mirage): v0.1.0 release

amd-arosa:
## Motivation

prepare to release mirage

### removed
- made hotswap no longer by default
- made webui no longer default

### feature
- protype rocjitsu-dbt support
- made it clear mirage is designed for rocjitsu
- "hack" system for special compatibility for workloads

### docs
- added demos
- added tutorials

### refactored
- use inventory system to make separation of rocjitsu

### changed
- consistent flags / layering

### test
- extensive bug bash
- aorta
- test matrix to test all configurations

### build
- add Dockerfile that builds rocjitsu + mirage using therock build image 

### fixed
- rccl support
- torch ddp support
- torchrun support
- full docker support
- better mounts
- correct shutdown
- and way more stuff i can't remeber

## Test Plan

- bug bash
- test matrix

## Test Result

- test matrix works
- bug bash works

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #7175, opened 2026-06-12
[rocjitsu] Add CI for rocjitsu-test-corpus.

amd-eochoalo:
## Motivation

Test rocjitsu with rocjitsu-test-corpus.

## Technical Details

* Not using submodules to avoid need to constantly update references

## Test Plan

* Tests execute on CI

## Test Result

Tests run! More tests timeout on CI than on my local machine. We can either let these tests fails now or just skip them. Other failing tests were skipped. One which hung the CI.

## Submission Checklist

- [X] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #7447, opened 2026-06-18
[rocjitsu] Use mirage instead of rocjitsu for testing corpus

amd-eochoalo:
## Motivation

Switch to mirage instead of rocjitsu. Do not merge yet. This PR depends on https://github.com/ROCm/rocm-systems/pull/7175 and https://github.com/ROCm/rocm-systems/pull/7352

## Technical Details

Build mirage

## JIRA ID

None applicable

## Test Plan

Test in CI

## Test Result

CI should pass

## Submission Checklist

### DIFF
--- a/.github/scripts/therock_configure_ci.py
+++ b/.github/scripts/therock_configure_ci.py
@@ -139,6 +139,7 @@ SKIPPABLE_PATH_PATTERNS = [
     ".github/label*.yml",
     ".github/workflows/labeler.yml",
     ".github/workflows/amdsmi-manylinux-build.yml",
+    ".github/workflows/rocjitsu-corpus-tests.yml",
 ]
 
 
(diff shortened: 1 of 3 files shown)


Groverkss:
Do we need to do it this way... Why are we adding more layers on top of rocjitsu to do rocjitsu tests?

amd-eochoalo:
@Groverkss I think both can work, but I see your point. I remember @amd-arosa mentioning something about this being preferred, but I don't recall the exact argument. @amd-arosa do you remember why mirage was preferred for these tests as opposed to rocjitsu?

atgutier:
> @Groverkss I think both can work, but I see your point. I remember @amd-arosa mentioning something about this being preferred, but I don't recall the exact argument. @amd-arosa do you remember why mirage was preferred for these tests as opposed to rocjitsu?

We don't want to test rocjitsu with mirage. I believe Ada has some proposal to replace the current rocjitsu CLI with one in Rust, and I am fine with that, but until that happens and it is in our directory, Mirage is a separate project.

amd-eochoalo:
Ok, I'll close this ticket then. Thanks for the clarification.

PR #7448, opened 2026-06-18
[rocprofiler-compute] Add missing CLI option unit tests

abchoudh-amd:
## Motivation

Add missing unit tests for the rocprof-compute CLI options.

## Technical Details

- Add tests/test_argparser.py covering the missing general CLI options.
- Parametrize options that have both short and long forms over both flags

## JIRA ID

AIROCVAL-128

## Test Plan

- python3 -m pytest tests/test_argparser.py

## Test Result

- [X] Tests passed.

## Submission Checklist


projects/rocprofiler-compute/tests/test_argparser.py
jamessiddeley-amd, line 1: Test file is not registered in `tests/test_categories.yaml`. Recommend since `test_argparser` isn't listed, it won't get a quick/standard/pr label and won't run in CI jobs. Since this test file is super fast since were not executing rocprof compute for it, i recommend adding it to the quick list.

jamessiddeley-amd, line 1: Also should add to `ALL_TEST_NAMES` for the coverage report so we can cover some argparser.py test coverage in cdash. The `generate_coverage_report` target depends on `ALL_TEST_NAMES` (check cmakelists.txt) 
Actually some other tests are also missing from this list, adding it keeps coverage accurate. Not a blocker though.

```
CMakeLists.txt
    set(ALL_TEST_NAMES
        test_profile_kernel_execution
        ...
        test_parser
```

jamessiddeley-amd, line 1: Also im a little confused on the purpose of this PR in the sense that if its for coverage vs general CLI options, the general options group also includes `--list-metrics`, `--list-blocks`, `-s/--specs`, and `--experimental`, etc. which aren't covered in this PR. Why have only these specific arguments youve added?  I feel like the PR title here implies full general-option coverage so just wondering why that is.

jamessiddeley-amd (commented):
Ran the test suite here in my own container and verified all the current tests pass 👍

projects/rocprofiler-compute/tests/test_argparser.py
abchoudh-amd: Thanks for mentioning it. Other options behaviors are already covered. I have updated the title and description with missing unit tests.

jamessiddeley-amd: Fair, ok sounds good

jamessiddeley-amd (approved):
My comments were addressed, LGTM otherwise pending CI jobs

projects/rocprofiler-compute/tests/test_argparser.py
```diff
@@ -0,0 +1,115 @@
...
+import pytest
+from common import SUPPORTED_ARCHS
+
+from argparser import omniarg_parser
+
+HOME = Path.cwd()
+VERSION = {"ver_pretty": "rocprof-compute (unit test)"}
+
+
+def build_args(argv, experimental=False):
```
vedithal-amd, line 20: Nit: the `experimental` param is unused. Every call site uses the default and this file only tests general config options, so drop it and pass `False` directly to `omniarg_parser`.

vedithal-amd (approved):
LGTM. Tests pass (16/16, 0.07s), ruff clean, no redundancy with existing argparser tests in test_utils.py, and registration in CMakeLists/test_categories.yaml is consistent. One non-blocking nit inline re: unused `experimental` param.

projects/rocprofiler-compute/tests/test_argparser.py
vedithal-amd: On reflection, keeping the `experimental` param is fine: it leaves `build_args` ready to cover experimental-gated options later without reworking the helper. Resolving.

PR #7455, opened 2026-06-18
[rocprofv3] Document GPU performance level requirement for Navi3x/Nav…

JasonTTang:
…i4x profiling

Add a new section explaining that on RDNA3 (Navi3x) and RDNA4 (Navi4x) GPUs, AUTO performance mode gates off the perfmon clock in hardware (RLC_CGTT_MGCG_OVERRIDE.PERFMON_CLOCK_STATE = 0), which disables GPU profiling entirely. Setting performance level to stable_std (or similar non-AUTO modes) turns the perfmon clock back on, enabling PMC profiling.

This is a hardware feature enablement requirement, not a clock-stability or result-reproducibility concern. The section documents two ways to set the performance level: via the power_dpm_force_performance_level sysfs entry and via amd-smi.


bing-ma (approved):
Looks good to me.

PR #7461, opened 2026-06-18
[rocprofiler-compute] Fix stale pmc perf bypassing kernel filtering

jamessiddeley-amd:
## Motivation

Roofline charts silently ignored kernel filtering after a re-profile into an already analyzed workload dir. 
For example, `profile` -> `analyze` -> `profile --roof-only -k <kernel>` -> `analyze` produced a roofline containing *all* kernels instead of the selected one with no error or warning. 
This made `--roof-only -k` effectively unusable in the normal workflow where the full run is analyzed first

## Technical Details

Root cause is that `join_workload_csvs()` in `src/rocprof_compute_analyze/analysis_base.py` reused any existing `pmc_perf.csv` unconditionally. A reprofile rewrites the per-pass `results_*.csv` just fine, but the stale `pmc_perf.csv` (left behind by the earlier `analyze`) was reused so the kernel-top table kept every kernel and `calc_ai_analyze()` fell back to plotting all of them unfiltered.

Fix is to only reuse `pmc_perf.csv` when it is at least as new as every `results_*.csv`, otherwise rejoin them. I added a small helper for this, everything in the rest of the analyze pipeline is preserved

## JIRA ID

AIPROFCOMP-253

## Test Plan

- Added unit test in `tests/test_analyze_commands.py` covering the following cases: 
   - missing `pmc_perf.csv`, 
   - no `results_*.csv`, 
   - `pmc_perf.csv` newer or same "age" than `results_*.csv` (basically just regular profile into analyze timeline), and 
   - `results_*.csv` newer than `pmc_perf.csv` (stale pmc perf).
- Ran the full `tests/test_analyze_commands.py` suite and `ruff check`
- Tested the origional repro on MI350 (`tests/occupancy`, filtering on `vgprbound`):
  - `profile → analyze → profile --roof-only -k vgprbound → analyze`
  - `profile --roof-only → analyze → profile --roof-only -k vgprbound → analyze`
  - Also ran regular re-analyze with no new profiling.

## Test Result

- All tests in `test_analyze_commands.py` pass (including the new test) pass, ruff clean.
- Both previously failing scenarios now rejoin and report 1 kernel (`vgprbound`) instead of 4 which is expected
- Control case still reports `Using existing pmc_perf.csv` with no needless rejoin

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_base.py
+++ b/projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_base.py
@@ -73,6 +73,16 @@ def test_df_column_equality(df: pd.DataFrame) -> bool:
     return df.eq(df.iloc[:, 0], axis=0).all(1).all()
 
 
+def _pmc_perf_is_current(pmc_perf: Path, results_files: list[Path]) -> bool:
+    """Return True if pmc_perf.csv can be reused (no newer results_*.csv)."""
+    if not pmc_perf.exists():
+        return False
+    if not results_files:
+        return True
+    newest_results_mtime = max(file.stat().st_mtime for file in results_files)
+    return pmc_perf.stat().st_mtime >= newest_results_mtime
+
+
 class OmniAnalyze_Base:
     def __init__(
         self, args: argparse.Namespace, supported_archs: dict[str, str]
@@ -700,7 +710,7 @@ class OmniAnalyze_Base:
             pmc_perf = directory / "pmc_perf.csv"
             results_files = list(directory.glob("results_*.csv"))
 
-            if pmc_perf.exists():
+            if _pmc_perf_is_current(pmc_perf, results_files):
                 console_debug(f"Using existing {pmc_perf}")
             elif results_files:
                 console_log(f"Joining results_*.csv for {directory}...")
--- a/projects/rocprofiler-compute/tests/test_analyze_commands.py
+++ b/projects/rocprofiler-compute/tests/test_analyze_commands.py
@@ -2185,3 +2185,30 @@ def test_join_prof_renames_sq_accum_prev_hires_to_bucket_target(tmp_path):
     assert "SQ_ACCUM_PREV_HIRES" not in merged.columns
     assert set(merged["SQ_LEVEL_WAVES_ACCUM"].tolist()) == {100, 200}
     assert "SQ_WAVES" in merged.columns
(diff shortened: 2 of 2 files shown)


vedithal-amd:
Thanks for tracking down the root cause, the analysis is solid.

Before we merge, I want to step back on the approach: I think this fixes a symptom of a larger problem rather than the underlying issue.

**The problem**

Re-profiling into a pre-existing workload folder is unsafe in more ways than the stale `pmc_perf.csv` this PR handles:
- It reuses the same perfmon files as the original run. If the counters to collect have changed, those files are now wrong and need regenerating.
- Stale CSVs and torch trace files linger. If the new profile doesn't overwrite every file, analysis silently mixes old and new data.

So `pmc_perf.csv` is just one of several stale artifacts. Fixing it alone closes one hole and leaves the rest open.

**The motivation**

The safe, predictable way to get a clean `--roof-only -k <id>` run today is simply: delete the workload folder, then profile again. A clean folder sidesteps perfmon regen, stale CSVs, and stale traces all at once.

**Suggested solution**

A broader-scoped change that clears the existing workload folder's contents and then re-profiles, handling perfmon regen, stale CSVs, and traces in one consistent place.

**This must always prompt for explicit user confirmation before deleting anything.** We are wiping the contents of a user's workload folder, so silently deleting their data is not acceptable. The user must opt in every time.

I'd lean toward closing this in favor of that approach, so we don't ship a partial fix that masks the remaining cases. Happy to discuss.

vedithal-amd (changes requested):
per previous comment

jamessiddeley-amd:
> per previous comment

This is a fair and a good time to propose the deletion and reprofiling idea properly, you bring up good points
 
But the confirmation step would break most of any CI jobs or any automated rocprof compute runs, we need to somehow handle non interactive confirmations as well as interactive ones
For example `test_roofline_kernel_filter` in `test_profile_general` re profiles `--roof-only [--kernel ...]` into the same populated dir 4 times, with the comment explicitly stating not to clean output dir and use the same workload. Other CIs run as well
 
A confirmation step would time out all CI and a hard stop would fail it so the interactive confirmation has to be optional in some capacity
 
An option is that we can add `--overwrite` to argparser for profile mode to overwrite automatically, and prompt if omitted?

jamessiddeley-amd:
Closed in favour of #7615

PR #7466, opened 2026-06-18
[rocjitsu] Add gfx1250 cluster load multicast

kuhar:
## Motivation

Add gfx1250 cluster load multicast support, including cluster-aware dispatch placement, AMD extended AQL packet parsing, cluster LDS multicast writeback, generated decode plumbing, and focused tests.
  
## Technical Details

The implementation models workgroup clusters at dispatch time because cluster load multicast depends on knowing the complete peer set, each peer workgroup’s CU, and each peer LDS base. Cluster shapes are validated up front, and the current implementation caps clusters at 16 workgroups to match the hardware multicast mask width.

Cluster dispatch reuses the normal wavefront initialization path so regular and clustered launches keep the same SGPR setup, EXEC mask initialization, plugin callbacks, and completion accounting. The clustered path only adds placement metadata and LDS lifetime management.

LDS multicast is routed through a `ClusterLdsMulticastTransaction` and engine interface instead of directly embedding peer writes in the memory pipeline. The default engine completes synchronously, but the transaction boundary preserves enough information for a future timing/fabric model to defer completion and hold ASYNCCNT until all target LDS writes are visible.

Peer LDS addresses are remapped from the source workgroup’s LDS window into each target workgroup’s actual LDS allocation. This avoids assuming identical LDS bases across CUs and keeps multicast semantics tied to workgroup-local LDS addresses.

## JIRA ID

N/A

## Test Plan

Unit tests + local CTS-style tests.
Regenerated all isas just in case.

## Test Result

All good.

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
+++ b/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
@@ -3972,13 +3972,23 @@ class CodeGenerator:
         L.append('  auto &cu = wf.cu();')
         L.append('  const auto &lds = cu.lds();')
         L.append('  uint64_t exec = wf.exec();')
+        L.append(
+            '  // VGLOBAL async-from-LDS applies ioffset to both destination and LDS source.'
+        )
+        L.append(
+            '  int64_t lds_offset = static_cast<int64_t>(static_cast<int32_t>(inst_.ioffset << 8) >> 8);'
+        )
         L.append(f"  uint32_t lds_addr_base = {self._vgpr_base_expr('vsrc')};")
         L.append(f'  d->store_data.resize(wf.wf_size() * {stride});')
         L.append('  for (uint32_t lane = 0; lane < wf.wf_size(); ++lane) {')
         L.append('    if (!(exec & (1ULL << lane))) continue;')
         L.append(
-            '    uint32_t lds_addr = wf.lds_base() + cu.read_vgpr(lds_addr_base, lane);'
+            '    uint32_t lds_addr = static_cast<uint32_t>(static_cast<int64_t>(wf.lds_base()) +'
         )
+        L.append(
+            '                                           static_cast<int64_t>(cu.read_vgpr(lds_addr_base, lane)) +'
+        )
+        L.append('                                           lds_offset);')
         L.append(f'    lds.read(lds_addr, &d->store_data[lane * {stride}], {stride});')
         L.append('  }')
         L.append('  set_data(std::move(d));')
--- a/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
+++ b/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
@@ -3852,6 +3852,8 @@ class CodeGenerator:
             L.append(f'  d->transpose = {sem.transpose_kind};')
         L.append(f'  d->mtype = {self._mtype_expr()};')
         L.append(f'  d->non_temporal = {nt};')
+        if sem.name.startswith('CLUSTER_LOAD_'):
+            L.append('  d->request_force_l1_bypass = true;')
         L.append('  flat_calculate_addresses(inst_, wf, *d);')
         L.append('  set_data(std::move(d));')
         return '\n'.join(L)
@@ -3927,27 +3929,20 @@ class CodeGenerator:
             L.append(
                 '  d->cluster_mcast_mask = wf.m0() & amdgpu::kClusterMulticastMask;'
             )
+            L.append('  d->request_force_l1_bypass = true;')
         L.append(f'  d->mtype = {self._mtype_expr()};')
         L.append(f'  d->non_temporal = {nt};')
         L.append('  flat_calculate_addresses(inst_, wf, *d);')
         L.append('  auto &cu = wf.cu();')
         L.append('  uint64_t exec = wf.exec();')
         L.append(
-            '  // VGLOBAL async-to-LDS applies ioffset to both source and LDS destination.'
-        )
-        L.append(
-            '  int64_t lds_offset = static_cast<int64_t>(static_cast<int32_t>(inst_.ioffset << 8) >> 8);'
+            '  // flat_calculate_addresses applies ioffset to the global side; the LDS operand is independent.'
         )
         L.append(f"  uint32_t lds_addr_base = {self._vgpr_base_expr('vdst')};")
         L.append('  for (uint32_t lane = 0; lane < wf.wf_size(); ++lane) {')
         L.append('    if (!(exec & (1ULL << lane))) continue;')
         L.append('    uint32_t lane_lds_addr = cu.read_vgpr(lds_addr_base, lane);')
-        L.append(
-            '    d->per_lane_lds_addr[lane] = static_cast<uint32_t>(static_cast<int64_t>(wf.lds_base()) +'
-        )
-        L.append(
-            '                                            static_cast<int64_t>(lane_lds_addr) + lds_offset);'
-        )
+        L.append('    d->per_lane_lds_addr[lane] = wf.lds_base() + lane_lds_addr;')
         L.append('  }')
         L.append('  set_data(std::move(d));')
         return '\n'.join(L)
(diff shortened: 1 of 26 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/amd_ext_aql_packet.h
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc.
```
atgutier, line 1: We have a copy of the HSA headers in lib/rocjitsu/external_headers that should have these definitions. We may need to refresh our copy from rocr-runtime.

Eventually, it'd be great if those headers moved out of the rocr-runtime project into a shared location in rocm-systems, but my attempt to put them into shared was nacked so we need to copy from rocr-runtime for now.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
atgutier: I think this may need updating as well. There should be an instruction offset applied that we may just not have stressed in tests yet.

Also, should we moving address calculating out of here and into our addr calc files?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
```diff
@@ -523,8 +689,65 @@ uint32_t CommandProcessor::dispatch_workgroups(DispatchEntry &entry) {
   // LDS is per-CU and s_barrier synchronises within a CU). Query each CU for
   // capacity before dispatching to guarantee all-or-nothing placement.
   uint32_t dispatched = 0;
+  auto dispatch_to_cu = [&](uint32_t local_wg_id, uint32_t global_wg_id, ComputeUnitCore *cu) {
```
atgutier, line 699: CP is getting complex. Now might be a good time to modularize this a bit and move dispatch-related things to a dispatch controller block (that should model the ACEs [1])

[1] https://rocm.docs.amd.com/projects/HIP/en/latest/understand/hardware_implementation.html

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/memory_pipeline.h
```diff
@@ -69,9 +76,13 @@ class MemoryPipeline {
     }
     wf.wait_counters().increment(issue_counter);
     initiate_access(*inst, wf);
-    complete_access(*inst, wf);
-    wf.release_wait_counter(issue_counter);
-    delete inst;
+    // The wait counter pins wf/inst ownership until this callback releases it.
+    MemoryAccessDeferredCompletion deferred_completion = [this, inst, &wf, issue_counter]() {
```
atgutier, line 82: This is an improvement over the pure synchronous path for now. When we add timing and have async execution we'll need to separate this out a bit such that initiate access gets the operands, does address calculation, and then dispatches to the request queues in the memory system.

Complete access should handle the responses from the response queue, do the writebacks on load, and update wait counters.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
```diff
@@ -33,42 +34,112 @@ namespace amdgpu {
...
+
 uint32_t read_memory_u32(GpuMemory *memory, uint64_t addr) {
   uint32_t value = 0;
   for (uint32_t i = 0; i < sizeof(value); ++i)
     value |= static_cast<uint32_t>(memory->read8(addr + i)) << (i * 8);
   return value;
 }
 
+uint32_t aligned_lds_bytes_per_workgroup(const DispatchEntry &entry) {
+  return util::align_up(entry.group_segment_fixed_size, 256u);
```
atgutier, line 96: Is this consistent across all ISAs?

atgutier (commented):
Some initial comments. Still need to do a deeper dive

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -3506,6 +3506,10 @@ def _gen_global_load_async_to_lds(
         self._append_wait_counter_type(L, 'global_load_async_to_lds')
         L.append('  d->lds_dst = true;')
         L.append('  d->lds_per_lane_addr = true;')
+        L.append('  d->lds_base = wf.lds_base();')
+        if sem.name.startswith('CLUSTER_LOAD_ASYNC_TO_LDS_'):
+            L.append('  d->cluster_multicast = true;')
+            L.append('  d->cluster_mcast_mask = wf.m0();')
```
atgutier: I believe this should only take bits 15:0 so we need & 0xffff;.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
kuhar: I documented this helper as mirroring `ComputeUnitCore::allocate_lds()`/`can_accept_workgroup()` 256-byte granularity, so clustered placement uses the same per-workgroup LDS accounting as ordinary dispatch across ISAs.

atgutier (commented):
Comments from the expert reviewer agent:

MAJOR (fix before merge)

  1. Per-load payload heap copy regression — memory_pipeline.cpp:84 / cluster_lds_multicast.cpp:48-50
  All LDS-destination loads (ordinary global/buffer→LDS, not just cluster) now route through
  make_cluster_lds_multicast_transaction, which copies the whole response_data vector + the 64-lane
  LDS-addr array and allocates a std::vector<ClusterLdsTarget> per load. The pre-PR path wrote bytes
  straight to LDS with zero allocation. The abstraction's cost lands on every LDS load, not just
  cluster ops.
  Fix: keep a fast path that writes directly to cu.lds() when !d.cluster_multicast; only build the
  transaction for the multicast case (or take payload by move / non-owning span, since response_data
  is dead after writeback).

  2. Overloaded lds_base semantics with an unenforced invariant — mem_state.h:106-117 /
  cluster_lds_multicast.cpp:27-32
  Cluster remap computes source_lds_addr - source_lds_base; if lds_base ever carries anything but the
  pure WG allocation base for a cluster load, remap_cluster_lds_addr throws and aborts the sim.
  Correctness hinges on codegen always setting d->lds_base = wf.lds_base() — true today
  (vglobal.cpp:3334) but nothing enforces it.
  Fix: assert(state.lds_base == wf.lds_base()) in make_cluster_lds_multicast_transaction when
  cluster_multicast, and tighten the doc to state the invariant as a hard requirement.

  ---
  MINOR

  - cluster_lds_multicast.cpp:64-72 — OOB peer LDS write is silently dropped (Lds::write8 ignores addr
   >= size), masking peer-window overflow bugs. Bounds-check + throw/log, like the payload check above
   it.
  - completion enums — ClusterLdsMulticastResult / MemoryAccessCompletion lack [[nodiscard]]; a
  dropped deferred result leaks a transaction (wait counter never released → wavefront never retires).
   Mark them [[nodiscard]].
  - command_processor.cpp:663 — dead mcast_mask == 0 branch (caller already guards). Drop or document
  as defensive.

  NIT

  - race_detector/plugin.cpp:20 — file-local helper has external linkage; move to anonymous namespace
  (project style).
  - command_processor.h:291 vs compute_unit.cpp — wg_key duplicated; factor into one shared inline
  constexpr (no copy-paste rule).
  - dispatch_entry.h:151 — cluster_peer_local_wg_id throws std::logic_error for a condition
  validate_cluster_shape already rejects; use assert per convention.
  - _generator.py:3954 — async-from-LDS stores don't apply ioffset to the LDS address, asymmetric with
   the new load path. Low confidence / out of this PR's load-only scope — flagging so it's a
  deliberate decision, not an oversight.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -3921,16 +3921,32 @@ def _gen_global_load_async_to_lds(
...
         L.append(f"  uint32_t lds_addr_base = {self._vgpr_base_expr('vdst')};")
         L.append('  for (uint32_t lane = 0; lane < wf.wf_size(); ++lane) {')
         L.append('    if (!(exec & (1ULL << lane))) continue;')
+        L.append('    uint32_t lane_lds_addr = cu.read_vgpr(lds_addr_base, lane);')
+        L.append(
+            '    d->per_lane_lds_addr[lane] = static_cast<uint32_t>(static_cast<int64_t>(wf.lds_base()) +'
+        )
         L.append(
-            '    d->per_lane_lds_addr[lane] = wf.lds_base() + cu.read_vgpr(lds_addr_base, lane);'
+            '                                            static_cast<int64_t>(lane_lds_addr) + lds_offset);'
```
atgutier: FYI, we will also need the lds_offset fix for stores too.

atgutier (approved):

PR #7467, opened 2026-06-18
[rocjitsu] Scale SMEM SBASE operand by 2

accauble:
## Motivation

When decoding smem instructions, the sbase operand was the raw value instead of the correct scaled value. This meant that the use/defs would be incorrect for those instructions.

## Technical Details

This commit fixes the codegen to set the sbase operand to the register that is actually used (see `_operand_encodeing_value_expr`).

## Test Plan

Added some unit tests for code gen (`amdisa/tests/test_smem_sbase_operand.py`) and reproducer tests (`tests/smem_sbase_operand_test.cpp`). These need to pass (as well as other repo tests).

## Test Result

Tests pass!

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
+++ b/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
@@ -4977,6 +4977,25 @@ class CodeGenerator:
             self.isa_spec.profile.vop3p_opsel_fields,
         )
 
+    @staticmethod
+    def _operand_encoding_value_expr(
+        opnd_name: str, is_smem: bool, packed_16bit: bool
+    ) -> str:
+        """C++ expression for an operand's raw encoding value.
+
+        SMEM SBASE is encoded in units of 2 SGPRs (raw N names s[2N:2N+1]), so it
+        is scaled here to a real SGPR index. This keeps the operand's
+        register-ref (disassembly, def/use, liveness) consistent with execution,
+        which scales the raw field independently (addr_calc_scalar.h:
+        ``sbase = base + inst.sbase * 2``).
+        """
+        expr = f'reinterpret_cast<const OpEncoding*>(inst)->{opnd_name}'
+        if is_smem and opnd_name == 'sbase':
+            expr = f'({expr} * 2)'
+        if packed_16bit:
+            expr = f'static_cast<unsigned short>({expr})'
+        return expr
+
     def gen_insts(self) -> None:
         """Generate instruction classes deriving from encoding classes.
 
@@ -5133,11 +5152,9 @@ class CodeGenerator:
                                 )
                                 else ''
                             )
-                            operand_value = f'reinterpret_cast<const OpEncoding*>(inst)->{opnd.name}'
-                            if packed_16bit_source_arg:
-                                operand_value = (
-                                    f'static_cast<unsigned short>({operand_value})'
-                                )
+                            operand_value = self._operand_encoding_value_expr(
+                                opnd.name, is_smem, bool(packed_16bit_source_arg)
+                            )
                             opnd_ctor_init.append(
                                 f'{opnd.name}({opnd_size_expr}, '
                                 f'OperandType::{opr_type}, '
--- a/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
+++ b/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
@@ -4981,13 +4981,14 @@ class CodeGenerator:
     def _operand_encoding_value_expr(
         opnd_name: str, is_smem: bool, packed_16bit: bool
     ) -> str:
-        """C++ expression for an operand's raw encoding value.
+        """C++ expression for the decoded value passed to an Operand constructor.
 
-        SMEM SBASE is encoded in units of 2 SGPRs (raw N names s[2N:2N+1]), so it
-        is scaled here to a real SGPR index. This keeps the operand's
-        register-ref (disassembly, def/use, liveness) consistent with execution,
-        which scales the raw field independently (addr_calc_scalar.h:
-        ``sbase = base + inst.sbase * 2``).
+        For most operands this is just the value. SMEM SBASE is an
+        exception, as it is encoded in units of 2 SGPRs (where N gets
+        s[2N:2N+1]), so this helper scales it to the real SGPR index.
+        This keeps the operand's register-ref (disassembly, def/use, liveness)
+        consistent with execution, which scales the raw field independently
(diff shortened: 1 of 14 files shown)


emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -4340,6 +4340,25 @@ def _can_force_shared_simd_probe(
             self.isa_spec.profile.vop3p_opsel_fields,
         )
 
+    @staticmethod
+    def _operand_encoding_value_expr(
+        opnd_name: str, is_smem: bool, packed_16bit: bool
+    ) -> str:
+        """C++ expression for an operand's raw encoding value.
```
kuhar: This docstring should not call the result a raw encoding value. For SMEM `sbase`, the helper returns the scaled SGPR index used to construct the operand, so wording like operand-constructor expression or decoded operand value would be clearer.

emulation/rocjitsu/tests/smem_sbase_operand_test.cpp
```diff
@@ -0,0 +1,105 @@
...
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    SmemFamilies, SmemSbaseOperandTest,
+    ::testing::Values(
+        SmemCase{ROCJITSU_CODE_ARCH_CDNA1, "cdna1", kCdnaSLoadDwordx2, "s_load_dwordx2"},
+        SmemCase{ROCJITSU_CODE_ARCH_CDNA2, "cdna2", kCdnaSLoadDwordx2, "s_load_dwordx2"},
+        SmemCase{ROCJITSU_CODE_ARCH_CDNA3, "cdna3", kCdnaSLoadDwordx2, "s_load_dwordx2"},
+        SmemCase{ROCJITSU_CODE_ARCH_CDNA4, "cdna4", kCdnaSLoadDwordx2, "s_load_dwordx2"},
+        SmemCase{ROCJITSU_CODE_ARCH_RDNA4, "rdna4", kRdna4SLoadB64, "s_load_b64"}),
```
kuhar: We should add a gfx1250 case here. The PR regenerates `gfx1250/smem.cpp`, but this test only covers the CDNA form and RDNA4; gfx1250 has its own SMEM encoding shape, so RDNA4 does not exercise that generated path. The other RDNA variants are lower value, but gfx1250 is the missing case worth keeping.

emulation/rocjitsu/tests/smem_sbase_operand_test.cpp
```diff
@@ -0,0 +1,105 @@
...
+  const SmemCase &tc = GetParam();
+
+  auto decoder = Decoder::create(tc.arch);
+  ASSERT_NE(decoder, nullptr) << "Decoder::create() failed for " << tc.arch_name;
+
+  std::unique_ptr<Instruction> inst(decoder->decode(tc.word.data()));
+  ASSERT_NE(inst, nullptr) << "decode() returned nullptr for " << tc.arch_name;
+  ASSERT_EQ(inst->mnemonic(), tc.mnemonic) << "unexpected mnemonic for " << tc.arch_name;
+
+  // The SBASE is the SGPR-pair source operand (sdata is s0, width 1).
```
kuhar: This comment should say `sdata` is excluded because it is a destination, not because it is width 1. For `s_load_dwordx2`, `sdata` is a 64-bit `s[0:1]` operand; this loop works because it only scans source operands. We should either fix the comment or count/assert the single SGPR-pair source so a future encoding cannot silently select the wrong source.

kuhar (changes requested):
I see some gfx1250 RocJITsu clang/SIMD FPSan failures are introduced by this PR; the focused base-ref run passes the same checks.

There are two concrete generated-code regressions to fix:

- F64 parity aborts in `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/vopd.cpp`: this patch deletes the f64 VOPD handling (`is_float64_op()`, `apply_neg64()`, `execute_slot64()`, and the 64-bit read/write dispatch), while `op_name()` still decodes f64 dual ops. As a result `v_dual_min_num_f64` reaches the 32-bit `execute_slot()` path and throws `Unimplemented instruction: v_dual_min_num_f64`.

- F16/BF16 parity mismatches in the packed-16 operand path generated from `emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py`: this patch removes `packed_16bit_vgpr_dst()` and the `Operand::write_lane()` half-merge. The regenerated `gfx1250/operand.cpp` now writes a whole 32-bit VGPR for 16-bit destinations, which matches the observed signature where expected storage such as `0x3c00` comes back as the packed dword `0xbc003c00`. The true16 VOP3 SIMD changes need the same treatment: they are only safe if the SIMD path preserves source-half selection and destination-half merge semantics.

We should restore f64 VOPD execution and packed-16 destination writes, then rerun the PR-head clang/SIMD FPSan parity and atomic cases.

accauble:
@kuhar I think this was because the branch was stale. The PR was submitted last week and it hadn't been rebased since. I have rebased and see the missing functionality now.

amd-eochoalo:
Just FYI, I'm fixing the rocjitsu corpus right now.

emulation/rocjitsu/tests/smem_sbase_operand_test.cpp
```diff
@@ -0,0 +1,122 @@
...
+  ASSERT_EQ(inst->mnemonic(), tc.mnemonic) << "unexpected mnemonic for " << tc.arch_name;
+
+  // SBASE is the only SGPR-pair source operand. sdata is the destination
+  // and thus not included in the following loop. Assert exactly one
+  // SGPR-pair source so a future encoding cannot silently select the wrong one.
+  std::optional<RegisterRef> base;
+  int sgpr_pair_srcs = 0;
+  for (int i = 0; i < inst->num_src_operands(); ++i) {
+    auto ref = inst->src_operand(i)->to_register_ref();
+    if (ref && ref->cls == RegClass::SGPR && ref->width == 2) {
```
kuhar: We should extend this regression test beyond the 64-bit load shape. The regenerated SMEM constructors also scale `sbase` for 128-bit buffer-descriptor operands (`sbase(128, ...)`) and for store-family operands where SBASE is not the first source operand. This test only counts `ref->width == 2` sources, so it would miss a future regression in those generated forms. One `s_buffer_load_dwordx4` case expecting `s[4:7]` plus one `s_store_dwordx2` case would cover the important missing shapes.

kuhar (approved):
LGTM

PR #7468, opened 2026-06-18
rocprofiler-register: fix deadlock error

meserve-amd:
- With attachment, it is now possible for new API tables to be registered while invoking all previous registrations
  - Previously, this would always result in ROCP_REG_DEADLOCK
  - A checked_lock is now used which allows multiple threads but still disallows recursion

## Motivation

Fixes an intermittent error in rocattach-execute tests that could occur if roctx was registered during attachment

Also ran `make format` which hadn't been done for a bit, evidently.

## Technical Details

Previously rocprofiler-register did not have to worry about multithreaded scenarios, so had a simple "no deadlock" check provided that also acted as a (very strict) mutex. This upgrades that check to allow for multiple threads while still giving the same error if the same thread accesses the lock.

## Test Plan

Ran rocattach-execute tests ~200 times locally

## Test Result

All passes no errors detected

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-register/source/lib/rocprofiler-register/details/checked_lock.hpp
+++ b/projects/rocprofiler-register/source/lib/rocprofiler-register/details/checked_lock.hpp
@@ -1,0 +1,90 @@
+// MIT License
+//
+// Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


ApoKalipse-V (commented):
Is the test failure real or unrelated?

meserve-amd:
> Is the test failure real or unrelated?

The rocprofiler-systems failures are resolved. The thread sanitizer failures are ignorable for now, they are in lower libraries

jrmadsen (approved):

ApoKalipse-V (approved):

PR #7469, opened 2026-06-18
[aqlprofile] Add SP block support for gfx1250

Sushma1203:
## Motivation
Add SP performance counter block support for gfx1250 in aqlprofile.

## Technical Details

The SP block shares the same physical `SQ_PERFCOUNTER` registers as the SQ block on gfx1250, so the change both defines the block and adds the register-sharing logic needed to avoid collisions between SP and SQ counters.

**Block definition:**
- `src/core/include/aqlprofile-sdk/aql_profile_v2.h` — add `AQLPROFILE_BLOCK_NAME_SP` to the `aqlprofile_block_name_t` enum.
- `gfxip/gfx12/gfx12_block_info.h` — add SP constants in the `gfx1250` namespace: `SpCounterBlockNumInstances = 1`, `SpCounterBlockNumCounters = 4`, `SpCounterBlockMaxEvent = 1023`.
- `gfxip/gfx12/gfx12_block_table.h` — add `SpCounterRegAddr[]` (4 entries reusing `regSQ_PERFCOUNTER` SELECT 0/4/8/12 → LO 0/2/4/6) and `SpCounterBlockInfo`, both guarded by `GFX12_VARIANT == GFX12_VARIANT_1250`. Attributes match SQ: `CounterBlockSeAttr|CounterBlockSaAttr|CounterBlockWgpAttr|CounterBlockSqAttr` with the `sq_select_value` primitive.
- `src/core/gfx1250_factory.cpp` — register the block: `block_table_[__BLOCK_ID(SP)] = &SpCounterBlockInfo;`.

**Counter index handling (SP/SQ shared registers):**
- `src/core/counters.cpp` — `GetCounter()` takes a new `num_sp_events` argument. Because SP and SQ share `SQ_PERFCOUNTER` registers, SQ events with `reg_index < num_sp_events` are remapped to odd register slots (`2*reg_index + 1`) and the remaining SQ events are offset by `num_sp_events`. `CountersVec()` computes `num_sp_events` by counting `AQLPROFILE_BLOCK_NAME_SP` events before assigning registers.
- `src/core/memorymanager.cpp` — change `std::sort` to `std::stable_sort` (both occurrences) so events that compare equal keep their relative order, which the SP/SQ register remapping depends on.

## Submission Checklist


projects/aqlprofile/src/core/include/aqlprofile-sdk/aql_profile_v2.h
```diff
@@ -91,6 +91,7 @@ typedef enum {
   AQLPROFILE_BLOCK_NAME_SQG,
 
   // New blocks for gc_12_1_x
+  AQLPROFILE_BLOCK_NAME_SP,
```
ApoKalipse-V: should be at the end

ApoKalipse-V (changes requested):
This needs to be in internal aqlprofile

ApoKalipse-V (approved):

Issue #7025, opened 2026-06-10
[rocjitsu] Liveness analysis treats every vector def as a non-kill


accauble:
# Liveness analysis treats every vector def as a non-kill

## Summary

Liveness (`lib/rocjitsu/src/rocjitsu/analysis/`) currently assumes that *every* VGPR / ACC_VGPR def is EXEC-masked, and so refuses to treat any vector def as a kill. The flag responsible, `InstDefUse::has_exec_masked_vector_def` (set in `def_use_chain.cpp ) flips on for any vector destination, and `kill_defs` in `liveness.cpp` responds by stripping every VGPR / ACC_VGPR from the kill set. The practical effect is that any vector register used anywhere downstream stays live all the way back to the top of the kernel scope, or to its first read that is itself a write. Vector liveness is essentially useless for scratch allocation today.

In other words, the analysis over-approximates liveness, so anything built on it (e.g. the planned DBI probe trampolines) gets safe-but-pessimistic answers and falls back to higher scratch indices than necessary.
Closing the gap unlocks tighter VGPR scratch allocation without changing any callers, and is the precondition for any liveness-driven register renaming or coalescing later on.

## The problem

The relevant snippets, in `def_use_chain.cpp`:

```cpp
// TODO: Replace this class-based approximation with instruction metadata
// that identifies vector defs whose inactive lanes are preserved under EXEC,
// and pair it with program-point EXEC state so full-EXEC writes can be
// treated as normal kills.
[[nodiscard]] bool is_exec_masked_def(RegisterRef ref) {
  return ref.cls == RegClass::VGPR || ref.cls == RegClass::ACC_VGPR;
}
```

and in `liveness.cpp:63–76`:

```cpp
[[nodiscard]] RegisterSet kill_defs(const InstDefUse &du) {
  RegisterSet kills = du.defs;
  if (du.has_predicated_def)
    return {};
  if (du.has_exec_masked_vector_def) {
    kills.clear_class(RegClass::VGPR);
    kills.clear_class(RegClass::ACC_VGPR);
  }
  return kills;
}
```

The TODO calls out the two missing pieces:

1. **Per-instruction metadata**: knowing which vector defs actually preserve inactive lane values under EXEC, rather than assuming all of them do. Almost all vector ALU ops are exec-masked, but a few aren't, and the flag is currently unable to make that distinction.
[...]

PR #7470, opened 2026-06-18
feat: [rocjitsu] Adding EXEC state tracking for liveness

theSK2005:
Exposed flags to label instructions that ignore EXEC or write to EXEC so it is accessible by liveness. Then iterate EXEC state per instruction in reverse post order so that liveness VGPR/accVGPR kills can be done properly. Moved some resolve_scalar functions into a shared header to reduce code overlap.

Defer handling instructions that require EXEC to be AllOnes.

Addresses Issue #7025.


emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -104,6 +104,68 @@
 }
 
 
+def _exec_mask_flag_stmts(sem) -> list[str]:
+    """Return ``flags_ |= ...;`` statements for EXEC-mask instruction metadata.
+
+    The flags are derived from the instruction's semantic AST so generic
+    liveness/dataflow analyses do not have to know AMDGPU instruction names:
+
+    * ``EXEC_MASKED``  - VECTOR exec-model: inactive lanes preserve old values.
```
accauble: Could you point me to where `EXEC_MASKED` and `READS_EXEC` are used and/or do you have plans to use them?

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -104,6 +104,68 @@
...
+            (InstructionProperty.READS_EXEC, 'READS_EXEC'),
+        )
+        if prop in props
+    ]
+
+
+def _result_combinator_flag_stmts(sem) -> list[str]:
+    """Return ``flags_ |= ...;`` for how the instruction forms its result value.
+
+    EXEC-state analysis uses this to prove an all-ones EXEC write:
```
accauble: Nit: This comment makes me think this is pretty specific to handling EXEC but the function is much more general than that. Could you rewrite the comment to describe the functionality without the analysis? A little comment at the end that says its currently used by EXEC-state analysis is fine.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -104,6 +104,68 @@
...
+    (and, xor, not, cmov, ...) get no flag, so the analysis cannot prove
+    all-ones and stays conservative.
+    """
+    if sem is None:
+        return []
+    cls = sem.semantic_class
+    op = (sem.operation or '').lower()
+    if cls == 'scalar_mov':
+        return ['flags_ |= RESULT_COPY;']
+    if cls in ('scalar_binop', 'scalar_saveexec') and op == 'or':
```
accauble, line 169: The comment here and in instruction.h do not specify scalar instructions but it appears you are limiting this to scalar instructions. Either this function is incomplete or it is described too generally.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/instruction.h
```diff
@@ -46,7 +46,24 @@ enum InstFlags : uint64_t {
...
+  /// @brief Reads the EXEC mask.
+  READS_EXEC = (1ULL << 15),
+  /// @brief The destination value is a plain copy of a single source operand
+  /// (e.g. s_mov). Lets EXEC-state analysis prove an all-ones EXEC write from an
+  /// all-ones source.
+  RESULT_COPY = (1ULL << 16),
+  /// @brief The destination value is the bitwise OR of the source operands
+  /// (e.g. s_or, s_or_saveexec). OR with an all-ones operand is all-ones
+  /// regardless of the others.
+  RESULT_OR = (1ULL << 17)
```
accauble: I think this only includes instructions that exactly do "or" (as opposed to "or" + "not". Since "RESULT_ORN" does not exist, it should probably be explicitly stated that this is for "pure or" instructions only.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -104,6 +104,68 @@
...
+            (InstructionProperty.EXEC_MASKED, 'EXEC_MASKED'),
+            (InstructionProperty.IGNORES_EXEC, 'IGNORES_EXEC'),
+            (InstructionProperty.WRITES_EXEC, 'WRITES_EXEC'),
+            (InstructionProperty.READS_EXEC, 'READS_EXEC'),
+        )
+        if prop in props
+    ]
+
+
+def _result_combinator_flag_stmts(sem) -> list[str]:
```
accauble, line 151: Is there a way that these could be derived from the AST like the EXEC properties above?

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -7450,6 +7541,36 @@ def gen_operand(self) -> None:
         resolve_code = cgen.Line(
             'namespace {\n'
             '\n'
+            '// Wavefront-free subset of resolve_src_scalar: the value of an inline\n'
+            '// constant (small integers 0..64 / -1..-16 and the inline float\n'
+            '// constants), or nullopt for any other encoding value. Negative inline\n'
+            '// integers are sign-extended to 64 bits so 64-bit consumers see all-ones.\n'
+            'std::optional<uint64_t> resolve_inline_const(int ev) {\n'
```
accauble: Nit: A better name might be "resolve_src_scalar_statically" or something to distinguish it as a wavefront-free version when someone is reading code.

Another nit: This looks like it could be shared by all archs at this time (at least how it is currently written). There used to be a way to do that and I would be interested to know what it would take to not duplicate the code across all the operand files.

accauble (changes requested):
This is my first pass of the codegen. Haven't reviewed the analysis yet. Mostly nit-picky things and a couple questions I have that could improve the code if it is feasible.

atgutier (approved):
This overall LGTM.

An agent review flagged a few things that would be good to address:

MINOR (recommended before merge)

```
  1. exec_state.cpp:131-161 — loop-header-as-scope-entry seeding edge case (soundness-adjacent). A
  block that is both the kernel-scope entry and the target of an in-scope back-edge (loop header at
  offset 0) is classified non-entry and seeded optimistically Full instead of Unknown — the only path
  by which the forward MUST analysis could compute Full where it should be Unknown (→ an unsound
  kill). In practice the absolute first block of a kernel cannot have an in-scope back-edge, so it's
  robustness, not an observed bug. Fix: treat a block as entry-for-seeding if it's the scope leader OR
   has no in-scope predecessor.
  2. RESULT_COPY/RESULT_OR derived by string-matching semantic_class/operation while
  WRITES_EXEC/IGNORES_EXEC come from the AST. A future EXEC-writing copy/or opcode under a class not
  in the hand-list silently loses the combinator flag (degrades safe, but latent drift). Fix: derive
  the combinator from the same AST path so the two flag sets can't diverge.

  NIT

  - exec_state.h:7-13 file-level @details ends mid-sentence ("two-point lattice:") without enumerating
   Full/Unknown.
  - Internal enums (ExecWrite, Combinator) lack explicit : uint8_t underlying type while public
  ExecState has it.
  - binary_translator.cpp:567 benign uint8_t→uint32_t wave-size widening; an assert(wave_size == 32 ||
   64) in the ExecMaskAnalysis ctor would document the invariant.
  - s_cbranch_execz/execnz get IGNORES_EXEC (correct — branches aren't lane-masked) but no READS_EXEC.
   These branches do test EXEC; whether that should mark EXEC live for liveness is a semantics call
  worth deciding. If yes, it's a generator change (emit READS_EXEC for those opcodes).
```

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/scalar_operand_resolve.h
```diff
@@ -0,0 +1,302 @@
+// Copyright (c) 2025-2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: MIT
+//
+// AUTO-GENERATED by the amdisa codegen pipeline. DO NOT EDIT.
```
accauble: If this is no longer auto-generated, the header will need to look like a non-autogenerated header.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/scalar_operand_resolve.h
```diff
@@ -0,0 +1,302 @@
...
+  if (ev == 248)
+    return static_cast<uint64_t>(0x3E22F983u); // 1/(2*pi)
+  return std::nullopt;
+}
+
+// The value of a scalar source operand resolvable from wavefront state alone
+// (SGPR/VCC/EXEC/M0 reads plus inline constants). `m0_ev` is the M0 encoding
+// value for this arch (124 on most arches; 125 on RDNA4 / GFX1250, where 124
+// is the NULL slot).
+inline uint32_t resolve_src_scalar(const Wavefront &wf, int ev, int m0_ev) {
```
accauble, line 28: It looks like the changes you made here are breaking the amdisa tests. Can you please fix those? As you're fixing them, I think it would be good to make sure all 10 ISAs have the right value for M0.

Alternatively, you can roll back the refactor and make it a separate PR to keep the changes contained.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/scalar_operand_resolve.h
```diff
@@ -0,0 +1,302 @@
...
+  if (ev == 247)
+    return static_cast<uint64_t>(0xC0800000u); // -4.0f
+  if (ev == 248)
+    return static_cast<uint64_t>(0x3E22F983u); // 1/(2*pi)
+  return std::nullopt;
+}
+
+// The value of a scalar source operand resolvable from wavefront state alone
+// (SGPR/VCC/EXEC/M0 reads plus inline constants). `m0_ev` is the M0 encoding
+// value for this arch (124 on most arches; 125 on RDNA4 / GFX1250, where 124
```
accauble: Please rewrite the comment so that it does not imply that M0 is 124 for rdna3 and rdna3.5

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/def_use_chain.cpp
```diff
@@ -10,32 +10,32 @@ namespace rocjitsu {
 
 namespace {
 
-// TODO: Replace this class-based approximation with instruction metadata that
-// identifies vector defs whose inactive lanes are preserved under EXEC, and pair
-// it with program-point EXEC state so full-EXEC writes can be treated as normal
-// kills.
-[[nodiscard]] bool is_exec_masked_def(RegisterRef ref) {
+// A vector def (VGPR/AccVGPR) normally preserves inactive lanes under EXEC, so
+// it cannot be treated as an unconditional kill. The exception is instructions
```
accauble: Nit: The comment should probably be closer to the relevant logic

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/exec_state.cpp
```diff
@@ -0,0 +1,263 @@
...
+enum class ExecWrite {
+  None,      ///< Does not write EXEC; state unchanged.
+  AllOnes,   ///< Writes the whole EXEC mask to all-ones -> Full.
+  Preserve,  ///< Sets a subset of EXEC to all-ones (e.g. only exec_lo on
+             ///< Wave64); keeps an already-Full mask Full and leaves Unknown
+             ///< as Unknown, but cannot establish Full on its own.
+  Narrowing, ///< Writes EXEC some other way -> Unknown.
+};
+
+[[nodiscard]] bool writes_exec(const Instruction &inst) {
```
accauble, line 31: Should the WRITES_EXEC flag be set for instructions that have it as a destination register? In other words, do we make the generator a bit more complex so that users do not have to remember to check the destination regs? Or perhaps moving it to `instruction.h`? @atgutier @kuhar I imagine you might have an opinion on this.

Either way, the only description for the flag is "Writes the EXEC mask." which is too broad for what it currently represents.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/exec_state.h
```diff
@@ -0,0 +1,64 @@
...
+/// participate, and edges leaving the scope are ignored. Blocks with no
+/// in-scope predecessor are treated as entries and seeded with `Unknown`.
+class ExecMaskAnalysis {
+public:
+  /// @brief Compute EXEC state for one kernel's block set.
+  /// @param wave_size Lanes per wavefront (EXEC bit width): 64 for Wave64, 32
+  ///        for Wave32. Used to tell a full-EXEC write from a partial half-write
+  ///        (e.g. `s_mov_b32 exec_lo` is the whole mask on Wave32 but only half
+  ///        on Wave64). Defaults to 64 (conservative for the common CDNA case).
+  explicit ExecMaskAnalysis(KernelBlockScope blocks, uint32_t wave_size = 64);
```
accauble: Can you actually default this to 0 and have it fail if it's not set? We can let liveness default to 64. (Update the comment accordingly too please). While you're there, making sure the wave size is an appropriate number would be good too.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/liveness.h
```diff
@@ -108,6 +113,7 @@ class LivenessAnalysis {
   void analyze(KernelBlockScope blocks);
 
   uint16_t min_free_vgpr_ = 0;
+  uint32_t wave_size_ = 64;
```
accauble: Let's make this 0 and keep the default setting to just one location (I think options makes the most sense).

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/instruction.h
```diff
@@ -46,7 +46,19 @@ enum InstFlags : uint64_t {
...
-  PREDICATED_DEF = (1ULL << 11)
+  PREDICATED_DEF = (1ULL << 11),
+  /// @brief Executes regardless of the EXEC mask (e.g. branches).
+  IGNORES_EXEC = (1ULL << 12),
+  /// @brief Writes the EXEC mask.
+  WRITES_EXEC = (1ULL << 13),
+  /// @brief The destination value is a plain copy of a single source operand
+  /// (e.g. s_mov). Lets EXEC-state analysis prove an all-ones EXEC write from an
+  /// all-ones source.
+  RESULT_COPY = (1ULL << 14),
```
accauble, line 58: RESULT_COPY and RESULT_OR are currently only set for scalar operations. That should be evident here. Additionally, do you expect things to break if you these eventually get generalized to vector operations? If so, then that should be clear and please make sure there will be a test to capture that failure.

emulation/rocjitsu/tests/analysis/liveness_test.cpp
```diff
@@ -361,6 +432,217 @@ TEST(LivenessAnalysis, ExecMaskedVgprDefDoesNotKillInactiveLaneValue) {
   EXPECT_NE(*free_vgpr, 0);
 }
 
+// Collect a block's instructions in order for index-based assertions.
```
accauble, line 3732: Pretty comprehensive testing. I'd love to see a some tests that make sure the wave size gets passed down correctly through Liveness? (Like, the option sets the wave size correctly, and that liveness passes the correct value to the exec analysis correctly.

emulation/rocjitsu/lib/python/amdisa/tests/test_exec_mask_flags.py
```diff
@@ -0,0 +1,151 @@
...
+            'S_OR_SAVEEXEC_B64', 'scalar_saveexec', operation='or', data_type='b64'
+        )
+        stmts = _exec_mask_flag_stmts(sem)
+        assert stmts
+        for s in stmts:
+            assert s.startswith('flags_ |= ')
+            assert s.endswith(';')
+
+
+class TestResultCombinatorFlagStmts:
```
accauble, line 127: If these need to remain scalar, please add non-scalar instructions that fail with the reasoning why.

accauble (changes requested):
Overall I think this is great. There is still some clean up required. I like the refactor that moves code to the scalar_operand_resolve header but I think @kuhar should weigh in on it once the tests all pass.

theSK2005:
> This overall LGTM.
> 
> An agent review flagged a few things that would be good to address:
> 
> MINOR (recommended before merge)
> 
> ```
>   1. exec_state.cpp:131-161 — loop-header-as-scope-entry seeding edge case (soundness-adjacent). A
>   block that is both the kernel-scope entry and the target of an in-scope back-edge (loop header at
>   offset 0) is classified non-entry and seeded optimistically Full instead of Unknown — the only path
>   by which the forward MUST analysis could compute Full where it should be Unknown (→ an unsound
>   kill). In practice the absolute first block of a kernel cannot have an in-scope back-edge, so it's
>   robustness, not an observed bug. Fix: treat a block as entry-for-seeding if it's the scope leader OR
>    has no in-scope predecessor.
>   2. RESULT_COPY/RESULT_OR derived by string-matching semantic_class/operation while
>   WRITES_EXEC/IGNORES_EXEC come from the AST. A future EXEC-writing copy/or opcode under a class not
>   in the hand-list silently loses the combinator flag (degrades safe, but latent drift). Fix: derive
>   the combinator from the same AST path so the two flag sets can't diverge.
> 
>   NIT
> 
>   - exec_state.h:7-13 file-level @details ends mid-sentence ("two-point lattice:") without enumerating
>    Full/Unknown.
>   - Internal enums (ExecWrite, Combinator) lack explicit : uint8_t underlying type while public
>   ExecState has it.
>   - binary_translator.cpp:567 benign uint8_t→uint32_t wave-size widening; an assert(wave_size == 32 ||
>    64) in the ExecMaskAnalysis ctor would document the invariant.
>   - s_cbranch_execz/execnz get IGNORES_EXEC (correct — branches aren't lane-masked) but no READS_EXEC.
>    These branches do test EXEC; whether that should mark EXEC live for liveness is a semantics call
>   worth deciding. If yes, it's a generator change (emit READS_EXEC for those opcodes).
> ```

Fixed the all the issues except comment #2. I think moving the RESULT_COPY and RESULT_R flags into the AST could be a follow-up PR.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
accauble: For a follow-up PR

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/instruction.h
accauble: Please address this

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/def_use_chain.cpp
accauble: Not addressed

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/exec_state.h
```diff
@@ -0,0 +1,64 @@
...
+private:
+  void analyze(KernelBlockScope blocks);
+
+  struct BlockExec {
+    ExecState in = ExecState::Full;
+    ExecState out = ExecState::Full;
+    bool is_entry = false;
+  };
+
+  uint8_t wave_size_ = 64;
```
accauble: Don't set a default here. Confusing since the default is supposed to be 0

accauble (changes requested):
Still need to update a default wave size in `exec_state.h`.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -104,6 +104,63 @@
 }
 
 
+def _exec_mask_flag_stmts(sem) -> list[str]:
+    """Return ``flags_ |= ...;`` statements for EXEC instruction metadata.
+
+    The flags are derived from the instruction's semantic AST so generic
+    liveness/dataflow analyses do not have to know AMDGPU instruction names:
+
+    * ``IGNORES_EXEC`` - branch-style instructions that run regardless of EXEC.
```
Groverkss, line 123: Some instructions require EXEC to be all true. Can we also add that? For example, ds_read_tr instructions and mfma instructions require EXEC to be all true. That is a good state point to add in the analysis (and is useful in general as well).

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/def_use_chain.h
```diff
@@ -25,10 +25,11 @@ class InstDefUse {
   /// @param inst Decoded instruction whose operands have stable lifetimes.
   InstDefUse(const Instruction &inst);
 
-  RegisterSet defs;                        ///< Registers overwritten by the instruction.
```
Groverkss, line 36: Why change this? The only thing changed here was more readable spacing -> less readable spacing

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/liveness.h
```diff
@@ -64,6 +64,11 @@ struct LivenessAnalysisOptions {
   /// while scratch allocation can be forced above a descriptor-declared VGPR
   /// range to test whether semantic lowerings clobber guest registers.
   uint16_t min_free_vgpr = 0;
+
+  /// @brief Lanes per wavefront (EXEC bit width), passed to the EXEC-state
+  /// analysis so it can tell full-EXEC writes from partial half-writes. 64 for
+  /// Wave64, 32 for Wave32. Defaults to 64.
```
Groverkss: I don't think we should make this default to 64. Instead require it to be passed to the constructor.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/liveness.cpp
```diff
@@ -110,6 +114,9 @@ void LivenessAnalysis::analyze(KernelBlockScope blocks,
       block_index_.emplace(blocks[i], i);
   }
 
+  // Program-point EXEC approximation: lets EXEC-masked vector defs count as
+  // kills where EXEC is provably full.
+  const ExecMaskAnalysis exec(blocks, wave_size_);
```
Groverkss: Instead of building it inside liveness, maybe pass it to liveness through a reference instead. If multiple analysis use Exec analysis (which they probably will do), you want to pass analysis by reference instead of building them again and again inside.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/def_use_chain.h
accauble: I'm guessing the formatter did that.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/liveness.cpp
```diff
@@ -60,14 +61,16 @@ std::vector<const Instruction *> instructions_in_order(BasicBlock &block) {
...
-  // writes cannot be treated as unconditional liveness kills.
+  // Predicated defs preserve old values on at least one control-flow path, so
+  // they can never be unconditional kills.
   if (du.has_predicated_def)
     return {};
-  if (du.has_exec_masked_vector_def) {
+  // EXEC-masked vector defs preserve inactive lanes' old values, so they are
+  // kills only where EXEC is provably full (every lane overwritten). Where the
+  // EXEC state is unknown we stay conservative and do not kill.
+  if (du.has_exec_masked_vector_def && exec_before != ExecState::Full) {
```
kuhar, line 71: We should only promote a vector def to a kill when it overwrites the entire destination, not just when EXEC is Full. CDNA4 v_writelane_b32 is a concrete counterexample: it exposes a VGPR destination, writes one selected lane, and does not model the preserved lanes as a use. With Full EXEC, kill_defs currently drops the whole VGPR from liveness, so scratch allocation can clobber the preserved lanes. Please add a whole-destination-write contract, or model partial defs as read-modify-write, and cover v_writelane_b32 with a real-decode regression.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/exec_state.h
```diff
@@ -0,0 +1,64 @@
...
+/// participate, and edges leaving the scope are ignored. Blocks with no
+/// in-scope predecessor are treated as entries and seeded with `Unknown`.
+class ExecMaskAnalysis {
+public:
+  /// @brief Compute EXEC state for one kernel's block set.
+  /// @param wave_size Lanes per wavefront (EXEC bit width): 64 for Wave64, 32
+  ///        for Wave32. Used to tell a full-EXEC write from a partial half-write
+  ///        (e.g. `s_mov_b32 exec_lo` is the whole mask on Wave32 but only half
+  ///        on Wave64). Defaults to 0 and fails if not set.
+  explicit ExecMaskAnalysis(KernelBlockScope blocks, uint8_t wave_size = 0);
```
kuhar: The EXEC analysis and liveness need to use the same scoped CFG. binary_translator.cpp passes scoped call and return edges only to LivenessAnalysis, while ExecMaskAnalysis meets and enqueues only BasicBlock predecessor and successor edges. A callee loop header or return continuation can therefore remain Full even when an omitted call or return predecessor carries Unknown, allowing liveness to turn an EXEC-masked vector def into an unsound kill. Please pass the same extra edges to ExecMaskAnalysis and add a call or return disagreement regression.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/exec_state.cpp
```diff
@@ -0,0 +1,267 @@
...
+      continue;
+    bool has_in_scope_pred = false;
+    for (const BasicBlock *pred : block->predecessors()) {
+      if (block_index_.contains(pred)) {
+        has_in_scope_pred = true;
+        break;
+      }
+    }
+    // consider entry to be scope leader and blocks with no predecessors in scope
+    states_[i].is_entry = (!has_in_scope_pred || i == 0);
```
kuhar: Please add a regression for a back-edge into the scope entry. The new i == 0 condition is the only reason an entry self-loop stays Unknown; without it, the default Full state meets with itself and can enable an unsafe vector kill. A direct ExecMaskAnalysis test using DefVgpr0 followed by BranchBackToStart and checking before(def) == Unknown would lock this in.

kuhar (commented):
A few comments

accauble:
@theSK2005 could you rebase please?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
```diff
@@ -488,7 +489,8 @@ InstrumentedCodeObjectDebug Instrumentor::patch_with_debug_summaries() {
   liveness_scope.reserve(blocks_.size());
   for (const auto &block : blocks_)
     liveness_scope.push_back(block.get());
-  const LivenessAnalysis liveness{KernelBlockScope(liveness_scope)};
+  const ExecMaskAnalysis exec{KernelBlockScope(liveness_scope), /*wave_size=*/64};
```
kuhar: We should pass the kernel's actual wave size to `ExecMaskAnalysis` here, or explicitly reject Wave32. `Instrumentor` accepts decoder-backed RDNA architectures, and RDNA kernel descriptors can select Wave32. On Wave32, an all-ones `exec_lo` write establishes `Full`; analyzing it as Wave64 leaves the state `Unknown`, so probe liveness can report dead VGPRs as live and reject instrumentation as requiring unsupported spills. This likely needs per-kernel scopes and descriptor-derived wave sizes rather than one analysis over every decoded block.

kuhar (commented):
A couple of comments, the other one is above: https://github.com/ROCm/rocm-systems/pull/7470#discussion_r3599252174

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/liveness.cpp
kuhar: We should reopen this because the merged partial-def support still does not cover `v_writelane_b32`. A real CDNA4 decode of `s_mov_b64 exec, -1; v_writelane_b32 v0, s0, s0; v_mov_b32 v2, v0; s_endpgm` reports `v0` in `defs` but not `uses`; after EXEC becomes `Full`, `live_before()` incorrectly drops `v0` even though the instruction writes one lane and preserves the rest. Please model `v_writelane_b32` (and equivalent lane-partial writers) as reading the destination, or gate kills on whole-destination coverage, and add this real-decode regression.

kuhar (commented):
We should account for the EXEC half selected by `RegisterRef::index` before treating an all-ones write as full. At `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/exec_state.cpp:66-76`, every 32-bit EXEC destination currently gets `mask == 0xffffffff` and `full == true` on Wave32, so `s_mov_b32 exec_hi, -1` can move `Unknown` to `Full` even though the active low half is unchanged. A later vector def may then be killed unsafely. Please add a Wave32 regression showing that an `exec_hi` write cannot establish `Full`, while it preserves an existing `Full` state.

It would be good to keep the EXEC operand-access regression when moving the resolver to the shared header. Commit `1f3c53ce` removed `test_operand_exec_register_access_is_wave32_gated`, and the remaining shared-header assertions only cover M0 (`test_generator_profile_gates.py:2657`); nothing checks the `exec_lo`/`exec_hi` read and write branches in `shared/scalar_operand_resolve.h:45-48,232-238`. Please relocate the deleted assertions alongside the M0 checks so changes to the Wave32/Wave64 split stay covered.

kuhar (commented):
We should keep a Wave32 `exec_hi` write state-neutral regardless of the value written. At `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/exec_state.cpp:79`, the current fix stops `exec_hi = -1` from establishing `Full`, but `exec_hi = 0` still falls through to `Narrowing` and changes an existing `Full` to `Unknown` even though the active low half is untouched. A focused `Full -> exec_hi = 0 -> vector def` repro sees `Unknown`. Please classify writes disjoint from `full_exec_mask(wave_size)` as preserving and add that regression.

It would be good to remove the orphaned `exec_register` and `set_exec_register` locals, plus their Wave32 comment, at `emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py:9508`. EXEC operand access now goes through the shared resolver and semantic-lowering paths, and a whole-word search at the current head finds only these two definitions.

newling:
The issue has 2 sub-problems:

> The TODO calls out the two missing pieces:
Per-instruction metadata: knowing which vector defs actually preserve inactive lane values under EXEC, rather than assuming all of them do. Almost all vector ALU ops are exec-masked, but a few aren't, and the flag is currently unable to make that distinction.
Program-point EXEC state: knowing when EXEC is statically all-ones at a given instruction. When it is, an exec-masked def is a full kill again (all lanes active -> all lanes written -> no preserved values).

This PR summary seems to only mention the second. Just wondering if the issue should be closed by this PR? For example, if EXEC is not -1 (i.e. there are inactive lanes) and an instruction that ignores EXEC (mfma maybe?) writes to a register, that register should be killed (whereas if a normal instruction writes to register it shouldn't be considered killed).

theSK2005:
> The issue has 2 sub-problems:
> 
> > The TODO calls out the two missing pieces:
> > Per-instruction metadata: knowing which vector defs actually preserve inactive lane values under EXEC, rather than assuming all of them do. Almost all vector ALU ops are exec-masked, but a few aren't, and the flag is currently unable to make that distinction.
> > Program-point EXEC state: knowing when EXEC is statically all-ones at a given instruction. When it is, an exec-masked def is a full kill again (all lanes active -> all lanes written -> no preserved values).
> 
> This PR summary seems to only mention the second. Just wondering if the issue should be closed by this PR? For example, if EXEC is not -1 (i.e. there are inactive lanes) and an instruction that ignores EXEC (mfma maybe?) writes to a register, that register should be killed (whereas if a normal instruction writes to register it shouldn't be considered killed).

There is an IGNORES_EXEC flag that is already part of the semantics that this PR accounts for by making them visible to liveness in the codegen. The codegen surfaces the flags [here](https://github.com/ROCm/rocm-systems/pull/7470/changes#diff-afcf004f21e3a5a09a268372fc931561b19d4cca9c5aff7e38d5bf3066fa5b4fR117) and an example of it showing up properly in codegen is [here](https://github.com/ROCm/rocm-systems/pull/7470/changes#diff-45880a4cb03f61b5aed9e8bf1e2deb02e4ad6d91348700839af37e176fb92d19R46). Liveness consumes it [here](https://github.com/ROCm/rocm-systems/pull/7470/changes#diff-44bc25b0a8853697a894578df634ac384d5874b89043de03c496ac48b2968a8fR67).

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/amdgpu_code_object.cpp
```diff
@@ -336,4 +336,36 @@ std::optional<uint32_t> AmdGpuCodeObject::min_kernel_sgpr_count(rj_code_arch_t a
...
+      KD desc;
+      std::memcpy(&desc, section->data() + off, sizeof(desc));
+      const bool wave32 = AMDHSA_BITS_GET(desc.kernel_code_properties,
+                                          kd::KERNEL_CODE_PROPERTY_ENABLE_WAVEFRONT_SIZE32);
+      saw_kernel = true;
+      all_wave32 = all_wave32 && wave32;
+      break;
+    }
+  }
+  return (saw_kernel && all_wave32) ? 32 : 64;
```
newling, line 350: saw_kernel == saw any kernel. 

all_wave32 == of the kernels we saw, they were all wave32. 

What happens if one of the kernels you didn't see is wave64?

If there is an implicit assumption about additional constraints/structure here, please add asserts to make it clear.

newling (commented):
I had an (arguably) interesting conversation with an agent about doing this analysis at a per-lane level. i.e. for each lane, keeping a bit (or 2 bits) specifying if the EXEC value for it 1, or unknown (or 1, 0, or unknown for a tighter analysis). It would allow a few more kill sites to be detected statically. I guess the questions are: would this help in reality? and does this simplify/complexify code? Dump from agent is here: https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/discussion_pr7470_exec_liveness_granularity.md

Also got an agent to review the PR: https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr7470.md. Apologies for it being so overly verbose (I need to change my prompt...) but 2 actionable items seem valid:

1) wave32 is being overly conservative on going FULL -> UNKOWN because in places where it needn't behave the same way as wave32

2) a case of confusion in choosing wave32 or wave64 (I've added a comment inline for this too). 

Neither is a blocker IMO but nice to haves. Feel free to address/ignore its other review comments too

theSK2005:
> I had an (arguably) interesting conversation with an agent about doing this analysis at a per-lane level. i.e. for each lane, keeping a bit (or 2 bits) specifying if the EXEC value for it 1, or unknown (or 1, 0, or unknown for a tighter analysis). It would allow a few more kill sites to be detected statically. I guess the questions are: would this help in reality? and does this simplify/complexify code? Dump from agent is here: https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/discussion_pr7470_exec_liveness_granularity.md
> 
> Also got an agent to review the PR: https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr7470.md. Apologies for it being so overly verbose (I need to change my prompt...) but 2 actionable items seem valid:
> 
> 1. wave32 is being overly conservative on going FULL -> UNKOWN because in places where it needn't behave the same way as wave32
> 2. a case of confusion in choosing wave32 or wave64 (I've added a comment inline for this too).
> 
> Neither is a blocker IMO but nice to haves. Feel free to address/ignore its other review comments too

Per lane EXEC state tracking would definitely be an upgrade, but I think that most of the EXEC defs that set it back to AllOnes would be setting EXEC to AllOnes directly rather than specifically turning the 0 lane to 1 (get AllOnes with EXEC = EXEC or !EXEC). Like, if there's an if/else where some lanes are active for the if and others are active for the else, when they merge back, I think that EXEC is set to AllOnes directly in the way that this PR sees. I'll look into those two issues.

newling:
> Per lane EXEC state tracking would definitely be an upgrade, but I think that most of the EXEC defs that set it back to AllOnes would be setting EXEC to AllOnes directly [...]

makes sense. Per lane tracking would help with something like

```
  s_mov_b32 exec_lo, -1
  s_mov_b32 exec_hi, -1
  v_mov_b32 v0, s0
```

but how often compilers generate this kind of sequence is unknown. I guess the normal lowering of if-else code is more like: 

```
  ; Then region
  ; EXEC contains one subset of lanes

  ; Else region
  ; EXEC contains the other lanes

  ; Reconvergence to full mask
  s_mov_b64 exec, -1

  v_mov_b32 v0, s0
 ```
which is iiuc what you describe above. And here per-lane tracking is not needed. Hmm, but what if instead of restoring to a full mask (-1) it stashes EXEC to 2 scalar registers, and restores with ` s_mov_b64 exec, s[10:11]` ? This is an orthogonal question to per-lane tracking though (and a question for another day..).

**Comment by theSK2005** [comment]
> > Per lane EXEC state tracking would definitely be an upgrade, but I think that most of the EXEC defs that set it back to AllOnes would be setting EXEC to AllOnes directly [...]
> 
> makes sense. Per lane tracking would help with something like
> 
> ```
>   s_mov_b32 exec_lo, -1
>   s_mov_b32 exec_hi, -1
>   v_mov_b32 v0, s0
> ```
> 
> but how often compilers generate this kind of sequence is unknown. I guess the normal lowering of if-else code is more like:
> 
> ```
>   ; Then region
>   ; EXEC contains one subset of lanes
> 
>   ; Else region
>   ; EXEC contains the other lanes
> 
>   ; Reconvergence to full mask
>   s_mov_b64 exec, -1
> 
>   v_mov_b32 v0, s0
> ```
> 
> which is iiuc what you describe above. And here per-lane tracking is not needed. Hmm, but what if instead of restoring to a full mask (-1) it stashes EXEC to 2 scalar registers, and restores with ` s_mov_b64 exec, s[10:11]` ? This is an orthogonal question to per-lane tracking though (and a question for another day..).

Keeping track of other register values in order to determine EXEC state would basically amount to running the code. I think the most that would be reasonable to do (in the future, not this PR) is keeping track of each EXEC lane state based on operations with immediate values.

**Review by atgutier — COMMENTED** [review]
This is from an expert agent reviewer: 

## BLOCKER

[BLOCKER] `emulation/rocjitsu/tests/analysis/liveness_test.cpp:2894` — the PR does not compile after making EXEC analysis mandatory

Reasoning: `LivenessAnalysis` now requires a `const ExecMaskAnalysis&` as its second constructor argument, but five tests still pass `LivenessAnalysisOptions` in that position.

Evidence:

- Stale calls remain at lines 2894, 2927, 2958, 3009, and 3041.
- `analysis/liveness.h:138-140` defines the new constructor as:

```cpp
LivenessAnalysis(KernelBlockScope blocks, const ExecMaskAnalysis &exec,
                 LivenessAnalysisOptions options = {},
                 std::span<const ScopedCfgEdge> extra_edges = {});
```

- Current release, ASan/UBSan, TSan, and GCC-sanitizer jobs all fail while compiling this test translation unit with the same “no matching constructor” errors. Tests never run.

Suggested fix: construct the appropriate Wave32 EXEC analysis in each test—or a shared helper—and pass it before `options`:

```cpp
ExecMaskAnalysis exec{KernelBlockScope(scope), /*wave_size=*/32};
LivenessAnalysis liveness{KernelBlockScope(scope), exec, options};
```

Confidence: High.

## MAJOR

[MAJOR] `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/exec_state.cpp:231-235` — block index zero is not necessarily the kernel entry

Reasoning: The forward must-analysis starts interior blocks optimistically at `Full`. Soundness therefore depends on pinning every external hardware/firmware entry to `Unknown`. The implementation substitutes `i == 0` for explicit entry identity, but callers do not guarantee that the hardware entry is the first scope element.

`reachable_kernel_blocks()` sorts blocks by global `.text` position. A helper or branch target located before the descriptor entry can therefore become index zero. If the real entry has an in-scope backedge, it is neither predecessor-free nor index zero; the cycle can remain at the optimistic `Full` fixed point even though initial entry state is `Unknown`. An EXEC-masked VGPR definition can then be treated as a kill, allowing scratch allocation to clobber preserved lanes.

Evidence:

- `binary_translator.cpp:613-619` sorts reachable blocks by their global block index.
- `binary_translator.cpp:650-663` retains the actual entry separately as `scope.entry`.
- Lines 653-659 can seed both the normal hardware entry and a kernarg-preload firmware entry.
- `binary_translator.cpp:1752-1755` passes none of those entry identities into `ExecMaskAnalysis`.
- The regression at `tests/analysis/liveness_test.cpp:3714-3730` tests a backedge only when the entry happens to be `blocks[0]`.

Suggested fix: make `ExecMaskAnalysis` accept an explicit span of external entry blocks. Pass the descriptor entry and any firmware entry from DBT, and construct per-kernel entry sets in the instrumentor. Pin those entries to `Unknown` regardless of ordering or predecessor count. Add a regression with a lower-address helper and a backedge into the actual kernel entry.

Confidence: High.

[MAJOR] `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp:533-534` — instrumentor analysis omits call and return edges

Reasoning: `BasicBlock` intentionally stores validated calls separately from ordinary CFG successors. The instrumentor supplies no `ScopedCfgEdge` set, so neither EXEC nor liveness analysis sees caller-to-callee or return-to-continuation flow.

For example, a caller can establish full EXEC, invoke a callee that narrows EXEC, and then return. The instrumentor sees only the syntactic caller-to-continuation edge and leaves the continuation `Full`. A subsequent EXEC-masked VGPR definition becomes an unconditional kill even though inactive lanes preserve their old values. Probe scratch/clobber selection may then reuse that live VGPR.

This omission predates the new analysis for ordinary liveness, but the PR materially worsens it by allowing the incomplete graph to produce new `Full`-dependent vector kills.

Evidence:

- `code/basic_block.cpp:434-438` adds the syntactic fallthrough continuation.
- `code/basic_block.cpp:599-604` stores returning callees in `call_edges()` instead of ordinary successors.
- `binary_translator.cpp:795-821` constructs caller-to-callee and return-to-continuation analysis edges.
- `binary_translator.cpp:1751-1755` correctly passes the same edges to both EXEC and liveness.
- The instrumentor passes neither at lines 533-534.
- `tests/analysis/liveness_test.cpp:3671-3707` explicitly demonstrates that omitting the return edge leaves the continuation incorrectly `Full`, but no corresponding instrumentor regression exists.

Suggested fix: expose/share `scoped_call_liveness_edges()` and pass its result to both instrumentor analyses. Prefer per-kernel scopes rather than one object-wide CFG. Add an instrumentor regression where a callee narrows EXEC and the continuation’s VGPR definition must remain a non-kill.

Confidence: High.

## MINOR

[MINOR] `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/amdgpu_code_object.cpp:327-344` — unreadable descriptors do not trigger the safe Wave64 fallback

Reasoning: `kernel_wavefront_size()` returns Wave32 when every descriptor it successfully reads is Wave32, but it silently ignores known `.kd` entries it cannot locate or read. If an ignored descriptor belongs to a Wave64 kernel, the object-wide instrumentor analyzes that kernel as Wave32. An all-ones `exec_lo` write may then incorrectly establish full EXEC and enable unsound VGPR kills.

Evidence:

- `load_sections()` collects symbols ending in `.kd` at lines 239-251 without validating that each symbol maps to a readable descriptor.
- The new helper silently continues when the symbol address cannot be mapped or the descriptor exceeds the section bounds.
- `saw_kernel` reflects only successful reads, so one readable Wave32 descriptor plus one skipped descriptor returns 32.
- Such an object remains `is_valid()`, and instrumentation does not require every decoded `.text` region to have a successfully parsed descriptor.

Suggested fix: record any unreadable known descriptor and return 64—or fail instrumentation—when one is encountered. Per-kernel descriptor-derived analysis would remove this whole-object ambiguity.

Confidence: Medium; it requires a partially readable or unusual code object, but that input is currently accepted rather than rejected.

atgutier (commented):
This is from an expert reviewer agent:


## BLOCKER

[BLOCKER] `emulation/rocjitsu/tests/analysis/liveness_test.cpp:2894` — the PR does not compile after making EXEC analysis mandatory

Reasoning: `LivenessAnalysis` now requires a `const ExecMaskAnalysis&` as its second constructor argument, but five tests still pass `LivenessAnalysisOptions` in that position.

Evidence:

- Stale calls remain at lines 2894, 2927, 2958, 3009, and 3041.
- `analysis/liveness.h:138-140` defines the new constructor as:

```cpp
LivenessAnalysis(KernelBlockScope blocks, const ExecMaskAnalysis &exec,
                 LivenessAnalysisOptions options = {},
                 std::span<const ScopedCfgEdge> extra_edges = {});
```

- Current release, ASan/UBSan, TSan, and GCC-sanitizer jobs all fail while compiling this test translation unit with the same “no matching constructor” errors. Tests never run.

Suggested fix: construct the appropriate Wave32 EXEC analysis in each test—or a shared helper—and pass it before `options`:

```cpp
ExecMaskAnalysis exec{KernelBlockScope(scope), /*wave_size=*/32};
LivenessAnalysis liveness{KernelBlockScope(scope), exec, options};
```

Confidence: High.

## MAJOR

[MAJOR] `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/exec_state.cpp:231-235` — block index zero is not necessarily the kernel entry

Reasoning: The forward must-analysis starts interior blocks optimistically at `Full`. Soundness therefore depends on pinning every external hardware/firmware entry to `Unknown`. The implementation substitutes `i == 0` for explicit entry identity, but callers do not guarantee that the hardware entry is the first scope element.

`reachable_kernel_blocks()` sorts blocks by global `.text` position. A helper or branch target located before the descriptor entry can therefore become index zero. If the real entry has an in-scope backedge, it is neither predecessor-free nor index zero; the cycle can remain at the optimistic `Full` fixed point even though initial entry state is `Unknown`. An EXEC-masked VGPR definition can then be treated as a kill, allowing scratch allocation to clobber preserved lanes.

Evidence:

- `binary_translator.cpp:613-619` sorts reachable blocks by their global block index.
- `binary_translator.cpp:650-663` retains the actual entry separately as `scope.entry`.
- Lines 653-659 can seed both the normal hardware entry and a kernarg-preload firmware entry.
- `binary_translator.cpp:1752-1755` passes none of those entry identities into `ExecMaskAnalysis`.
- The regression at `tests/analysis/liveness_test.cpp:3714-3730` tests a backedge only when the entry happens to be `blocks[0]`.

Suggested fix: make `ExecMaskAnalysis` accept an explicit span of external entry blocks. Pass the descriptor entry and any firmware entry from DBT, and construct per-kernel entry sets in the instrumentor. Pin those entries to `Unknown` regardless of ordering or predecessor count. Add a regression with a lower-address helper and a backedge into the actual kernel entry.

Confidence: High.

[MAJOR] `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp:533-534` — instrumentor analysis omits call and return edges

Reasoning: `BasicBlock` intentionally stores validated calls separately from ordinary CFG successors. The instrumentor supplies no `ScopedCfgEdge` set, so neither EXEC nor liveness analysis sees caller-to-callee or return-to-continuation flow.

For example, a caller can establish full EXEC, invoke a callee that narrows EXEC, and then return. The instrumentor sees only the syntactic caller-to-continuation edge and leaves the continuation `Full`. A subsequent EXEC-masked VGPR definition becomes an unconditional kill even though inactive lanes preserve their old values. Probe scratch/clobber selection may then reuse that live VGPR.

This omission predates the new analysis for ordinary liveness, but the PR materially worsens it by allowing the incomplete graph to produce new `Full`-dependent vector kills.

Evidence:

- `code/basic_block.cpp:434-438` adds the syntactic fallthrough continuation.
- `code/basic_block.cpp:599-604` stores returning callees in `call_edges()` instead of ordinary successors.
- `binary_translator.cpp:795-821` constructs caller-to-callee and return-to-continuation analysis edges.
- `binary_translator.cpp:1751-1755` correctly passes the same edges to both EXEC and liveness.
- The instrumentor passes neither at lines 533-534.
- `tests/analysis/liveness_test.cpp:3671-3707` explicitly demonstrates that omitting the return edge leaves the continuation incorrectly `Full`, but no corresponding instrumentor regression exists.

Suggested fix: expose/share `scoped_call_liveness_edges()` and pass its result to both instrumentor analyses. Prefer per-kernel scopes rather than one object-wide CFG. Add an instrumentor regression where a callee narrows EXEC and the continuation’s VGPR definition must remain a non-kill.

Confidence: High.

## MINOR

[MINOR] `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/amdgpu_code_object.cpp:327-344` — unreadable descriptors do not trigger the safe Wave64 fallback

Reasoning: `kernel_wavefront_size()` returns Wave32 when every descriptor it successfully reads is Wave32, but it silently ignores known `.kd` entries it cannot locate or read. If an ignored descriptor belongs to a Wave64 kernel, the object-wide instrumentor analyzes that kernel as Wave32. An all-ones `exec_lo` write may then incorrectly establish full EXEC and enable unsound VGPR kills.

Evidence:

- `load_sections()` collects symbols ending in `.kd` at lines 239-251 without validating that each symbol maps to a readable descriptor.
- The new helper silently continues when the symbol address cannot be mapped or the descriptor exceeds the section bounds.
- `saw_kernel` reflects only successful reads, so one readable Wave32 descriptor plus one skipped descriptor returns 32.
- Such an object remains `is_valid()`, and instrumentation does not require every decoded `.text` region to have a successfully parsed descriptor.

Suggested fix: record any unreadable known descriptor and return 64—or fail instrumentation—when one is encountered. Per-kernel descriptor-derived analysis would remove this whole-object ambiguity.

Confidence: Medium; it requires a partially readable or unusual code object, but that input is currently accepted rather than rejected.

theSK2005:
@atgutier I just fixed the liveness_test.cpp compilation issue, that came up as I was rebasing

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
```diff
@@ -529,7 +532,19 @@ InstrumentedCodeObjectDebug Instrumentor::patch_with_debug_summaries() {
...
     liveness_scope.push_back(block.get());
-  const LivenessAnalysis liveness{KernelBlockScope(liveness_scope)};
+  const std::span<const uint8_t> original_text = patcher.text_bytes();
+  const auto liveness_edges =
+      scoped_call_liveness_edges(KernelBlockScope(liveness_scope), original_text);
+  const auto entry_offsets = obj_.kernel_entry_text_offsets();
+  const std::unordered_set<uint64_t> entry_offset_set(entry_offsets.begin(), entry_offsets.end());
+  std::vector<const BasicBlock *> entry_blocks;
+  for (BasicBlock *block : liveness_scope) {
+    if (block != nullptr && entry_offset_set.contains(block->start_offset()))
```
kuhar, line 561: We should split the instrumentor CFG at every descriptor entry before running EXEC and liveness analysis. `ensure_blocks_built()` calls `BasicBlock::build()` without the offsets, then lines 538-543 keep only entries that already match a block start. A later entry inside a fallthrough block is therefore omitted, and a preceding `s_mov exec, -1` can make its first instruction appear `Full` even though hardware may enter there with unknown EXEC. Please obtain the entry offsets before block construction, pass them as `extra_leaders`, fail closed if a readable entry still has no exact block, and add an instrumentor regression for this layout.

kuhar (commented):
A few comments -- below and inline in https://github.com/ROCm/rocm-systems/pull/7470#discussion_r3687278205 and https://github.com/ROCm/rocm-systems/pull/7470#discussion_r3687278133

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/exec_state.cpp
kuhar: We should pass every hardware entry in this kernel scope to `ExecMaskAnalysis`, not only `scope.entry`. `kernel_translation_scopes()` already seeds reachability from both `entry_text_offset` and `kernarg_preload_firmware_entry_text_offset`, but the production call at `binary_translator.cpp:1597` pins only the primary block. If the ordinary path reaches the firmware entry after establishing full EXEC, direct firmware entry can inherit that `Full` state and license an unsafe vector kill. Please carry both entry blocks in the scope and add a two-entry regression where the firmware entry remains `Unknown`.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/liveness.cpp
kuhar: It would be useful to keep the requested real-decode liveness regression now that the underlying modeling has changed. `GeneratedInstDefUse.WritelaneReadsDestinationCdna4` and `GeneratedInstDefUse.WritelaneReadsDestinationGfx1250` verify that `v_writelane_b32` exposes its destination as both use and def, and the focused analysis suite passes, but the end-to-end `ExecFlagsRealDecode` tests still cover only full-width `v_mov` writers. Please add a `Full` EXEC plus `v_writelane_b32` case that checks the destination remains live before the lane write.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/liveness.cpp
```diff
@@ -154,11 +150,16 @@ std::vector<const BasicBlock *> reverse_post_order(KernelBlockScope blocks) {
...
+LivenessAnalysis::LivenessAnalysis(KernelBlockScope blocks, const ExecMaskAnalysis &exec,
+                                   LivenessAnalysisOptions options,
                                    std::span<const ScopedCfgEdge> extra_edges) {
   min_free_vgpr_ = options.min_free_vgpr;
   max_free_vgpr_ =
       static_cast<uint16_t>(std::min<size_t>(options.max_free_vgpr, REGISTER_SET_MAX_VGPRS));
+  // Capture the EXEC-state analysis now; the backward dataflow is deferred to the
+  // first query (ensure_analyzed), by which point the caller's exec temporary is
+  // gone, so hold a copy for the deferred analyze() to consult for kills.
+  exec_ = std::make_unique<ExecMaskAnalysis>(exec);
```
kuhar: We should avoid deep-copying `ExecMaskAnalysis` here. Its `before_` map is keyed per instruction, and both production callers create the analysis only to hand it to `LivenessAnalysis`, so an rvalue or by-value path that moves into `exec_` would remove one full map copy per kernel scope while preserving the deferred liveness design.

kuhar (commented):
A few comments (3 inline replies above)

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
kuhar: We should extend this fix to the kernarg-preload firmware entry. `kernel_entry_text_offsets()` reads only `kernel_code_entry_byte_offset`, while CDNA3/CDNA4 descriptors with a nonzero `KERNARG_PRELOAD_SPEC_LENGTH` also permit direct entry at +256. The existing Instrumentor entry tests pass, but they cover only the primary descriptor entry. Please enumerate both hardware entries and add a preload-descriptor case where the ordinary path establishes `Full` before +256; the second entry must remain `Unknown`.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/exec_state.cpp
kuhar: It would be useful to test the production boundary now that both entries are passed. `ExecMaskAnalysis.SecondHardwareEntryReachedAfterFullIsPinnedUnknown` hand-builds `entry_blocks`, so it also passes on the parent revision where `BinaryTranslator` supplied only `scope.entry`. Please add a `translate()`-level preload fixture that establishes `Full` on the ordinary path and verifies the firmware-entry vector def remains live.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/liveness.cpp
kuhar: We should make the new regression fail rather than skip when its metadata precondition is absent. Both `Cdna4WritelaneDestStaysLiveUnderFullExec` and `Cdna4SMovExecAllOnesPromotesVgprDefToKill` call `GTEST_SKIP()` when `s_mov` lacks `RESULT_COPY`, so a generator regression can disable both Full-EXEC checks while CI stays green. An `ASSERT_TRUE` on the flag would keep the contract load-bearing.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
accauble: @theSK2005 it seems like you chose to defer this?

accauble (approved):
LGTM - I would say that tracking the must be all ones is a good follow-up and not required for this PR.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
theSK2005: yes, this is a defer, where should I add that info?

kuhar (changes requested):
This seems to be ~10% slower on rccl hotswap translation -- we need to bring the overhead down before this can land.

accauble:
> This seems to be ~10% slower on rccl hotswap translation -- we need to bring the overhead down before this can land.

Does the rccl translation need the EXEC state tracking for correctness?

kuhar:
I found one meaningful optimization: https://github.com/ROCm/rocm-systems/commit/20a238e3dee02b2c3b6d9d191cf55e3de365cc6a

> > This seems to be ~10% slower on rccl hotswap translation -- we need to bring the overhead down before this can land.
> 
> Does the rccl translation need the EXEC state tracking for correctness?

If we can, we should have a very quick check that answers this and decides whether to skip the analysis or not.

kuhar (commented):
Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/exec_state.h:77`

Could we make `before_` sparse and store only instructions whose incoming state is `Full`? `before()` already returns `Unknown` for missing entries, and liveness only distinguishes the `Full` case, so an `unordered_set<const Instruction *> full_before_` preserves the query contract while avoiding a map node and state value for every instruction.

I prototyped this as local commit `20a238e3de` (5 insertions, 4 deletions). On the exact 224,676,512-byte RCCL object, three CPU-32-pinned runs reduced median CPU time from 184.46 s to 171.64 s (-6.95%) and wall time from 184.53 s to 171.72 s (-6.94%). Median peak RSS was effectively flat: 15,977,508 KiB to 15,977,872 KiB (+0.002%). A matched `perf stat` run showed task-clock -7.51%, cycles -7.93%, instructions -1.88%, branches -1.95%, branch misses -2.96%, and about 71k fewer page faults. All runs produced the same 224,676,512-byte output with SHA-256 `7c843d92dddbe69b1944a84eabd3d8d6a64304c6a84f1a03af638a16f7d15d9b`; all 76 focused liveness, EXEC, and gfx1250 translation tests passed.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
kuhar: I think the remaining test should exercise the Instrumentor path itself. The new preload test at instrumentor_test.cpp:1087 calls kernel_entry_text_offsets(), BasicBlock::build(), and ExecMaskAnalysis directly, so it stays green if ensure_blocks_built() stops forwarding the firmware offset as a leader or patch_with_debug_summaries() stops pinning that block as an entry. Could we add an end-to-end Instrumentor regression for the +256 entry and keep the lower-level assertions as supporting coverage?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/exec_state.cpp
kuhar: We should retarget the new translation test now that DBT deliberately passes no EXEC analysis. `BinaryTranslatorE2E.Cdna4ToCdna3KernargPreloadFirmwareEntryKeepsMaskedDefLive` still says it proves the +256 entry is pinned `Unknown`, but `binary_translator.cpp:2160` now makes every EXEC-masked vector def `Unknown`; changing `KERNARG_PRELOAD_SPEC_LENGTH` from 1 to 0 still leaves the test passing. Please rename this fixture to cover the conservative no-EXEC contract and add a direct `LivenessAnalysis(..., /*exec=*/nullptr)` unit test. The separate Instrumentor thread can retain the production entry-pinning regression.

kuhar (approved):

PR #7472, opened 2026-06-18
fix(rocprofv3): Output flag (-o) bug fix with multiple processes

itrowbri:
## Motivation

Setting the `-o` flag would cause multiple process traces to be written to the same file. This bug was discovered when attempting to use it with vllm, which uses multiple processes

## Technical Details

Added a new environment variable that stores the parent PID. If a child process being traced is found to have a different PID, the PID is inserted into the file name so as pipe the output to a different file. 

## JIRA ID

JIRA ID: ROCM-25327

## Test Plan

Added a fork-test that forks and runs a kernel for child and parent process. Test traces fork-test with `-o` flag set. Two process db files should be generated anyways, and the ran kernel should be in both traces.

## Test Result

New test should pass

## Submission Checklist


projects/rocprofiler-sdk/source/lib/output/output_stream.cpp
```diff
@@ -57,6 +60,20 @@ get_output_filename(const output_config& cfg, std::string_view fname, std::strin
...
+    // Descendants of the root process append their PID so they don't overwrite
+    // each other's output; the root keeps the user-specified name. Recomputed
+    // per call (not a static) so a process that forks later still gets its PID.
+    // Token check uses raw cfg.output_file since format_path() already expanded
+    // any %pid%/%p/{pid} in output_prefix.
+    auto _root_pid      = common::get_env_optional("ROCPROF_OUTPUT_ROOT_PID");
+    bool _is_descendant = _root_pid.has_value() && *_root_pid != std::to_string(getpid());
+    if(_is_descendant && cfg.output_file.find("%pid%") == std::string::npos &&
+       cfg.output_file.find("{pid}") == std::string::npos &&
+       cfg.output_file.find("%p") == std::string::npos)
```
MythreyaK: Please correct me if I'm wrong, but I think this check needs to be narrowed a little more because `-o my_%ppid%_output` will fail this check and child procs will use the same filename.

MythreyaK (changes requested):
Small change, but looks good to me otherwise. 

One thing I think is worth documenting is that this will work reliably iff the setenv happens before a fork and the output filename is lazily constructed (so after a potential fork). 

If in the future we update the tool to generate filenames eagerly, then this approach will probably fail (and the tests you added should catch that).

MythreyaK:
I looked at tmpfile buffer names and they do use pid-ppid format but are generated eagerly, so they will have the same filenames before/after a fork. But this shouldn't be a problem because using HIP/HSA before a fork without an exec is undefined behavior / unsupported anyway. Might be worth mentioning somewhere. 

Small nit, probably not an issue: the test has fork, no fork+exec, which should be okay I think. Up to you.

MythreyaK (commented):
Looks good. There are some rare edge-cases we can probably address in a follow-up (pls do create a ticket for it). 

Please do get an additional first review since there's a behavior change we collaborated on.

projects/rocprofiler-sdk/source/lib/output/output_config.cpp
```diff
@@ -66,7 +66,14 @@ output_config::parse_env()
     annotate_args  = common::get_env("ROCPROF_ANNOTATE_ARGS", false);
     annotate_kfd   = common::get_env("ROCPROF_ANNOTATE_KFD", false);
     annotate_pmc   = common::get_env("ROCPROF_ANNOTATE_PMC", false);
-    auto to_upper  = [](std::string val) {
+
+    // Record the root process's PID (inherited across fork/fork+exec, override=0 so
+    // descendants don't clobber it). get_output_filename() uses it to PID-suffix a
+    // descendant's output so it doesn't overwrite the root's. Relies on the final
+    // filename being built lazily, after any fork.
+    common::set_env("ROCPROF_OUTPUT_ROOT_PID", std::to_string(getpid()), 0);
```
MythreyaK: teeny-tiny nit: Should we make it `ROCPROF_INTERNAL_OUTPUT_ROOT_PID`? Also used in 2 places, so a constexpr value somewhere? It's good as-is; I'll leave it up to you/second reviewer's discretion. 

But we should document it somewhere that this is an internal env var. I'm also wondering if we should explicitly set this to `0` in rocprofv3 script.

MythreyaK (approved):

jrmadsen (commented):
I don't get the purpose of this PR whatsoever. As long as we don't cache the PID or cache the substitution before the fork, `getpid()` in the child process will return the child process id

jrmadsen (changes requested):
Need my question answered before this gets merged

jrmadsen:
User was directly encoding PID in the string instead of using `%pid%`.

itrowbri:
@jrmadsen The main point of the PR is the issue with the -o flag with multiple processes. Running `rocprofv3 -o output ... -- <process>` where <process> creates multiple processes (as often occurs with vLLM workloads), has all processes write to the same file ( `output_results.db` in the example above) regardless of how many processes are spawned in the workload. In contrast, running without the -o flag will have an output file spawn for each process with the pid (e.g. `<pid>_results.db`). If the description of the PR was not clear, I can update it. Alternatively, if you want to handle it another way, I can do that too.

itrowbri:
The PID matching behavior/files can be changed or dropped. That was to avoid "re-adding" the pid where the pid was already added as part of the output file name and not the main point of the PR

PR #7477, opened 2026-06-18
[rocprofiler-sdk] Fix PMC after-kernel packet ordering

mcao59:
## Motivation

Fix a potential rocprofv3 hang during PMC counter collection.

When profiling with counters such as SQ_WAVES, rocprofiler injects AQL packets around kernel dispatches to start, stop, and read counter collection. The after-kernel packet order should finalize counter collection before reading the results.

The previous order emitted the read packet before the stop packet, which could leave the queue waiting on a counter read before the collection had been stopped/finalized.

## Technical Details

Update `CounterAQLPacket::populate_after()` so the after-kernel PMC packet sequence is:

`stop -> read`

instead of:

`read -> stop`

This makes the full counter collection sequence:

`start -> kernel -> stop -> read`

which matches the expected start/stop/read flow for PMC collection.

Also add a regression test in `aql_test.cpp` to verify that `CounterAQLPacket::populate_after()` emits the stop packet before the read packet. The test marks the stop/read packets with unique completion-signal handles and checks the generated `after_krn_pkt` ordering without enqueueing packets.

## JIRA ID

ROCM-21782

## Test Plan

- run AQL packet tests.
- Manually validate the original rocprofv3 PMC hang reproducer with `--pmc SQ_WAVES`.

## Test Result

- `counter_after_packets_stop_before_read` passes.
- Existing AQL packet tests pass.
- Manual rocprofv3 PMC validation no longer reproduces the hang

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/aql_packet.hpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/aql_packet.hpp
@@ -167,8 +167,8 @@ public:
     void populate_after() override
     {
         if(empty) return;
-        after_krn_pkt.push_back(packets.read_packet);
         after_krn_pkt.push_back(packets.stop_packet);
+        after_krn_pkt.push_back(packets.read_packet);
     };
 
     aqlprofile_pmc_aql_packets_t packets{};
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/aql/tests/aql_test.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/aql/tests/aql_test.cpp
@@ -164,6 +164,52 @@ TEST(aql_profile, packet_generation_single)
     hsa_shut_down();
 }
 
+TEST(aql_profile, counter_after_packets_stop_before_read)
+{
+    ASSERT_EQ(hsa_init(), HSA_STATUS_SUCCESS);
+
+    struct HsaShutdownGuard
+    {
+        ~HsaShutdownGuard() { hsa_shut_down(); }
+    } cleanup;
+
+    rocprofiler::test_init();
+
+    auto agents = rocprofiler::hsa::get_queue_controller()->get_supported_agents();
+    ASSERT_GT(agents.size(), 0);
+
+    size_t tested_agents = 0;
+
+    for(const auto& [_, agent] : agents)
+    {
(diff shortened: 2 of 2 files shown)


projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/aql_packet.hpp
```diff
@@ -167,8 +167,8 @@ class CounterAQLPacket : public AQLPacket
     void populate_after() override
     {
         if(empty) return;
-        after_krn_pkt.push_back(packets.read_packet);
```
MythreyaK, line 170: The original code seems intentional, not sure?

github isn't [deep-linking to the relevant diff](https://github.com/ROCm/rocm-systems/commit/ebad2abe3c730ae20203f2a636fc337c35b77919#diff-1d87104cbb35ac16f678c595ff194910c4b0d5bd54efcd0035b0d858a2b7ef8aR130-R132) so adding a screenshot as well 

Image

mcao59: Hmm, from this page https://rocm.docs.amd.com/projects/aqlprofile/en/docs-7.2.1/examples/pmc-workflow.html, it  does put all stop packets before read packets in the code snippets, which also logically makes sense to me:
```
Generate command packets
hsa_ext_amd_aql_pm4_packet_t start_pkt, stop_pkt, read_pkt;
hsa_ven_amd_aqlprofile_start(&profile, &start_pkt);
hsa_ven_amd_aqlprofile_stop(&profile, &stop_pkt);
hsa_ven_amd_aqlprofile_read(&profile, &read_pkt);
Submit packets and run the kernel
// Pseudocode: inject packets into HSA queue
queue->Submit(&start_pkt);
queue->Submit(&kernel_pkt);
queue->Submit(&stop_pkt);
queue->Submit(&read_pkt);
```
@bwelton @ApoKalipse-V  Do you know if the original "read before stop" pattern is intentional?

ApoKalipse-V: Because of random failures to read counters with perfmon clock gating disabled

ApoKalipse-V (changes requested):
I dont agree with this change, and there is something wrong with the premise.
The read/stop order should not cause hangs.

PR #7487, opened 2026-06-19
feat: [rocshmem] add wave-level broadcast

kesag:
## Motivation

Develop `nvshmemx_##TYPE##_broadcast_warp()` and `nvshmemx_broadcastmem_warp()`to provide wave-level broadcast capabilities in rocSHMEM, matching NVSHMEM's API and semantics.

## Technical Details

Add wave-level broadcast implementation 
- `rocshmem_ctx_##TYPE##_broadcast_wave()` 
- `rocshmem_ctx_broadcastmem_wave()` 

Add broadcast mem wg
- `rocshmem_ctx_broadcastmem_wg()`

## JIRA ID

JIRA ID: AIROCSHMEM-432

## Test Plan

- [x] Use functional test to test the operation

## Test Result

- [x] Test matches broadcast_wg verification result

## Submission Checklist

### DIFF
--- a/projects/rocshmem/scripts/functional_tests/driver.sh
+++ b/projects/rocshmem/scripts/functional_tests/driver.sh
@@ -753,9 +753,9 @@ TestColl() {
   ExecTest  "teamreducescatter" 4      1            64        32768
   ExecTest  "teamreducescatter" 8      1            64        32768
 
-  if [[ $TEST != ro* ]]; then #AIROCSHMEM-432: wave tests not supported on RO
+  if [[ $TEST != ro* ]]; then #AIROCSHMEM-409: wave tests not supported on RO
     ExecTest  "broadcast_wave"   2       1            64        32768
-  else echo "Skip:   *_wave (AIROCSHMEM-408: wave tests not supported on RO)"; fi
+  else echo "Skip:   *_wave (AIROCSHMEM-409: wave tests not supported on RO)"; fi
 }
 
 TestOnStream() {
--- a/projects/rocshmem/scripts/functional_tests/driver.sh
+++ b/projects/rocshmem/scripts/functional_tests/driver.sh
@@ -164,6 +164,7 @@ declare -A TEST_NUMBERS=(
   ["host_wait_until_any_status"]="147"
   ["host_wait_until_some_status"]="148"
   ["teamreducescatter"]="149"
+  ["broadcast_wave"]="150"
 )
 
 # Detect which runtime to use
@@ -751,6 +752,7 @@ TestColl() {
   ExecTest  "teamreducescatter" 2      1            64        32768
   ExecTest  "teamreducescatter" 4      1            64        32768
   ExecTest  "teamreducescatter" 8      1            64        32768
+  ExecTest  "broadcast_wave"   2       1            64        32768
 }
 
 TestOnStream() {
--- a/projects/rocshmem/scripts/functional_tests/driver.sh
+++ b/projects/rocshmem/scripts/functional_tests/driver.sh
@@ -752,7 +752,10 @@ TestColl() {
   ExecTest  "teamreducescatter" 2      1            64        32768
   ExecTest  "teamreducescatter" 4      1            64        32768
   ExecTest  "teamreducescatter" 8      1            64        32768
-  ExecTest  "broadcast_wave"   2       1            64        32768
+
+  if [[ $TEST != ro* ]]; then #AIROCSHMEM-432: wave tests not supported on RO
+    ExecTest  "broadcast_wave"   2       1            64        32768
+  else echo "Skip:   *_wave (AIROCSHMEM-408: wave tests not supported on RO)"; fi
 }
 
 TestOnStream() {
--- a/projects/rocshmem/src/gda/context_gda_device.hpp
+++ b/projects/rocshmem/src/gda/context_gda_device.hpp
@@ -289,6 +289,17 @@ class GDAContext : public Context {
   __device__ void internal_get_broadcast_wave(T *dst, const T *src, int nelems,
       int pe_root, ActiveWFInfo &wf_info);  // NOLINT(runtime/int)
 
+  __device__ void internal_broadcastmem_wave(void *dst, const void *src,
+    int nelems, int pe_root, int pe_start, int stride, int pe_size,
+    long *p_sync);
+
+  __device__ void internal_get_broadcastmem_wave(void *dst, const void *src,
+    int nelems, int pe_root, ActiveWFInfo &wf_info);
+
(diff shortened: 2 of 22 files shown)


edgargabriel (approved):
LGTM, I think there are now conflicts however with the test numbers that need to be resolved first, but otherwise I think this is good to go.

projects/rocshmem/tests/functional_tests/broadcast_wave_tester.cpp
```diff
@@ -0,0 +1,217 @@
...
+                    std::is_same<T1, signed char>::value ||
+                    std::is_same<T1, unsigned char>::value) {
+        source_buf[idx] = static_cast<T1>('a' + n_pes + wg_id);
+        dest_buf[idx] = static_cast<T1>('a' + wg_id);
+      }
+      else if constexpr (std::is_floating_point<T1>::value) {
+        source_buf[idx] = static_cast<T1>(3.14 + n_pes + wg_id);
+        dest_buf[idx] = static_cast<T1>(3.14 + wg_id);
+      }
+      else if constexpr (std::is_integral<T1>::value) {
```
abouteiller: can't we still add 'a' when its an integral and collapse the cases?

projects/rocshmem/tests/functional_tests/broadcast_wave_tester.cpp
```diff
@@ -0,0 +1,217 @@
...
+template <typename T1>
+__global__ void BroadcastWaveTestKernel(int loop, int skip, long long int *start_time,
+                                  long long int *end_time, T1 *source_buf,
+                                  T1 *dest_buf, int size,
+                                  ShmemContextType ctx_type,
+                                  rocshmem_team_t *teams) {
+  __shared__ rocshmem_ctx_t ctx;
+  int wg_id = get_flat_grid_id();
+
+  rocshmem_wg_team_create_ctx(teams[wg_id], ctx_type, &ctx);
```
abouteiller: This is a bug: we need 1 team per wf.

abouteiller (changes requested):

projects/rocshmem/scripts/functional_tests/driver.sh
```diff
@@ -720,6 +721,10 @@ TestColl() {
   ExecTest  "fcollect"         2       1            64        32768
 
   ExecTest  "teamreduction"    2       1            64        32768
+
+  if [[ $TEST != ro* ]]; then #AIROCSHMEM-432: wave tests not supported on RO
```
abouteiller: numbers mismatch here 432 vs 408

projects/rocshmem/scripts/functional_tests/driver.sh
```diff
@@ -720,6 +721,10 @@ TestColl() {
...
+    ExecTest  "broadcast_wave"   2       1            64        32768
```
abouteiller, line 757: Same change must be applied to the ctest

projects/rocshmem/src/gda/context_gda_device.hpp
```diff
@@ -145,9 +145,19 @@ class GDAContext : public Context {
   __device__ int reduce(rocshmem_team_t team, T *dest, const T *source, int nreduce);
 
   template <typename T>
-  __device__ void broadcast(rocshmem_team_t team, T *dest, const T *source,
```
abouteiller, line 152: It looks like simple broadcast (non wave/wg) has been removed but it should still remain

projects/rocshmem/src/reverse_offload/context_ro_device.cpp
```diff
@@ -573,6 +573,15 @@ __device__ uint64_t broadcast(bool lowest_active, uint64_t value) {
   return broadcast_lds(lowest_active, value);
 }
 
+__device__ int ROContext::broadcastmem_wave([[maybe_unused]] rocshmem_team_t team,
+                                        [[maybe_unused]] void *dest, 
+                                        [[maybe_unused]] const void* source, 
+                                        [[maybe_unused]] int nelement, 
+                                        [[maybe_unused]] int PE_root) {
+  LOGD_WARN("Broadcastmem Wave API not implemented for reverse offload backend");
```
abouteiller, line 581: Isn't the implementation identical to the WG variant for RO?

projects/rocshmem/tests/functional_tests/broadcast_wave_tester.cpp
```diff
@@ -0,0 +1,215 @@
...
+    if (i == skip && t_id % wf_size == 0) {
+      start_time[flat_wf_id] = wall_clock64();
+    }
+    wave_broadcast<T1>(ctx, teams[flat_wf_id], dest_buf, source_buf, size, 0);
+  }
+
+  __syncthreads();
+
+  if (t_id % wf_size == 0) {
+    end_time[flat_wf_id] = wall_clock64();
```
abouteiller, line 86: Need to double check if this matches other wf primitive testers (in many tests we always use a timer per wg)

projects/rocshmem/tests/functional_tests/broadcast_wave_tester.cpp
```diff
@@ -0,0 +1,215 @@
...
+  dest_buf = (T1 *)alloc_test_buffer(buff_size);
+
+  char* value{nullptr};
+  if ((value = getenv("ROCSHMEM_MAX_NUM_TEAMS"))) {
+    num_teams = atoi(value);
+  }
+
+  if (num_teams < num_warps * args.num_wgs){
+    printf("not enough teams for each wavefront, try increasing ROCSHMEM_MAX_NUM_TEAMS\n");
+    abort();
```
abouteiller: lets use exit instead of abort we don't need a stack trace here. 
(if this matches existing pattern, we can keep as is, defer this change to a followup targeted pr)

projects/rocshmem/src/gda/context_gda_device.hpp
kesag: From what I can tell, the "simple" broadcast was always wg scoped, updated the name to fit with wg/wave naming

projects/rocshmem/tests/functional_tests/broadcast_wave_tester.cpp
kesag: From what I see, the wavefront primitives use the wf_id for the timers, so 1 timer per wavefront.

projects/rocshmem/src/reverse_offload/context_ro_device.cpp
abouteiller: deferred

abouteiller (approved):

PR #7454, opened 2026-06-18
hip-tests: Disable execution ctx test due to timeout in CI

manocharahul:
## Motivation

- Disable Unit_hipExecutionCtxRecordWaitEvent_Sanity test due to timeout in CI

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #7490, opened 2026-06-19
AIRUNTIME-2354 - Defer queue release to prevent deadlock for cu mask …

manocharahul:
…queues

## Motivation
1. Self deadlock on async completion handler thread caused by recursive callbacks (HsaAmdSignalHandler in hip and DynamicQueueEventsHandler in rocr ). 
2. HsaAmdSignalHandler callback triggers a queue destroy call which calls AqlQueue destructor in rocr and deadlocks waiting for DynamicQueueEventsHandler which can't run since original callback HsaAmdSignalHandler has not finished.
3. Issue only seen when CU mask queues are released in the VirtualGPU destructor path due to async completion handler thread being the last owner.

## Technical Details
1. This change defers release of queues with CU mask when async completion handler is the last owner by using a thead_local var set in HsaAmdSignalHandler.
2.  The actual drain is done in ~Device or in acquireQueue path on main thread. 
3. a batch threshold of 8 is used to do drain in acquireQueue to avoid accumulation of multiple queues in deferred list.
4. Coop queues stay inline (recycled via non-blocking GWSRelease, so safe and
  must not be deferred).
## Submission Checklist

### DIFF
--- a/projects/clr/hipamd/src/hip_executionctx.cpp
+++ b/projects/clr/hipamd/src/hip_executionctx.cpp
@@ -237,7 +237,7 @@ hipError_t ExecutionCtx::recordEvent(hipEvent_t event) {
 
   // For single stream, use the existing addMarker path directly.
   if (snapshot.size() == 1) {
-    hipError_t err = e->addMarker(snapshot[0], nullptr, true);
+    hipError_t err = e->addMarker(snapshot[0], nullptr, !hip::Event::kBatchFlush);
     snapshot[0]->release();
     return err;
   }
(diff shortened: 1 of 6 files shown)


jaydeeppatel1111:
Hi @manocharahul, This PR needs to enable back the the tests that it fix .. Right? Revert of https://github.com/ROCm/rocm-systems/pull/7454

gandryey (approved):

amd-hsivasun:
Bypass and Merging

PR #7492, opened 2026-06-19
rocr: Defer re-entrant frees to prevent ASAN deadlock

ApurvMishra-amd:
## Motivation

Fix a self-deadlock in `MemoryRegion::Free` when running under AddressSanitizer with quarantine enabled. When ASAN's quarantine is full during an allocate operation, it evicts an older GPU block, triggering a physical free that re-enters `MemoryRegion::Free` on the same thread. Since `agent_memory_lock_` is non-recursive, this causes a deadlock (futex_wait hang).

## Technical Details

Detect re-entrancy via a per-thread lock depth counter. If a Free() arrives while the thread already holds `agent_memory_lock_`, queue the physical free and defer it. Drain the queue once the outermost critical section exits (depth reaches zero). This preserves the "serial/exclusive memory ops" invariant by re-acquiring the lock for each deferred free, preventing allocator corruption while avoiding re-entrancy deadlock.

Fixes timeout in hip-tests _Unit_Warp_Shfl_Down_Positive_Basic_ when run with `ASAN_OPTIONS="quarantine_size_mb=600" and HSA_XNACK=1`.

Changes:
- Add `ScopedAgentMemoryLock` RAII guard to track lock depth
- Add thread-local depth counter and deferred-free queue
- Modify `Allocate()`, `Free()`, `IPCFragmentExport()` to use the guard
- `Free()` checks for re-entrancy first; if detected, queues instead of locking

rocr-runtime only; always-on (zero overhead when not under ASAN re-entrancy).

## JIRA ID

ROCM-26090

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/inc/amd_memory_region.h
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/inc/amd_memory_region.h
@@ -201,6 +201,28 @@ private:
   // Operational body for Free.  Recursive.
   hsa_status_t FreeImpl(void* address, size_t size) const;
 
+#if defined(SANITIZER_AMDGPU)
+  // ASAN re-entrancy guard: avoid deadlock when ASAN quarantine evicts a GPU
+  // block while this thread already holds an agent_memory_lock_.  Defer the
+  // re-entrant free until the outermost critical section exits.
+  class ScopedAgentMemoryLock {
+    public:
+      explicit ScopedAgentMemoryLock(std::mutex& mutex);
+      ~ScopedAgentMemoryLock();
+
+    private:
+      std::mutex& mutex_;
+      DISALLOW_COPY_AND_ASSIGN(ScopedAgentMemoryLock);
+  };
+
+  // Queue free if an agent_memory_lock_ is already held by this thread; returns
+  // true if the free was deferred, false if the caller should free normally.
+  bool DeferFreeIfLockHeld(void* address, size_t size) const;
+
+  // Drain queued frees after the outermost lock release.
+  static void DrainDeferredFrees();
+#endif  // SANITIZER_AMDGPU
+
   class BlockAllocator {
    private:
     MemoryRegion& region_;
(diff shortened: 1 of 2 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/inc/amd_memory_region.h
```diff
@@ -201,6 +201,24 @@ class MemoryRegion : public core::MemoryRegion {
   // Operational body for Free.  Recursive.
   hsa_status_t FreeImpl(void* address, size_t size) const;
 
+  // ASAN re-entrancy guard: avoid deadlock when ASAN quarantine evicts a GPU
```
b-sumner, line 205: Guarding this and the other changes with `#if __has_feature(address_sanitizer)` would limit the exposure of this change if desired.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_memory_region.cpp
```diff
@@ -133,8 +135,72 @@ MemoryRegion::MemoryRegion(bool fine_grain, bool kernarg, bool full_profile,
...
+  // queue terminates because the quarantine evicts a bounded number of blocks.
+  while (!tls_deferred_frees.empty()) {
+    const DeferredFree item = tls_deferred_frees.back();
+    tls_deferred_frees.pop_back();
+
+    std::lock_guard<std::mutex> guard(item.region->owner()->agent_memory_lock_);
+    ++tls_agent_lock_depth;
+    MAKE_SCOPE_GUARD([&]() { --tls_agent_lock_depth; });
+
+    try {
```
b-sumner, line 185: I'm not used to seeing try-catch in HSA runtime code.  Is this acceptable now?

cfreeamd: You might be thinking of the thunk, C code.  Look at hsa.cpp for example. Though, it's not used as much as it could be perhaps.

b-sumner (approved):
This looks good to me.

dayatsin-amd (approved):

dayatsin-amd:
Ideally, we should try to use recursive mutexes instead but this may expose some issues with the code where we assume some of the allocation tables cannot be modified while the agent_mutex is held. 
Created internal ticket: AILIROCR-523 as a future task to try to use recursive mutexes instead

PR #7499, opened 2026-06-19
[rocprofiler-systems] Skip shmem-pingpong tests if example binary is missing

mradosav-amd:
## Motivation

The `test_pingpong` test in `test_shmem.py` was failing in CI environments where the `shmem_pingpong` example binary had not been built or installed. Instead of a clean skip, the test would crash with an opaque `FileNotFoundError`, making it hard to distinguish a missing-binary situation from a genuine test regression.

## Technical Details

Added a pre-flight check at the start of `test_pingpong` that calls `rocprof_config.get_target_executable("shmem_pingpong")`. If the binary is absent, the test is skipped with the message `"shmem_pingpong binary not found"` rather than failing. The `rocprof_config` fixture was added to the method signature to enable this lookup. No test logic is changed for environments where the binary is present.

Changed file: `projects/rocprofiler-systems/tests/pytest/test_shmem.py`

## JIRA ID

AIPROFSYST-557

## Test Plan

- [x] Verified test is skipped cleanly when `shmem_pingpong` binary is absent
- [x] Verified test still runs and passes when `shmem_pingpong` binary is present
- [x] Confirmed no other `test_shmem.py` test parametrizations are affected

## Test Result

Tests skip gracefully with `"shmem_pingpong binary not found"` when the binary is missing, rather than crashing.

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-systems/tests/pytest/test_shmem.py
+++ b/projects/rocprofiler-systems/tests/pytest/test_shmem.py
@@ -93,10 +93,15 @@ class TestShmem(RocprofsysTest):
             pytest.param("sys_run", marks=pytest.mark.rocpd("shmem_env")),
         ],
     )
-    def test_pingpong(self, mode, shmem_env, shmem_validated, shmem_rules):
+    def test_pingpong(self, mode, shmem_env, shmem_validated, shmem_rules, rocprof_config):
         valid, reason = shmem_validated
         if not valid:
             pytest.skip(f"{reason}")
+
+        try:
+            rocprof_config.get_target_executable("shmem_pingpong")
+        except FileNotFoundError:
+            pytest.skip("shmem_pingpong binary not found")
 
         result = self.run_test(
             mode,


projects/rocprofiler-systems/tests/pytest/test_shmem.py
kcossett-amd, line 1: `run_test` should have caught this. The target is propagated to the runner and found using `get_target_executable()` which can throw a FNF:
 -  https://github.com/ROCm/rocm-systems/blob/develop/projects/rocprofiler-systems/tests/pytest/rocprofsys/runners.py#L187
 - https://github.com/ROCm/rocm-systems/blob/develop/projects/rocprofiler-systems/tests/pytest/rocprofsys/config.py#L121

This propagates to back to the `run_test` and it should do a `pytest.skip`.
https://github.com/ROCm/rocm-systems/blob/develop/projects/rocprofiler-systems/tests/pytest/conftest.py#L2160

Whilst your code will work, it should've been handled by `run_test`.

kcossett-amd:
The fact that it has to do with SHMEM makes me believe the error may be with the logic that appends an extra `--` to the command.

If `run_test` was flawed, a LOT more tests would fail. 
Please investigate before doing this. I believe the change, as it is now, is masking the real problem

mradosav-amd:
This is not solving the issue. Closing this PR.

PR #7501, opened 2026-06-19
[RCCL] enable WarpSpeed auto for grow communicators

atulkulk:
## Motivation

Grow tests fail on gfx950 because WarpSpeed is never enabled on grown (child) communicators. WarpSpeed auto-detection was only applied to top-level communicators, so any communicator created via the grow path ran without it, leading to the grow test failures.

## Technical Details

In `initTransportsRank` (projects/rccl/src/init.cc), the WarpSpeed auto-enable path was gated on `!parent`, which excludes grow communicators since they are created as children. The condition is relaxed to also cover grow communicators.

The explicit force-enable (`RCCL_WARP_SPEED_FORCE_ENABLE`) behavior is unchanged; only the auto path now applies to grow communicators.

## JIRA ID

Resolves AICOMRCCL-1335

## Test Plan

Run the RCCL grow communicator tests on gfx950.

## Test Result

All Grow tests passed.

## Submission Checklist

### DIFF
--- a/projects/rccl/src/init.cc
+++ b/projects/rccl/src/init.cc
@@ -1727,7 +1727,7 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, struct ncclComm* p
     }
   }
 #ifdef ENABLE_WARP_SPEED
-  comm->topo->warpSpeedEnabled = (rcclParamWarpSpeedForceEnable() > 0 || (!parent && rcclCanUseWarpSpeedAuto(comm, nNodes)));
+  comm->topo->warpSpeedEnabled = (rcclParamWarpSpeedForceEnable() > 0 || ((!parent || comm->isGrow) && rcclCanUseWarpSpeedAuto(comm, nNodes)));
 #endif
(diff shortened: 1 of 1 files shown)


nikitaxgusev:
The Ruby CI passing on gfx950 makes sense ,the librccl there isn't compiled with ENABLE_WARP_SPEED, so the
  #ifdef block containing this fix is a no-op and the bug never triggers in that environment. So the fix is
  correct and necessary for WarpSpeed-enabled builds.

  That said, could you add more context on what is actually being monitored and where? Right now I only see
  fixes appearing without visibility into the underlying signals, what failures triggered this, what's being
  watched, and what I should expect going forward.

nikitaxgusev (approved):
approve it, if grow tests are passing, but I would like to get what and how it should be monitored to prevent bugs.

atulkulk:
The fix is related to the issue caught during the execution of the unit/functional tests. After this fix, all Grow* tests were passed.

PR #7502, opened 2026-06-19
rocr: Fix coredump generation to pipes, and to file on lustrefs

lancesix:
## Motivation

This PR addresses 2 issues, with one series.

It was reported that coredump generation does not work on some file systems, such as [lustre](https://www.lustre.org/) (quite popular on HPC environments).

The ROCr runtime also has the ability to pipe the core dump content to a custom program.  However, coredumps produced that way are invalid, as reported by failed ROCgdb testsuite when exercising this functionality.

## Technical Details

For the lustre issue, it is caused by uses of the `posix_fallocate` call.  This is not supported on lustre.  Those call is not necessary anyway, drop them.

For the second issue, it is caused by the runtime failing to insert the required padding between sections.  As a result, section contents end up at invalid offsets in the coredump.  See the second patch of the series for further details.

## JIRA ID

- ROCM-25819
- AILIROCR-521

## Test Plan

Validate using ROCgdb's testsuite.  Testing coredump dumped to pipes requires to use https://github.com/ROCm/ROCgdb/pull/160.

## Test Result

Pass.

## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/lnx/amd_core_dump.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/lnx/amd_core_dump.cpp
@@ -631,9 +631,49 @@ struct LoadSegmentBuilder : public SegmentBuilder {
 
 // Write core dump to a file descriptor (for pipe handler)
 namespace {
+
+// Advance the write pointer in FD by SEEK bytes.  If FD is a regular file
+// (IS_REG_FILE is true), then use lseek(2).  Any non-written byte is
+// implicitly 0.  If FD is a pipe (IS_REG_FILE is false), then write null
+// bytes until we have advanced to the desired point.
+// When explicitly copying null bytes, use COPY_BUFFER (of size
+// COPY_BUFFER_SIZE).
+//
+// On error, return -1, with errno set to indicate the error.  On success
+// return the amount of bytes we have advanced in on the file descriptor.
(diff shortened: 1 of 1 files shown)


dayatsin-amd (approved):

cfreeamd (commented):
The two core bug fixes (lustre `posix_fallocate` and pipe alignment) look correct.

---

**[Important]** `seek_to()` partial-write on pipes not treated as error

When `is_reg_file=false`, `seek_to()` loops calling `write()` to emit padding. If a single `write()` returns fewer bytes than requested (`r > 0` but `r < c`), the loop breaks early and returns a positive value — the caller's `if (r < 0)` check doesn't catch this. In debug builds the `assert(total_written == phdr.p_offset)` fires (hard crash). In release builds the assert is compiled out and the coredump silently has wrong section offsets — defeating the fix.

In practice ELF alignment padding is ≤ 15 bytes (well below `PIPE_BUF`), so a partial write is extremely unlikely. Fix: `return -1` from `seek_to()` when the write count doesn't match the requested count.

---

**[Nit]** Error messages still say "to pipe" but this path now handles both pipes and regular files. Remove "to pipe" from `"Failed to write ELF header to pipe"`, `"Failed to write program header to pipe"`, `"Failed to write segment data to pipe"`.

---

**[Nit]** Typos in new comments:
- `"necessary paddint to respect the desired alinment"` → `"necessary padding to respect the desired alignment"`
- `"the amont of bytes we have advanced"` → `"the amount of bytes we have advanced"`

lancesix:
Addressed the various comments reported by the :robot:.  Unless there are concerns, I'll merge with clean CI.

cfreeamd (approved):

PR #7513, opened 2026-06-19
rocrtst: Add IPC tests for server restart and free-before-detach scenarios

shwetagkhatri:
Add extended IPC test coverage for the socket-server teardown/restart
fix and handle-lifetime rework.

## Submission Checklist


projects/rocr-runtime/rocrtst/suites/functional/ipc_extended.cc
```diff
@@ -0,0 +1,791 @@
...
+  FORK_ASSERT_EQ(HSA_STATUS_SUCCESS, err,
+                 "Child: Failed to attach to buffer 2 (server restart failed?)\n");
+  PROCESS_LOG("Child: Attached to buffer 2 at %p (server restart validated)\n", ipc_ptr2);
+
+  // Detach from buffer 2
+  err = hsa_amd_ipc_memory_detach(ipc_ptr2);
+  FORK_ASSERT_EQ(HSA_STATUS_SUCCESS, err);
+
+  shared_->phase = PHASE_ATTACH_2_DONE;
+
```
cfreeamd, line 288: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** `PHASE_ATTACH_2_DONE` is set here *after* `hsa_amd_ipc_memory_detach(ipc_ptr2)` completes — the actual invariant at this point is "detach 2 done", not "attach 2 done". The parent waits for `PHASE_ATTACH_2_DONE` before freeing `gpuBuf2`, so the logic is currently safe (detach has already happened), but the name is inverted. A future maintainer reading the name may reason that the child still holds the buffer attached when this signal fires, and could introduce a real free-while-attached bug if they reorder around it.

Suggest renaming to `PHASE_DETACH_2_DONE` throughout.

projects/rocr-runtime/rocrtst/suites/functional/ipc_extended.cc
```diff
@@ -0,0 +1,791 @@
...
+  err = hsa_amd_ipc_memory_attach(const_cast<hsa_amd_ipc_memory_t*>(&shared_->handle2),
+                                  shared_->size, 1, ag_list, &ipc_ptr2);
+  FORK_ASSERT_EQ(HSA_STATUS_SUCCESS, err, "Child: Failed to attach to buffer 2\n");
+  PROCESS_LOG("Child: Attached to buffer 2 at %p\n", ipc_ptr2);
+
+  err = hsa_amd_ipc_memory_detach(ipc_ptr2);
+  FORK_ASSERT_EQ(HSA_STATUS_SUCCESS, err);
+
+  shared_->phase = PHASE_ATTACH_2_DONE;
+
```
cfreeamd, line 519: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** Same `PHASE_ATTACH_2_DONE` naming inversion as the `IPCServerRestartTest` child — detach has already completed at this point. See the comment on the earlier instance.

cfreeamd (commented):
🤖 AI-assisted pre-review (for @cfreeamd) — `accept` → `accept4(SOCK_CLOEXEC)` hardening and `hsaKmtMemHandleFreePreserveMetadata` opt-in look correct. Inline comments above cover the phase-name inversion that could cause a future maintainer to introduce a real free-while-attached bug.

Issue #7520, opened 2026-06-19
[CLR/HIP] hipEventDestroy() on a recorded IPC event blocks on SyncAllStreams() — workload-proportional latency (AMD-only, ROCm 7.2.x)


andyluo7:
## Summary

Destroying a **recorded interprocess (IPC) HIP event** (`hipEventCreateWithFlags(..., hipEventInterprocess)`) blocks the calling CPU thread for the full duration of **all pending GPU work on the same device/process context** — even work completely unrelated to the event, and even when the event itself is already complete. CUDA's `cudaEventDestroy()` does not exhibit this. This shows up as 10–100 ms latency spikes in LMCache/vLLM-style cross-process workloads where IPC events are destroyed on the hot path.

## Environment

- ROCm 7.2.4 (also reproduced on 7.2.x), `hipamd` runtime (`projects/clr/hipamd`)
- AMD Instinct GPU; single process and real multi-process IPC both reproduce
- Comparison baseline: NVIDIA H200 + CUDA 13.0 — does **not** reproduce

## Reproduction

```cpp
hipEvent_t ev;
hipEventCreateWithFlags(&ev, hipEventDisableTiming | hipEventInterprocess);
hipEventRecord(ev, event_stream);
hipEventSynchronize(ev);
hipEventQuery(ev);              // hipSuccess — event is already complete

// unrelated work on a DIFFERENT stream of the SAME device
spin_kernel<<<grid, block, 0, unrelated_stream>>>();   // ~40 ms

hipEventDestroy(ev);           // BLOCKS ~40+ ms, despite ev being complete
```

Full standalone repro + benchmarks (14 HIP cases + matching CUDA baseline, CSV results, strace): https://github.com/giovanniguastiamd/hip-event

## Measured behavior

| Benchmark | scenario | `hipEventDestroy()` |
|---|---|---|
| bench_06 | IPC event ready, unrelated same-device work pending | **~43–47 ms** |
| bench_13 | event-stream pre-synced only, unrelated work pending | **~42–43 ms** |
| CUDA equiv | same pattern, H200/CUDA 13.0 | ~5 µs |

Properties: device-local, context-local, snapshot-fence (drains work queued at call time), workload-proportional, and **specific to recorded IPC events** (normal events and never-recorded IPC events are unaffected).

## Root cause (source-level, confirmed)

[...]

PR #7523, opened 2026-06-19
[CLR][HIP] Defer IPCEventEmulated physical cleanup; drop unconditional shm_unlink

andyluo7:
## Context

ROCm/rocm-systems#7520 reports that `hipEventDestroy()` on a recorded IPC event blocks for the full duration of unrelated GPU work on the device. The root cause is `~IPCEvent()` calling `ihipHostUnregister(&ipc_shmem_->signal)` in the destroy path; `ihipHostUnregister()` unconditionally calls `device->SyncAllStreams()`.

On current `develop` the **true ROCr-signal `IPCEvent`** (the ROCm/Linux path) has already been rewritten to use `amd::device::Signal` with a targeted `ipc_signal_->Wait(...)` — it no longer calls `ihipHostUnregister()`/`SyncAllStreams()`, so that path is already fixed. The original pattern survives only in **`IPCEventEmulated`** (the PAL/Windows fallback used when ROCr IPC signals are unavailable). This PR fixes that remaining case.

## Changes

`projects/clr/hipamd/src/hip_event.hpp` — `~IPCEventEmulated()`:
- Keep the cheap `synchronize()` (polls this event's own IPC signal; does not drain unrelated streams).
- **Defer** the physical cleanup (`ihipHostUnregister` → `SyncAllStreams`, `MemoryUnmapFile`, `shm_unlink`) out of the destroy path: decrement `owners`, detach the user-facing handle, and enqueue a `DeferredIpcEventCleanup{shmem, ipc_name, owners_after_decrement, device_id}`.
- **Remove the unconditional POSIX `shm_unlink()`** that ran at the end of the destructor. It fired even when `owners > 0` (peers still hold the shmem) or when the name was null/already unlinked, removing the shm name while other processes still needed to open it by name. The `amd::Os::shm_unlink` in the deferred `owners == 0` branch is the correct and sufficient unlink.

`projects/clr/hipamd/src/hip_event_ipc.cpp`:
- Add a small, mutex-guarded deferred-cleanup queue with `enqueueDeferredIpcEventCleanup()` / `drainDeferredIpcEventCleanup(device_id)` and the physical `cleanupDeferredIpcEvent()` (the original unregister/unmap/unlink sequence).
- Bounded: a threshold (256) triggers an inline drain on overflow so the queue can't grow unbounded if a process never reaches a drain point.

`projects/clr/hipamd/src/hip_device_runtime.cpp`:
- Drain the queue at `hipDeviceSynchronize()` (after `SyncAllStreams`) and `hipDeviceReset()` (before `Reset`) — points where a device-wide wait is already expected, so the `SyncAllStreams()` inside `ihipHostUnregister()` adds no surprise latency.

## Correctness

- Ownership/refcount semantics unchanged: `owners` is still decremented at destroy time; `shm_unlink` still happens only when `owners == 0`, just at the drain point.
- `synchronize()` is retained in the destructor, so the IPC signal is already complete by the time the item is enqueued; the deferred drain only performs resource release (no re-check needed).
- Physical resources stay mapped/registered until drained, so no use-after-free.
- Thread-safe (queue under a mutex; physical cleanup runs outside the lock).

## Validation

From the reproduction repo (AMD Instinct, ROCm 7.2.x) on the equivalent pre-refactor destructor:

| Benchmark | scenario | before | after |
|---|---|---|---|
| bench_06 | IPC event ready; unrelated same-device work pending | ~43–47 ms | **~1.0 µs** |
| bench_13 | event-stream pre-sync only; unrelated work pending | ~42–43 ms | **~0.81 µs** |

Total GPU work + IPC cleanup cost (~1.3–1.4 ms) is conserved — paid at the explicit sync point, not absorbed by `hipEventDestroy()`. Matches CUDA `cudaEventDestroy()` semantics.

## Test plan

- [ ] Existing HIP IPC event tests pass (single + multi-process), emulated path included.
- [ ] Regression: recorded emulated-IPC event ready + unrelated same-device work pending → `hipEventDestroy()` returns in µs; resources released by `hipDeviceSynchronize()`/`hipDeviceReset()`.
- [ ] Multi-process emulated IPC: `owners` honored; no premature `shm_unlink` (secondary-bug regression).
- [ ] No shmem/registration leak across create/record/destroy churn (queue drains; bounded threshold path exercised).
- [ ] Builds on Ubuntu/RHEL; MI-series + Radeon per CLR acceptance criteria.

## Note / follow-up

The customer in #7520 is on the ROCm 7.2.4 **release**, whose single `IPCEvent` still had the `SyncAllStreams` pattern. The `device::Signal` `IPCEvent` rewrite that fixes the ROCm path needs to land in a release / be backported to the 7.2.x branch — tracked separately in #7520.

### DIFF
--- a/projects/clr/hipamd/src/hip_device_runtime.cpp
+++ b/projects/clr/hipamd/src/hip_device_runtime.cpp
@@ -587,6 +587,11 @@ hipError_t hipDeviceGetSharedMemConfig(hipSharedMemConfig* pConfig) {
 hipError_t hipDeviceReset(void) {
   HIP_INIT_API(hipDeviceReset);
 
+  // Run any deferred emulated-IPC-event cleanup before tearing down the device,
+  // while it (and its registered host memory) is still valid. This is a safe
+  // drain point: device reset already implies a full device-wide wait/teardown.
+  hip::drainDeferredIpcEventCleanup(hip::getCurrentDevice()->deviceId());
+
   hip::getCurrentDevice()->Reset();
 
   HIP_RETURN(hipSuccess);
@@ -679,6 +684,9 @@ hipError_t hipDeviceSynchronize() {
   CHECK_SUPPORTED_DURING_CAPTURE();
   constexpr bool kDoWaitForCpu = false;
   hip::getCurrentDevice()->SyncAllStreams(kDoWaitForCpu);
+  // Safe drain point: the device was just synchronized, so the SyncAllStreams()
+  // inside the deferred IPC cleanup (ihipHostUnregister) is now cheap.
+  hip::drainDeferredIpcEventCleanup(hip::getCurrentDevice()->deviceId());
   HIP_RETURN_DURATION(hipSuccess);
 }
 
--- a/projects/clr/hipamd/src/hip_event.hpp
+++ b/projects/clr/hipamd/src/hip_event.hpp
@@ -244,19 +244,34 @@ class IPCEventEmulated : public Event {
       // work -- it does NOT drain unrelated streams.
       hipError_t status = synchronize();
       (void)status;
-      // Defer the physical IPC cleanup (ihipHostUnregister -> SyncAllStreams,
-      // munmap, shm_unlink) out of the destroy path so hipEventDestroy() does
-      // not block on unrelated device work. Drained later at a safe sync point.
+#if !defined(_MSC_VER)
+      // POSIX: defer the physical IPC cleanup (ihipHostUnregister ->
+      // SyncAllStreams, munmap, shm_unlink) out of the destroy path so
+      // hipEventDestroy() does not block on unrelated device work (the #7520
+      // stall). Drained later at a safe device-wide sync point.
       hip::ihipIpcEventShmem_t* shmem = ipc_evt_.ipc_shmem_;
       ipc_evt_.ipc_shmem_ = nullptr;  // detach the user-facing handle
       hip::enqueueDeferredIpcEventCleanup(
           {shmem, ipc_evt_.ipc_name_, owners, deviceId()});
+#else
+      // Windows/PAL: keep the original inline cleanup. The destroy-latency issue
+      // (#7520) was reported and validated only on Linux, so the emulated path
+      // here is left unchanged to avoid an untested behavior shift on a path
+      // that has no ROCr-IPC-signal alternative.
+      status = ihipHostUnregister(&ipc_evt_.ipc_shmem_->signal);
+      if (!amd::Os::MemoryUnmapFile(ipc_evt_.ipc_shmem_, sizeof(hip::ihipIpcEventShmem_t))) {
+        // print hipErrorInvalidHandle;
+      }
+      if (owners == 0) {
+        amd::Os::shm_unlink(ipc_evt_.ipc_name_);
+      }
+#endif
     }
-    // NOTE: the previous unconditional POSIX shm_unlink() that ran here was
-    // removed. It fired even when owners > 0 (other processes still hold the
-    // shmem) or when ipc_shmem_ was null / already unlinked, removing the shm
-    // name from the filesystem while peers still needed to open it by name. The
-    // amd::Os::shm_unlink in the deferred owners == 0 path is the correct and
+    // NOTE: the previous unconditional POSIX shm_unlink() that ran here (outside
+    // the ipc_shmem_ guard, under #if !defined(_MSC_VER)) was removed. It fired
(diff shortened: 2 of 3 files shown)


andyluo7:
Good catch — thanks Saleel. You're right that the emulated class is the only IPC-event path on Windows/PAL (the selection in `ihipEventCreateWithFlags` is runtime-probed: `dev->createIpcSignal()` succeeds → true ROCr `IPCEvent` (#7710's path), else → `IPCEventEmulated`; on Windows there are no ROCr IPC signals, so it always lands on the emulated path). So the original `~IPCEventEmulated()` deferral did change Windows behavior.

Two parts to this PR, with different platform reach:

1. **Removing the unconditional `shm_unlink`** — that block was already `#if !defined(_MSC_VER)`, i.e. POSIX-only and compiled out on Windows. No Windows impact.
2. **The deferred-cleanup refactor** — this was the part that touched Windows, and changing the Windows/PAL emulated destroy isn't necessary: the destroy-latency bug (#7520, the `ihipHostUnregister`→`SyncAllStreams` stall) was reported and validated only on Linux.

Pushed 43a240e to scope it: the deferred enqueue is now gated behind `#if !defined(_MSC_VER)`, and the `#else` keeps the **original inline cleanup** on Windows (synchronize → `ihipHostUnregister` → `MemoryUnmapFile` → `owners == 0` `shm_unlink`), byte-for-byte unchanged. On Windows the deferred queue is therefore always empty, so the drain calls in `hipDeviceSynchronize`/`hipDeviceReset` are no-ops there. Net effect: the PR now only changes the Linux/POSIX emulated path, which is the surface where #7520 was reproduced and fixed.

andyluo7:
### Ready for re-review

The Windows-path concern is addressed in 43a240e. Summary of where this PR now stands:

**What it changes (Linux/POSIX emulated path only):**
- `~IPCEventEmulated()` no longer runs the physical IPC cleanup (`ihipHostUnregister` → `SyncAllStreams`, `munmap`, `shm_unlink`) inline. It keeps the cheap `synchronize()` (this event's own completion) and enqueues the physical cleanup onto a bounded, mutex-guarded deferred queue, drained at points where a device-wide wait is already expected (`hipDeviceSynchronize` after `SyncAllStreams`, `hipDeviceReset` before `Reset`). This removes the `hipEventDestroy()` stall on unrelated device work (#7520).
- Removes the unconditional POSIX `shm_unlink()` at the destructor tail that fired even when `owners > 0` or `ipc_shmem_` was null/already-unlinked (racing peers opening the shm by name). The `owners == 0` unlink in the deferred path is the correct and sufficient unlink.

**Windows/PAL: unchanged.** The deferred enqueue is gated behind `#if !defined(_MSC_VER)`; the `#else` keeps the original inline cleanup byte-for-byte. The emulated path is the only IPC-event path on Windows (no ROCr IPC signals), and #7520 was reproduced/validated only on Linux, so Windows behavior is intentionally left as-is. The deferred queue is always empty there, making the drain calls no-ops.

**Scope vs. #7710:** that PR (merged) handles the true ROCr `IPCEvent` path; this one handles the disjoint `IPCEventEmulated` path it leaves untouched.

No outstanding review items that I'm aware of — happy to adjust if anything else stands out.

andyluo7:
> Do you need windows?

No — the driving use case is KV-cache IPC for LMCache/vLLM on ROCm **Linux** (MI355X); there's no Windows consumer here. The bug (#7520) was only ever reproduced on Linux.

> I would avoid touching windows unless there is a real use case

Agreed, and that's exactly the intent of 43a240e: it does **not** change Windows behavior. The `#else` branch isn't a new Windows code path — it just restores the original inline cleanup verbatim, so the Windows/PAL emulated destructor is byte-for-byte identical to `develop`. Only the Linux/POSIX path is modified (defer the cleanup + drop the buggy unconditional `shm_unlink`).

> The IPC event emulated design is not great, and soon rocr would be the default on Windows too

Makes sense — and that's another reason to leave the emulated Windows path alone rather than invest in it. Once ROCr is the default there too, this whole emulated path goes away. So this PR deliberately scopes itself to Linux/POSIX and leaves Windows untouched for the runtime to phase out.

Let me know if you'd prefer I drop the `#else` and structure it differently, but as-is it keeps Windows behavior frozen.

andyluo7:
Reopening: closed by the stale bot after 30 days, not on review. The Windows/PAL concern raised in review is addressed in 43a240e — the `#else` branch restores the original inline cleanup verbatim, so Windows behavior is byte-for-byte identical to `develop` and only the Linux/POSIX path changes. Still ready for re-review.

PR #5575, opened 2026-04-29
clr: Replace emulated IPC events with true ROCr IPC signals

saleelk:
Use hsa_amd_ipc_signal_create/attach for cross-process event synchronization instead of shared-memory slots and blit kernels. Record uses an async signal handler on the barrier's internal signal to set the IPC signal; StreamWait uses WaitCompleteSignal for GPU-side waits. The emulated path is preserved as fallback for PAL/Windows.

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #7526, opened 2026-06-19
[CLR][HIP] Defer IPCEvent physical cleanup; drop unconditional shm_unlink (7.2.x backport)

andyluo7:
## Context

Backport of the IPC-event `hipEventDestroy()` latency fix to `release/rocm-rel-7.2`. See ROCm/rocm-systems#7520 for the full root cause.

On 7.2.x, destroying a recorded IPC event (`hipEventInterprocess`) blocks the calling thread for the full duration of **unrelated** GPU work on the same device, even when the event is already complete. Root cause: `IPCEvent::~IPCEvent()` calls `ihipHostUnregister(&ipc_shmem_->signal)` in the destroy path, and `ihipHostUnregister()` unconditionally calls `g_devices[deviceId]->SyncAllStreams()` (a full device drain). There is also a secondary bug: an unconditional POSIX `shm_unlink()` at the destructor tail that fires even when `owners > 0` (peers still hold the shmem) or the name was already unlinked.

## Why a minimal backport (not #5575)

`develop` already fixes this comprehensively by replacing emulated IPC events with true ROCr IPC signals (#5575) — but that is a **cross-component** change (clr `hipamd` + `rocclr/device/rocm` + `rocr-runtime` HSA IPC signal create/attach) and is not a clean cherry-pick to 7.2.x. This PR is a **self-contained, low-risk** fix for the 7.2.x line with **no ROCr/rocclr dependency**.

## Change

- `~IPCEvent()`: keep the cheap `synchronize()` (polls this event's own signal; does not drain unrelated streams), then **defer** the physical cleanup (`ihipHostUnregister` / `MemoryUnmapFile` / `shm_unlink`) into a bounded, mutex-guarded queue instead of running it inline.
- **Remove** the unconditional POSIX `shm_unlink()` (the `amd::Os::shm_unlink` in the deferred `owners == 0` path is correct and sufficient).
- Drain the queue at `hipDeviceSynchronize()` (after `SyncAllStreams`) and `hipDeviceReset()` (before `Reset`) — points where a device-wide wait is already expected, so the `SyncAllStreams()` inside `ihipHostUnregister()` adds no surprise latency. Bounded threshold (256) inline-drains on overflow.

Ownership/refcount semantics unchanged; resources stay mapped/registered until drained (no use-after-free); thread-safe.

## Validation

From the reproduction repo (AMD Instinct, ROCm 7.2.x): `hipEventDestroy()` ~43–47 ms → **~1 µs** (bench_06/13). Total GPU + IPC cleanup cost (~1.3–1.4 ms) is conserved — paid at the explicit sync point, not absorbed by `hipEventDestroy()`. Matches CUDA `cudaEventDestroy()` semantics.

## Test plan

- [ ] Existing HIP IPC event tests pass (single + multi-process).
- [ ] Recorded IPC event ready + unrelated same-device work pending → `hipEventDestroy()` returns in µs; resources released by `hipDeviceSynchronize()`/`hipDeviceReset()`.
- [ ] Multi-process: `owners` honored; no premature `shm_unlink` (secondary-bug regression).
- [ ] No shmem/registration leak across create/record/destroy churn.

Relates to: ROCm/rocm-systems#7520 · equivalent develop fix for the emulated path: ROCm/rocm-systems#7523

### DIFF
--- a/projects/clr/hipamd/src/hip_device_runtime.cpp
+++ b/projects/clr/hipamd/src/hip_device_runtime.cpp
@@ -585,6 +585,11 @@ hipError_t hipDeviceGetSharedMemConfig(hipSharedMemConfig* pConfig) {
 hipError_t hipDeviceReset(void) {
   HIP_INIT_API(hipDeviceReset);
 
+  // Run any deferred IPC-event cleanup before tearing down the device, while it
+  // (and its registered host memory) is still valid. This is a safe drain point:
+  // device reset already implies a full device-wide wait/teardown.
+  hip::drainDeferredIpcEventCleanup(hip::getCurrentDevice()->deviceId());
+
   hip::getCurrentDevice()->Reset();
 
   HIP_RETURN(hipSuccess);
@@ -675,6 +680,9 @@ hipError_t hipDeviceSynchronize() {
   CHECK_SUPPORTED_DURING_CAPTURE();
   constexpr bool kDoWaitForCpu = false;
   hip::getCurrentDevice()->SyncAllStreams(kDoWaitForCpu);
+  // Safe drain point: the device was just synchronized, so the SyncAllStreams()
+  // inside the deferred IPC cleanup (ihipHostUnregister) is now cheap.
+  hip::drainDeferredIpcEventCleanup(hip::getCurrentDevice()->deviceId());
   HIP_RETURN_DURATION(hipSuccess);
 }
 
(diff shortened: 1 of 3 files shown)


andyluo7:
Reopening: this was closed by the stale bot, not on review. Still relevant and still applies cleanly to `release/rocm-rel-7.2`.

New context since June: this is now biting a downstream consumer. LMCache's KV-cache IPC path allocates and destroys `interprocess=True` events on the store/retrieve hot path, and the released ROCm 7.2.x that `vllm/vllm-openai-rocm` ships is exactly where the `hipEventDestroy()` -> `SyncAllStreams()` stall is still live — the `device::Signal` `IPCEvent` rewrite on `develop` fixed the ROCm/Linux path, but has not landed in a release. See LMCache/LMCache#4422 for the downstream discussion.

Happy to rebase or split this further if that helps it move.

PR #7534, opened 2026-06-19
[rocjitsu] DBI: Added call to an empty probe

accauble:
## Motivation

This PR implements the next step for binary instrumentation -- actually calling a probe. This includes resolving a compiled, self-contained (nop) probe, copying the body in the local code cave, and building a trampoline to call the probe via `s_swappc_b64`. To keep this as simple as possible, the probe (`rj_nop_probe`) does not yet do anything besides return back to the trampoline. This means that the Instrumentor will still look incomplete (e.g. register spilling is not necessary yet). The aim is to get the main framework in so we can keep building off of it.

## Technical Details

This PR includes:

The patching algorithm now looks like:

- `resolve_points`
  - Resolve probes (`resolve_probe_symbol`) -- Given a compiled device ELF and a symbol, find the symbol, validate it, and record where it is. See below for generating probe objects.
  - Make the probes callable (`build_probe_callable`) -- Given a resolved probe, capture the important information about it, like the calling convention (currently only supporting no arguments and a return with s[30:31]), the bytes that need to be copied, and (eventually) where that probe will sit in the `.text` section
  - Resolve instrumentation sites -- Verify that innnstrumentation anchors are valid and associated probes are callable (if it calls a probe)
- Create a code cave at the bottom of the `.text` file. The code cave will place the probe bodies and then the trampolines. Note that this does not currently take advantage of creating one local code cave per kernel. When that time comes, we can place all probes at the top of the first cave and emit long jumps to get to them.
- Determine where each probe will go
- Determine the layout for each trampoline. For probe-calling trampolines, this requires making a TrampolinePlan (`make_base_plan`) and selecting the registers required (`plan_probe_call`). The register selection, calling convention, etc will determine how big the trampoline will be. For the assembly trampoline, we still only support a nop so the layout is easy to calculate.
- Have the TrampolineBuilder emit the bytes for each trampoline
- If everything looks good, append each probe body and each trampoline. Update anchors.

`rj_nop_probe` looks like:

```C
extern "C" __device__ __attribute__((noinline, used, visibility("default")))
void rj_nop_probe() {}
```

* `extern "C"` keeps the symbol name intact so we can search for it
* `noinline, used` keeps the body intact and makes sure the compiler doesn't optimize it away
* `visibility("default")` exports the function
* Compile it with `-fgpu-rdc` to keep the `noinline, used` function from being dropped
* Unbundle the clang offload bundle to turn it into a device ELF with `clang-offload-bundler --offload`
* The format for the probes is probably a bit finicky for end users, but it sounds like we may not let them provide their own probe anyway

## Test Plan

- All previous tests still need to work
- Added new unit tests for new/updated classes
- gfx90a only (for now): Added end-to-end tests very similar to the ones implemented for the inlined nop. The only difference is that we need a version of vector_add that forces s30/31 to be allocated, since the is the only currently supported calling convention. We test that the nop probe instrumentation actually happens correctly by:
  - Saving the original kernel so that both original and instrumented can be executed
  - Making sure that the original kernel has output and then making sure that the instrumented kernel produces the same output
  - Statically checking that the instrumented instruction is now a branch, the probe was copied in, and the trampoline contains a call to the probe
  - Placing `endpgm` in the probe body and making sure the instrumented kernel exits without doing the add.

## Test Result

- All previous tests still pass
- Added unit tests pass
- End to end tests pass

## Submission Checklist


emulation/rocjitsu/cmake/rj_add_probe_object.cmake
```diff
@@ -0,0 +1,62 @@
...
+# before including this module.
+#
+# Usage: rj_add_probe_object(<name> <offload_arch> [OUTPUT_NAME <output_name>])
+#   name         - base name of the .hip source (without extension)
+#   offload_arch - GPU target (e.g. gfx90a)
+#   OUTPUT_NAME  - optional; basename of the output .hsaco (default: ${name}).
+#
+# Creates a custom target probe_${output_name} producing
+# ${KERNEL_OUTPUT_DIR}/${output_name}.hsaco.
+function(rj_add_probe_object name offload_arch)
```
atgutier: Could this live in rj_add_device_kernel.cmake as an update to the existing device kernel function or as a new one?

emulation/rocjitsu/docs/dbi-design.md
```diff
@@ -257,11 +260,49 @@ Branch offsets are computed in SOPP `simm16` units; the trampoline must lie with
...
+- **Static probe-call smoke (`tests/dbi/hsa_dbi_nop_probe_test.cpp`, `HsaDbiNopProbeStatic`, gated on `HAS_PROBE_FIXTURES`):** Instruments the register-padded `vector_add_probe` kernel with a probe call to `rj_nop_probe`, then asserts the patch shape — anchor decodes as `s_branch`, `.text` grew, the copied probe body matches the resolved body and ends in `s_setpc_b64`, and the trampoline contains the `s_swappc_b64` to it. No GPU required.
+- **Hardware probe-call smoke (`HsaDbiNopProbeHardware`, gated on `HAS_CDNA2_GPU`):** Load + validate; dispatch the probe-call-patched kernel and confirm bit-identical output to the original (the no-op probe is transparent); a *sabotage* test overwrites the copied probe body's first word with `s_endpgm` and asserts the wave terminates, proving the `s_swappc_b64` genuinely transfers control into the body. **Preconditions:** these dispatching cases require the [register-padded fixture kernel](#instrumentation-flow-probe-call) *and* the SMEM SBASE operand decode fix in the branch's base; without both the probe-call wave hangs the GPU.
```
atgutier, line 307: Just curious, have you tried running on the simulator too?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
```diff
@@ -3,19 +3,24 @@
 
 #include "rocjitsu/code/patch/instrumentor.h"
 
+#include "rocjitsu/analysis/liveness.h"
 #include "rocjitsu/code/amdgpu_code_object.h"
 #include "rocjitsu/code/basic_block.h"
 #include "rocjitsu/code/code_object.h"
 #include "rocjitsu/code/patch/code_object_patcher.h"
 #include "rocjitsu/code/patch/error_report.h"
```
atgutier, line 11: Just making a note for those working on DBIT that this should be removed in favor of using the existing utils log library (it can be improved, but should be the place for warning, erroring, etc.).

Not here, but at some point a dedicated refactor to DBIT isn't going its own way or duplicating things.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp
```diff
@@ -83,4 +118,155 @@ std::optional<TrampolineBytes> TrampolineBuilder::build(const TrampolinePlan &pl
   return out;
 }
 
+bool TrampolineBuilder::plan_probe_call(TrampolinePlan &plan, ProbeCallingConvention cc,
```
atgutier, line 121: This is only considering the live registers and not the kernel descriptor's actual allocation. Is that safe?

emulation/rocjitsu/docs/dbi-design.md
accauble: I had not at the time but I have since tried it and so far so good!

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp
accauble: That is correct. Right now I am requiring scalar registers up to s30/31 so we can assume that calling convention and dead registers are picked from the lowest available. It is currently safe, but a more general implementation should happen sometime soon-ish. There is a note documenting this around line 147.

newling:
This is my first look at the DBI work. After reading the clear PR description, the general flow mostly makes sense to me now. I have 2 fairly basic questions. 

> Create a code cave at the bottom of the .text file. The code cave will place the probe bodies and then the trampolines. Note that this does not currently take advantage of creating one local code cave per kernel. When that time comes, we can place all probes at the top of the first cave and emit long jumps to get to them.

What is the advantage of local caves? It sounds like like the plan here is to have probes that are used by multiple kernels inserted into the cave of first kernel (that uses the probe), and then have long jumps in the trampolines of other kernels to get to it. So is the thinking that trampolines are always local, but probes (that are used by different kernels) are not? 

> currently only supporting no arguments and a return with s[30:31]

and 

> The format for the probes is probably a bit finicky for end users, but it sounds like we may not let them provide their own probe anyway

Not actually a question but for supporting arguments, I'm trying to imagine how they can be supported and what the level of automation will be (will a probe creator need to manually modify probe assembly to match a calling convention?). I can see they will be very useful though (I'm imagining a probe that, give an LDS byte address as an argument, checks a data structure to see if there might be a race associated with it). 

I also got ai to do a review: https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr7534.md The only (optional) actionable item I see there some additional checks on the probe to check it is valid with the current spill-free logic. Specifically a check that there are no writes to s30 or s31.

accauble:
@newling Thanks for taking a look!

> What is the advantage of local caves?

The advantage of local caves is that they are closer to the anchor, which means they are more likely to be reachable via a 4-byte branch. The local cave structure actually comes from the DBT work (which is moving faster than DBI). The original implementation had trampolines and DBT code caves in their own section down at the bottom of the `.text` file. This worked well until they started translating real apps with `.text` sections that were too large. As an intermediate fix, the "one code cave at the bottom of each kernel" was implemented. This will not be enough for larger kernels, but it is working in the interim.

When we get to instrumenting larger apps, we will probably start with this local code cave implementation (or we'll see what functionality exists when we get there). The probes do not need to be local since they are called with the swappc (and we have the room to add what we want). When we are overwriting original code in the `.text` section, we have to be much more careful with how we overwrite the instructions.

> I'm trying to imagine how they can be supported and what the level of automation will be

I, personally, would love to be able to make custom probes available to advanced users, but it will take a backseat to ease-of-use. I expect to provide users with a set of useful probes; one that generates a memory access stream will be a top priority. Since much of the work will happen on the host, this will likely cover most use cases, at least for the medium term.

accauble:
~~Note: This is currently failing CI because of a pre-existing issue that will be fixed in PR #8149.~~

kuhar (commented):
I think it would help to wire `probe_fixture_test` and `hsa_dbi_nop_probe_test` through the same installed-test helpers as `hsa_dbi_nop_asm_test`.

Right now they use `add_executable()` plus `add_test()` directly (`emulation/rocjitsu/tests/CMakeLists.txt:394-420` and `emulation/rocjitsu/tests/CMakeLists.txt:923-965`). With `RJ_INSTALL_TESTS` enabled, that means the new probe tests do not install their binaries, get `ROCJITSU_KERNEL_DIR` for installed runs, or emit installed CTest entries.

If these are intended to run from installed test packages, we should also add `vector_add_probe_gfx90a.o` and `rj_nop_probe_gfx90a.hsaco` to the installed fixture list in `emulation/rocjitsu/tests/kernels/CMakeLists.txt:97-110`.

We should tidy the stale docs/comments that still describe the pre-probe-call behavior.

- `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.h:193` still says `pt.probe_obj` must be null and `pt.probe_symbol` empty, but `validate_anchor()` now accepts both when both are set and only rejects inconsistent pairs.
- `emulation/rocjitsu/tests/kernels/rj_nop_probe.hip:9` points at `cmake/rj_add_probe_object.cmake)` with a stray closing paren; the helper now lives in `cmake/rj_add_device_kernel.cmake`.
- `emulation/rocjitsu/docs/dbi-design.md:254` says the inline-nop placeholder "becomes probe call later", but the probe-call flow is documented immediately below.

These are small, but getting them in sync would make the new probe-call support easier to follow.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
```diff
@@ -309,43 +446,131 @@ InstrumentedCodeObjectDebug Instrumentor::patch_with_debug_summaries() {
...
+      // Callee clobbers (probe body) + liveness at the anchor feed envelope
+      // resource selection and the no-spill policy gate.
+      auto summary = build_probe_clobber_summary(probe, &err);
+      if (!summary) {
+        result.errors.push_back(std::move(err));
+        continue;
+      }
+
+      // Check for special machine state that has no save/restore path yet
+      if (!check_probe_special_state(*summary, &err)) {
```
kuhar, line 548: We should add one more probe-body validation for the return-link pair.

`build_probe_callable()` accepts a body as long as it ends in `s_setpc_b64 s[30:31]`, but a body that writes `s30` or `s31` earlier would still pass. `plan_probe_call()` only checks that the link pair is dead at the anchor, and `check_probe_special_state()` only covers EXEC/VCC/M0/FLAT_SCRATCH, so this kind of clobber never enters the spill set and the probe can return through a corrupted PC.

This is not reachable with the current `rj_nop_probe`, but it seems like the first thing that breaks when a second probe body lands. Since `summary->ordinary_clobbers` is already computed at `instrumentor.cpp:505-508`, a small `check_probe_link_pair()` next to `check_probe_special_state()` should be enough to fail closed when the body clobbers `s[link_base:link_base+1]`. That also lines up with newling's request for a no-writes-to-s30/s31 check.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp
```diff
@@ -83,4 +118,155 @@ std::optional<TrampolineBytes> TrampolineBuilder::build(const TrampolinePlan &pl
...
+  if (any_sgpr_in_range(live_at_anchor, kLinkPairBase, 2)) {
+    report(error_out, "probe-call resource planning: return-link pair s[30:31] is live at the "
+                      "anchor; cannot yet save a live link pair");
+    return false;
+  }
+
+  RegisterSet link_pair;
+  link_pair.expand(RegisterRef{RegClass::SGPR, kLinkPairBase, 2});
+
+  // NOTE: target/scc selection scans the conservative cross-ISA allocatable
```
kuhar, line 147: It would be good to turn the documented "kernel must already allocate through s31" precondition into a checked failure.

The register selection scans the cross-ISA `REGISTER_SET_ALLOCATABLE_SGPRS` bound (`trampoline_builder.cpp:147-151`), not the patched kernel's actual `.sgpr_count`. For a normal kernel that only allocates around 12 SGPRs, `s30/s31` can look dead at the anchor even though the wave does not own them. `plan_probe_call()` then emits a well-formed probe call and `patch()` reports success, but `vector_add_probe.hip`'s own comment says this faults or hangs the GPU.

This looks like the same concern atgutier raised on the thread: it is safe for the padded fixture, but not enforced for callers in general. Reading the kernel descriptor's `.sgpr_count` and failing closed when the chosen link/target/SCC registers fall outside it would turn that into a clean validation error now, and it gives the future auto-grow work a natural hook.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/probe_clobber.cpp
```diff
@@ -0,0 +1,118 @@
...
+    summary.touches_flat_scratch = true;
+    break;
+  default:
+    break;
+  }
+}
+
+// Scan every explicit operand and flag special-state register classes. This is
+// best-effort: it sees operands the decoder exposes, not implicit side effects.
+void scan_special_state(const Instruction &inst, ProbeClobberSummary &summary) {
```
kuhar, line 93: I think the special-state clobber check needs to fail closed here.

`scan_special_state()` only records a class when `to_register_ref()` returns a register, but the CDNA2 decoder returns `nullopt` for the special-state operand forms this gate is meant to catch: `OPR_SDST_EXEC`, `OPR_SDST_M0`, `OPR_FLAT_SCRATCH`, `OPR_VCC`, and `OPR_SSRC_SPECIAL_SCC`. The implicit-def path does not appear to rescue this either, because `RegisterSet::expand()` drops non-ordinary classes.

That means a probe body can write EXEC, M0, FLAT_SCRATCH, VCC, or SCC, still report every `touches_*` flag as false, pass `check_probe_special_state()`, and corrupt the host wave at runtime. We should either detect these operand classes structurally or conservatively reject unmapped special defs/uses until they are modeled.

It would also be worth adding decode-derived positive tests, for example bodies like `s_mov_b32 exec_lo, 0` and `s_mov_b32 m0, 0`, so this gate cannot silently become a no-op again.

emulation/rocjitsu/tests/patch/instrumentor_test.cpp
```diff
@@ -1348,5 +1364,368 @@ TEST_F(InstrumentorPatchDecoded, BranchesMultipleLandAtTheirIntendedTargets) {
...
+
+  // Forward branch from anchor (4) to trampoline (12): simm16 = (12-(4+4))/4 = 1.
+  const auto &p = result.patches[0];
+  ASSERT_EQ(p.patched_anchor_bytes.size(), 4u);
+  uint32_t w = 0;
+  std::memcpy(&w, p.patched_anchor_bytes.data(), sizeof(w));
+  EXPECT_EQ(w, build_s_branch(1, ROCJITSU_CODE_ARCH_CDNA4));
+}
+
+TEST(InstrumentorProbePatch, CopiesProbeBodyOnceAndCallTargetsIt) {
```
kuhar, line 1671: It would be useful to add patch-level coverage for multiple probe-call sites.

`CopiesProbeBodyOnceAndCallTargetsIt` is a good single-site check, but it cannot catch the two behaviors that make this layout tricky: deduplicating a shared probe body and targeting the right copied body from each trampoline.

Two synthetic-ELF cases would cover most of the risk:

- two anchors sharing the same `(probe_obj, probe_symbol)`, where the body is copied once and both trampolines target it
- two anchors calling distinct probes, where both bodies appear once in the expected order and each trampoline targets the right one

That should catch regressions where `resolve_probe_index()` starts copying per site or the trampoline materialization points at the wrong body.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instruction_builder.h
```diff
@@ -163,6 +172,51 @@ inline constexpr uint16_t kDelayAluSaluDep1 = 9;
   }
 }
 
+/// @brief Get the s_getpc_b64 (SOP1) opcode for a target ISA.
+[[nodiscard]] inline constexpr uint32_t sop1_op_getpc_b64(rj_code_arch_t arch) {
```
Groverkss: Don't we already have some of these being used in the indirect branch discovery analysis? Can we reuse them?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
```diff
@@ -230,61 +326,102 @@ const Instruction *Instrumentor::find_instruction_at_offset(uint64_t anchor_offs
...
-
   if (obj_.text_sections().empty()) {
-    result.errors.emplace_back("code object has no .text section");
-    return result;
+    out.errors.emplace_back("code object has no .text section");
+    return out;
   }
-  // TODO: support multi-text code objects. Anchor offsets
-  // would need to identify which .text section they belong to.
+  // TODO: support multi-text code objects. Anchor offsets would need to identify
```
Groverkss, line 361: Once i land the DBT work, we could reuse almost all of it for DBI here.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/probe_clobber.cpp
accauble: This is a real hole here. Thank you for pointing it out. This will hopefully be fixed soon (I'm hoping to add these operands to `to_register_ref` in a parallel PR). OTOH, this is all set up for the next PR which would include an actual useful probe (instead of just an empty probe) so I don't think it should block *this* PR.

I could add a temporary check for names which will catch many of these cases. I could go further and only allow certain instructions in the probe until the `to_register_ref` fix is complete/landed. Or we could just block the following PR until the `to_register_ref` is complete (I don't think it will work without that anyway).

accauble: I just re-read this and it sounds so much like an LLM. They are rubbing off on me :-P

Anyway, I added the temporary name check which should catch explicit operands. But I will frontload getting all operands available in `to_register_ref` so that it is done for the next PR.

emulation/rocjitsu/tests/patch/instrumentor_test.cpp
accauble: I added a little bit but multiple probe-call sites are going to require extra policies since it really depends on if the user wants one trampoline envelope for all calls or one per call.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instruction_builder.h
accauble: I do recall seeing that when looking over that PR. I have updated these per-arch opcode functions with the constants that are available. An actual refactor to completely remove duplication seemed like it was going to be a lot more work and not in scope of this already large PR.

accauble:
@Groverkss and @kuhar thanks for the reviews. I believe I have addressed all comments to at least some degree.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instruction_builder.h
accauble: I just saw #8408 is merged in so I will update the code to use that where possible.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/probe_clobber.cpp
kuhar: because it is :)

accauble: I meant my response! That I wrote entirely by my own hand!

kuhar:
@accauble can you rebase?

accauble:
Yes, I am currently rebasing and incorporating the changes from PR #8408 . I will notify you when that is done, assuming the CI also passes.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp
kuhar: The SGPR allocation gate still seems incomplete. It checks only the fixed return-link pair before resource planning, and it also skips the check entirely when `min_kernel_sgpr_count()` returns `nullopt`.

A 32-SGPR kernel with s0-s29 live and s30:s31 dead can still pass the link check, then `plan_probe_call()` may pick s32:s33 for the target pair or s32 for the SCC temp because it scans `REGISTER_SET_ALLOCATABLE_SGPRS`. That lets `patch()` report success for registers the wave does not own.

I think we should either bound resource planning by the decoded kernel SGPR count, or validate every selected SGPR resource (link pair, target pair, and SCC temp) after planning and fail closed when the count is unknown. It would also be good to add one patch-level under-allocated/unknown-descriptor case; the current test only exercises `probe_link_pair_fits_in_kernel()` directly.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/probe_clobber.cpp
kuhar: The remaining special-state risk is the implicit-write path. The new name fallback catches explicit operands, but `scan_special_state()` still documents that implicit writes such as `v_cmpx` -> EXEC and `v_cmp` -> VCC are invisible, while arbitrary probe objects are accepted in this PR.

We should fail closed for instructions/classes with unmodeled implicit EXEC/VCC effects until `to_register_ref()` or implicit special defs model them; otherwise a non-nop probe can still pass `check_probe_special_state()` and corrupt the host wave.

accauble: It's true but this PR would not work with a non-nop probe, even if the probe had no implicit operand effects. EXEC is going to have it's own special handling, since users will be able to configure how they want EXEC saved. I could handle VCC in the envelope (like how I handle SCC now), but since it doesn't actually matter for nop probes, I would rather clean up `to_register_ref` after this PR but before handling non-nop probes.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp
accauble: Passing this down looked like it was going to explode changes in a way that I was not thrilled about for an already-large PR that has been sitting here for a few weeks (I also thought I had written that decision somewhere but I'm not seeing it anymore so I apologize for that). That is why I put the gate where I did, which just halts it up at the instrumentor level. I believe it will make sense to add these checks either when we implement spilling, since at that time we will have a need to know which registers are available.

kuhar (approved):
Thanks for the fixes

jayhawk-commits:
Please use squash and merge in the future when all the commits, especially code clean-up, do not need to be preserved. This pull request left a lot of unnecessary commits in the history.

newling:
> Please use squash and merge in the future when all the commits, especially code clean-up, do not need to be preserved. This pull request left a lot of unnecessary commits in the history.

@jayhawk-commits I noticed this on one of my PRs recently too, I was wondering if there is a default setting that can be toggled somewhere to make this squash happen automatically?

accauble:
@jayhawk-commits Oh dear. I think this happened because I have a different workflow than the person that actually landed it. I should have warned him, my apologies.

Ugh, I guess it won't be possible to rewrite the history. Really embarrassed. Sorry again D-:

newling:
{2178C539-22C7-4402-9D57-7C812BF45664}

What I meant is the default at this point isn't 'squash and merge' but I wonder if it could be, it is easy when not paying close attention to 'merge' rather than 'squash and merge'

accauble:
I think I have heard from others that you cannot set the default in the list, you can only set the workflow which they do not want to do. :-\

PR #7539, opened 2026-06-19
fix(rocjitsu): repair sanitizer ISA execution paths

lialan:
## Motivation

Fix sanitizer-reported issues in the AMDGPU ISA execution and DBT translation paths so the gfx1250 execution, translation, and SIMD paths can be validated under sanitizer builds.

## Technical Details

This updates the:
* generated AMDISA execute helpers
* semantic lowering for affected AMDGPU instructions
* shared AMDGPU execute code
* SIMD glue used by the execution path
* DBT relocation expectations for the preserved descriptor-entry window
* sanitizer-stable gfx1250 simulation time checks
* related execution and translation tests

## Test Plan

Validated with focused rocjitsu sanitizer coverage across clang22 and GCC13:
* ASAN
* UBSAN
* FPSAN via `-fsanitize=undefined,float-divide-by-zero,float-cast-overflow`

Also reran clang22 TSAN coverage for the same focused slice

## Test Result

All relevant tests passing.

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/util/include/util/simd.h
+++ b/emulation/rocjitsu/lib/util/include/util/simd.h
@@ -18,10 +18,16 @@
 #include <experimental/simd>
 #endif
 
+// clang + libstdc++ <experimental/simd> on AVX-512 has produced incorrect
+// native 64-bit mask behavior on sanitizer CI hosts. Keep the default
+// workaround local to that stack, but leave it overrideable so fixed toolchains
+// can force vector coverage with -DUTIL_SIMD_BROKEN_NATIVE_64BIT_MASKS=0.
+#ifndef UTIL_SIMD_BROKEN_NATIVE_64BIT_MASKS
 #if defined(__clang__) && defined(__GLIBCXX__) && defined(__AVX512F__)
 #define UTIL_SIMD_BROKEN_NATIVE_64BIT_MASKS 1
 #else
 #define UTIL_SIMD_BROKEN_NATIVE_64BIT_MASKS 0
+#endif
 #endif
 
 namespace util {
--- a/emulation/rocjitsu/lib/util/include/util/simd.h
+++ b/emulation/rocjitsu/lib/util/include/util/simd.h
@@ -22,7 +22,10 @@
 // native 64-bit mask behavior on sanitizer CI hosts. Keep the default
 // workaround local to that stack, but leave it overrideable so fixed toolchains
 // can force vector coverage with -DUTIL_SIMD_BROKEN_NATIVE_64BIT_MASKS=0.
-#ifndef UTIL_SIMD_BROKEN_NATIVE_64BIT_MASKS
+#ifdef UTIL_SIMD_BROKEN_NATIVE_64BIT_MASKS
+#define UTIL_SIMD_BROKEN_NATIVE_64BIT_MASKS_USER_OVERRIDE 1
+#else
+#define UTIL_SIMD_BROKEN_NATIVE_64BIT_MASKS_USER_OVERRIDE 0
 #if defined(__clang__) && defined(__GLIBCXX__) && defined(__AVX512F__)
 #define UTIL_SIMD_BROKEN_NATIVE_64BIT_MASKS 1
 #else
--- a/emulation/rocjitsu/lib/util/include/util/simd.h
+++ b/emulation/rocjitsu/lib/util/include/util/simd.h
@@ -18,6 +18,12 @@
 #include <experimental/simd>
 #endif
 
+#if defined(__clang__) && defined(__GLIBCXX__) && defined(__AVX512F__)
+#define UTIL_SIMD_BROKEN_NATIVE_64BIT_MASKS 1
+#else
+#define UTIL_SIMD_BROKEN_NATIVE_64BIT_MASKS 0
+#endif
+
 namespace util {
 
 /// Compile-time switch for `<experimental/simd>` availability. Callers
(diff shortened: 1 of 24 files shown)


lialan:
@kuhar any chance we prioritize getting those sanitizers into CI? it is very easy to introduce new issues without sanitizers, and fixing those which could have been caught is annoying.

emulation/rocjitsu/lib/util/include/util/simd.h
```diff
@@ -18,6 +18,12 @@
 #include <experimental/simd>
 #endif
 
+#if defined(__clang__) && defined(__GLIBCXX__) && defined(__AVX512F__)
```
kuhar, line 29: Please make the SIMD fallback workaround explicit and testable. UTIL_SIMD_BROKEN_NATIVE_64BIT_MASKS disables broad f64 SIMD paths for clang+libstdc+++AVX512 without documenting the underlying bug, and some 32-bit shift/bitfield helpers are scalarized unconditionally. Add a short rationale or bug link, make the macro overrideable for CI/local coverage, and keep vectorized paths on unaffected toolchains unless the fallback is required.

emulation/rocjitsu/tests/dbt/translate_test.cpp
```diff
@@ -2112,7 +2150,7 @@ TEST(BinaryTranslatorE2E, Cdna4ToCdna3MfmaPartialScratchGrowsDescriptor) {
   expect_cdna3_translated_descriptor_vgprs_at_least(result.elf_bytes, kScratchFloor + 4);
 }
 
-TEST(BinaryTranslatorE2E, RelocatedKernelCompactsSourceGapsAndPatchesBranches) {
+TEST(BinaryTranslatorE2E, RelocatedKernelPreservesEntryWindowAndPatchesBranches) {
```
kuhar, line 2213: The DBT test rewrite is a real spec correction, not just sanitizer stabilization. The renamed relocation test now preserves the 256-byte kernarg preload window instead of compacting reachable blocks, and the entry-range test now keys on SHF_ALLOC rather than section name/range. Both may be correct, but please call them out in the PR or commit text; add coverage for reachable-block gap compaction beyond offset 256 if that behavior remains required.

kuhar (commented):
I ran the gfx1250 FPSan suite under both Clang 23 sanitizer builds through the rocjitsu CLI. Both completed CTest and failed broadly: ASan+UBSan 259/783 failed; TSan 260/783 failed. Failures span Parity/Atomic/Xlane, CVT, WMMA/SWMMAC, wave reductions, and AmdgcnLdexp, with many zeroed result or payload lanes. The PR is not sanitizer-clean for this validation target.

Scalar fallback paths still have sanitizer UB. The i24 multiply lowering can still overflow signed int32, and shift/bitfield scalar bodies can still use unmasked edge widths; these paths run under RJ_FORCE_SCALAR or when operands are not SIMD-capable. The ASan+UBSan FPSan run also repros signed overflow at execute_shared.h:1986 during AmdgcnLdexp.F16. Please share the same masked/unsigned helper semantics between SIMD probes and scalar lowering.

emulation/rocjitsu/lib/util/include/util/simd.h
lialan: Addressed in 9f8c0a724e. The clang+libstdc+++AVX512 native 64-bit mask workaround now has an explicit rationale, is guarded with `#ifndef` so CI/local builds can force `-DUTIL_SIMD_BROKEN_NATIVE_64BIT_MASKS=0`, and has default-selection coverage in `UtilSimd.Native64MaskWorkaroundMacroIsBooleanOverridePoint`. The 32-bit BFM/shift/i24 execution paths now keep the SIMD probes on vector-capable toolchains and route scalar fallback through the same unsigned helper semantics.

emulation/rocjitsu/tests/dbt/translate_test.cpp
lialan: Addressed in 9f8c0a724e. I added `BinaryTranslatorE2E.RelocatedKernelCompactsReachableGapAfterEntryWindow`, which preserves the 256-byte entry window and verifies that reachable blocks after that window still compact. The commit text also explicitly calls this out as a DBT relocation spec correction rather than only sanitizer stabilization.

lialan:
Follow-up pushed in af02a81547 to fix the remaining current sanitizer UB I could reproduce while rechecking the sanitizer review:

- v_bfe_i32 scalar fallback now uses unsigned mask/signbit math for sign extension.
- v_mad_legacy_u16 widens operands before multiply/add.
- signed/mixed VOP3P dot scalar fallbacks accumulate in uint32_t bits and clamp without signed overflow.

Validation:
- Python sema/codegen tests: 368 passed, 8 skipped.
- UBSan representative gfx1250/SIMD/MFMA/WMMA slice: 216 passed, 6 skipped.
- ASan same representative slice: 216 passed, 6 skipped, using direct binary run with detect_odr_violation=0:report_globals=0 because ASan gtest discovery still hits the existing ODR issue.
- TSan focused repro groups passed. The corrected wider 222-test representative run did not report a project race before ThreadSanitizer hit its own internal deadlock-detector CHECK in Gfx1250SimulationTest.GlobalStoreWritesVisibleMemory and hung.
- OPT-125M via /home/nod/workspace/rocm-utils/opt-125m.py on the freshly installed branch runtime passed; top-1 next token remains ' the'.

amd-eochoalo:
I've looked at the rocjitsu test corpus and the errors are introduced in a new TheRock release.

kuhar:
> I've looked at the rocjitsu test corpus and the errors are introduced in a new TheRock release.

Does this suggest we should pin it to a recent known-good release and periodically bump that?

kuhar:
@lialan let me rerun my local test suite...

amd-eochoalo:
> Does this suggest we should pin it to a recent known-good release and periodically bump that?

Yes, I've pinned it here https://github.com/ROCm/rocm-systems/pull/7609 . 

I think a better way to deal with this is figuring out what happened and trigger the CI when those files are modified.

emulation/rocjitsu/tests/util_simd_test.cpp
```diff
@@ -47,6 +47,17 @@ inline int sweep_iters() {
 
 constexpr std::size_t kW = util::native_width_v<uint32_t>;
 
+TEST(UtilSimd, Native64MaskWorkaroundMacroIsBooleanOverridePoint) {
+  constexpr int workaround = UTIL_SIMD_BROKEN_NATIVE_64BIT_MASKS;
+  static_assert(workaround == 0 || workaround == 1);
+
+#if defined(__clang__) && defined(__GLIBCXX__) && defined(__AVX512F__)
+  EXPECT_EQ(workaround, 1);
```
kuhar: We should make this test respect the documented override path. `simd.h` now lets fixed toolchains or CI force `-DUTIL_SIMD_BROKEN_NATIVE_64BIT_MASKS=0`, but this test still expects `workaround == 1` on every clang+libstdc+++AVX512 build. That forced-vector configuration would fail `UtilSimd.Native64MaskWorkaroundMacroIsBooleanOverridePoint` immediately. Please either test only that the macro is boolean here, or distinguish the auto-detected default from a user-supplied override before checking the clang/libstdc++/AVX512 default.

kuhar (approved):
LGTM overall

atgutier (approved):

PR #7540, opened 2026-06-19
[rocprofiler-compute] Preload native collector before rocprofiler-sdk tool

EvanPhoukong:
…t path is empty" abort with PyTorch workloads

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-compute/src/rocprof_compute_profile/profiler_rocprofiler_sdk.py
+++ b/projects/rocprofiler-compute/src/rocprof_compute_profile/profiler_rocprofiler_sdk.py
@@ -28,12 +28,20 @@ class rocprofiler_sdk_profiler(RocProfCompute_Base):
         args = self.get_args()
         app_cmd = shlex.split(args.remaining)
 
-        # Build LD_PRELOAD: preserve user's existing, then append our libs
-        # Order: [user's existing LD_PRELOAD] : [our profiler libs]
+        # Build LD_PRELOAD: preserve user's existing, then append our libs.
+        # Order: [user's existing LD_PRELOAD] : [native tool] : [rocprofiler-sdk tool]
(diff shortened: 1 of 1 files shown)


projects/rocprofiler-compute/src/rocprof_compute_profile/profiler_rocprofiler_sdk.py
```diff
@@ -28,12 +28,13 @@ def get_profiler_options(
         args = self.get_args()
         app_cmd = shlex.split(args.remaining)
 
-        # Build LD_PRELOAD: preserve user's existing, then append our libs
-        # Order: [user's existing LD_PRELOAD] : [our profiler libs]
+        # Native tool must precede the rocprofiler-sdk tool in LD_PRELOAD so it is the
+        # primary client; else workloads that force-configure on import (PyTorch libkineto)
```
vedithal-amd, line 32: The `output path is empty` was due to the native tool not able to get the ROCPROF_* env var which dictates the output directory

However, this problem was solved recently in the develop branch in  https://github.com/ROCm/rocm-systems/commit/f9d2a838c3e85f756d28260488b7f7a7b13b7e17

Also, one tool should not have side effects on other tools - so ordering should not matter - the elegant solution can be found by digging into the root cause. Ordering tools looks like a hacky solution.

Could you please provide the following debug information:
- Workload (so that we could run it) - can be a code with build instructions or docker container if heavy dependencies like pytorch
- Which GPU, Which OS, Which ROCm version? Basically the output of `rocprof-compute -s`
- The exact profiling command used with `--verbose`
- The log.txt generated in the profiled workload folder

Also please try using latest develop branch and see if the problem persists - if it persists we can debug based on the debug information.

Thanks!

vedithal-amd (changes requested):

PR #6997, opened 2026-06-09
Fix out-of-bounds read in ELF note section parsing.

dayatsin-amd:
Validate note entry bounds before reading headers and descriptor data to prevent heap OOB reads and infinite loops on malformed note sections.

Resolves SWSPLAT-24380

PR #7541, opened 2026-06-19
fix(rocr): Bound HSA note variable-length name reads

dayatsin-amd:
## Summary

Builds on #6997. That change hardens GElfNoteSection::getNote() so the returned descriptor pointer and n_descsz stay within the note section. But the note consumers one layer up still trust attacker-controlled size fields embedded inside the descriptor and pass them straight to GetNoteString, which reads that many bytes with no bound:

- GetNoteIsa: vendor_name_size / architecture_name_size
- GetNoteProducer: producer_name_size
- GetNoteProducerOptions: producer_options_size

A crafted code object can set these uint16 fields large so GetNoteString reads past the descriptor into adjacent heap memory (CWE-125).

Fix: expose the descriptor size from GetAmdNote() and, in each consumer, bound the claimed variable-length size against the room actually available in the descriptor before calling GetNoteString.

Addresses ROCM-26177 finding #3.

## JIRA ID

JIRA ID : ROCM-26177

## Test plan

- [ ] Load a normal code object; confirm no regression.
- [ ] Load a crafted code object with oversized note name fields; confirm rejection with HSA_STATUS_ERROR_INVALID_CODE_OBJECT.


cfreeamd (commented):
Changes to `GetAmdNote`, `GetNoteIsa`, `GetNoteProducer`, and `GetNoteProducerOptions` look correct; inline comments cover a few nits and one question about the arch-offset computation.

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_hsa_code.cpp
```diff
@@ -542,9 +543,22 @@ namespace code {
...
       amdgpu_hsa_note_isa_t *desc;
-      if (!GetAmdNote(NT_AMD_HSA_ISA_VERSION, &desc)) { return false; }
+      uint32_t desc_size = 0;
+      if (!GetAmdNote(NT_AMD_HSA_ISA_VERSION, &desc, &desc_size)) { return false; }
+      // The inner *_name_size fields are attacker-controlled. Bound the
+      // variable-length name array against the actual descriptor size before
+      // reading, otherwise GetNoteString reads past the descriptor into
+      // adjacent heap memory. See ROCM-26177 finding #3.
+      const size_t name_offset = offsetof(amdgpu_hsa_note_isa_t, vendor_and_architecture_name);
+      if (desc_size < name_offset) { return false; }
```

**[Optional]** This guard is unreachable dead code: `GetAmdNote` already guarantees `desc_size >= sizeof(S)`, and `sizeof(S) > offsetof(variable_array_field)` for all three structs (each ends in a `char[1]`). The check is harmless and adds defensive depth, but a comment noting the invariant — or a `static_assert(sizeof(S) > offsetof(variable_array_field))` — would clarify intent for future readers.

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_hsa_code.cpp
```diff
@@ -827,9 +841,16 @@ namespace code {
...
       amdgpu_hsa_note_producer_t* desc;
-      if (!GetAmdNote(NT_AMD_HSA_PRODUCER, &desc)) { return false; }
+      uint32_t desc_size = 0;
+      if (!GetAmdNote(NT_AMD_HSA_PRODUCER, &desc, &desc_size)) { return false; }
       *major = desc->producer_major_version;
       *minor = desc->producer_minor_version;
+      // Bound the attacker-controlled producer_name_size against the descriptor.
+      // See ROCM-26177 finding #3.
+      const size_t name_offset = offsetof(amdgpu_hsa_note_producer_t, producer_name);
+      if (desc_size < name_offset || desc->producer_name_size > desc_size - name_offset) {
```

**[Optional]** Minor style inconsistency: `GetNoteIsa` uses two separate `if` statements with an intermediate `name_room` variable, while this function uses a single combined expression. Both are correct (short-circuit evaluation prevents underflow). Pick one style and apply it uniformly for consistency.

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_hsa_code.cpp
```diff
@@ -542,9 +543,22 @@ namespace code {
...
+      // variable-length name array against the actual descriptor size before
+      // reading, otherwise GetNoteString reads past the descriptor into
+      // adjacent heap memory. See ROCM-26177 finding #3.
+      const size_t name_offset = offsetof(amdgpu_hsa_note_isa_t, vendor_and_architecture_name);
+      if (desc_size < name_offset) { return false; }
+      const size_t name_room = desc_size - name_offset;
+      if (desc->vendor_name_size > name_room) { return false; }
       vendor_name = GetNoteString(desc->vendor_name_size, desc->vendor_and_architecture_name);
-      architecture_name = GetNoteString(desc->architecture_name_size, desc->vendor_and_architecture_name + vendor_name.length() + 1);
+      const size_t arch_offset = vendor_name.length() + 1;
```
cfreeamd, line 582: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** `arch_offset` is derived from `vendor_name.length()` (the returned string, possibly with NUL stripped) rather than `desc->vendor_name_size` directly. For legitimate notes this is fine, but for a crafted note that omits the NUL, `arch_offset` would be 1 byte off. Still safe (the bounds check above prevents OOB reads), but the arch name read would be corrupted for crafted inputs. Was this intentional / pre-existing? A comment would help clarify.

dayatsin-amd:
Addressed review feedback in 4e5cd40:

- **Unreachable `desc_size < name_offset` guard**: Replaced with `static_assert(sizeof(NoteT) > offsetof(...))` on all three note types; `GetAmdNote` already guarantees `desc_size >= sizeof(S)`.
- **Style consistency**: All three consumers now use the shared `detail::IsNoteStringSizeWithinDescriptor` / `IsNoteStringSizeWithinRoom` helpers with the same two-step `name_room` pattern.
- **`arch_offset` question**: Added a comment explaining that the offset follows `GetNoteString`'s parsed length + 1 (NUL when present); well-formed notes include a trailing NUL in `vendor_name_size`, so `arch_offset == vendor_name_size` for legitimate inputs.
- **Unit test**: Added `test_amd_hsa_note_name_bounds.cpp` exercising the bounds helpers (satisfies therock-pr-bot).

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_hsa_code.cpp
dayatsin-amd: Agreed — replaced the runtime guard with static_assert in 7e36dbe.

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_hsa_code.cpp
dayatsin-amd: Done in 7e36dbe — all three consumers now use the same `detail::IsNoteStringSizeWithinDescriptor` helper (and `IsNoteStringSizeWithinRoom` for the ISA arch name). `GetNoteProducer`/`GetNoteProducerOptions` match the two-step `name_offset` → bounds-check style used in `GetNoteIsa`.

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_hsa_code.cpp
dayatsin-amd: Yes, pre-existing behavior — `AddNoteIsa` always writes a NUL-terminated vendor name with `vendor_name_size = len+1`, so legitimate notes have `arch_offset == vendor_name_size`. For crafted notes missing the NUL, the bounds check still prevents OOB reads but the arch name may parse incorrectly; documented in the comment above `arch_offset` (7e36dbe).

dayatsin-amd:
@cfreeamd All three review threads addressed in `7e36dbe`:

1. **Unreachable offset guard** → `static_assert(sizeof(NoteT) > offsetof(...))` on all three note types.
2. **Style consistency** → shared `detail::IsNoteStringSizeWithinDescriptor` / `IsNoteStringSizeWithinRoom` helpers used uniformly.
3. **`arch_offset` semantics** → comment added; pre-existing `GetNoteString` NUL-stripping behavior, safe for well-formed notes.

Also inlined the helpers into `amd_hsa_code.cpp` and dropped the separate header/test per follow-up feedback.

cfreeamd (approved):

PR #7544, opened 2026-06-19
Add RCCL_MAX_PAT_NCHANNELS to cap PAT channel count

wenkaidu:
## Motivation

PAT peer-to-peer connections follow a recursive halving-doubling pattern. It can unnecessarily use a lot of connection, up to NCC_MAX_NCHANNELS, if not capped. 

## Technical Details

Introduce the RCCL_MAX_PAT_NCHANNELS env var (default 4) to bound the number of channels used by the PAT algorithm, applied before the existing MIN/MAX_NCHANNELS clamps. Set to <= 0 to disable the cap.

## JIRA ID

AICOMRCCL-957

## Test Plan

Test PAT connections/QPs usage

## Test Result

Connections/QPs match expectation

## Submission Checklist

### DIFF
--- a/projects/rccl/src/enqueue.cc
+++ b/projects/rccl/src/enqueue.cc
@@ -1081,6 +1081,10 @@ NCCL_PARAM(P2pLLThreshold, "P2P_LL_THRESHOLD", 8192);
 RCCL_PARAM(P2pNetThreshold, "P2P_NET_THRESHOLD", 131072);
 NCCL_PARAM(ChunkSize, "CHUNK_SIZE", 0);
 
+// Upper bound on the number of channels used by the PAT algorithm.
+// Set to <= 0 to disable the cap.
+RCCL_PARAM(MaxPatNchannels, "MAX_PAT_NCHANNELS", 4);
+
 
 // Need this temporary parameter to disable p2p batching to avoid some dips at 4MB - 32 MB message size at large scale
 // This parameter must be removed after further investigation,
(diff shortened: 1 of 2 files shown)


projects/rccl/src/enqueue.cc
```diff
@@ -2445,6 +2449,11 @@ static ncclResult_t topoGetAlgoInfo(
   } else if (info->algorithm == NCCL_ALGO_PAT) {
     nc = (nBytes <= (32 << 10)) ? 1 : (nBytes <= (64 << 10)) ? 2 : (nBytes <= (1 << 20)) ? 4 : comm->nChannels;
 
+    int maxPatNChannels = rcclParamMaxPatNchannels();
+    if (maxPatNChannels > 0) {
+      nc = std::min(maxPatNChannels, nc);
```
nusislam, line 2454: PAT QPs are created during communicators creation. How is this helping with QP minimization?

srirajpaul: It seems to change usage and not creation. The QP creation is happening from here ncclTransportPatConnect (https://github.com/ROCm/rocm-systems/blob/develop/projects/rccl/src/transport/generic.cc#L70) . it is controlled by comm->nChannels which is common for ncclTransportRingConnect and ncclTransportTreeConnect. One way would be add a pat channels variable inside comm and use that inside ncclTransportPatConnect so that ring and tree are untouched.

The comm->nChannels is being set inside ncclTopoPostset (https://github.com/ROCm/rocm-systems/blob/develop/projects/rccl/src/graph/connect.cc#L1057) and this is where we seem to be diverging from nccl. There is a lot of rccl specific tuning from https://github.com/ROCm/rocm-systems/blob/develop/projects/rccl/src/graph/connect.cc#L1133-L1320. At the beginning of L1133 we have the nChannels as same as nccl. but at L1320, we have nChannels with a higher value than nccl. Probably around there we can add the RCCL_MAX_PAT_NCHANNELS and set some new variable comm->patChannels. 
But again this is only the creation part, and as Wenkai did, we many also have to find all places where pat code path uses comm->nChannels and change it to use comm->patChannels.

wenkaidu: fixed

PR #7547, opened 2026-06-20
[rocprofiler-systems] Report commit SHA in version banner when not on an exact tag

dgaliffiAMD:
## Motivation

When rocprofiler-systems starts it prints a version banner, e.g.:

```
rocprof-sys v1.3.0 (tag: v1.3.0, x86_64-linux-gnu, compiler: GNU v11.4.0, rocm: v7.2.x)
```

The `tag: v1.3.0` shown here does not exist in the repository. Builds that are not on an exact tagged commit reported a non-existent or stale tag name, making it impossible to identify the exact source used to produce the binary.

## Technical Details

The banner derived its `tag:` field from `git describe --tags` (nearest tag + offset) and fell back to a fabricated `v${VERSION}` when no `.git` directory was found under `PROJECT_SOURCE_DIR`. In the rocm-systems monorepo, `.git` lives at the repo root rather than the project subdirectory, so this **always** hit the fabricated fallback.

Changes in `CMakeLists.txt`:
- Detect the git checkout via `git` itself (using `WORKING_DIRECTORY`) instead of checking for a literal `.git` directory, so it works when the source is a subdirectory of a larger repo.
- Record a tag only when HEAD is an **exact** tag match (`git describe --tags --exact-match`); otherwise leave it empty so the banner reports the full commit SHA (`ROCPROFSYS_GIT_REVISION`) instead.
- Honor `ROCPROFSYS_GIT_DESCRIBE` / `ROCPROFSYS_GIT_REVISION` when provided on the command line, for tarball/CI builds with no checkout.

Resulting behavior:
- On an exact tag → `rocprof-sys v1.3.0 (tag: v1.3.0, ...)` (unchanged, correct).
- Non-tagged / dev / CI build → `rocprof-sys v1.7.0 (rev: <full-sha>, ...)` — no more fabricated tag.

## JIRA ID

Resolves AIPROFSYST-571

## Test Plan

- Reconfigured CMake in the monorepo checkout and inspected the computed git variables and the generated `defines.h`.
- Built the project to confirm it compiles and links.

## Test Result

- `git revision` populated with the full HEAD SHA; `git describe` empty (no fabricated `v1.7.0`).
- Generated `defines.h`: `ROCPROFSYS_GIT_REVISION` = full SHA, `ROCPROFSYS_GIT_DESCRIBE` = "".
- Core library builds and links cleanly.

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-systems/CMakeLists.txt
+++ b/projects/rocprofiler-systems/CMakeLists.txt
@@ -51,31 +51,47 @@ set(BINARY_NAME_PREFIX "rocprof-sys")
 
 find_package(Git)
 
-if(Git_FOUND AND EXISTS "${PROJECT_SOURCE_DIR}/.git")
+# ROCPROFSYS_GIT_DESCRIBE / ROCPROFSYS_GIT_REVISION may be provided on the command
+# line (e.g. tarball/CI builds with no git checkout); honor those if set.
+if(
+    NOT DEFINED ROCPROFSYS_GIT_DESCRIBE
+    AND NOT DEFINED ROCPROFSYS_GIT_REVISION
+    AND Git_FOUND
+)
+    # detect a git checkout via git itself rather than a literal .git directory so
+    # this works when the source is a subdirectory/submodule of a larger repo (the
+    # rocm-systems monorepo) where .git is not under PROJECT_SOURCE_DIR
     execute_process(
-        COMMAND ${GIT_EXECUTABLE} describe --tags
-        OUTPUT_VARIABLE ROCPROFSYS_GIT_DESCRIBE
+        COMMAND ${GIT_EXECUTABLE} rev-parse HEAD
+        WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+        OUTPUT_VARIABLE ROCPROFSYS_GIT_REVISION
         OUTPUT_STRIP_TRAILING_WHITESPACE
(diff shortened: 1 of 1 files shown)


projects/rocprofiler-systems/CMakeLists.txt
```diff
@@ -51,31 +51,47 @@ set(BINARY_NAME_PREFIX "rocprof-sys")
 
 find_package(Git)
 
-if(Git_FOUND AND EXISTS "${PROJECT_SOURCE_DIR}/.git")
+# ROCPROFSYS_GIT_DESCRIBE / ROCPROFSYS_GIT_REVISION may be provided on the command
+# line (e.g. tarball/CI builds with no git checkout); honor those if set.
+if(
+    NOT DEFINED ROCPROFSYS_GIT_DESCRIBE
```
sputhala-amd, line 57: The guard requires both to be undefined, is it possible that one is set and the other is not from the command line?

sputhala-amd (approved):
Other than one question regarding expectation around setting either or both of the variables, the change looks good to me.

projects/rocprofiler-systems/CMakeLists.txt
dgaliffiAMD: Both `rocprofsys_git_describe` and `rocprofsys_git_revision` are overrides used for test builds. If either are set, then that is what will be used in the banner. However, if neither is set, then we'll run `git` to query the data from the repository.

PR #7300, opened 2026-06-16
Support gfx12-5 generic code objects in ROCR

stellaraccident:
Changes:
- Add LLVM gfx1251 and gfx12-5-generic ELF machine values.
- Register gfx125x concrete and generic ISA variants with SRAM ECC/XNACK support.
- Update bootstrap target metadata for the current LLVM gfx12.5 generic target.

Testing:
- This was tested in a downstream hrx-system unit test suite which compiles code objects as gfx-12-5-generic. There appears to be non rocr-level test infrastructure for any of these categories of changes.

PR #4975, opened 2026-04-13
Add support for gfx1250

dayatsin-amd:
## Motivation

Add support for gfx1250 GPU

## Test Plan
Tested that these patches pass ROCRtst and HIP catch tests on existing GPUs.

## Test Result
Pass

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #7548, opened 2026-06-20
Fix ISA feature matching for gfx1250

ammarwa:
## Summary
- fix ISA lookup for gfx1250/gfx1251 entries registered with `IsaFeature::Any`
- add a defensive runtime error if GPU agent creation cannot resolve an ISA variant
- validated the patched runtime on a gfx1250 system with `rocminfo` and the shipped HIP empty-kernel gate

## Regression Source
This is a follow-up fix for the regression introduced by #7300.

PR #4975 originally added `gfx1250` to the ISA registry as:

```cpp
ISAREG_ENTRY_GEN("gfx1250", 12, 5, 0, unsupported, unsupported, 32, "gfx12-generic")
```

That shape worked with the existing feature-qualified lookup because `Unsupported` was already treated as a match for concrete runtime feature requests.

PR #7300 then changed the registration to support gfx12.5 generic code objects:

```cpp
ISAREG_ENTRY_GEN("gfx1250", 12, 5, 0, any, any, 32, "gfx12-5-generic")
ISAREG_ENTRY_GEN("gfx1251", 12, 5, 1, any, any, 32, "gfx12-5-generic")
```

That exposed a pre-existing mismatch in `IsaRegistry::GetIsa(version, sramecc, xnack)`: the lookup accepted `Unsupported` or exact equality only. It did not treat a registered `Any` feature as compatible with a concrete requested feature state.

On some systems, agent creation requests a concrete feature state, for example disabled/disabled. The registered `gfx1250` entry is now any/any, so the lookup returned `nullptr`. `GpuAgent::GpuAgent` then assumed `isa_` was valid and dereferenced it during `hsa_init`, causing `rocminfo` and HIP startup to segfault in `libhsa-runtime64.so.1`.

## Fix
This PR updates the feature-qualified ISA lookup so registered wildcard-style feature states match concrete runtime requests:

- `Unsupported` still matches, preserving existing behavior.
- `Any` now also matches a concrete requested feature state.
- Exact feature matches continue to work as before.
- The requested feature state is not treated as a wildcard, which avoids accidentally selecting an arbitrary enabled/disabled-specific ISA when the caller asks for `Any`.

It also replaces the release-build-only `assert(isa_ != nullptr)` with a runtime guard that raises `HSA_STATUS_ERROR_INVALID_ISA`. That prevents future ISA table mismatches from becoming a null-pointer crash.

## gfx1250 Validation
Validated using the TheRock gfx1250 build that previously failed.

Artifacts/logs are preserved under:

Validation results:

- Built patched `hsa-runtime64` from this PR branch at commit `55f6578b18bafb40888b97fe6df04e9771ed18df`.
- `ldd` confirmed the test binaries loaded the patched `libhsa-runtime64.so.1` from the PR build.
- `rocminfo` using the ROCm `rocminfo` binary exited `0` and reported `4` `gfx1250` agents.
- Shipped HIP gate passed: `KernelTest Unit_hipEmptyKernel`, `ROCR_VISIBLE_DEVICES=0`, `1 test case passed`.
- No new `rocminfo`/HIP segfault appeared in the post-run `dmesg` tail.

## Earlier Local Checks
- `git diff --check`
- `custom-ai-secret-scan --repo-root .codex_tmp/rocm-systems-develop`

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
@@ -197,7 +197,10 @@ GpuAgent::GpuAgent(HSAuint32 node, const HsaNodeProperties& node_props, bool xna
                              node_props.EngineId.ui32.Stepping), sramecc, xnack);
   }
 
-  assert(isa_ != nullptr && "ISA registry inconsistency.");
+  if (!isa_) {
+    throw AMD::hsa_exception(HSA_STATUS_ERROR_INVALID_ISA,
+                             "Agent creation failed.\nNo ISA variant matched requested feature state.\n");
+  }
 
   supported_isas_.push_back(isa_);
   if (!isa_->GetIsaGeneric().empty()) {
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/isa.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/isa.cpp
@@ -250,14 +250,17 @@ const Isa *IsaRegistry::GetIsa(const std::string &full_name) {
 }
 
 const Isa *IsaRegistry::GetIsa(const Isa::Version &version, IsaFeature sramecc, IsaFeature xnack) {
+  auto feature_matches = [](IsaFeature registered, IsaFeature requested) {
+    return registered == IsaFeature::Unsupported ||
+           registered == IsaFeature::Any ||
+           registered == requested;
+  };
   auto isareg_iter = std::find_if(GetSupportedIsas().begin(),
                                   GetSupportedIsas().end(),
                                   [&](const IsaMap::value_type& isareg) {
                                     return isareg.second.GetVersion() == version &&
-                                        (isareg.second.GetSramecc() == IsaFeature::Unsupported ||
-                                         isareg.second.GetSramecc() == sramecc) &&
-                                        (isareg.second.GetXnack() == IsaFeature::Unsupported ||
-                                         isareg.second.GetXnack() == xnack);
+                                        feature_matches(isareg.second.GetSramecc(), sramecc) &&
+                                        feature_matches(isareg.second.GetXnack(), xnack);
                                   });
   return isareg_iter == GetSupportedIsas().end() ?
                                               nullptr : &isareg_iter->second;


cfreeamd:
I'm still digesting what it's telling me (isa compatibility can get confusing), but I had Claude do a review. I have found it's reviews to be pretty good.  I will continue to look at it but thought I would put it out so others can evaluate as well:

> Background

  This PR fixes a segfault in hsa_init on gfx1250. PR #7300 changed the gfx1250/1251 ISA registry entries from
  Unsupported/Unsupported to Any/Any. The old GetIsa(version, sramecc, xnack) predicate only matched if the registered
   feature was Unsupported or exactly equal to the requested value — it didn't handle Any as a wildcard. So
  GetIsa(12,5,0, Disabled, Disabled) returned nullptr, and the subsequent assert(isa_ != nullptr) either aborted
  (debug) or null-deref'd (release).

  The fix is in the right place. But it introduces a new problem for gfx9xx GPUs. Request changes.

  ---
  Issues

  *Critical* — gfx9xx non-determinism in IsaRegistry::GetIsa (isa.cpp:253)

  Confirmed in code by Opus.

  The ISA registry is an unordered_map<string, Isa>. For gfx906, gfx908, gfx90a, gfx942, and gfx950, the registry
  contains both a bare Any/Any catch-all entry and feature-specific entries (e.g. gfx906:sramecc-:xnack-). Confirmed
  at isa.cpp:377–425.

  The GpuAgent constructor always calls the second GetIsa with concrete Disabled/Enabled values (confirmed:
  IsSrameccSupported() returns true for Any, so the concrete-feature branch is taken — isa.h:153,158).

  Under the old predicate: Any != Disabled and Any != Unsupported, so the bare entry never matched. find_if always
  found the feature-specific entry. Deterministic.

  Under the new feature_matches lambda: registered==Any → true, so both the bare gfx906 (Any/Any) entry and the
  feature-specific gfx906:sramecc-:xnack- entry satisfy the predicate. std::find_if returns whichever appears first in
   bucket traversal order — non-deterministic across stdlib versions and container sizes.
  Impact: If the bare entry wins, isa_->GetIsaName() returns "amdgcn-amd-amdhsa--gfx906" instead of
  "amdgcn-amd-amdhsa--gfx906:sramecc-:xnack-". Code objects compiled as gfx906:sramecc-:xnack- fail IsCompatible —
  silent code object rejection, no GPU dispatch.

  *Fix* — two-pass lookup (specificity-first):

  const Isa *IsaRegistry::GetIsa(const Isa::Version &version,
                                  IsaFeature sramecc, IsaFeature xnack) {
    // Pass 1: original semantics — Unsupported or exact match (preserves all pre-PR behavior)
    auto specific = std::find_if(GetSupportedIsas().begin(), GetSupportedIsas().end(),
      [&](const IsaMap::value_type& e) {
        return e.second.GetVersion() == version &&
               (e.second.GetSramecc() == IsaFeature::Unsupported || e.second.GetSramecc() == sramecc) &&
               (e.second.GetXnack()   == IsaFeature::Unsupported || e.second.GetXnack()   == xnack);
      });
    if (specific != GetSupportedIsas().end()) return &specific->second;

    // Pass 2: wildcard fallback — for ISAs with only an Any/Any entry (e.g. gfx1250)
    auto wildcard = std::find_if(GetSupportedIsas().begin(), GetSupportedIsas().end(),
      [&](const IsaMap::value_type& e) {
        return e.second.GetVersion() == version &&
               (e.second.GetSramecc() == IsaFeature::Unsupported || e.second.GetSramecc() == IsaFeature::Any || e.second.GetSramecc() == sramecc) &&
               (e.second.GetXnack()   == IsaFeature::Unsupported || e.second.GetXnack()   == IsaFeature::Any || e.second.GetXnack()   == xnack);
      });
    return wildcard == GetSupportedIsas().end() ? nullptr : &wildcard->second;
  }

  Opus confirmed: pass 1 is identical to the old predicate — every previously-matched ISA resolves identically. Only
  Any/Any-only parts (gfx1250/1251) newly resolve via the wildcard pass. Non-regressing.

 *Major* — IsCompatible still rejects gfx1250 code objects (isa.cpp:89–100)

  Even with the two-pass fix, gfx1250's isa_ points to the Any/Any registry entry. IsCompatible expects
  agent_isa.GetSramecc() != Any (there's a debug assert). In release builds, the comparison
  code_object_isa.GetSramecc() == agent_isa.GetSramecc() becomes Disabled == Any → false → code object rejected.

  Impact: The PR prevents the startup segfault but doesn't fully fix the feature — gfx1250 may fail to load code
  objects in release builds, or assert-abort in debug builds. Opus confirmed this is a gap the PR description doesn't
  address.

  *Fix options*:
  1. Update IsCompatible to treat agent_isa.GetSramecc() == Any as "match anything":
  // in IsCompatible, where it checks sramecc:
  if (agent_isa.GetSramecc() != IsaFeature::Any &&
      agent_isa.GetSramecc() != code_object_isa.GetSramecc()) return false;
  2. Or register explicit gfx1250:sramecc-:xnack- / gfx1250:sramecc+:xnack- entries alongside the Any/Any catch-all,
  so the lookup returns a feature-concrete entry for gfx1250 agents too.

  ---
  *Minor* — silent GPU disappearance on ISA mismatch (amd_topology.cpp:195)

  The throw AMD::hsa_exception(HSA_STATUS_ERROR_INVALID_ISA, ...) that replaced the assert causes DiscoverGpu to skip
  the agent silently in release builds — the warning is ifdebug-gated. A future registry misconfiguration would
  produce a system with zero visible GPUs and no diagnostic. The error message should be logged unconditionally (or
  surfaced through hsa_init's return code).

  ---
  Nit — no unit test for GetIsa with Any features

This function has now broken in two back-to-back PRs. A table-driven unit test covering Unsupported/exact/Any/fallback combinations would make the next breakage immediate and obvious.

cfreeamd:
I reviewed this more. I think the Claude analysis and suggestions are good for the Critical, Major and Minor issues flagged.

PR #6998, opened 2026-06-09
Validate ELF segment offsets before copying to GPU memory.

dayatsin-amd:
Reject program headers with out-of-bounds p_offset or p_filesz relative to the code object buffer before segment data is copied into allocations.

Resolves SWSPLAT-24378

PR #6996, opened 2026-06-09
Fix segment overflow when p_filesz exceeds p_memsz during load.

dayatsin-amd:
Reject crafted ELF program headers where the file size exceeds the memory size before copying segment data into GPU allocations.

Resolves SWSPLAT-24384

PR #6995, opened 2026-06-09
fix(rocr): Fix out-of-bounds read in ElfSize() for ELF objects

dayatsin-amd:
## Summary

Validate section table offsets and sizes against the known buffer size, require explicit buffer sizes for in-memory ELF initialization, and propagate code object sizes through the loader API to prevent parsing beyond the end of untrusted .hsaco input.

## JIRA ID

JIRA ID : SWSPLAT-24382

## Test plan

- [ ] Load a normal code object; confirm no regression.
- [ ] Load a crafted code object with truncated section table; confirm rejection instead of out-of-bounds read.

PR #6994, opened 2026-06-09
fix(rocr): Harden code-object loader against memory safety defects

dayatsin-amd:
## Summary

Fix memory-safety vulnerabilities detected in the ROCr code-object loader when parsing untrusted ELF/HSA code objects, including out-of-bounds reads and writes in segment loading, note parsing, and relocation application paths.

## JIRA ID

JIRA ID : ROCM-26177

## Test plan

- [ ] Load normal code objects and run existing kernels; confirm no regression.
- [ ] Load crafted malicious code objects; confirm clean rejection with HSA_STATUS_ERROR_INVALID_CODE_OBJECT.

PR #7552, opened 2026-06-20
fix(rocr): Bound code-object relocation writes and reject null segments

dayatsin-amd:
## Summary

Continuation of the ROCm code-object loader hardening work (untrusted .hsaco / code-object parsing). This addresses a vulnerability in the relocation application path not covered by existing segment-load / ELF-parser fixes (#6994, #6995, #6996, #6998, #7541, #7542).

`ExecutableImpl::ApplyStaticRelocation()` and `ApplyDynamicRelocation()` write resolved addresses into a loaded segment at offsets taken from attacker-controlled code-object fields without validating the destination range. In addition, null segment pointers were dereferenced unconditionally.

Fix:
- Add an overflow-safe `Segment::IsAddressInSegment(addr, size)` range check.
- `Segment::Copy()` validates the entire `[addr, addr + size)` destination.
- Reject null target/symbol segments and treat refused copy as `HSA_STATUS_ERROR_INVALID_CODE_OBJECT`.

## JIRA ID

JIRA ID : ROCM-26177

## Test plan

- [x] `g++ -fsyntax-only` of `loader/executable.cpp` passes (no new warnings).
- [ ] Load a normal code object / run an existing kernel; confirm no regression.
- [ ] Load a crafted code object with an out-of-range relocation offset; confirm rejection with HSA_STATUS_ERROR_INVALID_CODE_OBJECT.

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
@@ -790,14 +790,31 @@ bool Segment::IsAddressInSegment(uint64_t addr)
   return vaddr <= addr && addr < vaddr + size;
 }
 
-void Segment::Copy(uint64_t addr, const void* src, size_t size)
+bool Segment::IsAddressInSegment(uint64_t addr, size_t copy_size)
+{
+  if (addr < vaddr) { return false; }
+  const uint64_t offset = addr - vaddr;
+  // Overflow-safe check of offset + copy_size <= size.
+  return offset <= size && copy_size <= size - offset;
+}
+
+bool Segment::Copy(uint64_t addr, const void* src, size_t size)
 {
   // loader must do copies before freezing.
   assert(!frozen);
 
   if (size > 0) {
+    // addr/size can be derived from attacker-controlled code-object fields
+    // (e.g. a relocation's section sh_addr + r_offset). Offset() only asserts
+    // the start address via IsAddressInSegment() and that assert compiles out
+    // under NDEBUG, so validate the entire destination range here to prevent a
+    // heap out-of-bounds write when loading a crafted code object.
+    if (!IsAddressInSegment(addr, size)) {
+      return false;
+    }
     owner->context()->SegmentCopy(segment, agent, ptr, Offset(addr), src, size);
   }
+  return true;
 }
 
 void Segment::Print(std::ostream& out)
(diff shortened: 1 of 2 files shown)


projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
```diff
@@ -1828,14 +1848,14 @@ hsa_status_t ExecutableImpl::ApplyStaticRelocation(hsa_agent_t agent, amd::hsa::
       switch (rel->type()) {
         case R_AMDGPU_V1_32_HIGH:
           addr32 = uint32_t((addr >> 32) & 0xFFFFFFFF);
-          rseg->Copy(reladdr, &addr32, sizeof(addr32));
+          if (!rseg->Copy(reladdr, &addr32, sizeof(addr32))) { return HSA_STATUS_ERROR_INVALID_CODE_OBJECT; }
```
cfreeamd, line 1948: Do we need to handle lines 1483, 1501 and 1683 as we are handling these rseg->Copy() lines? If safe, maybe we should add a comment.

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
```diff
@@ -1791,6 +1808,9 @@ hsa_status_t ExecutableImpl::ApplyStaticRelocation(hsa_agent_t agent, amd::hsa::
   code::RelocationSection* rsec = rel->section();
   code::Section* sec = rsec->targetSection();
   Segment* rseg = SectionSegment(agent, sec);
+  // SectionSegment() returns nullptr when no loaded segment covers the target
+  // section (crafted sh_info/sh_addr); reject rather than dereferencing it.
+  if (!rseg) { return HSA_STATUS_ERROR_INVALID_CODE_OBJECT; }
   size_t reladdr = sec->addr() + rel->offset();
```
cfreeamd, line 1911: Line 1814: "sums two attacker-controlled values and can wrap, but it's safe because the overflow-safe range check rejects  the wrapped value. Non-obvious; worth a one-line comment."...Claude. Your choice--not a blocker.

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
```diff
@@ -767,14 +767,31 @@ bool Segment::IsAddressInSegment(uint64_t addr)
   return vaddr <= addr && addr < vaddr + size;
 }
 
-void Segment::Copy(uint64_t addr, const void* src, size_t size)
+bool Segment::IsAddressInSegment(uint64_t addr, size_t copy_size)
+{
+  if (addr < vaddr) { return false; }
+  const uint64_t offset = addr - vaddr;
+  // Overflow-safe check of offset + copy_size <= size.
+  return offset <= size && copy_size <= size - offset;
```
cfreeamd: The "<=" in "copy_size <= size - offset" inconsistent with the existing 1-argument version of the function which uses just "<"

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
dayatsin-amd: Added comments

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
dayatsin-amd: Added comment

cfreeamd (approved):

PR #7553, opened 2026-06-20
[hipfile] Make clang-tidy warning-free and add a CI gate

derobins:
## Motivation

Resolve every clang-tidy finding in hipFile and add a workflow that keeps the project clean going forward. With AISClangTidy.cmake passing -warnings-as-errors=*, a clean build now proves there are no remaining diagnostics.

## Technical Details

Source/build fixes:
- examples/async: reorder slice_state fields in the two multi-stream examples to group trailing bools, eliminating excessive struct padding (optin.performance.Padding).
- examples/async: drop dead stores in the close_file error branches of all three roundtrip examples; the assignments were never read because the goto skips the cleanup block that reads them (deadcode.DeadStores).
- cmake/AISUseGTest.cmake: mark GoogleTest include dirs as SYSTEM so clang-tidy suppresses diagnostics from its headers on every path (the SYSTEM keyword on FetchContent_Declare only applies to CMake >= 3.25, and a system find_package(GTest) may not mark them at all).
- test/stats.cpp: NOLINT the two intentional out-of-range enum casts that verify input rejection (optin.core.EnumCastOutOfRange).
- test/async.cpp: NOLINT two malloc sites whose ownership transfers to AsyncOpFallback and is freed via the mocked hipHostFree during async cleanup; verified leak-free under ASAN (unix.Malloc false positive).

CI:
- Add hipfile-clang-tidy.yml, a reusable workflow that compiles hipFile with -DAIS_USE_CLANG_TIDY=ON inside the CI container (no tests; we only care about clang-tidy diagnostics).
- Wire it into hipfile-ci-toplevel.yml as a run-once job, pinned to a single fixed image rather than fanning out across the build matrix since findings are compiler- and standard-independent.

## JIRA ID

AIHIPFILE-161

## Test Plan

CI passes with no clang-tidy issues

## Test Result

Passes

## Submission Checklist

### DIFF
--- a/.github/workflows/hipfile-ci-toplevel.yml
+++ b/.github/workflows/hipfile-ci-toplevel.yml
@@ -273,3 +273,20 @@ jobs:
     uses: ./.github/workflows/hipfile-sanitizers.yml
     with:
       ci_image: ${{ fromJSON(needs.Precheck.outputs.ci_images)['ubuntu-7.2.2'].ci_image }}
+
+  # clang-tidy findings are compiler- and standard-independent, so run the
+  # analysis exactly once against a single fixed image rather than fanning out
+  # across the build matrix. Change the image key below to retarget the ROCm
+  # version (the combo must exist in Precheck's matrix).
+  clang_tidy:
+    if: >-
+      ${{
+        !cancelled() && (
+          needs.build_CI_image.result == 'success' ||
+          needs.build_CI_image.result == 'skipped'
+        )
+      }}
+    needs: [Precheck, build_CI_image]
+    uses: ./.github/workflows/hipfile-clang-tidy.yml
+    with:
+      ci_image: ${{ fromJSON(needs.Precheck.outputs.ci_images)['ubuntu-7.2.2'].ci_image }}
--- a/projects/hipfile/examples/async/roundtrip-async-multi-stream-registered.cpp
+++ b/projects/hipfile/examples/async/roundtrip-async-multi-stream-registered.cpp
@@ -80,15 +80,15 @@ static_assert((SLICE_SIZE % BLOCK_ALIGN) == 0, "SLICE_SIZE must be a multiple of
  * must outlive submission. */
 struct slice_state {
     void       *devbuf;
-    bool        buf_registered;
     hipStream_t stream;
-    bool        stream_created;
-    bool        stream_registered;
     size_t      io_size;
     hoff_t      file_offset;
     hoff_t      buf_offset;
     ssize_t     bytes_read;
     ssize_t     bytes_written;
+    bool        buf_registered;
+    bool        stream_created;
+    bool        stream_registered;
 };
 
 int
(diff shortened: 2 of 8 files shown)


projects/hipfile/test/amd_detail/async.cpp
```diff
@@ -499,8 +499,10 @@ TEST_P(FallbackAsyncIO, attemptToQueueCleanupOnStreamSubmissionFailure)
     EXPECT_CALL(*mstream, fixedBufferOffset).Times(AnyNumber()).WillRepeatedly(Return(false));
 
     auto op_data = malloc(sizeof(AsyncOpFallback));
+    // NOLINTNEXTLINE(clang-analyzer-unix.Malloc): freed via mocked hipHostFree on cleanup
     ASSERT_NE(op_data, nullptr);
     auto bounce_buffer = malloc(size);
+    // NOLINTNEXTLINE(clang-analyzer-unix.Malloc): freed via mocked hipHostFree on cleanup
     ASSERT_NE(bounce_buffer, nullptr);
```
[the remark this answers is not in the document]
derobins: Copilot is wrong

.github/workflows/hipfile-ci-toplevel.yml
```diff
@@ -229,3 +229,20 @@ jobs:
       cxx_standard: ${{ matrix.cxx_standard }}
       platform: ${{ matrix.supported_platforms }}
       rocm_version: ${{ matrix.rocm_versions }}
+
+  # clang-tidy findings are compiler- and standard-independent, so run the
+  # analysis exactly once against a single fixed image rather than fanning out
+  # across the build matrix. Change the image key below to retarget the ROCm
+  # version (the combo must exist in Precheck's matrix).
```
riley-dixon, line 280: Based on this I would try to see if we could run this in or alongside Precheck. That seems cleaner rather than needing an arbitrary image. It would also mean you don't need to write all of the docker command wrappers for each step.

Clang may be pre-installed on the self-hosted CI image based on https://github.com/actions/runner-images#software-and-image-support

Do you want to use amdclang++ specifically though? That might change my answer since installing that is more involved/not as performant.

projects/hipfile/examples/async/roundtrip-async-multi-stream-registered.cpp
```diff
@@ -234,19 +234,13 @@ main(int argc, char *argv[])
      * file has undefined coherence per open(2); closing first sidesteps that
      * and also flushes file-size metadata for the fresh open() in the hasher.
      * Then hash the full TOTAL_SIZE and compare. */
-    if (close_file(write_path, wfd, whandle)) {
-        wfd          = -1;
-        whandle_open = false;
+    if (close_file(write_path, wfd, whandle))
         goto close_rfile;
```
riley-dixon, line 238: I would double check that the changing logic here is intentional? 
(if we fail to close the write file, skip the close_rfile and we don't need to bother setting wfd or whandle_open?)

kurtmcmillan (approved):
The commit message notes that you made several independent changes. In future PRs could you split each change into its own commit? A few times I wondered why a change was being made and had to re-scan the commit message for the reason.

.github/workflows/hipfile-ci-toplevel.yml
derobins: clang-tidy is a compile-time analyzer, so we need it to build the software and the best compiler for that is amdclang++.

PR #7556, opened 2026-06-20
[hipfile] Add CodeQL C/C++ analysis to CI

derobins:
## Motivation

The repo-wide codeql.yml only scans Python and GitHub Actions; C/C++ is uncovered because CodeQL needs to observe a real compiler build to analyze it. Add a reusable hipfile-codeql.yml that runs init -> build -> analyze inside the prebuilt hipFile CI image (job-level container, not the host + docker exec pattern used by clang-tidy, since the extractor's compiler tracing must share the build environment).

## Technical Details

Wire it into hipfile-ci-toplevel.yml as a single codeql job gated on the CI image, run once against ubuntu-7.2.2 since findings are compiler- and standard-independent. Uses the security-extended,security-and-quality query suite and a single gfx arch (host compilation is traced once; extra device-code targets add nothing for the C/C++ extractor).

## JIRA ID

AIHIPFILE-161

## Test Plan

CodeQL CI passes

## Test Result

Passes

## Submission Checklist

### DIFF
--- a/.github/workflows/hipfile-codeql.yml
+++ b/.github/workflows/hipfile-codeql.yml
@@ -1,0 +1,85 @@
+# Copyright (c) Advanced Micro Devices, Inc. All rights reserved.
+#
+# SPDX-License-Identifier: MIT
+name: hipfile-codeql
+run-name: Run CodeQL C/C++ analysis on hipFile
+
+# CodeQL's C/C++ extractor works by tracing the real compiler invocations, so
+# init -> build -> analyze must all run in the SAME environment. We therefore
+# run the whole job inside the prebuilt hipFile CI image (job-level container),
+# rather than the host-runner + `docker exec` pattern used by hipfile-clang-tidy
+# (the tracer's environment would not reach a compiler launched in a separate
+# `docker exec` process).
+#
+# Python and GitHub Actions are already covered repo-wide by codeql.yml; this
+# workflow exists to close the C/C++ gap, which needs an observed build.
+
+env:
+  # A single arch is enough: CodeQL traces the host compilation (done once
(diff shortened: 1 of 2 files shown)


.github/workflows/hipfile-codeql.yml
```diff
@@ -0,0 +1,85 @@
...
+    container:
+      image: ${{ inputs.ci_image }}
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Fetching code repository...
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          fetch-depth: 0
```
john00003, line 52: Why `fetch-depth: 0`? It clones the entire git history as opposed to just the head, see [here](https://github.com/actions/checkout#checkout-v4). IDK if this makes a measurable difference or not.  I see in some places in the repo we use the same convention, e.g., 
https://github.com/ROCm/rocm-systems/blob/6ca46b252eb18ffba40a15622777df291a4675eb/.github/workflows/hipfile-build.yml#L107-L114

But in other places we don't:
https://github.com/ROCm/rocm-systems/blob/6ca46b252eb18ffba40a15622777df291a4675eb/.github/workflows/hipfile-codespell.yml#L22-L30

john00003 (approved):
LGTM but I am not the CI guy

riley-dixon (approved):

.github/workflows/hipfile-codeql.yml
derobins: We'll look into it for a future PR

gaoikawa (approved):

PR #7564, opened 2026-06-21
[hipfile] Use move semantics in a few places

derobins:
## Motivation

Coverity (currently down) flagged a few places where we could use move semantics to avoid refcount bumps.

## Technical Details

- Backend::score — pass by const ref. score() inspected its file and buffer arguments but never retained them, yet took them by value. Switched to const&, updating the base virtual, both the fastpath and fallback overrides, and the gmock mock in mbackend.h to keep signatures in sync.
- BatchOperation constructor — move sink params. The constructor copied its buffer and file parameters into members. These are pass-by-value sink parameters, so they are now std::moved into the members, matching the existing pattern in the AsyncOp and AsyncOpFallback constructors.
- hipFileIo — move into the final io() dispatch. The selected backend's io() call received file and buffer as lvalues even though it was their last use. Since the io()/_io_impl chain is already a by-value move-sink, they are now std::moved in, extending the move-only path all the way down to the syscall leaf.

## JIRA ID

AIHIPFILE-161

## Test Plan

CI Passes, manual ASAN verification

## Test Result

Success

## Submission Checklist

### DIFF
--- a/projects/hipfile/src/amd_detail/backend.h
+++ b/projects/hipfile/src/amd_detail/backend.h
@@ -73,7 +73,7 @@ struct Backend {
     /// @param file_offset    Offset from the start of the file
     /// @param buffer_offset  Offset from the start of the buffer
     /// @return The eagerness "score"
-    virtual int score(std::shared_ptr<IFile> file, std::shared_ptr<IBuffer> buffer, size_t size,
+    virtual int score(const std::shared_ptr<IFile> &file, const std::shared_ptr<IBuffer> &buffer, size_t size,
                       hoff_t file_offset, hoff_t buffer_offset) const = 0;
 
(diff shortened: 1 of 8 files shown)


john00003 (approved):
Is my understanding correct?

1) score doesn't claim ownership over file or buffer, it's simply dereferenced as soon as the score function is done and the local variables go out of scope, so we avoid incrementing/decrementing the reference count by taking an alias of the original file and buffer.
2) the io() function just calls the internal _io_impl(), similarly no ownership, so don't touch reference count and instead move file and buffer.
3) BatchOperation constructor already takes _buffer and _file passed by value, so we move the local copies into the internal class fields.

If so, looks good!

jordan-turbofish (approved):

PR #7566, opened 2026-06-21
[hipfile] Add ASAN and TSAN unit-test CI

derobins:
## Motivation

hipFile needs to be sanitizer clean

## Technical Details

Add a reusable hipfile-sanitizers.yml workflow that builds the unit tests with sanitizers enabled and runs them via `ctest -L unit`. ASAN (AIS_USE_SANITIZERS, covering address/leak/UB/integer/etc.) and TSAN (AIS_USE_THREAD_SANITIZER) run as two legs of a matrix; they are mutually exclusive in CMake. Modeled on the clang-tidy workflow but runs the tests rather than compiling only.

Since the sanitizers are clang-only, the compiler is fixed to amdclang++ and the job runs once on a single image rather than fanning out across the build matrix.

The job runs on the self-hosted AIS runner where the system tests run; that is the controlled host we intend to eventually run the full ASAN/TSAN CI on.

Wire the new workflow into hipfile-ci-toplevel.yml alongside the Linux build matrix, using a fixed image and the same gating.

## JIRA ID

AIHIPFILE-154

## Test Plan

CI passes

## Test Result

Success

## Submission Checklist

### DIFF
--- a/.github/workflows/hipfile-sanitizers.yml
+++ b/.github/workflows/hipfile-sanitizers.yml
@@ -1,0 +1,139 @@
+name: hipfile-sanitizers
+run-name: Run sanitizer-instrumented unit tests on hipFile
+
+env:
+  AIS_INPUT_CI_IMAGE: ${{ inputs.ci_image }}
+  AIS_HIP_ARCHITECTURES: gfx950;gfx1201;gfx1200;gfx1101;gfx1100;gfx1030;gfx942;gfx90a;gfx908
+
+on:
+  workflow_call:
+    inputs:
+      ci_image:
+        description: "Full CI image name & tag"
+        required: true
+        type: string
+
+permissions:
+  contents: read
+  packages: read
+
+jobs:
+  sanitizers:
+    name: ${{ matrix.sanitizer.name }}
+    # Runs on the self-hosted AIS runner where the system tests run. The job is
+    # only compile + unit tests today, but this runner is the controlled host
+    # we intend to eventually run the full ASAN/TSAN CI on.
+    runs-on: [AIS]
+    strategy:
+      fail-fast: false
+      matrix:
+        # ASAN bundles address/leak/UB/integer/etc.; TSAN is mutually exclusive
+        # with it, so each runs as its own matrix leg with the appropriate flag.
+        sanitizer:
+          - name: asan
+            cmake_flag: -DAIS_USE_SANITIZERS=ON
+          - name: tsan
+            cmake_flag: -DAIS_USE_THREAD_SANITIZER=ON
+    steps:
+      - name: Set AIS CI container name
+        # On non-pull_request triggering workflows, AIS_PR_NUMBER may be empty.
+        run: |
+          echo "AIS_PR_NUMBER=$(echo "${GITHUB_REF}" | sed 's|[^0-9]||g')" >> "${GITHUB_ENV}"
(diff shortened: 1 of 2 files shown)


.github/workflows/hipfile-sanitizers.yml
```diff
@@ -0,0 +1,136 @@
...
+          if docker inspect "${AIS_CONTAINER_NAME}" >/dev/null 2>&1; then
+            docker stop "${AIS_CONTAINER_NAME}"
+          else
+            echo "Container ${AIS_CONTAINER_NAME} does not exist; skipping stop"
+          fi
+      - name: Cleanup self-hosted runner workspace
+        if: ${{ always() }}
+        run: |
+          shopt -s dotglob
+          rm -rf ${GITHUB_WORKSPACE}/*
```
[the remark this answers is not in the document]
derobins: The other system test workflows also do this. I will fix this here and set up a separate PR to harden the system tests.

derobins:
The TSAN action will not pass until the entropy has been updated on the self-hosted runner

.github/workflows/hipfile-sanitizers.yml
```diff
@@ -0,0 +1,139 @@
...
+  contents: read
+  packages: read
+
+jobs:
+  sanitizers:
+    name: ${{ matrix.sanitizer.name }}
+    # Runs on the self-hosted AIS runner where the system tests run. The job is
+    # only compile + unit tests today, but this runner is the controlled host
+    # we intend to eventually run the full ASAN/TSAN CI on.
+    runs-on: [AIS]
```
riley-dixon: Is there a reason why this needs to run a self-hosted runner vs a GitHub runner?

riley-dixon: Seems to be okay for github runners:

ASAN: https://github.com/ROCm/rocm-systems/actions/runs/27975496911/job/82792262033?pr=7628
TSAN: https://github.com/ROCm/rocm-systems/actions/runs/27975496911/job/82792262037?pr=7628

riley-dixon (approved):

.github/workflows/hipfile-sanitizers.yml
```diff
@@ -0,0 +1,139 @@
+name: hipfile-sanitizers
+run-name: Run sanitizer-instrumented unit tests on hipFile
+
+env:
+  AIS_INPUT_CI_IMAGE: ${{ inputs.ci_image }}
+  AIS_HIP_ARCHITECTURES: gfx950;gfx1201;gfx1200;gfx1101;gfx1100;gfx1030;gfx942;gfx90a;gfx908
```
kurtmcmillan, line 6: I've seen this twice now in PRs. @riley-dixon is this something that a top level (?) CI job can pass down into the lower level CI jobs?

derobins: I think we need to do a tidy pass over all the CI

riley-dixon: @kurtmcmillan Yes I think the easy thing would be to pass it through our CI via toplevel.

However, if multiple workflows outside of our main CI need to know about which architectures we should be building for, we should refactor it into a GitHub Action such that we only need to define it in one place. (This could be extended to other areas of CI as well, like enumerating CI images, it just would take some effort to get this right in the first place.)

PR #7568, opened 2026-06-21
refactor(hipfile): Guard system-test workspace cleanup against empty path

derobins:
## Motivation

This is a cleanup fix suggested by Copilot in #7566

## Technical Details

Harden the self-hosted runner cleanup in the system-test workflow: quote the path and add bash's ${VAR:?} guard so an unset or empty GITHUB_WORKSPACE aborts the step instead of letting the glob expand to `/*` and wipe the persistent runner.

## JIRA ID

AIHIPFILE-154

## Test Plan

CI passes

## Test Result

Success

## Submission Checklist

### DIFF
--- a/.github/workflows/hipfile-ci-system.yml
+++ b/.github/workflows/hipfile-ci-system.yml
@@ -230,6 +230,9 @@ jobs:
           fi
       - name: Cleanup self-hosted runner workspace
         if: ${{ always() }}
+        shell: bash
+        # ${VAR:?} aborts if GITHUB_WORKSPACE is unset/empty, so a missing value
+        # can never let the glob expand to `/*` and wipe the persistent runner.
         run: |
           shopt -s dotglob
-          rm -rf ${GITHUB_WORKSPACE}/*
+          rm -rf "${GITHUB_WORKSPACE:?}"/*


.github/workflows/hipfile-ci-system.yml
```diff
@@ -231,9 +231,12 @@ jobs:
           fi
       - name: Cleanup self-hosted runner workspace
         if: ${{ always() }}
+        shell: bash
+        # ${VAR:?} aborts if GITHUB_WORKSPACE is unset/empty, so a missing value
+        # can never let the glob expand to `/*` and wipe the persistent runner.
         run: |
           shopt -s dotglob
-          rm -rf ${GITHUB_WORKSPACE}/*
+          rm -rf "${GITHUB_WORKSPACE:?}"/*
```
john00003, line 238:
```suggestion
          rm -rf "${GITHUB_WORKSPACE:?Environment variable GITHUB_WORKSPACE not found, exiting to avoid nuking the computer!}"/*
```

~~Maybe? To make it easy to understand why this is happening if it ever does.~~

Nevermind, the default error message is `parameter null or unset`, so that should be fine.

john00003 (approved):

riley-dixon (approved):
If `${GITHUB_WORKSPACE}` is missing or malformed we have bigger issues, but I do agree that this kind of check is worthwhile.

kurtmcmillan (approved):

PR #7571, opened 2026-06-21
[hipfile] Fix IWYU issues and add CI action

derobins:
## Motivation

We should include only what we use. This makes the code more robust and avoids wasted CI time as the compiler parses unnecessary header files.

## Technical Details

Run include-what-you-use over hipFile and apply the verified findings, add mapping files to suppress structural false positives, and wire an advisory (non-blocking) IWYU job into the top-level CI.

Include fixes (library, tests, examples, tools):
- Add directly-used headers and drop unused/transitive ones across src/amd_detail, test/amd_detail, examples, and tools/ais-stats.
- Prefer C++ header forms (<cstdint>, <cstdio>, <cerrno>) in C++ files.
- Replace full includes with forward declarations in backend.cpp where only pointers/by-value declarations are used.

IWYU configuration (cmake/AISIWYU.cmake):
- Add cmake/iwyu-gtest.imp mapping GoogleTest/GoogleMock private sub-headers to their public umbrella headers, eliminating the spurious quoted-vs-angle gtest.h suggestions and gmock.h remove/re-add churn.
- Add cmake/iwyu-libc.imp mapping libstdc++/glibc private headers (<bits/chrono.h> -> <chrono>, <bits/statx*.h> -> <sys/stat.h>).
- Fail configure with a clear error if a mapping file is missing.

Pragmas for unavoidable false positives:
- "// IWYU pragma: keep" on stats.h in context.h/context.cpp, where the std::derived_from constraint requires the complete type.
- "// IWYU pragma: keep" on <array> in api-trace.cpp, used only under HIPFILE_ROCPROFILER_REGISTER (a config IWYU does not see enabled).

CI:
- Add .github/workflows/hipfile-iwyu.yml: compile-only, runs once against a fixed image, surfaces findings to the job summary and an artifact, and never fails the build (IWYU runs in launcher mode).
- Install the iwyu package in DOCKERFILE.ais_ci_ubuntu.
- Wire the advisory job into hipfile-ci-toplevel.yml.

Verified warning-free builds, clang-format-clean sources, and 522/522 unit tests passing throughout.

## JIRA ID

AIHIPFILE-161

## Test Plan

New and old CI needs to pass

## Test Result

Success

## Submission Checklist

### DIFF
--- a/.github/workflows/hipfile-ci-toplevel.yml
+++ b/.github/workflows/hipfile-ci-toplevel.yml
@@ -229,3 +229,22 @@ jobs:
       cxx_standard: ${{ matrix.cxx_standard }}
       platform: ${{ matrix.supported_platforms }}
       rocm_version: ${{ matrix.rocm_versions }}
+
+  # include-what-you-use findings are compiler- and standard-independent, so run
+  # the analysis exactly once against a single fixed image rather than fanning
+  # out across the build matrix. This job is advisory: IWYU runs in launcher
+  # mode and never fails the build, so it only surfaces suggestions. Change the
+  # image key below to retarget the ROCm version (the combo must exist in
+  # Precheck's matrix).
+  iwyu:
+    if: >-
+      ${{
+        !cancelled() && (
+          needs.build_CI_image.result == 'success' ||
+          needs.build_CI_image.result == 'skipped'
+        )
+      }}
+    needs: [Precheck, build_CI_image]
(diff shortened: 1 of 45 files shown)


derobins:
The IWYU action has a LOT of false positives, but we don't gate on it and it only turns red if the action fails, not if there are reported issues. It's mainly to ensure that our CMake option and mapping files stay functional.

PR #7574, opened 2026-06-21
[hipfile] Move C++23 from build matrix to single smoke test

derobins:
## Motivation

C++20 is the production standard and is built across the full platform/compiler matrix. C++23 was previously a second value on the cxx_standard axis, fanning it out across nearly every combo for what is really just a forward-compatibility check.

## Technical Details

Reduce CXX_STANDARD to [20] and validate C++23 with a single smoke-test job (ubuntu + amdclang++ + 7.2.2) that reuses hipfile-build.yml. The cxx_standard axis is retained (as a single element) since a future production-standard bump would reuse it.

## JIRA ID

AIHIPFILE-27

## Test Plan

CI passes

## Test Result

Success

## Submission Checklist

### DIFF
--- a/.github/workflows/hipfile-ci-toplevel.yml
+++ b/.github/workflows/hipfile-ci-toplevel.yml
@@ -37,10 +37,12 @@ jobs:
           "clang++",
           "g++"
         ]
+      # C++20 is the production standard, built across the full matrix. C++23 is
+      # validated separately as a single forward-compat smoke test (see the
+      # cxx23_smoke_test job) rather than fanned out across every combo.
       CXX_STANDARD: >-
         [
-          20,
-          23
+          20
         ]
       # Note: If adding distros or ROCm versions, ensure they have been properly excluded.
       SUPPORTED_PLATFORMS: >-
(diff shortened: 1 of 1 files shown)


.github/workflows/hipfile-ci-toplevel.yml
```diff
@@ -229,3 +230,27 @@ jobs:
       cxx_standard: ${{ matrix.cxx_standard }}
       platform: ${{ matrix.supported_platforms }}
       rocm_version: ${{ matrix.rocm_versions }}
+
+  # C++23 is a forward-compat smoke test, not a production target, so it runs
+  # exactly once (ubuntu + amdclang++ + 7.2.2) rather than fanning out across the
+  # build matrix. It reuses hipfile-build.yml with cxx_standard=23 and disables
+  # coverage and artifact upload; packaging and the install test still run (they
+  # are cheap and hipfile-build.yml has no toggle to skip them).
+  cxx23_smoke_test:
```
riley-dixon: May I suggest adding an entry in `MATRIX_INCLUDE` instead? The net effect is the entire CI runs for the C++23 leg, but like here it would only run for a single image.

riley-dixon: Only the single C++23 job remains https://github.com/ROCm/rocm-systems/actions/runs/27980092517/job/82807917066?pr=7574

riley-dixon (approved):

PR #7578, opened 2026-06-21
ci(hipfile): Add push + workflow_dispatch triggers to main CI

derobins:
## Motivation

We need to add a push trigger to hipFile's CI, like other libraries in rocm-systems

## Technical Details

- Add `push` (develop) and `workflow_dispatch` triggers to the hipFile CI workflow, alongside the existing `pull_request` trigger. This matches the push + pull_request convention used across rocm-systems, enables status badges on develop, and provides post-merge validation of the full matrix.
- Make the Dockerfile-change detection event-aware: only a `pull_request` diffs against its base ref (to build a private, PR-namespaced image for pre-merge testing). On push/dispatch, `changed_dockerfile` is forced to 0 so CI reuses the canonical `latest-rocm` images.

- `GITHUB_BASE_REF` only exists on `pull_request`; the short-circuit avoids diffing against an empty base on push/dispatch.
- The canonical `latest-rocm` images are (re)published by `hipfile-image-update.yml` on the triggering merge, so push-CI reusing them is correct and avoids a double-publish race on the shared tag/cache.
- Concurrency keys on `github.ref`, so push (refs/heads/develop) and PR runs don't cancel each other across event types.

## JIRA ID

AIHIPFILE-27

## Test Plan

- [ ] Util test suites pass: `ci-matrices.sh`, `ci-images.sh`, `build-ci-images-matrix.sh` (all `--test`)
- [ ] PR run still diffs against base and builds namespaced images on a Dockerfile change
- [ ] Push to develop runs the full matrix, reuses latest-rocm images, and skips the image-build job
- [ ] Status badge reflects develop post-merge state

## Test Result

We'll see!

## Submission Checklist

### DIFF
--- a/.github/workflows/hipfile-ci-toplevel.yml
+++ b/.github/workflows/hipfile-ci-toplevel.yml
@@ -12,6 +12,15 @@ on:
       # ais-check is a self-contained Python tool with no dependency on the C++
       # library, so its changes shouldn't trigger the full C++ build/test.
       - '!projects/hipfile/tools/ais-check/**'
+  push:
+    branches: [develop]
+    paths:
+      - 'projects/hipfile/**'
+      - '.github/workflows/hipfile-*.yml'
+      - '!**/*.md'
+      - '!**/*.rst'
+      - '!projects/hipfile/tools/ais-check/**'
+  workflow_dispatch:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/hipfile-ci-toplevel.yml
```diff
@@ -23,6 +32,10 @@ permissions:
 
 jobs:
   Precheck:
+    # Forks that sync develop must not run org-only infrastructure (self-hosted
+    # AMD runners, GHCR image push). On a fork this job is skipped, and every
+    # downstream job is gated on it via `needs`, so the whole run no-ops.
+    if: github.repository == 'ROCm/rocm-systems'
```
riley-dixon, line 38: Why? This is unnecessary and actively blocks being able to test CI on a fork. External forks do not use org-specific infrastructure.

riley-dixon (changes requested):
I would suggest combing over the other downstream workflow YAML files that toplevel will call into to make sure there are not any other lines dependent on a `pull_request` trigger.

Can you also please test the `push` trigger in a fork to verify that the whole CI works as intended prior to merging it here.

PR #5914, opened 2026-05-08
AIRUNTIME-171 - cooperative groups scan

g-h-c:
[scan_benchmark_Navi48.txt](https://github.com/user-attachments/files/28066073/scan_benchmark_Navi48.txt)
## Motivation

Implement `cooperative_groups::inclusive_scan()` and `exclusive_scan()`
Documentation: #7254
HIPIFY related ticket: https://github.com/ROCm/HIPIFY/pull/2448

## Technical Details

- This is an initial implementation of the scan operators for cooperative groups. Additional PRs will add optimizations for upcoming architectures
- This PR refactors `cooperative_groups::reduce` tests so they can be used to test scans too, as the test very similar operations (a reduce is the same as a `inclusive_scan`, but it returns the result the last lane would return in all lanes)
- The implementation is different from `cooperative_groups::reduce()`. The main difference is that `cg::reduce()` is based on  the __reduce_*sync() operators where the scans do not have an equivalent warp intrinsic. Because of that, the macro GENERATE_SCAN_FUNC() is introduced, which generates equivalent warp-wide scans based on the OCKL intrinsics 

## JIRA ID

Resolves AIRUNTIME-171

## Test Plan

Essentially the same tests as cg::reduce(), but now checking the result per lane as opposed as to checking a single value, as each lane in scan would produce different results most of the time.

## Test Result

Tested on Navi48, all tests pass. Also see attached benchmark.
[scan_benchmark_Navi48.txt](https://github.com/user-attachments/files/28066105/scan_benchmark_Navi48.txt)

Tests on MI300X pass successfully both with and without address sanitizer.

## Submission Checklist

- [X] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #7588, opened 2026-06-22
ROCM-26483 - fix cooperative_groups::reduce() would not work with 2D or 3D tiles

g-h-c:
## Motivation

When using cooperative_groups::reduce(block, value, operator), if the thread block contains y-dimensions or z-dimensions that are different from 1, the operation would produce incorrect results or lead to a crash.

## Technical Details

`mask <<= (((threadIdx.x % warpSize) / group.num_threads()) * group.num_threads());`

should have been

`mask <<= (((internal::workgroup::thread_rank() % warpSize) / group.num_threads()) * group.num_threads());`

(note that the previous expression cannot be simplified to mask <<= internal::workgroup::thread_rank() % warpSize` because the operator in that expression is an integer division; not a floating point division)

## JIRA ID

ROCM-26483

## Test Plan

Added test Unit_Thread_Block_Tile_Multi_Dimensional_Reduce

## Test Result

The test fails on current develop, but passes with the fix

## Submission Checklist


yxsamliu (approved):
LGTM

g-h-c:
Fix already included in https://github.com/ROCm/rocm-systems/pull/5914

PR #7590, opened 2026-06-22
[hipfile] Install relocatable unit tests to support TheRock packaging

derobins:
## Motivation

hipFile is being integrated into TheRock, which builds, packages, and tests each component as a relocatable artifact rather than from its build tree. Our unit tests could only run in-place: gtest_discover_tests() bakes absolute build-tree paths into its ctest fragments, so the suite could not run from a packaged/installed location. This change makes the unit tests relocatable so TheRock (and any packaged install) can run them from the artifact, where the shared libraries come from sibling subprojects' staged output rather than our build tree.

## Technical Details

Add AIS_INSTALL_TESTS (default OFF) to install the unit-test suite as a relocatable tree. It installs the test binaries to share/hipfile/test and ships a self-locating discovery script (hipfile_discover.cmake) that, at ctest-run time, finds itself, derives the runtime library path relative to its own location, enumerates each binary's GoogleTest cases, and registers one add_test per case (preserving the per-case process isolation the build-tree discovery provides). Scope is unit tests only; system and stress tests still run from the build tree. Run with:
    ctest --test-dir <install>/share/hipfile/test -L unit

The option is separate from building the tests and defaults OFF, matching the convention sibling rocm-systems projects use for installable test suites (e.g. rocprofiler-systems' ROCPROFSYS_INSTALL_TESTING); since installing requires building, AIS_INSTALL_TESTS forces BUILD_TESTING on. TheRock opts in by passing -DAIS_INSTALL_TESTS=${THEROCK_BUILD_TESTING} when configuring the subproject.

Define hipfile_tests only when UNIT_TEST_SOURCE_FILES is non-empty so the AMD build (whose unit tests live entirely in internal_tests) does not ship an empty test binary.

CI: hipfile-build.yml installs the test tree to a staging prefix, MOVES it, and runs ctest from the new location, so relocatability regressions are caught on every PR here instead of surfacing downstream in TheRock. The step fails loudly if zero tests are discovered.

Document AIS_INSTALL_TESTS in INSTALL.md.

## JIRA ID

AIHIPFILE-27
ROCM-1261

## Test Plan

CI Passes

## Test Result

Success

## Submission Checklist

### DIFF
--- a/projects/hipfile/CMakeLists.txt
+++ b/projects/hipfile/CMakeLists.txt
@@ -234,6 +234,19 @@ option(AIS_INSTALL_EXAMPLES "Install example programs" ON)
 #------------------
 option(AIS_INSTALL_TOOLS "Install tool programs" ON)
 
+#------------------
+# Test programs
+#------------------
+# Whether to install the relocatable unit-test assets (binaries + ctest files)
+# so the suite can run from an installed/packaged tree. Default OFF, separate
+# from the build of the tests, matching the convention used by sibling projects
+# (e.g. rocprofiler-systems' ROCPROFSYS_INSTALL_TESTING). Installing the tests
+# requires building them, so enabling this forces BUILD_TESTING on.
+option(AIS_INSTALL_TESTS "Install unit-test programs" OFF)
+if(AIS_INSTALL_TESTS)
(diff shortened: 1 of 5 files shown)


riley-dixon (approved):

ammallya (approved):
LGTM

PR #7577, opened 2026-06-21
[hotswap] Install and link libhsa-hotswap tool

harsh-amd:
ISSUE ID: ROCm/llvm-project#3000

## Stack
Layer 1 of the rocm-systems hotswap stack, paired with ROCm/llvm-project#3007.
- This layer: make rocm-systems build, link, and install `libhsa-hotswap.so`
- Layer 2: ROCm/rocm-systems#7593
- Layer 3: ROCm/rocm-systems#7592

## Summary
- Install `libhsa-hotswap.so` from `projects/hotswap` so TheRock can package it.
- Link the hotswap HSA tool against `hsa-runtime64::hsa-runtime64`.
- Keep `libhsa-hotswap.so` as the runtime HSA tools library while COMGR provides the rewrite API.

## Testing
- `cmake -S projects/hotswap -B build-hotswap-reorder-layer1 -DCMAKE_BUILD_TYPE=RelWithDebInfo -Damd_comgr_DIR=/home/harsh/llvm-pr3000-refactor/build-comgr-displacement-shared/lib/cmake/amd_comgr -Dhsa-runtime64_DIR=/opt/rocm-7.2.0/lib/cmake/hsa-runtime64`
- `cmake --build build-hotswap-reorder-layer1 --target hsa-hotswap hotswap_test hotswap_tool_test --parallel 32`
- `cmake --install build-hotswap-reorder-layer1 --prefix build-hotswap-reorder-layer1/install-check`

Full hotswap CTest is run at layer 2, where the same-ISA passthrough and entry-trampoline gating behavior are introduced.

PR #7593, opened 2026-06-22
fix(hotswap): gate entry trampoline rewrites for gfx12.5

harsh-amd:
## JIRA ID

JIRA ID - ROCM-27304

## Stack
Layer 2 of the rocm-systems HotSwap stack, paired with ROCm/llvm-project#3008.
- Previous layer: ROCm/rocm-systems#7577
- This layer: rocm-systems loader gating and COMGR request selection for the gfx12.5 entry-trampoline path
- Final layer: ROCm/rocm-systems#7592

## Summary
- Keep rocm-systems limited to loader routing and COMGR source/target ISA-pair construction; COMGR owns validation and all code-object transformations after the call.
- Use ROCm/llvm-project#3008's explicit COMGR options API: ROCr calls `amd_comgr_hotswap_rewrite_with_options` with `AMD_COMGR_HOTSWAP_REWRITE_FLAG_ENTRY_TRAMPOLINES` when it requests entry trampolines, and keeps `amd_comgr_hotswap_rewrite` for legacy/default rewrites.
- Keep `AMD_COMGR_HOTSWAP_ENTRY_TRAMPOLINES` as the ROCr policy knob: default enabled in ROCr, literal `0` disables, unset/empty/any other value enables the entry-trampoline request.
- Cover concrete `gfx125*` targets and `gfx12-5-generic` while keeping non-gfx12.5 agents and source code objects unchanged.
- Follow ROCm/rocm-systems#7581 by keeping the entry-trampoline path keyed to the code-object ISA rather than using this path for processor retargeting.
- Preserve the existing gfx1250 A0 source/target ISA pair and leave pass selection to COMGR.
- Add unit and loader-path coverage for the local routing behavior without testing COMGR pass internals.

## Testing
- `git diff --check`
- `cmake --build build-rocrtst-hotswap-review --target hotswap_rewrite -j 8`
- `ctest --test-dir build-rocrtst-hotswap-review -R hotswap --output-on-failure`
- `cmake --build build-rocr-hotswap --target hsa-runtime64 -j 8`

### DIFF
--- a/projects/rocr-runtime/rocrtst/suites/functional/hotswap/hotswap_rewrite_test.cc
+++ b/projects/rocr-runtime/rocrtst/suites/functional/hotswap/hotswap_rewrite_test.cc
@@ -31,6 +31,14 @@
 namespace {
 
 constexpr const char* kGfx1250Isa = "amdgcn-amd-amdhsa--gfx1250";
+constexpr const char* kGfx1251Isa = "amdgcn-amd-amdhsa--gfx1251";
+constexpr const char* kGfx12_5GenericIsa =
+    "amdgcn-amd-amdhsa--gfx12-5-generic";
+constexpr const char* kGfx942Isa = "amdgcn-amd-amdhsa--gfx942";
+constexpr const char* kGfx1250B0Isa =
+    "amdgcn-amd-amdhsa--gfx1250:gfx1250-b0-specific+";
+constexpr const char* kGfx1250A0Isa =
+    "amdgcn-amd-amdhsa--gfx1250:gfx1250-b0-specific-";
 
 struct FakeHsaEnv {
   std::string isa_name = kGfx1250Isa;
@@ -50,7 +58,11 @@ LibHandle LoadLib(std::string filename) {
 #if defined(_WIN32) || defined(_WIN64)
   return LoadLibraryA(filename.c_str());
 #else
-  return dlopen(filename.c_str(), RTLD_NOW | RTLD_LOCAL);
+  int flags = RTLD_LAZY;
+#ifdef RTLD_NODELETE
+  flags |= RTLD_NODELETE;
+#endif
+  return dlopen(filename.c_str(), flags);
 #endif
 }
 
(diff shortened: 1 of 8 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -129,7 +153,9 @@ bool ResolveComgrApi(os::LibHandle lib, ComgrApi* api) {
       ResolveComgrSymbol(lib, "amd_comgr_set_data", &api->set_data) &&
       ResolveComgrSymbol(lib, "amd_comgr_get_data", &api->get_data) &&
       ResolveComgrSymbol(lib, "amd_comgr_get_data_isa_name", &api->get_data_isa_name) &&
-      ResolveComgrSymbol(lib, "amd_comgr_hotswap_rewrite", &api->hotswap_rewrite);
+      ResolveComgrSymbol(lib, "amd_comgr_hotswap_rewrite", &api->hotswap_rewrite) &&
+      ResolveComgrSymbol(lib, "amd_comgr_hotswap_rewrite_with_options",
```
shwetagkhatri: Just noting - here is a hard dependency on PR - https://github.com/ROCm/llvm-project/pull/3008 - for  amd_comgr_hotswap_rewrite_with_options. Is there a possibility that COMGR maybe old? Do we need a fallback option? This comment is not a blocker. Just for my understanding.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -87,6 +96,11 @@ bool IsEnvFlagEnabled(const char* name) {
 
 bool IsHotswapDisabledByEnv() { return IsEnvFlagEnabled("HSA_HOTSWAP_DISABLE"); }
 
+bool IsGfx12_5RewriteRequested() {
+  constexpr char kEnvName[] = "AMD_COMGR_HOTSWAP_ENTRY_TRAMPOLINES";
+  return !os::IsEnvVarSet(kEnvName) || os::GetEnvVar(kEnvName) != "0";
```
shwetagkhatri, line 102: This only treats "0" as disabled. Values like "false", "off", "no" will ENABLE trampolines.
  Consider matching the IsEnvFlagEnabled() pattern above which handles common boolean strings:
  return val != "0" && val != "false" && val != "off" && val != "no";

  Or if this is intentional, add a comment explaining why only literal "0" disables.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
harsh-amd: Yes, this intentionally depends on ROCm/llvm-project#3008 and should land after it.

I do not think we should add an old-COMGR fallback. ROCR and COMGR should come from the same ROCm package set, and this path needs the new options API to request entry trampolines.

If an old COMGR is somehow present, it fails closed: `ResolveComgrApi()` stays not ready, hotswap returns false, and we load the original code object.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
harsh-amd: This is intentional for this env var. ROCR owns the default-on policy here: unset means enabled, and only `AMD_COMGR_HOTSWAP_ENTRY_TRAMPOLINES=0` opts out.

This is documented in `env_variables.rst` and covered by `RuntimeLoadNonA0UsesEntryTrampolinesUnlessEnvIsZero`. I added a code comment to make that explicit.

projects/rocr-runtime/rocrtst/suites/test_common/CMakeLists.txt
```diff
@@ -802,7 +802,7 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
             OUTPUT_VARIABLE _comgr_symbols
             ERROR_QUIET)
           if(_comgr_symbol_result EQUAL 0 AND
-             _comgr_symbols MATCHES "amd_comgr_hotswap_rewrite")
+             _comgr_symbols MATCHES "amd_comgr_hotswap_rewrite_with_options")
```
cfreeamd, line 806: The original MATCHES "amd_comgr_hotswap_rewrite" matched both symbols (prefix). The new check MATCHES "amd_comgr_hotswap_rewrite_with_options" silently produces an empty HOTSWAP_BUILD_RPATH on systems with older COMGR, causing runtime tests to fail with library-not-found. Consider either keeping the prefix check or add a CMake warning when the new symbol is absent.

projects/rocr-runtime/rocrtst/suites/functional/hotswap/hotswap_rewrite_test.cc
```diff
@@ -210,6 +222,77 @@ rocr::hotswap::CodeObjectView MakeRealCodeObjectView() {
...
+                                             uint32_t asic_revision,
+                                             bool has_asic_revision = true) {
+  rocr::hotswap::AgentGfxRevision revision;
+  revision.gfx_target = gfx_target;
+  revision.asic_revision = asic_revision;
+  revision.has_asic_revision = has_asic_revision;
+  return revision;
+}
+
+TEST(HotswapRewriteDecision, A0RetargetsWhenEntryTrampolinesDisabled) {
```
cfreeamd: A0RetargetsWhenEntryTrampolinesDisabled tests A0 retargeting with {false} options, but the name implies the behavior is conditional on trampolines being disabled. 
There's no test with default {} options to confirm A0 retargeting is unconditional w.r.t. gfx12_5_rewrite_enabled. Consider adding A0AlwaysRetargetsRegardlessOfOptions with {} asserting the same outcome.

cfreeamd (commented):
Added a few comments for consideration--not blocking. 

I noticed this is tagged not ready for review. Will approve when it's ready assuming no significant changes requiring re-review.

harsh-amd:
> Added a few comments for consideration--not blocking.
> 
> I noticed this is tagged not ready for review. Will approve when it's ready assuming no significant changes requiring re-review.

I am not sure why this is tagged not ready for review. It is ready for review. Do you know how I can get rid of the tag?

chinmaydd:
> I am not sure why this is tagged not ready for review. It is ready for review. Do you know how I can get rid of the tag?

I think we need to fix the PR title/description to keep the bot happy.

harsh-amd:
> > I am not sure why this is tagged not ready for review. It is ready for review. Do you know how I can get rid of the tag?
> 
> I think we need to fix the PR title/description to keep the bot happy.

The PR description is as per the bot's requirements. I know its asking for a JIRA ticket, but if you look here https://github.com/ROCm/rocm-systems/pull/7921 adding a JIRA ticket didn't remove the label here as well.

projects/rocr-runtime/rocrtst/suites/test_common/CMakeLists.txt
harsh-amd: Good point. I kept the new symbol check because this test path now requires the #3008 COMGR API, but added a warning when we find COMGR without `amd_comgr_hotswap_rewrite_with_options` so older COMGR does not fail silently.

projects/rocr-runtime/rocrtst/suites/functional/hotswap/hotswap_rewrite_test.cc
harsh-amd: Agreed. I renamed/expanded the test to cover both default and disabled options, so it is clear A0 retargeting uses the legacy path independent of entry-trampoline gating.

chinmaydd:
I see. In that case I think its tripping over the "Unit test" check ? Maybe a latent bug.

```
The following policy checks failed:

❌ Unit Test
The Not ready to Review label has been added to this PR.
Once all policies pass, the label will be removed automatically.
```

EDIT: Updated description with JIRA ID, FTFY

cfreeamd (approved):

harsh-amd:
> I see. In that case I think its tripping over the "Unit test" check ? Maybe a latent bug.
> 
> ```
> 
> ❌ Unit Test
> The Not ready to Review label has been added to this PR.
> Once all policies pass, the label will be removed automatically.
> ```
> 
> EDIT: Updated description with JIRA ID, FTFY

Thanks!

shwetagkhatri (approved):
LGTM

kentrussell (approved):

pmoutsias-amd (approved):

PR #7599, opened 2026-06-22
dbgapi memcache cleanups

palves:
## Motivation

Memcache cleanups, developed while working on the memory cache performance improvement patches.  These patches stand on their own.  They end up fixing a FIXME, and a GCC compilation problem, too.

## Technical Details

See individual commit logs.

## Test Plan

Run ROCgdb gdb.rocm/ tests.

## Test Result

Test results good.

### DIFF
--- a/projects/rocdbgapi/src/memory.cpp
+++ b/projects/rocdbgapi/src/memory.cpp
@@ -766,7 +766,7 @@ void
 memory_cache_t<AddressType>::prefetch (AddressType address,
                                        amd_dbgapi_size_t size)
 {
-  if (policy == policy_t::uncached || size == 0)
+  if (size == 0)
     return;
 
   dbgapi_assert (address < (address + size) && "invalid size");
@@ -815,7 +815,7 @@ memory_cache_t<AddressType>::write_back (AddressType address,
                                          amd_dbgapi_size_t size)
 {
   std::exception_ptr exception;
-  if (policy != policy_t::write_back || size == 0)
+  if (size == 0)
     return;
 
   dbgapi_assert (address < (address + size) && "invalid size");
@@ -937,8 +937,8 @@ memory_cache_t<AddressType>::xfer_global_memory (AddressType address,
   auto begin = m_cache_line_map.lower_bound (first_line);
   auto end = m_cache_line_map.upper_bound (last_line);
 
-  /* If uncached or there are no cache lines affected by this access.  */
-  if (policy == policy_t::uncached || begin == end)
+  /* If there are no cache lines affected by this access.  */
+  if (begin == end)
     return m_xfer_global_memory (address, read, write, size);
 
   /* For cached accesses, handle one cache line at a time.  */
@@ -976,10 +976,6 @@ memory_cache_t<AddressType>::xfer_global_memory (AddressType address,
   else
     {
       memcpy (&cache_line.m_data[0] + offset, write, size);
-
-      if (policy != policy_t::write_back)
-        return m_xfer_global_memory (address, nullptr, write, size);
-
       cache_line.m_dirty = true;
     }
 
--- a/projects/rocdbgapi/src/memory.cpp
+++ b/projects/rocdbgapi/src/memory.cpp
@@ -726,8 +726,8 @@ memory_cache_t::prefetch (agent_address_t address, amd_dbgapi_size_t size)
 
   try
     {
-      m_xfer_global_memory (cache_line_begin, &staging_buffer[0], nullptr,
-                            cache_line_end - cache_line_begin);
+      m_xfer_agent_memory (cache_line_begin, &staging_buffer[0], nullptr,
+                           cache_line_end - cache_line_begin);
     }
   catch (const memory_error_t &)
     {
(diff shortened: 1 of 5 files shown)


projects/rocdbgapi/src/memory.cpp
```diff
@@ -862,15 +861,23 @@ memory_cache_t::discard (agent_address_t address, amd_dbgapi_size_t size,
...
+  auto max = m_agent.agent_address_space ().last_address ();
+  if (address > max)
+    throw memory_access_error_t (m_agent.agent_address_space (), address);
+  /* Number of bytes available from ADDRESS to MAX inclusive, minus
+     one.  MAX is the last representable address, so the full byte
+     count (MAX - ADDRESS + 1) wraps to zero when ADDRESS is 0.  Using
+     the "minus one" form avoids that wrap.  */
+  auto available_minus_one = max - address;
+  if (size - 1 > available_minus_one)
+    size = available_minus_one + 1;
```
sebdar: I think `available_minus_one + 1` will wrap, then assign to size so we might end up with a size that's 0 ? But in a more realistic case the underlying type of agent_address_t is uint64_t so size may not be larger. This just reads a bit weird.

projects/rocdbgapi/src/memory.h
```diff
@@ -582,28 +582,17 @@ class address_class_t
                  void *value) const;
 };
 
-template <typename AddressType> class memory_cache_t
+/* A write-back memory cache.  Data is immediately updated in the cache, and
+   later updated in memory when the cache is flushed.  */
+
```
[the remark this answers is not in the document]
sebdar: Have to agree with copilot here, it's not obvious to someone discovering the codebase. Ask me how I know :)

sebdar (commented):
The clean-ups look good to me. I didn't notice any behavior changes from this. Some minor remarks regarding the comments.

lancesix (approved):
LGTM, thanks for working on this.

Tested with ROCgdb's testsuite, no regression spotted.

lumachad:
re-triggered rocgdb-gpu tests (probably bad runner).

projects/rocdbgapi/src/memory.cpp
palves: It won't wrap, as that line is not reached when ADDRESS == 0 and SIZE == MAX.

For 64-bit address space, the max SIZE possible is 0xffffffff'ffffffff, so worst case is:
``` 
size     = 0xffffffff'ffffffff
address  = 0
max      = 0xffffffff'ffffffff

size - 1 < max - address, so if-cond is false, and SIZE remains 0xffffffff'ffffffff.
``` 

For 32-bit address space, when SIZE covers the whole address space, we have for worst case:
``` 
size    = 0x1'00000000
address = 0
max     = 0x0'ffffffff

size - 1 == max - address, so if-cond is false, and SIZE remains 0x1'00000000.
``` 

For 32-bit address space, when SIZE covers MORE than the whole address space (when a clamp is needed), we have for worst case:
``` 
size    = 0x1'00000001
address = 0
max     = 0x0'ffffffff

size - 1 > max - address, so if-cond is true, so we end up with:

size'   = max - address + 1 = 0x1'00000000
``` 

I've pushed an update that drops the available_minus_one temp variable, and tries to clarify the comment.  Please take a look and let me know what you think.

projects/rocdbgapi/src/memory.h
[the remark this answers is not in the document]
palves: OK, I'll look at doing something here.

And how do you know?  :-)

palves:
The last update:
- drops the available_minus_one temp variable
- tries to clarify the clamping comment

palves:
In the last update:

- improve memory_cache_t intro comment (spread over two commits, as memory_cache_t only becomes agent-specific at the end.)

projects/rocdbgapi/src/memory.h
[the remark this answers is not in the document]
palves: Now done.  Please take a look.

projects/rocdbgapi/src/memory.cpp
sebdar: Thanks for the clarification, I do think the code is clearer this way.

My point was about a hypothetical case in which the address is stored in a type that's less than `size_t`. If the address is stored in a variable that's say, 32bits, then (max - address + 1) will overflow to 0 THEN will be converted to `size_t` .. as 0 : https://godbolt.org/z/9Kvnnec1q . This is fixed if you `static_cast<size_t>(max - address)`, then do `+ 1`

But then again, this is nitpicking and we will probably never a) change agent_address_t to something smaller and b) attempt to copy the entire device space.

sebdar (approved):
Thanks, this looks good. I answered regarding the overflow problem, but this is fine as is.

projects/rocdbgapi/src/memory.cpp
palves: Ah, I see.  Yeah, I wasn't thinking of that since I knew agent_address_t was uint64_t.  OK, no harm in adding the cast, it ends up self-documenting.  I've done that.

palves:
In the last update:

- Add static_cast<size_t> in clamping code.

sebdar:
Thanks. This looks good to merge :)

PR #7604, opened 2026-06-22
fix(rccl): CTA tuning -  respect min/maxCTAs (NCCL_MIN_CTAS/NCCL_MAX_CTAS)

Kapil-Shyam-Pawar:
## Motivation

Honor `maxCTAs/NCCL_MAX_CTAS` (and `minCTAs`) when capping the communicator's channel count, which RCCL currently ignores.

## Technical Details

In `ncclTopoPostset()` (src/graph/connect.cc), the `maxCTAs` clamp on `nChannels` was immediately undone by the following `copyChannels()` call, whose target `max(minNchannels, max(nc, minCTAs))` is dominated by the bandwidth channel count `nc` and isn't bounded by `maxCTAs`. 

The fix bounds that re-expansion target by `maxCTAs` (and `tpNChannels` for child comms) so the requested cap holds.
It's a no-op when `maxCTAs` is unset.

## JIRA ID

JIRA ID : AICOMRCCL-1167, AICOMRCCL-1362

## Test Plan

Validate that `minCTAs/maxCTAs` set via both `ncclConfig_t` and the `NCCL_MIN_CTAS/NCCL_MAX_CTAS` env knobs clamp `comm->nChannels` into [minCTAs, maxCTAs], non-positive env values are ignored, and that default config (no min/max set) does not reduce the channel count.

## Test Result

Added `CtaConfigMPITest` tests 
- `ConfigOverrideAppliesMinMaxCTAs`
- `EnvKnobsApplyMinMaxCTAs`
- `EnvKnobsIgnoreNonPositiveCTAs`
- `DefaultConfigDoesNotClampChannels`

## Submission Checklist

### DIFF
--- a/projects/rccl/docs/api-reference/env-variables.rst
+++ b/projects/rccl/docs/api-reference/env-variables.rst
@@ -50,6 +50,18 @@ in the following table.
         | ``1``: Enabled.
         | ``-2``: Auto-detect; enable when the platform supports VMM.
 
+    * - | ``NCCL_MIN_CTAS``
+        | Minimum number of CTAs (channels) used for a collective. Overrides
+          the ``minCTAs`` field of ``ncclConfig_t``.
+      - | Positive integer (values ``<= 0`` are ignored).
+        | Default: unset (uses the RCCL default).
+
+    * - | ``NCCL_MAX_CTAS``
+        | Maximum number of CTAs (channels) used for a collective. Overrides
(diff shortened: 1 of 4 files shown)


thomas-huber (approved):
Hey @Kapil-Shyam-Pawar great work. Tests and enablement look great. Only thing missing is some user docs (update the .rst for environment variables with entry for NCCL_MIN_CTAS/NCCL_MAX_CTAS)

PR #7606, opened 2026-06-22
[rocprofiler-compute] Roofline E2E test

abchoudh-amd:
## Motivation

Adds End-to-end roofline test with a sample workload.

## Technical Details

- Add templated FMA-throughput kernel.
- Cover FP16/BF16/FP32/FP64/I8/I32/I64 with native arithmetic
- Gate FP8/FP6/FP4 by offload arch via device-side macros
- Wire fma_throughput HIP target into tests CMake
- Add test_roofline_profile_to_analyze: profile fma_throughput then analyze across SoC-supported datatypes, asserting per-datatype HTML and the WMMA/MFMA legend
- Register the e2e test in CTest , the coverage list, and the nightly/weekly categories

## JIRA ID

AIROCVAL-124

## Test Plan

- [x] Compile-only across gfx950/gfx942/gfx90a/gfx1151
- [x] Run --all on gfx950 GPU
- [x] Run --dtype FP32/FP8 single-precision paths
- [x] Exercise bad dtype/arg error paths
- [x] Run e2e test on a roofline GPU; skips cleanly where unsupported

## Test Result

- [x] All four archs compile clean
- [x] All 10 precisions ran, exit 0, no HIP errors
- [x] Single-precision runs forwarded correctly
- [x] Bad dtype/arg exit 1 as expected
- [x] e2e test produces per-datatype HTML with correct legend; skips on MI100/no-GPU

## Submission Checklist


vedithal-amd (changes requested):
Blocking this PR due to discussion in AIROCVAL-124

cfallows-amd (commented):
In addition to the code block per the JIRA discussion, could we also add another topic to the table for offline discussion: we already have rocflop.cpp in sample directory- how does this differ from rocflop such that we need a separate fma throughput sample, and why we can build on the existing rocflop file. Thanks

PR #7609, opened 2026-06-22
[rocjitsu] Pin rocm's pip package

amd-eochoalo:
## Motivation

Pin rocm nightly to 7.14.0a20260619. This is a good known version and we don't want that are not introduced from rocjitsu to introduce failures to CI.

## Technical Details

Pinning a package. I noticed that since version 7.14.0a20260620, the pytest corpus is failing. I'd like to temporarily pin to a known working version while looking into:

* figuring out what was the root cause for these failures
* a good way to automate updating this automatically

I'll update again every Monday manually until we have an automated way of dealing with this.

## JIRA ID

N/A

## Test Plan

CI tests.

## Test Result

CI test pass

## Submission Checklist

### DIFF
--- a/.github/workflows/rocjitsu-corpus-tests.yml
+++ b/.github/workflows/rocjitsu-corpus-tests.yml
@@ -88,7 +88,7 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install --pre \
             --index-url https://rocm.nightlies.amd.com/whl-multi-arch/ \
-            "rocm[devel]"
+            "rocm[devel]==7.14.0a20260619"
 
           ROCM_ROOT="$(rocm-sdk path --root)"
           ROCM_LIB="$(rocm-sdk path --root)"/lib


kuhar (approved):
LGTM but can you update the motivation / technical details sections with an explanation for why we want this pinned and how often we want to bump this?

atgutier (approved):

amd-eochoalo:
Just for the record, I believe this was the commit that introduced the issue: 63fbc6a0024de3b72f96cf11dc0cbed663526f57 . It has already been reverted in https://github.com/ROCm/rocm-systems/commit/716e33876cfbbf22bbace3ca5ccaf356566822d1.

As to why is it still happening and why it started happening now and not when it was originally merged? TheRock merged this commit only on Friday https://github.com/ROCm/TheRock/commit/1dc7548c2622098e702e486ada0a462fcc5118bc and it still hasn't updated to a more recent version of rocm-system that includes the revert.

PR #7615, opened 2026-06-22
[rocprofiler-compute] Overwrite workload folder on re-profile via --overwrite

jamessiddeley-amd:
## Motivation

Re-profiling into an existing workload directory silently mixes runs, basically a second profile overwrites only the files whose names collide and leaves the rest in place, so analyze can consume a mix of old and new data with no error or warning.

What surfaced this was running `profile -> analyze -> profile --roof-only -k <kernel> -> analyze`. it produced a roofline containing every kernel instead of the selected one, because analyze reused a stale `pmc_perf.csv` left behind by the first analyze. Stale perfmon configs, orphan `results_*.csv`, traces, and `roofline.csv` are all left behind the same way, and this is because a workload directory has no owner and no lifecycle rule.

This PR gives the workload directory one owner and one rule where re-profiling in place requires explicit user authorization

Running `rocprofiler-compute/rocprof-compute profile --name vcopy -VVV --device 0 -- tests/vcopy -n 1048576 -b 256` twice back to back gives:
image

## Technical Details

- Add `--overwrite` to profile mode (`src/argparser.py`).
- `run_profiler()` now validates the output directory before profiling
  (`src/rocprof_compute_base.py`): if the directory is non-empty and
  `--overwrite` is absent, it errors and asks the user to re-run with
  `--overwrite` or choose a different `--output-directory`; with `--overwrite`
  it wipes the directory contents and profiles clean. Treating any non-empty
  directory as occupied avoids hardcoding a list of "profile data" filenames.
- `--bench-only` no longer overwrites `roofline.csv` with only an after-the-fact
  warning. It now errors before running the benchmark if `roofline.csv` exists
  and `--overwrite` is absent; with `--overwrite` it replaces only `roofline.csv`
  and leaves the rest of the workload untouched.
- Update `test_roofline_kernel_filter` to pass `--overwrite` on its re-profiles,
  since it intentionally re-profiles into the same directory four times.
- Docs: documented the fresh-directory-per-run expectation, `--overwrite`, and
  automated-run guidance in `docs/how-to/profile/mode.rst`.

## JIRA ID

AIPROFCOMP-253

## Test Plan

- `ruff check` and `ruff format --check` on all changed files.
- pytest (real GPU, gfx950): `test_roofline_kernel_filter`, `test_bench_only_basic`,
  `test_roof_workload_dir_validation`, `test_path`.
- Manual: profile into a fresh dir (succeeds), re-profile same dir without
  `--overwrite` (errors), re-profile with `--overwrite` (wipes and succeeds).
- Manual: `--bench-only` into a dir with existing `roofline.csv` without
  `--overwrite` (errors) and with `--overwrite` (replaces only `roofline.csv`).
- Original reproducer: `profile --roof-only -> analyze -> profile --roof-only
  -k vgprbound --overwrite -> analyze` on the `occupancy` workload.

## Test Result

- ruff check / format: all passed.
- pytest: 4 passed (test_roofline_kernel_filter, test_bench_only_basic,
  test_roof_workload_dir_validation, test_path).
- Manual directory policy: fresh profile exit 0; re-profile without `--overwrite`
  exit 1 ("is not empty. Re-run with --overwrite ..."); re-profile with
  `--overwrite` wiped the dir (seeded sentinel removed) and exited 0.
- Manual `--bench-only`: existing `roofline.csv` without `--overwrite` exit 1
  ("'roofline.csv' already exists ..."); with `--overwrite` regenerated only
  `roofline.csv` while unrelated files were preserved.
- Reproducer fixed: after the first analyze leaves a stale `pmc_perf.csv` with all
  4 kernels, the re-profile now errors without `--overwrite`; with `--overwrite`
  the final analyze processes only the filtered kernel

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-compute/docs/how-to/profile/mode.rst
+++ b/projects/rocprofiler-compute/docs/how-to/profile/mode.rst
@@ -264,6 +264,27 @@ The output directory can be parameterized with the following keywords:
 If MPI rank is detected and the output directory does not include ``%rank%``,
 ROCm Compute Profiler appends ``/<rank>`` to avoid collisions across ranks.
 
+Each profiling run should use a fresh output directory. A workload directory
+holds the output of exactly one profile run, so profiling into a directory that
+already contains files stops with an error instead of silently mixing two runs.
+To re-profile in place, re-run with ``--overwrite``, which wipes the directory's
+existing contents first:
+
+.. code-block:: shell-session
+
+   $ rocprof-compute profile --name vcopy --overwrite -- ./vcopy -n 1048576 -b 256
+
+.. warning::
+
+   ``--overwrite`` deletes the existing contents of the target directory. It is
+   an explicit authorization to discard prior data; nothing is removed without
+   it.
+
+.. note::
+
+   CI jobs and automated runs should never reuse a workload directory; either each run
+   should profile into a fresh directory or runs should utilize ``--overwrite``.
+
 Examples:
 
 * Profiling without MPI:
@@ -834,24 +855,28 @@ To target a specific GPU device, use ``--device``:
    $ rocprof-compute profile --name my_bench --bench-only --device 2
 
 To regenerate benchmark data in an existing profiled workload directory, use
-``--output-directory`` to point at the workload path directly:
+``--output-directory`` to point at the workload path directly. Because this
+replaces the existing ``roofline.csv``, it requires ``--overwrite`` to authorize
+overwriting that file. Unlike a full profile, ``--bench-only`` replaces only
+``roofline.csv`` and leaves the rest of the workload (counter data, traces)
+untouched:
 
 .. code-block:: shell-session
 
-   $ rocprof-compute profile --bench-only --output-directory workloads/vcopy/MI300X_A1
+   $ rocprof-compute profile --bench-only --overwrite --output-directory workloads/vcopy/MI300X_A1
 
 .. note::
 
    ``--bench-only`` writes ``roofline.csv`` only; rendering a roofline
    chart additionally requires application performance counters from a
    ``--roof-only`` (or regular profile) run. The intended workflow is to
-   profile first, then re-run ``--bench-only`` against that workload
+   profile first, then re-run ``--bench-only --overwrite`` against that workload
    later to refresh stale peak values before analyzing:
 
    .. code-block:: shell-session
 
       $ rocprof-compute profile --name vcopy --roof-only -- ./vcopy
-      $ rocprof-compute profile --bench-only --output-directory workloads/vcopy/MI300X_A1
+      $ rocprof-compute profile --bench-only --overwrite --output-directory workloads/vcopy/MI300X_A1
       $ rocprof-compute analyze --path workloads/vcopy/MI300X_A1
 
 .. _torch-operator-mapping:
--- a/projects/rocprofiler-compute/docs/how-to/profile/mode.rst
+++ b/projects/rocprofiler-compute/docs/how-to/profile/mode.rst
@@ -265,10 +265,9 @@ If MPI rank is detected and the output directory does not include ``%rank%``,
 ROCm Compute Profiler appends ``/<rank>`` to avoid collisions across ranks.
 
 Each profiling run should use a fresh output directory. A workload directory
-holds the output of exactly one profile run, so profiling into a directory that
-already contains files stops with an error instead of silently mixing two runs.
-To re-profile in place, re-run with ``--overwrite``, which wipes the directory's
-existing contents first:
+holds the output of exactly one run, so profiling into a non-empty directory
+fails with an error rather than silently mixing runs. To re-profile in place,
+add ``--overwrite``, which clears the directory first:
 
 .. code-block:: shell-session
 
@@ -276,14 +275,13 @@ existing contents first:
 
 .. warning::
 
-   ``--overwrite`` deletes the existing contents of the target directory. It is
-   an explicit authorization to discard prior data; nothing is removed without
-   it.
+   ``--overwrite`` deletes the target directory's existing contents. Nothing is
+   removed without it.
 
 .. note::
 
-   Automated runs should never reuse a workload directory; either each run
-   should profile into a fresh directory or use ``--overwrite``.
+   Automated runs should never reuse a workload directory: profile into a fresh
+   directory each time, or pass ``--overwrite``.
 
 Examples:
 
@@ -854,12 +852,11 @@ To target a specific GPU device, use ``--device``:
 
    $ rocprof-compute profile --name my_bench --bench-only --device 2
 
-To regenerate benchmark data in an existing profiled workload directory, use
-``--output-directory`` to point at the workload path directly. Because this
-replaces the existing ``roofline.csv``, it requires ``--overwrite`` to authorize
-overwriting that file. Unlike a full profile, ``--bench-only`` replaces only
-``roofline.csv`` and leaves the rest of the workload (counter data, traces)
-untouched:
+To regenerate benchmark data in an existing workload directory, point
+``--output-directory`` at the workload path. This replaces the existing
+``roofline.csv``, so it requires ``--overwrite``. Unlike a full profile,
+``--bench-only`` replaces only ``roofline.csv`` and leaves the rest of the
+workload (counter data, traces) untouched:
 
 .. code-block:: shell-session
 
--- a/projects/rocprofiler-compute/docs/how-to/profile/mode.rst
+++ b/projects/rocprofiler-compute/docs/how-to/profile/mode.rst
@@ -282,8 +282,8 @@ existing contents first:
 
 .. note::
 
-   CI jobs and automated runs should never reuse a workload directory; either each run
-   should profile into a fresh directory or runs should utilize ``--overwrite``.
+   Automated runs should never reuse a workload directory; either each run
+   should profile into a fresh directory or use ``--overwrite``.
 
 Examples:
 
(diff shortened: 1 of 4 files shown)


projects/rocprofiler-compute/docs/how-to/profile/mode.rst
```diff
@@ -264,6 +264,27 @@ The output directory can be parameterized with the following keywords:
...
+
+.. warning::
+
+   ``--overwrite`` deletes the existing contents of the target directory. It is
+   an explicit authorization to discard prior data; nothing is removed without
+   it.
+
+.. note::
+
+   CI jobs and automated runs should never reuse a workload directory; either each run
```
vedithal-amd: Just saying automated runs is enough - documentation should not care about CI concerns

projects/rocprofiler-compute/docs/how-to/profile/mode.rst
```diff
@@ -264,6 +264,27 @@ The output directory can be parameterized with the following keywords:
...
+   should profile into a fresh directory or runs should utilize ``--overwrite``.
```
vedithal-amd: Worth it to check if we have any tests which profile on to the same directory and fix them to use separate dirs

projects/rocprofiler-compute/src/argparser.py
```diff
@@ -297,6 +297,20 @@ def omniarg_parser(
             "\t\t\t <current-working-directory>/workloads/<name>/%%rank%% with MPI.)"
         ),
     )
+    profile_group.add_argument(
```
vedithal-amd, line 302: Also worth adding a one-liner in --name and --output-directory pointing to --overwrite

projects/rocprofiler-compute/src/rocprof_compute_base.py
```diff
@@ -552,6 +552,37 @@ def create_profiler(self) -> object:
             self.__soc[self.__mspec.gpu_arch],
         )
 
+    @staticmethod
+    def _clear_directory_contents(output_dir: Path) -> None:
```
vedithal-amd: Cleaning a dir should be a one liner in python and need not require a helper

projects/rocprofiler-compute/src/rocprof_compute_base.py
```diff
@@ -552,6 +552,37 @@ def create_profiler(self) -> object:
...
+
+        A workload directory holds the output of exactly one profile run, so
+        profiling refuses to write into a non-empty directory unless the user
+        has authorized overwriting it with --overwrite, in which case the
+        existing contents are wiped so the new run starts clean.
+        """
+        if output_dir.is_dir() and any(output_dir.iterdir()):
+            if not getattr(self.__args, "overwrite", False):
+                console_error(
+                    f"Workload directory '{output_dir}' is not empty. Re-run "
```
vedithal-amd: Can be simplified to just say `Existing workload directory {output_dir} is not empty, please use --overwrite`

projects/rocprofiler-compute/src/rocprof_compute_base.py
```diff
@@ -552,6 +552,37 @@ def create_profiler(self) -> object:
...
+                    "with --overwrite to wipe it and re-profile, or choose a "
+                    "different --output-directory."
+                )
+            console_warning(
+                f"--overwrite: clearing existing workload directory {output_dir}"
```
vedithal-amd: `clearing existing directory {output_dir} due to --overwrite`

projects/rocprofiler-compute/src/rocprof_compute_base.py
```diff
@@ -552,6 +552,37 @@ def create_profiler(self) -> object:
...
+            )
+            self._clear_directory_contents(output_dir)
```
vedithal-amd: We could simply delete and call mkdir with exist_ok = False, this ensures old dir is deleted and new dir is created. Deleting is one liner and need not require a helper. I dont we have symlinks in output directory but use some deletion mechanism in python which does not delete the target of the symlink

projects/rocprofiler-compute/src/rocprof_compute_base.py
```diff
@@ -571,13 +602,8 @@ def run_profiler(self) -> None:
...
 
-        # Create workload directory if it does not exist
-        p = Path(self.__args.output_directory)
-        if not p.exists():
-            try:
-                p.mkdir(parents=True, exist_ok=False)
-            except FileExistsError:
-                console_error("Directory already exists.")
+        # Validate and prepare the workload directory before profiling.
+        self._prepare_workload_directory(Path(self.__args.output_directory))
```
vedithal-amd: Can be a static method without _ prefix

projects/rocprofiler-compute/src/rocprof_compute_base.py
```diff
@@ -696,6 +722,11 @@ def _run_bench_only(self) -> None:
 
         roofline_csv = output_dir / "roofline.csv"
         existing_roofline = roofline_csv.is_file()
+        if existing_roofline and not getattr(self.__args, "overwrite", False):
+            console_error(
```
vedithal-amd, line 719: Simplify error message to not say --output-directory - see previous comments

projects/rocprofiler-compute/src/rocprof_compute_base.py
```diff
@@ -721,7 +752,7 @@ def _run_bench_only(self) -> None:
             shutil.move(str(tmp_csv), str(roofline_csv))
 
         if existing_roofline:
-            console_warning(f"Overwrote existing {roofline_csv}")
+            console_log("roofline", f"Overwrote existing {roofline_csv}")
```
vedithal-amd: Should we make this console_warning like the previous overwrite log?

vedithal-amd (changes requested):

projects/rocprofiler-compute/docs/how-to/profile/mode.rst
vedithal-amd: Done, reworded the note to drop CI and just reference automated runs.

projects/rocprofiler-compute/docs/how-to/profile/mode.rst
vedithal-amd: Checked: only test_roofline_kernel_filter re-profiles into the same directory, which is exactly what it tests, so it keeps --overwrite. All other profile tests use a fresh dir.

projects/rocprofiler-compute/src/argparser.py
vedithal-amd: Done, added an --overwrite pointer to both --name and --output-directory help.

projects/rocprofiler-compute/src/rocprof_compute_base.py
vedithal-amd: Done, dropped the helper and inlined cleanup as shutil.rmtree + mkdir.

projects/rocprofiler-compute/src/rocprof_compute_base.py
vedithal-amd: Done, simplified to: Existing workload directory {output_dir} is not empty, please use --overwrite.

projects/rocprofiler-compute/src/rocprof_compute_base.py
vedithal-amd: Done, reworded to: clearing existing directory {output_dir} due to --overwrite.

projects/rocprofiler-compute/src/rocprof_compute_base.py
vedithal-amd: Done, now shutil.rmtree(output_dir) + mkdir. rmtree unlinks symlink entries without following them, so symlink targets are left intact.

projects/rocprofiler-compute/src/rocprof_compute_base.py
vedithal-amd: Done, prepare_workload_directory is now a static method (no _ prefix) taking output_dir and overwrite.

projects/rocprofiler-compute/src/rocprof_compute_base.py
vedithal-amd: Done, dropped --output-directory from the message.

projects/rocprofiler-compute/src/rocprof_compute_base.py
vedithal-amd: Done, reverted to console_warning for consistency with the profile-path overwrite log.

projects/rocprofiler-compute/src/rocprof_compute_base.py
vedithal-amd: Update: final version clears contents in place (unlink files/symlinks, rmtree subdirs) rather than deleting the dir, to preserve its permissions/ownership. Still no helper.

projects/rocprofiler-compute/src/rocprof_compute_base.py
vedithal-amd: Update: switched away from rmtree(output_dir)+mkdir to clearing contents in place, since deleting+recreating drops the dir permissions/ownership. Symlinks are unlinked, so their targets are never followed.

xuchen-amd (approved):
looks good to me. 

a small nit: `delete`/`wipe`/`remove` terms are used interchangeably, maybe settle on one term? (personal preference and non blocking)

vedithal-amd (approved):
LGTM thanks!

cfallows-amd (approved):
Underscore prefix discussion is a todo for later, not necessary for this PR to go in. Thanks

PR #7616, opened 2026-06-22
[rocprofiler-compute][bugfix] fix db analyze empty metric filter

xuchen-amd:
## Motivation

Running `analyze` with `--output-format db` and a `--block` filter that matches no metrics on the GPU crashes instead of returning a result because the filter removes every metric. The CLI analyze mode handle this fine.

## Technical Details

When the filter removes all metrics, the code built empty DataFrames with no columns, which broke two steps downstream:

1. `KeyError: 'pct_of_peak'` — looking up a column that didn't exist on the empty frame.
2. `ValueError: Columns must be same length as key` — an empty `apply()` returned a DataFrame instead of a Series.

### Changes - 6b45f8b28cb8050400868d774e03b7b03ca4816e

- Build the metric DataFrames with explicit columns so they keep their shape even when empty.
- Return an empty Series (not a DataFrame) when there are no expressions.
- Print a warning when a workload matches no metrics.
- Move the shared metric-table loop into one helper (`_iter_metric_table_rows`) instead of duplicating it.

## JIRA ID

AIPROFCOMP-678

## Test Plan

added unit tests: a38d1a4c993a69395e181f5a20c77875f811e6a3

## Test Result

- [x] local pytest pass
- [x] local ctest pass
- [x] ci pass

## Submission Checklist


projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
```diff
@@ -559,6 +560,10 @@ def calc_dataframe_expressions(
         expression_df: pd.DataFrame,
         emit_variance_warnings: bool = False,
     ) -> pd.Series:
+        # apply(axis=1) on an empty frame returns a DataFrame, not a Series,
+        # which breaks the single-column assignment at the call site.
+        if expression_df.empty:
+            return pd.Series(index=expression_df.index, dtype=object)
```
vedithal-amd: Instead of the empty-frame guard, consider building the result directly:

```python
return pd.Series(
    [
        db_analysis.evaluate(
            f"{row.metric_id} - {row.value_name}",
            row.value, pmc_df, sys_info,
            emit_variance_warnings=emit_variance_warnings,
        )
        for row in expression_df.itertuples(index=False)
    ],
    index=expression_df.index,
)
```

This makes the `pd.Series` return type explicit at the call site rather than relying on `apply`'s shape inference, removes the empty special-case (no rows yields an empty Series), and `itertuples` avoids building a per-row Series the way `apply(axis=1)` does.

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
```diff
@@ -781,40 +800,59 @@ def calc_metrics_data(
...
+                    ).columns
+                ],
+                columns=["metric_id", "value_name", "value"],
+            )
+
+            if metrics_info_df.empty:
+                console_warning(
+                    f"No metrics matched the active filter for {workload_path}; "
+                    "the database will contain no metric values for this workload."
+                )
```
vedithal-amd: Three things here: drop the single-use helper, fix the empty schema without a parallel columns list, and drop the warning.

1. `_iter_metric_table_rows` (lines 761-773) is only called once, so it just relocates the loop. Inline the shared rows once instead. That dedupes the otherwise-duplicated loop header without a method + docstring.
2. Use a `NamedTuple` per frame as the single source of truth for field names. `columns=*._fields` preserves the schema when the result is empty (the actual fix for a `--block` filter matching nothing), and there's no separate `columns=[...]` to drift from the row construction.
3. Drop the no-metrics warning: an empty result from an explicit filter is expected and the empty DB is self-evident. The assertion in `test_calc_metrics_data_empty_filter_preserves_schema_and_warns` comes out with it.

```python
# module scope (add NamedTuple to the typing import)
class MetricInfoRow(NamedTuple):
    name: str
    metric_id: str
    description: Optional[str]
    unit: Optional[str]
    pct_of_peak: bool
    table_name: str
    sub_table_name: str


class ExpressionRow(NamedTuple):
    metric_id: str
    value_name: str
    value: str
```

```python
# in calc_metrics_data, replacing the helper call, both DataFrame builds, and the warning
metric_table_rows = [
    (metric_id, metric_df, row)
    for df_id, metric_df in self._arch_configs[gfx_arch].dfs.items()
    if df_id != 402  # roofline points handled in calc_roofline_data
    if set(metric_df.columns).intersection({"Metric", "Channel"})
    for metric_id, row in metric_df.iterrows()
]
metrics_info_df = pd.DataFrame(
    [
        MetricInfoRow(
            name=row.get("Metric") or row["Channel"].strip(),
            metric_id=metric_id,
            description=row.get("Description"),
            unit=row.get("Unit"),
            pct_of_peak=row.get("Percent of Peak") is True,
            table_name=table_names_map[int(metric_id.split(".")[0]) * 100],
            sub_table_name=table_names_map[
                int(metric_id.split(".")[0]) * 100
                + int(metric_id.split(".")[1])
            ],
        )
        for metric_id, _, row in metric_table_rows
    ],
    columns=MetricInfoRow._fields,
)
expression_df = pd.DataFrame(
    [
        ExpressionRow(
            metric_id=metric_id,
            value_name=value_name,
            value=row[value_name].strip(),
        )
        for metric_id, metric_df, row in metric_table_rows
        for value_name in metric_df.drop(
            columns=non_expression_columns, errors="ignore"
        ).columns
    ],
    columns=ExpressionRow._fields,
)
```

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
```diff
@@ -4,6 +4,7 @@
 import ast
 import re
 import warnings
+from collections.abc import Iterator
```
vedithal-amd: With `_iter_metric_table_rows` removed, this `Iterator` import is no longer used and should be dropped. Add `NamedTuple` to the `typing` import on the next line (`from typing import Any, NamedTuple, Optional`) for the schema tuples suggested above.

vedithal-amd (changes requested):
Fix is correct, but it can be simpler. See inline comments: build the Series via itertuples instead of the empty guard, drop the single-use helper + use a NamedTuple for schema, remove the no-metrics warning, and clean up the imports.

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
xuchen-amd: removed and added new import: 22729f8bc90b07a04e39e326507544cf76f4b509

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
xuchen-amd: addressed: 22729f8bc90b07a04e39e326507544cf76f4b509

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
xuchen-amd: addressed: 22729f8bc90b07a04e39e326507544cf76f4b509

feizheng10 (approved):
Great

vedithal-amd (changes requested):
will review soon

vedithal-amd (approved):
Renamed the empty-filter test to drop the stale `_and_warns` suffix (deab28eb63). LGTM.

PR #7617, opened 2026-06-22
[ROCM-26423] Fix driver version buffer

adam360x:
amdsmi_get_gpu_driver_info (amd_smi_utils.cc): When /sys/module/amdgpu/version is absent (in-tree KMD builds without MODULE_VERSION()), populate driver_version with "N/A" and return success instead of AMDSMI_STATUS_DIRECTORY_NOT_FOUND. 

This also fixes a secondary bug where the early-return prevented driver_name and driver_date from being populated. Verified by bind-mounting /dev/null over the version file to simulate the missing-file scenario.

### DIFF
--- a/projects/amdsmi/src/amd_smi/amd_smi_utils.cc
+++ b/projects/amdsmi/src/amd_smi/amd_smi_utils.cc
@@ -640,7 +640,10 @@ amdsmi_status_t smi_amdgpu_get_driver_version(amd::smi::AMDSmiGPUDevice* device,
   std::string empty = "";
   std::strncpy(version, empty.c_str(), len - 1);
   openFileAndModifyBuffer("/sys/module/amdgpu/version", version, static_cast<size_t>(len));
-  if (version[0] == '\0') return AMDSMI_STATUS_DIRECTORY_NOT_FOUND;
+  if (version[0] == '\0') {
+    std::strncpy(version, "N/A", len - 1);
+    version[len - 1] = '\0';
+  }
 
   return status;
 }


projects/amdsmi/src/amd_smi/amd_smi_utils.cc
```diff
@@ -640,7 +640,12 @@ amdsmi_status_t smi_amdgpu_get_driver_version(amd::smi::AMDSmiGPUDevice* device,
   std::string empty = "";
   std::strncpy(version, empty.c_str(), len - 1);
   openFileAndModifyBuffer("/sys/module/amdgpu/version", version, static_cast<size_t>(len));
-  if (version[0] == '\0') return AMDSMI_STATUS_DIRECTORY_NOT_FOUND;
+  // In-tree / upstream KMD builds may omit MODULE_VERSION(), leaving
+  // /sys/module/amdgpu/version absent.  Report "N/A" instead of failing.
+  if (version[0] == '\0') {
+    std::strncpy(version, "N/A", len - 1);
+    version[len - 1] = '\0';
+  }
```
marifamd, line 646: let's keep historical erroring. Cleaning the buffer is a separate issue vs returning AMDSMI_STATUS_DIRECTORY_NOT_FOUND

marifamd (changes requested):

projects/amdsmi/src/amd_smi/amd_smi_utils.cc
adam360x: Fixed all parts

projects/amdsmi/docs/reference/amdsmi-py-api.md
```diff
@@ -534,6 +534,7 @@ Exceptions that can be thrown by `amdsmi_get_gpu_driver_info` function:
 
 - `AMDSMI_STATUS_INVAL` - Invalid parameters
 - `AMDSMI_STATUS_TIMEOUT` - Timeout in API call
+- `AMDSMI_STATUS_DIRECTORY_NOT_FOUND` - `/sys/module/amdgpu/version` is absent (in-tree KMD builds without `MODULE_VERSION()`); `driver_date` and `driver_name` are still populated
```
marifamd: too verbose

adam360x: Shortened

projects/amdsmi/src/amd_smi/amd_smi_utils.cc
marifamd: I'm actually wrong here, if we do the return error then the python api will bubble up with an error as well

marifamd (approved):

gabrpham (approved):
Looks good to me. Approved.

PR #7618, opened 2026-06-22
feat(rocprofiler-systems): Allow trace cache to handle gotcha args

kcossett-amd:
## Motivation

This is the improved other half of #6116

`ROCPROFSYS_TRACE_LEGACY` encodes information that is missing from debug args when trace cache is used. For example, with `pthread_create`, we currently see:
image

But with `ROCPROFSYS_TRACE_LEGACY=ON`, we see:
image

There is a clear disconnect.

## Technical Details

`category_region.hpp`
Cache gotcha/perfetto region arguments into the trace cache. Category-independent helpers now live in a policy-templated helper `rocprofsys::utility::category_region<Policy>`, composed by `category_region<CategoryT>` via using `region_cache = utility::category_region<>`. The cache lifecycle's system seams (clock, trace-cache sink, thread metadata) are injected as value members through a Policy bundle

 - `serialize_name_value_pairs`: serializes the gotcha {"name", value} pairs into the wire format (idx;;type;;name;;value;;).
 - `serialize_annotation_args / serialize_return_arg`: serialize gotcha audit arguments (synthesized arg{N}-{type} names) and return values.
 - `has_trace_cache_arg_pairs_v`: detects the gotcha pair convention (even count, string-like name slots).
 - `next_arg_index`: derives the next index from the last record instead of storing a count.
 - `renumber_serialized_args`: renumbers an appended batch in a single string pass.
 - The above are static helpers; the stateful cache lifecycle (`cache_start` / `cache_stop` / `append_cache_args` / `flush_pending_cached_entries`) are instance methods reached through a per-thread `instance()`.
 - `append_cache_args`: appends a batch to the open entry (adopts the first batch as-is, else renumbers). 
 - `start / start_with_args` now funnel through one start_impl.
 - Hooked gotcha audit incoming/outgoing paths to cache argument and return values.

`regions.cpp`:
 -  Adjust `flush` call signature.

Added unit tests

Added pytests
 - two new binaries: `annotations.cpp` and `pthread.cpp`. The first tests perfetto annotations and the second verifies that gotcha arguments are captured.
 - Both ROCpd and perfetto outputs are verified.

**OTHER**
`test_shmem_gotcha.cpp`: Drop `name` and `[this]`. It was causing workflow to fail:
https://github.com/ROCm/rocm-systems/actions/runs/27972186388/job/82780906185

## JIRA ID

Jira ID : AIPROFSYST-519

## Test Plan

New unit tests
New pytests:
```
minimal-pthreads-binary-rewrite
minimal-pthreads-runtime-instrument
minimal-pthreads-sys-run
```

## Test Result

Unit tests pass:
```
Note: Google Test filter = category_region*
[==========] Running 29 tests from 3 test suites.

...

[----------] Global test environment tear-down
[==========] 38 tests from 4 test suites ran. (0 ms total)
[  PASSED  ] 38 tests.
```
New pytests pass:
```
    Start 182: minimal-pthreads-binary-rewrite
2/5 Test #182: minimal-pthreads-binary-rewrite .......   Passed    5.19 sec
    Start 183: minimal-pthreads-runtime-instrument
3/5 Test #183: minimal-pthreads-runtime-instrument ...   Passed    6.69 sec
    Start 184: minimal-pthreads-sys-run
4/5 Test #184: minimal-pthreads-sys-run ..............   Passed    2.14 sec
```

For `pthreads`:
image

## Submission Checklist


projects/rocprofiler-systems/source/lib/rocprof-sys/regions.cpp
```diff
@@ -43,6 +43,12 @@ invoke_category_region_start(rocprofsys_category_t _category, const char* name,
                         tracing::add_perfetto_annotation(ctx, _annotations[i]);
                 }
             });
+        // Annotation lambda cannot be decoded in category_region::start(),
+        // append args here
+        if(_annotations)
+            component::category_region<category_type>::append_cache_args(
+                name, serialize_annotation_args(std::span<const rocprofsys_annotation_t>{
+                          _annotations, _annotation_count }));
```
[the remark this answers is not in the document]
kcossett-amd: Perfetto exporter re-gates at export. Plus, this path also handles ROCpd, for which I assume we do not want `ROCPROFSYS_PERFETTO_ANNOTATIONS` to block.

projects/rocprofiler-systems/source/lib/rocprof-sys/regions.cpp
```diff
@@ -70,6 +76,12 @@ invoke_category_region_stop(rocprofsys_category_t _category, const char* name,
         // skip if category is disabled
         if(!trait::runtime_enabled<category_type>::get()) return;
 
+        // Append pop-time annotations to the cached entry before stop() flushes it
+        if(_annotations)
+            component::category_region<category_type>::append_cache_args(
+                name, serialize_annotation_args(std::span<const rocprofsys_annotation_t>{
+                          _annotations, _annotation_count }));
```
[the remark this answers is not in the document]
kcossett-amd: See previous.

kcossett-amd:
**Open question:** Arg serialization into the trace cache here is intentionally ungated. `ROCPROFSYS_PERFETTO_ANNOTATIONS` is perfetto-only and the perfetto exporter already re-gates at export (`handle(region_sample&)` checks `m_use_annotations`). Gating it here would also strip these args from ROCpd. 

Do we want a separate flag?

Edit: Perfetto annotation support removed as they come through user API, which will be removed.

projects/rocprofiler-systems/examples/minimal/annotations.cpp
```diff
@@ -0,0 +1,33 @@
...
+    int64_t status = 0;
+
+    rocprofsys_annotation_t push_annotations[] = {
+        { "size", ROCPROFSYS_INT64, &size },
+    };
+    rocprofsys_annotation_t pop_annotations[] = {
+        { "status", ROCPROFSYS_INT64, &status },
+    };
+
+    rocprofsys_user_push_annotated_region("annotated_region", push_annotations,
```
mradosav-amd: This API is deprecated and I guess it will be removed soon.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/category_region.hpp
```diff
@@ -77,14 +84,270 @@ cache_region(std::uint64_t thread_id, const std::string& name, std::uint64_t sta
             end_ts, CALLSTACK, args_str.c_str(), category.c_str() });
 }
 
+template <typename Tp>
```
mradosav-amd, line 75: Can we move those 'free' functions to `category_region` class as `private/public` static functions?

projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/category_region.hpp
```diff
@@ -77,14 +84,270 @@ cache_region(std::uint64_t thread_id, const std::string& name, std::uint64_t sta
...
+// Names are synthesized as "arg{N}-{demangled-type}" to mirror add_perfetto_annotation
+template <typename... Args>
+std::string
+serialize_annotation_args(Args&&... args)
+{
+    ROCPROFSYS_SCOPED_THREAD_STATE(ThreadState::Internal);
+
+    std::string   args_str = {};
+    std::uint32_t idx      = 0;
+    ROCPROFSYS_FOLD_EXPRESSION(
```
mradosav-amd: I think we can use `(<expression>, ...)` instead of this macro, since this is C++ project.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/category_region.hpp
```diff
@@ -77,14 +84,270 @@ cache_region(std::uint64_t thread_id, const std::string& name, std::uint64_t sta
...
+}
+
+template <typename Tp>
+std::string
+get_serialized_arg_value(Tp&& value)
+{
+    std::stringstream ss;
+    ss << std::forward<Tp>(value);
+    return ss.str();
+}
```
mradosav-amd: Use `fmt::format`, it's faster.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/category_region.hpp
```diff
@@ -77,14 +84,270 @@ cache_region(std::uint64_t thread_id, const std::string& name, std::uint64_t sta
...
+    {
+        return (
+            (Idx % 2 != 0 || trace_cache_arg_name<std::tuple_element_t<Idx, TupleT>>) &&
+            ...);
+    }
+}
+
+template <size_t Idx = 0, typename TupleT>
+void
+append_serialized_args(std::string& args_str, TupleT&& args)
```
mradosav-amd: Maybe do fold expression instead of recursive template. 

Something like this:
```cpp
template <typename TupleT, size_t... Idx>
void append_serialized_args_impl(std::string& args_str, TupleT&& args,
                                  std::index_sequence<Idx...>)
{
    (append_serialized_arg(args_str, Idx,
                           std::get<Idx * 2>(args),
                           std::get<Idx * 2 + 1>(args)), ...);
}

template <typename TupleT>
void append_serialized_args(std::string& args_str, TupleT&& args)
{
    constexpr auto N = std::tuple_size_v<std::remove_reference_t<TupleT>>;
    append_serialized_args_impl(args_str, std::forward<TupleT>(args),
                                std::make_index_sequence<N / 2>{});
}
```

projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/category_region.hpp
```diff
@@ -77,14 +84,270 @@ cache_region(std::uint64_t thread_id, const std::string& name, std::uint64_t sta
...
+                                     .arg_value  = std::string{ value } } });
+}
+
+// A type qualifies as a trace-cache argument "name" slot when it is string-like.
+template <typename Tp>
+concept trace_cache_arg_name = std::convertible_to<std::decay_t<Tp>, std::string_view>;
+
+// Args passed to start()/stop() qualify as cacheable ("arg-name", value) pairs only when
+// they are non-empty, grouped two-by-two, and every "name" slot is string-like
+template <typename TupleT, size_t... Idx>
```
mradosav-amd: You can hide the `index_sequence` from the call site with a variable template.
Right now we call it like:
```cpp
has_trace_cache_arg_pairs<tuple_type>(std::make_index_sequence<sizeof...(Args)>{})
```

We can hide `std::make_index_sequence<sizeof...(Args)` from user of this and just pass `Args...`.

Hint: You can use `consteval` and `constexpr`..  to achieve something like this:
`if constexpr (has_trace_cache_arg_pairs_v<Args...>)`

projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/category_region.hpp
```diff
@@ -77,14 +84,270 @@ cache_region(std::uint64_t thread_id, const std::string& name, std::uint64_t sta
...
+}
+
+// Decode a single rocprofsys_annotation_t into a trace-cache argument record.
+// Returns false (leaving out untouched) when the annotation is empty/short-circuited.
+// or its type is out of range.
+template <size_t Idx, size_t... Tail>
+bool
+annotation_to_argument(const rocprofsys_annotation_t& annotation,
+                       rocprofsys::argument_info& out, std::index_sequence<Idx, Tail...>)
+{
```
mradosav-amd: Maybe you can replace recursive overload chain with C++20 template lambda.
The existing implementation forces the caller to pass the index sequence explicitly.. 
In C++20 you can write `[]<size_t... Idx>(std::index_sequence<Idx...>)` which expresses the same expansion as a fold expression in a single function..

projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/category_region.hpp
```diff
@@ -77,14 +84,270 @@ cache_region(std::uint64_t thread_id, const std::string& name, std::uint64_t sta
...
+    if(args_str.empty()) return;
+
+    auto key = entry_key{ name, rocprofsys::trait::name<CategoryT>::value };
+    auto itr = map_name_to_args.find(key);
+    if(itr != map_name_to_args.end() && !itr->second.empty())
+    {
+        auto& entry = itr->second.back();
+        // If args were previously stored, the new ones must be renumbered to accommodate
+        auto appended_arg_count = renumber_serialized_args(args_str, entry.arg_count);
```
mradosav-amd: This deserializes and re-serializes the entire arg string just to renumber indices.. Maybe we can just process string since we know the format, or extend data to solve this with extended data structure.

projects/rocprofiler-systems/examples/minimal/annotations.cpp
kcossett-amd: Removed.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/category_region.hpp
kcossett-amd: Sure, but for arbitrary non-void* pointer types, I have to wrap them in `fmt::streamed` (routes through `operator<<`), since `fmt` rejects "{}" for those.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/category_region.hpp
kcossett-amd: ooh nice. However this function is part of the now removed code to handle perfetto annotations, which you have mentioned is deprecated.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/category_region.hpp
kcossett-amd: Processing the string is better. I'll also remove the stored `arg count`. Going through the string backwards to get that last idx does take some small amount of time, but the cost is really just felt when appending more args to an entry with already existing args (gotcha).

projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/category_region.hpp
kcossett-amd: I've (with the help of AI) moved them to static members on a non-template base `category_region_base` (inherited by `category_region<CategoryT>`) rather than onto the template directly, so the category-independent code compiles once instead of per category (same pattern as existing `get_thread_status`)

mradosav-amd: I understand what you wanted to do. I would suggest you not to go with inheritance.
It's more practical to go with composition pattern and with templates. Suggestion to rename `category_region_base` to `utility::category_region` . It will clearly show what it is used for. 

We would like to cover whole file with unit tests, this PR can enable that to be done in future. Because of that we should not instantiate dependencies. Dependencies should be mockable. For example, we don't want real `perfetto` call, we want to have mocked call for it.
We want to achieve what is described in [policy based dependency injection](https://github.com/ROCm/rocprofiler-systems-skills/blob/main/skills/programming-cpp-policy-based-di/SKILL.md).

kcossett-amd: Done. Renamed `category_region_base` to `utility::category_region<Policy>`. I had AI aid me to make policy-based DI with mockable dependencies + more tests.

mradosav-amd (approved):
Looks good, thanks.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/category_region.hpp
```diff
@@ -237,6 +507,10 @@ struct category_region : comp::base<category_region<CategoryT>, void>
 
     static void start_with_args(std::string_view name, std::string serialized_args);
 
+    // Appends pre-serialized args to the currently-open region with this name.
+    // Used by the gotcha audit and perfetto-annotation paths
```
dgaliffiAMD:
```suggestion
    // Used by the gotcha audit paths
```

projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/category_region.hpp
```diff
@@ -112,43 +112,310 @@ cache_stop(const char* name)
...
+
+        // Less than 5 delimiters means last record is the first one and starts at offset
+        // 0
+        std::size_t record_start = 0;
+        std::size_t search_end   = std::string::npos;
+        for(int i = 0; i < 5; ++i)
+        {
+            const std::size_t p = args_str.rfind(delim, search_end);
+            if(p == std::string::npos) break;
+            if(i == 4)
```
dgaliffiAMD: Can we define constant for some of these new "magic numbers"?

ex.,

``` 
// A serialized record is four delimiter-terminated fields: idx;;type;;name;;value;;                                                                                                                                                                                                    
static constexpr std::size_t fields_per_record = 4;  
```

```
// Step back over the trailing delimiter plus the last record's fields to find                                                                                                                                                                                                      // where its idx field begins. Fewer than that many delimiters means there is a                                                                                                                                                                                                     
// single record starting at offset 0. 
constexpr std::size_t delims_to_last_record = fields_per_record + 1; 
```

```
// each renumbered idx field can grow by a few digits; reserve a little slack                                                                                                                                                                                                       
constexpr std::size_t renumber_growth_slack = 16;    
```

projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/tests/test_category_region.cpp
```diff
@@ -0,0 +1,875 @@
...
+    EXPECT_EQ(args[2].arg_value, "3");
+}
+
+TEST(category_region_serialization, serialize_name_value_pairs_invalid_returns_empty)
+{
+    EXPECT_TRUE(serialize_name_value_pairs().empty());        // no args
+    EXPECT_TRUE(serialize_name_value_pairs("only").empty());  // odd count
+    EXPECT_TRUE(serialize_name_value_pairs(1, 2).empty());    // non-string name slot
+}
+
```
dgaliffiAMD, line 230:
```suggestion

// An argument value that itself contains the field delimiter ";;" must not corrupt
// the wire format: it should round-trip back as a single record whose value is
// preserved verbatim.
TEST(category_region_serialization, serialize_name_value_pairs_value_with_delimiter)
{
    auto args = parse(serialize_name_value_pairs("path", "a;;b"));
    ASSERT_EQ(args.size(), 1u);
    EXPECT_EQ(args[0].arg_number, 0u);
    EXPECT_EQ(args[0].arg_name, "path");
    EXPECT_EQ(args[0].arg_value, "a;;b");
}
```
I think something like this will fail. ie., if the string that your serializing has your delimiter.

You could go with a "lossy" approach and convert "a;;b" --> "a__b". 
You could go with a "lossless" approach by "escaping" the ";" in the string value. So, "a;;b" --> "a%3B%3Bb" and then can deserialized back to "a;;b". 

I tried the second option here: https://github.com/ROCm/rocm-systems/pull/new/users/dgaliffi/review/pr-7618.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/category_region.hpp
```diff
@@ -112,43 +112,310 @@ cache_stop(const char* name)
...
+            {
+                record_start = p + delim.size();
+                break;
+            }
+            if(p == 0) break;
+            search_end = p - 1;
+        }
+
+        std::uint32_t idx = 0;
+        std::from_chars(args_str.data() + record_start, args_str.data() + args_str.size(),
```
dgaliffiAMD: Forgot to check the error code: 

```suggestion
          const auto [ptr, ec] = std::from_chars(                                                                                                                                                                                                  
               args_str.data() + record_start, args_str.data() + args_str.size(), idx);                                                                                                                                                             
   
          if(ec != std::errc{})                                                                                                                                                                                                                    
          {                                                                                                                                                                                                                                        
              // The idx field did not parse: the wire string is malformed (e.g. an                                                                                                                                                                
              // unescaped delimiter shifted the fields). Fall back to 0 rather than                                                                                                                                                               
              // continuing from a garbage index.                                                                                                                                                                                                  
              LOG_WARNING("[category_region] next_arg_index: malformed record index in "                                                                                                                                                           
                          "\"{}\"; restarting numbering from 0",                                                                                                                                                                                   
                          args_str);                                                                                                                                                                                                               
              return 0;                                                                                                                                                                                                                            
          }                 
```

projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/category_region.hpp
kcossett-amd: Yes. I'll move `fields_per_record` to `common_types.hpp`, the rest can stay in `category_region.hpp`.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/tests/test_category_region.cpp
kcossett-amd: I'll adopt your lossless approach + add more tests.

kcossett-amd:
TheRock CI failures are unrelated (compute is failing)

dgaliffiAMD (approved):

PR #7621, opened 2026-06-22
fix(hsakmt): Fix shmem leak in reserved_aperture_release

amd-willpalacek:
Add munmap() before mmap() to release kernel shmem refcount for MAP_SHARED allocations. Without this,
  reserved_aperture_release() leaves shmem pages pinned, causing OOM during repeated allocations.

Fixes: ROCM-23563

  ## Motivation

  KFDMemoryTest.BigSysBufferStressTest fails with OOM kills on MI210. Shmem grows unbounded across allocation iterations
  because MAP_SHARED memory is not properly released.

  ## Technical Details

  `reserved_aperture_release()` in `projects/rocr-runtime/libhsakmt/src/fmm.c` remaps freed memory as PROT_NONE using
  `mmap(..., MAP_FIXED)`. For MAP_SHARED allocations, this does not decrement the kernel shmem refcount - only `munmap()` does.
  Added `munmap()` before the `mmap()` call. Also simplified the ENOMEM retry path since memory is already unmapped.

  ## JIRA ID

  JIRA ID: ROCM-23563

  ## Test Plan

  Run `kfdtest --gtest_filter='*BigSysBufferStressTest*'` on MI210 and monitor shmem via `/proc/meminfo`.

  ## Test Result

  | Metric | Before Fix | After Fix |
  |--------|------------|-----------|
  | Iteration 1 | 1953 × 128MB | 1953 × 128MB |
  | Iteration 2 | OOM killed | 1953 × 128MB |
  | Iteration 3 | - | 1953 × 128MB |
  | Iteration 4 | - | 1953 × 128MB |
  | Shmem after | 246 GiB (leaked) | 20 MiB |
  | Result | FAILED | PASSED (899s) |

  ## Submission Checklist

 **Unit Test Waiver:** This is a memory management bug fix in libhsakmt. The fix is validated by running `kfdtest --gtest_filter='*BigSysBufferStressTest*'` which
  exercises the affected code path. Adding a new unit test would require duplicating this existing coverage.

### DIFF
--- a/projects/rocr-runtime/libhsakmt/src/fmm.c
+++ b/projects/rocr-runtime/libhsakmt/src/fmm.c
@@ -712,29 +712,15 @@ static void reserved_aperture_release(manageable_aperture_t *app,
 		/* Reset NUMA policy */
 		mbind(address, MemorySizeInBytes, MPOL_DEFAULT, NULL, 0, 0);
 
-		/* Remove any CPU mapping, but keep the address range reserved */
+		munmap(address, MemorySizeInBytes);
+
 		mmap_ret = mmap(address, MemorySizeInBytes, PROT_NONE,
 			MAP_ANONYMOUS | MAP_NORESERVE | MAP_PRIVATE | MAP_FIXED,
 			-1, 0);
 		if (mmap_ret == MAP_FAILED && errno == ENOMEM) {
-			/* When mmap count reaches max_map_count, any mmap will
-			 * fail. Reduce the count with munmap then map it as
-			 * NORESERVE immediately.
-			 */
-			if (munmap(address, MemorySizeInBytes) == 0) {
-				/* After unmapping, try mmap again and handle failure
-				 * */
-				mmap_ret = mmap(address, MemorySizeInBytes, PROT_NONE,
-						MAP_ANONYMOUS | MAP_NORESERVE | MAP_PRIVATE | MAP_FIXED,
-						-1, 0);
-				if (mmap_ret == MAP_FAILED) {
-					/* Handle mmap failure gracefully, log if needed */
-					pr_err("Failed to remap memory after unmap\n");
-				}
-			} else {
-				/* Handle munmap failure if needed */
-				pr_err("Failed to unmap memory\n");
-			}
+			mmap_ret = mmap(address, MemorySizeInBytes, PROT_NONE,
+					MAP_ANONYMOUS | MAP_NORESERVE | MAP_PRIVATE | MAP_FIXED,
+					-1, 0);
 		}
 	}
 }
--- a/projects/rocr-runtime/libhsakmt/src/fmm.c
+++ b/projects/rocr-runtime/libhsakmt/src/fmm.c
@@ -712,8 +712,8 @@ static void reserved_aperture_release(manageable_aperture_t *app,
 		/* Reset NUMA policy */
 		mbind(address, MemorySizeInBytes, MPOL_DEFAULT, NULL, 0, 0);
 
+		/* Unmap first to drop shmem refcount for MAP_SHARED allocations */
 		munmap(address, MemorySizeInBytes);
-
 		mmap_ret = mmap(address, MemorySizeInBytes, PROT_NONE,
 			MAP_ANONYMOUS | MAP_NORESERVE | MAP_PRIVATE | MAP_FIXED,
 			-1, 0);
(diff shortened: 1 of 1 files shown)


kentrussell (approved):

projects/rocr-runtime/libhsakmt/src/fmm.c
```diff
@@ -711,30 +711,13 @@ static void reserved_aperture_release(manageable_aperture_t *app,
 		/* Reset NUMA policy */
 		mbind(address, MemorySizeInBytes, MPOL_DEFAULT, NULL, 0, 0);
 
-		/* Remove any CPU mapping, but keep the address range reserved */
+		/* Unmap first to drop shmem refcount for MAP_SHARED allocations */
+		munmap(address, MemorySizeInBytes);
```
cfreeamd, line 722: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** Unconditional `munmap` introduces race window for MAP_PRIVATE allocations

```c
munmap(address, MemorySizeInBytes);
mmap_ret = mmap(address, MemorySizeInBytes, PROT_NONE,
    MAP_ANONYMOUS | MAP_NORESERVE | MAP_PRIVATE | MAP_FIXED, -1, 0);
```

The fix is correct for MAP_SHARED-backed allocations, but the unconditional `munmap` introduces a race window for MAP_PRIVATE ones. Previously, `mmap(MAP_FIXED)` over an existing MAP_PRIVATE PROT_NONE mapping was atomic (single VMA replace). By always calling `munmap` first, there is now a window where another thread could fault-in or mmap over this address between `munmap` and the subsequent `mmap`.

Whether this matters depends on what locks are held in `reserved_aperture_release`. If the aperture mutex is held across both calls, the window is low-risk. **Suggestion:** Gate the `munmap` on the allocation type, or document that only MAP_SHARED allocations reach this code path.

projects/rocr-runtime/libhsakmt/src/fmm.c
```diff
@@ -711,30 +711,13 @@ static void reserved_aperture_release(manageable_aperture_t *app,
...
-					/* Handle mmap failure gracefully, log if needed */
-					pr_err("Failed to remap memory after unmap\n");
-				}
-			} else {
-				/* Handle munmap failure if needed */
-				pr_err("Failed to unmap memory\n");
-			}
-		}
+		if (mmap_ret == MAP_FAILED)
+			pr_warn("Failed to reserve VA range %p: %s\n", address, strerror(errno));
```

**[Optional]** `pr_warn` severity underrepresents the failure case

```c
if (mmap_ret == MAP_FAILED)
    pr_warn("Failed to reserve VA range %p: %s\n", address, strerror(errno));
```

After `munmap` succeeds but `mmap(MAP_FIXED)` fails, the aperture's VA range is permanently gone — the aperture metadata still believes the range is reserved, but it isn't. Subsequent allocations may hand out a VA that belongs to someone else. `pr_err` would better communicate that the aperture is now in a broken state.

cfreeamd (commented):

Reviewed `libhsakmt/src/fmm.c`. The shmem-leak fix (adding `munmap` before `mmap(MAP_FIXED)` for MAP_SHARED mappings) is correct — the test evidence (246 GiB → 20 MiB) is compelling.

projects/rocr-runtime/libhsakmt/src/fmm.c
amd-willpalacek: Verified the locking and kept the munmap unconditional, with a
code comment added to document why it is safe (commit e698edd).

- Mutex is held across both calls. reserved_aperture_release()
carries the contract  **Assumes that fmm_mutex is locked on entry**.
, and every caller reaches it under aperture->fmm_mutex (the
allocation side, aperture_allocate_area_aligned(), is also always
under that mutex). So the munmap→mmap window is serialized against
concurrent allocations into the same aperture, matching your "if
the aperture mutex is held across both calls, the window is
low-risk" note.
- No MAP_PRIVATE path reaches this branch to gate on.
reserved_aperture_release() works on vm_area_t, which records no
mmap flags; the only discriminator is the aperture-level
is_cpu_accessible. In practice every CPU-accessible
reserved-aperture mapping is MAP_SHARED (fmm_map_to_cpu() uses
MAP_SHARED|MAP_FIXED; fmm_allocate_host_gpu() uses
MAP_ANONYMOUS|MAP_SHARED|MAP_FIXED), so there is no MAP_PRIVATE
allocation here to race.
- The window is pre-existing, not new. The old ENOMEM retry branch
(removed earlier in this PR as dead code) already did the
identical munmap→mmap sequence.

Documented the contract rather than gating by type, since the gate
has nothing to key off and the MAP_PRIVATE case doesn't occur.

projects/rocr-runtime/libhsakmt/src/fmm.c
amd-willpalacek: Agreed, changed to pr_err in commit e698edd. If munmap() succeeds
but the re-reserving mmap() fails, the VA range is genuinely lost
while the aperture bookkeeping still treats it as reserved, so a
later allocation could hand out that VA, an error-level
inconsistency, not a warning

cfreeamd (approved):

PR #7622, opened 2026-06-22
rocjitsu: fix kernarg preload DBT entries and refactor tests cmake

Groverkss:
Kernels supporting kernarg-preloading have two entry points, one for old firmware that doesn't support kernarg-preloading and a new one for firmware that supports it. It look something like:

```
^actual_entry:
  // load kernargs into user sgprs
  <words>
  s_branch ^common_entry
  s_nop
  ... // padding to make actual_entry and preload_entry 256 bytes apart
^preload_entry:
  // kernargs are already in user sgprs
  <words>
  s_branch ^common_entry
^common_entry:
  ...
```

this patch adds proper support for it. The way we do it is by treating `actual_entry` and `preload_entry` both as kernel entry points and builds new entry points when rewriting:

```
^new_compat_actual_entry:
  <descriptor ABI prologue words>
  s_branch ^actual_entry
  s_nop
  ... // new_compat_firmware_entry is exactly 256 bytes later
^new_firmware_entry:
  <descriptor ABI prologue words>
  s_branch ^firmware_entry
```

This prologue is always added in front of a kernel to preserve the kernarg preloading 256 byte firmware contract.

Also refactors tests/CMakeLists.txt to make some tests that weren't running before run properly.

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/binary_translator.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/binary_translator.cpp
@@ -25,7 +25,6 @@
 #include <algorithm>
 #include <cassert>
 #include <cstring>
-#include <limits>
 #include <memory>
 #include <optional>
 #include <span>
@@ -37,6 +36,8 @@
 namespace rocjitsu {
 
 namespace {
+
+inline constexpr uint64_t kKernargPreloadSkipBytes = 256;
 
 EncodingTranslateFn select_encoding_translator(rj_code_arch_t guest, rj_code_arch_t host) {
   if (guest == ROCJITSU_CODE_ARCH_CDNA4 && host == ROCJITSU_CODE_ARCH_RDNA4)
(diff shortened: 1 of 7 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/kernel_descriptor_translator.cpp
```diff
@@ -520,6 +524,9 @@ translate_one_descriptor(rj_code_arch_t guest_arch, rj_code_arch_t host_arch,
   result.entry_text_offset = entry_text_offset;
   result.target_entry_text_offset = entry_text_offset;
   result.target_body_entry_text_offset = entry_text_offset;
+  result.has_kernarg_preload = kernarg_preload_length(src) != 0;
+  result.kernarg_preload_entry_text_offset =
+      result.has_kernarg_preload ? entry_text_offset + 256u : entry_text_offset;
```
accauble: Nit: If this and `kKernargPreloadSkipBytes` could share a source, we might avoid headaches if this number ever changes.

accauble (approved):
Seems like a nice contained change. Even being really picky, my only comments would be:

* The PR description mentions a refactor to `tests/CMakeLists.txt`, but I don't see that. Did that happen somewhere else or maybe you meant `tests/dbt/translate_test.cpp`?
* I don't know if you think the skip distance will ever change away from 256, but there's a nice variable for it in `binary_translator.cpp` but then no variable for it in `kernel_descriptor_translator.cpp`.
* Some tests to verify that the new code fails correctly (e.g. if the branch is too far away) would be nice?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/binary_translator.cpp
```diff
@@ -399,7 +523,24 @@ TranslatedCodeObject BinaryTranslator::translate(const AmdGpuCodeObject &obj) {
...
+      if (!preload_body_entry) {
+        append_error(result.diagnostics, DiagnosticKind::KernelDescriptor,
+                     "kernarg preload firmware entry offset is not present in the relocated body",
+                     source_preload_entry);
+        return leave_unchanged();
+      }
+      if (!write_launch_stub(layout, layout.target_entry, layout.target_body_entry) ||
+          !write_launch_stub(layout, layout.target_entry + kKernargPreloadSkipBytes,
+                             *preload_body_entry)) {
+        append_error(result.diagnostics, DiagnosticKind::ResourceLimit,
```
kuhar: Please add one regression for the preload launch-window branch-range failure path: use a large preload body that makes either synthesized stub's `s_branch` exceed simm16, and assert the diagnostic plus unchanged code object. The existing overflow test covers the non-preload prologue path, but this two-stub path has its own error return.

kuhar (commented):
Overall LGTM. Just some agent findings.

In `emulation/rocjitsu/tests/dbt/translate_test.cpp:1387`, the test only exercises the empty-prologue path: CDNA4 -> CDNA3 leaves `prologue_words` empty, so `write_launch_stub` emits just an `s_branch`. Please add a CDNA4 -> RDNA4 kernarg-preload case with workgroup-id SGPRs enabled and assert both synthesized stubs (+0 and +256) contain the descriptor ABI prologue before branching.

It would also be worth making the descriptor entry move, for example with a preceding kernel/body, so the `kernel_code_entry_byte_offset` assertion proves redirection instead of passing because source and target are both 0.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/kernel_descriptor_translator.cpp
Groverkss: I will fix this. kernarg preload skip bytes are always fixed, this is a known firmware contract.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/binary_translator.cpp
Groverkss: I'm going to make this an assert. This will never happen, because we emit blocks in source order and our entry blocks have a fixed size.

PR #7624, opened 2026-06-22
[rocprofiler-compute] refactor calc_metrics_data

xuchen-amd:
## Motivation

Refactor `calc_metrics_data` to compute table-level fields once per table instead of redundantly per row, improving clarity and reducing repeated work.

The original implementation computed several values redundantly for every metric row:

1. **`table_name` / `sub_table_name`**: Parsed from `metric_id.split(".")` on every row, even though all rows in the same table share the same table name
2. **`value_columns`**: Called `metric_df.drop(columns=...).columns` for every row in the nested comprehension, recomputing the same column list repeatedly
3. **`non_expression_columns`**: Rebuilt the list on every workload iteration

## Technical Details

- Compute `table_name`, `sub_table_name`, `value_columns` once per table, not per row
- Use `table_id // 100 * 100` instead of parsing `metric_id.split(".")`
- Keep `MetricInfoRow` / `ExpressionRow` NamedTuples for schema safety
- Replace nested list comprehensions with explicit loops for clarity

## JIRA ID

N/A

## Test Plan

- Existing test_calc_metrics_data_empty_filter_preserves_schema passes
- CI passes

## Test Result

- [x] Existing test_calc_metrics_data_empty_filter_preserves_schema passes
- [ ] CI passes

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
+++ b/projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
@@ -794,7 +794,7 @@ class db_analysis(OmniAnalyze_Base):
             gfx_arch = self._runs[workload_path].sys_info.iloc[0]["gpu_arch"]
             arch_config = self._arch_configs[gfx_arch]
 
-            # Build table_id -> title map (e.g. 201 -> "Wavefront").
+            # Build table_id -> title map (e.g. 700 -> "Wavefront", 701 -> "Wavefront Launch Stats").
             table_names_map: dict[int, str] = {}
             for panel_config in arch_config.panel_configs.values():
                 table_names_map[panel_config["id"]] = panel_config["title"]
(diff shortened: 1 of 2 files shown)


abchoudh-amd (approved):

vedithal-amd (changes requested):
reviewing

vedithal-amd (approved):
LGTM. Verified behavior-preserving: `table_id // 100 * 100` and `table_id` equal the old `metric_id.split(".")` parsing.

Follow-up pushed:
- **List comprehensions over a per-table context** — faster because it computes table fields/rows once per table (not per row) and skips the repeated `.append` attribute lookup, and reads as a single expression.
- `for table in source.values()` instead of `next(iter(...))` — more readable for the single-entry dicts.
- **Added populated-path coverage** to the `calc_metrics_data` test — the mapping was untested (branch had no net test changes), so this guards against future regressions in the rewritten logic.

PR #7627, opened 2026-06-22
[rocprofiler-systems] Update TheRock tarballs to use new index

jbonnell-amd:
## Motivation

Update tarball logic in Dockerfiles to use the new index to get the latest tarballs from TheRock

## Technical Details

- Updated `Dockerfile.*.ci` files in `rocprofiler-systems` to grab tarballs from `rocm.nightlies.amd.com` instead of S3 buckets directly
- Updated `rocprofiler-systems-continuous-integration.yml` workflow file to output `therock_manifest.json` to show detailed information on TheRock version used, in addition to some cleanup in the ROCm install step
- Updated `projects/rocprofiler-systems/docker/build-docker-ci.sh` to remove references to removed GPU_TARBALL argument

## Test Plan

- Ensure no regression issues with new Docker tarballs
- Make sure that `therock_manifest.json` output matches the latest from TheRock and not an older version

## Test Result

- Workflows pass, outputting the latest manifest information in `therock_manifest.json`

## Submission Checklist

### DIFF
--- a/.github/workflows/rocprofiler-systems-continuous-integration.yml
+++ b/.github/workflows/rocprofiler-systems-continuous-integration.yml
@@ -152,6 +152,10 @@ jobs:
           echo "ROCm installed to: ${{ env.ROCM_PATH }}"
           echo "✅ ROCm Installation Complete!"
 
+      - name: Output TheRock Manifest
+        run: |
+          cat ${{ env.ROCM_PATH }}/share/therock/therock_manifest.json
+
       - name: Install Python Test Dependencies
         timeout-minutes: 10
         shell: bash
--- a/.github/workflows/rocprofiler-systems-continuous-integration.yml
+++ b/.github/workflows/rocprofiler-systems-continuous-integration.yml
@@ -146,9 +146,8 @@ jobs:
         run: |
           set -e
           mkdir ${{ env.ROCM_PATH }}
-          TARBALL_ROCM_VERSION=$(ls /opt/*.tar.gz | grep -Eo '*([0-9]+\.[0-9]+\.[0-9]+)*')
+          TARBALL_ROCM_VERSION=$(ls /opt/*.tar.gz | grep -Eo '[0-9]+\.[0-9]+\.[0-9]+')
           tar -xf ${{ env.ROCM_PATH }}-${TARBALL_ROCM_VERSION}-multiarch.tar.gz -C ${{ env.ROCM_PATH }}
-          ln -s ${{ env.ROCM_PATH }}-${{ env.ROCM_VERSION }} ${{ env.ROCM_PATH }}
           echo "ROCm installed to: ${{ env.ROCM_PATH }}"
           echo "✅ ROCm Installation Complete!"
 
--- a/.github/workflows/rocprofiler-systems-ghcr.yml
+++ b/.github/workflows/rocprofiler-systems-ghcr.yml
@@ -100,22 +100,6 @@ jobs:
           fi
           echo "docker_file=${DOCKER_FILE}" >> $GITHUB_OUTPUT
 
-      - name: Get the latest build of The Rock tarball
-        id: therock
-        run: |
-          sudo apt-get install -y python3-pip
-          python3 -m pip install -U pip
-          python3 -m pip install -U awscli
-          export PATH=~/.local/bin:$PATH
-          KEY=$(aws s3api list-objects-v2 \
-                --bucket therock-nightly-tarball \
-                --no-sign-request \
-                --output json \
-                --query "sort_by(Contents[?contains(Key, 'linux-${{ matrix.gpu }}')], &LastModified)[-1].Key")
-          KEY=${KEY//\"/}
-          test -n "$KEY" || { echo "No ${{ matrix.gpu }} tarball found"; exit 1; }
-          echo "tarball=${KEY}" >> $GITHUB_OUTPUT
-
       - name: Extract metadata (tags, labels) for Docker
         id: meta
         uses: docker/metadata-action@030e881283bb7a6894de51c315a6bfe6a94e05cf # v6.0.0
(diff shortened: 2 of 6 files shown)


dgaliffiAMD (approved):
Thank you. The changes look good overall. I just left a couple of comments for feedback. 
I think the update to the `build-docker-ci` comment can be addressed now. The second can be a follow up pr.

dgaliffiAMD (commented):
.

dgaliffiAMD:
### 1. `build-docker-ci.sh` still references the removed `GPU_TARBALL` arg

  This PR drops `ARG GPU_TARBALL` from all three Dockerfiles and makes tarball selection self-contained (the Dockerfile queries the index using only `GPU_TYPE`). However, the local build script `docker/build-docker-ci.sh` still treats `GPU_TARBALL` as a live input:

  - Line 15: `: ${GPU_TARBALL:=""}`
  - Lines 111-115: the `--gpu-tarball` CLI option
  - Line 197: `--build-arg GPU_TARBALL=${GPU_TARBALL}`

  Two effects after this change:

  1. `docker build` now prints `WARNING: build-arg [GPU_TARBALL] was not consumed` on every local build (cosmetic).
  2. More importantly, `--gpu-tarball <key>` becomes a **silent no-op** - it's still parsed and passed as a build-arg, but nothing consumes it, so a developer who tries to pin a specific tarball will silently get the auto-selected latest instead.

  Suggest removing all three references so the local script matches the new index-based contract (and the GHCR workflow, which already dropped `GPU_TARBALL`). Not blocking - `build-docker-ci.sh` isn't in this diff - but worth folding in here or as a fast follow.

  ---

  ### 2. Optional / follow-up: consolidate ROCm provisioning into the Dockerfile

  Not for this PR - just flagging a simplification while this area is being touched.

  Right now ROCm provisioning is split across two places: the Dockerfile **downloads** the tarball to `/opt/rocm-<ver>-multiarch.tar.gz` at image-build time, and the workflow's `Install Latest Nightly ROCm` step later **extracts + symlinks** it at test time. Since the download already happens at
  image-build time, the split doesn't buy any freshness - both halves are pinned to the image build regardless. Doing download + extract + cleanup in the Dockerfile alone (or entirely in the workflow) would keep provisioning in one place and shrink the workflow.

  If you do consolidate, the current `Install Latest Nightly ROCm` step has a few latent bugs worth fixing in the same pass (all pre-existing, not introduced here):

  - `grep -Eo '*([0-9]+\.[0-9]+\.[0-9]+)*'` - the leading/trailing `*` aren't valid ERE quantifiers; version extraction should be `grep -Eo '[0-9]+\.[0-9]+\.[0-9]+'`.
  - `mkdir ${ROCM_PATH}` followed by `ln -s ${ROCM_PATH}-${ROCM_VERSION} ${ROCM_PATH}` - `/opt/rocm` already exists as a directory, so the symlink lands *inside* it rather than replacing it.
  - `${{ env.ROCM_VERSION }}` is referenced but doesn't appear in the job's `env:` block (only `ROCM_PATH` is defined), so it likely expands empty.

  These currently don't break the build because tests reference `${ROCM_PATH}` (the real dir from `mkdir` + `tar -C`), leaving the symlink effectively dead - but they'd bite anyone who later relies on the versioned path.

kcossett-amd:
> .

👀

dgaliffiAMD:
> > .
> 
> 👀

Sorry, the comments from VS Code didn't post properly.

jbonnell-amd:
> ### 1. `build-docker-ci.sh` still references the removed `GPU_TARBALL` arg
> This PR drops `ARG GPU_TARBALL` from all three Dockerfiles and makes tarball selection self-contained (the Dockerfile queries the index using only `GPU_TYPE`). However, the local build script `docker/build-docker-ci.sh` still treats `GPU_TARBALL` as a live input:
> 
> * Line 15: `: ${GPU_TARBALL:=""}`
> * Lines 111-115: the `--gpu-tarball` CLI option
> * Line 197: `--build-arg GPU_TARBALL=${GPU_TARBALL}`
> 
> Two effects after this change:
> 
> 1. `docker build` now prints `WARNING: build-arg [GPU_TARBALL] was not consumed` on every local build (cosmetic).
> 2. More importantly, `--gpu-tarball <key>` becomes a **silent no-op** - it's still parsed and passed as a build-arg, but nothing consumes it, so a developer who tries to pin a specific tarball will silently get the auto-selected latest instead.
> 
> Suggest removing all three references so the local script matches the new index-based contract (and the GHCR workflow, which already dropped `GPU_TARBALL`). Not blocking - `build-docker-ci.sh` isn't in this diff - but worth folding in here or as a fast follow.
> 
> ### 2. Optional / follow-up: consolidate ROCm provisioning into the Dockerfile
> Not for this PR - just flagging a simplification while this area is being touched.
> 
> Right now ROCm provisioning is split across two places: the Dockerfile **downloads** the tarball to `/opt/rocm-<ver>-multiarch.tar.gz` at image-build time, and the workflow's `Install Latest Nightly ROCm` step later **extracts + symlinks** it at test time. Since the download already happens at image-build time, the split doesn't buy any freshness - both halves are pinned to the image build regardless. Doing download + extract + cleanup in the Dockerfile alone (or entirely in the workflow) would keep provisioning in one place and shrink the workflow.
> 
> If you do consolidate, the current `Install Latest Nightly ROCm` step has a few latent bugs worth fixing in the same pass (all pre-existing, not introduced here):
> 
> * `grep -Eo '*([0-9]+\.[0-9]+\.[0-9]+)*'` - the leading/trailing `*` aren't valid ERE quantifiers; version extraction should be `grep -Eo '[0-9]+\.[0-9]+\.[0-9]+'`.
> * `mkdir ${ROCM_PATH}` followed by `ln -s ${ROCM_PATH}-${ROCM_VERSION} ${ROCM_PATH}` - `/opt/rocm` already exists as a directory, so the symlink lands _inside_ it rather than replacing it.
> * `${{ env.ROCM_VERSION }}` is referenced but doesn't appear in the job's `env:` block (only `ROCM_PATH` is defined), so it likely expands empty.
> 
> These currently don't break the build because tests reference `${ROCM_PATH}` (the real dir from `mkdir` + `tar -C`), leaving the symlink effectively dead - but they'd bite anyone who later relies on the versioned path.

Thank you! I addressed both feedback in this PR, since the changes were fairly minimal and related to the tarball install anyways.

PR #6686, opened 2026-06-02
[profiler-hub]: policy-based SQLite backend with mockable sqlite3 seam

adjordje-amd:
## Motivation

profiler-hub statically links sqlite3 into libprofiler-hub.so. To make the backend unit-testable and to enable a later step that seals the sqlite3 symbols so they cannot collide with other sqlite3 versions present on TheRock, the SQLite C API needs to be a swappable compile-time seam. This PR introduces that seam. It is a pure refactor with no behavior change; the symbol-hiding step is a follow-up.

## Technical Details

sqlite_backend becomes basic_sqlite_backend<SqlitePolicy>. The production policy sqlite_api_policy is the only translation unit that includes sqlite3.h and forwards each operation to the sqlite3 C API. A type alias using sqlite_backend = basic_sqlite_backend<sqlite_api_policy> keeps every existing call site unchanged. Heavy members (create, constructor, destructor, initialize_schema, execute, flush, discover_uuids) are defined out of line in sqlite_backend_impl.hpp and explicitly instantiated in sqlite_backend.cpp, keeping sqlite3 and the schema machinery out of the public headers. A gmock-backed mock_sqlite3 policy plus a small self-test are added so the backend can be driven without a real database in future tests.

## JIRA ID

N/A

## Test Plan

Full Ninja Release build, then run the complete unit suite via ctest.

## Test Result

Build clean; 260 of 260 tests pass (257 existing plus 3 mock self-tests). No behavior change.

## Submission Checklist

- [x] Builds cleanly
- [x] Unit tests pass
- [ ] sqlite3 symbol-hiding follow-up (separate PR)

PR #7631, opened 2026-06-22
[profiler-hub] Add profilers section to pr_category_label.py

jbonnell-amd:
## Motivation

We noticed that `profilers/profiler-hub` PRs do not automatically add the GitHub label for the project

## Technical Details

- Updated `pr_category_label.py` to add a new section for `profilers`
  - Since `profiler-hub` lives in `profilers/*` instead of `projects/*`, the current logic was unable to pick these changes up
  - This will also extend to other profiler components once they are migrated from `projects/*` to `profilers/*` directory
- Various pre-commit formatting changes to `pr_category_label.py` 
- Updated `labeller.yml` to use new version of `actions/labeler` that uses Node24 to address Node20 warnings in CI

## Test Plan

- Local testing using #6686 as a reference profiler-hub PR

## Test Result

- Using #6686 as a reference, testing locally using `--dry-run` option shows that the `profilers: profiler-hub` label would be added
image

## Submission Checklist

### DIFF
--- a/.github/scripts/pr_category_label.py
+++ b/.github/scripts/pr_category_label.py
@@ -68,7 +68,7 @@ def compute_desired_labels(file_paths: list) -> set:
             elif parts[0] == "shared":
                 desired_labels.add(f"shared: {parts[1]}")
             elif parts[0] == "profilers":
-                desired_labels.add(f"profilers: {parts[1]}")
+                desired_labels.add(f"project: {parts[1]}")
     logger.debug(f"Desired labels based on changes: {desired_labels}")
     return desired_labels
 
--- a/.github/scripts/pr_category_label.py
+++ b/.github/scripts/pr_category_label.py
@@ -38,14 +38,24 @@ from github_cli_client import GitHubCLIClient
 
 logger = logging.getLogger(__name__)
 
+
 def parse_arguments(argv: Optional[List[str]] = None) -> argparse.Namespace:
     """Parse command-line arguments."""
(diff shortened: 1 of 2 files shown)


.github/scripts/pr_category_label.py
```diff
@@ -57,10 +67,15 @@ def compute_desired_labels(file_paths: list) -> set:
                 desired_labels.add(f"project: {parts[1]}")
             elif parts[0] == "shared":
                 desired_labels.add(f"shared: {parts[1]}")
+            elif parts[0] == "profilers":
+                desired_labels.add(f"profilers: {parts[1]}")
```
jbonnell-amd: This will add a new "profilers: profiler-hub" label for profiler-hub PRs.

However, if we want to have profiler-hub PRs labeled using the existing "projects: profiler-hub" label, we can modify this label to use "projects: profiler-hub" instead.

dgaliffiAMD (approved):
nit-pick, but I would add to the technical details - "Various linting fixes", or something to that affect.

geomin12 (approved):

.github/scripts/pr_category_label.py
jbonnell-amd: I've updated the label to be "projects: profiler-hub" instead of "profilers: profiler-hub".

Once other profiler components move to this directory, we can consider then if the "profilers: <PROFILER_PROJECT>" template makes more sense. For now, this will keep it consistent with the existing profiler components.

PR #7633, opened 2026-06-22
Guard kpack splitting against empty target matches

dan-garvey:
## Summary
- fail artifact splitting when detected fat binaries produce no kpack entries after --gpu-targets filtering
- strengthen generic artifact verification to scan all binaries and require kpack markers plus HIPK wrappers
- fail verification loudly when an ELF/COFF binary is detected but cannot be inspected
- add regression coverage for the empty-filter guard, non-.so generic binaries, and unreadable ELF binaries

## Validation
- PYTHONPATH=shared/kpack/python uv run --allow-insecure-host pypi.org --allow-insecure-host files.pythonhosted.org --with pytest --with msgpack --with zstandard --with pyelftools pytest shared/kpack/tests/test_artifact_splitter_integration.py::TestArtifactSplitterIntegration::test_gpu_targets_rejects_generic_when_all_fat_binary_kernels_filtered shared/kpack/tests/test_artifact_splitter_integration.py::TestArtifactSplitterIntegration::test_verifier_rejects_raw_generic_executable_fat_binary shared/kpack/tests/test_artifact_splitter_integration.py::TestArtifactSplitterIntegration::test_verifier_rejects_unreadable_generic_binary
- PYTHONPATH=shared/kpack/python uv run --allow-insecure-host pypi.org --allow-insecure-host files.pythonhosted.org --with msgpack --with zstandard --with pyelftools python -m compileall -q shared/kpack/python/rocm_kpack shared/kpack/tests/test_artifact_splitter_integration.py
- Verified the guard on the known bad rocWMMA generic tree from ROCm/rocm-libraries#8660 by scanning the real artifact and mocking only the unavailable clang-offload-bundler output as filtered gfx908 code objects: split fails with "Found 37 fat binaries, but no device code objects matched --gpu-targets (gfx1030)".
- Verified artifact verifier passes the known good PR #5971 generic artifact: 37 kpack-transformed, 3 host-only, 0 failed.
- Verified artifact verifier rejects the known bad PR #6020 generic artifact: raw PROGBITS .hip_fatbin, missing .rocm_kpack_ref, HIPF wrappers.
- Full test_artifact_splitter_integration.py currently fails in this checkout because git-lfs is not installed and bundled binary fixtures are LFS pointer text files.

Fixes ROCm/rocm-libraries#8660.


bsyrowik:
I am not qualified to review the code changes, but the intent looks good to me!

dan-garvey:
@nunnikri @ScottTodd @marbre Related to https://github.com/ROCm/TheRock/pull/6042, would like a review

shared/kpack/python/rocm_kpack/tools/verify_artifacts.py
```diff
@@ -415,19 +445,167 @@ def _check_kpack_archives(self, artifacts: list[Path]) -> None:
...
+        if fatbin_section.raw_size == 0:
+            return "raw-size=0"
+        if fatbin_section.raw_size < fatbin_section.virtual_size:
+            return (
+                f"raw-size={fatbin_section.raw_size}, "
+                f"virtual-size={fatbin_section.virtual_size}"
+            )
+        return (
+            f"raw-size={fatbin_section.raw_size}, "
+            f"virtual-size={fatbin_section.virtual_size}"
```
scchan: It's returning the same thing regardless of what the previous condition is?

scchan (dismissed):

shared/kpack/python/rocm_kpack/tools/verify_artifacts.py
dan-garvey: good catch, thanks

nunnikri (commented):
Overall changes looks good. But might need more testing as we are moving from readelf based checks to directly calling ELF surgery functions. Which might be faster. But need more testing. 

Usually for these kind of issues, would suggest to add binaries which can reproduce the failures so that in future its not seen again. Adding the binaries here in shared/kpack/test_assets/bundled_binaries/linux/cov5/ and having test based on that will be required

dan-garvey:
Thanks for the suggestion, I added a test that reuses an existing binary

nunnikri:
> Thanks for the suggestion, I added a test that reuses an existing binary

Thanks @dan-garvey the changes look good. But there are no test for COEFF cases, if we can add those as well. Its good to merge. 
Additional good to have test case is for case:
    .hip_fatbin is NOBITS (section zeroed — looks transformed)
    .rocm_kpack_ref is present (marker is there)
    But wrapper magic in .hipFatBinSegment still contains HIPF_MAGIC instead of HIPK_MAGIC_

nunnikri (approved):
Please take care of the comments before merge

shared/kpack/python/rocm_kpack/tools/verify_artifacts.py
```diff
@@ -147,55 +177,56 @@ def _check_fat_binary_conversion(self, artifacts: list[Path]) -> None:
...
+                file_size = binary_path.stat().st_size
                 size_mb = file_size / (1024 * 1024)
-                rel_path = so_file.relative_to(artifact)
+                rel_path = binary_path.relative_to(artifact)
 
-                # Check if has .hip_fatbin section
-                section_info = self._get_hip_fatbin_section(so_file)
-                if section_info is None:
-                    host_only_binaries.append((rel_path, size_mb))
+                inspection, is_binary = self._inspect_fat_binary_conversion(
```
scchan, line 193: Does it support processing multiple fatbin bundles in a fatbin section?

dan-garvey:
> > Thanks for the suggestion, I added a test that reuses an existing binary
> 
> Thanks @dan-garvey the changes look good. But there are no test for COEFF cases, if we can add those as well. Its good to merge. Additional good to have test case is for case: .hip_fatbin is NOBITS (section zeroed — looks transformed) .rocm_kpack_ref is present (marker is there) But wrapper magic in .hipFatBinSegment still contains HIPF_MAGIC instead of HIPK_MAGIC_

I added COFF coverage using the existing Windows cov5 assets: raw untransformed COFF rejection, transformed COFF pass, stale HIPF wrapper rejection, and unreadable COFF rejection. I also added the suggested stale-HIPF case for ELF.

shared/kpack/python/rocm_kpack/tools/verify_artifacts.py
dan-garvey: The wrapper path does handle multiple wrappers by reading the full wrapper section in WRAPPER_SIZE chunks. I added coverage for a multi-arch ELF fatbin fixture, a multi-wrapper ELF fixture, and a transformed multi-wrapper pass case.

scchan (commented):
LGTM but someone else should approve

shared/kpack/python/rocm_kpack/artifact_splitter.py
ScottTodd, line 1: I only see unit tests as part of the validation steps in the PR description.

Did you check if this fixes (or would catch?) https://github.com/ROCm/rocm-libraries/issues/8660 through full CI testing? I see that we now have
* https://github.com/ROCm/rocm-systems/actions/workflows/therock-multi-arch-ci.yml 
* https://github.com/ROCm/rocm-systems/blob/develop/.github/workflows/therock-multi-arch-ci.yml

but those aren't yet fully working (cc @geomin12 )?

You can also test over in https://github.com/ROCm/TheRock as needed (via a redirected submodule or a patch file).

---

If I'm reading the PR description correctly, it sounds like this fails more loudly on errors without changing behavior, but is the code currently error-free, making this safe?

dan-garvey: This is both a preventive behavior change and a verifier safety net. The splitter now refuses to emit an untransformed generic artifact when fat binaries exist but `--gpu-targets` filters out every code object. The verifier then catches raw/stale generic artifacts if one is produced elsewhere.

For ROCm/rocm-libraries#8660 specifically, I validated against the known bad and known good rocWMMA artifacts: the bad generic is rejected, the good generic passes, and TheRock#6042 prevents unsupported rocWMMA shards from producing rocwmma_test_generic in the first place. Full multi-arch CI is still not the only signal here, but the issue path is covered by direct artifact validation plus the TheRock configure checks.

ScottTodd: A full CI run through TheRock's multi-arch CI would also check if artifacts other than rocWMMA have similar issues.

dan-garvey: I don't have permissions to trigger builds on either repo

dan-garvey: https://github.com/ROCm/TheRock/pull/6042 pulls this in as a submodule bump, so triggering it there should work

ScottTodd: I approved the workflow runs on that PR

dan-garvey: None of the tests failing on that PR appear to be related to this one. Are we okay to land this?

ScottTodd (approved):
Trusting the CI coverage here, I'm not that familiar with the intricacies of this code and we've seen a few other issues in it lately that only appeared in some specific circumstances (e.g. different Linux distros)

amd-chiranjeevi:
@dan-garvey 
reverting this  PR as it is blocking the rocm-systems bump 
more details here, Issue: https://github.com/ROCm/rocm-systems/issues/7862
revert PR: https://github.com/ROCm/rocm-systems/pull/7863

PR #7642, opened 2026-06-22
fix(rocprofiler-sdk): info log cleanup

meserve-amd:
- Demotes several log messages from info to trace level
- Removes extra newlines from kfd log messages
- Promotes kfd dropped events log message to warning
- Sanitizes newline characters on event_str coming from kfd- 

## Motivation

I found the log messages coming from our tests were very crowded, so I trimmed down some of the worst offenders.

## Technical Details

In most cases, there are simple demotions from info to trace for non-exceptional events. kfd.cpp's log messages all also had additional newlines which are not needed in our macro.

The KFD dropped events message is promoted to warning, this shouldn't usually happen.

The KFD event_str sometimes had newline characters in it which would be printed directly. This change makes it so they are printed as byte codes instead.

## JIRA ID

JIRA ID: AIPROFSDK-929

## Test Plan

Just CI, no new tests added

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hip/stream.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hip/stream.cpp
@@ -553,7 +553,7 @@ update_table(Tp* _orig, std::integral_constant<size_t, OpIdx>)
             }
             else
             {
-                ROCP_INFO << fmt::format(
+                ROCP_TRACE << fmt::format(
                     "[hip stream] {} has been designated as a stream set function", _info.name);
                 _func = create_read_functor<TableIdx, OpIdx>(_func);
             }
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hip/stream.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hip/stream.cpp
@@ -553,7 +553,7 @@ update_table(Tp* _orig, std::integral_constant<size_t, OpIdx>)
             }
             else
             {
-                ROCP_TRACE << fmt::format(
+                ROCP_INFO << fmt::format(
                     "[hip stream] {} has been designated as a stream set function", _info.name);
                 _func = create_read_functor<TableIdx, OpIdx>(_func);
             }
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/kfd/kfd.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/kfd/kfd.cpp
@@ -701,8 +701,7 @@ kfd_readlines(const std::string_view str, void(handler)(std::string_view))
         assert(char_count > 0);
         std::string_view event_str{cursor, char_count};
 
-        if(VLOG_IS_ON(ROCP_LOG_LEVEL_TRACE))
-        {
+        ROCP_TRACE << fmt::format("KFD event: [{}]", [&]() {
             auto escaped = std::string{};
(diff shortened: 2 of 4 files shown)


meserve-amd:
>     * ❌ Unit Test

No unit tests will be written for non-critical log message log levels or their contents; these are only for developers. Further, if any existing test relies on these messages we would see a failure in the normal test runs. Removing label and ignoring check.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hip/stream.cpp
```diff
@@ -553,7 +553,7 @@ update_table(Tp* _orig, std::integral_constant<size_t, OpIdx>)
             }
             else
             {
-                ROCP_INFO << fmt::format(
+                ROCP_TRACE << fmt::format(
```
ApoKalipse-V: is this one necessary?

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/kfd/kfd.cpp
```diff
@@ -631,8 +633,8 @@ parse_event<KFD_EVENT_DROPPED_EVENT>(const agent_id_map_t&, std::string_view str
         return {};
     }
 
-    ROCP_TRACE << fmt::format(
-        "Dropped events [ ts: {} pid: {} dropped count: {} ] \n", e.timestamp, e.pid, e.count);
+    ROCP_WARNING << fmt::format(
```
ApoKalipse-V, line 636: how often does this happen?

meserve-amd: We don't have good data on that... This should only fire if the buffer in KFD for these events is full (notifying us that some messages were dropped). If we (profiler) are keeping up it shouldn't happen. I'd like to keep as warning because this ultimately means these events won't be in our outputs.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hip/stream.cpp
meserve-amd: This reduces some noise but I suppose it only happens once per app. I'm also less sure of who is using it. I'll leave it for now

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/kfd/kfd.cpp
```diff
@@ -699,7 +701,19 @@ kfd_readlines(const std::string_view str, void(handler)(std::string_view))
         assert(char_count > 0);
         std::string_view event_str{cursor, char_count};
 
-        ROCP_INFO << fmt::format("KFD event: [{}]", event_str);
+        if(VLOG_IS_ON(ROCP_LOG_LEVEL_TRACE))
```
ApoKalipse-V: why is this necessary here? You dont see this check often

ApoKalipse-V (approved):
One strange part, else LGTM

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/kfd/kfd.cpp
meserve-amd: This work is unnecessary if the logging message doesn't fire, so I made it skip otherwise.

meserve-amd: Moved this inside the ROCP_TRACE statement after discussion

PR #1440, opened 2025-10-17
Fix dimension mismatch for multi-GPU systems with identical architect…

bwelton:
…ures

This change addresses an issue where counter dimensions were incorrectly shared across all GPU agents with the same architecture name, even when those agents had different hardware configurations (e.g., different CU counts).

Changes:
- Updated getBlockDimensions() to accept agent ID instead of architecture name
- Made dimension cache agent-specific instead of architecture-specific
- Updated set_dimensions() in AST evaluation to use specific agent ID
- Modified all API functions to handle agent-specific dimension lookups
- Updated tests to work with agent-specific dimensions

This fix ensures that dimensions accurately reflect the actual hardware configuration of each individual GPU agent, preventing dimension mismatches in multi-GPU systems where GPUs share the same architecture but have different physical configurations.

Counter ID Representation Changes:
- Modified counter_id encoding to include agent information in bits 37-32
- Agent logical_node_id is encoded as (value + 1) to ensure agent 0 is detectable
- Counter records internally store only 16-bit base metric IDs (bits 15-0)
- Tool reconstructs agent-encoded counter IDs from base metric ID & agent info
- Instance record counter_id field uses bitwise AND mask to extract base metric ID (counter_id.handle & 0xFFFF) to fit in 16-bit storage
- Output generators (CSV, JSON, Perfetto) use agent-encoded IDs for consistency
- Updated counter_config.cpp and metrics.cpp to extract base metric ID when needed
- All counter lookups now properly handle agent-encoded vs base metric IDs

This ensures counter IDs are consistent between metadata and output records while maintaining compact storage in instance records.

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #7645, opened 2026-06-23
[kpack] Keep xnack-variant Tensile kernels in per-arch artifacts

tony-davis:
## Motivation

Multi-arch ROCm packages were silently dropping xnack-variant Tensile kernel objects for datacenter GPUs, which broke rocBLAS at runtime on MI210 (gfx90a) and MI308 (gfx942) — the failures tracked in ROCM-25535. The kernels build correctly; they were lost at artifact-split time, so the defect only surfaces in packaged multi-arch builds, not in a single-arch developer build.

## Technical Details

The artifact splitter has two filtering paths. The fat-binary path normalizes architecture ids before matching against `--gpu-targets`; the database-handler path (rocBLAS/hipBLASLt/MIOpen Tensile files) did not. Tensile encodes xnack feature variants in filenames with a hyphen (`TensileLibrary_lazy_gfx90a-xnack-.hsaco`), so `DatabaseHandler.detect()` returns `gfx90a-xnack-`. `--gpu-targets` and the per-arch shard key are the bare base arch (`gfx90a`). The string `gfx90a-xnack-` is not in `{gfx90a}`, so the kernel failed the membership check in `FileClassificationVisitor` and was excluded from the per-arch artifact and from the generic artifact — and no separate `gfx90a-xnack-` build job exists to emit it elsewhere, so it was simply gone.

This PR adds a `base_arch()` helper that reduces an architecture id to its bare base, handling both the ELF colon form (`gfx942:xnack+`) and the Tensile hyphen form (`gfx90a-xnack-`). The database-handler path now normalizes the detected arch with `base_arch()` before the `gpu_targets` membership check and before using it as the shard key, so xnack+/- variants collapse onto the base-arch shard. The `gpu_targets` set construction is normalized the same way for symmetry.

Related (adjacent, not superseded by this change):

- ROCm/rocm-systems#7633 — guards the fat-binary split path against empty target matches; different path, complementary.
- ROCm/TheRock#6039 — fixes the fetch side (`fetch_artifacts.py`) to match colon-form xnack artifact variants; that is the ASAN/test-fetch consumer, not the nightly multi-arch tarball split addressed here.

## JIRA ID

Resolves ROCM-25535

## Test Plan

- Unit/integration (`shared/kpack`): new regression test `test_gpu_targets_keeps_xnack_variant_database_files` builds a `RocBLASHandler` layout with a bare `.dat` plus `gfx90a-xnack-` and `gfx90a-xnack+` kernels and an out-of-scope `gfx1100`, splits with `gpu_targets=["gfx90a"]`, and asserts the xnack variants land in the `gfx90a` shard, do not leak into generic, and do not spawn a spurious `gfx90a-xnack-` shard.
- Reproduction evidence (defect-fix): with the source fix stashed, the new test fails exactly as ROCM-25535 describes — both `gfx90a-xnack±` kernels are excluded from generic and absent from the shard (only the `.dat` moves). With the fix applied it passes.
- Command: `PYTHONPATH=shared/kpack/python python -m pytest shared/kpack/tests/test_database_handlers.py shared/kpack/tests/test_artifact_splitter_integration.py`

## Test Result

- Unit/integration: 84 relevant tests pass (database-handler + splitter-integration, including the new test). PR CI kpack lanes are green on ubuntu (3.10/3.12) and windows (3.10/3.12).
- End-to-end packaging validation (the path where the kernels were dropped): the fix was built through TheRock multi-arch CI ([run 27993529661](https://github.com/ROCm/TheRock/actions/runs/27993529661), building gfx90a + gfx94x together) and compared against a current unfixed `develop`/`main` multi-arch build ([run 27975666776](https://github.com/ROCm/TheRock/actions/runs/27975666776)), with both fetched via TheRock's installer for an apples-to-apples comparison.
- Static artifact diff (gfx90a): the unfixed build ships `librocblas.so` but the entire `lib/rocblas/library` tree is absent; the fix restores all 487 files, including the 174 hyphen-form `gfx90a-xnack±` kernels the un-normalized filter was dropping.

| build | gfx90a library files | gfx90a-xnack kernels |
| --- | --- | --- |
| baseline (27975666776) | 0 | 0 |
| fix (27993529661) | 487 | 174 |

- On-device functional (gfx90a / MI210-class, `rocminfo`-confirmed, `librocblas.so.5` resolving into the fetched tree): the unfixed build fails every Tensile-dispatching rocBLAS sample (sgemm, sgemm-strided-batched, c-dgeam, sscal, …) with `Cannot read .../library/TensileLibrary.dat`; only the non-Tensile samples pass. The fix build passes all non-fortran samples (fortran-* skipped for an unrelated missing `libgfortran.so.5`, identical in both builds).

| build | rocBLAS samples |
| --- | --- |
| baseline | 2 passed, 10 failed |
| fix | 12 passed, 0 failed |

- Scope confirmation: gfx942/MI308 is not affected — its Tensile kernels are packaged with plain `gfx942` names (no hyphen-form xnack token), so the buggy filter never matched them; on a gfx942 box both builds behaved identically. This matches the gfx942-passes / gfx90a-fails asymmetry reported on the original nightly.
- Not run locally: the fat-binary integration tests requiring LLVM `readelf`/`clang-offload-bundler` and `test_wheel_splitter.py` (needs optional `rocm-bootstrap`) — environment-limited, unrelated to this change.

## Risk

Level 3. Core build/packaging path; the change is narrow architecture-independent string normalization, well-covered by a reproducing unit test and confirmed end-to-end by a packaged multi-arch build plus an on-device rocBLAS run on the affected part (MI210). It alters which files land in per-arch shards only for arches whose Tensile kernels carry xnack suffixes (the datacenter parts), which is exactly what the e2e validation above exercises.

## Submission Checklist

### DIFF
--- a/shared/kpack/python/rocm_kpack/artifact_splitter.py
+++ b/shared/kpack/python/rocm_kpack/artifact_splitter.py
@@ -35,6 +35,19 @@ def strip_target_features(target: str) -> str:
     return target[:colon_pos] if colon_pos >= 0 else target
 
 
+def base_arch(arch: str) -> str:
+    """Reduce an architecture id to its bare base, dropping feature suffixes.
+
+    Handles both the ELF bundle/colon form ('gfx942:xnack+') and the
+    Tensile-style hyphen form found in kernel-database filenames
+    ('gfx90a-xnack-'). The bare base arch is the unit of artifact sharding
+    and of --gpu-targets matching, so xnack+/- variants of an in-scope base
+    arch must collapse onto it rather than be treated as distinct targets.
+    """
+    # Normalize the hyphen feature form to the colon form, then strip features.
+    return strip_target_features(arch.replace("-xnack", ":xnack"))
+
+
 @dataclass
 class ExtractedKernel:
     """Represents a kernel extracted from a fat binary."""
@@ -103,6 +116,15 @@ class FileClassificationVisitor:
         for handler in self.database_handlers:
             arch = handler.detect(file_path, prefix_path)
             if arch:
+                # Collapse feature variants onto the bare base arch. detect()
+                # returns the Tensile hyphen form for kernel objects (e.g.
+                # 'gfx90a-xnack-'), while gpu_targets and the per-arch shard key
+                # are bare ('gfx90a'). Without this, an xnack-suffixed kernel
+                # whose base arch IS in gpu_targets fails the membership check
+                # below and is dropped from both the generic and per-arch
+                # artifacts, even though no separate 'gfx90a-xnack-' build job
+                # exists to emit it (see ROCM-25535).
+                arch = base_arch(arch)
                 if self.gpu_targets is not None and arch not in self.gpu_targets:
                     if self.verbose:
                         print(
@@ -235,7 +257,7 @@ class ArtifactSplitter:
         self.database_handlers = database_handlers or []
         self.verbose = verbose
         self.gpu_targets: Optional[Set[str]] = (
-            {strip_target_features(t) for t in gpu_targets} if gpu_targets else None
+            {base_arch(t) for t in gpu_targets} if gpu_targets else None
         )
 
     def compute_kpack_search_pattern(self, binary_path: Path, prefix_root: Path) -> str:
--- a/shared/kpack/python/rocm_kpack/artifact_splitter.py
+++ b/shared/kpack/python/rocm_kpack/artifact_splitter.py
@@ -9,6 +9,7 @@ This module provides functionality to split TheRock build artifacts into:
 """
 
 import os
+import re
 import shutil
 from pathlib import Path
 from typing import List, Tuple, Dict, Optional, Set
@@ -35,9 +36,16 @@ def strip_target_features(target: str) -> str:
     return target[:colon_pos] if colon_pos >= 0 else target
 
 
+# Known LLVM AMDGPU target features that ride on a base arch, in either the
+# colon form (gfx942:xnack+) or the Tensile kernel-filename hyphen form
+# (gfx90a-xnack-). Normalizing the hyphen form to colon lets one colon-split
+# drop every feature variant regardless of how many features are present.
+_HYPHEN_FEATURE_RE = re.compile(r"-(xnack|sramecc)")
+
+
 def base_arch(arch: str) -> str:
-    """Strip features in colon or Tensile hyphen form (e.g. 'gfx90a-xnack-' -> 'gfx90a')."""
-    return strip_target_features(arch.replace("-xnack", ":xnack"))
+    """Strip known feature suffixes in colon or hyphen form (e.g. 'gfx90a-xnack-' -> 'gfx90a')."""
+    return strip_target_features(_HYPHEN_FEATURE_RE.sub(r":\1", arch))
 
 
 @dataclass
--- a/shared/kpack/python/rocm_kpack/artifact_splitter.py
+++ b/shared/kpack/python/rocm_kpack/artifact_splitter.py
@@ -36,15 +36,7 @@ def strip_target_features(target: str) -> str:
 
 
 def base_arch(arch: str) -> str:
-    """Reduce an architecture id to its bare base, dropping feature suffixes.
-
-    Handles both the ELF bundle/colon form ('gfx942:xnack+') and the
-    Tensile-style hyphen form found in kernel-database filenames
-    ('gfx90a-xnack-'). The bare base arch is the unit of artifact sharding
-    and of --gpu-targets matching, so xnack+/- variants of an in-scope base
-    arch must collapse onto it rather than be treated as distinct targets.
-    """
-    # Normalize the hyphen feature form to the colon form, then strip features.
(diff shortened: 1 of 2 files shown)


tony-davis:
Validating this through TheRock's multi-arch packaging path before merge. A throwaway TheRock branch pins rocm-systems to this fix (cherry-picked onto TheRock's current base so the only delta is this change) and dispatches Multi-Arch CI building `gfx90a` + `gfx94x` (gfx942) with the on-device rocBLAS tests — reproducing the conditions from the originating report (ROCM-25535, nightly #1440, MI210/MI308).

Test run: https://github.com/ROCm/TheRock/actions/runs/27993529661

Plan: inspect the multi-arch artifacts first to confirm the gfx90a/gfx942 Tensile xnack kernel objects are present (the files this fix stops dropping), then confirm the on-device rocBLAS run is green.

tony-davis:
Validated on hardware via TheRock multi-arch CI. Fix build = TheRock run 27993529661 (this commit pinned); baseline = current `main` multi-arch run 27975666776; both fetched identically with TheRock's installer. Environment confirmed `gfx90a` via `rocminfo`, `librocblas.so.5` resolving into the fetched trees (no system hijack).

Static (gfx90a): baseline ships `librocblas.so` but `lib/rocblas/library` is entirely missing; the fix restores it in full.

```
build                     total gfx90a library files    gfx90a-xnack kernels
baseline (27975666776)    0                             0
fix      (27993529661)    487                           174
```

On-device (gfx90a):

```
baseline: 2 passed, 10 failed   (only non-Tensile samples pass)
fix:      12 passed, 0 failed   (fortran-* skipped: unrelated libgfortran.so.5)
```

Baseline sgemm symptom vs. fix:

```
# baseline
rocBLAS error: Cannot read .../rocblas/library/TensileLibrary.dat: No such file or directory for GPU arch : gfx90a
rocBLAS error: Could not initialize Tensile host:
filesystem error: directory iterator cannot open directory: No such file or directory [.../rocblas/library]

# fix
sgemm example
NT: m, n, k, lda, ldb, ldc = 1023, 1024, 1025, 1023, 1024, 1023
PASS: max_relative_error = 1.17549e-38
```

gfx942 is unaffected (plain `gfx942` kernel names, no hyphen-xnack token to drop; both builds identical on a gfx942 box), confirming the fix's scope.

shared/kpack/python/rocm_kpack/artifact_splitter.py
```diff
@@ -35,6 +35,19 @@ def strip_target_features(target: str) -> str:
...
+    """Reduce an architecture id to its bare base, dropping feature suffixes.
+
+    Handles both the ELF bundle/colon form ('gfx942:xnack+') and the
+    Tensile-style hyphen form found in kernel-database filenames
+    ('gfx90a-xnack-'). The bare base arch is the unit of artifact sharding
+    and of --gpu-targets matching, so xnack+/- variants of an in-scope base
+    arch must collapse onto it rather than be treated as distinct targets.
+    """
+    # Normalize the hyphen feature form to the colon form, then strip features.
+    return strip_target_features(arch.replace("-xnack", ":xnack"))
```
nunnikri: Its assuming -xnack is the only possible gpu feature. It will fail if gfx90a-sramecc or gfx90a-sramecc:xnack+ comes. 
Not sure whether that will come in actual use case. If its a real case its good to handle using a regular expression

nunnikri (approved):
Overall looks good. Please take care of the comment in any case its a valid case.

shared/kpack/python/rocm_kpack/artifact_splitter.py
TorreZuk: @nunnikri  the arch string go into a helper function to remove a bunch of llvm known suffixes then?

TorreZuk (approved):
The pre-exising design is very weak as it sets a required convention for filenames to allow correct splitting?     If this really only needs to support Tensile and Tensilelite (another very bad design assumption) I think your changes look acceptable.

shared/kpack/python/rocm_kpack/artifact_splitter.py
tony-davis: Thanks @nunnikri and @TorreZuk. Pushed `934daa0fed` to address this. `base_arch()` no longer assumes `-xnack` is the only feature: it now collapses any known LLVM AMDGPU target feature (currently `xnack` and `sramecc`) onto the bare base arch, handling both the Tensile hyphen form (`gfx90a-sramecc-`) and the ELF colon form (`gfx942:sramecc+:xnack-`) via a small regex over the known-feature list — which is the "route the arch string through a helper that strips known llvm suffixes" approach @TorreZuk suggested; adding a future feature is a one-token change to that list. I also added `base_arch()` unit tests covering `xnack`/`sramecc` in both forms, the combined multi-feature case, and bare arches. There's no functional change for the target names actually shipped today (only `xnack+/-` appears in the rocBLAS/Tensile filenames), so this hardens the helper for the variants you flagged rather than changing behavior. The separate request to add a `cov5` bundled ELF asset plus a reproduction test I'll defer to a tracked follow-up, since it overlaps the integration-coverage gap already noted on ROCM-25535.

PR #7625, opened 2026-06-22
[roctjistu] Fix RDNA3-4 dispatch metadata handling

kuhar:
## Motivation

IREE-emitted RDNA kernels rely on packed TID values in v0, and wave64 RDNA descriptors encode VGPR counts differently from the runtime’s previous fixed granularity assumption. Without these fixes, dispatch setup can initialize the wrong workitem IDs or under/overcompute VGPR allocation, causing otherwise valid RDNA3/RDNA4 kernels to fail in simulation.

## Technical Details

- Enable packed workitem IDs for RDNA3/RDNA3.5/RDNA4 configs.
- Use 64-lane physical VGPR storage while keeping wave size as per-dispatch runtime state. This decouples the register-file layout from the ISA default wave size, allowing RDNA wave32 and wave64 kernels to share one CU/register-file representation. Wave32 dispatches use only the low active lanes via wf_size/EXEC masks, while wave64 dispatches have storage for lanes 32-63. This also gives checkpoint/restore and SIMD helpers a stable maximum-width VGPR layout without changing architectural VGPR counts.
- Decode AMDHSA VGPR descriptor counts using the wave-size-specific GFX11+ granularity.
- Add coverage for packed workitem IDs when only one TID VGPR is allocated.

The interposer now keeps the simulator’s view of host VMAs in sync with process mapping changes. GPU page-table entries can point at USERPTR or imported dma-buf host mappings, but those mappings may later become `PROT_NONE`, read-only, or unmapped. After successful `mmap`, `mprotect`, and `munmap` calls, the interposer invalidates the cached host-mapping snapshot so GPU memory translation can refresh permissions and fall back to sparse simulator memory instead of dereferencing inaccessible host pages.

In the daemon mode, synthetic DRM fds do not have local mmap state, so DRM mmap requests are forwarded through the remote KFD driver. This gives imported dma-buf mappings real host backing while preserving the same host-VMA permission tracking path as local simulation.

## JIRA ID

N/A

## Test Plan

Unit tests + local smoke tests.

## Test Result

All good.

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #7646, opened 2026-06-23
[rocjtisu] Fix isa semantics for rdna3 and rdna4

kuhar:
## Motivation

This extends RocJITsu’s generated AMDGPU ISA support so RDNA3/RDNA3.5/RDNA4 simulation can execute the instruction patterns needed by IREE e2e tests and the fpsan numeric suite.

The bulk of the diff is regenerated ISA output. The source changes are concentrated in the Python generator/profile logic, shared execution helpers, and regression tests.

This needs https://github.com/ROCm/rocm-systems/pull/7625 to work e2e.

## Technical Details

- Add generator-backed execution coverage for RDNA packed/vector paths:
  - packed 16-bit source/destination handling
  - true16 VOP3 `op_sel` handling
  - DPP-aware VOP3 paths
  - VOP3/VOP3P dot, compare, convert, and mixed-width operations

- Add RDNA WMMA/SWMMAC dispatch support:
  - RDNA3/RDNA3.5 GFX11 WMMA layout
  - RDNA4 GFX12 WMMA/SWMMAC layout
  - FP8/BF8/F16/BF16/IU matrix paths needed by smoke kernels

- Fix RDNA4 compatibility behavior:
  - inject opcode-9 `s_waitcnt` compatibility form where the XML only exposes split wait instructions
  - keep execution and disassembly formatting on the same GFX11 waitcnt bit layout
  - preserve generated decode/format tests for nonzero waitcnt immediates

- Refresh generated outputs from the same generator state:
  - RDNA3/RDNA3.5/RDNA4 generated ISA files
  - gfx1250 generated VOP3 split files
  - DBT legalization tables

- Expand regression coverage:
  - generator/profile tests for RDNA feature gates and matrix dispatch wiring
  - decode and execution harness tests for RDNA/RDNA4 paths
  - SIMD correctness coverage for wide MAD/shift behavior
  - value-level tests for dot2, SCC overflow, true16 modifiers, and waitcnt formatting

## JIRA ID

N/A

## Test Plan

Unit tests, IREE e2e tests, fpsan tests across rdan3, rdna4, and gfx1250.

Rerun the generator across all ISAs.

## Test Result

All good with the runtime patch.

## Submission Checklist


emulation/rocjitsu/lib/python/amdisa/codegen/execute/__init__.py
```diff
@@ -309,6 +310,7 @@ def _register_handlers() -> None:
     DISPATCH['dot2'] = lambda c: gen_dot2(
         c.dst_ops, c.src_ops, c.cls, opsel_exprs=c.opsel_exprs
     )
+    DISPATCH['dot2_true16'] = lambda c: gen_dot2_true16(c.dst_ops, c.src_ops, c.cls)
```
atgutier: Agent review:


  1. Dead DISPATCH['dot2_true16'] entry — latent ValueError (MINOR)
  execute/__init__.py:313. No instruction ever has semantic class dot2_true16 (semantics.py emits
  dot2_f16_f16/dot2_bf16_bf16, handled by an inline branch at _generator.py:3215). If this path were
  ever hit, gen_dot2_true16 would re-dispatch its own class string and raise ValueError
  (packed.py:847).
  → Either delete the registration, or register the two real classes and drop the inline special-case
  for consistency.

atgutier (approved):
This LGTM, but should probably fix the dot2_true16 dispatch entry problem.


Here are some other agent comments:

 2. Systemic copy-paste of per-arch operand logic (MINOR, the main maintainability risk)
  packed_16bit_vgpr_source/dst, the Packed16VgprSource struct, resolve_src_scalar16, the half-merge
  read_lane/write_lane, and the VCCZ/EXECZ active-mask fix are byte-for-byte identical across
  rdna3/rdna3_5/rdna4/gfx1250 (and resolve_src_scalar16 across cdna1–4 too — 7+ files). Confirmed via
  diff. This violates the project's stated style (factor reuse into templates+using), and a future fix
   to the true16 half-mapping would have to land in N places — real drift risk.
  → Hoist into a shared CRTP/base in shared/ (e.g. extend AmdgpuIsaOperand), parameterized by the
  per-ISA inline-constant table.

  ---
  Lower priority

  - packed.py:561-571 (+628, +700) (MINOR) — read_mix_src f16 path only treats selectors 240–248 as
  float inline constants; integer inline constants fed to an f16 mix get reinterpreted as raw f16
  bits. Nonsensical-but-legal encoding, low impact. Assert or document.
  - Wait-counter update inconsistency (MINOR) — _generator.py:2682-2691 emits lgkmcnt via a fragile
  read-modify-write of the combined set_wait_target(...) while the other three counters use
  single-field setters. Verified functionally correct, but a dedicated set_wait_target_lgkmcnt() would
   be cleaner. Mirror in rdna4/sopk.cpp etc. where expcnt uses the string-keyed
  set_wait_counter("wait_expcnt", …) vs typed siblings.
  - addr_calc_buffer.h:129 (MINOR, pre-existing) — static uint64_t pl_count in the MUBUF log block is
  non-thread_local; benign race across SIMD threads. Not a regression (this PR only added constexpr
  helpers nearby). Make thread_local while in the area, like mma_exec.h:895.
  - read_lane/write_lane half-select asymmetry (NIT) — write_lane calls packed_16bit_vgpr_dst
  unconditionally while read_lane gates on packed_16bit_source_. Currently unreachable (generated
  true16 ops always pass the flag or merge manually), but a foot-gun. Gate symmetrically or assert.
  - rdna4/operand.cpp:561-566 OPR_WAITCNT decode (NIT) — the combined S_WAITCNT is ignored on gfx12
  (separated into s_wait_loadcnt/dscnt/expcnt). The new vmcnt/expcnt/lgkmcnt bit-decode is disassembly
   cosmetics only, not a semantic fix. Optionally render as s_wait_idle/"(ignored)".

PR #7649, opened 2026-06-23
[rocjitsu] Native DRM ioctls; harden daemon-mode KFD lifecycle

atgutier:
## Motivation

The rocjitsu interposer layer intercepts at the system call level (to be precise, the C lib's syscall wrappers) and it doesn't intercept any user-mode driver API explicitly. As a temporary workaround we intercepted some libdrm amdgpu calls for expediency. This PR removes those cases in favor of native support for amdgpu DRM ioctls, etc.

## Technical Details

Replace the libdrm library-symbol shims with native handling of the amdgpu/DRM ioctl ABI so real libdrm_amdgpu runs unmodified over the simulator, and harden the existing daemon mode for multi-process clients with cross-process memory access, client-PID tracking, and process reuse.

Serve the amdgpu/DRM driver natively (interposer.cpp):
  - Remove all libdrm symbol interceptions (amdgpu_device_initialize/2, amdgpu_device_deinitialize/get_fd, amdgpu_query_gpu_info, amdgpu_get_marketing_name, amdgpu_bo_import/export/free/cpu_map/ cpu_unmap/va_op/query_info/set_metadata, drmCommandWriteRead, drmGetDevice(s)/drmFreeDevice(s)) and the dlsym override; real libdrm_amdgpu now runs and its calls are caught at the syscall layer
  - Handle DRM_IOCTL_AMDGPU_INFO for the queries libdrm issues during amdgpu_device_initialize: ACCEL_WORKING, DEV_INFO (with VA aperture), VRAM_GTT, MEMORY, and READ_MMR_REG; VERSION and PRIME_FD_TO_HANDLE remain handled
  - Interpose fcntl64 (libdrm imports fcntl64@GLIBC_2.28, not fcntl) so F_DUPFD_CLOEXEC on the render fd propagates DRM tracking; ioctls on the duped fd are otherwise unrecognized
  - Redirect individual /dev/dri/renderD*,card* node paths (not just the /dev/dri directory) into the synthetic dev_dri tree so libdrm's drmGetMinorType access() probe classifies every GPU's render node on multi-GPU configs
  - Drop the RjFakeBo/dmabuf bridge, Rj* libdrm mirror types, and the dlsym passthrough machinery left dead by the removals

Vendor kernel DRM UAPI headers (MIT):
  - Add external_headers/drm_headers/linux/uapi/drm/{drm.h,amdgpu_drm.h} (libdrm 2.4.125) and a DRM_INCLUDE_DIR include path, so the ioctl handlers use the real kernel structs (drm_amdgpu_info, drm_amdgpu_info_device, drm_amdgpu_info_vram_gtt, drm_amdgpu_memory_info, drm_version) rather than hand-mirrored copies

Harden daemon-mode KFD lifecycle, memory bridging, and process reuse (daemon mode and its RemoteDriver/RPC transport already exist; this adds the multi-process client support on top):
  - Interposer: remote_mutex_ guards RemoteDriver create/destroy, remote_kfd_fd_ made atomic, close_remote()/close_remote_fd_only() for clean shutdown, AMDKFD_IOCTL_BASE fallback for late ioctls, reset_after_fork() clears all containers and reinits remote_mutex_
  - SimulatedDriver: client_pid parameter on open_process() with process reuse, open_process() TOCTOU fix (pid search + create under one lock), unmap_memory_ioctl unmaps GPU pages without erasing the allocation record, set_process_client_pid() for post-hoc pid attachment
  - KfdProcess: client_pid_ plus retain_open()/release_open() refcount
  - gpu_memory.h: process_vm_readv/writev fallback for cross-process access, per-VMID client_pid tracking, warnings on access failure
  - remote_driver.cpp: extract dmabuf fd from EXPORT_DMABUF RPC response
  - rj_vm.h/cpp: rj_vm_device_open_for_client_pid() and rj_vm_device_set_client_pid() APIs
  - main.cpp: extract client PID via SO_PEERCRED, pass to rj_vm_device_open_for_client_pid()

## JIRA ID

N/A

## Test Plan

Run all CI tests, run OPT125, run hrx-systems tests.

## Test Result

CI tests: PASSED
OPT125: PASSED with 100% accurate tokens/logits compared to CPU reference run.
hrs-systems tests: 61/107 PASSED (known failures, no regressions)

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/CHANGELOG.md
+++ b/emulation/rocjitsu/CHANGELOG.md
@@ -11,9 +11,10 @@ Versioning follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   unmodified over the simulator; DRM device discovery and queries are handled at
   the syscall layer instead of by intercepting libdrm library symbols
 - **BREAKING:** `rj_vm_device_open()` now takes a connecting client's PID:
-  `rj_vm_device_open(vm, pid_t client_pid, uint32_t *process_id)`. Pass 0 in
-  local mode; a nonzero PID enables daemon-mode process reuse and cross-process
-  memory access
+  `rj_vm_device_open(vm, rj_client_pid_t client_pid, uint32_t *process_id)`. The
+  new `rj_client_pid_t` is a fixed-width (`int32_t`) typedef so the public header
+  stays OS- and ABI-width-agnostic. Pass 0 in local mode; a nonzero PID enables
+  daemon-mode process reuse and cross-process memory access
 
 ### Added
 - Daemon-mode multi-process client support: per-client process reuse,
--- a/emulation/rocjitsu/CHANGELOG.md
+++ b/emulation/rocjitsu/CHANGELOG.md
@@ -4,6 +4,23 @@ All notable changes to rocjitsu are documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 Versioning follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.2.0] - 2026-06-23
+
+### Changed
+- Serve the amdgpu/DRM ioctl ABI natively so upstream libdrm_amdgpu runs
+  unmodified over the simulator; DRM device discovery and queries are handled at
+  the syscall layer instead of by intercepting libdrm library symbols
+- **BREAKING:** `rj_vm_device_open()` now takes a connecting client's PID:
+  `rj_vm_device_open(vm, pid_t client_pid, uint32_t *process_id)`. Pass 0 in
+  local mode; a nonzero PID enables daemon-mode process reuse and cross-process
+  memory access
+
+### Added
+- Daemon-mode multi-process client support: per-client process reuse,
+  client-PID identification via SO_PEERCRED, and cross-process GPU memory access
+- Vendored kernel DRM UAPI headers (drm.h, amdgpu_drm.h, drm_mode.h) for native
+  amdgpu ioctl struct fidelity
+
 ## [0.1.0] - 2026-06-04
 
 ### Added
--- a/emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/vm/rj_vm.h
+++ b/emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/vm/rj_vm.h
@@ -12,6 +12,7 @@
 
 #include <stddef.h>
 #include <stdint.h>
+#include <sys/types.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -223,10 +224,14 @@ RJ_API_EXPORT rj_status_t rj_vm_execute(rj_vm_t *vm, rj_vm_cmd_t *cmd);
 /// @param[in,out] cmd Command descriptor.
 RJ_API_EXPORT rj_status_t rj_vm_execute_as(rj_vm_t *vm, uint32_t process_id, rj_vm_cmd_t *cmd);
 
-/// @brief Open the VM's simulated device and create a new KFD process.
+/// @brief Open the VM's simulated device and create (or reuse) a KFD process.
 /// @param[in] vm VM handle.
-/// @param[out] process_id The new process ID (may be NULL for legacy callers).
-RJ_API_EXPORT rj_status_t rj_vm_device_open(rj_vm_t *vm, uint32_t *process_id);
+/// @param[in] client_pid Connecting client's OS PID for daemon mode (enables
+///   cross-process memory access and process reuse); pass 0 in local mode where
+///   there is no separate client process. If a process already exists for a
+///   nonzero client_pid, it is reused.
+/// @param[out] process_id The simulator's process handle (may be NULL).
+RJ_API_EXPORT rj_status_t rj_vm_device_open(rj_vm_t *vm, pid_t client_pid, uint32_t *process_id);
 
 /// @brief Close a specific KFD process by ID.
 /// @param[in] vm VM handle.
--- a/emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/vm/rj_vm.h
+++ b/emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/vm/rj_vm.h
@@ -12,7 +12,6 @@
 
 #include <stddef.h>
 #include <stdint.h>
-#include <sys/types.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -30,6 +29,11 @@ typedef struct rj_vm_t rj_vm_t;
 
 /// @brief Platform-specific handle type (fd on Linux, HANDLE on Windows).
 typedef int rj_handle_t;
+
+/// @brief Connecting client's process ID, in fixed-width form.
+/// @details Kept ABI-stable and OS agnostic so we don't depend on the
+/// platform width of a process ID.
+typedef int32_t rj_client_pid_t;
 
 /// @brief VM creation mode.
 typedef enum rj_vm_mode_t {
@@ -231,7 +235,8 @@ RJ_API_EXPORT rj_status_t rj_vm_execute_as(rj_vm_t *vm, uint32_t process_id, rj_
 ///   there is no separate client process. If a process already exists for a
 ///   nonzero client_pid, it is reused.
 /// @param[out] process_id The simulator's process handle (may be NULL).
-RJ_API_EXPORT rj_status_t rj_vm_device_open(rj_vm_t *vm, pid_t client_pid, uint32_t *process_id);
+RJ_API_EXPORT rj_status_t rj_vm_device_open(rj_vm_t *vm, rj_client_pid_t client_pid,
+                                            uint32_t *process_id);
 
 /// @brief Close a specific KFD process by ID.
 /// @param[in] vm VM handle.
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
@@ -20,6 +20,18 @@
 #include "rocjitsu/vm/rj_vm.h"
 #include "rocjitsu/vm/rj_vm_impl.h"
 
+RJ_DIAGNOSTIC_PUSH
+RJ_DIAGNOSTIC_IGNORE_PEDANTIC
+#include "linux/uapi/kfd_ioctl.h"
+// Vendored kernel DRM/amdgpu UAPI (MIT). Provides the real drm_version,
+// drm_amdgpu_info, drm_amdgpu_info_device, drm_amdgpu_info_vram_gtt, and
+// drm_amdgpu_memory_info structs so the interposer services the amdgpu DRM
+// ioctl ABI directly. These are kernel ABI, not libdrm library types, so this
+// keeps the interposer independent of libdrm.
+#include "amdgpu_drm.h"
+#include "drm.h"
+RJ_DIAGNOSTIC_POP
+
 #include "util/dynamic_loader.h"
 #include "util/log.h"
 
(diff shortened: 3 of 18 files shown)


atgutier:
@copilot do another review

atgutier:
@copilot review again

kuhar (commented):
It would be nice to add a small amount of targeted coverage for the new daemon/libdrm surface. The important, practical tests would be:

1. A libdrm smoke test that runs `amdgpu_device_initialize` / `amdgpu_query_gpu_info` through the interposer. This is the headline behavior and the init cascade is field/order-sensitive. The existing `daemon_test.cpp` helper pattern looks suitable; if linking a libdrm helper is too heavy, factor the AMDGPU_INFO DEV_INFO filling into a helper and unit-test that path.
2. Retarget `drm_info_layout_test.cpp` to the vendored `drm_amdgpu_info_device` struct and assert the fields populated by DEV_INFO, including the virtual-address and high-VA fields. The current `LocalDevInfo` mirror no longer corresponds to production code.
3. Add one focused daemon/process lifecycle test for same-client `open_process(client_pid)` reuse and last-close teardown. This is easy in `KfdIoctlTest` and also guards the open/close refcount behavior added here.

Cross-process `process_vm_readv/writev` coverage would be useful, but it is a second-tier/Linux-environment-sensitive test. I would not block this PR on that alone if the smoke and lifecycle coverage above are added.

Agent comments below.

The new /dev/dri redirect should also be applied in the 64-bit and versioned stat wrappers. stat(), lstat(), access(), and opendir() now call redirect_dev_dri(), but stat64(), lstat64(), __xstat*(), and __lxstat*() still only check the sysfs/dev-char redirects. A libdrm probe using those entry points can still hit the host /dev/dri/renderD* and miss the synthetic render node.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
```diff
@@ -811,8 +709,8 @@ int openat64(int dirfd, const char *path, int flags, ...) {
 
 int close(int fd) {
   assert(InterposerContext::real.ready());
-  if (auto *remote = InterposerContext::ctx.remote_lookup(fd))
-    return remote->close();
+  if (InterposerContext::ctx.remote_lookup(fd))
+    return InterposerContext::ctx.close_remote_fd_only();
```
kuhar: We need the last daemon KFD close to tear down the remote driver, not just the synthetic fd. The close() path for remote_lookup(fd) calls close_remote_fd_only(), while close_remote()/try_close_remote() have no call sites and the destructor is empty. That means RemoteDriver::close()/RPC_CLOSE are not sent on /dev/kfd close, leaving daemon-side process state alive until process exit. Please ref-count the primary and dup-tracked KFD fds and call RemoteDriver::close() on the last close.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_driver.cpp
```diff
@@ -364,12 +376,25 @@ uint32_t SimulatedDriver::open_process() {
   uint32_t pid;
   {
     std::lock_guard<std::mutex> lk(process_mutex_);
+    if (client_pid > 0) {
```
kuhar: We should keep the client-pid reuse and close-refcount conditions symmetric. open_process() reuses and calls retain_open() for any client_pid > 0, but close() only calls release_open() when daemon_mode_ is true. If the public API is called with a nonzero client PID outside daemon mode, the first close can tear down a still-retained process. Gate reuse on daemon_mode_ too, or remove the daemon-mode guard from close; a focused same-client open/open/close/close test would lock this down.

emulation/rocjitsu/CHANGELOG.md
```diff
@@ -4,6 +4,23 @@ All notable changes to rocjitsu are documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 Versioning follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.2.0] - 2026-06-23
```
kuhar, line 7: We have a bunch of recent changes that should probably also be added here... Perhaps worth sending a reminder on Teams before we bump to 0.2.0?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
```diff
@@ -126,6 +134,9 @@ class GpuMemory : public simdojo::SparseMemory {
   uint8_t read8(uint64_t addr, uint32_t vmid = 0) const {
     if (auto *p = translate(addr, vmid))
       return p[addr & PAGE_MASK];
+    uint8_t val = 0;
+    if (vmid > 0 && read_client_memory(addr, &val, 1, vmid))
```
kuhar, line 138: nit / optional: in c++ >=17 you could define `uint8_t val = 0;` inside the if condition itself. Here and elsewhere below.

emulation/rocjitsu/CHANGELOG.md
atgutier: I have only been adding changes to this log that directly affect the versioning. Are you suggesting to log all changes here?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
atgutier: I had another commit that did this, so I just added it to this PR.

LMK if you're comments have been addressed.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
atgutier: Will cleanup stuff like this in a targeted refactor/cleanup. I have a list going...

emulation/rocjitsu/CHANGELOG.md
kuhar: I see, I thought this was for major user-facing changes and, for example, supported targets would go in here too

atgutier: Yes, that is the right idea. We can start that from now on. Technically, those should also bump the minor version. I think once we get to 1.0.0 we will be far more strict on this.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
```diff
@@ -913,6 +927,15 @@ int ioctl(int fd, unsigned long request, ...) {
     if (auto *remote = InterposerContext::ctx.remote_lookup(InterposerContext::ctx.remote_kfd_fd()))
       return remote->ioctl(request, arg);
   }
+  // Late-ioctl safety net: an AMDKFD ('K') ioctl may arrive on a tracked KFD fd
+  // whose primary remote handle changed underneath it (e.g. a close/dup race in
+  // daemon mode). Forward only AMDKFD-typed ioctls, and only on fds we already
+  // track as KFD (primary or dup), so an arbitrary unrelated fd carrying a
+  // type-'K' ioctl is never misrouted to the remote KFD driver.
+  if (_IOC_TYPE(request) == AMDKFD_IOCTL_BASE && InterposerContext::ctx.is_kfd_tracked(fd)) {
+    if (auto *remote = InterposerContext::ctx.remote())
```
kuhar, line 962: We should make `remote_` access consistently synchronized. `teardown_remote()` deletes and clears `remote_` under `remote_mutex_`, but `remote()`, `remote_lookup()`, the AMDKFD fallback, and the mmap path read the raw pointer without that mutex. The open refcount lowers the practical use-after-free risk, but it does not make the pointer read/write data-race-free. Either load/store `remote_` through an atomic pointer or take `remote_mutex_` in the read helpers.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
```diff
@@ -214,6 +243,42 @@ class GpuMemory : public simdojo::SparseMemory {
...
+    return (it != vmid_table_.end()) ? it->second.client_pid : 0;
+  }
+
+  bool read_client_memory(uint64_t addr, void *dst, size_t len, uint32_t vmid) const {
+    pid_t pid = client_pid_for_vmid(vmid);
+    if (pid <= 0)
+      return false;
+    iovec local{dst, len};
+    iovec remote{reinterpret_cast<void *>(addr), len};
+    ssize_t rc = process_vm_readv(pid, &local, 1, &remote, 1, 0);
```
kuhar, line 258: We should make the daemon cross-process memory contract explicit. `GpuMemory` now uses `process_vm_readv/writev` for `vmid > 0` misses, but on systems with Yama `ptrace_scope=1` a daemon child or unrelated attach process cannot read the client without ptrace permission; this runner has that setting. Today failure only logs and falls back to `SparseMemory`, so a real client-memory read can become zero or stale data. Please arrange the ptrace permission relationship, or use a shared-memory path, and treat `EPERM`/short transfer as a daemon-mode error instead of a cache miss.

emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/vm/rj_vm.h
```diff
@@ -223,10 +224,14 @@ RJ_API_EXPORT rj_status_t rj_vm_execute(rj_vm_t *vm, rj_vm_cmd_t *cmd);
...
+/// @brief Open the VM's simulated device and create (or reuse) a KFD process.
 /// @param[in] vm VM handle.
-/// @param[out] process_id The new process ID (may be NULL for legacy callers).
-RJ_API_EXPORT rj_status_t rj_vm_device_open(rj_vm_t *vm, uint32_t *process_id);
+/// @param[in] client_pid Connecting client's OS PID for daemon mode (enables
+///   cross-process memory access and process reuse); pass 0 in local mode where
+///   there is no separate client process. If a process already exists for a
+///   nonzero client_pid, it is reused.
+/// @param[out] process_id The simulator's process handle (may be NULL).
+RJ_API_EXPORT rj_status_t rj_vm_device_open(rj_vm_t *vm, pid_t client_pid, uint32_t *process_id);
```
kuhar: We should keep the public C API type-stable here. `rj_vm.h` otherwise exposes fixed-width types or project handles, but this change adds `<sys/types.h>` and `pid_t` to `rj_vm_device_open`, making the common VM header POSIX-specific and ABI-width-dependent. Please use a project typedef or fixed-width client PID type in the exported API, and translate to `pid_t` inside the Linux daemon path.

kuhar (commented):
One focused daemon lifecycle regression test would be worthwhile here, but I would not ask for a full libdrm/ioctl test stack in this PR.

The cheap path is in `KfdIoctlTest`: it already constructs a daemon-mode `SimulatedDriver` directly, and the file already has local KFD open-refcount coverage. A targeted test could call `open_process(client_pid)` twice with the same nonzero PID, assert the same process ID is reused, close it once, verify a process-scoped ioctl still succeeds, then close it again and verify the process is gone. That covers the client-PID reuse/refcount behavior added here without launching the daemon, linking libdrm, or refactoring `interposer.cpp`.

A real libdrm smoke test for `amdgpu_device_initialize` / `AMDGPU_INFO_DEV_INFO` would be useful follow-up coverage if we add optional libdrm test plumbing, but I would not make that a blocker for this PR.

Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp:965`

`dup2`/`dup3` should not mutate KFD tracking before the syscall succeeds. The current wrappers call `untrack_dup(newfd)` first, so `dup2(fd, fd)` drops tracking before the POSIX no-op and `dup3(fd, fd, ...)` drops it before the expected `EINVAL`. That can decrement the open ref while the descriptor remains live. Capture the old tracking state, skip the `dup2(fd, fd)` no-op case, and only replace tracking after a successful duplication/replacement.

Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp:803`

`DRM_IOCTL_VERSION` can copy one byte past a caller buffer. The guard accepts `name_len >= sizeof("amdgpu") - 1`, but the copy uses `sizeof(drv_name)`, including the NUL byte. A caller that allocates the exact length returned by the query path gets a 7-byte write into a 6-byte buffer. Either require space for the terminator or copy only the advertised length and NUL-terminate only when there is room.

kuhar (approved):
Overall LGTM, none of the agent issues posted above seem blocking

emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/vm/rj_vm.h
atgutier: Yeah, this is actually a big deal. I will fix this.

PR #7656, opened 2026-06-23
rccl: fix GDRDMA and NIC affinity on gfx1250 topology

VadimKutovoi:
## Motivation

This PR fixes topology discovery logic on the test platform

## JIRA ID

AICOMRCCL-1223

## Test Plan

Manual validation on a test platform, check that GDRDMA works without topology adjustments.

## Test Result

Pass

## Submission Checklist

### DIFF
--- a/projects/rccl/src/graph/paths.cc
+++ b/projects/rccl/src/graph/paths.cc
@@ -730,6 +730,28 @@ static bool rcclPathOverride(struct ncclTopoSystem* system, uint64_t distance) {
   }
 }
 
+// Rewrite GPU<->NIC paths of type `fromType` to `toType` when the two devices share a PCI domain.
+static void rcclRewriteSameDomainNetPaths(struct ncclTopoSystem* system, int fromType, int toType) {
+  for (int g=0; g<system->nodes[GPU].count; g++) {
+    struct ncclTopoNode* gpu = system->nodes[GPU].nodes+g;
+    int64_t gpuBusId = NCCL_TOPO_ID_LOCAL_ID(gpu->id);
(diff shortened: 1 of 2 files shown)


projects/rccl/src/graph/paths.cc
```diff
@@ -802,6 +763,35 @@ ncclResult_t ncclTopoComputePaths(struct ncclTopoSystem* system, struct ncclComm
...
+        if (gpuBusId != 0 && net->net.busId != 0 && gpuDomain == netDomain) {
+          if (gpu->paths[NET] && gpu->paths[NET][n].type == PATH_PHB) {
+            gpu->paths[NET][n].type = PATH_PXB;
+            INFO(NCCL_GRAPH, "gfx1250: corrected GPU %d -> NET %d path from PHB to PXB (domain %ld)", g, n, gpuDomain);
+          }
+          if (net->paths[GPU] && net->paths[GPU][g].type == PATH_PHB)
+            net->paths[GPU][g].type = PATH_PXB;
+        }
+      }
+    }
```
amd-arozanov: Please move it to a separate function. ncclTopoComputePaths is already big enough.

Also, there is function rcclPathOverride(), which is doing PHB->PXB correction for GPU-NET pairs. Does it do the same? Would it be possible to reuse it?

VadimKutovoi: Thanks, that's a good point to extract this code into separate helper function. rcclPathOverride logic is slightly different - it checks the distances between NIC and GPU. This fix is looking for GPU and NIC that shares the same PCIe domain, so I'm going to keep them separate.

alexander-sannikov (approved):

amd-arozanov (approved):

PR #7659, opened 2026-06-23
[RCCL] [AICOMRCCL-1175] Add UT for symmetric memory decoupled from NCCL_P2P_LEVEL

BertanDogancay:
## Motivation

Verifies the gating path for symmetric memory with NCCL_P2P_LEVEL, updates docs on symmetric memory enablment and add test cases to UT.

## Technical Details

- Added docs for symmetric memory enablement.
- Symmetric memory support should no longer depend on NCCL_P2P_LEVEL. Added test cases to test this change in rccl source code.

## JIRA ID

AICOMRCCL-1175

## Test Plan

Running newly written test cases for this change. All test cases passed.

## Test Result

All test cases passed.

## Submission Checklist


thomas-huber (approved):
Nice work @BertanDogancay .

projects/rccl/docs/how-to/rccl-usage-tips.rst
```diff
@@ -32,6 +32,52 @@ set the HSA environment variable as follows:
...
+(``ncclCommWindowRegister`` with the ``NCCL_WIN_COLL_SYMMETRIC`` flag, or
+buffers allocated with ``ncclMemAlloc``) so that every participating rank can
+address the buffer directly.
+
+Whether a communicator can use symmetric memory is decided once at
+``ncclCommInitRank`` time. The prerequisites are:
+
+- All local ranks are **CUDA peer-to-peer capable** with each other (on AMD
+  GPUs this means they are on the same host, or are reachable through a
+  Multi-Node NVLink/Infinity Fabric clique).
```
corey-derochie-amd: Should this be mentioning "CUDA peer-to-peer capable", or "MNNVL"? Perhaps this makes sense as we use some cross-over language, but saying that "on AMD GPUs this means they... are reachable through a [MNNVL]... clique" is not accurate.

projects/rccl/test/TopoEnvPolicyTests.cpp
corey-derochie-amd, line 1: Should we also validate expected behaviour for other P2P levels?

BertanDogancay: Good point. That would help with coverage but let's keep it in a separate PR though, since it's not really in scope for this work item.

projects/rccl/docs/how-to/rccl-usage-tips.rst
BertanDogancay: Removed cuda and NVLink mentions

PR #7661, opened 2026-06-23
[HIP][Graph] Support cross-device segment path and fix captureDeviceId_

anugodavar:
- Rename instantiateDeviceId_ to captureDeviceId_ to better convey that this holds the device ID where the graph was created/captured.

- Derive captureDeviceId_ from the first graph node (via Graph::GetNodes()) rather than hip::getCurrentDevice(), which may differ from the capture device. For single-device graphs, the authoritative key from max_streams_dev_ is used after FindStreamsReq* populates it.

- In FindStreamsReqPerDevForSegments(), set captureDeviceId_ from the single max_streams_dev_ key for single-device graphs. For multi-device graphs, log an error and return early (not supported on segment path). Init() checks for captureDeviceId_ == -1 and returns hipErrorNotSupported.

- Extend the segment scheduling path in Run() to handle cross-device launches (captureDeviceId_ != launch_stream->DeviceId()):
  - Pre-sync: marker on capture device stream waits for launch_stream's last command so graph starts after prior work on the launch device.
  - Enqueue graph via streams_[0] (capture device's internal stream).
  - Post-sync: marker on launch_stream waits for graph completion so subsequent work on the launch device starts after the graph is done.

- Update UpdateStreams() to accept nullptr for cross-device launches. When nullptr, uses captureDeviceId_ to populate streams_ from parallel_streams_ without including the launch stream.

- Fix CallbackCommand and cleanup->device to use captureDeviceId_ so signals are recycled to the correct device's pool in OnLaunchComplete.

## Submission Checklist

### DIFF
--- a/projects/clr/hipamd/src/hip_graph_internal.cpp
+++ b/projects/clr/hipamd/src/hip_graph_internal.cpp
@@ -443,7 +443,7 @@ void GraphExec::BuildSyncPlan() {
   sync_plan_.seg_to_hw_event.assign(segments_.size(), -1);
   sync_plan_.num_hw_events = 0;
 
-  auto* device = g_devices[instantiateDeviceId_]->devices()[0];
+  auto* device = g_devices[captureDeviceId_]->devices()[0];
 
   // PASS 1: Assign a compact HW-event slot only to segments whose completion
   // signal is consumed — cross-device/stream successor, or leaf when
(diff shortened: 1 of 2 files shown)


saleelk (approved):

saleelk (changes requested):
accidenrtally approved

PR #7674, opened 2026-06-23
[rocprofiler-sdk] Fix gfx1151/gfx1152 APU support in HSA test apps

bgopesh:
Grant memory access to the GPU agent only in the test apps' hsa_malloc helpers instead of passing the full agent list (CPU + GPU). On APUs such as gfx1151, hsa_amd_agents_allow_access rejects the CPU agent with HSA_STATUS_ERROR_INVALID_AGENT, causing the test apps to abort. This matches the production pattern in hsa_rsrc_factory.cpp.

Also add gfx1151 and gfx1152 to DEFAULT_GPU_TARGETS so the test HSACO kernels are generated for these architectures.

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-sdk/samples/openmp_target/CMakeLists.txt
+++ b/projects/rocprofiler-sdk/samples/openmp_target/CMakeLists.txt
@@ -30,8 +30,17 @@ target_link_libraries(
     PRIVATE rocprofiler-sdk::rocprofiler-sdk rocprofiler-sdk::samples-build-flags
             rocprofiler-sdk::samples-common-library)
 
-set(DEFAULT_GPU_TARGETS "gfx906" "gfx908" "gfx90a" "gfx942" "gfx950" "gfx1100" "gfx1101"
-                        "gfx1102")
+set(DEFAULT_GPU_TARGETS
+    "gfx906"
+    "gfx908"
+    "gfx90a"
+    "gfx942"
+    "gfx950"
+    "gfx1100"
+    "gfx1101"
+    "gfx1102"
+    "gfx1151"
+    "gfx1152")
 
 set(OPENMP_GPU_TARGETS
     "${DEFAULT_GPU_TARGETS}"
--- a/projects/rocprofiler-sdk/tests/bin/hsa-code-object/hsa_code_object_app.h
+++ b/projects/rocprofiler-sdk/tests/bin/hsa-code-object/hsa_code_object_app.h
@@ -276,7 +276,11 @@ public:
         hsa_status_t err = hsa_amd_memory_pool_allocate(mem.pool, size, 0, &ret);
         RET_IF_HSA_ERR(err);
 
-        err = hsa_amd_agents_allow_access(all_devices.size(), all_devices.data(), nullptr, ret);
+        // Grant access to the GPU agent only. On APUs (for example, gfx1151) the CPU
+        // agent is rejected by hsa_amd_agents_allow_access with
+        // HSA_STATUS_ERROR_INVALID_AGENT, so passing the full agent list fails.
+        hsa_agent_t ag_list[1] = {gpu[0].agent};
+        err                    = hsa_amd_agents_allow_access(1, ag_list, nullptr, ret);
         RET_IF_HSA_ERR(err);
         return ret;
     }
--- a/projects/rocprofiler-sdk/tests/bin/hsa-code-object/hsa_code_object_app.h
+++ b/projects/rocprofiler-sdk/tests/bin/hsa-code-object/hsa_code_object_app.h
@@ -276,12 +276,29 @@ public:
         hsa_status_t err = hsa_amd_memory_pool_allocate(mem.pool, size, 0, &ret);
         RET_IF_HSA_ERR(err);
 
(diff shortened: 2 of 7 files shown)


projects/rocprofiler-sdk/tests/bin/hsa-code-object/hsa_code_object_app.h
```diff
@@ -276,7 +276,11 @@ class MQDependencyTest
         hsa_status_t err = hsa_amd_memory_pool_allocate(mem.pool, size, 0, &ret);
         RET_IF_HSA_ERR(err);
 
-        err = hsa_amd_agents_allow_access(all_devices.size(), all_devices.data(), nullptr, ret);
+        // Grant access to the GPU agent only. On APUs (for example, gfx1151) the CPU
```
ApoKalipse-V: What about services that need to read the code object data?
We expose the pointer via the code object tracing API

ApoKalipse-V (changes requested):

projects/rocprofiler-sdk/tests/bin/hsa-code-object/hsa_code_object_app.h
bgopesh: This is a test only change and the code object data isn't affected by this change. `hsa_malloc` here only allocates kernel-argument/copy buffers from the CPU kernarg pool — it's never used for code object data. The code 
  object data comes from `hsa_code_object_reader_create_from_file` (file-backed) or a plain host new uint8_t[] buffer passed to `hsa_code_object_reader_create_from_memory`; neither goes through `hsa_amd_agents_allow_acces`.

projects/rocprofiler-sdk/tests/bin/hsa-code-object/hsa_code_object_app.h
```diff
@@ -276,7 +276,11 @@ class MQDependencyTest
...
+        // agent is rejected by hsa_amd_agents_allow_access with
+        // HSA_STATUS_ERROR_INVALID_AGENT, so passing the full agent list fails.
+        hsa_agent_t ag_list[1] = {gpu[0].agent};
```
ApoKalipse-V: Is gpu[0] guaranteed to exist here? What if we have none?

ApoKalipse-V: What if we have multiple? Is it going to be added to the correct agent?

ApoKalipse-V (changes requested):

projects/rocprofiler-sdk/tests/bin/hsa-code-object/hsa_code_object_app.h
bgopesh: This change is not introducing any new assumption. we already dereference dereference gpu[0].agent at startup, before any hsa_malloc call  `https://github.com/ROCm/rocm-systems/blob/7ae0ef15125a80c432e6c4b03854fef1fa74d58a/projects/rocprofiler-sdk/tests/bin/hsa-code-object/hsa_code_object_app.cpp#L46`

bgopesh: Concern is valid. But what I mean to say is there are pre existing gaps. We can of course harden these tests

projects/rocprofiler-sdk/tests/common/CMakeLists.txt
```diff
@@ -23,7 +23,9 @@ set(DEFAULT_GPU_TARGETS
     "gfx1010"
     "gfx1100"
     "gfx1101"
-    "gfx1102")
+    "gfx1102"
+    "gfx1151"
+    "gfx1152")
```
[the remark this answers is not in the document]
bgopesh: done

projects/rocprofiler-sdk/samples/openmp_target/CMakeLists.txt
```diff
@@ -30,8 +30,17 @@ target_link_libraries(
     PRIVATE rocprofiler-sdk::rocprofiler-sdk rocprofiler-sdk::samples-build-flags
             rocprofiler-sdk::samples-common-library)
 
-set(DEFAULT_GPU_TARGETS "gfx906" "gfx908" "gfx90a" "gfx942" "gfx950" "gfx1100" "gfx1101"
-                        "gfx1102")
+set(DEFAULT_GPU_TARGETS
```
ApoKalipse-V, line 33: I know the AI talked about this list, but I'm not sure about it.
Google says gfx115x is not supported on openmp. Can you verify?

ApoKalipse-V (changes requested):

ApoKalipse-V (approved):

projects/rocprofiler-sdk/samples/openmp_target/CMakeLists.txt
bgopesh: recent toolchain is supporting it seems: AOMP 20.0-0

itrowbri (approved):

PR #7677, opened 2026-06-23
rocjitsu: support indirect branches and s_call_b64

Groverkss:
Add DBT support for relocating scalar control flow that was previously unsupported: direct `s_call_b64`, static indirect `s_setpc_b64`, and static indirect `s_swappc_b64` call/return patterns.

The indirect branch recovery is conservative and dataflow-based:

- Decode `.text` and build a temporary CFG from direct branches, fallthroughs, and descriptor entry points.
- Track SGPR pairs that hold concrete PC values built from `s_getpc_b64`.
- Summarize each block as `SET`, `KILL`, or `PASS` for each tracked SGPR pair.
- Resolve `s_setpc_b64` / `s_swappc_b64` consumers from bounded block-entry facts.
- Fail closed when a pair is clobbered by unmodeled writes or a consumer has too many possible targets.

The canonical recovered indirect branch is:

```asm
s_getpc_b64  s[8:9]
s_add_u32    s8, s8, <lo-target-delta>
s_addc_u32   s9, s9, <hi-target-delta-or-0>
s_setpc_b64  s[8:9]
```

Tensile-style static skips are also recovered:

```asm
s_getpc_b64  s[8:9]
s_add_i32    s10, <literal-delta>, 4
s_add_u32    s8, s8, s10
s_addc_u32   s9, s9, 0
s_setpc_b64  s[8:9]
```

Direct scalar calls are relocated through the normal PC-relative branch fixup path, while their return SGPR pair is recorded:

```asm
s_call_b64   s[20:21], target
...
s_setpc_b64  s[20:21]
```

Recovered `s_swappc_b64` calls are modeled similarly:

```asm
s_getpc_b64  s[8:9]
s_add_u32    s8, s8, <callee-delta>
s_addc_u32   s9, s9, 0
s_swappc_b64 s[20:21], s[8:9]
...
s_setpc_b64  s[20:21]
```

Recovered non-call `s_setpc_b64` targets become ordinary CFG successors. Calls are kept separately as `BasicBlock::CallEdge` records so shared helper blocks do not gain every possible return continuation globally. DBT uses those call edges only inside the current kernel translation scope for reachability, liveness, and validation of return-like `s_setpc_b64` instructions.

After each kernel-local body is emitted, DBT patches direct PC-relative branch/call immediates and rewrites recovered PC-builder ranges using the final relocated target offsets. Shared source blocks are duplicated per translated kernel so branches, calls, and returns resolve through the current kernel's placement map.


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/rdna4/sopp.cpp
```diff
@@ -59,7 +59,7 @@ SSleepSopp::SSleepSopp(const MachineInst *inst)
   num_dst_ = 0;
 }
 
-void SSleepSopp::execute_impl(amdgpu::Wavefront &wf) { amdgpu::execute_s_sleep_sopp(*this, wf); }
+void SSleepSopp::execute_impl(amdgpu::Wavefront &wf) { (void)wf; }
```
atgutier: Looks like you have some manual edits to auto-generated files.

Please ensure all auto-generated files are modified only by running the amdisa lib to generate and that all generated diffs are included in this PR.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/binary_translator.cpp
```diff
@@ -754,6 +857,50 @@ TranslatedCodeObject BinaryTranslator::translate(const AmdGpuCodeObject &obj) {
...
+                     "target ISA cannot encode canonical recovered indirect branch builder",
+                     fixup.source_call_offset, "indirect branch");
+        return leave_unchanged();
+      }
+
+      [[maybe_unused]] const uint64_t recovery_size =
+          fixup.target_recovery_end_offset - fixup.target_recovery_begin_offset;
+      [[maybe_unused]] const uint64_t replacement_size =
+          replacement_words.size() * sizeof(uint32_t);
+      assert(replacement_size <= recovery_size &&
```
atgutier, line 1018: This may be better off as a runtime check. Can't the necessary number of words grow after cross-ISA relocation and cave growth?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/binary_translator.cpp
```diff
@@ -568,17 +635,49 @@ TranslatedCodeObject BinaryTranslator::translate(const AmdGpuCodeObject &obj) {
...
+              layout, inst, text_word_at(text, offset), offset, guest_arch_);
+          if (carried) {
+            layout.indirect_call_fixups.push_back(*carried);
+            recovered_indirect_call_offsets.insert(carried->source_call_offset);
+          }
+        }
+        const bool has_recovered_indirect_call = recovered_indirect_call_offsets.contains(offset);
+        const uint32_t word = text_word_at(text, offset);
+        const bool recovered_indirect_return =
+            s_setpc_from_sreg(guest_arch_, inst, word, 30) ||
```
atgutier: Agent review comment:

Hard-coded return reg 30 accepts any s_setpc s[30:31] as relocation-safe —
  binary_translator.cpp:~676
  recovered_indirect_return = s_setpc_from_sreg(..., 30) || recovered_direct_call_return(...). The
  first disjunct passes through any s_setpc_b64 s[30:31] untouched, without verifying a matching
  in-scope call set it and without relocating the held PC — emitting a stale source PC, contradicting
  the design doc's "fail closed" guarantee. recovered_direct_call_return
  (static_pc_recovery.cpp:345-354) is the principled call-paired check; the magic 30 short-circuits
  it.
  Fix: drop the unconditional s[30:31] disjunct, rely on recovered_direct_call_return; if a default
  ABI return reg must be honored, gate it on having actually seen a call that wrote s[30:31] in scope.

emulation/rocjitsu/tests/dbt/translate_test.cpp
```diff
@@ -504,10 +504,11 @@ bool has_error_containing(const TranslatedCodeObject &result, DiagnosticKind kin
                      });
 }
```
atgutier, line 505: Agent review comment:

 No positive test for the headline feature — tests/dbt/translate_test.cpp
  The PR removes s_call_b64 from the rejection test (implying it's now supported) but adds no E2E test
   translating an s_call_b64 + paired s_setpc return and asserting the relocated immediate/return.
  recovered_direct_call_return, recover_carried_indirect_call_target, and the call-return recording
  are entirely untested — the highest-risk paths in a control-flow PR.
  Fix: add E2E tests for (a) s_call_b64 to a reachable target + paired return, (b) carried recovery
  (two consumers of one builder), (c) a call/return where relocation shifts offsets.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/binary_translator.cpp
```diff
@@ -68,13 +70,31 @@ LegalizationLookupFn select_legalization(rj_code_arch_t guest, rj_code_arch_t ho
...
 }
 
-[[nodiscard]] BasicBlock *block_for_offset(const std::vector<std::unique_ptr<BasicBlock>> &blocks,
-                                           uint64_t offset) {
-  for (const auto &block : blocks) {
-    if (block && block->start_offset() <= offset && offset < block->end_offset())
-      return block.get();
-  }
-  return nullptr;
+[[nodiscard]] bool compute_sopp_branch_offset(uint64_t branch_pc, uint64_t target,
```
atgutier: Agent review:

Duplicated branch-offset helper — binary_translator.cpp:~72-92: new compute_sopp_branch_offset
  re-implements compute_sopp_branch_simm16 (still used in 6 sites) and drops the branch_pc%4/target%4
  alignment checks — a subtle divergence. Consolidate to one.
  - scalar_sop2_opcode switch fall-through — static_pc_recovery.cpp:103-108 /
  instruction_builder.cpp:49-53: relies on the inner switch being exhaustive; a new enumerator
  silently falls into the RISC-V cases. Add return std::nullopt; after. (Two near-identical copies —
  share one.)

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/static_pc_recovery.cpp
```diff
@@ -0,0 +1,357 @@
...
+      recovery_begin_offset = offsets[i + 1];
+      recovery_end_offset = offsets[i + 3] + insts[i + 3]->size();
+      matched_static_target = true;
+    }
+    if (!matched_static_target) {
+      continue;
+    }
+
+    const auto &call_indices = indirect_call_indices_by_sreg[*sdst];
+    auto first_call = std::ranges::lower_bound(call_indices, i + 3, std::less<>{},
```
atgutier: Agent review comment:

No-op ranges projection — static_pc_recovery.cpp:296: lower_bound(..., std::less<>{}, identity) →
  just std::ranges::lower_bound(call_indices, i+3).

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/static_pc_recovery.cpp
```diff
@@ -0,0 +1,357 @@
...
+    if (auto ssrc0 = scalar_pc_sreg(arch, *insts[i], word, ScalarPcOp::SwapPc64))
+      indirect_call_indices_by_sreg[*ssrc0].push_back(i);
+    if (auto ssrc0 = scalar_pc_sreg(arch, *insts[i], word, ScalarPcOp::SetPc64))
+      indirect_call_indices_by_sreg[*ssrc0].push_back(i);
+  }
+
+  for (size_t i = 0; i + 3 < insts.size(); ++i) {
+    auto sdst =
+        scalar_pc_sreg(arch, *insts[i], text_word_at(text, offsets[i]), ScalarPcOp::GetPc64);
+    if (!sdst || *sdst >= 127)
```
atgutier: Can you get 127 from a named value (maybe from the ISA traits)?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/static_pc_recovery.h
```diff
@@ -0,0 +1,74 @@
...
+#include <vector>
+
+namespace rocjitsu {
+
+class Instruction;
+struct KernelTextLayout;
+
+/// @brief Recovered indirect PC-relative branch through a statically-built PC register.
+///
+/// @details BasicBlock construction turns the recovered destination into an
```
atgutier: Can you comment on which targets are valid pre- and post-reloc?

atgutier (dismissed):

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/binary_translator.cpp
Groverkss: Will change this, but on targets that are not gfx1250, this is valid, because the minimum number of words needed to build a long jump address is same as the canonical long jump we emit. But I will turn it into a runtime failure instead, although not reproducable on non gfx1250 (gfx1250 has a add_pc_i64 instruction)

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/static_pc_recovery.h
Groverkss: Any branch targets that are static should be recoverable in theory, but we only match the ones we know tensile, rocisa and llvm emit. It's very hard to find static branch targets in general, but generators do not usually emit it like that.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
```diff
@@ -161,7 +161,7 @@ void insert_file_bytes(std::vector<uint8_t> &image, Elf64_Ehdr &ehdr,
 [[nodiscard]] bool target_supports_wave32(rj_code_arch_t arch) {
```
accauble, line 161: Changes appear to be breaking a test:

```
Note: Google Test filter = CodeObjectPatcher.AppliesArchSpecificWgpModeBit
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from CodeObjectPatcher
[ RUN      ] CodeObjectPatcher.AppliesArchSpecificWgpModeBit
/work/acaublec/workspaces/test-rocjitsu/rocm-systems/emulation/rocjitsu/tests/dbt/translate_test.cpp:1224: Failure
Expected equality of these values:
  ((*gfx1250_rsrc1 & COMPUTE_PGM_RSRC1_MEM_ORDERED) >> COMPUTE_PGM_RSRC1_MEM_ORDERED_SHIFT)
    Which is: 0
  1u
    Which is: 1

/work/acaublec/workspaces/test-rocjitsu/rocm-systems/emulation/rocjitsu/tests/dbt/translate_test.cpp:1225: Failure
Expected equality of these values:
  ((*gfx1250_rsrc1 & COMPUTE_PGM_RSRC1_FWD_PROGRESS) >> COMPUTE_PGM_RSRC1_FWD_PROGRESS_SHIFT)
    Which is: 0
  1u
    Which is: 1

[  FAILED  ] CodeObjectPatcher.AppliesArchSpecificWgpModeBit (0 ms)
```

Does this fail on your system?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp
```diff
@@ -0,0 +1,1563 @@
...
+    }
+  }
+
+  std::array<std::optional<PcValue>, REGISTER_SET_MAX_SGPRS> builders_;
+  std::vector<uint16_t> active_pairs_;
+  uint64_t dirty_lo_ = 0;
+  uint64_t dirty_hi_ = 0;
+};
+
+[[nodiscard]] uint32_t text_word_at(std::span<const uint8_t> text, uint64_t offset) {
```
accauble, line 361: You've got this function added to binary_translator.cpp and it seems to exist in kernel_text_layout too. I've been responsible for code duplication as well, so I think we need to think about refactoring soon.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp
```diff
@@ -0,0 +1,1563 @@
...
+  // naturally instead of needing a separate bounds status.
+  uint32_t word = 0;
+  if (offset + sizeof(word) <= text.size())
+    std::memcpy(&word, text.data() + offset, sizeof(word));
+  return word;
+}
+
+[[nodiscard]] std::optional<uint8_t> scalar_pc_opcode(rj_code_arch_t arch, ScalarPcOp op) {
+  // s_getpc/setpc/swappc are adjacent SOP1 opcodes within each AMDGPU ISA
+  // family currently supported by rocjitsu. Keep this mapping local to the
```
accauble, line 374: Nit: Not sure why the analysis should be local instead of in the generator.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp
```diff
@@ -0,0 +1,1563 @@
...
+    if (source_it == block_by_offset.end() || target_it == block_by_offset.end())
+      continue;
+
+    std::vector<size_t> &successors = blocks[source_it->second].successors;
+    if (std::ranges::find(successors, target_it->second) == successors.end())
+      successors.push_back(target_it->second);
+  }
+}
+
+void recover_signed_delta_templates(const AnalysisContext &ctx,
```
accauble, line 1434: My LLM flagged this as functionality worth of a test but without one.

accauble (changes requested):
Lots of cool work and I'm happy to see it! But there is a failing test that needs to be addressed.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp
```diff
@@ -0,0 +1,1563 @@
...
+         (BRANCH | COND_BRANCH | INDIRECT_BRANCH | INDIRECT_CALL | PROGRAM_TERMINATOR);
+}
+
+[[nodiscard]] bool is_direct_call(const Instruction &inst) {
+  return (inst.flags() & INDIRECT_CALL) != 0 && inst.branch_offset_bytes().has_value();
+}
+
+[[nodiscard]] bool has_no_direct_successor(const Instruction &inst) {
+  // Indirect branches have no known target until this analysis recovers one.
+  // Indirect calls still expose their ordinary fallthrough/return continuation
```
kuhar, line 602: The temporary CFG for direct `s_call` carries PC-builder facts through the fallthrough edge without applying callee effects. The builder can be materialized before the call, `build_analysis_blocks()` adds both the direct callee edge and the fallthrough continuation, and `scan_block()` only kills the call's return SGPR pair. There is no return edge from the callee back to this continuation, so writes in the callee never kill a builder pair that is later consumed by `s_setpc`/`s_swappc`.

For static recovery, that can turn a value clobbered by the callee into a relocated fixup instead of failing closed. Please either kill carried PC-builder facts across direct calls, or model call/return edges so callee transfers reach the continuation.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
```diff
@@ -161,7 +161,7 @@ void insert_file_bytes(std::vector<uint8_t> &image, Elf64_Ehdr &ehdr,
...
   return arch == ROCJITSU_CODE_ARCH_RDNA1 || arch == ROCJITSU_CODE_ARCH_RDNA2 ||
          arch == ROCJITSU_CODE_ARCH_RDNA3 || arch == ROCJITSU_CODE_ARCH_RDNA3_5 ||
-         arch == ROCJITSU_CODE_ARCH_RDNA4 || arch == ROCJITSU_CODE_ARCH_GFX1250;
+         arch == ROCJITSU_CODE_ARCH_RDNA4;
 }
 
 [[nodiscard]] bool target_uses_gfx10_plus_mode_bits(rj_code_arch_t arch) {
```
kuhar, line 167: `CodeObjectPatcher.AppliesArchSpecificWgpModeBit` is failing on this branch. I reproduced it locally with `ctest -R '^CodeObjectPatcher\.AppliesArchSpecificWgpModeBit$'`: for gfx1250, both `COMPUTE_PGM_RSRC1_MEM_ORDERED` and `COMPUTE_PGM_RSRC1_FWD_PROGRESS` read back as `0` when the test expects `1`.

The source path matches the failure: `target_uses_gfx10_plus_mode_bits()` now delegates to `target_supports_wave32()`, and this PR removed `GFX1250` from the wave32 predicate. Please decouple the gfx10+ RSRC1 mode-bit predicate from wave32 support, or otherwise keep gfx1250 covered for this mode-bit path.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp
```diff
@@ -0,0 +1,1563 @@
...
+      ++unresolved;
+      continue;
+    }
+    const auto &facts = entry_facts[consumer.block_index];
+    auto it = facts.find(consumer.pair_lo);
+    if (it == facts.end() || it->second.values.empty()) {
+      ++unresolved;
+      continue;
+    }
+    if (it->second.incomplete && it->second.values.size() >= kMaxIndirectTargetsPerConsumer) {
```
kuhar, line 1390: The incomplete lattice still emits recovered fixups after a real kill reaches the consumer. A `Kill` transfer becomes `{values={}, incomplete=true}`, `join_lattice_value()` preserves that flag, and `classify_pending_consumers()` still emits fixups whenever the joined fact has a non-empty sub-cap concrete set. That means "one predecessor built this PC" plus "one predecessor wrote the pair through unmodeled code" recovers the consumer instead of failing closed.

That also suppresses the later DBT diagnostic, because `binary_translator.cpp` marks the source offset as recovered. Please distinguish benign path-insensitive incompleteness from a killed/unmodeled path reaching the consumer, or fail closed when a kill participates in the join. A focused two-predecessor test would make this concrete.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/semantic/cdna4_to_cdna3.cpp
```diff
@@ -209,6 +214,12 @@ void emit_cdna3_exec_mask(std::vector<uint32_t> &words, uint64_t mask) {
   emit_s_mov_b32_lit(words, kExecLo + 1, static_cast<uint32_t>(mask >> 32));
 }
 
+[[nodiscard]] uint8_t choose_cdna3_exec_save_sgpr(const TranslationContext &context) {
```
kuhar: `choose_cdna3_exec_save_sgpr()` reuses `s[100:101]` when `context.num_sgprs >= 102`. The three callers only grow the descriptor when `saved_exec + 2 > context.num_sgprs`, so a kernel that already declares the CDNA3 limit of 102 ordinary SGPRs will save EXEC into registers inside the kernel's allocated SGPR window.

Unless those registers are proven dead at each lowering site, this can clobber high-SGPR kernels during the permlane, `ds_read_b64_tr_b16`, or MUBUF-to-LDS expansions. Please choose a liveness-proven free pair or fail closed when no fresh descriptor-backed pair is available.

kuhar (commented):
Looks very nice overall, I appreciate the docs! The review agents found some issues.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp
Groverkss: Yes, we need to do a bit of refactoring soon... analysis/ lives outside of code/ so it's hard right now. We need to move analysis/ inside code/ as it depends on code/ already and it's not very useful standalone. I'm going to do this eventually and improve things.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp
Groverkss: I will eventually move this to the generator. I have a pending issue of making the generator generate op codes for all operations and use that, but it's a larger change than this.

Groverkss: I will eventually move this to the generator. I have a pending issue of making the generator generate op codes for all operations and use that, but it's a larger change than this.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp
Groverkss: fixed

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/semantic/cdna4_to_cdna3.cpp
Groverkss: I accidently added this commit as part of this pr. I will remove this from here.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp
Groverkss: fixed by adding a "killed" bool to lattice.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/liveness.cpp
```diff
@@ -96,18 +96,43 @@ std::vector<const BasicBlock *> reverse_post_order(KernelBlockScope blocks) {
...
+LivenessAnalysis::LivenessAnalysis(KernelBlockScope blocks, LivenessAnalysisOptions options,
+                                   std::span<const ScopedCfgEdge> extra_edges) {
   min_free_vgpr_ = options.min_free_vgpr;
-  analyze(blocks);
+  analyze(blocks, extra_edges);
 }
 
-void LivenessAnalysis::analyze(KernelBlockScope blocks) {
+void LivenessAnalysis::analyze(KernelBlockScope blocks,
+                               std::span<const ScopedCfgEdge> extra_edges) {
```
kuhar, line 106: The new `extra_edges` path is the load-bearing part of call/return liveness, but it has no direct unit test. The only `LivenessAnalysis(` construction in the tests still goes through `analyze_scope()` with the default empty edge set, so the scoped `caller -> callee` and `return -> continuation` edges are covered only indirectly by DBT E2E tests.

Please add a small `liveness_test.cpp` case that constructs `LivenessAnalysis(scope, {}, edges)` directly and asserts `live_in`/`live_out` across a call edge and return-continuation edge. That would catch regressions in the new dataflow wiring without depending on byte-exact translated output.

kuhar (commented):
LGTM overall, just two comment threads worth addressing IMO: https://github.com/ROCm/rocm-systems/pull/7677#discussion_r3492687560 and https://github.com/ROCm/rocm-systems/pull/7677#discussion_r3492687656.

Testing strategy for `indirect_branch_discovery`: the E2E coverage is useful, but the new analysis is large and most behavior is only exercised through `BasicBlock::build` or the DBT pipeline. A small direct unit-test layer for `discover_indirect_branch_edges(...)` would make the fragile analysis cases much easier to pin down.

Two high-value cases to cover there are the straight-line temp-delta builders (`match_temp_add_pattern` / `match_temp_sub_pattern`) and bounded multi-target joins around `kMaxIndirectTargetsPerConsumer`. Those are currently checked only indirectly, yet they determine whether the pass recovers all intended static targets or correctly fails closed at the cap.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp
kuhar: The direct-call fix only invalidates PC builders that were created inside the call block. `BlockState` is reset per block, so a builder produced in an earlier block can enter a block containing `s_call_b64` as a carried fact; the call then produces no `Kill` for that pair and the fallthrough continuation can still recover a later `s_setpc` from the stale pre-call value.

That is still fail-open if the callee clobbers the builder pair: DBT relocates the builder instead of leaving the consumer unresolved. We should either treat direct calls as killing all carried PC-builder pairs on the continuation edge, or model the call/return edge so callee writes reach the continuation. A two-block regression test would pin the gap: build the PC in block A, branch to block B containing `s_call_b64`, then consume the pair after the call.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp
kuhar: The new `killed` flag fixes the explicit `Kill` join, but the incomplete join path still emits fixups when one predecessor has a concrete PC and another predecessor has no fact for the pair. `run_block_dataflow()` marks that case only as `incomplete=true`, and `classify_pending_consumers()` still emits when the concrete set is below the cap and `killed` is false.

If that missing predecessor is feasible, the consumer can read an unconstrained/unrelocated PC while DBT marks the source as recovered. We should either gate the incomplete-but-emit path on clear feasibility evidence, or document this as an intentional precision tradeoff at `classify_pending_consumers()` and add a focused test for the one-concrete / one-missing-predecessor join.

accauble (commented):
My comments were addressed (thanks!), but I do think @kuhar's comments are worth addressing before merging.

kuhar (approved):

accauble (approved):

PR #7679, opened 2026-06-23
clr: Collapse low-ROI graphs onto the launch stream

saleelk:
## Motivation
HIP graph launch is host-overhead-bound for launch-bound workloads (many tiny,
near-serial kernels). PyFR's 2D couette-flow case (ROCM-25757) spends most of a
step in hipGraphLaunch submitting cross-stream barriers/completion signals that
buy no real overlap.

## Technical Details
- BuildSyncPlan PASS 0 collapses low-ROI single-device graphs onto the launch
  stream (one in-order queue; zero cross-stream barriers/signals). Streams sized
  to the decision at instantiate.
- Redundant cross-stream barrier elimination (PASS 2 compute / PASS 3 materialize),
  indexed by segment id.
- ShouldCollapseToSingleStream gate (DEBUG_HIP_GRAPH_MIN_OVERLAP): collapse when
  overlappable work < min_overlap * cross-stream sync cost. Cost counts only the
  barriers/signals multi-stream actually emits (cross-stream deps +
  needs_completion_signal); work weighted by occupancy passes. Single-device only.
- AQL logging: correct ext kernel-dispatch detection in batched dispatch; gate
  decision logged at LOG_DETAIL_DEBUG.

## JIRA ID
ROCM-25757

## Test Plan
- hip-tests GraphPerformance: PyFR couette-flow launch-overhead benchmarks + a
  CollapseCorrectness test (diamond graph, cross-branch data deps, poisoned
  buffers, run with DEBUG_HIP_GRAPH_MIN_OVERLAP=0 and =2).
- GraphBench --verify ordering checks across topologies (gate on/off).

## Test Result
PyFR 2D couette-flow (MI300X, idle GPU 0), per RHS eval:
- Default (collapse on): 20.6 MDOF/s
- MIN_OVERLAP=0 (off): 11.9 MDOF/s

CollapseCorrectness + GraphBench --verify: all pass on both gate paths.


projects/clr/hipamd/src/hip_graph_internal.cpp
```diff
@@ -1248,7 +1479,36 @@ hipError_t GraphExec::Init() {
...
+            "[hipGraph] Init: %zu device(s), %u total stream(s) (per-device cap: %u)",
+            max_streams_dev_.size(), total_streams, DEBUG_HIP_FORCE_GRAPH_QUEUES);
+    for (auto const& [dev_id, num_streams] : max_streams_dev_) {
+      if (num_streams > 0) {
+        status = CreateStreams(num_streams, dev_id);
+        if (status != hipSuccess) {
+          return status;
+        }
+      }
+    }
```
LeoZhao-BJ, line 1524: why put createstreams() after  CaptureAQLPackets()?  raise risk of missing initialization cleanup.

projects/clr/hipamd/src/hip_graph_internal.cpp
```diff
@@ -1202,6 +1285,158 @@ void GraphExec::PrecomputeStreamAssignment() {
...
+    if (seg.segment_ids_edges.empty()) {
+      if (leaf_sync_required) ++signal_est;
+    } else {
+      ++signal_est;
+    }
+    for (Node n : seg.nodes) seg_work[i] += node_work(n);
+    total_work += seg_work[i];
+  }
+  // Total per-launch cross-stream overhead multi-stream pays on the host path.
+  const size_t sync_cost = barrier_est + signal_est;
```
LeoZhao-BJ, line 1413: is there assumption that one barrier and one signal have same cost?

projects/clr/hipamd/src/hip_graph_internal.cpp
```diff
@@ -1202,6 +1285,158 @@ void GraphExec::PrecomputeStreamAssignment() {
...
+    }
+    // A segment emits at most one completion signal (mirrors
+    // ComputeCompletionSignalFlags): a producer signals when a consumer may land
+    // on another stream — which multi-stream assignment will generally arrange,
+    // so any segment with downstream edges is counted — and a leaf signals when
+    // leaf-join back to the launch stream is required.
+    if (seg.segment_ids_edges.empty()) {
+      if (leaf_sync_required) ++signal_est;
+    } else {
+      ++signal_est;
```
LeoZhao-BJ, line 1407: seems here logic is different with ComputeCompletionSignalFlags, why here always ++

saleelk: so the original logic is every segment that has an edge would need a completion signal.  ie its edge would have to wait on this segment to finish.  that's why we need to ++. The computeCompletionSignal tries to reduce the impact by taking advantage of stream ordering ie segments on same stream would necessarily wait for previous to complete since stream works like a FIFO.

projects/clr/hipamd/src/hip_graph_internal.cpp
saleelk: yes, I just attribute a cost factor here, the actual overhead is not the same.  a barrier packet itself may have a signal. but in Graphs I avoid the signal on the barrier packet itself as a barrier is only used for a wait here. The signal is put on the last packet in a segment,

projects/clr/hipamd/src/hip_graph_internal.cpp
saleelk: There is no cleanup anywhere, here. its after the captureAQLPackets as we make the decision to collapse in there, if collapsed, we do not need multiple streams created.

projects/clr/hipamd/src/hip_graph_internal.cpp
LeoZhao-BJ: ok, here the comparing target is original multi-stream logic. computeCompletionSignal is new optimized version for segment.

projects/clr/hipamd/src/hip_graph_internal.cpp
LeoZhao-BJ: but here it covers use_segment_scheduling_, not just for collapse, CaptureAQLPackets() is also for use_segment_scheduling_ actually. 
so my understanding is you assume if any fails on createstream, the whole session will fail w/o opportunity to move forward?

saleelk: correct. if CreateStreams fails we have an underlying problem

projects/clr/hipamd/src/hip_graph_internal.cpp
saleelk: yes, this collapsing is rather hacky because our hardware sucks at resolving barriers. I want to disable this after testing on gfx1250

gandryey (approved):

PR #7683, opened 2026-06-23
AIRUNTIME-2362: Fix data race on gpu_error_ with atomic

jungx098:
## Motivation

`amd::Device::gpu_error_` is a process-global `static cl_int` read on every public HIP API entry and written from background threads (HSA async-event and queue-error callbacks, PAL flush failure). This concurrent access without synchronization is a data race and undefined behavior.

## Technical Details

Convert `gpu_error_` to `std::atomic<cl_int>` and update all access sites across 5 files. Writes use `store(memory_order_relaxed)` and reads use `load(memory_order_relaxed)`; the error code carries no associated payload, so only per-variable atomicity is required with no ordering guarantee. No functional or semantic change.

## JIRA ID

- AIRUNTIME-2362

## Test Plan

N/A - synchronization-only change with no functional behavior change.

## Test Result

N/A

## Submission Checklist


jungx098:
@copilot review

chrispaquot (approved):

PR #7686, opened 2026-06-23
[rocjitsu] Update corpus ref.

amd-eochoalo:
## Motivation

Greater coverage for CI.

## Technical Details

* Update reference
* Test sequentially instead of a test matrix
* We are moving away from test matrix because pytests can share the same rocjitsu build. This saves resources.

## JIRA ID

N/A

## Test Plan

Test on CI.

## Test Result

CI passes.

## Submission Checklist


kuhar (approved):
Would be also worth mentioning why we dropped the test matrix

PR #7687, opened 2026-06-23
[ci] Update thread limit env vars to 14

geomin12:
Increase OPENBLAS_NUM_THREADS and OMP_NUM_THREADS from 4 to 14 to better utilize available CPU cores during tests.


tony-davis (approved):
Thanks!

PR #7694, opened 2026-06-23
[ci] Adding test dependency to rocm-systems multi-arch CI 

geomin12:
Changes:
- Adding test dependency selection to multi-arch CI
- adding new `.github/scripts/get_changed_projects.py` to provide a clean git diff script, instead of relying on `therock_configure_ci.py` (plan to delete this post migration, it's a messy script)

test ran here: https://github.com/ROCm/rocm-systems/actions/runs/28048813827/job/83033807636 (cancelled to save resources)

### DIFF
--- a/.github/scripts/ci_utils.py
+++ b/.github/scripts/ci_utils.py
@@ -1,0 +1,76 @@
+"""Shared CI utilities for GitHub Actions workflow scripts."""
+
+import fnmatch
+import logging
+import os
+import subprocess
+import time
+from typing import Callable, Iterable, Mapping, Tuple, Type, TypeVar
+
+F = TypeVar("F", bound=Callable[..., object])
+
+logging.basicConfig(level=logging.INFO)
+
+
+def retry(
+    max_attempts: int,
+    delay_seconds: float,
+    exceptions: Tuple[Type[BaseException], ...],
+) -> Callable[[F], F]:
+    """Retry decorator with exponential backoff."""
+
+    def decorator(func: F) -> F:
+        def wrapper(*args, **kwargs):
+            attempt = 0
+            while attempt < max_attempts:
+                try:
+                    return func(*args, **kwargs)
+                except exceptions as e:
+                    print(
+                        f"Exception {str(e)} thrown when attempting to run, "
+                        f"attempt {attempt} of {max_attempts}"
+                    )
+                    attempt += 1
+                    if attempt < max_attempts:
+                        backoff = delay_seconds * (2 ** (attempt - 1))
+                        time.sleep(backoff)
+            return func(*args, **kwargs)
+
+        return wrapper
+
+    return decorator
+
+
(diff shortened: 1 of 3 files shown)


.github/scripts/ci_utils.py
```diff
@@ -0,0 +1,76 @@
...
+                        backoff = delay_seconds * (2 ** (attempt - 1))
+                        time.sleep(backoff)
+            return func(*args, **kwargs)
+
+        return wrapper
+
+    return decorator
+
+
+@retry(max_attempts=3, delay_seconds=2, exceptions=(TimeoutError,))
```
erman-gurses: `subprocess.run(..., timeout=60)` raises `subprocess.TimeoutExpired`, not `TimeoutError`. As written, the retry logic won't catch timeouts from `git diff`. Consider catching `subprocess.TimeoutExpired` (or both exceptions if that's intentional).

.github/scripts/ci_utils.py
```diff
@@ -0,0 +1,76 @@
...
+                except exceptions as e:
+                    print(
+                        f"Exception {str(e)} thrown when attempting to run, "
+                        f"attempt {attempt} of {max_attempts}"
+                    )
+                    attempt += 1
+                    if attempt < max_attempts:
+                        backoff = delay_seconds * (2 ** (attempt - 1))
+                        time.sleep(backoff)
+            return func(*args, **kwargs)
```
erman-gurses: This performs one additional invocation after the retry loop exits. With `max_attempts=3`, `func()` can actually be called 4 times. Would it make more sense to re-raise the last exception instead of calling `func()` again here?

.github/scripts/ci_utils.py
```diff
@@ -0,0 +1,76 @@
...
+        def wrapper(*args, **kwargs):
+            attempt = 0
+            while attempt < max_attempts:
+                try:
+                    return func(*args, **kwargs)
+                except exceptions as e:
+                    print(
+                        f"Exception {str(e)} thrown when attempting to run, "
+                        f"attempt {attempt} of {max_attempts}"
+                    )
```
erman-gurses: Since this module already uses logging, would it make sense to use `logging.warning()` (or `logging.exception()`) here instead of print() for consistency?

.github/scripts/ci_utils.py
```diff
@@ -0,0 +1,76 @@
...
+
+def retry(
+    max_attempts: int,
+    delay_seconds: float,
+    exceptions: Tuple[Type[BaseException], ...],
+) -> Callable[[F], F]:
+    """Retry decorator with exponential backoff."""
+
+    def decorator(func: F) -> F:
+        def wrapper(*args, **kwargs):
```
erman-gurses: Could be like `def wrapper(*args: object, **kwargs: object):`

erman-gurses (approved):
Looks good overall - please see my comments to address.

geomin12:
CI checks are irrelevant to changes. landing

PR #7234, opened 2026-06-14
[hotswap] Enable the hotswap tool via HSA_TOOLS_LIB and forward foreign-ISA code objects

nirmie:
## Motivation

Allow the comgr hotswap tool to be used purely via the existing `HSA_TOOLS_LIB` env variable — no `LD_PRELOAD` and no forked ROCR/CLR. Two pieces: (1) ROCR loads the tool via `HSA_TOOLS_LIB` even when rocprofiler-register (v3) succeeds; (2) CLR forwards a foreign-ISA code object to the HSA loader so the tool can transpile it to the device ISA. Every path is gated on the hotswap tool actually being loaded, so there is no behavior change when it is not.

## Technical Details

- **rocclr/device/hotswap.hpp** (new): single source of truth for hotswap routing in CLR — `Enabled()` (true only when `libamd_comgr_hotswap_tool.so` is named in `HSA_TOOLS_LIB`), a supported `(source -> target)` ISA allowlist (`IsSupportedPair`), and an ISA-name token helper (`IsaIsGfx`). Used by both paths below.
- **hipamd/src/hip_fatbin.cpp**: when the tool is loaded, also request the allowlisted source ISAs for each device target during unbundling, then select the first supported source bundle and forward it to the HSA loader. Gated on `Enabled()`; when the tool is not loaded the selection is bypassed and native/generic/SPIR-V handling is unchanged.
- **rocclr/device/devprogram.cpp**: when the tool is loaded, allow a foreign source ISA past the ISA-compatibility gate only when `(source -> device)` is in the allowlist, so the loader can transpile it. Unsupported pairs (and the no-tool case) error exactly as before.
- **rocr-runtime/.../runtime.cpp**: in `LoadTools`, enable v1 registration when the hotswap tool is named in `HSA_TOOLS_LIB`, so it loads alongside rocprofiler-register (v3) without `HSA_TOOLS_ROCPROFILER_V1_TOOLS`. v3 is trace-only; the tool must intercept and modify `hsa_code_object_reader_create_from_memory`, which requires the v1 path. Scoped to the hotswap tool, so other v1 tools are unaffected.

## JIRA ID

N/A

## Test Plan

- Unit-tested `hotswap.hpp` standalone (`g++ -std=c++17 -Wall -Wextra`): allowlist (`IsSupportedPair`), tool-name gating (`Enabled`), and ISA token matching (`IsaIsGfx`, including boundary cases) all pass.
[...]

PR #7629, opened 2026-06-22
[hotswap] Fixes to use projects/hotswap libhsa_hotswap.so  and resolve header paths

nirmie:
## Motivation
This PR contains two components. One component is to update to use the tooling bundled in projects/hotswap instead of the tool that exists at comgr (deprecated). The other component is to resolve amd_comgr header issues during the build. Additionally, there are some changes to the CMakeLists.txt to have the hotswap HSA tool against hsa-runtime64::hsa-runtime64.

## Technical Details
~~Builds on top of Harsh's PRs here: #7577~~ (deprecated, #7577 code is directly in this)
Relevant PRs: amd-llvm: https://github.com/ROCm/llvm-project/pull/3007

## JIRA ID
N/A

## Test Plan
Contained build for the tool (verifies if linking resolves)
```
git sparse-checkout add projects/rocr-runtime/runtime/hsa-runtime/inc 2>/dev/null || true

P=~/hotswap/npi-build/_gapfix/comgr-prefix
DIST=~/hotswap/npi-build/therock/build/dist/rocm

cmake -S projects/hotswap -B /tmp/build-hotswap -GNinja -DCMAKE_BUILD_TYPE=Release \
  -DCMAKE_PREFIX_PATH="$DIST" \
  -Damd_comgr_DIR="$P/lib/cmake/amd_comgr" \
  -Dhsa-runtime64_DIR="$DIST/lib/cmake/hsa-runtime64"

cmake --build /tmp/build-hotswap --target hsa-hotswap
```
Tool name testing must be done with full build

## Test Result
Ran:
ls -l /tmp/build-hotswap/libhsa-hotswap.so
nm -D /tmp/build-hotswap/libhsa-hotswap.so | grep hotswap_rewrite
```
nm -D /tmp/build-hotswap/libhsa-hotswap.so | grep hotswap_rewrite
-rwxrwxr-x 1 nisenthi nisenthi 50480 Jun 22 18:45 /tmp/build-hotswap/libhsa-hotswap.so
                 U amd_comgr_hotswap_rewrite@amd_comgr_3.2
```

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #7715, opened 2026-06-23
[hotswap][clr] Restrict HotSwap forwarding to validated source/target pairs

xintin:
## Motivation
HotSwap foreign-ISA forwarding in `FatBinaryInfo::ExtractFatBinaryUsingCOMGR` is gated by`amd::hotswap::kSupportedPairs`: a `(source -> device target)` pair must be listed for a foreign source bundle to be forwarded to the HSA loader for transpilation. The forwarding branch is intentionally evaluated before the native/generic branches, so any listed target is routed through the comgr hotswap tool whenever the tool is loaded (`HSA_TOOLS_LIB` names `libamd_comgr_hotswap_tool.so`).

With `gfx942` and `gfx950` in the allowlist, fatbins built for those devices (which also contain a `gfx1250` bundle) were force-transpiled from the `gfx1250` source instead of using their already-correct native code object. That needlessly invokes the tool and crashes a large fraction of HIP workloads on gfx942.

Until per-target transpilation is validated, restrict `kSupportedPairs` to `gfx1250 -> gfx1250` (B0 -> A0) only. `gfx942` and `gfx950` will be re-added as their transpilation paths are validated. No selection-order or `HIP_FORCE_SPIRV_CODEOBJECT` changes; the no-tool path is unchanged. Related: #7234.

## JIRA ID
NA

## Test Plan
Built ROCm via TheRock with the hotswap tool on gfx942 (MI300X). Ran the hip-tests catch suite (full 4191 + a 1048-test shard) with and without `HSA_TOOLS_LIB`, comparing pass/fail.
Will also run through the TheRock CI before merging.

## Test Result
With the tool enabled, the suite shows zero new failures versus the no-tool baseline; the remaining failures are pre-existing and reproduce without the tool. No `HotSwap: forwarding` occurs on gfx942 (the allowlist no longer matches that target).

## Submission Checklist


lamb-j:
We need to update the tool name to libhsa-hotswap.so

lamb-j (approved):

gandryey (approved):

nirmie (approved):

nirmie:
libhsa-hotswap.so rename @lamb-j mentions will be done in #7629

PR #7724, opened 2026-06-24
[rocjitsu][CI] Use asan and ubsan for CI

amd-eochoalo:
## Motivation

Detect address and undefined behaviour bugs.

## Technical Details

* Modify cmake to allow asan and ubsan to run together

## JIRA ID

N/A.

## Test Plan

Test on CI.

## Test Result

CI passes.

## Submission Checklist

### DIFF
--- a/.github/workflows/rocjitsu-corpus-tests.yml
+++ b/.github/workflows/rocjitsu-corpus-tests.yml
@@ -41,15 +41,31 @@ env:
   ROCJITSU_CORPUS_DIR: ${{ github.workspace }}/rocjitsu-test-corpus
 
 jobs:
-  pytest:
+  test:
     runs-on: azure-linux-scale-rocm
-    name: pytest
+    name: test (${{ matrix.name }})
     container:
       image: ghcr.io/rocm/no_rocm_image_ubuntu24_04@sha256:4150afe4759d14822f0e3f8930e1124f26e11f68b5c7b91ec9a02b20b1ebbb98
       options: --user 0:0
       env:
         DEBIAN_FRONTEND: noninteractive
     continue-on-error: true
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: release
+            build_type: Release
+            enable_asan: 0
+            enable_ubsan: 0
+            enable_tsan: 0
+            enable_msan: 0
+          - name: asan-ubsan
+            build_type: RelWithDebInfo
+            enable_asan: 1
+            enable_ubsan: 1
+            enable_tsan: 0
+            enable_msan: 0
 
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/rocjitsu-corpus-tests.yml
```diff
@@ -41,15 +41,25 @@ env:
...
     continue-on-error: true
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: release
+            build_type: Release
+            sanitizer: ""
+          - name: asan-ubsan
+            build_type: RelWithDebInfo
```
kuhar, line 64: is this enough to enable assertions or we need to pass `-O2 -g`?

emulation/rocjitsu/CMakeLists.txt
```diff
@@ -117,11 +121,62 @@ if(LTO)
 endif()
 
 if(RJ_SANITIZER)
-    message(STATUS "Sanitizer enabled: ${RJ_SANITIZER}")
+    string(TOLOWER "${RJ_SANITIZER}" _rj_sanitizer_value)
```
kuhar: Why do we need these changes?

amd-eochoalo: The CMake script already contained a way to use asan and ubsan, but it didn't enable them working together. So these changes allow one to build asan and ubsan together using the same CMake variable that was already defined.

The LLM added "address", "undefined", "thread", "memory" as synonyms for asan, ubsan, tsan, msan. I think it is ok to keep.

.github/workflows/rocjitsu-corpus-tests.yml
amd-eochoalo: The LLM suggested to change the following CMake flags: 

```
              -DCMAKE_C_FLAGS_RELWITHDEBINFO="-O2 -g"
              -DCMAKE_CXX_FLAGS_RELWITHDEBINFO="-O2 -g"
```

I've added them in https://github.com/ROCm/rocm-systems/pull/7724/changes/422427acf63433f2e513bdb56d3eebb058cb4fbd

emulation/rocjitsu/CMakeLists.txt
kuhar: IMO it's better to simplify this cmake logic and require the user to set these in a friendlier way, e.g.:
`-DRJ_ENABLE_UBSAN=1 -DRJ_ENABLE_ASAN=1`. WDYT?

amd-eochoalo: Thanks! I've updated the PR.

kuhar (approved):

PR #7727, opened 2026-06-24
feat(clr): Use device-resident queue ring buffers

saleelk:
- Switch CLR queue acquisition to the descriptor-based hsa_amd_queue_create API (nested hsa_amd_queue_create_desc_t). Ring-buffer placement is selected purely via the descriptor flags: HSA_AMD_QUEUE_CREATE_DEVICE_MEM_RING_BUF for device memory, no flag for system memory. Cache the per-queue metadata ring buffer and hardware doorbell in QueueExtras.
- Allocate the device-memory ring buffer (and appended metadata ring) with AllocateExecutable | AllocateUncached. AllocateExecutable is required because the CP hardware requests execute permissions for packet buffer accesses (the hardware treats AQL packets as an execution language). AllocateUncached propagates through ROCr → KFD (KFD_IOC_ALLOC_MEM_FLAGS_UNCACHED) → AMDGPU (AMDGPU_GEM_CREATE_UNCACHED) and sets the page MTYPE to UC in the *GPU* page tables, not the CPU side. Without this the coarse-grain VRAM default is MTYPE_NC; removing the uncached attribute causes GPU hangs on gfx1250 with older MEC firmware.
- Add DEBUG_CLR_AQL_DEV_QUEUE release flag (default 0): device-memory AQL ring buffers are auto-enabled on gfx94x/gfx125x when Large BAR is available; set the flag to 1 to force device-memory rings on other architectures, or to 0 to force a system-memory ring everywhere. Other architectures always use system memory unless overridden. Queue descriptors remain host-resident.
- Add DEBUG_CLR_DIRECT_DOORBELL release flag (default off): write the hardware doorbell directly from CLR through a cached uncacheable doorbell pointer, bypassing hsa_signal_store_screlease. On the MOVDIR64B path the write-combining buffer is already closed, so the pre-doorbell fence is skipped; falls back to the ROCr signal store when the doorbell pointer is unavailable.
- Cache MOVDIR64B CPU capability at device initialization (via Os::hasMovdir64b / CPUID) and use it only for Large BAR-backed device-memory ring buffers to publish full 64-byte AQL packets with a single atomic store. Add DEBUG_CLR_USE_MOVDIR64B release flag (default 1) to force the non-temporal store path even on MOVDIR64B-capable CPUs.
- Add nontemporal.hpp: non-temporal store helpers, MOVDIR64B 64-byte copy support, the direct-doorbell helper, and 64-byte aligned flat packet buffers, plus unified packet write helpers for AQL submission (writePacketToRingBuffer, ringQueueDoorbell).
- Align barrier_packet_ and barrier_value_packet_ to 64 bytes for MOVDIR64B source alignment.
- Stage metadata-prefetch packets for device-memory ring buffers on the single-dispatch path: MetaDataPreloader::SetPacket publishes full 64-byte segments to reduce scattered write-combining stores. Unify SetMetadata and CaptureMetadata through a shared FillMetadata helper that handles MOVDIR64B, legacy+device-mem (staged NT), and legacy+system-mem (direct write) paths for both kernel-dispatch and barrier metadata packets.
- Capture metadata-prefetch packets during graph build: each captured AQL dispatch also captures its corresponding metadata packet into a host buffer (Command::getMetadataPacket / MetaDataPreloader::CaptureMetadata), kept pointer-parallel with the AQL packet vector. Flatten captured metadata into a contiguous 64-byte-aligned buffer (flatMetadataData / filteredFlatMetadataData) index-aligned with the flat AQL packet buffer.
- Publish metadata to the metadata ring on graph batch-flat dispatch: dispatchAqlPacketBatchFlat accepts the pre-built flat metadata buffer and NT-copies it to the per-queue metadata ring at the corresponding slot indices, handling ring-buffer wrap, before the store fence — ensuring metadata is visible when AQL headers go valid.
- Propagate 64-byte aligned graph flat-packet buffers through the HIP graph dispatch paths.
- Remove the legacy DeviceKernelArgsHDP kernel-argument workaround and its DEBUG_CLR_KERNARG_HDP_FLUSH_WA flag. Also remove the DEBUG_CLR_ENABLE_PREFETCH_METADATA flag (metadata is now gated on whether the metadata ring buffer base is non-null).
- Add support for new qeueu create API in rocprof-sdk so that interception can work

## JIRA ID

JIRA ID : AIRUNTIME-2368

## Submission Checklist

- Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/projects/clr/rocclr/device/rocm/rocvirtual.cpp
+++ b/projects/clr/rocclr/device/rocm/rocvirtual.cpp
@@ -8,6 +8,7 @@
 #include "device/rocm/rocdevice.hpp"
 #include "device/rocm/rocvirtual.hpp"
 #include "device/rocm/rockernel.hpp"
+#include "utils/nontemporal.hpp"
 #include "device/rocm/rocmemory.hpp"
 #include "device/rocm/rocblit.hpp"
 #include "device/rocm/roccounters.hpp"
@@ -39,13 +40,6 @@
 #include <cstdarg>
 #include <mutex>
 
-#if defined(__AVX__)
-#if defined(__MINGW64__)
-#include <intrin.h>
-#else
-#include <immintrin.h>
-#endif
-#endif
 
 /**
  * HSA image object size in bytes (see HSA spec)
@@ -1231,8 +1225,9 @@ bool VirtualGPU::processMemObjects(const amd::Kernel& kernel, const_address para
 }
 
 // ================================================================================================
-void VirtualGPU::SetGpuQueue(hsa_queue_t* queue, void* metadata_ring_buffer) {
+void VirtualGPU::SetGpuQueue(hsa_queue_t* queue) {
   gpu_queue_ = queue;
+
   cached_read_dispatch_id_ = 0;
   // The cached queue-progress state belongs to the previously assigned HW queue. When the HW
   // queue changes (released back to / reacquired from the shared dynamic-queue pool), that state
@@ -1240,17 +1235,27 @@ void VirtualGPU::SetGpuQueue(hsa_queue_t* queue, void* metadata_ring_buffer) {
   // so IsQueueIdle() reports idle for the freshly assigned queue.
   last_write_index_ = kInvalidQueueIndex;
   last_packet_with_signal_index_ = kInvalidQueueIndex;
-  metadata_preloader_.SetQueueBase(metadata_ring_buffer,
-                                   roc_device_.MetadataVersionHeader());
+
+  if (queue) {
+    auto extras = roc_device_.GetQueueExtras(queue);
+    device_mem_ring_buf_ = extras.deviceMemRingBuf;
+    use_movdir64b_ = roc_device_.info().movdir64b_ && device_mem_ring_buf_;
+    doorbell_ptr_ = extras.doorbellPtr;
+    metadata_preloader_.SetQueueBase(extras.metadataRingBuffer,
+                                     roc_device_.MetadataVersionHeader());
+  } else {
+    device_mem_ring_buf_ = false;
+    use_movdir64b_ = false;
+    doorbell_ptr_ = nullptr;
+    metadata_preloader_.SetQueueBase(nullptr, roc_device_.MetadataVersionHeader());
+  }
 }
 
 // ================================================================================================
 void VirtualGPU::AcquireQueueWithPreference() {
   std::scoped_lock lock(execution());
   if (!dedicated_queue_ && gpu_queue_ == nullptr && last_hwq_ != nullptr) {
-    void* md_rb = nullptr;
-    auto* queue = roc_device_.AcquireActiveQueue(priority_, last_hwq_, nullptr, &md_rb);
-    SetGpuQueue(queue, md_rb);
+    SetGpuQueue(roc_device_.AcquireActiveQueue(priority_, last_hwq_));
     last_hwq_ = nullptr;
   }
 }
@@ -1266,9 +1271,7 @@ bool VirtualGPU::ReacquireQueueExcluding(const std::unordered_set<uint64_t>& exc
     roc_device_.releaseQueue(gpu_queue_, std::vector<uint32_t>{}, false, true);
     gpu_queue_ = nullptr;
   }
-  void* md_rb = nullptr;
-  auto* queue = roc_device_.AcquireActiveQueue(priority_, nullptr, &excluded_ids, &md_rb);
-  SetGpuQueue(queue, md_rb);
+  SetGpuQueue(roc_device_.AcquireActiveQueue(priority_, nullptr, &excluded_ids));
   return gpu_queue_ != nullptr;
 }
 
@@ -1277,9 +1280,7 @@ uint64_t VirtualGPU::getQueueID() {
   std::scoped_lock lock(execution());
   // Dedicated queues keep their HW queue, never acquire from pool
   if (!dedicated_queue_ && gpu_queue_ == nullptr) {
-    void* md_rb = nullptr;
-    auto* queue = roc_device_.AcquireActiveQueue(priority_, nullptr, nullptr, &md_rb);
-    SetGpuQueue(queue, md_rb);
+    SetGpuQueue(roc_device_.AcquireActiveQueue(priority_));
   }
   return gpu_queue_->id;
 }
@@ -1370,6 +1371,43 @@ std::string VirtualGPU::AnalyzeAqlQueue() const {
 }
 
 // ================================================================================================
+// Write an AQL packet to the ring buffer with metadata prefetch.
+//
+// MOVDIR64B + device memory path (2 sfences):
+//   metadata MOVDIR64B → sfence → AQL MOVDIR64B → ROCr doorbell sfence → doorbell
+//
+// Legacy + device memory (2 sfences):
+//   AQL body NT → metadata 4x64B NT (staged) → sfence → AQL header → sfence → doorbell
+//   (body+header within each 64B NT copy share one WC buffer fill — no internal sfence)
+//
+// Legacy + system memory (3 sfences):
+//   AQL body NT → metadata direct writes → sfence(body→hdr) → metadata headers →
+//   sfence → AQL header → sfence → doorbell
+template <typename AqlPacket>
+void VirtualGPU::writePacketToRingBuffer(AqlPacket* aql_loc, AqlPacket* packet,
+                                         uint16_t header, uint16_t rest, uint64_t slot_index) {
+  if (use_movdir64b_) {
+    metadata_preloader_.SetMetadata(packet, header, slot_index, true, device_mem_ring_buf_);
+    amd::nontemporalStoreFence();
+    amd::nontemporalWriteAQL(aql_loc, packet, header, rest);
+  } else {
+    amd::nontemporalCopyAQL(aql_loc, packet);
+    metadata_preloader_.SetMetadata(packet, header, slot_index, false, device_mem_ring_buf_);
+    amd::nontemporalStoreFence();
+    packet_store_release(reinterpret_cast<uint32_t*>(aql_loc), header, rest);
+  }
+}
+
+// ================================================================================================
+void VirtualGPU::ringQueueDoorbell(uint64_t index) {
+  if (doorbell_ptr_) {
+    amd::ringDoorbell(doorbell_ptr_, index, use_movdir64b_);
+  } else {
+    Hsa::signal_store_screlease(gpu_queue_->doorbell_signal, index);
+  }
+}
+
+// ================================================================================================
 template <typename AqlPacket>
 bool VirtualGPU::dispatchGenericAqlPacket(AqlPacket* packet, uint16_t header, uint16_t rest,
                                           bool blocking, bool attach_signal, bool cluster_launch) {
@@ -1417,12 +1455,8 @@ bool VirtualGPU::dispatchGenericAqlPacket(AqlPacket* packet, uint16_t header, ui
   TrackQueueProgress(*packet, index);
 
   AqlPacket* aql_loc = &((AqlPacket*)(gpu_queue_->base_address))[index & queueMask];
-  *aql_loc = *packet;
+  writePacketToRingBuffer(aql_loc, packet, header, rest, index & queueMask);
 
-  metadata_preloader_.Set(packet, header, index & queueMask);
-  if (header != 0) {
-    packet_store_release(reinterpret_cast<uint32_t*>(aql_loc), header, rest);
-  }
   if (IsLogEnabled(amd::LOG_DETAIL_DEBUG, amd::LOG_AQL)) {
     if (dev().settings().ext_dispatch_packet_) {
       logAqlDispatchPacketExtended(
@@ -1442,7 +1476,7 @@ bool VirtualGPU::dispatchGenericAqlPacket(AqlPacket* packet, uint16_t header, ui
   bool ring_doorbell = IS_LINUX || dev().IsPm4Emulation() || blocking ||
                        (skippedDispatches_ >= skip_limit) || ring_for_non_profiler_signal;
   if (ring_doorbell) {
-    Hsa::signal_store_screlease(gpu_queue_->doorbell_signal, index);
+    ringQueueDoorbell(index);
     skippedDispatches_ = 0;
   } else {
     ++skippedDispatches_;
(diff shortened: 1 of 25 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
```diff
@@ -610,7 +610,7 @@ void AqlQueue::AllocRegisteredRingBuffer(uint32_t queue_size_pkts) {
     // that select this path are responsible for the required packet-store ordering.
     ring_buf_ = agent_->coarsegrain_allocator()(
         ring_buf_alloc_bytes_ + ring_buf_metadata_alloc_bytes_,
-        core::MemoryRegion::AllocateExecutable);
+        core::MemoryRegion::AllocateExecutable | core::MemoryRegion::AllocateUncached);
```
cfreeamd, line 621: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]**

The pre-existing `AllocateExecutable` flag on a buffer that holds AQL packet data (not code) is worth a comment — is it used to get write-combining memory attributes, or something else? The new `AllocateUncached` addition makes sense for device-resident ring buffers where CPU writes need to bypass the CPU cache to be immediately visible to the GPU, but the combination `AllocateExecutable | AllocateUncached` isn't obviously correct without understanding what each flag actually controls in the coarsegrain allocator path.

cfreeamd (commented):

**ROCr-only review** (`projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp`). The CLR changes are out of scope.

The one-line change adds `AllocateUncached` to the ring buffer allocation on the device-resident path. This is the correct behavior: when the ring buffer lives in device-local (coarsegrain) memory, the GPU reads AQL packets from it directly; making it uncached from the CPU side ensures CPU writes are immediately visible to the GPU without an explicit cache flush. The approach is sound.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
saleelk: Valid, this uncached used to be present earlier. I removed it but looks like doing that hangs on gfx1250, I do not understand the exact reason for it yet as a lot of things are in flux there, I would certainly pull back and leave detailed comments once I am there.

cfreeamd (approved):
LGTM (only ROCr reviewed)

jlgreathouse:
> 🤖 AI-assisted pre-review (for @cfreeamd)
> 
> **ROCr-only review** (`projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp`). The CLR changes are out of scope.
> 
> The one-line change adds `AllocateUncached` to the ring buffer allocation on the device-resident path. This is the correct behavior: when the ring buffer lives in device-local (coarsegrain) memory, the GPU reads AQL packets from it directly; making it uncached from the CPU side ensures CPU writes are immediately visible to the GPU without an explicit cache flush. The approach is sound.

I do not believe this automated analysis is correct. AllocateUncached causes ROCr to ask for KFD to use uncached memory on the allocation through hsaKmtAllocMemory, which  https://github.com/AMD-ROCm-Internal/rocm-systems/blob/4560180bd3ea1a76dd1c2d06c16171cfe4f2e049/projects/rocr-runtime/runtime/hsa-runtime/core/driver/kfd/amd_kfd_driver.cpp#L268.

This eventually causes KMT to request KFD_IOC_ALLOC_MEM_FLAGS_UNCACHED from AMDKFD_IOC_ALLOC_MEMORY_OF_GPU ioctl, https://github.com/AMD-ROCm-Internal/rocm-systems/blob/4560180bd3ea1a76dd1c2d06c16171cfe4f2e049/projects/rocr-runtime/libhsakmt/src/fmm.c#L1234

This causes KFD to request AMDGPU_GEM_CREATE_UNCACHED https://github.com/ROCm/amdgpu/blob/8fa1981194836936acd57c4cf4b3fd694af47cc3/drivers/gpu/drm/amd/amdgpu/amdgpu_amdkfd_gpuvm.c#L1877

This causes the page to be marked uncached _in the GPU page tables_, not uncached to the CPU.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
jlgreathouse: If you're adding ca comment about this, the reason the ring buffer needs to be marked as executable is because the CP hardware requests execute permissions for packet buffer accesses. This is because the hardware team considered packets to be an execution language, and they want to be able to ensure we can't cause the CP to start "executing" packets from non-validated regions of memory.

chrispaquot (approved):

bwelton (commented):
I reviewed the rocprofiler-sdk impact of this change, focusing on queue interception, kernel timestamps, counter collection, HIP graph launches, and late attach.

My overall conclusion is that the normal profiling paths should work under the following conditions:

- `DEBUG_CLR_DIRECT_DOORBELL=0`, so doorbells continue through `hsa_signal_store_screlease`.
- CLR creates its current single compute-queue descriptor.
- The queue uses normal priority and does not require a CU mask to be preserved through profiler interposition.
- Late attach is excluded for now.

Under those conditions:

- Kernel-dispatch tracing should continue to use inline queue interposition.
- Kernels launched through HIP graphs are normal AQL kernel dispatches and should use the same inline interception and timestamp path.
- Kernel timestamps should remain correct because rocprofiler-sdk replaces or borrows the kernel completion signal and obtains the hardware start/end interval through `hsa_amd_profiling_get_dispatch_time`.
- Counter collection and other services that insert AQL packets can fall back to the ROCr `InterceptQueue` path.
- Losing the device-resident queue optimization while counter collection is active is acceptable for the initial implementation.
- Counter records should retain the kernel's hardware start/end timestamps; those timestamps are independent of the surrounding counter start/read/stop packet interval.

I left inline comments on the specific points below (direct doorbell profiling support, priority/CU-mask preservation, a suggested log for the system-memory fallback, partial batch handling, late-attach descriptor validation, non-compute queue registration, and the four `host_functions` test changes).

One more note that isn't tied to a line changed by this PR: `ROCPROFILER_BUFFER_TRACING_HIP_GRAPH` forces non-inline interception in `registration.cpp:1301` (`non_queue_interposition_contexts`), but HIP graph kernel dispatches are normal kernel-dispatch packets and don't inherently need the legacy `InterceptQueue` -- they're compatible with inline interposition and the normal timestamp path. That restriction exists only because `kernel_dispatch_count` accounting for this domain is currently implemented in the legacy `Queue::WriteInterceptor` (the `graph_launch_active` fast path in `hsa/queue.cpp`), and the inline interceptor in `hsa/queue_interposition.cpp` has no equivalent counting yet. This could be fixed as a follow-up by mirroring that fast path in `write_interceptor()` in `queue_interposition.cpp` (inspect `hip::graph::current_launch_state()`, count standard/extended kernel-dispatch packets before the `_contexts.empty()` early return, add to `dispatch_count`, forward unchanged when no kernel-tracing consumer exists) and then removing `ROCPROFILER_BUFFER_TRACING_HIP_GRAPH` from `non_queue_interposition_contexts`, so graph-only tracing can retain device-resident queues.

projects/clr/rocclr/device/rocm/rocvirtual.cpp
```diff
@@ -1369,6 +1370,43 @@ std::string VirtualGPU::AnalyzeAqlQueue() const {
...
+    metadata_preloader_.SetMetadata(packet, header, slot_index, false, device_mem_ring_buf_);
+    amd::nontemporalStoreFence();
+    packet_store_release(reinterpret_cast<uint32_t*>(aql_loc), header, rest);
+  }
+}
+
+// ================================================================================================
+void VirtualGPU::ringQueueDoorbell(uint64_t index) {
+  if (doorbell_ptr_) {
+    amd::ringDoorbell(doorbell_ptr_, index, use_movdir64b_);
```
bwelton, line 1404: Direct doorbell writes bypass the HSA signal-store path used by rocprofiler-sdk queue interception (both inline interposition and the ROCr `InterceptQueue` path key off `hsa_signal_store_*`). Can we document that `DEBUG_CLR_DIRECT_DOORBELL=1` is not supported with profiling, and that profiling requires `DEBUG_CLR_DIRECT_DOORBELL=0`?

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
```diff
@@ -112,6 +112,112 @@ destroy_queue(hsa_queue_t* hsa_queue)
...
+                const uint32_t ring_packets   = static_cast<uint32_t>(
+                    descs[desc_idx].queue_size_bytes / sizeof(hsa_kernel_dispatch_packet_t));
+
+                // Release the plain queue that hsa_amd_queue_create just handed us; the
+                // InterceptQueue built below replaces it. get_core_table() holds the real ROCr
+                // functions (saved before the interception wrappers were installed).
+                controller->get_core_table().hsa_queue_destroy_fn(queue);
+
+                hsa_queue_t* intercept_queue = nullptr;
+                new_queue                    = std::make_unique<Queue>(
```
bwelton, line 177: Falling back to a system-memory `InterceptQueue` is reasonable for counter collection and other services that require packet insertion. However, this replacement `Queue` constructor doesn't pass through `desc.priority` or `compute.cu_mask`.

This changes execution semantics beyond just losing the device-ring optimization (e.g. a CU-partitioned or high-priority stream could run differently while profiled). Can these attributes be applied to the replacement queue via `SetPriority`/`SetCUMasking`? If that's deferred, could we document it as a known limitation?

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
```diff
@@ -112,6 +112,112 @@ destroy_queue(hsa_queue_t* hsa_queue)
...
+                // A queue created by hsa_amd_queue_create is a plain queue that cannot carry a
+                // ROCr packet interceptor (hsa_amd_queue_intercept_register only accepts an
+                // InterceptQueue). In non-inline mode we therefore discard the plain queue and
+                // build a ROCr InterceptQueue that mirrors the descriptor's compute parameters,
+                // then hand that back to the caller in place of the original queue.
+                //
+                // NOTE: the InterceptQueue is created via the classic hsa_queue_create path, so
+                // the descriptor's device-memory ring-buffer flag is not honored while a
+                // non-inline profiling context (e.g. --sys-trace / HIP graph tracing) is active;
+                // the queue falls back to a system-memory ring.
```
bwelton, line 162: This fallback is acceptable for the initial counter-collection implementation. Can we emit a `ROCP_WARNING` here when a requested device-memory queue is replaced with a system-memory `InterceptQueue`? Having this in the SDK logs will make it clear that profiling intentionally disabled the device-resident queue optimization, which will help with future debugging.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
```diff
@@ -112,6 +112,112 @@ destroy_queue(hsa_queue_t* hsa_queue)
...
 
+#if defined(HSA_AMD_EXT_API_TABLE_STEP_VERSION) && HSA_AMD_EXT_API_TABLE_STEP_VERSION >= 0x10
+hsa_status_t
+create_amd_queue(hsa_agent_t                  agent,
+                 hsa_amd_queue_create_desc_t* descs,
+                 uint32_t                     num_descs)
+{
+    auto* controller = CHECK_NOTNULL(get_queue_controller());
+    auto  status     = controller->get_ext_table().hsa_amd_queue_create_fn(agent, descs, num_descs);
+    if(status != HSA_STATUS_SUCCESS) return status;
```
bwelton, line 123: `hsa_amd_queue_create` permits partial batch success: it can return the first error while leaving successful `descs[i].queue` entries valid. Returning immediately here leaves those successful queues unregistered with rocprofiler-sdk (no `Queue`, serializer entry, or `QueueState`). Could this process every non-null `descs[i].queue` and return the original status afterward, rather than bailing out on the first error?

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
```diff
@@ -112,6 +112,112 @@ destroy_queue(hsa_queue_t* hsa_queue)
...
+                    controller->get_core_table(),
+                    controller->get_ext_table(),
+                    &intercept_queue);
+
+                queue                 = intercept_queue;
+                descs[desc_idx].queue = intercept_queue;
+            }
+            else
+            {
+                // Non-compute engines (e.g. SDMA) do not dispatch kernels; keep the plain queue
```
bwelton, line 196: This comment says non-compute queues are registered without packet interception, but `add_queue()` unconditionally calls `queue_interposition::create_queue_state(id)`, which assumes fixed-size 64-byte AQL packets and a packet-count queue size. SDMA queue sizes are byte counts and use a different packet format. Could non-compute queues be excluded from inline `QueueState` registration (and the kernel-dispatch serializer) here?

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-attach/queue_registration.cpp
```diff
@@ -158,46 +210,75 @@ create_queue(hsa_agent_t        agent,
...
+// queue that cannot carry a ROCr packet interceptor, so for compute queues we build an
+// InterceptQueue that mirrors the descriptor's compute parameters instead (the InterceptQueue is
+// what allows kernel-dispatch tracing under attachment). Non-compute queues (e.g. SDMA) do not
+// dispatch kernels and are forwarded to the real implementation unchanged.
+//
+// NOTE: the InterceptQueue is created via the classic hsa_queue_create path, so the descriptor's
+// device-memory ring-buffer flag is not honored while the queue is being profiled; the queue falls
+// back to a system-memory ring.
+hsa_status_t
+create_amd_queue(hsa_agent_t agent, hsa_amd_queue_create_desc_t* descs, uint32_t num_descs)
```
bwelton, line 231: This wrapper bypasses the real `hsa_amd_queue_create` implementation for compute descriptors, so it also bypasses the runtime's descriptor validation (version, reserved fields, size, priority, CU-mask constraints) and partial-batch-success semantics.

Since late attach is excluded from the currently-supported configuration for this PR, could this limitation be documented explicitly? It may be safer to reject this path as unsupported until a descriptor-aware attach implementation lands, rather than return `HSA_STATUS_SUCCESS` with different API semantics than the real runtime.

projects/rocprofiler-sdk/tests/async-copy-tracing/validate.py
```diff
@@ -113,7 +113,7 @@ def test_data_structure(input_data):
     node_exists("names", sdk_data["callback_records"])
     node_exists("code_objects", sdk_data["callback_records"])
     node_exists("kernel_symbols", sdk_data["callback_records"])
-    node_exists("host_functions", sdk_data["callback_records"])
+    node_exists("host_functions", sdk_data["callback_records"], 0)
```
bwelton, line 116: This (and the equivalent changes in `hip-graph-tracing`, `kernel-tracing`, and `scratch-memory-tracing` validate.py) relaxes `host_functions` from "must contain records" to "key may be empty," which weakens existing host-kernel-symbol registration coverage. This doesn't appear related to the device-resident queue ring change. Did this PR expose a host-function registration regression? If not, could these four changes be reverted? If the records are no longer guaranteed, that behavior change would need a separate explanation and dedicated test coverage.

projects/rocprofiler-sdk/tests/hip-graph-tracing/validate.py
```diff
@@ -56,7 +56,7 @@ def test_data_structure(input_data):
     node_exists("names", sdk_data["callback_records"])
     node_exists("code_objects", sdk_data["callback_records"])
     node_exists("kernel_symbols", sdk_data["callback_records"])
-    node_exists("host_functions", sdk_data["callback_records"])
+    node_exists("host_functions", sdk_data["callback_records"], 0)
```
bwelton, line 59: Same `host_functions` relaxation as in `async-copy-tracing/validate.py` -- see that comment.

projects/rocprofiler-sdk/tests/kernel-tracing/validate.py
```diff
@@ -64,7 +64,7 @@ def test_data_structure(input_data):
     node_exists("names", sdk_data["callback_records"])
     node_exists("code_objects", sdk_data["callback_records"])
     node_exists("kernel_symbols", sdk_data["callback_records"])
-    node_exists("host_functions", sdk_data["callback_records"])
+    node_exists("host_functions", sdk_data["callback_records"], 0)
```
bwelton, line 67: Same `host_functions` relaxation as in `async-copy-tracing/validate.py` -- see that comment.

projects/rocprofiler-sdk/tests/scratch-memory-tracing/validate.py
```diff
@@ -60,7 +60,7 @@ def test_data_structure(input_data):
     node_exists("names", sdk_data["callback_records"])
     node_exists("code_objects", sdk_data["callback_records"])
     node_exists("kernel_symbols", sdk_data["callback_records"])
-    node_exists("host_functions", sdk_data["callback_records"])
+    node_exists("host_functions", sdk_data["callback_records"], 0)
```
bwelton, line 63: Same `host_functions` relaxation as in `async-copy-tracing/validate.py` -- see that comment.

projects/clr/rocclr/device/rocm/rocvirtual.cpp
saleelk: Correct I am aware, its experimental only, I want the initial set of changes in for test purposes, this would be remove later.

projects/rocprofiler-sdk/tests/async-copy-tracing/validate.py
saleelk: I see. thats easy fix, ACK

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-attach/queue_registration.cpp
saleelk: Sure, let me document that. Overall I would have to revisit all this and I want to separate out queue writing semantics and allow profiler to support true device side queues but in a separate change.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
saleelk: Ack

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
saleelk: thats the behavior I wanted, ie we return error if anything fails.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
saleelk: Let me document it, I have summed up below

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
saleelk: sure

bwelton (approved):

lumachad:
@saleelk 
FYI, this seems to have introduced a gfx942 regression. See AIRUNTIME-2569.

PR #7732, opened 2026-06-24
[RCCL] test: allow NET transport MPI tests to run with multiple ranks per node

atulkulk:
## Motivation
NET transport MPI tests required exactly one rank per node, preventing multi-rank-per-node runs.

## Technical Details
Relax the requirement to >=2 ranks across >=2 nodes and pair each rank with the same-local-rank peer on a partner node to keep transfers cross-node.

## JIRA ID
Resolves AICOMRCCL-1387.

## Test Plan
Ran NetMPITests with multiple ranks per node across 2 nodes.

## Test Result
Tests run and pass with multiple ranks per node while staying cross-node; single-rank-per-node still works.

### DIFF
--- a/projects/rccl/test/transport/NetMPITests.cpp
+++ b/projects/rccl/test/transport/NetMPITests.cpp
@@ -153,8 +153,8 @@ namespace
 inline constexpr size_t kTestBufferSize = 16384;
 
 // NET transport test requirements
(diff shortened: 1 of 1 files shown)


alex-breslow-amd (approved):
LGTM

PR #7734, opened 2026-06-24
[RCCL] rccl_ptr: enable global dwordx4 builtins on gfx1250

atulkulk:
## Motivation
On gfx1250, 128-bit global load/store used the faulting fallback path instead of the builtin.

## Technical Details
Add gfx1250 to the RCCL_HAVE_GLOBAL_DWORDX4_BUILTINS gate so dwordx4 global access uses the builtin path.

## JIRA ID
Resolves AICOMRCCL-1389.

## Test Plan
Built and exercised 128-bit global load/store paths on gfx1250.

## Test Result
gfx1250 now uses the dwordx4 builtin path with no fault; other arches unchanged.


atulkulk:
Closing this PR which is superseded by another branch.

PR #7739, opened 2026-06-24
test(rocprofiler-systems):  Test add rocprof-sys-sample kitchen-sink CLI pa…

anujshuk-amd:
## Motivation

This adds dedicated CLI argument-parser regression coverage for rocprof-sys-sample, closing a gap where only rocprof-sys-run had a "kitchen-sink" parser test. It guards against parser regressions (argument arity bugs, short-flag bundling, = vs space forms, and one new flag silently breaking another) when many options are combined in a single invocation.

## Technical Details

Add a TestRocprofilerSystemsSample class mirroring the rocprof-sys-run
kitchen-sink parser regression test: a broad combined-flag invocation, a
negative test proving run-only flags (-S) are rejected, and a boundary test
ensuring a greedy multi-value option stops at the `--` separator. Also expand
the run kitchen-sink coverage and hoist the shutil import to module scope.

## JIRA ID

AIROCVAL-115

## Test Plan

Ran tests locally and will check in CI/CD.

## Test Result

It passes on local mahcines.

## Submission Checklist


projects/rocprofiler-systems/tests/pytest/test_binaries.py
```diff
@@ -961,4 +1087,65 @@ def test_args(self, test_output_dir):
...
+        ``-E`` takes a list of values, so the parser keeps reading words until it
+        hits something that isn't a value. We put ``-E`` right before ``--`` to
+        make sure the parser stops at ``--`` instead of greedily swallowing it.
+        If it worked correctly, everything after ``--`` is treated as the program
+        to launch (``sleep 1``) and the run succeeds with no parser error.
+        """
+        sleep_cmd = shutil.which("sleep")
+        if not sleep_cmd:
+            pytest.skip("sleep command not found")
+
```
kcossett-amd, line 1071: Since `sleep_cmd` is not used three times throughout this file, how about we add module level function to get it and skip test:

```
def get_sleep_cmd():
    sleep_cmd = shutil.which("sleep")
    if not sleep_cmd:
        pytest.skip("sleep command not found")
    return sleep_cmd
```

You will have to move the `import shutil` to the module level.

projects/rocprofiler-systems/tests/pytest/test_binaries.py
```diff
@@ -921,35 +920,162 @@ def test_args(self, test_output_dir):
...
+            "baseline",
+            target=self.target,
+            run_args=args,
+            fail_on_not_found=True,
+        )
+        self.assert_regex(result)
+
+
+@pytest.mark.class_name("rocprofiler-systems-sample")
+class TestRocprofilerSystemsSample(RocprofsysTest):
```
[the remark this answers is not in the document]
kcossett-amd: Agreed, this is required when the target is a `rocprof-sys` binary (because you go through the Baseline runner)

projects/rocprofiler-systems/tests/pytest/test_binaries.py
```diff
@@ -921,35 +920,162 @@ def test_args(self, test_output_dir):
...
+
+        config_dir = test_output_dir / "config"
+        config_dir.mkdir(parents=True, exist_ok=True)
+        empty_cfg = config_dir / "empty.cfg"
+        empty_cfg.write_text("#\n# empty config file\n#\n")
+
+        tmpdir = (test_output_dir / "tmpdir").resolve()
+        tmpdir.mkdir(parents=True, exist_ok=True)
+
+        args = [
```
kcossett-amd: The args here are almost identical to the args from the equivalent `rocprof-sys-run` test. To avoid repetition, can we have a "base" arg list, and then you reuse that throughout the two tests?

projects/rocprofiler-systems/tests/pytest/test_binaries.py
```diff
@@ -959,3 +993,98 @@ def test_args(self, test_output_dir):
...
+            sleep_cmd=sleep_cmd,
+        )
+
+        result = self.run_test(
+            "baseline",
+            target=self.target,
+            run_args=args,
+            fail_on_not_found=True,
+        )
+        # exit 0 is enforced by run_test; also assert no parser error was printed.
```
kcossett-amd:
```suggestion
```
Redundant

kcossett-amd (approved):
Looks good. Just one small nit

PR #7744, opened 2026-06-24
feat(workflows): [rocprofiler-systems] Improve CI workflow reporting

marantic-amd:
## Motivation

As it stands now, the output of the CI workflow stages that is represented in the GitHub Actions is really hard to follow and look for specific failures, or just a single stage results. This PR breaks down everything into fine groupings, separating Configure, Build and Test as 3 stages of a run, each with it's own report. Also targets the repetition that existed between workflow files them self with a loot of duplicated code and core logic.

## Technical Details

- Consolidate four distro workflows into one matrix-driven pair: `rocprofiler-systems-{debian,redhat,ubuntu-jammy,ubuntu-noble}.yml` (~1,580 lines, mostly duplicated steps) are replaced by a single reusable `rocprofiler-systems-build-group.yml `called once per group from a thin `rocprofiler-systems-build.yml` orchestrator. Matrix data lives in one JSON file (`ci-build-matrix.json`) filtered per group by `generate-ci-build-matrix.py`.
- Custom **JUnit PR summary** (`summarize-junit-results.py`): one updating Markdown comment per push, grouped by OS and including sanitizer status, replacing the generic per-matrix-job comments.
- Split `run-ci.py's` dashboard invocation into discrete generate/configure/build/test stages, enabling separate GitHub Actions steps while keeping one CDash build entry (ctest_start(APPEND) across stages).
- Colored build-failure diagnostics:  the compiler output only exists inline in Testing/<TAG>/Build.xml, ANSI-escaped by CTest as literal [NON-XML-CHAR-...] tokens. `run-ci.py` now recovers and un-escapes that output so real color renders in the Actions log, and centralizes the "N error(s)/M warning(s)" annotation that three separate workflow steps previously computed independently (and, on build failure, silently never emitted).
- Skip-reason reporting: surfaces why a test was skipped (CDash's Completion Status measurement, falling back to pytest's captured output) instead of CTest's uninformative (Skipped).
- Sanitizer retry fix: --repeat-until-pass 1 was passed at the generate stage, where run-ci.py silently ignores it; moved to the test stage where retries are actually applied.
- Local CI contract test harness (`test_ci_workflow_contracts.py`): validates workflow YAML structure and exercises `run-ci.py's` stage behavior against fake ctest/cmake binaries.
- Pinned action SHAs for actions/checkout, actions/github-script, and EnricoMi/publish-unit-test-result-action per CodeQL/reviewer feedback.

## JIRA ID

JIRA ID : AIPROFSYST-391
JIRA ID: AIPROFSYST-627

## Test Plan

- python3 `projects/rocprofiler-systems/scripts/test_ci_workflow_contracts.py --strict-actionlint off` as a local contract harness covering workflow structure, matrix filtering, and `run-ci.py` stage behavior (colored-log recovery, build annotation, skip-reason reporting), fake-tool-driven so it runs without ROCm/CI infrastructure.
- Reproduced CTest's CTEST_USE_LAUNCHERS failure-log behavior in an isolated minimal repro to verify the `Build.xml` recovery and annotation fix before relying on live CI.
- Exercised the colored-failure path end-to-end via PR #8433 (an intentionally-broken test PR based on this branch).

## Test Result

- All local contract checks pass; `black --check` / `py_compile` clean on all modified Python scripts.

## Submission Checklist


mradosav-amd:
image

Can we split to multiple groups?

```

●                         ┌─► [RHEL (with/without sys-deps)]    ─────┐
                          │                                           │
                          ├─► [Ubuntu 22.04 (with/without sys-deps)] ┤
                          │                                           │
  [Matrix setup] ────────►├─► [Ubuntu 24.04 (with/without sys-deps)] ┼──► [summary report]
                          │                                           │
                          ├─► [Debian (with/without sys-deps)]  ─────┤
                          │                                           │
                          ├─► [Code coverage]               ─────────┤
                          │                                           │
                          └─► [Sanitizers]                  ─────────┘
```

marantic-amd:
> image
> Can we split to multiple groups?
> 
> ```
> 
> ●                         ┌─► [RHEL (with/without sys-deps)]    ─────┐
>                           │                                           │
>                           ├─► [Ubuntu 22.04 (with/without sys-deps)] ┤
>                           │                                           │
>   [Matrix setup] ────────►├─► [Ubuntu 24.04 (with/without sys-deps)] ┼──► [summary report]
>                           │                                           │
>                           ├─► [Debian (with/without sys-deps)]  ─────┤
>                           │                                           │
>                           ├─► [Code coverage]               ─────────┤
>                           │                                           │
>                           └─► [Sanitizers]                  ─────────┘
> ```

We can. I hoped to do that in a separate PR, but I can also do it as a part of this one, for sure. What I can do is to add a small  top-level orchestrator with a matrix setup job, then call one reusable rocprofiler-systems build-group workflow for each group: RHEL, Ubuntu 22.04, Ubuntu 24.04, and Debian. The same pattern can then feed coverage, sanitizers, and the final summary job. That should give the GitHub Actions graph the grouped shape shown here while keeping the build/test implementation centralized as intended.

.github/workflows/rocprofiler-systems-ubuntu-noble-sanitizers.yml
```diff
@@ -219,11 +181,69 @@ jobs:
...
+          printf '### CDash Results\n[View build on CDash](%s)\n' "${URL}" >> "${GITHUB_STEP_SUMMARY}"
+          echo "::notice title=CDash Results::${URL}"
+
+      - name: Test
+        timeout-minutes: 180
+        shell: bash
+        working-directory: projects/rocprofiler-systems/
+        run: |
+          python3 ./scripts/run-ci.py --stage test \
+            --name "${{ github.repository_owner }}-${{ github.ref_name }}-ubuntu-noble-${{ matrix.compiler }}-${{ matrix.sanitizer-type }}-sanitizer" \
```
sputhala-amd, line 229: If sanitizer builds needs to retry only once then we need to pass `--repeat-until-pass 1` and `--repeat-after-timeout 1` here in the test stage.

.github/workflows/rocprofiler-systems-ubuntu-noble-sanitizers.yml
```diff
@@ -174,28 +137,27 @@ jobs:
...
+          echo "Compiler: $(${{ matrix.compiler }} --version | head -n 1)"
           echo "Sanitizer type: ${{ matrix.sanitizer-type }}"
           echo "Sanitizer CMake flags: ${SANITIZER_CMAKE_FLAGS}"
-          TAG="-${{ matrix.sanitizer-type }}-sanitizer"
-          python3 ./scripts/run-ci.py -B build \
-            --name ${{ github.repository_owner }}-${{ github.ref_name }}-ubuntu-noble-${{ matrix.compiler }}${TAG} \
+          python3 ./scripts/run-ci.py --stage generate \
+            --name "${{ github.repository_owner }}-${{ github.ref_name }}-ubuntu-noble-${{ matrix.compiler }}-${{ matrix.sanitizer-type }}-sanitizer" \
             --build-jobs $(nproc) \
             --repeat-until-pass 1 \
```
sputhala-amd: It seems this ` --repeat-until-pass 1` is ignored.
run-ci.py applies `--repeat` settings at the test stage, so the override must be passed there.

sputhala-amd (commented):
Does this provide one updating comment per push in the PR rather than one per matrix job?
Is there a sample comment?
Currently the PR details are empty so I guess you may update the details there.

marantic-amd:
> Does this provide one updating comment per push in the PR rather than one per matrix job? Is there a sample comment? Currently the PR details are empty so I guess you may update the details there.

THere were a lot of trial-and-error runs here, so the first comments you see were by matrix job which was really overwhelming and not really any improvement. So, short answer, YES, there is a single summary comment that is updated after each rerun of the systems workflows. 
image
 
 The comment provides the summary per OS and for sanitizer runs. It references a commit in question, so you are always sure that the report is being made against the last run

.github/workflows/rocprofiler-systems-ubuntu-noble-sanitizers.yml
marantic-amd: I managed to miss that one. Thanks.

.github/workflows/rocprofiler-systems-continuous-integration.yml
```diff
@@ -200,6 +210,80 @@
...
+        run: |
+          PATH=${{ env.ROCM_PATH }}/bin:${{ env.ROCM_PATH }}/llvm/bin:$PATH \
+          LD_LIBRARY_PATH=${{ env.ROCM_PATH }}/lib:${{ env.ROCM_PATH }}/llvm/lib:$LD_LIBRARY_PATH \
+          python3 ./scripts/run-ci.py --stage test \
+            --name "${{ github.repository }}-${{ github.ref_name }}-ubuntu-${{ matrix.system.os-release }}-${{ matrix.system.gpu }}" \
+            -B build
+
+      - name: Publish test results
+        if: always()
+        uses: EnricoMi/publish-unit-test-result-action@v2
```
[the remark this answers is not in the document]
dgaliffiAMD: Yes, this is a requirement. Please use a pinned SHA. Something like this - `EnricoMi/publish-unit-test-result-action@d0a4676d0e0b938bc201470d88276b7c74c712b3 // v.2.24.0`.

dgaliffiAMD:
> Does this provide one updating comment per push in the PR rather than one per matrix job? Is there a sample comment? Currently the PR details are empty so I guess you may update the details there.

https://github.com/ROCm/rocm-systems/pull/8433 - Creating a test PR here.

.github/workflows/rocprofiler-systems-build.yml
```diff
@@ -0,0 +1,307 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      checks: write
+      contents: read
+      issues: write
+      pull-requests: write
+    steps:
+      - name: Checkout summary script
+        uses: actions/checkout@v4
```
dgaliffiAMD: The `checkout@v4` action is using a version of Node.JS that will soon be deprecated.

```suggestion
        uses: actions/checkout@v6
```
... to match the others

```suggestion
        uses: actions/checkout@v7
```
... to use the latest (but should also update the others).

.github/workflows/rocprofiler-systems-build.yml
```diff
@@ -0,0 +1,307 @@
...
+          python3 projects/rocprofiler-systems/scripts/summarize-junit-results.py \
+            --mode sanitizer \
+            --junit-glob 'sanitizer-junit-results/**/*.xml' \
+            --status-glob 'sanitizer-status/**/*.json' \
+            --output sanitizer-summary.md
+          cat sanitizer-summary.md >> "${GITHUB_STEP_SUMMARY}"
+
+      - name: Update PR CI result summary
+        if: github.event_name == 'pull_request'
+        uses: actions/github-script@v7
```
dgaliffiAMD: The `github-script@v7` action is using a version of Node.js that will soon be deprecated. We should be using `v8` or `v9`.

```suggestion
        uses: actions/github-script@v9
```

dgaliffiAMD:
> > Does this provide one updating comment per push in the PR rather than one per matrix job? Is there a sample comment? Currently the PR details are empty so I guess you may update the details there.
> 
> #8433 - Creating a test PR here.

Hi @marantic-amd, 

Can you help me with this one. I have my test which passes the configuration stage, but fails the build phase (expected), and this is the message that I get in the "conversation" - https://github.com/ROCm/rocm-systems/pull/8433#issuecomment-4939629901. It (correctly) states that 0 tests have run / passed but does not indicate that that is because the build step fails. I can click one of the actions (https://github.com/ROCm/rocm-systems/actions/runs/29121703353/job/86458188038?pr=8433) but I can't to find the logs. Neither from the passing configuration, nor the failed build. Where should I look to see why my test build fails.

marantic-amd:
> > > Does this provide one updating comment per push in the PR rather than one per matrix job? Is there a sample comment? Currently the PR details are empty so I guess you may update the details there.
> > 
> > 
> > #8433 - Creating a test PR here.
> 
> Hi @marantic-amd,
> 
> Can you help me with this one. I have my test which passes the configuration stage, but fails the build phase (expected), and this is the message that I get in the "conversation" - [#8433 (comment)](https://github.com/ROCm/rocm-systems/pull/8433#issuecomment-4939629901). It (correctly) states that 0 tests have run / passed but does not indicate that that is because the build step fails. I can click one of the actions (https://github.com/ROCm/rocm-systems/actions/runs/29121703353/job/86458188038?pr=8433) but I can't to find the logs. Neither from the passing configuration, nor the failed build. Where should I look to see why my test build fails.

The idea I had was to use the log files generated by `CTest`, since they keep the colour formatted output and don't insert so much noise as the XML one. The underlying issue was that for the sake of `per-rule CDash error/warning report` and for better optimization of parallel logs being generated we are using `CTest launchers` for the build step, and the caveat of using those is that the log file is only generated on successfully executed build step, hence why you could not see any output. I fixed that with the latest commit, and I think now we have everything in place. I checked on your test PR #8433 and to me it looks like it's behaving as intended.

sputhala-amd:
> > Does this provide one updating comment per push in the PR rather than one per matrix job? Is there a sample comment? Currently the PR details are empty so I guess you may update the details there.
> 
> THere were a lot of trial-and-error runs here, so the first comments you see were by matrix job which was really overwhelming and not really any improvement. So, short answer, YES, there is a single summary comment that is updated after each rerun of the systems workflows. image
> 
> The comment provides the summary per OS and for sanitizer runs. It references a commit in question, so you are always sure that the report is being made against the last run

Thanks @marantic-amd. I really liked this summary output. 
Does the `runs` in the table here mean the actual number of workflows which ran for the OS as per the matrix?

dgaliffiAMD:
> > > > Does this provide one updating comment per push in the PR rather than one per matrix job? Is there a sample comment? Currently the PR details are empty so I guess you may update the details there.
> > > 
> > > 
> > > #8433 - Creating a test PR here.
> > 
> > 
> > Hi @marantic-amd,
> > Can you help me with this one. I have my test which passes the configuration stage, but fails the build phase (expected), and this is the message that I get in the "conversation" - [#8433 (comment)](https://github.com/ROCm/rocm-systems/pull/8433#issuecomment-4939629901). It (correctly) states that 0 tests have run / passed but does not indicate that that is because the build step fails. I can click one of the actions (https://github.com/ROCm/rocm-systems/actions/runs/29121703353/job/86458188038?pr=8433) but I can't to find the logs. Neither from the passing configuration, nor the failed build. Where should I look to see why my test build fails.
> 
> The idea I had was to use the log files generated by `CTest`, since they keep the colour formatted output and don't insert so much noise as the XML one. The underlying issue was that for the sake of `per-rule CDash error/warning report` and for better optimization of parallel logs being generated we are using `CTest launchers` for the build step, and the caveat of using those is that the log file is only generated on successfully executed build step, hence why you could not see any output. I fixed that with the latest commit, and I think now we have everything in place. I checked on your test PR #8433 and to me it looks like it's behaving as intended.

Yes, that worked. I can now see my failing build logs and failing configuration logs.

.github/workflows/rocprofiler-systems-build-group.yml
```diff
@@ -0,0 +1,699 @@
...
+          RUNNING_PROCS=$(pgrep trace_processor_shell)
+          if [ -n "${RUNNING_PROCS}" ]; then kill -s 9 ${RUNNING_PROCS}; fi
+
+      - &_step-ctest-artifacts
+        name: CTest artifacts
+        if: failure()
+        continue-on-error: true
+        uses: actions/upload-artifact@v7
+        with:
+          name: ctest-${{ github.job }}-${{ strategy.job-index }}-log
```
dgaliffiAMD:
```suggestion
          name: ctest-${{ inputs.group_key }}-${{ github.job }}-${{ strategy.job-index }}-log
```
failure-diagnostic artifacts (`ctest-*`, `data-*`) omit `group_key`, so they collide across the 4 distro calls.

.github/workflows/rocprofiler-systems-build-group.yml
```diff
@@ -0,0 +1,699 @@
...
+          path: |
+            projects/rocprofiler-systems/build/*.log
+
+      - &_step-data-artifacts
+        name: Data artifacts
+        if: failure()
+        continue-on-error: true
+        uses: actions/upload-artifact@v7
+        with:
+          name: data-${{ github.job }}-${{ strategy.job-index }}-files
```
dgaliffiAMD:
```suggestion
          name: data-${{ inputs.group_key }}-${{ github.job }}-${{ strategy.job-index }}-files
```
failure-diagnostic artifacts (ctest-*, data-*) omit group_key, so they collide across the 4 distro calls.

projects/rocprofiler-systems/scripts/summarize-junit-results.py
```diff
@@ -0,0 +1,427 @@
...
+        if artifact.startswith(f"junit-build-{group}-"):
+            return group
+        if artifact.startswith(f"junit-{group}-"):
+            return group
+    if artifact.startswith("junit-system-deps-"):
+        return "system-deps"
+    if artifact.startswith("junit-build-"):
+        return "build"
+    return "other"
+
```
dgaliffiAMD, line 172:
```suggestion
    if "-system-deps-" in artifact:
        return "system-deps"
    for group in ("ubuntu-22.04", "ubuntu-24.04", "debian", "rhel"):
        if artifact.startswith(f"junit-build-{group}-"):
            return group
    if artifact.startswith("junit-build-"):
        return "build"
    return "other"
```
The `system-deps` artifacts are named `junit-build-<distro>-system-deps-<idx>`, so this must be tested before the leading-distro match below (otherwise e.g., `junit-build-rhel-system-deps-0` matches "junit-build-rhel-" first).

projects/rocprofiler-systems/scripts/summarize-junit-results.py
```diff
@@ -0,0 +1,427 @@
...
+def parse_junit(path: Path, mode: str) -> JUnitResult:
+    artifact = artifact_name(path)
+    if mode == "sanitizer":
+        group_key = sanitizer_from_artifact(artifact)
+        group_label = SANITIZER_LABELS.get(group_key, group_key.title())
+    else:
+        group_key = build_group_from_artifact(artifact)
+        group_label = BUILD_GROUP_LABELS.get(group_key, group_key.title())
+
+    root = ET.parse(path).getroot()
```
dgaliffiAMD: `parse_junit` will abort the entire summary if any single artifact's XML is malformed.

`ET.parse(path)` is called here for every downloaded artifact (via the comprehension in `main`), with no guard. A `ParseError` from one bad file propagates out and takes down the whole summary — no PR comment gets posted, and every *other* group's results are lost with it.

This isn't a rare edge case: the `test-results.xml` upload is `if: always()` with `if-no-files-found: ignore`, so a job killed by `timeout-minutes` or OOM still uploads a **partial** `test-results.xml` (ctest writes incrementally and never closed the root element). A 0-byte file from an upload racing the write does the same. So the exact conditions that cause a build/test failure — the thing this summary exists to report — are what produce the XML that blinds it.

Repro: an empty file gives `ParseError: no element found`; a truncated `<testsuites><testsuite tests="12"` gives `unclosed token`. Both currently abort the run.

Suggest catching per-file and skipping with a warning so one bad artifact doesn't sink the rest:

```python
try:
    root = ET.parse(path).getroot()
except ET.ParseError as exc:
    print(f"warning: skipping unparseable JUnit XML {path}: {exc}", file=sys.stderr)
    return None
```

(make `parse_junit` return `Optional[JUnitResult]` and filter `None` in `main`). `read_statuses`'s `json.loads` (~line 369) has the same exposure for the sanitizer status files — worth the same guard. This matches how `run-ci.py` already handles `ET.parse` (`_read_build_xml_failures`, `_collect_skip_reasons`).

I have a fix + regression test for this on a branch if useful:
- `2bb6f1651e2ecf7030cdd94760ac48c90fdc0193` — the guard (parse_junit + read_statuses)
- `3e59a9da3ad56499c042036dbe2faf5d33590708` — a contract-harness test that feeds a good + truncated + empty `test-results.xml` through the summarizer and asserts the bad ones are skipped rather than fatal.

projects/rocprofiler-systems/scripts/test_ci_workflow_contracts.py
```diff
@@ -0,0 +1,1123 @@
...
+
+_FAILURE_SENTINEL = "SENTINEL_BUILD_FAILURE"
+_WARNING_SENTINEL = "SENTINEL_BUILD_WARNING"
+_NON_XML_ESC = "[NON-XML-CHAR-0x1B]"
+
+
+def write_fake_build_xml_failure(bin_dir: Path, binary_dir: Path) -> None:
+    """Fake ctest reproducing a real CTEST_USE_LAUNCHERS build failure:
+    NO LastBuild*.log at all (verified against a live minimal repro — see
+    planning/bugfix-build-failure-colored-log-xml.md), only
```
dgaliffiAMD: Stale comment referencing a planning document. This can be removed.

projects/rocprofiler-systems/scripts/run-ci.py
```diff
@@ -24,29 +55,52 @@ def log(msg, level="info"):
...
+        bar = "=" * 60
+        print(_color(f"\n{bar}\n{title}\n{bar}", "cyan", "bold"), flush=True)
 
 
 def log_group_end():
     if IS_GITHUB_ACTIONS:
         print("::endgroup::", flush=True)
 
 
+def log_step(title):
```
dgaliffiAMD: ruff check: Unused function.

projects/rocprofiler-systems/scripts/test_ci_workflow_contracts.py
```diff
@@ -0,0 +1,1123 @@
...
+        result.append(as_mapping(step, f"job '{job_name}' step {index}"))
+    return result
+
+
+def step_uses(step: Mapping[str, Any], action_prefix: str) -> bool:
+    uses = step.get("uses")
+    return isinstance(uses, str) and uses.startswith(action_prefix)
+
+
+def step_run_contains(step: Mapping[str, Any], needle: str) -> bool:
```
dgaliffiAMD: ruff check: Unused function.

projects/rocprofiler-systems/scripts/run-ci.py
```diff
@@ -180,49 +217,249 @@ def generate_dashboard_script(args):
...
+        kw_lines.append(f'EXCLUDE "{exclude}"')
+    if include:
+        kw_lines.append(f'INCLUDE "{include}"')
+    if exclude_label:
+        kw_lines.append(f'EXCLUDE_LABEL "{exclude_label}"')
+    if include_label:
+        kw_lines.append(f'INCLUDE_LABEL "{include_label}"')
+
+    params = "\n            ".join(kw_lines)
+    return "", params
```
dgaliffiAMD: If the `preample` is alway empty, can it be removed?
see: https://github.com/ROCm/rocm-systems/commit/dd4032cf2e7998631a00f4cc5f7d3d99bd23d517

projects/rocprofiler-systems/scripts/summarize-junit-results.py
marantic-amd: Thanks, this was very useful!
 I added the checks for the same issues in the validation script for our workflows.

dgaliffiAMD (approved):
Thank you, @marantic-amd. A couple other nit-picks included, but nothing blocking.

kcossett-amd:
Thank you for implementing skipped test viewer!
```
2026-07-16T12:51:27.7108076Z ##[group]Skipped tests (10)
2026-07-16T12:51:27.7108460Z annotate-parallel-overhead-binary-rewrite      Requires either perf_event_paranoid <= 2 or CAP_SYS_ADMIN to be available
2026-07-16T12:51:27.7108806Z annotate-parallel-overhead-sampling            Requires either perf_event_paranoid <= 2 or CAP_SYS_ADMIN to be available
2026-07-16T12:51:27.7109131Z annotate-parallel-overhead-sys-run             Requires either perf_event_paranoid <= 2 or CAP_SYS_ADMIN to be available
2026-07-16T12:51:27.7109484Z overflow-parallel-overhead-binary-rewrite      Requires either perf_event_paranoid <= 2 or CAP_SYS_ADMIN to be available
2026-07-16T12:51:27.7109851Z overflow-parallel-overhead-runtime-instrument  Requires either perf_event_paranoid <= 2 or CAP_SYS_ADMIN to be available
2026-07-16T12:51:27.7110184Z overflow-parallel-overhead-sampling            Requires either perf_event_paranoid <= 2 or CAP_SYS_ADMIN to be available
2026-07-16T12:51:27.7110516Z overflow-parallel-overhead-sys-run             Requires either perf_event_paranoid <= 2 or CAP_SYS_ADMIN to be available
2026-07-16T12:51:27.7110710Z shmem-pingpong-baseline                        oshrun version 4.1 < required 5.0
2026-07-16T12:51:27.7110901Z shmem-pingpong-sampling                        oshrun version 4.1 < required 5.0
2026-07-16T12:51:27.7111080Z shmem-pingpong-sys-run                         oshrun version 4.1 < required 5.0
```

projects/rocprofiler-systems/scripts/run-ci.py
marantic-amd: It can be removed. It's a leftover from when I tried a little bit different argument passing to the actual workflows and ctest script. I will remove it.

.github/workflows/rocprofiler-systems-ubuntu-noble-sanitizers.yml
```diff
@@ -1,45 +1,7 @@
-name: rocprofiler-systems Ubuntu 24.04 Sanitizers (GCC)
-run-name: ubuntu-noble-sanitizers
+name: rocprofiler-systems CI / Sanitizers
```
jbonnell-amd, line 1: Minor nitpick (non-blocking): Perhaps we should rename the file `.github/workflows/rocprofiler-systems-ubuntu-noble-sanitizers.yml` to remove the noble aspect, to keep this as a more generic sanitizer workflow not specific to a particular Ubuntu version (only using noble for now, but could expand for others as needed)

jbonnell-amd (approved):
Looks good! Once merged, we can monitor the reporting and make any adjustments needed afterwards in a future PR. Thanks for the refactoring and improvements!

.github/workflows/rocprofiler-systems-ubuntu-noble-sanitizers.yml
marantic-amd: We could. I didn't rename, because to be honest I was more focused on reorganizing it properly. I don't have anything against renaming it, since as you said this can be expanded in the future to multiple ubuntu distributions, or a different one.

PR #7745, opened 2026-06-24
tests (rocprofiler-system): rocprof sys sample preset and domain flags with gpu workload 

anujshuk-amd:
## Motivation
rocprof-sys-sample presets and domain flags were only validated at the CLI/config level in test_presets.py, which runs a trivial ls workload and asserts that flags translate into the right configuration (e.g. ROCPROFSYS_AMD_SMI_METRICS=temp,power). That path never executes on a GPU, so it can't catch regressions in the actual tracing and metric-collection behavior.

## Technical Details

Change adds end-to-end coverage that runs the sample tool against real HIP workloads (transpose, roctx, scratch-memory) on GPU hardware and asserts the produced output: Perfetto trace categories (hip_runtime_api, kernel_dispatch, memory_copy, rocm_marker_api), a flat wall_clock profile for profile-only, and a rocpd database for workload-trace. It exercises the tracing presets (trace-gpu, trace-hpc, sys-trace, runtime-trace) and the --rocm/--gpu/--cpu domain flags, including compose recipes that combine them.

## JIRA ID

JIRA ID : AIROCVAL-116

## Test Plan

Ran tests locally.

## Test Result

Passes tests locally.

## Submission Checklist


projects/rocprofiler-systems/tests/pytest/test_sample_workloads.py
```diff
@@ -0,0 +1,206 @@
...
+from pathlib import Path
+
+import pytest
+from conftest import RocprofsysTest
+
+pytestmark = [
+    pytest.mark.sample_workloads,
+    pytest.mark.gpu,
+    pytest.mark.rocm,
+    pytest.mark.ci_enable,  # TODO: Deprecate once TheRock switches to CTest
```
kcossett-amd:
```suggestion
```

Do not use this. It has been deprecated.

projects/rocprofiler-systems/tests/pytest/test_sample_workloads.py
```diff
@@ -0,0 +1,206 @@
...
+
+from pathlib import Path
+
+import pytest
+from conftest import RocprofsysTest
+
+pytestmark = [
+    pytest.mark.sample_workloads,
+    pytest.mark.gpu,
+    pytest.mark.rocm,
```
kcossett-amd:
```suggestion
```

No longer needed to get tests running on the `gfx950` workflow.

projects/rocprofiler-systems/tests/pytest/test_sample_workloads.py
```diff
@@ -0,0 +1,206 @@
...
+        ["--preset=runtime-trace"],
+        "transpose",
+        _TRANSPOSE_ARGS,
+        ["hip_runtime_api"],
+        id="preset-runtime-trace-transpose",
+    ),
+]
+
+
+@pytest.mark.class_name("sample-workloads")
```
kcossett-amd:
```suggestion
@pytest.mark.timeout(120)
@pytest.mark.class_name("sample-workloads")
```

Put timeout at the class level since every method has the same timeout.

projects/rocprofiler-systems/tests/pytest/test_sample_workloads.py
```diff
@@ -0,0 +1,206 @@
...
+            "sampling",
+            target="transpose",
+            sampling_args=["--preset=profile-only"],
+            run_args=_TRANSPOSE_ARGS,
+            check_target_arch=True,
+        )
+        # Clean run, and with tracing off no perfetto trace should be written.
+        self.assert_regex(result, fail_regex=[r"\[perfetto\]> Outputting"])
+        # The flat profile is still produced.
+        self.assert_timemory(result, file_name="wall_clock.json", metric="wall_clock")
```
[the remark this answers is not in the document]
kcossett-amd: I guess you could add it. I doubt anything will ever change in logging but better safe then sorry.

projects/rocprofiler-systems/tests/pytest/test_sample_workloads.py
```diff
@@ -0,0 +1,206 @@
...
+
+# =============================================================================
+# Fixtures
+# =============================================================================
+
+
+@pytest.fixture
+def workload_trace_env() -> dict[str, str]:
+    # The rocpd marker sets ROCPROFSYS_USE_ROCPD on this dict before it reaches
+    # run_test; the workload-trace preset already turns on rocpd output.
```
kcossett-amd:
```suggestion
```

Unneeded. Reminder to myself to add a fixture that has empty env in the future.

projects/rocprofiler-systems/tests/pytest/test_sample_workloads.py
```diff
@@ -0,0 +1,206 @@
...
+    return {}
+
+
+@pytest.fixture
+def default_rules(validation_rules_dir: Path) -> list[Path]:
+    # The transpose runs here are tiny, so use the workload-agnostic rules.
```
kcossett-amd:
```suggestion
```

Unneeded comment

projects/rocprofiler-systems/tests/pytest/test_sample_workloads.py
kcossett-amd, line 1: Some of these tests overlap with `test_presets.py`. I guess I can understand why these tests exist, but I don't see why we need them. 

Take `sample-workloads-rocm-hip-kernel-memory-transpose`, it is basically the same as `sample-domain-flags-rocm` test except you check that the perfetto file has certain tracks, which can be trivially achieved in the `sample-domain-flags-rocm` by adding that same `assert_perfetto` check.

The only real difference is the binary used: `ls` vs `transpose` and if a GPU is involved.

kcossett-amd (changes requested):

sputhala-amd (commented):
What is the intent of these tests. If its to test the presets then its already covered. There is no need to run a real workload to tests the validity of presets.
Other parts of the testing with transpose is also covered in existing tests.

anujshuk-amd:
> What is the intent of these tests. If its to test the presets then its already covered. There is no need to run a real workload to tests the validity of presets. Other parts of the testing with transpose is also covered in existing tests.

Added new coverage. Removed overlapping tests.

projects/rocprofiler-systems/tests/pytest/test_sample_workloads.py
anujshuk-amd: Added new coverage. Removed overlapping tests.

projects/rocprofiler-systems/tests/pytest/test_scratch_memory.py
```diff
@@ -72,9 +93,59 @@ def test(self, mode, scratch_memory_env, scratch_memory_rules):
...
+
+# =============================================================================
+# rocprof-sys-sample CLI on scratch-memory
+# =============================================================================
+
+
+@pytest.mark.timeout(120)
+@pytest.mark.sampling
+@pytest.mark.class_name("scratch-memory-sample-cli")
+class TestScratchMemorySampleCLI(RocprofsysTest):
```
kcossett-amd, line 126: This entire class is not needed. You are just duplicating behavior of above scratch memory test as it by default sets `ROCPROFSYS_ROCM_DOMAINS="hip_api,hsa_api,kernel_dispatch,memory_copy,memory_allocation,scratch_memory"`

projects/rocprofiler-systems/tests/pytest/test_roctx.py
```diff
@@ -150,7 +171,50 @@ def test_sampling(
...
+
+# ============================================================================
+# Test Class: rocprof-sys-sample CLI on roctx
+# ============================================================================
+
+
+@pytest.mark.timeout(120)
+@pytest.mark.sampling
+@pytest.mark.class_name("roctx-sample-cli")
+class TestROCTxSampleCLI(RocprofsysTest):
```
kcossett-amd, line 205: This entire class is not needed. This is duplicating behavior from base `roctx` tests as it by default sets `ROCPROFSYS_ROCM_DOMANS="hip_runtime_api,marker_api,kernel_dispatch"`

projects/rocprofiler-systems/tests/pytest/test_transpose.py
```diff
@@ -180,18 +234,52 @@ def test_sampling(self, transpose_env, transpose_rules):
...
 
-    @pytest.mark.parametrize(
-        "mode",
-        [
-            pytest.param("sampling", marks=pytest.mark.timeout(120)),
-            pytest.param("sys_run"),
-        ],
-    )
+    @pytest.mark.sampling
+    def test_profile_only(self):
```
kcossett-amd, line 267: I don't think you need to sample `transpose` if you just want to test that no perfetto file is outputted and a flat wall_clock is. You could move this to a test under `test_binaries` and sample `ls/coreutils=ls`, that way, it won't require a GPU to run.

projects/rocprofiler-systems/tests/pytest/test_transpose.py
```diff
@@ -269,19 +354,94 @@ def test_hip_stream(self, mode, type):
...
 
+# ============================================================================
+# Test Class: rocprof-sys-sample CLI on transpose
+# ============================================================================
+
+
+@pytest.mark.timeout(120)
+@pytest.mark.sampling
+@pytest.mark.class_name("transpose-sample-cli")
+class TestTransposeSampleCLI(RocprofsysTest):
```
kcossett-amd, line 365: There seems to be large overlap here. Although I do notice unique coverage with `rocpd` generation via `workload-trace`.

However, IIRC, the flags set `env` vars, which can just be a unit test under `test_json_config` no?

kcossett-amd (changes requested):
I feel most of these tests are still uneeded. The new assertions in existing tests are good, but for the `SampleCLI`, the flags are mostly used to set env vars, which should be unit tested instead. You could for example, in `test_json_config.cpp`, put:
```

// ----------------------------------------------------------------------------
// expand_rocm_domains(): the --rocm=<tokens> CLI shorthand that resolves to
// ROCPROFSYS_ROCM_DOMAINS. These cover the flag->env mapping directly; the
// env->trace behavior for each domain is already exercised by the
// domain-driven sampling tests in test_roctx.py / test_scratch_memory.py.
// ----------------------------------------------------------------------------

// --rocm=hip,marker (the roctx workload's sample-CLI recipe)
TEST_F(json_config_test, expand_rocm_domains_hip_marker)
{
    EXPECT_EQ(expand_rocm_domains("hip,marker"), "hip_runtime_api,marker_api");
}

// --rocm=hip,kernel (the scratch-memory workload's sample-CLI recipe; mirrors
// the launcher-level check in test_presets.py::TestDomainFlags::test_rocm)
TEST_F(json_config_test, expand_rocm_domains_hip_kernel)
{
    EXPECT_EQ(expand_rocm_domains("hip,kernel"), "hip_runtime_api,kernel_dispatch");
}

TEST_F(json_config_test, expand_rocm_domains_aliases)
{
    EXPECT_EQ(expand_rocm_domains("roctx"), "marker_api");
    EXPECT_EQ(expand_rocm_domains("scratch"), "scratch_memory");
    EXPECT_EQ(expand_rocm_domains("mem"), "memory_copy");
    EXPECT_EQ(expand_rocm_domains("kernels"), "kernel_dispatch");
    EXPECT_EQ(expand_rocm_domains("hsa"), "hsa_api");
    EXPECT_EQ(expand_rocm_domains("rccl"), "rccl_api");
}

// Tokens are trimmed and lowercased before lookup.
TEST_F(json_config_test, expand_rocm_domains_case_and_whitespace_insensitive)
{
    EXPECT_EQ(expand_rocm_domains("HIP, Marker"), "hip_runtime_api,marker_api");
}

// Unrecognized tokens (e.g. already-canonical domain names) pass through as-is.
TEST_F(json_config_test, expand_rocm_domains_unknown_passthrough)
{
    EXPECT_EQ(expand_rocm_domains("hip,kfd_events"), "hip_runtime_api,kfd_events");
}
```

However, if you have specific reasons for these tests that I might have missed, please let me know.

PR #7747, opened 2026-06-24
fix(rocprofiler-systems): Recover missing UCX merged Perfetto trace

habajpai-amd:
## Motivation

UCX MPI sys-run tests were failing Perfetto validation because `merged.proto` was missing when pytest started validation.
The per-rank Perfetto traces were generated correctly, and manual merge produced a valid `merged.proto`, so this was a test/merge timing issue rather than a UCX tracing issue.

## Technical Details

 - Switch UCX tests from legacy Perfetto mode to cached Perfetto mode.
 - Enable Perfetto merge output for UCX tests.
 - Add a pytest-side recovery path for `merged.proto`:
   - wait briefly for `merged.proto`
   - if still missing, check for per-rank `perfetto-trace-*.proto`
   - run `rocprof-sys-merge-output.sh`
   - validate the recovered `merged.proto`
 - Keep recovery scoped to `merged.proto` only, so unrelated missing Perfetto files still fail normally.
 - Add regression coverage for the recovery helper:
   - successful recovery
   - non-`merged.proto` skip path
   - no per-rank traces path
   - merge timeout path
   - merge-script failure path

## JIRA ID

JIRA ID - AIPROFSYST-428

## Test Plan

Validated on gfx1250:
ctest --test-dir build/ucx-pr-clean -L '^ucx$' --output-on-failure -V

## Test Result

gfx1250 UCX validation passed:
- Targeted ucx-mpi-allreduce-sys-run passed.
- Full UCX CTest label set passed: 24/24 tests, 0 failures.
- Perfetto validation found UCX Comm Recv and UCX Comm Send in merged.proto.
 
image

## Submission Checklist


projects/rocprofiler-systems/CHANGELOG.md
```diff
@@ -99,6 +99,8 @@ Full documentation for ROCm Systems Profiler is available at [https://rocm.docs.
   `ROCPROFSYS_TRACE`; setting `ROCPROFSYS_USE_TRACE` had no effect.
 - Fixed a pre-main `rocprof-sys-run` `SIGSEGV` in `rocprofiler_configure()` when
   profiling OpenMPI GPU-aware MPI workloads.
+- Fixed UCX MPI Perfetto validation by recovering a missing `merged.proto` from
```
dgaliffiAMD: Can you move changelog to a new section for "ROCm Systems Profiler 1.8.0 for ROCm 7.15.0". I'll bump the version, now that it has branched.
Since only the tests are affected, a change log entry is complete optional.

projects/rocprofiler-systems/tests/pytest/conftest.py
```diff
@@ -2309,6 +2318,92 @@ def _assert_file_regex(
...
+            False, f"Recovery failed: could not execute {merge_script}: {exc}"
+        )
+
+    message = (
+        "Recovery attempted with "
+        f"{merge_script}; returncode={result.returncode}\n{result.stdout}"
+    )
+    if result.returncode == 0 and perfetto.exists():
+        return PerfettoMergeRecoveryResult(True, message)
+    return PerfettoMergeRecoveryResult(False, message)
```
dgaliffiAMD:
```suggestion
    detail = f"{merge_script}; returncode={result.returncode}\n{result.stdout}"
    if result.returncode == 0 and perfetto.exists():
        return PerfettoMergeRecoveryResult(True, f"Recovery succeeded with {detail}")
    if result.returncode == 0:
        return PerfettoMergeRecoveryResult(
            False,
            f"Recovery failed: {merge_script} exited 0 but produced no "
            f"{MERGED_PERFETTO_FILE}\n{detail}",
        )
    return PerfettoMergeRecoveryResult(False, f"Recovery failed with {detail}")
```
suggestion: slightly more detail in the error message.

habajpai-amd: addressed, thank you.

projects/rocprofiler-systems/tests/pytest/conftest.py
```diff
@@ -72,6 +75,12 @@
 # CTests set their timeout to DEFAULT_TIMEOUT + CTEST_TIMEOUT_BUFFER
 DEFAULT_TIMEOUT = 300
 CTEST_TIMEOUT_BUFFER = 30  # Not overridable
+MERGED_PERFETTO_FILE = "merged.proto"
+_PERFETTO_TRACE_GLOB = "perfetto-trace-*.proto"
+_PERFETTO_MERGE_SCRIPT = "rocprof-sys-merge-output.sh"
+_MERGE_WAIT_ATTEMPTS = 20
+_MERGE_WAIT_INTERVAL_S = 0.1
+_MERGE_SCRIPT_TIMEOUT_S = 60
```
kcossett-amd: Could we decrease this timeout? I see that the `perfetto-merge` tests take <1 second to run.
Of course, I assume time scales with data, but 60 seconds seems far too much.

projects/rocprofiler-systems/tests/pytest/conftest.py
```diff
@@ -2349,7 +2454,16 @@ def _assert_perfetto(
             else:
                 perfetto = result.perfetto_file
             if not perfetto.exists():
-                pytest.fail(f"Perfetto trace file {perfetto} not found")
+                if perfetto.name != MERGED_PERFETTO_FILE:
+                    pytest.fail(f"Perfetto trace file {perfetto} not found")
+
+                recovery = _recover_merged_perfetto_trace(perfetto, tests_dir)
```
kcossett-amd, line 2467: You never use `recovery.ok` in the actual code, only in the new tests that test the function itself.

habajpai-amd: Addressed, thank you.

habajpai-amd:
The TheRock CI failure appears unrelated to this PR; it fails in hip-tests due to a missing YAML config entry for `memory/Unit_hipMemset_UnalignedKernelCornerCases`, while this PR only changes rocprofiler-systems pytest UCX/Perfetto handling.

dgaliffiAMD (approved):

marantic-amd:
Do we actually need still the fallback mechanism for calling the merge script. With `PERFETTO_COMBINE_TRACES=ON` and on cached path you should have the merged proto file generated without any issues. SO I think that the whole "recovery" branch in the test is obsolete and in the first place it's not the right solution since it just masks the issue of our code not creating the merged file as it should

habajpai-amd:
> Do we actually need still the fallback mechanism for calling the merge script. With `PERFETTO_COMBINE_TRACES=ON` and on cached path you should have the merged proto file generated without any issues. SO I think that the whole "recovery" branch in the test is obsolete and in the first place it's not the right solution since it just masks the issue of our code not creating the merged file as it should

Thank you, @marantic-amd. Let me reassess this, and I’ll get back to you on this PR.

habajpai-amd:
Thanks @marantic-amd, I agree with the feedback.
The UCX test was still forcing legacy Perfetto, so it was not actually exercising the cached Perfetto path.
I’ll close this PR and open a smaller replacement PR that only updates the UCX test configuration to use cached Perfetto with combine traces enabled.

PR #7327, opened 2026-06-16
[rocrtst] ROCM-22477 new rcrtst

bimaniamd:
New rocrtst's to check race conditions and trigger TSAN warnings.

## Motivation

 New rocrtst's to check race conditions and trigger TSAN warnings

## Technical Details

concurrent_async_copy.cc: 5 specialized TSAN tests only (TestSdmaDoorbellRace, TestSdmaEngineExhaustion, TestSignalCompletionRace, TestVaryingCopySizeStress, TestConcurrentCopyAndKernelDispatch)

memory_async_copy.cc: Performance benchmarks + 2 TSAN tests (TestConcurrentAsyncCopy, TestMultiGpuP2P)

main.cc: Updated registrations.
## JIRA ID

AILIROCR-484

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #7754, opened 2026-06-24
Users/bimani/rocm 21691 cas implementation correctness

bimaniamd:
## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/rocrtst/suites/test_common/main.cc
+++ b/projects/rocr-runtime/rocrtst/suites/test_common/main.cc
@@ -780,6 +780,13 @@ TEST(rocrtstStress, Queue_LoadStore_Write_Index_ConcurrentTest) {
     RunCustomTestEpilog(&Qw);
 }
 
+TEST(rocrtstStress, Signal_CAS_No_Spurious_Wakeup) {
+    QueueWriteIndexConcurrentTest Qw(false, false, false);
+    if (!RunCustomTestProlog(&Qw)) return;
+    Qw.TestCasNoSpuriousWakeup();
+    RunCustomTestEpilog(&Qw);
+}
+
 TEST(rocrtstPerf, Memory_Async_Copy) {
   MemoryAsyncCopy mac;
   // To do full test, uncomment this:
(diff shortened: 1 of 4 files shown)


dayatsin-amd (approved):
Approving changes interrupt_signal. I will let Shweta/Chris review roctst changes

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/interrupt_signal.cpp
```diff
@@ -358,7 +362,9 @@ hsa_signal_value_t InterruptSignal::CasRelease(hsa_signal_value_t expected,
   hsa_signal_value_t ret = hsa_signal_value_t(
       atomic::Cas(&signal_.value, int64_t(value), int64_t(expected),
                   std::memory_order_release));
-  SetEvent();
+  // Only wake waiters if CAS succeeded (signal value actually changed)
+ if (ret == expected)
+    SetEvent();
```
cfreeamd, line 367: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** Indentation inconsistency: the `if` here has one leading space while the identical blocks in `CasRelaxed`, `CasAcquire`, and `CasAcqRel` all use two spaces. Looks like a copy-paste artifact.

```cpp
// should be:
  if (ret == expected)
    SetEvent();
```

projects/rocr-runtime/rocrtst/suites/test_common/main.cc
```diff
@@ -780,6 +780,13 @@ TEST(rocrtstStress, Queue_LoadStore_Write_Index_ConcurrentTest) {
     RunCustomTestEpilog(&Qw);
 }
 
+TEST(rocrtstStress, Signal_CAS_No_Spurious_Wakeup) {
+    QueueWriteIndexConcurrentTest Qw(false, false, false);
+    if (!RunCustomTestProlog(&Qw)) return;
```
cfreeamd, line 785: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** `rocrtstStress.Signal_CAS_No_Spurious_Wakeup` is registered here but not added to `projects/rocr-runtime/rocrtst/suites/test_common/test_categories.yaml`. Without that entry the test is invisible to CI tiers (standard, comprehensive, full). See #7327 reviewer comment from @shwetagkhatri for the same requirement.

cfreeamd (commented):
Files that look correct: `interrupt_signal.cpp` (the conditional `if (ret == expected) SetEvent()` logic is right — CAS returns the old value, so `ret == expected` means the swap succeeded and the value changed), `queue_write_index_concurrent_tests.h`, `queue_write_index_concurrent_tests.cc` (RunSignalOperationTest helper is sound).

PR #7756, opened 2026-06-24
fix(rocprofiler-compute): Update pip version in workflow

jbonnell-amd:
## Motivation

We noticed that Ubuntu 22.04 runs were failing all tests due to missing pytest module.

## Technical Details

- Update `.github/workflows/rocprofiler-compute-continuous-integration.yml`
  - Updated pip install step to show current pip version and perform a pip upgrade
    - Issue was with the default version of pip in 22.04 using an older version of pip that did not support the `--break-system-packages` argument
  - Update setup_env step to only add coverage and nightly label to CDash when on schedule, to support manual workflow_dispatch runs without overriding the nightly result 

## JIRA ID

JIRA ID : SWDEV-12345

## Test Plan

Ensure 22.04 workflows pass without failing tests

## Test Result

Workflows pass

## Submission Checklist


.github/workflows/rocprofiler-compute-continuous-integration.yml
```diff
@@ -116,13 +116,19 @@ jobs:
...
           else
             MODE=Continuous
             ADD_COVERAGE=""
           fi
+          if [ '${{ github.event_name }}' = 'schedule' ] || [ '${{ inputs.mode }}' = 'nightly' ]; then
+            EXCLUDED_TESTS=""
+          else
+            EXCLUDED_TESTS="test_profile_live_attach_detach"
+          fi
+
```
vedithal-amd: This seems un-related

vedithal-amd (changes requested):

.github/workflows/rocprofiler-compute-continuous-integration.yml
jbonnell-amd: Thanks, this was added before I updated the branch and these changes don't seem necessary anymore given the new state of the CI workflow file. Removed.

cfallows-amd (approved):

vedithal-amd (approved):

dgaliffiAMD (approved):

PR #1962, opened 2025-11-20
[Cherry-pick] rocr: Fix error when internal signal is destroyed

amd-alysaliu:
Fix error when we destroy internal signals during shutdown.

PR #7759, opened 2026-06-24
[RCCL] Validate NCCL 2.29.7 defect - memory init in P2P transport

JhaShweta1:
Add `Alloc.ncclCuMemAllocZeroInitialized` unit test that asserts the cuMem (VMM) allocation path returns zero-initialized memory, locking in the fix for NCCL GitHub issue [#1962](https://github.com/NVIDIA/nccl/issues/1962) ("Fix memory initialization in the P2P transport") and the RCCL ROCM-20370 workaround in `ncclCuMemAlloc`.

The test allocates via `ncclCuMemAlloc` and verifies every byte is zero. Verified it FAILS when the zeroing is removed/poisoned and PASSES with the fix in place. Gated on `ROCM_VERSION >= 70000` and skips gracefully when no GPU is present or the VMM driver API is unavailable.

## Motivation

NCCL 2.29.7 fixed issue [#1962](https://github.com/NVIDIA/nccl/issues/1962): the cuMem (VMM) allocation path in the P2P transport returned uninitialized memory (no `memset` after `cuMemAlloc`), whereas the legacy `cudaMalloc` path zeroed via `ncclCudaCalloc`. These shareable buffers back the P2P FIFO control structures (`ncclSendMem`/`ncclRecvMem` — `head`, `tail`, `ptrExchange`, `redOpArgExchange`), which must start at zero; otherwise a receiver can observe stale/garbage flags before the producer writes them, leading to data corruption or stalls.

RCCL already satisfies this requirement — but it does so centrally inside `ncclCuMemAlloc` (the ROCM-20370 workaround, where `hsa_amd_vmem_map` writes bookkeeping metadata into the user-visible buffer after the kernel driver's SDMA clear, leaving non-zero residue). This is an AMD-relevant correctness behavior with no automated coverage. This PR adds a deterministic regression guard so the zeroing can't be silently dropped (e.g. removed for perf) in the future.

## JIRA ID

AICOMRCCL-1111

## Test Plan

- Run: `./build/debug/test/rccl-UnitTestsFixturesDebug --gtest_filter='Alloc.ncclCuMemAllocZeroInitialized'`.
- Validate fail-before / pass-after by temporarily neutralizing the zeroing in `ncclCuMemAlloc` (`alloc.h`) and re-running.

## Submission Checklist

### DIFF
--- a/projects/rccl/test/AllocTests.cpp
+++ b/projects/rccl/test/AllocTests.cpp
@@ -8,6 +8,9 @@
 #include <gtest/gtest.h>
 #include <rccl/rccl.h>
 
+#include <cstdint>
+#include <vector>
+
 #include "TestBed.hpp"
 #include "common/ErrCode.hpp"
 #include "common/ProcessIsolatedTestRunner.hpp"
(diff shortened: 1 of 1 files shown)


mch:
The links to the NCCL issue are pointing at the ROCm/rocm-systems repo. Agents are kind sloppy about assuming NCCL issues and PRs are in this repo. Can you update? 1962 was re-opened in NCCL, not sure how we want to handle that...

thomas-huber:
> The links to the NCCL issue are pointing at the ROCm/rocm-systems repo. Agents are kind sloppy about assuming NCCL issues and PRs are in this repo. Can you update? 1962 was re-opened in NCCL, not sure how we want to handle that...

@JhaShweta1 +1

mch (commented):
I don't think this actually tests what was fixed. This appears to assert that `ncclCuMemAlloc` zeros out the memory it allocates. RCCL does do that, but not because of 2.29.7, but rather because of ROCM-20370: https://github.com/ROCm/rocm-systems/blob/d378a17c534e578842474541f0a836b4e980d105/projects/rccl/src/include/alloc.h#L695C3-L717C4

I think the actual fix for the issue described in NVIDIA/nccl#1962 is https://github.com/NVIDIA/nccl/blame/v2.29.7-1/src/transport/p2p.cc#L224 and https://github.com/NVIDIA/nccl/blob/v2.29.7-1/src/include/alloc.h#L463-L476, but those were reverted in https://github.com/NVIDIA/nccl/commit/d199bff4f1ee66c006cbafb80898dbc44d17e009

The revert was part of 2.30.7 and was noted in the release notes on github: 
> Fixes a deadlock caused by cuda stream allocation under PXN when memseting a buffer at runtime.

RCCL brought in that revert with the NCCL Sync for 2.30.7, and the related ticket is AICOMRCCL-1846.

We could reframe this as a test for ROCM-20370, but #10500 already added equivalent tests. 

I think we should close this PR, and close the ticket, linking it to AICOMRCCL-1846, since the change that AICOMRCCL-1111 is intended to verify has already been reverted.

PR #7762, opened 2026-06-24
[rocprofiler-sdk] keep ATT completion callback active after marker pause

venkat1361:
## Motivation

`rocprofv3 --att --marker-trace` failed when applications used `roctxProfilerPause` and `roctxProfilerResume`.
Stopping the profiling context immediately removed ATT queue callbacks and disabled serialization. ATT dispatches submitted before the pause could lose their completion callback, producing no `.att` output and leaving active traces during teardown.

## Technical Details

- Added an atomic `enabled` state to `DispatchThreadTracer`.
- `start_context()` enables ATT instrumentation.
- `pre_kernel_call()` skips ATT packet injection while paused.
- `stop_context()` disables only new ATT instrumentation.
- Queue completion callbacks and serialization remain active across pause/resume so in-flight ATT work can finish.
- Callback removal and serialization shutdown occur during `resource_deinit()`.
- Fix bug in profile_serializer

## JIRA ID

JIRA ID: AIPROFSDK-901

## Test Plan

- added tests with marker-trace and att

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
@@ -411,6 +411,18 @@ DispatchThreadTracer::resource_init()
 void
 DispatchThreadTracer::resource_deinit()
 {
+    enabled.store(false, std::memory_order_release);
+
+    if(auto* controller = hsa::get_queue_controller())
+    {
+        client.wlock([&](auto& client_id) {
+            if(!client_id) return;
+            controller->remove_callback(*client_id);
+            client_id = std::nullopt;
+        });
+        controller->disable_serialization();
+    }
+
     ROCP_TRACE << "Clearing agents";
     auto lk = std::unique_lock{agents_map_mut};
     agents.clear();
@@ -435,6 +447,8 @@ DispatchThreadTracer::pre_kernel_call(const hsa::Queue&              queue,
         rocprof_corr_id.internal = corr_id->internal;
     }
     // TODO: Get external
+
+    if(!enabled.load(std::memory_order_acquire)) return {nullptr, false};
 
     std::shared_lock<std::shared_mutex> lk(agents_map_mut);
 
@@ -494,6 +508,7 @@ DispatchThreadTracer::start_context()
     // Only installs queue-controller callbacks (cached and applied to queues as
     // they are created), so this is safe to call before hsa_init.
     CHECK_NOTNULL(hsa::get_queue_controller())->enable_serialization();
+    enabled.store(true, std::memory_order_release);
 
     // Only one thread should be attempting to enable/disable this context
     client.wlock([&](auto& client_id) {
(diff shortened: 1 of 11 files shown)


projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
```diff
@@ -529,18 +544,10 @@ DispatchThreadTracer::start_context()
...
-
-    client.wlock([&](auto& client_id) {
-        if(!client_id) return;
-
-        // Remove our callbacks from HSA's queue controller
-        controller->remove_callback(*client_id);
-        client_id = std::nullopt;
-    });
-
-    controller->disable_serialization();
```
ApoKalipse-V, line 543: Removing this will cause serialization to be enabled without the ATT context active.
maybe you should call enabled.exchange() and enable/disable serialization based on return value.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
```diff
@@ -529,18 +544,10 @@ DispatchThreadTracer::start_context()
...
 DispatchThreadTracer::stop_context()  // NOLINT(readability-convert-member-functions-to-static)
 {
-    auto* controller = hsa::get_queue_controller();
-    if(!controller) return;
-
-    client.wlock([&](auto& client_id) {
-        if(!client_id) return;
-
-        // Remove our callbacks from HSA's queue controller
-        controller->remove_callback(*client_id);
```
ApoKalipse-V, line 539: Is this how counter collection and SPM do it? Seems strange to keep the callback forever.
Are we adding multiple callbacks to the same service if we keep cycling start_context() + stop_context()?

ApoKalipse-V (changes requested):

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
venkat1361: added it back. there was minor bug in profile_serializer where queue.cpp completion only notifies the serializer for serialized packets.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
venkat1361: start_context() checks client_id, so repeated start/stop cycles do not register multiple callbacks. The callback remains registered only for the service lifetime and is removed in resource_deinit(). This is necessary so completion callbacks for ATT dispatches submitted before pause are not dropped.

ApoKalipse-V (approved):
Gonna be honest:
* I don't like that we have even more variables/states associated with context management
* I don't like that those callbacks stay there forever, though callback mechanisms is going away eventually.

I'm approving because we need it fixed, but under the condition we have a ticket to get a proper solution here.

venkat1361:
> Gonna be honest:
> 
> * I don't like that we have even more variables/states associated with context management
> * I don't like that those callbacks stay there forever, though callback mechanisms is going away eventually.
> 
> I'm approving because we need it fixed, but under the condition we have a ticket to get a proper solution here.

AIPROFSDK-1056

PR #7764, opened 2026-06-24
rocr: Drain ASAN quarantine on runtime teardown

ApurvMishra-amd:
Drain ASan quarantine on runtime teardown to avoid stale device chunks

## Motivation

On full teardown, `Unload()` unmaps ROCr device memory still held in the ASAN
device allocator's quarantine, leaving dangling chunks in the sanitizer's
process-global allocator. A later `hsa_amd_pointer_info()` then reads an
uninitialized chunk header and aborts in `DeviceAllocatorT::GetBlockBegin`.

Under ASAN, `hsa_amd_memory_pool_free` does not release the memory immediately.
The sanitizer holds it in a quarantine and performs the real release (and runs
ROCr's deallocation notifier) later. Tests that check the result of a free
right after calling it therefore fail: `Memory_Available` sees the freed VRAM as
still in use, and `Deallocation_Notifier` sees its callback as not yet run.

## Technical Details

Call `__sanitizer_purge_allocator()` in `Runtime::Release()` before `Unload()` so
the quarantine is drained while device memory is still mapped. Guarded by
`SANITIZER_AMDGPU`.

Drain the quarantine with __sanitizer_purge_allocator() so the deferred release
completes before each check (ROCRTST_ASAN-guarded; no-op otherwise).
The recursive and double-free notifier callbacks run while the quarantine is
already being drained, and the drain cannot be invoked again from inside
itself, so their inner same-callback checks are skipped under ASAN. The outer
notifier behavior is still validated.

## JIRA ID

ROCM-26384, 26746

## Test Plan

Running full rocrtst suite.

### DIFF
--- a/projects/rocr-runtime/rocrtst/suites/functional/deallocation_notifier.cc
+++ b/projects/rocr-runtime/rocrtst/suites/functional/deallocation_notifier.cc
@@ -77,6 +77,15 @@ struct callback_status {
 static callback_status notifiers[2];
 static hsa_amd_memory_pool_t pool;
 
+#ifdef ROCRTST_ASAN
+// ASAN defers the real free (and thus the notifier) via its quarantine; drain
+// it so callbacks have run before we check them.
+extern "C" void __sanitizer_purge_allocator(void);
+static inline void DrainAsanQuarantine() { __sanitizer_purge_allocator(); }
+#else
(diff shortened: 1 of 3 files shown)


dayatsin-amd (approved):

shwetagkhatri (approved):
LGTM

PR #7768, opened 2026-06-24
[rocprofiler-compute] Remove the CSV profile backend (Phase A)

jamessiddeley-amd:
## Motivation

Phase A of the profile-interface refactor: rocpd .db becomes the only profile output format, so there is one storage format before the profiling_data boundary lands.

rocpd was already the default; `--format-rocprof-output` let you pick CSV as well, but two formats meant a "not supported in CSV" caveat on every new feature, plus a parallel set of conversion code and tests. So we are just taking that out.

Design doc (Phase A): #7424

## Technical Details

**Profile side**

- --format-rocprof-output is removed. Profiling always writes rocpd.
- Analyze no longer reads CSV-shaped workload directories. Ones produced by the CSV backend are rejected with an error naming the directory which tells users to analyze those with an older release. Workloads predating the backend that ship a joined pmc_perf.csv still work.
- `--join-type` is deprecated. It only fed the removed wide-merge, so it now warns, does nothing, and is no longer recorded in profiling_config.yaml. Deleting the flag, its two golden workloads, their tests and the doc references is Phase A-2.
- Profile mode streams the counter collection into `results_*.csv` rather than holding every counter row in memory.

## Scope note

`results_*.csv` is intentionally kept, analyze still reads it. Removing the intermediate so analyze reads .db directly is Phase B.

## Diff size
Two regenerated `vcopy/MI350` workloads are 184/211 of the files and ~99% of the lines. 
**The change itself is +294 −1,152 across 14 files in src/; tests are +563 −2,181.**

## JIRA ID

Phased implementation for AIPROFCOMP-593

## Test Plan

- Profile and analyze unit suites: `test_utils`, `test_utils_profile_csv`, `test_file_io`, `test_analysis_base`, `test_argparser`, `test_profile_general`, `test_profiler_base`, `test_analyze_workloads`.
- `ruff check` and `ruff format` on changed files.
- Confirm no remaining references to `--format-rocprof-output` or the removed CSV conversion helpers.
- Confirm a legacy `format_rocprof_output: csv` workload is rejected in every analyze mode.
- Manual profile + analyze run on real hardware to verify rocpd output and report.

## Test Result

- `ruff check` / `ruff format`: clean on all changed files.
- Unit suites pass locally in the dev container (MI350): `test_utils_profile_csv`, `test_file_io`, `test_analysis_base`, `test_argparser`, and the `run_prof` / profiling-format / subprocess subsets of `test_utils`.
- Manual run: `profile -n csv_smoke2 --device 0 -- tests/occupancy` followed by `analyze -b 2`. The `results_*.csv` files carry no `PID`, have `Dispatch_ID` renumbered per unique dispatch and `Kernel_ID` per launch shape, none are zero-byte, and the report renders. `analyze -b 2.1.0` on the regenerated `tests/workloads/vcopy/MI350` fixture also renders.
- The full `test_profile_general.py` hardware suite does not finish in a reasonable time in the local container; relying on CI for that one.

## Submission Checklist

- [x] Looked over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.


jamessiddeley-amd:
A design decision I would appreciate some feedback on:

Since as of this PR, rocpd is the only profiling output, I removed the `--format-rocprof-output` flag but kept the `format_rocprof_output` config key, written from a single `_PROFILE_OUTPUT_FORMAT = "rocpd"` constant. Today it only does two things: analyze reads it to reject legacy CSV workloads, and it's the field Phase B's get_reader is meant to select on.

So it's basically a single valued "placeholder" for the future boundary, which is why I leaned toward keeping it rather than removing it entirely. The alternative is to remove it entirely now and reintroduce a real format discriminator in Phase B when `get_reader/get_writer` come in and there's an actual second format to choose.

I leaned toward keeping it (cheap, keeps legacy-reject working, matches the HLD), but happy to strip it if we'd rather not carry a placeholder across phases. Let me know your thoughts!

abchoudh-amd (approved):
LGTM.

One question: Do we have any policy on backward compatibility?

jamessiddeley-amd:
> LGTM.
> 
> One question: Do we have any policy on backward compatibility?

Consensus is that we will not be supporting backward compatibility; its a branch that will be required to be maintained which defeats the purpose of this PR. Users with outdated intermediate formats will have to re-profile or use an older version of ROCm.

projects/rocprofiler-compute/tests/test_utils.py
```diff
@@ -1082,1385 +950,355 @@ def test_run_prof_success_rocprofiler_sdk(tmp_path, monkeypatch):
...
-        )
+    assert update_calls == []
+    assert any("No native counter CSV for pid 12345" in m for m in debug_msgs)
 
 
-def test_process_kokkos_trace_output_csv_with_only_headers(tmp_path, monkeypatch):
-    """
-    Test process_kokkos_trace_output with CSV files that contain
-    only headers but no data.
+def _stub_run_prof_deps(monkeypatch, counter_csv_body, warnings):
```
xuchen-amd:
```suggestion
def stub_run_prof_deps(monkeypatch, counter_csv_body, warnings):
```

nit: `_` prefix is usually not used for test files.

projects/rocprofiler-compute/src/rocprof_compute_profile/profiler_base.py
```diff
@@ -291,10 +292,12 @@ def pre_processing(self) -> None:
             "w",
             encoding="utf-8",
         ) as f:
-            args_dict = vars(self.__args)
+            args_dict = dict(vars(self.__args))
             # Override filter_blocks when writing profiling config yaml
             args_dict["filter_blocks"] = self._filter_blocks
             args_dict["config_dir"] = str(args_dict["config_dir"])
+            args_dict["format_rocprof_output"] = PROFILE_OUTPUT_FORMAT
+            args_dict.pop("join_type", None)
```
xuchen-amd, line 306: just noting that no test verifies the written YAML contains the expected values and omits `join_type`. Not a blocker since `validate_profiling_format` tests would catch a wrong format value at analysis time.

projects/rocprofiler-compute/tests/test_profiler_base.py
xuchen-amd, line 1: just a note that `join_type_grid` and `join_type_kernel` are dead tests, removal would be in the next phase.

xuchen-amd (changes requested):
This looks good!

Validated the following on mi325x:

1. `--help` — `--format-rocprof-output` is gone, `--join-type` shows deprecation text
2. `analyze -p vcopy/MI350 -b 2` — golden workload analyzes correctly with rocpd format
3. `analyze -p no_roof/MI350 -b 2` — second golden workload also works
4. Legacy CSV rejection — `format_rocprof_output`: csv workload exits with a clear error naming the directory
5. `--join-type kernel` — deprecation warning fires before profiling proceeds
6. `--format-rocprof-output csv` — argparse rejects the removed flag

Requesting one coding style change.

projects/rocprofiler-compute/src/utils/utils_profile.py
```diff
@@ -275,202 +272,105 @@ def run_prof(
...
+            "--retain-rocpd-output is deprecated and will be removed in "
+            "a future release. .db files will be retained automatically."
+        )
+        for db_path in db_paths:
+            pid = db_path.stem.split("_")[0]
             shutil.copyfile(
-                f"{workload_dir}/out/pmc_1/results_{fbase}.csv",
-                f"{workload_dir}/results_{fbase}.csv",
+                db_path,
+                workload_dir + f"/{fbase}_{pid}.db",
```
xuchen-amd:
```suggestion
                Path(workload_dir) / f"{fbase}_{pid}.db",
```

xuchen-amd (approved):
approve after fix.

projects/rocprofiler-compute/tests/test_profiler_base.py
jamessiddeley-amd: Yup, that's addressed in #9372

vedithal-amd (changes requested):
will review

vedithal-amd:
@jamessiddeley-amd I will commit the implementation of my review comments directly over here to reduce back and forth time since this pr review has been delayed a lot by me. Will comment a summary of the changes after its pushed.

This biggest thing which I disagree with this PR is keeping backward compatibility for WIDE format pmc_perf.csv - this is exactly the complexity and maintenance cost which we should avoid.

I am going to push a change to check whether results csv and pmc per csv are LONG format and error out if they are WIDE format saying that the output is corrupted and re-profile is needed. The caches tests/workloads should have their pmc_perf.csv converted from WIDE -> LONG format.

Will implement this and connect over a call.

vedithal-amd:
Pushed two commits.

**d0a78e4 — fix test regression + simplify rocpd streaming**
- Fixed `test_utils` failure: `stub_run_prof_deps` rename missed two call sites (NameError ×3, ruff F821, pre-commit).
- Simplified ML-API-trace path: dropped `relabeled_counter_csv` + `.replace()`; `results_*.csv` already holds the relabeled counters.
- `stream_csv_to_files` → single-dest `stream_csv_to_file`.
- `concat_result_csvs` errors on zero counter rows; analyze no longer reuses a zero-byte `pmc_perf.csv`.
- Tightened streaming comment + CHANGELOG; fixed stale `create_df_pmc` example in `PYTHON_CODING_STYLE.md`; moved test globals.

**8379cea — drop wide `pmc_perf.csv`, standardize analyze on rocpd long format**
- Analyze accepts one shape: rocpd long (`Counter_Name`/`Counter_Value`).
- Replaced `validate_profiling_format` (deleted) with a shape check: `concat_result_csvs` and `create_df_pmc` reject non-long input with a re-profile error.
- Regenerated 62 wide golden `pmc_perf.csv` to long; deleted 230 orphan `SQ_*.csv`.

Verified: unit suite green; analyze integration run locally on 6 converted workloads (MI300X_A1/MI300A_A1). Full matrix on CI.

vedithal-amd (approved):
Reviewed and pushed fixes for the CI test_utils failure and the rocpd streaming path. See summary comment. LGTM.

PR #7773, opened 2026-06-24
ROCM-26531 - Fix capture graph leak on execution-context stream detach

victzhan:
## Motivation

Fix the ROCM-26531 LeakSanitizer regression where destroying an execution context that owns a stream with an active capture leaks the captured graph.

## Technical Details

Free the captured graph in Stream::Detach() when the capture is invalidated, since the detached stream can no longer reach hipStreamEndCapture's cleanup. Only the origin frees it; forked streams just clear their aliased pointer to avoid a double free.

## JIRA ID

ROCM-26531

## Test Plan

Ran the two failing capture-detach tests and all other relative tests

## Test Result

No regression

## Submission Checklist


anugodavar:
Good leak fix. Consider matching invalidated hipStreamEndCapture: EraseParallelCaptureStream on fork detach, ReleaseCaptureGraph + remove origin from global capture lists, then (void)EndCapture() for full state reset. Current patch frees the graph but leaves captureEvents_, parallelCaptureStreams_, and global lists uncleared. Execution-context tests can’t call EndCapture (hipErrorStreamDetached), so runtime must do this cleanup here.

anugodavar (approved):

chrispaquot (approved):

manocharahul (approved):

fabiomestre (approved):

PR #7774, opened 2026-06-24
feat(hipFile): Add a check for ais capable filesystem mounts

akolliasAMD:
## Motivation

Adding listing of available mounts 

## Technical Details

Python script addition to ais-check that prints out ais capable mounts. The ais-check will fail if it doesnt find them

## JIRA ID

JIRA ID : AIHIPFILE-153

## Submission Checklist

### DIFF
--- a/projects/hipfile/tools/ais-check/ais-check
+++ b/projects/hipfile/tools/ais-check/ais-check
@@ -344,7 +344,10 @@ def fs_supported(mount):
         return True
     if mount.fstype == "ext4":
         data = mount.option("data")
-        # Absent data= option means ordered (ext4 default), which qualifies.
+        # A missing data= is treated as ordered to match libmount/mountinfo.cpp,
+        # which defaults to ordered when the option is absent. The effective mode
+        # also lives in /proc/fs/ext4/<dev>/options, but we deliberately mirror
+        # the C reference so our verdict matches what hipFile decides at runtime.
         return data is None or data == "ordered"
     return False
 
@@ -372,7 +375,11 @@ def probe_odirect(mountpoint):
     """
     o_direct = getattr(os, "O_DIRECT", 0)
     if o_direct == 0:
-        return None  # platform has no O_DIRECT to test
+        # os.O_DIRECT is only defined when the C library Python was built
+        # against defines O_DIRECT; its absence reflects that build, not
+        # whether the kernel/FS supports O_DIRECT. Without the flag we can't
+        # construct the test open, so we can't verify support either way.
+        return None
 
     flags = os.O_RDWR | os.O_CREAT | os.O_EXCL | os.O_CLOEXEC | o_direct
     path = os.path.join(mountpoint, f".ais-odirect-{os.getpid()}-{os.urandom(4).hex()}")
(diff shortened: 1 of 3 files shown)


projects/hipfile/tools/ais-check/ais-check
```diff
@@ -234,11 +278,240 @@ def amdgpu_supports_ais():
...
+def probe_odirect(mountpoint):
+    """
+    Determine O_DIRECT support by opening a temp file with O_DIRECT.
+
+    Returns True (supported), False (rejected with EINVAL), or None when it
+    could not be verified (e.g. read-only or no write permission).
+    """
+    o_direct = getattr(os, "O_DIRECT", 0)
+    if o_direct == 0:
+        return None  # platform has no O_DIRECT to test
```
riley-dixon: I went down a bit of a rabbit hole to try to understand this condition. os.O_DIRECT is not present if the C library being used at compile time of the Python interpreter does not define O_DIRECT. This is subtly different from the OS/kernel not supporting O_DIRECT.

I think this supports returning None in this instance, but perhaps the comment can be rephrased to specify this nuance of the C library Python is linked to rather than making a comment about the platform itself? Since we only return False if the kernel/OS verifiably reject the IO for having a bad flag value (and we already checked that the FS supports O_DIRECT operations).

riley-dixon (approved):

projects/hipfile/tools/ais-check/ais-check
```diff
@@ -234,11 +277,243 @@ def amdgpu_supports_ais():
...
+    return mount.fstype
+
+
+def probe_odirect(mountpoint):
+    """
+    Determine O_DIRECT support by opening a temp file with O_DIRECT.
+
+    Returns True (supported), False (rejected with EINVAL), or None when it
+    could not be verified (e.g. read-only or no write permission).
+    """
```
jordan-turbofish, line 375: Do we want a check that likely requires root to perform?  Or maybe this should have a command line option to enable?

projects/hipfile/tools/ais-check/ais-check
```diff
@@ -7,21 +7,64 @@
...
+eligibility the hipFile library enforces for a regular file:
+
+  * filesystem is xfs, or ext4 mounted with data=ordered (the ext4 default), and
+  * direct I/O (O_DIRECT) opens succeed on the filesystem.
+
+See src/amd_detail/mountinfo.cpp (fstype/journaling classification) and
+src/amd_detail/backend/fastpath.cpp (the accept rule) for the authoritative
+logic the volume scan reproduces.
+
+The BACKING column (nvme, lvm, md, ...) is informational only: a local NVMe
```
jordan-turbofish, line 26: I think we need to check on this too.  Checking on alola-10:

```
Mounted volumes:
MOUNTPOINT     FSTYPE          DEVICE     BACKING  O_DIRECT    HIPFILE
/              ext4 (ordered)  dm-0       lvm      yes         yes
/boot          ext4 (ordered)  nvme1n1p2  nvme     yes         yes
/boot/efi      vfat            nvme1n1p1  nvme     unverified  no
/mnt/ais/ext4  ext4 (ordered)  nvme0n1p1  nvme     yes         yes
```

The root mount isn't supported because it is on LVM.

projects/hipfile/tools/ais-check/ais-check
jordan-turbofish: Oh, I see.  You just show unverified.  Maybe add some information to the help message?

projects/hipfile/tools/ais-check/ais-check
```diff
@@ -234,11 +277,243 @@ def amdgpu_supports_ais():
...
+
+    ext4 qualifies only with data=ordered (the default when no data= option is
+    present); xfs always qualifies.
+    """
+    if mount.fstype == "xfs":
+        return True
+    if mount.fstype == "ext4":
+        data = mount.option("data")
+        # Absent data= option means ordered (ext4 default), which qualifies.
+        return data is None or data == "ordered"
```
jordan-turbofish: I'm not sure how long this has been the case for, or if we should just assume it.  I'm not sure what libmount does to get this, but you can look at `/proc/fs/ext4/<block_device>/options` to get ext4 options.

john00003 (approved):

PR #7775, opened 2026-06-24
[rdc] Fix non-partitionable gpu support

dmitrii-galantsev:
## Motivation

RDC broke support for unpartitioned GPUs in #4551. Think NV21. Because it lacks partitions - RDC would assume GPU is in a bad state and fail metric fetching.

## Technical Details

`amdsmi_get_gpu_accelerator_partition_profile` returned `AMDSMI_STATUS_NOT_SUPPORTED` and caused RDC to return early.
Forcing the number to 1 solves the issue.

## Test Plan

- Run `rdcd -u` and in another terminal `rdci dmon -u -e 2 -c 1` on different types of machines

## Test Result

- [x] NV21
- [ ] MI300A
- [ ] MI350X


gabrpham (approved):
Looks fine to me. Approved.

marifamd (approved):

PR #7780, opened 2026-06-24
[rocjitsu] Label tests as either build or run-time.

amd-eochoalo:
## Motivation

This is part of cleaning up TheRock's build process for rocjitsu. Right now all of rocjitsu's tests are considered build tests. This means that tests that require a GPU or may use rocm libraries are skipped during the build process as they are assumed to be CPU-only and lack rocm libraries.

A future PR, will separate the tests in TheRock's build process to allow rocjitsu to test both build and run-time tests.

## Technical Details

Tests are now guarded with -DRJ_ENABLE_BUILD_TESTS and -DRJ_ENABLE_RUNTIME_TESTS. Selecting both is equivalent to selecting -DBUILD_TESTING=ON.

Both can also be built at the same time and we can use ctest -L to select whether to run build or run-time tests exclusively.

## JIRA ID

N/A

## Test Plan

Run on CI.

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/CMakeLists.txt
+++ b/emulation/rocjitsu/CMakeLists.txt
@@ -257,6 +257,16 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
 endif()
 # Tests and scaling tools.
 if(BUILD_TESTING)
+    set(_rj_enable_build_tests TRUE)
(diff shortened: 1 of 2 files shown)


amd-eochoalo:
After some discussion, we are moving them all to run-time tests as opposed to keeping some as compile time tests.

PR #7781, opened 2026-06-24
[rocjitsu] Fix gfx1250 flat private scratch decoding

kuhar:
## Motivation

Fix gfx1250 flat private scratch address decoding.

## Technical Details

LLVM lowers private-to-flat scratch addresses using gfx1250 lane bits in the flat scratch address encoding. We were treating those addresses as ordinary global addresses, so nonzero lanes read from the wrong backing location and the HIP repro returned incorrect values.

Add gfx1250 flat private address translation that strips the encoded lane bits and maps the private offset into RocJITsu’s per-lane scratch backing layout, with a regression test covering the lane-bit decoding path.

## JIRA ID

N/A

## Test Plan

Unit tests + repro in https://gist.github.com/lua1235/62fa1506c584b454e3135406eea75c42 + fpsan suite.
No generator changes.

## Test Result

All good.

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/addr_calc.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/addr_calc.cpp
@@ -83,9 +83,35 @@ void init_vector_mem_state(amdgpu::Wavefront &wf, amdgpu::VectorMemState &d) {
   d.cu_path = wf.cu().full_path();
 }
 
+bool decode_flat_private_address(amdgpu::Wavefront &wf, uint64_t addr, uint64_t *translated) {
+  uint32_t lane_stride = wf.scratch_lane_size();
+  if (lane_stride == 0)
+    return false;
+
+  uint32_t wf_size = wf.wf_size();
+  assert(wf_size == 32 || wf_size == 64);
+  uint32_t lane_shift = wf_size == 64 ? 51 : 52;
+  uint64_t lane_mask = static_cast<uint64_t>(wf_size - 1) << lane_shift;
+  uint64_t scratch_base = wf.scratch_base();
+  uint64_t base_without_lane = scratch_base & ~lane_mask;
+  uint64_t addr_without_lane = addr & ~lane_mask;
+  if (addr_without_lane < base_without_lane)
+    return false;
+
+  uint64_t private_offset = addr_without_lane - base_without_lane;
+  if (private_offset > 0xFFFF'FFFFULL)
+    return false;
+
+  if (translated != nullptr) {
+    uint32_t encoded_lane = static_cast<uint32_t>((addr & lane_mask) >> lane_shift);
+    *translated = scratch_base + static_cast<uint64_t>(encoded_lane) * lane_stride + private_offset;
+  }
+  return true;
+}
+
 template <typename Inst>
 void flat_global_calculate_addresses(const Inst &inst, amdgpu::Wavefront &wf,
-                                     amdgpu::VectorMemState &d) {
+                                     amdgpu::VectorMemState &d, bool decode_flat_private) {
   auto &cu = wf.cu();
   init_vector_mem_state(wf, d);
   uint64_t exec = d.exec_mask;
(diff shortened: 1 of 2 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/addr_calc.cpp
```diff
@@ -106,12 +134,7 @@ void flat_global_calculate_addresses(const Inst &inst, amdgpu::Wavefront &wf,
           (static_cast<uint64_t>(cu.read_vgpr(vbase + 1, lane)) << 32) | cu.read_vgpr(vbase, lane);
     }
     uint64_t addr = saddr_val + vaddr + offset;
-    uint32_t priv_hi = static_cast<uint32_t>(wf.private_aperture_base() >> 32);
-    if (priv_hi != 0 && static_cast<uint32_t>(addr >> 32) == priv_hi) {
-      uint64_t lane_base = wf.scratch_base() + static_cast<uint64_t>(lane) * wf.scratch_lane_size();
-      addr = lane_base + (addr & 0xFFFFFFFFULL);
-    }
-    d.per_lane_addr[lane] = addr;
+    d.per_lane_addr[lane] = translate_flat_private_address(wf, addr, lane);
```
atgutier: Flat and Globals use flat_global_calculate_addresses() correct? Do we need a flat segment check so Globals don't go this route? I don't think such addresses would be generated, so maybe an assert suffices?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/addr_calc.cpp
atgutier, line 131: Agent comment:

Legacy aperture-tag fallback can't correctly fire for gfx1250 → silent wrong address on any
  first-path miss — addr_calc.cpp:105-109
  If a real gfx1250 flat-private access misses the first path (e.g. private_offset >= lane_stride),
  control drops to the old logic comparing addr_hi32 == private_aperture_base()>>32 and remapping with
   the iteration lane. But on gfx1250, scratch_base() (~`0x1_0000_0000) is unrelated to
  private_aperture_base() (~0x2_0000_0000_0000), and the formed address carries the lane in bits
  [56:51], not the aperture tag — so the comparison fails, the fallback does nothing, and the function
   returns the raw *swizzled* address (exactly the bug being fixed). It also uses the wrong lane
  (iteration lane, not encoded lane). Verified against dispatch values in
  command_processor.cpp:255-273andsimulated_driver.cpp:375-378. **Fix:** drop the legacy aperture-tag
  branch for gfx1250 (it's dead/incorrect), and add an assert` so a decode miss is loud rather than a
  silent wrong address.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/addr_calc.cpp
```diff
@@ -83,6 +83,34 @@ void init_vector_mem_state(amdgpu::Wavefront &wf, amdgpu::VectorMemState &d) {
...
+    uint32_t wf_size = wf.wf_size();
+    assert(wf_size == 32 || wf_size == 64);
+    uint32_t lane_shift = wf_size == 64 ? 51 : 52;
+    uint64_t lane_mask = static_cast<uint64_t>(wf_size - 1) << lane_shift;
+    uint64_t scratch_base = wf.scratch_base();
+    uint64_t base_without_lane = scratch_base & ~lane_mask;
+    uint64_t addr_without_lane = addr & ~lane_mask;
+    if (addr_without_lane >= base_without_lane) {
+      uint64_t private_offset = addr_without_lane - base_without_lane;
+      if (private_offset < lane_stride) {
```
atgutier: Agent comment:

addr_calc.cpp:98 — the private_offset < lane_stride gate is correct as an in-bounds predicate, but
   on failure it falls to the dead fallback; an OOB scratch access should still map via the encoded
  lane. Once the lane field is confirmed present, use the encoded lane regardless of bounds; only
  fault behavior should depend on the range check.

atgutier (approved):
LGMT, but some comments that may need to be addressed.

PR #7784, opened 2026-06-24
[rocprofiler-compute] Bump scipy test pin for Python 3.13 support

vedithal-amd:
## Motivation

scipy 1.11.4 fails to install on Python 3.13: it pins legacy Cython <3.0, which cannot compile against the CPython 3.13 C-API. Bump the test pin to scipy 1.14.1, keeping numpy 1.26.4 unchanged.

## Technical Details

- requirements-test.txt: scipy 1.11.4 -> 1.14.1.
- 1.14.1 is the lowest scipy with a prebuilt cp313 wheel and runs against the existing numpy 1.26.4 pin.

## JIRA ID

ROCM-26919

## Test Plan

- Create a Python 3.13 venv, run pip install -r requirements-test.txt, then run the test suite.

## Test Result

Verified on Python 3.13: scipy 1.14.1 installs against the existing numpy 1.26.4 pin, with no change to test behavior.

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-compute/requirements-test.txt
+++ b/projects/rocprofiler-compute/requirements-test.txt
@@ -1,4 +1,5 @@
 pytest==8.0.2
 pytest-cov==5.0.0
 pytest-xdist==3.5.0
-scipy==1.14.1
+scipy==1.11.4; python_version < "3.10"
+scipy==1.14.1; python_version >= "3.10"
--- a/projects/rocprofiler-compute/requirements-test.txt
+++ b/projects/rocprofiler-compute/requirements-test.txt
@@ -1,4 +1,4 @@
 pytest==8.0.2
 pytest-cov==5.0.0
 pytest-xdist==3.5.0
-scipy==1.11.4
+scipy==1.14.1


feizheng10 (approved):

xuchen-amd (approved):

cfallows-amd:
Side note- approvals in place but there was offline conversation about backwards compatibility back to python3.9 version we have to support. Vignesh is checking in on this, we should not push until validated and is passing below PR checks.

vedithal-amd:
@cfallows-amd on the 3.9 question: scipy 1.14.1 declares `requires-python >=3.10`, so it's a hard install gate on 3.9 (not just a missing wheel, a flat refusal). Pushed 401650872a to split the pin by version:

```
scipy==1.11.4; python_version < "3.10"
scipy==1.14.1; python_version >= "3.10"
```

3.9 stays on 1.11.4 (unchanged), 3.10-3.13 get 1.14.1 with its cp313 wheel. numpy needs no change: 1.26.4 is `requires-python >=3.9` and source-builds on 3.13 (verified by the 3.13 run). Should now cover the full 3.9-3.13 range.

vedithal-amd:
Closing in favor of #7859, which removes scipy from the test deps entirely instead of bumping the pin. scipy was only used for one `scipy.stats.zscore` call in test_profile_general.py; #7859 computes it directly with pandas (`(s - s.mean()) / s.std(ddof=0)`, ddof=0 to match scipy's population std). That avoids the Python 3.9 vs 3.13 wheel conflict (scipy 1.11.4 has no cp313 wheel, 1.14.1 drops 3.9) with no version pin at all. Same JIRA (ROCM-26919).

PR #7786, opened 2026-06-24
rocr: Close shareable dmabuf fds after use instead of holding them open

dayatsin-amd:
## Summary
- Stop holding a shareable `dmabuf_fd` open for the entire lifetime of a VMM shareable handle; `CreateShareableHandle` now leaves `dmabuf_fd` as `-1` after converting the allocation to a driver handle.
- Lazily export a dmabuf fd in `VMemorySetAccessPerHandle` when peer import needs it, then close it again via a scope guard before returning.
- Pass `DriverMemoryHandle*` through `ImportMemoryHandle` (reading `.dmabuf_fd` or `.fabric_handle`) and close fds in `MemoryHandle` destructor / `DestroyMemoryHandle`.

Fabric-handle export/import is unchanged and continues to use the internal BO handle path.

## Test plan
- [x] Built `libhsa-runtime64` on S83-1 (MI300X) against TheRock toolchain
- [x] `Unit_hipMemExportToShareableHandle_Positive_Basic` — pass
- [x] `Unit_hipMemImportFromShareableHandle_Positive_Basic` — pass
- [x] Full `VirtualMemoryManagementTest` suite (85 cases, `HIP_VISIBLE_DEVICES=0`): 69 passed / 6 failed / 10 skipped
- [x] A/B vs pre-patch baseline ROCr: same 6 failures in full suite; all 6 previously-failing tests pass in isolation on both builds — failures are not regressions from this change (suite-order / VA exhaustion)

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
@@ -4022,11 +4022,11 @@ Runtime::MappedHandleAllowedAgent::MappedHandleAllowedAgent(
   if (memHandle->imported && memHandle->is_fabric_handle) {
     status = targetAgent->driver().ImportMemoryHandle(
         *targetAgent, &driver_handle, ShareType::FABRIC_HANDLE,
-        &memHandle->driver_handle.fabric_handle);
+        &memHandle->driver_handle);
   } else {
     status = targetAgent->driver().ImportMemoryHandle(
         *targetAgent, &driver_handle, ShareType::DMABUF_FD,
-        &memHandle->driver_handle.dmabuf_fd);
+        &memHandle->driver_handle);
   }
   if (status != HSA_STATUS_SUCCESS)
     throw AMD::hsa_exception(status, "Failed to import memory");
@@ -4056,7 +4056,7 @@ hsa_status_t Runtime::MappedHandleAllowedAgent::EnableAccess(hsa_access_permissi
 #endif
     auto agentOwner = mappedHandle->mem_handle->agentOwner();
     int mmap_fd = -1;
-    /* Do not check the return value of GetDeviceFd. We do not need mmap_fd so it is valid for mmap_fd to be -1*/
+    /* Do not check the return value of GetDeviceFd. We do not need mmap_fd in some cases, so it is valid for mmap_fd to be -1*/
     agentOwner->driver().GetDeviceFd(agentOwner->node_id(), &mmap_fd);
 
     if (!rocr::os::MapMemory(va, size, PermissionsToMemProt(perms), mmap_fd,
@@ -4153,6 +4153,11 @@ Runtime::MemoryHandle::MemoryHandle(hsa_fabric_handle_t fabric_handle)
 Runtime::MemoryHandle::~MemoryHandle() {
   if (driver_handle.handle != 0 && region != nullptr)
     agentOwner()->driver().DestroyMemoryHandle(&driver_handle);
+
+  if (driver_handle.dmabuf_fd >= 0) {
+    os::DmaBufClose(driver_handle.dmabuf_fd);
+    driver_handle.dmabuf_fd = -1;
+  }
 }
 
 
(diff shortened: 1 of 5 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
```diff
@@ -4148,6 +4153,11 @@ Runtime::MemoryHandle::MemoryHandle(hsa_fabric_handle_t fabric_handle)
 Runtime::MemoryHandle::~MemoryHandle() {
   if (driver_handle.handle != 0 && region != nullptr)
     agentOwner()->driver().DestroyMemoryHandle(&driver_handle);
```
cfreeamd, line 4155: Suggestion from Claude related to lines 4155 and the block at 4157. In a nutshell, it's saying if we derive from the base class in the future, we may miss doing the clean up that is happening here.

>  ~MemoryHandle dual-cleanup coupling — runtime.cpp:4153
> 
>   The destructor now has two cleanup sites:
>   if (driver_handle.handle != 0 && region != nullptr)
>       agentOwner()->driver().DestroyMemoryHandle(&driver_handle);  // (A)
> 
>   if (driver_handle.dmabuf_fd >= 0) {                              // (B)
>       os::DmaBufClose(driver_handle.dmabuf_fd);
>       driver_handle.dmabuf_fd = -1;
>   }
> 
> This is safe today because KfdDriver::DestroyMemoryHandle now zeroes dmabuf_fd before returning (added in this PR), so block (B) won't fire for
>   owned handles. But the two cleanup sites are coupled through an informal convention: any future Driver subclass that implements
>   DestroyMemoryHandle without zeroing dmabuf_fd will silently double-close the fd — a latent fd-aliasing bug with no compiler enforcement.
> 
> Suggestion: Either make the base class zero dmabuf_fd after the virtual call, or merge the dmabuf_fd close into DestroyMemoryHandle entirely and
>   remove block (B), with a brief comment explaining the imported-handle path.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
```diff
@@ -4156,6 +4166,32 @@ hsa_status_t
...
                                    const hsa_amd_memory_access_desc_t *desc,
                                    const size_t desc_cnt) {
+  MemoryHandle *memHandle = mappedHandle.mem_handle;
+
+  /*
+   * For locally-created shareable handles CreateShareableHandle leaves dmabuf_fd as -1 to avoid
+   * holding an fd open for the lifetime of the handle. Export it lazily here so the target agents
+   * can import the memory below, then close it again before returning.
+   */
+  bool created_dmabuf_fd = false;
```
cfreeamd, line 4176: I'm not sure how often a single MemoryHandle is mapped across multiple VA chunks--if not often, then this probably isn't worth worrying about, but here's Claude feedback on this:

> [MINOR] Per-chunk re-export amplification — runtime.cpp:4166
> 
>   When a single MemoryHandle is mapped across multiple VA chunks (e.g., the same handle mapped twice via VMemoryHandleMap), VMemorySetAccess calls
>   VMemorySetAccessPerHandle once per chunk. Each call finds dmabuf_fd == -1 (the scope guard reset it) and does a fresh ExportMemoryHandle → kernel
>    ioctl. Cost becomes O(chunks × agents) rather than O(agents). Not a correctness issue, but worth lifting the lazy-export out to the per-handle
>   level in VMemorySetAccess if this path gets hot.
>

cfreeamd (approved):
I don't see any big problems, but have a few suggestions for consideration, but nothing blocking.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
dayatsin-amd: Yes, I agree, it is not the cleanest solution. My goal was to have the Driver class decide when to close the dmabuf-fd's. But in this particular case, because we do not have a region on imported handles, we do not have a pointer to the driver class.
I created a follow-up patch to improve it a bit. [PR 7803](https://github.com/ROCm/rocm-systems/pull/7803/) 
As a further change maybe memory handles should have a pointer to the Driver class directly instead of relying on existence of the region to get the Driver class. I will think about that a bit more.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
dayatsin-amd: I think that situation is very rare. We can ignore it for now.

PR #7789, opened 2026-06-24
fix(rocprofiler-sdk): Fix OMPT-init deadlock in client discovery

itrowbri:
## Motivation

Fix deadlock that occurs due to ompt initialization when importing torch in some environments. 

## Technical Details

Deadlock occurs due to ompt initialization. When importing torch in some configurations, OpenBLAS (built with USE_OPENMP=1) can initialize ompt during its library constructor, causing libomp to acquire a non-recursive initialization lock. The sdk gets called with ompt_start_tool, which in turn calls ::rocprofiler::registration::initialize(), which eventually calls find_clients(), which calls dlopen() on a client library exporting rocprofiler_configure. That dlopen re-runs the client's still-pending constructor, which calls back into libomp and tries to reacquire the initialization lock on the same thread, hitting a deadlock. To fix this issue, this PR removes `ompt_start_tool` from the sdk and relocates it to the rocprofv3 tool

## JIRA ID

JIRA ID : ROCM-25067

## Test Plan

Add execution test only that fails when test hangs

## Test Result

New test should not hang.

## Submission Checklist


projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/registration.cpp
```diff
@@ -313,6 +316,33 @@ get_link_map()
...
+resolve_symbol_no_dlopen(const common::elf_utils::ElfInfo& _elf,
+                         uintptr_t                         _load_bias,
+                         std::string_view                  _symname)
+{
+    for(const auto& sym : _elf.dynamic_symbol_entries)
+    {
+        if(sym.name != _symname) continue;
+        if(sym.section_index == SHN_UNDEF) continue;  // undefined import, not a definition here
+        if(sym.type != STT_FUNC) continue;            // refuse IFUNC/exotic symbol types
+        if(sym.value == 0) continue;
```
jrmadsen:
```suggestion
        if(sym.section_index == SHN_UNDEF) continue;  // undefined import, not a definition here
        if(sym.type != STT_FUNC) continue;            // refuse IFUNC/exotic symbol types
        if(sym.value == 0) continue;
        if(sym.name != _symname) continue;
```
Move the string comparison to the end for performance -- we don't need to compare the name if the symbol is undefined/not a function/has no value.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/registration.cpp
```diff
@@ -501,38 +531,25 @@ find_clients()
...
-            auto* _sym        = rocprofiler_configure_dlsym(handle);
-            auto* _attach_sym = rocprofiler_configure_attach_dlsym(handle);
+            // resolve the configure symbols via the ELF symbol table + load bias instead of
+            // dlopen/dlsym; dlopen here can re-run a pending DT_INIT and deadlock libomp during
+            // OMPT bring-up (see resolve_symbol_no_dlopen)
+            auto  elfinfo = common::elf_utils::read(itr, optimize_elf_parsing);
+            auto* _sym    = resolve_symbol_no_dlopen<decltype(::rocprofiler_configure)*>(
+                elfinfo, load_bias, "rocprofiler_configure");
+            auto* _attach_sym = resolve_symbol_no_dlopen<decltype(::rocprofiler_configure_attach)*>(
+                elfinfo, load_bias, "rocprofiler_configure_attach");
```
jrmadsen: So... I understand why this seems like the solution but what about when a tool's `rocprofiler_configure` function depends on the `DT_INIT` / `DT_INIT_ARRAY` running??? `DT_INIT_ARRAY` is responsible for executing global C++ constructors (i.e. global static C++ objects) so it is quite likely that in many tools, any modification of global static C++ objects during the call to `rocprofiler_configure` will either (A) segfault due to operating on uninitialized memory or (B) get overwritten when `DT_INIT_ARRAY` runs later.

I think what we may want is:

1. Enable using this logic by default only when OpenMP is present 
2. Provide an environment variable to force this behavior (but have it disabled by default)
3. Log a warning when we use the logic which bypasses the initialization of the global static C++ objects.

If we go forward with these changes, due to the effect of bypassing `DT_INIT_ARRAY`, we might need to provide some way for a tool library to:

A. Communicate that `rocprofiler_configure` should be called after dlopening the library instead of directly called in this manner. E.g. the library provides some symbol like `rocprofiler_configure_parameters` which, by definition, must support being directly called and if the tool needs the library to be dlopened it returns that data via some struct. 
B. Communicate to that tool that rocprofiler-sdk bypassed `DT_INIT_ARRAY` to call `rocprofiler_configure`
C. Provide some way for them to check/verify their static global variables were not overwritten.
D. Detect if/when this will happen in the tool.

Frankly, this is problematic.

projects/rocprofiler-sdk/tests/bin/openmp/init-deadlock/blas_stub.cpp
```diff
@@ -0,0 +1,46 @@
...
+    fprintf(stderr, "[blas-stub] DT_INIT: entering omp parallel (triggers OMPT bring-up)\n");
+    fflush(stderr);
+#pragma omp parallel
+    {
+        if(omp_get_thread_num() == 0)
+            fprintf(stderr, "[blas-stub] in parallel, nthreads=%d\n", omp_get_num_threads());
+    }
+    fprintf(stderr, "[blas-stub] DT_INIT done\n");
+    fflush(stderr);
+}
```
jrmadsen, line 48: Is this what blas is actually doing? If not, what are they actually doing in their constructor function?

projects/rocprofiler-sdk/tests/bin/openmp/init-deadlock/client.cpp
jrmadsen, line 1: This file should really just be `client.cpp` for the sake of consistency.

projects/rocprofiler-sdk/tests/bin/openmp-init-deadlock/configure_client.cpp
```diff
@@ -0,0 +1,56 @@
...
+__attribute__((constructor)) void
+configure_client_init()
+{
+    fprintf(stderr, "[client] DT_INIT (still pending during discovery): omp_get_num_places()\n");
+    fflush(stderr);
+    int n = omp_get_num_places();
+    fprintf(stderr, "[client] omp_get_num_places() returned %d\n", n);
+    fflush(stderr);
+}
+}  // namespace
```
jrmadsen, line 49: Shouldn't this be in `openmp-init-deadlock.cpp`? Or in some other library? Putting it in the same library as the tool is not really representative of the apps causing the deadlocks. 

I would suggest putting this is a different file and compiled into a separate library which is linked to the executable built from `openmp-init-deadlock.cpp`:

```suggestion
namespace
{
int g_lib_num_places = 0;

__attribute__((constructor)) void
configure_client_init()
{
    fprintf(stderr, "[client] DT_INIT (still pending during discovery): omp_get_num_places()\n");
    fflush(stderr);
    g_lib_num_places= omp_get_num_places();
    fprintf(stderr, "[client] omp_get_num_places() returned %d\n", g_lib_num_places);
    fflush(stderr);
}
}  // namespace

extern "C"
{
int lib_get_num_places() __attribute__((visibility("default"));

int lib_get_num_places()
{
    return g_lib_num_places;
}
}
```

projects/rocprofiler-sdk/tests/bin/openmp/init-deadlock/openmp-init-deadlock.cpp
```diff
@@ -0,0 +1,44 @@
...
+// means no deadlock occurred.
+
+#include <cstdio>
+
+int
+main()
+{
+    printf("reached main: no OMPT-init deadlock\n");
+    return 0;
+}
```
jrmadsen, line 83:
```suggestion
#include <cstdlib>
#include <cstdio>
#include <omp.h>

extern "C"
extern int lib_get_num_places();

int
main()
{
    printf("reached main: no OMPT-init deadlock\n");
    int _omp_num_places = omp_get_num_places();
    int _lib_num_places = lib_get_num_places();
    printf("Number of places via OpenMP call: %d\n", _omp_num_places);
    printf("Number of places via library call: %d\n", _lib_num_places);
    return (_omp_num_places == _lib_num_places) ? EXIT_SUCCESS : EXIT_FAILURE;
}
```

projects/rocprofiler-sdk/tests/bin/openmp-init-deadlock/configure_client.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include <cstdio>
+
+extern "C" rocprofiler_tool_configure_result_t*
+rocprofiler_configure(uint32_t, const char*, uint32_t, rocprofiler_client_id_t*)
+{
+    // a no-op client: returning nullptr means "do not activate", which is sufficient for
+    // exercising the discovery path that previously deadlocked
+    return nullptr;
+}
```
jrmadsen: Your `rocprofiler_configure` should return an init function and a fini function. It should modify some global static variable in the init function and verify the variable has that value in the fini function -- in other words, for global static variable `int var_check = 0;`, set `var_check = 1` in the init function and throw an error if `var_check != 1` in the fini function (which means that the zero initialization of `var_check` ran after init was called).

projects/rocprofiler-sdk/tests/bin/openmp-init-deadlock/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+
+# This regression test needs the LLVM OpenMP runtime (libomp), which performs OMPT tool
+# discovery while holding its non-recursive init ticket lock. The GNU runtime (libgomp)
+# does not reproduce the deadlock the same way, so the test is only registered when libomp
+# is found.
+find_library(
+    openmp-init-deadlock-libomp
+    NAMES omp libomp.so.5 libomp.so
+    HINTS ${ROCM_PATH} ENV ROCM_PATH /opt/rocm
+    PATH_SUFFIXES llvm/lib lib NO_CACHE)
```
jrmadsen: You can prefer the LLVM OpenMP runtime but this test should still run if only the GNU OpenMP is available. Also, Intel provides an OpenMP library...

jrmadsen (changes requested):

projects/rocprofiler-sdk/tests/bin/openmp/init-deadlock/blas_stub.cpp
itrowbri: Changed it to `omp_get_num_places` since that's really what's triggering it in the original bug. But this mimics the stack trace from the original ticket and replicates the hang without the associated fix. I can look further into this if required?

jrmadsen:
This BLAS implementation causing the issue... do we own the SW causing it?

itrowbri:
> This BLAS implementation causing the issue... do we own the SW causing it?

I don't think so. I'm pretty sure is found here: https://github.com/OpenMathLib/OpenBLAS. I don't see a ROCm fork anywhere. It's open source, so we technically could propose a PR to change it, but it's up to the maintainers to accept or reject it.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/tool.cpp
```diff
@@ -4468,4 +4468,24 @@ rocprofv3_main(int argc, char** argv, char** envp)
...
+        "ompt_start_tool(omp_version={}, runtime_version=\"{}\") was invoked in rocprofv3 tool... "
+        "returning rocprofiler_ompt_start_tool(...) result",
+        omp_version,
+        runtime_version);
+
+    initialize_rocprofv3();
+
+    // test force configure
+    return rocprofiler_ompt_start_tool(omp_version, runtime_version);
+}
```
jrmadsen, line 4490: @itrowbri it needs to be documented that rocprofiler-sdk is no longer providing a `ompt_start_tool` symbol in the library. I have confirmed that `rocprofv3 -s ... -- ./bin/openmp-init-deadlock` works. Please add a test for rocprofv3 + ompt-init-deadlock in `tests/rocprofv3/ompt/...`

PR #7791, opened 2026-06-25
fix(rocr): Restrict HSA_MODEL_LIB dlopen to trusted install path (ROCM-26861)

dayatsin-amd:
## Summary

Addresses Mythos security finding SEC-00591 (ROCM-26861 / SWSPLAT-29283): `HSA_MODEL_LIB` was read with `getenv()` and passed directly to `dlopen()`, allowing arbitrary shared-library load in processes with elevated privileges.

- Use `secure_getenv()` for `HSA_MODEL_TOPOLOGY` and `HSA_MODEL_LIB` so both are ignored under `AT_SECURE`
- Validate `HSA_MODEL_LIB` against the build-time `HSAKMT_INSTALL_LIBDIR` prefix (reject `..` traversal)
- Document model-mode environment variable restrictions in `hsakmtmodel.h`

## JIRA ID

JIRA ID : ROCM-26861

## Test plan

- [x] Build `libhsakmt` (static) on Linux — compiles cleanly with `HAVE_SECURE_GETENV` detected
- [ ] Model-mode emulation smoke test with `HSA_MODEL_LIB` pointing to a library under the install libdir
- [ ] Confirm `HSA_MODEL_LIB=/tmp/evil.so` is rejected with a clear error before `dlopen`
- [ ] Confirm model mode is disabled when `HSA_MODEL_LIB` is set under a setuid/capability-elevated wrapper (`secure_getenv` returns NULL)


kentrussell (approved):

projects/rocr-runtime/libhsakmt/CMakeLists.txt
```diff
@@ -273,6 +273,17 @@ else()
...
     endif()
+
+    set(CMAKE_REQUIRED_INCLUDES stdlib.h)
+    check_function_exists(secure_getenv HAVE_SECURE_GETENV)
+    if(HAVE_SECURE_GETENV)
+      target_compile_definitions(${HSAKMT_TARGET} PRIVATE -DHAVE_SECURE_GETENV=1)
+    endif()
+
+    # Trusted search path for HSA_MODEL_LIB dlopen (see hsakmtmodel.c).
+    set(HSAKMT_INSTALL_LIBDIR "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
```
cfreeamd, line 290: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** `HSAKMT_INSTALL_LIBDIR` computed before `/opt/rocm` default is applied — silent mismatch on bare `cmake ..`

The macro is computed on line 284:
```cmake
set(HSAKMT_INSTALL_LIBDIR "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
```
The `/opt/rocm` default is applied 43 lines later:
```cmake
if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
  set(CMAKE_INSTALL_PREFIX "/opt/rocm")
endif()
```
CMake's built-in default is `/usr/local`. A developer who runs `cmake ..` without `-DCMAKE_INSTALL_PREFIX=/opt/rocm` gets `HSAKMT_INSTALL_LIBDIR="/usr/local/lib"` baked in, while the actual ROCm install goes to `/opt/rocm/lib`. The path validation then rejects every `HSA_MODEL_LIB` pointing at `/opt/rocm/lib/...`, causing `abort()` for all FFM users who built this way.

ROCm CI always passes `-DCMAKE_INSTALL_PREFIX=/opt/rocm` so CI is unaffected, but any developer doing a standalone libhsakmt build with default cmake flags will hit this silently.

**Suggestion:** Move the `HSAKMT_INSTALL_LIBDIR` computation to after the `/opt/rocm` default is finalized.

cfreeamd (commented):

Reviewed `libhsakmt/CMakeLists.txt`, `hsakmtmodel.c`, and `hsakmtmodel.h`. The security fix is correct and well-scoped: `secure_getenv()` with compile-time fallback is the standard idiom for setuid-safe env reads. `model_lib_path_allowed()` correctly bounds-checks and the `..`-component check covers all realistic traversal patterns.

projects/rocr-runtime/libhsakmt/CMakeLists.txt
dayatsin-amd: Fixed in e61a80e: moved the `/opt/rocm` `CMAKE_INSTALL_PREFIX` default to immediately after `GNUInstallDirs` (before `HSAKMT_INSTALL_LIBDIR` is computed), so standalone `cmake ..` builds now bake in `/opt/rocm/lib` instead of `/usr/local/lib`. Thanks for catching this.

dayatsin-amd:
Addressed the open review feedback in e61a80e:

1. **CMake prefix ordering** (@cfreeamd): `CMAKE_INSTALL_PREFIX` default (`/opt/rocm`) is now applied before `HSAKMT_INSTALL_LIBDIR` is computed, so bare `cmake ..` builds get the correct trusted lib path.
2. **Unit test policy** (therock-pr-bot): Re-added `test_hsakmtmodel_lib_path.c` with shared path-validation helpers in `hsakmtmodel_path.h` (used by both `hsakmtmodel.c` and the test). Verified locally with `gcc -Iinclude tests/test_hsakmtmodel_lib_path.c`.

dayatsin-amd:
Pushed 6ff7cdec: removed `test_hsakmtmodel_lib_path.c` and `hsakmtmodel_path.h`; path validation stays as static helpers in `hsakmtmodel.c`. The CMake `/opt/rocm` prefix-order fix from e61a80e is unchanged.

cfreeamd (approved):

ScottTodd:
This broke postsubmit builds: https://github.com/ROCm/rocm-systems/actions/runs/29343301396 (prior commit: https://github.com/ROCm/rocm-systems/actions/runs/29342189788)

```
[ROCR-Runtime] /__w/rocm-systems/rocm-systems/projects/rocr-runtime/libhsakmt/src/hsakmtmodel.c:56:9: error: call to undeclared function 'secure_getenv'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
[ROCR-Runtime]    56 |         return secure_getenv(name);
[ROCR-Runtime]       |                ^
[ROCR-Runtime] /__w/rocm-systems/rocm-systems/projects/rocr-runtime/libhsakmt/src/hsakmtmodel.c:56:9: error: incompatible integer to pointer conversion returning 'int' from a function with result type 'char *' [-Wint-conversion]
[ROCR-Runtime]    56 |         return secure_getenv(name);
[ROCR-Runtime]       |                ^~~~~~~~~~~~~~~~~~~
[ROCR-Runtime] 2 errors generated.
```

Please revert PRs that break the build instead of attempting fix forwards like https://github.com/ROCm/rocm-systems/pull/8555

PR #6837, opened 2026-06-05
[RCCL] Sync to NCCL v2.30.4-1 (v2.30.3-1, v2.30.4-1)

thomas-huber:
## Summary

Syncs upstream NCCL into RCCL through **v2.30.4-1**, via two subtree merges stacked on `develop`:

- `e04ba6e66d` — Merge NCCL **v2.30.3-1** (large feature release: Device-API/GIN enhancements, Elastic Buffers, one-sided RMA `src/rma/`, new NCCL parameter C API, IB port recovery, cross-clique).
- `8ccfcd2283` — Merge NCCL **v2.30.4-1** (follow-up: `nccl_param` header fix GH #2105 + elastic-buffer support for GIN).

All changes are confined to `projects/rccl/` (310 files, +40,661 / −4,733). `makefiles/version.mk` is now `2.30.4`; `CHANGELOG.md` has RCCL-centric entries for both releases.

## RCCL-side notes / known issues

**v2.30.4-1**
- Elastic-buffer support for GIN (multi-segment symmetric windows mixing device + CPU/`HOST_NUMA` memory; `NCCL_ELASTIC_BUFFER_REGISTER`, `NCCL_SYM_REUSE_SYSMEM_HANDLES`) compiles on ROCm but is **unverified on AMD hardware**.
- `CU_MEM_LOCATION_TYPE_HOST_NUMA` (CUDA driver enum, no ROCm equivalent on ROCm < 7.12) is shimmed in `nccl_device/hip_compat.h` under `__HIP_PLATFORM_AMD__ && ROCM_VERSION < 71200`; host-NUMA branches stay inert on AMD (GIN segments are always device-backed).

**v2.30.3-1**
- One-sided RMA subsystem (`src/rma`) newly synced; uses RCCL's direct-HIP batch-memop path. Unverified at scale on ROCm.
- Upstream Copy-Engine collective redesign **not adopted**; RCCL retains its existing HIP CE implementation. CE profiler path (`ncclProfiler_v6`) not enabled — RCCL stays on `ncclProfiler_v5`.
- GIN GDAKI host support uses the shared IB context; GDAKI is DOCA/Mellanox-specific and unverified on AMD NICs.
- RCCL IB GIN proxy backend ported to the new `ncclGin_v13_t` interface; GIN GET/FLUSH (`iget`/`iflush`) not implemented.

## Test plan

[...]

PR #7792, opened 2026-06-25
[RCCL] Adapt topology and P2P routing to NCCL 2.30 DEV nodes

rahulvaidya20:
## Motivation

RCCL 2.30.4 regresses collective bandwidth ~50–70% on MI250X (gfx90a) across all_reduce/reduce/reduce_scatter/all_gather/broadcast, and ~25–35% on alltoall/gather. Bisected to the NCCL v2.30.4-1 sync (#6837). sendrecv is unaffected. The same sync left several RCCL topology/P2P functions on the pre-2.30 model.

## Technical Details

NCCL 2.30 inserts a parent DEV topology node above each GPU and moves inter-GPU XGMI links onto DEV nodes. A GPU→GPU path is now 3 hops when direct (GPU-DEV-DEV-GPU) and 4 when indirect; `type == PATH_NVL` no longer implies a direct link, and a GPU node's raw `links` no longer point GPU→GPU. Four RCCL sites were not adapted for AMD:

- **`parseRomeSystem`** (rome_models.cc) — treated every `PATH_NVL` path as direct XGMI, over-populating the Rome `connMatrix`/`nLinks`. On MI250's sparse 4P2H fabric the Rome model stopped matching → generic ring search → channels halved (4→2) → ~−58%. Fix: count direct XGMI only when `path->count <= 3`.
- **`ncclTopoCountXGMI`** (search.cc) — gated on `path->count == 1`, never true post-DEV, so it always returned 0 and the XGMI tie-breaker in `ncclTopoCompareGraphs` was dead on every AMD arch. Fix: path-based direct test (`type == PATH_NVL && count <= 3`), neighbor read from the GPU node instead of `list[0]->remNode`.
- **`parseChordalRing`** (rome_models.cc) — walked `node->links` reading `remNode->gpu.dev`; XGMI links now live on DEV nodes, so chordal-ring (CR8G/8P6L) topologies stopped matching and fell back to generic search (channels 24→4). Fix: read direct XGMI neighbors from `node->paths[GPU]` with `count <= 3`.
- **`ncclTopoCheckP2p`** (paths.cc) — the upstream NCCL code routes a non-adjacent GPU pair (`path->count == 4`) through a software GPU relay (`*intermediateRank`), i.e. NCCL's NVB-style indirect P2P. This is correct for NVLink (non-adjacent peers are not directly reachable) but wrong for AMD: the xGMI fabric routes non-adjacent pairs directly in hardware (`isAllCudaP2p == 1`), so the relay is pure overhead — it flips `isAllDirectP2p` 1→0 and sends distant alltoall/gather traffic over a 2-hop relayed copy (~−27% alltoall, ~−35% gather). Fix: guard the relay block with `#if !defined(__HIP_PLATFORM_AMD__)` so AMD uses direct P2P, restoring `isAllDirectP2p == 1`.

Path-type comparisons use `PATH_NVL` (the `ncclTopoLinkList::type` domain). All changes are intra-node only — they affect per-node XGMI topology matching, graph search, and same-node GPU-GPU P2P. Cross-node pairs return early in `ncclTopoCheckP2p` and never reach the relay block; multi-node GPU→NIC (PXN) routing uses a separate path and is untouched. No-op on fully-connected arches (gfx942/gfx950), where every GPU pair is direct (`count==3`) so the sparse/relay branches never fire.

## JIRA ID

Resolves ROCM-26913 and AICOMRCCL-1394.

## Test Plan

- MI250X, 8 GPU: rccl-tests, all collectives, 8K–1G; clean 3-way (2.29.7 / develop / develop-fixed) on one node.
- P2P relay fix: alltoall/gather/all_reduce 256M–1G vs 2.29.7 baseline, with `NCCL_DEBUG=INFO` confirming `isAllDirectP2p` 0→1 and no `P2P/indirect` routing.
- topo_expl offline replay of the 8p6l (gfx908 1H6XGMI) chordal-ring models on a gfx950 host, before vs after the parseChordalRing fix.
- Unit tests (`rccl-UnitTestsFixturesDebug --gtest_filter=TopoTest.*`): updated the stale topology fixture for the 2.30 DEV-node model and added DEV-node path-hop invariant tests (direct = 3-hop PATH_NVL, indirect = 4-hop).

## Test Result

- MI250X ring collectives: large-message busbw ~2× restored — all_reduce 60→129, broadcast 58→119, reduce 62→129, all_gather 53→113, reduce_scatter 60→119 GB/s.
- MI250X P2P collectives (P2P relay fix): alltoall 45→62 and gather 89→137 GB/s @1G — fully back to 2.29.7 (alltoall 61.8, gather 136.6); all_reduce control unchanged (127); rings unaffected.
- topo_expl 8p6l models: chordal-ring match restored — Broadcast 1G maxChannels 4 (pre-fix) → 24 (post-fix) across all six 8p6l configs.
- TopoTest suite: 9/9 pass. Broader topology suites (NetDevsPolicy/Xml/NetDev) also pass.

## Submission Checklist

### DIFF
--- a/projects/rccl/src/graph/rome_models.cc
+++ b/projects/rccl/src/graph/rome_models.cc
@@ -2137,7 +2137,7 @@ ncclResult_t parseChordalRing(struct ncclTopoSystem* system, struct ncclTopoGrap
     for (int n = 0; n<ngpus; n++) {
       // Direct XGMI neighbor: post NCCL-2.30 a direct GPU->GPU path is GPU-DEV-DEV-GPU (count<=3).
       struct ncclTopoLinkList* path = node->paths[GPU] + n;
-      if (path->type != LINK_NVL || path->count > 3) continue;
+      if (path->type != PATH_NVL || path->count > 3) continue;
       sum -= system->nodes[GPU].nodes[n].gpu.dev;
       count ++;
     }
@@ -2277,7 +2277,7 @@ static ncclResult_t parseRomeSystem(struct ncclTopoSystem* system, struct rcclRo
       struct ncclTopoLinkList *path = node->paths[GPU] + gpu_scores[n].g;
       // Only count direct XGMI links: since NCCL 2.30, GPU->GPU routes via DEV nodes, so direct is
       // count==3 and indirect count==4. Counting indirect breaks Rome matching on sparse topos.
-      if (path->type != LINK_NVL || path->count > 3) continue;
+      if (path->type != PATH_NVL || path->count > 3) continue;
       romeTopo->connMatrix[i*romeTopo->nGpus+n] = path->bw/ncclTopoXGMISpeed(node->gpu.gcn);
       count ++;
     }
--- a/projects/rccl/src/graph/rome_models.cc
+++ b/projects/rccl/src/graph/rome_models.cc
@@ -2135,12 +2135,9 @@ ncclResult_t parseChordalRing(struct ncclTopoSystem* system, struct ncclTopoGrap
     int sum = ngpus*(ngpus-1)/2 - node->gpu.dev;
     int count = 0;
     for (int n = 0; n<ngpus; n++) {
-      struct ncclTopoLink* link;
-      for (link = node->links; link->remNode; link++) {
-        if (link->remNode->gpu.dev == n) break;
-      }
-      if (!link->remNode) continue;
-      if (link->type != LINK_NVL) continue;
+      // Direct XGMI neighbor: post NCCL-2.30 a direct GPU->GPU path is GPU-DEV-DEV-GPU (count<=3).
+      struct ncclTopoLinkList* path = node->paths[GPU] + n;
+      if (path->type != LINK_NVL || path->count > 3) continue;
       sum -= system->nodes[GPU].nodes[n].gpu.dev;
       count ++;
     }
--- a/projects/rccl/src/graph/rome_models.cc
+++ b/projects/rccl/src/graph/rome_models.cc
@@ -2278,7 +2278,9 @@ static ncclResult_t parseRomeSystem(struct ncclTopoSystem* system, struct rcclRo
     for (n = 0; n < romeTopo->nGpus; n++) {
       romeTopo->connMatrix[i*romeTopo->nGpus+n] = 0;
       struct ncclTopoLinkList *path = node->paths[GPU] + gpu_scores[n].g;
-      if (path->type != LINK_NVL) continue;
+      // Only count direct XGMI links: since NCCL 2.30, GPU->GPU routes via DEV nodes, so direct is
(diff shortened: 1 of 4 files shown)


projects/rccl/src/graph/rome_models.cc
```diff
@@ -2135,12 +2135,9 @@ ncclResult_t parseChordalRing(struct ncclTopoSystem* system, struct ncclTopoGrap
...
     int count = 0;
     for (int n = 0; n<ngpus; n++) {
-      struct ncclTopoLink* link;
-      for (link = node->links; link->remNode; link++) {
-        if (link->remNode->gpu.dev == n) break;
-      }
-      if (!link->remNode) continue;
-      if (link->type != LINK_NVL) continue;
+      // Direct XGMI neighbor: post NCCL-2.30 a direct GPU->GPU path is GPU-DEV-DEV-GPU (count<=3).
+      struct ncclTopoLinkList* path = node->paths[GPU] + n;
```
alex-breslow-amd, line 2139: This raw pointer arithmetic is a bit dangerous.  Do we do any checking in tests to make sure that we can't index past the end of the paths[GPU] buffer?

projects/rccl/src/graph/search.cc
```diff
@@ -367,14 +367,11 @@ static int ncclTopoCountXGMI(struct ncclTopoSystem* system, struct ncclTopoGraph
...
-          if (node->paths[GPU][k].count == 1) {
-            struct ncclTopoLink* link = node->paths[GPU][k].list[0];
-            struct ncclTopoNode* remNode = link->remNode;
-            if (remNode->gpu.rank == n) {
-              if (link->type == LINK_NVL)
-                count ++;
-            }
-          }
+          // Direct XGMI link: post NCCL-2.30 a direct GPU->GPU path is GPU-DEV-DEV-GPU (count<=3).
+          struct ncclTopoLinkList* path = node->paths[GPU] + k;
```
alex-breslow-amd, line 371: I'm not a huge fan of this raw pointer arithmetic, but it looks in keeping with what's already there, so I won't protest.

alex-breslow-amd (approved):
LGTM, but I'd like someone else to also take a look.

projects/rccl/src/graph/rome_models.cc
rahulvaidya20: The loop is bounded by `system->nodes[GPU].count`, and `paths[GPU]` is `ncclCalloc'd` to exactly that size in `ncclTopoSetPaths` ([paths.cc:40](https://github.com/ROCm/rocm-systems/blob/develop/projects/rccl/src/graph/paths.cc#L40))

thomas-huber:
I manually ran MI350 regression checks, no issues seen.

thomas-huber (approved):
Nice work.

PR #7794, opened 2026-06-25
hipGraph: Approach A - collapse captured segments to one queue using ...

jaydeeppatel1111:
any-order overlap

Adds DEBUG_HIP_GRAPH_COLLAPSE_ANYORDER. On capable hardware (gfx1250 / gfx12.5), collapses the fully-captured (compute-queue) segments of each device onto a single reserved stream so the CP overlaps them out-of-order on that one queue.

## Submission Checklist

### DIFF
--- a/projects/clr/hipamd/src/hip_graph_internal.cpp
+++ b/projects/clr/hipamd/src/hip_graph_internal.cpp
@@ -432,6 +432,162 @@ void Graph::ResolveSegmentDependencies() {
 }
 
 // ================================================================================================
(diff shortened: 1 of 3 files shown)


jaydeeppatel1111:
> image
> Here we can do dispatch without barrier packet from multiple queues. So the general approach of collapsing to single queue is not optimal design for all types of graphs.

yeah, @saleelk suggested to extend approach B. https://github.com/ROCm/rocm-systems/pull/7795

PR #7795, opened 2026-06-25
feat(clr): Apply concurrent any-order on oversubscribed queues for round robin.

jaydeeppatel1111:
## Motivation

HIP graphs schedule independent segments across a small pool of hardware queues.
When more parallel segments land on a queue than there are queues (oversubscription),
they serialize on that in-order queue even though they're independent — leaving GPU
capacity idle. Clearing the AQL barrier bit on the head packet of the later
same-queue collisions lets capable hardware overlap them instead of serializing.

## Technical Details

- Opt-in via DEBUG_HIP_GRAPH_ANYORDER_OVERLAP. No ISA gating — the bit is cleared
  through the existing AnyOrderLaunch dispatch path and the CP decides per-ISA
  whether it reorders; incapable HW serializes as before.
- Decision lives in RoundRobinStreamAssignment: a segment head clears its barrier
  iff it is a later same-queue collision (idx >= pool) with no off-queue dependency
  and is a kernel node. Only the head packet is affected; non-head packets keep
  barrier=1, preserving intra-segment order.
- Correctness: dispatch is level-ordered and the first-on-queue head per level keeps
  barrier=1, fencing all prior-level work; cross-queue deps keep the barrier
  (fenced by a prepended BARRIER_AND).
- Guarded off under DFS assignment and single-stream collapse. Multi-device graphs
  are rejected on this path.

## JIRA ID
JIRA ID : AIRUNTIME-74

## Test Plan
Added hipGraphAnyOrderOverlap.cc (correctness + perf).

## Test Result
Correctness cases pass; Performance_Graph_AnyOrderOverlap_Perf shows a perf gain on capable HW.

## Submission Checklist


jaydeeppatel1111:
@saleelk , Once you get a chance, Can you please review this PR? Thanks!

projects/clr/hipamd/src/hip_graph_internal.cpp
```diff
@@ -386,6 +386,103 @@ void Graph::ResolveSegmentDependencies() {
...
+// Helpers for the same-queue any-order overlap pass. AQL packet headers are
+// 16-bit little-endian: byte 0 holds the 8-bit packet type, bit 8 (byte 1,
+// bit 0) is the barrier bit, and byte 2 is the AMD vendor format for
+// VENDOR_SPECIFIC packets. We edit the raw bytes directly to avoid pulling in
+// HSA header dependencies here.
+namespace {
+constexpr uint8_t kAqlTypeVendorSpecific = 0;  // HSA_PACKET_TYPE_VENDOR_SPECIFIC
+constexpr uint8_t kAqlTypeKernelDispatch = 2;  // HSA_PACKET_TYPE_KERNEL_DISPATCH
+constexpr uint8_t kAmdFormatExtDispatch = 3;   // AMD_AQL_FORMAT_KERNEL_DISPATCH (ext)
+constexpr uint8_t kBarrierBitMask = 0x01;      // bit 8 lives in byte 1, bit 0
```
saleelk: this probably need to honor the bit shifts lke we have in the rocvirtual file

projects/clr/hipamd/src/hip_graph_internal.cpp
```diff
@@ -386,6 +386,103 @@ void Graph::ResolveSegmentDependencies() {
...
+  if (barrier) {
+    pkt[1] |= kBarrierBitMask;
+  } else {
+    pkt[1] &= static_cast<uint8_t>(~kBarrierBitMask);
+  }
+}
+}  // namespace
+
+// ================================================================================================
+bool GraphExecSegmented::DeviceHonorsSameQueueAnyOrder(int dev_id) {
```
saleelk: Is this needed? Could we not apply this optimization everywhere, its then upto CP to decide if it applies

projects/clr/hipamd/src/hip_graph_internal.cpp
```diff
@@ -386,6 +386,103 @@ void Graph::ResolveSegmentDependencies() {
...
+  if (!anyorder_enabled_ || !assignment_is_round_robin_ || collapsed_to_single_stream_) {
+    return;
+  }
+  for (auto& batch : sb.packet_batches) {
+    for (uint8_t* pkt : batch.dispatchPackets) {
+      if (packetIsDispatch(pkt)) {
+        writeDispatchBarrierBit(pkt, !seg.clear_head_barrier);
+        if (seg.clear_head_barrier) {
+          ClPrint(amd::LOG_DETAIL_DEBUG, amd::LOG_CODE,
+                  "[hipGraph] Approach B: CLEARED head barrier bit for segment %d "
```
saleelk: why does the comment ahve to say approach B

projects/clr/hipamd/src/hip_graph_internal.cpp
```diff
@@ -386,6 +386,103 @@ void Graph::ResolveSegmentDependencies() {
...
+
+// Force the barrier bit of a dispatch packet to `barrier`; no-op otherwise.
+// Writing the bit explicitly (rather than only clearing) keeps the overlap pass
+// idempotent across repeated instantiation / node updates.
+void writeDispatchBarrierBit(uint8_t* pkt, bool barrier) {
+  if (!packetIsDispatch(pkt)) {
+    return;
+  }
+  if (barrier) {
+    pkt[1] |= kBarrierBitMask;
```
saleelk: not sure if we need to set it explicitly, its set by default already

projects/clr/hipamd/src/hip_graph_internal.cpp
```diff
@@ -474,30 +574,81 @@ void GraphExecSegmented::BuildSyncPlan() {
       auto level_it = segments_per_level_.find(level);
       if (level_it == segments_per_level_.end()) continue;
 
+      // Approach B: the first segment to land on each (device, stream) at this
+      // level keeps barrier=1 to fence prior-level work; later same-queue
+      // collisions may drop it. Reset per level; the loop below walks the real
+      // dispatch order (the same vector EnqueueSegmentedGraph uses).
+      std::unordered_set<uint64_t> occupiedQueuesThisLevel;
```
saleelk: is this really needed? could you not figure out from the number segment at the level as in lets say there re 4 queues and 8 segments at that level . the first 4 need the barrier bit, the next 4 RR to the same and do not need it, index/4 >=1 would mean that.. isnt it?

jaydeeppatel1111: Replaced the set with your index arithmetic. In RoundRobinStreamAssignment the per-device, per-level index already encodes it: idx >= pool (i.e. idx / pool >= 1) means a later RR collision on an already-occupied queue. Folded the whole decision into the existing RR pass (no separate walk). Kept one extra guard — !dependsOffQueue — so a later same-queue segment that also has a cross-queue dependency still keeps its barrier (it's fenced by the prepended BARRIER_AND).

projects/clr/hipamd/src/hip_graph_internal.cpp
jaydeeppatel1111: Right — removed the explicit set path (writeDispatchBarrierBit) altogether. Default dispatch already sets barrier=1; we only tag the any-order heads so the dispatch path clears it. No redundant set/write.

projects/clr/hipamd/src/hip_graph_internal.cpp
jaydeeppatel1111: Removed — all "Approach B" naming is gone from comments, logs, and the flag description.

projects/clr/hipamd/src/hip_graph_internal.cpp
jaydeeppatel1111: Agreed — removed the ISA allow-list (DeviceHonorsSameQueueAnyOrder) completely. We now clear the bit whenever the feature is enabled and let the CP decide whether it actually reorders; incapable HW just serializes as before, so it's safe everywhere.

projects/clr/hipamd/src/hip_graph_internal.cpp
jaydeeppatel1111: Ok. Dropped the hand-rolled kBarrierBitMask and raw byte-poking entirely. The head now just signals intent via the existing hipExtAnyOrderLaunch → NDRangeKernelCommand::AnyOrderLaunch flag, and the actual barrier-bit clear happens in the canonical rocvirtual dispatch path (~(1 << HSA_PACKET_HEADER_BARRIER)). So we reuse the proper shift handling instead of duplicating it.

projects/clr/hipamd/src/hip_graph_internal.cpp
```diff
@@ -1227,14 +1229,60 @@ void GraphExecSegmented::RoundRobinStreamAssignment() {
...
+      const size_t pool = getPoolSize(seg.dev_id);
+      const size_t idx = dev_idx[seg.dev_id]++;
+      seg.stream_id = static_cast<int>(idx % pool);
+
+      // Same-queue any-order overlap: a later collision (idx >= pool) with no
+      // cross-queue dep on a kernel head drops its barrier via the AnyOrderLaunch
+      // path. Collapse guard applied after ComputeCompletionSignalFlags() below.
+      seg.clear_head_barrier = false;
+      if (anyorder_enabled_ && idx >= pool && seg.first_node != nullptr &&
+          seg.first_node->GetType() == hipGraphNodeTypeKernel) {
```
saleelk: why does it need to be kernel only, its totalyl fine to clear the barrier bit regardless of dependency, the way I see it is that anythign at the same level is already determined to go in parallel. the dependencies if any are built in the barrier packet dep signals or the kernel packet itself(for gfx1250), thats handled when the levels are created.  So basically if one is going from 

for each level
{
if  seg[i].index/ max_streams >1 (as in if you go in second iteration of RR, then the next set onwards can have the first packets barrier bit cleared)
}

I dont think we shoudl need any other checks

saleelk (commented):
I want the basic thing applied first for RR only, For the collapse it needs to be a separate change. Its easier to comprehend and isolate

projects/clr/hipamd/src/hip_graph_internal.cpp
jaydeeppatel1111: Ok. Simplified and moved to BuildSyncPlan. The clear just tags later same-level collisions (idx >= pool) and clears the head barrier bit in BuildSyncPlan; the collapse work is left out as a separate change. Dropped the kernel-only and dependency checks as suggested — same-level segments are already parallel and cross-level order is held by the barrier/dep signals. The one remaining check is a packet-type gate (clear only on kernel/ext-dispatch heads), since we edit the raw captured packet that goes to HW unvalidated and mustn't touch a barrier/SDMA head.

jaydeeppatel1111:
Closing as superseded by #9895 (merged to develop).

PR #6811, opened 2026-06-04
[rocr] produce coredumps without using debugger ioctls

amd-bfilipov:
The hsa-runtime has the ability to produce a core dump when an exception is detected on the GPU.  To do so, the runtime uses the debugger ioctls to gather the necessary information (i.e. the process starts debugging itself).

However, if the rocr-debug-agentis loaded, the debug agent is already using the debugger ioctls.  When the exception reaches ROCr and it tries to connect to the debugger ioctl, things fails as there can only be one debugger of the process at a time.  As a consequence, rocr fails to produce a coredump.  There are no fundamental reasons why ROCr should not be able to produce a GPU coredump if the debug agent is active, and this restriction is preventing users from leveraging the debug agent and core dump when they need.

This series proposes to fix this.  Instead of using the debugger ioctls, it teaches ROCr how to produce a core dump using information and tools it already has available.

When the exception is received, ROCr needs to:
- gather metadata regarding the runtime itself
- gather information regarding each agent
- gather information regarding every AQL queue
- suspend all of the AQL queues (to trigger CWSR, so the wave's workload can be captured in the core dump)

With the information gathered, ROCr can create the core dump note necessary for the debugger, and with suspended queues, we can capture the state of devices.

This testsuite was tested using ROCgdb's testsuite, which is the only component capable of opening and properly validating core dump.  The testsuite ran with and without the debug agent attached, exercising the case which fails without this series.

PR #6947, opened 2026-06-08
fix(rocdbgapi): Keep waves halted when resuming with exceptions

lancesix:
When resuming a wave after it received an exception, a client debugger can do 2 things:
- swallow the exception, and call amd_dbgapi_wave_resume with exceptions=AMD_DBGAPI_EXCEPTION_NONE
- keep the exception and transmit it to the runtime, i.e. call amd_dbgapi_wave_resume with a non-0 exception.

In the second case, dbgapi currently calls
architecture_t::wave_set_state which when called with STATE_RUN clears all the exceptions marker (calls architecture_::clear_stop_reasons). This is an issue if after we resume a wave with exceptions, the runtime decides to create a core dump.  In such scenario, the runtime can correctly create a core dump, but the wave which caused the core to be created has all its exception bits cleared, making it indistinguishable from a perfectly "healthy" wave.

This patch proposes to do the following changes:
- When calling amd_dbgapi_wave_resume with a non-0 exception, we do not actually resume the wave and keep it in the stopped state.  However, from dbgapi's point of view, we record the wave as resumed (m_state is RUNNING), so if we are to see this wave again, it looks like it re-raised the same exception (but this should never happen).
- When clearing exceptions from a wave, make sure to not clear exceptions associated with the exceptions we resume with, so the runtime can still observe those bits.

Tested with ROCgdb's testsuite of gfx1031.

PR #7799, opened 2026-06-25
fix(libhsakmt): fix gpuvm base/limit returning 0 on GPUVM64 dGPUs

amd-junhshen:
## Motivation

On GPUVM64 dGPUs (canonical address space mode), querying FMM_GPUVM via hsakmt_fmm_get_apert_base_and_limit() returns 0 for both gpuvm_base and gpuvm_limit. This breaks any consumer that needs the actual GPU virtual memory address range, including [core dump](https://github.com/ROCm/rocm-systems/pull/6811) device info collection.

## Technical Details

In GPUVM64 mode, `gpuvm_aperture` (a manageable_aperture_t) is intentionally left unpopulated to avoid conflicts with the SVM aperture's allocation management (rbtree, mutex, vm_find_object() lookups). However, the kernel reports valid gpuvm base/limit values in process_apertures — they were simply being discarded.

This fix adds a lightweight `aperture_t gpuvm_range` field to gpu_mem_t that unconditionally caches the kernel-returned gpuvm base/limit during topology init. The `FMM_GPUVM` query path now reads directly from gpuvm_range instead of gpuvm_aperture, providing correct boundaries without interfering with SVM aperture's allocation management.

## JIRA ID

JIRA ID :  NO_EXISTED-000

## Test Plan

kfdtest

## Test Result

- gpuvm_base/gpuvm_limit correctly return the kernel-reported range on GPUVM64 dGPU
- Core dump collection succeeds with valid gpuvm boundaries
- kfdtest passed with no regressions observed

### DIFF
--- a/projects/rocr-runtime/libhsakmt/src/fmm.c
+++ b/projects/rocr-runtime/libhsakmt/src/fmm.c
@@ -207,7 +207,8 @@ typedef struct {
 						 * dgpu_aperture. When requested by RT, each
 						 * GPU will get a differnt range
 						 */
-	manageable_aperture_t gpuvm_aperture;   /* used for GPUVM on APU, outsidethe canonical address range */
+	manageable_aperture_t gpuvm_aperture;   /* used for GPUVM on APU, outside the canonical address range */
+	aperture_t gpuvm_range;                 /* raw gpuvm base/limit from kernel, always valid */
 	int drm_render_fd;
 	uint32_t usable_peer_id_num;
 	uint32_t *usable_peer_id_array;
@@ -3074,6 +3075,11 @@ HSAKMT_STATUS hsakmt_fmm_init_process_apertures(HsaKFDContext *ctx,
 		gpu_mem[gpu_mem_id].scratch_aperture.limit =
 			PORT_UINT64_TO_VPTR(process_apertures[i].scratch_limit);
 
+		gpu_mem[gpu_mem_id].gpuvm_range.base =
+			PORT_UINT64_TO_VPTR(process_apertures[i].gpuvm_base);
+		gpu_mem[gpu_mem_id].gpuvm_range.limit =
+			PORT_UINT64_TO_VPTR(process_apertures[i].gpuvm_limit);
+
 		if (IS_CANONICAL_ADDR(process_apertures[i].gpuvm_limit)) {
 			uint64_t vm_alignment = get_vm_alignment(
 				gpu_mem[gpu_mem_id].device_id);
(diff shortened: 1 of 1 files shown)


kentrussell (commented):
Seems reasonable, but hoping to get some confirmation from @amd-yangp to make sure it won't cause any issues

amd-bfilipov:
Checked it in combination with #6811 and #6947. Works. Tests pass.

kentrussell (approved):

cfreeamd (commented):

Reviewed `libhsakmt/src/fmm.c`. The mechanism (adding `aperture_t gpuvm_range` populated unconditionally before the `IS_CANONICAL_ADDR` branch) is correct, minimal-footprint, and backward-safe for Kaveri.

projects/rocr-runtime/libhsakmt/src/fmm.c
```diff
@@ -3256,10 +3262,10 @@ HSAKMT_STATUS hsakmt_fmm_get_aperture_base_and_limit(HsaKFDContext *ctx,
 
 	switch (aperture_type) {
 	case FMM_GPUVM:
-		if (aperture_is_valid(fmm_ctx->gpu_mem[slot].gpuvm_aperture.base,
-			fmm_ctx->gpu_mem[slot].gpuvm_aperture.limit)) {
-			*aperture_base = PORT_VPTR_TO_UINT64(fmm_ctx->gpu_mem[slot].gpuvm_aperture.base);
-			*aperture_limit = PORT_VPTR_TO_UINT64(fmm_ctx->gpu_mem[slot].gpuvm_aperture.limit);
+		if (aperture_is_valid(fmm_ctx->gpu_mem[slot].gpuvm_range.base,
```
cfreeamd, line 3272: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** PR description overstates fixed call path / no test for the dGPU fix

The only in-tree caller of `FMM_GPUVM` is `topology.c:2461`, gated by `GFX_VERSION_KAVERI`. Kaveri is an APU (GFX7) with non-canonical GPUVM — the very case where `gpuvm_aperture` was already set correctly. No current in-tree code path exercises `FMM_GPUVM` for a GPUVM64 dGPU; the fix is defensive/preparatory for external consumers or `users/bfilipov/coredump-clean-v6`.

The PR description ('causes FMM_GPUVM queries to return 0...') is technically accurate in isolation but misleading about impact. **Suggestion:** Add to the commit message: which tool or call site actually observes the zero values, and note the relationship to `FMM_GPUVM_DBG` in `users/bfilipov/coredump-clean-v6`. No code change needed.

Also: no test calls `hsakmt_fmm_get_aperture_base_and_limit(FMM_GPUVM)` on a GPUVM64 dGPU and asserts non-zero results. A kfdtest case would validate the exact scenario claimed fixed.

projects/rocr-runtime/libhsakmt/src/fmm.c
```diff
@@ -207,7 +207,8 @@ typedef struct {
 						 * dgpu_aperture. When requested by RT, each
 						 * GPU will get a differnt range
 						 */
-	manageable_aperture_t gpuvm_aperture;   /* used for GPUVM on APU, outsidethe canonical address range */
+	manageable_aperture_t gpuvm_aperture;   /* used for GPUVM on APU, outside the canonical address range */
+	aperture_t gpuvm_range;                 /* raw gpuvm base/limit from kernel, always valid */
```
cfreeamd, line 211: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** Comment overstates unconditional validity

The comment `/* raw gpuvm base/limit from kernel, always valid */` is slightly overstated. The kernel *could* return 0/0 for `gpuvm_base/gpuvm_limit` on a CPU node or if a future firmware doesn't populate the field. Suggest: `/* raw gpuvm base/limit from kernel; unconditionally populated at init */`.

projects/rocr-runtime/libhsakmt/src/fmm.c
kentrussell: @lancesix said that this was required for some coredump patches. Philip also noted that it was gated on KV

lancesix: Indeed.  We have patches coming to have ROCr be able to produce coredump without using KFD's debug ioctls (this solves incompatibility between tools).  To produce the coredump, rocr needs to be able to describe the various GPU apertures, this includes GPU VM, scratch and shared.  Currently, for the GPUVM (`hsakmt_fmm_get_aperture_base_and_limit(FMM_GPUVM)`) is not returning the desired information, this is what this PR tries to fix.

The coredump PR was using another approach: define `FMM_GPUVM_DBG` only used when creating the core dumps, but if `FMM_GPUVM` can return the GPUVM range without causing other issues, that would make things easier.

amd-yangp: ok, understand the background now, this doesn't change the FMM_GPUVM API for GFX_VERSION_KAVERI, which is the only usage in current Thunk. Since this FMM_GPUVM API change is for new feature in ROCr to get GPU VM aperture base/limit, it is safe.

amd-yangp (approved):

cfreeamd (approved):

PR #7802, opened 2026-06-25
dbgapi: flags refactor/cleanups

palves:
## Motivation

This series removes the duplicated "walk the set bits of a flag value,
one at a time" loops scattered across the library, replacing them with
two small helpers in utils.h.  There is no functional change -- this is
cleanup only, the output and behavior are identical.

## Technical Details

See individual commits.

## Test Plan

Run ROCgdb gdb.rocm/ tests on gfx1030.

## Test Result

Pass.

### DIFF
--- a/projects/rocdbgapi/src/utils.h
+++ b/projects/rocdbgapi/src/utils.h
@@ -1103,6 +1103,43 @@ operator+= (T &lhs, int rhs)
   return lhs = lhs + rhs;
 }
 
+namespace utils
+{
+
+/* Build the "A | B | C" string representation of the flag set FLAGS,
+   converting each individual bit to string with ONE_FLAG_TO_STRING.
+   When FLAGS has no bit set, ONE_FLAG_TO_STRING is called once with
+   the zero value so that its "none" case is rendered.  */
+
+template <typename T, typename OneFlagToString>
+std::string
+flags_to_string (T flags, OneFlagToString one_flag_to_string)
+{
+  static_assert (is_flag_v<T>,
+                 "flags_to_string requires an is_flag-enabled type");
+
+  if (!flags)
+    return one_flag_to_string (flags);
+
+  std::string str;
+  /* The is_flag types provide operator! but no conversion to bool,
+     so test for "any bit set" with !!.  */
+  while (!!flags)
+    {
+      T one_flag = flags ^ (flags & (flags - 1));
+
+      if (!str.empty ())
+        str += " | ";
+      str += one_flag_to_string (one_flag);
+
+      flags ^= one_flag;
+    }
+
+  return str;
+}
+
+} /* namespace utils */
+
 /* Enable bitwise operations for amd_dbgapi_exceptions_t.  */
 template <> struct is_flag<amd_dbgapi_exceptions_t> : std::true_type
 {
--- a/projects/rocdbgapi/src/utils.h
+++ b/projects/rocdbgapi/src/utils.h
@@ -1106,6 +1106,25 @@ operator+= (T &lhs, int rhs)
 namespace utils
 {
 
+/* Call FUNC once for each bit (i.e., flag) set in FLAGS, lowest bit
+   first, passing it a value of type T with that single bit set.  T
+   may be an is_flag-enabled enum or a plain unsigned integer.  Does
+   nothing when FLAGS has no bit set.  */
+
+template <typename T, typename Func>
+constexpr void
+for_each_flag (T flags, Func func)
+{
+  /* The is_flag types provide operator! but no conversion to bool, so
+     test for "any bit set" with !!.  */
+  while (!!flags)
+    {
+      T one_flag = flags ^ (flags & (flags - 1));
+      func (one_flag);
+      flags ^= one_flag;
+    }
+}
+
 /* Build the "A | B | C" string representation of the flag set FLAGS,
    converting each individual bit to string with ONE_FLAG_TO_STRING.
    When FLAGS has no bit set, ONE_FLAG_TO_STRING is called once with
@@ -1122,18 +1141,12 @@ flags_to_string (T flags, OneFlagToString one_flag_to_string)
     return one_flag_to_string (flags);
 
   std::string str;
-  /* The is_flag types provide operator! but no conversion to bool,
-     so test for "any bit set" with !!.  */
-  while (!!flags)
+  for_each_flag (flags, [&] (T one_flag)
     {
-      T one_flag = flags ^ (flags & (flags - 1));
-
       if (!str.empty ())
         str += " | ";
       str += one_flag_to_string (one_flag);
-
-      flags ^= one_flag;
-    }
+    });
 
   return str;
 }
--- a/projects/rocdbgapi/src/architecture.cpp
+++ b/projects/rocdbgapi/src/architecture.cpp
@@ -1365,11 +1365,8 @@ amdgcn_architecture_t::set_exceptions (wave_t &wave, exception_mask_t mask,
 
   uint32_t trapsts;
   wave.read_register (amdgpu_regnum_t::trapsts, &trapsts);
-  while (mask != 0)
+  utils::for_each_flag (mask, [&] (exception_mask_t single_exception)
     {
-      /* Get the lowest bit that is set.  */
-      auto single_exception = mask ^ (mask & (mask - 1));
-
       /* For each exception, set or clear the corresponding bit in TRAPSTS.  */
       auto trapsts_bit = convert_exception (single_exception);
       if (trapsts_bit != 0)
@@ -1381,9 +1378,7 @@ amdgcn_architecture_t::set_exceptions (wave_t &wave, exception_mask_t mask,
         }
       else
         unhandled_exceptions |= single_exception;
-
-      mask ^= single_exception;
-    }
+    });
   wave.write_register (amdgpu_regnum_t::trapsts, trapsts);
 
   return unhandled_exceptions;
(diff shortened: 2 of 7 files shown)


projects/rocdbgapi/src/utils.h
```diff
@@ -1103,6 +1103,51 @@ operator+= (T &lhs, int rhs)
...
+    return one_flag_to_string (flags);
+
+  std::string str;
+  for_each_flag (flags, [&] (T one_flag)
+    {
+      if (!str.empty ())
+        str += " | ";
+      str += one_flag_to_string (one_flag);
+    });
+
```
sebdar, line 1150: `clang-format` seems to align the block with the opening paren : 

```
  for_each_flag (flags,
                 [&] (T one_flag)
                 {
                   if (!str.empty ())
                     str += " | ";
                   str += one_flag_to_string (one_flag);
                 });
```

which .. I'm not sure I like. I'd go with yours but we might want to format consistently with the rest of the codebase.

projects/rocdbgapi/src/utils.h
```diff
@@ -1103,6 +1103,51 @@ operator+= (T &lhs, int rhs)
...
+/* Call FUNC once for each bit (i.e., flag) set in FLAGS, lowest bit
+   first, passing it a value of type T with that single bit set.  T
+   may be an is_flag-enabled enum or a plain unsigned integer.  Does
+   nothing when FLAGS has no bit set.  */
+
+template <typename T, typename Func>
+constexpr void
+for_each_flag (T flags, Func func)
+{
+  /* The is_flag types provide operator! but no conversion to bool, so
```
sebdar, line 1118: I feel like this belongs to the `utils` namespace ?

sebdar (commented):
Thanks, this is much cleaner this way! I just want to open the discussion about formatting. I get you're trying to modify as little lines as possible (and what `clang-format` ends up formatting looks a bit weird, e.g. : 

```
  uint32_t trapsts;
  wave.read_register (amdgpu_regnum_t::trapsts, &trapsts);
  for_each_flag (mask,
                 [&] (exception_mask_t single_exception)
                 {
                   /* For each exception, set or clear the corresponding bit in
                    * TRAPSTS.  */
                   auto trapsts_bit = convert_exception (single_exception);
                   if (trapsts_bit != 0)
                     {
                       if ((exceptions & single_exception) != 0)
                         trapsts |= trapsts_bit;
                       else
                         trapsts &= ~trapsts_bit;
                     }
                   else
                     unhandled_exceptions |= single_exception;
                 });
```
.. but this just adds to the changes that would end up being overwritten if we enable auto formatting for files. Do we have a strict policy on this ?

sebdar:
Adding `LambdaBodyIntentation: OuterScope` looks a bit better, but not quite exactly what we expect : 
```
for_each_flag (mask, [&] (exception_mask_t single_exception)
{
  /* For each exception, set or clear the corresponding bit in TRAPSTS.  */
  auto trapsts_bit = convert_exception (single_exception);
  if (trapsts_bit != 0)
    {
      if ((exceptions & single_exception) != 0)
        trapsts |= trapsts_bit;
      else
        trapsts &= ~trapsts_bit;
    }
  else
    unhandled_exceptions |= single_exception;
});
```

like we're missing the extra indentation level when opening braces, from the GNU coding style

projects/rocdbgapi/src/utils.h
palves: Hmm, just to confirm, then every caller will look like "utils::for_each_flag".  Is that what we want?  E.g.:

```
  uint32_t trapsts;
  wave.read_register (amdgpu_regnum_t::trapsts, &trapsts);
  utils::for_each_flag (mask, [&] (exception_mask_t single_exception)
    {
      /* For each exception, set or clear the corresponding bit in TRAPSTS.  */
      auto trapsts_bit = convert_exception (single_exception);
      if (trapsts_bit != 0)
        {
          if ((exceptions & single_exception) != 0)
            trapsts |= trapsts_bit;
          else
            trapsts &= ~trapsts_bit;
        }
      else
        unhandled_exceptions |= single_exception;
    });
```

And what about flags_to_string?

palves:
> Adding `LambdaBodyIntentation: OuterScope` looks a bit better, but not quite exactly what we expect :

Ahah, yes, I had spent about an hour playing with clang-format around this and had tried that option too, before sending.  It doesn't show to me what you pasted (which looks like what we would want?), though.  I get this for example:

```
  for_each_flag (mask,
                 [&] (exception_mask_t single_exception)
                 {
    /* For each exception, set or clear the corresponding bit in
       TRAPSTS.  */
    auto trapsts_bit = convert_exception (single_exception);
    if (trapsts_bit != 0)
      {
        if ((exceptions & single_exception) != 0)
          trapsts |= trapsts_bit;
        else
          trapsts &= ~trapsts_bit;
      }
    else
      unhandled_exceptions |= single_exception;
  });
```

Note I wasn't just trying to minimize the delta, I was really trying to follow GDB's style, which dbgapi's was originally based on.  See https://sourceware.org/gdb/wiki/Internals%20GDB-C-Coding-Standards#Indentation_of_lambdas_as_parameters

The GDB style for lambda's was originally based on the LLVM style for the same, adjusted for GNU/GDB.  See https://llvm.org/docs/CodingStandards.html#format-lambdas-like-blocks-of-code

(which, OHOH, is designed to minimize deltas, so... yeah.  and I'm biased as I proposed that GDB rule.)

I'm surprised that clang format doesn't have better support for this.

clang-format anchors a lambda body's indent to one of two columns, either the lambda signature column, or the enclosing scope column.  The GNU-block style needs a third anchor: "indent as a block belonging to the enclosing statement", with brace at statement +2 and body at brace +2.

No combination of the current ContinuationIndenter knobs I tries manages to expresses it, as none of them treats the lambda's braces as a nested statement block of the call, AFAICT.

I think the ideal would be to fix clang-format.   I think we would need a new LambdaBodyIndentation value, e.g. Statement (or BlockNested).  The meaning would be: when a call's last argument is a multi-line lambda and BeforeLambdaBody puts the brace on its own line, indent that brace to the enclosing statement's indent + IndentWidth, and the body to brace + IndentWidth -- reusing the same path that indents an if/for/while braced block under BreakBeforeBraces.  I felt tempted to dig into the clang-format sources, but... I have enough distractions already.  And of course that wouln't be coded, accepted, released, shipped and in everyone's distros in time.  So... For a rainy day.  :-)

In the end, I gave up, and thought it best to leave it as is manually.  You're not supposed to run clang-format on the whole sources, only on the code that you're touching, with "git clang-format", or 
```
$ cat ~/bin/run-clang-format 
git diff -U0 --no-color --relative HEAD^.. | clang-format-diff -p1 -i
```

and if you're only touching a few lines within the body, then clang-format won't mess it up.  I feel like the ocasional need to indent manually when you add a new for_each_flag call is better than the super ugly output clang-format produces here.  I find it so ugly that I considered not sending the for_each_flag patch for fear of having to clang format it, even!  :-)

sebdar:
Thanks for looking into tweaking the clang-format file, it's a shame we can't get it to work. I'm okay with manually formatting as long as we stay consistent!

projects/rocdbgapi/src/utils.h
sebdar: I don't think the extra `utils::` is distracting, helps clearly identify that that's a "foreign" utils function and not something class specific. Skimming through the code I see free functions only for asserts, `fatal_error`, logging and the odd C function like `memcpy`

That would also go for flags_to_string.

projects/rocdbgapi/src/utils.h
```diff
@@ -1103,6 +1103,51 @@ operator+= (T &lhs, int rhs)
   return lhs = lhs + rhs;
 }
 
+/* Call FUNC once for each bit (i.e., flag) set in FLAGS, lowest bit
+   first, passing it a value of type T with that single bit set.  T
+   may be an is_flag-enabled enum or a plain unsigned integer.  Does
+   nothing when FLAGS has no bit set.  */
+
+template <typename T, typename Func>
```
lancesix, line 1114: Could pobably use a

```
std::enable_if_t<is_flag_v<T>, int> = 0
```

projects/rocdbgapi/src/utils.h
```diff
@@ -1103,6 +1103,51 @@ operator+= (T &lhs, int rhs)
...
+   converting each individual bit to string with ONE_FLAG_TO_STRING.
+   When FLAGS has no bit set, ONE_FLAG_TO_STRING is called once with
+   the zero value so that its "none" case is rendered.  */
+
+template <typename T, typename OneFlagToString>
+std::string
+flags_to_string (T flags, OneFlagToString one_flag_to_string)
+{
+  static_assert (is_flag_v<T>,
+                 "flags_to_string requires an is_flag-enabled type");
```
lancesix, line 1138: this is usually enforced with std::enable_if

lancesix (commented):
Thanks for doing this.

I think I had a patch along those lines at some point, thanks for pushing this through.

projects/rocdbgapi/src/utils.h
palves: I think both ways are valid, though std::enable_if is more for disabling a particular overload, which is not what we need to do here.  Since std::enable_if leads to uglier code, I'd prefer the static_assert.

palves:
In the latest update:

- move the new functions to the utils namespace.

projects/rocdbgapi/src/utils.h
palves: (Another advantage of static_assert here is that you get a clear error pointing at the right code instead of some "no candidate match" or some such which tend to leave you scratching your head about why the call isn't working.  Given the function that has "flags_" in it, there is no real risk of it masking some other function that works with non-flags types, which is what enable_if would allow.)

sebdar: I'm fine with the static_assert. Agreed that this is much more readable.

sebdar (approved):
This looks good to me. Thanks, Pedro.

lancesix (approved):
This LGTM, but CI seems to be failing. Can you check this?

PR #7803, opened 2026-06-25
refactor(rocr): Have DmaBufClose take int* and invalidate the fd after close

dayatsin-amd:
## Summary

- Change `os::DmaBufClose` to accept `int*`, close the fd when valid, and set `*dmabuf` to `-1` so callers cannot reuse a stale handle.
- Update all call sites and simplify `DestroyMemoryHandle` accordingly.

Stacks on #7786 (`users/dayatsin/close-dmabuf-handles`).

## JIRA ID

JIRA ID : ROCM-26177

## Test plan

- [ ] Build `libhsa-runtime64` on Linux
- [ ] Re-run VMM shareable-handle / IPC tests from #7786 to confirm no regressions

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/driver/kfd/amd_kfd_driver.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/driver/kfd/amd_kfd_driver.cpp
@@ -573,7 +573,7 @@ hsa_status_t KfdDriver::ImportMemoryHandle(const core::Agent& agent, core::Drive
     if (status != HSAKMT_STATUS_SUCCESS) return HSA_STATUS_ERROR;
 
     handle->handle = reinterpret_cast<uint64_t>(res.buf_handle);
-    rocr::os::DmaBufClose(res.dmabuf_fd);
+    rocr::os::DmaBufClose(&res.dmabuf_fd);
     handle->size = res.alloc_size;
     return HSA_STATUS_SUCCESS;
   }
@@ -637,7 +637,7 @@ hsa_status_t KfdDriver::CreateShareableHandle(void* va, void* mem, size_t size,
   hsa_status_t ret = ImportMemoryHandle(agent, &targetHandle, core::ShareType::DMABUF_FD,
                                         &source_handle, mem);
 #if defined(__linux__)
-  rocr::os::DmaBufClose(source_fd);
+  rocr::os::DmaBufClose(&source_fd);
 #endif
   if (ret != HSA_STATUS_SUCCESS)
     return ret;
@@ -672,11 +672,7 @@ hsa_status_t KfdDriver::CreateShareableHandle(void* va, void* mem, size_t size,
 }
 
 hsa_status_t KfdDriver::DestroyMemoryHandle(core::DriverMemoryHandle* handle) {
-  if (handle->dmabuf_fd >= 0) {
-    hsa_status_t status = rocr::os::DmaBufClose(handle->dmabuf_fd);
-    handle->dmabuf_fd = -1;
-    if (status != HSA_STATUS_SUCCESS) return status;
-  }
+  rocr::os::DmaBufClose(&handle->dmabuf_fd);
 
   auto memhandle = reinterpret_cast<HsaMemoryObjectHandle>(handle->handle);
   if (memhandle != nullptr) {
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/driver/kfd/amd_kfd_driver.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/driver/kfd/amd_kfd_driver.cpp
@@ -573,7 +573,7 @@ hsa_status_t KfdDriver::ImportMemoryHandle(const core::Agent& agent, core::Drive
     if (status != HSAKMT_STATUS_SUCCESS) return HSA_STATUS_ERROR;
 
     handle->handle = reinterpret_cast<uint64_t>(res.buf_handle);
-    rocr::os::DmaBufClose(&res.dmabuf_fd);
+    if (rocr::os::DmaBufClose(&res.dmabuf_fd) != HSA_STATUS_SUCCESS) return HSA_STATUS_ERROR;
     handle->size = res.alloc_size;
     return HSA_STATUS_SUCCESS;
   }
@@ -672,7 +672,7 @@ hsa_status_t KfdDriver::CreateShareableHandle(void* va, void* mem, size_t size,
 }
 
 hsa_status_t KfdDriver::DestroyMemoryHandle(core::DriverMemoryHandle* handle) {
-  rocr::os::DmaBufClose(&handle->dmabuf_fd);
+  hsa_status_t ret = rocr::os::DmaBufClose(&handle->dmabuf_fd);
 
   auto memhandle = reinterpret_cast<HsaMemoryObjectHandle>(handle->handle);
   if (memhandle != nullptr) {
@@ -682,7 +682,7 @@ hsa_status_t KfdDriver::DestroyMemoryHandle(core::DriverMemoryHandle* handle) {
     }
   }
   *handle = {};
-  return HSA_STATUS_SUCCESS;
+  return ret;
 }
 
 hsa_status_t KfdDriver::SPMAcquire(uint32_t preferred_node_id) const {
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa_ext_amd.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa_ext_amd.cpp
@@ -1844,6 +1844,7 @@ hsa_status_t hsa_amd_portable_export_dmabuf_v2(const void* ptr, size_t size,
 }
 
 hsa_status_t hsa_amd_portable_close_dmabuf(int dmabuf) {
+  /* dmabuf is passed by value; caller's copy is not zeroed — ABI constraint */
   TRY;
   return rocr::os::DmaBufClose(&dmabuf);
   CATCH;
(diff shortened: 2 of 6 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/util/lnx/os_linux.cpp
```diff
@@ -880,9 +880,13 @@ bool MapMemory(void* va, size_t size, MemProt perms, int fd, uint64_t cpu_addr)
...
 
-hsa_status_t DmaBufClose(int dmabuf) {
-  if (dmabuf < 0) return HSA_STATUS_SUCCESS;
-  return ::close(dmabuf) == 0 ? HSA_STATUS_SUCCESS : HSA_STATUS_ERROR_RESOURCE_FREE;
+hsa_status_t DmaBufClose(int* dmabuf) {
+  if (dmabuf == nullptr) return HSA_STATUS_ERROR_INVALID_ARGUMENT;
+  if (*dmabuf < 0) return HSA_STATUS_SUCCESS;
+  hsa_status_t status =
+      ::close(*dmabuf) == 0 ? HSA_STATUS_SUCCESS : HSA_STATUS_ERROR_RESOURCE_FREE;
+  *dmabuf = -1;
```
ypapadop-amd, line 892: If it fails to be closed, shouldn't we keep the existing value?

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
```diff
@@ -4163,10 +4163,9 @@ Runtime::MemoryHandle::~MemoryHandle() {
   if (driver_handle.handle != 0 && region != nullptr)
     agentOwner()->driver().DestroyMemoryHandle(&driver_handle);
 
-  if (driver_handle.dmabuf_fd >= 0) {
-    os::DmaBufClose(driver_handle.dmabuf_fd);
-    driver_handle.dmabuf_fd = -1;
-  }
+  /* FIXME: the Driver class should close the dmabuf_fd, but in case of imported handles,
+   * we do not have a region so we do not have an agentOwner() to call the driver.*/
+  os::DmaBufClose(&driver_handle.dmabuf_fd);
```
ypapadop-amd: Can we have the `DestroyMemoryHandle` close the fd and if we don't have a region, then explicitly close it in runtime.cpp?

projects/rocr-runtime/runtime/hsa-runtime/core/driver/kfd/amd_kfd_driver.cpp
```diff
@@ -682,11 +682,7 @@ hsa_status_t KfdDriver::CreateShareableHandle(void* va, void* mem, size_t size,
 }
 
 hsa_status_t KfdDriver::DestroyMemoryHandle(core::DriverMemoryHandle* handle) {
-  if (handle->dmabuf_fd >= 0) {
-    hsa_status_t status = rocr::os::DmaBufClose(handle->dmabuf_fd);
-    handle->dmabuf_fd = -1;
-    if (status != HSA_STATUS_SUCCESS) return status;
-  }
+  rocr::os::DmaBufClose(&handle->dmabuf_fd);
```
ypapadop-amd: Is it ok to ignore the return value here?

ypapadop-amd (changes requested):

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa_ext_amd.cpp
```diff
@@ -1845,7 +1845,7 @@ hsa_status_t hsa_amd_portable_export_dmabuf_v2(const void* ptr, size_t size,
 
 hsa_status_t hsa_amd_portable_close_dmabuf(int dmabuf) {
   TRY;
-  return rocr::os::DmaBufClose(dmabuf);
+  return rocr::os::DmaBufClose(&dmabuf);
```
cfreeamd, line 1849: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]**

`hsa_amd_portable_close_dmabuf` takes `int dmabuf` by value (ABI-fixed public API), so `DmaBufClose(&dmabuf)` writes `-1` back to a local stack copy — the caller's variable is never invalidated. The fd is closed correctly, but the "invalidate after close" guarantee that motivates this PR is silently not delivered to users of this public HSA API.

Two options:
1. Accept the ABI limitation and add a comment explicitly noting it: `/* dmabuf is passed by value; caller's copy is not zeroed — ABI constraint */`
2. If a `hsa_amd_portable_close_dmabuf_v2(int*)` variant is acceptable, expose the pointer form at the API level.

At minimum, a comment is needed so the next reader doesn't assume the write-back reaches the caller.

cfreeamd (commented):
os.h, os_linux.cpp, os_win.cpp, amd_kfd_driver.cpp, and runtime.cpp look correct; see inline comments for one call-site issue in hsa_ext_amd.cpp and a pre-existing fd-leak in IPCClientImport.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
```diff
@@ -1624,7 +1624,7 @@ int Runtime::IPCClientImport(uint32_t conn_handle, uint64_t dmabuf_fd_handle,
       } else {
         HSAKMT_CALL(hsaKmtMemHandleFreePreserveMetadata(res.buf_handle));
       }
-      os::DmaBufClose(static_cast<int>(dmabuf_fd));
+      os::DmaBufClose(&dmabuf_fd);
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]**

Pre-existing fd leak: when `hsaKmtRegisterGraphicsHandleToNodesExt` fails, `dmabuf_fd` has been received from the kernel but is not closed before `return err`. This predates the PR, but since the surrounding code is already being touched, adding `os::DmaBufClose(&dmabuf_fd);` before `return err` in this error branch would be a low-cost fix.

projects/rocr-runtime/runtime/hsa-runtime/core/util/lnx/os_linux.cpp
cfreeamd, line 892: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]**

`*dmabuf` is set to `-1` unconditionally after `::close()`, even when `close()` returns an error. This is probably intentional — a failed `close` on Linux still releases the fd number in most practical cases — but a comment would help document the intent: `/* Set to -1 even on close failure: the fd is no longer valid regardless of errno. */`

dayatsin-amd: My reasoning was that if we fail to close, then either this is a fd that was already closed, or something went wrong at the OS level. In either case, the fd is no longer valid and userspace should consider the fd as invalid.

dayatsin-amd: Added comment

projects/rocr-runtime/runtime/hsa-runtime/core/driver/kfd/amd_kfd_driver.cpp
dayatsin-amd: Ack. I think it is better to return the return-value.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa_ext_amd.cpp
dayatsin-amd: ACK

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
dayatsin-amd: ACK

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
dayatsin-amd: Ok. I moved this into an "else".

projects/rocr-runtime/runtime/hsa-runtime/core/driver/kfd/amd_kfd_driver.cpp
```diff
@@ -573,7 +573,7 @@ hsa_status_t KfdDriver::ImportMemoryHandle(const core::Agent& agent, core::Drive
     if (status != HSAKMT_STATUS_SUCCESS) return HSA_STATUS_ERROR;
 
     handle->handle = reinterpret_cast<uint64_t>(res.buf_handle);
-    rocr::os::DmaBufClose(res.dmabuf_fd);
+    if (rocr::os::DmaBufClose(&res.dmabuf_fd) != HSA_STATUS_SUCCESS) return HSA_STATUS_ERROR;
```
ypapadop-amd, line 576: Do we need to return the error that DmaBufClose returns here?

ypapadop-amd (approved):

ypapadop-amd:
LGTM

projects/rocr-runtime/runtime/hsa-runtime/core/driver/kfd/amd_kfd_driver.cpp
dayatsin-amd: I was thinking that this could help detect an unexpected call flow or bad state with the fd earlier

ypapadop-amd: My comment is very misleading, I'm sorry. I meant that should we do something like:

```cpp
auto err = rocr::os::DmaBufClose(res.dmabuf_fd);
if (err != HSA_STATUS_SUCCESS) {
  return err;
}
```

or do we not care about the error code from `DmaBufClose` and we want to swallow it?

cfreeamd (approved):

PR #7811, opened 2026-06-25
[rocjitsu] Use composite actions

amd-eochoalo:
## Motivation

With GitHub's composite action we can place GitHub action files in different directories other than .github/workflow. This allows dependabot updates to target specific directories that can be taken care by specific teams. Right now, there is this dependabot PR https://github.com/ROCm/rocm-systems/pull/7771 which has many different files owned by many different teams. By using composite actions dependabot can update in a single PR those which fall under the rocjistu/actions/ path. And can be updated by any member of the rocjitsu-core-team.

## Technical Details

Unfortunately, composite actions require that minimal .github/workflow/rocjitsu-*.yml files exist and these still use the checkout action. So dependabot, will still target these along with all the other ones.

## JIRA ID

None

## Test Plan

Test on CI

## Test Result

Ci passes.

## Submission Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,8 +6,20 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    exclude-paths:
+      - "emulation/rocjitsu/actions/**"
     groups:
       github-actions:
+        patterns:
+          - "*"
+
+  # Check for updates to RocJitsu GitHub Actions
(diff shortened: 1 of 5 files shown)


atgutier (approved):
LGTM, but I'm not an expert on CI.

amd-eochoalo:
Thanks Tony! I'm asking on gardening for some feedback!

Issue #7808, opened 2026-06-25
[Issue] Linux gfx94X-dcgpu, gfx950: rocprofiler-compute profile-mode tests fail — `PROFILE MODE DEPENDENCY VIOLATION: Forbidden package rocm_sdk (amdsmi_wrapper import)`


amd-chiranjeevi:
## Context

- **PR:** https://github.com/ROCm/TheRock/pull/6132
- **Workflow:** Linux::release / Test gfx94X-dcgpu / Test rocprofiler-compute
- **Failing run:** https://github.com/ROCm/TheRock/actions/runs/28149435342
- **Platform:** Linux
- **Impacted Arch:** gfx94X-dcgpu (gfx942) , gfx950

## Full log:
- https://github.com/ROCm/TheRock/actions/runs/28149435342/job/83413403455?pr=6132
- https://github.com/ROCm/TheRock/actions/runs/28149435342/job/83413404176?pr=6132

## Summary

All rocprofiler-compute profile-mode tests that import `amdsmi` fail with a `PROFILE MODE DEPENDENCY VIOLATION` because `amdsmi_wrapper.py` now attempts `import rocm_sdk` inside `find_smi_library()`. The rocprofiler-compute `ProfileModeImportGuard` intercepts this import — `rocm_sdk` is not in `ALLOWED_ROCM_MODULES` — raises an `ImportError`, and `amdsmi` fails to load, causing every affected test to exit with code 1.

**5 CTest suites failed (21 individual tests):**
- `test_profile_dispatch` — `test_dispatch_0`, `test_dispatch_0_1`, `test_dispatch_2`
- `test_profile_sort` — `test_roof_sort_dispatches`, `test_roof_sort_kernels`
- `test_profile_path` — `test_path`, `test_path_rocflop`, `test_path_no_native`, `test_path_rocpd`, `test_path_csv`, `test_output_directory_hostname`, `test_output_directory_rank_ignored_without_mpi`, `test_output_directory_rank_replaced_with_mpi`, `test_output_directory_env_variable`, `test_output_directory_env_variable_unset`
- `test_profile_sets_func` — `TestSetsIntegration::test_set_profiling[compute_thruput_util]`, `TestSetsIntegration::test_set_profiling[compute_thruput_flops]`, `TestSetsIntegration::test_set_profiling[mem_thruput]`, `TestSetsIntegration::test_set_profiling[launch_stats]`
[...]

PR #7813, opened 2026-06-25
Adding profiler tests for core builds

ammallya:
## Motivation

Adding profiler tests to core build.


jayhawk-commits (approved):
This would help catch https://github.com/ROCm/rocm-systems/issues/7808 sooner.

PR #7814, opened 2026-06-25
[rocprofiler-compute] improve yaml formatting

xuchen-amd:
## Motivation

This PR standardizes the formatting of metric equations in YAML configuration files across all GPU architectures (gfx115x, gfx942, gfx950).

**No functional changes** - this is purely formatting/style improvements to improve readability and maintainability.

Metric equations had inconsistent formatting that made them harder to read and review, no automated checks to prevent future inconsistencies

### 1. New Formatting Tool

Added `tools/format_yaml.py` - a Python formatter for YAML files.

### 2. Pre-commit Hook

Added automatic formatting check that runs on every commit to maintain consistency.

### 3. Applied Formatting

Updated YAML formatting, example changes

**Before:**
```yaml
vL1D Cache BW:
  value: SUM(TCP_TOTAL_CACHE_ACCESSES_sum * 128) / SUM(End_Timestamp - Start_Timestamp)
```

After:
```yaml
vL1D Cache BW:
  value: 128 * SUM(TCP_TOTAL_CACHE_ACCESSES_sum) / SUM(End_Timestamp - Start_Timestamp)
```

## JIRA ID

N/A

## Test Plan

N/A

## Test Result

N/A

## Submission Checklist


projects/rocprofiler-compute/tools/test_format_yaml.py
```diff
@@ -0,0 +1,142 @@
+#!/usr/bin/env python3
```
vedithal-amd, line 1: Suggest moving this to `tools/` alongside `format_yaml.py` (e.g. `tools/test_format_yaml.py`) and dropping the CTest entry in `CMakeLists.txt`. `tests/` is for production code (src Python + C++ under src/lib); developer tooling and pre-commit hooks aren't production code, so their tests belong next to the tool and can be run manually, as with `tools/test_counter_grouping_inspector_manual.py`.

projects/rocprofiler-compute/.pre-commit-config.yaml
```diff
@@ -81,3 +81,16 @@ repos:
         files: ^projects/rocprofiler-compute/
         pass_filenames: false
         stages: [pre-commit]
+
+  # Local hook: YAML and metric equation formatting
+  - repo: local
+    hooks:
+      - id: yaml-format-check
+        name: YAML metric equation formatting
```
vedithal-amd, line 89: This hook will silently trip contributors (and AI tools) who don't know the equation formatting rules it enforces (constant factoring out of aggregations, minimal parentheses, spacing, `Peak (Empirical)` as a literal, etc.). Please document these rules in a contributing guide or README section and hook that doc into the `AGENTS.md` redirect chain, so both humans and AI agents know the conventions up front instead of discovering them via a failed commit.

projects/rocprofiler-compute/.pre-commit-config.yaml
```diff
@@ -81,3 +81,16 @@ repos:
...
+        entry: python projects/rocprofiler-compute/tools/format_yaml.py --check
```
vedithal-amd: Consider running this hook with `--fix` instead of `--check`, to match the `ruff-check`/`ruff-format` hooks above which auto-fix in place. The hook would then correct equations itself and pre-commit's standard "files were modified" message tells the user to re-stage, so contributors don't have to know to run `--fix` manually.

vedithal-amd (changes requested):
Requesting changes per inline comments:
- Move `test_format_yaml.py` to `tools/` alongside the module (tooling tests don't belong in the production `tests/`/CTest suite).
- Run the pre-commit hook with `--fix` to match the ruff convention.
- Document the equation formatting rules and hook them into the AGENTS.md redirect chain.

projects/rocprofiler-compute/tools/test_format_yaml.py
xuchen-amd: changed in: fa0344ee71a5567937eda625f0d9c94981bb5bff

moved tooling test to under `tools/`, added production test cases (to verify analysis config format)

projects/rocprofiler-compute/.pre-commit-config.yaml
xuchen-amd: added a yaml equation format md and its references: a331661acdbaf588e29239e792044b3e4989bc49

projects/rocprofiler-compute/.pre-commit-config.yaml
xuchen-amd: done: 949482cafb1f69406089228ebc23cb4ee437aaf1

vedithal-amd:
Pushed a follow-up commit (0637cec8177) addressing the test-placement and hook-scoping feedback:

- **Removed `test_analysis_config_yaml_format` from the CTest suite.** Formatter output is developer tooling; enforcing it belongs in the pre-commit hook, not the production unit-test suite. This also drops the `--diff`-per-file `subprocess` gate that duplicated the hook.
- **Pinned `pyyaml==6.0.3`** in the hook (was `>=6.0`) to match `requirements.txt` and the existing hook, so equation parsing stays reproducible across environments.
- **Scoped the hook `files:` regex** to the two `gfx` config dirs and dropped the blanket-include + `exclude:` list. `format_yaml.py` already self-filters, so the exclude list was redundant maintenance debt.
- **Folded the equation rules into `CONTRIBUTING.md`** and deleted `YAML_EQUATION_FORMAT.md` + `.ai/rules/yaml-equation-format.md`. The `AGENTS.md` redirect chain already points at `CONTRIBUTING.md`, so no separate shim is needed.

Verified: hook runs green on all 130 gfx configs; `tools/test_format_yaml.py` 22/22 pass.

vedithal-amd (approved):
LGTM

projects/rocprofiler-compute/.pre-commit-config.yaml
```diff
@@ -81,3 +81,15 @@ repos:
...
         stages: [pre-commit]
+
+  # Local hook: YAML and metric equation formatting
+  - repo: local
+    hooks:
+      - id: yaml-format-fix
+        name: YAML metric equation formatting
+        entry: python projects/rocprofiler-compute/tools/format_yaml.py --fix
+        language: python
+        additional_dependencies: [pyyaml==6.0.3]
```
cfallows-amd, line 92: not specific to this PR but related, the pyyaml requirement in docs/sphinx /requirements.txt is 6.0.2...but in this hooks file, vendored files, general requirements.txt it is 6.0.3. Can we change the sphinx file as well to sync?

Also the cmake/Dockerfile line 39 pip install grabs pyyaml...no version locking. Can we lock the version here as well in case a newer version gets released?

vedithal-amd: The cmake/Dockerfile is possibly a stale artifact which is not being used anywhere - will check with DevOps and remove in separate PR

For docs sphinx requirements - there are usually depandabot prs asking us to approve the bump in dependency versions. Ideally, the pyyaml version in docs sphinx requirements is a separate concern and does not have to be strictly coupled with pre commit hook

xuchen-amd: marking as resolved to unblock this PR.
Will follow up on
> The cmake/Dockerfile is possibly a stale artifact which is not being used anywhere - will check with DevOps and remove in separate PR

PR #7815, opened 2026-06-25
rocdbgapi: add .gitignore and .gitattribute files

aktemur:
Add git files for developer convenience.

## Motivation

Avoid seeing `...~` files in git status and display `git diff` better using C++ context.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 .cline_storage
 /projects/hip/_build
 
+# Emacs backup files.
+*~
+
 # Visual Studio project files
 *.suo
 *.user
--- a/projects/rocdbgapi/.gitattributes
+++ b/projects/rocdbgapi/.gitattributes
@@ -1,0 +1,4 @@
+# Use cpp-style diffs.
+*.h     diff=cpp
+*.h.in  diff=cpp
+*.cpp   diff=cpp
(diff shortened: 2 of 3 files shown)


lumachad:
Should these go into the root directory of rocm-systems instead since rocdbgapi is now part of it?

Also, we might want similar changes to rocr-debug-agent too while at it. Might be convenient.

aktemur:
> Should these go into the root directory of rocm-systems instead since rocdbgapi is now part of it?

I split the patches to add `*~` to gitignore at the top level and to add gitattributes to rocdbgapi and rocr-debug-agent. I'm not sure if the gitattributes that is being added would apply to all the projects. Hence, I preferred to be conservative.

.gitignore
```diff
@@ -1,5 +1,6 @@
 .cline_storage
 /projects/hip/_build
+*~
```
lancesix, line 5: Might be worth a comment saying those are emacs backup files.

aktemur: > Might be worth a comment saying those are emacs backup files.

It was mentioned in the commit message. I also added it as a comment to the file.

lumachad (approved):
LGTM thanks.

lancesix (approved):

PR #7821, opened 2026-06-25
rocdbgapi: use bit_extract instead of bit_mask

aktemur:
This is a small code simplification.  No behavioral change should be seen.


sebdar (approved):
This looks good, thanks for the clean-up.

PR #7831, opened 2026-06-25
amdsmi: link libamd_smi with -Bsymbolic-functions to stop rsmi_init 

### LINKED ISSUE #7643: [Issue]: `amdsmi` returns 0 GPU handles when `librocm_smi64.so.1` is loaded with `RTLD_GLOBAL`
## Summary

Loading `librocm_smi64.so.1` with `RTLD_GLOBAL` before importing the `amdsmi` Python package causes `amdsmi_get_processor_handles()` to return **0 handles**, even on a system with 8 GPUs. This makes `torch.cuda.device_count()` return 0 on ROCm, breaking any downstream code that relies on it.

The real-world trigger is simply `pip install amdsmi` on a system where PyTorch is installed via the `rocm_sdk` pip packages (TheRock). The `rocm_sdk` package preloads `librocm_smi64.so.1` with `RTLD_GLOBAL` during `import torch`, and PyTorch then uses `amdsmi` for `device_count()` on ROCm — which silently returns 0 due to the symbol conflict.

> `/opt/rocm` was set as a symbolic link to `$(rocm-sdk path --root)`.  ROCm version: 7.14.0a20260612.

### Operating System

Ubuntu 24.04.4 LTS (Noble Numbat)

### CPU

AMD EPYC 9575F 64-Core Processor

### GPU

8x AMD Instinct MI355X (gfx950)

### ROCm Version

ROCm 7.14.0a20260612 from nightly build

### ROCm Component

amdsmi

### Reproducer

```python
"""
Reproducer: librocm_smi64.so.1 loaded with RTLD_GLOBAL breaks amdsmi.

Prerequisites:
    pip install amdsmi

Usage:
    python repro.py
"""
import ctypes
import os
import subprocess
import sys

ROCM_SMI_LIB = None
for candidate in [
    os.path.join(
        os.environ.get("ROCM_HOME", os.environ.get("ROCM_PATH", "")),
        "lib/librocm_smi64.so.1",
    ),
    "librocm_smi64.so.1",
]:
    if os.path.exists(candidate):
        ROCM_SMI_LIB = candidate
        break

if ROCM_SMI_LIB is None:
    try:
        ctypes.CDLL("librocm_smi64.so.1")
        ROCM_SMI_LIB = "librocm_smi64.so.1"
    except OSError:
        print("SKIP: librocm_smi64.so.1 not found")
        sys.exit(0)


def run_case(label, code):
    result = subprocess.run(
        [sys.executable, "-c", code],
        capture_output=True,
        text=True,
    )
    output = result.stdout.strip()
    ok = "PASS" if "handles=8" in output or "handles=7" in output else "FAIL"
    print(f"  [{ok}] {label}: {output}")
    if result.returncode:
        for line in result.stderr.strip().splitlines()[-3:]:
            print(f"         {line}")


print(f"librocm_smi64.so.1 path: {ROCM_SMI_LIB}")
print()

run_case(
    "amdsmi alone (baseline)",
    """
import amdsmi
amdsmi.amdsmi_init()
h = amdsmi.amdsmi_get_processor_handles()
print(f"handles={len(h)}")
amdsmi.amdsmi_shut_down()
""",
)

run_case(
    "librocm_smi64 RTLD_LOCAL then amdsmi",
    f"""
import ctypes
ctypes.CDLL("{ROCM_SMI_LIB}", mode=0)
import amdsmi
amdsmi.amdsmi_init()
h = amdsmi.amdsmi_get_processor_handles()
print(f"handles={{len(h)}}")
amdsmi.amdsmi_shut_down()
""",
)

run_case(
    "librocm_smi64 RTLD_GLOBAL then amdsmi  <-- BUG",
    f"""
import ctypes
ctypes.CDLL("{ROCM_SMI_LIB}", mode=256)
import amdsmi
amdsmi.amdsmi_init()
h = amdsmi.amdsmi_get_processor_handles()
print(f"handles={{len(h)}}")
""",
)

run_case(
    "import torch then amdsmi  <-- real-world trigger",
    """
import torch
import amdsmi
amdsmi.amdsmi_init()
h = amdsmi.amdsmi_get_processor_handles()
print(f"handles={len(h)}")
""",
)

run_case(
    "torch.cuda.device_count() vs _cuda_getDeviceCount()",
    """
import torch
dc = torch.cuda.device_count()
raw = torch._C._cuda_getDeviceCount()
print(f"handles=device_count={dc} _cuda_getDeviceCount={raw}")
""",
)

print()
print("Expected: all cases should report 8 handles.")
print("Bug: cases 3-5 report 0 because RTLD_GLOBAL on librocm_smi64.so.1")
print("     pollutes symbols that break libamd_smi.so's amdsmi_init().")
```

### Reproducer output

```
librocm_smi64.so.1 path: /opt/venv/lib/python3.12/site-packages/_rocm_sdk_devel/lib/librocm_smi64.so.1

  [PASS] amdsmi alone (baseline): handles=8
  [PASS] librocm_smi64 RTLD_LOCAL then amdsmi: handles=8
  [FAIL] librocm_smi64 RTLD_GLOBAL then amdsmi  <-- BUG: handles=0
  [FAIL] import torch then amdsmi  <-- real-world trigger: handles=0
  [FAIL] torch.cuda.device_count() vs _cuda_getDeviceCount(): handles=device_count=0 _cuda_getDeviceCount=8

Expected: all cases should report 8 handles.
Bug: cases 3-5 report 0 because RTLD_GLOBAL on librocm_smi64.so.1
     pollutes symbols that break libamd_smi.so's amdsmi_init().
```

<details>

```
Paste output here
```

</details>

### Additional Information

> Generated by Claude Code.

### Root cause analysis

The bug involves three components interacting:

**1. `rocm_sdk` preloads `librocm_smi64.so.1` with `RTLD_GLOBAL`**

During `import torch`, the call chain is:

```
torch.__init__
  → torch._rocm_init.initialize()
    → rocm_sdk.initialize_process()
      → rocm_sdk.preload_libraries()
        → ctypes.CDLL("librocm_smi64.so.1", mode=256)   # RTLD_GLOBAL
```

This exports all symbols from the old ROCm SMI library into the global symbol namespace.

**2. `torch.cuda.__init__` then imports `amdsmi`, which loads `libamd_smi.so`**

`libamd_smi.so` (the new AMD SMI library) and `librocm_smi64.so.1` (the old ROCm SMI library) share overlapping symbol names. Because `librocm_smi64.so.1` was loaded with `RTLD_GLOBAL`, the dynamic linker resolves the conflicting symbols from the old library rather than from `libamd_smi.so` itself.

As a result, `amdsmi_init()` silently "succeeds" (no exception), but `amdsmi_get_processor_handles()` returns 0 handles.

**3. `torch.cuda.device_count()` trusts the amdsmi result**

On ROCm, `torch.cuda.device_count()` calls `_device_count_amdsmi()` → `_raw_device_count_amdsmi()`. When this returns 0 (a non-negative value), PyTorch uses it as-is instead of falling through to `torch._C._cuda_getDeviceCount()` (which correctly returns 8 via the HIP runtime).

### User-visible symptoms

```
$ python bench_amd_deterministic_allreduce.py
  ...
  File "sgl_kernel/load_utils.py", line 22, in _get_compute_capability
    properties = torch.cuda.get_device_properties(device)
  AssertionError: Invalid device id
```

Any code that calls `torch.cuda.device_count()`, `torch.cuda.current_device()`, or `torch.cuda.get_device_properties()` will see 0 devices or crash.

### Why it works without `amdsmi` installed

When `amdsmi` is not installed, PyTorch sets `_HAS_PYNVML = False`. Then `_device_count_amdsmi()` returns -1, causing `device_count()` to fall through to `torch._C._cuda_getDeviceCount()` which returns the correct count via the HIP runtime API.

### Possible fixes

**Option A — `rocm_sdk`: stop preloading `librocm_smi64.so.1` with `RTLD_GLOBAL`**

`librocm_smi64.so.1` is the legacy library superseded by `libamd_smi.so`. If it must be loaded, use `RTLD_LOCAL` (mode=0) to prevent symbol pollution.

**Option B — `libamd_smi.so` / `librocm_smi64.so.1`: resolve the symbol conflict**

Use symbol versioning or hidden visibility to prevent overlapping symbols between the two libraries from interfering with each other.

**Option C — `torch.cuda`: treat 0-handle amdsmi result as unreliable**

In `_raw_device_count_amdsmi()`, if `amdsmi_get_processor_handles()` returns 0, return -1 to fall through to the HIP runtime:

```python
def _raw_device_count_amdsmi():
    if not _HAS_PYNVML:
        return -1
    try:
        amdsmi.amdsmi_init()
    except amdsmi.AmdSmiException:
        return -1
    count = len(amdsmi.amdsmi_get_processor_handles())
    return count if count > 0 else -1  # don't trust 0, fall through to HIP runtime
```

### Current workaround

```
pip uninstall amdsmi
```

This makes `_HAS_PYNVML = False`, bypassing the broken amdsmi code path entirely. `torch.cuda.device_count()` then uses `torch._C._cuda_getDeviceCount()` which works correctly.

Samiii777:
libamd_smi.so statically embeds rocm_smi and internally PLT-calls rsmi_init()/rsmi_is_P2P_accessible() (amd_smi_system.cc, amd_smi.cc) while also exporting those symbols at default visibility for external consumers. When librocm_smi64.so.1 (which also exports rsmi_init) is dlopen'd with RTLD_GLOBAL before libamd_smi.so loads -- exactly what rocm_sdk/TheRock does during 'import torch' -- ELF global-scope interposition redirects amd_smi's internal 'call rsmi_init@plt' to librocm_smi64's rsmi_init. The wrong rsmi_init initializes a different global state, so device discovery enumerates 0 devices and amdsmi_get_processor_handles() returns 0.

Add -Bsymbolic-functions so amd_smi's own references to globally-defined functions bind to its own definitions at link time; internal rsmi calls can no longer be interposed, while the symbols stay exported for external use.

Fixes: ROCm/rocm-systems#7643

### DIFF
--- a/projects/amdsmi/src/CMakeLists.txt
+++ b/projects/amdsmi/src/CMakeLists.txt
@@ -192,6 +192,12 @@ if(BUILD_BOTH_LIBS OR BUILD_SHARED_LIBS)
     )
     target_link_options(${AMD_SMI} PRIVATE "LINKER:--version-script=${VERSION_SCRIPT}")
     set_property(TARGET ${AMD_SMI} APPEND PROPERTY LINK_DEPENDS "${VERSION_SCRIPT}")
+    # Bind amd_smi's internal calls to its own embedded rsmi_init /
+    # rsmi_is_P2P_accessible at link time. Without this, dlopen'ing
+    # librocm_smi64.so.1 with RTLD_GLOBAL first (as rocm_sdk/TheRock does on
+    # `import torch`) interposes those calls and breaks device discovery
+    # (rocm-systems#7643). Symbols stay exported for external use.
+    target_link_options(${AMD_SMI} PRIVATE "LINKER:-Bsymbolic-functions")
 endif()
 
 # Add UALOE compile flags for netlink dependencies (e.g. -I.../libnl3 for


projects/amdsmi/src/CMakeLists.txt
```diff
@@ -192,6 +192,12 @@ if(BUILD_BOTH_LIBS OR BUILD_SHARED_LIBS)
     )
     target_link_options(${AMD_SMI} PRIVATE "LINKER:--version-script=${VERSION_SCRIPT}")
     set_property(TARGET ${AMD_SMI} APPEND PROPERTY LINK_DEPENDS "${VERSION_SCRIPT}")
+    # Bind amd_smi's internal calls to its own embedded rsmi_init /
+    # rsmi_is_P2P_accessible at link time. Without this, dlopen'ing
+    # librocm_smi64.so.1 with RTLD_GLOBAL first (as rocm_sdk/TheRock does on
+    # `import torch`) interposes those calls and breaks device discovery
+    # (rocm-systems#7643). Symbols stay exported for external use.
+    target_link_options(${AMD_SMI} PRIVATE "LINKER:-Bsymbolic-functions")
```
oliveiradan, line 200: @Samiii777, we are in the process of deprecating `rocm-smi` in favor of `amd-smi`. That is the reason why only `amdsmi_*` symbols get exported (hiding all others, including `rsmi_*`). 

Telling the linker to bind function references internally (so the dynamic linker resolves those at runtime via global symbol table), where `LD_PRELOAD` or another `library.so` loaded earlier can intercept that call by exporting its own function/API doesn't help with the deprecation process.

PR #5777, opened 2026-05-05
AI NIC changes for Pensando Phase 2

ajanicijamd:
## Motivation

In Phase 2, we are adding 4 new performance parameters.

## Technical Details

We are adding the following parameters:
- tx_rdma_ack_timeout - Local ACK Timeout Errors
- resp_tx_pkt_seq_err - Responder has detected a packet sequence error and issued a NAK back to the requester
- req_rx_pkt_seq_err - Requester has detected a sequence error via a NAK
- req_rx_impl_nak_seq_err - Requester has received an ACK with PSN larger than the expected PSN

## JIRA ID

- AIPROFSYST-351
- AIPROFSYST-109

## Test Plan

- Build `rocprofiler-systems` with AI NIC enabled
- Run `rocprof-sys-sample` on a program that runs long enough to collect the AI NIC metrics
  (e.g. `wget`), with collection enabled for both Perfetto and rocpd
- Verify that all AI NIC tracks (6 from Phase 1 and 4 from Phase 2) are present in `.proto` and `.db` files

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #7833, opened 2026-06-25
refactor(rocprofiler-systems): AI-NIC refactoring

dgaliffiAMD:
## Motivation

Reintroduce the NIC track-setup optimizations that were initially descoped from #5777, and fold in the review follow-ups requested on this PR.

## Technical Details

Deduplicates the NIC Perfetto policy and cleans up several adjacent issues:

- **Track setup / emit dedup** — replaces the ten repetitive per-metric `if`-blocks in `setup_counter_tracks` with a single loop over a `make_default_nic_tracks()` table, and the ten `TRACE_COUNTER` `if`-blocks in `post_process_device` with an inline function template `emit_nic_counter<CategoryTp>` (the category rides as a registered category trait so Perfetto still sees a compile-time literal — no macro). Addresses @mradosav-amd's review comment.

- **Performance** — resolves each metric's track index once before the per-sample flush loop instead of doing a `std::map::find` per metric per sample (`O(10·N·log n)` → `O(10·log n) + O(N)`); returns the default-track table by `const&` from a `static const` map instead of rebuilding it on every setup call.

- **Storage** — the per-device track map now stores only the resolved track index (`std::map<uint32_t, size_t>`); `nic_track_description` is a pure label/units row.

- **Static-analysis cleanup** — drops the anonymous namespace in the header (fixes `-Wsubobject-linkage`; helpers are `inline`), and renames the `enabled_metrics`/`supported_metrics` parameters that shadowed the type (`-Wshadow`).

All behavior-preserving: identical track names, units, categories, and emit conditions.

## JIRA ID

JIRA ID - AIPROFSYST-697

## Test Plan

 Covered by `tests/pytest/test_ainic_perf.py` (AI NIC hardware validation).

## Submission Checklist


ajanicijamd (approved):
LGTM.

Passing the test in a container with simulated AI NIC:
`pytest -m ainic test_ainic_perf.py -v`

projects/rocprofiler-systems/source/lib/rocprof-sys/library/pmc/collectors/nic/perfetto_policy.hpp
```diff
@@ -293,136 +233,46 @@ struct perfetto_policy
...
-                        counter_track::at(device_index, it->second.track_index), ts,
-                        static_cast<double>(sample.metric_values.req_rx_pkt_seq_err));
-                }
-            }
+            // Emit one counter sample per enabled metric. The category must be a
+            // compile-time literal (Perfetto resolves registered categories at build
+            // time), so this is a scoped macro rather than a runtime loop. Each row
+            // names a metric whose member is identically spelled in enabled_metrics,
+            // the metrics struct, and the *_VALUE track key.
+#define ROCPROFSYS_NIC_EMIT_COUNTER(MEMBER, VALUE_KEY, CATEGORY)                         \
```
mradosav-amd: can we use inline function instead of macro?

dgaliffiAMD: Good call — done. Replaced `ROCPROFSYS_NIC_EMIT_COUNTER` with an inline function template `emit_nic_counter<CategoryTp>`.

dgaliffiAMD:
Thank you for the improvement, @adjordje-amd.

PR #7835, opened 2026-06-25
fix(rocjitsu): Propagate VMID through mem hierarchy, fix coherency/SDMA

atgutier:
## Motivation

This PR tags every cache line by (vmid, addr) and threads the VMID end-to-end from L1 down to HBM,
eliminating cross-process data aliasing in daemon mode. It also replaces the simulator-only
per-range SDMA invalidates with the coarse cache maintenance real hardware performs, relying on the
dispatch-time acquire fence for coherence, and fixes the GCR packet sizing that desynced the SDMA
ring pointer on RDNA3/4.

## Technical Details

VMID tagging and propagation:
    - Tag every cache line by (vmid, addr), not addr alone. Guest VAs are
        per-process and can alias across processes, so two VMIDs sharing a GPU
        VA must occupy distinct cache entries. Add a vmid field to simdojo
        CacheTag and match it in lookup/allocate/invalidate/read_line/
        write_line/fill_line/line_data_for_write.
    - Carry vmid in the simdojo message header so it survives across links
        (L2 -> MSC -> HBM).
    - Thread vmid through L1 vector, L1 scalar, L2, and memory-side caches,
        the HBM controller, and GpuMemory's request handler.
    - Write back evicted/flushed dirty lines under the evicted line's own
        vmid rather than the current request's vmid, so cross-process aliased
        lines flush to the correct address space.
    - Pass vmid to kernarg preload reads and the atomic-RMW L1 invalidate
        so daemon mode resolves client-process memory.
    
MSC write-through:
  - Change the memory-side cache from write-back to write-through: stores
    are pushed straight to the backing store and the line is left clean
    (EXCLUSIVE) instead of dirty (MODIFIED), so the daemon's
    process_vm_readv bridge observes writes immediately.
    
HW-faithful SDMA coherence:
  - Replace the surgical per-range L2 invalidates after SDMA copy/fill/
    write/fence/atomic/GCR with a single coarse invalidate_gpu_caches()
    (whole L2 + L1 V$). Real SDMA does not snoop GL2 and HW cache
    maintenance is indiscriminate, not per-range; the consuming kernel's
    acquire fence at dispatch re-establishes coherence. Remove the
    simulator-only L2Cache::invalidate_range API.
    
AQL grid workgroup count:
  - Compute grid workgroup counts with ceiling division instead of
    truncating division, which undercounted workgroups when grid_size is
    not a multiple of workgroup_size. Make util::ceil_div strict
    (asserts on zero divisor) and add util::ceil_div_or_one for the
    guest-packet path, which must tolerate a malformed zero dimension
    without aborting.

 SDMA GCR packet size by dialect:
  - Split the SDMA GCR packet size: gfx1250 uses the 6-dword
    SDMA_PKT_GCR_GFX1250 layout while gfx11/gfx12 use the 5-dword layout.
    Previously all gfx11+ arches used 6 dwords, desyncing the SDMA ring
    read pointer on RDNA3/RDNA4. Add a Gfx1250 SdmaPacketDialect and
    select it for gfx1250 in the config loader.

## JIRA ID

JIRA ID N/A
PR #7835

## Test Plan

Run CI tests

## Test Result

CI tests: PASSED

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
@@ -14,6 +14,7 @@ RJ_DIAGNOSTIC_POP
 
 #include "simdojo/sim/message.h"
 #include "simdojo/sim/simulation.h"
+#include "util/bit.h"
 #include "util/log.h"
 
 #include <algorithm>
@@ -84,10 +85,10 @@ void validate_cluster_shape(const DispatchEntry &dp) {
   }
 }
 
-uint32_t read_memory_u32(GpuMemory *memory, uint64_t addr) {
+uint32_t read_memory_u32(GpuMemory *memory, uint64_t addr, uint32_t vmid = 0) {
   uint32_t value = 0;
   for (uint32_t i = 0; i < sizeof(value); ++i)
-    value |= static_cast<uint32_t>(memory->read8(addr + i)) << (i * 8);
+    value |= static_cast<uint32_t>(memory->read8(addr + i, vmid)) << (i * 8);
   return value;
 }
 
@@ -239,7 +240,8 @@ void CommandProcessor::init_wavefront_regs(ComputeUnitCore *cu, Wavefront *wf,
 
       uint64_t preload_addr = pkt.kernarg_addr + static_cast<uint64_t>(preload_offset) * 4;
       for (uint32_t i = 0; i < preload_length; ++i)
-        cu->write_sgpr(sbase + idx + i, read_memory_u32(memory_, preload_addr + i * 4));
+        cu->write_sgpr(sbase + idx + i,
+                       read_memory_u32(memory_, preload_addr + i * 4, pkt.process_id));
       util::Logger::vm("CP: init_wf kernarg preload s[", idx, ":", idx + preload_length - 1,
                        "] length=", preload_length, " offset=", preload_offset, " sbase=", sbase);
       idx += preload_length;
@@ -922,9 +924,12 @@ void CommandProcessor::process_aql_packet(const hsa_kernel_dispatch_packet_t &pk
   uint32_t wfs_per_wg = (wg_size + wave_size - 1) / wave_size;
 
   uint32_t num_dims = pkt.setup & 0x3;
-  uint32_t grid_wgs_x = pkt.workgroup_size_x > 0 ? pkt.grid_size_x / pkt.workgroup_size_x : 1;
-  uint32_t grid_wgs_y = pkt.workgroup_size_y > 0 ? pkt.grid_size_y / pkt.workgroup_size_y : 1;
-  uint32_t grid_wgs_z = pkt.workgroup_size_z > 0 ? pkt.grid_size_z / pkt.workgroup_size_z : 1;
+  uint32_t grid_wgs_x =
+      util::ceil_div_or_one(pkt.grid_size_x, static_cast<uint32_t>(pkt.workgroup_size_x));
+  uint32_t grid_wgs_y =
+      util::ceil_div_or_one(pkt.grid_size_y, static_cast<uint32_t>(pkt.workgroup_size_y));
+  uint32_t grid_wgs_z =
+      util::ceil_div_or_one(pkt.grid_size_z, static_cast<uint32_t>(pkt.workgroup_size_z));
   uint32_t total_wgs = grid_wgs_x * grid_wgs_y * grid_wgs_z;
 
   DispatchEntry dp{};
@@ -1478,7 +1483,7 @@ static void *resolve_sdma_ptr(GpuMemory *memory, uint64_t va, uint32_t vmid) {
   return page_base + (va & 0xFFF);
 }
 
-// SDMA opcodes (from sdma_registers.h).
+// SDMA opcodes.
 namespace sdma {
 constexpr uint8_t OP_NOP = 0;
 constexpr uint8_t OP_COPY = 1;
@@ -1506,7 +1511,26 @@ constexpr uint32_t ATOMIC_SIZE = 8;
 constexpr uint32_t CONST_FILL_SIZE = 5;
 constexpr uint32_t TIMESTAMP_SIZE = 3;
 constexpr uint32_t GCR_SIZE = 5;
-constexpr uint32_t GCR_GFX11_PLUS_SIZE = 6;
+constexpr uint32_t GCR_GFX1250_SIZE = 6;
+
+// GCR GL2 cache-op control bits. The control dword and bit positions genuinely
+// differ by dialect (the gfx1250 GCR packet is a distinct layout, not a resized
+// legacy packet). Legacy/gfx9-12 pack gcr_control[15:0] into DW2 starting at
+// bit 16 (DW2 low 16 bits hold BaseVA_HI); gfx1250 makes DW2 a full 25-bit base
+// VA and relocates the control field to DW3 starting at bit 0. Only the GL2
+// writeback/invalidate/discard bits matter for the functional GL2 model.
+//   Legacy DW2:  GL2_DISCARD=29, GL2_INV=30, GL2_WB=31  (= gcr_control 13/14/15 + 16)
+//   gfx1250 DW3: GL2_DISCARD=13, GL2_INV=14, GL2_WB=15
+// Layouts and sizes (legacy 5 dwords, gfx1250 6 dwords) match the vendored
+// runtime SDMA GCR packet definitions for each dialect.
+constexpr uint32_t GCR_LEGACY_CONTROL_DW = 2;
+constexpr uint32_t GCR_LEGACY_GL2_DISCARD_BIT = 1u << 29;
+constexpr uint32_t GCR_LEGACY_GL2_INV_BIT = 1u << 30;
+constexpr uint32_t GCR_LEGACY_GL2_WB_BIT = 1u << 31;
+constexpr uint32_t GCR_GFX1250_CONTROL_DW = 3;
+constexpr uint32_t GCR_GFX1250_GL2_DISCARD_BIT = 1u << 13;
+constexpr uint32_t GCR_GFX1250_GL2_INV_BIT = 1u << 14;
+constexpr uint32_t GCR_GFX1250_GL2_WB_BIT = 1u << 15;
 constexpr uint32_t COPY_LINEAR_WAITSIGNAL_GFX11_PLUS_SIZE = 19;
 constexpr uint32_t FENCE_64B_GFX11_PLUS_SIZE = 5;
 constexpr uint32_t POLL_MEM_64B_GFX11_PLUS_SIZE = 8;
@@ -1538,6 +1562,30 @@ bool sdma_compare_u64(uint32_t func, uint64_t value, uint64_t reference) {
 }
 
 } // namespace
+
+void CommandProcessor::flush_gpu_caches() {
+  // Write back dirty scalar L1 (K$) lines into L2 first, then flush L2 to
+  // backing, so a dirty K$ line overlapping an SDMA destination is published
+  // before the direct write (which happens after this helper returns) rather
+  // than being written out over it by a later K$ flush. Each line is written
+  // back under its own owning vmid. Vector L1 (V$) is write-through, so it only
+  // needs invalidation. Ordering: K$ -> L2 -> backing, then invalidate V$.
+  for (auto *cu : cus_) {
+    cu->l1_scalar().writeback_all();
+    cu->l1_scalar().invalidate_all();
+  }
+  for (auto *l2 : l2_caches_)
+    l2->flush_all();
+  for (auto *cu : cus_)
+    cu->l1_vector().invalidate_all();
+}
+
+void CommandProcessor::invalidate_gpu_caches() {
+  for (auto *l2 : l2_caches_)
+    l2->invalidate_all();
+  for (auto *cu : cus_)
+    cu->l1_vector().invalidate_all();
+}
 
 void CommandProcessor::process_sdma_ring(HwQueue &queue, uint64_t read_idx, uint64_t write_idx) {
   uint32_t ring_mask = (queue.ring_size / sizeof(uint32_t)) - 1;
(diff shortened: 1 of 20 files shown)


atgutier:
@copilot do another review

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
```diff
@@ -1539,6 +1544,13 @@ bool sdma_compare_u64(uint32_t func, uint64_t value, uint64_t reference) {
 
 } // namespace
 
+void CommandProcessor::invalidate_gpu_caches() {
+  for (auto *l2 : l2_caches_)
+    l2->invalidate_all();
```
kuhar, line 1585: `invalidate_gpu_caches()` now calls `L2Cache::invalidate_all()`, which clears tags and dirty bits without writeback. L2 can still hold dirty lines from `L1ScalarCache::writeback_all()` via `L2Cache::writeback_line()`, so a small SDMA write or signal can discard unrelated dirty L2 data for another address or VMID. Please flush dirty L2 lines before the coarse invalidate, or limit the SDMA maintenance to lines known to be clean/destination-owned.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
```diff
@@ -1891,19 +1896,11 @@ void CommandProcessor::process_sdma_ring(HwQueue &queue, uint64_t read_idx, uint
...
-        if (range > 0)
-          l2->invalidate_range(base_va, range);
-        else
-          l2->invalidate_all();
-      }
-      pkt_dwords = uses_gfx11_plus_sdma_packets() ? sdma::GCR_GFX11_PLUS_SIZE : sdma::GCR_SIZE;
+      // GCR is the real SDMA cache-maintenance packet. It carries a base/size,
+      // but GFX9+ HW services it at coarse (whole-cache) granularity, so we
+      // model it as a full GL2 invalidate regardless of the range fields.
+      invalidate_gpu_caches();
```
kuhar: `OP_GCR` should not unconditionally invalidate GL2. The runtime packet has separate `GCR_CONTROL_GL2_WB` and `GCR_CONTROL_GL2_INV` bits, but this decoder ignores the control word and always calls `invalidate_gpu_caches()`. A writeback-only GCR packet should flush dirty L2 lines; treating it as invalidate can drop simulator data instead of publishing it. Please decode the control bits and flush for `GL2_WB`, invalidating only for invalidate/discard requests.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
```diff
@@ -1891,19 +1896,11 @@ void CommandProcessor::process_sdma_ring(HwQueue &queue, uint64_t read_idx, uint
...
+      pkt_dwords = uses_gfx1250_gcr_packet() ? sdma::GCR_GFX1250_SIZE : sdma::GCR_SIZE;
```
kuhar: The GCR size split is correct, but it needs a regression test. `SDMA_PKT_GCR` in the runtime header is 5 dwords, while `SDMA_PKT_GCR_GFX1250` is 6; a wrong `pkt_dwords` silently desynchronizes the SDMA ring and corrupts the following packet. Please add a two-packet test, `OP_GCR` followed by a FENCE, asserting the read pointer advances by 5 dwords for the gfx11/12 dialect and 6 for gfx1250.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
```diff
@@ -922,9 +924,12 @@ void CommandProcessor::process_aql_packet(const hsa_kernel_dispatch_packet_t &pk
   uint32_t wfs_per_wg = (wg_size + wave_size - 1) / wave_size;
 
   uint32_t num_dims = pkt.setup & 0x3;
-  uint32_t grid_wgs_x = pkt.workgroup_size_x > 0 ? pkt.grid_size_x / pkt.workgroup_size_x : 1;
-  uint32_t grid_wgs_y = pkt.workgroup_size_y > 0 ? pkt.grid_size_y / pkt.workgroup_size_y : 1;
-  uint32_t grid_wgs_z = pkt.workgroup_size_z > 0 ? pkt.grid_size_z / pkt.workgroup_size_z : 1;
+  uint32_t grid_wgs_x =
```
kuhar, line 927: The floor-to-ceil workgroup-count fix is correct but untested. For example, `grid_size_x=65, workgroup_size_x=32` should dispatch 3 workgroups; the old code would dispatch 2. Existing `PartialWorkgroupMasksTailWaveExec` only covers a partial wave inside one workgroup (`33/33`), not a partial final workgroup. Please add a dispatch regression that asserts the rounded-up `total_wgs`/dispatched wave count, ideally with one 2D case as well.

kuhar (commented):
In `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l1_scalar_cache.cpp:81`, `L1ScalarCache::writeback_all()` still writes every dirty K$ line using the caller VMID. Cache tags now record the owning VMID, and the eviction path already uses `evicted.vmid`; bulk writeback should do the same. Otherwise a flush for one queue can write a dirty scalar line from another process through the wrong page table before invalidating it.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
```diff
@@ -1662,16 +1700,19 @@ void CommandProcessor::process_sdma_ring(HwQueue &queue, uint64_t read_idx, uint
...
-        for (auto *cu : cus_)
-          cu->l1_vector().invalidate_all();
+        // Emulated SDMA writes straight to the backing store, bypassing the GPU
+        // caches. Real SDMA does not snoop GL2; coherence with the destination
+        // is re-established by the consuming kernel's acquire fence (see the
+        // AGENT-scope GL2 invalidate at dispatch). We model that with a coarse,
+        // indiscriminate writeback+invalidate: the writeback publishes any dirty
+        // L2 lines (e.g. from K$ writeback) so they are not lost, and the
+        // invalidate forces the destination to re-read the fresh backing data.
+        flush_gpu_caches();
```
kuhar, line 1720: flush_gpu_caches runs after the SDMA path has already written directly to backing memory. If L2 holds a dirty line overlapping the destination, flush_all writes the stale cached line back over the SDMA result. I validated the sequence with the real L2Cache/GpuMemory classes: dirty L2 value 0x11111111, direct backing write 0x22222222, then flush_all left backing as 0x11111111. Please flush or invalidate the overlapping destination before the direct SDMA write, or make the post-write flush skip destination-overlapping dirty lines, and add a regression for that ordering.

emulation/rocjitsu/lib/simdojo/include/simdojo/components/cache.h
```diff
@@ -67,8 +67,13 @@ enum class CoherenceState : uint8_t {
 };
 
 /// @brief Tag and metadata for a single cache line.
+///
+/// @details @c vmid identifies the owning process address space. Lines with the
+/// same line address but different vmids are distinct entries, because guest VAs
+/// are per-process and may alias across processes.
 struct CacheTag {
   uint64_t tag = 0;
+  uint32_t vmid = 0;
```
kuhar, line 76: The vmid-tagged Cache behavior is the core invariant this PR depends on, but there is still no direct test for it. Please add small simdojo::Cache tests for: same line address under two VMIDs stores distinct data; eviction reports the evicted line owner VMID; invalidate(addr, vmidA) leaves vmidB intact. I validated these cases manually against the header-only Cache, so this is a coverage gap rather than a current functional failure.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
```diff
@@ -1891,19 +1923,34 @@ void CommandProcessor::process_sdma_ring(HwQueue &queue, uint64_t read_idx, uint
...
+      const uint32_t control =
+          gfx1250 ? dw(sdma::GCR_GFX1250_CONTROL_DW) : dw(sdma::GCR_LEGACY_CONTROL_DW);
+      const uint32_t wb_bit = gfx1250 ? sdma::GCR_GFX1250_GL2_WB_BIT : sdma::GCR_LEGACY_GL2_WB_BIT;
+      const uint32_t inv_bit =
+          gfx1250 ? sdma::GCR_GFX1250_GL2_INV_BIT : sdma::GCR_LEGACY_GL2_INV_BIT;
+      const uint32_t discard_bit =
+          gfx1250 ? sdma::GCR_GFX1250_GL2_DISCARD_BIT : sdma::GCR_LEGACY_GL2_DISCARD_BIT;
+      const bool gl2_wb = (control & wb_bit) != 0;
+      const bool gl2_inv = (control & (inv_bit | discard_bit)) != 0;
+      if (gl2_wb) {
```
kuhar, line 1970: The GCR test now catches packet-size/ring desync, but it still does not observe the new WB vs INV behavior. The decoder has three distinct outcomes now: WB flushes dirty L2 to backing, INV/DISCARD drops without writeback, and no GL2 bits is a no-op. Please add one behavioral regression that dirties an L2 line and issues WB-only, INV-only, and no-op GCR packets. That is the data-loss distinction this PR is trying to protect.

kuhar (commented):
Original anchor: emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l1_scalar_cache.cpp:81

writeback_all should write each dirty scalar-cache line back with the line owner VMID, not the caller VMID. The eviction path already does this with evicted.vmid, but this loop still passes vmid to L2Cache::writeback_line. I validated the failure with the real classes: two page tables mapped the same VA to different backing pages; after store under VMID 7 and writeback_all(8), the bytes landed in VMID 8 backing instead of VMID 7. Please pass tag.vmid here and add a small regression for aliased VAs across VMIDs.

kuhar:
We should probably land this first and then rebase #8118 on top:
Concrete places to watch:

- PR 8118’s new scalar flush_line should use VMID-tagged lookup/read_line/invalidate when combined with PR 7835.
- PR 8118’s writeback_all fix should not regress to writing with the caller VMID; after PR 7835 it should preserve the line-owner VMID
    behavior.
- PR 8118’s vector cache_.invalidate(ea) additions should become cache_.invalidate(ea, vmid) after PR 7835.

emulation/rocjitsu/tests/gfx1250_sim_test.cpp
```diff
@@ -689,6 +692,139 @@ TEST(Gfx1250SdmaTest, Fence64WritesFull64BitValue) {
...
+    queue.submit(6); // gfx1250 GCR is 6 dwords.
+    EXPECT_TRUE(sim.engine->step());
+    return sim.memory->read32(queue.dst_va(), kProcessId);
+  };
+
+  // Writeback publishes the dirty line to backing.
+  EXPECT_EQ(run(GcrKind::WritebackOnly), kDirtyWord);
+  // Invalidate/discard drops the dirty line; backing keeps its original value.
+  EXPECT_EQ(run(GcrKind::InvalidateOnly), kBackingWord);
+  // No GL2 bits: no cache maintenance; the dirty line is untouched in L2 and
```
kuhar: The no-op GCR case is not actually distinguished from invalidate-only here. Returning only the backing-store word means an implementation that incorrectly calls invalidate_gpu_caches for a no-GL2-bits packet would also leave backing as kBackingWord and this assertion would still pass. Please have the no-op case also observe that the dirty L2 line remains resident, or otherwise refactor the helper so the no-op path fails if it drops the cache line.

kuhar (commented):
I don't think https://github.com/ROCm/rocm-systems/pull/7835#pullrequestreview-4638966821 is addressed?

L1ScalarCache::writeback_all still writes every dirty scalar line through the caller-supplied vmid. With the new VMID-tagged cache, one CU can retain dirty K-cache lines from process A and then be flushed from an acquire fence or SDMA path while processing process B; this loop will install/write those A lines into L2 under B's address space. That breaks the core line-owner contract added in this patch and can corrupt another process VA.

emulation/rocjitsu/tests/gfx1250_sim_test.cpp
```diff
@@ -689,6 +692,139 @@ TEST(Gfx1250SdmaTest, Fence64WritesFull64BitValue) {
...
+  // gfx1250 GCR control dword (DW3) bit positions.
+  constexpr uint32_t kControlDw = 3;
+  constexpr uint32_t kGl2InvBit = 1u << 14;
+  constexpr uint32_t kGl2WbBit = 1u << 15;
+
+  enum class GcrKind { WritebackOnly, InvalidateOnly, Noop };
+  auto run = [&](GcrKind kind) -> uint32_t {
+    Gfx1250Sim sim;
+    auto *l2 = sim.xcd()->l2_cache();
+    EXPECT_NE(l2, nullptr);
```
kuhar: For consistency with the surrounding tests, make this a fatal guard before dereferencing l2 below.

emulation/rocjitsu/tests/gfx1250_sim_test.cpp
```diff
@@ -689,6 +692,139 @@ TEST(Gfx1250SdmaTest, Fence64WritesFull64BitValue) {
...
+
+  run_dialect(amdgpu::SdmaPacketDialect::Gfx11Plus, kGcrLegacySize, kLegacyGl2InvControlDw,
+              kLegacyGl2InvBit);
+  run_dialect(amdgpu::SdmaPacketDialect::Gfx1250, kGcrGfx1250Size, kGfx1250Gl2InvControlDw,
+              kGfx1250Gl2InvBit);
+}
+
+// SDMA writes go straight to backing while L2 may still hold a dirty line that
+// overlaps the destination (e.g. left by a prior K$ writeback). The post-write
+// cache maintenance must not write that stale line back OVER the SDMA result.
```
kuhar: Nit: the all-caps emphasis in these test comments does not match the surrounding sentence-case style.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
```diff
@@ -1539,6 +1559,23 @@ bool sdma_compare_u64(uint32_t func, uint64_t value, uint64_t reference) {
 
 } // namespace
 
+void CommandProcessor::flush_gpu_caches() {
```
kuhar, line 1566: `flush_gpu_caches()` still leaves dirty scalar L1/K$ lines resident. Scalar stores go through `L1ScalarCache::store()` and mark K$ lines dirty, but this helper only flushes L2 and invalidates vector L1. If SDMA writes directly to a backing page that overlaps a dirty K$ line, a later `flush_l1()`/acquire/release can write the stale scalar line into L2/backing after the SDMA write and clobber the SDMA result.

Please include scalar K$ writeback+invalidate in this pre-write maintenance, before the L2 flush, or add a test/proof that dirty scalar lines cannot overlap SDMA destinations.

kuhar (commented):
Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp:1930`

`OP_TIMESTAMP` is still a direct backing-store write with no cache maintenance. COPY/FENCE/ATOMIC/CONST_FILL/WRITE now flush before their direct stores so dirty cached data cannot later overwrite the SDMA result, but this path resolves the timestamp VA and stores through the host pointer directly.

If L2 already has a dirty line for the timestamp address, or a CU has a stale vector line, the timestamp can be hidden or later overwritten. Please use the same pre-write cache-maintenance path here and cover it with the dirty-line regression pattern used for CONST_FILL.

kuhar (commented):
Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp:1941`

The SDMA direct-write cache-maintenance issue is only partially addressed. COPY/FENCE/ATOMIC/CONST_FILL/WRITE now call `flush_gpu_caches()` before writing through `resolve()`, and `flush_gpu_caches()` now publishes dirty scalar K$ before flushing L2. `OP_TIMESTAMP` still does the same kind of direct backing-store write without that pre-write maintenance.

If L2/K$ holds a dirty line for the timestamp address, it can later flush over the timestamp value; a resident clean line can also hide the new value from a GPU read. Please apply the same pre-write maintenance here and add a small timestamp regression.

kuhar (approved):
Approving modulo the last remaining issue: https://github.com/ROCm/rocm-systems/pull/7835#pullrequestreview-4648511076

PR #7836, opened 2026-06-25
rocr: gfx1250 SDMA fused WaitSignal path - chunking, single-reservati…

saleelk:
## Summary

gfx1250 SDMA fused WaitSignal path: chunking, single-reservation notify, multicast consolidation, and broadcast routing tuning.

### Builders

* Fix `BuildWaitSignalCopyCommand` / `BuildWaitSignalSwapCommand` / multicast to emit **compacted variable-length packets** (only the WAIT/SIGNAL doubleword blocks actually present), matching the HW parser and fixing faults on multi-packet fused copies. Caller sizing reserves the WAIT block only when a dep is folded in, exactly matching the compacted emission.
* Fused linear-copy, swap, and multicast WaitSignal builders **chunk large transfers** via `num_copy_command`. Every chunk waits on the same input signal and signals the same output signal; the submit pre-loads `out_signal` by `N-1` so it only reaches 0 once all chunks complete — correct even when the HW overlaps chunk copies. Indirect packets still cannot chunk and reject `size > max`.

### Single-reservation notify

* `SubmitLinearCopyBodyWaitSignal` gains `fused_notify` (default true): it folds the GCR-invalidate prologue and the GCR-writeback/mailbox epilogue into the **same ring reservation**, so a standalone single-engine copy needs one reserve/release instead of three. The fan-out path passes `false` and keeps driving `SubmitNotifyPrologue/Epilogue` separately across engines.

### Multicast

* Consolidate multicast into a single `SubmitLinearCopyMulticastCommand` that branches on profiling: profiling uses the timestamp-capable plain packet via the classic submit path; non-profiling fuses GCR + copy/signal + GCR writeback/mailbox into one reservation using the multicast WaitSignal packet. Reject `> 1024` destinations (10-bit `num_of_destination` field).

### gfx1250 cache + epilogue

* **GCR off for gfx1250.** `BlitSdmaV6` now builds with `useGCR=false`: gfx1250 SDMA ops run on the DACC backend and are **not cached in GL2**, so the GCR invalidate/writeback packets are unnecessary; the per-packet system-scope fields handle coherency.
* **Drop the redundant epilogue self-poll on `out_signal`** from the fused linear and multicast WaitSignal paths. The ring is in-order and the packet's inline 64b-sub SIGNAL already executes after its copy, so the copy is drained before the epilogue. Worse than redundant on multicast: polling the very address the packet inline-signals **deadlocked the gfx1250 SDMA engine** (the inline-signal write was not observed by the immediately-following `POLL_REGMEM`), wedging every later packet. The cross-engine fan-out epilogue keeps its poll on the *other* engines' body signals, which is a legitimate cross-engine wait and is unchanged.

### Routing (`amd_gpu_agent`)

* gfx1250 single-copy fast path drops the `size <= MaxSingleLinearCopySize()` gate and the now-redundant prologue/epilogue calls; chunking is handled in the builder.
* `DmaCopyBroadcast` on gfx1250: multicast for copies `<= 256 KB`; larger copies now fall through to **fan-out** (was linearB2B). Fan-out parallelises the per-destination copies across SDMA engines and measured **~3–5% faster** than a single-engine linearB2B at large sizes. The non-gfx1250 path (linearB2B / broadcast / fan-out) is unchanged.
* `DmaCopyFanOutOp`: narrow the multi-packet fallback to indirect copies only, so large linear-copy and swap entries take the chunked fused WaitSignal path while oversized indirect copies still fall back and are rejected.

### CLR (`rocdevice` SdmaEngineAllocator)

* Enable the **simple round-robin** engine allocator on gfx1250+, whose SDMA engines have symmetric bandwidth, so broadcast fan-out actually spreads across engines. Older asics keep the preferred-engine heuristics.

### Debug

* `HSA_SDMA_FORCE_MULTICAST=1` forces the gfx1250 broadcast onto the multicast packet regardless of size. **Debug only**: the multicast engine drops destinations above 256 KB, so large copies fail validation by design.

## Motivation

gfx1250 multi-destination collectives (e.g. cuMem-fabric allgather) need a fused, chunked SDMA path that does not fault on large transfers, minimises ring reservations, and picks the fastest correct submission mode per size.

## Technical Details

See the per-area breakdown above. Key correctness fixes: (1) compacted variable-length packet emission now exactly matches caller-reserved sizes, eliminating uninitialized command-buffer regions that caused GPU faults on multi-packet fused copies; (2) the redundant epilogue self-poll that deadlocked the gfx1250 multicast engine is removed, and GCR cache packets are dropped on gfx1250 (uncached DACC backend + per-packet scopes). Key perf change: large broadcasts route through fan-out + round-robin engine allocation on gfx1250.

## Test Plan

* cuMem-fabric allgather (4 ranks, MI450), size sweep 512 B → 16 GiB.
* Pure-HSA `hsa_amd_memory_async_batch_copy` multicast sweep (512 B → 64 MiB), both GPU-only and interruptible completion signals.
* A/B of the three broadcast submission modes (multicast / linearB2B / fan-out) via a forced-path build to confirm correctness and relative throughput.

## Test Result

* Default routing: **Validation passed** across the full 512 B → 16 GiB sweep.
* Pure-HSA multicast sweep (MI450, gfx1250): **PASS** at every size for both GPU-only and interruptible signals — no hangs. (Previously the fused multicast epilogue self-poll wedged the SDMA engine; the poll removal fixes it.)
* Large-size throughput (4-rank allgather), fan-out vs previous linearB2B default:
| Size   | linearB2B | fan-out   | gain   |
| ------ | --------- | --------- | ------ |
| 1 GiB  | 630.3 µs  | 600.5 µs  | ~4.7% |
| 4 GiB  | 2454.7 µs | 2333.8 µs | ~4.9% |
| 8 GiB  | 4888.6 µs | 4645.0 µs | ~5.0% |
| 16 GiB | 9732.0 µs | 9278.1 µs | ~4.7% |
* `HSA_SDMA_FORCE_MULTICAST=1` reproduces the expected multicast > 256 KB destination-drop validation failures, confirming the debug knob (this is a HW limit, not a packet-encoding bug: the dumped packet carries all destinations and the correct count).

## Submission Checklist

* Look over the contributing guidelines at <https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests>.

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/inc/amd_blit_sdma.h
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/inc/amd_blit_sdma.h
@@ -81,6 +81,21 @@ class BlitSdmaBase : public core::Blit {
       std::vector<core::Signal*>& dep_signals,
       core::Signal& out_signal) = 0;
 
+  /// @brief Multicast (one source, N destinations) linear copy.  Only valid
+  /// when MulticastSupported() is true.  When @p profiling_enabled, the plain
+  /// multicast packet is submitted through the classic prologue/copy/epilogue
+  /// path so start/end timestamps are collected.  Otherwise it fuses the GCR
+  /// invalidate prologue, poll+copy+signal, and the GCR writeback + mailbox
+  /// epilogue into a single ring reservation: dep_signals[0] folds into each
+  /// chunk's packet WAIT and out_signal is decremented by every chunk's SIGNAL
+  /// (the count is pre-loaded accordingly); any remaining dep_signals are
+  /// emitted as poll commands.
+  virtual hsa_status_t SubmitLinearCopyMulticastCommand(
+      const std::vector<void*>& dsts, const void* src, size_t size,
+      std::vector<core::Signal*>& dep_signals,
+      core::Signal& out_signal,
+      bool profiling_enabled) = 0;
+
   /// @brief Pack N linear copy packets back-to-back in a single SDMA ring
   /// submission (linearB2BCopy path).  Each entry i copies srcs[i] -> dsts[i]
   /// of sizes[i] bytes.  For the broadcast case (same src/size for all dsts),
@@ -91,8 +106,17 @@ class BlitSdmaBase : public core::Blit {
       core::Signal& out_signal) = 0;
 
   virtual bool BroadcastSupported() const = 0;
+  virtual bool MulticastSupported() const = 0;
   virtual bool PlatformAtomicSupport() const = 0;
   virtual bool IsGfx1250() const = 0;
+
+  /// @brief Maximum byte count a single linear copy packet can describe
+  /// (the 30-bit count field, ~1 GiB).  Larger copies are split into multiple
+  /// chunked packets; both the plain (BuildCopyCommand) and fused WaitSignal
+  /// builders chunk internally.  The indirect WaitSignal packet is the lone
+  /// exception - it has no offset field, so callers must reject indirect copies
+  /// larger than this limit.
+  virtual size_t MaxSingleLinearCopySize() const = 0;
 
   virtual hsa_status_t SubmitPrologue(const std::vector<core::Signal*>& dep_signals,
                                       core::Signal& out_signal,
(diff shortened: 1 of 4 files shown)


saleelk:
Addressed the Copilot review comments (force-pushed):

- **amd_blit_sdma.h (multicast doc):** Updated `SubmitLinearCopyMulticastCommand` doc — the non-profiling path fuses the GCR-invalidate prologue and poll/GCR-writeback/mailbox epilogue into one reservation (no longer says "no prologue/epilogue").
- **amd_blit_sdma.h (`MaxSingleLinearCopySize` doc):** Reworded — it is the 30-bit count limit (~1 GiB); both plain and fused WaitSignal builders chunk internally. Only the indirect WaitSignal packet cannot chunk, so callers must reject oversized indirect copies.
- **amd_gpu_agent.cpp (`HSA_SDMA_LINEAR_B2B`):** Restored the env override in `DmaCopyBroadcast`, consistent with `DmaCopyMulti`.
- **amd_gpu_agent.cpp (fast-path comment):** Reworded — only the profiling timestamp path is skipped; GCR/mailbox sync is still issued via the coordinator's `SubmitNotifyPrologue/Epilogue`.
- **amd_blit_sdma.cpp (multicast dst count):** Added validation rejecting >1024 destinations (10-bit `num_of_destination`, encoded as count-1) to avoid field truncation.

Additionally, the fused chunking now uses per-chunk wait+signal (every chunk waits on the same input signal and decrements the same output signal, with out_signal pre-loaded by N-1). This keeps completion correct even when the HW overlaps chunk copies (MI450+), replacing the earlier wait-first/signal-last scheme.

gandryey (approved):

dayatsin-amd (approved):

PR #7840, opened 2026-06-25
ci(rocjitsu): Install tests as a build artifact

amd-eochoalo:
## Motivation

Adds the possibility to install tests as a build artifact. This is necessary to separate rocjitsu's testing from TheRock's build and test phases.

## Technical Details

We have a CMake function rj_add_test that is used instead of add_test to register tests. When RJ_INSTALL_TESTS=OFF, it is just add_test. When RJ_INSTALL_TESTS=ON, these tests are also installed under /share/rocjitsu/tests and binaries under /share/rocjitsu/tests/bin. Some tests are updated to take into account this.

## JIRA ID

#7891

## Test Plan

Locally built with RJ_INSTALL_TESTS=OFF and RJ_INSTALL_TESTS=ON, making sure to preserve the original behaviour and installing in a new directory

## Test Result

Test build artifacts are installed.

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/cmake/rj_generate_installed_gtest_ctest.cmake
+++ b/emulation/rocjitsu/cmake/rj_generate_installed_gtest_ctest.cmake
@@ -1,0 +1,97 @@
+# Copyright (c) 2025-2026 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+# Build-time helper for installed GoogleTest discovery.
+#
+# This script is run with `cmake -P` after TEST_EXECUTABLE has been built. It
+# asks GoogleTest for the concrete test list, including parameterized TEST_P
+# expansions, then writes a CTest fragment that runs the installed executable
+# once per discovered test case.
+#
+# Required variables:
+#   TEST_EXECUTABLE      Build-tree executable used for --gtest_list_tests.
+#   INSTALLED_EXECUTABLE Path CTest should run from the installed test dir.
+#   OUTPUT_FILE          Generated CTest fragment path.
+function(rj_generate_installed_gtest_ctest)
+    set(oneValueArgs TEST_EXECUTABLE INSTALLED_EXECUTABLE OUTPUT_FILE)
+    cmake_parse_arguments(ARG "" "${oneValueArgs}" "" ${ARGN})
+
+    if(NOT ARG_TEST_EXECUTABLE)
+        message(FATAL_ERROR "TEST_EXECUTABLE is required")
+    endif()
+    if(NOT ARG_INSTALLED_EXECUTABLE)
+        message(FATAL_ERROR "INSTALLED_EXECUTABLE is required")
+    endif()
+    if(NOT ARG_OUTPUT_FILE)
+        message(FATAL_ERROR "OUTPUT_FILE is required")
+    endif()
+
+    execute_process(
+        COMMAND
+            "${ARG_TEST_EXECUTABLE}" --gtest_list_tests
+            "--gtest_filter=-*Benchmark*"
+        RESULT_VARIABLE _result
+        OUTPUT_VARIABLE _tests
+        ERROR_VARIABLE _errors
+    )
+    if(NOT _result EQUAL 0)
+        message(
+            FATAL_ERROR
+            "Failed to list tests from ${ARG_TEST_EXECUTABLE}: ${_errors}"
+        )
+    endif()
+
+    file(WRITE "${ARG_OUTPUT_FILE}" "# Installed GoogleTest CTest tests\n")
+    set(_suite)
+    string(REPLACE "\n" ";" _lines "${_tests}")
+    foreach(_line IN LISTS _lines)
+        string(STRIP "${_line}" _stripped_line)
+        if(_line MATCHES "^([^# \t]+)\\.[ \t]*(#.*)?$")
+            set(_suite "${CMAKE_MATCH_1}")
+        elseif(_line MATCHES "^  ([^# \t]+)")
+            if(NOT _suite)
+                message(FATAL_ERROR "Found gtest case without suite: ${_line}")
+            endif()
+            set(_case_name "${CMAKE_MATCH_1}")
+            set(_pretty_case_name "${_case_name}")
+            if(
+                _case_name MATCHES "/[0-9]+$"
+                AND _line MATCHES "# GetParam\\(\\) = \"?([A-Za-z0-9_]+)\"?$"
+            )
+                string(
+                    REGEX REPLACE
+                    "/[0-9]+$"
+                    "/${CMAKE_MATCH_1}"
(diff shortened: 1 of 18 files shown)


emulation/rocjitsu/cmake/rj_generate_installed_gtest_ctest.cmake
```diff
@@ -0,0 +1,83 @@
+# Build-time helper for installed GoogleTest discovery.
```
kuhar, line 4: Do we need a copyright header?

emulation/rocjitsu/tests/daemon_test.cpp
```diff
@@ -46,6 +46,29 @@ bool daemon_ready(const std::string &path) {
   return ok;
 }
 
+const char *env_or_default(const char *name, const char *fallback) {
+  const char *value = std::getenv(name);
```
kuhar: If there any way to avoid these new env vars?

emulation/rocjitsu/cmake/rj_generate_installed_gtest_ctest.cmake
amd-eochoalo: Added.

emulation/rocjitsu/tests/daemon_test.cpp
amd-eochoalo: I now look for the files based on the location of the executable and where the installed location is specified in CMake.

emulation/rocjitsu/tests/CMakeLists.txt
```diff
@@ -679,6 +988,22 @@ if(HIPCC_EXECUTABLE AND HSA_RUNTIME64)
     # --- Daemon RPC tests (gtest-driven) ---
     # A gtest fixture manages the daemon lifecycle: fork+exec the daemon in
     # SetUp, run the HIP test binary as a subprocess, tear down in TearDown.
+    _rj_path_from_installed_test_bin(
+        _rj_installed_daemon_config
+        "${CMAKE_INSTALL_DATADIR}/rocjitsu/configs"
+        "amdgpu_cdna4_kmd.json"
```
RattataKing: config names are changed in https://github.com/ROCm/rocm-systems/commit/56c64d67f4f569049fe0e0f997f854264cac133b

RattataKing: if needed maybe it's possible to derive the installed config filename from the same build-tree config variable:
```python
get_filename_component(_rj_cdna4_kmd_config_name "${RJ_CDNA4_KMD_CONFIG}" NAME)
```

emulation/rocjitsu/tests/CMakeLists.txt
```diff
@@ -359,6 +610,9 @@ if(HSA_RUNTIME64)
         )
         add_dependencies(hsa_translate_test device_kernels)
         rj_configure_target(hsa_translate_test TEST)
+        if(RJ_INSTALL_TESTS)
+            rj_install_test_target(hsa_translate_test)
```
RattataKing, line 625: this installs the test binary, but some tests like `hsa_translate_test` load kernel fixtures (`.o`/`.hsaco` files) at runtime, should those also be installed with the test artifact?

RattataKing (approved):

emulation/rocjitsu/tests/CMakeLists.txt
amd-eochoalo: Thank you! There were a couple of others as well. :)

emulation/rocjitsu/tests/CMakeLists.txt
```diff
@@ -202,49 +449,40 @@ if(ROCMINFO_EXECUTABLE)
...
+            COMMAND
+                rocminfo_test
+                "--gtest_filter=RocminfoTest.${_rocminfo_test}"
+            WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+            ENVIRONMENT ${RJ_KMD_PRELOAD_ENV}
+            INSTALL_COMMAND
+                "bin/rocminfo_test"
+                "--gtest_filter=RocminfoTest.${_rocminfo_test}"
+            INSTALL_ENVIRONMENT
+                "ROCMINFO_PATH=\$ENV{ROCM_PATH}/bin/rocminfo"
```
kuhar: We should avoid forcing `ROCMINFO_PATH=$ENV{ROCM_PATH}/bin/rocminfo` when `ROCM_PATH` is unset. The installed CTest file expands that to `/bin/rocminfo`, which bypasses both PATH lookup and the fallback in `rocminfo_test.cpp`.

On my host `/bin/rocminfo` exists, so the exact unset-`ROCM_PATH` run still passed. A controlled run with `ROCMINFO_PATH` set to a bad absolute path fails with `rocjitsu: execvp failed`, so the generated environment is what decides the binary. Please leave `ROCMINFO_PATH` unset when `ROCM_PATH` is empty, or derive it from the installed/runtime prefix.

emulation/rocjitsu/cmake/rj_generate_installed_gtest_ctest.cmake
```diff
@@ -0,0 +1,87 @@
...
+            FATAL_ERROR
+            "Failed to list tests from ${ARG_TEST_EXECUTABLE}: ${_errors}"
+        )
+    endif()
+
+    file(WRITE "${ARG_OUTPUT_FILE}" "# Installed GoogleTest CTest tests\n")
+    set(_suite)
+    string(REPLACE "\n" ";" _lines "${_tests}")
+    foreach(_line IN LISTS _lines)
+        if(_line MATCHES "^([^# ]+)\\.$")
```
kuhar: We should either handle typed-test suite lines here or fail loudly on unknown non-empty `--gtest_list_tests` lines. GoogleTest prints typed suites as e.g. `MyTyped/0.  # TypeParam = int`; the current suite regex only accepts lines ending exactly at `.`, so the following cases are emitted under the previous suite.

I reproduced this with a canned listing: the script emitted two duplicate `NormalSuite.TestB` entries with `--gtest_filter=NormalSuite.TestB`. There are no typed tests in the current RocJITsu suite, so this is latent, but the parser should not silently generate wrong filters once one is added.

kuhar (commented):
`emulation/rocjitsu/tests/CMakeLists.txt:1105`: We should carry the 300s timeout for `RocblasGemmTest.Large_2048x2048` into the installed CTest entry too. The build-tree test gets it from this late `set_tests_properties(...)` call, but the generated installed `CTestTestfile.cmake` only mirrors properties passed through `rj_add_test` / `rj_add_hip_test_case`; the installed entry has no timeout.

Please pass `TIMEOUT 300` through the registration helper, or otherwise mirror late property updates, so build-tree and installed CTest behavior stay equivalent.

PR #7844, opened 2026-06-25
refactor(amdsmi): Sync with unified with known abi changes

amd-josnarlo:
## Motivation
The fabric (UALoE/IFoE) telemetry, fabric-info, and NIC firmware APIs were introduced in the in-progress ROCm 7.14.0 release but their declarations had drifted from the "unified" ABI that downstream consumers build against. The public surface carried inconsistencies that would force source churn on every caller and leave dead symbols in the header: a string-returning telemetry helper that couldn't report failure, a misspelled fabric-type enumerator (UALLINK), an intermediate versioning struct that added an extra access level, a NIC firmware typedef whose name didn't reflect that it describes a single entry, a duplicate telemetry-category enumerator, and a cluster of unreferenced max-count macros. Because the feature is still unreleased, this is the last clean opportunity to reconcile the headers with the agreed ABI before the symbols are frozen, avoiding a future breaking change once external code depends on them.

## Technical Details
- `amdsmi_fabric_telem_id_to_string()` signature changed from `const char* f(uint64_t)` to `amdsmi_status_t f(uint64_t, const char**)`, so the function can surface lookup failures via a status code and return the name through an out-parameter. Callers (the GTest, the example, and the Python interface) were updated to pass a `const char**` and check the status.
- Flattened `amdsmi_fabric_info_t` by removing the intermediate `amdsmi_fabric_info_ver_t` type. Its payload union is now the `fabric_info` member of `amdsmi_fabric_info_t`, and its version field is exposed directly as the top-level `fabric_version`. Field access simplifies from `fabric_info.fabric_version.v1.<field>` to `fabric_info.v1.<field>`, and `fabric_info.version` becomes `fabric_version`. The change is ABI-preserving: field offsets and overall struct size are unchanged, and the Python `amdsmi_get_gpu_fabric_info()` dictionary keys are unchanged.
- Typedef rename: `amdsmi_nic_fw_t` → `amdsmi_nic_fw_entry_t` (it models one firmware entry within `amdsmi_nic_fw_info_t::fw[]`), with the `AMDSmiAINICDevice` member updated to match.
- Enumerator rename: `AMDSMI_FABRIC_TYPE_UALLINK` → `AMDSMI_FABRIC_TYPE_UALINK`, including the `UALoeLinkTypeMap` entry.
- Removed symbols: the duplicate `AMDSMI_FABRIC_TELEMETRY_CATEGORY_UNKNOWN` enumerator (identical `0xFFFFFFFF` value to `..._INVALID`), and eight unreferenced public macros (`AMDSMI_MAX_VF_COUNT`, `AMDSMI_MAX_DRIVER_NUM`, `AMDSMI_DFC_FW_NUMBER_OF_ENTRIES`, `AMDSMI_MAX_WHITE_LIST_ELEMENTS`, `AMDSMI_MAX_BLACK_LIST_ELEMENTS`, `AMDSMI_MAX_TA_WHITE_LIST_ELEMENTS`, `AMDSMI_MAX_ERR_RECORDS`, `AMDSMI_MAX_PROFILE_COUNT`).
- Cascade: the ctypes wrapper (`amdsmi_wrapper.py`) and Rust bindings (`amdsmi_wrapper.rs`) were regenerated; `amdsmi_interface.py` was updated by hand; the CHANGELOG records the Changed/Removed entries.

## Issue Tracking

JIRA ID : AILITOOLS-270

## Test Plan
C++ functional tests
Python CLI test

## Test Result
All tests passed as expected


projects/amdsmi/src/amd_smi/amd_smi_ualoe.cc
```diff
@@ -561,12 +561,14 @@ static const std::unordered_map<uint64_t, const char*> telemetry_id_map = {
...
+amdsmi_status_t amdsmi_fabric_telem_id_to_string(uint64_t telem_id, const char** telem_name) {
   auto it = telemetry_id_map.find(telem_id);
   if (it != telemetry_id_map.end()) {
-    return it->second;
+    *telem_name = it->second;
+    return AMDSMI_STATUS_SUCCESS;
   }
-  return "UNKNOWN";
+  *telem_name = "UNKNOWN";
+  return AMDSMI_STATUS_API_FAILED;
```
gabrpham: We should have a null check for telem_name here which returns the invalid status if it's null.

Also, I think rather returning API FAILED here, we should instead return NOT FOUND.

projects/amdsmi/py-interface/amdsmi_interface.py
gabrpham: This is dead since we make name_ptr a ctypes.POINTER. This should be removed.

projects/amdsmi/include/amd_smi/amdsmi.h
```diff
@@ -5748,10 +5737,11 @@ amdsmi_status_t amdsmi_get_fabric_telemetry_data(amdsmi_processor_handle process
  *
  *  @param[in] telem_id The telemetry item ID for which the name is requested
  *
- *  @return const char* | Pointer to string containing the telemetry item name,
- *  or UNKNOWN if the category or telemetry ID is not recognized
+ *  @param[in] telem_name The telemetry item name
```
gabrpham: should probably be:

```suggestion
 *  @param[out] telem_name The telemetry item name
```

since we don't use it as input in that function

gabrpham (dismissed):
Please see comments below

projects/amdsmi/include/amd_smi/amdsmi.h
marifamd: fixed

projects/amdsmi/py-interface/amdsmi_interface.py
marifamd: fixed

projects/amdsmi/src/amd_smi/amd_smi_ualoe.cc
marifamd: fixed

marifamd (approved):

dmitrii-galantsev:
## Blocking issues (re-review at head `a1a1021`) — updated

Re-reviewed the squashed PR. **The four originally-flagged items are now resolved** (three fixed in `a1a1021`, one in `2e382b7`), but re-review surfaced **two behavioral regressions in the new `NOT_FOUND` telemetry contract** that still need the author's attention.

### ✅ Resolved
1. **clang-format** — include ordering in `clock_limit_read_write_test.cc` fixed (commit `2e382b7` "format").
2. **Rust wrapper drift** — the 9 removed macros (`AMDSMI_MAX_VF_COUNT` … `AMDSMI_PF_INDEX`) dropped from `amdsmi_wrapper.rs`; bindings now match the trimmed header (`a1a1021`).
3. **Dead private constants** — `_AMDSMI_STRING_LENGTH` / `_AMDSMI_MAX_STRING_LENGTH` removed from `amdsmi_interface.py` (`a1a1021`).
4. **Dangling `kSetGpuClkRange`** — orphaned `amdgpu_property_reinforcement_list` entry and the now-dead enumerator removed, so the `amdgpu_verb_check_list.at()` lookups can no longer throw `std::out_of_range` (`a1a1021`).

Minor follow-up on #2: the CHANGELOG "Removed unused public macros" bullet still lists 8 and omits `AMDSMI_PF_INDEX` — worth adding for completeness.

### 🔴 Still to fix — `NOT_FOUND` contract not tolerated by its two callers
`amdsmi_fabric_telem_id_to_string()` now returns `AMDSMI_STATUS_NOT_FOUND` for an unmapped telemetry id while still writing `"UNKNOWN"` (`src/amd_smi/amd_smi_ualoe.cc:573-574`). That's a reasonable C contract, but both updated callers treat the benign "unknown id" as a hard failure — a regression versus the old always-returns-a-string behavior:

1. **Python — `amdsmi_get_fabric_telemetry_data()` aborts the whole fetch.** The call is wrapped in `_check_res(...)` (`py-interface/amdsmi_interface.py:7031`), and `_check_res` raises `AmdSmiLibraryException` on any non-SUCCESS/RETRY/TIMEOUT code (`py-interface/amdsmi_interface_utils.py:96-97`). So a single unmapped id anywhere in the dataset throws and aborts the entire telemetry read; previously that item was just labeled `"UNKNOWN"` and iteration continued. As firmware adds ids not yet in `telemetry_id_map`, this public API starts raising instead of degrading. The `else: name_str = "UNKNOWN"` fallback (`amdsmi_interface.py:7036`) is now dead code. Suggest calling the wrapper directly and falling back to `"UNKNOWN"` on `NOT_FOUND`, raising only on genuinely unexpected codes.
2. **C++ test — `fabric_read_test.cc:236-237`** does `err = amdsmi_fabric_telem_id_to_string(it.id, &name); CHK_ERR_ASRT(err)`, and `CHK_ERR_ASRT` asserts `AMDSMI_STATUS_SUCCESS` (`tests/amd_smi_test/test_base.h:167-175`). The test fails the moment hardware reports any id absent from the map. The example got this right (`example/amd_smi_fabric_example.cc:205-210`: warn + `continue`); the test should be equally tolerant (accept `NOT_FOUND`, or don't assert on the lookup).

### 🟡 Minor / non-blocking
- **Undocumented buffer widening.** Commit `46f4ff1` repoints several getter buffers (`vendor_name`, `vram_vendor`, `subsystem_name`, compute/memory partition) from the private `80` to the public `AMDSMI_MAX_STRING_LENGTH` (`256`), and sets the length in-param to 256. Behaviorally safe, but it's a real change unrelated to the fabric/NIC ABI theme and isn't in the PR "Cascade" summary or CHANGELOG — worth a one-line note.
- **Private dlopen mirrors will drift in naming.** `projects/rccl` and `projects/rocshmem` keep the old `AMDSMI_FABRIC_TYPE_UALLINK` / `amdsmi_fabric_info_ver_t` / `.fabric_info.version` names, but they `dlopen` `libamd_smi.so.<MAJOR>` with their own ABI declarations and don't include this header, so they're unaffected — flagging so it's a conscious decision.

Positives verified at head: enum values preserved (`UALINK` == old `UALLINK` == 1; telemetry `INVALID` = 0xFFFFFFFF retained after dropping the duplicate `UNKNOWN`); `amdsmi_fabric_info_t` flatten is offset-preserving; removed public symbols (`amdsmi_get_gpu_vram_vendor`, `amdsmi_set_gpu_clk_range`, `amdsmi_get_cpusocket_handles`) are gone from header/impl/Python; both ABI Compliance Checks pass.

dmitrii-galantsev:
## Downstream consumer analysis + backward-compat fixes (head `a5a7748`)

Investigated every known consumer of the amdsmi C API and Python bindings against this PR's breaking changes (removed symbols, renamed enums/macros, the `amdsmi_fabric_info_t` flatten, the `amdsmi_fabric_telem_id_to_string` signature change, and the `amdsmi_gpu_metrics_t` counter widening).

### Consumer verdicts

| Consumer | Binding | Verdict |
|---|---|---|
| **RCCL** | hybrid include + dlopen | 🔴 was BREAKING (2 compile + 1 silent runtime) → **fixed** |
| **PAPI** (vendored + upstream) | dlopen, null-guarded | 🟢 SAFE — `vram_vendor` restored (see below) |
| **rocprofiler-systems** | links amd_smi | 🟢 SAFE — recompile only (`gpu_metrics_t` offsets), automatic under the `.so.27` SONAME bump; no source change |
| **rocprofiler-compute** | Python | 🟡 WARNS only (deprecated `compute_partition`) |
| **RDC** | links amd_smi 27.0.0 | 🟢 SAFE (already migrated, pins .27) |
| **rocSHMEM** | pure dlopen, private types | 🟢 SAFE |
| **PyTorch** | Python, defensively wrapped | 🟢 SAFE |
| **vLLM** (+ sglang, aphrodite) | Python (pins amdsmi 7.0.2) | 🟢 SAFE |
| **rocm-smi-lib, rocr-runtime, rocprofiler-sdk** | — | 🟢 SAFE |
| **hhk7734/amdsmi.go** (external cgo) | direct link | 🟢 SAFE — `vram_vendor` restored |
| **go_amd_smi / device-metrics-exporter** | cgo | 🟢 SAFE — recompile only (`gpu_metrics_t`); no source change |

Everything is 🟢 SAFE except RCCL (🔴, now fixed) and rocprofiler-compute (🟡 deprecation warnings only). The remaining "recompile only" cases involve **no source changes and no break** — the `amdsmi_gpu_metrics_t` field widening changes struct layout, and the `libamd_smi.so.26 → .so.27` SONAME bump already forces the relink in any coherent packaging (automatic in TheRock). The only residual risk is an out-of-tree binary built against pre-7.15 headers and then run against `.so.27` without rebuilding; the SONAME change is specifically what prevents that from happening silently.

### What was fixed (commits `057bc5d`, `a5a7748`)

**1. Two enum changes made backward-compatible instead of breaking.**
Retained as deprecated aliases with unchanged values (matching the PR's existing `clk_limit`/`io_bw` style), across the C header + generated Python + Rust bindings:
- `AMDSMI_FABRIC_TYPE_UALLINK` → alias of `AMDSMI_FABRIC_TYPE_UALINK`
- `AMDSMI_FABRIC_TELEMETRY_CATEGORY_UNKNOWN` → alias of `AMDSMI_FABRIC_TELEMETRY_CATEGORY_INVALID`

**2. `amdsmi_get_gpu_vram_vendor()` retained as a deprecated shim** (was removed).
It now delegates to `amdsmi_get_gpu_vram_info()` and copies `.vram_vendor` (Python still emits a `DeprecationWarning`). This keeps PAPI's `vram_vendor_hash:device=N` event alive (the dlsym lookup resolves again — no silent DEGRADE) and keeps external cgo bindings (`amdsmi.go`) linking. Restored across all layers: header, impl, Python interface + export, and both generated wrappers.

**3. Migrated RCCL — the one hard in-repo break.** Compiling against the new header (`AMDSMI_FABRIC_DIRECT=1`) it hit the two changes aliases can't cover: the flattened `amdsmi_fabric_info_t` field access, and a **silent runtime** ABI mismatch (its dlsym typedef declared `const char*(uint64_t)` vs the new `amdsmi_status_t(uint64_t, const char**)`). Typedef + caller + the `!AMDSMI_FABRIC_DIRECT` compat block updated; `AMDSMI_FABRIC_MAX_LOCAL_GPUS` bumped 8→16 to match the library.

**Generated wrappers** (`amdsmi_wrapper.py`, `amdsmi_wrapper.rs`) were regenerated with the official `tools/update_wrapper.sh` and `tools/update_rust_wrapper.sh` (Docker/bindgen), **not hand-edited** — each generator ran a full in-docker library build + `cargo build` to exit 0. `amdsmi.h` also verified to compile as C and C++, and the RCCL fabric TU compiles in both `AMDSMI_FABRIC_DIRECT` modes.

oliveiradan (approved):

dmitrii-galantsev (approved):

lamikr:
I am getting a following amdsmi install related build error when I try to build the latest therock version where I have checked out the "users/josnarlo/sync_with_unified_abi" branch for the rocm-systems.

Her are the build steps: Thet are the standard therock build steps except that I checkout the 7844 fix branch from rocm-systems:
 
```
1) git clone https://github.com/ROCm/therock.git
2) cd therock
3) python3 -m venv .venv && source .venv/bin/activate
4) pip install --upgrade pip
5) pip install -r requirements.txt
6) python build_tools/fetch_sources.py
7) cd rocm-systems
8) git checkout users/josnarlo/sync_with_unified_abi
9) cd ..
10) cmake -B build -GNinja . -DTHEROCK_AMDGPU_FAMILIES="gfx90a;gfx1100"
-DTHEROCK_AMDGPU_DIST_BUNDLE_NAME=gfx90a_gfx1100
11)cmake --build build

```

and here is the error:

```
[amdsmi install] -- Up-to-date: /home/butler/bugs/therock_6696/therock/build/core/amdsmi/stage/share/amd_smi/tests/run_amdsmi_dual_copy_test.py
[amdsmi install] -- Installing: /usr/lib/python3/dist-packages/amdsmi
[amdsmi install] CMake Error at py-interface/cmake_install.cmake:54 (file):
[amdsmi install]   file INSTALL cannot make directory "/usr/lib/python3/dist-packages/amdsmi":
[amdsmi install]   Permission denied.
[amdsmi install] Call Stack (most recent call first):
[amdsmi install]   cmake_install.cmake:67 (include)
[amdsmi install]
[amdsmi install]
[2/275] Building sub-project rocminfo (in background)
[rocminfo] [1/3] cd /home/butler/bugs/therock_6696/therock/build/core/rocminfo/build && /usr/bin/cmake -E rm -f /home/butler/bugs/therock_6696/therock/build/core/rocminfo/stamp/stage.stamp
[rocminfo] [2/3] Building CXX object CMakeFiles/rocminfo.dir/rocminfo.cc.o
[rocminfo] [3/3] Linking CXX executable rocminfo
[rocminfo SUCCEEDED in 1 seconds]
ninja: build stopped: subcommand failed.
```

nileshnegi (approved):
LGTM for RCCL changes

marifamd:
> I am getting a following amdsmi install related build error when I try to build the latest therock version where I have checked out the "users/josnarlo/sync_with_unified_abi" branch for the rocm-systems.
> 
> Her are the build steps: Thet are the standard therock build steps except that I checkout the 7844 fix branch from rocm-systems:
> 
> ```
> 1) git clone https://github.com/ROCm/therock.git
> 2) cd therock
> 3) python3 -m venv .venv && source .venv/bin/activate
> 4) pip install --upgrade pip
> 5) pip install -r requirements.txt
> 6) python build_tools/fetch_sources.py
> 7) cd rocm-systems
> 8) git checkout users/josnarlo/sync_with_unified_abi
> 9) cd ..
> 10) cmake -B build -GNinja . -DTHEROCK_AMDGPU_FAMILIES="gfx90a;gfx1100"
> -DTHEROCK_AMDGPU_DIST_BUNDLE_NAME=gfx90a_gfx1100
> 11)cmake --build build
> ```
> 
> and here is the error:
> 
> ```
> [amdsmi install] -- Up-to-date: /home/butler/bugs/therock_6696/therock/build/core/amdsmi/stage/share/amd_smi/tests/run_amdsmi_dual_copy_test.py
> [amdsmi install] -- Installing: /usr/lib/python3/dist-packages/amdsmi
> [amdsmi install] CMake Error at py-interface/cmake_install.cmake:54 (file):
> [amdsmi install]   file INSTALL cannot make directory "/usr/lib/python3/dist-packages/amdsmi":
> [amdsmi install]   Permission denied.
> [amdsmi install] Call Stack (most recent call first):
> [amdsmi install]   cmake_install.cmake:67 (include)
> [amdsmi install]
> [amdsmi install]
> [2/275] Building sub-project rocminfo (in background)
> [rocminfo] [1/3] cd /home/butler/bugs/therock_6696/therock/build/core/rocminfo/build && /usr/bin/cmake -E rm -f /home/butler/bugs/therock_6696/therock/build/core/rocminfo/stamp/stage.stamp
> [rocminfo] [2/3] Building CXX object CMakeFiles/rocminfo.dir/rocminfo.cc.o
> [rocminfo] [3/3] Linking CXX executable rocminfo
> [rocminfo SUCCEEDED in 1 seconds]
> ninja: build stopped: subcommand failed.
> ```

Not related to this PR, this is an issue on develop, but we have always required sudo for building

dmitrii-galantsev (approved):

lamikr:
With current therock version in main sudo is not needed, but the amd-smi version on rocm-systems develop requires that build-user has write permissions to folder /usr/lib/python3/dist-packages/amdsmi .

That issue is no related directly to this change, so I just finished some testing with 7844. I created first the /usr/lib/python3/dist-packages/amdsmi with root and then gave the permissions for that folder for the user that was doing the build. In that way it passed and I was able to test the amd-smi that got build.

I am fine to approve/merge this.

lamikr (approved):
tested with local build to work. requirement for the build user to have write permission to /usr/lib/python3/dist-packages/amdsmi folder is not directly related to this pr so I will approve.

lamikr:
Update, latest change fixes the requirement for build user being able to write to dir  /usr/lib/python3/dist-packages/amdsmi. Tested that build worked and amd-smi also worked on build after that change.

PR #7846, opened 2026-06-25
[rocprofiler-sdk] Fall back to cloning externals when submodule update is unavailable

venkat1361:
## Motivation

- Building rocprofiler-sdk from a source tarball can fail during CMake configure when the archive contains `.gitmodules` but has no `.git` metadata or populated submodules.
- In that case, `rocprofiler_checkout_git_submodule()` fails because the source tree is not a Git worktree. Previously this was treated as fatal, so CMake never reached the existing REPO_URL clone fallback.
- This change allows that fallback to run: if `git submodule update --init` fails and `REPO_URL` is available, CMake now warns and proceeds to clone. If no fallback URL exists, the fatal error behavior is unchanged.
- This also updates the `fmt` fallback ref from master to main, since upstream `fmt` no longer provides master.
## JIRA ID

JIRA ID : ROCM-25707

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-sdk/cmake/rocprofiler_utilities.cmake
+++ b/projects/rocprofiler-sdk/cmake/rocprofiler_utilities.cmake
@@ -229,8 +229,16 @@ function(ROCPROFILER_CHECKOUT_GIT_SUBMODULE)
         if(RET GREATER 0)
             set(_CMD "${GIT_EXECUTABLE} submodule update --init ${_RECURSE}
                 ${CHECKOUT_ADDITIONAL_CMDS} ${CHECKOUT_RELATIVE_PATH}")
-            message(STATUS "function(rocprofiler_checkout_git_submodule) failed.")
-            message(FATAL_ERROR "Command: \"${_CMD}\"")
+            if(_HAS_REPO_URL)
+                message(
+                    WARNING
+                        "Command failed: \"${_CMD}\". Falling back to cloning ${CHECKOUT_REPO_URL}."
+                    )
+                set(_TEST_FILE_EXISTS OFF)
+            else()
+                message(STATUS "function(rocprofiler_checkout_git_submodule) failed.")
+                message(FATAL_ERROR "Command: \"${_CMD}\"")
+            endif()
         else()
             set(_TEST_FILE_EXISTS ON)
         endif()
--- a/projects/rocprofiler-sdk/cmake/rocprofiler_utilities.cmake
+++ b/projects/rocprofiler-sdk/cmake/rocprofiler_utilities.cmake
@@ -234,7 +234,6 @@ function(ROCPROFILER_CHECKOUT_GIT_SUBMODULE)
                     WARNING
                         "Command failed: \"${_CMD}\". Falling back to cloning ${CHECKOUT_REPO_URL}."
                     )
-                set(_TEST_FILE_EXISTS OFF)
             else()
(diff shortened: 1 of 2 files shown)


projects/rocprofiler-sdk/cmake/rocprofiler_utilities.cmake
```diff
@@ -229,8 +229,16 @@ function(ROCPROFILER_CHECKOUT_GIT_SUBMODULE)
...
+            if(_HAS_REPO_URL)
+                message(
+                    WARNING
+                        "Command failed: \"${_CMD}\". Falling back to cloning ${CHECKOUT_REPO_URL}."
+                    )
+                set(_TEST_FILE_EXISTS OFF)
+            else()
+                message(STATUS "function(rocprofiler_checkout_git_submodule) failed.")
+                message(FATAL_ERROR "Command: \"${_CMD}\"")
+            endif()
```
venkat1361, line 240: Issue with this approach is fallback cloning uses REPO_BRANCH, not the original submodule gitlink commit SHA, because that SHA is not available without Git metadata.

projects/rocprofiler-sdk/cmake/rocprofiler_utilities.cmake
```diff
@@ -229,8 +229,16 @@ function(ROCPROFILER_CHECKOUT_GIT_SUBMODULE)
...
             set(_CMD "${GIT_EXECUTABLE} submodule update --init ${_RECURSE}
                 ${CHECKOUT_ADDITIONAL_CMDS} ${CHECKOUT_RELATIVE_PATH}")
-            message(STATUS "function(rocprofiler_checkout_git_submodule) failed.")
-            message(FATAL_ERROR "Command: \"${_CMD}\"")
+            if(_HAS_REPO_URL)
+                message(
+                    WARNING
+                        "Command failed: \"${_CMD}\". Falling back to cloning ${CHECKOUT_REPO_URL}."
+                    )
+                set(_TEST_FILE_EXISTS OFF)
```
jrmadsen: This does nothing... `_TEST_FILE_EXISTS` is already set to `OFF` here.

venkat1361: yes, removed.

MythreyaK (approved):
Looks good to me. 

Though the lines on 230-231 seem a bit confusing to me. It can go in a different PR, this looks good 👍

projects/rocprofiler-sdk/cmake/rocprofiler_utilities.cmake
```diff
@@ -229,8 +229,16 @@ function(ROCPROFILER_CHECKOUT_GIT_SUBMODULE)
         if(RET GREATER 0)
             set(_CMD "${GIT_EXECUTABLE} submodule update --init ${_RECURSE}
                 ${CHECKOUT_ADDITIONAL_CMDS} ${CHECKOUT_RELATIVE_PATH}")
-            message(STATUS "function(rocprofiler_checkout_git_submodule) failed.")
-            message(FATAL_ERROR "Command: \"${_CMD}\"")
+            if(_HAS_REPO_URL)
+                message(
+                    WARNING
+                        "Command failed: \"${_CMD}\". Falling back to cloning ${CHECKOUT_REPO_URL}."
+                    )
```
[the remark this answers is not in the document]
venkat1361: aware of this possibility

ApoKalipse-V (approved):

PR #7847, opened 2026-06-25
[rocjitsu] Fix cdna3/4 DPP semantics and fpsan numerical failures

lua1235:
## Motivation

Fix cdna3/4-specific fp8 format, MFMA NEG, and DPP row broadcast emulation issues found by the hip-fpsan suite.

## Technical Details
- Added CDNA3 FNUZ FP8/BF8 handling for generated matrix paths and shared data-type helpers.
- Fixed CDNA3/4 F64 MFMA NEG handling by passing the encoded BLGP/NEG modifier.
- Fixed DPP row-broadcast source behavior and destination write-mask cleanup for CDNA3/4, including invalid shared data plus BOUND_CTRL.
- Added runtime coverage for CDNA3 predicate-mask behavior, CDNA3/CDNA4 F64 MFMA NEG, CDNA4 generated DPP cleanup, DPP write masks, FNUZ data formats, and MAD64 carry-out.

## JIRA ID
N/A

## Test Plan
Unit tests, rocjitsu-test-corpus, fpsan tests.

Rerun the generator across all isa targets.

## Test Result

All pass

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
+++ b/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
@@ -5609,20 +5609,31 @@ class CodeGenerator:
                             )
                             _dpp_preamble = ''
                             if _is_vopc:
-                                _dpp_preamble += (
-                                    '  uint64_t dpp_old_vcc_ = wf.vcc();\n'
-                                    '  uint64_t dpp_write_mask_ = ~0ULL;\n'
-                                    '  if (inst_.src0 == amdgpu::SRC_DPP) {\n'
-                                    '    dpp_write_mask_ = 0;\n'
-                                    '    for (uint32_t ln = 0; ln < wf.wf_size(); ++ln) {\n'
-                                    '      uint32_t row = ln / 16;\n'
-                                    '      uint32_t bank = (ln % 16) / 4;\n'
-                                    '      if ((dpp_row_mask_ & (1u << row)) &&\n'
-                                    '          (dpp_bank_mask_ & (1u << bank)))\n'
-                                    '        dpp_write_mask_ |= (1ULL << ln);\n'
-                                    '    }\n'
-                                    '  }\n'
-                                )
+                                if self.isa_spec.arch_name == 'cdna3':
+                                    _dpp_preamble += (
+                                        '  uint64_t dpp_old_vcc_ = wf.vcc();\n'
+                                        '  uint64_t dpp_write_mask_ = ~0ULL;\n'
+                                        '  if (inst_.src0 == amdgpu::SRC_DPP) {\n'
+                                        '    dpp_write_mask_ = amdgpu::dpp::dpp_write_mask(\n'
+                                        '        wf.wf_size(), dpp_ctrl_, dpp_row_mask_, dpp_bank_mask_,\n'
+                                        '        dpp_bound_ctrl_);\n'
+                                        '  }\n'
+                                    )
+                                else:
+                                    _dpp_preamble += (
+                                        '  uint64_t dpp_old_vcc_ = wf.vcc();\n'
+                                        '  uint64_t dpp_write_mask_ = ~0ULL;\n'
+                                        '  if (inst_.src0 == amdgpu::SRC_DPP) {\n'
+                                        '    dpp_write_mask_ = 0;\n'
+                                        '    for (uint32_t ln = 0; ln < wf.wf_size(); ++ln) {\n'
+                                        '      uint32_t row = ln / 16;\n'
+                                        '      uint32_t bank = (ln % 16) / 4;\n'
+                                        '      if ((dpp_row_mask_ & (1u << row)) &&\n'
+                                        '          (dpp_bank_mask_ & (1u << bank)))\n'
+                                        '        dpp_write_mask_ |= (1ULL << ln);\n'
+                                        '    }\n'
+                                        '  }\n'
+                                    )
                             elif not _is_vopc:
                                 if _has_sdwa_encoding:
                                     _dpp_preamble += (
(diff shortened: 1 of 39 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/dpp_sdwa_ops.h
```diff
@@ -146,14 +146,23 @@ inline int dpp_permute(uint32_t dpp_ctrl, int lane, int wf_size, bool &out_of_bo
   }
 
   if (dpp_ctrl == ROW_BCAST15) {
-    // Broadcast lane 15 of the current row to all lanes in the row.
-    return row_num * ROW_SIZE + 15;
+    // Broadcast lane 15 of each half-wave to lanes 16-31, and lane 47 to
+    // lanes 48-63. Lanes in the lower row of each half-wave have invalid
+    // shared data.
+    if ((lane & 16) == 0) {
```
atgutier: Is this value correct for the upper half? This is true for 0-15 and 32-47.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -5590,6 +5590,10 @@ def gen_insts(self) -> None:
                         _dpp_struct, _dpp8_struct = self._vop_dpp_struct_names(
                             _enc_upper
                         )
+                        _uses_full_dpp_write_mask = self.isa_spec.arch_name in (
+                            'cdna3',
```
atgutier: What about CDNA1/2? They also have some DPP ops, do they need to be considered here?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/dpp_sdwa_ops.h
```diff
@@ -185,11 +194,38 @@ inline bool dpp_lane_masked(int lane, uint32_t row_mask, uint32_t bank_mask) {
...
+                                   uint32_t bank_mask, uint32_t bound_ctrl) {
+  if (dpp_lane_masked(lane, row_mask, bank_mask))
+    return false;
+
+  bool oob = false;
+  (void)dpp_permute(dpp_ctrl, lane, wf_size, oob);
+  return !oob || bound_ctrl != 0;
+}
+
+inline uint64_t dpp_write_mask(uint32_t wf_size, uint32_t dpp_ctrl, uint32_t row_mask,
```
atgutier, line 238: Can you also add doxygen for this too?

emulation/rocjitsu/lib/python/amdisa/codegen/execute/matrix.py
```diff
@@ -37,6 +37,18 @@ def _f8_bools(input_type: str) -> tuple[str, str]:
     )
 
 
+def _cdna3_fnuz_name(arch_name: str, name: str) -> str:
+    """Select CDNA3's FNUZ f8 helpers while preserving OCP helper names elsewhere."""
+    if arch_name == 'cdna3' and name in (
```
atgutier: Can you add a frozen set (like how we do elsewhere) for _FNUZ_FP8_ARCHES (I guess it is only CDNA3) and then do something like;

arch_name in _FNUZ_FP8_ARCHES and name in (

atgutier (commented):
I have some questions about the correctness in some cases and a few other minor comments.

Agents flagged some minor/nits:

 - MINOR matrix.py:159 — throwaway dict-comprehension over a 2-entry literal; write the dict
  directly.
  - NIT _generator.py:5613-5806 — old/new write-mask branches duplicate the whole preamble+cleanup;
  only the mask-derivation line differs.
  - NIT data_types.h:295,533 — unrelated em-dash→hyphen churn.
  - NIT test struct Fixture with methods vs the struct=data/class=behavior convention (acceptable for
  test-local fixtures).

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/dpp_sdwa_ops.h
```diff
@@ -146,14 +146,23 @@ inline int dpp_permute(uint32_t dpp_ctrl, int lane, int wf_size, bool &out_of_bo
   }
 
   if (dpp_ctrl == ROW_BCAST15) {
```
kuhar, line 156: ROW_BCAST15 still looks wrong for row 2. The new helper treats lanes 32-47 as out-of-bounds, but row_bcast:15 should broadcast lane 31 into that row; only row 0 is invalid. We should change the source calculation to use the previous row's lane 15 for every row after row 0, then update the new shared_infra tests so lanes 32 and 47 expect source lane 31 instead of OOB or old-value behavior.

kuhar (commented):
CDNA3 fp8/bf8 conversion instructions should use the same FNUZ format as MFMA/SMFMAC. This patch routes the matrix paths through `_fnuz` helpers, but the CDNA3 VOP1/VOP3 CVT generators still emit the legacy `util::fp8_e4m3_to_f32` and `util::bf8_e5m2_to_f32` aliases, which are OCP. That lets the same byte decode differently between CVT and MFMA. We should route the CDNA3 CVT, packed, and scaled conversion paths through the same arch-format selector and add coverage that CVT and MFMA agree.

Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/cdna3/vopc.cpp:263`

VCMPX_DPP also needs to preserve EXEC for lanes the DPP write mask disables. The generated bodies set both VCC and EXEC to `result`, but the new cleanup merges only VCC with the old mask; a lane masked out by row/bank or invalid-data handling keeps its old VCC bit while losing its old EXEC bit. We should snapshot old EXEC and merge it with the same `dpp_write_mask_` for VCMPX, in both generated CDNA3 and CDNA4 VOPC output.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
lua1235: Resolved, CDNA1/2 can be included

lua1235:
> VCMPX_DPP also needs to preserve EXEC for lanes the DPP write mask disables. The generated bodies set both VCC and EXEC to `result`, but the new cleanup merges only VCC with the old mask; a lane masked out by row/bank or invalid-data handling keeps its old VCC bit while losing its old EXEC bit. We should snapshot old EXEC and merge it with the same `dpp_write_mask_` for VCMPX, in both generated CDNA3 and CDNA4 VOPC output.

This also applies to the RDNA family, although their DPP is seperate from CDNA. I extended it to the general case, which slightly expands the scope of this PR.

emulation/rocjitsu/tests/data_types_test.cpp
```diff
@@ -206,6 +208,54 @@ TEST(Fp8E4M3, KeyValues) {
...
+  EXPECT_EQ(util::f32_to_fp8_e4m3_fnuz_rne(1.0f), 0x40);
+  EXPECT_EQ(util::f32_to_fp8_e4m3_fnuz_rne(240.0f), 0x7F);
+
+  float largest_denorm = util::fp8_e4m3_fnuz_to_f32(0x07);
+  float smallest_normal = util::fp8_e4m3_fnuz_to_f32(0x08);
+  float midpoint = (largest_denorm + smallest_normal) / 2.0f;
+  EXPECT_EQ(util::f32_to_fp8_e4m3_fnuz_rne(midpoint + 1e-9f), 0x08);
+}
+
+TEST(Fp8E4M3Fnuz, SrNarrow) {
```
kuhar, line 251: Please add one seed-dependent FNUZ SR assertion for each new encoder. The current `Fp8E4M3Fnuz.SrNarrow` and `Bf8E5M2Fnuz.SrNarrow` cases only call the SR helpers with `seed=0`, which makes the stochastic increment path unreachable. The existing `Fp8E5M3.SrUsesSeedForNormalAndSubnormalRounding` test is a good template: choose a value between two codes and check both a truncating seed and a rounding seed.

emulation/rocjitsu/tests/amdgpu_vm_test.cpp
```diff
@@ -1155,6 +1156,101 @@ TEST_P(IsaTest, MfmaF16AccumulationPatterned) {
...
+  auto *wf = cu->wf(0);
+  uint32_t vb = wf->vgpr_alloc().base;
+  uint32_t dst = vb + amdgpu::ACC_VGPR_OFFSET;
+  uint32_t s0 = vb + 10;
+  uint32_t s1 = vb + 20;
+  uint32_t s2 = dst;
+
+  init_mfma_f64_neg_inputs(cu, s0, s1, s2);
+
+  // CDNA f64 MFMA uses the BLGP bit range as NEG[2:0]. NEG=5 negates A and C:
```
kuhar, line 1215: Please add one f64 MFMA case that isolates the B-negate bit (`NEG=2`). The current tests use `NEG=5`, which exercises A and C negation; with A=B=C=1.0 the expected `-5` result would not change if the B-bit path were dropped or mapped incorrectly.

kuhar (commented):
Overall, looks good, just a few remaining issues. Can you also rebase this on top of origin/develop? The base ref is a bit stale. 

Review agent findings below.

Original anchor: `emulation/rocjitsu/lib/python/amdisa/codegen/execute/sema_lower.py:452`

Please carry `ctx.arch_name` into the nested `LoweringContext`s used for `if`, `for`, and `while` bodies. The top-level lowering now sets it for CDNA3 FNUZ helper selection, but those inner contexts rebuild the dataclass without `arch_name`, so a nested `cvt_f32_fp8` still lowers to `util::fp8_e4m3_to_f32` instead of `util::fp8_e4m3_fnuz_to_f32` on CDNA3. A small nested-control-flow regression test would cover this.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -5590,6 +5597,9 @@ def gen_insts(self) -> None:
                         _dpp_struct, _dpp8_struct = self._vop_dpp_struct_names(
                             _enc_upper
                         )
+                        _uses_full_dpp_write_mask = (
```
kuhar, line 5600: We should use `amdgpu::dpp::dpp_write_mask(...)` for the RDNA/GFX1250 DPP paths too, or make those paths explicitly unsupported. The generator still gates the shared helper on `_CDNA_ARCHES`, so generated RDNA code computes a row/bank-only mask while still calling `apply_dpp(..., dpp_bound_ctrl_, ...)`.

For `BOUND_CTRL=0`, invalid shared-data lanes read the old value but remain writable, so the ALU result can overwrite a lane that should preserve its old destination/VCC/EXEC state.

emulation/rocjitsu/tests/shared_infra_test.cpp
```diff
@@ -819,6 +970,313 @@ TEST(DppPermuteTest, DppRead) {
...
+}
+
+TEST(DppPermuteTest, CdnaGeneratedVopcDppWriteMaskHonorsBoundCtrl) {
+  cdna_generated_vopc_dpp_write_mask_honors_bound_ctrl<Cdna1DppTraits>();
+  cdna_generated_vopc_dpp_write_mask_honors_bound_ctrl<Cdna2DppTraits>();
+  cdna_generated_vopc_dpp_write_mask_honors_bound_ctrl<Cdna3DppTraits>();
+  cdna_generated_vopc_dpp_write_mask_honors_bound_ctrl<Cdna4DppTraits>();
+}
+
+TEST(DppPermuteTest, CdnaGeneratedVcmpxDppWriteMaskPreservesExec) {
```
kuhar, line 1273: Since this PR extends the VCMPX DPP EXEC-preservation fix to RDNA/GFX1250, we should add one behavioral RDNA Wave32 regression test. The generated-code grep in `test_generator_profile_gates.py` already checks that every arch emits the `dpp_old_exec_` / `merged_exec` pattern, so this is not an untested-code-path blocker.

The remaining gap is runtime behavior for RDNA Wave32 write-mask calculation, which the CDNA Wave64 test does not exercise.

emulation/rocjitsu/tests/data_types_test.cpp
```diff
@@ -206,6 +208,66 @@ TEST(Fp8E4M3, KeyValues) {
...
+  for (size_t i = 0; i < std::size(src); ++i) {
+    float scalar = util::fp8_e4m3_fnuz_to_f32(src[i]);
+    if (std::isnan(scalar))
+      EXPECT_TRUE(std::isnan(dst[i])) << "i=" << i;
+    else
+      EXPECT_EQ(dst[i], scalar) << "i=" << i;
+  }
+}
+
+TEST(Fp8E4M3Fnuz, RneNarrow) {
```
kuhar, line 233: Please add a few FNUZ saturation/underflow assertions for finite values, not only `inf` and exact max-finite. The current E4M3/BF8 FNUZ tests cover `inf`, `240.0f`, and `57344.0f`, but they miss the distinct finite-overflow and deep-underflow branches in `util/data_types.h`.

A local standalone check shows the current behavior is correct: `e4m3_fnuz rne(300.0f)=0x7F`, `rne(-300.0f)=0xFF`, `rne(1e-12f)=0x00`, and `bf8_e5m2_fnuz rne(70000.0f)=0x7F`. Adding those as one-liners would lock down the branches without expanding the test surface much.

kuhar (commented):
A few comments

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
lua1235: Would it be reasonable to extend the scope of this PR to DPP coverage for all archs, instead of restricting to CDNA3/4?

The main difference is that RDNA/GFX1250 introduces DPP16 and DPP8. DPP16 matches CDNA DPP, and needs to be fixed to preserve its old EXEC state. DPP8 does not have these masks, so is fine as is.

One other gap I found is that RDNA DPP16 and DPP8 add a FI instruction field to control fetch inactive lane behavior, which is currently decoded but ignored by implementation. 

I can fix this as well, and extend test coverage as necessary in this PR.

kuhar: OK, let's land that separately.

kuhar (approved):

PR #7851, opened 2026-06-25
[RCCL] Add multi-node hierarchical ReduceScatter algorithm

nawrinsu:
## Motivation

Implement a multi-phase hierarchical ReduceScatter algorithm that decomposes the collective into:
1. Local data shuffle kernel (pre-shuffle send buffer from node-major to local-rank-major)
2. Intra-comm RS (Ring)
3. Intra-comm RS (Direct/PAT/Ring)

This decomposition helps reduce the traffic over network.
* Hierarchical RS is disabled by default
* Use env var `RCCL_HIERARCHICAL_REDUCE_SCATTER=1` to enable for >= 8 nodes
* Enabled for message sizes up to 128MB for >= 16 nodes and 64MB >= 8 nodes
* Allocate 128MB temporary buffer

Additional changes:
* Direct RS use ncclEnqueueCheck instead of the loop wrapped in ncclGroupStart/End.
* rccl-tests `-A` option will show `hier` as algo and inter-comm protocol and channels

## JIRA ID

JIRA ID: AICOMRCCL-763
## Test Plan

Performance and functionality tested on MI350 cluster.
Unit tests added - 
- Selection logic of hierarchical RS. Function tested is `rcclUseHierarchicalReduceScatter`.
- hierarchical shuffle kernel
- temp buffer allocation
- algo/proto/channel display

## Test Result

Hierarchical RS shows up to 47% improvement on 8 nodes and up to 68% improvement on 16 nodes over default RCCL RS. Performance number collected on MI350 nodes.

## Submission Checklist


projects/rccl/src/collectives.cc
```diff
@@ -535,6 +534,98 @@ ncclResult_t ncclReduce_impl(const void* sendbuff, void* recvbuff, size_t count,
...
+    int chunkSteps, int sliceSteps) {
+  if (recvcount == 0) return ncclSuccess;
+  void* tempbuff = comm->tempBuff;
+
+  comm->enableDirectReduceScatter = 1;
+  struct ncclInfo infoP2P = { ncclFuncReduceScatter, "ReduceScatter",
+    sendbuff, tempbuff, recvcount, datatype, op, 0, comm, stream,
+    chunkSteps, sliceSteps, nullptr };
+  infoP2P.useDirect = true;
+  NCCLCHECK(ncclEnqueueCheck(&infoP2P));
```
nusislam: You moved the Direct RS to enqueue. I assume you have manually tested Direct RS is working properly.

projects/rccl/src/collectives.cc
```diff
@@ -535,6 +534,98 @@ ncclResult_t ncclReduce_impl(const void* sendbuff, void* recvbuff, size_t count,
...
+    tempBuf, intraOut, intraOutCount, datatype, op, 0, intraComm, stream,
+    REDUCESCATTER_CHUNKSTEPS,
+    intraComm->rcclUseOneSlice ? REDUCESCATTER_SLICESTEPS_SINGLE_NODE : REDUCESCATTER_SLICESTEPS,
+    nullptr };
+  NCCLCHECK(ncclEnqueueCheck(&infoIntraRS));
+
+  // Step 3: Inter-node ReduceScatter  intraOut -> recvbuff.
+  size_t interMsgSize = recvcount * typeSize * (size_t)nNodes;
+  interComm->enableDirectReduceScatter = 0;
+  if (rcclUseReduceScatterDirect(interComm, interMsgSize)) {
```
nusislam, line 856: Up to what node count is direct being used for `inter-comm`?

projects/rccl/src/enqueue.cc
```diff
@@ -3611,6 +3621,15 @@ static ncclResult_t taskAppend(struct ncclComm* comm, struct ncclInfo* info) {
             NCCLCHECK(p2pTaskAppend(comm, info, ncclFuncSend, collAPI, (void*)info->sendbuff, info->count, info->datatype, r, allowUB));
             NCCLCHECK(p2pTaskAppend(comm, info, ncclFuncRecv, collAPI, (void*)((char*)info->recvbuff + r*rankOffset), info->count, info->datatype, r, allowUB));
           }
+        } else if (info->coll == ncclFuncReduceScatter && info->useDirect) {
+          NCCLCHECK(ncclRegFind(comm, info->sendbuff, comm->nRanks * info->count * ncclTypeSize(info->datatype), &sendReg));
+          NCCLCHECK(ncclRegFind(comm, info->recvbuff, comm->nRanks * info->count * ncclTypeSize(info->datatype), &recvReg));
+          allowUB = captured || (sendReg != NULL && recvReg != NULL);
+          size_t rankOffset = info->count * ncclTypeSize(info->datatype);
+          for (int r=0; r<comm->nRanks; r++) {
+            NCCLCHECK(p2pTaskAppend(comm, info, ncclFuncSend, collAPI, (void*)((char*)info->sendbuff + r*rankOffset), info->count, info->datatype, r, allowUB));
```
nusislam: I see the `p2p` transfers being issued for direct RS. Where are we doing the local reduction?

projects/rccl/src/collectives.cc
nawrinsu: Yes, I tested Direct RS for 4 and 8 nodes.

projects/rccl/src/collectives.cc
nawrinsu: Direct will be used up to 16 nodes as direct RS is enabled up to 16 nodes.

projects/rccl/src/enqueue.cc
nawrinsu: The local reduction is enqueued from collectives.cc via ncclEnqueueCheck() immediately after the p2p staging enqueue.

srirajpaul (approved):

AbandiGa (approved):
LGTM

projects/rccl/src/rccl_wrap.cc
```diff
@@ -626,6 +668,23 @@ bool rcclUseReduceScatterDirect(struct ncclComm* comm, size_t& msgSize) {
   return false;
 }
 
+RCCL_PARAM(HierarchicalReduceScatter, "HIERARCHICAL_REDUCE_SCATTER", 0);
```
thomas-huber, line 727: It seems that the four `Enabled_*` test cases expect this to be true, but the tests themselves never seem to set `RCCL_HIERARCHICAL_REDUCE_SCATTER=1`.

projects/rccl/src/rccl_wrap.cc
```diff
@@ -626,6 +668,23 @@ bool rcclUseReduceScatterDirect(struct ncclComm* comm, size_t& msgSize) {
...
+
+bool rcclUseHierarchicalReduceScatter(struct ncclComm* comm, size_t msgSize) {
```
thomas-huber, line 729: OPUS 5 flagged this:

2. User-defined reduction ops will fail with ncclInvalidArgument. An op from ncclRedOpCreatePreMulSum is a comm-scoped handle, mangled by the comm pointer:

ncclHierarchicalReduceScatter_Impl forwards op verbatim to intraComm and interComm, so ArgsCheck (src/misc/argcheck.cc:214) rejects it with "reduction operation N unknown to this communicator". Even if the handle resolved, applying a PreMulSum scalar at both levels would multiply by the scalar twice. A working NCCL program breaks the moment this env var is turned on. The fix is one line in rcclUseHierarchicalReduceScatter: bail out when int(op) >= int(ncclNumOps), which means threading op into that predicate and into rcclGetAlgoInfo's call.

projects/rccl/src/device/hierarchical_shuffle.h
```diff
@@ -0,0 +1,65 @@
...
+                                                                  size_t rankOffset, int cols, int rows) {
+  int totalPairs = rows * cols;
+  size_t numVec = rankOffset / sizeof(v4u);
+
+  for (int pair = blockIdx.x; pair < totalPairs; pair += gridDim.x) {
+    int i = pair / cols;
+    int j = pair - i * cols;
+    int dstIdx = j * rows + i;
+
+    v4u_gptr srcV = (v4u_gptr)(src + (size_t)pair * rankOffset);
```
thomas-huber: Looks like we are assuming 16-byte alignment but that might not be guaranteed/required as the code is now. 

I think we need to add `rankOffset % 16 == 0` to the eligibility check or handle the potential misalignment

thomas-huber:
With both flags on, the parent comm holds 128MB (AG) + 128MB (RS) + 16MB (tempBuff), plus 16MB for each of intraComm and interComm. The AG and RS temp buffers can never be live simultaneously on a given stream, is there a reason not to share one allocation?

projects/rccl/src/collectives.cc
```diff
@@ -666,6 +665,141 @@ ncclResult_t ncclReduce_impl(const void* sendbuff, void* recvbuff, size_t count,
...
+                                 stream,
+                                 REDUCESCATTER_CHUNKSTEPS,
+                                 intraComm->rcclUseOneSlice ? REDUCESCATTER_SLICESTEPS_SINGLE_NODE :
+                                                              REDUCESCATTER_SLICESTEPS,
+                                 nullptr};
+  NCCLCHECK(ncclEnqueueCheck(&infoIntraRS));
+
+  // Step 3: Inter-node ReduceScatter  intraOut -> recvbuff.
+  size_t interMsgSize = recvcount * typeSize * (size_t)nNodes;
+  interComm->enableDirectReduceScatter = 0;
```
thomas-huber: I think this is dead code since `rcclUseReduceScatterDirect` doesn't read the flag

projects/rccl/src/collectives.cc
```diff
@@ -666,6 +665,141 @@ ncclResult_t ncclReduce_impl(const void* sendbuff, void* recvbuff, size_t count,
...
+  size_t rankOffset = recvcount * typeSize;
+
+  void* tempBuf = comm->hierarchicalRSTempBuffer;
+  size_t intraOutCount = recvcount * nNodes; // per-rank intra-RS output
+  void* intraOut = (char*)tempBuf + (size_t)intraComm->rank * intraOutCount * typeSize;
+
+  // Step 1: Pre-shuffle sendbuff (node-major) -> tempBuf (local-rank-major).
+  size_t totalBytes = (size_t)nNodes * localRanks * rankOffset;
+  int numBlocks = hierarchicalShuffleNumBlocks(totalBytes);
+  hierarchicalShuffle<<<numBlocks, 256, 0, stream>>>((const char*)sendbuff, (char*)tempBuf, rankOffset, localRanks,
```
thomas-huber: Should we have a constant for 256? Might be overkill

projects/rccl/src/rccl_wrap.cc
alex-breslow-amd: Good.  Off by default.  For new features, we will be changing the process so that new features stay off by default.

nawrinsu:
@thomas-huber I addressed your comments. Please take a look. Thanks.

projects/rccl/src/init.cc
```diff
@@ -2748,12 +2750,19 @@ static ncclResult_t ncclCommInitRankFunc(struct ncclAsyncJob* job_) {
         comm->forcePatEnable = !userDisabledPat && !rcclUseAinic();
         NCCLCHECKGOTO(ncclCommSplit(comm, local_rank, node_id, &comm->hierarchicalInterComm, NULL), res, fail);
         comm->forcePatEnable = false;
-        size_t tempBufSize = (comm->nNodes >= 32) ? HIERARCHICAL_AG_TEMP_BUFFER_SIZE
-                             : (comm->nNodes >= 16) ? HIERARCHICAL_AG_TEMP_BUFFER_SIZE / 2
-                                                    : HIERARCHICAL_AG_TEMP_BUFFER_SIZE / 4;
-        NCCLCHECKGOTO(ncclCudaMalloc(&(comm->hierarchicalAGTempBuffer), tempBufSize, comm->memManager), res, fail);
+        // inherit PXN disable from parent comm
+        comm->hierarchicalInterComm->pxnDisable = comm->pxnDisable;
```
thomas-huber, line 2754: Seems like this overrides PXN after the inter-comms topology and transports were initialized, is this safe?

projects/rccl/src/init.cc
```diff
@@ -2748,12 +2750,19 @@ static ncclResult_t ncclCommInitRankFunc(struct ncclAsyncJob* job_) {
...
+        size_t tempBufSize;
+        if (rcclParamHierarchicalReduceScatter() == 1) {
+          tempBufSize = (comm->nNodes >= 16) ? HIERARCHICAL_TEMP_BUFFER_SIZE : HIERARCHICAL_TEMP_BUFFER_SIZE / 2;
+        } else {
+          tempBufSize = (comm->nNodes >= 32) ? HIERARCHICAL_TEMP_BUFFER_SIZE :
+                        (comm->nNodes >= 16) ? HIERARCHICAL_TEMP_BUFFER_SIZE / 2 :
+                                               HIERARCHICAL_TEMP_BUFFER_SIZE / 4;
+        }
```
thomas-huber: nit: trailing whitespace

projects/rccl/src/include/comm.h
```diff
@@ -692,8 +692,10 @@ struct ncclComm {
   struct ncclComm* hierarchicalInterComm;
   bool hierarchicalCommsInitialized;
 
-  // Hierarchical AG temporary buffers
-  void* hierarchicalAGTempBuffer;
+  // Hierarchical temporary buffer
+  // Both hierarchical AG and RS use the same temp buffer,
+  // so running them concurrently on same communicator across two streams is unsafe
+  void* hierarchicalTempBuffer;
```
thomas-huber, line 698: Should we implement some sort of split or safe guard to prevent a race when users enqueue ops on one communicator with different streams?

projects/rccl/src/rccl_wrap.cc
```diff
@@ -406,6 +406,45 @@ ncclResult_t rcclGetAlgoInfo(struct ncclComm* comm, ncclFunc_t coll, uint64_t co
...
+    ncclComm* interComm = comm->hierarchicalInterComm;
+    ncclComm* intraComm = comm->hierarchicalIntraComm;
+    int nNodes = interComm->nRanks;
+
+    size_t interMsgSize = count * ncclTypeSize(dataType) * nNodes;
+    if (rcclUseReduceScatterDirect(interComm, interMsgSize)) {
+      *protocol = NCCL_PROTO_SIMPLE;
+      *maxChannels = interComm->p2pnChannels;
+    } else {
+      struct ncclTaskColl task;
```
thomas-huber: Seems like this and the other `ncclTaskColl` instances should be initialized. I think `getAlgoInfo` can read un-initialized fields as things currently stand.

projects/rccl/src/collectives.cc
```diff
@@ -759,6 +758,141 @@ ncclResult_t ncclReduce_impl(const void* sendbuff, void* recvbuff, size_t count,
...
+
+enum rcclReduceScatterAlgo {
+  RCCL_RS_RING,
+  RCCL_RS_DIRECT,
+  RCCL_RS_HIERARCHICAL
+};
+
+static rcclReduceScatterAlgo rcclSelectReduceScatterAlgo(struct ncclComm* comm, size_t msgSize, ncclRedOp_t op) {
+  // User-defined op is not supported for hierarchical reduce scatter
+  if (ncclGroupDepth == 0 && (int(op) < int(ncclNumOps)) && rcclUseHierarchicalReduceScatter(comm, msgSize)) {
```
thomas-huber: Hey looks like this allows `ncclAvg`. Is there any potential issue where performing the average separately in the intra and internode stage might produce a different result than averaging once across all ranks?

nawrinsu: Good point. `avg` with `float` type is actually exact for hierarchical but `Integer` types aren't. Avg lowers to `ncclDevSumPostDiv`, so each node truncates its partial sum before inter-node sees it. There's also a pre-existing issue where direct RS only handles `ncclSum` and inter-comm RS uses Direct RS, so I've gated the hierarchical to only `ncclSum`.

projects/rccl/src/include/comm.h
nawrinsu: My understanding is that applications aren't expected to run collectives concurrently on the same communicator. Two overlapping calls would already race on the intraComm/interComm channel and proxy state, not just the shared scratch, so a buffer-level guard wouldn't make that case safe.
@nusislam, any insight here?

projects/rccl/src/init.cc
nawrinsu: Yes, it's a selection gate override, not a topology change. The only post-init reader is the Direct RS gate, which is the point for this override: the 8-rank inter-comm falls below rcclSetPxn's gfx950 threshold of 32 even though the 64-rank parent has PXN enabled. The other two readers are gated on maxLocalRanks > 1, which is 1 here.

projects/rccl/src/include/comm.h
nusislam: NCCL does not support multiple concurrent collectives in the same communicator.

thomas-huber (approved):

projects/rccl/src/device/hierarchical_shuffle.h
```diff
@@ -0,0 +1,72 @@
...
+      }
+      copied += numWords * WordBytes;
+    }
+
+    for (size_t b = copied + threadIdx.x; b < rankOffset; b += blockDim.x) {
+      dstTile[b] = srcTile[b];
+    }
+  }
+}
+
```
alex-breslow-amd, line 71: Can you add a unit test to validate that this does what you want it to?  I didn't see one.  Let me know if I missed it.

projects/rccl/src/rccl_wrap.cc
```diff
@@ -691,6 +733,23 @@ bool rcclUseReduceScatterDirect(struct ncclComm* comm, size_t& msgSize) {
   return false;
 }
 
+RCCL_PARAM(HierarchicalReduceScatter, "HIERARCHICAL_REDUCE_SCATTER", 0);
+
+bool rcclUseHierarchicalReduceScatter(struct ncclComm* comm, size_t msgSize) {
+  if (comm->nNodes < 8) return false;
+  if (rcclParamHierarchicalReduceScatter() != 1) return false;
+  if (!comm->hierarchicalCommsInitialized) return false;
```
alex-breslow-amd: This might be cleaner as an if any statement.  Not blocker.

projects/rccl/src/init.cc
```diff
@@ -2748,12 +2750,19 @@ static ncclResult_t ncclCommInitRankFunc(struct ncclAsyncJob* job_) {
...
```
alex-breslow-amd: This logic looks similar to logic you have below where you are dividing by constants.  It seems like there's an opportunity to add a function.  Then you can unit test the function too to verify it behaves as expected.

projects/rccl/src/rccl_wrap.cc
```diff
@@ -406,13 +406,52 @@ ncclResult_t rcclGetAlgoInfo(struct ncclComm* comm, ncclFunc_t coll, uint64_t co
...
+      intraProto = task.protocol;
+      intraChan = task.nMaxChannels;
+    }
+
+    // For hierarchical algorithm, only the inter-comm protocol/channels are
+    // reported in rccl-tests -A output.
+    // The intra-comm values are logged below for debugging purposes
+    INFO(NCCL_COLL, "Hierarchical RS inter: proto=%d channels=%d, intra: proto=%d channels=%d", *protocol, *maxChannels,
+         intraProto, intraChan);
+    return ncclSuccess;
```
alex-breslow-amd: This seems function worthy and test worthy

projects/rccl/src/collectives.cc
```diff
@@ -312,12 +312,11 @@ static ncclResult_t ncclHierarchicalAllGather_Impl(const void* sendbuff, void* r
...
   }
 
-  // Step 3: Shuffle tempBuffer to recvbuff
+  // Step 3: Shuffle tempBuffer (local-rank-major) -> recvbuff (node-major).
   size_t totalAGBytes = (size_t)nNodes * localRanks * rankOffset;
   int numBlocks = hierarchicalShuffleNumBlocks(totalAGBytes);
-  int threadsPerBlock = 1024;
-  hierarchicalAGShuffle<<<numBlocks, threadsPerBlock, 0, stream>>>((const char*)tempBuffer, (char*)recvbuff, rankOffset,
-                                                                   nNodes, localRanks);
+  hierarchicalShuffle<<<numBlocks, HIERARCHICAL_SHUFFLE_THREADS, 0, stream>>>((const char*)tempBuffer, (char*)recvbuff,
```
alex-breslow-amd, line 318: QQ: What's the maximum number of GPU kernels that are in flight at a time on a GPU from this hierarchical algorithm implementation?  My reading is 1.  Is that correct?

projects/rccl/src/device/hierarchical_shuffle.h
nawrinsu: Test added

projects/rccl/src/collectives.cc
nawrinsu: Yes, that's correct.

alex-breslow-amd: Ok, that allays my potential concerns.  Thanks.

projects/rccl/src/rccl_wrap.cc
```diff
@@ -480,6 +494,9 @@ ncclResult_t rcclGetAlgoName(int algo, const char** algoName) {
     case rcclAddonAlgos_t::RCCL_HIERARCHICAL_ALLGATHER:
       *algoName = "Hier";
       break;
+    case rcclAddonAlgos_t::RCCL_HIERARCHICAL_REDUCESCATTER:
+      *algoName = "Hier";
```
alex-breslow-amd, line 498: Thank you for adding this. :-)

projects/rccl/src/rccl_wrap.cc
```diff
@@ -518,22 +535,38 @@ bool rcclUseAlltoAllGda(struct ncclComm* comm) {
   return false;
 }
 
+size_t rcclHierarchicalTempBufferSize(int nNodes, bool allGather, bool reduceScatter) {
```
alex-breslow-amd, line 538: Nice, thank you. :-)

alex-breslow-amd (approved):
LGTM, thank you for the quick turnaround time on my feedback.

PR #7852, opened 2026-06-25
test(rccl): Add test for p2p buffer reuse bug fixed with NCCL 2.29.2

mch:
## Motivation

The fix for the P2P-then-collective same-buffer crash (NCCL GH#1859) was already
brought into RCCL with the NCCL 2.29.2 sync, but that sync left behind a now-dead
block of code in `src/transport/p2p.cc`. The primary goal of this PR is to add a
regression test that covers the fixed behaviour so the crash cannot silently
return; it also removes the leftover redundant block.

## Technical Details

In `src/transport/p2p.cc`, `ipcRegisterBuffer`'s `NCCL_IPC_COLLECTIVE` branch
contained a leftover block that allocated `devPeerRmtAddrs` and only copied
`hostPeerRmtAddrs → devPeerRmtAddrs` when `needUpdate == true`. With the 2.29.2
fix already in place this block is redundant and is removed here.

The new regression test in `test/RegisterTests.cpp` reproduces the exact
P2P-then-collective sequence on a shared registered buffer. Because the relevant
code path is only reached when ranks live in separate processes
(`directMode == false`), the test forks one process per rank and uses
`ncclCommInitRank` with a shared `ncclUniqueId` rather than `ncclCommInitAll`.
The test is skipped when fewer than 2 GPUs are visible or when the GPUs are not
direct-P2P capable.

## JIRA ID

JIRA ID: AICOMRCCL-1100

## Test Plan

Build RCCL and run the new `RegisterTests` regression case on a ≥2-GPU,
P2P-capable node. To verify the test actually catches the defect, a mutant 
can be introduced that restores the original bug:

```
-      if (needUpdate) {
+      if (needUpdate && type == NCCL_IPC_COLLECTIVE) {
```

## Test Result

With the mutant in place the new test reproduces the crash (illegal memory
access); on the fixed tree the P2P-then-collective sequence completes with
`ncclSuccess` and correct AllReduce output. Full Register suite passes.


projects/rccl/src/transport/p2p.cc
corey-derochie-amd, line 1: ~This appears to restore a purposeful AMD workaround back to original NCCL code. Is it clear why this workaround is no longer valid?~
I read your description, looks like the NCCL sync missed this deletion, is that correct?

projects/rccl/test/RegisterTests.cpp
```diff
@@ -23,6 +31,42 @@ static bool hasGpuAvailable() {
...
+
+// Queries every visible device and returns true only if all of them are a
+// supported architecture for the P2P+collective registration test.
+static bool allDevicesSupportP2pCollTest(int numDevices, char* unsupportedArchOut,
+                                         size_t outLen) {
+    for (int dev = 0; dev < numDevices; dev++) {
+        hipDeviceProp_t prop;
+        if (hipGetDeviceProperties(&prop, dev) != hipSuccess) {
+            return false;
+        }
```
corey-derochie-amd, line 58: So, this isn't the behaviour under test, just a prerequisite that we assume must succeed. If this returns false it's not clear that this is what failed. Does gtest support a better way to fail under this condition?

corey-derochie-amd (approved):
lgtm, just a couple questions.

corey-derochie-amd:
I'm a bit apprehensive about the test plan, I probably misunderstand it. What is the relationship to `NCCL_LOCAL_REGISTER`? If that flag is required to reproduce the issue under test, should it be set within the test code?

projects/rccl/src/transport/p2p.cc
mch: Yes, it was a missed deletion.

projects/rccl/test/RegisterTests.cpp
mch: If this returns false, the message on line 338 will be printed ("Unsupported GPU architecture...") and the child process will update the shared state to say that it skipped. If any child skips, that is handled on line 518, and will cause GTest to show that the test was skipped and that the precondition was not met.

mch:
> I'm a bit apprehensive about the test plan, I probably misunderstand it. What is the relationship to `NCCL_LOCAL_REGISTER`? If that flag is required to reproduce the issue under test, should it be set within the test code?

I revised the test plan to be clearer: `NCCL_LOCAL_REGISTER=1` is set within the test itself, line 600 of RegisterTests.cpp.

projects/rccl/test/RegisterTests.cpp
mch: It looks like this on my laptop for example:
```
[ RUN      ] Register.ProcessIsolatedRegisterTests
/home/mahughes/Work/ROCm/rocm-systems/projects/rccl/test/RegisterTests.cpp:519: Skipped
Test requires 2 supported GPUs (gfx942 / gfx950) with direct GPU-to-GPU P2P access; preconditions not met.
```

mch:
> 
> 
> * ❌ Unit Test
> 

This is incorrect, the bot is not matching RCCL test filename conventions. This PR specifically adds tests.

PR #7371, opened 2026-06-17
rocrtst: Fix/omit some tests running in FFM (#1075)

cfreeamd:
Also, added a summary at the end of test runs that show which tests were skipped due to platform filtering.

PR #7853, opened 2026-06-25
rocr: Improvements when running rocrtst with FFM 

cfreeamd:
- Added and use wildcards for rocrtst filtering of tests not to run
- Fixed how some tests use rocrtst filtering

### DIFF
--- a/projects/rocr-runtime/rocrtst/config/platform_config.yaml
+++ b/projects/rocr-runtime/rocrtst/config/platform_config.yaml
@@ -182,4 +182,11 @@ platforms:
   # FFM Simulator - only Core and Extended groups allowed
   FFM_SIMULATOR:
     allowed_groups: [Core, Extended]
-    blocked_tests: []
+    blocked_tests:
+      - rocrtstFunc.Memory_Max_Mem
+      - rocrtstFunc.Memory_Available  # Uses max & available mem; not accurate
+      - rocrtstPerf.Memory_Async_Copy # Requires hwloc support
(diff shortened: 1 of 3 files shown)


shwetagkhatri (approved):
LGTM

projects/rocr-runtime/rocrtst/config/platform_config.yaml
```diff
@@ -182,4 +182,11 @@ platforms:
   # FFM Simulator - only Core and Extended groups allowed
   FFM_SIMULATOR:
     allowed_groups: [Core, Extended]
-    blocked_tests: []
+    blocked_tests:
```
cfreeamd, line 185: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** PR #7371 independently modifies this same FFM_SIMULATOR `blocked_tests` section from the same base commit, listing the individual `rocrtstNeg.Queue_Validation_*` tests (plus `Memory_Max_Mem`, `Memory_Available`, `Memory_Async_Copy`). These PRs will conflict at merge time. If this PR lands first (adding fnmatch infrastructure and the `rocrtstNeg.*` wildcard), then #7371 will need to be rebased: its specific `rocrtstNeg.Queue_Validation_*` entries can be dropped (superseded by the wildcard), but the three `Memory_*` blocked entries must be preserved.

cfreeamd (commented):

`platform_filter.cc` fnmatch integration looks correct. `signal_allocation_validation.cc` early-return guard is the right pattern.

PR #7857, opened 2026-06-26
[rocjitsu] Track implicit operands, enable PC_OPERAND flag and remove DBI denylist

accauble:
## Motivation

DBI cannot currently relocate PC-relative instructions or instructions that read/write the PC because  this requires more care. PC read/writes were originally rejected through a fragile denylist based on mnemonic spelling. This refactors the denylist by adding a `is_pc_operand` function to `instruction.h` which checks if the PC is an operand of the instruction (i.e., if it reads or writes the PC).

## Technical Details

* `is_pc_operand` just checks the new `PC_OPERAND` flag, which gets set by instructions that have the PC as in implicit operand.
* I couldn't find an example of the PC being written through an explicit operand so I `is_pc_operand` does not search the explicit operands.
* To set PC_OPERAND, I needed to add `implicit_operands` to gpuisa, since `operands` only included explicit ones.
* I chose to set a PC_OPERAND flag rather than something broader that might include all branches (which would change the PC) since branches are already covered by their own flags and because this was not possible to gather from the MR ISA. 

## Test Plan

* Implement amdisa tests for `implicit_operands` and `PC_OPERAND`.
* Update Instrumentation tests to forgo denylist and make sure that the instructions originally denied are still being denied.
* Make sure all amdisa tests pass
* Make sure all overall tests pass

## Test Result

All tests pass.

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/docs/dbi-design.md
+++ b/emulation/rocjitsu/docs/dbi-design.md
@@ -120,9 +120,11 @@ Rules enforced:
 - `anchor.size()` is 4 or 8 and fits inside `text_bytes` (subtraction-based bounds check; resists overflow when `anchor_offset` is huge).
 - `anchor.raw_encoding()` is non-null.
 - `anchor` is not a branch / cond branch / indirect branch / indirect call / program terminator, and `branch_offset_bytes()` is `nullopt`.
-- `anchor.mnemonic()` is not on the small PC-relative denylist (`s_getpc_b64`, `s_call_b64`, `s_setpc_b64`, `s_swappc_b64`, `s_rfe_*`) — these may not surface as flag bits on every ISA.
+- `anchor.is_pc_operand()` is false — i.e. the instruction does not read or write the program counter (`s_getpc` / `s_setpc` / `s_swappc` / `s_call` / `s_addpc`, the `s_rfe` family, and `s_cbranch_g_fork`/`join`). Relocating such an instruction's bytes would change the PC value it reads or the target it computes.
 
-`arch` is accepted now so a future ISA-specific denylist can grow without an API change.
+The PC check uses the `PC_OPERAND` `InstFlag`, a structural property set in the ISA layer from the MR ISA spec's implicit `OPR_PC` operand (captured by the `amdisa` parser as `Instruction.has_pc_operand` and emitted into the generated per-ISA constructors). `s_call` / `s_setpc` / `s_swappc` are additionally caught by their `INDIRECT_BRANCH` / `INDIRECT_CALL` flags above. PC-relative *branches* are deliberately **not** `PC_OPERAND` — they carry a label operand, not `OPR_PC`, and are handled by the branch/`branch_offset_bytes()` rules instead.
+
+`arch` is accepted now so a future ISA-specific check can grow without an API change.
 
 ### `validate_anchor(anchor, anchor_offset, text_bytes, pt, arch, error_out)`
 
--- a/emulation/rocjitsu/lib/python/amdisa/parser.py
+++ b/emulation/rocjitsu/lib/python/amdisa/parser.py
@@ -864,6 +864,9 @@ class Parser:
                     continue
                 opcode = int(xs.get_node_text(opcode_node))
                 opnds = []
+                # Operands the spec declares implicit and are currently
+                # excluded from `opnds`.
+                implicit_opnds = []
                 for opnd in operands_node:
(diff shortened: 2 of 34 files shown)


accauble:
~~Note: This is currently failing CI because of a pre-existing issue that will be fixed in PR #8149.~~

emulation/rocjitsu/lib/python/amdisa/parser.py
```diff
@@ -894,14 +897,33 @@ class with a literal constant).
                                 order,
                             )
                         )
+                    else:
```
kuhar, line 900: We should make the `implicit_operands` contract match the parser split here, or pin the invariant with a parser/spec test. This branch classifies operands by the absence of `<FieldName>`, not by the XML `IsImplicit` attribute: the checked-in MR ISA XML has many field-less operands with `IsImplicit="false"` (for example `OPR_VCC` on `V_CNDMASK_B32`), while the current `OPR_PC` operands are all field-less.

Since `Instruction.has_implicit_operand('OPR_PC')` now drives `PC_OPERAND`, a small parser-level/spec-driven test for `OPR_PC` classification, or wording that explicitly says "field-less operands", would keep future callers from assuming `implicit_operands == is_implicit`.

emulation/rocjitsu/docs/dbi-design.md
```diff
@@ -120,9 +120,11 @@ Rules enforced:
...
-- `anchor.mnemonic()` is not on the small PC-relative denylist (`s_getpc_b64`, `s_call_b64`, `s_setpc_b64`, `s_swappc_b64`, `s_rfe_*`) — these may not surface as flag bits on every ISA.
+- `anchor.is_pc_operand()` is false — i.e. the instruction does not read or write the program counter (`s_getpc` / `s_setpc` / `s_swappc` / `s_call` / `s_addpc`, the `s_rfe` family, and `s_cbranch_g_fork`/`join`). Relocating such an instruction's bytes would change the PC value it reads or the target it computes.
 
-`arch` is accepted now so a future ISA-specific denylist can grow without an API change.
+The PC check uses the `PC_OPERAND` `InstFlag`, a structural property set in the ISA layer from the MR ISA spec's implicit `OPR_PC` operand (captured by the `amdisa` parser as `Instruction.has_pc_operand` and emitted into the generated per-ISA constructors). `s_call` / `s_setpc` / `s_swappc` are additionally caught by their `INDIRECT_BRANCH` / `INDIRECT_CALL` flags above. PC-relative *branches* are deliberately **not** `PC_OPERAND` — they carry a label operand, not `OPR_PC`, and are handled by the branch/`branch_offset_bytes()` rules instead.
```
kuhar, line 125: Please update this doc reference to the helper that actually exists: `Instruction.has_implicit_operand('OPR_PC')`, not `Instruction.has_pc_operand`.

emulation/rocjitsu/tests/patch/instrumentor_test.cpp
```diff
@@ -236,25 +236,82 @@ TEST(Validator, RejectsForceFullExec) {
...
+  // The validator's verdict must follow the flag: PC_OPERAND anchors rejected,
+  // ordinary ones accepted. Use a .text big enough for the (4- or 8-byte) inst.
+  const auto size = static_cast<size_t>(inst->size());
+  std::vector<uint8_t> text(size + 8, 0xCCu);
+  std::string err;
+  const bool ok = is_relocatable_anchor(*inst, 0, text, tc.arch, &err);
+  EXPECT_EQ(ok, !tc.expect_pc_operand) << tc.mnemonic << " on " << tc.arch_name;
+}
+
+INSTANTIATE_TEST_SUITE_P(
```
kuhar, line 301: We should add at least one decode-level case for the PC-only instructions this PR newly rejects, rather than only `s_getpc`/`s_rfe`. On CDNA4, `s_cbranch_join`, `s_cbranch_g_fork`, and `s_rfe_restore_b64` now get `PC_OPERAND` in their generated constructors without another control-flow flag, and their encodings are already in `test_encodings.h` (`0xBE802E00`, `0x94800000`, `0x95800000`).

Adding these to `PcOperandDecodeTest` would lock in the actual denylist-removal behavior: a dropped `OPR_PC` flag would otherwise re-admit those anchors without this test noticing.

kuhar (commented):
A few comments

accauble:
Redoing how operands look on the python side in PR #8426. Closing this and will resubmit when the time is right.

PR #7859, opened 2026-06-26
[rocprofiler-compute] Drop scipy test dependency, use pandas z-score

vedithal-amd:
## Motivation

The only scipy usage is a single z-score call in the test suite. Computing it directly with pandas removes scipy from the test requirements and supports the full Python 3.9-3.13 range without pinning a scipy version (scipy 1.11.4 has no cp313 wheel; 1.14.1 drops 3.9).

## Technical Details

- tests/test_profile_general.py: replace `scipy.stats.zscore` with `(s - s.mean()) / s.std(ddof=0)` (ddof=0 matches scipy's population std); drop unused numpy/scipy imports.
- requirements-test.txt: remove scipy.

## JIRA ID

ROCM-26919

## Test Plan

- Run the profiling test suite on Python 3.9 through 3.13; confirm the stochastic-counter similarity check is unchanged.

## Test Result

- Verified the pandas z-score matches scipy.stats.zscore numerically, including the zero-variance NaN case.

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-compute/requirements-test.txt
+++ b/projects/rocprofiler-compute/requirements-test.txt
@@ -1,4 +1,3 @@
 pytest==8.0.2
 pytest-cov==5.0.0
 pytest-xdist==3.5.0
-scipy==1.11.4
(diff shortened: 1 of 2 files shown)


feizheng10 (approved):

cfallows-amd (approved):
Thanks for changing over to pandas option. Because this is for our unit test the data set should not be large enough that we have a time impact in not using scipy.

PR #7861, opened 2026-06-26
[rocSHMEM] Fix misaligned 64-bit atomics in work/sync pool for odd PE counts

avinashkethineedi:
## Motivation

Collectives crashed with HSA_STATUS_ERROR_MEMORY_APERTURE_VIOLATION on odd PE counts (3, 5, 7, ...) in the IPC and GDA backends.

## Technical Details

The fence region (sizeof(int) * num_pes) is 4-mod-8 for odd num_pes, leaving barrier_sync and the pSync pools 4-byte misaligned. The 64-bit atomics used on those regions then faulted.

- Carve the fence region last so its non-8-byte size can't misalign anything.
- Add __builtin_align_up guards (with reserved padding) before the atomic regions as a backstop.
- Add odd-rank (3, 5) collective tests to driver.sh and CTestFunctionalTests.cmake.

## JIRA ID
- AIROCSHMEM-441

## Test Plan

- coll functional suite at 2, 3, and 5 ranks.

## Test Result

- All rank counts pass; previously 3/5/7 faulted.
- coll suite reports TESTS PASSED.

## Submission Checklist


projects/rocshmem/src/gda/backend_gda.cpp
```diff
@@ -672,10 +675,14 @@ void GDABackend::setup_wrk_sync_buffer() {
...
   /**
-   * Size of fence array
+   * Size of fence array: the only region not sized as a multiple of 8
+   * (4*num_pes), so it is carved last (see init()).
    */
   wrk_sync_pool_size_ += sizeof(int) * num_pes; //TODO: do we need a fence array?
 
+  /* Worst-case padding for the alignment guards in the carve functions. */
+  wrk_sync_pool_size_ += 2 * (alignof(long) - 1);
+
```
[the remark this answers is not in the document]
avinashkethineedi: commit: https://github.com/ROCm/rocm-systems/pull/7861/changes/2de7b10e3637e4cfb01b9dd80b553d4b7b6d4645

edgargabriel (changes requested):
you also need to add the new tests to the `tests/test_categories.yaml` in addition to the `driver.sh` and the `CTestFunctionalTests.cmake`, otherwise theRock CI will not be able to pick the new tests up. (I know its complicated and we need to find a different way to simplify things)

Otherwise LGTM

projects/rocshmem/scripts/functional_tests/driver.sh
```diff
@@ -702,11 +710,20 @@ TestColl() {
...
 
   ExecTest  "teambroadcast"    2       1            64        32768
+  ExecTest  "teambroadcast"    3       1            64        32768
+  ExecTest  "teambroadcast"    5       1            64        32768
 
   ExecTest  "fcollect"         2       1            64        32768
+  ExecTest  "fcollect"         3       1            64        32768
+  ExecTest  "fcollect"         5       1            64        32768
 
+  # NOTE: teamreduction at rank counts > 2 currently fails a data validation
```
abouteiller, line 736: Do we have an AIROCSHMEM number to add to that comment?

abouteiller (approved):

projects/rocshmem/src/ipc/backend_ipc.cpp
```diff
@@ -408,9 +408,12 @@ void IPCBackend::setup_wrk_sync_buffers() {
 
   /**
    * Size of fence array
-  */
+   */
   wrk_sync_pool_size_ += sizeof(int) * num_pes;
 
+  /* Worst-case padding for the alignment guards in the carve functions. */
+  wrk_sync_pool_size_ += 2 * (alignof(int64_t) - 1);
```
omor1: Add `IPCBackend::wrk_sync_pool_alignment` in `backend_ipc.hpp`:
```C++
class IPCBackend : public Backend {
  /* ... */
private:
  static constexpr wrk_sync_pool_alignment = alignof(max_align_t);
};
```
`max_align_t` is probably a 64-bit type, the same as `size_t` or `ptrdiff_t`, but this is "technically correct". We can use `(u)int64_t` instead if that makes more sense to others; I mostly want the alignment to be a constant we defined once and use multiple times instead of writing `alignof(int64_t)` multiple times.

Then rather than manually computing the worst-case padding, just use Clang's `__builtin_align_up`; it can handle integers (sizes) as well as pointers:
```suggestion
  wrk_sync_pool_size_ = __builtin_align_up(wrk_sync_pool_size_, wrk_sync_pool_alignment);
```

projects/rocshmem/src/ipc/backend_ipc.cpp
```diff
@@ -492,6 +497,8 @@ void IPCBackend::rocshmem_collective_init() {
   size_t one_sync_size_bytes {sizeof(*barrier_sync)};
   size_t sync_size_bytes {one_sync_size_bytes * ROCSHMEM_BARRIER_SYNC_SIZE};
 
+  /* Guard: barrier_sync is accessed with 64-bit atomics; keep it 8-byte aligned. */
+  wrk_sync_pool_top_ = __builtin_align_up(wrk_sync_pool_top_, alignof(int64_t));
```
omor1:
```suggestion
  wrk_sync_pool_top_ = __builtin_align_up(wrk_sync_pool_top_, wrk_sync_pool_alignment);
```

projects/rocshmem/src/ipc/backend_ipc.cpp
```diff
@@ -519,6 +526,8 @@ void IPCBackend::teams_init() {
    */
   auto max_num_teams{team_tracker.get_max_num_teams()};
 
+  /* Guard: the pSync pools are accessed with 64-bit atomics; keep them 8-byte aligned. */
+  wrk_sync_pool_top_ = __builtin_align_up(wrk_sync_pool_top_, alignof(int64_t));
```
omor1:
```suggestion
  wrk_sync_pool_top_ = __builtin_align_up(wrk_sync_pool_top_, wrk_sync_pool_alignment);
```

projects/rocshmem/src/ipc/backend_ipc.cpp
```diff
@@ -408,9 +408,12 @@ void IPCBackend::setup_wrk_sync_buffers() {
 
   /**
    * Size of fence array
-  */
+   */
```
omor1, line 429: Noticed the comment alignment is off on lines 395 and 405 as well as here, fix them at the same time?

More broadly we should (at some point) decide on a code style and use clang-format to fix a whole bunch of inconsistencies.

projects/rocshmem/src/ipc/backend_ipc.cpp
```diff
@@ -116,10 +116,10 @@ void IPCBackend::init() {
 
   rocshmem_collective_init();
 
-  setup_fence_buffer();
-
   teams_init();
 
+  setup_fence_buffer();
```
omor1, line 126: Add or move comment from inside the definition of `setup_fence_buffer` to the caller here.

projects/rocshmem/src/ipc/backend_ipc.cpp
omor1: I would at minimum like to see the following, even without `wrk_sync_pool_alignment`:
```suggestion
  wrk_sync_pool_size_ = __builtin_align_up(wrk_sync_pool_size_, alignof(int64_t));
```

omor1 (approved):
LGTM, just some minor comments and recommendations.

I would like to at least see the `wrk_sync_pool_size_` change though, manually computing the "worst-case padding" feels error-prone to me.

projects/rocshmem/scripts/functional_tests/driver.sh
avinashkethineedi: There isn't one at the movement

projects/rocshmem/src/ipc/backend_ipc.cpp
avinashkethineedi: commit: https://github.com/ROCm/rocm-systems/pull/7861/changes/9f5a07a881196428d3a4d263cdeb919933e39f59

projects/rocshmem/src/ipc/backend_ipc.cpp
avinashkethineedi: commit: https://github.com/ROCm/rocm-systems/pull/7861/changes/9f5a07a881196428d3a4d263cdeb919933e39f59

projects/rocshmem/src/ipc/backend_ipc.cpp
avinashkethineedi: commit: https://github.com/ROCm/rocm-systems/pull/7861/changes/9f5a07a881196428d3a4d263cdeb919933e39f59

projects/rocshmem/src/ipc/backend_ipc.cpp
avinashkethineedi: commit: https://github.com/ROCm/rocm-systems/pull/7861/changes/9f5a07a881196428d3a4d263cdeb919933e39f59

projects/rocshmem/src/ipc/backend_ipc.cpp
avinashkethineedi: commit: https://github.com/ROCm/rocm-systems/pull/7861/changes/9f5a07a881196428d3a4d263cdeb919933e39f59

edgargabriel (approved):

PR #7869, opened 2026-06-26
rocr: Add gfx125x kernel-entry trampolines (disabled by default) with env var to enable

dayatsin-amd:
## Summary
Add kernel-entry trampolines for gfx125x - Trampolines redirect each kernel's dispatch entry through a small stub to allow instrumentation to run before the real kernel.

Trampolines are disabled by default for gfx125x. For testing purposes, set LOADER_ENABLE_TRAMPOLINE=1 to enable trampolines.

## Test plan
  - Load a gfx125x code object without the env var set and verify trampolines are disabled by default (trampoline_enabled_gfx125x_ = false, no trampoline fixups applied).
  - Set LOADER_ENABLE_TRAMPOLINE=1 and verify trampolines are enabled for gfx125x code objects (trampoline_enabled_gfx125x_ = true, trampoline fixups applied).
  - Load a non-gfx125x code object (e.g., gfx942) and verify trampolines remain disabled regardless of env var (ISA check gates it).
  - Confirm other loader behavior is unchanged when the env var is unset.


projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
```diff
@@ -1318,7 +1326,8 @@ hsa_status_t ExecutableImpl::LoadCodeObject(
 
   // Kernel-entry trampolines (gfx125x). Gate on this code object's ISA and reset
   // the per-object fixup list collected by LoadDefinitionSymbol.
-  trampoline_enabled_gfx125x_ = CodeObjectIsaIsGfx125Family(codeIsa);
+  trampoline_enabled_gfx125x_ =
+      !loaderOptions.TrampolineDisabled() && CodeObjectIsaIsGfx125Family(codeIsa);
   kd_fixups_.clear();
```
hkasivis, line 1334: May be not an issue but I see this was also added as part of Trampoline code. Should this be also guarded

kentrussell (approved):

hkasivis (approved):

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
```diff
@@ -118,6 +118,8 @@ class LoaderOptions {
   const amd::options::ValueOption<std::string>* DumpDir() const { return &dump_dir; }
   const amd::options::PrefixOption* Substitute() const { return &substitute; }
 
+  bool TrampolineEnabled() const { return trampoline_enabled_; }
```
cfreeamd, line 121: Consider making trampoline_enabled_ a local static so it's only evaluated once; e.g.,

>   static const bool kTrampolineEnvEnabled = [] {
>     const char* e = getenv("LOADER_ENABLE_TRAMPOLINE");
>     return e && std::strcmp(e, "0") != 0;
>   }();

cfreeamd (approved):
Minor things, but no blockers.

Consider accepting only "1", or document the "any non-zero string" contract in the comment. Otherwise things like "false" will be considered true.

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
shwetagkhatri: Done

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
shwetagkhatri: The code is already guarded.  The current code requires both the env var (LOADER_ENABLE_TRAMPOLINE=1) and gfx125x ISA to enable trampolines.

shwetagkhatri (changes requested):
Not needed anymore. Original code has been reverted.

shwetagkhatri (approved):

PR #7871, opened 2026-06-26
finx(libhsakmt): ROCM-26866 Fix integer overflow in event_limit * 8

amd-alysaliu:
- Change event_limit from unsigned int to uint64_t
- Add overflow guard before mmap to prevent undersized allocation

JIRA ID : [ROCM-26866]

[ROCM-26866]: https://amd-hub.atlassian.net/browse/ROCM-26866?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/projects/rocr-runtime/libhsakmt/src/events.c
+++ b/projects/rocr-runtime/libhsakmt/src/events.c
@@ -25,6 +25,7 @@
 
 #include "libhsakmt.h"
 #include <stdlib.h>
+#include <stdint.h>
 #include <string.h>
 #include <errno.h>
 #include <unistd.h>
@@ -71,7 +72,7 @@ HSAKMT_STATUS HSAKMTAPI hsaKmtCreateEventCtx(HsaKFDContext *ctx,
 						 bool ManualReset, bool IsSignaled,
 						 HsaEvent **Event)
 {
-	unsigned int event_limit = KFD_SIGNAL_EVENT_LIMIT;
+	uint64_t event_limit = KFD_SIGNAL_EVENT_LIMIT;
 
 	CHECK_KFD_OPEN();
 
@@ -101,7 +102,7 @@ HSAKMT_STATUS HSAKMTAPI hsaKmtCreateEventCtx(HsaKFDContext *ctx,
 
 	if (hsakmt_is_dgpu && !events_page) {
 		events_page = hsakmt_allocate_exec_aligned_memory_gpu(ctx,
-			KFD_SIGNAL_EVENT_LIMIT * 8, PAGE_SIZE, 0, 0, true, false, true);
+			KFD_SIGNAL_EVENT_LIMIT * sizeof(*events_page), PAGE_SIZE, 0, 0, true, false, true);
 		if (!events_page) {
 			free(e);
 			pthread_mutex_unlock(&hsakmt_mutex);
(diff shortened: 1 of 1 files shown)


projects/rocr-runtime/libhsakmt/src/events.c
```diff
@@ -122,6 +123,11 @@ HSAKMT_STATUS HSAKMTAPI hsaKmtCreateEventCtx(HsaKFDContext *ctx,
 	e->EventId = args.event_id;
 
 	if (!events_page && args.event_page_offset > 0) {
+		if (event_limit > SIZE_MAX / sizeof(*events_page)) {
+				free(e);
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** The new guard block is indented at 4 tabs but the surrounding code uses 3. This makes the `free`/`unlock`/`return` body look like it is nested inside an additional level of control flow that does not exist. Should be 3 tabs to match the enclosing `if` body style throughout this function.

projects/rocr-runtime/libhsakmt/src/events.c
```diff
@@ -122,6 +123,11 @@ HSAKMT_STATUS HSAKMTAPI hsaKmtCreateEventCtx(HsaKFDContext *ctx,
...
+				pthread_mutex_unlock(&hsakmt_mutex);
+				return HSAKMT_STATUS_NO_MEMORY;
+		}
 		events_page = mmap(NULL, event_limit * 8, PROT_WRITE | PROT_READ,
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** The overflow guard above correctly uses `sizeof(*events_page)`, but this `mmap` call still uses the literal `8`. If the element type of `events_page` ever changes, the guard and the allocation will silently diverge. Use `sizeof(*events_page)` here too:
```c
events_page = mmap(NULL, event_limit * sizeof(*events_page), PROT_WRITE | PROT_READ,
    MAP_SHARED, ctx->fd, args.event_page_offset);
```
Also note: the dGPU path (~line 99) has the same class of problem — `KFD_SIGNAL_EVENT_LIMIT * 8` with no overflow guard and a bare signed-int multiply. Worth fixing at the same time.

cfreeamd (commented):
🤖 AI-assisted pre-review (for @cfreeamd) — inline comments above. Note: the dGPU path (~line 99, `KFD_SIGNAL_EVENT_LIMIT * 8`) has the same unguarded signed-int multiply and is not covered by this diff.

projects/rocr-runtime/libhsakmt/src/events.c
kentrussell: Yeah, it might seem silly, but can we get rid of the extra tab for the 3 lines?

cfreeamd (commented):

The rebased commit addresses all three prior findings: the `mmap` call now uses `event_limit * sizeof(*events_page)` (consistent with the overflow guard), the dGPU path uses `KFD_SIGNAL_EVENT_LIMIT * sizeof(*events_page)` instead of the bare `* 8` literal, and the guard block indentation is corrected to match the surrounding `mmap` line (also resolving the extra-tab comment on-thread). The fix looks complete.

cfreeamd (approved):

PR #7873, opened 2026-06-26
[rocprofiler-compute] Stream counter CSVs instead of loading into memory

xuchen-amd:
## Motivation

`update_rocpd_pmc_events()` and `convert_native_counter_collection_csv()` previously loaded full CSV contents into `list[dict]` via `csv_ops.read_csv_as_dicts` before processing. For large counter collection CSVs (common with many GPU dispatches), this creates unnecessary memory pressure. Streaming rows through `csv.DictReader` keeps memory usage bounded regardless of file size.

### `rocpd_data.py`
- `update_rocpd_pmc_events` now takes a CSV file path instead of a pre-loaded list of rows
- Rows are read via `csv.DictReader` and inserted into SQLite in batches within a single transaction
- Extracted two helpers: `_resolve_pmc_event_metadata` (table/GUID lookup) and `_stream_csv_to_pmc_event_table` (batched insert)

### `utils_profile.py`
- Replaced `csv_ops.groupby_aggregate` and `csv_ops.merge_rows` with streaming equivalents
- Extracted `_stream_aggregate_counter_csv` (single-pass group-by with summing) and `_build_rocprofv3_counter_rows` (dict-based inner 
join)
- Updated `run_prof` call site to pass the CSV path directly

## JIRA ID

N/A

## Test Plan

Memory usage comparison with example workload.

## Test Result

Tested with a single-pass collection producing a 4.3 GB native counter CSV (~72M rows).

| | Before | After |
|---|---|---|
| **Peak memory** | 49 GB | 427 MB |
| **Processing time** | ~404s | ~424s |
| **Output rows** | 2,507,249 | 2,507,249 |

Memory reduced **116x** with equivalent processing time and identical output. The previous implementation would OOM on machines with less than ~50 GB RAM; the streaming version completes on any machine.

## Submission Checklist


projects/rocprofiler-compute/src/utils/rocpd_data.py
```diff
@@ -100,69 +101,98 @@ def convert_dbs_to_csv(
...
+def _stream_csv_to_pmc_event_table(
+    conn: sqlite3.Connection,
+    counter_csv_path: str,
+    insert_sql: str,
+    guid: str,
+    dispatch_to_event: dict[str, str],
+    batch_size: int,
+) -> None:
+    """Read counter CSV in batches and insert into the pmc_event table."""
+    batch: list[tuple[Optional[str], ...]] = []
```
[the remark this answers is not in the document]
xuchen-amd: drop

projects/rocprofiler-compute/src/utils/utils_profile.py
```diff
@@ -615,85 +615,105 @@ def v3_counter_csv_to_v2_csv(
...
             kernel_data_filename.replace("kernel_trace", "counter_collection"),
             rocprofv3_counter_data,
         )
 
 
+def _stream_aggregate_counter_csv(
+    csv_path: Path,
+    groupby_columns: tuple[str, ...],
+) -> dict[tuple[str, ...], dict]:
+    """Stream CSV and group rows by key columns, summing counter_value."""
```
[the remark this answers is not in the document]
xuchen-amd: drop

projects/rocprofiler-compute/src/utils/utils_profile.py
```diff
@@ -615,85 +615,105 @@ def v3_counter_csv_to_v2_csv(
...
+    groups: dict[tuple[str, ...], dict] = {}
+    with open(csv_path, newline="", encoding="utf-8") as f:
+        for row in csv.DictReader(f):
```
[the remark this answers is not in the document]
xuchen-amd: fixed: 2f313a163cea382318ce33504f0397b5d3c9ebb9

projects/rocprofiler-compute/src/utils/rocpd_data.py
```diff
@@ -100,69 +101,98 @@ def convert_dbs_to_csv(
...
+    with conn:
+        with open(counter_csv_path, newline="", encoding="utf-8") as f:
+            for row in csv.DictReader(f):
```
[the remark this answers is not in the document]
xuchen-amd: fixed: 2f313a163cea382318ce33504f0397b5d3c9ebb9

projects/rocprofiler-compute/src/utils/rocpd_data.py
```diff
@@ -100,69 +101,98 @@ def convert_dbs_to_csv(
...
+            _stream_csv_to_pmc_event_table(
+                conn,
+                counter_csv_path,
+                insert_sql,
+                guid,
+                dispatch_to_event,
+                batch_size,
+            )
     except OSError as e:
         console_error(f"Database error while updating pmc_event table: {e}")
```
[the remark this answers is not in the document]
xuchen-amd: fixed: 2f313a163cea382318ce33504f0397b5d3c9ebb9

projects/rocprofiler-compute/src/utils/utils_profile.py
```diff
@@ -615,85 +615,108 @@ def v3_counter_csv_to_v2_csv(
...
+                try:
+                    row["counter_value"] = float(row.get("counter_value", 0))
+                except (ValueError, TypeError):
+                    row["counter_value"] = 0.0
+                groups[key] = row
+                continue
+            try:
+                groups[key]["counter_value"] += float(row.get("counter_value", 0))
+            except (ValueError, TypeError):
+                pass
```
jamessiddeley-amd: Something to note here in a logic change that is different than the old `csv_ops.groupby_aggregate(...)` implementation, is that the old path built a list of all values first, then sums them in one shot with `sum(float(v) for v in values)` so if any single value in the group can't be turned into a number the whole thing throws. But this new path handles the first row of a group and later rows separately.

So if we have some mismanaged or malformed input, the returned value would be different since this new implementation skips on valueerrors and type errors while the old path completely throws. Low risk though as I imagine its rare since `counter_value` is always numeric but worth either noting or fixing.

projects/rocprofiler-compute/src/utils/rocpd_data.py
```diff
@@ -100,69 +101,101 @@ def convert_dbs_to_csv(
...
+    dispatch_to_event: dict[str, str],
+    batch_size: int,
+) -> None:
+    """Read counter CSV in batches and insert into the pmc_event table."""
+    batch: list[tuple[Optional[str], ...]] = []
+    with conn:
+        with open(counter_csv_path, newline="", encoding="utf-8") as f:
+            reader = csv.DictReader(f)
+            if reader.fieldnames is None:
+                return
```
jamessiddeley-amd: Before, a broken CSV stopped the run with a clear "no header row" error
```
            fieldnames = reader.fieldnames
            if fieldnames is None:
                raise ValueError(f"CSV file {csv_file} has no header row")
```
While now it returns instead of throwing. If we follow the return the caller logs success with
```
rocpd_data.update_rocpd_pmc_events(str(counter_csv), str(db_name))
console_debug(f"Updated rocpd db {db_name} with native tool counters.")
```
So if the csv was broken the function doesnt throw and the log still says that debug message _"Updated rocpd db ... with native tool counters."_ which would be a false success message. 
I recommend we call `console_error(...)` (or raising) when `fieldnames` is `None` like how it was before

projects/rocprofiler-compute/src/utils/rocpd_data.py
```diff
@@ -100,69 +101,101 @@ def convert_dbs_to_csv(
                         )
 
 
-def update_rocpd_pmc_events(counter_info: list[dict], rocpd_db_path: str) -> None:
-    """Updates pmc_event table in the given rocpd database path."""
+def update_rocpd_pmc_events(
+    counter_csv_path: str,
+    rocpd_db_path: str,
+    batch_size: int = 50_000,
```
jamessiddeley-amd: Magic number here with no validation (like technically you could pass `batch_size <= 0` and that would probably crash), although I guess no callers on it that actually do that, so sort of a nitpick.

Why 50,000 as the number you landed on, any specific reason? Just wondering.

jamessiddeley-amd (commented):
Couple nitpicks and some comments. Looks good over first pass thanks Xuan!

vedithal-amd (changes requested):
We should update unit/integration (as required) tests when we update source code

projects/rocprofiler-compute/src/utils/rocpd_data.py
```diff
@@ -100,69 +101,101 @@ def convert_dbs_to_csv(
...
-        console_error(f"Database error while updating pmc_event table: {e}")
+        console_error(f"Error while updating pmc_event table: {e}")
     except Exception as e:
         console_error(f"Unexpected error updating pmc_event table: {e}")
+
+
+def _resolve_pmc_event_metadata(
+    conn: sqlite3.Connection,
+) -> Optional[tuple[str, str, dict[str, str]]]:
+    """Return (table_name, guid, dispatch_to_event) or None on failure."""
```
vedithal-amd: Can you elaborate more on what this function is doing

projects/rocprofiler-compute/src/utils/rocpd_data.py
```diff
@@ -100,69 +101,101 @@ def convert_dbs_to_csv(
...
+
+    guid = table_name[len(ROCPD_PMC_EVENT_TABLE_NAME_PREFIX) :].replace("_", "-")
+
+    # event_id may differ from dispatch_id when marker API tracing is enabled
+    with closing(conn.execute(KERNEL_DISPATCH_QUERY, (guid,))) as cursor:
+        db_rows = cursor.fetchall()
+
+    if not db_rows:
+        console_error("No kernel dispatch data found.")
+        return None
```
vedithal-amd, line 165: return will never hit after console_error since it immidiately exists unless exit=False is given

projects/rocprofiler-compute/src/utils/rocpd_data.py
```diff
@@ -100,69 +101,101 @@ def convert_dbs_to_csv(
...
     except OSError as e:
-        console_error(f"Database error while updating pmc_event table: {e}")
+        console_error(f"Error while updating pmc_event table: {e}")
     except Exception as e:
         console_error(f"Unexpected error updating pmc_event table: {e}")
+
+
+def _resolve_pmc_event_metadata(
+    conn: sqlite3.Connection,
+) -> Optional[tuple[str, str, dict[str, str]]]:
```
vedithal-amd, line 138: See my comment below, i dont think it can ever actually return None without terminating the execution

vedithal-amd: But this is OK

projects/rocprofiler-compute/src/utils/rocpd_data.py
```diff
@@ -100,69 +101,101 @@ def convert_dbs_to_csv(
...
-                conn.executemany(
-                    INSERT_QUERY.format(
-                        table_name=table_name,
-                        columns=", ".join(columns),
-                        placeholders=placeholders,
-                    ),
-                    values,
-                )
+            placeholders = ", ".join(["?"] * len(columns))
+            insert_sql = INSERT_QUERY.format(
```
vedithal-amd: insert_sql can be folded into streaming helper

projects/rocprofiler-compute/src/utils/utils_profile.py
```diff
@@ -615,85 +615,108 @@ def v3_counter_csv_to_v2_csv(
...
-            rocprofv3_counter_data.append(new_row)
+        rocprofv3_counter_data = _build_rocprofv3_counter_rows(groups, kernel_lookup)
 
         csv_ops.write_csv_from_dicts(
             kernel_data_filename.replace("kernel_trace", "counter_collection"),
             rocprofv3_counter_data,
         )
 
 
+def _stream_aggregate_counter_csv(
```
vedithal-amd: The name is a bit misleading here - this is not actually batch streaming anything - its aggregating the csvs

projects/rocprofiler-compute/src/utils/utils_profile.py
```diff
@@ -615,85 +615,108 @@ def v3_counter_csv_to_v2_csv(
...
-                col: "first"
-                for col in counter_data[0].keys()
-                if col not in groupby_cols
-            }
-        else:
-            agg_dict = {}
-        # Overwrite counter_value aggregation to sum
-        agg_dict["counter_value"] = "sum"
-        counter_data = csv_ops.groupby_aggregate(counter_data, groupby_cols, agg_dict)
+        groups = _stream_aggregate_counter_csv(native_path, groupby_columns)
```
vedithal-amd: groups is a vague variable name

projects/rocprofiler-compute/src/utils/rocpd_data.py
```diff
@@ -100,69 +101,101 @@ def convert_dbs_to_csv(
...
+    insert_sql: str,
+    guid: str,
+    dispatch_to_event: dict[str, str],
+    batch_size: int,
+) -> None:
+    """Read counter CSV in batches and insert into the pmc_event table."""
+    batch: list[tuple[Optional[str], ...]] = []
+    with conn:
+        with open(counter_csv_path, newline="", encoding="utf-8") as f:
+            reader = csv.DictReader(f)
```
vedithal-amd: This loop and `_stream_aggregate_counter_csv` (utils_profile.py:655) both hand-roll the same `DictReader` + header guard. Extract one helper in utils_profile_csv.py:

```python
def iter_csv_rows(csv_file: str) -> Iterator[dict]:
    with open(csv_file, newline="", encoding="utf-8") as f:
        reader = csv.DictReader(f)
        if reader.fieldnames is None:
            raise ValueError(f"CSV file {csv_file} has no header row")
        yield from reader
```

Both sites reuse it. Here, wrap it in the 50k batch loop. In `_stream_aggregate_counter_csv`, iterate directly. Validation lives in one place, no duplicated boilerplate. Batching stays here since it's a SQLite concern.

projects/rocprofiler-compute/src/utils/utils_profile.py
```diff
@@ -615,85 +615,108 @@ def v3_counter_csv_to_v2_csv(
...
 
-        # Merge counter_data with kernel_data on dispatch_id
-        merged_data = csv_ops.merge_rows(
-            counter_data,
-            kernel_data,
-            left_on="dispatch_id",
-            right_on="Dispatch_Id",
-            how="inner",
-        )
+        kernel_lookup: dict[str, list[dict]] = {}
```
vedithal-amd: `kernel_lookup` + `_build_rocprofv3_counter_rows` reimplements `merge_rows` inline. Same O(n+m) join, not faster. The only reason not to call `merge_rows` is that it returns a list, forcing an intermediate copy. Fix that instead of bypassing it: make it lazy.

```python
def iter_merge_rows(left_rows, right_rows, left_on, right_on, how="inner"):
    if how not in ("inner", "left", "right", "outer"):
        raise ValueError(f"Unsupported join type: {how}")
    # build right_lookup as today, then yield instead of append
    for left_row in left_rows:
        ...
        yield {**left_row, **right_row}

def merge_rows(...):
    return list(iter_merge_rows(...))
```

Then this block drops `kernel_lookup` entirely:

```python
rows = []
for merged in csv_ops.iter_merge_rows(
    groups.values(), kernel_data,
    left_on="dispatch_id", right_on="Dispatch_Id", how="inner",
):
    rows.append(_project_rocprofv3_row(merged))
```

This reuses the tested join and keeps the no-intermediate-list memory win. The existing `merge_rows` caller and its tests keep working through the wrapper unchanged. Keep the `how` check at the top of `iter_merge_rows`: `merge_rows` wraps it in `list(...)`, which consumes the generator immediately, so the invalid-how case still raises eagerly.

projects/rocprofiler-compute/src/utils/rocpd_data.py
xuchen-amd: fixed in: 6172dfcd9d

projects/rocprofiler-compute/src/utils/rocpd_data.py
xuchen-amd: fixed in: 0f23cf5148

projects/rocprofiler-compute/src/utils/rocpd_data.py
xuchen-amd: fixed in: 0f23cf5148

projects/rocprofiler-compute/src/utils/rocpd_data.py
xuchen-amd: improved docstring in: 0f23cf5148

projects/rocprofiler-compute/src/utils/rocpd_data.py
xuchen-amd: avoiding magic number with comments to add context with constant var: 865b19a8d14a9380c1c259197f40fa115167d836, basically trying to avoid row-by-row (size = 1) processing or entire file (no streaming).

projects/rocprofiler-compute/src/utils/rocpd_data.py
xuchen-amd: fixed in: e31ba72ebd

projects/rocprofiler-compute/src/utils/rocpd_data.py
xuchen-amd: fixed in: e31ba72ebd

projects/rocprofiler-compute/src/utils/utils_profile.py
xuchen-amd: fixed in: 87387b8c89

projects/rocprofiler-compute/src/utils/utils_profile.py
xuchen-amd: fixed in: 87387b8c89

projects/rocprofiler-compute/src/utils/utils_profile.py
xuchen-amd: `counter_value` in native counter CSVs is always a C++ double (written by `CsvCountersWriter` in `counters_writer.cpp`), so non-numeric values can't occur in practice. The try/except is just a defensive guard, if the value is somehow corrupted, the CSV itself would be broken thus the skip.

projects/rocprofiler-compute/src/utils/utils_profile.py
xuchen-amd: added the refactor in: 84dc2b97146ce53528e2f7695b43099383983281 and 819a654763e1175fe628e6a8e2f12789f007450b

jamessiddeley-amd (approved):
LGTM pending other reviewers' comments and CI!

abchoudh-amd (approved):

vedithal-amd (changes requested):
Blocking for final review

vedithal-amd:
Pushed a follow-up (e5db7fbe633) collapsing the streaming work into the existing csv_ops rather than parallel helpers:

- Folded `iter_merge_rows` into a single lazy `merge_rows` (dropped the list-wrapping shim). Callers that re-iterate now wrap in `list()`.
- Dropped `_aggregate_counter_csv`; the call site uses `groupby_aggregate(iter_csv_dicts(...), ...)`. `groupby_aggregate` is back to its original body with only the input widened to `Iterable`, since the whole-file load (fixed by `iter_csv_dicts`) was the OOM cause, not per-group buffering.
- Removed `test_native_counter_streaming.py`: its csv_ops cases belong in the 1:1 `test_utils_profile_csv.py` (added there), and the rest is covered by integration tests.

Rationale: keep one aggregation/merge path in csv_ops instead of duplicating logic. Same memory win, smaller surface.

vedithal-amd (approved):
LGTM

cfallows-amd (approved):
nitpick- the dict type annotations added in this PR lack the inner key value types.
Otherwise this is great memory savings. Thanks

PR #7881, opened 2026-06-26
[RCCL] Add opt-in LL128 protocol for send/recv kernels

PedramAlizadeh:
Added opt-in LL128 protocol for send/recv kernels (gfx1250) gated behind NCCL_P2P_LL128_ENABLE (default off) and limited to gfx1250, with NCCL_P2P_LL128_THRESHOLD (default 0 = no upper bound) to bound the per-channel payload size.

## Motivation

The send/recv kernels currently only support the LL and SIMPLE protocols, and this effort investigates whether the more bandwidth-efficient LL128 could help on MI450X (gfx1250) even for large messages; this PR adds an opt-in, gfx1250-only LL128 path for send/recv so we can measure that impact while leaving default behavior on all existing architectures unchanged.

## Technical Details

LL128 is added as a third protocol in the per-direction/per-connection selection in addP2pToPlan() (precedence LL128 → LL → SIMPLE) and dispatched in the send/recv device kernel via new sendProtoLL128/recvProtoLL128 work flags; no primitives changes are needed since the Primitives wrapper already maps directSend/directRecv to LL128's send/recv. It is gated by NCCL_P2P_LL128_ENABLE (default 0), bounded by NCCL_P2P_LL128_THRESHOLD (default 0 = no upper bound), and requires arch gfx1250 plus an allocated LL128 connection buffer; when selected, the LL128 chunk math chunkDataSize = (chunkSize / ll128LineElems) * ll128DataElems is applied, matching the collective path and the device-side ProtoLL128::calcBytePerStep().

## JIRA ID

AICOMRCCL-1402

## Test Plan

On gfx1250, run the 2-rank single-node sendrecv perf benchmark with NCCL_P2P_LL128_ENABLE=1 NCCL_P2P_LL128_THRESHOLD=0 and sweep message sizes, confirming correctness and bandwidth vs the default LL/SIMPLE path.

## Submission Checklist

### DIFF
--- a/projects/rccl/CMakeLists.txt
+++ b/projects/rccl/CMakeLists.txt
@@ -90,7 +90,8 @@ option(ENABLE_CODE_COVERAGE                    "Enable code coverage"           
 option(ENABLE_IFC                              "Enable indirect function call"                 OFF)
 option(ENABLE_DEVICE_LINKER                    "Build with assembly-extract device linker"     ON)
 option(EMIT_LLVM_IR                            "Emit LLVM IR/bitcode for nccl_device APIs"     OFF)
-option(GENERATE_SYM_KERNELS                    "Generate symmetric memory kernels"             ON)
+option(DISABLE_KERNARG_PRELOAD                 "Disable amdgpu-kernarg-preload-count compile/link flag" OFF)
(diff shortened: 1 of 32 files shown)


atulkulk:
@PedramAlizadeh Is this PR still relevant?

PR #7887, opened 2026-06-26
ci(therock-ci): pin TheRock checkout to release/therock-* branch for release PRs

araravik-psd:
## Summary

Adds a composite action `.github/actions/ci-env/action.yml` as the
single source of truth for the TheRock git ref used across CI. When a
PR targets a `release/therock-*` branch in rocm-systems, all TheRock
checkouts automatically use the matching branch in the TheRock
repository (e.g. `release/therock-7.14`). PRs targeting `develop` or
any other branch are unaffected and continue using the pinned commit
SHA in the composite action.

## Changes

- **`.github/actions/ci-env/action.yml`** *(new)*: Composite action
  that resolves the TheRock ref — release branch name if on a
  `release/therock-*` branch, otherwise the pinned commit SHA. This is
  the only place the SHA lives, making bump automation updates simpler.

- **`therock-ci.yml`**, **`therock-ci-linux.yml`**,
  **`therock-ci-windows.yml`**, **`therock-rccl-ci-linux.yml`**,
  **`therock-test-packages.yml`**: Each job now calls
  `./.github/actions/ci-env` and uses `steps.ci-env.outputs.therock-ref`
  for the TheRock checkout ref.

## Behavior

| PR base branch | TheRock ref used |
|---|---|
| `develop` | pinned SHA in `action.yml` (no change) |
| `release/therock-7.14` | `release/therock-7.14` |
| any other branch | pinned SHA in `action.yml` (no change) |

## Follow-up required in TheRock

`bump_automation.py` needs to add `.github/actions/ci-env/action.yml`
to `ROCM_SYSTEMS_FILES` and remove the individual workflow files from
it, so the automated SHA bump only updates the composite action — same
pattern as `rocm-libraries`.


.github/workflows/therock-ci-linux.yml
```diff
@@ -50,7 +59,7 @@ jobs:
         with:
           repository: "ROCm/TheRock"
           path: "TheRock"
-          ref: f1e3fc2311fde45bf3e17b0e3349200be9caa4b4 # 2026-06-23
+          ref: ${{ inputs.therock_ref || 'f1e3fc2311fde45bf3e17b0e3349200be9caa4b4' }} # 2026-06-23
```
ScottTodd: These lines are updated by automation that currently does a full overwrite of the entire line. We'd need to either change that to
1. Do a substring replacement (looking for commit hash patterns)
2. Start generating this format (potentially brittle, that could assume that the input is available)

https://github.com/ROCm/TheRock/blob/3b102113a13cc8c33fd9d3e87ac9a0e65d9f38ab/build_tools/github_actions/bump_automation.py#L135-L138

.github/workflows/therock-ci.yml
```diff
@@ -60,12 +61,31 @@ jobs:
...
+          # Use base_ref for PRs; ref_name for push/dispatch events.
+          BASE_BRANCH="${{ github.base_ref || github.ref_name }}"
+          echo "Base branch: ${BASE_BRANCH}"
+          if [[ "${BASE_BRANCH}" =~ ^release/therock-.+$ ]]; then
+            THEROCK_REF="${BASE_BRANCH}"
+            echo "Release branch detected – using same branch in TheRock: ${THEROCK_REF}"
+          else
+            THEROCK_REF=""   # empty → child workflows use their hardcoded default pin
+          fi
+          echo "therock_ref=${THEROCK_REF}" >> "$GITHUB_OUTPUT"
```
ScottTodd: If we add a composite action like https://github.com/ROCm/rocm-libraries/blob/develop/.github/actions/ci-env/action.yml we could put this logic all in one place, shared across other entry points like
* https://github.com/ROCm/rocm-systems/blob/develop/.github/workflows/therock-multi-arch-ci.yml
* https://github.com/ROCm/rocm-systems/blob/develop/.github/workflows/therock-ci-nightly.yml

ScottTodd (commented):
This logic looks okay to me, with a few details that would help to address.

Let's also run this by @geomin12 too.

.github/workflows/therock-ci.yml
araravik-psd: yes this is a good idea, I was just doing this pinning for rocm-libraries, I saw the added composite action, let me make the changes accordingly

araravik-psd: @ScottTodd  please take a look I have updated rocm-systems similar to how we have it in rocm-libraries with a composite action

ScottTodd: Use for _all_ entry points?

https://github.com/search?q=repo%3AROCm%2Frocm-systems+f1e3fc2311fde45bf3e17b0e3349200be9caa4b4&type=code

otherwise lgtm

araravik-psd: @ScottTodd  Updated therock-ci-nightly.yml to use the composite action as well.
therock_multi_arch_ci_asan.yml cannot be updated with this approach, the SHA there appears in uses: ROCm/TheRock/.github/workflows/...@<sha> syntax for calling external reusable workflows, where GitHub requires a literal value and doesn't support expressions. That file will need a separate solution.

ScottTodd:
This has merge conflicts after @geomin12 merged https://github.com/ROCm/rocm-systems/pull/7850

.github/workflows/therock-ci.yml
ScottTodd: Oh... uhhhhh, well... this isn't going to work well: https://github.com/ROCm/rocm-systems/blob/bac4e814c611a14ff7ca128a3d34bdf577c1a81c/.github/workflows/therock-multi-arch-ci.yml#L60

The workflows are *not* stable using a floating ref like `main`. We already coordinate cross-repository changes to workflow files in https://github.com/ROCm/rockrel (which only runs jobs once a day), can't easily roll out changes a repository like this one that runs workflows throughout the day.

So if we pinned a commit ref there we wouldn't be able to use this `actions/ci-env` setup, and we're migrating _all_ workflows to multi-arch, so should we not implement the `actions/ci-env` changes then?

Need your input here @geomin12

araravik-psd: @geomin12 ping, might have missed this

geomin12: ah the `@main` tag is temporary as we haven't fully migrated to multi arch yet. I know that pointing to main will be unstable! 

currently, that job only runs on nightly

.github/actions/ci-env/action.yml
```diff
@@ -0,0 +1,34 @@
...
+      shell: bash
+      run: |
+        # Use base_ref for PRs; ref_name for push/dispatch events.
+        BASE_BRANCH="${{ github.base_ref || github.ref_name }}"
+        if [[ "${BASE_BRANCH}" =~ ^release/therock-.+$ ]]; then
+          # PR targets a release branch — check out the same branch in TheRock.
+          THEROCK_REF="${BASE_BRANCH}"
+        else
+          # Default: use the pinned commit SHA (updated by bump automation).
+          THEROCK_REF="f1e3fc2311fde45bf3e17b0e3349200be9caa4b4" # 2026-06-23
```
geomin12: let's update this to the most recent

.github/actions/ci-env/action.yml
```diff
@@ -0,0 +1,34 @@
...
+        fi
+        echo "Resolved TheRock ref: ${THEROCK_REF}"
+        echo "therock_ref=${THEROCK_REF}" >> "$GITHUB_OUTPUT"
+
```
geomin12: for rocm-systems, we will always use `f1e3fc2311fde45bf3e17b0e3349200be9caa4b4`

no need for THEROCK_REF="${BASE_BRANCH}"

geomin12 (commented):
we should look into #8215, there are a few missing files!

once we propogate to all the files, i can close my PR

PR #7888, opened 2026-06-26
fix(rccl): Fix devrState leak on the non-symmetric window path

Kapil-Shyam-Pawar:
## Motivation

On a `!symmetricSupport comm` (e.g. cuMem disabled), `ncclCommWindowRegister` still inits `devrState` but `commFree` only freed it under `if (symmetricSupport)`, leaking lsaRankList (24 B/comm).
```
Direct leak of 24 byte(s) in 6 object(s) allocated from:
    #1 ... ncclDevrInitOnce(ncclComm*) dev_runtime.cc
```
## Technical Details

`ncclCommWindowRegister_impl` calls `ncclDevrInitOnce` unconditionally, so non-sym comms allocate devrState too. 
Fix: call `ncclDevrFinalize` unconditionally in `commFree` (self-guards on bigSize == 0).

It doesn't happen on NCCL because:
- NCCL's `ncclCommWindowRegister` returns early for non-symmetric comms before `ncclDevrInitOnce`, so it never allocates this state. RCCL can't, because it falls back to a non-symmetric window path (IPC/proxy) that needs devrState.
- NCCL defaults `NCCL_CUMEM_ENABLE` to auto-on (-2) vs RCCL's 0, so NCCL comms are usually symmetric and rarely hit this path.

## JIRA ID

ROCM-27249

## Test Plan

Add a test to exercise the non-symmetric window register/deregister/destroy path on a single-rank comm (cuMem disabled).

## Test Result

Added test `Register.ProcessIsolatedRegisterTests (WindowRegisterSingleRankNonSymTeardown)` - passes. 
Reproducer's `lsaRankList/devrState` leaks are gone

## Submission Checklist

### DIFF
--- a/projects/rccl/src/init.cc
+++ b/projects/rccl/src/init.cc
@@ -455,8 +455,9 @@ static ncclResult_t commFree(ncclComm_t comm) {
 
   if (comm->symmetricSupport) {
     NCCLCHECK(ncclSymkFinalize(comm));
-    NCCLCHECK(ncclDevrFinalize(comm));
   }
+  // RCCL: !symmetricSupport comms still init devrState via the non-sym window-register path (dev_runtime.cc), so finalize unconditionally to free lsaRankList.
+  NCCLCHECK(ncclDevrFinalize(comm));
   NCCLCHECK(ncclRasCommFini(comm));
 
   /* in commReclaim, we have guaranteed only last rank which calls ncclCommDestroy() will
--- a/projects/rccl/test/RegisterTests.cpp
+++ b/projects/rccl/test/RegisterTests.cpp
@@ -212,7 +212,7 @@ static void testWindowRegisterSingleRankNonSymTeardown() {
     ncclComm_t comm;
     NCCLCHECK(initSingleRankComm(&comm));
 
-    const size_t bufferSize = 4096; // NCCL_WIN_REQUIRED_ALIGNMENT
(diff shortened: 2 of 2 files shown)


atulkulk (approved):

projects/rccl/src/init.cc
```diff
@@ -455,8 +455,9 @@ static ncclResult_t commFree(ncclComm_t comm) {
 
   if (comm->symmetricSupport) {
     NCCLCHECK(ncclSymkFinalize(comm));
-    NCCLCHECK(ncclDevrFinalize(comm));
   }
+  // RCCL: !symmetricSupport comms still init devrState via the non-sym window-register path (dev_runtime.cc), so finalize unconditionally to free lsaRankList.
```
PJAvinash, line 459: init devrState via the non-sym window-register, does this happen unconditionally too? 
and is ncclDevrFinalize(comm) idempotent?

Kapil-Shyam-Pawar: 1. Yes `ncclCommWindowRegister_impl` calls `ncclDevrInitOnce`  unconditionally (dev_runtime.cc:1826), so `devrState (lsaRankList)` is allocated on both the sym and non-sym paths
2. Its not but its only called once (from `commFree`, once per comm)

PJAvinash (approved):

Kapil-Shyam-Pawar:
therock-pr-bot flag on unit test is invalid
its unable to detect rccl test files
this pr includes a test

PR #7889, opened 2026-06-26
fix(rocjpeg): Fix integer overflow in SOF parser and CopyChannel

AryanSalmanpour:
## Motivation

JIRA ID : ROCM-26827 - Fix integer overflow in SOF parser and CopyChannel Fix

## Technical Details

rocjpeg_parser.cpp — ParseSOF():
- Reject width/height > 65500 or == 0 before any arithmetic (ISO 10918-1 allows up to 65535, but values that large overflow downstream 32-bit expressions)
- Guard against zero luma sampling factors (divide-by-zero in MCU calculation)
- Compute width * height * num_components in uint64_t and reject if it exceeds SIZE_MAX/4, preventing undersized allocations from crafted JPEG files
- Compute MCU grid dimensions in uint64_t and range-check before narrowing to uint32_t

rocjpeg_decoder.cpp — CopyChannel():
- Compute pitch * height in uint64_t and check for uint32_t overflow before passing the size to the HIP copy API

## JIRA ID

Resolves ROCM-26827

## Test Plan

rocJPEG CTEST

## Submission Checklist

### DIFF
--- a/projects/rocjpeg/src/rocjpeg_decoder.cpp
+++ b/projects/rocjpeg/src/rocjpeg_decoder.cpp
@@ -21,7 +21,6 @@ THE SOFTWARE.
 */
 
 #include "rocjpeg_decoder.h"
-#include <limits>
 
 RocJpegDecoder::RocJpegDecoder(RocJpegBackend backend, int device_id) :
     num_devices_{0}, device_id_ {device_id}, hip_stream_ {0}, backend_{backend} {}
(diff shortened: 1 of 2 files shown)


projects/rocjpeg/src/rocjpeg_decoder.cpp
```diff
@@ -482,7 +483,15 @@ RocJpegStatus RocJpegDecoder::CopyChannel(HipInteropDeviceMem& hip_interop_dev_m
...
-            uint32_t channel_size = destination->pitch[channel_index] * channel_height;
+            // Use 64-bit arithmetic to detect pitch*height overflow before passing
+            // to the HIP copy API, which takes a size_t.
+            uint64_t channel_size_64 = static_cast<uint64_t>(destination->pitch[channel_index]) * static_cast<uint64_t>(channel_height);
+            if (channel_size_64 > static_cast<uint64_t>(std::numeric_limits<uint32_t>::max())) {
+                ErrorLog(g_rocjpeg_logger, "Channel size overflows uint32_t (pitch=" +
+                    ROCJPEG_TOSTR(destination->pitch[channel_index]) + " height=" + ROCJPEG_TOSTR(channel_height) + ")!");
+                return ROCJPEG_STATUS_INVALID_PARAMETER;
+            }
+            uint32_t channel_size = static_cast<uint32_t>(channel_size_64);
```
rrawther: why are we converting this to 32bit? does this mean we will lose precision? If so, why we need to do the math in 64?

projects/rocjpeg/src/rocjpeg_decoder.cpp
```diff
@@ -482,7 +483,15 @@ RocJpegStatus RocJpegDecoder::CopyChannel(HipInteropDeviceMem& hip_interop_dev_m
         }
 
         if (destination->pitch[channel_index] == hip_interop_dev_mem.pitch[channel_index]) {
-            uint32_t channel_size = destination->pitch[channel_index] * channel_height;
+            // Use 64-bit arithmetic to detect pitch*height overflow before passing
+            // to the HIP copy API, which takes a size_t.
+            uint64_t channel_size_64 = static_cast<uint64_t>(destination->pitch[channel_index]) * static_cast<uint64_t>(channel_height);
```
rrawther: I think pitch and height can never exceed 32bit. Only one of those needs to be casted to get the result in 64bit

rrawther (changes requested):
Added some comments

projects/rocjpeg/src/rocjpeg_decoder.cpp
AryanSalmanpour: Done.

projects/rocjpeg/src/rocjpeg_decoder.cpp
AryanSalmanpour: That's correct. Even if one of them is cast, the other will be implicitly cast by the compiler. Keeping the cast for both is harmless and improves readability.

rrawther (approved):

PR #7890, opened 2026-06-26
feat(rocjitsu): Write-after-write for race detection

newling:
## Summary

Detect write-after-write hazards where an instruction overwrites a VGPR that still has a pending asynchronous load.

This is the race-detector follow-up to #8798, which added execution-plugin observation for instruction-visible VGPR writes. This PR uses that hook without changing the instruction execution or plugin APIs.

ISSUE ID: #7950

## Motivation

A memory instruction can initiate a write to a VGPR before its result is available. If another instruction writes the same VGPR before the corresponding wait count is satisfied, the two writes can complete in the opposite order from which they were issued.

For example, this sequence is racy:

```asm
global_load_dword v2, v[0:1], off
v_add_f32 v2, v3, v4
```

The `v_add_f32` writes `v2` immediately, but the earlier global load may still be in flight. If that load completes afterward, it overwrites the result of the add.

The ordered version waits for the asynchronous write first:

```asm
global_load_dword v2, v[0:1], off
s_waitcnt vmcnt(0)
v_add_f32 v2, v3, v4
```

The same issue exists for LDS reads, which are tracked by `lgkmcnt`:

```asm
ds_read_b32 v2, v0
v_mov_b32 v2, v3
```

This is racy until an appropriate wait is inserted:

```asm
ds_read_b32 v2, v0
s_waitcnt lgkmcnt(0)
v_mov_b32 v2, v3
```

## What this PR detects

The detector now compares instruction-visible VGPR destination writes with pending global and LDS loads. A race is reported when the destination overlaps the pending load in register, lane, and byte range.

This applies to ordinary VALU destinations:

```asm
global_load_dword v2, v[0:1], off
v_add_f32 v2, v3, v4
```

It also applies to multi-register destinations. For example, an MFMA can overwrite registers targeted by a pending wide load:

```asm
global_load_dwordx4 v[0:3], v[24:25], off
v_mfma_f32_4x4x1_16b_f32 v[0:3], v4, v5, 0
```

The gfx1151 WMMA path is covered as well:

```asm
global_load_dwordx4 v[0:3], v[24:25], off
v_wmma_f32_16x16x16_f16 v[0:7], v[8:15], v[16:23], 0
```

Sub-dword overlap is handled using byte masks. A pending low-half load does not conflict with an instruction that writes only the high half, but it does conflict with a write to the same low-half bytes.

Each violation now records the exact pending memory event that caused it. This matters when several pending loads target the same register: diagnostics and instruction traces identify the actual conflicting load rather than searching for an arbitrary matching event.

Structured reports also distinguish the access that exposed the hazard:

```text
type=VGPR access=write
```

## What this PR does not detect

This PR does not attempt to compare two asynchronous VGPR writers.

For example, a global load and an LDS read can target the same VGPR while using different wait-count domains:

```asm
global_load_dword v2, v[0:1], off
ds_read_b32 v2, v3
```

The relative completion order of those writes requires additional memory-event and wait-domain reasoning. That case remains outside this PR.

LDS write-after-write hazards are also not included. Two waves can write the same LDS address without a barrier:

```asm
; wave 0
ds_write_b32 v0, v1

; wave 1, with v0 resolving to the same LDS address
ds_write_b32 v0, v2
```

The current change does not report that collision.

Similarly, it does not compare a pending direct-to-LDS global write with a later LDS instruction write:

```asm
buffer_load_dword v0, s[0:3], 0 offen lds
ds_write_b32 v0, v1
```

Those writes use different wait-count domains (`vmcnt` and `lgkmcnt`) and need separate mixed-domain LDS handling.

SGPR write-after-write is not included either:

```asm
s_load_dword s4, s[0:1], 0
s_mov_b32 s4, s5
```

The detector also does not identify collisions between lanes of one vector instruction. If several lanes execute:

```asm
ds_write_b32 v0, v1
```

and their `v0` values resolve to the same LDS address, that same-instruction collision is outside the current event model.

Finally, DPP and SDWA destination masks are still conservative in the execution-plugin layer. DPP instructions may report more destination lanes than they architecturally modify, while SDWA preserve-mode operations may report more destination bytes. The race detector uses the masks supplied by that layer, so these instructions can conservatively produce false positives for preserved lanes or bytes. Improving those masks is separate execution-effects work.

The broader asynchronous-writer, LDS WAW, and SGPR WAW cases remain follow-up work under #7950.

## Testing

The unit coverage checks wait-count retirement, clean destination registers, lane-mask overlap, sub-dword byte overlap, wave32 behavior, and multiple pending events targeting the same register.

End-to-end HIP coverage exercises:

- a gfx950 VALU destination;
- a gfx950 MFMA destination;
- a gfx950 64-bit destination;
- a gfx1151 WMMA destination.

The focused result was 101 unit tests passed with one architecture-dependent MFMA fast-path test skipped. All four WAW HIP tests passed. The branch-diff pre-commit checks also passed.


emulation/rocjitsu/tests/race-detector/hip_race_tests.hip
```diff
@@ -1162,6 +1354,21 @@ __global__ void f64_vgpr_safe_kernel(const double *src, double *dst) {
...
+  double val;
+  double tmp = static_cast<double>(tid) + 1.0;
+  asm volatile(
+      "global_load_dwordx2 %0, %1, off\n"
+      // BUG: no s_waitcnt before f64 ALU write to the same VGPR pair.
+      // Use tmp as the source so this isolates WAW rather than RAW.
+      "v_fma_f64 %0, %2, 1.0, 1.0\n"
+      : "=&v"(val) : "v"(&src[tid]), "v"(tmp)
+  );
+  asm volatile("s_waitcnt vmcnt(0)\n" ::: "memory");
```
bjacob: Just checking that you are aware of the gotchas with inline asm on AMDGPU:  each asm statement will insert a certain opaque number of NOPs in an attempt to protect against hazards; that number of NOPs isn't based on the actual instructions and is in general neither minimal not sufficiently large to be correct;  this implies in particular that two back-to-back asm volatile statements are not equivalent to one merged statement (there may be extra NOPs)  (and this is on top of the other reasons why back-to-back asm volatile aren't equivalent to single asm volatile even on traditional CPU architectures, because the no-reordering claim of "asm volatile" is next to impossible to represent and enforce in the compiler anyway).

bjacob (commented):
I'm honestly not sufficiently proficient with the ISA-level memory model to review this.  For example, I didn't realize before reading your PR description that loads-to-VGPR counts as "write to VGPR" and are thus susceptible of being WAW races.

Happy to help with any specific question, or if the LLVM-IR-level memory model is the topic.

emulation/rocjitsu/tests/race-detector/hip_race_tests.hip
newling: That's a good point, and I have had unexpected results with inline assembly working on this plugin. Specifically, I've seen the compiler insert synchronization operations very conservatively around inline assembly, which has undone my attempt at creating races. My general approach currently (although semi-violated in the code we're looking at now) is to make the entire racy code contained in a single volatile block. In the above case, the second inline assembly block is not required to induce the race, it is there to ensure there are no additional races over and above the one I'm trying to create. 

(code update since this comment to make it clearer)

newling:
Quick status update: I pushed an update to try and get the CI failures to stop. The change is orthogonal to this PR though, so I will factor it out. But the fix should land after #7840, as that is a large refactor and will clash. So I will wait for that to land first (this WAW PR is not urgent, I am happy wait a few days before landing it). 
 
 @bjacob not reviewing this is totally fine, I understand your perspective! Thus far Tony has reviewed all my plugin PRs. I am in no rush to land this. 
 
 > I didn't realize before reading your PR description that loads-to-VGPR counts as "write to VGPR" and are thus susceptible of being WAW races.
 
I probably don't use verbs consistently and, TBH, still need to check the ISA sometimes to determine which of the write/load/read instructions are 'towards' registers or away from. But essentially the source of a data movement **to** a VGPR can be 1 of 3 places  (in my mental model for race detection work). (1) another register (2) LDS memory (3) global memory. Any 2 writes from any 2 (different) memory spaces can race with each other in certain situations.

amd-eochoalo:
#7840 has been merged

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/cdna4/operand.cpp
```diff
@@ -1666,7 +1666,11 @@ void Operand::write_scalar(amdgpu::Wavefront &wf, uint32_t val) const {
 void Operand::write_lane(amdgpu::Wavefront &wf, uint32_t lane, uint32_t val) const {
   if (auto off = Isa::resolved_vgpr_offset(opr_type_, encoding_value_)) {
     uint32_t voff = wf.gpr_idx_en() ? amdgpu::apply_gpr_idx(wf, *off, true) : *off;
-    wf.cu().write_vgpr(wf.vgpr_alloc().base + voff, lane, val);
+    uint32_t idx = wf.vgpr_alloc().base + voff;
+    // TODO(newling): audit other AMDGPU Operand::write_lane implementations
+    // before claiming non-SIMD VGPR WAW coverage outside CDNA4/gfx950.
+    wf.cu().notify_vgpr_write(&wf, idx, lane, lane + 1);
```
accauble: This is an auto-generated file. If you want to change it you gotta update amdisa.

accauble (changes requested):
This makes changes to an auto-generated file. When that is fixed I can do an actual review.

atgutier (commented):
The agent found some issues:

The detector core is correct; the two problems are at the edges                   16:43:01 [50/1961]

  The WAW event-lifecycle model, LDS domain logic, barrier/waitcnt scoping, byte/lane overlap tests,
  thread-safety (single-RaceDetector-per-workgroup, single-threaded per CU), and the RAW/WAR
  non-regression are all verified correct. But there's a false-negative gap in hook coverage and a
  false-positive gap in the VGPR check — and they're mirror images of the same root cause.

  BLOCKER

  MFMA VGPR/AccVGPR destination writes bypass the write hook — shared/mma_exec.h (write sites
  ~1090/1210/1306). The whole "instruction-destination VGPR WAW" case depends on every VGPR dst write
  firing the hook, but MFMA — the flagship VGPR/accumulator-writing class on gfx950 — writes results
  through the shared matrix engine via bare cu.write_vgpr(...) with no notify_vgpr_write. So a
  buffer_load/global_load into an accumulator VGPR followed by an MFMA writing the same VGPR before
  s_waitcnt vmcnt(0) — the exact WAW pattern in real GEMM prologues — is silently undetected on the
  ISA this PR claims to support. Fix: route MFMA writeback through notify_vgpr_write (or a
  write_vgpr_hooked helper) before each store; add a gfx950 load-into-acc→MFMA→no-waitcnt test.

  MAJOR

  1. VGPR WAW check omits the wait-domain gate → false positives — wave_race_state.cpp:200
  (checkVgprWrite). The LDS path correctly treats same-wave same-domain writes as ordered (they retire
   in issue order), but the VGPR path applies no domain distinction, so two global_load v0 (both
  vmcnt) with no intervening waitcnt are flagged as a race when they're not. Root cause is the
  copy-paste-by-omission the split was designed to surface: waitDomain() is anonymous-namespace-local
  to race_detector.cpp, so the VGPR check in wave_race_state.cpp can't reuse it. Fix: hoist waitDomain
   to a shared constexpr header; gate the memory-vs-memory registration path on domain-match (skip
  same-domain pending writes) while keeping the ALU-dst path flagging all domains (pass std::nullopt,
  since a synchronous ALU write can be clobbered by any later-landing async load).
  2. v_cvt_scalef32_* VGPR-dst writes bypass the hook — cdna4/vop3.cpp (18 sites). These microscaling
  FP-conversion ops write via a hand-written wr lambda calling cu.write_vgpr directly, bypassing both
  the hooked Operand::write_lane and notify_vgpr_write — and inconsistently, since sibling writes on
  the same instruction use the hooked vdst.write_lane. Another gfx950 false-negative. Fix: add
  notify_vgpr_write inside the wr lambda or convert to write_lane.
  3. Tests exercise only the hooked ALU/mem paths — the HIP WAW cases (waw_global_load_then_alu, etc.)
   all use v_add/v_fma which go through the hooked path, so the suite proves the easy cases and never
  touches MFMA or scalef32. The PR frames the coverage gap as a non-gfx950 concern ("audit other ISAs
  later"), but the two findings above show gfx950 itself has unhooked instruction-dst writes. Either
  extend the hook to those classes or add explicit "known-gap" negative tests so the limitation is
  visible rather than implied-covered.

MINOR

  - checkVgprWrite doxygen documents only the mixed-domain case and omits the (buggy) same-domain
  over-flagging — update once the MAJOR is fixed.
  - notify_vgpr_write hardcodes 0xF instead of ExecutionPlugin::kFullByteMask (duplicated magic
  constant, can drift).
  - test_generator_profile_gates.py asserts only the packed-16bit notify emission, not the plain
  write_lane/write_lane64 notifies — a regression dropping those wouldn't be caught by codegen tests.
  - Hardcoded waveSize=64 in the VGPR WAW registration loop (safe today via exec-mask-zero, but
  hardcodes CDNA wave size in the RDNA-capable detector).
  - docs/race-detector.md describes the current over-broad VGPR behavior — tighten after the fix.

  NIT

  No negative "instruction VGPR write over a clean register produces no race" test (would catch a
  future double-fire hook).

newling:
Thank you for the review @atgutier 
I have decided to put this PR into draft mode, as it does 2 things (1) adds plugin hooks for the writes (2) uses those hooks in the race detector. I want to make a PR with just (1), and then rebase this PR onto that once it lands. I think I'll be able to reason about (2) better and address your review comments better.

newling:
The plugin-hook portion has now been split into draft PR #8798.

Once #8798 lands, I will rebase this PR onto the updated `develop` and keep this PR focused on the race-detector behavior.

newling:
Just an FYI to previous reviewers in case I can tempt you into another look at this PR: It is significantly simpler now. I think you can get a fairly accurate idea of what it does from the description now. Before it was covering more racy cases, I've decided to split the more tricky ones out for a future PR.

emulation/rocjitsu/tests/race-detector/hip_race_gfx950_test.hip
```diff
@@ -1120,39 +1122,60 @@ TEST_F(Gfx950RaceTest, scratch_vmcnt_race) {
...
-  // TODO(newling): this should detect a VGPR WAW race, but currently
-  // doesn't. When WAW detection is implemented, change to:
-  //   auto r = records();
-  //   ASSERT_GE(r.size(), 1u);
-  //   EXPECT_NE(r[0].type.find("VGPR"), npos);
-  ExpectNoRace();
+  auto r = records();
+  ASSERT_GE(r.size(), 1u);
+  EXPECT_EQ(r[0].type, "VGPR");
+  EXPECT_EQ(r[0].access, "write");
```
kuhar, line 1178: It would be useful to round out the end-to-end WAW coverage in two places. First, assert the exact trace markers for the ALU and f64 WAW cases, as the MFMA/WMMA tests already do: `global_load_dword` → `v_add_f32`, and `global_load_dwordx2` → `v_fma_f64`. Checking only `type` and `access` would miss a regression that reports the wrong pending event. Second, add a safe `global_load` → `s_waitcnt vmcnt(0)` → same-VGPR instruction write case with `ExpectNoRace()`; the unit test covers the core state transition, but the existing HIP safe cases do not exercise that write hook end to end.

kuhar (commented):
A few comments

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/race_detector/core/types.h
```diff
@@ -67,6 +67,7 @@ struct RaceViolation {
   int lane;     ///< Lane within the wave, or -1 for scalar.
   bool isWrite; ///< True if the violating access was a write.
   Dim3d workgroupId;
+  EventId conflictingEvent; ///< Exact pending memory event that caused the violation.
```
kuhar, line 74: I think we should make the required `conflictingEvent` invariant explicit instead of letting an omitted aggregate initializer silently become `EventId{0}`. `findConflict()` now unconditionally indexes the event registry and always constructs a `MarkedPc`, so a missing field would produce a plausible but wrong first-event PC. Please give the field an invalid sentinel (or require it through a constructor), validate it before lookup, and return `MarkedPc` directly so the API matches the invariant.

emulation/rocjitsu/tests/race-detector/hip_race_gfx950_test.hip
newling: The ALU and f64 tests now check the exact trace markers. There's also now a (HIP) safe load -> waitcnt -> same-VGPR write case that expects no race.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/race_detector/core/types.h
newling: EventId now defaults to a invalid value. RaceViolation requires the conflicting event, and lookup validates it. findConflict() now returns MarkedPc directly, with a test for invalid events.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/race_detector/core/race_detector.cpp
```diff
@@ -127,7 +127,7 @@ void RaceDetector::adjustByteCounts(const IntervalSet &ivs, std::vector<int> &co
 
 std::string
 RaceDetector::decorateException(const RaceViolation &e, uint64_t wavePc,
-                                WaveRaceState *waveRaceState, int numSourceLines,
+                                WaveRaceState * /*waveRaceState*/, int numSourceLines,
```
accauble: Is there a good reason to keep the `WaveRaceState*` around?

accauble (approved):
Looks like you sufficiently addressed @kuhar's comments and otherwise LGTM.

Edit: There is an asan/ubsan test that timed out which really shouldn't have anything to do with this PR but could you please confirm that it is unrelated?

newling:
> Edit: There is an asan/ubsan test that timed out which really shouldn't have anything to do with this PR but could you please confirm that it is unrelated?

It seems to be unrelated, the last run through CI was green. Thanks for offering in chat to investigate this flakiness!

PR #7892, opened 2026-06-26
feat(rocr): AILIROCR-13 Vmem host memory support

pghoshamd:
## Motivation

JIRA ID: AILIROCR-13 
Rocr virtual memory APIs currently only work with GPU agents. Extend them to also work with host memory. 

## JIRA ID

Resolves: AILIROCR-13

## Submission Checklist


projects/rocr-runtime/runtime/hsa-runtime/core/driver/kfd/amd_kfd_driver.cpp
```diff
@@ -618,12 +613,24 @@ hsa_status_t KfdDriver::Unmap(const core::DriverMemoryHandle& handle, void *mem,
...
+  Use the first available GPU agent. */
+  core::Agent* import_agent = const_cast<core::Agent*>(&agent);
+  if (agent.device_type() == core::Agent::DeviceType::kAmdCpuDevice) {
+    const auto& gpus = core::Runtime::runtime_singleton_->gpu_agents();
+    if (gpus.empty()) return HSA_STATUS_ERROR_OUT_OF_RESOURCES;
+    import_agent = gpus.front();
+  }
+
+  if (import_agent_used) *import_agent_used = import_agent;
+
```
ypapadop-amd: Can we handle this before this function is called? I.e., detect it's a CPU agent and then pick the first GPU agent?

We won't have to have the additional argument and touch any of the drivers.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
```diff
@@ -3889,13 +3889,18 @@ hsa_status_t Runtime::VMemoryHandleCreate(const MemoryRegion* region, size_t siz
     uint64_t offset;
     core::DriverMemoryHandle driver_handle = {};
     auto agentOwner = region->owner();
-    auto ret = agentOwner->driver().CreateShareableHandle(nullptr, mem, size, *agentOwner, &driver_handle, &offset);
+    core::Agent* drm_agent = nullptr;
+
```
ypapadop-amd, line 3889: Following up from the previous comment,
```
if (agentOwner == CpuAgent)
  drm_owner = first-gpu-agent
  drm_owner->driver().CreateShareableHandle(...);
  ...
```

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
```diff
@@ -4028,10 +4033,21 @@ Runtime::MappedHandleAllowedAgent::MappedHandleAllowedAgent(
...
+  if (targetAgent->device_type() == core::Agent::DeviceType::kAmdCpuDevice) {
+    return;
+  }
 
   MemoryHandle *memHandle = mappedHandle->mem_handle;
 
+  /* Avoid creating multiple amdgpu bos in the same gpu agent that was used 
+  for drm import of host memory during the creation of a shareable_handle */
+  if (memHandle->agentOwner()->device_type() == core::Agent::DeviceType::kAmdCpuDevice &&
+      memHandle->drm_owner && memHandle->drm_owner == targetAgent) {
```
ypapadop-amd: Isn't `memHandle->agentOwner()->device_type() == core::Agent::DeviceType::kAmdCpuDevice` the definite property that causes the different code path? Maybe we don't need to check the other conditions.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
```diff
@@ -4056,20 +4072,24 @@ Runtime::MappedHandleAllowedAgent::~MappedHandleAllowedAgent() {
...
+    }
   }
 }
 
 hsa_status_t Runtime::MappedHandleAllowedAgent::EnableAccess(hsa_access_permission_t perms) {
   if (targetAgent->device_type() == core::Agent::DeviceType::kAmdCpuDevice) {
     if (core::Runtime::runtime_singleton_->thunkLoader()->IsWslDxg()) return HSA_STATUS_ERROR;
 
-    auto agentOwner = mappedHandle->mem_handle->agentOwner();
+    core::Agent* agent = mappedHandle->mem_handle->drm_owner
```
ypapadop-amd: This ternary is used in multiple places. It'll be easier to understand the logic if it's a new member function of the MemoryHandle.

ypapadop-amd (changes requested):

projects/rocr-runtime/libhsakmt/src/fmm.c
```diff
@@ -2154,7 +2154,17 @@ static void *fmm_allocate_host_gpu(HsaKFDContext *ctx,
...
 
+		if (mflags.ui32.NoAddress) {
+			aperture = &fmm_ctx->mem_handle_aperture;
+		}
+
 		if (mem && mflags.ui32.HostAccess) {
+			/* We do not need to map to cpu for GTT system memory 
+			that has been allocated from mem_handle_aperture */
+			if (mflags.ui32.NoAddress) {
+				return mem;
```
cfreeamd: Early return mem skips vm_obj book-keeping.

The new host NoAddress path returns immediately after __fmm_allocate_device, skipping the vm_obj->mflags /node_id population that the reference device-memory NoAddress path (lines ~1838-1900) sets before falling through. The DMABUF/IPC export path later reads obj->mflags.Value and obj->node_id, so a host memory-only allocation gets exported with uninitialized/zeroed metadata — corrupting the host interprocess export flow that VirtMemory_Interprocess_HostPool_Test exercises. 

Fix:  Remove the if-stmt and return above and add it to the mapping below (2158-2165), like:

>   if (mem && mflags.ui32.HostAccess) {
>       /* GTT system memory from mem_handle_aperture has no VA, so skip CPU mapping */
>       if (!mflags.ui32.NoAddress) {
>           void *ret = fmm_map_to_cpu(mem, MemorySizeInBytes,
>                                      mflags.ui32.HostAccess,
>                                      gpu_drm_fd, mmap_offset);
>           if (ret == MAP_FAILED) {
>               __fmm_release(ctx, vm_obj, aperture);
>               return NULL;
>           }
>       }

cfreeamd (commented):
You will need to update rocrtst/config/platform_config.yaml. It still references the (old) renamed test. It should use the new names.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
pghoshamd: This check is for a situation when user tries to `hsa_amd_vmem_set_access(gpu_agent)` on the same gpu_agent that we internally used for DRM import. Without this, the first available GPU agent would end up with 2 DRM BO's. 

If we remove `memHandle->drm_owner && memHandle->drm_owner == targetAgent`, all GPU agents would try to reuse the same driver_handle when accessing host memory, instead of just the one GPU that created it. Each GPU needs its own import (except the drm_owner), so this would be incorrect.

projects/rocr-runtime/runtime/hsa-runtime/core/inc/runtime.h
```diff
@@ -1032,13 +1032,22 @@ class Runtime {
 
     __forceinline core::Agent* agentOwner() const { return region->owner(); }
 
+    /** 
+     * @brief For host owned memory, resolve to the GPU agent that imported the memory. 
+     * For device owned memory, return the GPU agent that owns the memory.
```
ypapadop-amd:
```suggestion
     * For device owned memory, return the agent that owns the memory.
```

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
```diff
@@ -4114,15 +4159,15 @@ Runtime::MappedHandle::MappedHandle(MemoryHandle *mem_handle, AddressHandle *add
      * need this in the process that owns this memory allocation.
      */
     auto cpu_agent = agentOwner()->GetNearestCpuAgent();
-    auto agentPermsIt = allowed_agents.emplace(std::piecewise_construct,
-                        std::forward_as_tuple(cpu_agent),
-                        std::forward_as_tuple(this, cpu_agent, va,
-                                              size, HSA_ACCESS_PERMISSION_NONE))
-                        .first;
+      auto agentPermsIt = allowed_agents.emplace(std::piecewise_construct,
```
ypapadop-amd: Something happened with formatting here. Maybe run the format script?

ypapadop-amd (approved):
LGTM, but I'd wait for review from another owner too.

projects/rocr-runtime/runtime/hsa-runtime/core/inc/runtime.h
```diff
@@ -1032,13 +1032,22 @@ class Runtime {
 
     __forceinline core::Agent* agentOwner() const { return region->owner(); }
 
+    /** 
+     * @brief For host owned memory, resolve to the GPU agent that imported the memory. 
+     * For device owned memory, return the agent that owns the memory.
+     */
+    __forceinline core::Agent* GetDrmAgent() const {
```
ypapadop-amd: Maybe `drmAgent()` to match the style of `agentOwner()` above?

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_cpu_agent.cpp
```diff
@@ -428,6 +428,18 @@ hsa_status_t CpuAgent::GetInfo(hsa_agent_info_t attribute, void* value) const {
...
+    case HSA_AMD_AGENT_INFO_HOST_ALLOC_DMABUF_SUPPORTED: {
+      /* CPU agents support host memory DMA-BUF allocations if:
+        - Virtual memory APIs are supported
+        - At least one GPU agent exists (needed for drm ops) */
+      bool supported = false;
+      if (core::Runtime::runtime_singleton_->VirtualMemApiSupported()) {
+        const auto& gpus = core::Runtime::runtime_singleton_->gpu_agents();
+        supported = !gpus.empty();
+      }
+      *((bool*)value) = supported;
```
ypapadop-amd:
```suggestion
      *static_cast<bool*>(value) = supported;
```

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -2576,6 +2576,10 @@ hsa_status_t GpuAgent::GetInfo(hsa_agent_info_t attribute, void* value) const {
         *((uint32_t*)value) = 0;
       }
       break;
+    case HSA_AMD_AGENT_INFO_HOST_ALLOC_DMABUF_SUPPORTED:                                                                                                                                        
+      // GPU agents can participate in host memory DMA-BUF export if the system supports virtual memory APIs                                                                                                                                         
+      *((bool*)value) = core::Runtime::runtime_singleton_->VirtualMemApiSupported();
```
ypapadop-amd:
```suggestion
      *static_cast<bool*>(value) = core::Runtime::runtime_singleton_->VirtualMemApiSupported();                                                                                           
```

projects/rocr-runtime/runtime/hsa-runtime/core/inc/runtime.h
```diff
@@ -1032,13 +1032,22 @@ class Runtime {
...
+    }
+
     const MemoryRegion* region;
     int ref_count;
     int use_count;
     DriverMemoryHandle driver_handle;  // handle returned by Driver::Allocate(NoAddress = 1)
     bool imported; // True if this BO was imported from another process
     bool is_fabric_handle;
     MemoryRegion::AllocateFlags alloc_flag;
+    core::Agent* drm_owner; // Gpu agent used for import of host memory, NULL for device memory
```
ypapadop-amd: I think that if we say for device memory, `drm_owner = agentOwner()`, this would simplify the logic below: we never have to check if it's `nullptr`, we always get `drm_owner` and if we ever need to determine if it's device or host memory, we can check if `agentOwner() == drm_owner`.

cfreeamd (approved):

projects/rocr-runtime/runtime/hsa-runtime/inc/hsa_ext_amd.h
```diff
@@ -969,6 +970,12 @@ typedef enum hsa_amd_agent_info_s {
    * Returns uint32_t. Zero if the device does not support dynamic data prefetch.
    */
   HSA_AMD_AGENT_INFO_MAX_DATA_PREFETCH_REGIONS = 0xA123,
+  /**
+   * Queries whether the agent supports allocating host memory that can be
```
cfreeamd: Is this description correct? Is this really supporting allocation of host memory from GPU agents, or is it really meant to say, this agent supports access to host-allocated memory? I'm asking because I'm not sure, but if I had to guess, I would say the latter.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -2576,6 +2576,10 @@ hsa_status_t GpuAgent::GetInfo(hsa_agent_info_t attribute, void* value) const {
         *((uint32_t*)value) = 0;
       }
       break;
+    case HSA_AMD_AGENT_INFO_HOST_ALLOC_DMABUF_SUPPORTED:                                                                                                                                        
+      // GPU agents can participate in host memory DMA-BUF export if the system supports virtual memory APIs                                                                                                                                         
+      *((bool*)value) = core::Runtime::runtime_singleton_->VirtualMemApiSupported();                                                                                           
+    break;
```
cfreeamd: Indent is off here.

projects/rocr-runtime/runtime/hsa-runtime/inc/hsa_ext_amd.h
pghoshamd: The latter is right. This query has different meanings for CPU and GPU agents. I've updated the description to make that distinction more explicit.

cfreeamd (approved):

PR #7898, opened 2026-06-26
fix(rocr): throttle AsyncEventsLoop polling-fallback spin on WSL and mixed batches (closes ROCm/librocdxg#60)

### LINKED ISSUE #60: ROCr AsyncEventsLoop threads busy-spin (~2 CPU cores) for the lifetime of any GPU context under WSL2 — reproduced on system ROCm 7.2.4; interrupt-backed signal events apparently unavailable through /dev/dxg
### Summary

Under WSL2, any process that touches the GPU through ROCm spins two `rocr::core::Runtime::AsyncEventsLoop` threads at \~91% CPU each (\~182% of a core total) for the entire lifetime of its GPU context — completely idle, no work submitted. GPU compute itself is correct and fast; the cost is \~2 host CPU cores burned for as long as a context exists. Reproduced with **stock system ROCm 7.2.4** and independently with the ROCm 7.2 runtime bundled by Ollama 0.30.7 — same behavior on both, so this is not a packaging issue.

### Environment

- GPU: AMD Radeon RX 7900 XTX (24 GB, RDNA 3, gfx1100), Adrenalin 26.6.1
- Host: Windows 11 + WSL2, kernel `6.6.114.1-microsoft-standard-WSL2`, Ubuntu 24.04
- `librocdxg.so.1.2.0`, `libdxcore.so`
- Reproduced against two HSA runtimes: system ROCm 7.2.4 (`libhsa-runtime64.so.1.18.70204`) and Ollama 0.30.7's bundled ROCm 7.2 (`libhsa-runtime64.so.1.18.70201`)

### Minimal reproduction (20 lines, no dependencies beyond hipcc)

```cpp
#include <hip/hip_runtime.h>
#include <cstdio>
#include <unistd.h>

__global__ void noop() {}

int main() {
    if (hipSetDevice(0) != hipSuccess) { fprintf(stderr, "no device\n"); return 1; }
    void* p = nullptr;
    if (hipMalloc(&p, 1 << 20) != hipSuccess) { fprintf(stderr, "alloc failed\n"); return 1; }
    noop<<<1, 1>>>();
    if (hipDeviceSynchronize() != hipSuccess) { fprintf(stderr, "sync failed\n"); return 1; }
    fprintf(stderr, "GPU touched; idling 120s, pid=%d\n", getpid());
    sleep(120);
    hipFree(p);
    return 0;
}
```

```bash
hipcc --offload-arch=gfx1100 -o hipidle hipidle.hip && ./hipidle
# then watch it with top -H while it sleeps
```

Measured while the program sits in `sleep(120)`: **181.8% CPU — two threads at 90.9% each, `wchan=0`** (never blocked in the kernel). The spin starts at context creation and ends only when the process exits.

### Attribution

The two hot threads are ROCr's async-event handlers. With Ollama's bundled runtime (which ships symbol names) the stacks are explicit — both spinners sit in `AsyncEventsLoop`, and every other thread in the process is properly parked in `cond_wait`/`accept4`:

```text
Thread 24 (LWP 328349):
#0  rocr::core::Runtime::AsyncEventsLoop(void*) ()   from libhsa-runtime64.so.1
#1  rocr::os::ThreadTrampoline(void*) ()             from libhsa-runtime64.so.1

Thread 23 (LWP 328350):
#0  rocr::core::Runtime::AsyncEventsLoop(void*) ()   from libhsa-runtime64.so.1
#1  rocr::os::ThreadTrampoline(void*) ()
```

The stripped system runtime shows the identical signature on the minimal reproduction — two threads, two nearby offsets in `libhsa-runtime64.so.1`, each one frame above `start_thread`:

```text
Thread 3 (LWP 336579):
#0  0x00007ffff569ac37 in ?? () from /opt/rocm-7.2.4/lib/libhsa-runtime64.so.1
#1  0x00007ffff56f9a31 in ?? () from /opt/rocm-7.2.4/lib/libhsa-runtime64.so.1
#2  start_thread (...) at ./nptl/pthread_create.c:447

Thread 2 (LWP 336578):
#0  0x00007ffff569ac84 in ?? () from /opt/rocm-7.2.4/lib/libhsa-runtime64.so.1
#1  0x00007ffff56f9a31 in ?? () from /opt/rocm-7.2.4/lib/libhsa-runtime64.so.1
#2  start_thread (...) at ./nptl/pthread_create.c:447
```

Neither spinner ever enters a kernel wait — whatever wait they call returns immediately, so the event loop degenerates to a userspace poll.

### Supporting evidence from the original (Ollama) observation

The bug was first noticed on Ollama 0.30.7: a GGUF model merely loaded to GPU (zero requests) pins the same two threads; evicting the model returns CPU to baseline instantly. Reloading the **same model with GPU offload disabled** (`num_gpu: 0`) idles at **0.0%** CPU — same binary, same server — so the spin is exclusively on the GPU path.

<details>
<summary>Full captured thread dump from the Ollama runner (main thread + first 80 lines of <code>thread apply all bt 8</code> — includes both spinners and representative parked threads; lower-numbered threads fell outside the capture window, and per-thread CPU sampling showed no thread other than LWPs 328349/328350 above 1%)</summary>

```text
--- main thread (tid 328346) ---
#0  0x0000706b99098d71 in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x7ffd2ad39a70, op=393, expected=0, futex_word=0x15761b98) at ./nptl/futex-internal.c:57
#1  __futex_abstimed_wait_common (cancel=true, private=0, abstime=0x7ffd2ad39a70, clockid=0, expected=0, futex_word=0x15761b98) at ./nptl/futex-internal.c:87
#2  __GI___futex_abstimed_wait_cancelable64 (futex_word=futex_word@entry=0x15761b98, expected=expected@entry=0, clockid=clockid@entry=0, abstime=abstime@entry=0x7ffd2ad39a70, private=private@entry=0) at ./nptl/futex-internal.c:139
#3  0x0000706b9909bc8e in __pthread_cond_wait_common (abstime=0x7ffd2ad39a70, clockid=0, mutex=0x15761b48, cond=0x15761b70) at ./nptl/pthread_cond_wait.c:503
#4  ___pthread_cond_timedwait64 (cond=0x15761b70, mutex=0x15761b48, abstime=0x7ffd2ad39a70) at ./nptl/pthread_cond_wait.c:652
#5  0x0000706b9a3ca9f1 in server_queue::start_loop(long) () from /usr/local/lib/ollama/libllama-server-impl.so
#6  0x0000706b9a2b1608 in llama_server(int, char**) () from /usr/local/lib/ollama/libllama-server-impl.so
#7  0x0000706b9902a1ca in __libc_start_call_main (...) at ../sysdeps/nptl/libc_start_call_main.h:58
#8  0x0000706b9902a28b in __libc_start_main_impl (...) at ../csu/libc-start.c:360
#9  0x000000000040107e in ?? ()

Thread 25 (Thread 0x706b98fff6c0 (LWP 328348) "llama-server"):
#0  0x0000706b99098d71 in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x14e4b0e8) at ./nptl/futex-internal.c:57
#1  __futex_abstimed_wait_common (cancel=true, private=0, abstime=0x0, clockid=0, expected=0, futex_word=0x14e4b0e8) at ./nptl/futex-internal.c:87
#2  __GI___futex_abstimed_wait_cancelable64 (futex_word=futex_word@entry=0x14e4b0e8, expected=expected@entry=0, clockid=clockid@entry=0, abstime=abstime@entry=0x0, private=private@entry=0) at ./nptl/futex-internal.c:139
#3  0x0000706b9909b7ed in __pthread_cond_wait_common (abstime=0x0, clockid=0, mutex=0x14e4b090, cond=0x14e4b0c0) at ./nptl/pthread_cond_wait.c:503
#4  ___pthread_cond_wait (cond=0x14e4b0c0, mutex=0x14e4b090) at ./nptl/pthread_cond_wait.c:627
#5  0x0000706b994b9047 in std::condition_variable::wait(std::unique_lock<std::mutex>&) () from /lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x0000706b99e51233 in common_log::resume()::{lambda()#1}::operator()() const () from /usr/local/lib/ollama/libllama-common.so.0
#7  0x0000706b994ecdb4 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6

Thread 24 (Thread 0x706b1a3ff6c0 (LWP 328349) "llama-server"):
#0  0x0000706b98286aa6 in rocr::core::Runtime::AsyncEventsLoop(void*) () from /usr/local/lib/ollama/rocm_v7_2/libhsa-runtime64.so.1
#1  0x0000706b982e34dd in rocr::os::ThreadTrampoline(void*) () from /usr/local/lib/ollama/rocm_v7_2/libhsa-runtime64.so.1
#2  0x0000706b9909caa4 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:447
#3  0x0000706b99129c6c in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:78

Thread 23 (Thread 0x706b19bfe6c0 (LWP 328350) "llama-server"):
#0  0x0000706b982868b7 in rocr::core::Runtime::AsyncEventsLoop(void*) () from /usr/local/lib/ollama/rocm_v7_2/libhsa-runtime64.so.1
#1  0x0000706b982e34dd in rocr::os::ThreadTrampoline(void*) () from /usr/local/lib/ollama/rocm_v7_2/libhsa-runtime64.so.1
#2  0x0000706b9909caa4 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:447
#3  0x0000706b99129c6c in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:78

Thread 22 (Thread 0x706b18bfc6c0 (LWP 328352) "llama-server"):
#0  0x0000706b9912b914 in accept4 (fd=7, addr=..., addr_len=0x0, flags=524288) at ../sysdeps/unix/sysv/linux/accept4.c:31
#1  0x0000706b9a3e5d71 in httplib::Server::listen_internal() () from /usr/local/lib/ollama/libllama-server-impl.so
#2  0x0000706b994ecdb4 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x0000706b9909caa4 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:447
#4  0x0000706b99129c6c in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:78

Threads 21-17 (LWPs 328353-328357): all identical —
#0-#4 __futex_abstimed_wait_common64 / pthread_cond_wait
#5  std::condition_variable::wait
#6  httplib::ThreadPool::worker(bool) () from /usr/local/lib/ollama/libllama-server-impl.so
```

</details>

### Suspected mechanism

On bare-metal Linux, `AsyncEventsLoop` blocks in the kernel via interrupt-backed KFD signal events. The polling fallback that pins a core when interrupt-backed events are unavailable is long-known (ROCm/ROCm#748). Under WSL2 there is no `/dev/kfd`; the thunk layer is librocdxg over `/dev/dxg` — and the observed behavior matches interrupt-backed signal events not being available on this path, leaving every signal on the memory-polling fallback. Notably, both reproducing runtimes postdate the ROCm 6.3.2 `AsyncEventsLoop` anti-spin improvements, which apparently cannot engage here.

### Impact

\~2 CPU cores burned continuously for the lifetime of any GPU context under WSL2. For inference servers this is severe: Ollama's default keep-alive holds a model (and the spin) for minutes after every request, and a server keeping a model warm spins indefinitely.

---

🤖 This investigation and report were generated with [Claude Code](https://claude.com/claude-code); the measurements, bisect, and thread dumps were captured on the reporter's machine and reviewed by the reporter.

### LINKED ISSUE #5534: [Issue]: [clr][Windows] Two busy-looping threads during idle (related to TheRock Issue: PyTorch high CPU usage after 7.13.0a20260408 #4825)
### Problem Description

Simply initializing the ROCm runtime environment results in two threads continuously running in the background.
(related to [[Issue]: Pytorch high CPU usage after 7.13.0a20260408 #4825](https://github.com/ROCm/TheRock/issues/4825))

### Operating System

Windows 11 10.0.26200

### CPU

AMD Ryzen 5 7600 6-Core Processor

### GPU

AMD Radeon RX 9060 XT

### ROCm Version

7.13.0a20260416

### ROCm Component

clr

### Steps to Reproduce

Start Python:
```
>>> import torch
>>> torch.cuda.is_available()
```

_No response_

### What I found during debugging
[hsa_init() in rocr-runtime/runtime/hsa-runtime/core/runtime/hsa.cpp](https://github.com/ROCm/rocm-systems/blob/7d1b6c033cdcc61b437f4a6210ba65e5862ef0e8/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa.cpp#L204) calls [Runtime::Load() in rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp](https://github.com/ROCm/rocm-systems/blob/7d1b6c033cdcc61b437f4a6210ba65e5862ef0e8/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp#L2361).
Inside this function, the [following lines](https://github.com/ROCm/rocm-systems/blob/7d1b6c033cdcc61b437f4a6210ba65e5862ef0e8/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp#L2395):
```
  asyncSignals_.reset(new AsyncEventsInfo(false));
  asyncExceptions_.reset(new AsyncEventsInfo(g_use_interrupt_wait));
```
create two threads that execute [Runtime::AsyncEventsLoop](https://github.com/ROCm/rocm-systems/blob/7d1b6c033cdcc61b437f4a6210ba65e5862ef0e8/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp#L1702), and both threads keep looping indefinitely.

On my Windows environment, [amd::Device::init in rocm-systems/projects/clr/rocclr/device/device.cpp](https://github.com/ROCm/rocm-systems/blob/7d1b6c033cdcc61b437f4a6210ba65e5862ef0e8/projects/clr/rocclr/device/device.cpp#L671) calls [roc::Device::init](https://github.com/ROCm/rocm-systems/blob/7d1b6c033cdcc61b437f4a6210ba65e5862ef0e8/projects/clr/rocclr/device/rocm/rocdevice.cpp#L346), which then calls [Hsa::init](https://github.com/ROCm/rocm-systems/blob/7d1b6c033cdcc61b437f4a6210ba65e5862ef0e8/projects/clr/rocclr/device/rocm/rocrctx.hpp#L164) (which calls hsa_init).
After that, the [following code](https://github.com/ROCm/rocm-systems/blob/7d1b6c033cdcc61b437f4a6210ba65e5862ef0e8/projects/clr/rocclr/device/rocm/rocdevice.cpp#L461) returns false:
```
if (IS_WINDOWS && (GPU_ENABLE_PAL == 2) &&
    (std::string_view(roc_device->info().name_).find("gfx942") == std::string_view::npos)) {
  return false;
}
```
This causes the runtime to switch to PAL.
However, [Hsa::shut_down()](https://github.com/ROCm/rocm-systems/blob/7d1b6c033cdcc61b437f4a6210ba65e5862ef0e8/projects/clr/rocclr/device/rocm/rocrctx.hpp#L165) is not called during this transition, so the two HSA threads remain running.

### Suspected Cause
Previously, the lifetime of clr matched either the HSA runtime or PAL, so calling hsa_shutdown() during execution was unnecessary.
But after [TheRock PR "Enable PAL + Rocr backend for HIP/OCL" #4044](https://github.com/ROCm/TheRock/pull/4044), this assumption no longer holds.

When HSA initialization fails and the system switches to PAL, the HSA runtime now needs to be shut down explicitly — but this is not happening.

### Proposed Fix
Hsa::shut_down() should be called at an appropriate point.
Since Hsa::shut_down() is already implemented inside roc::Device::tearDown(), it may be better to call roc::Device::tearDown() instead.

For example, [adding roc::Device::tearDown() inside amd::Device::init()](https://github.com/ROCm/rocm-systems/blob/7d1b6c033cdcc61b437f4a6210ba65e5862ef0e8/projects/clr/rocclr/device/device.cpp#L703):
```
ret = roc::Device::init();
if (!ret) {
  // abort() commented because this is the only indication
  // that KFD is not installed.
  // Ignore the failure and assume KFD is not installed.
  // abort();
  ClPrint(amd::LOG_DETAIL_DEBUG, amd::LOG_INIT, "KFD is not installed");
  // Disable direct dispatch if ROC initialization wasn't successful
  if (flagIsDefault(AMD_DIRECT_DISPATCH)) {
    AMD_DIRECT_DISPATCH = false;
  }
  roc::Device::tearDown();  //<-
  GPU_ENABLE_PAL = 1;
} else {
```

### Additional Notes
The fact that Runtime::AsyncEventsLoop keeps looping indefinitely is also concerning, but this should probably be treated as a separate issue.

seanthegeek:
## Problem

When `Runtime::AsyncEventsLoop` (`core/runtime/runtime.cpp`) cannot arm an interrupt-backed wait, it falls back to polling: `PrepareInterrupt()` returns `false`, the loop sets `polling = true`, and the inner loop re-scans signal values in userspace. In that mode the loop had **no sleep**, so each `AsyncEventsLoop` thread monopolized a CPU core for the entire lifetime of its async-events context.

The polling fallback is reachable in two distinct situations:

1. **Globally** — the thunk exposes no interrupt-backed signal events at all (`g_use_interrupt_wait == false`): WSL2/dxg (where `librocdxg` stubs `hsaKmtCreateEvent`), DTIF, KFD 1.0, or `HSA_ENABLE_INTERRUPT=0`. Every signal forces polling. On WSL2 any process that has merely touched the GPU burns **~2 host CPU cores** while completely idle — reported and root-caused in ROCm/librocdxg#60.
2. **Per wait batch, on ordinary interrupt-capable Linux** — a single registered signal with no `EopEvent()` (an IPC signal registered via `hsa_amd_signal_async_handler`, which explicitly permits `IPCSignal`, or an internal `DefaultSignal` such as gang-copy signals) forces the *entire* batch into polling, dragging unrelated interrupt-backed handlers on the shared thread with it. This pre-existing behavior was identified during review ([expert-agent review](https://github.com/ROCm/rocm-systems/pull/7898#pullrequestreview-4856707566)); earlier revisions of this PR wrongly claimed the branch was unreachable on interrupt-backed platforms.

## Fix

Nap between scans in the forced-polling branch with an escalating backoff: **20 µs doubling up to a ceiling**, re-initialized at every new wait batch so escalation only compounds within a single idle wait. The ceiling depends on *why* the loop is polling:

- **Interrupts globally unavailable** (`g_use_interrupt_wait == false`): ceiling **2 ms** (`kPollNapCeilingUs`). Every signal is polling-only, so the only cost of a long nap is the napping wait's own observation latency.
- **Mixed batch on an interrupt-capable system** (`g_use_interrupt_wait == true`): ceiling **200 µs** (`kPollNapCeilingMixedUs`). The nap here delays *unrelated* interrupt-backed handlers on the shared thread, so it is capped at the same 200 µs the interrupt path already spends in its active-poll window before `hsaKmtWaitOnMultipleEvents` — the thread's pre-existing latency noise floor.

```cpp
const int poll_nap_ceiling_us =
    g_use_interrupt_wait ? kPollNapCeilingMixedUs : kPollNapCeilingUs;
...
} else if (polling && !finish) {
  os::uSleep(poll_nap_us);
  poll_nap_us = NextPollNapUs(poll_nap_us, poll_nap_ceiling_us);
}
```

Pure interrupt-backed batches take the `WaitForInterrupt()` (KFD sleep) path and never enter this branch — verified by measurement below. Kernel-completion waits use the separate userspace `BusyWaitSignal` path and are likewise untouched.

## Measurements

**WSL2** — RX 7900 XTX (gfx1100), ROCm 7.2.4, native `libhsa-runtime64` swapped via `LD_LIBRARY_PATH`:

| scenario | stock | patched |
| --- | ---: | ---: |
| idle with a model/context resident | **182%** CPU | **1.07%** CPU |
| 200 back-to-back kernel + `hipDeviceSynchronize` iterations | 196% CPU, 0.22 s | **44% CPU, 0.23 s** |

Wall-clock for back-to-back work is unchanged (within noise) because every new wait resets to the 20 µs floor.

**Native Linux, mixed batch** — Ryzen 9950X3D, Radeon AI PRO R9700 (gfx1201) + RX 7900 XTX (gfx1100), interrupts enabled, stock KFD. Interrupt-backed async-handler callback latency over 300 sparse completions while a never-satisfied IPC-signal handler is parked on the same async-events thread:

| build | idle CPU (async thread) | callback latency p50 / p99 / max |
| --- | ---: | ---: |
| pre-PR spin | 99.9% of a core | 0.2 / 0.7 / 1.8 µs |
| 2 ms cap (earlier revision of this PR) | 0.2% | 1219 / 2046 / 2053 µs |
| **this PR (200 µs mixed cap)** | 0.6% | **119 / 249 / 251 µs** |

Pure interrupt-backed batches are unchanged in all three builds (p50 ≈ 10 µs, 0% idle CPU).

## Reproduction

A 20-line HIP program that touches the GPU once and then sleeps reproduces the spin on stock ROCm under WSL2 (full source in ROCm/librocdxg#60); with this patch the same program idles at ~1% CPU instead of ~182%. The native-Linux mixed-batch case is reproduced by registering `hsa_amd_signal_async_handler` callbacks on one interrupt-backed signal plus one IPC-attached signal and timing the interrupt signal's callbacks.

## Testing

The nap escalation is factored into a small dependency-free helper (`core/util/poll_backoff.h`: `kPollNapFloorUs`, `kPollNapCeilingUs`, `kPollNapCeilingMixedUs`, `NextPollNapUs`) and covered by `rocrtst/common/utils_test/poll_backoff_test.cpp` (gtest): bound ordering (floor < mixed < full), the doubling escalation, saturation and fixed-point behavior at both ceilings, and the per-wait reset to the floor. All 5 cases pass.

Closes ROCm/librocdxg#60

---

🤖 This patch was developed with [Claude Code](https://claude.com/claude-code) (Claude Fable 5); the diagnosis, builds, and before/after measurements on WSL2 and native Linux were captured and reviewed on the contributor's machines.


seanthegeek:
Supersedes #7214, which GitHub auto-closed when its head branch was renamed to satisfy the branch-name policy. Same fix, now with a compliant branch name (`users/seanthegeek/…`), a Conventional Commits title, and an accompanying unit test (`poll_backoff_test.cpp`). Prior review discussion is on #7214.

projects/rocr-runtime/runtime/hsa-runtime/core/util/poll_backoff.h
```diff
@@ -0,0 +1,75 @@
+////////////////////////////////////////////////////////////////////////////////
```
dayatsin-amd, line 1: License should be MIT

/*
 * Copyright © Advanced Micro Devices, Inc., or its affiliates.
 *
 * SPDX-License-Identifier: MIT
 */

dayatsin-amd (commented):
Other than the license header, then changes look ok.
But we would need to some performance tests on Linux to make sure this is not causing perf regressions.

projects/rocr-runtime/runtime/hsa-runtime/core/util/poll_backoff.h
seanthegeek: Fixed

projects/rocr-runtime/runtime/hsa-runtime/core/util/poll_backoff.h
```diff
@@ -0,0 +1,66 @@
...
+//
+//
+// Copyright © Advanced Micro Devices, Inc., or its affiliates.
+/// SPDX-License-Identifier: MIT
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** SPDX line uses `///` (Doxygen triple-slash) instead of `//`. Other utility headers in this directory (`os.h`, `memory.h`) use the standard NCSA template with `//` throughout. Either use the NCSA template to match the rest of the directory, or at minimum fix the `///` to `//`.

projects/rocr-runtime/runtime/hsa-runtime/core/util/poll_backoff.h
```diff
@@ -0,0 +1,66 @@
...
+
+#include <algorithm>
+
+namespace rocr {
+namespace core {
+
+// Default floor and ceiling (microseconds) for the polling-fallback nap.
+constexpr int kPollNapFloorUs = 20;
+constexpr int kPollNapCeilingUs = 2000;
+
```
cfreeamd, line 42: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** `current_us * 2` is a signed `int` multiply. With the current constants (floor 20, ceiling 2000) this saturates safely, but `ceiling_us` is a parameter — a caller passing `current_us > INT_MAX/2` gets UB before `std::min` can clamp it, and `os::uSleep` receiving a wrapped negative value would sleep for a very long time. Consider using `unsigned int` or adding a comment documenting the valid input range.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
```diff
@@ -2003,6 +2010,22 @@ void Runtime::AsyncEventsLoop(void* _eventsInfo) {
...
+          // that is every signal, since it implements no KFD events at all.
+          // Without a sleep this loop monopolizes a CPU core re-scanning the
+          // signal values for the whole lifetime of the async-events thread,
+          // idle or not; the two async-events threads cost ~2 cores in every
+          // process that has merely touched the GPU
+          // (https://github.com/ROCm/librocdxg/issues/60). Nap between scans,
+          // doubling from 20us up to a 2ms cap, so a wait that completes
+          // quickly keeps low observation latency while a long-lived idle
+          // wait costs almost no CPU. Progress resets the escalation: a
+          // processed event leaves this inner loop, and the next wait starts
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** "a processed event leaves this inner loop, and the next wait starts again from the floor" — the second half is accurate but the first half overstates. `poll_nap_us` resets at the outer-loop boundary (once per wait batch), not per processed event within a batch. In non-`wait_any` mode a handler can fire without immediately leaving the inner loop. Suggest: "the nap is re-initialized at the outer-loop boundary so escalation only compounds within one idle wait batch."

projects/rocr-runtime/rocrtst/common/utils_test/poll_backoff_test.cpp
```diff
@@ -0,0 +1,65 @@
...
+// long-lived idle wait stays at the cheap-CPU cap rather than overflowing.
+TEST(PollBackoffTest, CeilingIsFixedPoint) {
+  EXPECT_EQ(NextPollNapUs(kPollNapCeilingUs), kPollNapCeilingUs);
+  EXPECT_EQ(NextPollNapUs(2 * kPollNapCeilingUs), kPollNapCeilingUs);
+}
+
+// AsyncEventsLoop resets the nap to the floor on every new wait batch (via
+// block scope). Emulate that reset and confirm a fresh wait starts cheap again
+// regardless of how far a previous idle wait had escalated -- the escalated
+// value never carries across waits.
```
cfreeamd, line 76: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** Hard-coded `10` assumes 10 iterations reaches ceiling (true today: floor=20, ceiling=2000 saturates in 7 steps). Will give a false pass if constants change. More robust:
```cpp
int nap = kPollNapFloorUs;
while (nap < kPollNapCeilingUs) nap = NextPollNapUs(nap);
EXPECT_EQ(nap, kPollNapCeilingUs);
```

projects/rocr-runtime/rocrtst/common/utils_test/CMakeLists.txt
```diff
@@ -4,13 +4,16 @@
...
+set (rocrtstUtilsTestSrcs ${rocrtstUtilsTestSrcs} poll_backoff_test.cpp)
 
 #
 # Header files include path(s).
 #
 include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 include_directories(${PROJECT_SOURCE_DIR}/utils)
 include_directories(${PROJECT_SOURCE_DIR}/gtest/include)
+# hsa-runtime source root, for the core/util/poll_backoff.h helper under test.
+include_directories(${CMAKE_CURRENT_SOURCE_DIR}/../../../runtime/hsa-runtime)
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** Three `..` hops is fragile — will silently break if the directory structure is reorganized. Prefer anchoring to a project-root CMake variable that's already defined (e.g. `${ROCR_INC_PATH}` or equivalent) rather than counting relative levels.

cfreeamd (commented):
🤖 AI-assisted pre-review (for @cfreeamd) — `runtime.cpp`, `poll_backoff_test.cpp`, and `CMakeLists.txt` changes look correct. Inline comments above cover items worth addressing before merge.

projects/rocr-runtime/runtime/hsa-runtime/core/util/poll_backoff.h
seanthegeek: Fixed in 4bc6552 — the `///` on the SPDX line is now `//`, consistent with the rest of the NCSA/MIT template this header already uses.

projects/rocr-runtime/runtime/hsa-runtime/core/util/poll_backoff.h
seanthegeek: Fixed in 4bc6552. Kept `int` and made the multiply overflow-safe by only evaluating `current_us * 2` when `current_us <= ceiling_us / 2`:

```cpp
return (current_us > ceiling_us / 2) ? ceiling_us
                                     : std::min(current_us * 2, ceiling_us);
```

This is equivalent to the previous expression for all inputs (checked the odd/even `ceiling_us` boundary cases) and removes the UB for any `ceiling_us` up to `INT_MAX`. The gtest still passes.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
seanthegeek: Fixed in 4bc6552. Reworded to: "The nap is re-initialized at the outer-loop boundary (see poll_nap_us above), so the escalation only compounds within a single idle wait batch."

projects/rocr-runtime/rocrtst/common/utils_test/poll_backoff_test.cpp
seanthegeek: Fixed in 4bc6552 — `ResetReturnsToFloor` now escalates until the ceiling instead of a fixed count:

```cpp
int nap = kPollNapFloorUs;
while (nap < kPollNapCeilingUs) nap = NextPollNapUs(nap);
EXPECT_EQ(nap, kPollNapCeilingUs);
```

so it stays correct if the constants change.

projects/rocr-runtime/rocrtst/common/utils_test/CMakeLists.txt
seanthegeek: Fixed in 4bc6552. `${ROCR_INC_PATH}` is not defined in this repo, so I anchored to `${PROJECT_SOURCE_DIR}` — the same project-root variable this file already uses on lines 13-14 — dropping the three `..` hops to one:

```cmake
include_directories(${PROJECT_SOURCE_DIR}/../runtime/hsa-runtime)
```

seanthegeek:
Pushed 4bc6552 addressing the review feedback:

- **License** (@dayatsin-amd): header now carries `SPDX-License-Identifier: MIT` (the stray `///` on that line is also fixed).
- **`NextPollNapUs` signed-multiply** (@cfreeamd, *Important*): made overflow-safe while keeping `int` — the multiply is only evaluated when `current_us <= ceiling_us / 2`, so no signed overflow for any `ceiling_us` up to `INT_MAX`. Behavior is unchanged for the shipping constants.
- Comment wording in `runtime.cpp`, the loop-until-ceiling test in `poll_backoff_test.cpp`, and the CMake include anchored to `${PROJECT_SOURCE_DIR}` — all addressed (replies on the individual threads).

The `poll_backoff` gtest still passes (all 4 cases).

seanthegeek:
**Re: native-Linux perf check**

I ran a targeted measurement on native Linux (no WSL) to check for a regression. Short version: on mainstream interrupt-backed Linux the changed branch is never entered, and I verified that empirically.

**Reachability.** The nap lives in the `polling && !finish` branch of `AsyncEventsLoop`, which is only reached when `PrepareInterrupt()` returns false — i.e. when `g_use_interrupt_wait` is off. On native Linux that happens in exactly three places:

- `core/driver/kfd/amd_kfd_driver.cpp:135-136` — KFD interface version **1.0**
- `core/runtime/amd_topology.cpp:317` — **DTIF** platforms (`enable_dtif`)
- `core/runtime/amd_topology.cpp:320-327` — the **DXG/WSL** thunk (the reported case)

With current KFD on a normal GPU, interrupts stay enabled, the loop takes the `WaitForInterrupt()` (KFD sleep) path, and the new code does not execute — so there is no perf surface to regress there.

**Empirical check** (gfx1201 / Radeon AI PRO R9700, native Linux, ROCm 7.x). A small HSA program registers an idle `hsa_amd_signal_async_handler`, which is serviced by `AsyncEventsLoop`, and I measure this process's CPU over the window:

| Path | idle CPU (of one core) |
|---|---:|
| Interrupts enabled (native default) | **0.0 %** |
| `HSA_ENABLE_INTERRUPT=0` (forces the polling branch; stock/pre-patch lib) | **99.9 %** |

The `0.0%` row confirms the branch isn't entered on the interrupt path — nothing for the change to affect. The `99.9%` row reproduces the exact idle spin this PR targets, on native hardware, using the current released library (so it's the pre-patch "before").

**Where the branch *is* reached** (KFD 1.0 / DTIF / WSL), the change only swaps the busy-spin for an escalating nap that resets to the 20 µs floor on every new wait — so it lowers idle CPU without slowing back-to-back work; the throughput row in the PR description shows wall-clock unchanged. I kept the forced-polling before/after harness and can post a full patched-vs-stock table for that path if wanted, but since native interrupt-backed Linux never runs the code, the measurement above should cover the regression question.

---

🤖 This comment (analysis and measurements) was generated by Claude Code and reviewed on the contributor's machine before posting.

projects/rocr-runtime/runtime/hsa-runtime/core/util/poll_backoff.h
```diff
@@ -0,0 +1,69 @@
+////////////////////////////////////////////////////////////////////////////////
+//
+//
+// Copyright © Advanced Micro Devices, Inc., or its affiliates.
+// SPDX-License-Identifier: MIT
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
```
cfreeamd: With the SPDX part you correctly added above, you should remove this part (lines 7-31). I'm not a legal person, but they might be contradictory, and the SPDX lines 4-5 are all that's needed.

kentrussell: Agreed. Hopefully we'll have some time to get someone to clean all those up.

cfreeamd (approved):

seanthegeek:
Thanks for the approval @cfreeamd. Just curious, what are the next steps before this lands in a release? Also, can someone look at my other PR, which addresses another longstanding bug? https://github.com/ROCm/rocm-systems/pull/7989

cc: @dayatsin-amd

projects/rocr-runtime/runtime/hsa-runtime/core/util/poll_backoff.h
```diff
@@ -0,0 +1,42 @@
...
+#ifndef HSA_RUNTIME_CORE_UTIL_POLL_BACKOFF_H_
+#define HSA_RUNTIME_CORE_UTIL_POLL_BACKOFF_H_
+
+#include <algorithm>
+
+namespace rocr {
+namespace core {
+
+// Default floor and ceiling (microseconds) for the polling-fallback nap.
+constexpr int kPollNapFloorUs = 20;
```
atgutier: Can you make these constexpr variables inline?

atgutier (commented):
Here is a review from an expert agent:


 [MAJOR] projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp:2117 — One polling-only
  signal throttles the entire shared async-event thread on native Linux

  Reasoning: The new backoff is not restricted to WSL or configurations where interrupts are globally
  disabled. PrepareInterrupt() returns false whenever any registered signal has no EopEvent()
  (runtime.cpp:1938-1943). That makes the batch-wide polling flag sticky (:2056-2062), discards
  already-collected interrupt events (:2063-2065), and eventually sleeps for 2 ms before every scan
  (:2117-2132). Consequently, one long-lived polling signal can delay unrelated interrupt-backed
  callbacks, async-function requests, control wakes, and teardown by roughly 2 ms plus scheduler
  latency.

  Evidence:

  - Ordinary native Linux can contain such mixed batches even with g_use_interrupt_wait == true.
  - The public async-handler API explicitly permits IPCSignal at hsa_ext_amd.cpp:1155-1159. An IPC
    signal derives from BusyWaitSignal, whose EopEvent() is null at default_signal.h:154.

  - Native multi-GPU gang copies create a DefaultSignal at amd_gpu_agent.cpp:1221 and register it
    directly with the same shared async-event thread at :1230-1233. A DefaultSignal also has no
    interrupt event.

  - The PR’s native-Linux benchmark registered only an interrupt-backed signal, so it never exercised
    this mixed-batch path. Its claim that interrupt-backed Linux “never enters” the new branch is
    therefore incorrect.

  - The backoff reaches its steady 2 ms sleep after 20+40+80+160+320+640+1280 µs, about 2.54 ms of
    idle time. Sparse completions thereafter can incur approximately 0–2 ms observation latency, with
    no strict upper bound once scheduler delay is included.

  - The unit tests validate only backoff arithmetic. They do not exercise mixed signal types, callback
    latency, control wakes, or shutdown.

  This is unlikely to affect ordinary single-GPU workloads whose async handlers are all interrupt-
  backed. It is a credible performance regression for IPC async handlers, native multi-GPU gang
  copies, polling dependencies, intercepted queues, and callback-driven work chains. Repeated
  dependent callbacks can accumulate the delay.

  Suggested fix: distinguish fully polling runtimes from mixed native batches. Reasonable approaches
  include:

  - Apply the 2 ms ceiling only when interrupts are globally unavailable, while using a much smaller
    cap for mixed batches.

  - Separate polling-only signals from interrupt-backed signals.
  - Wait on the interrupt subset with a finite timeout tied to the polling budget, then scan polling-
    only signals.


  Add an integration test containing an unsatisfied IPC/Default signal plus an interrupt-backed signal
  and verify prompt callback, async-function wake, and shutdown. Benchmark sparse IPC callbacks and
  native gang copies, not only back-to-back WSL kernel synchronization.

  Confidence: High for reachability and the shared-thread latency mechanism; medium for workload-level
  magnitude because no mixed-batch benchmark was supplied.

projects/rocr-runtime/runtime/hsa-runtime/core/util/poll_backoff.h
seanthegeek: Done in bc2227f744 — both existing constants are now `inline constexpr`, along with the new `kPollNapCeilingMixedUs` added in the same commit.

seanthegeek:
## Mixed-batch latency fix (bc2227f744) — response to the [expert-agent review](https://github.com/ROCm/rocm-systems/pull/7898#pullrequestreview-4856707566)

The review's MAJOR finding is correct, and I want to walk back my earlier claim that the backoff only affects WSL. I reproduced the mechanism on native Linux, measured it, and pushed a fix. Details below.

### Confirming the finding

I verified each cited code path and then benchmarked on a native Linux box (Ryzen 9950X3D, Radeon AI PRO R9700 gfx1201 + RX 7900 XTX gfx1100, interrupts enabled, stock KFD). The benchmark registers an async handler on an interrupt-backed signal and measures store→callback latency over 300 sparse completions (randomized 3–8 ms gaps), while a never-satisfied handler on an **IPC-attached signal** — polling-only, `EopEvent() == nullptr`, and explicitly permitted by `hsa_amd_signal_async_handler` even with `g_use_interrupt_wait` — is parked on the same async-events thread. Exactly as the review predicted, that one signal forces `PrepareInterrupt` to fall back to polling for the whole batch:

| Build | Idle CPU (async thread) | Interrupt-signal callback latency p50 / p99 / max |
|---|---|---|
| pre-PR spin (baseline) | **99.9% of a core** | 0.2 / 0.7 / 1.8 µs |
| 2 ms cap (PR before this commit) | 0.2% | **1219 / 2046 / 2053 µs** |
| **this commit (200 µs mixed cap)** | 0.6% | **119 / 249 / 251 µs** |

Pure interrupt-backed batches never enter the nap branch and are unchanged in all three builds (p50 ≈ 10 µs, 0% idle CPU) — so the common single-GPU workload was never affected. But the review is right that IPC async handlers and internal `DefaultSignal` registrations (e.g. gang copies) reach this path on ordinary native Linux, and the original benchmark, which registered only an interrupt-backed signal, could not see it.

### The fix

Following the review's first suggested approach, the nap ceiling now depends on **why** the loop is polling:

- **`g_use_interrupt_wait == false`** (WSL/dxg, DTIF, KFD 1.0, `HSA_ENABLE_INTERRUPT=0`): every signal is polling-only, so a long nap costs only the napping wait's own observation latency. The 2 ms ceiling (`kPollNapCeilingUs`) is kept — the librocdxg#60 fix is unchanged on WSL.
- **`g_use_interrupt_wait == true`** (interrupt-capable system, polling forced by an `EopEvent()`-less signal in a mixed batch): the nap delays *unrelated* interrupt-backed handlers on the shared thread, so it is capped at **200 µs** (`kPollNapCeilingMixedUs`) — the same 200 µs the interrupt path already spends in its active-poll window before `hsaKmtWaitOnMultipleEvents`, i.e. the pre-existing latency noise floor of this thread.

Compared to the pre-PR code, the mixed-batch case on native Linux trades the old busy-spin (sub-µs latency at 99.9% of a core, the librocdxg#60 pathology) for a ≤ 250 µs worst-case observation latency at 0.6% of a core. Escalation still starts at the 20 µs floor and resets every new wait batch, so short-lived mixed batches (a gang copy completing quickly) mostly see naps well under the cap.

Also in this commit:

- The backoff constants are now `inline constexpr` (@atgutier's inline comment).
- `poll_backoff_test.cpp` gains coverage for the mixed ceiling: ordering (floor < mixed < full), saturation at the mixed ceiling, and its fixed-point property. All 5 tests pass locally.

### Remaining caveats, honestly stated

- The batch-wide "one polling signal disables interrupt waits for everyone" behavior is pre-existing and untouched here; this commit only bounds its latency cost. The review's more invasive suggestion (wait on the interrupt subset with a finite timeout, then scan polling-only signals) would remove even the 200 µs bound and could be follow-up work.
- The 200 µs figure for the mixed ceiling is a judgment call anchored to the existing active-poll window; I'm happy to tune it if the team prefers a different bound.
- My benchmark exercises the long-lived IPC-signal case directly; gang-copy `DefaultSignal`s are transient and were not benchmarked separately (this hardware has no XGMI links), but they run through the identical code path with strictly shorter escalation.

---

🤖 This analysis, the reproduction benchmark, and the fix in bc2227f744 were developed with [Claude Code](https://claude.com/claude-code) (Claude Fable 5); the code-path verification and before/after measurements were run and reviewed on the contributor's native Linux machine.

seanthegeek:
**Re-verified on WSL2 after `bc2227f744`** (nap-cap split: 2 ms when `g_use_interrupt_wait == false`, 200 µs otherwise). As expected, WSL behavior is unchanged since it always takes the 2 ms polling branch:

| | Idle CPU (30 s window) | 200× kernel + `hipDeviceSynchronize` |
|---|---|---|
| Stock ROCm 7.2.4 | 205% (~2 cores spinning) | ~18 ms |
| This branch | **0.76%** | ~18 ms (unchanged, within noise) |

Test: minimal HIP program from ROCm/librocdxg#60 (touch the GPU once, then sleep), CPU measured from `/proc/<pid>/stat` over 30 s. Combined with the earlier native-Linux mixed-batch result (p99 2046→249 µs at 0.6% idle CPU), both sides of the `bc2227f744` split are now verified.

Testing note (not specific to this PR): ROCr on develop now requires `hsaKmtGetKernelQueueId` from librocdxg (added in `55c34d6ea3`), so develop builds segfault against the stock ROCm 7.2.4 `librocdxg.so` on WSL. I built librocdxg from the same tree (`libhsakmt` with `WIN_SDK` pointing at the Windows SDK `shared;um` include dirs) and put both libs on `LD_LIBRARY_PATH` for the measurements above.

---
*This comment was drafted by Claude Code on my behalf; the measurements are from its runs on my WSL2 machine, reviewed by me before posting.*

cfreeamd (commented):

Re-reviewed the changes since my last pass (commit bc2227f744, 2026-08-05, plus @atgutier's request). All resolved: the original signed-int overflow in `NextPollNapUs` is now definitively closed (the `current_us > ceiling_us/2 ? ceiling : min(current_us*2, ceiling)` guard means the multiply only runs when `current_us <= ceiling/2`, so no overflow for any ceiling up to INT_MAX); the constants are now `inline constexpr` as requested; and the new mixed-batch ceiling (`kPollNapCeilingMixedUs = 200`, selected when `g_use_interrupt_wait` is true) correctly bounds the added latency to interrupt-backed handlers sharing the async-events thread while keeping the 2 ms ceiling for the all-polling WSL case. Tests extended to cover both ceilings and the reset-to-floor behavior. No new concerns.

seanthegeek:
@cfreeamd @dayatsin-amd @kentrussell @atgutier @fcui-amd — asking for help getting this over the line, because the process itself has become the blocker.

**Current state:** approved (July 8), conflict-free against `develop` as of today, and re-verified on WSL2 after the latest fix (`bc2227f744`). Everything remaining requires maintainer permissions:

- Six workflow runs (including `ROCR Runtime WSL` and `TheRock CI`) are sitting at **"action required"** — fork PRs need a maintainer to click *Approve and run* after every push.
- `therock-pr-bot` shows red only because it times out after 15 minutes waiting for those unapproved checks; it needs a re-run once they finish.
- The merge itself is maintainer-only.

**The loop we're stuck in:** `runtime.cpp` averages a commit on `develop` every ~4 days. Each cycle so far has gone: I resolve the conflict → days pass before CI is approved → `develop` moves → new conflict → repeat. This has now happened three times since approval; the PR has been mergeable after every resolution, just never merged before going stale again.

Three ways to break the cycle, any one works:

1. **Approve the CI runs and merge the PR once checks pass** — the whole thing done in one sitting closes the window where `develop` can drift.
2. **Enable auto-merge** at the same time as approving the CI runs — same effect without having to watch the checks; the PR merges the moment they go green (just the `therock-pr-bot` re-run noted above).
3. **"Allow edits by maintainers" is enabled** on this PR, so if another conflict does appear, any of you can push the resolution to my branch and merge in the same sitting — no round-trip through my timezone.

Happy to keep resolving conflicts as needed; I just can't do any of the merge-side steps from an external fork.

---

🤖 Drafted with [Claude Code](https://claude.com/claude-code) and reviewed by the author.

fcui-amd:
@seanthegeek The branch is not conflict-free against current `develop`. `runtime.cpp` still conflicts with the latest AsyncEventsLoop / event-age changes.

I already resolved it and pushed to [`users/fcui/fix-conflict`](https://github.com/ROCm/rocm-systems/tree/users/fcui/fix-conflict) (`00884b043b`, 6 commits on current `develop`). Please update your branch from that branch (or rebase onto `develop` the same way) and force-push. This PR cannot merge until the head is conflict-free.

seanthegeek:
@fcui-amd Done — the PR head is now exactly your `00884b043b` from `users/fcui/fix-conflict`, so it should match what you verified. Thanks for pushing the resolution.

For what it's worth, `git merge-tree` against current `develop` showed the previous head (`c796654`) merging cleanly too, and GitHub reported the PR as mergeable — but either way, the branch is now rebased and linear on top of `develop`.

The CI workflow runs will need *Approve and run* again after this push — if you can approve them and merge (or enable auto-merge) in the same pass, this should finally land.

---

🤖 Drafted with [Claude Code](https://claude.com/claude-code) and reviewed by the author.

fcui-amd:
@seanthegeek GitHub is currently unable to load the merge status for me, as shown in the screenshot. This may be a temporary GitHub issue, so I’ll check again later. In the meantime, another maintainer may be able to help with the remaining steps.
image

seanthegeek:
@fcui-amd The error in your screenshot was a GitHub incident that began at 13:40 UTC today ([status page](https://stspg.io/y1fl26l6wpzr)) — your attempt at 14:20 was in the middle of it. As of this writing GitHub is still reporting residual impact on Git Operations and Pull Requests, so merge status may not load reliably for a little while yet.

On the branch itself: I've since merged the latest `develop` into the PR, so the head is currently **0 commits behind `develop`** and ready to go. Once the incident fully clears, the remaining steps are the same maintainer-only pair as before: *Approve and run* on the workflow runs (the new push reset them again), then merge or enable auto-merge.

PR #3703, opened 2026-03-03
[rocprofiler-systems] AMD SMI collector using policy-based design

adjordje-amd:
## Motivation

The current AMD SMI implementation faces several challenges related to its maintainability and architecture, which this refactor aims to address:

Monolithic Design: The original amd_smi.cpp file, comprising 1354 lines, merges sampling, caching, and output logic. This tight coupling makes it challenging to modify or extend individual components.

Lack of Testability: Due to the integration with AMD SMI hardware APIs, the implementation is difficult to unit test and necessitates real hardware for verification.

Limited Flexibility: Introducing support for new output formats, such as RocPD alongside Perfetto, involves substantial code duplication.

Accumulated Bugs: The implementation contains several persistent bugs

The refactoring process aims to tackle these issues by replacing the monolithic setup with a contemporary, policy-based design. This new approach emphasizes modularity, testability, and enhanced performance

## Technical Details
New architecture
image

## JIRA ID

- AIPROFSYST-13
- AIPROFSYST-28
- AIPROFSYST-29

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #7310, opened 2026-06-16
feat(rocprofiler-systems): Introduce device backend for amdsmi  

mradosav-amd:
## Motivation

The PMC backend for AMD SMI had `amdsmi_*` types leaking through every layer of the stack — the `backend.hpp` session, the `device_backend`, the provider, and down into the collector traits. This made the backend untestable without real AMD hardware: every unit test had to link against `amdsmi` and hope it initialised. The goal of this PR is to draw a strict seam between the AMD SMI C API and the rest of the PMC stack, enabling full mock-based unit testing of every layer.

## Technical Details

**New `amdsmi_backend` struct (`backends/amd_smi/amdsmi_backend.hpp`)** — a 1:1 thin wrapper around the AMD SMI C API. Every method maps directly to one `amdsmi_*` call and returns the raw status code. Public type aliases (`processor_handle`, `status_t`, `gpu_metrics_t`, etc.) hide `amdsmi_*` names from all callers above this file. This is the single place in the codebase where `amdsmi.h` types cross into our own names.

**`amdsmi_backend_policy` concept (`backend.hpp`)** — the `backend<T>` session template now gates on this C++20 concept, which enumerates every expression the template evaluates on its policy type. Mismatches are caught at the template boundary with a clear error rather than deep inside a nested instantiation. The concept covers lifecycle, enumeration, per-device GPU queries, and conditionally SDMA and NIC calls based on feature-flag macros.

**`backend_session_contract` concept (`device_backend.hpp`)** — the `device_backend<Backend>` per-device proxy gates on a parallel concept that covers the session interface it actually calls. This ensures a mock session can be injected into `device_backend` in tests without involving the provider or AMD SMI at all.

**Provider refactoring (`device_providers/amd_smi/provider.hpp`)** — the provider was thinned by ~200 lines. It now only uses `backend<amdsmi_backend>` as a factory type parameter; `amdsmi_*` names no longer appear in provider code.

**`mock_amdsmi_backend` rename** — `mock_backend` in test files was renamed `mock_amdsmi_backend` to make its role in the seam explicit.

**New concept on provider** — a `provider_concept` was added to let the collector traits check provider interface compliance at compile time.

**Unit tests** — two new test suites and one expanded suite cover the full stack without AMD hardware:
- `test_backend.cpp` — tests `backend<mock_amdsmi_backend>` lifecycle, enumeration, error propagation, GPU/NIC/SDMA query paths.
[...]

PR #7903, opened 2026-06-26
fix(rocprofiler-systems): initialize GPU and NIC when initializing AMD SMI

ajanicijamd:
## Motivation

AI NIC device enumeration was silently failing when rocprofiler-systems was built from an
official package. The root cause is a regression introduced during the PMC policy-based
design refactor (PR #3703, March 2026): the original `amd_smi::setup()` called
`gpu::initialize_amdsmi()` which passed both `AMDSMI_INIT_AMD_GPUS` and
`AMDSMI_INIT_AMD_NICS` to `amdsmi_init()`. The refactored PMC provider called
`backend::init()` with a default argument of `AMDSMI_INIT_AMD_GPUS` only, dropping the
NIC flag.

Since `amdsmi_init()` uses a ref-count guard — the first call initializes devices and
subsequent calls return success without re-initializing — NIC devices were never populated
regardless of whether a second call with the NIC flag was made later.

## Technical Details

- **`source/lib/backends/amd_smi/backend.hpp`**: Removed the default argument from
  `backend::init()`. Callers must now explicitly pass initialization flags, turning future
  omissions into compile-time errors rather than silent runtime failures.

- **`source/lib/rocprof-sys/library/pmc/device_providers/amd_smi/provider.hpp`**: The
  `provider` constructor now includes `backends/amd_smi/ainic_feature.hpp` and
  conditionally ORs `AMDSMI_INIT_AMD_NICS` into the flags when `AINIC_SUPPORTED` is
  defined (AMD SMI ≥ 26.3 and `ROCPROFSYS_USE_AINIC` enabled), mirroring the existing
  logic in `core/gpu.cpp`.

## JIRA ID

* ROCM-25722

## Test Plan

* Build with `ROCPROFSYS_USE_AINIC=ON` against AMD SMI ≥ 26.3 and verify
  `ROCPROFSYS_BUILD_AINIC=1` is set
* Run rocprofiler-systems against a simulated NIC sysfs and confirm AI NIC devices are
  enumerated
* Verify GPU enumeration is unaffected (no regression)
* Confirm the build fails to compile if any caller of `backend::init()` omits the flags
  argument

## Test Result

Tested against a simulated NIC sysfs on a development system with AMD SMI 26.5. AI NIC
devices are now correctly enumerated. GPU monitoring is unaffected.

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-systems/source/lib/backends/amd_smi/backend.hpp
+++ b/projects/rocprofiler-systems/source/lib/backends/amd_smi/backend.hpp
@@ -25,10 +25,10 @@ struct backend
 {
     /**
      * @brief Initialize the AMD SMI library.
-     * @param init_flags Initialization flags (default: AMDSMI_INIT_AMD_GPUS).
+     * @param init_flags Initialization flags (e.g. AMDSMI_INIT_AMD_GPUS, AMDSMI_INIT_AMD_NICS).
      * @return AMD SMI status code indicating success or failure.
      */
-    static amdsmi_status_t init(std::uint64_t init_flags = AMDSMI_INIT_AMD_GPUS)
+    static amdsmi_status_t init(std::uint64_t init_flags)
     {
         return amdsmi_init(init_flags);
     }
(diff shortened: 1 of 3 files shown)


mradosav-amd:
@ajanicijamd Please take a look at https://github.com/ROCm/rocm-systems/pull/7310. That PR already cover change from this PR.

ajanicijamd:
PR #7310 has been merged, so this PR is closed as not needed any more.

PR #7904, opened 2026-06-26
fix(rocjitsu): avoid stale race-detector event ids

### LINKED ISSUE #7946: [rocjitsu] race detector fails to retire even on barrier
A bug was detected while running rocjitsu with the race detector plugin on a workload. The issue is related to the periodical removal of fully retired events. This issue tracks the bug.

newling:
Fixes [#7946](https://github.com/ROCm/rocm-systems/issues/7946)

## Summary

Fix a race-detector lifecycle bug where wave-local memory events could remain queued for later barrier retirement after `EventRegistry` had already trimmed them.

`EventRegistry` intentionally treats `WAVE_COMPLETE` non-LDS events as trimmable, because they are wave-local and do not need a workgroup barrier. However, `WaveRaceState::resolveWaitCnt()` was previously pushing every completed event into the barrier-retirement queue. After trimming, that queue could contain stale `EventId`s; a later barrier flush could then retire invalid ids, causing heap corruption and corrupting race-detector bookkeeping.

This change renames the queue to `barrierPendingLdsEvents` and only stores LDS-involved events there. LDS events still remain live after `s_waitcnt` until `s_barrier`, which preserves cross-wave race detection semantics.

The regression test is in `race_detector_event_registry_test.cpp`
## Testing

```bash
cmake --build /home/jnewling/workspace/builds/rocm-systems-augustus
/home/jnewling/workspace/builds/rocm-systems-augustus/tests/rocjitsu_tests --gtest_filter='RaceDetector.EventRegistry_TrimmedWaveLocalEventsAreNotBarrierQueued:RaceDetector.*'
ctest --test-dir /home/jnewling/workspace/builds/rocm-systems-augustus -R '^RaceTest\.' --output-on-failure
git diff --check
```

Also ran the rank050 PyTorch/rocjitsu emulator repro in Docker after the fix; it completed with `inner_exit=0`, no reported races, and no NaN/Inf output.


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/race_detector/core/wave_race_state.cpp
```diff
@@ -141,7 +141,16 @@ template <typename Pred> void WaveRaceState::resolveWaitCnt(int limit, Pred isTa
...
-      waveCompleteMemoryEvents.push_back(eid);
+      // Non-LDS events are wave-local: once s_waitcnt completes them, the
+      // owning wave can safely use their results and no workgroup barrier is
+      // needed. EventRegistry is therefore allowed to trim those WAVE_COMPLETE
+      // events immediately. LDS events are different: a completed LDS access
+      // can still conflict with another wave until the workgroup reaches
+      // s_barrier, so only LDS-involved events belong in this barrier-retire
+      // queue. Queuing wave-local events here would leave stale EventIds after
+      // registry trimming, and a later barrier would try to retire invalid ids.
+      if (isLdsInvolved(detector->events().type(eid)))
```
accauble: Nit: I read this PR as "only queue events that won't be trimmed". If you ever change what is trimmable, you would need to update this too, so I'm wondering if it makes sense to queue if `!isTrimmable` instead of specifying `isLds`?

emulation/rocjitsu/tests/race-detector/race_test_builder.h
```diff
@@ -114,7 +114,7 @@ class RaceTestBuilder {
   /// Barrier: flush all waves' WAVE_COMPLETE events (simulates s_barrier).
```
accauble: Nit: Comment says it flushes WAVE_COMPLETE events but now its just LDS events

accauble (approved):
LGTM. Just a couple nits to look over. Oh and the issue # in the description is wrong ;-)

PR #7905, opened 2026-06-26
feat(rocr): add per-queue error callback to AIE queues

### LINKED ISSUE #7908: AIE queues drop the hsa_queue_create error callback
### Problem
AIE queues created via `hsa_queue_create` silently drop the error callback.
`AMD::AieAgent::QueueCreate` receives the `event_callback`/`data` arguments but
never forwards them to `AieAqlQueue`, and `AieAqlQueue` has no error-reporting
path. As a result, applications that register an error callback on an AIE queue
never receive command-submission failures (unlike GPU queues via `AqlQueue`).

### Expected
An AIE queue should report submission/command failures through the per-queue
error callback registered at `hsa_queue_create`, mirroring the GPU `AqlQueue`
behavior (`Suspend` + `errors_callback_`).

### Notes
This is a prerequisite for asynchronous AIE command execution, which needs a way
to surface command errors that occur off the dispatch thread.

aamarnat:
Closes #7908

## Summary
AIE queues created via `hsa_queue_create` silently dropped the error callback:
`AieAgent::QueueCreate` received `event_callback`/`data` but never passed them to
`AieAqlQueue`, which had no error-reporting path. This wires up a per-queue error
callback, mirroring the GPU `AqlQueue`.

## Changes
- `AieAgent::QueueCreate` now forwards `callback`/`data` to the `AieAqlQueue` ctor.
- `AieAqlQueue` stores `errors_callback_`/`errors_data_` and adds `InvokeErrorCallback()`
  (same shape and guards as `AqlQueue`).
- On a `SubmitCmdChain` failure, `SubmitPackets` now suspends the queue and invokes the
  callback instead of throwing, leaving the read index unadvanced so the failed packets
  are not marked consumed.
- Adds a self-contained rocrtst test (`suites/aie/error_callback_test.cc`) that registers a
  callback and asserts it fires when a packet with an unregistered buffer is rejected.

## Notes
- Mirrors the GPU `AqlQueue` error path (`Suspend` + `errors_callback_`); constructor
  signature, member types, and `InvokeErrorCallback` match `AqlQueue`.
- The test is standalone and needs no AIE kernel artifacts (no aiecc/NPU toolchain). Its
  CMake will be reconciled with the broader AIE rocrtst suite when that lands on `develop`.


ypapadop-amd (approved):

projects/rocr-runtime/rocrtst/suites/aie/error_callback_test.cc
```diff
@@ -0,0 +1,142 @@
...
+
+// Creating a queue with an error callback must succeed, and the callback must not fire while no
+// error has occurred.
+TEST(ErrorCallback, QueueCreateWithCallback) {
+  ASSERT_EQ(hsa_init(), HSA_STATUS_SUCCESS);
+
+  bool found = false;
+  const hsa_agent_t agent = first_aie_agent(found);
+  ASSERT_TRUE(found);
+
```
cfreeamd, line 85: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** Tests fail (not skip) on machines without AIE hardware.

`ASSERT_TRUE(found)` when `found` is false registers as a test FAILURE, not a skip, so every CI machine without an NPU will show these tests as FAILED. Replace with:
```cpp
if (!found) GTEST_SKIP() << "No AIE device found; skipping test";
```
`GTEST_SKIP()` is available since GTest 1.10.0, which predates the pinned FetchContent commit.

cfreeamd, line 85: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** `hsa_init` is called before the agent-discovery check and is leaked if the test exits early (or skips). `hsa_shut_down()` at the bottom is never reached in that path, leaving the HSA runtime initialized for the rest of the process lifetime. Move `hsa_init` after the agent-discovery/skip check, or use a GTest fixture with `SetUpTestSuite`/`TearDownTestSuite` shared across both tests.

projects/rocr-runtime/runtime/hsa-runtime/core/inc/amd_aie_aql_queue.h
```diff
@@ -118,6 +126,13 @@ class AieAqlQueue : public core::Queue,
   void* kmq_metadata_ = nullptr;
   /// @brief Indicates if queue is active.
   std::atomic<bool> active_ = false;
+  /// @brief Set when the queue is suspended after an error; no further packets are submitted.
+  /// Only accessed on the packet-submission path, so it does not need to be atomic.
+  bool suspended_;
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** `bool suspended_;` has no in-class default initializer. `std::atomic<bool> active_ = false;` above it carries one; `suspended_` does not. The initializer-list entry `suspended_(false)` is correct for this constructor, but any future constructor that omits it would leave `suspended_` indeterminate. Add:
```cpp
bool suspended_ = false;
```
at the declaration site for safety and consistency with the surrounding members.

cfreeamd (commented):
Projects/rocr-runtime files reviewed. Inline comments cover the issues found. `amd_aie_aql_queue.cpp`, `amd_aie_agent.cpp`, and `CMakeLists.txt` look structurally correct.

aamarnat:
Thanks for the review @cfreeamd — all three comments addressed in dabcd8b. Summary and on-hardware verification below.

**1. Tests fail (not skip) on machines without AIE hardware** ✅
Both tests are now a `TEST_F(ErrorCallback, …)` fixture whose `SetUp()` resolves the AIE agent and calls `GTEST_SKIP()` when none is present.

**2. `hsa_init` leaked on early exit/skip** ✅
Went with the fixture approach you suggested rather than the "move `hsa_init` after the discovery check" wording — the latter isn't viable because `hsa_iterate_agents` requires the runtime to already be initialized. Instead, `SetUpTestSuite`/`TearDownTestSuite` init once and **unconditionally** shut down, so there's no leak on the pass *or* skip path.

**3. `bool suspended_;` has no in-class default initializer** ✅
Now `bool suspended_ = false;` at the declaration site.

### Verification (on an NPU box)

Normal run — hardware present:
```
[ RUN      ] ErrorCallback.QueueCreateWithCallback
[       OK ] ErrorCallback.QueueCreateWithCallback (0 ms)
[ RUN      ] ErrorCallback.InvalidDispatchInvokesCallback
AIE queue: command submission failed (0x1003)
[       OK ] ErrorCallback.InvalidDispatchInvokesCallback (0 ms)
[  PASSED  ] 2 tests.
```
(`0x1003` = `HSA_STATUS_ERROR_INVALID_PACKET_FORMAT` — the expected error path: `suspended_` set, callback invoked with the right status/source.)

Forced no-agent path (temporary local override, reverted before commit) — confirms the skip branch and that CI stays green:
```
[  SKIPPED ] ErrorCallback.QueueCreateWithCallback
[  SKIPPED ] ErrorCallback.InvalidDispatchInvokesCallback
[  PASSED  ] 0 tests.  [  SKIPPED ] 2 tests
# process exit code: 0
```
Global teardown ran on the skip path too, so `hsa_shut_down` fires and the runtime isn't leaked.

Compiled clean under `-Wall -Wextra -Wpedantic`.

cfreeamd (approved):

dayatsin-amd (approved):

aamarnat:
Added NCSA copyright headers to the new files in this PR:

- `projects/rocr-runtime/rocrtst/suites/aie/CMakeLists.txt`
- `projects/rocr-runtime/rocrtst/suites/aie/error_callback_test.cc`

aamarnat:
Added SPDX license headers to the new files in this PR:

- `projects/rocr-runtime/rocrtst/suites/aie/CMakeLists.txt` (SPDX/MIT)
- `projects/rocr-runtime/rocrtst/suites/aie/error_callback_test.cc` (already had SPDX/MIT)

PR #7906, opened 2026-06-26
fix(rocjitsu): improve kernel name reporting

### LINKED ISSUE #7948: [rocjitsu] Improved kernel name extraction for debugging
## Problem

rocjitsu has code for extracting a kernel symbol from an in-memory AMDHSA code object, but kernel names are not always surfaced clearly in dispatch metadata and logs.

The current helper is `find_kernel_symbol()`:

https://github.com/ROCm/rocm-systems/blob/56c64d67f4f569049fe0e0f997f854264cac133b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/kernel_symbol.h#L11-L24

The implementation first attempts to resolve the kernel descriptor through ELF dynamic symbols:

https://github.com/ROCm/rocm-systems/blob/56c64d67f4f569049fe0e0f997f854264cac133b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/kernel_symbol.cpp#L40-L97

It then falls back to scanning AMDGPU metadata notes for kernel descriptor symbol strings:

https://github.com/ROCm/rocm-systems/blob/56c64d67f4f569049fe0e0f997f854264cac133b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/kernel_symbol.cpp#L99-L135

The dispatch path currently attempts to resolve a kernel name while processing an AQL kernel dispatch packet:

https://github.com/ROCm/rocm-systems/blob/56c64d67f4f569049fe0e0f997f854264cac133b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp#L1008-L1038

## Desired Outcome

Improve kernel name extraction and reporting so dispatch metadata can expose a useful kernel name, with a clear fallback when no name is available.

Add focused tests covering the symbol extraction and dispatch metadata paths.

newling:
Fixes #7948

## Summary

Improve kernel-name reporting for rocjitsu dispatches and race reports.

The main correctness fix is for KFD/daemon VMID-mapped dispatches: the AQL
`kernel_object` field is a GPU virtual address, not a directly readable host
pointer. Kernel-symbol lookup now resolves that GPU VA through the process page
table, finds the daemon-accessible host range for the loaded code object, and
then scans ELF metadata from the translated host pointer.

The reporting change separates two concepts:

- `kernel_symbol`: the exact ELF symbol, useful for identity/debugging.
- `kernel_name`: a compact readable display name, useful in logs and structured
  race reports.

Race reports now include `kernel=...`, `symbol=...`, and `dispatch=...` fields,
so a race can be tied back to the dispatch and exact kernel symbol while keeping
the common log header readable.

## Changes

- Add VMID-aware host-range lookup in `GpuMemory`.
- Translate VMID-mapped `pkt.kernel_object` before ELF symbol lookup.
- Add exact `kernel_symbol` dispatch metadata alongside readable `kernel_name`.
- Demangle C++/HIP symbols for display, with fallback to the original symbol.
- Include kernel name, exact symbol, and dispatch id in race report headers.
- Add focused coverage for VMID host ranges, VMID-mapped dispatch metadata,
  demangling/display names, unresolved-name fallback, and race-log parsing.

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/kernel_symbol.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/kernel_symbol.cpp
@@ -5,6 +5,9 @@
 
 #include "rocjitsu/code/amdgpu_elf.h"
 
+#include <cctype>
+#include <cstdlib>
+#include <cxxabi.h>
 #include <string_view>
 
 namespace rocjitsu {
@@ -133,6 +136,33 @@ std::string find_kernel_symbol(const uint8_t *kernel_object_ptr, const uint8_t *
       return best_name;
   }
   return {};
+}
+
+std::string demangle_kernel_symbol(std::string_view symbol) {
+  if (symbol.empty())
+    return {};
+
+  std::string symbol_string(symbol);
+  int status = 0;
+  char *demangled = abi::__cxa_demangle(symbol_string.c_str(), nullptr, nullptr, &status);
+  if (status != 0 || demangled == nullptr)
+    return symbol_string;
+
+  std::string result(demangled);
+  std::free(demangled);
+  return result;
+}
+
+std::string kernel_display_name(std::string_view symbol) {
+  std::string result = demangle_kernel_symbol(symbol);
+  size_t parenthesis_position = result.find('(');
+  if (parenthesis_position != std::string::npos)
+    result.resize(parenthesis_position);
+  if (result.starts_with("void "))
+    result.erase(0, 5);
+
+  std::erase_if(result, [](unsigned char c) { return std::isspace(c); });
+  return result;
 }
 
 } // namespace rocjitsu
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/kernel_symbol.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/kernel_symbol.cpp
@@ -20,6 +20,38 @@ bool in_range(const uint8_t *ptr, uint64_t size, const uint8_t *base, uint64_t r
          static_cast<uint64_t>(ptr - base) <= range_size - size;
 }
 
+size_t find_outer_argument_list(std::string_view text) {
+  size_t argument_list = std::string_view::npos;
+  int angle_depth = 0;
+  int parenthesis_depth = 0;
+
+  for (size_t pos = 0; pos < text.size(); ++pos) {
+    switch (text[pos]) {
+    case '<':
+      if (parenthesis_depth == 0)
+        ++angle_depth;
+      break;
+    case '>':
+      if (angle_depth > 0 && parenthesis_depth == 0)
+        --angle_depth;
+      break;
+    case '(':
+      if (angle_depth == 0 && parenthesis_depth == 0)
+        argument_list = pos;
+      ++parenthesis_depth;
+      break;
+    case ')':
+      if (parenthesis_depth > 0)
+        --parenthesis_depth;
+      break;
+    default:
+      break;
+    }
+  }
+
+  return argument_list;
+}
+
 } // namespace
 
 std::string find_kernel_symbol(const uint8_t *kernel_object_ptr, const uint8_t *elf_base,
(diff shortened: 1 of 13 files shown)


emulation/rocjitsu/tests/execution_plugin_test.cpp
```diff
@@ -462,12 +471,80 @@ struct PluginFixture {
...
+std::vector<uint8_t> make_loaded_kernel_symbol_elf(uint64_t kernel_descriptor_offset,
+                                                   std::string_view symbol_name);
+
 TEST(ExecutionPluginTest, NoPluginNoCrash) {
   PluginFixture f;
   const uint32_t code[] = {S_NOP, S_ENDPGM};
   f.run_kernel(code, 2);
 }
 
+TEST(ExecutionPluginTest, DispatchPacketNameResolvesForVmidMappedCodeObject) {
```
accauble, line 876: It seems that this test is [failing with asan](https://github.com/ROCm/rocm-systems/actions/runs/28477308569/job/84405032887?pr=7906#step:8:2185).

newling: This seems to be resolved now

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/kernel_symbol.cpp
```diff
@@ -135,4 +147,33 @@ std::string find_kernel_symbol(const uint8_t *kernel_object_ptr, const uint8_t *
...
+      abi::__cxa_demangle(symbol_string.c_str(), nullptr, nullptr, &status), &std::free);
+  if (status != 0 || demangled == nullptr)
+    return symbol_string;
+
+  return std::string(demangled.get());
+}
+
+std::string kernel_display_name(std::string_view symbol) {
+  std::string result = demangle_kernel_symbol(symbol);
+  size_t parenthesis_position = result.find('(');
```
kuhar: We should identify the outer function-argument list with nesting awareness before trimming the demangled spelling. Cutting at the first `(` makes anonymous-namespace kernels resolve to an empty display name (`(anonymous namespace)::kernel(float*)`) and truncates valid template arguments (`void foo<void (*)(int)>(...)` becomes `foo<void`). Please cover both shapes with unit tests.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/kernel_symbol.cpp
```diff
@@ -73,21 +78,28 @@ std::string find_kernel_symbol(const uint8_t *kernel_object_ptr, const uint8_t *
...
+      auto *sym = reinterpret_cast<const Elf64_Sym *>(elf_base + symtab_off + uint64_t{s} * syment);
+      if (sym->st_value != kd_offset)
+        continue;
+      if (sym->st_name >= strsz)
         continue;
-      if (syms[s].st_name >= strsz)
+      const char *symbol_name = strtab + sym->st_name;
+      size_t max_name_size = strsz - sym->st_name;
+      size_t name_size = strnlen(symbol_name, max_name_size);
+      if (name_size == max_name_size)
```
kuhar, line 132: I think we should add a regression test for a symbol whose name reaches the end of `.dynstr` without a NUL terminator. This new `strnlen` guard is what prevents the old out-of-bounds scan, but the fixture always supplies a trailing NUL, so the safety branch is not exercised. Setting `DT_STRSZ` to the leading NUL plus the symbol bytes and checking that `find_kernel_symbol` returns empty would cover it.

kuhar (commented):
It would be good to update `docs/race-detector.md` with the new report header. The documented example still starts `RACE type=...`, while this change emits and tests `RACE kernel=... symbol=... dispatch=... type=...`; please also document the `?` fallback for unresolved names.

newling:
> It would be good to update docs/race-detector.md with the new report header. The documented example still starts RACE type=..., while this change emits and tests RACE kernel=... symbol=... dispatch=... type=...; please also document the ? fallback for unresolved names.

That's a very thorough review! Maybe it's time for me to update my agent...

please take another look, I believe all comments are addressed.

newling:
@kuhar Can I get another review of this, please?

accauble (approved):
Looks like you appropriately addressed @kuhar's comments. I only have nit a which you can take or leave: the default unknown string, "?" is commented and defined in a few places. Could be worth having a single definition.

PR #7907, opened 2026-06-26
fix(rocjitsu): use absolute-PC keys in race disassembly cache

### LINKED ISSUE #7949: rocjitsu: make disassembly cache robust to non-monotonic PCs
## Problem

rocjitsu caches instruction disassembly so reports can include assembly text around interesting PCs.

The current race detector disassembly cache is shared per dispatch:

https://github.com/ROCm/rocm-systems/blob/56c64d67f4f569049fe0e0f997f854264cac133b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/race_detector/plugin.h#L56-L63

It currently indexes entries with `(pc - base) / 4`, where `base` is the first PC recorded for that dispatch:

https://github.com/ROCm/rocm-systems/blob/56c64d67f4f569049fe0e0f997f854264cac133b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/race_detector/plugin.h#L65-L81

This assumes later PCs are near the first recorded PC and not below it. That assumption is not always valid.

## Desired Outcome

Make the cache robust to PCs that arrive in non-monotonic order or are far from the first PC recorded for a dispatch.

Add a focused regression test for non-monotonic PCs.

newling:
Fixes #7949

## Motivation

`RaceDetectorPlugin::DisasmCache` assumed instruction PCs for a dispatch formed one compact range starting at the first PC observed. Some dispatches can execute code at lower or far-away PCs after the first recorded instruction. In that case, indexing by `(pc - base) / 4` can underflow or request an enormous vector resize.

## Changes

- Store disassembly by absolute PC in an `unordered_map`.
- Remove the compact-range vector indexing assumption from `DisasmCache`.
- Keep trace formatting behavior unchanged.

## Notes

The original vector representation was chosen for speed on compact PC ranges. This change prioritizes correctness for non-monotonic and far-apart PCs. We may want to revisit the performance tradeoff if disassembly caching shows up in profiles.

## Tests

- Add `DisasmCacheTest.HandlesNonMonotonicPcOrder`.

## Testing

- `pre-commit` on touched files
- Focused `ctest` including `DisasmCacheTest` and race detector tests


amd-eochoalo (approved):

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/race_detector/plugin.h
```diff
@@ -62,40 +61,25 @@ std::string formatTrace(const RingBuffer<uint64_t, 256> &trace,
...
+/// compact, monotonic text range; helper/trampoline code can be far away from
+/// the first PC observed for the dispatch.
 struct DisasmCache {
-  void record(uint64_t pc, const Instruction &inst) {
-    size_t idx = pc_to_idx(pc);
-    if (idx < size_.load(std::memory_order_acquire) && !entries_[idx].empty())
-      return;
+  void record(uint64_t pc, const Instruction &inst) { record(pc, inst.disassemble()); }
+
+  void record(uint64_t pc, std::string disasm) {
```
kuhar, line 70: nit: pass `const std::string &` / `std::string_view` to avoid copies?

kuhar (approved):

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/race_detector/plugin.h
newling: My understanding here is that the string is not copied inside the function, it is moved to its final location in the map. If I replaced this with a string_view or a const-ref, we'd need to create a new string. Ideally the map could store a string_view where the underlying memory allocation is still in the .co (or somewhere), but I'm not sure that's safe and don't want to optimize for that until we know it'll help. 

So I'm going to leave this as is...

newling: FWIW I don't know why my ai agent uses std::string function arguments as liberally as it does, 'before ai' const-ref was my default choice. But I do think that in this case const-ref wouldn't be more efficient

PR #5726, opened 2026-05-02
[ROCM-3941] PyPi Wheel Gen

amd-juwillia:
## Motivation

`amd-smi` previously shipped only as a system `.deb`/`.rpm`, and its postinst pivoted through `pip`/`.pth` to register the Python module. Two problems:

- Mixing packaging ecosystems — a system `.deb`/`.rpm` invoking `pip install`/`pip uninstall` and dropping an `amdsmi.pth` — was fragile and drew reviewer pushback.
- The loader carried a ~200-line ROCm-path-walking resolver (`ROCM_HOME`/`ROCM_PATH`/`LD_LIBRARY_PATH` probing) that was hard to reason about.

This PR drops all `pip`/`.pth` plumbing from the system package and installs the amdsmi module **directly** into the target interpreter's `dist-packages`/`site-packages` at package-build time, paired with a flat 3-case import-time loader and a separate SONAME-isolated `libamd_smi_python.so` so the manylinux wheel (tooling already landed in develop via **#5725**) can coexist with the system package without double-load.

> **PR split:** #5725 (merged into develop) added the wheel build tooling (`tools/build_wheel.py`) and the `amdsmi-manylinux-build.yml` CI. **This PR (#5726)** is the consuming side: the loader rewrite + dist-packages install pivot, sitting 1 commit on top of develop.

### 3-case loader (`py-interface/amdsmi_wrapper.py`)

The auto-generated wrapper resolves the `.so` at import time via a ~30-line ladder (replacing the old dual-context resolver):

1. `AMDSMI_LIB_OVERRIDE` env var — short-circuits everything (dev / ABI tests).
2. `libamd_smi_python.so` next to the wrapper — pip-wheel context.
3. `ctypes.CDLL("libamd_smi.so.26")` — system context, resolved by the dynamic linker via the `ld.so.conf.d` entry installed by postinst.

If all three fail, a `_MissingLibrary` sentinel keeps import tolerant (doc/lint/multi-stage builds). `tools/generator.py` is updated in lockstep.

[...]

PR #7912, opened 2026-06-28
feat(amdsmi): [ROCM-3941] Add SONAME conflict test and build harness

amd-juwillia:
## Motivation

Guard the SONAME-distinctness invariant that lets the system and wheel libraries coexist without double-loading. (The loader/ABI-parity test ships with the loader change in #7914.)

## Technical Details

**Test** (`tests/run_amdsmi_pkg_conflict_test.py`):
- Assert the system `libamd_smi.so` and the wheel `libamd_smi_python.so` keep distinct SONAMEs so they never double-load in one process.

**Harness** (`tests/amdsmi_build/run_amdsmi_build.py`):
- Extend the local build/install/verify harness mirroring the CI matrix.

> Split 3 of 3 from #5726. **Depends on #7914 and #7913** and is stacked on them; all target `develop`. Collapses to 2 files once the parents merge. Merge order: #7914 → #7913 → this.

## JIRA ID

Resolves ROCM-3941

## Test Plan

- `python3 projects/amdsmi/tests/run_amdsmi_pkg_conflict_test.py`

## Test Result

- SONAME-distinctness check passes against a `-DBUILD_PYTHON_WHEEL=ON` build.

## Submission Checklist

### DIFF
--- a/projects/amdsmi/tests/amdsmi_build/run_amdsmi_build.py
+++ b/projects/amdsmi/tests/amdsmi_build/run_amdsmi_build.py
@@ -1186,7 +1186,12 @@ def summarize_results(results_dir: Path, os_label: str, summary_file: Optional[P
 
     # 4. Python test outputs
     fail_re = _re.compile(r"^(FAIL|ERROR):", _re.MULTILINE)
-    for test_file in ("integration_test_output.txt", "unit_test_output.txt", "perf_test_output.txt"):
+    for test_file in (
+        "integration_test_output.txt",
+        "unit_test_output.txt",
+        "perf_test_output.txt",
+        "abi_compat_output.txt",
+    ):
         full = results_dir / test_file
         if not full.exists():
             continue
--- a/projects/amdsmi/tests/run_amdsmi_pkg_conflict_test.py
+++ b/projects/amdsmi/tests/run_amdsmi_pkg_conflict_test.py
@@ -1,0 +1,128 @@
+#!/usr/bin/env python3
+#
+# Copyright (C) Advanced Micro Devices. All rights reserved.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy of
+# this software and associated documentation files (the "Software"), to deal in
+# the Software without restriction, including without limitation the rights to
+# use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+# the Software, and to permit persons to whom the Software is furnished to do so,
+# subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+# FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+# COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+# IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+# CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+"""
+run_amdsmi_pkg_conflict_test.py
+================================
+
+Sanity-check that the system-package shared object and the pip wheel's
+bundled shared object ship under DIFFERENT SONAMEs, so a process that
+somehow has both installed cannot accidentally double-load the same
+library and crash during static C++ initialisation.
+
+This is the only multi-package invariant we still validate. amd-smi
+follows an "either you install the rpm/deb OR you install the wheel"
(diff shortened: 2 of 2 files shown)


.github/workflows/amdsmi-build.yml
```diff
@@ -38,6 +38,62 @@ defaults:
...
+          PY_BUILD=""
+          for _py in cp313 cp312 cp311 cp310 cp39; do
+            _cand="/opt/python/${_py}-${_py}/bin/python3"
+            if [ -x "$_cand" ]; then PY_BUILD="$_cand"; break; fi
+          done
+          if [ -z "$PY_BUILD" ]; then
+            echo "ERROR: no suitable CPython interpreter found under /opt/python" >&2
+            exit 1
+          fi
+          echo "Using build interpreter: $PY_BUILD ($("$PY_BUILD" --version 2>&1))"
```
geomin12, line 77: let's split this up into separate github action steps. it should be obvious if there was a failure, it should be directly related to build in this step

this is just selecting python

.github/workflows/amdsmi-build.yml
```diff
@@ -38,6 +38,62 @@ defaults:
...
+
+          echo "Installing build prerequisites..."
+          dnf -y install git make gcc gcc-c++ cmake ninja-build openssl-devel libnl3-devel libmnl-devel
```
geomin12: separate step pls, this is installing build prereqs

.github/workflows/amdsmi-build.yml
```diff
@@ -38,6 +38,62 @@ defaults:
...
+          echo "Configuring and building AMDSMI..."
+          rm -rf "$BUILD_DIR"
+          mkdir -p "$BUILD_DIR"
+          cd "$BUILD_DIR"
+          cmake "$PROJECT_DIR" \
+            -DBUILD_TESTS=ON \
+            -DENABLE_ESMI_LIB=ON \
+            -DBUILD_PYTHON_WHEEL=ON \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DPython3_EXECUTABLE="$PY_BUILD"
```
geomin12, line 94: configure should be separate too

.github/workflows/amdsmi-build.yml
```diff
@@ -38,6 +38,62 @@ defaults:
...
+          make -j"$(nproc)"
+
+          echo "Building Python wheel..."
+          $PY_BUILD -m pip install --upgrade pip setuptools wheel
+          cd "$BUILD_DIR/py-interface/python_package"
+          rm -rf build dist *.whl *.egg-info
+          $PY_BUILD -m pip wheel --no-deps --no-build-isolation -w . .
+          echo "Python wheel built: $(ls *.whl)"
+          echo "Build gate passed"
```
geomin12: this should be separate as well

geomin12 (commented):
comments regards to yml. for a workflow github action step, it should be very clear on what the failure is based on what was run. we shouldn't have to parse through a lot of logs to figure out "oh it was an install error"

i would prefer amdsmi folks to review the amdsmi related code changes

dmitrii-galantsev:

Top of the ROCM-3941 chain: #7914 -> #7913 -> #7912 (this). Rebased onto the updated #7913; history rewritten and force-pushed, so please re-pull.

Fixes applied to `run_amdsmi_pkg_conflict_test.py`:

- Replaced the lexicographic `sorted(...)[0]` library selection with a numeric version key. Lexically `libamd_smi.so.24` sorts before `.so.9`, so the old code could pick the wrong major when several coexist; it now picks the highest numeric version.
- Robust `readelf` handling: guard with `shutil.which` and catch `CalledProcessError` instead of an opaque traceback with stderr discarded.
- Honor `ROCM_PATH` / `ROCM_HOME` for the system lib directory instead of hardcoding `/opt/rocm`.
- Dropped `from __future__ import annotations` (not needed; the style guide avoids it).

`_version_key` verified: with a decoy `.so.9` present it correctly selects `.so.26`.


---

**Update:** rebased on the follow-up #7914. The install-verify interpreter fix now lives in #7914 (the base layer) so every stacked PR's CI picks it up; this commit retains only its original `run_amdsmi_build.py` change (adding `abi_compat_output.txt` to the scanned results) plus the SONAME conflict-test fixes. CI green: Manylinux Gate, RHEL8, SLES and the RPM/Debian wheel jobs all pass.

dmitrii-galantsev (approved):

.github/workflows/amdsmi-build.yml
marifamd: fixed in 6491aa1ada6

.github/workflows/amdsmi-build.yml
marifamd: fixed in 6491aa1ada6

.github/workflows/amdsmi-build.yml
marifamd: fixed in 6491aa1ada6

.github/workflows/amdsmi-build.yml
marifamd: fixed in 6491aa1ada6

marifamd (approved):

geomin12 (approved):

marifamd (approved):

dmitrii-galantsev:
~~ci failing, latest rebase should have fixed it: https://github.com/ROCm/rocm-systems/pull/9061~~
see the same problem pop up in #7914

dmitrii-galantsev:
converted to draft because we're reverting #7914 and #7913

PR #7913, opened 2026-06-28
feat(amdsmi): [ROCM-3941] Add manylinux wheel job to the amdsmi CI matrix

amd-juwillia:
## Motivation

Wire the SONAME-isolated wheel build into amd-smi's own per-distro CI matrix so every change is checked against a manylinux wheel build.

## Technical Details

**CI** (`.github/workflows/amdsmi-build.yml`):
- Add a `manylinux-build` job that runs `tools/build_wheel.py` with `-DBUILD_PYTHON_WHEEL=ON` alongside the existing build-and-test matrix.

> Split 2 of 3 from #5726. **Depends on #7822** (the wheel build it exercises) and is stacked on it. Both target `develop`, so until #7822 merges this PR's diff also shows #7822's changes; it collapses to 2 files once #7822 lands. Merge order: #7822 → this.

## JIRA ID

Resolves ROCM-3941

## Test Plan

- Trigger the `manylinux-build` job on this branch and confirm a `.whl` is produced and import-smoke-tested.

## Test Result

- The manylinux job builds and installs the wheel on the ubuntu-22.04 manylinux_2_28 target.

## Submission Checklist


geomin12 (commented):
seems this is same as #7914 and #7912 , same comments apply here

but amdsmi code should be reviewed by amdsmi folks

dmitrii-galantsev:

Part of the ROCM-3941 chain: #7914 -> #7913 (this) -> #7912. Rebased onto the updated #7914; history rewritten and force-pushed, so please re-pull.

Fixes applied to the manylinux job:

- `runs-on: ubuntu-latest` -> `ubuntu-24.04` (style guide requires pinned runners).
- Removed the redundant manual `pip wheel` invocation. The CMake `python_wheel` target is part of `ALL` (gated on `BUILD_PYTHON_WHEEL`), so the build already produces the wheel via its proper path (which runs `disable_system_fallback.py`). The job now just builds and asserts the `.whl` exists, instead of hand-rolling a second wheel that bypassed that step.
- Split the single `run:` block into discrete steps (select interpreter, install prereqs, configure, build, verify) so a failure points at the right stage, per review feedback.
- Dropped `-DBUILD_TESTS=ON` (the gate never runs the tests); consistent quoting.
- CHANGELOG wording: "workflow" -> "job".


---

**Update:** rebased on the follow-up #7914 (which adds the RHEL CLI-shebang / interpreter-detection fix). No changes to this commit itself. CI green: Manylinux Gate, RHEL8, Debian10 and the wheel-build jobs all pass.

dmitrii-galantsev (approved):

.github/workflows/amdsmi-build.yml
```diff
@@ -38,6 +38,76 @@ defaults:
...
+  manylinux-build:
+    name: Build (Manylinux Gate)
+    runs-on: ubuntu-24.04
+    container:
+      image: quay.io/pypa/manylinux_2_28_x86_64
+    env:
+      BUILD_DIR: /tmp/amdsmi-build
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
```
geomin12, line 52: let's checkout specific paths! so we save time

.github/workflows/amdsmi-build.yml
```diff
@@ -38,6 +38,76 @@ defaults:
...
+        run: |
+          set -euo pipefail
+          wheel_dir="$BUILD_DIR/py-interface/python_package"
+          shopt -s nullglob
+          wheels=("$wheel_dir"/*.whl)
+          if [ "${#wheels[@]}" -eq 0 ]; then
+            echo "ERROR: no wheel produced under $wheel_dir" >&2
+            exit 1
+          fi
+          printf 'Python wheel built: %s\n' "${wheels[@]}"
```
geomin12, line 114: can we try checking if we can install this wheel? not just verify it is there?

.github/workflows/amdsmi-build.yml
```diff
@@ -38,6 +38,76 @@ defaults:
...
+      - name: Verify wheel artifact
+        run: |
+          set -euo pipefail
+          wheel_dir="$BUILD_DIR/py-interface/python_package"
+          shopt -s nullglob
+          wheels=("$wheel_dir"/*.whl)
+          if [ "${#wheels[@]}" -eq 0 ]; then
+            echo "ERROR: no wheel produced under $wheel_dir" >&2
+            exit 1
+          fi
```
geomin12, line 113: can this wheel be re-used anywhere? i would at least want to test if we can install it or test it

.github/workflows/amdsmi-build.yml
marifamd: 9222b9797cb
Done

.github/workflows/amdsmi-build.yml
marifamd: 9222b9797cb
Now checks if pip install works and trys to `import amdsmi` after

.github/workflows/amdsmi-build.yml
marifamd: build-verification & debug only; the reusable wheels come from the `amdsmi-manylinux-build.yml`

geomin12 (approved):

marifamd (approved):

PR #5725, opened 2026-05-02
[AMD-SMI] tools: add manylinux wheel build scripts

amd-juwillia:
Add wheel build infrastructure for producing manylinux_2_28 wheels for cp38-cp313 from both Debian and RPM-based environments.

- tools/_build_wheel_common.py: shared helpers
- tools/build_wheel_debian.py: Debian/Ubuntu wheel builder
- tools/build_wheel_rpm.py: RHEL/SLES wheel builder
- .github/workflows/manylinux-build.yml: CI workflow

PR #7914, opened 2026-06-28
ci(amdsmi): [ROCM-3941] Python packaging rework

amd-juwillia:
## Motivation

`amd-smi` shipped its Python module by having the system `.deb`/`.rpm` postinst call `pip install` and drop an `amdsmi.pth`, with a ~200-line ROCm-path-walking loader. This drops all `pip`/`.pth` plumbing, installs the module directly into the system interpreter's `site-packages`, and adds an opt-in SONAME-isolated wheel build so a PyPI wheel can coexist in-process with the system library.

> **Why loader + wheel build ship together:** develop's `amdsmi-manylinux-build.yml` (from #5725) runs on every `projects/amdsmi/**` PR and configures `-DBUILD_PYTHON_WHEEL=ON`. A loader-only PR would either fail that job (no wheel target) or emit a wheel that silently falls back to a system `libamd_smi.so`. The complete, correct wheel build therefore belongs in this base PR; CI wiring and tests stack on top (#7823, #7824).

## Technical Details

**Loader** (`py-interface/amdsmi_wrapper.py`, `tools/generator.py` — regenerated in lockstep):
- Flat 3-case ladder: `AMDSMI_LIB_OVERRIDE`, a bundled `libamd_smi_python.so`, then system `libamd_smi.so`; a `_MissingLibrary` sentinel keeps import tolerant.

**Install layout** (`py-interface/CMakeLists.txt`):
- Query `/usr/bin/python3` for its real on-`sys.path` dir and `install(DIRECTORY ...)` the module there, excluding `libamd_smi*.so`.

**Packaging** (`DEBIAN/postinst.in`, `prerm.in`, `RPM/post.in`, `preun.in`, `CMakeLists.txt`):
- Remove pip install/uninstall + `.pth`; drop the `python3-pip`/`setuptools`/`wheel` deps and `setup.py.in`.

**Wheel build** (`src/CMakeLists.txt`, `py-interface/CMakeLists.txt`, `tools/disable_system_fallback.py`, `CMakeLists.txt`):
- `BUILD_PYTHON_WHEEL` (default OFF, requires `BUILD_SHARED_LIBS`): a SONAME-isolated `libamd_smi_python.so` linked `-Bsymbolic-functions`, and a `python_wheel` target that flips `_AMDSMI_ALLOW_SYSTEM_FALLBACK` to `False` in the staged wrapper so the wheel never loads a system lib.

**CLI / build** (`amdsmi_cli/amdsmi_cli.py`, `amdsmi_init.py`, `src/CMakeLists.txt`):
- Pin shebang to `/usr/bin/python3`; drop a dead guard; strip the system lib in Release.

> Split 1 of 3 from #5726. #7823 (CI job) and #7824 (tests) stack on this; all target `develop`.

## JIRA ID

Resolves ROCM-3941

## Test Plan

- `python3 projects/amdsmi/tools/build_wheel.py --project-dir projects/amdsmi --no-repair` (mirrors the manylinux CI).
- Multi-OS Docker build/install/verify.

## Test Result

- `build_wheel.py` builds clean; the wheel bundles only `libamd_smi_python.so` and ships `_AMDSMI_ALLOW_SYSTEM_FALLBACK = False`.
- `import amdsmi` + `amdsmi_init()` succeed via `/usr/bin/python3` with no `.pth`/pip across the AMD CI images.

## Submission Checklist

> Update: `test_abi_compat.py` (loader-contract/ABI test) now ships here with the loader code it covers, and runs in the build-and-test matrix.


projects/amdsmi/tests/python/test_abi_compat.py
```diff
@@ -0,0 +1,373 @@
...
+#
+# When a user pins amdsmi_interface.py from one ROCm version against a
+# libamd_smi.so from another, the contract we promise is:
+#   * Common amdsmi_* symbols still in the older .so keep working.
+#   * Symbols added in a newer wrapper but missing from the older .so
+#     fail cleanly via ctypes' AttributeError -- no silent corruption,
+#     no fake success.
+#   * If the .so cannot be loaded at all, importing the wrapper still
+#     succeeds (degraded mode) so doc/lint tooling keeps working; the
+#     first call into a wrapped symbol raises OSError with a diagnostic.
```
marifamd, line 31: look into the edge cases where this is applicable.
- When a user has a python install that pip installs on the system and the .so it references is out of date.
    - This should go away with system level python installs and pypi
    - @amd-juwillia Will the system level python installer reference the .so from the rocm path or will it have it's own .so?
    - However system level multi-rocm API usage
        - This doesn't apply to the CLI since the CLI now references the current "alternative"/Multi-rocm that has been set

amd-juwillia: @marifamd Tested this locally to confirm.

System install: uses the .so from the ROCm path. The module installs to dist-packages with no bundled .so, so import amdsmi resolves libamd_smi.so.26 via the linker to /opt/rocm/lib. Confirmed with /proc/self/maps.

Stale .so:

pip wheel: ships its own libamd_smi_python.so and never falls back to a system lib. Stays bound to its bundled .so even with /opt/rocm present, and fails loudly if that .so is missing. So an out-of-date system .so can't affect it.

system package: the SONAME major bounds the ABI. Only .26 loads (.24/.25/.27 are rejected), and the major comes from AMDSMI_LIB_VERSION_MAJOR, so any matching .26 is compatible.

.github/workflows/amdsmi-build.yml
```diff
@@ -222,6 +223,10 @@ jobs:
         if: always()
         run: cat /tmp/test-results-${{ matrix.os }}/perf_test_output.txt || echo "No perf test results"
 
+      - name: ABI Compat Test Results
+        if: always()
+        run: cat /tmp/test-results-${{ matrix.os }}/abi_compat_output.txt || echo "No ABI compat results"
```
geomin12, line 227: let's make sure this is readable format

geomin12 (commented):
yml file changes lgtm, but i would get some AMDSMI folks to review as majority of the changes are amdsmi related

projects/amdsmi/py-interface/CMakeLists.txt
```diff
@@ -98,34 +113,145 @@ add_custom_command(
...
+# means "auto-detect from /usr/bin/python3 below".
+set(AMDSMI_SYSTEM_PYTHON_SITELIB
+    ""
+    CACHE PATH
+    "Install destination for the amdsmi Python module (system site-packages). Empty = auto-detect from /usr/bin/python3."
+)
+if(NOT AMDSMI_SYSTEM_PYTHON_SITELIB)
+    if(EXISTS "/usr/bin/python3")
+        execute_process(
+            COMMAND /usr/bin/python3 -c "${_amdsmi_sitelib_query}"
```
amd-nicknick: RPM packaging should not do this. The system site-packages path usually contains version number of Python (at least in RHEL / Fedora AFAIK)
Together with package declared minimum version of python3 > 3.6.8, this introduces a potential runtime mismatch.
https://github.com/ROCm/rocm-systems/blob/986c7acfae45cf60249f0351ba72779479bb5f58/projects/amdsmi/CMakeLists.txt#L561

For example, if build system has system Python 3.14 installed, on Fedora / RHEL the system site package path will be `/usr/lib64/python3.14/site-packages`, this will be the path resolved by `_amdsmi_sitelib_query`, and will be hardcoded into the RPM install file paths.

But since dependency is declared as python3 > 3.6.8, this RPM can later be installed on a system with system Python 3.6. On the running system the site package path will be `/usr/lib64/python3.6/site-packages`, and the Python would not see the installed 3.14 amdsmi wheel.

See https://fedoraproject.org/wiki/User:Torsava/PackagingDrafts/PythonProvides
We probably want to use `%{python3_sitearch}` macro, or at least tighten the version requirement to 3.14.

amd-nicknick: On Debian / Ubuntu the current impl is not broken, but ideally if we could use PyBuild / dh_python3 that would be better here (Or just hardcode `/usr/lib/python3/dist-packages`). Again, potentially polluting the package with build system config is probably not a good idea.

amd-nicknick (dismissed):
Native package building needs some fixes.
.so fallback logic & pip wheel handling LGTM. Thanks!

dmitrii-galantsev:

Restructured the ROCM-3941 work into a proper PR chain so each PR shows a single commit: #7914 (this) -> #7913 -> #7912. History rewritten and force-pushed; please re-pull your local branches.

Fixes applied to this commit on top of the original:

**RPM/CLI interpreter mismatch (main fix).** The module installed into a version-specific `site-packages` detected from `/usr/bin/python3` at build time. On RHEL, CI/build images repoint `/usr/bin/python3` to a newer interpreter via `alternatives`, so the package landed where the system default `python3` (and the `amd-smi` CLI `#!/usr/bin/python3` shebang) never look. Result: `import amdsmi` and the CLI both failed with `ModuleNotFoundError` on stock el8/el9. Detection now anchors on `/usr/libexec/platform-python` (RHEL's default interpreter, immune to `alternatives`), falling back to `/usr/bin/python3` on Debian/SUSE/Azure.

**RPM dependency.** Interpreter dependency is now distro-aware and matches the detected default interpreter: `python(abi) = X.Y` on RHEL/Fedora/Azure, `pythonXY` on SUSE, loose `python3 >= 3.6.8` on Debian (version-agnostic `dist-packages`).

**Minor.** Type hints on `disable_system_fallback.py`; `except Exception` -> `except ImportError` and type hints in `test_abi_compat.py`.

Verified in Docker against real consumers (not CI): built on el8 with the harness switching build-python to 3.12, installed on stock el8 (default 3.6), confirmed `import amdsmi` under the default python and the `amd-smi` CLI both work (reach GPU detection, no import error). Debian `dist-packages` behavior unchanged; el9 resolves to 3.9 correctly. The wheel path (bundled SONAME-isolated `libamd_smi_python.so`) was verified to coexist in-process with a separately-loaded `libamd_smi.so` (the PyTorch scenario) with no symbol interposition.


---

**Update (follow-up round, CI-verified):** the initial `python(abi)` pin fixed `import` under the pinned interpreter but the `amd-smi` CLI still failed on el8/el9 in CI. Root cause: the module installs into a version-specific site-packages, but detection trusted `/usr/bin/python3`, which build/CI images repoint via `alternatives`, and the CLI shebang `#!/usr/bin/python3` then ran under an interpreter that could not import amdsmi.

Fixes added to this commit:
- Detect the site-packages path from the distro-DEFAULT interpreter (`/usr/libexec/platform-python` on RHEL, immune to `alternatives`), not the possibly-repointed `/usr/bin/python3`.
- Pin the installed `amd-smi` CLI shebang to the concrete interpreter the module was built for, derived from that anchor's real path: `/usr/libexec/platform-python3.6` on el8, `/usr/bin/python3.9` on el9, etc. Debian keeps `#!/usr/bin/python3` (version-agnostic dist-packages). Note: a naive `#!/usr/bin/python3.6` would fail on el8, where that path does not exist (the interpreter is at `/usr/libexec/platform-python3.6`).
- Fix the install-verify step to check `import amdsmi` under the interpreter the CLI actually resolves to (read from the installed CLI shebang), not a bare `/usr/bin/python3` the harness may have switched.

Verified in Docker across el8/el9/SLES/Azure/Debian and confirmed green in this PR's CI: AlmaLinux8, RHEL8/9/10, SLES, AzureLinux3, Debian10, Ubuntu20/22 all pass (AlmaLinux8 and RHEL8 were previously failing). Approach cross-checked against the Fedora/RHEL Python packaging guidelines.


---

**Update 2 (cross-consumer regression fix):** the previous round moved the `amdsmi` Python module out of `share/amd_smi` into site-packages only. That broke a downstream consumer: **rocprofiler-compute** (and omnistat) do `sys.path.insert(0, ROCM_PATH + "/share/amd_smi"); import amdsmi`, and the TheRock artifact manifest (`core/artifact-core-amdsmi.toml`) captures `share/amd_smi/**` but not the interpreter's site-packages, so the module vanished from the artifact and its test failed with `ModuleNotFoundError`.

Investigated the full consumer set (PyTorch, vLLM, JAX, Triton, DeepSpeed, omnistat, distro packagers, and every in-repo consumer). Conclusion: site-packages is the correct primary (PyTorch/vLLM and all major ML consumers use plain `import amdsmi`), but a few first-party ROCm tools and the artifact flow still rely on `share/amd_smi`.

Fix (additive, keeps site-packages primary):
- Also stage the module under `share/amd_smi` in addition to site-packages, so the TheRock artifact and the `ROCM_PATH/share/amd_smi` consumers keep working.
- Restore a `share/amd_smi` sys.path fallback in the CLI (`amdsmi_init.py`) for the artifact layout, tried only if the site-packages `import amdsmi` fails.

Verified in Docker: (a) TheRock artifact layout — module present at `share/amd_smi`, rocprofiler-compute's `sys.path.insert(...share/amd_smi); import amdsmi` succeeds; (b) RPM — ships both copies, plain `import amdsmi` (site-packages), the `share/amd_smi` path import, and the `amd-smi` CLI all work on stock el8.

projects/amdsmi/py-interface/CMakeLists.txt
dmitrii-galantsev: https://github.com/ROCm/rocm-systems/blob/71e8dde53691e47ad420f82ac3d1bd5fd7a4ba1e/projects/amdsmi/py-interface/CMakeLists.txt#L162

Ended up solving this with some insane cmake magic after doing a lot of testing in docker and CI. It WORKS!  
basically we try and get the default python path and install there. But it super specific to each distribution. So we try and handle all the nonsense ourselves.

dmitrii-galantsev (approved):

dmitrii-galantsev:
blocked by https://github.com/ROCm/rocm-systems/pull/9061

marifamd (approved):

dmitrii-galantsev:
#9061 resolved, ci should work

.github/workflows/amdsmi-build.yml
marifamd: 2f32e4b443a - Collapsed repetitive result steps into one step that prints each log in a labeled collapsible group in the gh actions logging

.github/workflows/amdsmi-build.yml
geomin12, line 52: should we just add the label name here? instead of using a var? (var disable forked PR checks)

geomin12: oh this is intended? maybe explanation on why?

geomin12, line 68: same here. we could just hard-code it?

geomin12, line 74: should we checkout specific paths? bc this is costly (as rocm-systems is pretty big)

geomin12: since this is a docker container, it'll clean itself up after run! no need for this

geomin12: tbh no need for this? if the script tries to get called and fails, we already know the error. i think removing this and adding the script below will be easier for reproduction locally

geomin12, line 133: why *99999?

geomin12, line 136: just curious, do we need to upload? is this re-used anywhere?

.github/workflows/amdsmi-build.yml
```diff
@@ -206,29 +207,24 @@ jobs:
...
-        if: always()
-        run: cat /tmp/test-results-${{ matrix.os }}/perf_test_output.txt || echo "No perf test results"
-
-      - name: Example DRM Test Results
-        if: always()
-        run: cat /tmp/test-results-${{ matrix.os }}/amd_smi_drm_ex.log || echo "No DRM example results"
-
-      - name: Example NoDRM Test Results
-        if: always()
-        run: cat /tmp/test-results-${{ matrix.os }}/amd_smi_nodrm_ex.log || echo "No NoDRM example results"
```
geomin12, line 231: i might actually prefer this. it depends on how large the test results are, but keeping it as individual steps may help, so if i need to look at only unit tests, i shouldn't have to parse through

.github/workflows/amdsmi-build.yml
```diff
@@ -206,29 +207,24 @@ jobs:
           ./amd_smi_drm_ex > "$RESULTS_DIR/amd_smi_drm_ex.log" 2>&1 || echo 'amd_smi_drm_ex failed'
           ./amd_smi_nodrm_ex > "$RESULTS_DIR/amd_smi_nodrm_ex.log" 2>&1 || echo 'amd_smi_nodrm_ex failed'
```
geomin12, line 207: personally, let's separate these test results out to individual steps. maybe with `Test Results`, each step should have its own

let's say `amd_smi_drm_ex` failed, we would need to dig pretty deep in the logs to find what went wrong, and we would also not know what step even failed

projects/amdsmi/tests/amdsmi_build/run_amdsmi_build.py
geomin12: The verify_wheel_site_packages() function claims to verify PyPI takes priority, but the test is
  flawed:

  # line 970-972 comment:
  # so a coexisting system module is fine (whichever sys.path entry wins
  # is whichever is searched first by the active python).

  The actual check (lines 978-982) only verifies the module is in some valid location:
  "ok = 'site-packages' in p or 'dist-packages' in p or '/opt/rocm/' in p; "

  Problem: This does NOT verify PyPI takes priority over system packages. It just checks the import
  came from any of those locations. If the system package is loaded instead of the pip-installed
  wheel, the test still passes.

geomin12, line 937: --force-reinstall Masks the Issue (line 937)

  Using pip install --force-reinstall will overwrite any existing pip installation, but it won't
  remove the system package. Both can coexist, and Python's sys.path order determines which one wins.

  Recommendation

  If the goal is to verify PyPI takes priority over system packages, you need to:
  1. Install system package (already happening)
  2. Install pip wheel (already happening)
  3. Explicitly check that amdsmi.__file__ points to site-packages (NOT dist-packages or /opt/rocm)

  The current assertion allows /opt/rocm/ as valid, which defeats the purpose of the test.

geomin12 (commented):
some cleanup comments and also some AI generated comments about order priority

.github/workflows/amdsmi-build.yml
marifamd: 66e6f9749ad
Done

projects/amdsmi/tests/amdsmi_build/run_amdsmi_build.py
marifamd: 9d9a7858d22
Now asserts `import amdsmi` resolves under pip's reported location (not rocm), so a shadowing system copy fails the check.

marifamd: 9d9a7858d22
The check now verifies the winning module is pip's, so a coexisting system copy is caught rather than masked.

.github/workflows/amdsmi-build.yml
marifamd: 66e6f9749ad
Reverted to individual steps so a single failure is directly attributable.

.github/workflows/amdsmi-build.yml
marifamd: 66e6f9749ad
Reverted to individual steps so a single failure is directly attributable.

.github/workflows/amdsmi-build.yml
marifamd: af9857b3d49
Commented that it is for debug, no job consumes it.

marifamd: 66e6f9749ad
Added comment, it's the local (non-release) version CPack stamps when ROCM_LIBPATCH_VERSION is unset.

marifamd: 3252cec741f
Removed it so any failure on our self-hosted runners is visible; these runners are expected to be stable.

marifamd: 66e6f9749ad
Removed

marifamd: 66e6f9749ad
Dropped

.github/workflows/amdsmi-build.yml
```diff
@@ -85,26 +90,16 @@ jobs:
       - name: Set Project Directory
         run: |
           set -euo pipefail
-          TARGET_DIR=$(find $GITHUB_WORKSPACE -path "*/projects/amdsmi/CMakeLists.txt" -exec dirname {} \;)
-          if [ -z "$TARGET_DIR" ]; then
-            TARGET_DIR=$(find $GITHUB_WORKSPACE -maxdepth 2 -name "CMakeLists.txt" -exec dirname {} \; | head -n 1)
-          fi
-          echo "PROJECT_DIR=$TARGET_DIR" >> $GITHUB_ENV
+          # sparse-checkout places the project at a fixed path.
+          echo "PROJECT_DIR=$GITHUB_WORKSPACE/projects/amdsmi" >> $GITHUB_ENV
```
geomin12, line 94: you could also make this a global env variable! so no need for extra step

geomin12 (approved):
lgtm! thanks for addressing comments!

.github/workflows/amdsmi-build.yml
marifamd: Fixed it up the stack in #9033 
[b03bbc6](https://github.com/ROCm/rocm-systems/pull/9033/commits/b03bbc61874cc4d8dcd834efb62c3f46ab6bd514)

PR #7916, opened 2026-06-28
fix(rccl): [ROCM-27155] false "iommu" warning when using config var

dlamd1dai:
**Summary**
Two helpers are added before ncclInit():
1) ncclKernelHasConfigOption() — searches kernel config files (same paths as rocmwrap.cc: /proc/config.gz, /boot/config-<release>, etc.) for a given CONFIG_* option.
2) ncclIommuPassthroughOk() — returns true if either:
   2.a)  iommu=pt is present in /proc/cmdline, or
   2.b)  CONFIG_IOMMU_DEFAULT_PASSTHROUGH=y is found in the kernel config.

Modified: projects/rccl/src/init.cc

**Motivation**
LLNL El Capitan (MI300A) users frequently see this warning during PyTorch / RCCL init:

NCCL WARN Missing "iommu=pt" from kernel command line which can lead to system instablity or hang!
On their RHEL8 kernels, IOMMU passthrough is already the default via CONFIG_IOMMU_DEFAULT_PASSTHROUGH=y, even when iommu=pt is absent from /proc/cmdline. The warning is a false positive and creates unnecessary support noise; workloads run correctly if the warning is ignored.

This change suppresses the warning when passthrough is already guaranteed by kernel build configuration, while preserving the warning for hosts that lack both explicit iommu=pt and default passthrough.

**Technical Details**
Previously, ncclInit() only checked /proc/cmdline for the literal substring iommu=pt. That misses kernels built with CONFIG_IOMMU_DEFAULT_PASSTHROUGH=y, where passthrough is active without an explicit boot parameter.

Summary of changes:
1. Shared kernel config helper
projects/rccl/src/include/kernel_config.h — public API
projects/rccl/src/misc/kernel_config.cc — implementation using C++17 std::ifstream and full-file std::string search (fixes mch’s fgets boundary issue)
2. init.cc
Uses ncclIommuPassthroughOk() from shared module
Cmdline bug fix: dedicated cmdline[] buffer; passes NULL when /proc/cmdline read fails (no BIOS string reuse)
3. rocmwrap.cc
DMA-BUF kernel config check now uses ncclKernelConfigReadFirstAvailable() + ncclKernelConfigContentHasOption()
/proc/kallsyms fallback preserved
4. Unit tests added
projects/rccl/test/IommuPassthroughTests.cpp — deterministic tests for content search, file read, and iommu=pt cmdline
Wired into rccl-UnitTestsFixtures with kernel_config.cc compiled directly (same pattern as AltRsmiTests)

Expected behavior:
1) If iommu=pt is on the cmdline, unchanged behavior as previously.
2) Otherwise, scans kernel config (using the same search paths as rocmwrap.cc) for CONFIG_IOMMU_DEFAULT_PASSTHROUGH=y.
3) Emits the existing WARN only when neither condition is met.

Related context:
1) ROCm install FAQ — [Issue 5: Application hangs on multi-GPU systems](https://rocm.docs.amd.com/projects/install-on-linux/en/latest/reference/install-faq.html#issue-5-application-hangs-on-multi-gpu-systems)
2) Prior warning location: projects/rccl/src/init.cc (~L301–302)

JIRA ID : ROCM-27155

**Test Plan**
- [ ] On a host with iommu=pt on cmdline: run minimal RCCL init (ncclGetUniqueId / PyTorch init_process_group) — no WARN (unchanged).
- [ ]  On an LLNL-like host without iommu=pt but with CONFIG_IOMMU_DEFAULT_PASSTHROUGH=y in /boot/config-$(uname -r) or /proc/config.gz: run RCCL init — WARN should not appear.
- [ ]  On a host without iommu=pt and without CONFIG_IOMMU_DEFAULT_PASSTHROUGH=y: run RCCL init — WARN should still appear.
- [ ]  Run a short multi-GPU collective (e.g. all_reduce) on an affected El Capitan node to confirm no functional regression.
- [ ]  Run tests (if you have RCCL test build): # from RCCL build directory with BUILD_TESTS=ON
./rccl-UnitTestsFixtures --gtest_filter='IommuPassthroughTest.*'

**Test Result**
1) Verified on smci355-ccs-aus-m03-17.cs-aus.dcgpu (has explicit iommu=pt): no WARN, behavior unchanged.
2) El Capitan / LLNL validation pending on a node with CONFIG_IOMMU_DEFAULT_PASSTHROUGH=y and no iommu=pt on cmdline.


projects/rccl/src/init.cc
```diff
@@ -266,6 +267,63 @@ static void initOnceFunc() {
...
+    if (strstr(path, "/proc/config.gz") != NULL) {
+      pclose(fp);
+    } else {
+      fclose(fp);
+    }
+
+    if (found) return true;
+  }
+  return false;
+}
```
nileshnegi: we have a similar check for kernel config in https://github.com/ROCm/rocm-systems/blob/develop/projects/rccl/src/misc/rocmwrap.cc#L259

might benefit to unify all kernel config checks in one place.

nileshnegi (approved):

projects/rccl/src/init.cc
dlamd1dai: That is a good idea for RCCL. We might have to take El Capitan as an exception.

prasanna-amd (approved):

projects/rccl/src/init.cc
```diff
@@ -266,6 +267,63 @@ static void initOnceFunc() {
...
+      if (!hasZcat || access("/proc/config.gz", R_OK) != 0) continue;
+      fp = popen("zcat /proc/config.gz 2>/dev/null", "r");
+    } else {
+      fp = fopen(kernelConfFile, "r");
+    }
+    if (fp == NULL) continue;
+
+    bool found = false;
+    while (fgets(buf, sizeof(buf), fp) != NULL) {
+      if (strstr(buf, option) != NULL) {
```
mch: I think this will fail to find `option` if it is over the boundary two `fgets` calls. Either read line by line or using a sliding window.

projects/rccl/src/init.cc
```diff
@@ -266,6 +267,63 @@ static void initOnceFunc() {
...
+    "/usr/src/linux-%s/.config",
+    "/usr/src/linux/.config",
+    "/usr/lib/modules/%s/config",
+    "/usr/lib/ostree-boot/config-%s",
+    "/usr/lib/kernel/config-%s",
+    "/usr/src/linux-headers-%s/.config",
+    "/lib/modules/%s/build/.config",
+  };
+
+  char kernelConfFile[128];
```
mch: I think using C++17 constructs like `std::filestream` and `std::ifstream` and avoiding buffers of a specific size would be safer and more concise.

projects/rccl/src/init.cc
```diff
@@ -78,6 +78,7 @@
 #include "dda_all_reduce_ipc.h"
 #include "ipc_init.h"
 #include  <cpuid.h>
+#include <sys/utsname.h>
```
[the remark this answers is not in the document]
dlamd1dai: Those libs are included earlier in build flow. Compilation passed without any issue.

PR #7917, opened 2026-06-28
feat(FreeBSD-Support-hsa-types): Add FreeBSD support to hsakmttypes.h

Yohello1:
Sponsored By: FreeBSD Foundation

## Motivation

Working on supporting FreeBSD, just needed to add a condition to a macro to also include certain libraries if on FreeBSD> 

## Technical Details

Modified this line so the correct libraries also get imported on freebsd;
```diff
 
-#elif defined(__linux__)
+#elif defined(__linux__) || defined (__FreeBSD__)
```

## JIRA ID

ISSUE ID: #9999999

## Test Plan

Added test to check types, and tested it by running;
 `LD_LIBRARY_PATH=/opt/rocm ./rocrtst64`

## Test Result

The test results were the exact same as before & after this change, including the test I added

It can still run my fluid simulation on Fedora 43 Linux. 

## Submission Checklist


Yohello1:
@dayatsin-amd Should I open an issue? I don't want to clog up the issues queue
I guess I'm writing tests lol.

Yohello1:
@kentrussell Should I rename the test file to comply with the bot? I didn't want to deviate from the naming convention too much

kentrussell:
> @kentrussell Should I rename the test file to comply with the bot? I didn't want to deviate from the naming convention too much

I wouldn't worry about the rename. I'll defer to David or @cfreeamd for the actual code review of the rocrtst part. The thunk change is fine with me.

dayatsin-amd (changes requested):
The change to hsakmttypes.h is ok.

But hsakmttypes.h is a private header within the libhsa-runtime library, so it should not be included by components other that the libhsa-runtime library itself (e.g: rocrtst).

I don't think we should have a new unit test for this. If anything these should be compile-time checks. 

@Yohello1  Could you please remove the changes to rocrtst/* in this PR. we can see how to override the PR check that requires the Unit Test on our side.

Yohello1:
@dayatsin-amd Sounds good, I just reverted the change.

projects/rocr-runtime/libhsakmt/include/hsakmt/hsakmttypes.h
```diff
@@ -52,7 +52,7 @@ extern "C" {
     typedef signed __int64     HSAint64;
     typedef unsigned __int64   HSAuint64;
 
-#elif defined(__linux__)
+#elif defined(__linux__) || defined (__FreeBSD__)
```
cfreeamd, line 56: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `defined (__FreeBSD__)` has a space before `(` that is inconsistent with the style of `defined(__linux__)` on the same line. Consider removing the space: `defined(__FreeBSD__)`.

cfreeamd (commented):
The `hsakmttypes.h` condition and surrounding C/C++ type aliases are the only file touched; the change is minimal.

Yohello1:
@dayatsin-amd Changes have been made, let me know if any other changes are needed

dayatsin-amd (approved):
Looks good to me. Thank you

cfreeamd (approved):

Yohello1:
@dayatsin-amd So how do we merge it in then given the bot doesn't like stuff

Yohello1:
Merged into https://github.com/ROCm/rocm-systems/pull/8430

PR #7924, opened 2026-06-29
[rocprofiler-sdk] Fix queue interposition completion waiter ordering

venkat1361:
## Motivation

- `transpose-sampling` test from rocprofiler-systems can hang because queue interposition may start the async completion-signal waiter before the final real doorbell is published, so the waiter can block waiting for GPU completion while the GPU has not yet been told about the packet.

## Technical Details

- Defer completion-waiter enqueue from `write_interceptor()` to `process_doorbell_impl()`, after `publish_submitted_packets()` publishes the real write index and rings the doorbell.

## Test Result

- https://github.com/ROCm/rocm-systems/actions/runs/28333482927/job/83974368084(includes commits from PR which is causing test failures).

## Submission Checklist


minseobshin11 (approved):
LGTM

bwelton:
While i don't have opposition to this change. 

`queue interposition may start the async completion-signal waiter before the final real doorbell is published. so the waiter can block waiting for GPU completion while the GPU has not yet been told about the packet.`

This shouldn't matter. The async just wont trigger until the completion signal is set. Its not clear why this change fixes anything other than changing the ordering and holding the lock for potentially less time.

bwelton (approved):

rahulc-gh (approved):
LGTM - wait for CI to Pass

PR #7927, opened 2026-06-29
fix(rocprofiler-systems): Replace oshrun version gate with live -- strip probe

mradosav-amd:
## Motivation

The `shmem_pingpong` test was failing on machines with Open MPI 5.0.9 installed from source. A previous commit had added an extra `--` decoy to the oshrun command for any `oshrun_version[0] >= 5`, to work around PRRTE-based launchers that strip the first `--` from the application argv (breaking `rocprof-sys-run -- <binary>`). However, empirical testing across Open MPI 5.0.0–5.0.10 showed that only 5.0.0 strips `--` when built from upstream source. RHEL 10's distro package of Open MPI 5.0.2 independently bundles PRRTE 3.0.2 (which strips), while upstream 5.0.2 does not. A major-version comparison is therefore an unreliable discriminant — the stripping depends on which PRRTE version a given build actually links, not on the Open MPI version number.

## Technical Details

Adds `oshrun_strips_double_dash` as a `cached_property` on `SystemCapabilities` (`capabilities.py`). At session startup it spawns `oshrun -n 1 probe.sh -- SENTINEL` and checks whether the probe script receives `SENTINEL` as `$1` (stripped) or `--` (preserved). It retries with `--allow-run-as-root` if the first attempt fails due to a root-guard rejection. The probe runs once and is memoised for the rest of the session. `runners.py` now gates the decoy `--` insertion on `capabilities.oshrun_strips_double_dash` instead of `oshrun_version[0] >= 5`. `conftest.py` adds a `Strips '--':` row to the test session header so the behaviour is visible in CI logs.

## Test Plan
- [x] Verified probe returns `STRIPPED` in `dgaliffiamd/rocprofiler-systems:ci-rocm-7.1-rhel-10` (Open MPI 5.0.2, PRRTE 3.0.2)
- [x] Verified probe returns `PRESERVED` in Ubuntu 22.04/24.04 CI images (Open MPI 4.1.x, ORTE)
- [x] Built and probed all 11 upstream Open MPI releases 5.0.0–5.0.10 in a container; only 5.0.0 strips
- [x] `shmem_pingpong` test passes on machine with Open MPI 5.0.9 from source

---
JIRA ID - AIPROFSYST-557

### DIFF
--- a/projects/rocprofiler-systems/tests/pytest/rocprofsys/capabilities.py
+++ b/projects/rocprofiler-systems/tests/pytest/rocprofsys/capabilities.py
@@ -462,6 +462,43 @@ class SystemCapabilities:
             return None
 
     @cached_property
+    def oshrun_strips_double_dash(self) -> bool:
+        """Return True if this oshrun strips the first '--' from application argv.
+
+        Probes the live binary by running:
+            oshrun -n 1 probe.sh -- SENTINEL
+        and checking whether the script receives 'SENTINEL' (stripped) or '--'
+        (preserved).  Falls back to False when oshrun is absent or the probe
+        fails.
+        """
+        if not self.oshrun_exec:
+            return False
+        import tempfile
+
+        with tempfile.NamedTemporaryFile(
+            mode="w", suffix=".sh", delete=False
+        ) as probe_file:
+            probe_file.write("#!/bin/sh\nprintf '%s\\n' \"$1\"\n")
+            probe_path = probe_file.name
+        os.chmod(probe_path, 0o755)
+
+        try:
+            for extra in ([], ["--allow-run-as-root"]):
+                cmd = (
+                    [str(self.oshrun_exec)]
+                    + extra
+                    + ["-n", "1", probe_path, "--", "SENTINEL"]
+                )
+                result = subprocess.run(cmd, capture_output=True, text=True, timeout=30)
+                if result.returncode == 0:
+                    return result.stdout.strip() == "SENTINEL"
+            return False
+        except (subprocess.SubprocessError, OSError):
+            return False
+        finally:
+            os.unlink(probe_path)
+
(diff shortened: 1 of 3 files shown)


mradosav-amd:
[rhel (g++, 8.10, 7.2, Release)](https://github.com/ROCm/rocm-systems/actions/runs/28360016186/job/84012167793?pr=7927#logs)
    Strips '--':          No

[rhel (g++, 10, 7.2, Release)](https://github.com/ROCm/rocm-systems/actions/runs/28360016186/job/84012167810?pr=7927#logs)
    Strips '--':          Yes (decoy '--' inserted)

projects/rocprofiler-systems/tests/pytest/rocprofsys/capabilities.py
```diff
@@ -461,6 +461,43 @@ def oshrun_version(self) -> Optional[tuple[int, ...]]:
...
+        if not self.oshrun_exec:
+            return False
+        import tempfile
+
+        with tempfile.NamedTemporaryFile(
+            mode="w", suffix=".sh", delete=False
+        ) as probe_file:
+            probe_file.write("#!/bin/sh\nprintf '%s\\n' \"$1\"\n")
+            probe_path = probe_file.name
+        os.chmod(probe_path, 0o755)
```
kcossett-amd: `os.chmod` can fail. Should move this into the `try/except` block.

kcossett-amd (approved):
Thank you for the fix

projects/rocprofiler-systems/tests/pytest/rocprofsys/capabilities.py
mradosav-amd: fixed

dgaliffiAMD (approved):

PR #7546, opened 2026-06-20
ci: RCCL paired A/B perf-regression gate (advisory mode)

speriaswamy-amd:
## Summary

Adds a GitHub Actions workflow that gates RCCL PRs in the rocm-systems monorepo on a paired A/B librccl.so perf-regression check running on TensorWave MI355X (gfx950) nodes.

## How it works

1. Triggered on PRs touching \`projects/rccl/**\` against \`main\` or \`develop\`
2. Builds \`librccl.so\` for **merge-base** (reference) and **PR head** (candidate) from \`projects/rccl\` inside a single Slurm allocation on pinned nodes
3. Runs the A/B regression detector — compares busBw across message sizes for each collective
4. Posts a per-collective verdict table as a sticky PR comment
5. Gates the merge on a confirmed-regression-free result

## Active collectives

| Collective | dtypes | Status |
|---|---|---|
| all_reduce_perf | float, bfloat16 | ✅ gated |
| reduce_scatter_perf | float, bfloat16 | ✅ gated |
| broadcast_perf | float, bfloat16 | ✅ gated |
| alltoall_perf | both | ⏭ skipped — multi-node gfx950 RCCL bootstrap deadlock |
| all_gather_perf | both | ⏭ skipped — multi-node gfx950 RCCL bootstrap deadlock |

## Validation

End-to-end validated on RCCL 2.30.4 (gfx950) with new rccl-tests binary built against ROCm 7.14:

- **Calibration** (job 11566, 2026-07-07): 6/6 collectives PASS, 0 regressions
- **Detect A=A** (job 11569, 2026-07-07): 6/6 collectives PASS, 0 confirmed regressions across 276 message-size/dtype keys
- **Thresholds**: \`{small: 30.3%, mid: 24.7%, large: 8.7%}\` (2× safety factor over observed noise)

## Deployment notes

- Running in **advisory/draft mode** — promoting to hard merge gate after reviewer approval
- \`concurrency\` group serialises all runs on the pinned nodes (no cancel-in-progress)
- Self-hosted runner: \`tw-gfx950-ainic-ROCm-scale-runner\` (TensorWave org)
- CI workspace: \`/it-share/rccl-ci\` on TW shared NFS (configs, scripts, rccl-tests binaries)
- rccl-tests binary: rebuilt at \`/it-share/rccl-ci/rccl-tests-2.30.4/bin/\` (gfx950, MPI-enabled, ROCm 7.14)

## Skipped collectives — known upstream bugs

[...]

PR #7929, opened 2026-06-29
ci: add AINIC test runner config and workflow (non-gate)

nikitaxgusev:
## Summary

- Adds `projects/rccl/tools/scripts/test_runner/configs/ainic_abc.json` — AINIC A/B/C test plan:
  - **Group A**: NetIbMPI + GIN component/unit tests (2 nodes x 1 GPU = 2 ranks)
  - **Group B**: NCCL_NET plugin x knobs sweep on rccl-tests perf (2 nodes x 8 GPU = 16 ranks), incl. out-of-the-box autodetect
  - **Group C**: IONIC hardware counter validation
- Adds `.github/workflows/ainic_test_runner.yml` — manual + CI-branch GitHub Actions workflow for the AINIC cluster

## Non-gate

This workflow has **no pull_request trigger** and does not block any PR merge. Same pattern as PR #7546 (`rccl_perf_regression.yml`).


projects/rccl/tools/scripts/test_runner/configs/ainic_abc.json
```diff
@@ -0,0 +1,462 @@
...
+        },
+        {
+          "name": "A12_Gin_IPutSignal",
+          "description": "GIN put-with-signal (atomic inc/add) completion semantics.",
+          "test_filter": "*GinMPITest.IPutSignalInc*:*GinMPITest.IPutSignalAdd*"
+        },
+        {
+          "name": "A13_Gin_IFlushAfterIGet",
+          "description": "GIN GPU flush ordering after iget.",
+          "test_filter": "*GinMPITest.IFlushAfterIGet*"
```
nikitaxgusev: remove

.github/workflows/ainic_test_runner.yml
```diff
@@ -0,0 +1,382 @@
...
+        required: false
+        default: '2'
+      salloc_time:
+        description: 'Max wall clock for the SLURM allocation (HH:MM:SS)'
+        required: false
+        default: '04:00:00'
+      salloc_exclude:
+        description: 'SLURM nodes to exclude from allocation (known-broken hardware)'
+        required: false
+        default: 'mia1-p01-g[24-25]'
```
nikitaxgusev: they were broken, now ok?

.github/workflows/ainic_test_runner.yml
```diff
@@ -0,0 +1,382 @@
...
+  push:
+    branches:
+      - users/nikitaxgusev/ainic-test-runner
```
nikitaxgusev: when we should change it to origin/develop?

projects/rccl/tools/scripts/test_runner/configs/ainic_abc.json
```diff
@@ -0,0 +1,436 @@
...
+
+    "a_base": {
+      "is_gtest": true,
+      "binary": "rccl-UnitTestsMPI",
+      "num_ranks": 2,
+      "num_nodes": 2,
+      "num_gpus": 1,
+      "timeout": 600,
+      "env_variables": {
+        "NCCL_LAUNCH_MODE": "GROUP",
```
nikitaxgusev: remove?

projects/rccl/tools/scripts/test_runner/lib/test_executor.py
```diff
@@ -1357,9 +1357,34 @@ def run_test(self, test_config, suite_config):
...
+        # RDMA connection setup (ibv_create_qp / ibv_create_cq) pins memory and
+        # fails with "Cannot allocate memory" when the locked-memory soft rlimit
+        # is too low. Under SLURM the soft limit is inherited from the submit
+        # host (often a couple of GB), while compute nodes allow an unlimited
+        # hard limit. Launch each MPI rank through a shell that raises the soft
+        # limit to the hard maximum first. `set -f` keeps gtest filter globs
+        # (e.g. "*Foo*") literal; harmless where the limit is already unlimited.
+        if num_ranks > 1:
+            mpi_prefix = f"{mpi_cmd} {mpi_args} "
+            if cmd.startswith(mpi_prefix):
```
speriaswamy-amd: This looks fragile, even whitespace difference could break this condition

projects/rccl/tools/scripts/test_runner/lib/test_executor.py
```diff
@@ -1357,9 +1357,34 @@ def run_test(self, test_config, suite_config):
...
+                program = cmd[len(mpi_prefix):]
+                inner = f"ulimit -l unlimited 2>/dev/null; set -f; exec {program}"
```
araravik-psd: The ulimit -l unlimited fix only gets applied if cmd.startswith(mpi_prefix) matches exactly (f"{mpi_cmd} {mpi_args} "). There's no else branch, if that string match ever fails (e.g. mpi_cmd/mpi_args
  construction changes elsewhere, an extra space, empty args, a wrapper prepended), the code silently falls through and runs the original unwrapped command. The RDMA "Cannot allocate memory" failure
  this PR fixes would then quietly reappear with no indication that the fix stopped applying, it'd just look like flaky infra.

  Can we add a warning log in the else case so a future break in this match is visible, e.g.:
  else:
      print(f"WARNING: mpi_prefix mismatch, ulimit fix not applied: {cmd!r}")

projects/rccl/tools/scripts/test_runner/configs/ainic_abc.json
```diff
@@ -0,0 +1,436 @@
...
+
+    { "name": "B1. IB baseline",                     "config": "b1_ib",             "enabled": true },
+    { "name": "B1f. CAST alltoall full (1K-16G)",     "config": "b_alltoall_cast_full", "enabled": true },
+    { "name": "B2. CAST WRR / CTS off",              "config": "b2_cast_wrr_ctsoff","enabled": true },
+    { "name": "B3. CAST no-WRR / CTS off",           "config": "b3_cast_nowrr_ctsoff","enabled": true },
+    { "name": "B4. ROCM-IB / CTS off",               "config": "b4_rocmib_ctsoff",  "enabled": true },
+    { "name": "B5. ROCM-IB / CTS on",               "config": "b5_rocmib_ctson",   "enabled": true },
+    { "name": "B6. CAST WRR / CTS on",               "config": "b6_cast_wrr_ctson", "enabled": true },
+    { "name": "B7. Autodetect (out-of-box)",         "config": "b7_autodetect",     "enabled": true },
+    { "name": "B7a. Pure PCI autodetect",            "config": "b7a_pure_autodetect","enabled": true },
```
speriaswamy-amd: Minor: b7a_pure_autodetect is still an acknowledged near-duplicate of b7_autodetect (description says so) and stays enabled: true — a judgment call, not a defect, but flagging since it burns another full 1024-16G sweep for no new coverage.

.github/workflows/ainic_test_runner.yml
```diff
@@ -0,0 +1,360 @@
...
+          # Phase 0: build RCCL with --enable-mpi-tests on THIS compute node
+          # rather than the 8-core submit node. No RCCL_BUILD_DIR here, so
+          # test_runner.py invokes install.sh, which defaults -j to $(nproc) and
+          # therefore uses ALL of the compute node's cores. WORKDIR points at this
+          # run's own checkout so install.sh and the build tree resolve there (on
+          # /it-share, shared with every node in the allocation). Both groups run
+          # against this fresh build (Group A: rccl-UnitTestsMPI gtest binary;
+          # Group B/C: perf binaries loading the fresh librccl.so), so the build
+          # runs whenever either group is selected.
+          if [ "${RUN_GROUP_A}" = "1" ] || [ "${RUN_GROUP_BC}" = "1" ]; then
```
araravik-psd: One thing I wasn't sure about while reading this—could this end up rebuilding RCCL even when a prebuilt `rccl_build_dir` has been provided for a Group B/C-only run?

From what I can tell, this gate only checks `RUN_GROUP_A`/`RUN_GROUP_BC`, not whether `rccl_build_dir` was explicitly supplied (via `GROUP_BC_RCCL_BUILD_DIR` above). If that's the case, it seems like we might still perform a full RCCL build even though the prebuilt directory is used later.

If my understanding is correct, it may be worth gating Phase 0 on whether a build is actually required (i.e., only build when no override was provided). If I've misunderstood the flow and this is already handled elsewhere, feel free to ignore this comment.

.github/workflows/ainic_test_runner.yml
```diff
@@ -0,0 +1,360 @@
...
+        description: 'Max wall clock for the SLURM allocation (HH:MM:SS)'
+        required: false
+        default: '04:00:00'
+      salloc_exclude:
+        description: 'SLURM nodes to exclude from allocation (empty = use all reservation nodes)'
+        required: false
+        default: ''
+  push:
+    branches:
+      - users/nikitaxgusev/ainic-test-runner
```
araravik-psd: As you have already noted please remove the user branch push trigger before merging

araravik-psd (commented):
Please take a look at the review comments I've left on this PR. I'd also recommend getting a review from the RCCL maintainers to ensure the component-specific changes receive a deeper review.

projects/rccl/tools/scripts/test_runner/configs/ainic_abc.json
atulkulk: Yes, pls remove it.

projects/rccl/tools/scripts/test_runner/configs/ainic_abc.json
nikitaxgusev: @speriaswamy-amd 
Removed B7a entirely. Under --system ainic the config always sets RCCL_AINIC_ROCE=1, so "pure PCI autodetect" is indistinguishable from B7 (b7_autodetect), a true duplicate.

projects/rccl/tools/scripts/test_runner/lib/test_executor.py
nikitaxgusev: Reworked, no more startswith on the assembled command. The program is wrapped at construction time:
```
inner = f"ulimit -l unlimited 2>/dev/null; set -f; exec {program}"
cmd = f"{mpi_cmd} {mpi_args} bash -c {shlex.quote(inner)}"
```
So there's no string match that spacing/layout can break.
final line:
```
/it-share/ompi-5.0.8/bin/mpirun -np 2 --host h1:1,h2:1 --bind-to numa --mca pml ob1 --mca btl tcp,self -x NCCL_DEBUG=INFO -x RCCL_AINIC_ROCE=1 -x LD_LIBRARY_PATH=... bash -c 'ulimit -l unlimited 2>/dev/null; set -f; exec /it-share/.../test/rccl-UnitTestsMPI --gtest_filter=NetIbMPITest.SimpleSendRecv --gtest_output=json:/tmp/rccl_gtest_ab12.json'
```

.github/workflows/ainic_test_runner.yml
nikitaxgusev: @araravik-psd Fixed. Phase 0 now only builds when a build is actually needed.

rccl_build_dir is a workflow_dispatch input (manual runs), default empty. On push/auto runs it's always empty, then we build fresh. It's only non-empty if someone manually kicks off the workflow and types a prebuilt lib path. That value flows into RCCL_BUILD_OVERRIDE.

Gate:
```
NEED_BUILD=0
[ "$RUN_GROUP_A" = 1 ] && NEED_BUILD=1                                   # A always needs a fresh build
[ "$RUN_GROUP_BC" = 1 ] && [ -z "$RCCL_BUILD_OVERRIDE" ] && NEED_BUILD=1 # B/C: build only if no override
```
Group A always builds (its rccl-UnitTestsMPI binary comes from that build). Group B/C loads librccl.so via LD_LIBRARY_PATH = its build dir, which defaults to the fresh build and is only the prebuilt path when the rccl_build_dir input is provided. So the only skip case is a Group B/C-only run with an explicit prebuilt override..

projects/rccl/tools/scripts/test_runner/lib/test_executor.py
nikitaxgusev: @araravik-psd 
Fixed at the root: the wrapper is applied unconditionally to program for every MPI launch (num_ranks > 1), so there's no conditional match that can silently skip, no else warning needed.
added:
```
inner = f"ulimit -l unlimited 2>/dev/null; set -f; exec {program}"
cmd = f"{mpi_cmd} {mpi_args} bash -c {shlex.quote(inner)}"
```

```
inner = ulimit -l unlimited 2>/dev/null; set -f; exec /it-share/.../build/debug/test/rccl-UnitTestsMPI --gtest_filter=NetIbMPITest.SimpleSendRecv --gtest_output=json:/tmp/rccl_gtest_ab12.json
```

final line:
```
/it-share/ompi-5.0.8/bin/mpirun -np 2 --host h1:1,h2:1 --bind-to numa --mca pml ob1 --mca btl tcp,self -x NCCL_DEBUG=INFO -x RCCL_AINIC_ROCE=1 -x LD_LIBRARY_PATH=... bash -c 'ulimit -l unlimited 2>/dev/null; set -f; exec /it-share/.../test/rccl-UnitTestsMPI --gtest_filter=NetIbMPITest.SimpleSendRecv --gtest_output=json:/tmp/rccl_gtest_ab12.json'
```

.github/workflows/ainic_test_runner.yml
nikitaxgusev: still broken

nikitaxgusev:
@amd-arozanov @VadimKutovoi @alexander-sannikov Could you please review?

nikitaxgusev:
## AINIC test-runner - timing breakdown

**Pipeline stages** (single `sbatch` job, 2 nodes):

| Stage | Tests | Wall time |
|---|---:|---:|
| Phase 0 — RCCL build (`--enable-mpi-tests`, `nproc=114`) | — | ~2 min |
| Group A — gtest MPI component tests | 13 | ~1 min (57 s) |
| Group B/C — rccl-tests perf binaries | 27* | ~92 min |
| **End-to-end (`sbatch`)** |  | **~1 h 36 min** |

**Group B/C per category**  
Each category is `1 alltoall + 1 allreduce` unless noted.

| Category | # | Time |
|---|---:|---:|
| B1. IB baseline | 2 | 6m 21s |
| B1f. CAST alltoall full range (1K–16G) | 1 | 6m 01s |
| B2. CAST WRR / CTS off | 2 | 6m 38s |
| B3. CAST no-WRR / CTS off | 2 | 6m 38s |
| B4. ROCM-IB / CTS off | 2 | 6m 44s |
| B5. ROCM-IB / CTS on | 2 | 6m 54s |
| B6. CAST WRR / CTS on | 2 | 6m 36s |
| B7. Autodetect out-of-box | 2 | 6m 06s |
| B7b. Autodetect off / fallback | 2 | 6m 23s |
| B8. ROCM-IB / no CTS offload | 2 | 6m 32s |
| B9. CTS force override  | 2 | 6m 43s |
| B-large. alltoall full-range guard | 3 | 18m 05s |
| C1. IONIC counter validation | 4 | 2m 34s |

**Main takeaway:** most of the time is spent in the full-range `alltoall` sweeps from `1K` to `16G`. Each sweep takes about `5.5–6.5 min`, while the `allreduce` legs are much shorter

amd-arozanov:
@nikitaxgusev could you please describe how to run these tests in other PRs when you merge it? And what are the next plans? Are you going to enable these tests for every RCCL PR?

nikitaxgusev:
@amd-arozanov 

There are three realistic options, depending on how often we want this to run:

Option 1 - Nightly (check develop state)

we shoudl enable

```
on:
  workflow_dispatch:
  schedule:
    - cron: '0 2 * * *'   # daily at 02:00 UTC
  ```
schedule always runs against the HEAD of the default branch (develop), i.e. exactly the current develop state.
Cron is in UTC; the workflow must be merged into develop for the schedule to activate.
One run per day, predictable load on the shared reservation nodes. This is the safest option.

Option 2  On every PR

```
on:
  workflow_dispatch:
  pull_request:
    branches: [ develop ]
    paths:
      - 'projects/rccl/**'
      - '.github/workflows/ainic_test_runner.yml'
```
Caveats that make this hard in practice:

The full A/B/C sweep takes ~1h40m on 2 shared reservation nodes, and runs are serialized (concurrency), so multiple PRs would queue for hours.

Option 3 - per-PR: label-gated smoke subset

```
on:
  workflow_dispatch:
  pull_request:
    types: [ labeled ]
jobs:
  ainic-test-runner:
    if: github.event.label.name == 'ainic-test'
```
Runs only when someone adds the ainic-test label - not automatically on every PR.

My suggestion: start with Option 1  + add Option 3.

File: .github/workflows/ainic_test_runner.yml [hunk 10]
```diff
@@ -0,0 +1,392 @@
...
+  DEFAULT_SALLOC_PARTITION: amd-tw
+  DEFAULT_SALLOC_RESERVATION: rccl_dev
+  DEFAULT_SALLOC_NODES: '2'
+  DEFAULT_SALLOC_TIME: '04:00:00'
+  # Exclude known-broken GPU nodes: g24 regressed (every rccl-tests perf run
+  # aborts with HIP 'out of memory' even under an --exclusive allocation, a
+  # symptom of leftover/orphaned GPU state), and g25 is its historically-flaky
+  # pair. The rccl_dev reservation still has g27/g29/g30 for the 2-node job.
+  # Clear this (or override the salloc_exclude input) once the nodes recover.
+  DEFAULT_SALLOC_EXCLUDE: 'mia1-p01-g[24-25]'
```
amd-arozanov: why do we need to specify all these parameters twice? They are already specified above

      salloc_exclude:
        description: 'SLURM nodes to exclude from allocation (known-broken hardware)'
        required: false
        default: 'mia1-p01-g[24-25]'

.github/workflows/ainic_test_runner.yml
```diff
@@ -0,0 +1,392 @@
...
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: ainic-test-report
+          path: |
+            ${{ runner.temp }}/ainic_build.log
+            ${{ runner.temp }}/ainic_run_groupA.log
+            ${{ runner.temp }}/ainic_run_groupBC.log
+            ${{ runner.temp }}/reports
+          if-no-files-found: ignore
```
amd-arozanov: would it be better to fail if no files are found?

projects/rccl/tools/scripts/test_runner/configs/ainic_abc.json
```diff
@@ -0,0 +1,418 @@
...
+    "test_binary_dir": "${RCCL_TESTS_BIN:-/it-share/rccl-ci/rccl-tests-2.30.4/bin}"
+  },
+  "auto_detect_hosts": true,
+  "mpi_args": {
+    "ainic": "--bind-to numa --mca pml ob1 --mca osc ^ucx --mca btl tcp,self --mca btl_tcp_if_include eno0"
+  },
+  "env_variables": {
+    "NCCL_DEBUG": "INFO",
+    "HSA_NO_SCRATCH_RECLAIM": "1",
+    "RCCL_MODEL_MATCHING_DISABLE": "1"
```
amd-arozanov: If there are some differences in topology between nodes, then we need to ask TW admins to align it.  

RCCL_MODEL_MATCHING_DISABLE=1 is not the default value. So, it can hide some errors or produce the new ones, which we do not observe on the production environment

projects/rccl/tools/scripts/test_runner/configs/ainic.json
```diff
@@ -0,0 +1,418 @@
+{
```
amd-arozanov, line 1: for the third person it would be not obvious what 'abc' means when he/she sees ainic_abc.json config name. Please consider renaming it. Some suggestions: 
ainic.json 
ainic_nightly.json 
ainic_sannity.json 
ainic_full.json 
or yours

.github/workflows/ainic_test_runner.yml
```diff
@@ -0,0 +1,392 @@
...
+              case "${tok}" in
+                "A. Component"*) GROUP_A_LIST="${GROUP_A_LIST:+${GROUP_A_LIST}:}${tok}" ;;
+                *)                GROUP_BC_LIST="${GROUP_BC_LIST:+${GROUP_BC_LIST}:}${tok}" ;;
+              esac
+            done
+            if [ -n "${GROUP_A_LIST}" ]; then GROUP_A_FILTER="${GROUP_A_LIST}"; else RUN_GROUP_A=0; fi
+            if [ -n "${GROUP_BC_LIST}" ]; then GROUP_BC_FILTER="${GROUP_BC_LIST}"; else RUN_GROUP_BC=0; fi
+          fi
+
+          # The self-hosted runner is a SLURM submit node without local GPUs, so
```
amd-arozanov: Some information is repeated in many comments. Please clean the comments up

For example:
- Group A always needs a fresh --enable-mpi-tests build
- The self-hosted runner is a SLURM submit node without local GPUs
- check other comments

.github/workflows/ainic_test_runner.yml
```diff
@@ -0,0 +1,392 @@
+# AINIC RCCL test runner — manual trigger only (non-gate).
+#
+# DEPLOYMENT: RCCL lives in the ROCm/rocm-systems monorepo at projects/rccl.
+# The test_runner framework lives at
+# projects/rccl/tools/scripts/test_runner/ and its entry point is test_runner.py
+# (run from that directory so its `from lib...` imports resolve, matching the
+# README). The AINIC A/B/C plan is configs/ainic_abc.json.
```
amd-arozanov: please specify what A, B, C groups mean

projects/rccl/tools/scripts/test_runner/configs/ainic.json
```diff
@@ -0,0 +1,418 @@
...
+      "num_ranks": 2,
+      "num_nodes": 2,
+      "num_gpus": 1,
+      "timeout": 600,
+      "env_variables": {
+        "NCCL_IB_MERGE_NICS": "1",
+        "NCCL_IB_MERGE_VFS": "1",
+        "NCCL_DMABUF_ENABLE": "1"
+      }
+    },
```
amd-arozanov, line 46: is there some wiki page where you describe all configurations you test in this file?

projects/rccl/tools/scripts/test_runner/configs/ainic.json
```diff
@@ -0,0 +1,418 @@
...
+          "description": "CTS-off P2P stress path passes cleanly (1024 outstanding recvs drain fully).",
+          "test_filter": "NetIbMPITest.CtsDepthStress"
+        }
+      ]
+    },
+
+    "a_cts_depth32": {
+      "extends": "a_base",
+      "timeout": 600,
+      "description": "A6: CTS-on at depth 32 (expected PASS). With CTS ON, CTS_QP_DEPTH=32 is well UNDER the AINIC CTS match-table overflow boundary, so this must PASS cleanly.",
```
amd-arozanov: numbering jumps A4 to A6 and A12 to A14. Were A5/A13 dropped intentionally?

projects/rccl/tools/scripts/test_runner/lib/test_executor.py
```diff
@@ -1333,33 +1346,40 @@ def run_test(self, test_config, suite_config):
...
+            if custom_args:
+                program += f" {custom_args}"
+            program += gtest_out_arg
+
+            # RDMA QP/CQ creation pins memory and fails with "Cannot allocate
+            # memory" under SLURM's low inherited locked-memory soft limit, so
+            # raise it per rank before exec. `set -f` keeps gtest filter globs
+            # literal; wrapping the program (not the assembled command) means it
+            # always applies for MPI launches.
+            inner = f"ulimit -l unlimited 2>/dev/null; set -f; exec {program}"
```
amd-arozanov, line 1365: The locked-memory wrapper is only applied on the MPI path, so a single-rank RDMA test would still hit "Cannot allocate memory". Not a problem for the AINIC plan (all multi-rank), maybe just a short comment here noting the fix is MPI-only or add it for nrank==1 also?

amd-arozanov (changes requested):
Thanks! A few minor comments.

About when we should run these tests:
I think we can start with Option 1 - Nightly (check develop state) and Option 3 - per-PR: label-gated smoke subset.

But our final goal should be running these tests for each RCCL PR to avoid any regressions.  For that we may need to add more compute nodes and create a new config without long running tests

.github/workflows/ainic_test_runner.yml
nikitaxgusev: Cleaned up. De-duplicated the repeated sentences ("self-hosted submit node without local GPUs", "Group A always needs a fresh --enable-mpi-tests build", the build-strategy paragraph) so each is stated once in the header and referenced elsewhere instead of repeated.

.github/workflows/ainic_test_runner.yml
nikitaxgusev: Cleaned up. The workflow_dispatch input defaults are now empty and fall back to the `env.DEFAULT_*` values, so each value lives in exactly one place. The env defaults have to stay because schedule/push/pull_request events have no inputs, so `env` is now the single source of truth.

.github/workflows/ainic_test_runner.yml
nikitaxgusev: Changed to `warn` in 0f7892e. Not `error`: when only one group is selected (e.g. the PR smoke subset, or any `--suite-name` filter) the other group's log legitimately does not exist, so `error` would false-fail the job. `warn` still surfaces a genuinely empty upload.

projects/rccl/tools/scripts/test_runner/configs/ainic.json
nikitaxgusev: Renamed to `ainic.json` and the config `name` field to `RCCL-AINIC`. The meaning of Group A/B/C is now spelled out in the workflow header and the config `description`.

projects/rccl/tools/scripts/test_runner/lib/test_executor.py
nikitaxgusev: Fixed - the `ulimit -l unlimited` wrapper now also wraps the single-rank (non-MPI) path, so a 1-rank RDMA test raises the locked-memory limit too instead of hitting "Cannot allocate memory".

projects/rccl/tools/scripts/test_runner/configs/ainic_abc.json
nikitaxgusev: Removed `RCCL_MODEL_MATCHING_DISABLE=1` in 0f7892e. Added because of g24/25 nodes as WA.

projects/rccl/tools/scripts/test_runner/configs/ainic.json
nikitaxgusev: No wiki page yet - each suite/knob is documented inline via its `description` field in `ainic.json`.

.github/workflows/ainic_test_runner.yml
nikitaxgusev: Added to the workflow header:
- Group A - NetIbMPI + GIN component/unit gtests (rccl-UnitTestsMPI).
- Group B - NCCL_NET plugin x knobs sweep on rccl-tests perf binaries.
- Group C - IONIC hardware counter validation.

nikitaxgusev:
Thanks for the review! Addressed the inline comments. On your questions about how to run this and the rollout plan:

**How to run it (after merge, from any PR/branch):**
- **Manually:** Actions -> AINIC RCCL Test Runner -> Run workflow, pick the branch, optionally set `test_suites` / `salloc_*`. It always builds RCCL fresh from the chosen ref, so it tests that PR's actual code.
- **Locally:** same runner on a SLURM allocation: `python test_runner.py -c configs/ainic.json --system ainic` (filter with `--suite-name`).

**per your suggestion - Option 1 + Option 3**
1. **Nightly** (`schedule`, 02:00 UTC) runs the full plan against the default branch.
2. **Label-gated per-PR smoke:** adding the `ainic-test` label runs a fast subset

projects/rccl/tools/scripts/test_runner/configs/ainic.json
nikitaxgusev: Aligned

.github/workflows/ainic_test_runner.yml
```diff
@@ -0,0 +1,390 @@
...
+  # runtime via LD_LIBRARY_PATH; see the build-strategy note in the header).
+  DEFAULT_RCCL_TESTS_BIN: /it-share/rccl-ci/rccl-tests-2.30.4/bin
+  # SLURM allocation defaults (single source of truth; workflow_dispatch inputs
+  # override them when non-empty).
+  DEFAULT_SALLOC_PARTITION: amd-tw
+  DEFAULT_SALLOC_RESERVATION: rccl_dev
+  DEFAULT_SALLOC_NODES: '2'
+  DEFAULT_SALLOC_TIME: '04:00:00'
+  # Exclude known-broken GPU nodes: g24/g25
+  DEFAULT_SALLOC_EXCLUDE: 'mia1-p01-g[24-25]'
```
VadimKutovoi: These nodes should be included. NIC firmware was updated, so we should be able to run tests.

projects/rccl/tools/scripts/test_runner/configs/ainic.json
```diff
@@ -0,0 +1,417 @@
...
+    }
+  },
+
+  "test_suites": [
+    { "name": "A1-A3. Component - NetIb core",       "config": "a_netib_component", "enabled": true },
+    { "name": "A4. Component - CTS workaround (off)", "config": "a_cts_workaround", "enabled": true },
+    { "name": "A5. Component - CTS depth 32 (on, passes)", "config": "a_cts_depth32", "enabled": true },
+    { "name": "A6-A7. Component - CAST WRR",          "config": "a_cast_wrr",        "enabled": true },
+    { "name": "A8. Component - Fault injection",      "config": "a_fault_inject",    "enabled": true },
+    { "name": "A9-A13. Component - GIN proxy",        "config": "a_gin",             "enabled": true },
```
VadimKutovoi, line 400: Should we also include NIC fusion and resiliency unit tests or you are going to extend it in the scope of another PR? I think there are more tests to include.

VadimKutovoi (changes requested):

projects/rccl/tools/scripts/test_runner/configs/ainic.json
nikitaxgusev: @VadimKutovoi Yes, I plan to initiate the json file here that everybody then can use it to extend that and run more tests.

VadimKutovoi: Got it, thanks!

.github/workflows/ainic_test_runner.yml
nikitaxgusev: @VadimKutovoi fixed

VadimKutovoi (approved):

amd-arozanov (approved):

atulkulk (approved):

.github/workflows/ainic_test_runner.yml
```diff
@@ -0,0 +1,390 @@
...
+  # PR subset for the label-gated path: Group A component gtests + selected B
+  # suites (IB baseline, CAST full alltoall, autodetect on/off, ROCM-IB no
+  # offload, CTS force-override, and the full-range guard). Skips B3-B7 and C1.
+  SMOKE_SUITES: 'A*. Component*,B1.*,B2.*,B8.*,B9.*,B10.*,B11.*,B12.*'
+
+jobs:
+  ainic-test-runner:
+    # On PRs, run only when the "ainic-test" label is present. All other events
+    # (workflow_dispatch, schedule) always run.
+    if: >-
```
jayhawk-commits, line 119: This label-gated PR path runs the PR checkout on a self-hosted AINIC runner. Can we add the same same-repo/fork guard used by other self-hosted workflows? As written, a fork PR could be labeled once and then later `synchronize` events would keep running updated fork code on the self-hosted runner while the `ainic-test` label remains.

nikitaxgusev: Good catch, agreed, the label alone doesn't protect against the fork + synchronize re-run case. I'll add the same-repo guard so the self-hosted job only runs for branches from this repo, not forks:
```
if: >-
  github.event_name != 'pull_request' ||
  (github.event.pull_request.head.repo.full_name == github.repository &&
   contains(github.event.pull_request.labels.*.name, 'ainic-test'))
```
Non-PR triggers (schedule, workflow_dispatch) are unaffected. Fork PRs won't auto-run on the AINIC runner anymore; a maintainer can push the branch into the repo or use workflow_dispatch when a fork needs a run.

jayhawk-commits (approved):
Thanks, the same-repo guard addresses my concern about label-gated PR execution on the self-hosted AINIC runner. I re-reviewed the latest head and don't have further blocking comments.

PR #7930, opened 2026-06-29
feat(rocprofiler-compute): Add PC sampling source snapshots

abchoudh-amd:
## Motivation

Save source files so generated PC sampling output's source code can be inspected.

## Technical Details

- Collect deduplicated source paths from decoded PC sampling instruction comments.
- Add a source snapshotter that copies source files under the configured `src` output directory.
- Skip empty, missing, non-regular, and copy-failing source paths without failing.
- Wire source snapshotting into PC sampling finalization after code object info is written.
- Add filesystem and SDK wrapper seams for focused collector and tool tests.

JIRA ID: AIPROFCOMP-634

## Test Plan

- [X] Run ctests

## Test Result

- [x] Ctests pass.

## Submission Checklist


projects/rocprofiler-compute/src/lib/pc_sampling_collector/pc_sampling_collector.cpp
```diff
@@ -3,21 +3,78 @@
...
+
+    return !token.empty() &&
+           std::all_of(token.cbegin(),
+                       token.cend(),
+                       [](char ch) { return std::isdigit(static_cast<unsigned char>(ch)) != 0; });
+}
+
+std::string_view pc_sampling_collector_impl_t::source_frame_separator()
+{
+    static const auto sdk_wrapper = sdk_wrapper_t::create();
```
vedithal-amd: The new `sdk_wrapper_t` (sdk_wrapper.h:10-17) is a virtual interface meant for test substitution, but it isn't wired as a seam: `source_frame_separator()` builds the concrete impl via a static local here, and the collector ctor (pc_sampling_collector.h:39) takes no wrapper, so no test can inject a mock through it. This matches the interface+impl+factory pattern used elsewhere (snapshotter, translator, the tool's `SdkWrapper` in sdk_callbacks.cpp:20), so the fix is to finish the wiring: inject it through the constructor like those, and add a test that substitutes a mock separator.

projects/rocprofiler-compute/src/lib/pc_sampling_collector/source_snapshotter.cpp
```diff
@@ -0,0 +1,130 @@
...
+    const std::filesystem::path& absolute_source_path,
+    const std::filesystem::path& destination_root) const
+{
+    std::error_code error;
+    const auto canonical_source_path = m_filesystem->weakly_canonical(absolute_source_path, error);
+    if (error)
+        throw std::runtime_error("Failed to get canonical source path: " + absolute_source_path.string());
+
+    return destination_root / m_filesystem->relative_path(canonical_source_path);
+}
```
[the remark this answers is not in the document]
vedithal-amd: Agreed. This throw aborts the whole finalize on one bad path, dropping every other snapshot too. Suggest making `get_destination_path` non-throwing: log and return an empty path on `weakly_canonical` failure, then skip in the `snapshot()` loop like the other per-file checks.

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_source_snapshotter.cpp
```diff
@@ -0,0 +1,262 @@
...
+{
+    const std::filesystem::path symlink_source_path = "/linked_sources/kernel.cpp";
+    set_regular_source(symlink_source_path);
+
+    EXPECT_NO_THROW(m_snapshotter->snapshot({symlink_source_path}, m_destination_root));
+
+    ASSERT_EQ(m_filesystem->get_copy_file_calls().size(), 1);
+    EXPECT_EQ(m_filesystem->get_copy_file_calls()[0].source, symlink_source_path);
+    EXPECT_EQ(m_filesystem->get_copy_file_calls()[0].destination, destination_path(symlink_source_path));
+}
```
vedithal-amd: These symlink tests don't actually exercise symlink behavior. The mock just returns a regular/directory status for a path named `/linked_sources/...`, so this is effectively a duplicate of the regular-file / directory tests. Suggest renaming to reflect what's tested, or dropping them.

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_source_snapshotter.cpp
```diff
@@ -0,0 +1,262 @@
...
+              std::vector<std::filesystem::path>{relative_source_path});
+    EXPECT_EQ(m_filesystem->get_status_calls(), std::vector<std::filesystem::path>{absolute_source_path});
+    EXPECT_EQ(m_filesystem->get_weakly_canonical_calls(),
+              std::vector<std::filesystem::path>{absolute_source_path});
+    EXPECT_EQ(m_filesystem->get_relative_path_calls(),
+              std::vector<std::filesystem::path>{absolute_source_path});
+    ASSERT_EQ(m_filesystem->get_copy_file_calls().size(), 1);
+    EXPECT_EQ(m_filesystem->get_copy_file_calls()[0].source, relative_source_path);
+    EXPECT_EQ(m_filesystem->get_copy_file_calls()[0].destination, destination);
+}
```
vedithal-amd, line 161: No test covers the `weakly_canonical` failure path, which is the throw flagged in `get_destination_path`. The mock already supports `set_weakly_canonical_error`, so once that path is changed to skip-instead-of-throw, please add a test asserting the file is skipped and no copy happens.

projects/rocprofiler-compute/src/lib/rocprofiler_compute_tool/rocprofiler_compute_tool.cpp
```diff
@@ -195,6 +195,15 @@ static std::string generate_output_filename(std::string_view output_path, std::s
...
 }
 
+static std::string generate_output_directory(std::string_view output_path, std::string_view directory_name)
+{
+    std::string directory{output_path};
+    if (directory.back() != '/')
+        directory += '/';
+    directory.append(directory_name);
+    return directory;
+}
```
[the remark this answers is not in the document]
vedithal-amd: Not reachable in practice: `output_path` comes from `get_output_path()`, which returns the default `"./"` whenever the env var is unset or empty, so it's never empty here. The pre-existing `generate_output_filename` uses the same `.back()` pattern too. Leaving as-is.

vedithal-amd (approved):
LGTM with minor comments. Clean, well-tested change. A few non-blocking items left inline:

- `sdk_wrapper_t` interface isn't wired as an injectable seam; suggest finishing the wiring to match the existing pattern.
- `get_destination_path` throws on `weakly_canonical` failure, which aborts the whole finalize; suggest skip-instead-of-throw (agreeing with Copilot).
- Symlink tests don't actually exercise symlink behavior; rename or drop.
- Add a test for the `weakly_canonical` error path once the throw is changed.

Built locally and both affected unit suites pass (test-pc-sampling-collector: 58, test-rocprofiler-compute-tool: 64).

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/mocks.cpp
```diff
@@ -121,3 +143,179 @@ uint32_t mock_code_object_writer_t::get_end_symbol_count() const
 {
     return m_end_symbol_count;
 }
+
+std::string_view mock_sdk_wrapper_t::source_frame_separator()
```
feizheng10: const

projects/rocprofiler-compute/src/lib/rocprofiler_compute_tool/rocprofiler_compute_tool.cpp
```diff
@@ -197,6 +197,15 @@ static std::string generate_output_filename(std::string_view output_path, std::s
     return filename;
 }
 
+static std::string generate_output_directory(std::string_view output_path, std::string_view directory_name)
```
feizheng10, line 196: Why not using fs::path::operator/ ?

feizheng10 (approved):

feizheng10:
Please fix those 2 minor comments and make sure CI passed.


And some suggestions from my agent:

[SUGGESTION] Untested edge case: frame with trailing colon ("kernel.cpp:")
In path_from_source_frame, a frame like "kernel.cpp:" returns the whole string including the colon as the path. This fails lookup gracefully (skipped as missing), but a test documenting this behavior would help future maintainers.

[SUGGESTION] m_source_paths accumulates across multiple finalize() calls
m_source_paths is a member field with no reset. If finalize() were ever called twice, paths from the first run would persist. Consider documenting one-shot semantics or using a local accumulator in finalize().

vedithal-amd:
> Please fix those 2 minor comments and make sure CI passed.
> 
> And some suggestions from my agent:
> 
> [SUGGESTION] Untested edge case: frame with trailing colon ("kernel.cpp:") In path_from_source_frame, a frame like "kernel.cpp:" returns the whole string including the colon as the path. This fails lookup gracefully (skipped as missing), but a test documenting this behavior would help future maintainers.
> 
> [SUGGESTION] m_source_paths accumulates across multiple finalize() calls m_source_paths is a member field with no reset. If finalize() were ever called twice, paths from the first run would persist. Consider documenting one-shot semantics or using a local accumulator in finalize().

@abchoudh-amd I am not sure these two hypothetical cases will ever occur, and these seem like nitpicks, please feel free to pushback.

kernel.cpp: ending in colon and not followed by > or line number is a problem in SDK since it does not mean anything

finalize is supposed to be called only once per tool

abchoudh-amd:
> Please fix those 2 minor comments and make sure CI passed.
> 
> And some suggestions from my agent:
> 
> [SUGGESTION] Untested edge case: frame with trailing colon ("kernel.cpp:") In path_from_source_frame, a frame like "kernel.cpp:" returns the whole string including the colon as the path. This fails lookup gracefully (skipped as missing), but a test documenting this behavior would help future maintainers.
> 
> [SUGGESTION] m_source_paths accumulates across multiple finalize() calls m_source_paths is a member field with no reset. If finalize() were ever called twice, paths from the first run would persist. Consider documenting one-shot semantics or using a local accumulator in finalize().

Added a test for suggestion 1 for completeness.

finalize is supposed to be called once and is evident from the name, so documentation is redundant,

PR #7931, opened 2026-06-29
feat: Python API to rocprof-trace-decoder

ApoKalipse-V:
## Motivation

Initial python API to rocprof-trace-decoder.
* Integration tests converted to python in the name of simplicity.
* Samples added

JIRA ID : AIPROFSDK-928

TheRock PR: https://github.com/ROCm/TheRock/pull/6365

## Test Plan

All integration tests converted to use the python att_tool.py

## Test Result

All tests pass

## Submission Checklist


projects/rocprof-trace-decoder/python/rocprof_trace_decoder/bindings.py
```diff
@@ -0,0 +1,728 @@
...
+        isa: IsaProvider | None = None,
+        on_batch: Callable[[RecordType, list[object] | int], None] | None = None,
+    ) -> TraceRecords:
+        out = TraceRecords()
+        for chunk in chunks:
+            partial = self.parse(chunk, isa=isa, on_batch=on_batch)
+            merge_records(out, partial)
+        return out
+
+    def parse(
```
feizheng10, line 512: _callbacks is only cleared in close() ([bindings.py:276]). A long-lived Decoder that parses many files — e.g. csv_test.py loops over all .att files on one decoder, and att.py reuses one decoder per run — accumulates 1–2 CFUNCTYPE wrappers per parse. Memory growth is bounded per wrapper size but unbounded in count.
Please replace list append with per-parse slots (self._isa_cb, self._trace_cb) or overwrite the list each parse. One live reference per callback type is enough to prevent GC during the C call.

feizheng10 (dismissed):

projects/rocprof-trace-decoder/python/rocprof_trace_decoder/code_index.py
```diff
@@ -0,0 +1,272 @@
...
+
+    def isa_for_pc(self, pc: Pc) -> tuple[str, int] | None:
+        entry = self.entries.get(pc)
+        if entry is None:
+            return None
+        return entry.inst, entry.memory_size
+
+    def get_or_create(self, pc: Pc) -> CodeEntry:
+        entry = self.entries.get(pc)
+        if entry is not None:
```
feizheng10, line 150: _set_memory_sizes() sorts all entries per code object on every new PC. accumulate_wave() calls get_or_create() per instruction ([code_index.py:133]), so a trace with N unique PCs does O(N²·log N) work.

Suggestion: Defer with a dirty flag; rebuild only in isa_for_pc() or at write time.

projects/rocprof-trace-decoder/python/rocprof_trace_decoder/codegen.py
```diff
@@ -0,0 +1,589 @@
...
+                    if shndx not in ("SHN_UNDEF", "SHN_ABS"):
+                        labels.setdefault(vaddr, [])
+                        if name not in labels[vaddr]:
+                            labels[vaddr].append(name)
+    return functions, labels
+
+
+def parse_disassembly(elf_path: str) -> list[tuple[int, str]]:
+    """Return [(vaddr, instruction_text)] from llvm-objdump output."""
+    proc = subprocess.run(
```
feizheng10, line 483: Should we have a timeout here?

projects/rocprof-trace-decoder/python/rocprof_trace_decoder/codegen.py
```diff
@@ -0,0 +1,589 @@
...
+
+    with Path(elf_path).open("rb") as f:
+        try:
+            elf = ELFFile(f)
+            if not elf.has_dwarf_info():
+                return []
+            dwarfinfo = elf.get_dwarf_info()
+            cu_iter = iter(dwarfinfo.iter_CUs())
+        except Exception:
+            return []
```
feizheng10, line 321: Should we catch expected errors and log other errors?

projects/rocprof-trace-decoder/python/rocprof_trace_decoder/code_index.py
```diff
@@ -0,0 +1,272 @@
...
+            inst="",
+            line_number=max(self.line_numbers.values(), default=0) + 1,
+        )
+        self.entries[pc] = entry
+        self.line_numbers[pc] = entry.line_number
+        self._set_memory_sizes()
+        return entry
+
+    def line_number(self, pc: Pc) -> int:
+        return self.entries[pc].line_number
```
feizheng10, line 168: return int | None or document the KeyError?

projects/rocprof-trace-decoder/python/rocprof_trace_decoder/codegen.py
ApoKalipse-V: I think not? Should llvm-objdump hangs be a concern?

projects/rocprof-trace-decoder/python/rocprof_trace_decoder/code_index.py
```diff
@@ -257,6 +277,12 @@ def _is_code_label(inst: str) -> bool:
     return inst.startswith(";") or inst.startswith("label")
 
 
+def _clear_document_counts(document: dict) -> None:
+    for row in document.get("code", []):
+        if len(row) >= 10:
+            row[6:10] = [0, 0, 0, 0]
```
ApoKalipse-V: TODO: Remove these hardcoded indices

projects/rocprof-trace-decoder/python/rocprof_trace_decoder/code_index.py
ApoKalipse-V: Thanks. AI liked the dirty option.

I'm considering adding the sizes to code.json so we don't have to do this.

projects/rocprof-trace-decoder/python/rocprof_trace_decoder/bindings.py
```diff
@@ -0,0 +1,727 @@
...
```
meserve-amd, line 512: Nit: add a docstring for this

projects/rocprof-trace-decoder/python/rocprof_trace_decoder/codegen.py
```diff
@@ -0,0 +1,632 @@
...
+                f"{seen_ids[code_object_id]} and {path}."
+            )
+        seen_ids[code_object_id] = path
+        out.append(CodeObject(path, code_object_id))
+    if not out:
+        raise ValueError("No code objects were provided.")
+    return out
+
+
+def generate_code_artifacts(code_objects: Iterable[CodeObject]) -> CodeArtifacts:
```
meserve-amd, line 560: Nit: Add docstring, maybe comments too

projects/rocprof-trace-decoder/python/rocprof_trace_decoder/codegen.py
```diff
@@ -0,0 +1,632 @@
...
+                    if shndx not in ("SHN_UNDEF", "SHN_ABS"):
+                        labels.setdefault(vaddr, [])
+                        if name not in labels[vaddr]:
+                            labels[vaddr].append(name)
+    return functions, labels
+
+
+def parse_disassembly(elf_path: str) -> list[tuple[int, str]]:
+    """Return [(vaddr, instruction_text)] from llvm-objdump output."""
+    proc = subprocess.run(
```
meserve-amd, line 483: Minor: We discard proc.stderr here, could be useful to capture for errors

projects/rocprof-trace-decoder/python/rocprof_trace_decoder/code_index.py
```diff
@@ -0,0 +1,310 @@
...
+
+    def write_code_json(self, path: str | Path) -> None:
+        out_doc = dict(self.document)
+        rows = []
+        seen: set[Pc] = set()
+        for row in self.document.get("code", []):
+            if len(row) < 10 or _is_code_label(str(row[0])):
+                rows.append(row)
+                continue
+            pc = Pc(address=int(row[5]), code_object_id=int(row[4]))
```
meserve-amd: Minor: Should this use constants defined at the top of this file? (and similar code in write_stats_csv)

projects/rocprof-trace-decoder/python/rocprof_trace_decoder/bindings.py
```diff
@@ -0,0 +1,727 @@
...
+            ptr = ctypes.cast(events, ctypes.POINTER(_OtherSimdInstruction))
+            return [_convert_other_simd(ptr[i]) for i in range(size)]
+        if record_type == RecordType.DISPATCH:
+            ptr = ctypes.cast(events, ctypes.POINTER(_Dispatch))
+            return [_convert_dispatch(ptr[i]) for i in range(size)]
+        return []
+
+
+def _convert_perf(c: _PerfEvent) -> PerfEvent:
+    return PerfEvent(c.time, c.events0, c.events1, c.events2, c.events3, c.CU, c.bank)
```
meserve-amd: Minor: Other _convert functions use named parameters, but this, _convert_shaderdata, and _convert_realtime don't

meserve-amd (approved):
I think this is pretty good overall. Some more comments in the API code would improve readability greatly, but I wouldn't hold the change over it.

.github/workflows/rocprof-trace-decoder-continuous_integration.yml
```diff
@@ -128,7 +129,7 @@ jobs:
         cmake --build build -j $(nproc)
 
     - name: Test
-      timeout-minutes: 15
+      timeout-minutes: 45
```
jbonnell-amd: How long do the runs usually take? I noticed the run in the PR took around 9 minutes in total, so curious why we're going from a 15-minute to a 45-minute timeout.

Same applies for the other timeout changes - if we're not approaching the current timeouts, we should keep them as is.

jbonnell-amd (approved):
Approved for `.github/workflows` changes

feizheng10:
Some minor comments:

1. Decoder.close() marks the handle closed before verifying destroy succeeded
[bindings.py:326-333]


bindings.py
Lines 326-333
    def close(self) -> None:
        if self._closed:
            return
        status = self._lib.rocprof_trace_decoder_destroy_handle(self._handle)
        self._closed = True
        self._isa_cb = None
        self._trace_cb = None
        self._check(status)
If rocprof_trace_decoder_destroy_handle returns a non-success status, _closed is already True and callbacks are cleared, so callers cannot retry or distinguish a half-torn-down handle from a clean one. Consider calling _check(status) first, or only setting _closed = True after a successful destroy.

2. load_code_object* can pass inconsistent load_size vs actual bytes
[bindings.py:469-474]

The native call receives both caller-supplied load_size and len(blob), with no validation that they match:

status = self._lib.rocprof_trace_decoder_codeobj_load(
    self._handle, load_id, load_addr, load_size, buf, len(blob)
)
That is an easy footgun for API users. Either validate (and raise DecoderError / ValueError), or document clearly that load_size must equal the buffer length.

3. Code-object ID inference is duplicated with slightly different logic
[att_tool.py:187-213], [generate_code.py:34-51], [samples/common.py:119-148]

The same _code_objects_from_paths logic appears in three places. samples/common.py additionally recognizes code_object_id_<N> via regex; att_tool.py and generate_code.py only use stem.rfind("_"). Over time this can diverge (as it already has). Worth extracting one shared helper used by CLI, samples, and generate_code.py.

feizheng10 (approved):

PR #7936, opened 2026-06-29
ci(hip-tests): Add option to enforce label settings for new hip-tests

litvaOo:
## Motivation
Enforcement of the label's creation for new hip-tests was disabled by default by https://github.com/ROCm/rocm-systems/pull/6018/changes. This PR adds environment variable to enforce it in CI. 

## Technical Details
Adds a new variable, THEROCK_REQUIRE_HIP_YAML_ENTRIES, which will enforce all hip-tests have an entry in the yaml configs, with at minimum test level label set.

## JIRA ID
NA

## Test Plan
- Default build should not be impacted
- Build should fail if any test is added to the *.cpp files, but not present in hip-tests config

## Test Result
- Waiting for PSDB with new labels
- The tests without labels were added to develop already, see https://github.com/ROCm/rocm-systems/actions/runs/28375497488/job/84064154510?pr=7936#step:12:302

## Submission Checklist

### DIFF
--- a/.github/workflows/therock-ci-linux.yml
+++ b/.github/workflows/therock-ci-linux.yml
@@ -41,6 +41,7 @@ jobs:
       AMDGPU_FAMILIES: ${{ inputs.amdgpu_families }}
       TEATIME_FORCE_INTERACTIVE: 0
       AWS_SHARED_CREDENTIALS_FILE: /home/awsconfig/credentials.ini
+      THEROCK_REQUIRE_HIP_YAML_ENTRIES: 1
     steps:
       - name: "Checking out repository for rocm-systems"
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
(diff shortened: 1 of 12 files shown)


cjatin (approved):
hip-tests part looks good, need someone else to validate the ci part.

.github/workflows/therock-ci-linux.yml
```diff
@@ -41,6 +41,7 @@ jobs:
       AMDGPU_FAMILIES: ${{ inputs.amdgpu_families }}
       TEATIME_FORCE_INTERACTIVE: 0
       AWS_SHARED_CREDENTIALS_FILE: /home/awsconfig/credentials.ini
+      THEROCK_REQUIRE_HIP_YAML_ENTRIES: 1
```
geomin12, line 44: is this env variable meant to be for tests? or during build?

geomin12 (approved):

.github/workflows/therock-ci-linux.yml
litvaOo: It's only required in build, test scripts do not consume this variable

PR #7937, opened 2026-06-29
[kpack] Pin relocated PHDR vaddr to file offset for old-kernel exec compat (#7681)

marbre:
Cherry-picks 872b4f6 into the 7.14 release branch.

---

## Motivation

Fixes https://github.com/ROCm/TheRock/issues/5430

kpack-split ROCm **executables** SIGSEGV in the dynamic loader before `main` when run directly on RHEL/Rocky 8.x (kernel 4.18) baremetal. `test_hip_api`/`clang-offload-bundler` crash with no output, while `/lib64/ld-linux-x86-64.so.2 <binary>` works.

## Technical Details

When kpack relocates the program header table to a trailing `PT_LOAD`, that segment's `p_vaddr` is only page-*congruent* with its file offset (`p_vaddr != p_offset`). Kernel 4.18 computes `AT_PHDR = load_bias + e_phoff` **without** translating `e_phoff` through the covering segment, so `ld.so` is handed an unmapped address and crashes before `main`. Newer kernels (>= 5.x) translate correctly, so it only reproduces on EL8.

Fix: a dedicated `normalize_phdr_vaddr()` pass (in `elf/phdr_manager.py`), called as the **last** step of `kpack_offload_binary` after all size-changing phases, pins the relocated PHDR `PT_LOAD`/`PT_PHDR` to `p_vaddr == p_offset`.

- **Executables only** (gated on `PT_INTERP`). Shared libraries are always mapped by `ld.so`, which handles a relocated PHDR correctly, so they are left untouched — avoiding needless file growth.
- **No size change for real binaries:** when the table already clears every other segment's address range (the case for shipped binaries, whose host image dwarfs the residual stub `.hip_fatbin`), this is a pure metadata edit.
- It only moves the table to a fresh end-of-file offset (growing the file) when a *fully zero-paged* binary's file has shrunk below the address ceiling — which in practice is just the small synthetic test fixtures, not shipped binaries.

Single chokepoint, so it covers both the wheel and tarball paths.

## Test Plan

- Reproduced on Rocky Linux 8.10, kernel 4.18.0-553 — both a GPU-free KVM VM and RHEL 8.10 baremetal with a gfx1100.
- Applied the resulting `p_vaddr == p_offset` layout to a real kpack-split `test_hip_api` and re-ran on kernel 4.18 (direct exec and via `ld.so`).
- Added a regression test (`TestFullPipelineAlignment::test_relocated_phdr_vaddr_equals_offset`) asserting the relocated PHDR `PT_LOAD` satisfies `p_vaddr == p_offset` after the full pipeline.

## Test Result

- **Before:** direct exec -> SIGSEGV (exit 139) in `ld-2.28.so` before `main`.
- **After:** direct exec runs normally. On a real RHEL 8.10 gfx1100 box, `test_hip_api` went from segfault to a full pass (`[ OK ] HIPTests.Saxpy`, exit 0), matching the `ld-linux` workaround.
- The normalization is verified (via kpack's own loader) to be a zero-byte metadata edit on real binaries, idempotent, and a no-op on shared libraries and on non-relocated binaries.
- CI: Python tests (ubuntu + windows, 3.10/3.12) and C++ runtime tests pass.

### Test changes

- Two `.exe`-fixture size assertions (`test_kpack_fat_binary`, `test_unbundling_visitor`) were relaxed: a tiny *fully zero-paged* executable can grow when its PHDR is normalized for old-kernel exec, so they now assert validity + device-code removal (`.hip_fatbin` -> NOBITS) instead of raw shrink. Real binaries are unaffected.
- The shared-library size test (`test_artifact_with_fat_binaries`) is unchanged and still passes, since libraries are no longer touched.

## Submission Checklist

---------

## Submission Checklist

### DIFF
--- a/shared/kpack/python/rocm_kpack/elf/__init__.py
+++ b/shared/kpack/python/rocm_kpack/elf/__init__.py
@@ -31,6 +31,7 @@ from .phdr_manager import (
     ProgramHeaderManager,
     PhdrResizeResult,
     create_load_segment,
+    normalize_phdr_vaddr,
 )
 from .operations import (
     map_section_to_load,
@@ -146,6 +147,7 @@ __all__ = [
     "SetPointerResult",
     "UpdateRelocationResult",
     "create_load_segment",
+    "normalize_phdr_vaddr",
     # Zero-page operations
     "conservative_zero_page",
     "zero_page_section",
--- a/shared/kpack/python/rocm_kpack/elf/kpack_transform.py
+++ b/shared/kpack/python/rocm_kpack/elf/kpack_transform.py
@@ -21,6 +21,7 @@ import msgpack
 from ..kpack_transform import NotFatBinaryError
 from .surgery import ElfSurgery, SectionInfo, AddSectionResult
 from .operations import map_section_to_load, set_pointer
+from .phdr_manager import normalize_phdr_vaddr
 from .verify import ElfVerifier
 from .zero_page import conservative_zero_page
 from .types import SHT_PROGBITS
@@ -383,6 +384,10 @@ def kpack_offload_binary(
                 "\nPhases 2-3: No .hip_fatbin section, skipping semantic "
                 "transformation and zero-page"
             )
+
+    # Pin the relocated PHDR table to p_vaddr == p_offset so the binary execs
+    # on old kernels (e.g. EL8 4.18). Must run after all size-changing phases.
+    normalize_phdr_vaddr(surgery)
 
     # Post-transform verification: catch structural corruption at build time
     verify_result = ElfVerifier.verify_data(surgery.data)
--- a/shared/kpack/python/rocm_kpack/elf/phdr_manager.py
+++ b/shared/kpack/python/rocm_kpack/elf/phdr_manager.py
@@ -13,6 +13,9 @@ Key invariants:
 - (p_offset % PAGE_SIZE) == (p_vaddr % PAGE_SIZE) for all PT_LOAD segments
 - PT_PHDR (if present) must point to the program header table location
 - The program header table must be covered by a PT_LOAD segment
+
+See normalize_phdr_vaddr() for the additional p_vaddr == p_offset constraint
+that old kernels (e.g. EL8 4.18) require on the relocated PHDR segment.
 """
 
 from dataclasses import dataclass
@@ -24,6 +27,7 @@ from .types import (
     PAGE_SIZE,
     PT_LOAD,
     PT_PHDR,
+    PT_INTERP,
     PF_R,
     round_up_to_page,
     page_align_offset,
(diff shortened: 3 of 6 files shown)


marbre:
@amd-hsivasun as you're still gardener today and @arvindcheru, you are for rocm-systems starting tomorrow, please confirm that the CI failures are pre-existing as both of you have more context.

ScottTodd (commented):
This cherry-pick is borderline in my opinion.

* We don't yet have a nightly release that includes this commit
  * https://github.com/ROCm/rockrel/actions/runs/28341070120 from last night used https://github.com/ROCm/rocm-systems/commit/cd52904912a17fa006bcebddc3ac7e4e32274ab1 from 2026-06-25
* We don't yet even have this commit on `main` in TheRock (rocm-systems submodule updates are running a few days behind)
  * Current `main` was last updated to https://github.com/ROCm/rocm-systems/commit/72822631d427f33ab85fdb7f54689b2092df6413 in https://github.com/ROCm/TheRock/commit/dbeeb2a981c92441638cad96076079c56b5a3368 (from 12 hours before the commit this is cherry-picking)
* While this commit did add new unit tests, we don't have an automated test (unit or integration) for the symptom https://github.com/ROCm/TheRock/issues/5430 where simple programs segfault on RHEL 8.10 (see https://github.com/ROCm/TheRock/issues/6065 as one idea)

To approve this cherry-pick I would want to see all of these items:
1. Confirmation of the severity of the issue this fixes
    * Is it worth being a release blocker?
    * Is it a regression compared to prior releases? I presume it was there in 7.13 multi-arch releases (which used kpack) but not 7.13 per-family releases (which did NOT use kpack)
2. Strong evidence from manual testing that this sufficiently fixes the issue (or ideally a commitment to adding automated testing)
3. Strong evidence from automated testing that this does not cause other issues (e.g. existing ROCm and PyTorch tests do not regress when this and other rocm-systems commits are included in TheRock)

marbre:
Agreed. If we want to see it in nightlies, we need https://github.com/ROCm/TheRock/pull/6216 to go into main first.

arvindcheru:
> @amd-hsivasun as you're still gardener today and @arvindcheru, you are for rocm-systems starting tomorrow, please confirm that the CI failures are pre-existing as both of you have more context.

### CI triage Details Learned

Reviewed CI on #7937 vs `release/therock-7.14` @ `a5cc428`. (a5cc428 = “release branch before/with merge target”
dce4965 = “release branch + this PR’s one commit”)

**PR scope:** 6 files, all under `shared/kpack/` (phdr normalization + tests).

**Checks directly covering this change:** all green
- Python tests (ubuntu/windows, 3.10/3.12)
- C++ runtime tests (ubuntu/windows)

**Failing on PR head (5):**
- TheRock CI Summary
- rocprofiler-sdk, rocprofiler-compute (shards 1–2), rocprofiler-systems

**Base branch (`release/therock-7.14`) already failing (3):**
- TheRock CI Summary
- hip-tests shard 3/4 (Linux gfx94X)
- hip-tests (ROCR) shard 3/4 (Windows gfx1151)

**Comparison:** `TheRock CI Summary` is red on both PR and base. Profiler jobs fail on the PR run but pass on the last base run; hip-tests shard 3 fails on base but passes on the PR run — inconsistent sub-job failures, no path overlap with `shared/kpack/`.

**Conclusion:** CI failures appear **pre-existing / flaky on the release branch**, not introduced by this kpack cherry-pick. Not blocking on PR logic from a CI-triage perspective.

**Workflow run (PR):** https://github.com/ROCm/rocm-systems/actions/runs/28375918394

Also Note:
**Log review (rocprofiler-sdk job):** Root failure is
`tests.integration.execute.openmp-tools` — OpenMP integration data mismatch
(`c_i[55424](0) != validate_i[55424](-1222981246)`, 64 validation failures).
9 dependent openmp-tools validation tests did not run. 99% of ctest passed (866/876).
No references to `phdr_manager`, `normalize_phdr`, or `shared/kpack` code in the
failure output (only kpack-named CI artifacts). **Not introduced by this cherry-pick.**

astyrrian1 (approved):

ScottTodd (approved):
We'll get some automated test signal once this gets pulled into TheRock on the release branch there. I think that, together with the manual test coverage, will have to be enough for now.

PR #7760, opened 2026-06-24
Fix device side memory faults with Asan

fabiomestre:
## Motivation

Building HIP tests with ASan produced device-side memory faults: HIPRTC device code wasn't compiled with ASan

## Technical Details

* hiprtcInternal.cpp: pass -fsanitize=address to HIPRTC options when clang has the ASan feature enabled.
* Pass -fsanitize=address to HIPRTC preprocessed builtins object.
* catch/CMakeLists.txt: keep ASan device flags for GPU targets; add a SPIR-V-only flag set with ASan stripped (unsupported on SPIR-V), used when ENABLE_SPIRV.
* module/CMakeLists.txt: build the addKernel.spv/addKernel-bundle.spv targets with the SPIR-V flag set.
* rtc.yaml: disable Unit_hiprtc_devicemalloc under ASan (COMGr mishandles ASan flags, ROCM-27071).
* hipMemcpyPeerAsync.cc: sync stream/device before disabling peer access to avoid a use-after-free under ASan.
* Excludes the `fini` and `init` internal kernels from `hipLibraryGetKernelCount`
* Disable RTC tests that are failing due to assertion in COMGr

## JIRA ID

ROCM-26928

## Test Plan

Tested locally with Asan.

## Test Result

Tests pass locally. New tickets opened for new RTC sanitizer issues.

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #7938, opened 2026-06-29
[hip-tests] [ASAN] fix hiprtc kernel count when asan is enabled

cjatin:
## Motivation

Fix the count of kernels in library apis

## Technical Details

When user enables ASAN, we might get incorrect kernel count since compiler embeds a few kernel to help with its init/fini.
This takes care of it.

## JIRA ID

Resolves ROCM-27236

## Test Plan

These are existing tests, which fail with ASAN. Retest with ASAN.

## Test Result

test passes although one of them is failing, awaiting inputs from compiler team.

## Submission Checklist


cjatin:
Pushed a commit to https://github.com/ROCm/rocm-systems/pull/7760 which should sort this.

PR #7951, opened 2026-06-29
build(hipfile): Remove `fetch-depth: 0` where unnecessary

derobins:
## Motivation

A comment in an earlier PR noticed that we were using `fetch-depth: 0` in GitHub actions that don't inspect the git history. That level of fetch-depth pulls in the entire history of rocm-systems, which is enormous.

## Technical Details

Remove `fetch-depth: 0` from these actions, so they default to a sparse checkout:

* hipfile-build-nvidia.yml
* hipfile-build.yml
* hipfile-ci-system.yml
* hipfile-image-build.yml
* hipfile-image-update.yml
* hipfile-python-bindings.yml
* hipfile-sanitizers.yml

## JIRA ID

AIHIPFILE-232

## Test Plan

CI Passes

## Test Result

Success

## Submission Checklist

### DIFF
--- a/.github/workflows/hipfile-build-nvidia.yml
+++ b/.github/workflows/hipfile-build-nvidia.yml
@@ -45,7 +45,6 @@ jobs:
       - name: Fetching code repository...
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
-          fetch-depth: 0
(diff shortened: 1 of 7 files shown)


john00003 (approved):
Not the most familiar with the workflows and if they would use the Git history anywhere, but LGTM!

zbyrne (approved):

PR #7953, opened 2026-06-29
[rocjitsu][nfc] replace hardcoded constants with generated constants

Groverkss:
Currently we have various places where we are using hardcoded opcodes. This PR changes it to use generated opcodes, which makes the code more readable and more robust.


kuhar (approved):
This seems like a very good idea to make the codebase more readable without agent help

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -163,6 +166,216 @@ def __init__(
...
+
+        if not constants:
+            return None
+
+        lines = [
+            'namespace encoding {',
+            '',
+            '/// @brief Primary decode selector constants generated from the ISA XML.',
+            '///',
+            '/// These values match Instruction::encoding_id(), which is word0 >> 23.',
```
kuhar, line 303: We should make this generated comment explicit that `encoding::k*` values are decode selectors (`word0 >> 23`), not necessarily the narrower machine-inst encoding field. SOP1/SOPP happen to coincide, but formats like DS/MUBUF/SMEM/SOP2 do not; a one-line warning would keep future handwritten code from assigning `encoding::kDs` or `encoding::kMubuf` into the wrong bitfield.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instruction_builder.h
```diff
@@ -100,65 +111,184 @@ inline constexpr uint16_t kDelayAluSaluDep1 = 9;
...
+    return rdna4::kSDelayAluSopp;
+  case ROCJITSU_CODE_ARCH_GFX1250:
+    return gfx1250::kSDelayAluSopp;
+  default:
+    return rdna4::kSDelayAluSopp;
+  }
+}
+
+/// @brief Get the s_mov_b32 opcode for a target ISA.
+[[nodiscard]] inline constexpr uint32_t sop1_op_mov_b32(rj_code_arch_t arch) {
```
kuhar, line 272: We should add a small regression test for the RDNA1/RDNA2 `build_s_mov_b32` opcode now that this helper is arch-aware. The generated headers show `rdna1::kSMovB32Sop1` / `rdna2::kSMovB32Sop1` are `3`, while the old hardcoded builder emitted `0`; since RDNA1/RDNA2 are accepted RocJITsu target arches, this is a real behavior change worth pinning.

kuhar (commented):
It would be nice add some focused tests for the new opcode/encoding generator before relying on the checked-in headers. The new pure helpers decide alias emission, collision errors, `Hi`/`OpHi` expansion, and name fragmenting, but none of that logic has direct coverage today.

A small Python test in the existing `test_generator_profile_gates.py` style could cover `gen_opcode_constants`, `_opcode_name_fragment`, `_encoding_constants_block`, and the `ValueError` paths. It would also be worth pinning one representative regenerated `opcodes.h` / `encodings.h` case so generator/header drift is caught before the next regen.

atgutier (commented):
This is a great improvement. Just some agent review comments:

 MAJOR — must address before merge

  build_s_mov_b32 silently gains arch-dependent behavior — instruction_builder.h:222,341. The base
  version always emitted opcode 0; the new version calls sop1_op_mov_b32(arch), which returns 3 for
  RDNA1/RDNA2 (0 elsewhere). So for those two targets the encoded opcode changes 0→3 — a real behavior
   change in a PR labeled NFC. It's latent (every current caller passes RDNA4, so no live regression),
   and opcode 3 is actually correct per the ISA spec (0 was a pre-existing latent bug). But it
  undermines blind trust in the "nfc" label. Fix: split it into a separate non-NFC commit (or annotate
   the PR) calling out that it fixes RDNA1/2 s_mov_b32, so reviewers don't assume every change is
  behavior-preserving.

  MINOR

  - Redundant per-arch ternary with identical arms — cdna4_to_rdna_common.cpp:55-60. Selects rdna3::
  vs rdna4::kVAddCoU32Vop3SdstEnc by host arch, but both equal 768 (and the CI variant both 288).
  NFC-correct but reads as if RDNA3/4 differ when they don't; collapse to one constant or comment it
  as future-proofing.
  - op & 0xFF on a 7-bit MUBUF field — cdna4_to_cdna3.cpp:100 + test. The MUBUF op bitfield is :7; a
  value ≥128 would pass the mask then truncate on assignment. All live values (20/22/23) are <128 so
  NFC holds, but & 0x7F would match the encoding honestly.

  NIT

  - kSop2EncodingPrefix = 0x2 left as a raw literal while the SOPP/SOP1 prefixes were migrated —
  inconsistent (though SOP2's 2-bit prefix may not map to the same generated shape; verify before
  swapping).
  - Codegen not live-verified — the reviewer confirmed the generated files are generator-shaped
  (DO-NOT-EDIT banner, correct guards/namespaces, inline constexpr uint16_t, zero duplicate names
  across all 10 opcodes.h/encodings.h) and that values match ISA-spec ground truth, but could not run
  the generator against this commit (working tree is a different branch; the tmp/ regen scripts aren't
   present here), so byte-for-byte reproducibility is reasoning-based, not proven. Recommend the
  author confirm git diff is empty after a clean regen.
  - Generator determinism is fine (insertion-ordered defaultdict) but a sorted() or comment would make
   the stability guarantee explicit against future refactors.

PR #7955, opened 2026-06-29
feat(rocSHMEM): Add unaligned fallback for device memcpy helper

amirsojoodi:
## Motivation

This PR adds an unaligned fallback path for the device memcpy helpers in `src/util.hpp`. The existing 16-byte chunk path is preserved for aligned operands, while unaligned operands are copied through the fallback chain. 

## Technical Details

The memcpy helpers use 16-byte chunks for efficient device copies. If either operand is not aligned for that access size, the vectorized load/store can fault and appear as an illegal memory access. The fallback avoids issuing those 16-byte accesses for unaligned copies

## JIRA ID

JIRA ID: AIROCSHMEM-445

## Test Result

- Unit tests and functional tests pass on gfx950 and gfx942
- DeepEP works fine
- Although the all-to-all test should not produce unaligned addresses in the first place, but the illegal memory access causing from that function is now fixed as a side effect of this PR.

### DIFF
--- a/projects/rocshmem/src/util.hpp
+++ b/projects/rocshmem/src/util.hpp
@@ -531,15 +531,6 @@ __device__ __forceinline__ void copy_remainder(uint8_t* dst, uint8_t* src, int s
   }
 }
 
-// Prefix/tail bytes (< ChunkSize) are copied via copy_bulk<1, ...> rather
-// than a single tid==0 lane running copy_remainder: with group-cooperative
-// callers (stride > 1), gating a multi-step byte copy behind one lane forces
-// every other active lane to sit masked-off through the branch until
-// reconvergence -- paid on *every* call. Routing it through copy_bulk<1>
-// spreads the (at most ChunkSize-1) leftover bytes one-per-lane across
-// tid/stride instead, so it costs a single parallel step. For memcpy_lane
-// (tid=0, stride=1) this degenerates back to the same serial byte-by-byte
-// copy it already did.
 template <int ChunkSize, CachePolicy LP, CachePolicy SP, int Unroll,
           bool UniformBase = false>
 __device__ __forceinline__ void copy_aligned_body(uint8_t* dst, uint8_t* src,
@@ -578,8 +569,6 @@ __device__ __forceinline__ void copy_aligned_body(uint8_t* dst, uint8_t* src,
 // LANE, WAVE, AND WG IMPLEMENTATIONS
 // ==============================================================================
 
-// Blocking variants additionally drain all in-flight VMEM ops before returning.
-//
 // memcpy_lane is per-lane: many lanes may call it concurrently, each with its
 // own independent dst/src (see e.g. tile_put's element-by-element fallback).
 // It must NOT set UniformBase=true on copy_aligned_best_effort/copy_bulk --
@@ -627,7 +616,7 @@ template <MemcpyKind Kind = MemcpyKind::Put>
 
 // memcpy_wave is group-cooperative: every lane in the wave passes the same
 // dst/src and only differs by wave_tid/wave_size (folded into the per-access
-// offset inside copy_bulk), so UniformBase=true is safe here.
+// offset inside copy_bulk), so UniformBase=true
 template <MemcpyKind Kind = MemcpyKind::Put>
 [[maybe_unused]] __device__ __forceinline__ void memcpy_wave(void* dst, void* src, size_t size) {
   if (size == 0) return;
@@ -645,7 +634,7 @@ template <MemcpyKind Kind = MemcpyKind::Put>
                                                 static_cast<uint8_t*>(src),
                                                 size, wave_tid, wave_size);
   } else {
-    // The "Whatever Else" Path: Small sizes skip bulk setup entirely
+    // Small sizes skip bulk setup entirely
     if (wave_tid == 0) {
       copy_remainder<LP, SP>(static_cast<uint8_t*>(dst),
                              static_cast<uint8_t*>(src),
@@ -656,7 +645,7 @@ template <MemcpyKind Kind = MemcpyKind::Put>
 
 // memcpy_wg is group-cooperative: every lane in the workgroup passes the same
 // dst/src and only differs by tid/stride (folded into the per-access offset
-// inside copy_bulk), so UniformBase=true is safe here.
+// inside copy_bulk), so UniformBase=true 
 template <MemcpyKind Kind = MemcpyKind::Put>
 [[maybe_unused]] __device__ __forceinline__ void memcpy_wg(void* dst, void* src, size_t size) {
   if (size == 0) return;
--- a/projects/rocshmem/src/util.hpp
+++ b/projects/rocshmem/src/util.hpp
@@ -400,43 +400,17 @@ constexpr bool is_blocking(MemcpyKind k) {
   return k == MemcpyKind::PutBlocking || k == MemcpyKind::GetBlocking;
 }
 
-// UniformBase: true iff every active lane calling this copy_bulk instance is
-// guaranteed to pass the *same* dst/src (a group-cooperative copy, where only
-// tid/stride vary the per-lane offset -- e.g. memcpy_wave/memcpy_wg). This is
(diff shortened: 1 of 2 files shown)


omor1:
Broad recommendation for the memcpy:
```C++
// memcpy at min(count, sizeof(T)) bytes where dst and src have alignof(T)
template <typename T>
__device__ T* memcpy_single_aligned(T* dst, const T* src, size_t count);

// memcpy count bytes where dst and src have alignof(T)
template <typename T>
__device__ T* memcpy_loop_aligned(T* dst, const T* src, size_t count);

__device__ void* memcpy(void* dst, const void* src, size_t count) {
  uintptr_t s = reinterpret_cast<uintptr_t>(src);
  uintptr_t d = reinterpret_cast<uintptr_t>(dst);

  /*
   * the alignment of an address is the number of least-significant zeros in the address:
   * 0 ->  1-byte aligned
   * 1 ->  2-byte aligned
   * 2 ->  4-byte aligned
   * 3 ->  8-byte aligned
   * 4 -> 16-byte aligned
   */
  int src_alignment = countr_zero(s);
  int dst_alignment = countr_zero(d);
  int alignment = min(src_alignment, dst_alignment);
  // possibly compute minimum alignment of src and dst for all lanes

  if (__builtin_expect(alignment >= 4 || src_alignment == dst_alignment, true)) {
    // step up through alignments until maximum copy width (16 bytes)
    switch (alignment) {
    case 0:
      // copy (at most) 1 byte, with 1-byte alignment of dst and src
      assert(__builtin_is_aligned(d, alignof(uint8_t)));
      assert(__builtin_is_aligned(s, alignof(uint8_t)));
      memcpy_single_aligned(reinterpret_cast<uint8_t*>(d), reinterpret_cast<const uint8_t*>(s), count);
      d += sizeof(uint8_t);
      s += sizeof(uint8_t);
      if (count > sizeof(uint8_t)) {
        count -= sizeof(uint8_t);
        // dst and src now aligned to 2 bytes
        [[fallthrough]]
      else {
        return dst;
      }
    case 1:
      // copy (at most) 2 bytes, with 2-byte alignment of dst and src
      assert(__builtin_is_aligned(d, alignof(uint16_t)));
      assert(__builtin_is_aligned(s, alignof(uint16_t)));
      memcpy_single_aligned(reinterpret_cast<uint16_t*>(d), reinterpret_cast<const uint16_t*>(s), count);
      d += sizeof(uint16_t);
      s += sizeof(uint16_t);
      if (count > sizeof(uint16_t)) {
        count -= sizeof(uint16_t);
        // dst and src now aligned to 4 bytes
        [[fallthrough]]
      else {
        return dst;
      }
    case 2:
      // copy (at most) 4 bytes, with 4-byte alignment of dst and src
      assert(__builtin_is_aligned(d, alignof(uint32_t)));
      assert(__builtin_is_aligned(s, alignof(uint32_t)));
      memcpy_single_aligned(reinterpret_cast<uint32_t*>(d), reinterpret_cast<const uint32_t*>(s), count);
      d += sizeof(uint32_t);
      s += sizeof(uint32_t);
      if (count > sizeof(uint32_t)) {
        count -= sizeof(uint32_t);
        // dst and src now aligned to 8 bytes
        [[fallthrough]]
      else {
        return dst;
      }
    case 3:
      // copy (at most) 8 bytes, with 8-byte alignment of dst and src
      assert(__builtin_is_aligned(d, alignof(uint64_t)));
      assert(__builtin_is_aligned(s, alignof(uint64_t)));
      memcpy_single_aligned(reinterpret_cast<uint64_t*>(d), reinterpret_cast<const uint64_t*>(s), count);
      d += sizeof(uint64_t);
      s += sizeof(uint64_t);
      if (count > sizeof(uint64_t)) {
        count -= sizeof(uint64_t);
        // dst and src now aligned to 16 bytes
        [[fallthrough]]
      else {
        return dst;
      }
    default:
      // copy with 16-byte alignment of dst and src
      assert(__builtin_is_aligned(d, 16));
      assert(__builtin_is_aligned(s, 16));
      return memcpy_16byte_aligned(reinterpret_cast<void*>(d), reinterpret_cast<const void*>(s), count);
    }
  } else {
    /*
     * handle differing src and dst alignments < 16-byte
     * this is hard to implement correctly & efficiently,
     * since correcting alignment for one will misalign the other
     * just do a loop correct for the alignment
     */
    switch (alignment) {
    case 0:
      return memcpy_loop_aligned(reinterpret_cast<uint8_t*>(dst), reinterpret_cast<const uint8_t*>(src), count);
    case 1:
      return memcpy_loop_aligned(reinterpret_cast<uint16_t*>(dst), reinterpret_cast<const uint16_t*>(src), count);
    case 2:
      return memcpy_loop_aligned(reinterpret_cast<uint32_t*>(dst), reinterpret_cast<const uint32_t*>(src), count);
    case 3:
      return memcpy_loop_aligned(reinterpret_cast<uint64_t*>(dst), reinterpret_cast<const uint64_t*>(src), count);
    default:
      assert(false /* alignment >= 4 is handled elsewhere, this can't occur);
      __builtin_unreachable();
    }
  }
}
```

projects/rocshmem/src/util.hpp
```diff
@@ -441,53 +441,35 @@ __device__ __forceinline__ void copy_bulk(void* dst, void* src,
...
-  if (remainder & 8) {
-    auto val = AsmAccess<8, LP, SP>::load(src);
-    if constexpr (LP != CachePolicy::Standard) {
-      wait_on_vmem_and_lds(0);
+__device__ __forceinline__ void copy_remainder(uint8_t* dst, uint8_t* src, int size) {
+  while (size > 0) {
+    // Dynamically check alignment at the current pointer position
+    uintptr_t align = reinterpret_cast<uintptr_t>(dst) | reinterpret_cast<uintptr_t>(src);
+
+    if (size >= 8 && (align & 7) == 0) {
```
abouteiller, line 511: builtin_expect we are aligned?

projects/rocshmem/src/util.hpp
```diff
@@ -497,108 +479,134 @@ __device__ __forceinline__ void copy_remainder(uint8_t* dst,
...
+      copy_bulk<16, LP, SP, 16>(dst, src, n_chunks, wave_tid, wave_size);
+
+      if (wave_tid == 0) {
+        copy_remainder<LP, SP>(static_cast<uint8_t*>(dst) + n_chunks * 16,
+                               static_cast<uint8_t*>(src) + n_chunks * 16,
+                               remainder);
+      }
+    } else {
+      // Safety Net: Large unaligned transfer
+      copy_bulk<1, LP, SP, 4>(dst, src, size, wave_tid, wave_size);
```
abouteiller: Can we transform a large unaligned copy into a preamble and large aligned copy instead

projects/rocshmem/src/util.hpp
```diff
@@ -441,53 +441,35 @@ __device__ __forceinline__ void copy_bulk(void* dst, void* src,
...
-      return;
-    }
-    dst += 4;
-    src += 4;
-  }
-  if (remainder & 8) {
-    auto val = AsmAccess<8, LP, SP>::load(src);
-    if constexpr (LP != CachePolicy::Standard) {
-      wait_on_vmem_and_lds(0);
+__device__ __forceinline__ void copy_remainder(uint8_t* dst, uint8_t* src, int size) {
```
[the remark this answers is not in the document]
amirsojoodi: `copy_remainder` is only called for `size < 16`

projects/rocshmem/src/util.hpp
```diff
@@ -434,156 +437,202 @@ __device__ __noinline__ void copy_bulk(void* dst, void* src,
...
-    dst += 4;
-    src += 4;
   }
-  if (remainder & 8) {
-    auto val = AsmAccess<8, LP, SP>::load(src);
-    AsmAccess<8, LP, SP>::store(dst, val);
+}
+
+template <int ChunkSize, CachePolicy LP, CachePolicy SP, int Unroll>
+__device__ __forceinline__ void copy_aligned_body(uint8_t* dst, uint8_t* src,
```
avinashkethineedi, line 509: only `dst` ptr is aligned.

PR #7956, opened 2026-06-29
fix(clr): bound fat-binary parsing by readable image size

fabiomestre:
## Motivation

The pointer-based module load APIs (`hipModuleLoadData`, `hipModuleLoadDataEx`, `hipModuleLoadFatBinary`) receive an image pointer with no length. While parsing a fat binary, the runtime read fixed-size structures (offload-bundle magic strings, the compressed bundle header, the declared `totalSize`, and the ELF header) without verifying they fit within the actual image. A malformed or truncated image could therefore cause out-of-bounds reads past the supplied buffer. This PR hardens the fat-binary parsing path against those over-reads and adds regression tests so the behavior cannot silently regress.

## Technical Details
Tests (hip-tests):
- Added to `catch/unit/oob/oob_module.cc` (Linux/AMD-only, matching the `unit/oob` gating):
  - `OOB_hipModuleLoadData_Negative_TruncatedImages` — feeds truncated/malformed images (short compressed magic, out-of-bounds `totalSize`, sub-`Elf64_Ehdr` AMDGPU ELF, truncated uncompressed magic, and an AMDGPU ELF whose section table is declared past the image) and expects `hipErrorInvalidImage`. Each image is placed flush against the end of a file-backed page followed by a no-access guard page, so an over-read deterministically faults rather than silently succeeding.
  - `OOB_hipModuleLoadData_Positive_ValidFileBackedImage` — false-positive guard: loads a valid file-backed compressed module and runs its kernel, ensuring the new checks do not reject legitimate images.
  - `OOB_hipModuleLoad_Negative_OutOfBoundsTotalSize` — file-path counterpart of the `totalSize` check.

## JIRA ID

ROCM-26907

## Test Plan

CI is expected to pass with no regressions. New tests are expected to pass on both Windows and Linux CI runs.

## Test Result

CI passed successfully.

## Submission Checklist


projects/hip-tests/catch/unit/module/hipFatBinaryBounds.cc
```diff
@@ -0,0 +1,463 @@
...
+                        FILE_ATTRIBUTE_NORMAL, nullptr);
+    REQUIRE(file_ != INVALID_HANDLE_VALUE);
+    DWORD written = 0;
+    REQUIRE(WriteFile(file_, file_bytes.data(), static_cast<DWORD>(file_bytes.size()), &written,
+                      nullptr));
+    REQUIRE(written == file_bytes.size());
+
+    // Only this one page is backed; the pages after it stay unmapped and act as
+    // the guard, so any read past the page faults.
+    mapping_ = CreateFileMappingA(file_, nullptr, PAGE_READONLY, 0, 0, nullptr);
```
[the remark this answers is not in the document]
fabiomestre: OOB tests don't run on windows

projects/clr/hipamd/src/hip_fatbin.cpp
```diff
@@ -426,18 +465,24 @@ hipError_t FatBinaryInfo::ExtractFatBinaryUsingCOMGR(const std::vector<hip::Devi
...
   if (!is_compressed && !is_uncompressed) {
-    if (IsCodeObjectElf(image_)) {
+    if (IsCodeObjectElf(image_, readable_size)) {
       // Load the binary directly
       auto elf_size = amd::Elf::getElfSize(image_);
+      if (readable_size != 0 && (elf_size == 0 || elf_size > readable_size)) {
+        LogPrintfError("ELF size out of bounds: elf_size=%zu mapped=%zu",
+                       static_cast<size_t>(elf_size), readable_size);
+        return hipErrorInvalidImage;
+      }
```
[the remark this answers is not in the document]
fabiomestre: Outdated comment. Elf is handled by https://github.com/ROCm/rocm-systems/pull/9019/changes#diff-68b2e4890322fbbe7148739ea39b0454471352575560bda9b9d1766499e030be

projects/hip-tests/catch/unit/oob/hipFatBinaryBounds.cc
```diff
@@ -0,0 +1,463 @@
...
+
+#endif  // defined(_WIN32)
+
+}  // namespace
+
+/**
+ * Feeds guard-page-backed malformed images to the pointer load APIs and
+ * expects hipErrorInvalidImage instead of an out-of-bounds read.
+ */
+HIP_TEST_CASE(Unit_hipModuleLoadData_Negative_TruncatedImages) {
```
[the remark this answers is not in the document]
fabiomestre: Done

projects/clr/hipamd/src/hip_fatbin.cpp
```diff
@@ -429,28 +466,30 @@ hipError_t FatBinaryInfo::ExtractFatBinaryUsingCOMGR(const std::vector<hip::Devi
     }
     image_size_ = fsize;
     image_mapped_ = true;
+    readable_size = fsize;
```
cjatin: minor nit: we can use `readable_size` in this section instead of `fsize`

projects/clr/hipamd/src/hip_fatbin.cpp
```diff
@@ -429,28 +466,30 @@ hipError_t FatBinaryInfo::ExtractFatBinaryUsingCOMGR(const std::vector<hip::Devi
...
   }
   guarantee(image_ != nullptr, "Image cannot be nullptr, file:%s did not map for some reason",
```
cjatin, line 475: should we have a check before this line on the size?
we check it on line 460. Was thinking if it made sense to have a single check outside the `if`/`else`

projects/clr/hipamd/src/hip_fatbin.cpp
```diff
@@ -429,28 +466,30 @@ hipError_t FatBinaryInfo::ExtractFatBinaryUsingCOMGR(const std::vector<hip::Devi
...
   if (!is_compressed && !is_uncompressed) {
-    if (IsCodeObjectElf(image_)) {
-      // Use the exact file size when known, else the mapping bound derived above.
-      size_t buf_size = image_size_ != 0 ? image_size_ : image_region_bound;
-      if (buf_size == 0) {
+    if (IsCodeObjectElf(image_, readable_size)) {
+      // readable_size bounds the ELF parse: it is the exact file size on the
+      // file path and the mapping's readable extent on the in-memory path.
+      // getElfSize fails closed, so a zero (unknown) bound must be rejected here.
+      if (readable_size == 0) {
```
cjatin: if we add the check above. This can be removed. `IsCodeObjectElf` does not modify the size, so it should be the same.

projects/clr/hipamd/src/hip_fatbin.cpp
fabiomestre: The issue is that the size can be 0 when the image is not a mapped file. So we can't really guarantee anything on this line.

projects/clr/hipamd/src/hip_fatbin.cpp
fabiomestre: I don't think the check above can be added unless I'm missing something

cjatin (approved):

mangupta (approved):

PR #7863, opened 2026-06-26
Revert "Guard kpack splitting against empty target matches"

amd-chiranjeevi:
Reverts PR: ROCm/rocm-systems#7633
Fixes Issue: https://github.com/ROCm/rocm-systems/issues/7862

Issue #7862, opened 2026-06-26
[Issue] Windows gfx101X-dgpu, gfx103x, gfx900, gfx906 math-libs stage fails — r`ocwmma_test artifact split error: no device code objects matched` gfx101X targets


amd-chiranjeevi:
## Context

- **Workflow:** Windows::release / Build Multi-Arch Stages / math-libs (gfx101X-dgpu, false)
- **GPU arch:** gfx101X-dgpu, gfx103x, gfx900, gfx906
- **Failing PR:** https://github.com/ROCm/TheRock/pull/6163 — Bump rocm-systems `cd5290491` → `b0a543eb5`
- **Failing run:** https://github.com/ROCm/TheRock/actions/runs/28210131602/job/83571640061
- **Failing step:** Ninja build — `split_artifacts.py` for `rocwmma_test_gfx101X-dgpu`

## Summary

After the rocm-systems bump, `split_artifacts.py` now raises `RuntimeError` instead of silently emitting an untransformed generic artifact when fat binaries contain no device code matching `--gpu-targets`. For the gfx101X-dgpu build shard, rocWMMA is excluded from gfx1010/1011/1012 (all RDNA1, no MFMA support, per [#1944](https://github.com/ROCm/TheRock/issues/1944)), so its GPU_TARGETS list is empty after filtering. rocWMMA falls back to its own default targets (gfx1100, gfx90a, etc.) and compiles 37 fat binaries — none of which contain gfx101X device code. The new strict guard then refuses to split the `rocwmma_test` artifact.

## Error

```
FAILED: artifacts/rocwmma_test_generic/artifact_manifest.txt B:/build/artifacts/rocwmma_test_generic/artifact_manifest.txt
Error: Found 37 fat binaries, but no device code objects matched --gpu-targets (gfx1010, gfx1011, gfx1012). Refusing to emit an untransformed generic artifact.
ninja: build stopped: cannot make progress due to previous errors.
```

## Suspected rootcause:
-  [rocm-systems#7633](https://github.com/ROCm/rocm-systems/pull/7633) ("Guard kpack splitting against empty target matches"), merged 2026-06-25

## Impact:
Blocker for promotion of rocm-systems: https://github.com/ROCm/TheRock/pull/6163

amd-chiranjeevi:
revrt PR: https://github.com/ROCm/rocm-systems/pull/7863

testing the revert: https://github.com/ROCm/TheRock/actions/runs/28221819677

[...]

PR #7957, opened 2026-06-29
fix(kpack): guard splitting against empty target matches

dan-garvey:
## Summary
- re-land the kpack split guard from ROCm/rocm-systems#7633 after its revert in ROCm/rocm-systems#7863
- fail artifact splitting when fat binaries are present but `--gpu-targets` filters out every device code object, instead of emitting an untransformed generic artifact
- strengthen generic artifact verification across ELF and PE/COFF binaries, including kpack marker and wrapper magic checks
- restore regression coverage for the empty-filter guard and transformed/raw/stale-wrapper verifier cases

## Sequencing
This PR is predicated on the order agreed from ROCm/rocm-systems#7862:
1. Land ROCm/TheRock#6042 first so unsupported rocWMMA target shards (`gfx101X-dgpu`, `gfx103X-all`, `gfx900`, `gfx906`) no longer build or publish `rocwmma_test_generic`.
2. Re-land this rocm-systems kpack guard.
3. Bump TheRock rocm-systems submodule to the re-landed commit.

Without ROCm/TheRock#6042 first, TheRock bump validation can hit this intended guard while rocWMMA is still producing test artifacts for unsupported target shards.

## Validation
- `git diff --check origin/develop...HEAD`
- `uv run --with pytest --with msgpack --with zstandard --directory shared/kpack pytest tests/test_artifact_splitter_integration.py -q` (42 passed)
- `uv run --with pre-commit --directory shared/kpack pre-commit run --files python/rocm_kpack/tools/verify_artifacts.py tests/test_artifact_splitter_integration.py`

ISSUE ID: https://github.com/ROCm/rocm-libraries/issues/8660

### DIFF
--- a/shared/kpack/python/rocm_kpack/artifact_splitter.py
+++ b/shared/kpack/python/rocm_kpack/artifact_splitter.py
@@ -799,6 +799,21 @@ class ArtifactSplitter:
                 all_kernels_by_arch, output_dir
             )
 
+        if (
+            fat_binaries_by_prefix
+            and self.gpu_targets is not None
+            and not kpack_info_by_arch
+        ):
+            target_list = ", ".join(sorted(self.gpu_targets))
+            fat_binary_count = sum(
+                len(paths) for paths in fat_binaries_by_prefix.values()
+            )
+            raise RuntimeError(
+                f"Found {fat_binary_count} fat binaries, but no device code objects "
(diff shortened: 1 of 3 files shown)


ScottTodd (approved):
I think this should be safe to merge now that https://github.com/ROCm/TheRock/pull/6042 is in.

PR #7964, opened 2026-06-29
fix(rocjitsu): Avoid signed overflow UB in scalar add/sub SCC execution

accauble:
## Motivation
The autogenerated executors for `s_add_co_i32`, `s_sub_co_i32`, `s_add_i32`, and `s_sub_i32` computed the result with a bare signed int32. This failed with GCC13+ builds with optimizations because signed integer overflow is undefined and assumed that it does not happen, and optimized out our overflow check.

Edit: To fix this, we perform the operation unsigned and check whether to set SCC via the sign-bit identity, as suggested by reviewers.

NOTE: This fix was not done to avoid sanitizers. ASAN was a red herring. It was done to fix a bug that surfaces when using GCC13 with optimizations.

## Technical Details

The original code looked like this:

```C
1: inline void execute_s_add_i32_sop2([[maybe_unused]] Inst &inst, [[maybe_unused]] Wavefront &wf) {
2:  int32_t s0 = static_cast<int32_t>(inst.ssrc0.read_scalar(wf));
3:  int32_t s1 = static_cast<int32_t>(inst.ssrc1.read_scalar(wf));
4:  int32_t result = s0 + s1;
5:  inst.sdst.write_scalar(wf, result);
6:  wf.write_scc(
      ((static_cast<int64_t>(s0) + static_cast<int64_t>(s1)) != static_cast<int64_t>(result)));
```

The important lines of this code snippet are lines 4 and 6. Line 4 does the addition. Line 6 is where the bug surfaced. Line 6 checks and sets SCC if we carry.

### Why unit tests were failing

We have 2 unit tests (`Rdna4ScalarSccTest.AddSubCoI32UseSignedOverflow` and `Rdna4ScalarSccTest.AddSubCoI32UseSignedOverflow`) which made sure that SCC was set correctly in the event that the result overflowed. When rocjitsu is compiled with clang or gcc + O0, there is no issue here. However, when optimizations are enabled, GCC optimizes out line 6, because according to C++, line 4 is undefined behavior and it assumes that line 4 will never overflow. This allows it to say that `(static_cast<int64_t>(s0) + static_cast<int64_t>(s1)) != static_cast<int64_t>(result))` is FALSE every time. Which means the SCC bit never gets set.

### The Fix

Since we allow GCC as a compiler, we have to fix line 4 to do the sum and avoid line 6 from being optimized out. I originally chose to do this by casting to unsigned ints to do the sum (not undefined behavior) and casting back to the signed integer to get the actual sum. Reviewers suggested to instead just do the sum unsigned and check the SCC bit via the sign-bit identity, so the code and tests have been updated to reflect that.

## JIRA ID

Resolves JIRA ID : ROCM-27230

## Test Plan

Python tests should still pass.
Build with gcc13. All tests should pass.

## Test Result

Python tests still pass.
Tests pass with gcc13.

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/python/amdisa/sema_derive.py
+++ b/emulation/rocjitsu/lib/python/amdisa/sema_derive.py
@@ -507,6 +507,18 @@ class _ScalarBinop(_ScalarDeriver):
             result = SemaNode(
                 SemaNodeKind.CALL, ty=ty, call_name=fn, children=(_id(fn), src0, src1)
             )
+        elif op in ('add', 'sub') and ty.base == 'I':
+            # Perform the wrapping add/sub in unsigned to avoid signed-overflow
+            # UB. GCC13+ assumes no overflow with a bare signed `s0 + s1` and
+            # so the overflow check is always false if not done unsigned. Then,
+            # reinterpret the unsigned result back to signd for the overflow check.
+            u_ty = SemaType.U32 if ty.size == 32 else SemaType.U64
+            arith = SemaNode(
+                kind,
+                ty=u_ty,
+                children=(_cast(src0, u_ty), _cast(src1, u_ty)),
+            )
+            result = _cast(arith, ty)
         else:
             result = SemaNode(kind, ty=ty, children=(src0, src1))
 
--- a/emulation/rocjitsu/lib/python/amdisa/sema_derive.py
+++ b/emulation/rocjitsu/lib/python/amdisa/sema_derive.py
@@ -559,35 +559,14 @@ class _ScalarBinop(_ScalarDeriver):
                     ),
                 )
             elif sem.sets_scc == 'overflow':
-                # Signed overflow via the sign-bit identity:
-                # For add, overflow iff the operands share a sign that differs
-                # from the result's.
-                # For sub, overflow iff the operands differ in sign and src0's
-                # sign differs from the result's.
-                res_id = _id('result', result_ty)
-                if op == 'sub':
-                    x0 = SemaNode(SemaNodeKind.XOR, ty=result_ty, children=(src0, src1))
-                    x1 = SemaNode(
-                        SemaNodeKind.XOR, ty=result_ty, children=(src0, res_id)
-                    )
-                else:
-                    x0 = SemaNode(
-                        SemaNodeKind.XOR, ty=result_ty, children=(src0, res_id)
-                    )
-                    x1 = SemaNode(
-                        SemaNodeKind.XOR, ty=result_ty, children=(src1, res_id)
-                    )
-                sign_mask = _lit(
-                    '0x80000000u' if result_ty.size == 32 else '0x8000000000000000',
-                    result_ty,
-                )
+                # Signed-overflow SCC detected in unsigned via the sign-bit
+                # identity
+                fn = 'signed_sub_overflows' if op == 'sub' else 'signed_add_overflows'
                 scc_expr = SemaNode(
-                    SemaNodeKind.AND,
-                    ty=result_ty,
-                    children=(
-                        SemaNode(SemaNodeKind.AND, ty=result_ty, children=(x0, x1)),
-                        sign_mask,
-                    ),
+                    SemaNodeKind.CALL,
+                    ty=SemaType.U1,
+                    call_name=fn,
+                    children=(_id(fn), src0, src1),
                 )
             elif sem.sets_scc == 'compare':
                 cmp_kind = SemaNodeKind.GE if op == 'max' else SemaNodeKind.LT
--- a/emulation/rocjitsu/lib/python/amdisa/sema_derive.py
+++ b/emulation/rocjitsu/lib/python/amdisa/sema_derive.py
@@ -398,7 +398,9 @@ class _ScalarBinop(_ScalarDeriver):
     def derive(sem: InstructionSemantics) -> SemaBlock:
         op = sem.operation
         ty = _dtype_to_sema(sem.data_type)
-        raw_carry_bits = sem.sets_scc in ('carry', 'borrow') and ty.base == 'I'
+        raw_carry_bits = (
+            sem.sets_scc in ('carry', 'borrow', 'overflow') and ty.base == 'I'
+        )
         calc_ty = SemaType('U', ty.size) if raw_carry_bits else ty
         src0 = _cast(_src(0), calc_ty)
         src1 = _cast(_src(1), calc_ty)
@@ -508,22 +510,21 @@ class _ScalarBinop(_ScalarDeriver):
                 SemaNodeKind.CALL, ty=ty, call_name=fn, children=(_id(fn), src0, src1)
             )
         elif op in ('add', 'sub') and ty.base == 'I':
-            # Perform the wrapping add/sub in unsigned to avoid signed-overflow
-            # UB. GCC13+ assumes no overflow with a bare signed `s0 + s1` and
-            # so the overflow check is always false if not done unsigned. Then,
-            # reinterpret the unsigned result back to signd for the overflow check.
+            # Emulate the hardware's wrap-around arithmetic in unsigned. A bare
+            # signed `s0 + s1` is undefined on overflow, which GCC -O1+ exploits
+            # to fold away the signed-overflow SCC check. Derive SCC from the
+            # the sign-bit identity
             u_ty = SemaType.U32 if ty.size == 32 else SemaType.U64
-            arith = SemaNode(
+            result = SemaNode(
                 kind,
                 ty=u_ty,
                 children=(_cast(src0, u_ty), _cast(src1, u_ty)),
             )
-            result = _cast(arith, ty)
         else:
             result = SemaNode(kind, ty=ty, children=(src0, src1))
 
         result_ty = calc_ty if raw_carry_bits else ty
-        if op == 'mul' and ty.base == 'I':
+        if op in ('mul', 'add', 'sub') and ty.base == 'I':
             result_ty = result.ty or result_ty
         if ty.base in ('F', 'BF') and ty.size == 16:
             result_ty = SemaType.F32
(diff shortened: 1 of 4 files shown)


amd-eochoalo (approved):
I tested this locally and it is working. 

A note for other reviewers / maintainers: I have a branch adding CI for building rocjitsu sanitized with asan and ubsan compiled with gcc. However, this error only shows when compiling with asan and not ubsan. Just pointing out that maybe we shouldn't use both at the same time.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/execute_shared.h
```diff
@@ -408,7 +408,7 @@ template <typename Inst>
 inline void execute_s_add_co_i32_sop2([[maybe_unused]] Inst &inst, [[maybe_unused]] Wavefront &wf) {
   int32_t s0 = static_cast<int32_t>(inst.ssrc0.read_scalar(wf));
   int32_t s1 = static_cast<int32_t>(inst.ssrc1.read_scalar(wf));
-  int32_t result = (s0 + s1);
+  int32_t result = static_cast<int32_t>((static_cast<uint32_t>(s0) + static_cast<uint32_t>(s1)));
```
bjacob: That is hiding the overflow from the sanitizer, effectively implementing wrap-around overflow for int32.  Is this the desired behavior?

emulation/rocjitsu/lib/python/amdisa/sema_derive.py
```diff
@@ -507,6 +507,18 @@ def derive(sem: InstructionSemantics) -> SemaBlock:
             result = SemaNode(
                 SemaNodeKind.CALL, ty=ty, call_name=fn, children=(_id(fn), src0, src1)
             )
+        elif op in ('add', 'sub') and ty.base == 'I':
+            # Perform the wrapping add/sub in unsigned to avoid signed-overflow
+            # UB. GCC13+ assumes no overflow with a bare signed `s0 + s1` and
```
amd-eochoalo: Thinking a little bit more, why didn't ubsan complained about this? Let me run with just ubsan.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/execute_shared.h
bjacob: Self-answering: yes, this is intended because this is emulating a hardware instruction that presumably has that wrap-around behavior.

My question becomes:  why is this code using `int32_t` at all?  An emulator implementation of an instruction that performs wrap-around arithmetic naturally wants to be using `uint32_t`.

kuhar: It would be easier to tell what the intention is if we defined helpers for zero/sign-extension and truncation and relied on these over static_casts... but that's probably well outside of the scope of this PR

accauble: It seems there was some confusion over the sanitizers so I have updated the PR description to clarify that.

As for why use `int32_t` at all? That is a good question. The instruction itself returns a signed int so I returned a signed int, but that's probably not the best answer.

kuhar (approved):

emulation/rocjitsu/lib/python/amdisa/sema_derive.py
amd-eochoalo: Just as a note, if we run with:

```
UBSAN_OPTIONS=halt_on_error=1 ctest --test-dir build -R Rdna4 --output-on-failure

 8/19 Test #525: Rdna4ScalarSccTest.AddSubCoI32UseSignedOverflow ..............................***Failed    0.01 sec
Running main() from /home/eochoalo/code/rocm-systems/emulation/rocjitsu/third_party/googletest-src/googletest/src/gtest_main.cc
Note: Google Test filter = Rdna4ScalarSccTest.AddSubCoI32UseSignedOverflow
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from Rdna4ScalarSccTest
[ RUN      ] Rdna4ScalarSccTest.AddSubCoI32UseSignedOverflow
/home/eochoalo/code/rocm-systems/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/execute_shared.h:411:24: runtime error: signed integer overflow: 2147483647 + 1 cannot be represented in type 'int32_t' (aka 'int')
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/eochoalo/code/rocm-systems/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/execute_shared.h:411:24

      Start 527: Rdna4Dot2True16ExecutionTest.F16AppliesVop3ModifiersAndSelectedHalves
 9/19 Test #527: Rdna4Dot2True16ExecutionTest.F16AppliesVop3ModifiersAndSelectedHalves ........   Passed    0.01 sec
 ```

File: emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/execute_shared.h (hunk 1 above)
Comment (bjacob) [reply emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/execute_shared.h]: Can you elaborate on "The instruction itself returns a signed int" ?

At the ISA level, this instruction is likely signless.

At the Rocjitsu level, I am not familiar with the code, but reading line 409,

```c++
 int32_t s0 = static_cast<int32_t>(inst.ssrc0.read_scalar(wf));
```

I looked into the expression `inst.ssrc0.read_scalar(wf)`.  Here, the template parameter `Inst` seems the be instantiated as `SAddCoI32Sop2`, I can see that its `ssrc0` member has type `Operand` and that at operand.h:116, we have

```c++
  virtual uint32_t read_scalar(const amdgpu::Wavefront &wf) const;
```

So this returns a `uint32_t` to begin with. Why cast it to `int32_t`?

bjacob: Oh, I think I understand where signed arithmetic enters the picture now.

Below the write_scc needs to write (IIUC) whether the addition overflows *signed* i32.

But that is still no reason for signed arithmetic to enter the picture above.  Signed arithmetic is at most needed only below in the SCC computation.  The above casts were still a footgun:  the above code subtly requires C++20 to be fully defined to behave as intended.  Before C++20, the above uint32_t-to-int32_t cast has implementation-defined behavior when the uint32_t value is outside of the int32_t range. While that is going to be fine in practice, it's best to have code that doesn't rely on such subtleties.

Simplify to:

```c++
inline void execute_s_add_co_i32_sop2([[maybe_unused]] Inst &inst, [[maybe_unused]] Wavefront &wf) {
  uint32_t s0 = inst.ssrc0.read_scalar(wf);
  uint32_t s1 = inst.ssrc1.read_scalar(wf);
  uint32_t sum = s0 + s1;
  inst.sdst.write_scalar(wf, sum);
  wf.write_scc((s0 ^ sum) & (s1 ^ sum) & 0x80000000u);
}
```

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/execute_shared.h
```diff
@@ -2479,12 +2477,11 @@ inline void execute_s_sub_co_ci_u32_sop2([[maybe_unused]] Inst &inst,
...
-  int32_t s0 = static_cast<int32_t>(inst.ssrc0.read_scalar(wf));
-  int32_t s1 = static_cast<int32_t>(inst.ssrc1.read_scalar(wf));
-  int32_t result = (s0 - s1);
+  uint32_t s0 = inst.ssrc0.read_scalar(wf);
+  uint32_t s1 = inst.ssrc1.read_scalar(wf);
+  uint32_t result = (s0 - s1);
   inst.sdst.write_scalar(wf, result);
-  wf.write_scc(
-      ((static_cast<int64_t>(s0) - static_cast<int64_t>(s1)) != static_cast<int64_t>(result)));
+  wf.write_scc((((s0 ^ s1) & (s0 ^ result)) & 0x80000000u));
```
bjacob: Good! That is the correct adaptation of the above add-overflow check, now to subtraction instead of addition.

Could you introduce helper inline functions to contain these nontrivial pieces of arithmetic, thus avoiding having to write each twice and making the intent more explicit?

```c++
inline bool overflow_add_signed_u32(uint32_t a, uint32_t b) {
  return ((s0 ^ result) & (s1 ^ result)) & 0x80000000u;
}

inline bool overflow_sub_signed_u32(uint32_t a, uint32_t b) {
  return ((s0 ^ s1) & (s0 ^ result)) & 0x80000000u;
}
```

Note: please find better functin names, I'm not good with naming.  The point is that the function name needs to somehow conveys that it checks for *signed* overflow,  and that its parameter types are *unsigned* (and thus will cause implict casting)  (actually, a conceivable way of guarding against the implicit casting would be to templatize them, and then have them static_assert(std::is_same_v<T, uint32_t>....)  up to you.

bjacob (approved):

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/execute_shared.h
accauble: Done!

amd-eochoalo:
I think updating the branch should get rid of the PR bot failing check.

PR #7979, opened 2026-06-29
fix(rocjitsu): Add InstDefUse Checks to Handle Partial Defs

theSK2005:
## Motivation

Some instructions have dst operands that are under 32 bits wide. These will be a partial write for 32 bit registers and InstDefUse/liveness need to account for this and not have incorrect kills.

## Technical Details

Added a check for dst operand size to add as uses if their bit size is less than 32. Remaining logic is handled by current liveness implementation

## Test Plan

Write tests to make sure partial defs don't cause incorrect kills and that full defs continue to cause correct kills. All prior tests pass.

## Test Result

New tests pass, prior tests all pass.

## Submission Checklist


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/def_use_chain.cpp
```diff
@@ -18,8 +18,18 @@ namespace {
   return ref.cls == RegClass::VGPR || ref.cls == RegClass::ACC_VGPR;
 }
 
-void add_def(InstDefUse &du, RegisterRef ref) {
+// A 32-bit register lane is the finest granularity RegisterSet tracks. An
+// operand narrower than that (e.g. a 16-bit dst) writes only part of its
+// lowest lane and leaves the remaining bits intact, so the instruction
+// read-modify-writes that register rather than fully redefining it.
+[[nodiscard]] bool is_partial_def(int size_bits) { return size_bits > 0 && size_bits < 32; }
```
accauble: Nit: I know the magic number is pre-existing but it probably would be worth creating a named constant for the register land width...

emulation/rocjitsu/tests/analysis/liveness_test.cpp
```diff
@@ -45,13 +48,16 @@ class TestInstruction : public Instruction {
...
                   std::optional<int64_t> branch_delta = std::nullopt,
-                  std::initializer_list<RegisterRef> implicit_uses = {})
+                  std::initializer_list<RegisterRef> implicit_uses = {}, int def_size_bits = 0)
       : Instruction(mnemonic, nullptr), implicit_uses_(implicit_uses), branch_delta_(branch_delta) {
     size_ = 4;
     flags_ = flags;
 
     for (RegisterRef ref : defs) {
-      dst_storage_[num_dst_] = TestOperand(ref);
+      // def_size_bits == 0 keeps the default full-lane width; a non-zero value
```
accauble, line 59: Why 0 and not just default to 32?

accauble (approved):
Looks good but consider addressing a couple nits.

emulation/rocjitsu/tests/analysis/liveness_test.cpp
```diff
@@ -33,6 +33,9 @@ class TestOperand : public Operand {
 public:
   TestOperand() = default;
   explicit TestOperand(RegisterRef ref) : Operand(ref.width * 32, ref.index), ref_(ref) {}
+  // Sub-lane operand: same RegisterRef, but a caller-chosen bit width so partial
+  // (less-than-32-bit) defs can be exercised.
```
newling, line 39: My understanding is that 
1) The number of lanes might be 32 or 64 depending on the architecture. 
2) The register size is 32-bits. 

I'm confused by ' Sub-lane operand', is this an operand that is less than 32-bits (i.e. a strict subset of bits of the register are read/written)? It's the connection between 'number of lanes' and 'bits in a register' that I'm unsure of.

emulation/rocjitsu/tests/analysis/liveness_test.cpp
theSK2005: if they have a TestOperand where they set the RegisterRef's width to 2 and don't declare a def size, then it should default to using the RegisterRef's width to determine the TestOperand width of 64 bits rather than the default 32. Need to distinguish for when no def_size_bits are given because of this.

emulation/rocjitsu/tests/analysis/liveness_test.cpp
theSK2005: This is supposed to refer to sub-register operands. I just clarified the comment.

newling (approved):
LGTM

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/def_use_chain.cpp
```diff
@@ -10,6 +10,8 @@ namespace rocjitsu {
 
 namespace {
 
+constexpr int MaxLaneWidth = 32;
```
Groverkss: I don't think this is the right term. This should be RegisterGranularity or something. Also this should live in IsaTraits.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/def_use_chain.cpp
```diff
@@ -18,8 +20,20 @@ namespace {
   return ref.cls == RegClass::VGPR || ref.cls == RegClass::ACC_VGPR;
 }
 
-void add_def(InstDefUse &du, RegisterRef ref) {
+// A 32-bit register lane is the finest granularity RegisterSet tracks. An
```
Groverkss: All of this should be done in the generator. I had a local patch that is closer to what we want:

https://gist.github.com/Groverkss/6ab61183c791aafd53d26fce591baf1a

I'd recommend doing something like that instead. We should get everything from the dynamic register usage stuff.

Groverkss (changes requested):

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/def_use_chain.cpp
```diff
@@ -10,6 +10,8 @@ namespace rocjitsu {
 
 namespace {
 
+constexpr int MaxLaneWidth = 32;
+
 // TODO: Replace this class-based approximation with instruction metadata that
```
accauble: This TODO can probably go

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/def_use_chain.cpp
theSK2005: What's the reasoning for moving it to the generator? The is_partial_def implementation relies on op->size_bits, so isn't the generated code an input to this? Keeping is_partial_def by the liveness analysis things makes more sense to me than keeping it with codegen.

Groverkss: Generally it's nicer to keep things in the generator so we can reuse it across the codebase. It's also useful for the simulator side of things that we share a common implementation.

Groverkss: Generator is supposed to contain "known facts" about the ISA and generate things from there.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -5659,6 +5659,30 @@ def gen_insts(self) -> None:
...
                         cgen.Statement('void execute_impl(amdgpu::Wavefront &wf)')
                     )
+                    # A sub-dword (< 32-bit) destination writes only part of its
+                    # 32-bit register lane, so the old value survives and the
+                    # register is also a read. Surface these partial defs as
+                    # implicit uses. Runtime-sized outputs are never sub-dword,
+                    # so a static size check suffices. Immediate/label outputs
+                    # (e.g. the S_SETREG hwreg selector) never name a register,
+                    # so skip them to avoid dead overrides.
+                    _partial_def_outputs = [
```
Groverkss, line 5669: I don't think it's a implicit_use really. It is explicit in the operand list. I think we just need to add where we generate source operands if the output is also read.

theSK2005: just to make sure, could adding the dst to the src operand list have any knock on effects?

Groverkss: I don't think it should really.

Groverkss: I don't think it should really. I have it in my local branch.

theSK2005: discussed offline

Groverkss (approved):

PR #7980, opened 2026-06-29
fix(rocSHMEM): correct size for non-double element size in reduce

amirsojoodi:
## Motivation

In the _reduce device_ paths we use `ROCSHMEM_REDUCE_MIN_WRKDATA_SIZE` for deciding whether the provided work buffer is large enough for the direct allreduce path and for sizing the ring allreduce work chunk. That constant represents the minimum work-data capacity in double-sized units, but the IPC and GDA reduce implementations were using it directly as a count of `T` elements.

For element types smaller than `double`, this underestimates the number of available work elements and can cause the reduce implementation to select a segment size that causes unaligned work-buffer accesses. This contributes to incorrect indexing and illegal memory accesses in IPC backend. 

## Technical Details

This issue will be silenced with another in-progress PR: https://github.com/ROCm/rocm-systems/pull/7955
But we can still improve this issue here with aligned memory accesses.

## JIRA ID

JIRA ID: AIROCSHMEM-437

## Test Result

- Functional tests pass with no change in performance on gfx950 and gfx942

### DIFF
--- a/projects/rocshmem/src/gda/context_gda_tmpl_device.hpp
+++ b/projects/rocshmem/src/gda/context_gda_tmpl_device.hpp
@@ -516,7 +516,8 @@ __device__ int GDAContext::reduce(rocshmem_team_t team, T *dest,
   size_t direct_pWrk = PE_size * nreduce;
   size_t direct_pSync = PE_size;
   size_t ring_pSync = 2 * PE_size;
-  size_t provided_pWrk = max(nreduce / 2 + 1, ROCSHMEM_REDUCE_MIN_WRKDATA_SIZE);
+  size_t provided_pWrk = max(nreduce / 2 + 1,
+          (ROCSHMEM_REDUCE_MIN_WRKDATA_SIZE * sizeof(double)) / sizeof(T));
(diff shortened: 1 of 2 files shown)


projects/rocshmem/src/gda/context_gda_tmpl_device.hpp
```diff
@@ -516,7 +516,8 @@ __device__ int GDAContext::reduce(rocshmem_team_t team, T *dest,
   size_t direct_pWrk = PE_size * nreduce;
   size_t direct_pSync = PE_size;
   size_t ring_pSync = 2 * PE_size;
-  size_t provided_pWrk = max(nreduce / 2 + 1, ROCSHMEM_REDUCE_MIN_WRKDATA_SIZE);
+  size_t provided_pWrk = max(nreduce / 2 + 1,
+          (ROCSHMEM_REDUCE_MIN_WRKDATA_SIZE * sizeof(double)) / sizeof(T));
```
abouteiller, line 520: would it make more sense to have MIN_WRKTDATA_SIZE be in bytes?

PR #7982, opened 2026-06-29
fix(rccl): disable DDA IPC init on gfx1100, since there is no xgmi 

prasanna-amd:
## Motivation

RCCL's DDA IPC all-reduce optimization unconditionally attemps cross-GPU IPC memory mapping during communicator init.

## Technical Details
vLLM's 8-GPU tensor-parallel setup crashes with hipErrorInvalidDevicePointer on gfx1100 (Radeon Pro W7900) GPUs. Root cause: RCCL's DDA IPC all-reduce optimization unconditionally attempts cross-GPU IPC memory mapping during communicator init. This requires XGMI interconnects (present on CDNA/MI-series GPUs) but gfx1100 is RDNA3 connected via PCIe, where P2P memory mapping is unsupported. The failed hipIpcOpenMemHandle poisons the HIP device context, crashing all workers. vLLM and PyTorch are not at fault — a minimal torchrun    reproducer confirms the issue is purely in RCCL. Fix: skip DDA IPC init on gfx1100.

## JIRA ID

JIRA ID : ROCM-26074

## Test Plan

Verify disabling DDA IPC gfx1100 fixes crash in minimal torchrun reproducer

## Test Result

Verified that DDA IPC init removal in gfx1100 fixes crash in minimal torchrun reproducer

## Submission Checklist

### DIFF
--- a/projects/rccl/src/ipc_init.cu
+++ b/projects/rccl/src/ipc_init.cu
@@ -56,11 +56,22 @@ ncclResult_t ncclDdaIpcCommInit(ncclComm* comm) {
   }
 
   // DDA IPC requires cross-GPU IPC memory mapping (hipIpcOpenMemHandle).
-  // On non-P2P topologies (e.g. RDNA3 over PCIe) this fails with
-  // hipErrorInvalidDevicePointer and poisons the HIP device context.
-  if (!comm->isAllCudaP2p) {
-    INFO(NCCL_INIT, "ncclDdaIpcCommInit: skipping DDA IPC (no P2P between all GPUs)");
-    return ncclSuccess;
+  // comm->isAllCudaP2p is set via ncclTopoCheckP2p which on AMD/HIP returns true
+  // whenever ranks share a hostHash, regardless of actual P2P support (paths.cc).
+  // Use hipDeviceCanAccessPeer directly — the authoritative runtime check for IPC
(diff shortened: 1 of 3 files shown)


projects/rccl/src/ipc_init.cu
```diff
@@ -55,6 +55,14 @@ ncclResult_t ncclDdaIpcCommInit(ncclComm* comm) {
     return ncclSuccess;
   }
 
+  // DDA IPC requires cross-GPU IPC memory mapping (hipIpcOpenMemHandle).
+  // On non-P2P topologies (e.g. RDNA3 over PCIe) this fails with
+  // hipErrorInvalidDevicePointer and poisons the HIP device context.
+  if (!comm->isAllCudaP2p) {
```
Kapil-Shyam-Pawar: are we sure that this check will work on a single node gfx1100?
because the hip branch just checks if ranks are on the same host while setting it right?
https://github.com/ROCm/rocm-systems/blob/d1ce0ec5778d4fb5a639847a5abb01a537c8a697/projects/rccl/src/graph/paths.cc#L416-L434

not sure if the flag is disabled somewhere else

prasanna-amd: you're right, good catch. Let me rework it.

Kapil-Shyam-Pawar (approved):

PR #7983, opened 2026-06-29
[ci] Add MI455 (gfx125X) bringup to multi-arch CI

geomin12:
Configure gfx125X-dcgpu for rocm-systems nightly multi-arch CI with:
- Custom test runner: linux-mi455-gpu-rocm
- Limited tests: hip-tests, rocrtst, sanity only
- Uses TheRock branch with external_family_overrides support

working here: https://github.com/ROCm/rocm-systems/actions/runs/28422102951

### DIFF
--- a/.github/workflows/therock-multi-arch-ci.yml
+++ b/.github/workflows/therock-multi-arch-ci.yml
@@ -67,7 +67,7 @@ concurrency:
 
 jobs:
   setup:
-    uses: ROCm/TheRock/.github/workflows/setup_multi_arch.yml@users/geomin12/mi455-bringup
+    uses: ROCm/TheRock/.github/workflows/setup_multi_arch.yml@main
     with:
       build_variant: "release"
       # Limit GPU families for rocm-systems CI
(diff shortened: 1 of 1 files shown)


rahulc-gh (commented):
Why are you moving to therock ref to your private branch ??
Also we cannot gfx125x until https://github.com/ROCm/TheRock/pull/6082 lands , else your pytorch buidl will be attempted for 2.10 and 2.12 torch branches and it will fail.

geomin12:
> Why are you moving to therock ref to your private branch ?? Also we cannot gfx125x until [ROCm/TheRock#6082](https://github.com/ROCm/TheRock/pull/6082) lands , else your pytorch buidl will be attempted for 2.10 and 2.12 torch branches and it will fail.

this nightly behaves differently than TheRock nightly. this is just build and test signal! no releases here

rahulc-gh (approved):
LGTM, just move theRock ref to main before merging. If it sjust build and test with no pytorch this can be merged for gfx125x as well builds should be good.

geomin12:
landing this to get mi455 coverage

PR #7984, opened 2026-06-30
[rocprofiler-compute] Correct gfx115x benchmarking and --roofline-data-type support

cfallows-amd:
## Motivation

Using --roofline-data-type option on gfx115x products was revealing inconsistent supported/unsupported precisions supported on the archs. This change corrects the list for benchmarking Int ops, and filtering types in analyze mode for the int ops as well as bf16 matrix ops.

## Technical Details

- Add I8/I32/I64 to benchmarking for gfx11.5 - should not have been marked as unsupported

- Remove BF16 supported datatype from all gfx11.5 arch - we cannot collect WMMA counters for this type, and since it is not a VALU op we should mark it as unsupported
  - Other precisions that also have matrix op support are not affected by this because they also have VALU ops for the precision, and are correctly filtered out
  - This BF16 issue is exclusive to the gfx115x lineup and will not resurface for past or future archs.

## JIRA ID

AIPROFCOMP-677

## Test Plan

- [X] On a gfx1150 system:
  - full profile mode
  - `profile --bench-only`
  - `analyze -b 4` output per usual
  - `analyze --roofline-data-type FP16 BF16 I8 I32 I64` -> filters out bf16 with clear message to user:
    `  ERROR BF16 is not a supported datatype for roofline profiling on RDNA35_POINT_1 (arch: gfx1150)`
    and correct HTML output file

- [ ] Ctest passing for gfx950 and gfx1151 on the runners

## Test Result

All manual testing above (CLI input and visual checking of completion and error logs) completed.

## Submission Checklist


projects/rocprofiler-compute/src/utils/roofline_calc.py
```diff
@@ -97,7 +98,8 @@
     ],  # Unsupported: F4, F6, F8
     "gfx1151": [
         "FP16",
-        "BF16",
+        # "BF16", # Technically BF16 is a supported matrix precision on this arch,
+        # but we are turning it off because we lack the metrics counters
         "FP32",
```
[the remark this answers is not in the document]
vedithal-amd: I think we can merge this with the unsupported comment below - across all 3 places.
Feel free to use multi line comments for readability

vedithal-amd (approved):
LGTM

projects/rocprofiler-compute/src/utils/roofline_calc.py
```diff
@@ -86,33 +86,35 @@
         "I32",
         "I64",
     ],
+    # *For gfx115x series, BF16 is a supported matrix ops precision on this arch, but
+    # we are turning it off because:
+    # a) it has no VALU ops, therefore no PEAK VALU roofline, and
```
mgehre-amd: V_DOT2_BF16_BF16 is a VALU op for BF16, and V_DUAL_DOT2ACC_F32_BF16 is a dual issue VOPD BF16 op
V_WMMA_F32_16X16X16_BF16 also exists.
It's not clear to me why those should be disabled?

vedithal-amd: @cfallows-amd, I believe we don't have VALU ops counters for gfx115x but please feel free to elaborate here.

cfallows-amd: Thank you for bringing this to my attention- this is RDNA3.5 specific addition that is not listed in the CDNA ISAs. Missed in our enablement because VALU is a collective counter for gfx115x, unlike separated precision VALU counters for CDNA that are easier to identify support.
We have the ability to collect VALU flops for gfx115x series, so I will enable it for the roofline and update this PR. Unfortunately there are no WMMA counters for gfx115x so we mark it as disabled and skip benchmarking for time-saving purposes during profiling. If there is reason to still benchmark wmma we can assess adding the kernel sources to run.

abchoudh-amd (approved):

vedithal-amd:
@cfallows-amd seems like this PR needs the post sprint merge treatment where we rebase on latest rocprofiler-compute-develop branch and cherry pick only the actual commits this PR contributes

projects/rocprofiler-compute/src/utils/roofline_calc.py
```diff
@@ -27,92 +27,95 @@
...
-        "I8",
-        "I32",
-        "I64",
-    ],  # Unsupported: F4, F6, F8
+# SUPPORTED_DATATYPES table indicates which datatypes per gpu arch can be generated by
+# the GPU benchmarking for roofline feature.
+# Data type value supports bitwise checks:
+# 1 - only Peak VALU Ops supported
+# 2 - only Matrix Ops supported
+# 3 - VALU and Matrix operations supported
```
vedithal-amd: instead of using magic number can we define enumerators in a core module and use it everywhere? I am afraid using magic numbers 1, 2, 3 will make reading and understanding the code really difficult!

```
VALU_OPS_SUPPORTED = 1
MATRIX_OPS_SUPPORTED = 2
VALU_AND_MATRIX_OPS_SUPPORTED = 3
```

vedithal-amd (commented):
Please use enumerators instead of magic numbers

projects/rocprofiler-compute/src/utils/roofline_calc.py
```diff
@@ -297,7 +300,7 @@ def calc_ceilings(
     ops_flops = "Ops" if dtype.startswith("I") else "Flops"
 
     peak_ops = 0.0
-    if dtype in PEAK_OPS_DATATYPES:
+    if SUPPORTED_DATATYPES[mspec.gpu_arch][dtype] & 1:
```
vedithal-amd: Please use enemurators (imported from core module) everywhere - wherever we are doing comparison.

Also, just wondering shouldn't a `==` equality comparison be sufficient here instead of `&` AND operator - equality comparisons are easier to read and understand

projects/rocprofiler-compute/src/utils/roofline_calc.py
cfallows-amd: We have updated our support chart and created a gfx115x-specific benchmark to measure BF16 through VALU with the dot2 builtin. We are still keeping WMMA benchmarking off for gfx115x series (counter collection issue).

projects/rocprofiler-compute/src/utils/roofline_calc.py
cfallows-amd: Support clarity of enum class- added a Flag type class "OpsSupport" with VALU and MATRIX. Support is identified in SUPPORTED_DATATYPES dict by specifying VALU or MATRIX, or stacking both with the | (OR) operator in bitwise fashion.

projects/rocprofiler-compute/src/utils/roofline_calc.py
cfallows-amd: Using the above mentioned OpsSupport Flag type enum class, using "in" keyword to check if VALU or MATRIX in the SUPPORTED_DATATYPES[gpu_arch][datatype] value. We only ever check for one ops type at a time so this works  with a bitwise check, even if we stacked "VALU | MATRIX" as the value.

PR #7989, opened 2026-06-30
fix(clr): make hipFree's global de-index non-fatal and identity-checked

### LINKED ISSUE #6224: [Issue]: rocclr "Memobj map does not have ptr" abort on 2nd Z-Image gen (Windows, dGPU + iGPU)
### Summary
On Windows, `amdhip64_7.dll` (rocclr) aborts the whole process on the **second**
image generation of a ComfyUI Z-Image workflow with:

```
:0:...\rocclr\device\device.cpp:373 : <ts> us: Memobj map does not have ptr: 0x2adf8000
```

The first generation succeeds; the second aborts. This is not a recoverable
`hipError_t` — it is a hard `guarantee()`/abort that kills the process.

### Key condition: more than one HIP device must be visible
This only reproduces when **both** the discrete GPU **and** the CPU's integrated
GPU are registered as HIP devices:

- Discrete: **AMD Radeon RX 7900 XTX** (gfx1100) → `cuda:0`
- Integrated: **Radeon 780M** in the Ryzen 8700G (gfx1103) → `cuda:1`

ComfyUI (Desktop) enumerates both by default, so it aborts on the 2nd run.

**Diagnostic:** launching with `HIP_VISIBLE_DEVICES=0` (discrete GPU only) makes
the crash disappear entirely — dozens of back-to-back generations succeed. This
points squarely at the multi-device, per-device-VA free path.

### Root cause
In the free path, `ihipFree()` resolves an allocation via `FindMemObj()`, which
matches **any** address in `[base, base + size)` and, on Windows with multiple
devices, falls back to per-device VA maps with host-pointer/base adjustment. It
then called `RemoveMemObj(ptr)`, which erased the global map only by the **exact
base key** and `guarantee()`d the result. A pointer resolved by range or adjusted
base freed correctly but missed the exact-key erase, so the guarantee aborts.
With a single device the per-device VA fallback is not exercised, so the
asymmetry never surfaces.

### Steps to reproduce
1. Windows 11, a system with a discrete AMD GPU **and** an active integrated
   Radeon GPU (e.g. Ryzen 8700G / 780M), ROCm 7.2.1 wheels, PyTorch `2.9.1+rocm7.2.1`.
2. ComfyUI with the **Z-Image (Turbo)** workflow; do **not** set
   `HIP_VISIBLE_DEVICES` (both GPUs visible — the default).
3. Generate one image (succeeds), then generate again.
4. The process aborts on the 2nd generation with `Memobj map does not have ptr: 0x...`
   (the address varies per run).

Confirmed with a minimal API driver (two `EmptySD3LatentImage` → KSampler runs):
crashes on gen 2 with both GPUs visible; never crashes with `HIP_VISIBLE_DEVICES=0`.

### Environment
- **OS:** Windows 11
- **GPUs:** RX 7900 XTX (gfx1100) + Ryzen 8700G integrated Radeon 780M (gfx1103)
- **ROCm:** 7.2.1 (TheRock `rocm-sdk` wheels), PyTorch `2.9.1+rocm7.2.1`
- **Driver:** AMD Adrenalin 26.6.2
- **App:** ComfyUI (Desktop) — Z-Image Turbo template

### Expected
A missing/absent pointer on a user-facing free should not abort the process.

### Fix
Proposed in ROCm/rocm-systems#7989 — makes `MemObjMap` removal range-aware
(symmetric with `FindMemObj`) and adds a non-fatal `TryRemoveMemObj` for
user-facing frees. Verified on this exact multi-GPU Windows/ROCm 7.2.1 setup:
`amdhip64_7.dll` rebuilt from the `rocm-7.2.1` tag + the fix → both generations
complete with no abort (A/B tested against the unpatched 7.2.1 DLL, which
crashes on generation 2).

seanthegeek:
## Associated JIRA ticket number/Github issue number

Fixes ROCm/TheRock#6224 (cross-repo — will not auto-close on merge).

### Likely closes

These reports all hit the same `guarantee()` abort from the asymmetric free
path — `Memobj map does not have ptr: 0x...` out of
`clr/rocclr/device/device.cpp` — and should be resolved by making the
user-facing free path non-fatal. They are cross-repo, so none will auto-close
on merge:

- Comfy-Org/ComfyUI#11551 — Z-Image Turbo VAE decode aborts on Windows / RX 7900
  XTX / ROCm 7.1 at `device.cpp:379`, same `Requested to load AutoencodingEngine`
  point. Matches this PR's setup closely.
- Comfy-Org/ComfyUI#11622 — reported as an "unhandled OOM", but the console shows
  the same `device.cpp:379` abort on Windows / RX 7900 XTX for Z-Image (and
  Wan2.2) workflows.
- patientx/ComfyUI-Zluda#372 — same abort on Windows / RX 7800 XT running Flux
  above 896×896. Reached through the ZLUDA translation layer rather than native
  HIP, so likely-but-not-certain.
- nod-ai/amd-shark-ai#2165 — same abort (`ptr: 0x0`) on Linux / gfx942 over many
  iterations; notably `HIP_VISIBLE_DEVICES=0` makes it pass, the same
  multiple-visible-device correlation described below.

## What type of PR is this? (check all applicable)

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Continuous Integration

## What were the changes?

- `ihipFree()` now de-indexes the pointer through a new
  **`MemObjMap::TryRemoveMemObj(ptr, memory_object)`**: non-fatal, range-aware
  (the same `[base, base + size)` test `FindMemObj` uses), and
  **identity-checked** — it erases the covering global-map entry only if that
  entry *is* the object being freed, decided under the map lock. A pointer that
  is legitimately absent from the global map (per-device VA on Windows,
  external memory) degrades to a logged warning instead of aborting the
  process, and an unrelated live allocation whose address range numerically
  overlaps the freed pointer is never removed by mistake.
- **`RemoveMemObj` is unchanged from develop** (exact-key erase + fatal
  `guarantee`). Every existing caller passes the tracked base key, so the
  strict corruption detector stays exactly as-is for all non-user-facing
  removal paths.
- `FindAndRemoveMemObj` is refactored onto a shared helper
  (`EraseCoveringMemObj` / `EraseCoveringMemObjIf` in
  `rocclr/device/memobjmap_erase.hpp`) with identical behavior, so the
  range-erase logic lives in one unit-tested place.
- Unit tests live in `rocclr/tests/unit/`, gated behind
  `option(BUILD_TESTS OFF)` — the same switch that gates the OpenCL `ocltst`
  suite — and require only GoogleTest. 10 cases cover exact-base, interior,
  one-past-end, gap, below-all, and empty-map erases, plus the
  identity-predicate behavior including the wrong-erase regression (a covering
  entry that is not the expected object must survive untouched).

## Why are these changes needed?

The free path was asymmetric. `ihipFree()` resolves allocations via
`FindMemObj()`, which matches any address in `[base, base + size)` and, on
Windows with **multiple HIP devices**, falls back to per-device VA maps with
base/host-pointer adjustment. It then called `RemoveMemObj(ptr)`, which erased
the global map only by the exact base key and `guarantee()`d the result — so any
pointer resolved by range or adjusted base freed correctly but missed the erase
and aborted the process:

```
rocclr/device/device.cpp:373 : Memobj map does not have ptr: 0x...
```

This reproduces as a crash on the **second** ComfyUI Z-Image generation on
Windows/ROCm 7.2.1, but only when a discrete GPU and an integrated GPU are both
visible (e.g. RX 7900 XTX + Ryzen 8700G/780M); `HIP_VISIBLE_DEVICES=0` hides it.

Making the user-facing free non-fatal fixes the false abort while leaving the
`guarantee` untouched as a true corruption detector everywhere else. The
identity check exists because per-device VA ranges can numerically overlap an
unrelated allocation's range in the global map (the reason the per-device maps
exist at all) — a blind range-erase there could silently de-index a live
allocation; with the check, that case is a warning and nobody else's entry is
touched. The warning also logs the pointer and object, which pins down the
allocation class on any future report.

Verified end-to-end: `amdhip64_7.dll` rebuilt from the `rocm-7.2.1` tag with an
earlier revision of this change (same `ihipFree` mechanism, before the identity
check was added on review), dropped into the ROCm 7.2.1 PyTorch runtime, runs
the Z-Image workflow across repeated generations with both GPUs visible and no
longer aborts; the unpatched 7.2.1 DLL crashes on generation 2 under the
identical setup. Re-verification with the current head is planned.

## Updated CHANGELOG?

- [x] Yes
- [ ] No, Does not apply to this PR.

## Added/Updated documentation?

- [ ] Yes
- [x] No, Does not apply to this PR.

## Additional Checks

- [x] I have added tests relevant to the introduced functionality, and the unit tests are passing locally.
- [x] Any dependent changes have been merged.

---


seanthegeek:
This PR is in a good state to merge

seanthegeek:
Can someone at AMD please look at this? It resolves a longstanding bug.

projects/clr/hipamd/src/hip_memory.cpp
```diff
@@ -175,7 +175,13 @@ hipError_t ihipFree(void* ptr) {
...
       // Non SVM memory free, such as external memory
       if (memory_object->getSvmPtr() == nullptr) {
-        amd::MemObjMap::RemoveMemObj(ptr);
+        // User-facing free: the pointer may live only in a per-device VA map
+        // (Windows) or be external memory not indexed in the global map, so a
+        // miss is a logged no-op rather than a fatal guarantee. The erase is
+        // identity-checked against memory_object so an unrelated allocation
+        // whose range covers ptr is never de-indexed. The object itself is
+        // still released exactly once below.
+        amd::MemObjMap::TryRemoveMemObj(ptr, memory_object);
```
mangupta: 🤖 AI-assisted pre-review (for @mangupta)

**[BLOCKING]** Failed global de-index now falls through to release(), leaving a stale amd::Memory* in the lookup maps

When `TryRemoveMemObj(ptr, memory_object)` finds no covering entry for `ptr` that is identical to `memory_object`, it logs a warning and returns false, and `ihipFree` unconditionally calls `memory_object->release()` on the next line. Any tracking entry that still points at `memory_object` — the global `MemObjMap_` entry recorded under a base key that does not cover `ptr` (the adjusted/host-pointer case), or the per-device VA entry that actually resolved the lookup — survives the release and becomes a dangling pointer that a later `FindMemObj`/`FindDevMemObj` can dereference.

**Suggested fix:** Make the free path remove every alias that resolves to `memory_object` before releasing it: de-index the per-device entry (the `RemoveDevMemObj`-style path) and the global entry under the object's own base key (as `Device::svmFree` does, by keying on the object's base rather than the user pointer), all under the map lock; if no alias can be removed and the object is genuinely tracked, either keep the object alive or fail the free with an error instead of releasing it. At minimum, check `TryRemoveMemObj`'s `bool` result and do not `release()` when a tracked alias could not be removed.

mangupta: 🤖 AI-assisted pre-review (for @mangupta)

**[IMPORTANT]** hipFree on an interior pointer now silently frees the whole allocation instead of failing

`TryRemoveMemObj` erases by the `[base, base + size)` range test, so `hipFree(base + offset)` for any offset inside a live allocation now matches the covering entry, passes the identity check (the same object `getMemoryObject()` resolved), is de-indexed, and is released — returning `hipSuccess`. A pointer that was never returned by an allocation API silently destroys the entire allocation, and the rest of the application keeps using addresses in that range.

**Suggested fix:** Distinguish the two miss classes the PR conflates: an adjusted/per-device base (legitimate, must still free) from a true interior pointer (invalid). Compare the resolved offset — e.g. require `offset == 0` / `ptr == memory_object`'s tracked base for the non-SVM path — and return `hipErrorInvalidDevicePointer` for a genuine interior address rather than de-indexing and releasing the covering allocation.

projects/clr/rocclr/CMakeLists.txt
```diff
@@ -10,3 +10,9 @@ endif()
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 include(ROCclr)
+
+option(BUILD_TESTS "Enable building ROCclr unit tests" OFF)
+if(BUILD_TESTS)
+  enable_testing()
+  add_subdirectory(tests/unit)
+endif()
```
mangupta: 🤖 AI-assisted pre-review (for @mangupta)

**[IMPORTANT]** Unnamespaced BUILD_TESTS option plus find_package(GTest REQUIRED) can break superbuild configuration

`option(BUILD_TESTS "Enable building ROCclr unit tests" OFF)` declares a generic, unnamespaced name. When CLR is configured as part of a larger build (ROCm superbuild/TheRock, or any parent that already sets `-DBUILD_TESTS=ON` for an unrelated component), `option()` is a no-op against the existing cache entry, `add_subdirectory(tests/unit)` is taken, and `find_package(GTest REQUIRED)` aborts configuration of the entire CLR build on any machine without an installed GoogleTest.

**Suggested fix:** Namespace the switch (e.g. `ROCCLR_BUILD_TESTS`, or gate on `BUILD_TESTS AND PROJECT_IS_TOP_LEVEL`) so an unrelated parent cannot turn it on implicitly, and use `find_package(GTest)` with a status message that skips the target when GoogleTest is absent instead of `REQUIRED`. Move `enable_testing()` to the top-level CLR CMakeLists (or rely on the parent's) so the registered test is visible to `ctest` from the build root.

mangupta (commented):

PR #7990, opened 2026-06-30
feat(rocr): expose graphics-aware VRAM availability query

### LINKED ISSUE #7993: [Issue]: HSA_AMD_AGENT_INFO_MEMORY_AVAIL over-reports free VRAM on discrete GPUs (ignores DRM graphics)
### Summary
On discrete AMD GPUs, `hipMemGetInfo` / `hsa_agent_get_info(HSA_AMD_AGENT_INFO_MEMORY_AVAIL)` reports more free VRAM than is actually available. It is backed by KFD `AMDKFD_IOC_AVAILABLE_MEMORY`, which counts only ROCm/KFD allocations and ignores DRM graphics allocations (Vulkan/GL/desktop compositor). With a graphics session holding VRAM the gap can be multiple GiB, and frameworks that size GPU offload from this value over-commit and OOM at load.

- Downstream report: https://github.com/ggml-org/llama.cpp/issues/24906 (a llama.cpp-side workaround https://github.com/ggml-org/llama.cpp/pull/25123 was closed in favor of fixing this in ROCm).
- Prior discrepancy report: ROCm/ROCm#1909 (2023), closed as a metric-difference note for a small idle delta; the actionable problem here is the multi-GiB over-report under graphics load causing OOM.

### Root cause
`HSA_AMD_AGENT_INFO_MEMORY_AVAIL` -> `KfdDriver::AvailableMemory` -> `hsaKmtAvailableMemory` -> `AMDKFD_IOC_AVAILABLE_MEMORY` (KFD compute-available; excludes DRM graphics). `rocm-smi` reads amdgpu `mem_info_vram_used` (graphics-inclusive).

### Reproduce (gfx1100 RX 7900 GRE, ROCm 7.2.1)
| state | hipMemGetInfo free | rocm-smi / amdgpu |
|---|---|---|
| idle | 16332 MiB | 16170 MiB |
| ~11 GiB resident | 5696 MiB | 5420 MiB |

### Fix
In ROCR `GpuAgent::GetInfo` (`HSA_AMD_AGENT_INFO_MEMORY_AVAIL`, `projects/rocr-runtime` in this super-repo), on discrete GPUs cross-check amdgpu's graphics-inclusive VRAM usage via the agent's existing libdrm handle and take the conservative value; APUs keep the KFD UMA value.

Fix PR: #7990

liminfei-amd:
## Summary

This PR fixes the free-VRAM over-reporting problem without changing the existing `HSA_AMD_AGENT_INFO_MEMORY_AVAIL` contract.

On discrete GPUs, the current `MEMORY_AVAIL` value is KFD/compute-oriented and does not include DRM graphics allocations such as Vulkan, GL, or the desktop compositor. That can make frameworks overestimate available VRAM and over-offload models.

The initial version of this PR tried to clamp `MEMORY_AVAIL` directly to amdgpu's graphics-inclusive heap usage. CI exposed why that is not safe: `MEMORY_AVAIL` is also used as a compute allocator accounting value, and existing tests expect it to recover synchronously after HIP/ROCR alloc/free. amdgpu/TTM heap usage can lag after frees, so changing the existing attribute breaks that contract.

The updated approach keeps `HSA_AMD_AGENT_INFO_MEMORY_AVAIL` unchanged and adds a separate opt-in query:

```c
HSA_AMD_AGENT_INFO_MEMORY_AVAIL_GRAPHICS_AWARE
```

On Linux discrete GPUs with an available libdrm handle, it returns the graphics-inclusive VRAM free value from `amdgpu_query_heap_info(..., AMDGPU_GEM_DOMAIN_VRAM, ...)`. Platforms where this query is not supported return `HSA_STATUS_ERROR_INVALID_ARGUMENT`.

## Validation

Validated on a gfx1100 lab system with a locally built `libhsa-runtime64.so` from this branch:

- Existing `HSA_AMD_AGENT_INFO_MEMORY_AVAIL` still recovers immediately across HIP alloc/free:
  - `17125343232 -> 16588472320 -> 17125343232`
- The new graphics-aware query is supported and returns a DRM-aligned value.
- With a 2 GiB DRM/amdgpu BO allocated:
  - system runtime: DRM free drops to `14269882368`, but old HIP/HSA free remains `17125343232` (over-report reproduced)
  - updated runtime: graphics-aware query drops to `12122398720`, matching DRM free, while old `MEMORY_AVAIL` remains `17125343232`

This preserves the compute allocator semantics while exposing the conservative graphics-aware value needed by upper layers for capacity planning.

Fixes #7993


liminfei-amd:
**Problem (from the field).** On discrete GPUs, `hipMemGetInfo` / `HSA_AMD_AGENT_INFO_MEMORY_AVAIL` over-reports free VRAM — it counts only KFD/compute allocations and ignores DRM graphics (Vulkan/GL/compositor). A user on the downstream issue (ggml-org/llama.cpp#24906) on the real impact:

> you could have 5 GiB free, but llama.cpp will see your total VRAM and try to overfill it ... it often results in a GPU reset or a total system freeze that requires a restart. I wish there were some kind of early OOM killer for VRAM.

So the over-report can hard-hang the machine, not just OOM. This PR makes `MEMORY_AVAIL` graphics-aware (clamped to amdgpu's VRAM usage, the same source `rocm-smi` reads), validated on gfx1100.

**Why CI is red.** `VirtMemory_Accounting_Test` and `Unit_hipMalloc3D_Basic` do alloc → free → **immediately** re-read `MEMORY_AVAIL` and expect exact recovery. amdgpu reclaims a freed BO asynchronously (KFD frees synchronously), so the immediate re-read is transiently low — only ever **under**-reporting, never re-introducing the over-report. It settles to baseline once TTM catches up; the tests just race it.

**Fix (follow-up commit): await the reclaim before the check.**

`virtual_memory.cc` (final check, L1247):
```diff
-  ASSERT_SUCCESS(hsa_agent_get_info(agent, (hsa_agent_info_t)HSA_AMD_AGENT_INFO_MEMORY_AVAIL, &amount_current));
-  ASSERT_EQ(amount_begin, amount_current);
+  for (int i = 0; i < 100; ++i) {  // await async DRM reclaim; MEMORY_AVAIL only ever lags low
+    ASSERT_SUCCESS(hsa_agent_get_info(agent, (hsa_agent_info_t)HSA_AMD_AGENT_INFO_MEMORY_AVAIL, &amount_current));
+    if (amount_current >= amount_begin) break;
+    std::this_thread::sleep_for(std::chrono::milliseconds(10));
+  }
+  ASSERT_EQ(amount_begin, amount_current);
```

`hipMalloc3D.cc` (post-`hipFree()` re-read):
```diff
-  HIP_CHECK(hipMemGetInfo(&avail, &tot));
+  for (int i = 0; i < 100; ++i) {  // await async DRM reclaim
+    HIP_CHECK(hipMemGetInfo(&avail, &tot));
+    if (avail >= pavail - height * width * depth - 4 * 1024 * 1024) break;
+    std::this_thread::sleep_for(std::chrono::milliseconds(10));
+  }
```

Both keep their exact leak detection — a real leak never settles, so it still fails. If you'd rather keep `MEMORY_AVAIL` strictly KFD-synchronous, I can surface the graphics-aware value through a separate path instead.

cc @dayatsin-amd @HongleiHuang-amd — you two have the most history on `amd_gpu_agent.cpp`; would appreciate your steer on the `MEMORY_AVAIL` semantics question above.

projects/rocr-runtime/rocrtst/suites/functional/memory_avail_test.cc
```diff
@@ -0,0 +1,80 @@
+// Regression test for graphics-aware available memory on discrete GPUs.
```
cfreeamd, line 7: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** Missing AMD copyright / SPDX-License-Identifier header.

Every other source file in `rocrtst/suites/functional/` carries:
```c
/* Copyright © Advanced Micro Devices, Inc., or its affiliates.
 * SPDX-License-Identifier: MIT */
```
This is required for IP compliance. Should be added before merge.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -2473,6 +2474,20 @@ hsa_status_t GpuAgent::GetInfo(hsa_agent_info_t attribute, void* value) const {
       *((uint64_t*)value) = availableBytes;
       break;
     }
+    case HSA_AMD_AGENT_INFO_MEMORY_AVAIL_GRAPHICS_AWARE: {
+#if defined(__linux__)
+      if (properties_.Integrated == 0 && ldrm_dev_ != nullptr) {
```

**[Nit]** `ldrm_dev_ != nullptr` is always true at this point. `InitLibDrm()` is called unconditionally in the constructor and throws on failure, so `ldrm_dev_` is never null for a live `GpuAgent`. The `Integrated == 0` guard is the meaningful condition. Either remove the null check or add a comment explaining when it could be null.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -2473,6 +2474,20 @@ hsa_status_t GpuAgent::GetInfo(hsa_agent_info_t attribute, void* value) const {
...
+        struct amdgpu_heap_info heap = {};
+        if (amdgpu_query_heap_info(ldrm_dev_, AMDGPU_GEM_DOMAIN_VRAM, 0, &heap) == 0 &&
+            heap.heap_size > 0) {
```
cfreeamd, line 2555: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** `heap.heap_size > 0` conflates two distinct failure modes. When `amdgpu_query_heap_info()` succeeds but returns `heap_size == 0`, the code falls through to `HSA_STATUS_ERROR_INVALID_ARGUMENT` — the same status as "platform doesn't support this query at all" (Integrated == 1, or DRM missing). A caller cannot tell the difference between "DRM query succeeded but returned zero VRAM" and "this attribute is unsupported on this platform."

In practice `heap_size == 0` is pathological, but a comment explaining the intent of the guard would clarify the semantics.

cfreeamd (commented):
The `hsa_ext_amd.h` enum definition and doc block look correct; the existing `MEMORY_AVAIL` path is untouched. DRM path gating and include guards are correct.

liminfei-amd:
Updated the PR to address the review:

- Added the `SPDX-License-Identifier: MIT` header to `memory_avail_test.cc`.
- Dropped the redundant `ldrm_dev_ != nullptr` check (`InitLibDrm()` runs in the ctor and throws on failure, so a live `GpuAgent` always has a valid handle) and documented the invariant.
- Added a comment on the `heap_size == 0` fall-through in the GRAPHICS_AWARE path.

cfreeamd (approved):
lgtm, assuming it passes CI. Probably should get @dayatsin-amd's blessing since it's adding to the api, to make sure this right way to address this.

liminfei-amd:
Rebased onto current `develop` and registered `MemoryAvailGraphicsAware_NotOverReported` in the ROCr `Core` group and the standard/comprehensive/full CTest categories, so CI now actually executes the regression.

The previous red jobs were unrelated stale-base failures: an existing VM aliasing SIGBUS, Windows `OccupancyTest.exe` discovery failure, and the WSL `rocdxg` link issue fixed upstream by #8343. The rebased runtime builds locally, and the registered regression passes on gfx1201 with the real platform configuration.

dayatsin-amd (changes requested):
NACK. 
This should be fixed in the KFD memory available IOCTL. 

The KFD memory available IOCTL should factor in memory usage from DRM interface and DRM API (amdgpu_query_heap_info)  should also factor in memory usage from KFD interface.

liminfei-amd:
Thanks for the direction. I agree that this belongs in shared KFD/DRM
accounting rather than a second ROCr query, so I will close this PR and retain
#7993 as the symptom and evidence record. Any follow-up will target
`amdgpu_amdkfd_get_available_memory()` and preserve the ioctl's immediate
alloc/free semantics while validating both interfaces' usage.

PR #7991, opened 2026-06-30
feat(rccl): Enable AllGatherV as a fusion of grouped broadcasts

rahulvaidya20:
## Motivation

Enable transparent runtime fusion of grouped `ncclBroadcast` calls (>=2 distinct roots in one group) into a single `ncclFuncAllGatherV` ring kernel (NCCL 2.29.2's scalable allgatherv pattern). Gated by `NCCL_ALLGATHERV_ENABLE` (default off).

## Technical Details

**Core fusion path:**
- Task producer in `enqueue.cc` intercepts grouped broadcasts (when `NCCL_ALLGATHERV_ENABLE=1`) and enqueues `ncclTaskBcast` instead of `ncclTaskColl`, scheduled by `scheduler/allgatherv_sched.cc` into a ring `all_gather_v.h` device kernel. Kernel + scheduler core kept byte-identical to upstream NCCL for clean future syncs.
- Proxy steps wired from per-bcast slice counts (`specifics.bcast.sendSlices` / `recvSlices`).
- WarpSpeed excluded for AllGatherV-fused plans — kernel does not implement WarpSpeed's warp-level channel distribution, which caused launch faults on gfx950.

**Correctness fixes (applied over initial implementation):**
- **Ring desync:** fused works were capped at `NCCL_MAX_DEV_WORK_BATCH_BYTES` (192 = 4 bcast works). A 5th work split into a second device batch, which the kernel runs as an independent ring pass — desyncing the ring across ranks and corrupting recv buffers. Fix bounds bcast batching by device shmem capacity so all fused works stay in one batch.
- **Batch loader truncation:** AllGatherV packs up to 64 works per batch (vs 1-2 for coll/P2P), exceeding the loader subgroup size. Added stride loop in `loadWorkBatchToShmem` with a break for coll/P2P to preserve single-pass behavior.
- **Device workStorage size / warp rotation:** fixed `ncclMaxDevWorkBatchBytes` returning wrong size at device compile time on HIP (`NCCL_CUDA_ARCH=0`); fixed warp rotation using `2*WARP_SIZE`.
- **FIFO recycling race:** `cudaEventRecord` on `finishedEvent` was missing before the recycling callback was enqueued, allowing the callback to fire before kernels completed.
- **Budget workBytes:** `ncclTestBudget` call in scheduler was missing the `nChannels` multiplier on bcast byte accounting.
- **Uninitialized work struct:** added `memset(work, 0, sizeof(*work))` before populating `ncclDevWorkBcast` from the memory pool.
- **`ncclMaxDevWorkBatchBytes` dead code:** HIP early-return left the CUDA arch ladder unreachable (no `#else`). Fixed to proper `#if/#else/#endif`.
- **`comm->opCount`:** bcast path in `collTaskAppend` was not incrementing `comm->opCount`, unlike the coll and P2P paths.

**Downgrade path (BcastPeers==1):**
- Single-root broadcast in an AllGatherV-enabled group downgrades back to plain `ncclFuncBroadcast` in `ncclPrepareTasks`. Profiler field propagation on the downgrade path tracked in AICOMRCCL-1509 (upstream NCCL has the same gap).

**Default off:**
- `NCCL_ALLGATHERV_ENABLE` defaults to 0 pending profiler plugin compatibility fix (NCCL profiler support for ncclBroadcast ops is broken when fusion is active).

## JIRA ID 

JIRA ID: AICOMRCCL-467

## Test Plan

**Functional (unit tests):**
- `GroupCall.MultiRootBroadcast`: AllGatherV fused path, all GPU counts SP + MP.
- `GroupCall.MultiRootBroadcast` with `NCCL_ALLGATHERV_ENABLE=0`: individual broadcast fallback path.
- `GroupCall.SingleRootBroadcastDowngrade`: BcastPeers==1 downgrade to plain broadcast.
- `GroupCall.MultiRootBroadcastConsecutive`: two consecutive AllGatherV group calls on same communicator, validates `reclaimPlannerState` reset.

**Performance (rccl-tests, msg 16K–16G, `-f 2`, `-N 10` stress cycles, scaling 1 → 16 nodes @ 8 GPUs/node)** on two multi-node AMD GPU systems: gfx942 over an InfiniBand fabric and gfx950 over a RoCE fabric.
- *Regression:* full collective sweep (all_reduce, all_gather, broadcast, reduce_scatter, reduce, alltoall, alltoallv, scatter, gather) with the feature **disabled**, this branch vs `develop`.
- *Feature:* grouped/fused broadcast with `NCCL_ALLGATHERV_ENABLE` **off vs on**.

## Test Result

**Functional:**
- All four new `GroupCall` tests PASS across 1-8 ranks SP + MP.
- nWorks 5/6/8 multi-chunk correctness PASS.

**Performance:**
- *Regression:* no regression against `develop` — every collective stays within run-to-run noise (bus-bandwidth means within ±2%) across all node counts (1–16) on both gfx942/InfiniBand and gfx950/RoCE, feature disabled.
- *Feature:* enabling fused grouped broadcast delivers substantial large-message bus-bandwidth gains at multi-node scale (mean +12–25%, peaks approaching ~2× on gfx950/RoCE), with the expected minor small-message overhead below the fusion size crossover.

*Note: a follow-up tuning PR is planned to refine the size/algorithm crossover thresholds (small-message fusion overhead and per-fabric channel tuning) so the feature is a net win across the full message-size range prior to enabling it by default.*

## Submission Checklist


thomas-huber:
@rahulvaidya20 can we add some documentation of the new env var to the .rst for env vars? Similar to how you did for previous validation/enablement tickets (I thik)

projects/rccl/src/include/device.h
```diff
@@ -533,7 +533,13 @@ constexpr size_t ncclDevWorkSize(enum ncclDevWorkType type) {
 }
 
 __host__ __device__ constexpr int ncclMaxDevWorkBatchBytes(int cudaArch = NCCL_CUDA_ARCH) {
+#if defined(__HIP_PLATFORM_AMD__) || defined(__HIPCC__)
```
thomas-huber, line 544: This ungated change grows every AMD kernel's static LDS ~24.7→39.7 KiB (measured on gfx950, still 123 KiB headroom, no launch issue). Please add a comment explaining the HIP value.

thomas-huber: @alex-breslow-amd @corey-derochie-amd is this problematic?

projects/rccl/src/device/common.h
```diff
@@ -303,12 +309,14 @@ __device__ __forceinline__ void loadWorkBatchToShmem(int tid, int tn, struct ncc
...
       dst += (workCursor + dstWork) * workSize + packInWork * 16;
       *(ulonglong2*)dst = tmp;
+      if (!isBcast) break; // coll/P2P fit in one pass; only AGV-fused bcast strides
     }
     workCursor += nWorks;
 
     if (batch.nextExtends) {
       batchIx += batch.nextJump;
-      tid -= 64; // Rotate threads so we use the next two warps for next batch struct.
+      // Rotate by 2*WARP_SIZE (not hardcoded 64) so the next batch uses two warps correctly.
```
thomas-huber: Seems like we are now at 128 vs the old 64. I believe this is a behavior change that is not exercised by the unit tests. Hopefully the perf impact, if any, would've been reflected in the scaling multi-node regression experiment you performed? @rahulvaidya20

projects/rccl/src/include/device.h
rahulvaidya20: The guard was needed because `NCCL_CUDA_ARCH` is 0 at device compile time on HIP (no `__CUDA_ARCH__` equivalent), causing the function to return 1 KiB instead of 16 KiB. Newer Nvidia architectures use 16KiB. The bcast work batches were sized too small and would crash without increasing this value. L537-538 already has the comment.

projects/rccl/src/device/common.h
rahulvaidya20: On NVIDIA, `WARP_SIZE` is 32, so the hardcoded 64 correctly represents 2 warps. However, on certain AMD gfx architectures where `WARP_SIZE` is 64, the hardcoded 64 would represent only 1 warp, causing the next batch struct read to alias the current one. Using `2*WARP_SIZE` correctly rotates threads by 2 warps on all gfx targets.

The stride loop itself only activates for `workType == ncclDevWorkTypeBcast`, the break at the end of the first iteration keeps coll/P2P on the original single-pass path, so there is no behavior change for any existing collective. No regressions observed.

projects/rccl/src/device/common.h
```diff
@@ -268,9 +268,15 @@ __device__ __forceinline__ void loadWorkBatchToShmem(int tid, int tn, struct ncc
...
 
-    // We can only assume we have 64 threads, which means we can read at most 1024 bytes
-    // here which is the per batch maximum.
-    if (tid < nPacks) {
+    // AGV bcast fusion packs up to 64 works (offsetBitset bits) per batch; at 3 packs/work
+    // (48B/16B each) that is 192 packs, which can exceed the loader subgroup size
+    // (tn = threadPerBlock - 2*WARP_SIZE), so bcast requires a stride loop. Coll/P2P
+    // hold at most 1-2 works and always fit in one pass; the break below enforces that.
+    // tid is preserved (not pk) so the nextExtends warp-rotation below stays correct.
+    const int bcastPacks = sizeof(struct ncclDevWorkBcast)/16; // 3 packs/work
```
PJAvinash: While sizeof(struct ncclDevWorkBcast) > 16 in our case, We need to assure this is always the case by definition. ( alignas (16) or with compile time assertions)

dstWork = pk/bcastPacks may result in division by zero errors and it can be hard to trace.

projects/rccl/src/scheduler/allgatherv_sched.cc
```diff
@@ -80,11 +66,23 @@ ncclResult_t ncclScheduleBcastTasksToPlan(struct ncclComm* comm, struct ncclKern
     if (proto == NCCL_PROTO_LL128) chunkSize = (chunkSize / NCCL_LL128_LINEELEMS) * NCCL_LL128_DATAELEMS;
     size_t grainSize = rcclProtoGrainSize(proto, comm);
     nChannels = tcoll.nMaxChannels;
+#ifdef ENABLE_WARP_SPEED
+    // AllGatherV launches grid=nChannels without WarpSpeed, so undo the tuner's
+    // warpSpeedChannelMultiplier inflation (else the oversized grid faults on gfx950).
+    if (comm->warpSpeedChannelMultiplier > 1) {
+      nChannels = std::max(1, nChannels / comm->warpSpeedChannelMultiplier);
```
PJAvinash: ncclTestBudget() is invoked with a value nChannels and, it is being overwritten in subsequent lines. Looking at final value of nChannels and, how ncclTestBudget returned true, It can be difficult to reason where this modification happens. We should prefer to keep things unmutable as much as possible.  Add logging here or make sure nChannels remains constant throughout the execution of this function, and delegate the responsibility of modification to some other block of code.

projects/rccl/src/enqueue.cc
```diff
@@ -54,19 +54,7 @@ static ncclKernelMatch const ncclKerns[3] = {
...
-  case NCCL_PROTO_LL128:
-    return comm->WarpSize * NCCL_LL128_SHMEM_ELEMS_PER_THREAD * comm->ll128DataElems * sizeof(uint64_t) /
-           comm->ll128LineElems;
-  case NCCL_PROTO_SIMPLE:
-    return 512;
-  default:
-    return -1;
-  }
-}
+
```
PJAvinash: You may remove this unwanted change.

projects/rccl/src/enqueue.cc
```diff
@@ -269,6 +265,9 @@ static void addWorkBatchToPlan(struct ncclComm* comm, struct ncclKernelPlan* pla
     // of the same round since they would use the same connections.
     chan->wipBatch.p2pRounds[chan->wipBatch.nP2ps++] = p2pRound;
   }
+  if (workType == ncclDevWorkTypeBcast) {
+    chan->wipBatch.nBcasts += 1;
```
PJAvinash, line 269: You may reduce a branch here.

PJAvinash (commented):
Added few comments. 

Also testcases can be included to validate the behavior for different values of NCHANNELS, unaligned buffer sizes (data).

projects/rccl/src/enqueue.cc
rahulvaidya20: This is directly from upstream.

projects/rccl/src/enqueue.cc
rahulvaidya20: Done.

projects/rccl/src/scheduler/allgatherv_sched.cc
rahulvaidya20: Agreed. Added logging.

projects/rccl/src/device/common.h
rahulvaidya20: Added an assert now.

rahulvaidya20:
> Added few comments.
> 
> Also testcases can be included to validate the behavior for different values of NCHANNELS, unaligned buffer sizes (data).

Added new tests.

thomas-huber (approved):

alex-breslow-amd (approved):
LGTM, synced with @rahulvaidya20 regarding testing on gfx942 and gfx950.  Satisfied with performance testing. 

Since this PR is higher risk, please refrain from merging until all CI has passed.

rahulvaidya20:
Merged directly in the july drop branch #9963.

PR #8000, opened 2026-06-30
Users/anujshuk amd/link shared lib and mock database tests

anujshuk-amd:
This pull request improves the way the `libprofiler-hub.so` shared library handles its bundled SQLite dependency to prevent symbol collisions with other SQLite versions in the same process. The bundled SQLite symbols are now sealed (hidden), and a regression test ensures this property is maintained. These changes make the library safer to use in environments where multiple components may include their own SQLite versions.

**Bundled SQLite symbol sealing:**
* The bundled `sqlite3_*` symbols are now hidden from export in `libprofiler-hub.so` by setting hidden visibility during compilation (`-fvisibility=hidden`) and using the linker option `--exclude-libs` to prevent them from being exported. [[1]](diffhunk://#diff-cc47e3a9b14388381b9af0aad27a286d14f4a77ee1d43c6abcc74ed70b70628bL148-R154) [[2]](diffhunk://#diff-97004d9ad3a7bcf8eb48287d46d1bc9feea9fa337459b9df791cb41ec06f8569R219-R228)
* The changelog documents that `libprofiler-hub.so` no longer exports the bundled `sqlite3_*` symbols, preventing symbol collisions.

**Testing and build system updates:**
* A new regression unit test (`symbol_seal_test.cpp`) is added to verify that no `sqlite3_*` symbols are exported from the built shared library. This test inspects the build artifact using `nm` and fails if any such symbols are found. [[1]](diffhunk://#diff-c9327bf9c2cb48455f13117f41064ae8dc11a2d9d8a0e0761169307a67e63ff1R1-R142) [[2]](diffhunk://#diff-d9197fa803aec35a3a9464bffd4cdc7dd30a40d79df46c58252a472e3624920dR19-R30)
* The test build is updated to include the regression guard only when the system SQLite is not used, and the path to the built shared library is injected for the test to inspect.

These changes collectively ensure that the library does not leak its bundled SQLite symbols, preventing runtime conflicts with other components.## Motivation

## Submission Checklist

### DIFF
--- a/profilers/profiler-hub/CHANGELOG.md
+++ b/profilers/profiler-hub/CHANGELOG.md
@@ -55,6 +55,11 @@ downstream consumer of the library.
   callers that need bundled builds can simply remove the system package or set
   `CMAKE_DISABLE_FIND_PACKAGE_<name>=ON`.
 
+### Changed
+
+- `libprofiler-hub.so` no longer exports the bundled `sqlite3_*` symbols (sealed via
+  hidden visibility + `--exclude-libs`), preventing collisions with other SQLite versions.
+
 ## [0.1.0] - 2026-05-05
 
 Initial release.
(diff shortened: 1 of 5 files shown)


anujshuk-amd:
changes taken with other PR

Issue #7973, opened 2026-06-29
[CI] Core-only builds select profiler test legs (aqlprofile/rocprofiler-*) that are never built


JeniferC99:
### Summary

For PRs that change only `projects/amdsmi/**` (and other `core` subtrees), TheRock CI builds **core-only** (`-DTHEROCK_ENABLE_CORE=ON -DTHEROCK_ENABLE_ALL=OFF`) but the "Configure test matrix" step still selects the **aqlprofile / rocprofiler-compute / rocprofiler-sdk / rocprofiler-systems** test legs. Those components are never built, so the legs fail at artifact/test setup because their test-support files don't exist. This makes core/amdsmi PRs show red CI for reasons unrelated to the change.

Introduced by https://github.com/ROCm/rocm-systems/pull/7813

### Affected PRs (reproduced)

- #7870 — run [28391115791](https://github.com/ROCm/rocm-systems/actions/runs/28391115791)
- #7942 — run [28384092658](https://github.com/ROCm/rocm-systems/actions/runs/28384092658)

Both build with `THEROCK_ENABLE_CORE=ON` / `THEROCK_ENABLE_ALL=OFF` and fail the same four legs.

### Root cause

`.github/scripts/therock_matrix.py` — the `core` project group lists the profiler projects under `projects_to_test`, but a core-only build doesn't produce them:

```python
"core": {
    "cmake_options": ["-DTHEROCK_ENABLE_CORE=ON", "-DTHEROCK_ENABLE_ALL=OFF"],
    "projects_to_test": "aqlprofile, rocprofiler-compute, rocprofiler-sdk, rocprofiler-systems",  # will run sanity test to cover rocminfo and amdsmi
},
```

The build populates only core artifacts (`core-amdsmi`, `core-runtime`, `core-hip`, `rocrtst`, …) — no aqlprofile/rocprofiler artifacts — yet the matrix selects `PROJECTS_TO_TEST: aqlprofile, rocprofiler-compute, rocprofiler-sdk, rocprofiler-systems`.

### Failing steps / errors

```
Test aqlprofile (step "Test"):
  FileNotFoundError: .../build/share/hsa-amd-aqlprofile/run_tests.sh

Test rocprofiler-sdk / -compute / -systems (step "Setup Additional Requirements"):
  File not found: build/share/rocprofiler-sdk/tests/requirements.txt
```

PR #8005, opened 2026-06-30
Adding profiler build to core to catch breaks during submodule bumps

ammallya:
## Motivation

Add build and test of profilers to core to catch profiler breaks on amdsmi changes. https://github.com/ROCm/rocm-systems/issues/7973


JeniferC99 (approved):
lgtm

PR #8006, opened 2026-06-30
rocjitsu: keep L2 coherent for uncached accesses

bjacob:
RocJITsu's L2 model treated Mtype::UC reads and writes as pure bypasses to backing memory. That allowed stale clean lines to remain resident in L2. A failed global atomic compare-and-swap could pull a line into L2, then a later uncached global store to the same address would update backing memory while leaving the stale L2 line in place. Subsequent atomics could keep reading the old value and spin forever.

Fix the UC read/write paths to operate in cache-line chunks, flush and invalidate any resident L2 line first, and then bypass to backing memory. This preserves dirty resident data before UC reads and makes UC stores visible to later cached or atomic accesses.

Add focused L2Cache regression tests for a UC store followed by atomic_rmw, and for a UC read of a dirty resident line.

This was exposed by https://github.com/bjacob/hip-moi gfx950 RocJITsu coverage: the relaxed compare-exchange lock handoff tests passed on real gfx950 but timed out under RocJITsu before this fix.

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l2_cache.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l2_cache.cpp
@@ -72,12 +72,21 @@ void L2Cache::ensure_line(uint64_t addr, uint32_t vmid) {
 }
 
 void L2Cache::read(uint64_t addr, uint8_t *dst, uint32_t size, Mtype mtype, uint32_t vmid) {
+  uint32_t copied = 0;
   if (mtype == Mtype::UC) {
-    send_backing(addr, dst, size, simdojo::MessageOp::READ, vmid);
+    while (copied < size) {
+      const uint64_t ea = addr + copied;
+      const uint32_t line_offset = CacheStore::line_offset(ea);
+      const uint32_t chunk = std::min(size - copied, LINE_SIZE - line_offset);
+
+      flush_line(ea, vmid);
+      send_backing(ea, dst + copied, chunk, simdojo::MessageOp::READ, vmid);
+      copied += chunk;
+    }
     return;
   }
 
-  uint32_t copied = 0;
+  copied = 0;
   while (copied < size) {
     const uint64_t ea = addr + copied;
     const uint32_t line_offset = CacheStore::line_offset(ea);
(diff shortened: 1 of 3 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l2_cache.cpp
```diff
@@ -72,12 +72,21 @@ void L2Cache::ensure_line(uint64_t addr, uint32_t vmid) {
 }
 
 void L2Cache::read(uint64_t addr, uint8_t *dst, uint32_t size, Mtype mtype, uint32_t vmid) {
+  uint32_t copied = 0;
   if (mtype == Mtype::UC) {
-    send_backing(addr, dst, size, simdojo::MessageOp::READ, vmid);
+    while (copied < size) {
+      const uint64_t ea = addr + copied;
+      const uint32_t line_offset = CacheStore::line_offset(ea);
+      const uint32_t chunk = std::min(size - copied, LINE_SIZE - line_offset);
```
kuhar, line 80: Please add a focused regression for the UC chunking path: make two adjacent L2 lines resident/dirty, then do an 8-byte UC read/write starting at base + LINE_SIZE - 4 and verify both halves. The current tests only use aligned 4-byte accesses, so the new loop never executes more than one iteration. For the write case, keep at least one dirty byte outside the UC store so the test proves flush_line preserves untargeted dirty data before the bypass write.

kuhar (commented):
An agent fidning

kuhar (commented):
Have you considered the symmetric case one level up, at L1? The L1 UC paths forward straight to L2 with Mtype::UC but do NOT invalidate the resident L1 line first, unlike the CC paths right next to them which do cache_.invalidate(ea):

- l1_scalar_cache.cpp:53 (store UC), :101 (load UC), :127 (load_bytes UC)
- l1_vector_cache.cpp:68 (read UC), :119 (write UC)

Compare the CC branches (e.g. l1_scalar_cache.cpp:59-64, l1_vector_cache.cpp:74-79) which invalidate the L1 line before delegating to L2. The premise motivating this PR -- a line resident from a prior cached access can shadow a later UC access to the same address -- applies identically at L1: a dirty or stale-clean L1 line can shadow a UC read (returns stale L1 data) or survive a UC write (later evicted over the UC store). Since mtype is derived per-access from the PTE (pte_mtype) and can differ from a prior access to the same line, this is reachable.

I do not think it needs to block this PR (scope is clearly L2), but it is the natural follow-up so the two levels stay consistent. The same flush/invalidate-before-bypass primitive would drop in cleanly. Worth a follow-up PR? Happy to review it.

kuhar (approved):

PR #8007, opened 2026-06-30
fix(rocprofiler-sdk): Make thread trace usable before hsa_init()

minseobshin11:
## Motivation

Starting a device thread trace context before the application calls `hsa_init()`
did not work. Thread trace needs HSA-allocated memory for its packets and trace buffer, so the per-agent resources cannot be built until HSA is registered. A `rocprofiler_start_context()` issued before `hsa_init()` silently dropped the trace and was never retried once HSA came up.

This PR caches the start request and starts the trace automatically once HSA is
available.

## Technical Details

Mirrors the device-counting deferred-start pattern:

* `DeviceThreadTracer::start_context()` returns early when HSA is not yet registered; the context is already recorded in the active-context array.
* `thread_trace::initialize()` (at HSA table registration, after `resource_init()` builds the per-agent resources) replays `start_context()` for every active device thread trace context.

## JIRA ID

JIRA ID : AIPROFSDK-887

## Test Plan

Added integration test `thread-trace-api-agent-before-hsa-init-test`, which starts the device thread trace context during `tool_init` (before `hsa_init()`) and
asserts trace data is collected. Verified existing thread trace tests still pass.

## Test Result

On MI300X (gfx942): `thread-trace-api-agent-before-hsa-init-test`, plus the existing `agent`, `single`, and `multi` thread trace tests all PASS.

## Submission Checklist

- [V] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
@@ -74,10 +74,8 @@ struct cbdata_t
 // operations so we can gate new traces while one is active.
 common::Synchronized<std::optional<int64_t>> client;
 
-// Tracks whether the HSA runtime has been registered with rocprofiler. Thread
-// trace cannot touch HSA (queues, memory pools, packet buffers) before this is
-// true, so start requests issued earlier are cached in the active-context array
-// and replayed from initialize() once HSA becomes available.
+// True once the HSA runtime is registered. Gates start_context() so pre-init
+// start requests are deferred and replayed by initialize().
 std::atomic<bool>&
 hsa_inited()
 {
@@ -483,15 +481,8 @@ DispatchThreadTracer::start_context()
 {
     using corr_id_map_t = hsa::queue_info_session_t::external_corr_id_map_t;
 
-    // The queue controller does not exist until HSA is registered. The request
-    // is already recorded in the active-context array, so defer here and let
-    // initialize() replay start_context() once HSA is available.
-    if(!hsa_inited().load())
-    {
-        ROCP_INFO << "Dispatch thread trace start requested before hsa_init; deferring";
-        return;
-    }
-
+    // Only installs queue-controller callbacks (cached and applied to queues as
+    // they are created), so this is safe to call before hsa_init.
     CHECK_NOTNULL(hsa::get_queue_controller())->enable_serialization();
 
     // Only one thread should be attempting to enable/disable this context
@@ -581,9 +572,8 @@ DeviceThreadTracer::resource_deinit()
 void
 DeviceThreadTracer::start_context()
 {
-    // Before HSA is registered the per-agent resources do not exist yet. The
-    // request is already recorded in the active-context array, so simply return
-    // here; initialize() replays start_context() once HSA is available.
+    // Per-agent resources don't exist until HSA is registered; the request is
+    // cached in the active-context array and replayed by initialize().
     if(!hsa_inited().load())
     {
         ROCP_INFO << "Device thread trace start requested before hsa_init; deferring";
@@ -650,13 +640,11 @@ initialize(HsaApiTable* table)
     // HSA resources now exist; allow start_context() to program the hardware.
     hsa_inited().store(true);
 
-    // Replay any thread trace contexts that the user started before hsa_init().
-    // Those start_context() calls returned early (cached) above, so start them
-    // now that the per-agent resources have been constructed.
+    // Replay device contexts started before hsa_init() (their start_context()
+    // returned early above). Dispatch mode needs no replay.
     for(auto& ctx : context::get_active_contexts())
     {
         if(ctx->device_thread_trace) ctx->device_thread_trace->start_context();
-        if(ctx->dispatch_thread_trace) ctx->dispatch_thread_trace->start_context();
     }
 }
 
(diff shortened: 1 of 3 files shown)


projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
```diff
@@ -472,6 +483,15 @@ DispatchThreadTracer::start_context()
 {
     using corr_id_map_t = hsa::queue_info_session_t::external_corr_id_map_t;
 
+    // The queue controller does not exist until HSA is registered. The request
+    // is already recorded in the active-context array, so defer here and let
+    // initialize() replay start_context() once HSA is available.
+    if(!hsa_inited().load())
```
ApoKalipse-V: I dont think we need the lock on dispatch thread trace. It's not possible to see a dispatch before hsa_init()

The tool.cpp already did enable the context before hsa_init() without issues.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
```diff
@@ -617,6 +646,18 @@ initialize(HsaApiTable* table)
...
+
+    // HSA resources now exist; allow start_context() to program the hardware.
+    hsa_inited().store(true);
+
+    // Replay any thread trace contexts that the user started before hsa_init().
+    // Those start_context() calls returned early (cached) above, so start them
+    // now that the per-agent resources have been constructed.
+    for(auto& ctx : context::get_active_contexts())
+    {
+        if(ctx->device_thread_trace) ctx->device_thread_trace->start_context();
```
ApoKalipse-V, line 647: Is this what the other services do for cached start?

Asking because it feels not very robust  for a start_context() return silently and trust it'll get recalled.
If there is precedent for this, I'm okay with it.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
```diff
@@ -74,6 +74,17 @@ struct cbdata_t
...
 common::Synchronized<std::optional<int64_t>> client;
 
+// Tracks whether the HSA runtime has been registered with rocprofiler. Thread
+// trace cannot touch HSA (queues, memory pools, packet buffers) before this is
+// true, so start requests issued earlier are cached in the active-context array
+// and replayed from initialize() once HSA becomes available.
+std::atomic<bool>&
+hsa_inited()
+{
+    static std::atomic<bool> inited{false};
```
ApoKalipse-V, line 82: is this what we have for other services that requires delayed start?

ApoKalipse-V (changes requested):

ApoKalipse-V (changes requested):
Global question: What happens if we call start_context() then stop_context() before hsa_init()? It is also protected? Stop_context() normally sends aql packets too.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
minseobshin11: Thanks for letting me know! I agree and will be removing them

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
minseobshin11: I believe device counting service has a similar approach

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
minseobshin11: Same here - device counting uses hsa_inited(). i mirrored it so they are consistent

minseobshin11:
Hi, if we call ```start_context()``` then ```stop_context()``` before ```hsa_init()```, it is protected - verified with a new test (```thread-trace-api-agent-start-stop-before-hsa-init-test```). ```start_context()``` is guarded with ```if(!hsa_inited().load()) return;```. And ```stop_context()``` returns at ```if(agents.empty()) return```

ApoKalipse-V (approved):
LGTM

projects/rocprofiler-sdk/tests/thread-trace/CMakeLists.txt
```diff
@@ -34,6 +34,24 @@ rocprofiler_add_integration_execute_test(
...
+    ENVIRONMENT "ATT_START_BEFORE_HSA_INIT=1")
+
+# Start+stop before hsa_init(): verifies the cached request is dropped and not replayed.
+rocprofiler_add_integration_execute_test(
+    thread-trace-api-agent-start-stop-before-hsa-init-test
+    TARGET thread-trace-test-app
+    TIMEOUT 10
+    LABELS "integration-tests"
+    PRELOAD "$<TARGET_FILE:thread-trace-agent-tool>"
+    ENVIRONMENT "ATT_START_BEFORE_HSA_INIT=1;ATT_STOP_BEFORE_HSA_INIT=1")
```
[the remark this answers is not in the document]
minseobshin11: Fixed!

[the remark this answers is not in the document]
minseobshin11: added an explicit assertion/observation in the tool for this test

venkat1361 (approved):

PR #8010, opened 2026-06-30
[rocprofiler-compute][bugfix] Fix --block filter crash and derive alias help at runtime

xuchen-amd:
## Motivation

The `--help` for profile lists block aliases users can pass to `-b/--block`. Some of those examples (SQ, TCC, l1i) no longer exist, so following the help text crashes the tool:
```
$ rocprof-compute profile -n test -b SQ -- ./vcopy
KeyError: "Unknown panel alias: 'SQ'"   # full traceback, exit
```
Two issues behind this:
1. Invalid filter values crash instead of showing a helpful error.
2. Help examples are hardcoded and silently go stale as configs change.

## Technical Details

1. Invalid `-b` values now fail gracefully with no traceback and a clear help message:

```
$ rocprof-compute profile -n test -b SQ -- ./vcopy
ERROR Invalid --block value 'SQ'.
      Run rocprof-compute --list-blocks <arch> to see all block ids/aliases.
```

2. Help examples only reference `sol`/System Speed-of-Light, which exists for all archs.

## JIRA ID

AIPROFCOMP-697

## Test Plan

- `profile -b SQ`: clean error with valid options, no traceback.
- `profile -h`: examples show only real aliases.
- `profile -b sol`: valid alias works, profiling runs.

## Test Result

- [x] `profile -b SQ`: clean error with valid options, no traceback.
- [x] `profile -h`: examples show only real aliases.
- [x] `profile -b sol`: valid alias works, profiling runs.

## Submission Checklist


projects/rocprofiler-compute/src/argparser.py
```diff
@@ -220,17 +234,19 @@ def omniarg_parser(
 
     ## Profile Command Line Options
     ## ----------------------------
+    # Build the --block example from real aliases so it never goes stale.
+    block_example = block_alias_examples(2, sep=" ")
     profile_parser = subparsers.add_parser(
         "profile",
         help="Profile the target application",
-        usage="""
+        usage=f"""
```
vedithal-amd: I don't think we should make the help message display depend on reading analysis config template. I mean, these are just examples and we don't make any guarantee that they will work on any gfx arch. This kind of over-engineering is a bad design IMO and not worth the extra code and maintenance. I mean, if the user passes an unsupported block, we should show them the block is not supported for the gfx arch and ask them to use --list-blocks.

projects/rocprofiler-compute/src/utils/utils_common.py
```diff
@@ -960,18 +971,29 @@ def load_yaml(filepath: Union[str, os.PathLike]) -> dict[str, Any]:
         return yaml.safe_load(f)
 
 
+def _load_panel_templates() -> list[tuple[str, list[dict[str, Any]]]]:
```
vedithal-amd: This helper can be removed and the caller in-lined given my comment below about help message

projects/rocprofiler-compute/src/utils/utils_common.py
```diff
@@ -879,6 +879,20 @@ def format_scientific_notation_if_needed(
...
 
 
+def console_error_invalid_block(
+    token: str, arch: Optional[str], alias_map: dict[str, str]
+) -> None:
+    """Report an invalid --block token and exit. Lists the arch's valid
+    aliases when available so alias typos are easy to correct."""
+    detail = ""
+    if arch and alias_map:
+        detail = f"\n\tValid aliases for {arch}: {', '.join(sorted(alias_map))}."
```
vedithal-amd: We are already telling user to use --list-blocks which shows all aliases, so this detail is overkill, please remove.

projects/rocprofiler-compute/src/utils/utils_common.py
```diff
@@ -879,6 +879,20 @@ def format_scientific_notation_if_needed(
...
+    console_error(
+        f"Invalid --block value {token!r}.{detail}"
```
vedithal-amd: i think the only thing remains here is the console_error - see my previous comment.
so this helper does not gain its existence, and it can be in-lined in the callers.
yes the error string will be duplicated, but you can hoist that up as a public global var on top of utils common

vedithal-amd (changes requested):

projects/rocprofiler-compute/src/argparser.py
xuchen-amd: I don't agree with the statement
>these are just examples and we don't make any guarantee that they will work on any gfx arch

We shouldn't show users stale/wrong information. We could either not show any examples and ask users to use `--list-blocks`, or we show what's correct for the use case. It is bad user experience for users to be given a help message that isn't actually helpful.

In this case, since the list of possible blocks is statically defined and a short list, this shouldn't be too much maintenance. But I think the better practice would be using `choices` arg.

vedithal-amd: I see your point, in this case can we just say in the help message of --block to use --list-blocks to check what blocks are supported and stop showing examples in the first place?

vedithal-amd: making `choices` work of --block is going to be tricky since there is no static set of allowed arguments as it depends on the gfx arch. I think for cases like this, we should have a sanitization function which processes the arguments for this option and returns a helpful error message to the end user, also, validation of arguments can also be folded into this function. I think we might already have such a function

xuchen-amd: fixed in: daa996211966c8c706ecafa4fed68b21121f1168

Note that validation of block aliases cannot be folded into this function since it requires gpu arch to be known.

projects/rocprofiler-compute/src/utils/utils_common.py
xuchen-amd: removed in: 6d130dcf4f822fa991a464c685d64b7b266d0d69

projects/rocprofiler-compute/src/utils/utils_common.py
xuchen-amd: updated: daa996211966c8c706ecafa4fed68b21121f1168

projects/rocprofiler-compute/src/utils/utils_common.py
xuchen-amd: updated: daa996211966c8c706ecafa4fed68b21121f1168

abchoudh-amd (approved):
LGTM

projects/rocprofiler-compute/src/rocprof_compute_soc/soc_base.py
```diff
@@ -301,9 +302,10 @@ def _append_analysis_yaml_for_filter_token(
             pass
         else:
             alias = block_id
-            panel_alias_dict = get_arch_alias_to_panel_id(self._mspec.gpu_arch)
+            arch = self._mspec.gpu_arch
```
vedithal-amd: remove unnecessary vairable

projects/rocprofiler-compute/tests/test_analyze_commands.py
```diff
@@ -926,31 +926,33 @@ def test_missing_file_handling(binary_handler_analyze_rocprof_compute):
 @pytest.mark.misc
 def test_filter_combinations_coverage(binary_handler_analyze_rocprof_compute, capsys):
     """Test basic filters that should work"""
-    for dir in ["tests/workloads/vcopy/MI100", "tests/workloads/vcopy/MI200"]:
-        if os.path.exists(dir):
-            workload_dir = common.setup_workload_dir(dir)
+    candidates = ["tests/workloads/vcopy/MI100", "tests/workloads/vcopy/MI200"]
+    dir = next((d for d in candidates if os.path.exists(d)), None)
+    if dir is None:
+        pytest.skip(f"no vcopy workload available: {candidates}")
```
vedithal-amd: i dont think this will every arise that vcopy workload is not available

projects/rocprofiler-compute/tests/test_parser.py
```diff
@@ -108,6 +110,54 @@ def test_unknown_alias_raises(self, monkeypatch):
             convert_filter_blocks_to_panel_ids(["bogus"], arch="gfx942")
 
 
+# =============================================================================
```
vedithal-amd: I dont think soc_base py unit tests should go in test_parser.py

projects/rocprofiler-compute/src/rocprof_compute_soc/soc_base.py
vedithal-amd: Done in bfd6e8c70b — inlined into get_arch_alias_to_panel_id(self._mspec.gpu_arch).

projects/rocprofiler-compute/tests/test_analyze_commands.py
vedithal-amd: Reverted in bfd6e8c70b — restored the original for/break loop; only the assertion string change ('SQ' -> quoted, needed by the !r formatting) remains.

projects/rocprofiler-compute/tests/test_parser.py
vedithal-amd: Moved in bfd6e8c70b — relocated to test_soc_base.py as flat test_* functions (matching that file's style) and dropped the redundant third test that hit the same console_error branch.

vedithal-amd:
@xuchen-amd pushed bfd6e8c70b addressing the three review comments on your behalf:

- **soc_base.py**: inlined the single-use `arch` local.
- **test_analyze_commands.py**: restored the original `for`/`break` loop; only the assertion string change remains (`'SQ'` quoted, required by the `!r` error formatting).
- **tests**: moved the profile-mode alias tests from `test_parser.py` to `test_soc_base.py` as flat `test_*` functions matching that file's style, and dropped the redundant third test (it exercised the same `console_error` branch as the unknown-alias test).

Unit tests pass and ruff is clean.

vedithal-amd (approved):
LGTM. Addressed the three outstanding comments directly in bfd6e8c70b (see summary above).

PR #8011, opened 2026-06-30
[rccl-tests] JIRA ID ROCM-25479 Add support for static linking against librccl.a

wenkaidu:
## Motivation

Resolve rccl-tests issues with static linking for reproduce performance issue in statically linked executables

## Technical Details

Add an opt-in STATIC_RCCL=1 build mode that statically links rccl-tests against a librccl.a built with -fgpu-rdc. This compiles the test objects with -fgpu-rdc and performs the HIP device link (--hip-link) at final link time, and pulls in -lroctx64.

Also give the test-local IsArchMatch internal linkage to avoid a duplicate symbol clash with rccl's own IsArchMatch when statically linked.

## Test Plan

Build static rccl library and rccl-tests

## Test Result

Pass

## Submission Checklist

### DIFF
--- a/projects/rccl-tests/src/Makefile
+++ b/projects/rccl-tests/src/Makefile
@@ -142,6 +142,14 @@ ifeq ($(ENABLE_DEVICE_API), 1)
 HIPCUFLAGS += -DENABLE_DEVICE_API
 endif
 
+# Statically link against a librccl.a that was built with -fgpu-rdc.
+# This requires compiling the test objects with -fgpu-rdc as well and
+# performing the HIP device link step (--hip-link) at final link time.
+ifeq ($(STATIC_RCCL), 1)
+HIPCUFLAGS += -fgpu-rdc
+HIPLDFLAGS += -fgpu-rdc --hip-link $(GPU_TARGETS_FLAGS) -lroctx64
+endif
+
 LIBRARIES += rccl dl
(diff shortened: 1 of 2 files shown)


alex-breslow-amd:
I'm concerned that this will only work without the parallel device linker.  Any thoughts on making this more general?  I thought @nileshnegi had already banged something out, which accomplished this.  We just never got around to merging it.

projects/rccl-tests/src/Makefile
```diff
@@ -142,6 +142,14 @@ ifeq ($(ENABLE_DEVICE_API), 1)
 HIPCUFLAGS += -DENABLE_DEVICE_API
 endif
 
+# Statically link against a librccl.a that was built with -fgpu-rdc.
+# This requires compiling the test objects with -fgpu-rdc as well and
+# performing the HIP device link step (--hip-link) at final link time.
+ifeq ($(STATIC_RCCL), 1)
+HIPCUFLAGS += -fgpu-rdc
+HIPLDFLAGS += -fgpu-rdc --hip-link $(GPU_TARGETS_FLAGS) -lroctx64
```
alex-breslow-amd, line 150: This will break the device linker I think unless this is a non-device linker feature.

wenkaidu: yes, you are right. When I tested this, I wasn't using device linker libs. Need to be tested/modified for device linker

PR #7818, opened 2026-06-25
Debugging and testing coverage expansion

ApoKalipse-V:
## Motivation

This PR adds
* Logging of SQTT tokens with the SQTT_LOGGING environment variable, for debugging.
* Improved code coverage with expanded unit tests
* Minor fixes for issues caught with said unit tests
* Removes a unused AVX512 no-vbmi path for quick scan.

## JIRA ID

AIPROFSDK-934

## Test Plan

Added extensive unit tests. More than that, it's coverage bloat.

## Test Result

Screenshot 2026-06-26 171216

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #8014, opened 2026-06-30
fix: Compilation error on windows

ApoKalipse-V:
## Motivation

The quick_scan component does not compile on windows due to a single variable being declared inside the #if condition.

JIRA ID : AIPROFSDK-830

### DIFF
--- a/projects/rocprof-trace-decoder/source/quick_scan_export.cpp
+++ b/projects/rocprof-trace-decoder/source/quick_scan_export.cpp
@@ -604,12 +604,12 @@ ROCPROF_TRACE_DECODER_API rocprofiler_thread_trace_decoder_status_t rocprof_trac
     if (!quick_scan::avx512_available()) return ROCPROFILER_THREAD_TRACE_DECODER_STATUS_ERROR_NOT_IMPLEMENTED;
 
     size_t ntokens = 0;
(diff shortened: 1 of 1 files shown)


EvanPhoukong (approved):

ApoKalipse-V:
Note: PR bot is incorrect, the unit tests do cover this.

arvindcheru:
**Note on Reason for Override:**
Moving header_offset above #if ROCPROF_TRACE_DECODER_QUICK_SCAN_HAS_SIMD is the correct minimal fix — MSVC sets HAS_SIMD=0, so the variable must be in scope for process_events_gfx9 after #endif. No runtime change on Linux/GCC builds.

Existing coverage is sufficient: BuildStandaloneApiTest.DoesNotCopyGfx9ChunkHeaderIntoStandalonePayload and ReplaysGfx9CodeObjectAndDispatchContextFromCutTrace exercise the gfx9 header_offset path, not a functional gap.

Lesson Learnt and suggestion by AI for the observation: (from #7818: to Linux CI: PASS (GCC, SIMD=1) and Windows MSVC: FAIL (SIMD=0)) 
add Windows MSVC compile coverage (or a Linux SIMD=0 compile check) so similar #if/scoping issues are caught before merge.

PR #8022, opened 2026-06-30
fix(rocprofiler-sdk): PMC FETCH_SIZE returns 0 on gfx12xx

JasonTTang:
JIRA ID: ROCM-15816

root cause: FETCH_SIZE formula is missing the GL2C_EA_RDREQ_256B component for gfx12xx.


ApoKalipse-V (approved):

bing-ma (commented):
This updates FETCH_SIZE for gfx1250, but FetchSize still doesn’t list gfx1250, so the alias metric looks left behind.

bing-ma (approved):
Looks good to me.

PR #8023, opened 2026-06-30
feat(rocprofiler-sdk): Introduce thread trace rocpd support.

srihari-uttanur:
## Motivation

This PR adds thread trace support for rocpd output format. When --att is combined with --output-format rocpd, the profiler now persists raw ATT chunk metadata, decoder-generated text artifacts, and DWARF-derived source snapshots directly into the rocpd SQLite database.

## Technical Details

**New schema tables**

rocpd_gpu_thread_trace_chunk
One row per raw ATT data chunk. Each chunk corresponds to one shader engine's raw SQTT binary output for one dispatch.

Fields:
agent_id, shader_engine_id, chunk_id — uniquely identify the chunk within the session
time_begin, time_end — wall-clock timestamps bounding the collection window
trace_mode — whether collection was dispatch-triggered or agent-wide
parameters — semicolon-separated key=value string capturing all ATT session parameters (shader engine mask, perfcounters, serialize_all, inline storage mode) so a DB consumer can reproduce the collection context without the config file
att_blob_id — colon-joined identity string dispatch:agent:se:chunk for external cross-referencing
binary_chunk_path — relative filesystem path to the .att file (default mode)
binary_chunk — raw .att bytes stored inline (optional, enabled via ROCPROF_ATT_PARAM_STORE_CHUNK_INLINE)
binary_chunk_size — size in bytes of the raw binary payload

the rocpd database has no record that thread trace was collected, which kernels produced trace data, where those files are, or their timing relationship to other GPU events. This table is the primary anchor for all thread trace information in the DB. It also allows portable DB packages where the .att binary is embedded rather than left on disk.

rocpd_info_text_blob:
One row per unique text artifact file, deduped by output-relative path within the session.

Fields:
path (relative to output root), 
md5 (hex digest for integrity verification), 
size (byte count), 
text (full UTF-8 content of the file).

The ATT decoder emits human-readable artifacts — an instruction-level decode map (code.json), a source snapshot manifest (snapshots.json), and copies of DWARF-referenced source files (source_*). If these are left only on disk, a consumer holding only the .db file cannot access them. Persisting them inline makes the database self-contained: instruction maps, source attribution, and snapshot metadata are all available without the original filesystem layout or source tree.

rocpd_gpu_thread_trace_text_link
Junction table linking each chunk row to one or more text artifact rows.

Fields:
chunk_id (FK to chunk table), text_blob_id (FK to text blob table), kind (one of code_json, snapshot_manifest, snapshot_source), ordinal (position within kind, stable across runs because source paths are sorted before assignment).

A single chunk maps to multiple artifacts of different types. Without this table a consumer cannot determine which code.json belongs to which chunk, or enumerate all source snapshots for a given dispatch in order. The kind column distinguishes artifact roles and the ordinal provides a stable ordering for multi-file kinds, enabling queries like "give me all source files for chunk X, in order" with a single join.

## JIRA ID

AIPROFSDK-152

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-sdk/source/lib/output/generateRocpd.cpp
+++ b/projects/rocprofiler-sdk/source/lib/output/generateRocpd.cpp
@@ -24,6 +24,7 @@
 #include "lib/common/uuid_v7.hpp"
 #include "metadata.hpp"
 #include "output_stream.hpp"
+#include "lib/rocprofiler-sdk-tool/config.hpp"
 #include "statistics.hpp"
 #include "stream_info.hpp"
 #include "timestamps.hpp"
@@ -65,10 +66,13 @@
 #include <cctype>
 #include <chrono>
 #include <filesystem>
+#include <fstream>
 #include <initializer_list>
 #include <limits>
+#include <map>
 #include <set>
 #include <type_traits>
+#include <tuple>
 #include <unordered_map>
 #include <unordered_set>
 #include <utility>
@@ -109,7 +113,7 @@ using function_args_t = std::vector<argument_info>;
 struct sql_insert_value
 {
     std::string_view                                                                     name  = {};
-    std::variant<std::monostate, int64_t, uint64_t, double, std::string, std::nullptr_t> value = {};
+    std::variant<std::monostate, int64_t, uint64_t, double, std::string, std::vector<uint8_t>, std::nullptr_t> value = {};
 };
 
 struct pending_insert_batch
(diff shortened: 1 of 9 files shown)


projects/clr/rocclr/device/rocm/rocblit.cpp
ApoKalipse-V, line 2006: Why do we have changes in rocr blit here?

projects/clr/CHANGELOG.md
ApoKalipse-V, line 9: Why the CLR change?

projects/rocprofiler-sdk/source/lib/output/generateRocpd.cpp
```diff
@@ -2016,6 +2101,175 @@ write_rocpd(
...
+                                        tool::get_config().att_param_shader_engine_mask));
+        params.emplace_back(
+            fmt::format("store_chunk_inline={}", tool::get_config().att_param_store_chunk_inline ? 1 : 0));
+
+        auto output_root = fs::path{tool::format_path(cfg.output_path)};
+        auto params_text = fmt::format("{}", fmt::join(params, ";"));
+        auto inline_blob = tool::get_config().att_param_store_chunk_inline;
+
+        // index chunk row id by chunk identity to support text linkage
+        auto chunk_id_map = std::map<std::tuple<uint64_t, int64_t, uint32_t, uint64_t>, uint64_t>{};
```
ApoKalipse-V, line 2142: I think we need a struct definition here instead of tuple

projects/rocprofiler-sdk/source/lib/output/generateRocpd.cpp
```diff
@@ -2016,6 +2101,175 @@ write_rocpd(
...
+        }
+
+        // ATT decoder emits text artifacts under ui_output_agent_*_dispatch_* directories.
+        // Persist text content and link to thread-trace chunks so DB consumers can resolve snapshots.
+        auto text_blob_id_counter = uint64_t{1};
+        auto text_blob_by_path    = std::unordered_map<std::string, uint64_t>{};
+
+        for(const auto& chunk : chunks)
+        {
+            auto ui_dir = output_root / fmt::format("ui_output_agent_{}_dispatch_{}",
```
ApoKalipse-V, line 2208: I dont think this is the way.
We dont want to link to json contents or ui_ dirs, we want to remove them entirely (future PR).

What we need is sufficient metadata in the database for the RCV to reconstruct the contents of the ui_ dir.

The .att + .out contains most of it, but the rocprofv3 input used and the source links/snapshots need to be supplied by rocprofv3

ApoKalipse-V: att-tool/code.cpp wil dump the ISA and DWARF information into json files.
The RCV is capable of disassembly so you dont need to dump the entire the dissassembly, only the DWARF range addresses.
However, you CAN if you think it's easier.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/tool.cpp
```diff
@@ -3405,6 +3438,40 @@ generate_output(cleanup_mode _cleanup_mode)
                              rocjpeg_output.get_generator());
     }
 
+    if(tool::get_config().advanced_thread_trace)
```
ApoKalipse-V, line 3540: NIT: Do you really need to move this up?

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/config.hpp
```diff
@@ -136,6 +136,8 @@ struct config : output_config
     bool   list_metrics_output_file      = get_env("ROCPROF_OUTPUT_LIST_METRICS_FILE", false);
     bool   advanced_thread_trace         = get_env("ROCPROF_ADVANCED_THREAD_TRACE", false);
     bool   att_serialize_all             = get_env("ROCPROF_ATT_PARAM_SERIALIZE_ALL", false);
+    bool   att_param_store_chunk_inline  =
+        get_env("ROCPROF_ATT_PARAM_STORE_CHUNK_INLINE", false);
```
ApoKalipse-V, line 141: Needs to be exposed in rocprofv3?

projects/rocprofiler-sdk/source/lib/output/generateRocpd.cpp
```diff
@@ -2016,6 +2101,175 @@ write_rocpd(
...
         };
 
+    auto insert_thread_trace_chunks = [&db, &tool_metadata, &cfg, node_id, this_pid]() {
+        auto _sqlgenperf_rocpd = get_simple_timer("rocpd_thread_trace_chunks");
+
+        const auto chunks = tool_metadata.get_att_trace_chunks();
+        if(chunks.empty()) return;
+
+        auto params = std::vector<std::string>{};
+        params.emplace_back("trace_mode=mixed");
```
ApoKalipse-V, line 2116: whats mixed?

ApoKalipse-V (changes requested):

projects/clr/rocclr/device/rocm/rocblit.cpp
srihari-uttanur: It was not rebased properly, I have fixed it now.

projects/clr/CHANGELOG.md
srihari-uttanur: Same rebase issue, I have fixed it now.

PR #8024, opened 2026-06-30
feat(amdsmi) : env var skipping gpu_metrics reads on a idle GPU

marifamd:
## Motivation

On Navi, reading `gpu_metrics` while the GPU is runtime-suspended forces a wake (D3->D0). The first firmware sample after that wake reports gfx activity and clock latched high for ~9-15s, so a tool polling an idle GPU sees false spikes and keeps the GPU awake. The fix belongs on the tool side: don't trigger a wakeup just to report activity.

## Technical Details

Add an opt-in env var `AMDSMI_SKIP_GPU_METRICS_ON_IDLE` (disabled by default). When enabled and the GPU is runtime-suspended, `amdsmi_get_gpu_metrics_info` returns `AMDSMI_STATUS_BUSY` instead of reading `gpu_metrics`.

- Gated at the single `gpu_metrics` read, so `amdsmi_get_gpu_activity` and `amdsmi_get_clock_info` inherit it.
- Suspend is detected via the non-waking `power/runtime_status` sysfs node. Instinct has no runtime PM, so it never reads `suspended` and the gate never fires there (no-op, metrics unaffected).
- No signature change (no ABI / wrapper impact). The skip is logged under `RSMI_LOGGING`.
- CLI: with the var set, `amd-smi metric`/`monitor` show `N/A` for metrics-derived fields on a suspended GPU. No CLI code change.

```text
  amdsmi_get_gpu_metrics_info()        also: amdsmi_get_gpu_activity()
            │                                amdsmi_get_clock_info()
            ▼
  AMDSMI_SKIP_GPU_METRICS_ON_IDLE enabled?
            │
            ├── no ───────────────► read gpu_metrics        (default, unchanged)
            │
            └── yes
                 │
                 ▼
        runtime_status == "suspended"?   (non-waking read; never true on Instinct)
                 │
                 ├── no ──────────► read gpu_metrics
                 │
                 └── yes ─────────► return AMDSMI_STATUS_BUSY
                                    (skip read; GPU stays asleep, no false spike)
```

## JIRA ID

JIRA ID : ROCM-26020

## Test Plan

- C++ `TestIdleMetricsRead` (helper parsing + per-device gating/no-op) and Python `test_skip_gpu_metrics_on_idle`, plus the existing `TestGpuMetricsRead`/`TestGPUBusyRead` regression.
- Hardware: Navi31 (runtime-suspended), Navi32, MI210, MI300X, MI300A.
- `pre-commit run` on all changed files.

## Test Result

- Suspended Navi31: returns `AMDSMI_STATUS_BUSY`, device stays suspended (no wake).
- Instinct (MI210/MI300X/MI300A) and awake GPUs: no-op, identical result flag on or off.
- Default (flag unset): unchanged; tests and pre-commit pass.

## Submission Checklist

### DIFF
--- a/projects/amdsmi/CHANGELOG.md
+++ b/projects/amdsmi/CHANGELOG.md
@@ -8,6 +8,12 @@ Full documentation for amd_smi_lib is available at [https://rocm.docs.amd.com/pr
 
 ### Added
 
+- **Added opt-in idle gating for GPU metrics reads (`AMDSMI_SKIP_GPU_METRICS_ON_IDLE`)**.  
+  - On GPUs with runtime power management / GFXOFF (for example Navi/RDNA), reading `gpu_metrics` while the GPU is runtime-suspended forces a wake whose first firmware sample latches GFX activity and clock high for several seconds, producing false spikes on an otherwise idle GPU.
+  - Set `AMDSMI_SKIP_GPU_METRICS_ON_IDLE` (to `1`/`true`/`on`/`yes`) so `amdsmi_get_gpu_metrics_info` (and APIs that read metrics internally, such as `amdsmi_get_gpu_activity` and `amdsmi_get_clock_info`) return `AMDSMI_STATUS_BUSY` on a runtime-suspended device instead of reading `gpu_metrics` and waking it. Disabled by default, so existing behavior is unchanged.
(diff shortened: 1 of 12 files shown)


marifamd:
Resolved within GPUagent

PR #8025, opened 2026-06-30
rocjitsu: preserve SCC across scalar conversions

bjacob:
Scalar conversion SOP1 instructions write their destination register but do not update SCC. The old shared implementation set SCC from the converted value, which can corrupt compare/select sequences when a scalar conversion is scheduled between the compare and the consuming s_cselect.

Drop the SCC side effect from the shared scalar conversion helpers and add a gfx1250 regression that runs every scalar conversion form with SCC initialized both ways.

This is PR 1 out of 2 needed to have rocjitsu/gfx1250 pass https://github.com/bjacob/hip-moi tests.


amd-eochoalo:
Should these changes be driven by changes in the code generator?

kuhar (approved):
Fixed agent comments, rebased, and made the changes generator driver like @amd-eochoalo requested

PR #8026, opened 2026-06-30
rocjitsu: block gfx1250 split barrier waits

bjacob:
Treat gfx1250 s_barrier_wait as the workgroup rendezvous point in the shared AMDGPU execution helper. Split barrier signal instructions remain non-blocking; the wait instruction now enters the existing WfState::BARRIER path so the scheduler holds the wavefront until the workgroup reaches the rendezvous.

Add a gfx1250 regression that decodes s_barrier_wait and checks that executing it moves a running wavefront into WfState::BARRIER.

This is PR 2 out of 2 needed to have rocjitsu/gfx1250 pass https://github.com/bjacob/hip-moi tests.


kuhar (approved):
Thanks, I reviewed this locally and addressed the issues in https://github.com/ROCm/rocm-systems/pull/8026/commits/669b42abced8447d736b95c6610aa58f6291567f

The main issue was bypassing the generator.

PR #7383, opened 2026-06-17
rocjitsu: Rocjitsu DBT Guest Mode 

Groverkss:
A rocjitsu tool path where an unmodified ROCm process can see a
guest GPU, but translate kernels and execute on a real host GPU.

The design has three main components with their own responsibilities:

1. **GuestKfd Interposer**: Shows applications a guest GPU device that looks
   like a real KFD/DRM device.
2. **HSA Hooks**: Intercepts HSA Runtime calls for the guest GPU and forwards
   them to the host GPU, translating kernels as needed.
3. **DBT**: Translates guest GPU kernels to host GPU kernels.

## High Level Flow

User:
```bash
rocjitsu --config configs/guest_gfx950_on_gfx942.json -- ./app
```

Underneath:
```
  LD_PRELOAD=librocjitsu_kmd.so
      open/read sysfs topology -> host topology plus synthetic gfx950 node
      open/ioctl /dev/kfd      -> real KFD for host nodes, GuestKfd for guest discovery/startup

  HSA_TOOLS_LIB=librocjitsu_hooks.so
      ROCR internally sees     -> selected host agent plus synthetic guest agent
      hsa_iterate_agents       -> public host slot is replaced by the guest agent
      hsa_agent_get_info       -> guest still reports gfx950 identity
      hsa_queue_create         -> guest agent is replaced with host agent
      hsa_executable_load...   -> guest ELF translated to host ELF, then loaded on host
      hsa_executable_get...    -> guest symbol lookup is replaced with host symbol lookup
      AMD extension calls      -> guest agents are replaced with host agents where needed
```

More design docs can be found in docs/rocjitsu_dbt_guest.md

Original Design Document: https://gist.github.com/Groverkss/0f13e40c2d1f0d0a2ec31f7850317269 (may have drifted a bit after review, but the principles are same)

PR #8035, opened 2026-06-30
fix(rocjitsu): tolerate early mmap during interposer load

newling:
ISSUE ID: #8037

This fix is suggested by AI. The goal is to address an interposer startup ordering issue.

## Summary

Fix an interposer startup issue where another preloaded library can call
`mmap()` or `munmap()` before rocjitsu has resolved its libc passthrough
symbols.

The fix adds a small bootstrap path for early `mmap()` / `munmap()` calls. Before
`InterposerContext::real` is ready, these wrappers bypass rocjitsu state and call
`SYS_mmap` / `SYS_munmap` directly. Once initialization is complete, both
wrappers continue through the normal interposer logic.

## Rationale

A loader-ordering fix is not general enough here. An explicit ELF dependency can
force constructor order for libraries rocjitsu controls, but the early caller may
be an arbitrary third-party preload or application dependency. Rocjitsu cannot
add dependency edges from every possible DSO back to itself, and constructor
priorities only order constructors within one shared object.

The direct-syscall fallback is intentionally narrow: it avoids rocjitsu state,
allocation, logging, and `dlsym()` during dynamic-loader startup.

## Testing

Built focused targets:

```bash
cmake --build <rocjitsu-build> --parallel 8 --target \
  early_mmap_preload early_mmap_driver rocjitsu_shared
```

Ran targeted test:

```bash
ctest --test-dir <rocjitsu-build> \
  -R 'InterposerTest\.EarlyMmapBeforeRocjitsuConstructor' \
  --output-on-failure
```

Result:

```text
1/1 tests passed
```

Verified the regression test exercises the intended constructor order with
`LD_DEBUG=libs`: the helper preload initializes before `librocjitsu.so`, and the
program exits successfully.

Ran pre-commit on the changed files:

```text
Passed
```

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
@@ -142,6 +142,7 @@ std::optional<std::string> child_config_path() {
 }
 
 void *raw_mmap_syscall(void *addr, size_t length, int prot, int flags, int fd, off_t offset) {
+  // syscall(2) is the libc wrapper: on kernel errors it returns -1 and sets errno.
   long rc = syscall(SYS_mmap, addr, length, prot, flags, fd, offset);
   if (rc == -1)
     return MAP_FAILED;
@@ -149,7 +150,9 @@ void *raw_mmap_syscall(void *addr, size_t length, int prot, int flags, int fd, o
 }
 
 int raw_munmap_syscall(void *addr, size_t length) {
-  return static_cast<int>(syscall(SYS_munmap, addr, length));
+  long rc = syscall(SYS_munmap, addr, length);
+  assert(rc == 0 || rc == -1);
+  return static_cast<int>(rc);
 }
 
 void rj_sigsegv_handler(int, siginfo_t *, void *) {
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
@@ -148,6 +148,10 @@ void *raw_mmap_syscall(void *addr, size_t length, int prot, int flags, int fd, o
   return reinterpret_cast<void *>(static_cast<uintptr_t>(rc));
 }
 
+int raw_munmap_syscall(void *addr, size_t length) {
+  return static_cast<int>(syscall(SYS_munmap, addr, length));
+}
+
 void rj_sigsegv_handler(int, siginfo_t *, void *) {
   signal(SIGSEGV, SIG_DFL);
   raise(SIGSEGV);
@@ -1247,6 +1251,9 @@ RJ_INTERPOSER_EXPORT int madvise(void *addr, size_t length, int advice) {
 }
 
 RJ_INTERPOSER_EXPORT int munmap(void *addr, size_t length) {
+  if (!InterposerContext::real().ready() || !InterposerContext::real().munmap)
+    return raw_munmap_syscall(addr, length);
+
   assert(InterposerContext::real().ready());
   if (auto *remote = InterposerContext::ctx.remote_lookup(InterposerContext::ctx.remote_kfd_fd())) {
     int ret = remote->munmap(addr, length);
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
@@ -141,6 +141,13 @@ std::optional<std::string> child_config_path() {
   return std::string(cfg_buf);
 }
 
+void *raw_mmap_syscall(void *addr, size_t length, int prot, int flags, int fd, off_t offset) {
+  long rc = syscall(SYS_mmap, addr, length, prot, flags, fd, offset);
+  if (rc == -1)
+    return MAP_FAILED;
+  return reinterpret_cast<void *>(static_cast<uintptr_t>(rc));
+}
+
 void rj_sigsegv_handler(int, siginfo_t *, void *) {
   signal(SIGSEGV, SIG_DFL);
   raise(SIGSEGV);
@@ -1176,6 +1183,9 @@ RJ_INTERPOSER_EXPORT int fcntl64(int fd, int cmd, ...) {
 
 RJ_INTERPOSER_EXPORT void *mmap(void *addr, size_t length, int prot, int flags, int fd,
                                 off_t offset) {
+  if (!InterposerContext::real().ready() || !InterposerContext::real().mmap)
+    return raw_mmap_syscall(addr, length, prot, flags, fd, offset);
+
   assert(InterposerContext::real().ready());
   if (auto *remote = InterposerContext::ctx.remote_lookup(fd))
     return remote->mmap(addr, length, prot, flags, offset);
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
@@ -142,7 +142,9 @@ std::optional<std::string> child_config_path() {
 }
 
 void *raw_mmap_syscall(void *addr, size_t length, int prot, int flags, int fd, off_t offset) {
-  // syscall(2) is the libc wrapper: on kernel errors it returns -1 and sets errno.
+  // syscall(2) is the libc wrapper, not a raw inline syscall instruction: on
+  // kernel errors it returns -1 and sets errno. For mmap(2), success returns
+  // the mapped address; for munmap(2), success returns exactly 0.
   long rc = syscall(SYS_mmap, addr, length, prot, flags, fd, offset);
   if (rc == -1)
     return MAP_FAILED;
--- a/emulation/rocjitsu/tests/CMakeLists.txt
+++ b/emulation/rocjitsu/tests/CMakeLists.txt
@@ -687,7 +687,7 @@ set_tests_properties(
     "InterposerTest.EarlyMmapBeforeRocjitsuConstructor"
     PROPERTIES
         ENVIRONMENT
-            "LD_PRELOAD=$<TARGET_FILE:early_mmap_preload>:$<TARGET_FILE:rocjitsu_shared>"
+            "LD_PRELOAD=$<TARGET_FILE:rocjitsu_shared>:$<TARGET_FILE:early_mmap_preload>"
 )
 
 # --- rocminfo test (requires rocminfo binary + HSA runtime + CLI launcher) ---
--- a/emulation/rocjitsu/tests/CMakeLists.txt
+++ b/emulation/rocjitsu/tests/CMakeLists.txt
@@ -670,6 +670,23 @@ target_link_libraries(
 rj_configure_target(hsa_hooks_unit_test TEST)
 gtest_discover_tests(hsa_hooks_unit_test)
 
+# --- Interposer loader initialization regression tests ---
+
+add_library(early_mmap_preload SHARED early_mmap_preload_test.cpp)
+rj_configure_target(early_mmap_preload TEST)
+add_dependencies(early_mmap_preload rocjitsu_kmd_shim)
+
+add_test(
+    NAME "InterposerTest.EarlyMmapBeforeRocjitsuConstructor"
+    COMMAND /bin/true
+)
+set_tests_properties(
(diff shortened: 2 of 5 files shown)


newling:
TODO(newling) see how this is related to the comment https://github.com/ROCm/rocm-systems/pull/7383#discussion_r3540074744

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
```diff
@@ -88,6 +88,17 @@ static int connect_to_daemon() {
 
 namespace {
 
+void *raw_mmap_syscall(void *addr, size_t length, int prot, int flags, int fd, off_t offset) {
+  long rc = syscall(SYS_mmap, addr, length, prot, flags, fd, offset);
+  if (rc == -1)
```
atgutier, line 149: -1 is just one particular errno so this will miss other returned error codes.

Also, because we're using a raw syscall (and not using libc) we need to set the errno otherwise callers that expect libc to set the errno will get stale errnos.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
```diff
@@ -88,6 +88,17 @@ static int connect_to_daemon() {
...
+    return MAP_FAILED;
+  return reinterpret_cast<void *>(static_cast<uintptr_t>(rc));
+}
+
+int raw_munmap_syscall(void *addr, size_t length) {
+  return static_cast<int>(syscall(SYS_munmap, addr, length));
```
atgutier: I think this also leaks a bad errno return value and also doesn't set the errno variable.

atgutier (commented):
This is a good find and I think the reason mmap/munmap are special is because dlsym can use them itself, but I think we could get rid of our existing asserts and do a lazy init for the other calls and rely on our libc passthrough without readiness checks. This could also avoid some extra complexity here.

An agent suggests a more comprehensive approach and provided this fix. I haven't vetted it myself, but I agree with it that a larger rework is required.

Now I have the actual LibcPassthrough (interposer.cpp:117-172) and the real readiness model. Here's
  what that consolidated fix looks like, grounded in the code that's actually there.

  The key realization

  Once accessors self-resolve, the entire ready() / assert(ready()) machinery disappears. There is no
  longer a "not ready" state for any wrapper to guard against — so you delete the per-wrapper guards
  (the mmap/munmap ones this PR adds and the bare asserts on open/ioctl/dup/fork/…) rather than adding
   more of them.

  But there's one real nuance the reviewer's "one stroke" phrasing glossed over, and it's the whole
  reason mmap/munmap are special: dlsym can allocate (it may call mmap/calloc). So the fallback splits
   into two categories, and that split is fundamental, not accidental:

  - Most wrappers (open, close, ioctl, dup*, fcntl, fork, fopen, stat, …): if not yet resolved, lazily
   dlsym(RTLD_NEXT, …) and cache. Safe because dlsym is fine to call this way (the fs wrappers already
   do exactly this).
  - mmap / munmap only: must never dlsym (that would re-enter allocation during loader startup). Use
  the cached pointer if present, else a raw syscall.

  The consolidation is: route both categories through one uniform accessor shape, with atomic
  publication, so the three divergent styles collapse into one.

  1. Errno-correct raw helpers (fixes the BLOCKER + munmap MAJOR)

```
  namespace {
  // Raw syscalls translate the kernel's [-4095,-1] error return into the
  // libc -1/errno contract. Used only as the mmap/munmap bootstrap fallback,
  // because dlsym() itself may allocate and cannot run before resolution.
  [[nodiscard]] void *raw_mmap(void *addr, size_t len, int prot, int flags, int fd, off_t off) {
    long rc = syscall(SYS_mmap, addr, len, prot, flags, fd, off);
    if (rc < 0 && rc >= -4095) {          // kernel error range
      errno = static_cast<int>(-rc);
      return MAP_FAILED;
    }
    return reinterpret_cast<void *>(static_cast<uintptr_t>(rc));
  }

  [[nodiscard]] int raw_munmap(void *addr, size_t len) {
    long rc = syscall(SYS_munmap, addr, len);
    if (rc < 0 && rc >= -4095) {
      errno = static_cast<int>(-rc);
      return -1;
    }
    return static_cast<int>(rc);
  }
  } // namespace
```

  2. Self-resolving accessors on LibcPassthrough

  Make the members std::atomic (fixes the ordering MINOR: release on publish, acquire on read) and add
   a lazy resolver. Members become private; every call site goes through an accessor.

```
  class LibcPassthrough {
    template <typename Fn>
    static Fn resolve(std::atomic<Fn> &slot, const char *name) {
      if (Fn fn = slot.load(std::memory_order_acquire)) return fn;   // fast path
      Fn fn = util::lookup_symbol<Fn>(RTLD_NEXT, name);              // dlsym — may allocate
      slot.store(fn, std::memory_order_release);
      return fn;
    }

    std::atomic<int (*)(int, const char *, int, ...)> openat_{nullptr};
    std::atomic<int (*)(int)> close_{nullptr};
    std::atomic<int (*)(int, unsigned long, ...)> ioctl_{nullptr};
    std::atomic<void *(*)(void *, size_t, int, int, int, off_t)> mmap_{nullptr};
    std::atomic<int (*)(void *, size_t)> munmap_{nullptr};
    // ... dup_, dup2_, dup3_, fcntl_, fopen_, fork_, stat_, ...

  public:
    // dlsym-safe wrappers: guaranteed non-null (assert inside resolve()).
    int openat(int d, const char *p, int f, mode_t m) { return resolve(openat_, "openat")(d, p, f, m);
   }
    int close(int fd)                                  { return resolve(close_,  "close")(fd); }
    int ioctl(int fd, unsigned long r, void *a)        { return resolve(ioctl_,  "ioctl")(fd, r, a); }
    // ... etc for dup/fcntl/fork/stat/...

    // mmap/munmap: NEVER dlsym (would re-enter allocation). Cached ptr or raw syscall.
    void *mmap(void *a, size_t n, int p, int f, int fd, off_t o) {
      if (auto fn = mmap_.load(std::memory_order_acquire)) return fn(a, n, p, f, fd, o);
      return raw_mmap(a, n, p, f, fd, o);
    }
    int munmap(void *a, size_t n) {
      if (auto fn = munmap_.load(std::memory_order_acquire)) return fn(a, n);
      return raw_munmap(a, n);
    }

    // Optional eager warm-up at constructor time (pure optimization now, not a
    // correctness requirement): resolve() each dlsym-safe slot + mmap_/munmap_.
    void warm_up() { resolve(openat_, "openat"); /* ... */ ; resolve(mmap_, "mmap"); resolve(munmap_,
  "munmap"); }
  };
```

 Note mmap_/munmap_ are still populated (by warm_up() or lazily by a dlsym-safe accessor that happens
   to run after startup), so the steady-state path uses the real libc pointer — the raw syscall is
  strictly the pre-resolution bootstrap.

  3. What the call sites become — the actual de-duplication

  Every wrapper loses its guard and its assert. Before/after for three representative cases:

```
  // mmap — BEFORE (this PR):
  if (!InterposerContext::real.ready() || !InterposerContext::real.mmap)
    return raw_mmap_syscall(addr, length, prot, flags, fd, offset);
  assert(InterposerContext::real.ready());
  ...
  return InterposerContext::real.mmap(addr, length, prot, flags, fd, offset);
  // AFTER: guard + assert gone; the accessor handles bootstrap internally
  return InterposerContext::real.mmap(addr, length, prot, flags, fd, offset);

  // open — BEFORE: assert(real.ready()); ... real.openat(...)   ← crashes early in release
  // AFTER: real.openat(...)   ← lazily resolves if constructor hasn't run

  // ioctl, dup, dup2, dup3, fcntl, fork, mprotect, madvise — same: drop assert(real.ready()),
  // the call already goes through the self-resolving accessor.
```

  Deleted outright: ready(), initialized_, all eleven assert(InterposerContext::real.ready()) sites,
  and the RTLD_NEXT-fallback blocks currently duplicated inside
  fopen/stat/lstat/access/opendir/fstat/readlink (that logic now lives once, in resolve()).

  Why this closes all four findings at once

  - BLOCKER + munmap MAJOR — errno-correct raw_mmap/raw_munmap.
  - Completeness MAJOR — open/close/ioctl/dup*/fcntl/fork/mprotect/madvise now tolerate early calls
  too, because the accessor self-resolves; the fix is no longer mmap/munmap-only.
  - Divergence MAJOR — one accessor shape replaces the three styles (raw-fast-path /
  RTLD_NEXT-fallback / bare-assert).
  - Ordering MINOR — atomic members with release/acquire give the fn-ptr stores proper happens-before.
The one honest caveat

  This is a bigger change than a 63-line hotfix — it touches every wrapper and reworks
  LibcPassthrough. Two things to verify before adopting it: (a) that lazy dlsym is genuinely safe for
  open/ioctl/fork at early startup (the fs wrappers already rely on it, and since mmap never dlsyms,
  dlsym-calls-mmap can't recurse — so it's self-consistent), and (b) that util::lookup_symbol doesn't
  itself route through an interposed call. If the team wants to keep #8035 as a minimal targeted
  patch, the must-do subset is just the errno fix in the two raw helpers (BLOCKER + MAJOR); the
  accessor refactor can be a follow-up that retires the asserts.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
newling: Agent says:

"""This is using libc's syscall(2) wrapper, not an inline raw syscall instruction. The wrapper preserves the usual libc error contract: on syscall failure it returns -1 and sets errno."""

So I think this is fine. Perhaps I shouldn't have called these methods raw_*, I can change that.

newling:
> I think we could get rid of our existing asserts and do a lazy init for the other calls and rely on our libc passthrough without readiness checks. This could also avoid some extra complexity here.

The following alternative to this PR is hopefully going in the direction you have in mind here: https://github.com/ROCm/rocm-systems/pull/8278 At least superficially, it removes readiness asserts and appears to not introduce new complexity (I say 'superficially' as I am not comfortable with this part of the codebase to review the agents work reliably). 

The following alternative to this PR is generated based on the agent's review comment: https://github.com/ROCm/rocm-systems/pull/8279 so that is the third option. 

I've confirmed that all 3 fix my initial reproducer.

emulation/rocjitsu/tests/CMakeLists.txt
```diff
@@ -442,6 +442,26 @@ set(RJ_KMD_PRELOAD_ENV
...
+
+add_test(
+    NAME "InterposerTest.EarlyMmapBeforeRocjitsuConstructor"
+    COMMAND $<TARGET_FILE:early_mmap_driver>
+)
+set_tests_properties(
+    "InterposerTest.EarlyMmapBeforeRocjitsuConstructor"
+    PROPERTIES
+        ENVIRONMENT
+            "LD_PRELOAD=$<TARGET_FILE:rocjitsu_shared>:$<TARGET_FILE:early_mmap_preload>"
```
kuhar: This test would be safer if the setup made the two test contracts explicit.

First, please register/install it through the usual test path, or state why it is intentionally build-tree-only: the direct `add_test` at `emulation/rocjitsu/tests/CMakeLists.txt:454` bypasses `rj_add_test`, and the new `early_mmap_driver`/`early_mmap_preload` targets are not installed when `RJ_INSTALL_TESTS` is enabled.

Second, the `LD_PRELOAD` order at `emulation/rocjitsu/tests/CMakeLists.txt:462` is what makes `early_mmap_preload`'s constructor run before `init_interposer`. If the order changes, or `early_mmap_preload` gains a `DT_NEEDED` edge on `librocjitsu`, the test can still pass while using the already-initialized `mmap` path. Please add a comment near the preload list and, if practical, a check in `early_mmap_preload_test.cpp:22` that fails unless the pre-init fallback was actually taken.

kuhar (approved):
LGTM

emulation/rocjitsu/tests/CMakeLists.txt
newling: > First, 

I updated the test to use the normal `rj_add_test` path and added install handling for the helper targets when installed tests are enabled.

> Second, 

Good spot. I documented the `LD_PRELOAD` ordering contract next to the test setup. I looked into making the helper assert that the pre-init fallback was definitely taken, but the only option I found required adding observable test state/exported probes to the interposer itself. I’d rather avoid instrumenting library code for a test. Hopefully the comment will prevent regressing on this test.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
newling: I see the idea of setting errno is done in https://github.com/ROCm/rocm-systems/pull/8357. My understanding is that the difference is that there it is not using libc's syscall(2) wrapper.

PR #8036, opened 2026-06-30
ci(amdsmi): Change mutable git tag to immutable commit hash

gabrpham:
## Motivation
Security scan discovered supply chain vulnerability where attacker could inject own code if the tag was changed to point to said code.

## Technical Details
Swapped download and installation via mutable git tag to via immutable git hash commit.

## JIRA ID
JIRA ID : ROCM-27024

### DIFF
--- a/projects/amdsmi/CMakeLists.txt
+++ b/projects/amdsmi/CMakeLists.txt
@@ -259,21 +259,38 @@ set(ROCM_SRC_DIR "${PROJECT_SOURCE_DIR}/rocm_smi/src")
 set(ROCM_INC_DIR "${PROJECT_SOURCE_DIR}/rocm_smi/include/rocm_smi")
 set(SHR_MUTEX_DIR "${PROJECT_SOURCE_DIR}/third_party/shared_mutex")
 if(ENABLE_ESMI_LIB)
-    # Pin to an immutable commit, not a mutable tag (ROCM-27024).
+    # Pin to an immutable commit hash so the sources can't be swapped by a moved tag.
     set(ESMI_GIT_HASH "d494a3194ceb4cc4dbb2debf9fcbe8773c6d3bef")
     set(ESMI_SOURCE_DIR "${PROJECT_SOURCE_DIR}/esmi_ib_library")
 
-    # Use existing sources if present (vendored/offline); only fetch when absent.
+    # Drop an existing checkout pinned to a different commit so a hash bump
+    # re-fetches instead of silently building stale sources.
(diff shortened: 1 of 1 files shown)


projects/amdsmi/CMakeLists.txt
```diff
@@ -259,54 +259,19 @@ set(ROCM_SRC_DIR "${PROJECT_SOURCE_DIR}/rocm_smi/src")
...
+        esmi_ib_library
+        GIT_REPOSITORY https://github.com/amd/esmi_ib_library.git
+        GIT_TAG d494a3194ceb4cc4dbb2debf9fcbe8773c6d3bef
+        GIT_SHALLOW FALSE
+        SOURCE_DIR
+        ${PROJECT_SOURCE_DIR}/esmi_ib_library
+    )
+    FetchContent_GetProperties(esmi_ib_library)
+    if(NOT esmi_ib_library_POPULATED)
+        FetchContent_Populate(esmi_ib_library)
```
marifamd: Can you test locally and see if the git tag actually updates if you give it a new hash. The reasoning for the bottom half of the code you removed was to ensure that the locally built esmi lib was actually getting updated whenever the tag updated.

gabrpham: local test confirmed that the pinned hash version is picked up regardless of what local version you currently have.

marifamd (approved):

adam360x (approved):

dmitrii-galantsev:
hmmm will this break offline installs more than they're already broken?

dmitrii-galantsev (approved):

projects/amdsmi/CMakeLists.txt
```diff
@@ -259,65 +259,33 @@ set(ROCM_SRC_DIR "${PROJECT_SOURCE_DIR}/rocm_smi/src")
...
+        FetchContent_Declare(
+            esmi_ib_library
+            GIT_REPOSITORY https://github.com/amd/esmi_ib_library.git
+            GIT_TAG ${ESMI_GIT_HASH}
+            SOURCE_DIR
+            ${ESMI_SOURCE_DIR}
+            # Download-only: sources are compiled into amd_smi, not added as a
+            # subdir, so point SOURCE_SUBDIR at a dir with no CMakeLists.txt.
+            SOURCE_SUBDIR
+            __download_only__
```
marifamd, line 295: This will break rebuilding locally, I'm writing a patch

marifamd (changes requested):

marifamd (approved):

PR #8041, opened 2026-06-30
fix(rocjitsu): make MFMA SIMD reads visible to plugins

### LINKED ISSUE #8040: rocjitsu: MFMA SIMD reads bypass VGPR read plugins
MFMA SIMD fast paths stage inputs through raw VGPR storage to avoid per-lane read overhead. That keeps execution fast, but it also bypasses VGPR read plugin hooks, so plugins such as the race detector can miss hazards when an MFMA source register is read while an outstanding memory operation still owns that VGPR value.\n\nThe fix should keep the raw staging model while explicitly observing the MFMA source register reads with lane-mask VGPR read notifications before raw storage is consumed.\n\nA regression should cover the case where an unresolved write to an MFMA source VGPR is followed by a specialized MFMA SIMD fast-path read.

newling:
fixes : #8040
replaced by: https://github.com/ROCm/rocm-systems/pull/8075

## Motivation

The MFMA SIMD fast paths deliberately avoid `read_vgpr()` in their inner loops. They stage matrix inputs through `raw_vgpr_data()` so the emulator can keep MFMA execution vectorized instead of paying per-lane virtual read overhead. That is good for performance, but it currently means there is a gap for the race detector plugin.

A pending memory operation can still own the value of a VGPR lane when a later instruction reads that VGPR. For ordinary instruction paths, the VGPR read hook gives the race detector a chance to validate the access. For the specialized MFMA SIMD paths, the raw pointer staging skipped that hook, so an MFMA could read an unresolved source register without the race detector seeing the read.

## Technical Details

This PR keeps the raw staging model, but makes the read observation explicit. Before the MFMA fast path reads through raw VGPR storage, it observes the contiguous VGPR windows used for A, B, and the accumulator source. Those windows are reported through lane-mask VGPR read notifications, so the detector still validates with its existing mask-based logic while the MFMA implementation remains vectorized.

I do not see this as the final shape for all register observation. Long term, I would like the emulator to have a tighter bottleneck for instruction-visible register reads so that raw storage access is harder to use accidentally. This PR is a focused step for the MFMA fast-path blind spot.

The supporting VGPR read-hook changes in this branch make that representation natural. A lane mask is a better fit than a contiguous lane range for vectorized instruction paths, and it lets MFMA describe full-window reads without turning the observation path back into a loop of scalar hook calls.

## Test Plan

The tests cover the change at two levels.

The build-independent tests exercise the MFMA read-observation layer itself. One checks that observation emits multi-lane VGPR read hooks. Another wires that observation into the race detector and shows that an unresolved source register read is reported as a race.

The fast-path regression test executes the specialized MFMA SIMD path. In an AVX-512 build, that test fails on `origin/develop` because the race detector sees no read, and passes with this PR because the MFMA source read is now explicitly observed.

## Test Result

Normal Clang build: focused MFMA, race-detector, and plugin tests pass. The AVX-512-only fast-path regression test skips in this build when the 16-lane native float fast path is not active.

GCC `-march=native` AVX-512 regression check: the fast-path regression test fails on `origin/develop` with no race violation reported, and passes with this PR.

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/checkpoint.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/checkpoint.cpp
@@ -119,7 +119,8 @@ void save_checkpoint(const std::string &path, const SoC &soc, uint64_t tick,
               builder.CreateVector(cu->sgpr_data(w->sgpr_alloc().base), w->num_sgprs());
           size_t vgpr_bytes = static_cast<size_t>(cu->vgpr_allocation_block_size()) *
                               static_cast<size_t>(w->wf_size()) * sizeof(uint32_t);
-          auto vgprs_vec = builder.CreateVector(cu->vgpr_data(w->vgpr_alloc().base), vgpr_bytes);
+          auto vgprs_vec =
+              builder.CreateVector(cu->raw_vgpr_data(w->vgpr_alloc().base), vgpr_bytes);
 
           auto wfs = fb::CreateWavefrontState(builder, w->wf_id(), w->wg_id(), w->pc, w->exec(),
                                               w->vcc(), w->m0(), w->is_halted(), w->status_raw(),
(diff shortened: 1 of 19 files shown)


newling:
Superseded by #8075, which keeps the same changes on the descriptive branch `mfma-plugin-read-observation`.

PR #8048, opened 2026-07-01
Enable Fabric handles support when using ncclMemAlloc

AbandiGa:
## JIRA ID
AICOMRCCL-928

## Submission Checklist

### DIFF
--- a/projects/rccl/src/allocator.cc
+++ b/projects/rccl/src/allocator.cc
@@ -36,13 +36,16 @@ ncclResult_t  ncclMemAlloc_impl(void **ptr, size_t size) {
 
   if (ncclCuMemEnable()) {
     size_t handleSize = size;
-    int requestedHandleTypes = CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR;
-#if CUDART_VERSION >= 12030
-    // Query device to see if FABRIC handle support is available
-    flag = 0;
-    (void) CUPFN(cuDeviceGetAttribute(&flag, CU_DEVICE_ATTRIBUTE_HANDLE_TYPE_FABRIC_SUPPORTED, currentDev));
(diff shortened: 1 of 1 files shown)


nileshnegi (approved):

projects/rccl/src/allocator.cc
```diff
@@ -36,13 +36,16 @@ ncclResult_t  ncclMemAlloc_impl(void **ptr, size_t size) {
...
     size_t handleSize = size;
-    int requestedHandleTypes = CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR;
-#if CUDART_VERSION >= 12030
-    // Query device to see if FABRIC handle support is available
-    flag = 0;
-    (void) CUPFN(cuDeviceGetAttribute(&flag, CU_DEVICE_ATTRIBUTE_HANDLE_TYPE_FABRIC_SUPPORTED, currentDev));
-    if (flag) requestedHandleTypes |= CU_MEM_HANDLE_TYPE_FABRIC;
-#endif
+    int requestedHandleTypes = ncclCuMemHandleType;
+    if (requestedHandleTypes == CU_MEM_HANDLE_TYPE_FABRIC) {
```
atulkulk, line 40: Are we going to support everything via only fabric handles if available else fallback to only POSIX?

nileshnegi: seems like HIP does not support a combined bitmask like `CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR | CU_MEM_HANDLE_TYPE_FABRIC`

so we will try fabricHandles first, and if unavailable try posixFD

atulkulk: Ok. We should bring this up with the HIP team to support it.

atulkulk (approved):

PR #8055, opened 2026-07-01
clr: fix HIP graph host callback visibility for GPU writes to CPU-accessible memory

LeoZhao-BJ:
## Summary

  This change fixes HIP graph host callback visibility for GPU writes to CPU-accessible memory.
  
  JIRA ID:  ROCM-26916

  The regression was exposed after removing the pre/post marker path around uncaptured graph nodes. In segmented graph launch, an uncaptured host node can
  run after a captured GPU producer, but the host callback may not observe GPU writes to host-visible memory unless the handoff uses system scope.

  This patch updates `GraphHostNode` to use a system-scope acquire marker before invoking the host callback. This ensures preceding GPU writes to CPU-
  accessible memory are visible before the callback reads them.

  The fix covers cases such as:

  - captured blit/shader D2H copy writes host memory, followed by a host node reading that host memory
  - kernel writes directly to managed memory, followed by a host node reading that memory
  - kernel writes directly to pinned host memory, followed by a host node reading that memory

  The patch also keeps the single-packet flat AQL batch system-scope release fix. For a single-packet batch, `firstHeader` and `lastHeader` refer to the
  same packet, but only `firstHeader` is committed through `packet_store_release()`. When system scope is requested, the release fence must therefore also
  be applied to `firstHeader`.

## Details

2 Issues as shown below

### Host callback visibility

  Previously, `GraphHostNode::CreateCommand()` created a marker only to trigger the host callback and used:

  ```cpp
  command->setCommandEntryScope(amd::Device::kCacheStateIgnore);
  ```
  That is not sufficient when the callback reads memory recently written by GPU work.
  This patch changes the marker to:

  ```cpp
  command->setCommandEntryScope(amd::Device::kCacheStateSystem);
  ```
  so the host callback is preceded by a system-scope acquire.

  This is needed because the host node itself does not dispatch an AQL packet that could consume addSystemScope_. The marker before the callback is the correct synchronization point for making prior GPU writes visible to the CPU callback.

### Single-packet flat batch system release

  For flat AQL graph batches, system scope is applied to the first packet acquire and last packet release. However, when the batch contains only one packet,
  the first and last packet are the same physical packet, and only firstHeader is committed with packet_store_release().

  This patch applies the system release scope to firstHeader as well when:

  numPackets == 1

  so single-packet batches correctly publish their writes.

## Tests

  Added HIP graph regression coverage for host callback visibility:

  - Unit_hipLaunchHostFunc_BlitDtoHHostCallbackVisibility
      - captured blit D2H writes host memory
      - host node reads the host buffer
      - verifies callback-time visibility and post-sync visibility
  - Unit_hipLaunchHostFunc_SdmaDtoHHostCallbackVisibilityNoProducer
      - SDMA D2H writes host memory
      - host node reads the host buffer
      - verifies SDMA D2H completion already provides the required host visibility
  - Unit_hipLaunchHostFunc_KernelWritesManagedMem_HostNodeReads
      - kernel writes managed memory
      - host node reads managed memory directly
  - Unit_hipLaunchHostFunc_KernelWritesPinnedMem_HostNodeReads
      - kernel writes pinned host memory
      - host node reads pinned host memory directly

## Test Plan
update hipLaunchHostFuncWithGraph.cc, and create new test case unit_hipLaunchHostFunc_SameStreamCapturedDtoH_UncapturedHost for uncovered path.

 Built `clr` and `hip-tests` in Debug mode for `gfx950`.

  ```bash
  AMD_LOG_LEVEL=4 ctest -V \
    -R "^(Unit_hipGraphInstantiateWithFlags_AutoFreeOnLaunchDoubleKernel|Unit_hipLaunchHostFunc_CrossStreamDep_UncapturedFirstNode|Unit_hipLaunchHostFunc_BlitDtoHHostCallbackVisibility|Unit_hipLaunchHostFunc_SdmaDtoHHostCallbackVisibilityNoProducer|Unit_hipLaunchHostFunc_KernelWritesManagedMem_HostNodeReads|Unit_hipLaunchHostFunc_KernelWritesPinnedMem_HostNodeReads)$" \
    --output-on-failure
  ```

## Test Result

100% tests passed

## Submission Checklist


projects/clr/hipamd/src/hip_graph_internal.cpp
```diff
@@ -2320,6 +2320,24 @@ hipError_t GraphExec::EnqueueSegment(const Segment& segment, hip::Stream* stream
 
   size_t batchIndex = 0;
 
+  // A downstream segment can begin with an uncaptured host node whose callback
+  // reads host memory produced by this segment's captured AQL batch.  When this
+  // segment reaches its tail batch, publish those GPU writes with system scope
+  // before the host callback is allowed to run.
+  const bool host_first_consumer = [&]() -> bool {
```
jaydeeppatel1111: Just minor design nit, Is it more appropriate to move system scope for host node on consumer side instead? In that case, we don't need to find following call is host call everytime.

LeoZhao-BJ: removed on latest change

anugodavar (approved):

projects/clr/rocclr/device/rocm/rocvirtual.cpp
```diff
@@ -1546,6 +1546,13 @@ bool VirtualGPU::dispatchAqlPacketBatchFlat(const std::vector<uint8_t>& flatPack
     firstHeader |= (HSA_FENCE_SCOPE_SYSTEM << HSA_PACKET_HEADER_SCACQUIRE_FENCE_SCOPE);
     lastHeader &= ~(HSA_FENCE_SCOPE_AGENT << HSA_PACKET_HEADER_SCRELEASE_FENCE_SCOPE);
     lastHeader |= (HSA_FENCE_SCOPE_SYSTEM << HSA_PACKET_HEADER_SCRELEASE_FENCE_SCOPE);
```
saleelk, line 1548: when numPackets=1 , lastHeader=ValidFullHeader[0] so it should point to the same packet. I dont think we need that if there

saleelk (changes requested):

projects/clr/rocclr/device/rocm/rocvirtual.cpp
LeoZhao-BJ: here, firstHeader/lastHeader are just local value variables, equals but not points to same ValidFullHeader[0], so value updating can't reflect on same place.

LeoZhao-BJ: the place that real problem happens is here,  i starts from 1 if only one packet, then no chance to store release from last_header.

        for (size_t i = (isFirstChunk ? 1 : chunkStart); i < chunkEnd; ++i) {
      const uint64_t idx = startIndex + i;
      auto* aql_loc =
          reinterpret_cast<uint32_t*>(queueBase + (idx & queueMask) * kPacketSize);
      const uint32_t dword = validFullHeaders[i];
      const uint16_t hdr = (i == numPackets - 1) ? lastHeader : static_cast<uint16_t>(dword);
      packet_store_release(aql_loc, hdr, static_cast<uint16_t>(dword >> 16));
    }

saleelk (approved):

PR #8057, opened 2026-07-01
[clr] [hip-tests] Flush caches with system scope before graph host-no…

jaydeeppatel1111:
…de callback.

## Motivation

A HIP graph host node's callback runs on the CPU and often reads host memory produced by an upstream GPU node (e.g. a device-to-host copy feeding the callback). The callback marker was created with a NOP-scope fence (kCacheStateIgnore), which only orders GPU work and does not flush the GPU L2 to system-visible memory. On GPUs where the producer's agent-scope release is not host-visible (observed on gfx1201 / NAVI48), the CPU callback reads stale data, while gfx1100 / NAVI3x happens to pass. This is a silent, architecture-specific data-correctness bug. Ordering was never the problem — the marker correctly waited for the producer to complete — the produced data simply wasn't made visible to the CPU before the callback ran.

## Technical Details

Fix (clr): in GraphHostNode::CreateCommand (hipamd/src/hip_graph_internal.hpp), the callback marker's command entry scope is changed from kCacheStateIgnore to kCacheStateSystem.

## JIRA ID

ROCM-27311

## Test Plan

New test added to test coherency
Unit_hipLaunchHostFunc_CrossStreamDep_UncapturedFirstNode

## Test Result

Unit_hipLaunchHostFunc_CrossStreamDep_UncapturedFirstNode pass
Unit_hipGraphAddHostNode_CallbackSeesUpstreamDtoH pass

## Submission Checklist

### DIFF
--- a/projects/clr/hipamd/src/hip_graph_internal.hpp
+++ b/projects/clr/hipamd/src/hip_graph_internal.hpp
@@ -2901,8 +2901,13 @@ class GraphHostNode : public GraphNode {
     amd::Command::EventWaitList waitList;
     commands_.reserve(1);
     amd::Command* command = new amd::Marker(*stream, !kMarkerDisableFlush, waitList);
-    // This is just to invoke a callback, so no need to flush caches.
-    command->setCommandEntryScope(amd::Device::kCacheStateIgnore);
+    // The host callback runs on the CPU and may read memory produced by an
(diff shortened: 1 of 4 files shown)


jaydeeppatel1111:
Superseded by https://github.com/ROCm/rocm-systems/pull/8065

PR #8061, opened 2026-07-01
fix(rocprofiler-sdk): harden env-var dlopen and pickle attach cache (…

bgopesh:
…SEC-00398, SEC-00752)

Address two AI security-scan findings:

SEC-00398: An unprivileged local user could inject an arbitrary shared library into a privileged (setuid/setgid/file-capabilities) process via ROCP_TOOL_LIBRARIES or ROCPROF_ATTACH_TOOL_LIBRARY, which are passed to dlopen. Add a common::is_at_secure() helper (getauxval(AT_SECURE)) and ignore both environment variables in secure-execution contexts.

SEC-00752: The attach-mode cache file at the predictable path /tmp/rocprofv3_attach_<pid>.pkl was deserialized with pickle, enabling arbitrary code execution if another user planted a payload. Create the file exclusively (O_CREAT|O_EXCL|O_WRONLY|O_NOFOLLOW, 0600) and only unpickle it when opened with O_RDONLY|O_NOFOLLOW and owned by the current user, so we only ever load a file we created ourselves. Failures are converted into clear fatal_error messages.

## JIRA ID
JIRA ID: ROCM-26858
JIRA ID: ROCM-26869

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-sdk/source/bin/rocprofv3.py
+++ b/projects/rocprofiler-sdk/source/bin/rocprofv3.py
@@ -2390,17 +2390,52 @@ def main(argv=None):
             if args.collection_period:
                 fatal_error("--collection-period is not compatible with attach mode")
 
+            # SECURITY: this cache file lives at a predictable path in a world-writable
+            # directory (/tmp) and is deserialized with pickle, which would execute
+            # arbitrary code if another user on a shared system planted a malicious
+            # payload. To prevent this, the file is created exclusively (O_EXCL) with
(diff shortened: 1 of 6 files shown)


meserve-amd (commented):
This change is good. However, I think similar vulnerabilities exist with ROCPROF_ATT_LIBRARY_PATH that should be addressed as well.

meserve-amd (approved):

jrmadsen (approved):

PR #8064, opened 2026-07-01
fix(rocprofiler-systems): handle indexed GPU counter files

habajpai-amd:
## Motivation

ROCProfiler counter validation tests were failing on gfx1250 even though the requested counters were available and counter output files were generated.
The failing tests were:
- `transpose-rocprofiler-sampling`
- `transpose-rocprofiler-binary-rewrite`
The runtime generated files such as:
- `rocprof-device-0-GRBM_COUNT-0.txt`
- `rocprof-device-0-SQ_WAVES-0.txt`
- `rocprof-device-0-SQ_INSTS_VALU-0.txt`
- `rocprof-device-0-TX_VCA_VCA_BUSY-0.txt`
but the pytest validation expected filenames without the trailing output index, for example:
- `rocprof-device-[0-9]-GRBM_COUNT.txt`
This caused a Systems-side validation failure even though the counter data was present.

## Technical Details

Updated the ROCProfiler counter test harness to accept indexed counter output filenames by changing the expected glob from: rocprof-device-[0-9]-<counter>.txt
to:
rocprof-device-[0-9]-<counter>*.txt
Also made ROCpd validation more robust for MPI runs. MPI tests can emit multiple rank/process ROCpd databases. The validation now considers all emitted ROCpd DB candidates and passes if one validates, avoiding failures caused by selecting a rank DB that does not contain the GPU counter rows.

Updated lightweight pytest coverage for:
- gfx1250 expected counter file patterns
- deterministic ROCpd DB candidate discovery

## JIRA ID

JIRA ID - AIPROFSYST-590

## Test Plan

gfx1250 targeted validation with public TheRock ROCm 7.15.0a20260630 and ROCprofiler-SDK 1.3.2:
ctest --test-dir build/public-gfx125x-rocm715 \
   -V -R '^(transpose-rocprofiler-sampling|transpose-rocprofiler-binary-rewrite|transpose-gpu-perf-counters)$' \
   --output-on-failure

## Test Result

gfx1250 targeted CTest passed:
- transpose-rocprofiler-sampling
- transpose-rocprofiler-binary-rewrite
- transpose-gpu-perf-counters

image

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-systems/tests/pytest/conftest.py
+++ b/projects/rocprofiler-systems/tests/pytest/conftest.py
@@ -2441,14 +2441,20 @@ def assert_rocpd(subtests, tests_dir, request):
                     gpu_category_to_skip=gpu_category_to_skip,
                 )
 
-            passing_output, failures = _validate_rocpd_candidates(
+            passing_output, failures, global_failure = _validate_rocpd_candidates(
                 rocpd_files,
                 validate_candidate,
                 pass_regex=pass_regex,
                 fail_regex=fail_regex,
             )
 
-            if passing_output is None:
+            if global_failure is not None:
+                msg = fail_message or f"ROCpd validation failed:\n{global_failure}"
+                if skip_on_fail:
+                    pytest.skip(msg)
+                else:
+                    pytest.fail(msg, pytrace=False)
+            elif passing_output is None:
                 output = "\n\n--- Next ROCpd candidate ---\n\n".join(failures)
                 msg = fail_message or f"ROCpd validation failed:\n{output}"
                 if skip_on_fail:
@@ -2465,18 +2471,23 @@ def _validate_rocpd_candidates(
     validate_candidate: Callable[[Path], ValidationResult],
     pass_regex: Optional[list[str]] = None,
     fail_regex: Optional[list[str]] = None,
-) -> tuple[Optional[str], list[str]]:
+) -> tuple[Optional[str], list[str], Optional[str]]:
     """Validate ROCpd candidates and return the first passing output.
 
     Multi-process runs can emit multiple ROCpd databases. Some rank-local
     databases may not contain the GPU rows required by a rule set, so the
-    validation succeeds if any emitted candidate fully validates.
+    validation succeeds if any emitted candidate fully validates. A fail regex
+    match is a global failure and stops validation immediately.
     """
     failures: list[str] = []
 
     for rocpd_file in rocpd_files:
         validation = validate_candidate(rocpd_file)
         output = f"Command: {validation.command}\n\n{validation.message}"
+        if fail_regex:
+            for pattern in fail_regex:
+                if re.search(pattern, validation.stdout):
+                    return None, failures, f"Fail regex found: {pattern}\n{output}"
         if not validation.is_valid:
             failures.append(output)
             continue
@@ -2487,18 +2498,13 @@ def _validate_rocpd_candidates(
                 if not re.search(pattern, validation.stdout):
                     regex_failure = f"Pass regex not found: {pattern}"
                     break
-        if regex_failure is None and fail_regex:
-            for pattern in fail_regex:
-                if re.search(pattern, validation.stdout):
-                    regex_failure = f"Fail regex found: {pattern}"
-                    break
         if regex_failure is not None:
             failures.append(f"{regex_failure}\n{output}")
             continue
 
-        return output, failures
+        return output, failures, None
 
-    return None, failures
+    return None, failures, None
 
 
 @pytest.fixture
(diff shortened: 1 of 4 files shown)


dgaliffiAMD (approved):

projects/rocprofiler-systems/tests/pytest/rocprofsys/runners.py
```diff
@@ -90,12 +90,15 @@ def perfetto_file(self) -> Optional[Path]:
 
     @property
     def rocpd_file(self) -> Optional[Path]:
```
kcossett-amd: This is now unused (I'm not counting `test_pytest_impl.py`).

projects/rocprofiler-systems/tests/pytest/conftest.py
```diff
@@ -2431,37 +2432,75 @@ def _assert_rocpd(
...
-                            f"Fail regex found: {pattern}\n{output}", pytrace=False
-                        )
-            _print_subtest_output(request, subtest_name, output)
+                    pytest.fail(msg, pytrace=False)
+            _print_subtest_output(request, subtest_name, passing_output)
 
     return _assert_rocpd
 
 
+def _validate_rocpd_candidates(
```
kcossett-amd, line 2469: `fail_regex` is evaluated per candidate `db`, and only disqualifies that `db` alone.
The function returns as soon as any candidate passes, and only reports failure if every candidate is disqualified.

I'd prefer that `fail_regex` is stricter. I.e. a single failure fails the entire thing. That matches behavior of other fail regexes.

projects/rocprofiler-systems/tests/pytest/test_pytest_impl.py
```diff
@@ -6,9 +6,11 @@
 """
 
 from __future__ import annotations
+from pathlib import Path
+
 import pytest
-from conftest import RocprofsysTest
-from rocprofsys import GPUInfo
+from conftest import RocprofsysTest, _validate_rocpd_candidates
+from rocprofsys import GPUInfo, TestResult, ValidationResult
```
kcossett-amd: Importing `TestResult` causes a warning during build:
```
[100%] Generating CTest definitions from pytest suite
============================================================================= test session starts =============================================================================
collected 612 items                                                                                                                                                           
Generated 612 CTest definitions -> /home/rocm/rocm-systems-branches/harsha-review/projects/rocprofiler-systems/rocprof-sys-build/tests/CTestTestfile.cmake

============================================================================== warnings summary ===============================================================================
rocprofsys/runners.py:45
  /home/rocm/rocm-systems-branches/harsha-review/projects/rocprofiler-systems/rocprof-sys-build/share/rocprofiler-systems/tests/pytest/rocprofsys/runners.py:45: PytestCollectionWarning: cannot collect test class 'TestResult' because it has a __init__ constructor (from: rocprof-sys-build/share/rocprofiler-systems/tests/pytest/test_pytest_impl.py)
    @dataclass

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
============================================================================= 1 warning in 0.52s ==============================================================================
```

I'd suggest in `runners.py`, that you mark `TestResult` class with `__test__ = False` so it does not think it is a pytest test class.

sputhala-amd (approved):
The code changes look good.
However, I don't think the underlying issue is specific to `gfx1250` but more around using MPI for transpose tests, which actually surfaced pre-existing gaps in our tests that our MPI tests should have caught. On CIs we do not build transpose with MPI.
`MPI_Init` triggers the MPI gotcha, which forces `use_output_suffix()` on and appends the per-rank suffix to all outputs.
So good that you caught this issue and fixed it.

projects/rocprofiler-systems/tests/pytest/rocprofsys/runners.py
```diff
@@ -90,12 +90,15 @@ def perfetto_file(self) -> Optional[Path]:
...
+    @property
+    def rocpd_files(self) -> list[Path]:
         candidate = self.output_dir / "rocpd.db"
         if candidate.exists():
-            return candidate
-        # Try globbing
-        dbs = list(self.output_dir.glob("*.db"))
-        return dbs[0] if dbs else None
+            return [candidate]
+        return sorted(self.output_dir.glob("*.db"))
```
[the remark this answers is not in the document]
habajpai-amd: For this PR, the ordering only needs to be deterministic because ROCpd validation now iterates over all emitted DB candidates and succeeds when any candidate validates. The validation path no longer depends on selecting the numerically lowest PID DB.
I’m going to keep the lexicographic sort as-is for this fix to avoid adding a separate natural-sort behavior change. We can add natural sorting later if we need `rocpd_file` itself to represent numeric PID order.

projects/rocprofiler-systems/tests/pytest/test_pytest_impl.py
habajpai-amd: Thank you for the review comments!

PR #8069, opened 2026-07-01
[CLR] Enable staticCO to allow concurrent reads

AidanBeltonS:
## Motivation

Multiple threads all accessing different GPUs with different streams are all serialized when they try retrieving the function. This slows submission and increases variability in submission,

## Technical Details

The regular mutex is replaced with a shared mutex allowing concurrent accesses from multiple threads. Additionally `managedVarsDevicePtrInitalized_` now stores atomic values, which are atomically updated, this produces very little contention after setup. 

Results from multi-gpu submissions benchmarking. Significant decrease in latency and submission variablitiy

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

  | **Averages** |  |  | %   **Change** | |  | 
-- | -- | -- | -- | -- | -- | --
  | Kernel Submission Times | | Total Run | Kernel Submission Times | | Total Run
  | Mean (Us) | Stdev | Ops/Sec | Mean | Stdev | Ops/Sec
**Baseline** | 2220.5 | 2210.5 | 1575776.5 |   |   |  
**StatCo   Concurrent** | 1108.5 | 645.5 | 2176200.5 | -50% | -71% | 38%

</body>

</html>

## JIRA ID

ROCM-27455

## Test Plan

Additional testing added to harden the feature

## Test Result

Pass locally

## Submission Checklist

### DIFF
--- a/projects/clr/hipamd/src/hip_code_object.cpp
+++ b/projects/clr/hipamd/src/hip_code_object.cpp
@@ -525,76 +525,77 @@ const char* StatCO::GetFuncName(const void* hostFunction) {
 }
 
 hipError_t StatCO::GetFunc(hipFunction_t* hfunc, const void* hostFunction, int deviceId) {
-  std::shared_lock<std::shared_mutex> lock(sclock_);
+  {
+    std::shared_lock<std::shared_mutex> lock(sclock_);
 
+    auto it = functions_.find(hostFunction);
+    if (it == functions_.end()) {
+      return hipErrorInvalidSymbol;
+    }
+
+    FatBinaryInfo** module = it->second->ModuleInfo();
+    if (module == nullptr) {
+      return hipErrorInvalidDeviceFunction;
+    }
+    if (*module != nullptr) {
+      return it->second->GetStatFunc(hfunc, deviceId);
+    }
+  }
+
+  // Lazy load: upgrade to exclusive lock.
+  std::unique_lock<std::shared_mutex> wlock(sclock_);
+  // Re-find after acquiring exclusive lock (concurrent removal may have occurred).
   auto it = functions_.find(hostFunction);
   if (it == functions_.end()) {
     return hipErrorInvalidSymbol;
   }
-
   FatBinaryInfo** module = it->second->ModuleInfo();
   if (module == nullptr) {
     return hipErrorInvalidDeviceFunction;
   }
-
   if (*module == nullptr) {
-    // Lazy load: upgrade to exclusive lock.
-    lock.unlock();
-    std::unique_lock<std::shared_mutex> wlock(sclock_);
-    // Re-find after acquiring exclusive lock (concurrent removal may have occurred).
-    it = functions_.find(hostFunction);
-    if (it == functions_.end()) {
-      return hipErrorInvalidSymbol;
+    hipError_t err = DigestFatBinary(module_to_hostModule_[module], *module);
+    if (err != hipSuccess) {
+      return err;
     }
-    module = it->second->ModuleInfo();
-    if (module == nullptr) {
-      return hipErrorInvalidDeviceFunction;
-    }
-    if (*module == nullptr) {
-      hipError_t err = DigestFatBinary(module_to_hostModule_[module], *module);
-      if (err != hipSuccess) {
-        return err;
-      }
-    }
-    return it->second->GetStatFunc(hfunc, deviceId);
   }
-
   return it->second->GetStatFunc(hfunc, deviceId);
 }
 
 hipError_t StatCO::GetFuncAttr(hipFuncAttributes* func_attr, const void* hostFunction,
                                int deviceId) {
-  std::shared_lock<std::shared_mutex> lock(sclock_);
+  {
+    std::shared_lock<std::shared_mutex> lock(sclock_);
 
+    auto it = functions_.find(hostFunction);
+    if (it == functions_.end()) {
+      return hipErrorInvalidSymbol;
+    }
+
+    FatBinaryInfo** module = it->second->ModuleInfo();
+    if (module == nullptr) {
+      return hipErrorInvalidDeviceFunction;
+    }
+
+    if (*module != nullptr) {
+      return it->second->GetStatFuncAttr(func_attr, deviceId);
+    }
+  }
+
+  // Lazy load: upgrade to exclusive lock.
+  std::unique_lock<std::shared_mutex> wlock(sclock_);
+  // Re-find after acquiring exclusive lock (concurrent removal may have occurred).
   auto it = functions_.find(hostFunction);
   if (it == functions_.end()) {
     return hipErrorInvalidSymbol;
   }
-
   FatBinaryInfo** module = it->second->ModuleInfo();
   if (module == nullptr) {
     return hipErrorInvalidDeviceFunction;
   }
-
   if (*module == nullptr) {
-    // Lazy load: upgrade to exclusive lock.
-    lock.unlock();
-    std::unique_lock<std::shared_mutex> wlock(sclock_);
-    // Re-find after acquiring exclusive lock (concurrent removal may have occurred).
-    it = functions_.find(hostFunction);
-    if (it == functions_.end()) {
-      return hipErrorInvalidSymbol;
-    }
-    module = it->second->ModuleInfo();
-    if (module == nullptr) {
-      return hipErrorInvalidDeviceFunction;
-    }
-    if (*module == nullptr) {
-      std::ignore = DigestFatBinary(module_to_hostModule_[module], *module);
-    }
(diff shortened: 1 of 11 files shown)


AidanBeltonS:
> ### ❌ PR Check — Action Required
> Check	Status	Details
> 🌿 **Branch Name**	✅ Pass	—
> 📝 **PR Title/Description**	❌ Fail	**Error:** Title does not follow Conventional Commits style.
> **Expected:** start with a valid type (feat, fix, docs, …).
> **Desired format:** `type(optional-scope): short description`
> ───
> **Error:** PR description must reference a JIRA ID, ISSUE ID, or a GitHub closing keyword.
> **Expected:** include a `JIRA ID` / `ISSUE ID` line (separator `:` or `-`, or omitted; value may be a JIRA key, a number with/without `#`, or a link), OR a closing keyword + issue reference. Accepted examples:
> • `JIRA ID : TESTAUTO-6039`
> • `JIRA ID - #330`
> • `JIRA ID #330`
> • `ISSUE ID : TESTUTO-3334`
> • `ISSUE ID #3334`
> • `ISSUE ID - TESTAUTO-3433`
> • `ISSUE ID : https://github.com/<org_name>/<repo_name>/issues/1234`
> • `Closes #10`
> • `Fixes octo-org/octo-repo#100`
> • `Resolves: #123`
> • `#123`
> • `https://github.com/<org_name>/<repo_name>/issues/123`
> **Current:** no valid JIRA/ISSUE/closing-keyword reference found
> ⛔ **Forbidden Files**	✅ Pass	—
> 🧪 **Unit Test**	❌ Fail	**Error:** Source/code files changed without an accompanying unit test.
> **Expected:** add at least one test file named like `test_<name>.py` / `test_<name>.cpp` (or `<name>_test.*`).
> **Current:** code file(s) changed: `projects/clr/hipamd/src/hip_code_object.cpp`, `projects/clr/hipamd/src/hip_code_object.hpp`, `projects/clr/hipamd/src/hip_global.cpp`, `projects/clr/hipamd/src/hip_global.hpp`, `projects/hip-tests/catch/unit/dynamicLoading/concurrentDynamicLoad.cc` (+2 more); no test file found
> 🔎 **pre-commit**	⏳ Pending	⏳ Still running…
> 🚫 **Draft PR**	🔜 To Be Enabled	—
> 🚩 **Feature Flag**	🔜 To Be Enabled	—
> 📊 **Code Coverage**	🔜 To Be Enabled	—
> > ⚠️ **2 policy check(s) failed.** Please address the issues above before this PR can be Reviewed.
> > 🚫 **Please fix the failed policies**
> > 
> > * ❌ PR Title/Description
> > * ❌ Unit Test
> > 
> > The **`Not ready to Review`** label was added to this PR. Once all policies pass, the label is removed automatically.
> 
> 📖 **Need help?** See the [Policy FAQ](https://github.com/ROCm/rocm-systems/tree/develop/docs/SYSTEMS_PR_BOT_FAQ.md) for details on every check and how to fix failures.

All false positives, this tool needs to be properly developed to reflect the projects coding standards

projects/clr/hipamd/src/hip_code_object.cpp
```diff
@@ -289,8 +289,6 @@ StatCO::~StatCO() {
 }
 
 hipError_t StatCO::DigestFatBinary(const void* data, FatBinaryInfo*& programs) {
-  std::scoped_lock lock(sclock_);
```
cjatin, line 302: we dont need a lock here?

projects/clr/hipamd/src/hip_code_object.cpp
```diff
@@ -517,45 +515,77 @@ const char* StatCO::GetFuncName(const void* hostFunction) {
 }
 
 hipError_t StatCO::GetFunc(hipFunction_t* hfunc, const void* hostFunction, int deviceId) {
-  const auto it = functions_.find(hostFunction);
+  {
+    std::shared_lock<std::shared_mutex> lock(sclock_);
```
cjatin, line 551: we are kind of doing similar ops in shared_lock vs unique_lock path (except the Digest part).
can we simplify this a bit?

projects/clr/hipamd/src/hip_code_object.cpp
```diff
@@ -517,45 +515,77 @@ const char* StatCO::GetFuncName(const void* hostFunction) {
...
 
 hipError_t StatCO::GetFuncAttr(hipFuncAttributes* func_attr, const void* hostFunction,
                                int deviceId) {
-  std::scoped_lock lock(sclock_);
+  {
+    std::shared_lock<std::shared_mutex> lock(sclock_);
 
-  const auto it = functions_.find(hostFunction);
+    auto it = functions_.find(hostFunction);
+    if (it == functions_.end()) {
```
cjatin: same comment here, different locks have same path.

projects/clr/hipamd/src/hip_code_object.cpp
```diff
@@ -574,34 +604,55 @@ hipError_t StatCO::RegisterGlobalVar(const void* hostVar, Var* var) {
...
-  *dev_ptr = memDevPtr(mem);
-  *size_ptr = mem->getSize();
+  // Handle size-0 globals: return null device pointer and size 0.
+  *dev_ptr = (mem == nullptr) ? 0 : memDevPtr(mem);
+  *size_ptr = (mem == nullptr) ? 0 : mem->getSize();
   return hipSuccess;
 }
 
 hipError_t StatCO::RegisterManagedVar(Var* var) {
+  std::scoped_lock lock(sclock_);
```
cjatin, line 673: I expect this to be called from `__hipRegisterManagedVar`. can we have issues with multiple access from this?

projects/hip-tests/catch/unit/dynamicLoading/concurrentDynamicLoad.cc
```diff
@@ -0,0 +1,270 @@
...
+/**
+ * @addtogroup hipLaunchKernelGGL
+ * @{
+ * @ingroup DynamicLoading
+ */
+
+/**
+ * Test Description
+ * ------------------------
+ * - Stress-tests the thread-safety of the HIP runtime's StatCO class by
```
cjatin, line 24: should this be in Unit tests?

projects/clr/hipamd/src/hip_code_object.cpp
AidanBeltonS: `DigestFatBinary` was a public function, but was only used internally to the class. It requires a lock to operate, the previous solution was to use a recursive_mutex, this adds overhead and confuses who "owns" the object under lock. So by removing the recursive_mutex and making this private, it is the responsibility of any StatCO:: function to lock before calling this helper. 

TLDR; Function is now internal helper func, lock must be acquired before calling it

projects/clr/hipamd/src/hip_code_object.cpp
AidanBeltonS: Done good suggestion. I have added the `FindRegistered()` helper which handles retrieving check and error code handling.

projects/clr/hipamd/src/hip_code_object.cpp
AidanBeltonS: Done

projects/clr/hipamd/src/hip_code_object.cpp
AidanBeltonS: Yes this is called from `__hipRegisterManagedVar`, it can in theory. Two separate threads could dlopen different libraries that contain `__managed__` variables. In practice this is very unlikely to occur. However, it is also not a hot path, so it makes sense to lock and be safe

projects/hip-tests/catch/unit/dynamicLoading/concurrentDynamicLoad.cc
AidanBeltonS: Its somewhat up to debate. I wanted to test the race conditions in StatCO but because the code objects are static this is the best way I have found to test this. dlopen triggers  AddFatBinary and RegisterFunction through __hipRegisterFatBinary and __hipRegisterFunction internal API paths. dlcose triggers __hipUnregisterFatBinary. 

The number of threads and the barriers are to try maximise contention to tigger failures. Which did actually work, the leak fixed in #8477 was found in this func.

TLDR; I think it is useful, as it already caught one bug. However, I agree it is not a normal unit test. When we start the process of setting test levels, this should be set to run very infrequently

cjatin:
Think it looks ok now.

AidanBeltonS:
Closing in favour of #10381

PR #8075, opened 2026-07-01
fix(rocjitsu): make MFMA SIMD reads visible to plugins

### LINKED ISSUE #8040: rocjitsu: MFMA SIMD reads bypass VGPR read plugins

newling:
Replaces #8041.

fixes : #8040

## Motivation

The MFMA SIMD fast paths deliberately avoid `read_vgpr()` in their inner loops. They stage matrix inputs through `raw_vgpr_data()` so the emulator can keep MFMA execution vectorized instead of paying per-lane virtual read overhead. That is good for performance, but it currently means there is a gap for the race detector plugin.

A pending memory operation can still own the value of a VGPR lane when a later instruction reads that VGPR. For ordinary instruction paths, the VGPR read hook gives the race detector a chance to validate the access. For the specialized MFMA SIMD paths, the raw pointer staging skipped that hook, so an MFMA could read an unresolved source register without the race detector seeing the read.

## Technical Details

This PR keeps the raw staging model, but makes the read observation explicit. Before the MFMA fast path reads through raw VGPR storage, it observes the contiguous VGPR windows used for A, B, and the accumulator source. Those windows are reported through lane-mask VGPR read notifications, so the detector still validates with its existing mask-based logic while the MFMA implementation remains vectorized.

I do not see this as the final shape for all register observation. Long term, I would like the emulator to have a tighter bottleneck for instruction-visible register reads so that raw storage access is harder to use accidentally. This PR is a focused step for the MFMA fast-path blind spot.

The supporting VGPR read-hook changes in this branch make that representation natural. A lane mask is a better fit than a contiguous lane range for vectorized instruction paths, and it lets MFMA describe full-window reads without turning the observation path back into a loop of scalar hook calls.

## Test Plan

The tests cover the change at two levels.

The build-independent tests exercise the MFMA read-observation layer itself. One checks that observation emits multi-lane VGPR read hooks. Another wires that observation into the race detector and shows that an unresolved source register read is reported as a race.

The fast-path regression test executes the specialized MFMA SIMD path. In an AVX-512 build, that test fails on `origin/develop` because the race detector sees no read, and passes with this PR because the MFMA source read is now explicitly observed.

## Test Result

Normal Clang build: focused MFMA, race-detector, and plugin tests pass. The AVX-512-only fast-path regression test skips in this build when the 16-lane native float fast path is not active.

GCC `-march=native` AVX-512 regression check: the fast-path regression test fails on `origin/develop` with no race violation reported, and passes with this PR.

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/simd_glue.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/simd_glue.h
@@ -638,7 +638,7 @@ template <typename T, typename Inst, typename BinOp>
   const uint64_t exec = wf.exec();
   // Resolve each operand's `VgprStorage*` ONCE (a single virtual
   // simd_vgpr_storage dispatch -> resolved_vgpr_offset -> the localized
-  // vgpr_reg<64> cast). The resolved register object is chunk-independent, so the
+  // raw_vgpr_reg<64> cast). The resolved register object is chunk-independent, so the
   // per-chunk loop issues a value-semantic `r->simd_load<T>(base)` off it instead
   // of re-dispatching the resolution every chunk — and no raw base pointer
   // escapes the kernel body. Operands that aren't contiguous VGPR storage
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/simd_glue.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/simd_glue.h
@@ -364,6 +364,23 @@ inline util::native<float> apply_vop3_dst_mod_f32(util::native<float> v, uint32_
   return apply_vop3_dst_mod<float>(v, omod, clamp);
 }
 
+// Transitional observed-storage helpers.
+//
+// SIMD instruction bodies need zero-copy access to resolved VGPR storage, but
+// the long-term invariant should be "acquire a register read/read-write view
+// and the view acquisition observes the read." Until that register-access facade
+// exists, keep all SIMD plugin notifications centralized here rather than
+// sprinkling `notify_vgpr_read` through individual execute helpers.
+//
+// The distinction between source and mutable-destination helpers is deliberate:
+// AMDGPU GPR indexing uses different role bits for read operands and write
+// operands. Dst-accumulate forms read the destination through the same
+// write-role resolution that is later used for the store, so they need the
+// `_dst_` helpers below to observe the actual register being loaded.
+//
+// These helpers should shrink away when SIMD glue is migrated to the unified
+// register-access facade.
+
 /// Resolve an operand's read-side VGPR storage and notify plugins for the lanes
 /// being read. Returns null for non-VGPR operands (SGPR / immediate /
 /// inline-const / DPP delegate). This is the observed-read bottleneck for SIMD
@@ -391,6 +408,17 @@ template <typename Op> VgprStorage *simd_dst_reg(const Op &op, Wavefront &wf) {
   return SimdAccess::vgpr_storage_mut(op, wf);
 }
 
+/// Mutable counterpart used only by dst-accumulate operations, where the
+/// destination register is also an instruction-visible source operand.
+template <typename Op>
+VgprStorage *observed_simd_dst_reg(const Op &op, Wavefront &wf, uint64_t lane_mask,
+                                   uint8_t byte_mask = 0xF) {
+  VgprStorage *r = SimdAccess::vgpr_storage_mut(op, wf);
+  if (r)
+    SimdAccess::notify_read_mut(op, wf, lane_mask, byte_mask);
+  return r;
+}
+
 /// Resolve a 64-bit-lane source operand's lo/hi VGPR pair and notify plugins
 /// for the lanes being read.
 template <typename Op>
@@ -398,7 +426,7 @@ ConstVgprStoragePair64 observed_simd_src_reg64(const Op &op, const Wavefront &wf
                                                uint64_t lane_mask, uint8_t byte_mask = 0xF) {
   ConstVgprStoragePair64 p = SimdAccess::vgpr_storage64(op, wf);
   if (p.lo)
-    SimdAccess::notify_read(op, wf, lane_mask, byte_mask);
+    SimdAccess::notify_read64(op, wf, lane_mask, byte_mask);
   return p;
 }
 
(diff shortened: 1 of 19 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/simd_glue.h
```diff
@@ -364,36 +364,51 @@ inline util::native<float> apply_vop3_dst_mod_f32(util::native<float> v, uint32_
...
 }
 
+/// Resolve a 64-bit-lane source operand's lo/hi VGPR pair and notify plugins
+/// for the lanes being read.
+template <typename Op>
+ConstVgprStoragePair64 observed_simd_src_reg64(const Op &op, const Wavefront &wf,
+                                               uint64_t lane_mask, uint8_t byte_mask = 0xF) {
+  ConstVgprStoragePair64 p = SimdAccess::vgpr_storage64(op, wf);
+  if (p.lo)
+    SimdAccess::notify_read(op, wf, lane_mask, byte_mask);
```
kuhar: The 64-bit SIMD read path still only notifies the base operand even though `simd_vgpr_storage64` binds `{reg, reg + 1}` and `simd_load64` consumes both halves. A plugin/race detector will see reads from the low VGPR but not the high VGPR, so an outstanding load to `src + 1` can be missed on f64 SIMD paths.

We should notify both physical registers here, or add a 64-bit notify helper that reports the high half as well.

emulation/rocjitsu/tests/execution_plugin_test.cpp
```diff
@@ -468,6 +517,119 @@ TEST(ExecutionPluginTest, NoPluginNoCrash) {
...
+                                   /*data_bits=*/16, wf->wf_size());
+
+  ASSERT_FALSE(race_plugin->violations.empty());
+  const auto &violation = race_plugin->violations.front();
+  EXPECT_EQ(violation.space, RaceViolation::Space::VGPR);
+  EXPECT_EQ(violation.index, static_cast<int>(S0));
+  EXPECT_EQ(violation.lane, 0);
+}
+
+TEST(ExecutionPluginTest, MfmaFastPathReadHookReportsRace) {
```
kuhar, line 814: The only test that exercises the new `observe_mfma_*` calls through an `exec_*_mfma_*_spec` fast path is skipped in the default build: `ExecutionPluginTest.MfmaFastPathReadHookReportsRace` reports `MFMA fast path requires 16-lane native<float>`. The helper tests still pass, but they do not cover the per-call-site arguments in the five spec functions.

Could we add a non-AVX-512 seam that validates the observed register set from those spec call sites, or make sure an AVX-512-enabled build runs this test? Otherwise CI can pass after dropping or misparameterizing one of the fast-path `observe_*` calls.

kuhar (commented):
Scope question: this PR documents `raw_vgpr_data` as bypassing plugin read hooks and adds `observe_mfma_*` calls for the MFMA fast paths, but the WMMA fast paths were also renamed to `raw_vgpr_data` and still have no read observation.

Is WMMA intentionally out of scope for this PR? If so, a short TODO or PR note would make the asymmetry explicit; otherwise we should extend the same observation/test coverage to WMMA so the race detector does not stay blind to WMMA fast-path input reads.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/simd_glue.h
```diff
@@ -446,11 +461,11 @@ template <typename T, typename Op>
   requires(util::has_stdx_simd)
 inline util::native<T> read_simd(const Op &op, const Wavefront &wf, uint32_t lane_base) {
   static_assert(sizeof(T) == sizeof(uint32_t), "read_simd: T must be a 32-bit lane type");
-  if (const VgprStorage *r = SimdAccess::vgpr_storage(op, wf)) {
-    constexpr auto W = static_cast<uint32_t>(util::native_width_v<T>);
-    SimdAccess::notify_read(op, wf, lane_base, lane_base + W, 0xF);
+  constexpr auto W = static_cast<uint32_t>(util::native_width_v<T>);
+  const uint64_t lane_mask = ((wf.exec() >> lane_base) & util::mask<uint64_t>(static_cast<int>(W)))
```
kuhar, line 503: This PR changes `read_simd`/`read_narrow` from reporting the whole SIMD chunk to reporting `exec & chunk`. That is the right direction, but it changes the race-detector contract and there is no regression test for partial EXEC on the generic SIMD read path.

Please add a small VALU test that runs with partial EXEC and checks that only active lanes are observed/raced. That would lock in the new lane-mask behavior outside the MFMA helper tests.

emulation/rocjitsu/tests/execution_plugin_test.cpp
```diff
@@ -468,6 +517,119 @@ TEST(ExecutionPluginTest, NoPluginNoCrash) {
...
+  for (const HookEvent &e : plugin->events) {
+    if (e.kind != HookEvent::READ_VGPR)
+      continue;
+    ++read_hooks;
+    saw_multi_lane |= std::popcount(e.lane_mask) > 1;
+    EXPECT_NE(e.lane_mask, 0u);
+    EXPECT_NE(e.byte_mask, 0u);
+  }
+  EXPECT_GT(read_hooks, 0u);
+  EXPECT_LT(read_hooks, 32u);
```
kuhar: The helper-level MFMA tests are still loose for the behavior they are meant to pin down. For the current shape, `s0`, `s1`, and `acc` should each produce four full-lane VGPR read hooks: exactly 12 hooks across `[S0, S0+4)`, `[S1, S1+4)`, and `[ACC, ACC+4)` with `byte_mask == 0xF`.

Could we assert that exact set, and add the matching negative case for `const_acc != ACC_FROM_VGPR` so `s2` is not observed when the accumulator comes from the immediate path?

newling:
@kuhar thanks for your review. In light of your comments, I'm going to move this PR into draft mode. I am experimenting with a more comprehensive fix to the slightly leaky/sprawling API we have for reading/writing registers (I have this [draft PR](https://github.com/ROCm/rocm-systems/pull/8231) for example, but I am still seeing how far the agent can go in tightening this). At some point I'll move a draft PR into non-draft mode. This (current) PR does fix a real bug, but it doesn't really fix the API, so I'm trying to do that now.

newling:
Closing in favor of #8417.

#8417 keeps this MFMA read-observation fix, folds in the broader `RegisterAccess` work from #8231, and adds the C++ enforcement layer for the instruction register access boundary.

PR #8080, opened 2026-07-01
fix: [rocr]Raise multicast threshold to 8MB

saleelk:
gfx1250 SDMA multicast is now used for multi-destination copies up to 8MB instead of 256KB. With the updated SDMA firmware the multicast engine no longer drops destinations above 256KB, and measurements show multicast matches or beats the fan-out path up to ~8MB (e.g. ~13% faster at 8MB), while fan-out across multiple SDMA engines still wins above that due to higher aggregate bandwidth.

Replace the boolean HSA_SDMA_FORCE_MULTICAST debug flag with a tri-state HSA_SDMA_MULTICAST override (SDMA_OVERRIDE), matching HSA_SDMA_LINEAR_B2B:
  1 = force multicast at any size
  0 = force off (always fan-out)
  unset = auto (multicast <= kMulticastMaxSize, fan-out above)

Also scope SDMA async-copy profiling to the lifetime of the profiled command. Previously SetSDMAProfiling(true) was latched on the first profiled SDMA command and never cleared, which left the runtime's profiling_enabled() flag on permanently and forced the slower timestamp-capable SDMA packet paths for all subsequent copies. It is now enabled in profilingBegin() and cleared in profilingEnd() per command.

## JIRA ID

JIRA ID : AIRUNTIME-0

## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
@@ -1813,6 +1813,23 @@ hsa_status_t GpuAgent::DmaCopyFanOutOp(
   return coordinator->SubmitEpilogue(out_signal, kWaitValue, body_raw);
 }
 
+// Formats a destination pointer list for SDMA debug logging. Only called from
+// within LogPrint (guarded by the log flag), so it costs nothing when SDMA
+// logging is disabled. Caps the output so a large fan-out cannot flood the log.
+static std::string FormatDstList(void* const* dsts, uint32_t num) {
+  constexpr uint32_t kMaxShown = 16;
+  std::string out = "[";
+  const uint32_t shown = std::min(num, kMaxShown);
+  char buf[32];
+  for (uint32_t i = 0; i < shown; ++i) {
+    snprintf(buf, sizeof(buf), "%s%p", i ? ", " : "", dsts[i]);
+    out += buf;
+  }
+  if (num > kMaxShown) out += ", ...";
+  out += "]";
+  return out;
+}
+
 hsa_status_t GpuAgent::DmaCopyBroadcast(
     const hsa_amd_memory_copy_op_t& op,
     std::vector<core::Signal*>& dep_signals) {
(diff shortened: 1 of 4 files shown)


saleelk:
> 
> 
> * ❌ Unit Test
> 

No test needed

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
cfreeamd: The linearB2B is also using kMulticastMaxSize with its new value. Is that what we want? The comment on line 1889 says 256KB. If we don't want to change it for the non multicast case, maybe we should have 2 different constants.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -1813,6 +1813,23 @@ hsa_status_t GpuAgent::DmaCopyFanOutOp(
   return coordinator->SubmitEpilogue(out_signal, kWaitValue, body_raw);
 }
 
+// Formats a destination pointer list for SDMA debug logging. Only called from
+// within LogPrint (guarded by the log flag), so it costs nothing when SDMA
```
cfreeamd, line 1817: This function isn't free when logging is not on. Even if log print is disabled, the macro args are still evaluated. You could create a local (to this file) wrapper for LogPrint to check if logging is enabled before invoking LogPrint.

projects/rocr-runtime/runtime/hsa-runtime/core/util/flag.h
```diff
@@ -346,10 +346,10 @@ class Flag {
     var = os::GetEnvVar("HSA_SDMA_LINEAR_B2B");
     sdma_linear_b2b_ = (var == "0") ? SDMA_DISABLE : ((var == "1") ? SDMA_ENABLE : SDMA_DEFAULT);
 
-    // HSA_SDMA_FORCE_MULTICAST: 1=force gfx1250 broadcast onto the multicast
-    // packet regardless of copy size (debug only; the multicast engine drops
-    // destinations above 256 KB, so large copies will produce wrong data).
-    sdma_force_multicast_ = (os::GetEnvVar("HSA_SDMA_FORCE_MULTICAST") == "1");
```
cfreeamd, line 352: Claude:
Any CI script or debug workflow using HSA_SDMA_FORCE_MULTICAST=1 silently gets threshold-based (auto) behavior instead of forced-on. No warning is emitted. Additionally, with the old variable gone, the "large copies will fail validation by design" safety caveat disappears — users who set HSA_SDMA_MULTICAST=1 on a kernel with old firmware get silent data corruption at > 256 KB.

Fix: At minimum, read the old name and emit a fprintf(stderr, ...) deprecation warning for one release cycle. Reinstate the "must have updated firmware" caution near the force-enable code path.

chrispaquot (approved):

projects/rocr-runtime/runtime/hsa-runtime/core/util/flag.h
saleelk: I just added that couple days ago, its not out yet , hence changing it

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
saleelk: LogPrint is a function-like macro, not a function:

#define LogPrint(flag, format, ...)                          \
  do {                                                       \
    if (hsa_flag_isset64(log_flags, flag))                   \
      rocr::log_printf(__FILENAME__, __LINE__, format, ##__VA_ARGS__); \
  } while (false);
The variadic args  including FormatDstList are substituted inside the if hsa_flag_isset64 guarded log_printf call, so they're only evaluated when the SDMA log flag is set. When logging is off, FormatDstList is never called and no std::string is constructed, so there's no hot-path cost. This is the same pattern used by the other LogPrint sites in this file. So I dont think we need wrapper

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
saleelk: Ack, I just repurposed it but i should clarify that

chrispaquot (approved):

cfreeamd (approved):

PR #8085, opened 2026-07-01
rocrtst: Print FFM model info on FFM runs

cfreeamd:


projects/rocr-runtime/rocrtst/suites/test_common/main.cc
```diff
@@ -837,6 +842,64 @@ TEST(rocrtstPerf, Agent_Preload_Latency) {
...
+  }
+
+  std::string version_path = ffm_root + "/VERSION";
+  std::ifstream f(version_path);
+  if (!f.is_open())
+    return distro_name.empty() ? "(no VERSION file)" : distro_name + " (no VERSION file)";
+
+  std::map<std::string, std::string> fields;
+  std::string line;
+  while (std::getline(f, line)) {
```
shwetagkhatri, line 885: Here we continue to read whithout checking stream state. 
If the file exists but cannot be read (permissions, I/O error, etc.) the ifstream will open successfully but getline will fail silently. The function would return the distro_name with whatever fields were partially parsed.

Not a blocker - this is diagnostic output only, and partial information is better than crashing. Also the VERSION file is expected to be readable.

shwetagkhatri (approved):
Nothing major except 1 comment. LGTM.

projects/rocr-runtime/rocrtst/suites/test_common/main.cc
```diff
@@ -43,9 +43,14 @@
  *
  */
 
+#include <climits>
+#include <cstdlib>
+#include <fstream>
+#include <map>
```
cfreeamd, line 49: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `<unistd.h>` is likely unused — on Linux `realpath()` is declared in `<cstdlib>` (already included above). Copilot flagged the same. Verify with `grep -n unistd` that nothing else in this file needs it; if not, the include can be dropped.

cfreeamd (commented):

`main.cc` change looks correct — `GetFFMVersionInfo()` path derivation logic and the conditional print are straightforward. One nit inline on an include.

PR #8086, opened 2026-07-01
refactor(hipfile): Use dependency injection for the global state

derobins:
## Motivation

hipFile's driver state was reached through a process-global singleton (Context<DriverState>::get()) touched from ~15 call sites across the C-API implementation, plus batch and async paths. This coupling caused several problems:

- Tests fought the global. Mocking DriverState required a ContextOverride that swapped the global in and out. MDriverState inherited the concrete DriverState, so a test that injected the mock but forgot to EXPECT_CALL a method would silently fall through to real production code instead of failing loudly.
- The backend collection was cached twice — once in DriverState::getBackends() and again in a file-local static in hipfile.cpp (getCachedBackends()).
- The singleton was ambient, so any code below the C-API boundary could reach global state directly, making dependencies implicit and per-test isolation dependent on ctest's process-per-test model.

This PR pushes the singleton to the boundary and injects dependencies everywhere below it.

## Technical Details

- Introduced Runtime (runtime.h/runtime.cpp) as the single process-global root. It owns the one DriverState and the one backend collection, and is the only global touched by the C-API shims in hipfile.cpp. Backend construction (Fallback + Fastpath) now happens exactly once via std::call_once in Runtime::backends() — collapsing the former double cache into a single home.
- Threaded state as an explicit parameter. The internal helpers (hipFileIo, hipFileIOAsync, ensureDriverInit, BatchContext::submit_operations) now take an IDriverState& instead of fetching the global. The C-API shims grab Runtime::instance() once and pass state()/backends() down.
- Extracted an IDriverState interface. DriverState is now final : public IDriverState, and the injection seam depends on the abstraction. MDriverState was reparented onto IDriverState, so the mock has no production body to fall through to — an un-mocked call fails loudly. All 18 methods on the interface are pure virtual and fully mocked.
- Extracted batch helpers. hipFileBatchSetUp/hipFileBatchSubmit (taking IDriverState&) are declared in hipfile-private.h; the hipFileBatchIO* shims are thin hipFileInit() + delegate wrappers, and hipFileIOAsync was made non-static so tests can inject a mock directly.
- Deleted the dead singleton machinery. DriverState::getBackends(), its backends member, the friend Context<DriverState>, the double-cache getCachedBackends(), MDriverState's ContextOverride, and orphaned includes/forward-decls are all gone. The ContextOverride/Context<T> template stays — it's still used by the Hip/Sys/Stats mocks.
- Public C headers are untouched and the C11 boundary is unchanged.
- Test migration: shim-level mock tests now inject the mock directly; real-state integration tests (buffer, handle, fallback, state_mt, system async) read the same object the shims use via Runtime::instance().state(). Added test/amd_detail/runtime.cpp to unit-cover Runtime::backends() and Runtime::state(), which lost coverage when the shims switched to injecting backends.

## JIRA ID

JIRA ID: AIHIPFILE-238

## Test Plan

- [x] ctest -L unit passes (526/526) on a CPU-only node
- [x] Warning-free build with amdclang++ (-Werror posture)
- [x] Clean under ASAN (touches shared_ptr lifetimes and reference threading)
- [x] Clean under TSAN (Runtime::backends() uses std::call_once; AsyncMonitor thread still in play)
- [x] util/format-source.sh (clang-format 18) and clang-tidy clean
- [x] On real hardware: system tests pass, exercising the live async read/write round-trips (test/system/async.cpp) and the real hipFileRead/hipFileWrite → Runtime::backends() → Fallback/Fastpath selection path that CPU-only unit tests cannot cover

## Test Result

Everything passes

## Submission Checklist

### DIFF
--- a/projects/hipfile/src/amd_detail/CMakeLists.txt
+++ b/projects/hipfile/src/amd_detail/CMakeLists.txt
@@ -26,6 +26,7 @@ set(HIPFILE_SOURCES
     hipfile-stats.cpp
     io.cpp
     mountinfo.cpp
+    runtime.cpp
     state.cpp
     stats.cpp
     stream.cpp
--- a/projects/hipfile/src/amd_detail/batch/batch.cpp
+++ b/projects/hipfile/src/amd_detail/batch/batch.cpp
@@ -105,7 +105,7 @@ BatchContext::get_capacity() const noexcept
 }
 
 void
-BatchContext::submit_operations(const hipFileIOParams_t *params, unsigned num_params, DriverState &state)
+BatchContext::submit_operations(const hipFileIOParams_t *params, unsigned num_params, IDriverState &state)
 {
     std::unique_lock<std::shared_mutex> _ulock{context_mutex};
 
--- a/projects/hipfile/src/amd_detail/batch/batch.cpp
+++ b/projects/hipfile/src/amd_detail/batch/batch.cpp
@@ -5,7 +5,6 @@
 
 #include "batch.h"
 #include "buffer.h"
-#include "context.h"
 #include "file.h"
 #include "hipfile.h"
 #include "state.h"
@@ -106,7 +105,7 @@ BatchContext::get_capacity() const noexcept
 }
 
 void
-BatchContext::submit_operations(const hipFileIOParams_t *params, unsigned num_params)
+BatchContext::submit_operations(const hipFileIOParams_t *params, unsigned num_params, DriverState &state)
 {
     std::unique_lock<std::shared_mutex> _ulock{context_mutex};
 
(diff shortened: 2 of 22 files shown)


kurtmcmillan:
Before this can go in we need to have a discussion with the team. Jordan and I looked into dependency injection before we reluctantly settled on the implementation of `Context`. We shouldn't use both dependency injection and `Context`.

One of the challenges with dependency injection is that it is sometimes difficult to inject the dependency when it is needed (e.g. destructors).

kurtmcmillan (changes requested):
This shouldn't go in until we have a plan to fully migrate from `Context` to dependency injection

derobins:
This isn't stale. I was on vacation and now we're waiting for the next release to address this.

PR #8087, opened 2026-07-01
fix(rccl) : [ROCM-26926] [RCCL] Gate P2P batch auto-enable on multi-node for gfx950

prasanna-amd:
## Motivation

Observing Perf Drop of ~5% to ~22% on few RCCL-Perf tags in  DPX + NPS2 mode

## Technical Details

P2P batching was re-enabled by default for gfx950 + non-AINIC in d05a88560d.  The auto-enable unconditionally changes the channel-base computation in ncclP2pChannelBaseForRound (via batchP2PEnableEnv) even when actual batching never fires — batching only activates for nNodes >= 16 with messages below the 64 KB per-rank threshold.

On single-node DPX + NPS2 topologies (16 GPUs, non-uniform XGMI) the altered channel-base mapping distributes P2P traffic sub-optimally across the non-uniform cross-socket/cross-NUMA links, causing a 5-22% bandwidth regression on alltoall, alltoallv, gather, and scatter at 1 GB.  SPX mode is unaffected because its uniform XGMI mesh is insensitive to channel assignment.

Add a multi-node guard to rcclEffectiveP2pBatchEnable so the auto-enable path returns 0 for single-node communicators.  Explicit user overrides (RCCL_P2P_BATCH_ENABLE=0 or =1) are still honoured.
## JIRA ID

JIRA ID : ROCM-26926

## Submission Checklist

### DIFF
--- a/projects/rccl/test/transport/P2pMPITests.cpp
+++ b/projects/rccl/test/transport/P2pMPITests.cpp
@@ -1708,6 +1708,12 @@ TEST_F(P2pMPITest, IpcGraphRegisterBufferTest)
 // (comm->nNodes <= 1), so the channel-base mapping matches pre-7.14 behaviour.
 TEST_F(P2pMPITest, P2pBatchAutoDisableOnSingleNode)
 {
+    if(getenv("RCCL_P2P_BATCH_ENABLE"))
+    {
+        GTEST_SKIP() << "RCCL_P2P_BATCH_ENABLE is explicitly set — "
+                        "cannot test auto-detect path";
+    }
+
     ASSERT_TRUE(validateTestPrerequisites(kMinProcessesForMPI,
                                           kNoProcessLimit,
                                           kRequirePowerOfTwo,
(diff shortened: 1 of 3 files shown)


projects/rccl/test/transport/P2pMPITests.cpp
```diff
@@ -1702,5 +1702,96 @@ TEST_F(P2pMPITest, IpcGraphRegisterBufferTest)
...
+
+    setupP2PBuffers();
+    ASSERT_MPI_SUCCESS(createTestCommunicator());
+
+    // Save and clear RCCL_P2P_BATCH_ENABLE so the auto-detect path (-1) is
+    // exercised.  Restored after the test to avoid polluting subsequent tests.
+    const char*       saved_env = getenv("RCCL_P2P_BATCH_ENABLE");
+    const std::string saved_val = saved_env ? saved_env : "";
+    const bool        had_env   = (saved_env != nullptr);
+    unsetenv("RCCL_P2P_BATCH_ENABLE");
```
Kapil-Shyam-Pawar: I don't think we need to unset / set the env manually
just skipping the test if the env is set should be fine

Kapil-Shyam-Pawar (approved):

prasanna-amd:
Unit test has been created, naming schema differs from  therrockbot check

PR #8089, opened 2026-07-02
fix(hsakmt) : [ROCM-27124] fix lib loading/unloading errors in dxg fo…

manocharahul:
…r wsl

### Issue 
- rocminfo throws errors when running in WSL without LD_LIBRARY_PATH set
  "dlopen libhsa-runtime64.so failed - libhsa-runtime64.so: cannot open shared object file: No such file or directory"
   "[Shutdown] [DxcoreLoader] Cannot unload libdxcore.so: (null)"

### Fix
- since rocminfo already links against libhsa-runtime64.so , dlopen can use RTLD_NOLOAD with exact lib name, to load this lib
- second error is due to incorrect error handling from dlclose.

## Submission Checklist


mangupta (approved):

litvaOo:
Override and merge to address WSL issue

PR #8092, opened 2026-07-02
refactor(hipfile): hold Runtime state behind unique_ptr for in-process test isolation

derobins:
## Motivation

Runtime owned its DriverState by value, so C-API-level tests stayed isolated only because ctest forks a fresh process per test. That made per-test isolation a property of the process model rather than of the fixtures — the test binaries could not be safely run directly, and a test/common/magic-word.* helper existed solely to warn against doing so.

Holding the state behind a unique_ptr lets a test-only Runtime::resetForTesting() rebuild it, making isolation a property of the fixtures. Getting the full binary to pass when run directly (not via ctest) also surfaced three latent sources of process-wide or thread-wide state that had been masked by process-per-test, which this change fixes.

NOTE: Builds on top of #8086. DO NOT MERGE UNTIL 8086 is in.

### Core change
- runtime.h/.cpp: state_ is now std::unique_ptr<DriverState>; state() returns *state_, so the signature is unchanged and no production call site churns. Added an AIS_TESTING-only resetForTesting() that re-seats the pointer. backends() is deliberately not reset (it is stateless configuration, not per-test state).
- hipfile-test.h: HipFileOpened/HipFileUnopened call resetForTesting() at construction and teardown for a guaranteed-clean state per test.
- hipfile-api.cpp: HipFileIoParam resets in its constructor and destructor so the Read and Write parameter instantiations don't share registered files/buffers.
- Retired test/common/magic-word.* (and its references in the CMakeLists and test-common.h) — with in-process isolation the guard is no longer load-bearing.

### Latent state fixes uncovered by direct-run
- configuration.h/.cpp: fastpath(), fallback(), statsLevel(), and unsupportedFileSystems() used function-local statics that cached the first test's mock values for the whole process. Replaced with mutable members plus per-instance std::once_flag.
- hip.h/.cpp: getHipAmdFileReadPtr()/getHipAmdFileWritePtr() used function-local statics with the same problem; removed static so every call routes through Context<Hip>::get(). Cached the resolved proc pointers on the Hip instance (via a resolveProcPtrs() helper guarded by std::once_flag) so hipAmdFileRead/Write resolve the symbol once per instance rather than on every I/O.
- fastpath.h/.cpp: the thread_local bool hip_inited guard persisted across tests sharing a thread. Replaced with a per-instance uint64_t instance_id_ (from a monotonic counter) compared against a thread_local inited_for_instance. This preserves the per-thread production invariant — every new thread calls hipInit() on its first fastpath I/O — while a fresh Fastpath forces re-init, giving per-test isolation.

## JIRA ID

JIRA ID: AIHIPFILE-238

## Test Plan

- [x] Warning-free build with amdclang++; both AIS_TESTING and non-AIS_TESTING configs compile (the reset seam is #ifdef-guarded).
- [x] ctest -L unit: 530/530 passing.
- [x] Added FastpathHipInitThreading tests (parameterized over Read/Write) that drive io() from two std::threads behind a std::latch start barrier, verifying both the per-thread hipInit() invariant and the per-instance reset. These fail against a naive atomic<bool> guard and pass with the instance_id_ approach.
- [x] ASAN build + ctest -L unit clean — no use-after-free from the reset seam.
- [x] TSAN build to validate the threading guard, plus GPU-dependent system/stress tests, run manually on a GPU host.

## Test Result

Success

## Submission Checklist


derobins:
This isn't stale. I was on vacation and now we're waiting for the next release to address this.

PR #8093, opened 2026-07-02
[rocprofiler-sdk] Bypass idle inline queue interposition without active consumers

bwelton:
## Summary

Fix ROCM-27116 by bypassing queue interposition (QI) on the hot path when no active profiling context needs kernel-dispatch or scratch-memory tracing. This prevents idle QI from virtualizing write indices and processing doorbells when auto-registered clients only use code-object callbacks.

## Changes

- `queue_interposition.cpp` / `queue_interposition.hpp` — track active queue-interposition consumers with an atomic counter; extend `should_bypass_inline_intercept()` to passthrough when the count is zero
- `queue_controller.cpp` / `queue_controller.hpp` — add `context_needs_queue_interposition_tracing()` helper
- `context.cpp` — increment/decrement the counter on context start/stop and on `deactivate_client_contexts()`

Also bundled in this branch (surfaced during review/testing of the above):
- Saturating CAS-loop decrement for the consumer counter (avoids underflow on unbalanced stop notifications)
- Resync of `virtual_wptr`/`next_scan_pos`/`next_submit_pos` to the real write-dispatch-id when the consumer count transitions 0→1, so QI doesn't publish against stale bookkeeping after bypass doorbells advanced the real queue
- Ordering fix for context start/stop notifications relative to active-context publish, and CAS-based slot claiming in `deactivate_client_contexts()`, to close a consumer-count race
- Replace the persistent `thread_local` doorbell snapshot buffer with a fixed-capacity inline snapshot (falling back to a local vector for rare large batches), removing a heap buffer implicated in the ROCM-27116 crash

## Motivation

PyTorch auto-registration installs QI by default even when only code-object tracing is registered. In a fork+exec child that inherits `ROCPROFILER_REGISTER_LIBRARY`, idle QI still intercepts HSA queue write-index and doorbell operations during minimal GPU work, corrupting the heap before exit (`corrupted double-linked list` / SIGABRT). Setting `ROCPROFILER_QUEUE_INTERPOSITION=0` avoids the crash; this change makes that behavior automatic when no QI consumer is active.

## Testing

- Built `librocprofiler-sdk.so` on banff (MI325X) and deployed into `rocm27116-parallel` PyTorch container
- PyTorch fork+exec repro (`torch.cuda.synchronize()` → `subprocess` child with `env=None` → `torch.rand`) **15/15 pass**
- Stock image reproduces ~90%+ crash rate without the fix
- Full local build + `ctest` (gfx1200): 378/379 unit tests pass (1 pre-existing, unrelated failure: `unit.rocprofiler_lib.agent`); all queue-interposition/queue-controller/context/intercept-table tests and HSA multiqueue integration tests pass

## Related Issues

JIRA ID : ROCM-27116

### DIFF
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
@@ -569,13 +569,13 @@ enable_queue_intercept()
 }
 
 bool
-context_needs_inline_qi_tracing(const context::context* ctx)
+context_needs_queue_interposition_tracing(const context::context* ctx)
 {
     if(!ctx) return false;
-    return ctx->is_tracing(ROCPROFILER_CALLBACK_TRACING_KERNEL_DISPATCH) ||
-           ctx->is_tracing(ROCPROFILER_BUFFER_TRACING_KERNEL_DISPATCH) ||
-           ctx->is_tracing(ROCPROFILER_CALLBACK_TRACING_SCRATCH_MEMORY) ||
-           ctx->is_tracing(ROCPROFILER_BUFFER_TRACING_SCRATCH_MEMORY);
+    return ctx->is_tracing_one_of(ROCPROFILER_CALLBACK_TRACING_KERNEL_DISPATCH,
+                                  ROCPROFILER_BUFFER_TRACING_KERNEL_DISPATCH,
+                                  ROCPROFILER_CALLBACK_TRACING_SCRATCH_MEMORY,
+                                  ROCPROFILER_BUFFER_TRACING_SCRATCH_MEMORY);
 }
 
 void
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
@@ -568,6 +568,16 @@ enable_queue_intercept()
     return false;
 }
 
+bool
+context_needs_inline_qi_tracing(const context::context* ctx)
+{
+    if(!ctx) return false;
+    return ctx->is_tracing(ROCPROFILER_CALLBACK_TRACING_KERNEL_DISPATCH) ||
+           ctx->is_tracing(ROCPROFILER_BUFFER_TRACING_KERNEL_DISPATCH) ||
+           ctx->is_tracing(ROCPROFILER_CALLBACK_TRACING_SCRATCH_MEMORY) ||
+           ctx->is_tracing(ROCPROFILER_BUFFER_TRACING_SCRATCH_MEMORY);
+}
+
 void
 queue_controller_init(HsaApiTable* table)
 {
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
@@ -571,10 +571,10 @@ enable_queue_intercept()
 bool
 context_needs_queue_interposition_tracing(const context::context* ctx)
 {
-    return ctx && ctx->is_tracing_one_of(ROCPROFILER_CALLBACK_TRACING_KERNEL_DISPATCH,
-                                         ROCPROFILER_BUFFER_TRACING_KERNEL_DISPATCH,
-                                         ROCPROFILER_CALLBACK_TRACING_SCRATCH_MEMORY,
-                                         ROCPROFILER_BUFFER_TRACING_SCRATCH_MEMORY);
+    return ctx != nullptr && ctx->is_tracing_one_of(ROCPROFILER_CALLBACK_TRACING_KERNEL_DISPATCH,
+                                                    ROCPROFILER_BUFFER_TRACING_KERNEL_DISPATCH,
+                                                    ROCPROFILER_CALLBACK_TRACING_SCRATCH_MEMORY,
+                                                    ROCPROFILER_BUFFER_TRACING_SCRATCH_MEMORY);
 }
 
 void
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
@@ -571,11 +571,10 @@ enable_queue_intercept()
 bool
 context_needs_queue_interposition_tracing(const context::context* ctx)
 {
-    if(!ctx) return false;
-    return ctx->is_tracing_one_of(ROCPROFILER_CALLBACK_TRACING_KERNEL_DISPATCH,
-                                  ROCPROFILER_BUFFER_TRACING_KERNEL_DISPATCH,
-                                  ROCPROFILER_CALLBACK_TRACING_SCRATCH_MEMORY,
-                                  ROCPROFILER_BUFFER_TRACING_SCRATCH_MEMORY);
+    return ctx && ctx->is_tracing_one_of(ROCPROFILER_CALLBACK_TRACING_KERNEL_DISPATCH,
+                                         ROCPROFILER_BUFFER_TRACING_KERNEL_DISPATCH,
+                                         ROCPROFILER_CALLBACK_TRACING_SCRATCH_MEMORY,
+                                         ROCPROFILER_BUFFER_TRACING_SCRATCH_MEMORY);
 }
 
 void
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_interposition.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_interposition.cpp
@@ -71,7 +71,7 @@ namespace queue_interposition
 {
 namespace
 {
-auto s_active_inline_qi_consumers = std::atomic<uint32_t>{0};
+auto s_active_queue_interposition_consumers = std::atomic<uint32_t>{0};
 
 // NOTE:
 //  - "installed" is for checking whether HSA functions have been passed
@@ -83,9 +83,9 @@ auto s_intercept_active    = std::atomic<bool>{false};  // actively intercepting
 auto s_intercept_dynamic   = std::atomic<bool>{false};  // dynamically add queue states
 
 bool
-has_active_inline_qi_consumers()
+has_active_queue_interposition_consumers()
 {
-    return s_active_inline_qi_consumers.load(std::memory_order_acquire) > 0;
+    return s_active_queue_interposition_consumers.load(std::memory_order_relaxed) > 0;
 }
 
 bool
@@ -95,7 +95,7 @@ should_bypass_inline_intercept()
             !s_intercept_active.load(std::memory_order_acquire) ||
             registration::get_fini_status() != 0 ||
             // TODO: debug and enable queue interposition for attachment
-            registration::supports_attachment() || !has_active_inline_qi_consumers());
+            registration::supports_attachment() || !has_active_queue_interposition_consumers());
 }
 
 auto*&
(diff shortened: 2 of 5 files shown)


projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
```diff
@@ -568,6 +568,16 @@ enable_queue_intercept()
     return false;
 }
 
+bool
+context_needs_inline_qi_tracing(const context::context* ctx)
```
jrmadsen: Naming-wise, let's just use "queue_interposition" instead of "inline_qi".

```suggestion
context_needs_queue_interposition_tracing(const context::context* ctx)
```

Furthermore, we need to centralize the logic for which services use queue interposition. Is this an attempt to do so?

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
```diff
@@ -568,6 +568,16 @@ enable_queue_intercept()
...
+{
+    if(!ctx) return false;
+    return ctx->is_tracing(ROCPROFILER_CALLBACK_TRACING_KERNEL_DISPATCH) ||
+           ctx->is_tracing(ROCPROFILER_BUFFER_TRACING_KERNEL_DISPATCH) ||
+           ctx->is_tracing(ROCPROFILER_CALLBACK_TRACING_SCRATCH_MEMORY) ||
+           ctx->is_tracing(ROCPROFILER_BUFFER_TRACING_SCRATCH_MEMORY);
```
jrmadsen: I think we have a variadic template for this. Something like:

```suggestion
    return ctx->is_tracing(ROCPROFILER_CALLBACK_TRACING_KERNEL_DISPATCH, 
                           ROCPROFILER_BUFFER_TRACING_KERNEL_DISPATCH, 
                           ROCPROFILER_CALLBACK_TRACING_SCRATCH_MEMORY, 
                           ROCPROFILER_BUFFER_TRACING_SCRATCH_MEMORY);
```

or maybe it is `is_tracing_one_of(...)`

jrmadsen (changes requested):

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
bwelton: Done — renamed to `context_needs_queue_interposition_tracing` in ec0b6ca237, and consistently renamed the related "inline_qi" identifiers (`s_active_inline_qi_consumers`, `has_active_inline_qi_consumers`, `notify_inline_qi_consumer_context_started/stopped`) to use "queue_interposition" naming as well for consistency. Left the unrelated `qi_dispatch_trace` debug-tracing facility untouched since it's a separate, pre-existing tool.

Yes, this function is exactly meant to centralize that logic — it's the single predicate now used both to gate the consumer-counter increment/decrement (`notify_queue_interposition_consumer_context_started/stopped`) and to decide which contexts count as active queue-interposition consumers for the bypass gate.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
bwelton: Done — switched to `ctx->is_tracing_one_of(ROCPROFILER_CALLBACK_TRACING_KERNEL_DISPATCH, ROCPROFILER_BUFFER_TRACING_KERNEL_DISPATCH, ROCPROFILER_CALLBACK_TRACING_SCRATCH_MEMORY, ROCPROFILER_BUFFER_TRACING_SCRATCH_MEMORY)` in ec0b6ca237. Confirmed it compiles fine mixing the callback- and buffer-tracing enum kinds in the same variadic call since `is_tracing_one_of`'s `Args...` pack deduces each argument's type independently and dispatches to the templated `is_tracing<KindT>` overload per-arg (there's existing precedent for this same heterogeneous-kind pattern elsewhere in the codebase, e.g. queue_interposition.cpp).

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
```diff
@@ -568,6 +568,16 @@ enable_queue_intercept()
     return false;
 }
 
+bool
+context_needs_queue_interposition_tracing(const context::context* ctx)
+{
+    if(!ctx) return false;
+    return ctx->is_tracing_one_of(ROCPROFILER_CALLBACK_TRACING_KERNEL_DISPATCH,
```
jrmadsen: Nitpick:

```suggestion
    return ctx && ctx->is_tracing_one_of(ROCPROFILER_CALLBACK_TRACING_KERNEL_DISPATCH,
```

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_interposition.cpp
```diff
@@ -105,6 +116,80 @@ get_next_table()
 }
 }  // namespace
 
+namespace qi_dispatch_trace
+{
+inline bool
+enabled()
+{
+    static const auto _enabled = common::get_env("ROCPROFILER_QI_DISPATCH_TRACE", 0) != 0;
```
jrmadsen:
```suggestion
    static const auto _enabled = common::get_env("ROCPROFILER_QUEUE_INTERPOSITION_DISPATCH_TRACE", false);
```
The problem with "QI" as an abbreviation is that it could also stand for "Queue Interception".

Also, `common::get_env` handles all the common representations of specifying a boolean (0/1/y/n/true/false).

Additional note... I don't love the idea of this solely being controlled via env variable. It doesn't need to be addressed in this PR necessarily but we may need a way to query/set this programmatically

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_interposition.cpp
```diff
@@ -105,6 +116,80 @@ get_next_table()
 }
 }  // namespace
 
+namespace qi_dispatch_trace
```
jrmadsen: Why is this in a namespace? And even if a namespace is desirable here, it should still be within an anonymous namespace since it isn't exposed outside this translation unit.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_interposition.cpp
```diff
@@ -105,6 +116,80 @@ get_next_table()
...
+    indices.next_submit_pos = state->next_submit_pos;
+    if(state->real_wdid) indices.real_wdid = __atomic_load_n(state->real_wdid, __ATOMIC_ACQUIRE);
+    if(state->real_rdid) indices.real_rdid = __atomic_load_n(state->real_rdid, __ATOMIC_ACQUIRE);
+    indices.bypass    = should_bypass_inline_intercept();
+    indices.consumers = s_active_queue_interposition_consumers.load(std::memory_order_acquire);
+    return indices;
+}
+
+template <typename... Args>
+void
```
jrmadsen: I think this should return a string  so things using it for logging can control the log level (i.e. trace instead of info) and the line attribution of the log message isn't always this line (i.e. log message at the line it was logged at).

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_interposition.cpp
```diff
@@ -222,8 +308,21 @@ publish_submitted_packets(QueueState* state, uint64_t submit_pos)
...
 
     __atomic_store_n(state->real_wdid, submit_pos, __ATOMIC_RELEASE);
-    (*tls.ring_doorbell)(state->doorbell_signal, static_cast<hsa_signal_value_t>(submit_pos - 1));
+    const auto doorbell_idx = static_cast<hsa_signal_value_t>(submit_pos - 1);
+    (*tls.ring_doorbell)(state->doorbell_signal, doorbell_idx);
     tls.last_published_submit_pos = submit_pos;
+
+    if(qi_dispatch_trace::enabled())
+    {
+        qi_dispatch_trace::log("publish",
```
jrmadsen: If log returns a string, this can be `ROCP_LOG_IF(enabled()) << log("publish", ...)`

jrmadsen (changes requested):

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_interposition.cpp
```diff
@@ -1048,6 +1056,21 @@ supports_queue_interposition()
...
+    s_active_inline_qi_consumers.fetch_add(1, std::memory_order_release);
+}
+
+void
+notify_inline_qi_consumer_context_stopped(const context::context* ctx)
+{
+    if(!context_needs_inline_qi_tracing(ctx)) return;
+    auto prev = s_active_inline_qi_consumers.fetch_sub(1, std::memory_order_release);
+    if(prev == 0) s_active_inline_qi_consumers.store(0, std::memory_order_release);
+}
```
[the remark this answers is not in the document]
bwelton: Fixed in 8f2c7e9b5e — switched to a saturating CAS-loop decrement so the counter never wraps.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_interposition.cpp
```diff
@@ -77,14 +79,20 @@ auto s_intercept_installed = std::atomic<bool>{false};  // installed (may not be
 auto s_intercept_active    = std::atomic<bool>{false};  // actively intercepting
 auto s_intercept_dynamic   = std::atomic<bool>{false};  // dynamically add queue states
 
+bool
+has_active_inline_qi_consumers()
+{
+    return s_active_inline_qi_consumers.load(std::memory_order_acquire) > 0;
+}
```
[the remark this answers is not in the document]
bwelton: Fixed in ec0b6ca237 — switched to `memory_order_relaxed` in `has_active_queue_interposition_consumers()`.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_interposition.cpp
```diff
@@ -66,6 +66,8 @@ namespace hsa
 {
 namespace queue_interposition
 {
+auto s_active_inline_qi_consumers = std::atomic<uint32_t>{0};
+
```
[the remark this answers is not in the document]
bwelton: Fixed in 8f2c7e9b5e — moved into the anonymous namespace.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.cpp
bwelton: Done — applied in 86222cff76.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_interposition.cpp
bwelton: Good points all around on naming, scoping, and the log-as-string idea. Rather than rework it, removed the whole `qi_dispatch_trace` facility in 7edea8bcca — it was ad-hoc debug instrumentation added while bisecting the ROCM-27116 crash and was never meant to ship as part of this feature. Happy to revisit a proper env-var/programmatic-toggle diagnostic facility as a separate follow-up if there's a real need for it.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_interposition.cpp
bwelton: Removed entirely in 7edea8bcca — this was ad-hoc debug instrumentation added while bisecting the ROCM-27116 crash and shouldn't have shipped in the PR.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_interposition.cpp
bwelton: Removed entirely in 7edea8bcca (see above) rather than reworking the return type.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_interposition.cpp
bwelton: Removed entirely in 7edea8bcca (see above).

jrmadsen (approved):

venkat1361 (approved):

marbre:
Overriding as test failures are unrelated.

PR #8094, opened 2026-07-02
rocr: Fix out-of-bounds access in hsa_amd_signal_wait_any

bwelton:
## Summary
Fixes an out-of-bounds read/write in `hsa_amd_signal_wait_any` that occurs whenever the satisfying signal in a multi-signal batch is at any index other than 0.

## Root Cause
`hsa_amd_signal_wait_any` allocated `satisfying_value_vec` with a fixed size of `1`:

```cpp
std::vector<hsa_signal_value_t> satisfying_value_vec(1);
uint32_t satisfying_signal_idx =
    core::Signal::WaitMultiple(valid_signals.size(), valid_signals.data(), conds, values,
                               timeout_hint, wait_hint, satisfying_value_vec, false);
...
if (satisfying_value) *satisfying_value = satisfying_value_vec.at(0);
```

But `core::Signal::WaitMultiple()` (called with `wait_on_all=false`, i.e. "wait for any") writes `satisfying_values[i] = value` and returns `i` directly, where `i` can be any index in `[0, valid_signals.size())` — whichever signal is satisfied first. Whenever `i != 0`, `WaitMultiple()` writes out of bounds into the 1-element vector, and the subsequent `.at(0)` read (even if the OOB write somehow didn't corrupt anything) returns the wrong value regardless of which signal actually satisfied.

This also missed handling `WaitMultiple()`'s `uint32_t(-1)` timeout/invalid-signal sentinel return, which would separately index `valid_signal_ids` out of bounds when mapping the result back to the caller's original array position.

## Fix
- Size `satisfying_value_vec` to `valid_signals.size()` (matching `WaitMultiple()`'s local index space), not a fixed `1`.
- Return early on the `uint32_t(-1)` sentinel before doing any indexing.
- Read the satisfying value using `WaitMultiple()`'s *local* index into `satisfying_value_vec`, then map the *index returned to the caller* through `valid_signal_ids` back to the caller's original array position — these two index spaces are different and must not be conflated (the fix separates them explicitly with comments).

Sanity-checked `hsa_amd_signal_wait_all`'s analogous code path; it already sizes its output vector correctly (`valid_signal_count`), so no equivalent fix was needed there.

## How this was found
Discovered while prototyping a batched-signal-wait completion mechanism for rocprofiler-sdk's Queue Interposition feature (tracked separately, ROCM-27290 investigation), which calls `hsa_amd_signal_wait_any` with a multi-signal batch where the satisfying index is very commonly greater than 0 (a dedicated "wake" signal is always appended at the end of the batch). Any existing caller of `hsa_amd_signal_wait_any` with more than one signal in its batch is exposed to this bug today, independent of that follow-on work.

## Testing
- Manual code-review verification of the corrected index-space separation (implementer + independent reviewer pass).
- Exercised indirectly via a downstream rocprofiler-sdk prototype that calls `hsa_amd_signal_wait_any` with multi-signal batches where the satisfying index is >0 in the common case; before this fix, that prototype produced incorrect/garbage wake-signal values, consistent with this bug. After the fix: correct behavior confirmed across `ctest` runs and repeated stress runs.
- No dedicated unit test for `hsa_amd_signal_wait_any` with a non-zero satisfying index currently exists in this repo's test suite; recommend adding one as a follow-up (happy to add if requested).

## Related
Related to ROCM-27290 (rocprofiler-sdk Queue Interposition performance investigation) but is a standalone, independently-valid bug fix.

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa_ext_amd.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa_ext_amd.cpp
@@ -1079,10 +1079,16 @@ uint32_t hsa_amd_signal_wait_all(uint32_t signal_count, hsa_signal_t* hsa_signal
   uint32_t valid_signal_count = valid_signals.size();
 
   std::vector<hsa_signal_value_t> satisfying_values_vec(valid_signal_count);
-  uint32_t first_satysifying_signal_idx =
+  uint32_t first_satisfying_signal_idx =
       core::Signal::WaitMultiple(valid_signal_count, valid_signals.data(), valid_conds.data(),
                                  valid_values.data(), timeout_hint, wait_hint, satisfying_values_vec, true);
 
+  // Note: on timeout (or if a signal became invalid mid-wait), WaitMultiple() returns
+  // uint32_t(-1) and satisfying_values_vec is only partially filled -- entries for
+  // signals whose condition was never met remain at their zero-initialized value and do
+  // not represent a real satisfying value. satisfying_values is still populated below in
+  // that case; callers must check the return value before treating its contents as
+  // meaningful.
   if (satisfying_values) {
     // Set 0 as satisfying value for NULL and invalid signals
     std::vector<hsa_signal_value_t> satisfying_values_vec_result(signal_count, 0);
@@ -1092,7 +1098,7 @@ uint32_t hsa_amd_signal_wait_all(uint32_t signal_count, hsa_signal_t* hsa_signal
     std::copy(satisfying_values_vec_result.begin(), satisfying_values_vec_result.end(), satisfying_values);
   }
 
-  return first_satysifying_signal_idx;
+  return first_satisfying_signal_idx;
   CATCHRET(uint32_t);
 }
 
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa_ext_amd.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa_ext_amd.cpp
@@ -1117,14 +1117,21 @@ uint32_t hsa_amd_signal_wait_any(uint32_t signal_count, hsa_signal_t* hsa_signal
     return std::numeric_limits<uint32_t>::max();
   }
 
-  std::vector<hsa_signal_value_t> satisfying_value_vec(1);
-  uint32_t satisfying_signal_idx =
+  // Sized to valid_signals.size() since WaitMultiple() may return/write any index in
+  // [0, valid_signals.size()), not just 0.
+  std::vector<hsa_signal_value_t> satisfying_value_vec(valid_signals.size());
+  uint32_t local_satisfying_signal_idx =
       core::Signal::WaitMultiple(valid_signals.size(), valid_signals.data(), conds, values, timeout_hint, wait_hint,
                                  satisfying_value_vec, false);
-  //  Map back the index
-  satisfying_signal_idx = valid_signal_ids[satisfying_signal_idx];
 
-  if (satisfying_value) *satisfying_value = satisfying_value_vec.at(0);
+  if (local_satisfying_signal_idx == uint32_t(-1)) {
+    return local_satisfying_signal_idx;
+  }
+
+  if (satisfying_value) *satisfying_value = satisfying_value_vec.at(local_satisfying_signal_idx);
+
+  // Map from WaitMultiple()'s local index back to the caller's original signal array index.
+  uint32_t satisfying_signal_idx = valid_signal_ids[local_satisfying_signal_idx];
 
   return satisfying_signal_idx;
   CATCHRET(uint32_t);
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa_ext_amd.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa_ext_amd.cpp
@@ -1056,10 +1056,14 @@ uint32_t hsa_amd_signal_wait_all(uint32_t signal_count, hsa_signal_t* hsa_signal
   // Treat NULL and invalid signals as already satisfied their condition and skip them
   std::vector<hsa_signal_t> valid_signals;
   std::vector<uint32_t> valid_signal_ids;
+  std::vector<hsa_signal_condition_t> valid_conds;
+  std::vector<hsa_signal_value_t> valid_values;
   for (uint32_t i = 0; i < signal_count; i++){
     if (hsa_signals[i].handle != 0 && core::SharedSignal::Convert(hsa_signals[i])->IsValid()){
       valid_signals.emplace_back(hsa_signals[i]);
       valid_signal_ids.emplace_back(i);
+      valid_conds.emplace_back(conds[i]);
+      valid_values.emplace_back(values[i]);
     }
   }
 
@@ -1076,8 +1080,8 @@ uint32_t hsa_amd_signal_wait_all(uint32_t signal_count, hsa_signal_t* hsa_signal
 
   std::vector<hsa_signal_value_t> satisfying_values_vec(valid_signal_count);
   uint32_t first_satysifying_signal_idx =
-      core::Signal::WaitMultiple(valid_signal_count, valid_signals.data(), conds, values, timeout_hint, wait_hint,
-                                 satisfying_values_vec, true);
+      core::Signal::WaitMultiple(valid_signal_count, valid_signals.data(), valid_conds.data(),
+                                 valid_values.data(), timeout_hint, wait_hint, satisfying_values_vec, true);
 
   if (satisfying_values) {
     // Set 0 as satisfying value for NULL and invalid signals
@@ -1104,10 +1108,14 @@ uint32_t hsa_amd_signal_wait_any(uint32_t signal_count, hsa_signal_t* hsa_signal
   // Ignore NULL and invalid signals
   std::vector<hsa_signal_t> valid_signals;
   std::vector<uint32_t> valid_signal_ids;
+  std::vector<hsa_signal_condition_t> valid_conds;
+  std::vector<hsa_signal_value_t> valid_values;
   for (uint32_t i = 0; i < signal_count; i++){
     if (hsa_signals[i].handle != 0 && core::SharedSignal::Convert(hsa_signals[i])->IsValid()){
       valid_signals.emplace_back(hsa_signals[i]);
       valid_signal_ids.emplace_back(i);
+      valid_conds.emplace_back(conds[i]);
+      valid_values.emplace_back(values[i]);
     }
   }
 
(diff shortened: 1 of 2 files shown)


bwelton:
Trimmed the inline comments down to the essentials (kept the two non-obvious points: sizing rationale and the local→original index mapping), removed the more verbose/redundant explanatory prose. No logic changes.

bwelton:
Addressed two additional issues found by Copilot's review (both verified as real, pre-existing bugs in this same area, independent of my confirmation): 

1. `conds`/`values` were not compacted alongside `valid_signals` in either `hsa_amd_signal_wait_any` or `hsa_amd_signal_wait_all` — fixed by compacting both arrays in lockstep.
2. `core::Signal::WaitMultiple()`'s wait-any branch was writing `satisfying_values[i]` instead of `satisfying_values[0]`, violating its own documented contract in `signal.h` and causing a separate real OOB at `runtime.cpp`'s `AsyncEventsLoop` caller. Fixed by honoring the documented slot-0 contract; simplified `hsa_amd_signal_wait_any()` back to a 1-element vector as a consequence. No change needed in `runtime.cpp` — it already assumed the correct contract.

Went through a code-review pass (no blockers/majors found) before landing. Rebuilt + retested rocr-runtime and the downstream rocprofiler-sdk batched-wait consumer (#8095):
- `ctest -R queue_interposition`: 14/14 passed
- Phase 1 sanity benchmark: unchanged (~1.33x vs QI-disabled), as expected
- 10/10 repeated shutdown-stress runs clean — this specifically re-exercises the non-zero-satisfying-index scenario bug 2 was about (a dedicated wake signal is always appended at the end of the wait batch in #8095's design)

No existing rocr-runtime test coverage was found for either bug case; flagged as a worthwhile follow-up (targeted regression tests) but kept out of scope for this fix to stay focused.

cfreeamd (commented):
signal.cpp and hsa_ext_amd.cpp both look correct for the stated fixes; see inline comments for one pre-existing wait_all timeout behavior worth noting and a few optional nits.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/signal.cpp
```diff
@@ -297,10 +297,13 @@ uint32_t Signal::WaitMultiple(uint32_t signal_count, const hsa_signal_t* hsa_sig
...
         it = unmet_condition_ids.erase(it);
-        satisfying_values[i] = value;
-        if (!wait_on_all)
+        if (!wait_on_all) {
+          // wait-any callers only need satisfying_values[0] per WaitMultiple()'s contract.
+          satisfying_values[0] = value;
           return i;
-        else if (unmet_condition_ids.empty())
+        }
+        satisfying_values[i] = value;
```
cfreeamd, line 305: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]**

In the `wait_on_all=true` branch, `satisfying_values[i]` is written before checking `unmet_condition_ids.empty()`. The variable `i` here is the stable value stored in `unmet_condition_ids` (initialized 0..N-1 via `std::iota`), not a position index that shifts on erase — so array bounds are safe. Stating this for reviewer confidence: is there any case where an entry is erased from `unmet_condition_ids` mid-iteration in a way that could cause index reuse for a different signal?

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa_ext_amd.cpp
```diff
@@ -1076,8 +1080,8 @@ uint32_t hsa_amd_signal_wait_all(uint32_t signal_count, hsa_signal_t* hsa_signal
 
   std::vector<hsa_signal_value_t> satisfying_values_vec(valid_signal_count);
   uint32_t first_satysifying_signal_idx =
-      core::Signal::WaitMultiple(valid_signal_count, valid_signals.data(), conds, values, timeout_hint, wait_hint,
-                                 satisfying_values_vec, true);
+      core::Signal::WaitMultiple(valid_signal_count, valid_signals.data(), valid_conds.data(),
+                                 valid_values.data(), timeout_hint, wait_hint, satisfying_values_vec, true);
 
   if (satisfying_values) {
```
cfreeamd, line 1092: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]**

`hsa_amd_signal_wait_all` does not guard the `uint32_t(-1)` timeout return from `WaitMultiple` before copying `satisfying_values_vec` into the caller's output array. On timeout the vector is partially filled (zero-initialized for unsatisfied signals) and the copy still runs. This is pre-existing and arguably acceptable (caller can check return value), but the analogous timeout guard added for `wait_any` in this PR makes the contrast jarring. A comment noting that on timeout `satisfying_values` contains partial/meaningless results would help future readers.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa_ext_amd.cpp
```diff
@@ -1076,8 +1080,8 @@ uint32_t hsa_amd_signal_wait_all(uint32_t signal_count, hsa_signal_t* hsa_signal
 
   std::vector<hsa_signal_value_t> satisfying_values_vec(valid_signal_count);
   uint32_t first_satysifying_signal_idx =
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]**

Pre-existing typo: `first_satysifying_signal_idx` ("satysifying"). Not introduced by this PR, but since this function is already being touched, worth fixing opportunistically.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa_ext_amd.cpp
```diff
@@ -1117,15 +1125,25 @@ uint32_t hsa_amd_signal_wait_any(uint32_t signal_count, hsa_signal_t* hsa_signal
     return std::numeric_limits<uint32_t>::max();
   }
 
+  // For wait-any, WaitMultiple() only ever writes the satisfying value to slot 0.
```
cfreeamd, line 1134: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]**

The comment "For wait-any, WaitMultiple() only ever writes the satisfying value to slot 0" is now true *because of* the fix to `signal.cpp` in this same PR. Consider rephrasing as a forward contract assertion rather than a behavioral observation — e.g., "WaitMultiple() contract: in wait-any mode, only satisfying_values[0] is written (see signal.cpp WaitMultiple)." This makes it clear the comment documents an invariant, not just coincidence.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa_ext_amd.cpp
```diff
@@ -1117,14 +1117,21 @@ uint32_t hsa_amd_signal_wait_any(uint32_t signal_count, hsa_signal_t* hsa_signal
...
   }
 
-  std::vector<hsa_signal_value_t> satisfying_value_vec(1);
-  uint32_t satisfying_signal_idx =
+  // Sized to valid_signals.size() since WaitMultiple() may return/write any index in
+  // [0, valid_signals.size()), not just 0.
+  std::vector<hsa_signal_value_t> satisfying_value_vec(valid_signals.size());
+  uint32_t local_satisfying_signal_idx =
       core::Signal::WaitMultiple(valid_signals.size(), valid_signals.data(), conds, values, timeout_hint, wait_hint,
                                  satisfying_value_vec, false);
```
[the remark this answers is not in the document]
bwelton: This looks already addressed by commit 3e2d84265e — that commit compacts `conds`/`values` alongside `valid_signals` in both `hsa_amd_signal_wait_all()` and `hsa_amd_signal_wait_any()`, passing the compacted `valid_conds`/`valid_values` to `WaitMultiple()` instead of the raw unfiltered arrays. Let me know if I'm missing something.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa_ext_amd.cpp
```diff
@@ -1117,14 +1117,21 @@ uint32_t hsa_amd_signal_wait_any(uint32_t signal_count, hsa_signal_t* hsa_signal
     return std::numeric_limits<uint32_t>::max();
   }
 
-  std::vector<hsa_signal_value_t> satisfying_value_vec(1);
-  uint32_t satisfying_signal_idx =
+  // Sized to valid_signals.size() since WaitMultiple() may return/write any index in
+  // [0, valid_signals.size()), not just 0.
+  std::vector<hsa_signal_value_t> satisfying_value_vec(valid_signals.size());
+  uint32_t local_satisfying_signal_idx =
       core::Signal::WaitMultiple(valid_signals.size(), valid_signals.data(), conds, values, timeout_hint, wait_hint,
```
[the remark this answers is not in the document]
bwelton: This looks already addressed by commit 3e2d84265e — `Signal::WaitMultiple()`'s wait-any branch now always writes `satisfying_values[0]` (never `satisfying_values[i]`) while still returning the true local index, so the 1-element vector at this call site and the `runtime.cpp` one you flagged are both safe regardless of which signal actually satisfies. Let me know if I'm missing something.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/signal.cpp
bwelton: Confirmed — there's no case where an index gets reused for a different signal. `unmet_condition_ids` is populated exactly once via `std::iota(..., 0)` right before the wait loop, giving a fixed set of distinct values, one per signal, that persists across all iterations. Entries are only ever removed via `erase()` when a condition is satisfied — nothing re-inserts or resets the container, so once index `i` is erased it can never reappear against a different signal later. `i` is a stable signal identifier, not a position that shifts meaning after erase, so `satisfying_values[i]` is always a valid, correctly-targeted write. No bug here.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa_ext_amd.cpp
bwelton: Fixed in e49b0841: added a comment noting that on timeout, `satisfying_values` is only partially filled (zero for signals whose condition was never met) and callers must check the return value before treating its contents as meaningful — mirrors the guard/comment already added for wait_any in this PR.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa_ext_amd.cpp
bwelton: Fixed in e49b0841: renamed `first_satysifying_signal_idx` to `first_satisfying_signal_idx`.

cfreeamd (approved):

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/signal.cpp
cfreeamd: thanks

bwelton:
Closing in favor of #8564 — same 3 commits (rebased to include a new regression test, squashed into the corresponding fix commits) from a correctly-named branch (`users/bewelton/fix-hsa-signal-wait-any-oob`), with a Conventional-Commits title and a `JIRA ID : ROCM-27290` reference in the description, to satisfy the Systems PR Bot's Branch Name / PR Title-Description / Unit Test checks that were failing here.

Root-causing this was a good lesson: GitHub's branch-rename API does **not** carry an already-open PR over to the renamed branch (it closes the PR against the deleted old ref instead) — new PR was required.

PR #8095, opened 2026-07-02
[rocprofiler-sdk] Replace Queue Interposition's completion waiter with a batched hsa_amd_signal_wait_any waiter

bwelton:
## Summary
Replaces inline Queue Interposition's completion-waiter mechanism with a single batched `hsa_amd_signal_wait_any`-based waiter, addressing the CPU overhead reported in ROCM-27290. This is now the **only** completion-waiter mechanism (no opt-in flag/alternate mode) — the previous per-task, one-signal-at-a-time waiter has been removed entirely.

**Depends on:** #8094 (companion rocr-runtime fix for a `hsa_amd_signal_wait_any` out-of-bounds bug this waiter relies on).

## Motivation
The previous waiter used a 4-thread PTL `TaskGroup` where each worker thread called `hsa_signal_wait_relaxed_fn` on exactly one dispatch completion signal at a time, fully blocking on that one wait before it could look at any other outstanding completion. Under bursty/idle-to-busy load this caused severe head-of-line blocking: a handful of individually-slow KFD waits per worker thread stalled the entire completion pipeline, even though many later dispatches were already GPU-complete.

Measured impact on a synthetic 200K-kernel reproducer (gfx1201): the signal pool grew up to 48x (200,704 vs a nominal 8,192 signals) and host CPU overhead was ~2.3x versus Queue Interposition disabled.

## Design
A single long-lived waiter thread batches all currently-outstanding completion signals into one `hsa_amd_signal_wait_any()` call. That API's underlying implementation (`core::Signal::WaitMultiple`) does a cheap non-blocking scan of the entire batch first (pure atomic loads, no syscall), and only falls back to a real blocking KFD wait — itself batched across the whole outstanding set via a single `hsaKmtWaitOnMultipleEvents_Ext` ioctl — when nothing in the batch is yet satisfied. This gives the same "wake on any of N outstanding signals, one syscall" property already used internally by ROCR's own `AsyncEventsLoop`, via a stable public API.

Explicitly avoided:
- `hsa_amd_signal_async_handler` (callback-registration API) — avoided due to a prior history of bugs with that specific mechanism in this codebase.
- Linking `libhsakmt` directly or issuing raw KFD ioctls — unnecessary given the above.

## Results
- Host CPU overhead: **2.33x (previous waiter) → 1.33x (this change)** vs QI disabled — eliminates ~75.5% of the previous waiter's excess overhead.
- Signal-pool growth: **200,704 → 8,192** (matches nominal, i.e. no growth-driven pathology).
- Confirmed via a second, independent benchmark harness (a multi-process crash/hang reproducer) with matching ratios.

## Modified Files
- `source/lib/rocprofiler-sdk/hsa/queue_interposition.cpp` / `.hpp`:
  - `batched_waiter_t` owns a mutex-guarded pending-entry vector (keyed by a stable per-entry id, not vector index, since indices can shift under concurrent enqueue/erase), a dedicated wake signal, and one waiter thread.
  - Producer-side wiring in `process_doorbell_impl()`/`write_interceptor()` preserves the existing "publish real doorbell, unlock gate_lock, THEN arm waiter" ordering.
  - `complete_queue_info_session()` is a shared completion helper used by both the normal waiter path and the shutdown-fallback path.
  - Explicit finalization handling: `shutdown_and_join()` does not pre-drain (which could hang forever on a stuck signal); it sets a shutdown flag and bumps the wake signal, then runs a short bounded final-drain pass (20 attempts × 5ms) to catch any very-late-arriving entries before exiting. The wake-signal bump/push on the producer side and the shutdown-flag-set/wake-bump/destroy sequence on the shutdown side are scoped under the same mutex, so a producer can never observe `shutdown==false` and then touch an already-destroyed wake signal.
- `source/lib/rocprofiler-sdk/tests/queue_interposition.cpp`: removed 3 tests that referenced the now-deleted thread-count configuration of the previous waiter.

## Testing
- `ctest -R queue_interposition`: 14/14 passed (down from 17 — 3 tests tied to the removed waiter's configuration were deleted).
- 10/10 repeated tight-loop process-lifecycle stress runs: no hangs, no crashes.
- A/B benchmark across QI-disabled / QI-enabled, 3 reps each, plus a second independent crash/hang reproducer harness (both produced consistent ~2.3x / ~1.3x ratios before/after this waiter change).
- This change went through multiple implementer → reviewer → implementer correction cycles; review caught and fixed a real out-of-bounds bug in ROCR itself (#8094) and a wake-signal lifetime race in this code.

## Known Limitations
- A separate, pre-existing heap-corruption bug (intermittent "corrupted double-linked list" `SIGABRT` at process teardown under concurrent load with Queue Interposition enabled) was discovered while re-testing the original Jira crash/hang reproducer. It reproduces at a similar rate with both the old and new waiter and is **not** addressed by this change; it needs its own separate investigation/ticket.

## Related
ROCM-27290 ("Impact on JAX unit test execution time and deadlock issues with ROCPROFILER_QUEUE_INTERPOSITION=1")

### DIFF
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_interposition.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_interposition.cpp
@@ -41,7 +41,6 @@
 #include "lib/rocprofiler-sdk/hsa/hsa.hpp"
 #include "lib/rocprofiler-sdk/hsa/queue_controller.hpp"
 #include "lib/rocprofiler-sdk/hsa/signal_pool.hpp"
-#include "lib/rocprofiler-sdk/internal_threading.hpp"
 #include "lib/rocprofiler-sdk/kernel_dispatch/tracing.hpp"
 #include "lib/rocprofiler-sdk/registration.hpp"
 #include "lib/rocprofiler-sdk/tracing/tracing.hpp"
@@ -55,8 +54,13 @@
 #include <hsa/hsa_api_trace.h>
 #include <pthread.h>
 
+#include <algorithm>
 #include <atomic>
+#include <chrono>
+#include <condition_variable>
+#include <cstdint>
 #include <cstring>
+#include <mutex>
 #include <thread>
 #include <vector>
 
@@ -207,9 +211,6 @@ get_doorbell_tls()
     return _v;
 }
 
-using async_signal_task_t        = std::function<void()>;
-using async_signal_task_vector_t = std::vector<async_signal_task_t>;
-
 inline void
 publish_submitted_packets(QueueState* state, uint64_t submit_pos)
 {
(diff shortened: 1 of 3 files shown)


bwelton:
Updated: this now replaces the previous waiter entirely as the sole completion-waiter mechanism (removed the `ROCPROFILER_QI_BATCHED_WAIT` opt-in flag and the old PTL-TaskGroup-based waiter/tests). Also stripped remaining investigation-only scaffolding (health-counter struct/`get_stats()`/shutdown-summary log, and two leftover defensive null-checks inconsistent with the rest of the file's HSA-table-call convention). Comments trimmed dramatically throughout. Rebuilt and retested: `ctest -R queue_interposition` 14/14 (down from 17 — 3 tests tied to the removed waiter's config were deleted), 10/10 repeated shutdown-stress runs clean, perf numbers unchanged (~1.33x vs QI off).

PR #8097, opened 2026-07-02
test(rocprofv3): add integration tests for PC sampling with --selected-regions

julijose-amd:
## Motivation

rocprofv3 pc-sampling with `--selected-regions` had no integration coverage. The existing `roctx-pause-resume/pc-sampling` test uses `--marker-trace`, which does not gate collection, so pc-sampling combined with selected-regions was untested.

## Technical Details

Each method covers:
- `--selected-regions` and `--selected-regions` with `--selected-regions-ref-count`
- CSV and JSON output
- Kernel-dispatch tracing for sample-to-kernel attribution

There are 18 tests total:
- 4 execute tests
- 14 validation tests
Shared validators live under `tests/pytest-packages/pc_sampling/selected_regions/`. The host-trap and stochastic `validate.py` files import these helpers, while fixtures remain local to each test directory.

The common validators check:
- Exact method-specific CSV schema
- Non-trivial sample volume
- CSV/JSON sample-count parity
- Undecoded internal and blit-kernel sample sentinels
- At least 100 `v_mov_b32` samples and a minimum 30% instruction ratio
- Strict mapping from PC-sample dispatch IDs to kernel-dispatch records
- Samples are restricted to kernels launched inside selected regions

Undecoded internal samples, including blit-kernel samples, are validated using sentinel values. CSV requires `Correlation_Id == 0` and `Dispatch_Id == 0`; JSON additionally requires `inst_index == -1` and `code_object_id == 0`. These samples remain included in volume and CSV/JSON sample-count parity, but are excluded from kernel gating because they cannot be attributed to a dispatched kernel.

Stochastic validation additionally checks:
- `Wave_Issued_Instruction`
- `Instruction_Type`
- `Stall_Reason`
- `Wave_Count`
- `v_mov_b32` issue and arbiter semantics

For issued `v_mov_b32` samples, the test requires a VALU issue without a VALU stall. For non-issued samples, it validates the reported stall reason and associated arbiter state.

Following the existing stochastic exec-mask and transpose validators, the gfx9-specific arbiter semantic test is skipped on gfx12. Common PC-sampling, selected-region gating, and stochastic field checks continue to run there.

## Selected-Region Gating
Each decoded PC sample is mapped through: `dispatch_id` → `kernel_dispatch` → `kernel_symbols`

The tests assert that:
- `kernel_add` and `kernel_mult`, launched while profiling is paused, produce no samples.
- Only resume-window kernels are sampled.
- Without reference counting, `nested_kernel` is neither sampled nor traced.
- With reference counting, exactly one `nested_kernel` appears in the kernel-dispatch trace.

> **Note:** `nested_kernel` is too short to reliably produce PC samples, so reference-count behavior is validated through the kernel-dispatch trace. Extending the workload to make `nested_kernel` reliably sampleable remains a follow-up.

## JIRA ID

JIRA ID: AIROCVAL-54

## Test Plan
ctest -R pc-sampling-selected-regions

## Test Result

All 18 tests pass on MI350 (gfx950) from a clean build.


bgopesh (changes requested):
host-trap/conftest.py & stochastic/conftest.py are identical. 

Hoist a single conftest.py (and templated pytest.ini) into the parent pc-sampling-selected-regions/ directory — pytest auto-discovers parent conftest.py, so both subdirs inherit the fixtures with zero copies.  
  The parent CMakeLists.txt currently only does add_subdirectory(...); it can rocprofiler_configure_pytest_files the shared conftest instead

projects/rocprofiler-sdk/tests/rocprofv3/roctx-pause-resume/pc-sampling-selected-regions/host-trap/validate.py
```diff
@@ -0,0 +1,88 @@
...
+    assert (
+        list(pc_csv.columns) == HOST_TRAP_COLUMNS
+    ), f"unexpected columns: {list(pc_csv.columns)}"
+    assert len(pc_csv) >= MIN_SAMPLES, f"too few samples: {len(pc_csv)}"
+    assert (pc_csv["Exec_Mask"] > 0).all(), "Exec_Mask must be > 0"
+    assert (pc_csv["Dispatch_Id"] > 0).all(), "Dispatch_Id must be > 0"
+    assert (pc_csv["Correlation_Id"] >= 0).all(), "Correlation_Id must be >= 0"
+
+
+def test_validate_pc_sampling_selected_regions_json(pc_csv, json_data):
```
rpathani: The --selected-regions-ref-count runs (...-ref-count-host-trap and ...-ref-count-stochastic) use the exact same assertions as the non-ref-count runs, so nothing here actually verifies ref-counting.

This is documented behavior — see docs/how-to/using-rocprofiler-sdk-roctx.rst, "Nested pause-resume pairs": --selected-regions-ref-count "uses reference counting ... and only toggles profiling when the outermost pair boundaries are crossed." The app is built to exercise exactly this: after resume, resume, pause, the first nested_kernel (marked // Kernel should appear only with --selected-regions-ref-count) should be sampled only with --selected-regions-ref-count and absent without it. (

Suggest: in the ref-count validator, assert nested_kernel appears in the records (via the kernel-name / pc_sample_instructions strings in the JSON); in the plain validator, assert it does not. Otherwise a broken --selected-regions-ref-count would still pass everything.

projects/rocprofiler-sdk/tests/rocprofv3/roctx-pause-resume/pc-sampling-selected-regions/host-trap/validate.py
```diff
@@ -0,0 +1,88 @@
...
+    tool = _tool(json_data)
```
rpathani: These tests confirm "valid samples exist" but never confirm collection was actually restricted to the selected regions — which is the whole point of --selected-regions. Per docs/how-to/using-rocprofiler-sdk-roctx.rst (Mode 2): "profiling is disabled by default ... data collection takes place only for regions explicitly enclosed within roctxProfilerResume() and roctxProfilerPause()."

The app runs kernel_add/kernel_mult 64× each while paused, plus a trailing nested_kernel marked // Should not appear in trace, and none are asserted absent. Because pc_sampling_kernel is such a dense v_mov_b32 loop, even a broken gate (collecting everything) could still clear the ≥100-samples and ≥30% v_mov_b32 bar.

Suggest asserting samples come only from the expected kernels (pc_sampling_kernel/target_kernel) — i.e. the paused kernels and the final nested_kernel contribute zero samples.

rpathani (changes requested):
One thing worth strengthening is that the tests verify "selected-regions produced valid-looking output" but not "selected-regions actually did its job." Two things specifically:

The ref-count tests don't test ref-counting. The --selected-regions-ref-count runs use the identical assertions as the plain runs, so a completely broken ref-count feature would still pass. The neat part is the test app is already built for this: it has a nested_kernel that — per the docs and the code — should be sampled only with ref-count on. The tests just never check for it. Asserting that kernel is present (ref-count) / absent (plain) would make these tests actually prove the feature.

Gating isn't verified, only presence. Nothing checks that the kernels running outside the marked regions (the paused kernel_add/kernel_mult and the trailing nested_kernel) produced zero samples. Since the sampled loop is so instruction-dense, even a broken gate could still pass the current thresholds. A check that samples come only from the expected kernels would close this.

+1 and agree with @bgopesh   on de-duplicating the identical conftest.py/pytest.ini into the parent folder

julijose-amd:
> host-trap/conftest.py & stochastic/conftest.py are identical.
> 
> Hoist a single conftest.py (and templated pytest.ini) into the parent pc-sampling-selected-regions/ directory — pytest auto-discovers parent conftest.py, so both subdirs inherit the fixtures with zero copies. The parent CMakeLists.txt currently only does add_subdirectory(...); it can rocprofiler_configure_pytest_files the shared conftest instead

Thanks for the review @bgopesh.  
Duplicate validators/conftest: Moved the shared validators into a pytest-packages module (`tests/pytest-packages/pc_sampling/selected_regions/, csv.py + json.py`). Both validate.py files now import from it instead of duplicating the checks. 
I did look at using a single shared conftest at the parent, but our test harness sets pytest's `rootdir` to each subdir, and pytest only picks up conftests at or below rootdir — so a parent one wouldn't get loaded. Keeping the fixtures inline in each subdir conftest.py turned out to be the cleaner fit, and it's also how the existing pc-sampling tests are set up— e.g. `pc-sampling/host-trap/exec-mask-manipulation/` keeps its own conftest while pulling shared checks from `pytest-packages/pc_sampling/exec_mask_manipulation/`. After we talked through a few options offline and this felt like the best fit with the repo conventions.

projects/rocprofiler-sdk/tests/rocprofv3/roctx-pause-resume/pc-sampling-selected-regions/host-trap/validate.py
julijose-amd: Thanks for catching that. 
On the ref-count runs asserting the same as the plain runs: you're right, they weren't actually testing differently. They now diverge, using nested_kernel as the signal, without `--selected-regions-ref-count` it's neither sampled nor in the kernel-dispatch trace, and with ref counting exactly one nested_kernel shows up. So a broken ref-count would now be caught.
One thing worth flagging: the nested_kernel ref-count check leans on the kernel-dispatch trace rather than the PC samples, because nested_kernel is very short (~35 µs) and doesn't reliably produce samples. If we extend it later to generate its own samples, we can assert ref-count directly on the PC samples too.

projects/rocprofiler-sdk/tests/rocprofv3/roctx-pause-resume/pc-sampling-selected-regions/host-trap/validate.py
julijose-amd: Both runs now map each PC sample back to its kernel (dispatch_id → kernel_dispatch → kernel_symbols) and check that the paused kernels (kernel_add/kernel_mult) produce no samples, and that only the resume-window kernels are sampled.

bgopesh (approved):

projects/rocprofiler-sdk/tests/pytest-packages/pc_sampling/selected_regions/csv.py
```diff
@@ -0,0 +1,59 @@
...
+    "Stall_Reason",
+    "Wave_Count",
+]
+
+MIN_SAMPLES = 100
+
+
+def validate_columns(df, method):
+    # CSV columns match the exact schema for this sampling method
+    expected = STOCHASTIC_COLUMNS if method == "stochastic" else HOST_TRAP_COLUMNS
```
vlaindic: I'd recommend we define `STOCHASTIC_METHOD_STR = "stochastic"` and then compare against it instead of hardcoding values.

projects/rocprofiler-sdk/tests/pytest-packages/pc_sampling/selected_regions/csv.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def validate_sample_volume(df):
+    # enough samples were collected
+    assert len(df) >= MIN_SAMPLES, f"too few samples: {len(df)}"
+
+
+def validate_values(df):
+    # per-row fields are within valid ranges
+    assert (df["Exec_Mask"] > 0).all(), "Exec_Mask must be > 0"
```
vlaindic: I think EXEC_MASK could be zero around branches, but I believe the code we execute lacks them, correct?

projects/rocprofiler-sdk/tests/pytest-packages/pc_sampling/selected_regions/csv.py
```diff
@@ -0,0 +1,59 @@
...
+    assert (df["Dispatch_Id"] > 0).all(), "Dispatch_Id must be > 0"
+    assert (df["Correlation_Id"] >= 0).all(), "Correlation_Id must be >= 0"
```
vlaindic: Are we sure we're not sampling blit kernels in any situations (e.g., copies)?

projects/rocprofiler-sdk/tests/pytest-packages/pc_sampling/selected_regions/json.py
```diff
@@ -0,0 +1,172 @@
...
+def _tool(json_data):
+    tool = json_data["rocprofiler-sdk-tool"]
+    return tool[0] if isinstance(tool, list) else tool
+
+
+def _records_key(method):
+    return "pc_sample_host_trap" if method == "host_trap" else "pc_sample_stochastic"
+
+
+def validate_csv_json_parity(df, json_data, method):
```
vlaindic: I think this only ensures that the sampling num matches, so I think it's worth renaming function.

```suggestion
def validate_csv_json_parity_num_samples(df, json_data, method):
```

projects/rocprofiler-sdk/tests/rocprofv3/roctx-pause-resume/pc-sampling-selected-regions/stochastic/validate.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def test_validate_pc_sampling_stochastic_specific_json(json_data):
+    # stochastic-only fields are well-formed in the JSON records
+    for rec in _records(json_data):
+        r = rec["record"]
+        assert r["wave_issued"] in (0, 1)
+        assert r["wave_cnt"] > 0
+        assert r["inst_type"].startswith(INSTRUCTION_TYPE_PREFIX)
+        assert r["snapshot"]["stall_reason"].startswith(STALL_REASON_PREFIX)
```
vlaindic, line 129: Doesn't need to be part of this PR, but I'd recommend adding the following checks:
- filter out all v_mov_b32 instructions (by inst string) and then ensure:
- if wave_issue=1, inst_type == VALU, arbiter_state_issue_value=1, arbiter_statet_stall=0
- else stall_reason is either no_instruction_available, arbiter_not_win, aludep.

This should give us some more confidence that stochastic part is fine.

vlaindic (commented):
Overall, I'm fine with the implementation. I don't mind landing as is, although I added a few comments that I believe are worth addressing and can be done as a separate effort in another PR (maybe opening a JIRA/issue would be good). 

Also, adding AI review concerns:


### Comment for Finding #1 — `dispatch_info` inconsistency (`json.py`)

> **`_dispatch_id_to_kernel_name` and `_kernel_dispatch_names` handle the `kernel_dispatch` record shape differently.**
>
> `_dispatch_id_to_kernel_name` (line 81) falls back to a `dispatch_data.dispatch_info` nesting:
> ```python
> di = kd.get("dispatch_info") or kd.get("dispatch_data", {}).get("dispatch_info", {})
> ```
> `_kernel_dispatch_names` (line 97) does not:
> ```python
> di = kd.get("dispatch_info") or {}
> ```
>
> Based on other tests in the repo (`tracing-hip-in-libraries`, `hip-stream-display`), `--kernel-trace` output always puts dispatch info at `kd["dispatch_info"]` — the `dispatch_data` nesting appears only in counter-collection records. So the fallback in `_dispatch_id_to_kernel_name` is dead code today, and the inconsistency is the real risk: if the shape ever did use `dispatch_data`, `_kernel_dispatch_names` would silently return an empty list, making the "nested_kernel not in trace" assertion pass incorrectly.
>
> Please unify both functions to the same access pattern and add a short comment stating which JSON shape `--kernel-trace` produces. If the `dispatch_data` fallback is intentional (e.g., forward-compat), document why and apply it to both functions.

---

### Comment for Finding #2 — dead `_records()` helper (`stochastic/validate.py`)

> **`_records()` is a local re-implementation of `_tool()` from `selected_regions/json.py`, used in exactly one place.**
>
> ```python
> def _records(json_data):
>     tool = json_data["rocprofiler-sdk-tool"]
>     tool = tool[0] if isinstance(tool, list) else tool
>     return tool["buffer_records"]["pc_sample_stochastic"]
> ```
>
> `pcs_json` is already imported at the top of this file. The test that uses `_records()` could call `pcs_json._tool(json_data)["buffer_records"]["pc_sample_stochastic"]` directly, or you could expose a small `get_records(json_data, method)` helper in `json.py` and reuse it here. As-is, there are two independent code paths that decode the top-level JSON shape — they will drift if the shape changes.

vlaindic (approved):
Please find comments here: https://github.com/ROCm/rocm-systems/pull/8097#pullrequestreview-4789599169

projects/rocprofiler-sdk/tests/pytest-packages/pc_sampling/selected_regions/csv.py
julijose-amd: Added `HOST_TRAP_METHOD_STR` and `STOCHASTIC_METHOD_STR`, along with explicit method-to-schema and method-to-record-key mappings. Method selection no longer relies on hardcoded comparisons, and unsupported method values now fail instead of silently falling back to another format.

projects/rocprofiler-sdk/tests/pytest-packages/pc_sampling/selected_regions/csv.py
julijose-amd: Good catch. Eventhough the instruction sequence is `v_mov_b32`, `pc_sampling_kernel` contains a non-unrolled loop, so the generated ISA includes control-flow/branch instructions. A blanket `Exec_Mask > 0` assertion for every decoded sample is therefore not safe and has been removed.

Exact execution-mask correctness remains covered by the dedicated exec-mask tests, while this test focuses on selected-region gating and sample attribution.

projects/rocprofiler-sdk/tests/pytest-packages/pc_sampling/selected_regions/csv.py
julijose-amd: You're right, we could observe undecoded internal samples, including samples from blit kernels, so the validation no longer assumes they are absent.

The CSV check requires `Correlation_Id == 0` iff the instruction is not decoded, with `Dispatch_Id == 0` for those samples.

The JSON check requires the corresponding sentinel values: `corr_id.internal == 0`, `dispatch_id == 0`, `inst_index == -1`, and `code_object_id == 0`.

These samples remain included in total sample volume and CSV/JSON sample-count parity. They are excluded only from kernel mapping because they cannot be attributed to a dispatched kernel.

projects/rocprofiler-sdk/tests/pytest-packages/pc_sampling/selected_regions/json.py
julijose-amd: Renamed `validate_csv_json_parity()` to `validate_csv_json_parity_num_samples()` and updated both host-trap and stochastic callers.

projects/rocprofiler-sdk/tests/rocprofv3/roctx-pause-resume/pc-sampling-selected-regions/stochastic/validate.py
julijose-amd: The stochastic validator now filters to decoded `v_mov_b32` samples by instruction string and checks:
- When `wave_issued == 1`:
  - `inst_type == VALU`
  - `arb_state_issue_valu == 1`
  - `arb_state_stall_valu == 0`
- When `wave_issued == 0`:
  - `stall_reason` must be one of the recognized reasons.
  
I also included `ARBITER_WIN_EX_STALL`, which is accepted by the existing gfx9 VALU validator and was observed on gfx950. For this reason, `arb_state_issue_valu == 1` is required.
These arbiter checks seems to be architecture-sensitive, so they are isolated in a separate test that skips on gfx12. All common selected-region and stochastic field validation continues to run on gfx12.

julijose-amd:
> Overall, I'm fine with the implementation. I don't mind landing as is, although I added a few comments that I believe are worth addressing and can be done as a separate effort in another PR (maybe opening a JIRA/issue would be good).
> 
> Also, adding AI review concerns:
> 
> ### Comment for Finding #1 — `dispatch_info` inconsistency (`json.py`)
> > **`_dispatch_id_to_kernel_name` and `_kernel_dispatch_names` handle the `kernel_dispatch` record shape differently.**
> > `_dispatch_id_to_kernel_name` (line 81) falls back to a `dispatch_data.dispatch_info` nesting:
> > ```python
> > di = kd.get("dispatch_info") or kd.get("dispatch_data", {}).get("dispatch_info", {})
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `_kernel_dispatch_names` (line 97) does not:
> > ```python
> > di = kd.get("dispatch_info") or {}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Based on other tests in the repo (`tracing-hip-in-libraries`, `hip-stream-display`), `--kernel-trace` output always puts dispatch info at `kd["dispatch_info"]` — the `dispatch_data` nesting appears only in counter-collection records. So the fallback in `_dispatch_id_to_kernel_name` is dead code today, and the inconsistency is the real risk: if the shape ever did use `dispatch_data`, `_kernel_dispatch_names` would silently return an empty list, making the "nested_kernel not in trace" assertion pass incorrectly.
> > Please unify both functions to the same access pattern and add a short comment stating which JSON shape `--kernel-trace` produces. If the `dispatch_data` fallback is intentional (e.g., forward-compat), document why and apply it to both functions.
> 
> ### Comment for Finding #2 — dead `_records()` helper (`stochastic/validate.py`)
> > **`_records()` is a local re-implementation of `_tool()` from `selected_regions/json.py`, used in exactly one place.**
> > ```python
> > def _records(json_data):
> >     tool = json_data["rocprofiler-sdk-tool"]
> >     tool = tool[0] if isinstance(tool, list) else tool
> >     return tool["buffer_records"]["pc_sample_stochastic"]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `pcs_json` is already imported at the top of this file. The test that uses `_records()` could call `pcs_json._tool(json_data)["buffer_records"]["pc_sample_stochastic"]` directly, or you could expose a small `get_records(json_data, method)` helper in `json.py` and reuse it here. As-is, there are two independent code paths that decode the top-level JSON shape — they will drift if the shape changes.

Thanks for the review, @vlaindic. I have addressed all the comments.
- **`dispatch_info` inconsistency:** Both kernel-trace consumers now use the documented top-level `kernel_dispatch["dispatch_info"]` shape. The counter-collection-only `dispatch_data` fallback was removed, so missing or malformed kernel-trace fields fail explicitly instead of silently producing empty mappings.

- **Duplicate `_records()` helper:** Removed the local helper from the stochastic validator. Both validation paths now use the shared `get_tool()` and `get_records()` helpers from `json.py`; `get_tool()` also verifies that exactly one tool entry is present.

vlaindic (approved):
LGTM, thanks!

rpathani (approved):
Approving it. look good to me.

rpathani (approved):
Looks good to me. Approving it.

PR #8099, opened 2026-07-02
fix(libhsakmt): [ROCM-26862] Add integer checks to prevent  buffer overflow

andmar-amd:
Add integer overflow checks to prevent stack buffer overflow.

Problem:
In libhsakmt, the expression `nattr * sizeof(*attrs)` can wrap on a 32-bit size_t multiplication, producing a much smaller value than expected. The result is used directly as the allocation size, creating an undersized buffer. The subsequent memcpy() then writes past the end of the allocation, causing a stack-based out-of-bounds write.

Solution:
Replaced alloca() with malloc() and added overflow checks before the multiplication in all affected locations. Uses goto-based cleanup pattern for single-point resource management. Returns HSAKMT_STATUS_INVALID_PARAMETER on overflow and HSAKMT_STATUS_NO_MEMORY on allocation failure.

Testing:
- All kfdtest smoke tests pass with no regressions
- Build completes successfully with expected warnings on 64-bit systems

Notes:
- goto cleanup pattern provides single free() point, preventing resource leaks
- Compiler warnings "comparison is always false" are expected on 64-bit systems where overflow is mathematically impossible with 32-bit inputs
- On 32-bit systems, the checks will properly prevent overflow

Security Impact:
Fixes stack-based buffer overflow vulnerability (ROCM-26862)
JIRA ID : ROCM-26862

## JIRA ID

Resolves ROCM-26862

## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/libhsakmt/src/events.c
+++ b/projects/rocr-runtime/libhsakmt/src/events.c
@@ -273,6 +273,7 @@ static HSAKMT_STATUS get_mem_info_svm_api(HsaKFDContext *ctx, uint64_t address, 
 	uint32_t node_id = 0;
 	HSAuint32 s_attr;
 	HSAuint32 i;
+	HSAKMT_STATUS ret;
 	HSA_SVM_ATTRIBUTE attrs[] = {
 					{HSA_SVM_ATTR_PREFERRED_LOC, 0},
 					{HSA_SVM_ATTR_PREFETCH_LOC, 0},
@@ -284,7 +285,12 @@ static HSAKMT_STATUS get_mem_info_svm_api(HsaKFDContext *ctx, uint64_t address, 
 	CHECK_KFD_MINOR_VERSION(5);
 
 	s_attr = sizeof(attrs);
-	args = alloca(sizeof(*args) + s_attr);
+
+	/* s_attr is a compile-time constant (32 bytes); no overflow possible */
+	args = malloc(sizeof(*args) + s_attr);
+	if (!args)
+		return HSAKMT_STATUS_NO_MEMORY;
+
 	args->start_addr = address;
 	args->size = PAGE_SIZE;
 	args->op = KFD_IOCTL_SVM_OP_GET_ATTR;
@@ -292,7 +298,8 @@ static HSAKMT_STATUS get_mem_info_svm_api(HsaKFDContext *ctx, uint64_t address, 
 	memcpy(args->attrs, attrs, s_attr);
 	if (hsakmt_ioctl(ctx->fd, AMDKFD_IOC_SVM + (s_attr << _IOC_SIZESHIFT), args)) {
 		pr_debug("op get range attrs failed %s\n", strerror(errno));
-		return HSAKMT_STATUS_ERROR;
+		ret = HSAKMT_STATUS_ERROR;
+		goto out;
 	}
 
 	pr_err("GPU address 0x%lx, is Unified memory\n", address);
@@ -337,11 +344,16 @@ static HSAKMT_STATUS get_mem_info_svm_api(HsaKFDContext *ctx, uint64_t address, 
 			break;
 		default:
 			pr_debug("get invalid attr type 0x%x\n", args->attrs[i].type);
-			return HSAKMT_STATUS_ERROR;
+			ret = HSAKMT_STATUS_ERROR;
+			goto out;
 		}
 	}
 
-	return HSAKMT_STATUS_SUCCESS;
+	ret = HSAKMT_STATUS_SUCCESS;
+
+out:
+	free(args);
+	return ret;
 }
 //Analysis memory exception data, print debug messages
 static void analysis_memory_exception(HsaKFDContext *ctx,
--- a/projects/rocr-runtime/libhsakmt/src/fmm.c
+++ b/projects/rocr-runtime/libhsakmt/src/fmm.c
@@ -1123,12 +1123,17 @@ static HSAKMT_STATUS fmm_register_mem_svm_api(HsaKFDContext *ctx,
 	HSAuint64 aligned_addr = (HSAuint64)address - page_offset;
 	HSAuint64 aligned_size = PAGE_ALIGN_UP(page_offset + size);
 	struct hsa_kfd_fmm_context *fmm_ctx = ctx->fmm_context;
+	HSAKMT_STATUS ret;
 
 	if (!fmm_ctx->first_gpu_mem)
 		return HSAKMT_STATUS_ERROR;
 
+	/* s_attr is a compile-time constant (16 bytes); no overflow possible */
 	s_attr = 2 * sizeof(struct kfd_ioctl_svm_attribute);
-	args = alloca(sizeof(*args) + s_attr);
+	args = malloc(sizeof(*args) + s_attr);
+	if (!args)
+		return HSAKMT_STATUS_NO_MEMORY;
+
 	args->start_addr = aligned_addr;
 	args->size = aligned_size;
 	args->op = KFD_IOCTL_SVM_OP_SET_ATTR;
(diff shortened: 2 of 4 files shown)


projects/rocr-runtime/libhsakmt/src/fmm.c
```diff
@@ -1159,13 +1172,21 @@ static HSAKMT_STATUS fmm_map_mem_svm_api(HsaKFDContext *ctx,
...
+	HSAKMT_STATUS ret;
 
 	if (!fmm_ctx->first_gpu_mem)
 		return HSAKMT_STATUS_ERROR;
 
 	nattr = nodes_array_size;
+
+	/* Check for integer overflow in multiplication */
+	if (nattr > (SIZE_MAX - sizeof(*args)) / sizeof(struct kfd_ioctl_svm_attribute))
+		return HSAKMT_STATUS_INVALID_PARAMETER;
```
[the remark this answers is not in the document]
andmar-amd: Greetings 

if you want a more minimalistic approach instead of the malloc/free combo, please see the attached: `alloca-overflow-checks.patch`
[alloca-overflow-checks.patch](https://github.com/user-attachments/files/29608190/alloca-overflow-checks.patch)

projects/rocr-runtime/libhsakmt/src/events.c
```diff
@@ -284,15 +285,24 @@ static HSAKMT_STATUS get_mem_info_svm_api(HsaKFDContext *ctx, uint64_t address,
 	CHECK_KFD_MINOR_VERSION(5);
 
 	s_attr = sizeof(attrs);
-	args = alloca(sizeof(*args) + s_attr);
+
+	/* Check for integer overflow before malloc */
+	if (s_attr > SIZE_MAX - sizeof(*args))
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** This `SIZE_MAX` check is dead code — `s_attr` is `sizeof(attrs)` where `attrs` is a fixed four-element stack array (32 bytes, a compile-time constant). `32 > SIZE_MAX - sizeof(*args)` is always false on any 32-bit or 64-bit platform. The subsequent `malloc` is safe regardless. The guard misleads future readers into thinking there is a real overflow risk at this call site.

Suggest either removing the guard entirely (just `malloc(sizeof(*args) + sizeof(attrs))`), or adding a comment like: `/* s_attr is a compile-time constant (32 bytes); guard retained for pattern consistency */`

projects/rocr-runtime/libhsakmt/src/fmm.c
```diff
@@ -1122,12 +1122,20 @@ static HSAKMT_STATUS fmm_register_mem_svm_api(HsaKFDContext *ctx,
 	HSAuint64 aligned_addr = (HSAuint64)address - page_offset;
 	HSAuint64 aligned_size = PAGE_ALIGN_UP(page_offset + size);
 	struct hsa_kfd_fmm_context *fmm_ctx = ctx->fmm_context;
+	HSAKMT_STATUS ret;
 
 	if (!fmm_ctx->first_gpu_mem)
 		return HSAKMT_STATUS_ERROR;
 
+	/* Check for integer overflow: sizeof(*args) + 2 * sizeof(struct kfd_ioctl_svm_attribute) */
+	if (2 * sizeof(struct kfd_ioctl_svm_attribute) > SIZE_MAX - sizeof(*args))
```

**[Important]** Same issue as `events.c`: `2 * sizeof(struct kfd_ioctl_svm_attribute)` is 16 bytes — a compile-time constant. `16 > SIZE_MAX - sizeof(*args)` is always false. This function has no caller-controlled `nattr`; the attribute count is always 2, so there is nothing to guard. Either remove the check, or add a comment explaining it is retained for pattern consistency only.

projects/rocr-runtime/libhsakmt/src/svm.c
```diff
@@ -56,8 +56,16 @@ hsaKmtSVMSetAttrCtx(HsaKFDContext *ctx,
 	if (size & (PAGE_SIZE - 1))
 		return HSAKMT_STATUS_INVALID_PARAMETER;
 
+	/* Check for integer overflow in multiplication and ioctl size-field limit */
+	if (nattr > (SIZE_MAX - sizeof(*args)) / sizeof(*attrs))
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** This `SIZE_MAX` check is dead on 64-bit: `nattr` is `unsigned int` (max ~4 billion), and `(SIZE_MAX - 24) / 8 ≈ 2.3×10¹⁸` on 64-bit — far larger than any `uint32_t` value. The `_IOC_SIZEBITS` check on the next line is the actual effective guard and fires at ~2044 attributes.

On 32-bit (`size_t` = 32 bits), this check IS meaningful: `nattr > (0xFFFFFFFF - 24) / 8 = 536870910`. If 32-bit builds are a supported target, the check is justified — but it needs a comment explaining this, otherwise readers on 64-bit will rightly wonder why it exists:
```c
/* SIZE_MAX check is meaningful only on 32-bit (size_t overflow);
 * the _IOC_SIZEBITS check below is the effective guard on 64-bit. */
```

cfreeamd (commented):
🤖 AI-assisted pre-review (for @cfreeamd) — `fmm_map_mem_svm_api` and `svm.c` SetAttr/GetAttr guards, goto cleanup, and alloca→malloc conversion look correct. Inline comments above cover the dead-code guards worth addressing.

projects/rocr-runtime/libhsakmt/src/events.c
andmar-amd: Greetings

I had dealt with much of this already, but I still cleaned up things on a more explicit way.

cfreeamd (approved):
LGTM

projects/rocr-runtime/libhsakmt/src/fmm.c
```diff
@@ -1159,13 +1169,27 @@ static HSAKMT_STATUS fmm_map_mem_svm_api(HsaKFDContext *ctx,
...
+	HSAKMT_STATUS ret;
 
 	if (!fmm_ctx->first_gpu_mem)
 		return HSAKMT_STATUS_ERROR;
 
 	nattr = nodes_array_size;
+
+	/* Check for integer overflow and ioctl size-field limit */
+#if SIZE_MAX == UINT32_MAX
+	/* 32-bit: check for size_t overflow (~536M attrs) */
```
xiaogang-chen-amd: _IOC_SIZEBITS is the data size limit that one ioctl can pass to kernel. You have that check at following. Do need use SIZE_MAX to check for 32 bits system? 

SIZE_MAX represents the maximum possible value that can be stored in a variable of type size_t. not size of data that ioctl can use.

andmar-amd: Greetings @xiaogang-chen-amd  

Thanks for the comments, I have already dealt with the SIZE_MAX so the way the code is written now make sure that only on 32 bit system this is applicable.

The _IOC_SIZEBITS is a good dimension in  incorporate!

andmar-amd: Fixed, the _IOC_SIZEBITS  does make things simpler!

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDSVMRangeTest.cpp
```diff
@@ -2212,4 +2212,123 @@ TEST_P(KFDSVMRangeTest, MapAllHighAddr) {
     TEST_END
 }
```
xiaogang-chen-amd, line 2214: I wonder it makes much sense to make a kfdtest case that intentionally has big nattr_too_large  value. This patch replaces buffer allocation from stack to heap. malloc already checks if it can allocate buffer.

The changes on libhsakmt for system call parameter allocations looks good for me. The kfdtest case adding may not be necessary.

andmar-amd: Thanks

I think the idea is good, but certainly not for this PR.  So that it is not lost , do create a "add new test case feature ticket" and we can do it another time.

Since you
> The changes on libhsakmt for system call parameter allocations looks good for me

Please give me an approval, I cannot merge w/o your approval.

kentrussell (approved):
Not sure if we need to pare down the kfdtest output, but that's a future task

PR #8109, opened 2026-07-03
chore(rocprofiler-sdk): remove libatomic dependency from ROCm

srihari-uttanur:
## Motivation

The libatomic package cannot be part of ROCm install because of licensing issues, so this commit introduces code changes which allows to remove the libatomic dependency. 
rocprofiler-sdk was unconditionally linking libatomic via the rocprofiler-sdk-atomic interface target, which caused link failures on images/toolchains where libatomic1 was not installed.

**ld.lld: error: unable to find library -latomic**

## Technical Details

rocprofv3 had ptrace_data_t which was atomic struct, this commit removes the atomicity by introducing memory ordering synchronization to get rid of the atomic dependency. 

The only reason libatomic was needed was a single std::atomic<ptrace_data_t> in the rocattach ptrace runner. Since
ptrace_data_t is about 40 bytes, those operations are not lock-free and the compiler lowers them to libatomic calls.
This change removes that dependency end-to-end.

Build system updates:

Remove atomic linkage from interface configuration.
Remove the now-unused rocprofiler-sdk-atomic interface target.
Remove downstream links to rocprofiler-sdk-atomic in common and aqlprofile.

Source updates:

Replace std::atomic<ptrace_data_t> with plain ptrace_data_t in the ptrace worker handoff path.
Preserve cross-thread correctness using the existing running flag with explicit release/acquire ordering:
Producer writes request data, then running=true (release).
Worker observes running=true (acquire), reads request, executes ptrace, writes result, then running=false (release).
Producer waits until running==false and then reads result.
Add an in-flight guard in PTraceRunner::run(): if running is still true, return ROCATTACH_STATUS_ERROR instead of publishing a new request.
This fixes the timeout-path race where a new request could overwrite
shared data while the worker still owned it.

## JIRA ID

JIRA ID : ROCM-27086

## Submission Checklist


projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/ptrace_runner.cpp
```diff
@@ -221,7 +229,9 @@ class PTraceRunner
                 return;
             }
 
-            ptrace_data = runner_data.data.load();
+            // Plain read: wait_for_eq uses seq_cst load, which provides acquire semantics,
+            // so all result writes by the worker thread are already visible here.
+            ptrace_data = runner_data.data;
```
subbul-amd, line 234: assuming underlying "wait_for_eq" doesn't change from default "seq_cst" to relaxed memory order

srihari-uttanur: I have updated the comment, the wait_for_eq is only safe under the default seq_cst load behavior and it must be revisited if that helper changes to a relaxed load.

meserve-amd (approved):
Good use of the existing atomic. LGTM

dgaliffiAMD:
Updated title to satisfy the "PR-Bot".

jrmadsen (approved):

PR #8110, opened 2026-07-03
style(rccl): adopt NCCL clang-format style and tooling for src/

mch:
Adopt NCCL's clang-format configuration and tooling in RCCL and reformat
the src/ tree to match, in preparation for upcoming NCCL syncs (reduces
formatting-only merge conflicts).

Changes:
- Add the NCCL clang-format config and associated tooling.
- Reformat src/ with clang-format to match NCCL style.
- Add the reformat commit to .git-blame-ignore-revs so blame skips it.
- Fix cmake/scripts/add_unroll.sh: the reformat changed device template
  declarations from `template<typename ...>` to `template <typename ...>`.
  Two perl substitutions in the script matched only the no-space form and
  silently stopped injecting the RCCL-only USE_ACC/COLL_UNROLL/Pipeline
  template parameters into the primary templates, while the trailing sed
  rules still appended them to the partial specializations — producing
  specializations that referenced undeclared identifiers and broke the
  device build. The regexes now accept either spacing (`template\s*<`).

Testing:
- `./install.sh -l -t --no_clean` builds cleanly (librccl.so + rccl-UnitTests).

JIRA ID : AICOMRCCL-1472


MythreyaK (approved):
`api_trace.h` looks good from rocprofiler-sdk side.

alex-breslow-amd (approved):
Going to stamp.  @wenkaidu, I think this might increase the difficulty of merges.  Let's sync if you have concerns.

PR #8114, opened 2026-07-03
rocr: make HotSwap entry trampolines opt-in

ammarwa:
## Summary
- keep the existing B0-to-A0 HotSwap translation path enabled by default
- make gfx12.5 HotSwap entry-trampoline rewrites opt-in via AMD_COMGR_HOTSWAP_ENTRY_TRAMPOLINES=1
- update ROCr tests and environment variable docs for the new default

## Notes
This leaves HSA_HOTSWAP_DISABLE behavior unchanged; it only changes the default for the entry-trampoline mode that replaced the older ROCr trampoline path.

### DIFF
--- a/projects/rocr-runtime/rocrtst/suites/functional/hotswap/hotswap_rewrite_test.cc
+++ b/projects/rocr-runtime/rocrtst/suites/functional/hotswap/hotswap_rewrite_test.cc
@@ -255,9 +255,16 @@ TEST(HotswapRewriteDecision, A0RetargetsThroughLegacyPathRegardlessOfOptions) {
   }
 }
 
-TEST(HotswapRewriteDecision, EntryTrampolinesDefaultOnRoutesNonA0Gfx1250) {
+TEST(HotswapRewriteDecision, EntryTrampolinesDefaultOffBlocksNonA0Gfx1250) {
   const auto decision = rocr::hotswap::DecideHotswapRewriteForTesting(
       MakeRevision("gfx1250", 1), kGfx1250Isa, kGfx1250Isa, {});
+
+  EXPECT_FALSE(decision.has_value());
+}
+
+TEST(HotswapRewriteDecision, EntryTrampolinesExplicitlyEnabledRoutesNonA0Gfx1250) {
+  const auto decision = rocr::hotswap::DecideHotswapRewriteForTesting(
+      MakeRevision("gfx1250", 1), kGfx1250Isa, kGfx1250Isa, {true});
 
   ASSERT_TRUE(decision.has_value());
   EXPECT_EQ(decision->source_isa, kGfx1250B0Isa);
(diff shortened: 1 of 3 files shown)


projects/rocr-runtime/rocrtst/suites/functional/hotswap/hotswap_rewrite_test.cc
```diff
@@ -255,10 +255,17 @@ TEST(HotswapRewriteDecision, A0RetargetsThroughLegacyPathRegardlessOfOptions) {
   }
 }
 
-TEST(HotswapRewriteDecision, EntryTrampolinesDefaultOnRoutesNonA0Gfx1250) {
+TEST(HotswapRewriteDecision, EntryTrampolinesDefaultOffBlocksNonA0Gfx1250) {
   const auto decision = rocr::hotswap::DecideHotswapRewriteForTesting(
       MakeRevision("gfx1250", 1), kGfx1250Isa, kGfx1250Isa, {});
 
+  EXPECT_FALSE(decision.has_value());
```
cfreeamd, line 262: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** This test passes `{}` (brace-initialized `RewriteOptions`) and expects `EXPECT_FALSE(decision.has_value())`. But `RewriteOptions::gfx12_5_rewrite_enabled` still defaults to `true` in `hotswap.hpp` (not changed by this PR), so `{}` == `{true}`. The new `EntryTrampolinesExplicitlyEnabledRoutesNonA0Gfx1250` test also passes `{true}` and asserts `ASSERT_TRUE(decision.has_value())` — these two tests assert contradictory outcomes for identical inputs.

If the intent is that the default should now be `false` (opt-in), the fix is to change the default in `hotswap.hpp`:
```cpp
struct RewriteOptions {
  bool gfx12_5_rewrite_enabled = false;  // opt-in via AMD_COMGR_HOTSWAP_ENTRY_TRAMPOLINES=1
};
```

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -98,8 +98,9 @@ bool IsHotswapDisabledByEnv() { return IsEnvFlagEnabled("HSA_HOTSWAP_DISABLE");
 
 bool IsGfx12_5RewriteRequested() {
   constexpr char kEnvName[] = "AMD_COMGR_HOTSWAP_ENTRY_TRAMPOLINES";
-  // Default-on policy owned by ROCR: only literal "0" opts out.
-  return !os::IsEnvVarSet(kEnvName) || os::GetEnvVar(kEnvName) != "0";
+  // Entry-trampoline rewrites are opt-in. The legacy B0-to-A0 HotSwap
+  // translation path remains enabled by default for supported agents.
```
cfreeamd, line 102: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** The env-var logic is now opt-in (only `"1"` enables). This is correct per the PR intent. But `hotswap.hpp`'s `RewriteOptions::gfx12_5_rewrite_enabled` still defaults to `true`, which means callers that construct `RewriteOptions{}` and pass it to `DecideHotswapRewrite` directly (bypassing the env-var check, e.g. in tests) get the old opt-out semantics. The test `EntryTrampolinesDefaultOffBlocksNonA0Gfx1250` relies on the struct default being `false` — see inline comment on the test file.

cfreeamd (commented):
🤖 AI-assisted pre-review (for @cfreeamd) — `hotswap.cpp` env-var logic and docs look correct. See inline comments: the `RewriteOptions` struct default in `hotswap.hpp` needs to change to `false` for the new tests to be consistent.

PR #8118, opened 2026-07-03
fix(rocjitsu): keep L1 coherent for uncached accesses

bjacob:
This acts on https://github.com/ROCm/rocm-systems/pull/8006#pullrequestreview-4627248305.


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l1_scalar_cache.cpp
```diff
@@ -83,6 +84,19 @@ void L1ScalarCache::writeback_all(uint32_t vmid) {
...
 
+void L1ScalarCache::flush_line(uint64_t addr, uint32_t vmid) {
+  simdojo::CacheTag *tag = nullptr;
+  if (!cache_.lookup(addr, &tag))
+    return;
+
+  if (tag->dirty) {
+    uint8_t line_buf[CacheStore::LINE_SIZE];
+    cache_.read_line(addr, line_buf, 0, CacheStore::LINE_SIZE);
+    l2_->writeback_line(CacheStore::line_address(addr), line_buf, Mtype::RW, vmid);
```
kuhar: `flush_line` passes a 64-byte scalar K$ line to `L2Cache::writeback_line`, which expects a full 128-byte L2 line. That makes the new UC scalar flush path overread the stack buffer and can also corrupt the neighboring half of the L2 line.

I reproduced this locally with the ASan build:

`./rocjitsu_tests --gtest_filter='L1ScalarCacheTest.*'`

It aborts in `L2Cache::writeback_line` from `L1ScalarCache::flush_line` with a stack-buffer-overflow on `line_buf`. Please route scalar K$ writeback through a byte-granular/partial L2 write, or assemble a full 128-byte L2 line before calling `writeback_line`.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l1_vector_cache.cpp
```diff
@@ -117,6 +119,8 @@ void L1VectorCache::write_bytes(uint64_t addr, const uint8_t *src, uint32_t size
       chunk_mtype = effective_mtype(inst_mtype, memory_->pte_mtype(ea, vmid));
 
     if (chunk_mtype == Mtype::UC || non_temporal) {
```
kuhar, line 120: The new UC paths invalidate a resident V$ line before bypassing L1, but the adjacent `non_temporal` / `request_l1_bypass` paths still bypass L1 without updating or invalidating any resident line.

That leaves stale data resident in V$: a non-temporal store can write the new value to L2, then a later cacheable read from the same cache can still hit the old L1 line. The read-side bypass path has the same contract question if L2 changes while a clean V$ line is resident.

If NT/bypass accesses are intentionally coherence-relaxed, please document and test that contract here. Otherwise, these paths should mirror the UC invalidation policy.

kuhar (commented):
A few comments

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l1_scalar_cache.cpp
```diff
@@ -125,6 +140,7 @@ void L1ScalarCache::load_bytes(uint64_t addr, uint32_t num_bytes, uint8_t *dst,
       mtype = memory_->pte_mtype(ea, vmid);
 
     if (mtype == Mtype::UC) {
+      flush_line(ea, vmid);
```
kuhar, line 150: Please add one focused UC load_bytes regression. This branch now calls flush_line before bypassing to L2, but the new scalar tests only exercise the load/store dword paths. load_bytes is the SMEM byte path used for sub-dword scalar loads and has its own chunking loop, so a small UcLoadBytesFlushesDirtyResidentLine case would cover a distinct new path.

emulation/rocjitsu/tests/shared_infra_test.cpp
```diff
@@ -771,6 +775,247 @@ TEST(L2CacheTest, UcWriteCrossingLineBoundaryFlushesBothDirtyResidentLines) {
...
+
+  {
+    std::unique_lock lock(page_table_mutex);
+    page_table[kAddr >> KfdProcess::kPageShift].mtype = amdgpu::Mtype::UC;
+  }
+  uint32_t read_value = 0;
+  l1.load(kAddr, /*num_dwords=*/1, &read_value, kVmid);
+
+  EXPECT_EQ(read_value, kDirtyValue);
+  EXPECT_EQ(mem.read32(kAddr, kVmid), kDirtyValue);
```
kuhar, line 830: This scalar UC-load test proves the dirty writeback, but not the invalidate half of flush_line. If cache_.invalidate(addr) were removed after the writeback, read_value and mem.read32 would still match kDirtyValue. Please add the same final step the vector tests use: mutate backing, switch the PTE back to RW, then do a cacheable l1.load and assert it sees the new backing value.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l1_scalar_cache.cpp
```diff
@@ -26,7 +26,7 @@ void L1ScalarCache::ensure_line(uint64_t addr, uint32_t vmid) {
     uint64_t evicted_line_addr =
         (evicted.tag << (LINE_SIZE_BITS + set_bits)) |
         (static_cast<uint64_t>(CacheStore::set_index(addr)) << LINE_SIZE_BITS);
-    l2_->writeback_line(evicted_line_addr, evicted_data, Mtype::RW, vmid);
+    l2_->write(evicted_line_addr, evicted_data, CacheStore::LINE_SIZE, Mtype::RW, vmid);
```
kuhar: The dirty K$ eviction/writeback paths changed from L2Cache::writeback_line to L2Cache::write, but the new tests never force a dirty scalar victim and the writeback_all call happens after the only dirty line was already flushed and invalidated. A small eviction/writeback_all regression would make this safer: dirty four lines in one set, touch a fifth to evict one, then assert the evicted dirty bytes reached backing or are visible after reload.

kuhar (commented):
A few comments -- looks good overall, but could use more tests

kuhar (commented):
Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l1_scalar_cache.cpp:61`

The scalar CC bypass paths still drop dirty resident K$ data. UC now calls `flush_line()` before bypassing to L2, but the CC branches in `store`, `load`, and `load_bytes` still do a bare `cache_.invalidate(...)` first.

I validated this with focused repro tests: dirty a K$ line under RW, flip the PTE to CC, then issue a CC `load`, `load_bytes`, or partial `store`. Current code returns the old backing value (`0x11111111`) instead of the dirty values (`0x22222222` / `0x44332211`), and the CC store drops the dirty dword outside the store range. Replacing the three CC invalidates with `flush_line(...); l2_->read/write(..., Mtype::CC, ...)` makes the same repros pass under both the GCC and ASan builds.

Please route scalar CC bypass through `flush_line()` before the L2 access, and add CC variants of the dirty-line UC tests so this does not regress again.

bjacob:
> Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l1_scalar_cache.cpp:61`
> 
> The scalar CC bypass paths still drop dirty resident K$ data. UC now calls `flush_line()` before bypassing to L2, but the CC branches in `store`, `load`, and `load_bytes` still do a bare `cache_.invalidate(...)` first.
> 
> I validated this with focused repro tests: dirty a K$ line under RW, flip the PTE to CC, then issue a CC `load`, `load_bytes`, or partial `store`. Current code returns the old backing value (`0x11111111`) instead of the dirty values (`0x22222222` / `0x44332211`), and the CC store drops the dirty dword outside the store range. Replacing the three CC invalidates with `flush_line(...); l2_->read/write(..., Mtype::CC, ...)` makes the same repros pass under both the GCC and ASan builds.
> 
> Please route scalar CC bypass through `flush_line()` before the L2 access, and add CC variants of the dirty-line UC tests so this does not regress again.

@kuhar , addressed in https://github.com/ROCm/rocm-systems/pull/8118/changes/3b6adb1277d7b86e54939e0c7148b2b465da1e98 , is that what you had in mind?

kuhar (approved):
LGTM

PR #8122, opened 2026-07-03
[rocjitsu] Fix RDNA/gfx1250 DPP

lua1235:
## Motivation
Extend DPP fixes to RDNA + gfx1250

## Technical Details
 - Changed RDNA/gfx1250 DPP to the full dpp_write_mask(...) path so BOUND_CTRL=0 preserves old
    destination/VCC/EXEC state for invalid shared-data lanes.
  - Added fetch-inactive (fi) handling for DPP16 and DPP8 source reads. (DPP8 FI is derived from SRC_DPP8_FI_0/1 source marker).
  - Split RDNA VOPC behavior by ISA support: RDNA3+ and gfx1250 support VOPC DPP paths, while RDNA1/RDNA2 now explicitly reject unsupported VOPC DPP/DPP8 markers.
  - Added corresponding tests covering DPP write-mask preservation, FI behavior, EXEC preservation, RDNA1/
    RDNA2 unsupported VOPC DPP cases.
## JIRA ID
N/A

## Test Plan
Unit tests, rocjitsu-test-corpus, fpsan tests.

Rerun the generator.

## Test Result
All pass

## Submission Checklist


kuhar (commented):
Can we add a small generated-instruction regression for the remaining DPP edge cases?

The current tests cover the main wave32 RDNA/gfx1250 paths, but a few important cases are still only covered indirectly:

- `dpp_read` now gives inactive source lanes with `fi=0` a zero value independent of `bound_ctrl`, while the FI generated-instruction tests use `bound_ctrl=1`. A `fi=0` plus `bound_ctrl=0` case would pin the intended precedence.
- `rdna1/rdna2/vop1.cpp` now contain the generated DPP FI/write-mask paths, but the new RDNA1/2 tests only instantiate the unsupported VOPC-DPP throw path. A VOP1 runtime instantiation would catch regressions there.
- The new RDNA/gfx1250 generated-instruction tests assert wave32, while the implementation uses 64-bit EXEC masks. One wave64 case with an inactive upper-32 source lane would cover the shift/merge path that wave32 cannot.

kuhar (approved):

PR #8128, opened 2026-07-05
fix(rccl): avoid address-taken DevFunc table in -fgpu-rdc build

### LINKED ISSUE #8129: [Issue]: RCCL -fgpu-rdc build regression
### Problem Description

RCCL -fgpu-rdc code generation regressed with addition of device linker build.

### Operating System

Ubuntu 22.04 LTS

### CPU

AMD EPYC 9654 96-Core Processor

### GPU

gfx90a

### ROCm Version

rocm 7.1.0

### ROCm Component

_No response_

### Steps to Reproduce

_No response_

<details>

```
Paste output here
```

</details>

### Additional Information

_No response_

lfmeadow:
Fixes #8129

Code generation in the `-fgpu-rdc` build was constrained because the device
function-pointer table caused address-taken to be set on the primary
`ncclDevFunc_*` device functions, blocking launch-bound/register propagation in
the whole-program link. Also removes `-w` from the device-linker build (it was
hiding warnings), and guards a Linux-only `alias(...)` in `debug.cc` out of the
HIP device compilation pass (pre-existing breakage that blocked the RDC build).

## Motivation

Restore the performance of the non device-linker (`-fgpu-rdc`) build.

## Technical Details

- Use a compile-time templated binary search in the non device-linker build,
  eliminating the function-pointer table and thus avoiding address-taken on the
  internal `ncclDevFunc_*`; the dispatch leaves call each function directly by name.
- Emit the function-pointer table only for the runtime-dispatch builds
  (`RCCL_DEVICE_LINKER` / legacy `USE_INDIRECT_FUNCTION_CALL`) and declare it
  `static` so unused copies are dead-stripped -- this retires `RCCL_DEVICE_TABLE_OMIT`.
- Apply `noinline` to `ncclDevFunc_*` only under `RCCL_DEVICE_LINKER` via a shared
  `RCCL_DEVFUNC_ATTR` macro (matched on both the forward declaration and definition).
- Arch-guarded-out dispatch leaves `__builtin_trap()` (fail-fast, matching the old
  `nullptr` table entries) rather than silently no-op.

## Test Plan

- New `src/device/test_generate_device_table.py` unit test locks in the dispatch
  gating (static/runtime-only table, pure-RDC by-name dispatch, noinline gating,
  no `RCCL_DEVICE_TABLE_OMIT`, guarded-out leaves trap, IFC still emits the table).
- RCCL unit tests on gfx90a.

## Test Result

- Generator unit test: pass.
- Verified the built `-fgpu-rdc` `librccl.so` device code has no `ncclDevFuncTable_*`
  symbol and no table-indexed indirect dispatch.
- RCCL unit tests: pass on 7 GPUs.

### DIFF
--- a/projects/rccl/src/device/generate.py
+++ b/projects/rccl/src/device/generate.py
@@ -398,20 +398,15 @@ with open(os.path.join(gensrc, "device_table.h"), "w") as f:
   print("-- Generating %s" % os.path.join(gensrc, "device_table.h"))
   out = f.write
 
-  # noinline is applied to every ncclDevFunc_* ONLY in the device-linker build.
-  out("#if defined(RCCL_DEVICE_LINKER)\n")
-  out("#define RCCL_DEVFUNC_ATTR __attribute__((noinline))\n")
-  out("#else\n")
-  out("#define RCCL_DEVFUNC_ATTR\n")
-  out("#endif\n\n")
-
+  # Plain forward declarations; noinline (device-linker only) is controlled
+  # solely by DEFINE_ncclDevFunc in common.h.
   for fn in primary_funcs:
     sym = paste("_", "ncclDevFunc", *fn)
     guard = get_arch_guard(fn)
     if guard:
-      out("#if %s\n__device__ RCCL_DEVFUNC_ATTR void %s();\n#endif\n" % (guard, sym))
+      out("#if %s\n__device__ void %s();\n#endif\n" % (guard, sym))
     else:
-      out("__device__ RCCL_DEVFUNC_ATTR void %s();\n" % sym)
+      out("__device__ void %s();\n" % sym)
   out("\n")
 
   index = {val: None for val in all_unrolls}
@@ -452,9 +447,8 @@ with open(os.path.join(gensrc, "device_table.h"), "w") as f:
           f"      : Caller{unroll}<m, l>::call{unroll}(funcIndex);\n"
           "  }\n"
           "};\n\n")
-      i = 0
-      for fn in primary_funcs:
-        if fn.unroll != unroll: continue
+      unroll_fns = [fn for fn in primary_funcs if fn.unroll == unroll]
+      for i, fn in enumerate(unroll_fns):
         sym = paste("_", "ncclDevFunc", *fn)
         guard = get_arch_guard(fn)
         spec = f"template<> struct Caller{unroll}<{i}, {i+1}> {{ static __forceinline__ __device__ void call{unroll}(unsigned short) noexcept"
@@ -469,10 +463,9 @@ with open(os.path.join(gensrc, "device_table.h"), "w") as f:
           out("#endif\n")
         else:
           out(f"{spec} {{ {sym}(); }} }};\n")
-        i += 1
       out("\n")
       out(f"__forceinline__ __device__ void NCCL_CALL_FUNCTIONS_{unroll}(unsigned short funcIndex) noexcept {{\n")
-      out(f"  Caller{unroll}<0, {i}>::call{unroll}(funcIndex);\n")
+      out(f"  Caller{unroll}<0, {len(unroll_fns)}>::call{unroll}(funcIndex);\n")
       out("}\n\n")
     out("#endif // !USE_INDIRECT_FUNCTION_CALL && !RCCL_DEVICE_LINKER\n")
 
--- a/projects/rccl/src/device/generate.py
+++ b/projects/rccl/src/device/generate.py
@@ -462,7 +462,10 @@ with open(os.path.join(gensrc, "device_table.h"), "w") as f:
           out(f"#if {guard}\n")
           out(f"{spec} {{ {sym}(); }} }};\n")
           out("#else\n")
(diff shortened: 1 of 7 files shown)


projects/rccl/src/device/generate.py
```diff
@@ -427,34 +434,44 @@ def get_arch_guard(fn):
...
+        if fn.unroll != unroll: continue
+        sym = paste("_", "ncclDevFunc", *fn)
+        guard = get_arch_guard(fn)
+        spec = f"template<> struct Caller{unroll}<{i}, {i+1}> {{ static __forceinline__ __device__ void call{unroll}(unsigned short) noexcept"
+        if guard:
+          out(f"#if {guard}\n")
+          out(f"{spec} {{ {sym}(); }} }};\n")
+          out("#else\n")
+          out(f"{spec} {{}} }};\n")
+          out("#endif\n")
```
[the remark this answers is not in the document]
lfmeadow: Good catch — addressed in db8db42df0. The arch-guarded-out `Caller<i,i+1>` specialization now calls `__builtin_trap()` instead of an empty no-op, so an out-of-range/inconsistent `funcId` fails fast, matching the old `nullptr` entries of the function-pointer table. Added `test_generate_device_table.py` which asserts guarded-out leaves trap.

lfmeadow:
Apparently Systems PR Bot is broken somehow (something about testbranch), so CI will never be clean. I just keep rebasing this to keep it up to date incase CI ever gets fixed.

projects/rccl/src/device/generate.py
```diff
@@ -398,24 +398,31 @@ def get_arch_guard(fn):
...
 
   for fn in primary_funcs:
     sym = paste("_", "ncclDevFunc", *fn)
     guard = get_arch_guard(fn)
     if guard:
-      out("#if %s\n%s %s();\n#endif\n" % (guard, func_declaration, sym))
+      out("#if %s\n__device__ RCCL_DEVFUNC_ATTR void %s();\n#endif\n" % (guard, sym))
     else:
-      out("%s %s();\n" % (func_declaration, sym))
+      out("__device__ RCCL_DEVFUNC_ATTR void %s();\n" % sym)
```
alex-breslow-amd: Is there a way to make this part of common.h, or do we need to have this in generate.py?

projects/rccl/src/device/generate.py
```diff
@@ -427,34 +434,47 @@ def get_arch_guard(fn):
...
-      out(f"template<unsigned short f>\n"
-          f"struct Caller{unroll}<f, f + 1> {{\n"
-          "  static __forceinline__ __device__ __host__\n"
-          f"  void call{unroll}(unsigned short funcIndex) noexcept {{\n"
-          f"    ncclDevFuncTable_{unroll}[f]();\n"
-          "  }\n"
-          "};\n\n")
-
-      # emit NCCL_CALL_FUNCTIONS_<unroll> wrapper using last index value
+      i = 0
```
alex-breslow-amd: enumerate?  Seems cleaner

projects/rccl/src/device/test_generate_device_table.py
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env python3
+"""Unit tests for src/device/generate.py device_table.h dispatch generation.
```
alex-breslow-amd, line 2: Glad this exists

alex-breslow-amd (approved):
LGTM.  I would appreciate if someone else can also review.

projects/rccl/src/device/generate.py
lfmeadow: Moved to common.h: the noinline attribute is now folded into the two RCCL_DEVICE_LINKER-gated DEFINE_ncclDevFunc variants, and generate.py emits plain forward declarations. Nothing devfunc-attr-related left in generate.py.

projects/rccl/src/device/generate.py
lfmeadow: Done — now enumerate() over a per-unroll filtered list (byte-identical output).

projects/rccl/src/device/test_generate_device_table.py
lfmeadow: Thanks!

lfmeadow:
Fixed Alex's two comments. I'm ready when you are. CI running.

projects/rccl/src/device/test_generate_device_table.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env python3
```
alex-breslow-amd, line 1: I didn't catch this when I first looked at it, but nothing runs this as part of CI, right?

Can you add that @lfmeadow?

lfmeadow: Good catch. Registered it as a CPU-only CTest (rccl-generate-device-table), gated on BUILD_TESTS — no GPU or built library needed. Added a top-level enable_testing() so it registers from the build root; verified it runs and passes via ctest -R rccl-generate-device-table.

PR #8136, opened 2026-07-06
Fix ROCM-27034: P2P/IPC transport setup fails under asymmetric HIP_VISIBLE_DEVICES

amd-arozanov:
## Motivation

Fixes ROCM-27034: `ncclCommInitRank` leaves a sticky HIP error behind under an asymmetric `HIP_VISIBLE_DEVICES` topology (each rank sees a different device subset). Init still returns `ncclSuccess`, but the pending error surfaces later as an "invalid device ordinal" failure on the first framework allocation (e.g. PyTorch), well away from its real cause.

The PR contains three changes:
- Bounds-check peer ordinals in the clique peer-access scan (`init.cc`).
- Clear the pending HIP error left by the optional MNNVL fabric-support probe (`mnnvl.cc`).
- Add a self-contained regression test reproducing the asymmetric topology.

## Technical Details

- `src/init.cc`: query `cudaGetDeviceCount` once and skip any peer pair whose ordinal falls outside `[0, localDevCount)`, treating it as no peer access instead of calling `hipDeviceCanAccessPeer` with an invalid device. This removes the `hipErrorInvalidDevice` at its source.
- `src/mnnvl.cc`: the `cuDeviceGetAttribute(CU_DEVICE_ATTRIBUTE_HANDLE_TYPE_FABRIC_SUPPORTED)` probe is best-effort and its result is intentionally ignored, but on ROCm builds where the attribute is unsupported it records a pending HIP error. Clear it with `hipGetLastError` so a successful init leaves clean error state.
- `test/AsymmetricVisibilityTests.cpp`: an orchestrator forks two workers via `fork()+execv()`, each with a distinct `HIP_VISIBLE_DEVICES` set before HIP initializes (rank0 sees `{0,1}` and binds ordinal 1, rank1 sees `{2}` and binds ordinal 0). Each worker joins a 2-rank communicator, asserts `hipGetLastError() == hipSuccess` after init (the ROCM-27034 invariant), and runs a small `AllReduce` to confirm the communicator is functional. The test requires at least 3 GPUs and runs with `NCCL_CUMEM_ENABLE=1`, since the IPC path relies on the cuMem/VMM handle path under asymmetric visibility. Registered in `test/CMakeLists.txt` and enabled in the `mi300x_mellanox_ib` test suite.

## JIRA ID

JIRA ID : ROCM-27034

## Test Plan

CI tests complition

## Test Result

In progress

## Submission Checklist

### DIFF
--- a/projects/rccl/src/init.cc
+++ b/projects/rccl/src/init.cc
@@ -1372,6 +1372,7 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, struct ncclComm* p
   int* nvbPeers = NULL;
   struct ncclProxyConnector proxyConn;
   int* pxnPeers = NULL;
+  int64_t* localBusIds = NULL;
   int *topParentLocalRanks = NULL;
   int p2pLevel = -1;
   bool globalNicFused = false;
@@ -1637,7 +1638,22 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, struct ncclComm* p
   allXgmi = true;
   hasPeerAccess = true;
 
-  if (cudaGetDeviceCount(&localDevCount) != cudaSuccess) localDevCount = 0;
+  if (CUDACLEARERROR(cudaGetDeviceCount(&localDevCount)) != cudaSuccess) {
+    WARN("cudaGetDeviceCount failed; treating all peers as non-accessible for "
+         "clique setup");
+    localDevCount = 0;
+  }
+
+  // RCCL: HIP_VISIBLE_DEVICES may differ per rank, so precompute each rank's
+  // busId so we compare them later.
+  NCCLCHECKGOTO(ncclCalloc(&localBusIds, nranks), ret, fail);
+  for (int j = 0; j < nranks; j++) {
+    int cudaDevJ = comm->peerInfo[j].cudaDev;
+    if (cudaDevJ < 0 || cudaDevJ >= localDevCount ||
+        getBusId(cudaDevJ, &localBusIds[j]) != ncclSuccess) {
+      localBusIds[j] = -1;
+    }
+  }
 
   // Check that all the GPUs have peer access to one another and are XGMI connected
   for (int i = 0; i < nranks && hasPeerAccess; i++) {
@@ -1647,17 +1663,21 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, struct ncclComm* p
       int cudaDev2 = comm->peerInfo[j].cudaDev;
       int p2p;
 
-      // RCCL: under asymmetric HIP_VISIBLE_DEVICES a peer ordinal may be out
-      // of range here; hipDeviceCanAccessPeer would return hipErrorInvalidDevice
-      // and leave a sticky error. Treat such a pair as no peer access.
+      // RCCL: HIP_VISIBLE_DEVICES may differ per rank, so check that the peer
+      // device is visible in this process and that its busId matches.
       if (cudaDev1 < 0 || cudaDev1 >= localDevCount || cudaDev2 < 0 ||
           cudaDev2 >= localDevCount) {
         hasPeerAccess = false;
         break;
       }
 
-      if (hipDeviceCanAccessPeer(&p2p, cudaDev1, cudaDev2) != hipSuccess || !p2p)
-      {
+      if (localBusIds[j] == -1 || localBusIds[j] != comm->peerInfo[j].busId) {
+        hasPeerAccess = false;
+        break;
+      }
+
+      if (hipDeviceCanAccessPeer(&p2p, cudaDev1, cudaDev2) != hipSuccess ||
+          !p2p) {
         hasPeerAccess = false;
         break;
       }
@@ -2263,6 +2283,7 @@ exit:
   free(rings);
   free(nvbPeers);
   free(pxnPeers);
+  free(localBusIds);
   return ret;
 fail:
   goto exit;
--- a/projects/rccl/src/init.cc
+++ b/projects/rccl/src/init.cc
@@ -1633,8 +1633,12 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, struct ncclComm* p
   timers[TIMER_INIT_GRAPHS] = clockNano() - timers[TIMER_INIT_GRAPHS];
 
   bool allXgmi, hasPeerAccess;
+  int localDevCount;
   allXgmi = true;
   hasPeerAccess = true;
+
+  if (cudaGetDeviceCount(&localDevCount) != cudaSuccess) localDevCount = 0;
+
   // Check that all the GPUs have peer access to one another and are XGMI connected
   for (int i = 0; i < nranks && hasPeerAccess; i++) {
     int cudaDev1 = comm->peerInfo[i].cudaDev;
@@ -1642,6 +1646,16 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, struct ncclComm* p
       if (i == j) continue;
       int cudaDev2 = comm->peerInfo[j].cudaDev;
       int p2p;
+
+      // RCCL: under asymmetric HIP_VISIBLE_DEVICES a peer ordinal may be out
+      // of range here; hipDeviceCanAccessPeer would return hipErrorInvalidDevice
+      // and leave a sticky error. Treat such a pair as no peer access.
+      if (cudaDev1 < 0 || cudaDev1 >= localDevCount || cudaDev2 < 0 ||
+          cudaDev2 >= localDevCount) {
+        hasPeerAccess = false;
+        break;
+      }
+
       if (hipDeviceCanAccessPeer(&p2p, cudaDev1, cudaDev2) != hipSuccess || !p2p)
       {
         hasPeerAccess = false;
--- a/projects/rccl/test/AsymmetricVisibilityTests.cpp
+++ b/projects/rccl/test/AsymmetricVisibilityTests.cpp
@@ -31,6 +31,8 @@
 #include <sys/wait.h>
 #include <unistd.h>
 
+#include <csignal>
+
 #include <cerrno>
 #include <cstdio>
 #include <cstdlib>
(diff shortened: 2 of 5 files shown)


amd-arozanov:
%RCCL_TEST_CONFIG
mi300x_mellanox_ib.json

projects/rccl/test/AsymmetricVisibilityTests.cpp
```diff
@@ -0,0 +1,246 @@
...
+  if (pid != 0) return pid; // parent (or fork error, reported by caller)
+
+  // Child: set the per-rank environment BEFORE HIP initializes, then re-exec
+  // a fresh image so the restricted visibility takes effect cleanly.
+  setenv("HIP_VISIBLE_DEVICES", visibleDevices, 1);
+  setenv("CUDA_VISIBLE_DEVICES", visibleDevices, 1);
+  setenv(kEnvRank, std::to_string(rank).c_str(), 1);
+  setenv(kEnvLocalDev, std::to_string(localDev).c_str(), 1);
+  setenv(kEnvUid, uidHex.c_str(), 1);
+  setenv("NCCL_CUMEM_ENABLE", "1", 0);
```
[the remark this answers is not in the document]
amd-arozanov: It is not a good idea to override NCCL_CUMEM_ENABLE. NCCL_CUMEM_ENABLE can be set by higher layer

projects/rccl/test/AsymmetricVisibilityTests.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+  ncclUniqueId id;
+  ASSERT_EQ(ncclGetUniqueId(&id), ncclSuccess);
+  const std::string uidHex = toHex(id);
+
+  // rank0: HIP_VISIBLE_DEVICES=0,1 -> bind ordinal 1; rank1: =2 -> bind ordinal 0.
+  const pid_t child0 = spawnWorker(0, "0,1", 1, uidHex);
+  ASSERT_GT(child0, 0) << "fork failed for rank 0";
+  const pid_t child1 = spawnWorker(1, "2", 0, uidHex);
+  ASSERT_GT(child1, 0) << "fork failed for rank 1";
```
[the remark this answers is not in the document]
amd-arozanov: All GPUs are available in CI

amd-arozanov:
%RCCL_TEST_CONFIG
mi300x_mellanox_ib.json

projects/rccl/src/init.cc
```diff
@@ -1633,15 +1633,30 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, struct ncclComm* p
   timers[TIMER_INIT_GRAPHS] = clockNano() - timers[TIMER_INIT_GRAPHS];
 
   bool allXgmi, hasPeerAccess;
+  int localDevCount;
   allXgmi = true;
   hasPeerAccess = true;
+
+  if (CUDACLEARERROR(cudaGetDeviceCount(&localDevCount)) != cudaSuccess)
```
nikitaxgusev: CUDACLEARERROR(cudaGetDeviceCount(&localDevCount)) clears the error and returns it, but the return value check is for success. If the macro returns non-success, localDevCount is set to 0 which silently disables
all peer access rather than propagating the error. Consider CUDACHECK or at least a WARN log when this fails.

projects/rccl/src/init.cc
```diff
@@ -1633,15 +1633,30 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, struct ncclComm* p
...
       if (i == j) continue;
       int cudaDev2 = comm->peerInfo[j].cudaDev;
       int p2p;
+
+      // RCCL: under asymmetric HIP_VISIBLE_DEVICES a peer ordinal may be out
+      // of range here; hipDeviceCanAccessPeer would return hipErrorInvalidDevice
+      // and leave a sticky error. Treat such a pair as no peer access.
+      if (cudaDev1 < 0 || cudaDev1 >= localDevCount || cudaDev2 < 0 ||
+          cudaDev2 >= localDevCount) {
+        hasPeerAccess = false;
```
nikitaxgusev, line 1670: The inner loop breaks on the first out-of-range pair, but the outer loop over i continues. A rank pair (i=0,j=out-of-range) sets hasPeerAccess=false and breaks the j-loop, then i increments, so redundant iterations follow since hasPeerAccess is already false.

projects/rccl/test/AsymmetricVisibilityTests.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+// Worker entry point: only active when re-exec'd by the orchestrator (i.e. when
+// the rank env var is present). In a normal full-suite run it skips.
+TEST(AsymmetricVisibilityWorker, Run) {
+  const char* rankEnv = getenv(kEnvRank);
+  if (rankEnv == nullptr)
+    GTEST_SKIP() << "worker entry point, driven by "
+                    "AsymmetricVisibility.CommInitRankAllReduce";
+
+  const int rank = atoi(rankEnv);
```
nikitaxgusev, line 302: atoi without validation silently returns 0 on a malformed value. Should you check iret value?

projects/rccl/test/AsymmetricVisibilityTests.cpp
```diff
@@ -0,0 +1,255 @@
...
+    GTEST_SKIP() << "requires at least 3 GPUs for an asymmetric topology";
+
+  ncclUniqueId id;
+  ASSERT_EQ(ncclGetUniqueId(&id), ncclSuccess);
+  const std::string uidHex = toHex(id);
+
+  // rank0: HIP_VISIBLE_DEVICES=0,1 -> bind ordinal 1; rank1: =2 -> bind ordinal 0.
+  const pid_t child0 = spawnWorker(0, "0,1", 1, uidHex);
+  ASSERT_GT(child0, 0) << "fork failed for rank 0";
+  const pid_t child1 = spawnWorker(1, "2", 0, uidHex);
```
nikitaxgusev: spawnWorker on child1 after child0 failure If child0 fork fails (ASSERT_GT fails), child1 is still spawned and will
hang waiting for rank 0 to connect. The orchestrator exits via the failed assertion but child1 leaks. Consider forking both before asserting, or  killing child0 on child1 fork failure.

projects/rccl/test/AsymmetricVisibilityTests.cpp
```diff
@@ -0,0 +1,255 @@
...
+  ASSERT_GT(child1, 0) << "fork failed for rank 1";
+
+  auto waitOk = [](pid_t pid, int rank) {
```
nikitaxgusev: lambda and ASSERT_* inside lambda  ASSERT_* macros inside a lambda call return from the lambda, not from the  enclosing test. The test will continue past a failed waitOk(child0, 0)  and call waitOk(child1, 1) even if child0 failed.

projects/rccl/src/init.cc
```diff
@@ -1633,15 +1633,30 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, struct ncclComm* p
...
+        break;
+      }
+
       if (hipDeviceCanAccessPeer(&p2p, cudaDev1, cudaDev2) != hipSuccess || !p2p)
```
nikitaxgusev: For checking: 

peerInfo[j].cudaDev is the device ordinal local to rank j's process. hipDeviceCanAccessPeer(cudaDev1,
cudaDev2) is called in rank i's process, where both arguments are interpreted as ordinals in rank i's
visible device set. Under asymmetric HIP_VISIBLE_DEVICES, even when cudaDev2 is within bounds, it may refer
to a completely different physical GPU than what rank j actually uses, so the P2P check tests the wrong
device pair and can silently return a wrong answer.

The bounds check at line 1654 prevents the hipErrorInvalidDevice crash, but doesn't fix the semantic
incorrectness. Wouldn't the right fix be to detect asymmetric visibility earlier (e.g. by comparing busId
fields exchanged during AllGather, peerInfo[j].busId is already available) and skip the clique P2P check
 entirely when asymmetry is detected?

projects/rccl/src/init.cc
amd-arozanov: The outer loop is already gated on hasPeerAccess: for (int i = 0; i < nranks && hasPeerAccess; i++). Once the inner loop sets it to false and breaks, the outer loop's condition fails on the next check and the scan stops, so no additional pairs are examined

projects/rccl/test/AsymmetricVisibilityTests.cpp
amd-arozanov: These env vars are not external input: the orchestrator sets them itself right before execv via std::to_string(rank/localDev), so the value is always a well-formed integer we produced

projects/rccl/src/init.cc
amd-arozanov: Added WARN log

projects/rccl/test/AsymmetricVisibilityTests.cpp
amd-arozanov: Fixed

projects/rccl/test/AsymmetricVisibilityTests.cpp
amd-arozanov: Fixed

projects/rccl/src/init.cc
amd-arozanov: Fixed. Added bus_id checks

amd-arozanov:
%RCCL_TEST_CONFIG
mi300x_mellanox_ib.json

projects/rccl/src/init.cc
```diff
@@ -1633,17 +1633,43 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, struct ncclComm* p
...
+      if (cudaDev1 < 0 || cudaDev1 >= localDevCount || cudaDev2 < 0 ||
+          cudaDev2 >= localDevCount) {
+        hasPeerAccess = false;
+        break;
+      }
+
+      int64_t busId1 = 0, busId2 = 0;
+      if (getBusId(cudaDev1, &busId1) != ncclSuccess ||
+          getBusId(cudaDev2, &busId2) != ncclSuccess ||
+          busId1 != comm->peerInfo[i].busId ||
```
nikitaxgusev: busId1 != comm->peerInfo[i].busId check seems always false and can never fire. cudaDev1 is
peerInfo[i].cudaDev the same value used in fillInfo() to compute peerInfo[i].busId via
getBusId(comm->cudaDev) ? So getBusId(cudaDev1) here returns the same value that's already
stored in peerInfo[i].busId, making the comparison a no-op that adds an unnecessary syscall on every inner
loop iteration. Could you check and remove if it is not needed?

amd-arozanov: Yes, you are right, checking busId1 is not necessary

amd-arozanov: Fixed

amd-arozanov:
%RCCL_TEST_CONFIG
mi300x_mellanox_ib.json

amd-arozanov:
%RCCL_TEST_CONFIG
mi300x_mellanox_ib.json

amd-arozanov:
%RCCL_TEST_CONFIG
mi300x_mellanox_ib.json

amd-arozanov:
%RCCL_TEST_CONFIG
mi300x_mellanox_ib.json

nikitaxgusev (approved):

PR #8140, opened 2026-07-06
[rocprofiler-compute] Fix Memory Chart headings

abchoudh-amd:
## Motivation

Align Memory Chart CLI output across RDNA3.5 and CDNA so each renderer exposes the numbered panel title, normalization scope, rendered components, and connector directions.

Bugfix: CDNA Memory Chart output now prints the `3. Memory Chart` header once outside the diagram without repeating the default normalization label.

## Technical Details

- Make `chart_title` a required arg on both renderers; `tty._mem_chart_heading` is the single heading formatter for the CLI path.
- Remove the unused `format_mem_chart_heading` helper and the `normal_unit` parameter from `plot_mem_chart`.
- Move CDNA chart title outside the diagram.
- Add architecture component and connector assertions.
- Add TTY renderer dispatch coverage.

## JIRA ID

- JIRA ID: AIROCVAL-168
- JIRA ID: AIPROFCOMP-726

## Test Plan

- Run `pytest tests/test_mem_chart.py`.
- Run `pytest tests/test_tty.py`.

## Test Result

- [x] `pytest tests/test_mem_chart.py` passes
- [x] `pytest tests/test_tty.py` passes

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-compute/CHANGELOG.md
+++ b/projects/rocprofiler-compute/CHANGELOG.md
@@ -44,6 +44,8 @@ Full documentation for ROCm Compute Profiler is available at [https://rocm.docs.
 
 * Fixed multi-user roofline benchmarking on shared systems: the per-GPU lock file under `/tmp/rocprof-compute-benchmark/` is now created world-readable/writable (0666) so any user can acquire it, regardless of which user created it first or the active umask. Stale unreadable lock files left by older versions in a sticky `/tmp` cannot be repaired automatically and must be removed manually by their owner or an administrator.
 
+* Fixed CDNA memory chart CLI output to show the numbered `3. Memory Chart` header without repeating the default per-kernel normalization label.
+
 ### Upcoming changes
 
 ### Known issues
--- a/projects/rocprofiler-compute/src/utils/mem_chart_gfx11.py
+++ b/projects/rocprofiler-compute/src/utils/mem_chart_gfx11.py
@@ -12,7 +12,6 @@ USAGE:
 API:
     normalize_mem_chart_metrics(metric_dict) -> flat ordered dict for UIs
     plot_mem_chart(..., *, chart_title=...) -> str
-    format_mem_chart_heading(normal_unit, *, panel_id=300, section_label=...) -> str
 
 Metric dict keys must match the Memory Chart panel YAML for RDNA3.5:
 
@@ -242,21 +241,6 @@ def _print_mem_chart_scope_bar(console: Console) -> None:
         + "-" * 4
         + "|"
     )
-
-
-def format_mem_chart_heading(
-    normal_unit: str,
-    *,
-    panel_id: int = 300,
-    section_label: str = "Memory Chart",
-) -> str:
-    """Build CLI diagram title: ``{panel_id//100}. {label} (Normalization: …)``.
-
-    Matches other panels (e.g. ``3. System Speed-of-Light``) where the leading
-    number is ``Panel Config id // 100`` (panel 300 → ``3.``).
-    """
-    section = max(0, int(panel_id)) // 100
-    return f"{section}. {section_label} (Normalization: {normal_unit})"
 
 
 def normalize_mem_chart_metrics(metric_dict: dict[str, Any]) -> dict[str, Any]:
@@ -740,7 +724,7 @@ def plot_mem_chart(
     normal_unit: str,
     metric_dict: dict[str, Any],
     *,
-    chart_title: Optional[str] = None,
+    chart_title: str,
 ) -> str:
     """Plot the memory chart and return as string.
 
@@ -748,22 +732,16 @@ def plot_mem_chart(
     ``MEM_CHART_PANEL_METRIC_KEYS``. Values for bandwidth metrics are in **Bytes/s**.
     Input is normalized to a flat ordered dict before rendering.
 
-    ``chart_title``: full heading line; if omitted, uses ``format_mem_chart_heading``
-    with ``panel_id=300`` (section ``3.``).
+    ``chart_title``: full heading line printed above the diagram.
     """
     flat = normalize_mem_chart_metrics(metric_dict)
-    resolved_heading = (
-        format_mem_chart_heading(normal_unit, panel_id=300)
-        if chart_title is None
-        else chart_title
-    )
     buf = StringIO()
     console = Console(file=buf, force_terminal=True, width=200, height=80)
     create_mem_chart_diagram(
         flat,
         console,
         show_debug=False,
-        chart_title=resolved_heading,
+        chart_title=chart_title,
     )
     return buf.getvalue()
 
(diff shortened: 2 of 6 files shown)


feizheng10 (approved):

vedithal-amd (changes requested):
blocking for review

projects/rocprofiler-compute/tests/test_mem_chart_cli.py
```diff
@@ -0,0 +1,138 @@
...
+    "tests/workloads/vcopy/MI300X_A1",
+    "tests/workloads/vcopy/MI350",
+    "tests/workloads/vcopy/RDNA35_HALO",
+]
+
+
+class TestMemChartCLI:
+    """CLI integration tests for Memory Chart analysis."""
+
+    def test_memory_chart_runs_for_supported_workloads(
```
vedithal-amd: Drop this. Exit-0 for block 3 over these workloads is already covered by `test_analyze_workload` in `test_analyze_workloads.py` (full analyze over a superset) and `test_filter_block_3` in `test_analyze_commands.py` (same `indirs` list).

projects/rocprofiler-compute/tests/test_mem_chart_cli.py
```diff
@@ -0,0 +1,138 @@
...
+                    f"{expected_label}"
+                )
+        finally:
+            common.clean_output_dir(config["cleanup"], workload_dir)
+
+    @pytest.mark.parametrize(
+        "normal_unit",
+        ["per_wave", "per_cycle", "per_second", "per_kernel"],
+    )
+    def test_memory_chart_accepts_normal_unit(
```
vedithal-amd: Drop this. `test_normal_unit_per_wave/per_cycle/per_second/per_kernel` in `test_analyze_commands.py` already loop all 4 units over the same 6 vcopy workloads.

projects/rocprofiler-compute/tests/test_mem_chart_cli.py
```diff
@@ -0,0 +1,138 @@
...
+                    "GL2 Cache",
+                    "SQC",
+                    "GCEA",
+                    "DRAM",
+                ],
+                id="RDNA35_HALO",
+            ),
+        ],
+    )
+    def test_memory_chart_output_includes_memory_hierarchy(
```
vedithal-amd: The asserted labels are static diagram scaffolding drawn regardless of data, so this passes even if every metric is N/A. As written it is not worth its runtime over the exit-code coverage above. If it stays, it needs to assert enough of the scaffolding (headings, block frames, wires) to be a genuine render check. Metric-value correctness is a separate concern, out of scope here.

vedithal-amd (changes requested):
Most of this suite duplicates existing analyze integration coverage. See inline comments.

projects/rocprofiler-compute/tests/test_mem_chart_cli.py
abchoudh-amd: `test_filter_block_3` covers block `5.2.2`. The test numbers do not correspond to the block but serial numbers.

`test_analyze_workload` doesn't test for  `--block` filtering

projects/rocprofiler-compute/tests/test_mem_chart_cli.py
abchoudh-amd: This test combines `--normal_unit` with `--block`.

projects/rocprofiler-compute/tests/test_mem_chart_cli.py
abchoudh-amd: This test neither intends to check the correctness of the metric data nor the specifics of the diagram. It's sole purpose is to verify that the correct memory hierarchy components are present in the memory chart.

projects/rocprofiler-compute/tests/test_mem_chart_cli.py
vedithal-amd: `test_analyze_workload` (test_analyze_workloads.py:105) runs unfiltered analyze over a superset of these workloads, which renders every panel including Memory Chart (panel 300) and asserts exit 0. This test only asserts `code == 0` with `--block 3`, so it adds no assert and no code path beyond what's already covered. The single-integer filter path is also already exercised by `test_filter_block_1`/`_5` (test_analyze_commands.py:192,256). Any real value would come from block-3-specific asserts, which is the concern in the other thread. Suggest dropping this one.

projects/rocprofiler-compute/tests/test_mem_chart_cli.py
vedithal-amd: `test_normal_unit_per_wave/cycle/second/kernel` (test_analyze_commands.py:449-512) run full analyze under all 4 units over these same workloads, which already renders the Memory Chart panel (300) under every normalization unit. `--block` and `--normal-unit` are orthogonal (panel selection vs metric denominator), and the existing tests pass `--normal-unit` without `--block`, so the combination adds no new code path. This test only asserts `code == 0`, so it adds no assertion either. If there's a specific normalization interaction worth guarding, a targeted unit test would cover it more cheaply. Suggest dropping this one.

projects/rocprofiler-compute/tests/test_mem_chart_cli.py
vedithal-amd: Keeping this one makes sense — it's the only test in the suite with assertions beyond exit code. Two things to make it a real render check rather than a partial one:

1. The asserted labels are a subset. The gfx9 renderer (`mem_chart_gfx9.py`) draws Exec, Instr Buff, Instr Dispatch, GDS, LDS, Vector/Scalar/Instr L1, L2 Cache, Fabric, HBM, GMI, xGMI/PCIe; this asserts 7 of them. gfx11 (`mem_chart_gfx11.py`) additionally has a Kernel box that isn't asserted. Since every label is a static constant drawn unconditionally, a partial set passes even if whole boxes are removed. Suggest asserting the complete set of component boxes per renderer.

2. The two params already correctly cover both scaffolding variants (gfx9 for MI300X_A1, gfx11/gfx115x for RDNA35_HALO), so the structure is right — just complete the label set for each.

vedithal-amd:
@xuchen-amd  and @feizheng10  could you please have a look at the memory chart scaffolding tests here - do you think it adds value? how can we improve this? Any other ideas on how to test memory chart?

xuchen-amd:
> @xuchen-amd and @feizheng10 could you please have a look at the memory chart scaffolding tests here - do you think it adds value? how can we improve this? Any other ideas on how to test memory chart?

replied offline.

vedithal-amd (approved):
Approving. Pushed the heading cleanup we discussed:

- `chart_title` is now a required arg on both `plot_mem_chart` renderers, with `tty._mem_chart_heading` as the single heading formatter for the CLI path.
- Removed the now-unused `mem_chart_gfx11.format_mem_chart_heading` helper (its last use, `main()`, is now an inline f-string) and the `normal_unit` parameter, which only existed to build the old default heading.
- Dropped the dead empty-title branch in gfx9.
- Tests: parametrized the `chart_title` override across both renderers, merged the two gfx9 heading/connector tests, renamed tests by arch, and reworked the tty dispatch case to derive the renderer via `is_gfx115x` with monkeypatch stubs (no `Mock`). Simplified the LDS check to a plain membership assert.

PR #8141, opened 2026-07-06
fix(rocprofiler-sdk): prevent roctxMark from leaking kernel renames

bgopesh:
Root cause:                                                                                                                                                                                                                                                                                                                                                                                                                              
  Under --kokkos-trace (which implies --kernel-rename), the KokkosP connector emitted roctxMark for its init/finalize/profile_event milestones. rocprofv3 treats roctxMark (roctxMarkA) as a sticky kernel rename:               
  kernel_rename_callback pushes the mark's message onto the thread-local rename stack on the EXIT phase and never pops it (a mark is instantaneous). As a result, every subsequent kernel dispatched on that thread was relabeled
   to the last mark's message. This made all non-Kokkos (raw HIP) kernels "disappear" from --kernel-trace, showing up as e.g. Kokkos::Finalization Complete instead of their real names. 

Fix:
Rather than change the generic roctxMark rename semantics — which other tools and integration tests (kernel-rename, summary) intentionally rely on to label groups of kernels — the fix is on the Kokkos side. The three       
  milestone marks are emitted as a balanced roctxRangePush + roctxRangePop pair (kokkosp_emit_milestone). Nothing is dispatched between the push and pop, so the rename stack is left net-unchanged while the milestone still    
  appears in the marker timeline. Kokkos parallel regions continue to rename their kernels via their existing roctxRangePush/roctxRangePop scopes, and non-Kokkos kernels retain their real names. 
## Motivation


<!-- Explain the purpose of this PR and the goals it aims to achieve. -->

## Technical Details

<!-- Explain the changes along with any relevant GitHub links. -->

## JIRA ID
JIRA ID : ROCM-27097
<!-- If applicable, mention the JIRA ID resolved by this PR (Example: Resolves SWDEV-12345). -->
<!-- Do not post any JIRA links here. -->

## Test Plan

<!-- Explain any relevant testing done to verify this PR. -->

## Test Result

<!-- Briefly summarize test outcomes. -->

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/kokkosp/kokkosp.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/kokkosp/kokkosp.cpp
@@ -60,6 +60,19 @@ struct Section
 
 bool tool_globfences  = false;
 auto kokkosp_sections = std::vector<Section>{};
+
+// Emit an instantaneous Kokkos milestone into the marker timeline without perturbing kernel
+// renaming. rocprofv3's --kernel-rename (implied by --kokkos-trace) treats roctxMark as a
+// sticky rename that relabels every subsequent kernel dispatched on the thread. Using a
+// balanced roctxRangePush/roctxRangePop instead keeps the timeline marker but leaves the
+// kernel-rename stack net-unchanged (nothing is dispatched between the push and pop), so
+// non-Kokkos kernels retain their real names under --kokkos-trace.
+void
+kokkosp_emit_milestone(const char* name)
+{
+    roctxRangePush(name);
+    roctxRangePop();
+}
 }  // namespace
 
 extern "C" {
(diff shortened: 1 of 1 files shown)


rpathani (approved):
I went through this to check my understanding — please correct me if I've got anything wrong:

The Kokkos milestones (like "Initialization Complete") were being announced in a way that acts like a permanent label — once set, it stuck to every GPU kernel that ran afterward. So under kernel-rename, real kernels were showing up with the milestone's name instead of their own.
The fix announces the milestone as a label that's turned on and off right away, with nothing running in between. So the milestone still shows up on the timeline, but it no longer sticks to the kernels that come after it.
And this was done only on the Kokkos side, without changing the general "sticky label" behavior, since other tools depend on that behavior on purpose.
**With this understanding, this looks good to me.** 
Currently we don't have any test that runs under --kokkos-trace — the only coverage is the --kokkos-trace=no disable case in negate-aggregate-tracing-options. Since running a full Kokkos app is heavy for CI, we could consider a lightweight test as a follow-up. Happy to discuss.

jrmadsen (approved):

arvindcheru:
CI Test Triage: 
Request to ignore rocprofiler-compute test-pc-sampling-collector failure — known LD_LIBRARY_PATH/sysdeps issue, tracked in ROCm/rocm-systems#8186 (not introduced by this PR).

PR #8143, opened 2026-07-06
test(rocprofv3): expand cli-multipass multi-pass PMC counter-collection coverage

minseobshin11:
## Motivation

rocprofiler-sdk's `rocprofv3` **cli-multipass** counter-collection test only exercised two `--pmc` passes via CLI flags, against hardcoded `pass_1`/`pass_2` fixtures. It did not cover the **input-file** path for defining pass groups, the **CLI + input-file combined** path, or the **error paths** for illegal option combinations, and it did not verify that pass output directories are contiguously 1-indexed (`pass_1/`, `pass_2/`, … with no `pass_0`). There were also no GPU-free tests for the launcher's own parsing logic.

This PR closes those coverage gaps. It contains **no product changes** — `source/bin/rocprofv3.py` is byte-identical to `develop`. The product fix that previously lived here has moved to #9795, stacked on this branch, so the behavior change can be reviewed on its own.

## Technical Details

**Integration coverage (`counter-collection/cli-multipass/`)**
- Adds a third `--pmc` group (`GRBM_GUI_ACTIVE`) so the CLI test spans three passes.
- Adds JSON- and YAML-input multi-pass tests (`-i input.json`, `-i input.yml`), one counter per pass.
- Adds a **combined** test: two CLI `--pmc` groups plus one input-file group (`input-combined.json`) in a single run → three passes, verifying CLI groups run first and the input-file group last.
- Adds three **negative** tests for the illegal combinations, asserting the exact message via `PASS_REGULAR_EXPRESSION` rather than `WILL_FAIL`, which would accept any non-zero exit: multiple `--pmc` + `--pid`; multiple `--pmc` + `--collection-period`; `--pmc` + input-file `pmc_groups` (`input-pmc-groups.json`).
- Adds `validate_no_output.py`, asserting the rejected runs leave behind no output directory, no `pass_*` subdirectory and no CSV. It carries a positive control so a wrong `--binary-dir` fails loudly instead of passing vacuously.
- Parametrizes the validator: it takes an ordered `--expected-counters` list and derives the number of passes from its length, validating `pass_i` against `expected_counters[i-1]`. Adding or removing a pass is a CMake-only change, with no edits to `validate.py`/`conftest.py`.
- Adds `test_pass_count` asserting the `pass_*` directories are contiguously 1-indexed, exactly one per expected counter, with no `pass_0`; generalizes the agent-info, per-pass counter, dispatch-consistency and counter-separation checks from two passes to N.

**GPU-free unit tests (`counter-collection/cli-multipass-unit/`)**
- New suite that imports the `rocprofv3` launcher as a module and exercises its parsing helpers with no GPU and no application launch: single- vs multi-pass `--pmc` normalization, and `parse_input` for JSON/YAML (→ `pass_`) and the deprecated text format (→ `pmc_`).

## Issue Tracking

JIRA ID: AIROCVAL-59

## Test Plan

```bash
cmake -S projects/rocprofiler-sdk -B build -DCMAKE_PREFIX_PATH=/opt/rocm \
      -DROCPROFILER_BUILD_TESTS=ON
cmake --build build --parallel
ctest --test-dir build -R "cli-multipass" --output-on-failure
ctest --test-dir build -R "counter-collection" --output-on-failure
```

## Test Result

Ran on gfx942 (AMD Instinct MI300X):
- `cli-multipass`: 35/35 pass. `counter-collection`: 65/65 pass, no regressions.
- CLI (3-pass), JSON-input, YAML-input and combined multi-pass all produce contiguously 1-indexed `pass_1..N` directories with pairwise-disjoint counters and identical dispatch IDs across passes.
- The three negative tests fail fast with a non-zero exit code and the exact human-readable message, and create no output directories.
- The GPU-free unit tests pass without a GPU.

## Submission Checklist


projects/rocprofiler-sdk/tests/rocprofv3/counter-collection/cli-multipass/validate.py
meserve-amd: Nit: Copyright year here and in other modified files

projects/rocprofiler-sdk/tests/rocprofv3/counter-collection/cli-multipass/CMakeLists.txt
```diff
@@ -53,7 +57,156 @@ rocprofiler_add_integration_validate_test(
...
+rocprofiler_add_integration_execute_test(
+    rocprofv3-test-counter-collection-cli-multipass-json
+    COMMAND
+        $<TARGET_FILE:rocprofiler-sdk::rocprofv3> -i
+        ${CMAKE_CURRENT_BINARY_DIR}/input.json -d
+        ${CMAKE_CURRENT_BINARY_DIR}/%argt%-cc-json -o out -f csv --
+        $<TARGET_FILE:simple-transpose>
+    DEPENDS simple-transpose
+    TIMEOUT 45
+    LABELS "integration-tests"
```
meserve-amd: Nit: This label is already added by the add_integration macro, so can be dropped. (here and in other tests)

projects/rocprofiler-sdk/tests/rocprofv3/counter-collection/cli-multipass/CMakeLists.txt
```diff
@@ -53,7 +57,156 @@ rocprofiler_add_integration_validate_test(
...
+    PRELOAD "${PRELOAD_ENV}"
+    FIXTURES_SETUP rocprofv3-test-counter-collection-cli-multipass-json)
+
+rocprofiler_add_integration_validate_test(
+    rocprofv3-test-counter-collection-cli-multipass-json
+    TEST_PATHS validate.py
+    COPY conftest.py input.json input.yml
```
meserve-amd: Minor: is this COPY already done by rocprofiler_configure_pytest_files ?

projects/rocprofiler-sdk/tests/rocprofv3/counter-collection/cli-multipass/CMakeLists.txt
```diff
@@ -53,7 +57,156 @@ rocprofiler_add_integration_validate_test(
...
+    rocprofv3-test-counter-collection-cli-multipass-rejects-pid
+    COMMAND
+        $<TARGET_FILE:rocprofiler-sdk::rocprofv3> --pmc SQ_WAVES --pmc GRBM_COUNT --pid
+        12345 -d ${CMAKE_CURRENT_BINARY_DIR}/%argt%-cc-err-pid -o out -f csv --
+        $<TARGET_FILE:simple-transpose>
+    DEPENDS simple-transpose
+    TIMEOUT 45
+    LABELS "integration-tests"
+    PRELOAD "${PRELOAD_ENV}"
+    PASS_REGULAR_EXPRESSION "multiple --pmc flags. is not compatible with attach mode")
```
meserve-amd: Major: Here and in other regex, avoid using `.` when the exact character is known. You can escape characters to match exactly with \ i.e. `multiple --pmc flags\) is not`

meserve-amd (approved):
Interestingly, this is the first instance of unit testing being added for rocprofv3. I like it.

projects/rocprofiler-sdk/tests/rocprofv3/counter-collection/cli-multipass/validate.py
minseobshin11: fixed

projects/rocprofiler-sdk/tests/rocprofv3/counter-collection/cli-multipass/CMakeLists.txt
minseobshin11: fixed

projects/rocprofiler-sdk/tests/rocprofv3/counter-collection/cli-multipass/CMakeLists.txt
minseobshin11: Got it! Fixed it

projects/rocprofiler-sdk/tests/rocprofv3/counter-collection/cli-multipass/CMakeLists.txt
minseobshin11: You are right, they are unnecessary. Removed!

projects/rocprofiler-sdk/source/bin/rocprofv3.py
```diff
@@ -1300,6 +1300,31 @@ def has_set_attr(obj, key):
...
+            f"Multi-pass counter collection ({source}) is not compatible "
+            "with attach mode (--pid)",
+        )
+    if collection_period:
+        return (
+            "--collection-period",
+            f"Multi-pass counter collection ({source}) is not compatible "
+            "with --collection-period",
+        )
+    return None
```
bgopesh: source is derived from one boolean but use_multipass has three causes. The combined case (cli_has_pmc and input_has_pmc with a single input job) falls into the else and reports "multiple input-file jobs" for a file that has exactly one job — pointing the user at the wrong thing.

 Also: element [0] of the returned tuple is never read by production code, only by the unit tests, and the three-way condition is now stated twice across the two helpers.

Returning the reason instead of a bool collapses both helpers' shared condition into one place and fixes the message. The if order matches the original or short-circuit exactly, so the truth value is provably unchanged.

```suggestion
def multipass_source(cli_multipass, num_input_jobs, cli_has_pmc, input_has_pmc):
    """Return why this run is multi-pass, or None if it is single-pass.

    Every case here means the application is launched more than once: multiple
    --pmc flags, more than one input-file job, or CLI --pmc groups concatenated
    with input-file groups.
    """
    if cli_multipass:
        return "multiple --pmc flags"
    if num_input_jobs > 1:
        return "multiple input-file jobs"
    if cli_has_pmc and input_has_pmc:
        return "--pmc combined with input-file pmc"
    return None


def multipass_incompatible_message(source, pid, collection_period):
    """Rejection message when multi-pass is combined with --pid/--collection-period."""
    if source is None:
        return None
    if pid:
        return (
            f"Multi-pass counter collection ({source}) is not compatible "
            "with attach mode (--pid)"
        )
    if collection_period:
        return (
            f"Multi-pass counter collection ({source}) is not compatible "
            "with --collection-period"
        )
    return None


def first_set_attr(objs, key):
    """First non-None value of `key` across `objs`, or None if none set it."""
    for itr in objs:
        if has_set_attr(itr, key):
            return getattr(itr, key)
    return None
```

projects/rocprofiler-sdk/source/bin/rocprofv3.py
```diff
@@ -2338,16 +2363,18 @@ def main(argv=None):
...
-    if cli_multipass and cmd_args.collection_period:
-        fatal_error(
-            "Multi-pass counter collection (multiple --pmc flags) is not compatible with --collection-period"
-        )
+    # multi-pass reruns the app per pass; reject --pid / --collection-period
+    _incompatible = multipass_incompatible_option(
+        use_multipass, cli_multipass, cmd_args.pid, cmd_args.collection_period
+    )
+    if _incompatible is not None:
+        fatal_error(_incompatible[1])
```
bgopesh: The guard reads cmd_args only, but inp_args entries are merged into the effective args by get_args(cmd_args, pass_config["config"]) at line 2482, and any argparse dest can be set from the input file. So this still slips through:

 {"jobs":[{"pmc":["SQ_WAVES"],"pid":12345},{"pmc":["GRBM_COUNT"]}]}

 use_multipass is True, cmd_args.pid is None → guard passes → reaches run() and line 2029 elif args.pid: update_env("ROCPROF_ATTACH_PID", args.pid). Attach mode, executed twice. Same for collection_period.

This is the same class of bug the PR closes, one level down: the PR broadened which multi-pass sources count but not which config sources count. Checking the merged view puts the guard and the thing it protects on the same input.
```suggestion
   cli_has_pmc = has_set_attr(cmd_args, "pmc")
    input_has_pmc = len(inp_args) > 0 and has_set_attr(inp_args[0], "pmc")
    multipass_reason = multipass_source(
        cli_multipass, len(inp_args), cli_has_pmc, input_has_pmc
    )
    use_multipass = multipass_reason is not None

    # multi-pass reruns the app once per pass; attach mode and periodic collection
    # cannot be replayed, so reject them wherever they were specified
    incompatible = multipass_incompatible_message(
        multipass_reason,
        cmd_args.pid or first_set_attr(inp_args, "pid"),
        cmd_args.collection_period or first_set_attr(inp_args, "collection_period"),
    )
    if incompatible is not None:
        fatal_error(incompatible)
```

bgopesh (changes requested):
The title is test(rocprofv3): expand … coverage, but this changes which CLI invocations the tool rejects — rocprofv3 -i two_jobs.json --pid 12345 -- ./app worked before and now exits non-zero.

minseobshin11:
> The title is test(rocprofv3): expand … coverage, but this changes which CLI invocations the tool rejects — rocprofv3 -i two_jobs.json --pid 12345 -- ./app worked before and now exits non-zero.

Hi, I've split the changes into test and fix and this PR now contains only test changes. On --pid, I don't think it worked before (using-rocprofv3.rst:1137 states "Multi-pass counter collection is not compatible with attach mode (--pid)" for all multi-pass, not just the CLI form). And on develop the command is accepted but does nothing useful -LD_PRELOAD is skipped whenever --pid is set, so the application launches uninstrumented while rocprof-attach targets a different PID.

projects/rocprofiler-sdk/source/bin/rocprofv3.py
minseobshin11: fixed the things that you have mentioned.

projects/rocprofiler-sdk/source/bin/rocprofv3.py
minseobshin11: applied the suggestion

bgopesh (changes requested):
Unit tests are located in a tests subdirectory of the units being tested." 
"cli-multipass-unit" test source/bin/rocprofv3.py, so their home ideally should be source/bin/tests/. 

Placing them under tests/rocprofv3/counter-collection/ forces registration through rocprofiler_add_integration_validate_test, which unconditionally appends integration-tests and validation to LABELS (tests/common/CMakeLists.txt:448-454). A parser test that needs no GPU and launches no application is now labeled as an integration test and filed under counter-collection.

bgopesh (approved):
LGTM now. Please make sure you see all the runners passing the tests before merging.

PR #8144, opened 2026-07-06
[rocprofiler-sdk] Enable ATT double buffering for gfx11.5

EvanPhoukong:
## Motivation

Enable ATT double buffering on gfx11 and gfx11.5, reusing the gfx12 design wherever
the hardware allows.

## Technical Details

Most of this follows gfx12 directly and is now shared rather than duplicated: STATUS2
buffer-full detection, buffer-swap and ring handling, skipping disabled shader engines,
and resetting the write pointer at start.

The parts gfx11 needed that gfx12 doesn't:

**Packed high address bits.** gfx12 has a separate `BASE_HI` register per buffer; gfx11
packs those bits into `SIZE`. The last write is what commits the address, so gfx11 writes
`BASE` first and `SIZE` last.

**Buffer registers only latch while the trace is off.** Reprogramming these registers under a live
trace leaves it running on the old configuration/addresses, so `Begin()` disables the trace and
waits for it to go idle first.

**GFXIP Changes - Liveness check** gfx11 can stop tracing partway through a dispatch. STATUS2 only
reports full/not-full, so a stalled trace is indistinguishable from one that hasn't
filled a buffer yet. This means the poll loop waits forever and returns nothing. The status packet
now also reads the owner id from STATUS, which is set while a trace is running and
cleared when it stops, and the producer saves the partial data and restarts the trace
when it sees that. Exposed as `GetTraceOwnerMask()`, which returns 0 on architectures
without the field, so no other target changes behavior.

**THREAD_TRACE_FINISH.** The gfx11 event-write packet was a stub; it's implemented so the
stop sequence can wait for the trace to drain.

**core.cpp** The producer thread must be polling before the trace is enabled,
otherwise buffers can fill before anything is watching them.

## JIRA ID

JIRA ID: AIPROFSDK-858

## Test Plan

Built rocprofiler-sdk and the ATT test targets on gfx1151, then ran:

- the thread-trace and aqlprofile unit tests
- the ATT API tests, including every triple-buffer case, 25 times each
- the rocprofv3 ATT integration and validation tests

## Test Result

- Unit tests: 101/101 passed.
- ATT API tests, 25x each: 300/300 passed.
- ATT integration tests: 32/32 passed.

## Submission Checklist


projects/rocprofiler-sdk/source/lib/aqlprofile/core/aql_profile.cpp
```diff
@@ -863,12 +863,15 @@ hsa_ven_amd_aqlprofile_iterate_data(const hsa_ven_amd_aqlprofile_profile_t* prof
                 // Check if SQTT buffer was wrapped
                 for(size_t se_index = 0; se_index < se_number_total; se_index++)
                 {
+                    auto full_status = sqttbuilder->SupportsDoubleBuffer()
```
ApoKalipse-V: aql_profile.cpp is not used

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/threading.cpp
```diff
@@ -96,6 +96,9 @@ copy_data_sync(void*         dst,
     // Workaround for ROCM-25606
     if(dependency) signal_wait(*dependency);
 
+    // A 0-byte async copy can wedge on gfx11.5 (completion signal never raised); skip it.
```
ApoKalipse-V: Is this a bug or are 0-byte copies not allowed by HSA?
Why are 0-byte copies present?

projects/rocprofiler-sdk/source/lib/aqlprofile/pm4/sqtt_builder.h
```diff
@@ -252,7 +268,11 @@ class GpuSqttBuilder
         // to 4KB per thread trace specification
         const uint64_t se_number_xcc = se_number_total / GetXCCNumber();
         uint64_t       base_addr     = reinterpret_cast<uint64_t>(config->data_buffer_ptr);
-        if(Primitives::GFXIP_LEVEL == 10 || Primitives::GFXIP_LEVEL == 11)
+        // gfx10/gfx11 single-buffer mode reserves a minimal stub per disabled SE. In double-buffer
```
ApoKalipse-V: You can remove this, or request more memory instead if it doesn't work.

projects/rocprofiler-sdk/source/lib/aqlprofile/pm4/sqtt_builder.h
```diff
@@ -471,24 +491,48 @@ class GpuSqttBuilder
...
-                    // If we are in double buffer mode
-                    if(!config->buffer_data.empty())
+                    // If we are in double buffer mode. Only masked-in SEs have a second buffer in
+                    // buffer_data; on gfx11 disabled SEs are not skipped above
+                    // (capacity_per_disabled_se is non-zero) so they must not index into
+                    // buffer_data here.
+                    if(bMaskedIn && !config->buffer_data.empty())
                     {
-                        if(Primitives::GFXIP_LEVEL != 12) throw std::runtime_error("Not supported");
+                        if(!SupportsDoubleBuffer())
```
ApoKalipse-V: You can just change to != 10

projects/rocprofiler-sdk/source/lib/aqlprofile/pm4/sqtt_builder.h
```diff
@@ -471,24 +491,48 @@ class GpuSqttBuilder
...
+                        {
+                            // gfx11 packs BASE_HI into the SIZE register and has no separate
+                            // BUF1_BASE_HI register, so program BUF1_SIZE (size + BASE_HI) then
+                            // BUF1_BASE (low bits).
+                            WriteConfigPacket(
+                                cmd_buffer,
+                                Primitives::SQ_THREAD_TRACE_BUF1_SIZE_ADDR,
+                                Primitives::sqtt_buffer_size_value(sqtt_size, buff1_hi));
+                            builder.BuildWriteWaitIdlePacket(cmd_buffer);
+                            WriteConfigPacket(cmd_buffer,
```
ApoKalipse-V: Program this one first, since its the same for gfx11 and gfx12. You dont need it in both if and else conditions.

projects/rocprofiler-sdk/source/lib/aqlprofile/pm4/sqtt_builder.h
```diff
@@ -471,24 +491,48 @@ class GpuSqttBuilder
...
+                                              Primitives::SQ_THREAD_TRACE_BUF1_BASE_HI_ADDR,
+                                              buff1_hi);
+                        }
+                        else
+                        {
+                            // gfx11 packs BASE_HI into the SIZE register and has no separate
+                            // BUF1_BASE_HI register, so program BUF1_SIZE (size + BASE_HI) then
+                            // BUF1_BASE (low bits).
+                            WriteConfigPacket(
```
ApoKalipse-V, line 538: This is also common in both

projects/rocprofiler-sdk/source/lib/aqlprofile/pm4/sqtt_builder.h
```diff
@@ -824,6 +870,24 @@ class GpuSqttBuilder
...
         }
+        else if(Primitives::GFXIP_LEVEL == 11)
+        {
+            // gfx11 has no separate BUF0 base register (buf0 uses SQ_THREAD_TRACE_BASE/SIZE) and
+            // packs the high address bits into the SIZE register, so reprogram the SIZE register
+            // (size + BASE_HI) and the low base register for the buffer being swapped.
+            const unsigned base_lo = Low32(base_addr >> Primitives::TT_BUFF_ALIGN_SHIFT);
+            const unsigned base_hi = High32(base_addr >> Primitives::TT_BUFF_ALIGN_SHIFT);
+
+            auto reg_base = buf1 ? Primitives::SQ_THREAD_TRACE_BUF1_BASE_LO_ADDR
```
ApoKalipse-V, line 895: Same comments as the Start() packet

ApoKalipse-V (changes requested):

projects/rocprofiler-sdk/source/lib/aqlprofile/core/aql_profile.cpp
EvanPhoukong: Removed SupportsDoubleBuffer().

projects/rocprofiler-sdk/source/lib/aqlprofile/pm4/sqtt_builder.h
EvanPhoukong: Removed. Should mirror gfx12 now, disabled SEs are skipped.

projects/rocprofiler-sdk/source/lib/aqlprofile/pm4/sqtt_builder.h
EvanPhoukong: Done.

projects/rocprofiler-sdk/source/lib/aqlprofile/pm4/sqtt_builder.h
EvanPhoukong: Done. BASE_LO is written before the BASI_HI/SIZE latch.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/threading.cpp
EvanPhoukong: When stopping thread trace, a final drain will happen even if the last full buffer was already emptied. This 0-byte copy will sometimes hang on gfx11.5. This just treats the 0-byte copy as no-op for safety.

projects/rocprofiler-sdk/source/lib/aqlprofile/pm4/sqtt_builder.h
EvanPhoukong: The BASE_LO write is no longer duplicated, only the final latch differs.

projects/rocprofiler-sdk/source/lib/aqlprofile/pm4/sqtt_builder.h
EvanPhoukong: Fixed.

projects/rocprofiler-sdk/source/lib/aqlprofile/pm4/sqtt_builder.h
```diff
@@ -446,12 +469,26 @@ class GpuSqttBuilder
                     {
                         const uint32_t sqtt_reg_size =
                             Primitives::sqtt_buffer_size_value(sqtt_size, baddr_hi);
-                        // Program size of buffer to use for thread trace
-                        WriteConfigPacket(
-                            cmd_buffer, Primitives::SQ_THREAD_TRACE_SIZE_ADDR, sqtt_reg_size);
-                        // Program base address of buffer to use for thread trace
-                        WriteConfigPacket(
-                            cmd_buffer, Primitives::SQ_THREAD_TRACE_BASE_ADDR, baddr_lo);
+                        if(Primitives::GFXIP_LEVEL == 11)
```
ApoKalipse-V, line 485: I dont think we need the if/else separation here, as only one item is different.

projects/rocprofiler-sdk/source/lib/aqlprofile/aql_profile_v2.h
```diff
@@ -627,6 +627,9 @@ typedef struct aqlprofile_att_buffer_status_t
     bool     is_too_late;
     bool     error;
     uint64_t read_offset;
+    // Set when the hardware is no longer running the trace; the caller must restart it to
+    // collect more data. Only meaningful if _size covers this field.
+    bool trace_stopped;
```
ApoKalipse-V, line 632: Is this different than the "error" or "is_too_late" bits?

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
```diff
@@ -343,6 +348,28 @@ ThreadTracerAgent::start_thread_trace(std::shared_ptr<std::atomic<int>> _flag)
...
+        // Start the worker first, then enable the trace, and do not return to the application
+        // until the producer has finished its first status poll.
+        while(!worker_data->producer_waiting.load(std::memory_order_acquire))
+            std::this_thread::yield();
+
+        // The producer is waiting on this signal, so it must be raised even when the trace is
+        // already shutting down, or that thread would never exit.
+        att_queue_submit_signal_last(*queue, start_packets, *shared_signal);
+
+        while(!worker_data->producer_ready.load(std::memory_order_acquire) &&
```
ApoKalipse-V: If we already submitted the packet, do we need this?

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
```diff
@@ -343,6 +348,28 @@ ThreadTracerAgent::start_thread_trace(std::shared_ptr<std::atomic<int>> _flag)
...
+              worker_flag->load() == WORKER_FLAG_RUNNING)
+            std::this_thread::yield();
+    }
+    else
```
ApoKalipse-V: To avoid this big if/else, maybe we can move the producer initialization earlier

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/threading.cpp
```diff
@@ -335,6 +346,17 @@ producer_loop(
             do_sleep = false;
             submit_wait_timeout();
         }
+        else if(buffer_packet.trace_stopped)
```
ApoKalipse-V, line 342: This is what the cpu_full and gpu_full status are supposed to cover. Why is this logic different?

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/aql_packet.hpp
```diff
@@ -284,6 +284,9 @@ class SQTTBufferingPackets
     hsa_ext_amd_aql_pm4_packet_t                query_status{};
     virtual std::optional<sqtt_buffer_status_t> query_buffer_status();
 
+    // Refreshed by every query_buffer_status(): set while the hardware is not running the trace.
+    bool trace_stopped{false};
```
ApoKalipse-V, line 288: This should be only possible as a valid status if both buffers were full, and we have a bool for that.

projects/rocprofiler-sdk/source/lib/aqlprofile/core/threadtrace.cpp
```diff
@@ -112,12 +112,6 @@ _internal_aqlprofile_att_iterate_data(aqlprofile_handle_t            handle,
         size_t sample_size =
             (control_ptr[se_index].wptr & wptr_mask) * sqttbuilder->GetWritePtrBlk();
 
-        if(pm4_factory->GetGpuId() == aql_profile::GFX11_GPU_ID)
```
ApoKalipse-V, line 115: This is required for gfx11.0

ApoKalipse-V (changes requested):

projects/rocprofiler-sdk/source/lib/aqlprofile/pm4/sqtt_builder.h
EvanPhoukong: Fixed. Only the write pointer reset is gfx11 specific.

projects/rocprofiler-sdk/source/lib/aqlprofile/aql_profile_v2.h
EvanPhoukong: `is_too_late` means we overflowed and lost data. `trace_stopped` means the hardware stopped tracing on its own with no overflow.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
EvanPhoukong: You're right, we don't. I removed it

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
EvanPhoukong: Done!

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/threading.cpp
EvanPhoukong: Correct me if I'm wrong, but cpu_full and gpu_full status cover overflow. Here, tracing stops early with no overflow at all, so neither flag ever gets set. During testing, both FULL bits read 0x0 at the stop, with PACKET_LOST_BUF_NO_LOCKDOWN never firing once.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/aql_packet.hpp
EvanPhoukong: Both bits are never simultaneously FULL as long as the drain keeps up, and when tracing stops they both read 0x0, so a stopped trace is indistinguishable from one that just hasn't filled a buffer yet. We need to also check whether TT is running using STATUS.BUSY.

projects/rocprofiler-sdk/source/lib/aqlprofile/core/threadtrace.cpp
EvanPhoukong: Readded!

PR #8150, opened 2026-07-06
fix(rocjitsu): clean the abi

### LINKED ISSUE #8145: rocjitsu: clean up exposed symbols
```
nm --demangle --dynamic --defined-only $(rocm-sdk path --bin)/../lib/librocjitsu.so
```
prints 1028 symbols, it should be 61

amd-arosa:
## Motivation
rocjitsu should have a clean abi.
- mark the ld_preload symbols with RJ_INTERPOSER_EXPORT
- librocjitsu.so reexports the flatbuff abi
- librocjitsu.so had 1028 public symbols most of which are not part of our abi
- TheRock builds with `-fvisibility=hidden` that was causing the LD_PRELOAD symbols to not be exported.

```
nm --demangle --dynamic --defined-only $(rocm-sdk path --bin)/../lib/librocjitsu.so
```

## Technical Details

- set our build to hide symbols by default
- explicitly export the LD_PRELOAD symbols

## JIRA ID

Resolves https://github.com/ROCm/rocm-systems/issues/8145

## Test Plan

- ran local tests
- ci cd
- check the symbols exported by therock artifact

## Test Result

pending build

## Submission Checklist


emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/base/rj_compiler.h
```diff
@@ -14,6 +14,13 @@
 #define RJ_API_EXPORT __attribute__((visibility("default")))
 #endif
 
+/// @brief Marks a libc entry point interposed by the LD_PRELOAD shim for export.
```
atgutier, line 17: Can you use RJ_API_EXPORT so we don't duplicate this?

emulation/rocjitsu/tests/interposer_test.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+#include <elf.h>
+
+namespace {
+
+// The complete set of libc entry points the interposer overrides. This MUST be
+// kept in sync with the extern "C" definitions in
+// lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp. When you add or remove an
+// interposed libc function, update both.
+constexpr const char *kInterposedLibcSymbols[] = {
```
atgutier: This seems fragile as it needs to be kept in sync with the exported functions otherwise we get false passes.

emulation/rocjitsu/tests/interposer_test.cpp
```diff
@@ -0,0 +1,270 @@
...
+  for (const std::string &name : symbols_.exported()) {
+    if (name.rfind("rj_", 0) == 0)
+      continue; // public C API
+    if (allowed.count(name) != 0)
+      continue; // interposed libc entry point
+    // Tolerate C++ standard-library runtime instantiations: they are mangled
+    // (_Z...) and never in our own namespace. Anything else -- an unmangled
+    // extern "C" symbol or a mangled rocjitsu:: symbol -- is a real leak.
+    const bool cxx_mangled = name.rfind("_Z", 0) == 0;
+    const bool ours = name.find("rocjitsu") != std::string::npos;
```
atgutier: Are you sure this is comprehensive (no false negatives)?

atgutier (commented):
The commit message still references the LD script.

emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/base/rj_compiler.h
amd-arosa: would you be okay with `#define RJ_INTERPOSER_EXPORT RJ_API_EXPORT`

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
```diff
@@ -794,9 +798,9 @@ int close(int fd) {
   return static_cast<int>(InterposerContext::real.close(fd));
 }
 
-__attribute__((destructor(101))) void rj_interposer_shutdown() {}
+RJ_INTERPOSER_EXPORT __attribute__((destructor(101))) void rj_interposer_shutdown() {}
```
kuhar: Should we keep `rj_interposer_shutdown` hidden? It is only a destructor/lifecycle hook, not an interposed libc entry point or public C API symbol, but `RJ_INTERPOSER_EXPORT` makes it a dynamic `T` export in the current build (`nm -D --defined-only build/librocjitsu.so` shows it). The new over-exposure test also allows every `rj_` symbol, so it will not catch this leak. We should either remove the export marker from the destructor or tighten the test to allow only the explicit public C API names.

emulation/rocjitsu/tests/CMakeLists.txt
```diff
@@ -408,6 +408,16 @@ rj_add_test(
...
 
+add_executable(interposer_test interposer_test.cpp)
+target_compile_definitions(
+    interposer_test
+    PRIVATE RJ_LIBROCJITSU_SO_PATH="$<TARGET_FILE:rocjitsu_shared>"
+)
+target_link_libraries(interposer_test PRIVATE GTest::gtest_main)
+rj_configure_target(interposer_test TEST)
+add_dependencies(interposer_test rocjitsu_shared)
+gtest_discover_tests(interposer_test)
```
kuhar: Please register this with `rj_add_gtest_tests(interposer_test)` instead of raw `gtest_discover_tests`. This build has `RJ_INSTALL_TESTS=ON`; the build-tree CTest file includes `interposer_test[1]_include.cmake`, but `build/tests/install/CTestTestfile.cmake` has no `InterposerExportsTest` entries. The wrapper also installs the test target and generates installed GTest cases, so packaged/install-tree validation currently skips the new export-surface test.

emulation/rocjitsu/tests/interposer_test.cpp
```diff
@@ -0,0 +1,270 @@
...
+    EXPECT_TRUE(symbols_.is_exported(name))
+        << "librocjitsu.so must export interposed libc symbol '" << name
+        << "'. Without it the LD_PRELOAD shim cannot intercept that part of "
+           "the amdgpu/KFD syscall surface. Check the RJ_INTERPOSER_EXPORT "
+           "marker on its definition in interposer.cpp.";
+  }
+}
+
+// The fcntl helpers are implementation details and must stay hidden.
+TEST_F(InterposerExportsTest, InternalHelpersStayHidden) {
```
kuhar: `InternalHelpersStayHidden` is not a useful guard for visibility. Both `fcntl_arg_kind` and `fcntl_impl` are in anonymous namespaces, so they have internal linkage and cannot become dynamic exports regardless of `-fvisibility=hidden` or `RJ_INTERPOSER_EXPORT`; `nm --defined-only build/librocjitsu.so` shows them only as local `t` symbols. If this test is meant to prove the new visibility machinery hides unmarked code, use an external-linkage helper that would otherwise be exportable, or drop this case and rely on `ExportSurfaceIsNotOverExposed` for the negative guard.

emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/base/rj_compiler.h
```diff
@@ -14,6 +14,13 @@
...
+#if defined(_WIN32) || defined(__CYGWIN__)
+#define RJ_INTERPOSER_EXPORT __declspec(dllexport)
+#else
+#define RJ_INTERPOSER_EXPORT __attribute__((visibility("default")))
+#endif
```
kuhar: How is this different from RJ_API_EXPORT?

kuhar (commented):
A few comments

emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/base/rj_compiler.h
kuhar: Why do we need to distinguish these two cases?

amd-arosa: replace with #define RJ_INTERPOSER_EXPORT RJ_API_EXPORT

amd-arosa: so we can find them easily, and because it's they aren't used the same way.

emulation/rocjitsu/tests/interposer_test.cpp
amd-arosa: adding auto parser

amd-arosa: i replaced with a parser.

emulation/rocjitsu/tests/interposer_test.cpp
amd-arosa: done

emulation/rocjitsu/tests/interposer_test.cpp
amd-arosa: de mangling is hard.
I came up with regex `^_Z[A-Z]*St` i THINK works?
switching to require positive match.

emulation/rocjitsu/tests/interposer_test.cpp
```diff
@@ -0,0 +1,351 @@
...
+/// of libc entry points (open/ioctl/mmap/stat/...) so an LD_PRELOAD of
+/// librocjitsu.so can redirect the amdgpu/KFD syscall surface to the simulated
+/// driver. Those overrides only take effect while they remain *exported*
+/// dynamic symbols: the project-wide -fvisibility=hidden preset plus
+/// --exclude-libs (see the rocjitsu_shared link options in CMakeLists.txt) strip
+/// almost everything, so a mistake there silently turns the preload into a
+/// no-op (the library loads but never redirects a single syscall).
+///
+/// This test parses the ELF dynamic symbol table of the freshly built
+/// librocjitsu.so and asserts every interposed libc symbol is present as a
```
atgutier: I don't see it doing this. It only appears to check for negative cases to ensure we aren't exporting when we don't want to. Can you double check or point me to which test does the assert?

emulation/rocjitsu/tests/interposer_test.cpp
```diff
@@ -0,0 +1,351 @@
...
+      continue;
+
+    // The parameter list opens at the first '(' after the marker.
+    const size_t paren = source.find('(', pos);
+    if (paren == std::string::npos)
+      break;
+    // A real definition has only "<return type> <name>" on the marker line
+    // between the marker and '('. A statement terminator or line break there
+    // means this is not an interposer definition (e.g. a mention in a comment).
+    if (source.substr(pos, paren - pos).find_first_of(";{}#\n") != std::string::npos)
```
atgutier: Probably ok as I don't imagine the list growing much longer than what we have currently, but is there a way to make this more robust? Will it work with line-wrapped functions?

atgutier (commented):
Some symbols are conditionally exported, like dup3, based on versioning etc. How will this work with that in the scenarios when it is not actually exported?

Overall the source scanning approach is very fragile and hard to get right such that it makes the test itself tricky. Can you use a more robust macro-based approach to define the interposer exports and track them in a list all at once?

kuhar:
> Can you use a more robust macro-based approach to define the interposer exports and track them in a list all at once?

If we are thinking about the same thing, I implemented something like this here in the past: https://github.com/google/vulkan-performance-layers/blob/9b3d9d22549c2bd67371bf71fced8b8c97ce6fe3/layer/support/layer_utils.h#L188-L289

emulation/rocjitsu/tests/interposer_test.cpp
```diff
@@ -0,0 +1,350 @@
...
+// The library's real ABI is entirely extern "C": the rj_* C API plus the
+// interposed libc entry points. Without a linker version script the dynamic
+// table cannot be made perfectly pristine -- libstdc++ marks namespace std as
+// _GLIBCXX_VISIBILITY(default), so std:: template instantiations emitted into
+// our own objects stay exported even under -fvisibility=hidden. Those residuals
+// are all std::-rooted Itanium C++ manglings, so we tolerate exactly that shape
+// (a positive std:: match) while still flagging any unmangled (extern "C") leak
+// or any other mangled C++ symbol (e.g. an accidentally exported rocjitsu:: one).
+TEST_F(InterposerExportsTest, ExportSurfaceIsNotOverExposed) {
+  const std::regex std_mangled("^_Z[A-Z]*St");
```
kuhar: `ExportSurfaceIsNotOverExposed` is too narrow for libstdc++ builds. It only tolerates std-rooted Itanium names via `^_Z[A-Z]*St`, but the repo's gcc-13 build exports a legitimate `__gnu_cxx::__stoa` instantiation (`_ZN9__gnu_cxx...`), so this test fails under gcc while the clang build passes.

Repro: `RJ_LIBROCJITSU_SO=/home/jakub/rocjitsu/review/build-gcc-13/librocjitsu.so ./build/tests/interposer_test --gtest_filter='*ExportSurfaceIsNotOverExposed*'` fails with sample `_ZN9__gnu_cxx6__stoa...`. Please either tolerate libstdc++/ABI helper residue such as `^_ZN9__gnu_cxx`/`__cxa`, or narrow this test to the real ABI risk: unexpected unmangled C symbols and rocjitsu-owned C++ exports.

emulation/rocjitsu/tests/CMakeLists.txt
```diff
@@ -408,6 +408,18 @@ rj_add_test(
...
+target_compile_definitions(
+    interposer_test
+    PRIVATE
+        RJ_LIBROCJITSU_SO_PATH="$<TARGET_FILE:rocjitsu_shared>"
+        RJ_INTERPOSER_SOURCE_PATH="${CMAKE_SOURCE_DIR}/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp"
+)
+target_link_libraries(interposer_test PRIVATE GTest::gtest_main)
+rj_configure_target(interposer_test TEST)
+add_dependencies(interposer_test rocjitsu_shared)
+rj_add_gtest_tests(interposer_test)
```
kuhar: Backtracking on my earlier suggestion to use `rj_add_gtest_tests` here: for this target, the plain `gtest_discover_tests(interposer_test)` hook is the better fit. `interposer_test` bakes in `RJ_LIBROCJITSU_SO_PATH` and `RJ_INTERPOSER_SOURCE_PATH`, so it is intentionally checking the freshly built library and source tree. The RocJITsu wrapper also generates installed-test fragments that run `bin/interposer_test` with only the standard installed config/kernel environment, so they do not point the test at installed `librocjitsu.so` or an installed source snapshot.

Please keep this test build-tree-only, and add a short CMake comment near the registration explaining why it intentionally uses the generic GoogleTest hook instead of `rj_add_gtest_tests`.

kuhar (commented):
A few comments

emulation/rocjitsu/tests/interposer_test.cpp
amd-arosa: as you mentioned above, some symbols aren't always exported so need to come up with a more robust way to figure out what should be exported.

emulation/rocjitsu/tests/interposer_test.cpp
amd-arosa: will try and do macros

emulation/rocjitsu/tests/interposer_test.cpp
amd-arosa: it was originally just checking for unmangled C symbols and rocjitsu-owned C++ exports.

https://github.com/ROCm/rocm-systems/pull/8150#discussion_r3531611799

ill tolerate libstdc++/ABI helper residue. though i would prefer to go back to the very lax  approach.

amd-arosa:
removed interposer_test.cpp because it was silly.

kuhar (approved):

PR #8151, opened 2026-07-06
test(rocprofv3): expand PMC kernel-filter coverage and add unit tests

minseobshin11:
## Motivation

The `rocprofv3` PMC kernel filters (`--kernel-include-regex`, `--kernel-exclude-regex`, `--kernel-iteration-range`) were under-tested: the iteration-range checks only compared subsets (a filter that profiled *every* launch still passed), several flag combinations had no CLI coverage, and the range-parsing / iteration-counting logic had no unit tests.

## Technical Details

**Integration coverage (`tests/rocprofv3/counter-collection/`):**

- **range_filtering** – validator now asserts the *exact* set of profiled iterations per pass, plus an exact record count so duplicates cannot hide behind the set comparison.
- **kernel_filtering** – captured dispatch counts are tied to the configured `kernel_iteration_range`, read from the input config so the CLI argument and the assertion cannot drift. The range applies per kernel per device, so both the CSV and JSON validators count launch ordinals within each `(kernel, agent)` pair — the assertions hold on a multi-GPU host, not just where the workload pins one device.
- **CLI tests** for: include + iteration range; multi-token range (`"1" "2"` ≡ `"[1-2]"`); exclude; exclude-cancels-the-include-set; `--mangled-kernels` vs `--truncate-kernels`; empty-match regex; out-of-bounds range; malformed / non-integer range aborts.

**Unit tests (GPU-independent):**

- Extracted `get_kernel_filter_range()` (from `config.cpp`) and the `is_targeted_kernel()` decision core (from `tool.cpp`) into `kernel_filter_range.*` / `kernel_iteration_filter.*`; `is_targeted_kernel()` is now a locking wrapper.
- **The locking also changed:** the iteration increment and the profile/skip decision now share one write lock. Previously they were separate acquisitions, so two concurrent dispatches of the same kernel could read the same iteration number — with range `{1}`, both skip and iteration 1 is never captured. That race would make the new exact-iteration assertions flaky. Decision logic is unchanged, verified against the old implementation over an exhaustive set of range/mode/interleaving cases.
- New `tool-tests` gtest target: range parsing (with death tests for malformed / non-integer input) and 1-based per-kernel counting (PMC all-iterations, thread-trace first-only, exact range, independent per-kernel counters). Only the test sources go to `rocprofiler_add_unit_test(SOURCES ...)`, since that list is installed into the generated `tool-tests.cmake`; including the tool sources embedded build-tree paths and broke the shipped test package.

## JIRA ID

JIRA ID : AIROCVAL-61

## Test Plan

```bash
cmake -S projects/rocprofiler-sdk -B build \
  -DCMAKE_BUILD_TYPE=RelWithDebInfo -DROCPROFILER_BUILD_TESTS=ON -DGPU_TARGETS=gfx942
cmake --build build -j

ctest --test-dir build -R "kernel-filtering|range-filtering" --output-on-failure
ctest --test-dir build -R "unit.kernel_filter_range|unit.kernel_iteration_filter" --output-on-failure
ctest --test-dir build -R "counter-collection"   # regression scope
```

## Test Result

gfx942 (MI300X): integration **45/45**, unit **17/17**, both passing twice back-to-back. Wider `counter-collection` suite **102/102**.

The full `rocprofv3` sweep shows 12 failures / 437, all from two pre-existing root causes (an ATT attachment timeout and a HIP-graph segfault, each cascading to its dependent validates). Both reproduce identically on an unmodified `develop` build.

Non-vacuity checked by mutating real output: a dropped dispatch, a duplicated record, a relabelled dispatch id, and a never-collected counter each fail with a per-`(kernel, agent)` diagnostic.

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests

### DIFF
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/CMakeLists.txt
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/CMakeLists.txt
@@ -4,11 +4,19 @@
 
 rocprofiler_activate_clang_tidy()
 
-set(TOOL_HEADERS att_no_intercept.hpp att_no_intercept_impl.hpp config.hpp
-                 execution_profile.hpp graph_stack.hpp helper.hpp stream_stack.hpp)
+set(TOOL_HEADERS
+    att_no_intercept.hpp
+    att_no_intercept_impl.hpp
+    config.hpp
+    execution_profile.hpp
+    graph_stack.hpp
+    helper.hpp
+    kernel_filter_range.hpp
+    kernel_iteration_filter.hpp
+    stream_stack.hpp)
 
-set(TOOL_SOURCES att_no_intercept.cpp config.cpp graph_stack.cpp main.c tool.cpp
-                 stream_stack.cpp)
+set(TOOL_SOURCES att_no_intercept.cpp config.cpp graph_stack.cpp kernel_filter_range.cpp
+                 kernel_iteration_filter.cpp main.c stream_stack.cpp tool.cpp)
 
 if(rocprof-trace-decoder_FOUND)
     list(APPEND TOOL_SOURCES att_no_intercept_quick_scan.cpp)
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/config.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/config.cpp
@@ -22,6 +22,7 @@
 //
 
 #include "config.hpp"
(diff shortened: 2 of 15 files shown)


samcrater-amd:
nit: per rocprof's contributing.md, unit tests that are scoped within one file should be in a file with the same name as the source file (i.e. no _test suffix)

samcrater-amd (approved):

bwelton (approved):
LGTM, Thanks for the change!

PR #8155, opened 2026-07-06
[RCCL] ADD gfx1250 support to DDA kernels

nawrinsu:
## Motivation

DDA fast-path collectives are supported only on gfx942/gfx950 via a single-node, IPC-based path (hipIpcGetMemHandle). That path does not work on gfx1250 as it requires cuMem and fabric handle support. This PR adds a parallel fabric DDA path for AllGather, AllReduce, AllToAll, and ReduceScatter collectives.

## Technical Details

Add a fabric-based DDA path so gfx1250 can use DDA kernels for AllGather, AllReduce, AlltoAll, and ReduceScatter.

 - New fabric kernels/host code per collective (dda_<collective>_fabric.cu)
 - Fabric memory handler (fabric_mem_handler.cc) exporting/importing VMM handles and a device-side fabric barrier (fabric_gpu_barrier.cu)
 - Fabric comm init/fini (fabric_init.cu), gated on MNNVL && gfx1250, requires cuMem/VMM; allocates the VMM scratch and peer-pointer buffers
 - Fabric kernels are templated on a compile-time size NRANKS_CT. NRANKS_CT = 4/8 are specialized with a fully unrolled peer loop; NRANKS_CT=0 is a runitme fallback that partially unrolls to 8.
 - Fabric kernels support any ranks between 2 to max (72 for scale-up)

Existing gfx942/950 IPC behavior is unchanged.

## JIRA ID

Resolves AICOMRCCL-1154
## Test Result

Collected performance numbers for all collectives on 1P4G system.
## Submission Checklist


projects/rccl/src/include/fabric_gpu_barrier.h
```diff
@@ -0,0 +1,122 @@
...
+#include "device_buffer.h"
+#include "fabric_mem_handler.h"
+#include "ipc_gpu_barrier.h"
+
+struct ncclMemManager;
+
+namespace meta::comms {
+
+// Upper bound on the number of ranks the fabric DDA path supports.
+constexpr int kDdaMaxNranks = 72;
```
nileshnegi, line 25: does it need to be 72? or can it be 144 for DPX mode?

projects/rccl/src/collectives.cc
```diff
@@ -243,16 +244,31 @@ ncclResult_t ncclAllGather_impl(const void* sendbuff, void* recvbuff, size_t sen
...
-        datatype,
-        comm,
-        stream));
-    return ncclSuccess;
+  if (rcclDdaEnabled(comm, nRanks * sendcount * ncclTypeSize(datatype), 8388608)) {
+    if (IsArchMatch(comm->archName, "gfx1250")) {
+      if (ncclAllGatherDdaFabricEligible(comm, sendbuff, recvbuff, sendcount, datatype)) {
+        INFO(NCCL_COLL,
+             "AllGather: taking DDA fabric (VMM) path: nRanks=%d nNodes=%d sendcount=%zu datatype=%d bytes=%zu",
+             comm->nRanks, comm->nNodes, sendcount, (int)datatype, sendcount * ncclTypeSize(datatype));
```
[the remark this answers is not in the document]
nusislam: Tests should be added to check fabric eligibility.

projects/rccl/src/include/fabric_gpu_barrier.h
nusislam: DDA is not enabled in partitioned mode at the moment.

projects/rccl/src/include/fabric_gpu_barrier.h
```diff
@@ -0,0 +1,122 @@
...
+
+      // Wait for the peer's signal in this rank's own buffer.
+      if constexpr (fenceType == MemFenceType::RELEASE_ONLY) {
+        waitFlag<std::memory_order_relaxed>(
+            selfBuf + getFlagIdx(peerRank, blockIdx.x));
+      } else {
+        waitFlag<std::memory_order_acquire>(
+            selfBuf + getFlagIdx(peerRank, blockIdx.x));
+      }
+    }
```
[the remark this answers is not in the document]
nusislam: This is a valid concern. Needs a fix.

projects/rccl/src/collectives.cc
```diff
@@ -132,12 +132,13 @@ RCCL_PARAM(DdaThreshold, "DDA_THRESHOLD", (size_t)(67108864));
 // threshold for gfx942; gfx950 uses the user-configurable rcclParamDdaThreshold();
 // all other architectures return false (threshold 0).
 static bool rcclDdaEnabled(const ncclComm* comm, size_t totalBytes, size_t gfx942Default) {
-  if (!rcclParamDdaEnable() || ncclParamLaunchOrderImplicit() || ncclGroupDepth != 0 || comm->nRanks < 8 || comm->symmetricSupport) return false;
+  if (!rcclParamDdaEnable() || ncclParamLaunchOrderImplicit() || ncclGroupDepth != 0) return false;
```
nusislam, line 136: I assume you have tested for `< 8` ranks. Also, the symmetric check is needed. Otherwise, even for symmetric registration DDA kernel was used instead of symmetric kernels.

projects/rccl/src/include/algorithms/CollCommon.h
```diff
@@ -129,34 +139,38 @@ static inline __device__ void reduceScatter(
...
     *reinterpret_cast<uint4*>(&srcVals[0]) =
         reinterpret_cast<const uint4*>(&ipcbuffs[0][srcIdx])[0];
-#pragma unroll NRANKS - 1
-    for (int r = 0; r < NRANKS - 1; ++r) {
+#pragma unroll kUnroll
+    for (int r = 0; r < nRanks - 1; ++r) {
       *reinterpret_cast<uint4*>(&srcVals[(r + 1) & 1]) =
-          reinterpret_cast<const uint4*>(
-              &ipcbuffs[(r + 1) % NRANKS][srcIdx])[0];
+          reinterpret_cast<const uint4*>(&ipcbuffs[r + 1][srcIdx])[0];
```
nusislam, line 145: The indexing is changed here. Why was it needed? Need to confirm that it does not cause any regression.

projects/rccl/src/include/fabric_mem_handler.h
```diff
@@ -0,0 +1,52 @@
...
+  ncclFabricMemHandler(const ncclFabricMemHandler&) = delete;
+  ncclFabricMemHandler& operator=(const ncclFabricMemHandler&) = delete;
+  ncclFabricMemHandler(ncclFabricMemHandler&&) = delete;
+  ncclFabricMemHandler& operator=(ncclFabricMemHandler&&) = delete;
+  ~ncclFabricMemHandler();
+
+  // Register this rank's local VMM allocation: mapped pointer, its cuMem
+  // allocation handle (for export), and the requested size (the import side
+  // re-aligns to the allocation granularity).
+  ncclResult_t addSelfDeviceMem(
```
nusislam, line 32: It would be good to add tests for the `fabric_mem_handler`.

projects/rccl/src/include/fabric_gpu_barrier.h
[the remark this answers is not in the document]
nawrinsu: Addressed

projects/rccl/src/collectives.cc
nawrinsu: `< 8` ranks and symmetric check is valid for gfx942 and gfx950. I moved these checks for those architectures. gfx1250 can have any number of ranks (we tested for 4 ranks). `comm->symmetricSupport` is always `true` on gfx1250 as we need `cuMem` and `NCCL_WIN_ENABLE` is set to `1` by default. That's why I removed it for gfx1250 arch. But I see the point and will add a symmetric eligible check.

projects/rccl/src/include/algorithms/CollCommon.h
nawrinsu: Since `r` only goes up to `nRanks - 2`, `r + 1` is always in `[1, nRanks - 1]` range (`< nRanks`), so the `%NRANKS` never wrapped around. Dropping it doesn't change the indexing behavior.

nusislam (approved):

nawrinsu:
%RCCL_TEST_CONFIG
mi300x_mellanox_ib.json‎

nawrinsu:
The `Gather.OutOfPlace` test is failing across multiple PRs and not related to this PR changes.

PR #8163, opened 2026-07-06
feat(rocjitsu): DBI: Add host-side logging ring buffer for probes

accauble:
## Motivation

Adds the host-side logging ring buffer for eventual use by device probes to pass dynamic instrumentation data back to the host. Vague plan for complete implementation in Issue #8574.

## Technical Details

* Added an ABI in `log_abi.h`. Defines a `RjLogBufferHeader` which includes the buffer metadata (e.g. `read_ptr` and `write_ptr`) and a `RjLogRecord` which includes the pertinent information at an instrumentation point (`record_type`, `site`, and the actual `payload`). Currently includes a way to create versions for future compatibility issues.
* Added the logging buffer functionality in `log_buffer.{h,cpp}`. Currently assumes one dispatch at a time (will need to reinitialize for a second dispatch), single-threaded, AND that drain runs after dispatch is complete. The expectation is that we will add these improvements once everything is working.
* Added CPU-only tests

## Test Plan

* Added tests should pass
* All other tests should pass

## Test Result

* All tests pass

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/log_abi.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/log_abi.h
@@ -1,0 +1,92 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: MIT
+
+/// @file log_abi.h
+/// @brief Fixed host/device ABI for the DBI logging buffer.
+///
+/// This header is intentionally dependency-light: it pulls in only <cstddef>
+/// and <cstdint> so the exact same struct definitions can be included from
+/// host C++ and from HIP device code (the `rj_log_write` probe fixture). Keep
+/// it that way -- do not add host-only includes here. The host ring-buffer
+/// class and drain logic live in log_buffer.h.
+///
+/// The type names are currently unversioned, but versioned type names can be
+/// introduced to extend the ABI while keeping compatibility.
+
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+
+namespace rocjitsu {
+
+/// @brief Magic value stamped into RjLogBufferHeader::magic ("RJLG",
+///        little-endian byte order 'R','J','L','G').
+inline constexpr uint32_t kRjLogMagic =
+    static_cast<uint32_t>('R') | (static_cast<uint32_t>('J') << 8) |
+    (static_cast<uint32_t>('L') << 16) | (static_cast<uint32_t>('G') << 24);
+
+/// @brief Current logging ABI version.
+inline constexpr uint16_t kRjLogAbiVersion = 0;
+
+/// @brief Fixed record size in bytes (power of two).
+inline constexpr uint32_t kRjLogRecordSize = 64;
+
+/// @brief Control header for a logging ring buffer.
+///
+/// The three monotonic counters (@c write_ptr, @c read_ptr, @c overflow_count)
+/// are each placed on their own 64-byte cache line to avoid false sharing
+/// between GPU producers and the host consumer. Counters are record *indices*
+/// (not byte offsets); the live slot for index @c i is @c i&(slot_count-1).
+struct alignas(64) RjLogBufferHeader {
+  // Cache line 0: immutable control fields after initialization.
+  uint32_t magic;       ///< kRjLogMagic.
+  uint16_t abi_version; ///< kRjLogAbiVersion.
+  uint16_t header_size; ///< sizeof(RjLogBufferHeader).
+  uint32_t record_size; ///< kRjLogRecordSize.
+  uint32_t slot_count;  ///< Number of records; power of two.
+  uint64_t flags;       ///< Reserved
+  uint8_t reserved_control[40];
+
+  // Cache line 1: producer-owned hot counter.
+  uint64_t write_ptr; ///< Next record index to claim.
+  uint8_t reserved_producer[56];
+
+  // Cache line 2: consumer-owned counter.
+  uint64_t read_ptr; ///< Next record index to drain.
+  uint8_t reserved_consumer[56];
+
+  // Cache line 3: diagnostics written by producers, read by host after drain.
+  uint64_t overflow_count; ///< Records dropped because the ring was full.
+  uint8_t reserved_diagnostics[56];
+};
+
+static_assert(sizeof(RjLogBufferHeader) == 256);
+static_assert(alignof(RjLogBufferHeader) == 64);
+static_assert(offsetof(RjLogBufferHeader, write_ptr) == 64);
+static_assert(offsetof(RjLogBufferHeader, read_ptr) == 128);
+static_assert(offsetof(RjLogBufferHeader, overflow_count) == 192);
+
+/// @brief One logging record.
+struct alignas(64) RjLogRecord {
+  uint32_t valid;       ///< 0 = free, 1 = record written.
+  uint16_t abi_version; ///< kRjLogAbiVersion.
+  uint16_t record_size; ///< kRjLogRecordSize.
+  uint32_t record_type; ///< Tool-defined record kind.
+  uint32_t site;        ///< Original .text byte offset of the anchor.
+  uint64_t payload;     ///< Tool-defined payload.
+  uint64_t exec_mask;   ///< Original EXEC mask at the probe site.
+  uint32_t wave_id;     ///< 0 if unavailable in the first probe.
+  uint32_t workgroup_x; ///< 0 if unavailable.
+  uint32_t workgroup_y; ///< 0 if unavailable.
+  uint32_t workgroup_z; ///< 0 if unavailable.
+  uint32_t active_lane_count;
+  uint32_t writer_lane; ///< First active lane, or 0xffffffff if unknown.
+  uint64_t reserved0;
+};
+
+static_assert(sizeof(RjLogRecord) == 64);
+static_assert(alignof(RjLogRecord) == 64);
+static_assert(sizeof(RjLogRecord) == kRjLogRecordSize);
+
+} // namespace rocjitsu
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/log_abi.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/log_abi.h
@@ -45,7 +45,7 @@ struct alignas(64) RjLogBufferHeader {
   uint16_t header_size; ///< sizeof(RjLogBufferHeader).
   uint32_t record_size; ///< kRjLogRecordSize.
   uint32_t slot_count;  ///< Number of records; power of two.
-  uint64_t flags;       ///< Reserved
+  uint64_t flags;       ///< Reserved; must be zero for now. Readers ignore unknown bits.
   uint8_t reserved_control[40];
 
   // Cache line 1: producer-owned hot counter.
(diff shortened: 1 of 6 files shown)


atgutier (commented):
Thought I'd share some initial agent comments:


Overall a well-considered foundation — no BLOCKER. The expensive-to-change axis (the ABI) is largely
   right; the issues are under-specified layout pinning and two release-build robustness gaps at the
  device boundary.

  MAJOR — must fix before a device writes this ABI

  1. Record field offsets aren't pinned with static_assert(offsetof(...)) — log_abi.h (RjLogRecord).
  The layout is pinned only by static_assert(sizeof == 64), which locks the total but not the position
   of site/payload/exec_mask/wave_id/workgroup_*. A future reorder or width-change of any middle field
   that preserves total size compiles clean, passes the sizeof assert, and silently desyncs from a
  device that writes the old offsets. Notably, the header struct does get per-field offsetof asserts —
   so this is an asymmetry, not a project non-convention. Fix: add offsetof asserts for every record
  field, exactly as the header does. Cheap insurance that turns an accidental device-desync into a
  build failure.
  2. pending > slot_count in drain() is only guarded by assert — log_buffer.cpp. pending = write_ptr -
   read_ptr comes from the device-written header. The code trusts the producer contract with an
  assert, which compiles out in release. If a buggy device advances write_ptr past read_ptr +
  slot_count, the loop runs pending iterations over only slot_count physical slots — (read_ptr+n) &
  mask aliases earlier slots and the same records get delivered to the callback more than once (silent
   duplicate instrumentation data). At a cross-agent boundary where the reader can't trust the
  writer's counters, an assert-only guard is insufficient. Fix: clamp in all builds (pending =
  min(pending, slot_count_)), surface it via a new DrainStats::overrun/dropped field, keep the assert
  for debug visibility.

  MINOR

  - No comment/assert freezing the offsets of the version discriminant fields
  (magic/abi_version/header_size) across ABI versions — log_abi.h. The whole versioning scheme depends
   on a host locating abi_version at a fixed offset before trusting anything else, but nothing states
  offsets 0/4/6 are frozen for all future versions. Add the comment + offsetof asserts.
  - record_size width differs between header (uint32_t) and record (uint16_t) — harmless today (both
  hold 64) but a wart; a record size >65535 would silently truncate in the record's field. Make both
  uint32_t.
  - Hole-skip policy in drain() will be wrong under concurrency — skipping a valid==0 slot inside
  [read_ptr, write_ptr) and still advancing read_ptr to write_ptr is fine for the current
  post-completion model, but is exactly the behavior that permanently loses a not-yet-published slot
  once producers publish valid concurrently. Correct now; add a @note that this must be revisited for
  the concurrent producer so it isn't mistaken for a settled invariant.
  - create_for_host_tests is the only factory and uses plain aligned operator new — fine for a
  host-side PR, but the eventual version needs the same [header][slots] image in
  device-mappable/pinned memory. Consider factoring the header-init out of the allocator so a future
  create_device_shared can reuse it (no retrofit).

NIT

  format_log_line's 128-byte stack buffer is a magic number (prefer std::format per rocjitsu style);
  the flags reserved field should document "must be 0 in v0, readers ignore unknown bits"; no test
  pins record field offsets or checks a wrong-magic/abi_version (drain currently trusts the header
  blindly — consider a validate() before first drain).

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/log_abi.h
```diff
@@ -0,0 +1,92 @@
...
+  uint32_t magic;       ///< kRjLogMagic.
+  uint16_t abi_version; ///< kRjLogAbiVersion.
+  uint16_t header_size; ///< sizeof(RjLogBufferHeader).
+  uint32_t record_size; ///< kRjLogRecordSize.
+  uint32_t slot_count;  ///< Number of records; power of two.
+  uint64_t flags;       ///< Reserved
+  uint8_t reserved_control[40];
+
+  // Cache line 1: producer-owned hot counter.
+  uint64_t write_ptr; ///< Next record index to claim.
```
atgutier, line 52: Do either the read/write_ptr need to be volatile?

atgutier (commented):
My own comment here

accauble:
Thanks for a first look! @atgutier! Before I submit my changes, I would like to discuss the record size comment. Adding @kuhar, @bjacob, and @lua1235 so they are at least aware of the discussion.

> * record_size width differs between header (uint32_t) and record (uint16_t) — harmless today (both
>   hold 64) but a wart; a record size >65535 would silently truncate in the record's field. Make both
>   uint32_t.

Thinking this over caused me to redesign the record. If the record size ever grows, it is most likely to grow because of the payload, so I have moved that to the end of the record where it would hopefully cause fewer headaches. The cost of a switch to u32 from u16 would mean either changing the record size right now to 128 or dropping the reserved space from 8 bytes to 4 bytes (2 bytes will serve as padding after the u16 abi_version). The latter looks like this: 

```
   0  valid                    u32    // must be u32 for device handshake
   4  abi_version          u16
   6  reserved0            u16   // added u16 buffer
   8  record_size          u32   // from u16 to u32 -- accommodates a record of > 65535 bytes
  12  record_type        u32
  16  exec_mask          u64
  24  site                     u32
  28  wave_id              u32
  32  workgroup_x      u32
  36  workgroup_y      u32
  40  workgroup_z      u32
  44  active_lane_count u32
  48  writer_lane         u32
  52  reserved1           u32  // only 4 bytes reserved
  56  payload              u64  // moved this here
  = 64 bytes
```

On the other hand, while a 64-bit payload is good for a basic block counter or something, it is not going to be enough for the concurrent sanitizer tools prototyped. The LLM thinks that the consan payload is about 68 bytes (including a dispatch id which could be included in the record instead). So, it might be worth expanding this to a 128 byte record size immediately.

But then this begs the question...does the record size really need to be 32 bits? The only use case I can think of for such a large payload (> 64KB) would be some debugging, like dumping the contents of memory at a point of execution.

atgutier:
> Thanks for a first look! @atgutier! Before I submit my changes, I would like to discuss the record size comment. Adding @kuhar, @bjacob, and @lua1235 so they are at least aware of the discussion.
> 
> > * record_size width differs between header (uint32_t) and record (uint16_t) — harmless today (both
> >   hold 64) but a wart; a record size >65535 would silently truncate in the record's field. Make both
> >   uint32_t.
> 
> Thinking this over caused me to redesign the record. If the record size ever grows, it is most likely to grow because of the payload, so I have moved that to the end of the record where it would hopefully cause fewer headaches. The cost of a switch to u32 from u16 would mean either changing the record size right now to 128 or dropping the reserved space from 8 bytes to 4 bytes (2 bytes will serve as padding after the u16 abi_version). The latter looks like this:
> 
> ```
>    0  valid                    u32    // must be u32 for device handshake
>    4  abi_version          u16
>    6  reserved0            u16   // added u16 buffer
>    8  record_size          u32   // from u16 to u32 -- accommodates a record of > 65535 bytes
>   12  record_type        u32
>   16  exec_mask          u64
>   24  site                     u32
>   28  wave_id              u32
>   32  workgroup_x      u32
>   36  workgroup_y      u32
>   40  workgroup_z      u32
>   44  active_lane_count u32
>   48  writer_lane         u32
>   52  reserved1           u32  // only 4 bytes reserved
>   56  payload              u64  // moved this here
>   = 64 bytes
> ```
> 
> On the other hand, while a 64-bit payload is good for a basic block counter or something, it is not going to be enough for the concurrent sanitizer tools prototyped. The LLM thinks that the consan payload is about 68 bytes (including a dispatch id which could be included in the record instead). So, it might be worth expanding this to a 128 byte record size immediately.
> 
> But then this begs the question...does the record size really need to be 32 bits? The only use case I can think of for such a large payload (> 64KB) would be some debugging, like dumping the contents of memory at a point of execution.

I'd say we can refine and stabilize the record after testing. We have some flexibility while in beta to muck with these things. Hard to say how it should formalize until we test and establish some use cases that can give us clarity. So I'm good with getting the best effort design and altering it later if need be.

accauble:
Alright sounds good @atgutier. I have kept the record like above. I was planning on keeping this as a draft until I had an actual consumer for it, but since you've already reviewed, I'll just mark it ready. Thanks again for looking it over.
 
> 1. Fix: add offsetof asserts for every record field, exactly as the header does.

Done.

> 2. Fix: clamp in all builds (pending = min(pending, slot_count_)), surface it via a new DrainStats::overrun/dropped field, keep the assert for debug visibility.

First half done and put into `dropped_overrun`. Did however drop the assert because it seemed wrong to abort the *host* for bad data from the device and is left for the user to decide how to handle it.

> * No comment/assert freezing the offsets of the version discriminant fields
>   (magic/abi_version/header_size) across ABI versions —  Add the comment + offsetof asserts.

Done

> * record_size width differs between header (uint32_t) and record (uint16_t) — Make both
>   uint32_t.

Done. Can revisit later.

> * Hole-skip policy in drain() will be wrong under concurrency —  add a @note that this must be revisited for
>   the concurrent producer so it isn't mistaken for a settled invariant.

Done.

> * create_for_host_tests is the only factory and uses plain aligned operator new — Consider factoring the header-init out of the allocator so a future
>   create_device_shared can reuse it (no retrofit).

Done.

> format_log_line's 128-byte stack buffer is a magic number (prefer std::format per rocjitsu style); the flags reserved field should document "must be 0 in v0, readers ignore unknown bits"; no test pins record field offsets or checks a wrong-magic/abi_version (drain currently trusts the header blindly — consider a validate() before first drain).

Updated `format_log_line`; documented flags; added `validate()`.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/log_abi.h
accauble: I think the answer is no, but fair question. For this version of the code, we drain once so this is not currently an issue. When we implement the drain while executing, caching/ordering/etc will become important but it's my understanding that volatile will not be enough and that we will have to use something like std::atomic_ref.

atgutier (commented):
[MINOR] lib/rocjitsu/src/rocjitsu/code/patch/log_buffer.cpp:140 — Overrun recovery drains the wrong
  logical record window

  Reasoning: When write_ptr - read_ptr > slot_count, the code clamps only the count but continues
  draining from the old read_ptr. In a genuinely overrun four-slot ring after six writes, slots
  contain logical records [4,5,2,3]. The current loop returns 4,5,2,3, although the surviving
  chronological window is records 2,3,4,5.

  Evidence:

  - The loop always indexes from the original read_ptr at line 148.
  - The comments and DrainStats::dropped_overrun promise defensive handling of a producer overrun.
  - DrainClampsProducerOverrun at tests/patch/log_buffer_test.cpp:364 writes only records 0–3 and then
    changes write_ptr to 6. It does not model records 4 and 5 overwriting slots 0 and 1, so it misses
    the ordering failure.

  Suggested fix: On overrun, set the drain start to write_ptr - slot_count_, set pending =
  slot_count_, and iterate from that adjusted start. Add a test that actually publishes six records
  into four slots and expects records 2–5 in order with dropped_overrun == 2.

  Confidence: High.

  [MINOR] lib/rocjitsu/src/rocjitsu/code/patch/log_buffer.cpp:149 — Record-level ABI version and size
  are ignored

  Reasoning: RjLogRecord contains abi_version and record_size, but drain() accepts every record with
  valid == 1 and passes it to the callback using the current compiled layout. A stale or malformed
  device probe can therefore be silently decoded using incompatible offsets.

  Header validation does not fully cover this: the host initializes the header, while the embedded
  probe writes individual records. The host and probe versions can diverge while the header remains
  valid.

  Evidence:

  - Record discriminants are defined at log_abi.h:86–88.
  - The only record gate in drain() is valid == 1.
  - validate() checks only header fields.
  - InvalidThenValidSlotBothHandled at tests/patch/log_buffer_test.cpp:243 leaves record_size == 0,
    sets valid == 1, and expects the record to be delivered.

  Suggested fix: Before invoking the callback, require:

 record.abi_version == kRjLogAbiVersion &&
  record.record_size == kRjLogRecordSize

  Clear and skip incompatible records and expose a dedicated malformed/incompatible-record statistic.
  Add wrong-version and wrong-size tests. If mixed versions are later supported, route them to
  version-specific decoders.

  Confidence: High.

  [MINOR] lib/rocjitsu/src/rocjitsu/code/patch/log_buffer.cpp:57 — Wrapper-allocation failure leaks
  the backing buffer

  Reasoning: The aligned backing allocation remains a raw pointer until the separate LogBuffer
  allocation succeeds at line 72. If new LogBuffer throws, the potentially large backing allocation
  leaks and std::bad_alloc escapes, even though the factory contract says allocation failure returns
  nullptr.

  Evidence:

  - Storage is allocated at lines 57–61.
  - No RAII owner exists before the throwing allocation at line 72.
  - log_buffer.h:49–50 promises nullptr when allocation fails.

  Suggested fix: Put the backing allocation under AlignedStorageDeleter immediately, use non-throwing
  allocation for the wrapper, and transfer ownership only after construction succeeds. Alternatively,
  explicitly release the backing allocation in a caught failure path. A fault-injection test would
  verify the contract.

  Confidence: High.

atgutier (approved):
Approved with comments, can ship once my minor comments are addressed and CI passes.

accauble:
@atgutier re-requesting your review since I didn't do exactly as you asked...

>  Overrun recovery drains the wrong logical record window

I did not implement the suggested test. As written in the comments, `write_ptr` should not advance with a full buffer, which means it should not get to 6, like in the suggested test. The overrun was added for debugging purposes which, I admit, is not clear at all from the code. I've updated the comments and the DrainClampsProducerOverrun test to clarify that and I've added a TODO to implement something that logs or fails when the producer is not acting correctly.

> Record-level ABI version and size are ignored

This was sort of intentional since they aren't used yet, but it was easy to do now even easier to forget later on so it's done.

> Wrapper-allocation failure leaks the backing buffer

Implemented the first fix. Did not implement the fault-injection test because it was not trivial and IMO not worth it.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/log_buffer.cpp
```diff
@@ -0,0 +1,213 @@
...
+    return nullptr;
+  }
+
+  // Header and records are each 64-byte sized/aligned, so the contiguous
+  // [header][slot_count * record] image is exactly the sum of the two.
+  const size_t total_bytes =
+      sizeof(RjLogBufferHeader) + static_cast<size_t>(slot_count) * sizeof(RjLogRecord);
+
+  // Own the backing allocation via RAII immediately: the wrapper allocation
+  // below is a throwing new, and until ownership transfers into the LogBuffer a
```
kuhar: Please update the allocation comment here. The wrapper now uses non-throwing new and handles a null result, so lines 57-60 still describe the old throwing path and contradict the accurate explanation at lines 76-79.

kuhar (commented):
Should we trim or explicitly mark the public API that is still being shaped by tests rather than a consumer? There is no in-tree caller yet, but log_buffer.h already exposes a test-named factory at line 65, mutable header and record access at lines 73-78, and a fixed presentation helper at line 150. I think it would help to either keep those seams private or experimental until the first consumer establishes the contract, or document that consumer and narrow each API around what it actually needs.

emulation/rocjitsu/tests/patch/log_buffer_test.cpp
```diff
@@ -0,0 +1,431 @@
...
+TEST(LogBufferTest, ValidateRejectsWrongMagic) {
+  auto buf = LogBuffer::create_for_host_tests(8);
+  ASSERT_NE(buf, nullptr);
+  buf->header()->magic = 0xdeadbeef; // as if written by an incompatible producer
+  std::string err;
+  EXPECT_FALSE(buf->validate(&err));
+  EXPECT_NE(err.find("magic"), std::string::npos) << err;
+}
+
+TEST(LogBufferTest, ValidateRejectsWrongAbiVersion) {
```
kuhar: It would be good to cover the remaining header-validation branches. The current negative tests exercise magic and ABI version, but not header_size, record_size, or slot_count; a small table-driven test could corrupt each field and check both the false result and diagnostic.

accauble:
> Should we trim or explicitly mark the public API that is still being shaped by tests rather than a consumer?

Good idea. Done.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/log_buffer.cpp
```diff
@@ -0,0 +1,213 @@
...
+      // completion, not a race to spin on.
+      ++stats.invalid_slots;
+      continue;
+    }
+    // The header passing validate() does not vouch for individual records: the
+    // host writes the header while the embedded probe writes each record, so a
+    // probe built against a different layout can publish a mismatched record
+    // under a valid header. Decoding it with our offsets would be silently
+    // wrong, so drop it rather than hand it to the callback.
+    if (record.abi_version != kRjLogAbiVersion || record.record_size != kRjLogRecordSize) {
```
kuhar, line 180: The per-record `record_size` check should not be presented as sufficient protection against a layout mismatch. `drain()` always strides by `sizeof(RjLogRecord)`, so with a future 128-byte producer the host index for slot 1 reads byte 64 of record 0; any values interpreted there are accidental. I think we should document the required handshake now: a producer must compare the header ABI and record size before writing, and any externally initialized image must pass `validate()` before `drain()`. The per-record check is then defense against corruption after both sides have agreed on the slot geometry.

kuhar (commented):
Just one suggestion

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/log_buffer.cpp
accauble: Added comments to document that

PR #8164, opened 2026-07-06
fix: Finddw from rocm sysdeps first

ApoKalipse-V:
## Motivation

TheRock builds libdw.so in lib/rocm_sysdeps/lib.
Rocprofiler-sdk does not need to use the system library in that case.

JIRA ID : AIPROFSDK-957

## Submission Checklist

### DIFF
--- a/.github/workflows/rocprofiler-sdk-continuous_integration.yml
+++ b/.github/workflows/rocprofiler-sdk-continuous_integration.yml
@@ -205,7 +205,7 @@ jobs:
         run: |
           git config --global --add safe.directory '*'
           apt-get update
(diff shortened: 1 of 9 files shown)


jbonnell-amd (approved):
Approved for the workflow and Docker changes. 

The `Findlibdw.cmake` change also looks fine but would like a review from someone more involved with rocprofiler-sdk to verify (want to make sure the new change won't break anything with non-TheRock installs of SDK)

lumachad:
Why not use find_package according to https://github.com/ROCm/TheRock/blob/main/docs/development/dependencies.md#elfutils instead of all the directory manipulation?

PR #8165, opened 2026-07-06
kfdtest: Add AMD Infinity Storage (AIS) test

amd-willpalacek:
Add a kfdtest component test for AMD Infinity Storage (AIS), which transfers data directly between a PCIe-backed file (e.g. NVMe) and GPU VRAM via hsaKmtAisReadWriteFile, bypassing system-memory bounce buffers.

The suite has five tests, all gated on hsakmt_is_dgpu() since AIS is only supported on dGPUs:

- AISGracefulFailureTest: exercises the API on /tmp and checks it either succeeds (NVMe-backed) or fails without crashing (non-NVMe).
- AISUnregisteredMemoryTest: deterministic negative case that passes an unregistered host pointer and expects the thunk address-range guard in ais.c to reject it without crashing. Runs on any dGPU.
- AISReadTest / AISWriteTest: read a file into VRAM and write VRAM back to a file, validating the full buffer contents with memcmp.
- AISMultiThreadedTest: threads each transfer a distinct file section to a separate VRAM buffer concurrently, then write them back. Thread count is configurable via --ais_threads (default 8) and the total workload via --ais_size_mb (default 1024), clamped to free disk and 1/4 of VRAM so it never overcommits on any machine.

Temp files are removed in TearDown.

## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDAISTest.cpp
+++ b/projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDAISTest.cpp
@@ -1,0 +1,690 @@
+/*
+ * Copyright (C) 2026 Advanced Micro Devices, Inc. All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+ * OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ * ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ *
+ */
+
+#include <sys/types.h>
+#include <sys/stat.h>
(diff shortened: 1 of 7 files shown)


projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDAISTest.cpp
```diff
@@ -0,0 +1,643 @@
...
+
+    for (int i = 0; i < N_BUFFERS_BASIC; i++) {
+        ssize_t bytesRead = read(m_fdOut, readBuf.data(), BUFFER_SIZE);
+        ASSERT_EQ(bytesRead, BUFFER_SIZE) << "Failed to read back buffer " << i;
+
+        std::fill(expected.begin(), expected.end(), m_patterns[i]);
+        ASSERT_EQ(memcmp(readBuf.data(), expected.data(), BUFFER_SIZE), 0)
+            << "Output file pattern mismatch at buffer " << i;
+    }
+
```
cfreeamd, line 535: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** `m_fdOut` is opened with `O_DIRECT` (via `createTestFile`), but `readBuf` is a `std::vector<uint32_t>` whose backing allocation is guaranteed only ~16-byte aligned. O_DIRECT requires the buffer address to be aligned to the device's logical block size (512B–4096B). This will fail with `EINVAL` on most NVMe devices, making `AISWriteTest` permanently fail on hardware.

Same root cause applies to `fillFileWithPatterns`' write buffers and the multi-threaded test's read-back. Fix: use `posix_memalign` / `std::aligned_alloc` for all staging buffers, or open a separate non-O_DIRECT fd for read-back verification.

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDAISTest.hpp
```diff
@@ -0,0 +1,132 @@
...
+
+    int m_fd;                       // Input file descriptor
+    int m_fdOut;                    // Output file descriptor (for write tests)
+    int m_gpuNode;                  // Default GPU node
+    bool m_isOnNVME;                // True if test file is on NVME
+    std::string m_testDir;          // Directory for test files
+    std::string m_testFilePath;     // Full path to input test file
+    std::string m_testFilePathOut;  // Full path to output test file
+    std::vector<void*> m_pBufs;     // VRAM buffer pointers
+    size_t m_bufSize;               // Size of each buffer in m_pBufs (for TearDown)
```
cfreeamd, line 122: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `isAISSupported()` is declared here but has no implementation in `KFDAISTest.cpp`. No test calls it, so there's no link error, but the declaration implies a utility that doesn't exist. Remove the declaration or add an implementation.

cfreeamd (commented):

`CMakeLists.txt`, `KFDTestFlags.hpp`, `KFDTestMain.cpp`, `OSWrapper.hpp` look correct. Issues in `KFDAISTest.cpp` and `KFDAISTest.hpp` noted inline.

PR #3933, opened 2026-03-10
[rocprofiler-sdk] Improving attachment tests

meserve-amd:
## Motivation

Fix many race conditions in attachment

Improve and reenable attachment tests

Sanitizer tests deferred for now

## Technical Details

This change includes a massive overhaul of rocprofiler-sdk-attach to make this possible and also patches some potential race conditions. Fixes #4636 These library changes include:

- Instead of SDK modifying the attach table to add a "notify" callback, SDK modules now directly call an "add callback" function which registers a function to be called on every object creation or destruction
  - This applies for both queues and code_objects
  - Added a new "phase" parameter to indicate which event is occurring (creation or destruction)
- To protect against some race conditions between iterate and register, registering a callback will now cause that callback to be immediately called for each existing object while a lock is held. This will ensure the SDK cannot possibly miss an object
  - The existing iterate_all calls remain so a snapshot can be taken and used at any time

I've bumped the version for the table, though we aren't doing any checks for the version as this is a fully internal API. This is an internal ABI breaking change, but if these libraries are incorrectly paired there is a fundamental installation issue.

Fixes three intermittent failures caused by race conditions unique to attachment. One is caused by the order queues are added to rocprofiler-sdk. One is caused by ordering of rocpd output steps. The last is caused by initialization order where queue interceptions could occur without kernel_id_map being populated. Fixes #5544

Rearchitects the attach tests to run indefinitely and receive SIGINT when they need to be stopped. Reduces the size and runtime of the test overall. Reduces runtime of a execute tests to 2-4 seconds.

Remove the separate json tests, matching the pattern of other rocprofv3 validation tests.

## JIRA ID

Resolves AIPROFSDK-856

## Test Plan

Ran attach tests locally to see pass
Validated intermittent tests across 1800 repeated runs on MI325X

## Test Result

All passed

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #8171, opened 2026-07-06
fix(rocprofiler-sdk): code object cb ordering

meserve-amd:
- Attach library now calls destroy callbacks in reverse order of registration
- Matches existing behavior of SDK
- Fixes an issue in PC sampling where code object data was never deleted

## Motivation

Fixes an issue where PC sampling code objects were never deleted found during validation of AIPROFSDK-856. Patches bug introduced by new features in #3933

## Technical Details

In typical initialization, when PC sampling is enabled, initialize will replace the hsa_executable_freeze and hsa_executable_destroy after code object initialization. The destroy implementation in particular expects this order. The attach library reversed this order, which caused the code object destroy callback to run first. The destroy would remove code objects from the shared storage and then the pc sampling callback would be unable to find the entry and fail. 

## JIRA ID

JIRA ID: AIPROFSDK-856

## Test Plan

Existing CI

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-attach/code_object_registration.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-attach/code_object_registration.cpp
@@ -1,6 +1,6 @@
 // MIT License
 //
-// Copyright (c) 2025 Advanced Micro Devices, Inc. All rights reserved.
+// Copyright (c) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


meserve-amd:
>     * ❌ Unit Test

This fixes an existing bug and relies on existing tests. Ignoring check and removing tag.

julijose-amd:
The load and destroy paths appear to require opposite callback orders:

**Load:** pc_sampling's `executable_freeze_internal()` depends on the shared `get_code_objects()` list populated by the main module, so the main module must execute before pc_sampling.
**Destroy:** pc_sampling still needs its entry to exist in that list, so the main module must execute after pc_sampling.

Since a single callback ordering cannot satisfy both requirements, switching to LIFO ordering may resolve the destroy path but would likely introduce a regression in the load path.

meserve-amd:
> The load and destroy paths appear to require opposite callback orders:

You're correct. I've updated the PR for this.

EvanPhoukong (approved):
LGTM - LIFO destruction order corresponds correctly to FIFO creation order and consistent in SDK.

ApoKalipse-V (approved):
LGTM though I think there needs to be something more robust than relying on the order a different library does something.

vlaindic (approved):
LGTM, thanks! @julijose-amd could you please test? Thanks!

PR #7217, opened 2026-06-13
refactor(rocshmem/gda): refactor QueuePair class

omor1:
## Motivation

The QueuePair class [gda/queue_pair.{hpp,cpp}] functions as three classes in a trenchcoat that masquerade as a single class; it simultaneously has functionality for:
* common device-side QP initialization and management, including host functions for registering additional local buffers; and
* provider-specific send and completion queue management, including WQE posting and CQ polling; and
* a multiplexing low-level interface that dispatches to the provider-specific functions based on which is at use at execution time; and
* a high-level SHMEM-like communication interface that dispatches to the low-level interface.

Maintaining this complex class and its many interactions is complicated and has led to changes intended for one GDA provider causing unintended impacts on the others, as well making several types of optimizations far more difficult. The goal of this PR is to refactor this single class into several distinct classes that will help improve code readability and maintainability, as well as restoring the ability to compile directly for a specific GDA provider without runtime dispatch.

The end goal is to allow much of the QueuePair and GDA initialization code to be abstracted as a separate library that can be consumed by both rocSHMEM and other users of similar functionality, such as the RCCL-GIN GDA backend.

## Technical Details

* `template <typename Provider> class QueuePairSHMEM`
  * gda/queue_pair.hpp
  * High-level SHMEM-like communication interface
  * Uses CRTP to dispatche to low-level functions implemented by `Provider`
  * Contains no non-static data members
* `template <typename Provider> class QueuePairBase`
  * gda/queue_pair.hpp
  * Common device-side QP initialization and management
  * Inherits from `QueuePairSHMEM<Provider>`
* `class QueuePairIONIC`
  * gda/ionic/queue_pair_ionic.{hpp,cpp}
  * Implements send and completion queue management for AMD Pensando 'ionic' RDMA NICs
  * Inherits from `QueuePairBase<QueuePairIONIC>`
* `class QueuePairBNXT`
  * gda/bnxt/queue_pair_bnxt.{hpp,cpp}
  * Implements send and completion queue management for Broadcom 'bnxt' RDMA NICs
  * Inherits from `QueuePairBase<QueuePairBNXT>`
* `class QueuePairMLX5`
  * gda/mlx5/queue_pair_mlx5.{hpp,cpp}
  * Implements send and completion queue management for NVIDIA/Mellanox 'mlx5' RDMA NICs
  * Inherits from `QueuePairBase<QueuePairMLX5>`
* `class QueuePairMux`
  * gda/queue_pair_mux.{hpp,cpp}
[...]

PR #8176, opened 2026-07-07
 Add SQTT marker support to rocSHMEM to improve device-level performance tracing

mberenjk:
## Motivation

Adding sqtt_marker to device functions and critical sections of the code to mprove device-level performance tracing

## Technical Details
Add SQTT marker support to rocSHMEM to enable device-side performance tracing. Similar to ROCTX, this provides annotations for performance analysis, but unlike ROCTX, it works on the device side, allowing correlation of rocSHMEM operations directly with GPU execution. This will improve trace visibility and help identify kernel-level bottlenecks more effectively.

## JIRA ID : AIROCSHMEM-431
JIRA ID - #431

## Test Plan

A complete test will be added in the CI with rocprofiler-sdk

## Submission Checklist


omor1 (changes requested):
Please hold until #7217 is merged to prevent conflicts.

projects/rocshmem/src/gda/queue_pair.cpp
ApoKalipse-V, line 160: Is the switch case itself a performance concern?
If not, I'd recommend to have just enter post_wqe_rma here
(it can be if gda_provider_ can be slow to load)

ApoKalipse-V, line 265: Would it be useful if we provided a way for enter/exit to contain extra information? E.g.

sqtt_marker_enter("mymarker")

if (a)
 sqtt_marker_exit("mymarker", "A")
else
 sqtt_marker_exit("mymarker", "B")

projects/rocshmem/src/backend_type.hpp
ApoKalipse-V: I think you need a flag here to switch these off when needed, like you do in #ifdef ENABLE_ROCTX

Imagine we changed the path, or sqtt_trace.hpp wasn't available for some reason

projects/rocshmem/CMakeLists.txt
ApoKalipse-V: We plan on having this available globally in rocm soon. I dont think it hurts for now, though.

projects/rocshmem/scripts/build_configs/all_backends
ApoKalipse-V: Doesn't hurt, but it's off by default.

projects/rocshmem/src/gda/queue_pair.cpp
```diff
@@ -31,6 +31,13 @@
 #include "constants.hpp"
 #include "util.hpp"
 
+#if SQTT_ENABLED
+#include </opt/rocm/include/rocprof-trace-decoder/rocprof_trace_decoder/cxx/markers.hpp>
```
mshanthagit, line 35: Do we need hard-coded path? Also, is there a check that ensures rocprof is installed?

projects/rocshmem/src/backend_type.hpp
```diff
@@ -39,6 +39,15 @@
 #include "rocshmem/rocshmem_config.h"  // NOLINT(build/include_subdir)
 #include "constmem.hpp"
 
+#if SQTT_ENABLED
+#include </opt/rocm/include/rocprof-trace-decoder/rocprof_trace_decoder/cxx/markers.hpp>
```
mshanthagit, line 43: Same as above

projects/rocshmem/src/gda/context_gda_device.cpp
```diff
@@ -32,6 +32,13 @@
 #include "context_gda_device.hpp"
 #include "context_gda_tmpl_device.hpp"
 
+#if SQTT_ENABLED
+#include </opt/rocm/include/rocprof-trace-decoder/rocprof_trace_decoder/cxx/markers.hpp>
```
amirsojoodi: This will probably cause issues if rocprof decoder is not installed there. For instance, on recent docker images rocm is shipped with python wheels under `/opt/venv/lib/python3.12/site-packages/_rocm_sdk_devel/`.

You might be able to handle this through the main CMakeLists to update the include paths such that this headerfile can be found without absolute addressing.

amirsojoodi (changes requested):
There might be some functions from the IPC backend that needs these markers. There are too many to name here. See if you can support some of those too. But if not, we can probably visit them later.

projects/rocshmem/src/gda/queue_pair.cpp
mberenjk: Yes, the switch is reported as a light performance concern. We have a plan to remove the switch in the future.

Issue #6447, opened 2026-05-27
[rocjitsu] Parallelization at the workgroup level (finer grained)


newling:
**James's description:** My understanding is that the granularity of multithreading in rocjitsu is at the XCD level. With only 8 XCDs, this leaves alot of potential parallelization on the table. For the functional emulation use case I currently have, I think each workgroup could run in a separate CPU thread. 

**Agent's description:**

## Current state

rocjitsu uses simdojo's PDES (Parallel Discrete Event Simulation) engine for multi-threaded execution. Components are partitioned across worker threads, synchronized via LBTS (Lower Bound on Time Stamp) barriers — each partition publishes the timestamp of its earliest pending event, and the global minimum determines how far partitions can safely advance in parallel.

**Threading granularity:** All CUs in an XCD share one L2 cache, and L2's non-atomic methods are not thread-safe. This forces all CUs in an XCD onto the same partition thread. Maximum parallelism = number of XCDs (8 on CDNA3/CDNA4). Current KMD configs use `num_threads: 1`.

**Multi-XCD dispatch:** The KMD/LD_PRELOAD path routes all work through a single CP. `SoC::distribute_cus()` registers CUs from all XCDs with the primary CP so its round-robin dispatch spreads workgroups across XCDs. This is intended to approximate hardware MES behavior for functional mode.

## Opportunity

rocjitsu supports two execution modes controlled by the `functional_quantum` setting:
- **quantum=0** (default): each CU runs all its wavefronts to completion in a single call — no yielding, no event-loop interleaving.
- **quantum>0**: each CU executes up to N instructions then yields, allowing other CUs to make progress. Believed to be required for correctness when workgroups synchronize via global memory (e.g., spin-locks).

In functional mode (quantum=0), workgroups are largely independent:
- LDS is per-workgroup
- Wavefront slots and registers are per-CU
- Barriers (`s_barrier`) are workgroup-internal
- Only global GPU memory is shared across workgroups

The natural parallelism unit appears to be the **workgroup**, not the XCD. A thread pool could potentially run N workgroups on N OS threads, far exceeding the 8-XCD limit. Modern server CPUs have 100+ hardware threads.

## What's already thread-safe (to our knowledge)

- **GpuMemory** (via `SparseMemory`): read/write methods appear to be protected by `std::shared_mutex` (reader-writer lock). Concurrent reads should be lock-free; writes serialize on page allocation.
[...]

PR #8177, opened 2026-07-07
test(rocjitsu): add execution quality gates

lialan:
## Summary

Add rocjitsu quality gates for safe scalar instruction execution and representative valid-address memory execution. The VGPR read-observation and D16 masked-read work has been split out to users/lialan/rocjitsu_8075_read_observation_split.

## Changes

- Require every safe scalar encoding to decode and execute without unexpected failures, with exact per-ISA expected unimplemented-instruction lists.
- Add a gfx1250 valid-address memory execution harness.
- Document dispatch/thread configuration fields without changing the shipped gfx950 CDNA4 KMD config.

## Validation

- pre-commit run --files emulation/rocjitsu/docs/configuration.md emulation/rocjitsu/tests/instruction_execution_harness_test.cpp
- git diff --check
- cmake --build /tmp/rocjitsu-pr8177-rework-20260707-213115-build -j 16
- ctest --test-dir /tmp/rocjitsu-pr8177-rework-20260707-213115-build -R '^(InstructionExecutionHarness\.|Gfx1250MemoryExecutionHarness\.)' --output-on-failure

ISSUE ID : #6447

### DIFF
--- a/emulation/rocjitsu/docs/configuration.md
+++ b/emulation/rocjitsu/docs/configuration.md
@@ -48,16 +48,14 @@ Pre-built configs are in `configs/`:
 }
 ```
 
-The example above is intentionally minimal and single-threaded. Shipped
-multi-XCD KMD configs may set `num_threads` to the number of simulated XCD
-partitions.
+The example above is intentionally minimal and single-threaded.
 
 ### Top-level fields
 
 | Field | Type | Description |
 |---|---|---|
 | `max_ticks` | int | Maximum simulation ticks (0 = unlimited) |
-| `num_threads` | int | PDES engine partitions/worker threads. For partitioned multi-XCD configs this is typically one partition per XCD. |
+| `num_threads` | int | Number of PDES engine partitions/worker threads. |
 | `exec_mode` | string | `"functional"` or `"cycle"` |
 | `vm.arch` | string | Architecture: `cdna3`, `cdna4`, etc. |
 
--- a/emulation/rocjitsu/docs/configuration.md
+++ b/emulation/rocjitsu/docs/configuration.md
@@ -48,12 +48,16 @@ Pre-built configs are in `configs/`:
 }
 ```
 
+The example above is intentionally minimal and single-threaded. Shipped
+multi-XCD KMD configs may set `num_threads` to the number of simulated XCD
+partitions.
+
 ### Top-level fields
 
 | Field | Type | Description |
 |---|---|---|
 | `max_ticks` | int | Maximum simulation ticks (0 = unlimited) |
-| `num_threads` | int | Worker threads for PDES engine |
+| `num_threads` | int | PDES engine partitions/worker threads. For partitioned multi-XCD configs this is typically one partition per XCD. |
 | `exec_mode` | string | `"functional"` or `"cycle"` |
 | `vm.arch` | string | Architecture: `cdna3`, `cdna4`, etc. |
 
--- a/emulation/rocjitsu/tests/instruction_execution_harness_test.cpp
+++ b/emulation/rocjitsu/tests/instruction_execution_harness_test.cpp
@@ -179,6 +179,25 @@ inline bool should_skip_inst(std::string_view mn) {
   return false;
 }
 
+struct HarnessCoverageExpectation {
+  std::string_view arch_name;
+  double min_coverage;
+  size_t max_unimplemented;
+};
+
+constexpr HarnessCoverageExpectation HARNESS_COVERAGE_EXPECTATIONS[] = {
+    {"cdna1", 99.0, 1},  {"cdna2", 99.0, 1},    {"cdna3", 99.0, 1},  {"cdna4", 99.0, 1},
+    {"rdna1", 98.0, 3},  {"rdna2", 98.5, 2},    {"rdna3", 100.0, 0}, {"rdna3_5", 100.0, 0},
+    {"rdna4", 100.0, 0}, {"gfx1250", 100.0, 0},
+};
+
+const HarnessCoverageExpectation *harness_coverage_expectation(std::string_view arch_name) {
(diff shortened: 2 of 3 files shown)


newling:
I'll take a look at this today, but just to note that for 
> Add VGPR write hooks so the race detector catches same-wave WAW hazards while preserving memory writeback semantics.

I have a PR out for this: https://github.com/ROCm/rocm-systems/pull/7890

I'd quite like to try and get that over the line (landed) soon. 

I also had a PR to fix some plugin leaks, I'm not sure if it related to this PR of yours: https://github.com/ROCm/rocm-systems/pull/8075 (I am working through some issues on that, and reconsidering how registers are accessed across the board).

lialan:
> I'll take a look at this today, but just to note that for
> 
> > Add VGPR write hooks so the race detector catches same-wave WAW hazards while preserving memory writeback semantics.
> 
> I have a PR out for this: #7890
> 
> I'd quite like to try and get that over the line (landed) soon.

@newling I removed the overlapping parts of #7890 from this PR, so now #8177 and #7890 no longer intersecting. Feel free to continue with it. 

> I also had a PR to fix some plugin leaks, I'm not sure if it related to this PR of yours: #8075 (I am working through some issues on that, and reconsidering how registers are accessed across the board).

Looks like they partially overlaps. I split the patch to `users/lialan/rocjitsu_8075_read_observation_split`, fell free to take what it can into your branch.

now this PR should be free of overlapping with #7890 and #8075

lialan:
@newling want to review it again?

emulation/rocjitsu/tests/race-detector/hip_race_gfx950_test.hip
```diff
@@ -16,6 +16,13 @@
 //   _race suffix = expects race, no suffix = expects no race.
 //   Paired tests share a base name (e.g. vgpr_waitcnt / vgpr_waitcnt_race).
 //
+// TODO: Add wave-32 (RDNA 3.5) cross-wave LDS test. Requires:
```
newling: I recently added tests for gfx1151: https://github.com/ROCm/rocm-systems/pull/8034, not sure what this comment is for?

emulation/rocjitsu/tests/instruction_execution_harness_test.cpp
```diff
@@ -162,6 +162,25 @@ inline bool should_skip_inst(std::string_view mn) {
   return false;
 }
 
+struct HarnessCoverageExpectation {
+  std::string_view arch_name;
+  double min_coverage;
+  size_t max_unimplemented;
+};
+
```
newling, line 196: What is min_coverage, and how is it computed? I seems to be inversely proportional to max_unimplemented, are the independent measures (if not do we want both?). 

What is the long term plan with this test, it seems like we're enforcing some quality control like ("at no point in the future should we have fewer unimplemented instructions than we have on July 13 2026" which seems a bit arbitrary). Is it possible to rather just have tests that execute all the implemented instructions?

newling (commented):
Thanks for working on improving testing. I have two questions to help me understand what is being added here.

emulation/rocjitsu/tests/race-detector/hip_race_gfx950_test.hip
lialan: ahh this is outdated, removed.

emulation/rocjitsu/tests/instruction_execution_harness_test.cpp
lialan: makes sense.. I replaced them with exact per-ISA expected exceptions; admitted encodings must now decode and execute without unexpected failures, while stateful instructions remain covered by targeted tests.

emulation/rocjitsu/docs/configuration.md
```diff
@@ -48,12 +48,16 @@ Pre-built configs are in `configs/`:
 }
 ```
 
+The example above is intentionally minimal and single-threaded. Shipped
+multi-XCD KMD configs may set `num_threads` to the number of simulated XCD
```
kuhar: I think we should describe `num_threads` only as the number of PDES partitions/worker threads here. All shipped configs currently set it to 1, including the 8-XCD configs, and `SimulationEngine::build()` invokes the generic graph partitioner; nothing assigns one partition per XCD. The XCD mapping would need an explicit assignment contract before we document it as typical behavior.

kuhar (commented):
We should verify the decoded mnemonic before counting an encoding as covered (`emulation/rocjitsu/tests/instruction_execution_harness_test.cpp:248`). The generated table pairs a claimed mnemonic with sample words, but this path accepts any successful decode and records `UnimplementedInst` under `te.mnemonic`. If the sample words decode to a different instruction, the new gate can pass under the wrong name; comparing `inst->mnemonic()` with `te.mnemonic` keeps the gate tied to the advertised instruction.

emulation/rocjitsu/docs/configuration.md
lialan: Updated the doc to remove the stale wording, and `num_threads` now describes only the number of PDES partitions/worker threads.

lialan:
> We should verify the decoded mnemonic before counting an encoding as covered (`emulation/rocjitsu/tests/instruction_execution_harness_test.cpp:248`). The generated table pairs a claimed mnemonic with sample words, but this path accepts any successful decode and records `UnimplementedInst` under `te.mnemonic`. If the sample words decode to a different instruction, the new gate can pass under the wrong name; comparing `inst->mnemonic()` with `te.mnemonic` keeps the gate tied to the advertised instruction.

Done. updated so that harness now verifies `inst->mnemonic()` against `te.mnemonic` before counting or executing an encoding.

newling (approved):
LGTM

kuhar (approved):

PR #8179, opened 2026-07-07
rocr: Add debug logging for trampoline/managed variable investigation

shwetagkhatri:
Add comprehensive debug instrumentation to trace managed variable address calculation when trampolines are enabled. Enable with ROCR_LOADER_DEBUG=1.

Logs added to:
- SymbolAddress(): vaddr, section addr/offset, segment info, result
- LoadDefinitionSymbol(): variable name, address, trampoline state
- InstallTrampolinesGfx125x(): kd_vaddr, entry_off, stub addresses, offset delta
- ApplyStaticRelocation(): symbol, type, reladdr, offset, addend
- ApplyDynamicRelocation(): symbol, rel_type, offset, addend, device_addr
- R_AMDGPU_ABS64/RELATIVE64: final symAddr and write location

This helps trace the 0x100 offset bug where managed variable addresses are miscalculated when trampolines modify kernel_code_entry_byte_offset.

## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
@@ -67,6 +67,24 @@
 
 #include "AMDHSAKernelDescriptor.h"
 
+// Debug logging for trampoline/managed variable investigation
+// Enable with: ROCR_LOADER_DEBUG=1
+static bool IsLoaderDebugEnabled() {
+  static int enabled = -1;
+  if (enabled < 0) {
+    const char* env = getenv("ROCR_LOADER_DEBUG");
+    enabled = (env && strcmp(env, "1") == 0) ? 1 : 0;
+  }
+  return enabled == 1;
+}
+
+#define LOADER_DEBUG(fmt, ...) \
+  do { \
+    if (IsLoaderDebugEnabled()) { \
+      fprintf(stderr, "[LOADER_DEBUG] " fmt "\n", ##__VA_ARGS__); \
+    } \
+  } while (0)
+
 using namespace rocr::amd::hsa;
 using namespace rocr::amd::hsa::common;
 
@@ -1549,6 +1567,9 @@ hsa_status_t ExecutableImpl::InstallTrampolinesGfx125x(hsa_agent_t agent) {
   objects.push_back(tramp);               // freed via Destroy() in ~ExecutableImpl
   trampoline_segments_.push_back(tramp);  // frozen in ExecutableImpl::Freeze
 
+  LOADER_DEBUG("InstallTrampolinesGfx125x: installing %zu trampolines, pool_size=0x%lx",
+               n, (unsigned long)pool);
+
   for (size_t i = 0; i < n; ++i) {
     const KdFixup& f = kd_fixups_[i];
     const uint64_t stub_off = i * kTrampolineStubStride;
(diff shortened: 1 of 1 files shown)


shwetagkhatri:
Hi @lmoriche ,
Please let me know if this helps with your investigation. Thanks.

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
```diff
@@ -67,6 +67,24 @@
 
 #include "AMDHSAKernelDescriptor.h"
 
+// Debug logging for trampoline/managed variable investigation
```
cfreeamd, line 70: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** Is this intended as a permanent `ROCR_LOADER_DEBUG` debug facility, or temporary investigation code for @lmoriche?

The PR title says "investigation" and the author's comment says "let me know if this helps with your investigation." If temporary, merging to develop creates cleanup debt. If permanent, consider adding a comment marking it as such and polishing the format strings (see below).

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
```diff
@@ -67,6 +67,24 @@
...
+// Enable with: ROCR_LOADER_DEBUG=1
+static bool IsLoaderDebugEnabled() {
+  static int enabled = -1;
```
cfreeamd, line 73: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** Benign data race on `enabled`. The `static int enabled = -1` declaration is thread-safe (C++11 guarantees thread-safe init for function-local statics). But the subsequent `if (enabled < 0) { enabled = ...; }` is an unprotected read-modify-write if multiple threads race here at startup. In practice benign on x86/aarch64 (all racers read the same `getenv` result and write the same idempotent value), but technically UB. Consider `std::atomic<int>` if strict conformance matters.

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
```diff
@@ -2091,6 +2196,15 @@ hsa_status_t ExecutableImpl::ApplyDynamicRelocation(hsa_agent_t agent, amd::hsa:
     {
       int64_t baseDelta = reinterpret_cast<uint64_t>(relSeg->Address(0)) - relSeg->VAddr();
       uint64_t relocatedAddr = baseDelta + rel->addend();
+
+      LOADER_DEBUG("  -> R_AMDGPU_RELATIVE64: baseDelta=0x%lx addend=0x%lx "
```
cfreeamd, line 2200: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `baseDelta` is `int64_t` (signed), but is formatted as `0x%lx` with `(long)` cast — `%lx` expects `unsigned long`. For a negative delta this prints the two's complement, which can look confusing in traces. Use `%ld` for explicitly signed relocation addends/deltas, or `(unsigned long)` if hex display is intentional (and add a note that it's two's complement).

cfreeamd (commented):
InstallTrampolinesGfx125x, LoadDefinitionSymbol, SymbolAddress (both overloads), ApplyStaticRelocation, and ApplyDynamicRelocation additions look mechanically correct — no debug path modifies program state.

PR #8181, opened 2026-07-07
fix(rocr/hotswap): Register rewritten code object under a debugger-visib…

lmoriche:
…le URI

When the HotSwap loader replaces an agent code object with a COMGR-rewritten one, LoadAgentCodeObjectWithHotswap loaded the rewritten bytes but registered them under the *original* code object's URI:

    callbacks.load_rewritten_code_object(
        ..., rewritten_code_object, rewritten_elf_size, options,
        code_object.uri,          // <-- the ORIGINAL reader's URI
        loaded_code_object);

amd-dbgapi / rocgdb resolve a loaded code object's contents by fetching the bytes named by its URI, then disassemble and read symbols from them. Because the URI still pointed at the original (un-rewritten) ELF, the debugger saw a different code object than the one the GPU executes.

Observed on gfx1250 with Unit_hipModuleGetGlobal_Functional and entry trampolines enabled:

  Thread ... in GPU_func () from
    file:///.../managed_kernel.code#offset=86016&size=3960   <- original (3960),
                                                                not the 4216-byte
                                                                transpiled object
  (gdb) p &x
  $1 = ... 0x...b878 <x>     <- original layout (x at .bss 0x3878)

while the runtime places x at 0x...b978 (transpiled layout). The debugger reported the pre-transpilation address for every symbol the rewrite moved and disassembled code that is not what runs (it had no knowledge of the appended entry-trampoline stubs the kernel is redirected through).

Fix: name the loaded rewritten object by where its bytes actually live -- this process's memory -- with a memory://<pid>#offset=0x<addr>&size=<size> URI. The rewritten buffer is kept alive for the executable's lifetime by RetainRewrittenElfBuffer, and its address is unchanged by the std::move, so the URI stays valid for the debugger. The original-object fallback path is unchanged.

The URI builders (GetUriFromMemoryAddress and the GetUriFromMemory*/GetUriFromFile family, plus their EncodePathname / NormalizeWindowsPathToUtf8 helpers) lived in an anonymous namespace in amd_hsa_loader.cpp and so were unreachable from other translation units. Rather than duplicate the memory:// format in hotswap.cpp, extract the whole family into core/inc/code_object_uri.hpp + core/runtime/code_object_uri.cpp (namespace rocr::amd::hsa::loader). The functions depend only on libc/OS and hsa_file_t, so the new .cpp drops into both the runtime library and the lightweight hotswap_rewrite unit test without pulling in loader/runtime internals. amd_hsa_loader.cpp keeps PAGE_SIZE_MASK and its call sites unchanged; hotswap.cpp now calls the shared GetUriFromMemoryAddress.

After the fix rocgdb resolves the loaded object to memory://<pid>#offset=0x...&size=4216 (the transpiled object) and reports &x at the transpiled address, matching the runtime.

Also adds/updates coverage in hotswap_rewrite_test.cc:
  * RuntimeLoadRewrittenCodeObjectIsDebuggerVisible: drives the real rewrite path through LoadAgentCodeObjectWithHotswap and asserts the URI handed to the loader (what the debugger fetches) is not the original code object's.
  * RuntimeLoadUsesRewrittenCodeObject: updated to assert the new memory:// URI (scheme + rewritten size) instead of the original URI it previously pinned.


dayatsin-amd (approved):

projects/rocr-runtime/rocrtst/suites/functional/hotswap/hotswap_rewrite_test.cc
```diff
@@ -409,7 +409,14 @@ TEST(HotswapRewrite, RuntimeLoadUsesRewrittenCodeObject) {
   EXPECT_EQ(load.calls[0].path, LoadPath::kRewritten);
   EXPECT_NE(load.calls[0].code_object, static_cast<const void*>(kGfx1250MinCo));
   EXPECT_GT(load.calls[0].code_object_size, 0u);
-  EXPECT_EQ(load.calls[0].uri, "memory://gfx1250_min.hsaco");
+  // The rewritten object is registered under a memory:// URI that resolves to
+  // the rewritten bytes in this process (see MakeMemoryUri in hotswap.cpp), not
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** Stale function reference: `MakeMemoryUri` does not exist in `hotswap.cpp`. The actual function used is `GetUriFromMemoryAddress`, defined in `code_object_uri.cpp`.

cfreeamd (commented):

Files that look correct: `code_object_uri.cpp`/`.hpp` (clean extraction, unchanged logic), `amd_hsa_loader.cpp` (deletion + include is clean), `hotswap.cpp` (minimal fix, buffer lifetime comment is accurate), both `CMakeLists.txt` updates, and the test additions in `hotswap_rewrite_test.cc`.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/code_object_uri.cpp
```diff
@@ -0,0 +1,389 @@
+////////////////////////////////////////////////////////////////////////////////
+//
+// The University of Illinois/NCSA
+// Open Source License (NCSA)
+//
+// Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved.
```
lancesix: I guess it might be a lawyer question, but even if the file is new (created in 2026), the contents of the file is not (got moved from amd_hsa_loader.cpp).  The copyright being about the content, it starts before 2026, probably at least 2020 for some bits.

projects/rocr-runtime/rocrtst/suites/functional/hotswap/hotswap_rewrite_test.cc
lmoriche: done

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/code_object_uri.cpp
lmoriche: done

shwetagkhatri (approved):

cfreeamd (approved):

PR #8182, opened 2026-07-07
fix(amdsmi): initialize Device metrics fields to avoid uninitialized reads

SakaSitharammurthy:
## Motivation

Due to uninitialized read of few variables in this api's "amdsmi_topo_get_p2p_status () and amdsmi_get_gpu_memory_partition_config ()" creates memleak.

## JIRA ID

ISSUE ID : https://github.com/ROCm/amdsmi/issues/129

## Test Result
->After running valgrind tool on this example "amd_smi_partition_example.cc" and observed no memory leaks.


projects/amdsmi/src/amd_smi/amd_smi.cc
```diff
@@ -3456,6 +3464,12 @@ amdsmi_status_t amdsmi_get_gpu_accelerator_partition_profile_config(
       rsmi_accelerator_partition_resource_type_t type =
           static_cast<rsmi_accelerator_partition_resource_type_t>(r);
       rsmi_accelerator_partition_resource_profile_t profile;
+      profile.partition_resource = 0;
+      profile.num_partitions_share_resource = 0;
+      profile_config->resource_profiles[resource_index].profile_index = 0;
+      profile_config->resource_profiles[resource_index].resource_type = AMDSMI_ACCELERATOR_MAX;
+      profile_config->resource_profiles[resource_index].partition_resource = 0;
+      profile_config->resource_profiles[resource_index].num_partitions_share_resource = 0;
```
marifamd: @gabrpham Does this impact partitioning in a failure case? Should we max out the profile_index, partition_resource, and num_partitions_share_resource instead?

projects/amdsmi/rocm_smi/src/rocm_smi_device.cc
```diff
@@ -691,7 +691,14 @@ static const std::map<const char*, dev_depends_t> kDevFuncDependsMap = {
...
 Device::Device(std::string p, RocmSMI_env_vars const* e)
-    : monitor_(nullptr), path_(p), env_(e), evt_notif_anon_fd_(-1), m_gpu_metrics_header{0, 0, 0} {
+    : monitor_(nullptr),
+      path_(p),
+      env_(e),
+      evt_notif_anon_fd_(-1),
+      m_gpu_metrics_header{0, 0, 0},
+      m_gpu_metrics_updated_timestamp(0),
+      m_device_id(0),
+      m_partition_id(0) {
```
marifamd, line 706: Justify these three new values

marifamd (changes requested):

projects/amdsmi/src/amd_smi/amd_smi.cc
gabrpham: No impact on failure cases. However, we're not doing bounds checking on resource_profiles[resource_index] and resource_index could go as large as 5 x (profile_config->num_profiles) since the incrementing is occurring within a nested for loop. If that's more than the length of resource_profiles array, then we'll be getting index out of bounds or buffer overflows.

gabrpham (changes requested):

projects/amdsmi/rocm_smi/src/rocm_smi_device.cc
SakaSitharammurthy: POD members with no in-class initializer — read as indeterminate values before their setters run.
0 is the natural default: m_gpu_metrics_updated_timestamp=0 is the "never fetched" sentinel for metrics staleness.
 m_device_id/m_partition_id=0 are the primary device/partition defaults, set later via set_smi_device_id()/set_smi_partition_id().

gabrpham (approved):
Potential for out of bounds error has been addressed. Approved.

marifamd (approved):

PR #8189, opened 2026-07-07
feat(profiler-hub): add Windows/MSVC build support

anujshuk-amd:
## Motivation

This pull request introduces major improvements for Windows support in the profiler-hub library, while also modernizing and simplifying cross-platform code. Key changes include a native MSVC build process with automatic dependency management, Windows-specific handling for SQLite3, and several codebase updates to use C++17 filesystem features. The documentation has been significantly expanded to cover both Linux and Windows workflows, and CMake logic has been updated for more robust and portable builds.

## Technical Details

profiler-hub currently only builds on Linux — the CMake and source code
depend on POSIX-only facilities (`pthread`, `libgen.h`, `sys/stat.h`,
`unistd.h`) and on building SQLite via `git clone` + autotools (`./configure`
+ `make`), none of which work with MSVC on Windows.

roc-optiq needs to consume profiler-hub's writer/reader APIs on Windows, so
this PR adds native Windows/MSVC build support while leaving the existing
Linux build and behavior unchanged. The goals are:

- Build the `profiler-hub` shared and static libraries with MSVC.
- Remove POSIX-only dependencies by moving to portable C++17 `std::filesystem`
  and guarding platform-specific code behind `_WIN32`.
- Obtain SQLite on Windows from the official amalgamation zip (no git, nmake,
  or Tcl toolchain required), keeping the Linux git-clone path intact.
- Ensure the unit tests and benchmarks build and pass on Windows.
- Document the Windows (MSVC) build alongside the existing Linux instructions.
## JIRA ID

TBA

## Test Plan
Same tests will be run.

## Test Result

All existing tests are passing on window

## Submission Checklist


profilers/profiler-hub/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ cmake_minimum_required(VERSION 3.21 FATAL_ERROR)
 project(profiler-hub VERSION 0.1.0)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
```
avansick-amd, line 8: This bumps profiler-hub to C++20 on Linux too, not just Windows —
was that intentional? The PR mentions Linux behavior being unchanged. (No consumer impact that
I can see — it's contained to profiler-hub's own build — but worth calling out if intended, or
scoping back if it just came along for the MSVC path.)

profilers/profiler-hub/source/common/directory.hpp
```diff
@@ -3,59 +3,64 @@
...
 namespace profiler_hub::common
 {
 
 inline std::string
 dirname(const std::string& path)
 {
-    std::string path_copy = path;
-    return ::dirname(path_copy.data());
+    return std::filesystem::path(path).parent_path().string();
 }
```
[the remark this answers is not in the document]
avansick-amd: Good catch that parent_path() returns "" for a bare filename where POSIX dirname returned ".". I traced
the one caller (create_directory_for_database_file in sqlite_backend_impl.hpp) and it's inert
here: the bare-filename case targets the current working directory, which already exists, so
nothing needs creating. makedir("") returns false but the caller discards the return, and
is_directory("") returns false without throwing — so the DB still opens in the cwd exactly as
before. No change needed. (A parent.empty() ? "." : parent fallback would be harmless if we
ever want to be defensive for future callers, but it's not required for correctness today.)

profilers/profiler-hub/cmake/sqlite3.cmake
```diff
@@ -47,81 +55,147 @@ if(PROFILER_HUB_SQLITE3_USE_SYSTEM)
...
+                "https://www.sqlite.org/${SQLITE3_AMALGAMATION_YEAR}/sqlite-amalgamation-${_sqlite3_amalg_version}.zip"
+            )
+
             message(
                 STATUS
-                "[profiler-hub] Optimized clone failed; retrying full clone"
+                "[profiler-hub] Downloading SQLite3 amalgamation ${_sqlite3_version} from ${_sqlite3_download_url}"
             )
+            file(
+                DOWNLOAD ${_sqlite3_download_url}
```
avansick-amd, line 101: Not sure if you know about this, but file(DOWNLOAD) takes an EXPECTED_HASH argument that checks the downloaded
bytes against a pinned hash and hard-fails on a mismatch. On Linux we're effectively
version-pinned already — git clone of version-3.45.3 over a cert-verified transport — but
the Windows path derives that same version into a sqlite.org URL and fetches it with nothing
verifying the bytes are really that release. Since we're pinning the version on Linux, it'd be
good to give Windows the same guarantee. We already do exactly this for the OTF2 source archive
(projects/rocprofiler-sdk/external/otf2, URL_HASH SHA256=...), and sqlite.org publishes a
SHA3-256 for each amalgamation. For the current default (version-3.45.3 →
sqlite-amalgamation-3450300.zip) the drop-in line is:

cmake   EXPECTED_HASH SHA3_256=c514a2640102a43adb1ea17056df3e6772c3b5aea94e0e64d5c819156f540952
(I computed that from the canonical sqlite.org /2024/ artifact — openssl and Python agree —
so it's ready to paste; it'll just need to move whenever SQLITE3_GIT_TAG bumps.) You could
also pair it with TLS_VERIFY ON, since our CMake floor is 3.21 where that defaults off.
One small thing while you're in here: after this, a version bump touches three values that have to
agree — the tag, the SQLITE3_AMALGAMATION_YEAR folder, and this hash. Might be worth keeping them
in one commented block so it's obvious they move together.

profilers/profiler-hub/cmake/sqlite3.cmake
```diff
@@ -145,7 +219,18 @@ else()
...
+    if(NOT MSVC)
+        target_compile_options(profiler-hub-sqlite3-static PRIVATE -O2 -fPIC)
+    endif()
+
+    if(WIN32)
+        # WINDOWS_EXPORT_ALL_SYMBOLS won't re-export symbols from a static lib,
+        # so mark the SQLite API dllexport to share one copy across the process.
+        target_compile_definitions(
+            profiler-hub-sqlite3-static
+            PRIVATE "SQLITE_API=__declspec(dllexport)"
```
avansick-amd, line 231: This export is only here to feed writer_test's direct SQLite calls.
That test should be reworked as an integration test, which removes
the need to export SQLite from the DLL at all.

profilers/profiler-hub/tests/unit/CMakeLists.txt
```diff
@@ -26,11 +26,20 @@ target_link_libraries(
...
 
+if(WIN32)
+    target_include_directories(
+        profiler-hub_unit_tests
+        PRIVATE
+            $<TARGET_PROPERTY:profiler-hub-sqlite3,INTERFACE_INCLUDE_DIRECTORIES>
+    )
+else()
+    target_link_libraries(profiler-hub_unit_tests PRIVATE profiler-hub-sqlite3)
+endif()
```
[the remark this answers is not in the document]
avansick-amd: This include problem is a symptom. The underlying issue is that writer_test.cpp calls into the
SQLite C API directly — it re-opens the produced .db and reads it back to check the writer's
output — which is what creates the <sqlite3.h> dependency this line is trying to satisfy. That
makes it an integration test (writer + real SQLite + real schema), not a unit test.
Suggest pulling it out of tests/unit/ and reworking it as an integration test — a target that
exercises the writer end to end, not an isolated unit. As for how it might read the output back:
rather than calling raw sqlite3_*, it could go through profiler-hub's own public reader_t —
write with the writer, read with the reader, assert the round trip. tests/unit/reader_test.cpp
is already exactly this shape: it opens a storage_t from the db path, wraps it in reader_t, and
asserts on typed results (get_all_nodes(), get_all_processes(), etc.) with no direct SQLite. Following that pattern, the reworked test needs no
<sqlite3.h>, no bundled-SQLite link, and no DLL symbol export — the include wiring here and the symbol-export point above both fall away. (If some assertion in writer_test targets a table the reader doesn't surface,
that's worth a separate look — either it was checking a storage detail no consumer reads, or it points
at a real gap in the reader API — but neither is a reason to keep raw SQLite in the test.) If the team
later wants a check that verifies storage independently of the reader, that belongs in a separate
integration target carrying its own SQLite — e.g. shelling out to the sqlite3 CLI — which stays
test-only and still needs no DLL export. Not this PR's job to do the rework — flagging as a follow-up.

profilers/profiler-hub/source/common/directory.hpp
```diff
@@ -3,59 +3,64 @@
...
-            {
-                status = false;
-                break;
-            }
-        }
-    } while(pos != std::string::npos);
-
-    if(!direxists(path))
+        created.push_back(p);
+    }
```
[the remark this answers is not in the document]
avansick-amd: Agree with Copilot's error_code fix on both lines. Worth noting it's already inconsistent within this PR — create_directories/fs::permissions just below and benchmarks/utility.hpp both use the non-throwing overload, so this is just direxists and the ancestor loop getting missed.

profilers/profiler-hub/CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ downstream consumer of the library.
 
 ### Added
 
+- Windows/MSVC build support (Visual Studio 2022 or 2026 Build Tools).
```
avansick-amd, line 25: The ### Added entry records the Windows/MSVC
support but not the C++ standard bump to C++20. That bump is arguably the more consequential line
for downstream consumers — it raises their compiler floor — so it's worth a changelog entry of its own,
so anyone scanning for a toolchain-requirement change actually sees it.

profilers/profiler-hub/tests/benchmarks/CMakeLists.txt
```diff
@@ -13,12 +13,17 @@ target_link_libraries(
         profiler-hub
         benchmark::benchmark
         benchmark::benchmark_main
-        pthread
         spdlog::spdlog
         profiler-hub-sqlite3
```
avansick-amd, line 17: This link looks redundant — the benchmark doesn't use SQLite directly and
profiler-hub already links it. Worth a look while you're in this block.

avansick-amd (approved):
No Windows job ships with this PR, so the new MSVC build path has no CI coverage — should we add Windows CI in a follow-up PR?

PR #8190, opened 2026-07-07
fix(rocr): Large BAR detection for APU devices with no device-local memory

B0b0A:
## Motivation

Without this fix the following rocrtst tests fail on gfx1151 (Strix Point):
```
  rocrtstFunc.Queue_Create_DeviceMem_RingBuf_Test
  rocrtstFunc.Queue_Create_Batch_Test
```
Both time out after 5 s waiting for a completion signal because the GPU CP cannot fetch packets from the incorrectly-placed ring buffer.

## Technical Details
`GpuAgent` currently sets `large_bar_enabled_` solely from the CPU->GPU link hop count (`num_hop >= 1`). On APU devices such as gfx1151 (Strix Point), the CPU and GPU may appear on separate NUMA nodes, so `num_hop == 1`, but there is no discrete-GPU VRAM behind a PCIe BAR: memory is system RAM shared on-die.

This causes `hsa_amd_queue_create` requests using device-memory queue placement flags to be accepted because `LargeBarEnabled()` returns `true`. In particular, `HSA_AMD_QUEUE_CREATE_DEVICE_MEM_RING_BUF` can allocate the ring buffer from the coarse-grain device-memory region even on a zero-local-memory APU. The GPU command processor then cannot fetch packets from that memory, so kernel dispatch through the queue times out after 5 s. These requests should instead be rejected up front with `HSA_STATUS_ERROR_INVALID_QUEUE_CREATION`.

Using `link_type` is not sufficient here: modern AMD APUs (gfx11.5+) can report `HSA_AMD_LINK_INFO_TYPE_PCIE` in KFD topology even though the GPU is on-die. Similarly, the existing `is_apu_node` heuristic based on `NumCPUCores > 0` does not reliably identify this case, because in SPX compute-partition mode gfx1151 can expose `cpu_cores_count == 0` in KFD node properties.

`HsaNodeProperties::LocalMemSize > 0` is a better runtime discriminator for allowing device-memory queue placement. Large BAR is only meaningful when the GPU exposes device-local memory (VRAM); on APUs in SPX mode, `local_mem_size` is reported as 0 because there is no dedicated VRAM.

This change therefore restricts `large_bar_enabled_` to GPUs that are CPU-reachable and also report nonzero local memory, which causes unsupported device-memory queue-placement requests to fail early instead of succeeding and hanging later. Because `LargeBarEnabled()` is the common gate for device-memory queue placement, this affects both `HSA_AMD_QUEUE_CREATE_DEVICE_MEM_RING_BUF` and `HSA_AMD_QUEUE_CREATE_DEVICE_MEM_QUEUE_DESCRIPTOR`.

## Test Plan

Running rocrtst ([rocr-runtime~git20260702.34b11ec813](https://launchpad.net/~bullwinkle-team/+archive/ubuntu/rocm-exp)) on Ubuntu Resolute machine with gfx1151 before and after the patch.

## Test Result

Before the patch
```
  rocrtstFunc.Queue_Create_DeviceMem_RingBuf_Test
  rocrtstFunc.Queue_Create_Batch_Test
```  
are failing, while after the patch:
```
[ RUN      ] rocrtstFunc.Queue_Create_DeviceMem_RingBuf_Test
###############################################################################

	#### TEST NAME ####
RocR Queue Create API Test

	#### TEST DESCRIPTION ####
Tests hsa_amd_queue_create with various device-memory placement flags:
system memory (default) and device-resident ring buffer.

	#### TEST SETUP ####

	#### TEST EXECUTION ####
The gpu device name is gfx1151
Target HW Profile is HSA_PROFILE_BASE
Test can run on any profile. OK.
  DeviceMemRingBufQueueTest: SKIPPED (agent does not support device-memory ring buffers or Large BAR is unavailable)

	#### TEST RESULTS ####

	#### TEST CLEAN UP ####
[       OK ] rocrtstFunc.Queue_Create_DeviceMem_RingBuf_Test (470 ms)
[ RUN      ] rocrtstFunc.Queue_Create_Batch_Test
###############################################################################

	#### TEST NAME ####
RocR Queue Create API Test

	#### TEST DESCRIPTION ####
Tests hsa_amd_queue_create with various device-memory placement flags:
system memory (default) and device-resident ring buffer.

	#### TEST SETUP ####

	#### TEST EXECUTION ####
The gpu device name is gfx1151
Target HW Profile is HSA_PROFILE_BASE
Test can run on any profile. OK.
  BatchQueueCreateTest: SKIPPED (agent does not support device-memory flags or Large BAR is unavailable)

	#### TEST RESULTS ####

	#### TEST CLEAN UP ####
[       OK ] rocrtstFunc.Queue_Create_Batch_Test (473 ms)
```

## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
@@ -263,7 +263,7 @@ GpuAgent::GpuAgent(HSAuint32 node, const HsaNodeProperties& node_props, bool xna
   auto link_info = core::Runtime::runtime_singleton_->GetLinkInfo(first_cpu->node_id(), node_id());
   xgmi_cpu_gpu_ = (link_info.info.link_type == HSA_AMD_LINK_INFO_TYPE_XGMI);
 
-  if (link_info.num_hop >= 1) {
+  if (link_info.num_hop >= 1 && !properties_.Integrated) {
     large_bar_enabled_ = true;
   }
 


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
dayatsin-amd: Can you try to use properties_.Integrated to determine APU instead:

 if (link_info.num_hop >= 1 && !properties_.Integrated) {

dayatsin-amd (changes requested):

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -238,7 +238,8 @@ GpuAgent::GpuAgent(HSAuint32 node, const HsaNodeProperties& node_props, bool xna
   auto link_info = core::Runtime::runtime_singleton_->GetLinkInfo(first_cpu->node_id(), node_id());
```
cfreeamd, line 263: test

cfreeamd (commented):
test 1

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
cfreeamd, line 263: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** `LocalMemSize` is hardcoded to `0` in `libhsakmt/src/dxg/topology.cpp:745` for all Windows/DXG-attached devices. After this change, a Windows dGPU that previously had `large_bar_enabled_ = true` (via `num_hop >= 1`) will now have `large_bar_enabled_ = false`. The device-mem ring-buffer queue path is separately gated off for DXG via `disable_dev_mem_queue_buf()`, so queue creation is unaffected. However `amd_loader_context.cpp:356` also branches on `LargeBarEnabled()` to choose between `memcpy` and DMA-copy for code-object loading — with this change DXG dGPUs would always take the slower DMA-copy path. Not a correctness regression (both paths produce the same result) but potentially a latency regression for code-object loading on Windows discrete GPUs.

Suggested follow-up: populate `LocalMemSize` from the Windows adapter query in `dxg/topology.cpp` for discrete GPUs, or add a comment documenting that the DXG topology always forces the DMA-copy code-object path.

cfreeamd (commented):
Files that look correct: the change to `projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp` is the only file modified and it uses the right discriminator — `LocalMemSize` comes from KFD sysfs `local_mem_size`, which is 0 on APUs with no dedicated VRAM and non-zero on dGPUs and APUs with dedicated HBM (e.g. MI300A).

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
B0b0A: Tried with  `&& !properties_.Integrated` and it is OK:
```
[ RUN      ] rocrtstFunc.Queue_Create_DeviceMem_RingBuf_Test
###############################################################################

	#### TEST NAME ####
RocR Queue Create API Test

	#### TEST DESCRIPTION ####
Tests hsa_amd_queue_create with various device-memory placement flags:
system memory (default) and device-resident ring buffer.

	#### TEST SETUP ####

	#### TEST EXECUTION ####
The gpu device name is gfx1151
Target HW Profile is HSA_PROFILE_BASE
Test can run on any profile. OK.
  DeviceMemRingBufQueueTest: SKIPPED (agent does not support device-memory ring buffers or Large BAR is unavailable)

	#### TEST RESULTS ####

	#### TEST CLEAN UP ####
[       OK ] rocrtstFunc.Queue_Create_DeviceMem_RingBuf_Test (439 ms)
[ RUN      ] rocrtstFunc.Queue_Create_Batch_Test
###############################################################################

	#### TEST NAME ####
RocR Queue Create API Test

	#### TEST DESCRIPTION ####
Tests hsa_amd_queue_create with various device-memory placement flags:
system memory (default) and device-resident ring buffer.

	#### TEST SETUP ####

	#### TEST EXECUTION ####
The gpu device name is gfx1151
Target HW Profile is HSA_PROFILE_BASE
Test can run on any profile. OK.
  BatchQueueCreateTest: SKIPPED (agent does not support device-memory flags or Large BAR is unavailable)

	#### TEST RESULTS ####

	#### TEST CLEAN UP ####
[       OK ] rocrtstFunc.Queue_Create_Batch_Test (442 ms)
```
Changes have been pushed. Thank you for the suggestion!

dayatsin-amd (approved):

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -263,7 +263,7 @@ GpuAgent::GpuAgent(HSAuint32 node, const HsaNodeProperties& node_props, bool xna
...
   xgmi_cpu_gpu_ = (link_info.info.link_type == HSA_AMD_LINK_INFO_TYPE_XGMI);
 
-  if (link_info.num_hop >= 1) {
+  if (link_info.num_hop >= 1 && !properties_.Integrated) {
```
cfreeamd, line 266: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** Switching from `LocalMemSize > 0` to `!properties_.Integrated` resolves the gfx1151 case cleanly, but `Integrated` is set from `AMDGPU_IDS_FLAGS_FUSION` and the header explicitly documents it as covering both small APU and **APP APU** (MI300A). MI300A has `Integrated=1` and `num_hop>=1` (XGMI/Infinity Fabric), so before this PR it had `large_bar_enabled_=true` (HBM → LocalMemSize>0); after this PR it becomes `false`.

That flips `LargeBarEnabled()` across all 6 call sites — device-mem ring buffer (`amd_aql_queue.cpp:604`), code-object DMA-copy path (`amd_loader_context.cpp:356`), SDMA path selection (`amd_sdma_queue.cpp:32,96`), DMA-BUF export (`runtime.cpp:3756`), PM4 perf-counter fallback (`amd_gpu_agent.cpp:3662`).

On MI300A the HBM is coherent and reachable by the CP, so the device-mem ring buffer and CPU-memcpy code-object paths were valid there — disabling Large BAR forces slower fallbacks. Is this regression on MI300A intentional (treat all APUs uniformly), or does MI300A need to remain `large_bar_enabled_=true`? Worth confirming before merge given MI300A is a shipping product.

cfreeamd (commented):

Prior [Important] (DXG `LocalMemSize=0` gap) and comment nits are resolved. The switch from `LocalMemSize>0` to `!properties_.Integrated` is correct for the stated gfx1151 target. One new question raised above about MI300A (APP APU) scope — see inline comment.

Issue #8186, opened 2026-07-07
Issue] Linux gfx94X-dcgpu rocprofiler-compute test: test-pc-sampling-collector fails — `librocm_sysdeps_dw.so.1: cannot open shared object file`


TriveniTadapaneni:
## Context:

- Workflow: Multi-Arch CI
- Run: https://github.com/ROCm/TheRock/actions/runs/28826002620
- Platform/arch: Linux gfx94x-dcgpu
- Failing job: https://github.com/ROCm/TheRock/actions/runs/28826002620/job/85557849505?pr=6371

## Summary

The `test-pc-sampling-collector` test (CTest entry 51/49) in the rocprofiler-compute test suite fails at runtime on gfx94X-dcgpu runners because `librocm_sysdeps_dw.so.1` cannot be found by the dynamic linker.

## Observed error

```
/__w/TheRock/TheRock/build/libexec/rocprofiler-compute/tests/test-pc-sampling-collector: \
  error while loading shared libraries: librocm_sysdeps_dw.so.1: cannot open shared object file: \
  No such file or directory
49/49 Test #51: test-pc-sampling-collector .......................***Failed    0.00 sec
98% tests passed, 1 tests failed out of 49
```

Job: https://github.com/ROCm/TheRock/actions/runs/28826002620/job/85557849505

## Suspected Root cause

Rocm-systems bump PR https://github.com/ROCm/TheRock/pull/6371 included commit [`91e88d7f`](https://github.com/ROCm/rocm-systems/commit/91e88d7f99638f584820751b6a88dd54cd4a0bd3) ("Register missing unit tests in CTest #7427"), which:

1. Registered `test-pc-sampling-collector` as a CTest test for the first time.
2. Fixed its CMake install destination (`PROJECT_NAME` → `CMAKE_PROJECT_NAME`), so the binary is now correctly packaged in `rocprofiler-compute_test_generic.tar.zst` and executes in CI.

The binary links (transitively via the static `pc-sampling-collector` lib) against `libdw::libdw`, which TheRock provides as `librocm_sysdeps_dw.so.1` installed into `lib/rocm_sysdeps/lib/`. At test time, `LD_LIBRARY_PATH` for rocprofiler-compute tests only includes `{ROCM_PATH}/lib` (configured in `test_runner.py`), so the loader cannot find the library even though `sysdeps_lib_generic.tar.zst` is downloaded and the file is present.

## Impact:

[...]

PR #8191, opened 2026-07-07
fix(rocprofiler-compute) Remove redundant test INSTALL_RPATH hacks

abchoudh-amd:
## Motivation

Bugfix: `test-pc-sampling-collector` fails in CI with `error while loading shared libraries: librocm_sysdeps_dw.so.1: cannot open shared object file`. The hardcoded `INSTALL_RPATH` on the test targets does not fix this: TheRock's post-process overrides any local `INSTALL_RPATH`, and the hardcoded path omitted `lib/rocm_sysdeps/lib` where the missing library lives.

## Technical Details

- Remove the redundant `INSTALL_RPATH` blocks from `pc_sampling_collector/tests/CMakeLists.txt` and `rocprofiler_compute_tool/tests/CMakeLists.txt`.
- TheRock computes the authoritative RPATH (including sysdeps) for these test binaries relative to their install origin. Origin handling for both tests lives in TheRock's post-hook (paired PR ROCm/TheRock#6386).

JIRA ID - AIPROFCOMP-710

## Test Plan

- [ ] Run CI

## Test Result

- [ ] CI passes

## Submission Checklist


jamessiddeley-amd (approved):
LGTM

vedithal-amd (changes requested):
blocking for review - think this needs change in therock

vedithal-amd:
Fixes https://github.com/ROCm/rocm-systems/issues/8186

vedithal-amd:
Expanding the scope of this PR: instead of adding an install RPATH, we should **remove** the redundant `INSTALL_RPATH` hacks from both test CMakeLists and fix this properly in TheRock.

Why the current hardcoded RPATH doesn't work: TheRock's global post-process (`therock_global_post_subproject.cmake`) overrides any `INSTALL_RPATH` set here, so it's a no-op in TheRock builds. And the path points to `lib/rocprofiler-compute`, but the missing `librocm_sysdeps_dw.so.1` lives in `lib/rocm_sysdeps/lib`, which it doesn't cover.

The real mechanism: TheRock computes the full RUNPATH (including `rocm_sysdeps/lib`) relative to a target's `THEROCK_INSTALL_RPATH_ORIGIN`. `post_hook_rocprofiler-compute.cmake` sets this for `test-rocprofiler-compute-tool` (verified against a build: its RUNPATH includes `$ORIGIN/../../../lib/rocm_sysdeps/lib` and `ldd` resolves cleanly). `test-pc-sampling-collector` is just missing from that hook.

Updated plan for this PR:
- Remove the manual `INSTALL_RPATH` block from `pc_sampling_collector/tests/CMakeLists.txt` and `rocprofiler_compute_tool/tests/CMakeLists.txt`.

Paired TheRock PR adds `test-pc-sampling-collector` to the post-hook so RPATH (including sysdeps) is computed correctly: ROCm/TheRock#6386

vedithal-amd:
Pushed the scope change on your behalf: dropped the hardcoded `INSTALL_RPATH` from both test CMakeLists, since TheRock's post-process is authoritative and the paired hook change (ROCm/TheRock#6386) sets the correct origin (incl. `lib/rocm_sysdeps/lib`) for both tests. Updated the title/body to match.

vedithal-amd (approved):
Approving: correct fix. The hardcoded RPATH was a no-op under TheRock and missed the sysdeps dir; removing it and letting TheRock's post-hook (ROCm/TheRock#6386) own the RPATH is the right approach. Merge once both CIs are green.

PR #8193, opened 2026-07-07
fix (ROCM-27029): net_ib_cast BY_INDEX AllToAll hang from cross-cache-line CTS-fifo

nikitaxgusev:
## Motivation

Fix net_ib_cast BY_INDEX AllToAll hang from cross-cache-line CTS-fifo reads

## Technical Details

For BY_INDEX matching the sender gates on a fresh CTS-fifo readiness word (idx == fifoHead+1) and then reads the receiver request-slot index (rxReqIndex) to echo in the RDMA-Write-with-IMM that routes the receiver's completion. Both values come from a fifo slot the receiver recycles every NET_IB_MAX_REQUESTS posts via a single RDMA WRITE of the whole element.

With a 64-bit idx the element placed rxReqIndex (byte 64) in a different 64-byte cache line than idx (byte 56). The NIC's RDMA write is observed by the polling sender atomically only within one cache line, so on a recycled slot the sender could observe a fresh idx paired with the previous occupant's rxReqIndex, misroute the completion to the wrong recv request and deadlock large AllToAll transfers. Disabling PCI relaxed ordering does not help: the polling sender reads target memory directly and never waits for the write completion, so it is cross-cache-line visibility, not write ordering, that tears the read.

Shrink idx from 64 to 32 bits so idx (@56) and rxReqIndex (@60) fall in the same 64-byte cache line and become visible together. The gate and the receiver publish truncate the generation consistently, and it stays unique across the 256-deep ring (wrap period 2^32 >> ring depth).

## JIRA ID

Fixes ROCM-27029

## Test Plan

```
$OMPI/bin/mpirun --np 16 -H "$HOSTS"   --bind-to numa   --mca pml ob1 --mca osc ^ucx --mca btl tcp,self --mca btl_tcp_if_include ens50f0   -x LD_LIBRARY_PATH=$RCCL:$OMPI/lib:/opt/rocm/lib   -x PATH=$OMPI/bin:/opt/rocm/bin:/usr/bin:/bin   -x NCCL_SOCKET_IFNAME=ens50f0   -x NCCL_NET=IB-CAST   -x NCCL_IB_QPS_PER_CONNECTION=1   -x RCCL_IB_QPS_PER_P2P=1   -x RCCL_IB_QP_SCHED_ENABLE=0   -x NCCL_IB_PCI_RELAXED_ORDERING=1   -x RCCL_CTS_OFFLOAD_ENABLED=0   -x NCCL_NET_OPTIONAL_RECV_COMPLETION=1   -x NCCL_GDR_FLUSH_DISABLE=1   -x NCCL_DEBUG=WARN $TESTS/alltoall_perf -b 1K -e 16G -f 2 -g 1 -n 20 -c 1 -w 2
```

## Test Result

alltoall_perf for 16 ranks two nodes - pass

## Submission Checklist

### DIFF
--- a/projects/rccl/src/transport/net_ib_cast/common_cast.h
+++ b/projects/rccl/src/transport/net_ib_cast/common_cast.h
@@ -356,9 +356,9 @@ struct alignas(64) ncclIbSendFifo {
   uint32_t rkeys[NCCL_IB_MAX_DEVS_PER_NIC];
   uint32_t nreqs;
   uint32_t tag;
-  uint64_t idx;
+  uint32_t idx;
   uint16_t rxReqIndex;
-  char padding[14];
+  char padding[2];
 };
 
 #define MAX_INLINE_DATA_SIZE 24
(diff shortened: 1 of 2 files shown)


alexander-sannikov (approved):

nikitaxgusev:
%RCCL_TEST_CONFIG
mi300x_mellanox_ib.json

alexander-sannikov:
%RCCL_TEST_CONFIG
mi300x_mellanox_ib.json

PR #8200, opened 2026-07-07
libhsakmt: Add size validation to IPC import path

Sundance636:
## Motivation
The IPC handle import path in hsakmt_fmm_register_shared_memory reads a size value from userspace-supplied HsaSharedMemoryStruct without validation. While the export path (hsakmt_fmm_share_memory) enforces a 16 TiB limit and ensures safe truncation to 32-bit page count, an attacker can bypass this by manually constructing a HsaSharedMemoryStruct with a valid kernel handle but a maliciously small SizeInPages value.

Attack scenario:
1. Attacker allocates 100 GB GPU memory via alternative path (direct ioctl, dmabuf, or other legitimate allocation)
2. Attacker obtains valid kernel IPC handle for this allocation
3. Attacker manually constructs HsaSharedMemoryStruct with:
   - ShareHandle = valid kernel handle (100 GB allocation)
   - SizeInPages = 1 (claims only 4 KB)
4. Victim calls hsakmt_fmm_register_shared_memory with malicious handle
5. Victim allocates VA space for 4 KB based on SizeInPages
6. Kernel maps actual 100 GB allocation into 4 KB VA reservation
7. Result: 100 GB mapping overlaps adjacent VA regions, causing memory corruption and potential information disclosure

The root cause is that the import side trusts the size field without verification. While the kernel validates the handle itself, it has no way to verify that the VA reservation is correctly sized because the AMDKFD_IOC_IPC_IMPORT_HANDLE ioctl does not receive a size parameter.

This patch adds validation on the import side:
- Enforce the same 16 TiB limit as the export side
- Check for arithmetic overflow in shift operations
- Validate that size is non-zero
- Verify that requested size fits within the target aperture
## JIRA ID
JIRA ID: ROCM-26865

## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/libhsakmt/src/fmm.c
+++ b/projects/rocr-runtime/libhsakmt/src/fmm.c
@@ -4309,8 +4309,16 @@ HSAKMT_STATUS hsakmt_fmm_register_shared_memory(HsaKFDContext *ctx,
 	const HsaSharedMemoryStruct *SharedMemoryStruct =
 		to_const_hsa_shared_memory_struct(SharedMemoryHandle);
 	HSAuint64 SizeInPages = SharedMemoryStruct->SizeInPages;
+	HSAuint64 SizeInBytesCalc;
 	HsaMemFlags mflags;
 	struct hsa_kfd_fmm_context *fmm_ctx = ctx->fmm_context;
+
+	SizeInBytesCalc = SizeInPages << PAGE_SHIFT;
+
+	if (SizeInPages == 0) {
+		pr_err("IPC import: size cannot be zero\n");
+		return HSAKMT_STATUS_INVALID_PARAMETER;
+	}
 
 	if (gpu_id_array_size > 0 && !gpu_id_array)
 		return HSAKMT_STATUS_INVALID_PARAMETER;
@@ -4322,6 +4330,14 @@ HSAKMT_STATUS hsakmt_fmm_register_shared_memory(HsaKFDContext *ctx,
 	aperture = fmm_get_aperture(fmm_ctx, SharedMemoryStruct->ApeInfo);
 	if (!aperture)
 		return HSAKMT_STATUS_INVALID_PARAMETER;
+
+	HSAuint64 aperture_size = VOID_PTRS_SUB(aperture->limit, aperture->base) + 1;
+	if (SizeInBytesCalc > aperture_size) {
+		pr_err("IPC import: size 0x%llx exceeds aperture range 0x%llx\n",
+				(unsigned long long)SizeInBytesCalc,
+		(unsigned long long)aperture_size);
+		return HSAKMT_STATUS_INVALID_PARAMETER;
+	}
 
 	pthread_mutex_lock(&aperture->fmm_mutex);
 	reservedMem = aperture_allocate_area(aperture, NULL,


dayatsin-amd (approved):

cfreeamd (commented):

Reviewed `libhsakmt/src/fmm.c`. The aperture-range check (lines 4328–4334) is the only genuinely useful new guard: it prevents a fabricated handle from reserving more VA than the target aperture holds.

projects/rocr-runtime/libhsakmt/src/fmm.c
```diff
@@ -4291,9 +4291,29 @@ HSAKMT_STATUS hsakmt_fmm_register_shared_memory(HsaKFDContext *ctx,
 	const HsaSharedMemoryStruct *SharedMemoryStruct =
 		to_const_hsa_shared_memory_struct(SharedMemoryHandle);
 	HSAuint64 SizeInPages = SharedMemoryStruct->SizeInPages;
+	HSAuint64 SizeInBytesCalc;
 	HsaMemFlags mflags;
 	struct hsa_kfd_fmm_context *fmm_ctx = ctx->fmm_context;
 
+	if (SizeInPages > (UINT64_MAX >> PAGE_SHIFT)) {
```

**[Important]** Overflow check is dead code — HSAuint32 can never trigger it

`SizeInPages` is `HSAuint32` (line 353). After widening to `HSAuint64` at line 4293, its maximum value is `UINT32_MAX = 0xFFFFFFFF`. `UINT64_MAX >> PAGE_SHIFT` (≈ `0x000FFFFFFFFFFFFF` for 4KB pages) is vastly larger, so this condition is always false. The check is harmless but misleading — it implies a 64-bit value could arrive here. Either document why this is future-proofing, or remove it.

projects/rocr-runtime/libhsakmt/src/fmm.c
```diff
@@ -4291,9 +4291,29 @@ HSAKMT_STATUS hsakmt_fmm_register_shared_memory(HsaKFDContext *ctx,
...
+		pr_err("IPC import: SizeInPages 0x%llx causes overflow in shift\n",
+				(unsigned long long)SizeInPages);
+		return HSAKMT_STATUS_INVALID_PARAMETER;
+	}
+
+	SizeInBytesCalc = SizeInPages << PAGE_SHIFT;
+
+	if (SizeInBytesCalc >= (1ULL << ((sizeof(HSAuint32) * 8) + PAGE_SHIFT))) {
```

**[Important]** 16 TiB ceiling also unreachable for HSAuint32

Same reason as above. The largest possible `SizeInBytesCalc` from a 32-bit `SizeInPages` is `UINT32_MAX << PAGE_SHIFT = 0xFFFFF000_0000` — exactly 4KB below 16 TiB — so this `>=` comparison is always false. Consider a code comment explaining this is a defensive assertion for future type widening.

projects/rocr-runtime/libhsakmt/src/fmm.c

**[Important]** Stated attack vector (undersized SizeInPages) not addressed

The commit message describes the primary attack as: attacker sets `SizeInPages = 1` (too-small) to make the victim reserve only 4KB of VA, while the kernel maps a 100GB BO into that region — causing adjacent-VA aliasing. None of the four new checks address this direction: the overflow guard, 16 TiB limit, zero check, and aperture check all catch values that are *too large*. A maliciously small `SizeInPages` still passes all of them.

A real fix requires the kernel to return the actual BO size from `AMDKFD_IOC_IPC_IMPORT_HANDLE` so userspace can verify the claimed `SizeInPages` against the actual allocation. This PR adds useful defense-in-depth for the opposite direction only, and should not be presented as fixing the described attack.

Sundance636: for the AI review's second part there, the kernel already returns the BO size to user space via libdrm in the ipc dmabuf path, 
> HSAKMT_STATUS status = HSAKMT_CALL(hsaKmtHandleImport(&desc, &res, &hflags));
> :
> :
> handle->handle = reinterpret_cast<uint64_t>(res.buf_handle);
> handle->size = res.alloc_size;

kentrussell (approved):

PR #6055, opened 2026-05-13
feat(rocprofiler-sdk): attach ATT support

meserve-amd:
## Motivation

Add ATT support to attach.

## Technical Details

Makes thread_trace aware of the attach code_object scheme and registers it for code_object iteration and creation.

TODO: add description of binary saving 

## JIRA ID

JIRA ID: AIPROFSDK-802

## Test Plan

Existing attach tests
More tests to be added

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #8202, opened 2026-07-07
fix(rocprofiler-sdk): block final output in attach

meserve-amd:
- Defensive fix for services that still have output after detach calls flush
- Prevents an issue where output from an attach session was overwritten when the app exited
  - Outputs a warning in this scenario instead

## Motivation

In some cases, output is generated after a flush. While this is incorrect and should be fixed on its own, we also want to change current behavior where a new output file is generated overwriting any output from the attach/detach cycle. This PR makes it so a warning is generated instead and output is suppressed.

This bug was encountered while implementing #6055, and the output for ATT was fixed there. However, it has been reported in other places as well.

## Technical Details

Scenarios considered for this change
Attach -> App Exits -> Finalize should generate output
Attach -> Detach -> App Exits -> Finalize should NOT generate output
These cases should also hold for any number of attach/detach cycles before

## JIRA ID

JIRA ID: AIPROFSDK-931

## Test Plan

Existing CI

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/tool.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/tool.cpp
@@ -137,6 +137,10 @@ namespace
 // Thread for safe cleanup output generation
 auto output_generation_thread = common::Synchronized<std::optional<std::thread>>{};
 
+// Tracks whether tool_detach produced output for the current attach session.
+// Set true by tool_detach, reset false by tool_attach (new session), read by tool_fini.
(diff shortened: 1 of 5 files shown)


meserve-amd:
>     * ❌ Unit Test

Fixes an existing intermittent issue and is also a defensive fix for handling changes in the future. No test needs to be added for speculative behaviors. Ignoring check and removing tag.

bgopesh (commented):
No regression test for the attach/detach/exit state machine

bgopesh (approved):

ApoKalipse-V (approved):

PR #8203, opened 2026-07-07
fix(rocprofiler-sdk): include fmt/format.h instead of fmt/core.h for fmt::format

### LINKED ISSUE #8183: [Issue]: fmt changed behavior of `fmt/core.h` which is heavily used
### Problem Description

fmt makes `fmt/core.h` include only `fmt/base.h` by default since v12.2.0 (via https://github.com/fmtlib/fmt/commit/0007426c2d8d22df6e56d3b4d109415242e683e0). 

The header is used in some rocm packages and the change is causing them fails to be compiled with fmt v12.2.0. The use of the header should be replaced with `fmt/format.h`.

### Operating System

Ubuntu 26.04 LTS

### CPU

AMD Ryzen 5600g

### GPU

AMD Instinct MI50

### ROCm Version

master

### ROCm Component

rocprofiler-register

### Steps to Reproduce

_No response_

### Additional Information

_No response_

darren-amd:
## Motivation

Since fmt v12.2.0, `<fmt/core.h>` was made equivalent to `<fmt/base.h>` by default (fmtlib/fmt@0007426c). `base.h` is a lightweight subset that no longer declares the `std::string`-returning `fmt::format` / `fmt::vformat`, which now live only in `<fmt/format.h>`.

As a result, any source that does `#include <fmt/core.h>` and calls `fmt::format(...)` fails to compile against system fmt 12.2.0:

```
error: 'format' is not a member of 'fmt'
```

Resolves ROCm/rocm-systems#8183.

## Technical Details

Replace `#include <fmt/core.h>` with `#include <fmt/format.h>` in the files that call `fmt::format` / `fmt::vformat`, across `rocprofiler-register`, `rocprofiler-sdk`, and `profiler-hub` (44 files).

- Only files that use the format.h-only symbols are changed; files that use just `fmt::print` / `fmt::format_to` (satisfied by `base.h`) are left on `core.h`.
- Where a file already included `<fmt/format.h>` next to `<fmt/core.h>`, the redundant `core.h` line is dropped.
- The change is backward compatible: `<fmt/format.h>` is a strict superset of `<fmt/core.h>` and has provided `fmt::format` in every fmt version these components build against. Verified against the pinned submodule (fmt 11.1.4) and system fmt 12.2.0.

## Test Plan

Built each affected component against a locally-installed fmt 12.2.0, forcing the `find_package(fmt)` path (`*_BUILD_FMT=OFF`) that the report hits:

- `rocprofiler-register` — full library build
- `rocprofiler-sdk` — full object/shared libraries plus the changed test targets (`counter-tests`, `aql-test`, `parser-test`)
- `profiler-hub` — full static + shared library build

## Test Result

- With the fix: all three components build cleanly against fmt 12.2.0 (0 compile errors).

## Submission Checklist


jrmadsen (approved):

avansick-amd (approved):
Confirmed the 5 profiler-hub include swaps are unchanged since jrmadsen's approval (the 2026-07-20 merge-from-develop didn't touch anything under profilers/profiler-hub/), and the Copilot-flagged scanner.l/scanner.cpp generator-sync concern was already addressed in a same-day follow-up commit. LGTM.

PR #8206, opened 2026-07-07
clr: Defer HostQueue teardown off the async-events thread

jaydeeppatel1111:
## Motivation
Running the HIP graph test Unit_hipGraphAddMemAllocNode_Negative_With_Cloneed_Graph (and, more generally, any workload that destroys a GraphExec whose launch completion is signaled by ROCr) intermittently hangs at process shutdown.

Root cause — self-deadlock on the async-events thread
A graph launch completes. ROCr delivers that completion on its async-events thread (HsaAmdSignalHandler).
The completion callback cascades into ~GraphExec, which drops the last reference to the graph's internal HostQueue (stream).
Dropping the last reference calls ReferenceCountedObject::release() → HostQueue::terminate().
terminate() flushes a marker and calls awaitCompletion(), i.e. it blocks waiting for a queue completion to be signaled — but the thread that signals completions is the very async-events thread we are currently running on.
The thread ends up waiting on itself → deadlock, and the process never exits.

This is a lifetime/threading issue: the blocking drain in terminate() is correct on an app thread, but must never run on the async handler thread.

## Technical Details

Regression from https://github.com/ROCm/rocm-systems/commit/5de80e1c1c611c0ec6c4c8c0e59cbea8e51f3d6e.

Keep terminate()'s existing blocking drain, but never execute it on the async-events thread. When terminate() is entered from the async handler, re-arm the object and hand it to an app thread, which re-invokes terminate() off the async thread and completes the normal drain.

## JIRA ID

[ROCM-27565](https://amd-hub.atlassian.net/browse/ROCM-27565)

## Test Plan

Run Unit_hipGraphAddMemAllocNode_Negative_With_Cloneed_Graph for multiple times.

## Test Result

ran Unit_hipGraphAddMemAllocNode_Negative_With_Cloneed_Graph 30 times and it passed.

## Submission Checklist

[ROCM-27565]: https://amd-hub.atlassian.net/browse/ROCM-27565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/projects/clr/rocclr/device/device.hpp
+++ b/projects/clr/rocclr/device/device.hpp
@@ -2409,6 +2409,18 @@ class Device : public RuntimeObject {
   static bool IsGPUInError() { return (gpu_error_.load(std::memory_order_relaxed) != CL_SUCCESS); }
   static cl_int GetGPUError() { return gpu_error_.load(std::memory_order_relaxed); }
 
+  //! True if the calling thread is inside a device async completion handler,
+  //! where blocking on GPU completion self-deadlocks. Base returns false.
+  virtual bool isInAsyncSignalHandler() const { return false; }
+
+  //! Release `obj`, deferring to an app thread if called from an async handler
+  //! (where blocking teardown self-deadlocks). Base releases inline.
+  virtual void deferReleaseObject(ReferenceCountedObject* obj) {
+    if (obj != nullptr) {
+      obj->release();
+    }
+  }
+
   bool GetHandleForAddressRange(void* dev_ptr, size_t size, void* handle);
 
   // Registers a memory object allocated via hostcall for later cleanup.
--- a/projects/clr/rocclr/device/rocm/rocdevice.cpp
+++ b/projects/clr/rocclr/device/rocm/rocdevice.cpp
@@ -228,7 +228,10 @@ Device::~Device() {
   delete xferQueue_;
   xferQueue_ = nullptr;
 
-  // Final drain of deferred destroys on the main thread before hsa_shut_down.
+  // Final drain before hsa_shut_down. Objects first: a deferred HostQueue
+  // teardown re-runs terminate() and may release HW queues that the destroy
+  // drain below then frees.
+  DrainDeferredObjectReleases();
   DrainDeferredQueueDestroys();
 
   for (auto& it : queuePool_) {
@@ -3240,10 +3243,12 @@ hsa_queue_t* Device::acquireQueue(uint32_t queue_size_hint, bool coop_queue,
                                   bool dedicated_queue, hsa_queue_t* preferred,
                                   const std::unordered_set<uint64_t>* excluded_ids,
                                   void** metadata_ring_buffer) {
-  // App-thread context: flush queues deferred from the async thread once they
-  // reach the batch threshold, bounding live deferred queues.
+  // App thread: bound deferred queues/objects by draining past a threshold.
   if (DeferredQueueCount() >= kDeferredQueueDrainThreshold) {
     DrainDeferredQueueDestroys();
+  }
+  if (DeferredObjectReleaseCount() >= kDeferredQueueDrainThreshold) {
+    DrainDeferredObjectReleases();
   }
 
   hsa_amd_queue_priority_t queue_priority;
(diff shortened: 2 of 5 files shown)


anugodavar:
Ideally we are destroying the graph internal streams only after all the gpu work is done. We are doing finish in destructor just to clean up cpu commands. If we defer HW-queue slots held longer, latency injected into an app thread and delay app's work. I think deferring to app thread is not good idea.

jaydeeppatel1111:
> Ideally we are destroying the graph internal streams only after all the gpu work is done. We are doing finish in destructor just to clean up cpu commands. If we defer HW-queue slots held longer, latency injected into an app thread and delay app's work. I think deferring to app thread is not good idea.

Ok..Alternatives already tried and merited as below. **Are you suggesting #1**?

Context: Shutdown hang = self-deadlock. A graph launch completion on ROCr's async-events thread cascades into ~GraphExec → last-ref drop of the internal HostQueue → terminate() → awaitCompletion(), which blocks waiting for a completion that only that same async thread can signal. #3 defers the HostQueue teardown off the async thread. Two other fixes were prototyped and rejected:

#1 — Guard-only (skip flush/await on the async thread)
In HostQueue::terminate(), when on the async thread, just skip the marker flush + awaitCompletion() and proceed to delete.
Pro: smallest diff; kills the hang (verified 30/30).
Con: silently skips draining an in-flight submission batch (commands left in SUBMITTED state). No crash in the repro, but it drops the completion guarantee → latent risk of leaks / unfinished work. It trades a deadlock for a correctness hole.
Verdict: rejected — fixes the symptom, not safe in general.

#2 — Defer the GraphExec release itself
Push the GraphExec/AccumulateCommand last-ref drop to an app thread (defer higher up, at the graph object rather than the queue).
Pro: kills the hang; conceptually closest to the culprit commit that tied GraphExec lifetime to the async handler.
Con: moves ~GraphExecSegmented past HSA teardown at shutdown → segfault (rc=139): the destructor allocates/touches HSA after it's torn down. Would need an eager app-thread drain at multiple hot paths to bound lifetime and avoid running destructors too late — more surface area and fragile ordering.
Verdict: parked — removes the hang but introduces a shutdown crash; not worth the added complexity.

jaydeeppatel1111:
> Ideally we are destroying the graph internal streams only after all the gpu work is done. We are doing finish in destructor just to clean up cpu commands. If we defer HW-queue slots held longer, latency injected into an app thread and delay app's work. I think deferring to app thread is not good idea.

JFR, This changes are inline to merge https://github.com/ROCm/rocm-systems/commit/4e2396f8a845203beb62f864dd917065a9013447.

manocharahul (commented):
Anusha's concern seems valid that we may delay actual work on main app thread even though this PR is using a batch flush of pending deferred cleanups.
Is it possible to somehow prevent async handler thread to get the last reference count of the HostQueue?

jaydeeppatel1111:
> Anusha's concern seems valid that we may delay actual work on main app thread even though this PR is using a batch flush of pending deferred cleanups. Is it possible to somehow prevent async handler thread to get the last reference count of the HostQueue?

>> Is it possible to somehow prevent async handler thread to get the last reference count of the HostQueue?
 -- @manocharahul , This change mirrors same deferring approach as https://github.com/ROCm/rocm-systems/pull/7490.
I think, Destruction is not serial and hence we have ref counted. Can you elaborate more if you have any idea?
And checkout alternatives https://github.com/ROCm/rocm-systems/pull/8206#issuecomment-4914423750.

manocharahul:
@jaydeeppatel1111 there is a major difference from the original change #7490 
#7490 only defers queue destroy(which is much more deterministic ops) whereas deferring entire HostQueue terminate flushes markers and awaits completion, delaying main app thread work. 
Can we try to stress test this scenario and see if it causes any other issues?

jaydeeppatel1111:
> @jaydeeppatel1111 there is a major difference from the original change #7490 #7490 only defers queue destroy(which is much more deterministic ops) whereas deferring entire HostQueue terminate flushes markers and awaits completion, delaying main app thread work. Can we try to stress test this scenario and see if it causes any other issues?

Yeah, I think, Deferring anything in general should have impact. Exploring alternative that does not require deferring. https://github.com/ROCm/rocm-systems/pull/8254

jaydeeppatel1111:
Fix is already merged.

PR #8208, opened 2026-07-07
Enable Host vmem tests on CI

pghoshamd:
## Submission Checklist


projects/rocr-runtime/rocrtst/suites/functional/virtual_memory.cc
```diff
@@ -1736,6 +1736,20 @@ void VirtMemoryTestBasic::TestVirtAddressAlias(void) {
...
+    }
+  }
+
+  if (verbosity() > 0) {
+    std::cout << "    Host Memory VA alias test PASSED" << std::endl;
+  }
+
+  // Run on GPU pools
   for (unsigned int i = 0; i < gpus.size(); ++i) {
     hsa_amd_memory_pool_t gpu_pool;
```
cfreeamd, line 1754: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** The `"    Host Memory VA alias test PASSED"` message fires unconditionally after the CPU pool loop even when every iteration returned early (e.g., no CPU agent has `HSA_AMD_AGENT_INFO_HOST_ALLOC_DMABUF_SUPPORTED = true`). Technically the test did pass (by skipping), but printing PASSED after zero meaningful assertions can be misleading. Consider printing the message only when at least one pool was actually exercised, or rewording it to something like `"Host Memory VA alias test done"` that is accurate in both the exercised and fully-skipped cases.

cfreeamd (commented):
One file looks straightforward: `projects/rocr-runtime/rocrtst/suites/test_common/test_categories.yaml` — the three additions of `VirtMemory_GPUtoHostAccess_Test` and `VirtMemory_Interprocess_HostPool_Test` are consistent across all three CI categories and match the test names registered in `main.cc` by the parent commit.

cfreeamd (approved):

PR #8212, opened 2026-07-07
[rocprofiler-sdk] Bypass idle inline queue interposition without active consumers (#8093)

bwelton:
## Summary

Cherry-pick of 6168477 (#8093) from `develop` onto `release/therock-7.14`.

Fixes ROCM-27116 by bypassing queue interposition (QI) on the hot path when no active profiling context needs kernel-dispatch or scratch-memory tracing. This prevents idle QI from virtualizing write indices and processing doorbells when auto-registered clients only use code-object callbacks.

## Changes

- `queue_interposition.cpp` / `queue_interposition.hpp` — track active queue-interposition consumers with an atomic counter; extend `should_bypass_inline_intercept()` to passthrough when the count is zero
- `queue_controller.cpp` / `queue_controller.hpp` — add `context_needs_queue_interposition_tracing()` helper
- `context.cpp` — increment/decrement the counter on context start/stop and on `deactivate_client_contexts()`

Also bundled in this change (surfaced during review/testing of the above):
- Saturating CAS-loop decrement for the consumer counter (avoids underflow on unbalanced stop notifications)
- Resync of `virtual_wptr`/`next_scan_pos`/`next_submit_pos` to the real write-dispatch-id when the consumer count transitions 0→1, so QI doesn't publish against stale bookkeeping after bypass doorbells advanced the real queue
- Ordering fix for context start/stop notifications relative to active-context publish, and CAS-based slot claiming in `deactivate_client_contexts()`, to close a consumer-count race
- Replace the persistent `thread_local` doorbell snapshot buffer with a fixed-capacity inline snapshot (falling back to a local vector for rare large batches), removing a heap buffer implicated in the ROCM-27116 crash

## Motivation

PyTorch auto-registration installs QI by default even when only code-object tracing is registered. In a fork+exec child that inherits `ROCPROFILER_REGISTER_LIBRARY`, idle QI still intercepts HSA queue write-index and doorbell operations during minimal GPU work, corrupting the heap before exit (`corrupted double-linked list` / SIGABRT). Setting `ROCPROFILER_QUEUE_INTERPOSITION=0` avoids the crash; this change makes that behavior automatic when no QI consumer is active.

## Testing

Cherry-picked cleanly from `develop` (auto-merged, no conflicts). Original testing (from #8093):
- Built `librocprofiler-sdk.so` on banff (MI325X) and deployed into `rocm27116-parallel` PyTorch container
- PyTorch fork+exec repro (`torch.cuda.synchronize()` → `subprocess` child with `env=None` → `torch.rand`) 15/15 pass
- Stock image reproduces ~90%+ crash rate without the fix
- Full local build + `ctest` (gfx1200): 378/379 unit tests pass (1 pre-existing, unrelated failure: `unit.rocprofiler_lib.agent`); all queue-interposition/queue-controller/context/intercept-table tests and HSA multiqueue integration tests pass

## Related Issues

JIRA ID : ROCM-27116

Original PR: #8093
Original commit: 6168477bc5d2f5f1a0ca533f6deef136c44327d1

### DIFF
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/context/context.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/context/context.cpp
@@ -28,6 +28,7 @@
 #include "lib/common/utility.hpp"
 #include "lib/rocprofiler-sdk/buffer.hpp"
 #include "lib/rocprofiler-sdk/counters/core.hpp"
+#include "lib/rocprofiler-sdk/hsa/queue_interposition.hpp"
 #include "lib/rocprofiler-sdk/pc_sampling/service.hpp"
 #include "lib/rocprofiler-sdk/thread_trace/core.hpp"
 
@@ -325,6 +326,8 @@ start_context(rocprofiler_context_id_t context_id)
         get_num_active_contexts().fetch_add(1, std::memory_order_release);
     }
 
+    rocprofiler::hsa::queue_interposition::notify_queue_interposition_consumer_context_started(cfg);
+
     // atomic swap the pointer into the "active" array used internally
     const context* _expected = nullptr;
     bool           success   = active_contexts.at(idx).compare_exchange_strong(
@@ -332,6 +335,8 @@ start_context(rocprofiler_context_id_t context_id)
 
     if(!success)
     {
+        rocprofiler::hsa::queue_interposition::notify_queue_interposition_consumer_context_stopped(
+            cfg);
         get_num_active_contexts().fetch_sub(1, std::memory_order_release);
         return ROCPROFILER_STATUS_ERROR_CONTEXT_NOT_STARTED;
     }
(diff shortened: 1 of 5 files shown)


marbre (approved):
I asked my agent to help comparing as the base is not identical and it went a bit further not just comparing the cherry-pick with the original commit but analyzed the logic. This is not blocking as the cherry-picked commit was approved but it might at least need a follow up to main @bwelton. Leaving the review here:

```
  Bug 1 — relaxed load on the bypass check (ARM/weak-memory)

  has_active_queue_interposition_consumers() uses memory_order_relaxed, but the counter is incremented with memory_order_release. On ARM (and any non-TSO architecture), there's no guarantee a thread doing the
  relaxed load will observe the release store in bounded time. A thread could see count == 0 after another thread has started a context, bypass QI, and miss packet dispatch data. Should be
  memory_order_acquire.

  ---
  Bug 2 — Window between resync and fetch_add in notify_started

  // prev == 0, so resync shadow state to current real queue position
  resync_all_queue_shadow_states();

  // counter still 0 here — QI still bypassed
  s_active_queue_interposition_consumers.fetch_add(1, std::memory_order_release);
  // QI now active, but real_wdid may have advanced since resync

  Between the resync and the counter increment, HSA operations still bypass QI and advance the real queue pointer. When QI becomes active, the shadow state from the resync is already stale. This is the exact
  problem resync was supposed to prevent, but the ordering inverts the fix. The window is small but real.
```

astyrrian1 (approved):

jrmadsen (commented):
Approved

PR #8213, opened 2026-07-07
fix(hotswap): disable entry trampolines by default 

xintin:
## Motivation

gfx12.5 entry-trampoline HotSwap rewriting was enabled by default, which meant the gfx1250 B0->A0 retarget (and the generic gfx12.5 rewrite path) went through the entry-trampoline path automatically. Entry trampolines are still being validated, so they should be opt-in rather than default-on to avoid changing behavior for existing gfx12.5 users.

## Technical Details

- Collapsed the entry-trampoline controls into a single `RewriteOptions::entry_trampolines_enabled` flag, defaulting to **off** (`kDefaultEntryTrampolinesEnabled = false`).
- A single env reader `AreEntryTrampolinesRequested()` drives it from `AMD_COMGR_HOTSWAP_ENTRY_TRAMPOLINES`: unset/false-like -> disabled, truthy (`1`/`true`/`on`/…) -> enabled.
- `DecideHotswapRewrite`: the gfx1250 B0->A0 retarget now requests entry trampolines only when opted in (otherwise it still runs the legacy rewrite path); the generic gfx12.5 path is gated on the same flag.
- Updated `env_variables.rst` (default `1`->`0`, opt-in wording) and the `rocrtst` hotswap unit tests to the default-off semantics.
- Note: overall HotSwap remains enabled by default (`HSA_HOTSWAP_DISABLE` unchanged); only entry trampolines become opt-in.

## Test Plan
- Built and ran the `hotswap_rewrite` gtest suite (`rocrtst/suites/functional/hotswap`) with `ROCR_HOTSWAP_TESTING`, against a hotswap-enabled COMGR.
- Verified default-off behavior (B0->A0 legacy path; generic gfx12.5 not rewritten by default) and opt-in behavior via `AMD_COMGR_HOTSWAP_ENTRY_TRAMPOLINES=1`.

## Test Result

- `hotswap_rewrite`: **21/21 tests passed**.

## Submission Checklist


dayatsin-amd (approved):

shwetagkhatri (approved):
LGTM

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -293,14 +297,15 @@ std::optional<RewriteDecision> DecideHotswapRewrite(
...
       target_gfx == kGfx1250) {
-    return RewriteDecision{
-        WithGfx1250SteppingFeature(source_isa, Gfx1250Stepping::kB0),
-        WithGfx1250SteppingFeature(target_isa, Gfx1250Stepping::kA0),
-        false};
+    // B0->A0 retarget defaults to the legacy (non entry-trampoline) rewrite
+    // path and only requests entry trampolines when explicitly opted in.
+    return RewriteDecision{WithGfx1250SteppingFeature(source_isa, Gfx1250Stepping::kB0),
+                           WithGfx1250SteppingFeature(target_isa, Gfx1250Stepping::kA0),
+                           options.entry_trampolines_enabled};
```
cfreeamd, line 304: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** Previously the B0→A0 retarget hardcoded `request_entry_trampolines = false`; now it propagates `options.entry_trampolines_enabled`. This means `AMD_COMGR_HOTSWAP_ENTRY_TRAMPOLINES=1` will request entry trampolines for B0→A0 as well. The new test `A0RetargetsUsesLegacyPathUnlessOptedIn` covers this case, so it appears intentional — does COMGR's B0→A0 path handle `request_entry_trampolines=true` correctly, or does it silently ignore the flag?

cfreeamd (commented):

Files that look correct: `hotswap.hpp` (rename + named constant are clean), `env_variables.rst` (accurate opt-in semantics with full truthy/falsy list), `hotswap.cpp` (`AreEntryTrampolinesRequested` using `IsEnvFlagEnabled` is consistent with other env-var parsing in the runtime), `hotswap_rewrite_test.cc` (test restructuring correctly reflects opt-in semantics; `NewComgrHotswapApiAvailable` guard correctly removed from default-OFF path), `gfx1250_min.s` (kernel descriptor + metadata make this a valid AMDHSA code object).

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
xintin: it's handled, not silently ignored. In COMGR, RunB0A0Patches and RunEntryTrampolines are independent options (comgr-hotswap.cpp), and retargetCodeObject (comgr-hotswap-b0a0.cpp) runs both passes: the B0->A0 stepping patches populate the deferred-trampoline growth, then appendKernelEntryTrampolines appends kernel-entry stubs to the same growth list and both are emitted together via one growWithTrampolines. 
The entry-trampoline pass walks the object's kernel descriptors (not gated to generic targets) and is idempotent, so with AMD_COMGR_HOTSWAP_ENTRY_TRAMPOLINES=1 a B0->A0 object gets both the stepping patches and entry-stub redirection. It's intentional that the flag now applies uniformly to B0->A0 and the generic gfx12.5 path.

cfreeamd (approved):

rahulc-gh:
overding the resolve conversation gate, for urgency

PR #8214, opened 2026-07-07
feat(ci): Enable THEROCK_ENABLE_ALL for gfx125X (MI455) builds

geomin12:
For MI455/gfx125X builds, always use -DTHEROCK_ENABLE_ALL=ON instead of the normal cmake_options. This ensures we build all projects for MI455 bringup rather than just core components.

Other architectures (gfx94X, gfx950, etc.) continue to use the cmake_options determined by therock_configure_ci.py based on which subtrees changed.


geomin12:
not needed

PR #8215, opened 2026-07-07
feat(ci): using `ci-env` for TheRock hash updating

geomin12:
Similar to rocm-libraries, we are now going to use ci-env action to obtain TheRock hash. 

this makes it easier to update and test hashes and don't need to update ~ 10 separate files


geomin12:
closing as #7887 handles this

PR #8219, opened 2026-07-07
fix(rocprofiler-sdk): pc sampling service check

meserve-amd:
- Fixes implementation of is_pc_sample_service_configured
  - Now matches what is required for get_pcs_session_of()
- Fixes an issue in attach where combining kernel trace and pc sampling crashed
- WriteInterceptor now only adds the pc_sampling marker packet if its callback would succeed

## Motivation

Fixes a bug where combining pc sampling and kernel trace options in attach would crash.

## Technical Details

The bug is ultimately in WriteInterceptor in queue.cpp.

Without --kernel-trace, get_active_contexts (line 307) returns false and the write interceptor returns early, so the bug is never reached.

With --kernel-trace, the tool registers a kernel dispatch buffer/callback tracer, so get_active_contexts returns true, and the write interceptor proceeds into the packet transformation loop. Line 629 is reached where is_pc_sample_service_configured() returns true causing a marker callback to be registered. However, if pc_sampling_service_finish_configuration has not run yet, get_pcs_session_of() will return nullptr and the registered callback will ultimately assert (hsa_adapter.cpp:101).

This change blocks this by simply checking if get_pcs_session_of() will succeed as part of is_pc_sample_service_configured()

## JIRA ID

JIRA ID: ROCM-26595

## Test Plan

Existing CI

Will use reproducer from JIRA issue to verify

## Test Result

Pending CI

Pending running reproducer

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/pc_sampling/service.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/pc_sampling/service.cpp
@@ -234,8 +234,11 @@ is_pc_sample_service_configured(rocprofiler_agent_id_t agent_id)
     // Require HSA-level init to be complete so that get_pcs_session_of() will succeed
     // and the ROCr session exists for every agent this function returns true for.
     if(!is_hsa_initialized().load()) return false;
-    return get_global_pc_sampling_sessions().rlock(
-        [agent_id](const auto& sessions) { return sessions.find(agent_id) != sessions.end(); });
+    return get_global_pc_sampling_sessions().rlock([agent_id](const auto& sessions) {
+        // If the agent_id is in the global sessions map,
+        // then the PC sampling service is configured on this agent.
+        return sessions.find(agent_id) != sessions.end();
+    });
 }
 
 PCSAgentSession*
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/pc_sampling/service.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/pc_sampling/service.cpp
@@ -231,12 +231,11 @@ configure_pc_sampling_service(context::context*                ctx,
 bool
 is_pc_sample_service_configured(rocprofiler_agent_id_t agent_id)
 {
-    return get_global_pc_sampling_sessions().rlock([agent_id](const auto& sessions) {
-        // Require hsa_agent to be set so that get_pcs_session_of() (which matches by
-        // hsa_agent handle) will succeed whenever this function returns true.
-        auto it = sessions.find(agent_id);
-        return it != sessions.end() && it->second->hsa_agent.has_value();
-    });
+    // Require HSA-level init to be complete so that get_pcs_session_of() will succeed
+    // and the ROCr session exists for every agent this function returns true for.
+    if(!is_hsa_initialized().load()) return false;
+    return get_global_pc_sampling_sessions().rlock(
+        [agent_id](const auto& sessions) { return sessions.find(agent_id) != sessions.end(); });
 }
 
 PCSAgentSession*
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/pc_sampling/service.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/pc_sampling/service.cpp
@@ -1,6 +1,6 @@
 // MIT License
 //
-// Copyright (c) 2023-2025 ROCm Developer Tools
+// Copyright (c) 2023-2026 ROCm Developer Tools
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -232,9 +232,10 @@ bool
 is_pc_sample_service_configured(rocprofiler_agent_id_t agent_id)
 {
     return get_global_pc_sampling_sessions().rlock([agent_id](const auto& sessions) {
-        // If the agent_id is in the global sessions map,
-        // then the PC sampling service is configured on this agent.
-        return sessions.find(agent_id) != sessions.end();
+        // Require hsa_agent to be set so that get_pcs_session_of() (which matches by
+        // hsa_agent handle) will succeed whenever this function returns true.
+        auto it = sessions.find(agent_id);
+        return it != sessions.end() && it->second->hsa_agent.has_value();
     });
 }
 


meserve-amd:
>     * ❌ Unit Test

Fixes a potential race failure in existing tests. No new tests are added. Ignoring check and removing tag.

vedithal-amd (approved):
It works
 
I built SDK using your PR
 
Ran the workload LD_LIBRARY_PATH=/app/projects/rocprofiler-sdk/build/lib:/rocm/lib:$LD_LIBRARY_PATH ROCP_TOOL_ATTACH=1 ./tests/hip_dynamic_shared
 
Did the profile
 
/app/projects/rocprofiler-sdk/build/bin/rocprofv3 --attach-duration-msec 1000 --pc-sampling-beta-enabled --pc-sampling-method stochastic --pc-sampling-unit cycles --pc-sampling-interval 1048576 --output-format json -d out_b -o ps_file --kernel-trace --pid 207797
 
And it created out_b/ps_file_results.json

vlaindic (dismissed):
Thanks for the fix — the crash is real and the intent is right, but I think the guard has a residual race that makes it still unsafe in attach mode. See the inline comment for details and a proposed alternative.

vlaindic (changes requested):
Thanks for the fix — the crash is real and the intent is right, but I think the guard has a residual race that makes it still unsafe in attach mode. See the inline comment for details and a proposed alternative.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/pc_sampling/service.cpp
```diff
@@ -232,9 +232,10 @@ bool
...
 {
     return get_global_pc_sampling_sessions().rlock([agent_id](const auto& sessions) {
-        // If the agent_id is in the global sessions map,
-        // then the PC sampling service is configured on this agent.
-        return sessions.find(agent_id) != sessions.end();
+        // Require hsa_agent to be set so that get_pcs_session_of() (which matches by
+        // hsa_agent handle) will succeed whenever this function returns true.
+        auto it = sessions.find(agent_id);
+        return it != sessions.end() && it->second->hsa_agent.has_value();
     });
```
vlaindic, line 241: **Residual race with `hsa_agent.has_value()`**

The crash is caused by the attach / late-start ordering:

1. `rocprofiler_force_configure` → `tool_init` runs → `configure_pc_sampling_on_all_agents` completes → agent session is in the global map but `hsa_agent = nullopt` (IOCTL phase only).
2. Kernel-trace context is already active → WriteInterceptor is live and processing dispatches.
3. `invoke_register_propagation` hasn't run yet → `finish_configuration` hasn't run → `hsa_agent` still `nullopt`.
4. A dispatch hits the WriteInterceptor → old `is_pc_sample_service_configured` returns `true` → `get_pcs_session_of` returns `nullptr` → assert crash.

The `hsa_agent.has_value()` guard closes that specific window, but there is still a narrower race **inside `finish_configuration` itself**:

```cpp
// For each agent_session in the loop:
agent_session->hsa_agent = get_hsa_agent(...);   // (A) guard now passes
hsa_ven_amd_pcs_create_from_id_fn(...);          // (B) ROCr session created — not yet!
parser->register_buffer_for_agent(...);
```

Between (A) and (B), `hsa_agent.has_value()` is `true` but the ROCr-level session does not exist yet. If the WriteInterceptor fires in this window the marker packet gets injected for a session ROCr doesn't know about. With multiple agents the problem is worse: once agent 0 passes (A), the guard returns `true` for agent 0 while agents 1..N are still at `nullopt`.

**Proposed fix**

Gate on `is_hsa_initialized()` instead — it is set to `true` only **after** the entire agent loop in `finish_configuration` completes, meaning all agents have `hsa_agent` set AND their ROCr sessions created:

```cpp
bool
is_pc_sample_service_configured(rocprofiler_agent_id_t agent_id)
{
    // Require HSA-level initialization to be complete (i.e. finish_configuration has run
    // for all agents) so that get_pcs_session_of() will succeed and the ROCr session
    // actually exists for every agent this function returns true for.
    if(!is_hsa_initialized().load()) return false;
    // If the agent_id is in the global sessions map,
    // then the PC sampling service is configured on this agent.
    return get_global_pc_sampling_sessions().rlock([agent_id](const auto& sessions) {
        return sessions.find(agent_id) != sessions.end();
    });
}
```

`is_hsa_initialized` is the only flag that precisely means "PC sampling is fully configured at the ROCr level for all agents" and it is already set in `post_hsa_init_start_active_service` right after the `finish_configuration` loop completes.

vlaindic (changes requested):
Thanks for catching this issue. However, I believe there might be another subtle race that we might not resolve, so I propose an inline comment that hopefully closes the problem. Please provide me your opinion.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/pc_sampling/service.cpp
```diff
@@ -232,9 +232,10 @@ bool
 is_pc_sample_service_configured(rocprofiler_agent_id_t agent_id)
 {
     return get_global_pc_sampling_sessions().rlock([agent_id](const auto& sessions) {
-        // If the agent_id is in the global sessions map,
-        // then the PC sampling service is configured on this agent.
-        return sessions.find(agent_id) != sessions.end();
+        // Require hsa_agent to be set so that get_pcs_session_of() (which matches by
+        // hsa_agent handle) will succeed whenever this function returns true.
+        auto it = sessions.find(agent_id);
+        return it != sessions.end() && it->second->hsa_agent.has_value();
```
vlaindic: **Residual race with `hsa_agent.has_value()`**

The crash is caused by the attach / late-start ordering:

1. `rocprofiler_force_configure` → `tool_init` runs → `configure_pc_sampling_on_all_agents` completes → agent session is in the global map but `hsa_agent = nullopt` (IOCTL phase only).
2. Kernel-trace context is already active → WriteInterceptor is live and processing dispatches.
3. `invoke_register_propagation` hasn't run yet → `finish_configuration` hasn't run → `hsa_agent` still `nullopt`.
4. A dispatch hits the WriteInterceptor → old `is_pc_sample_service_configured` returns `true` → `get_pcs_session_of` returns `nullptr` → assert crash.

The `hsa_agent.has_value()` guard closes that specific window, but there is still a narrower race **inside `finish_configuration` itself**:

```cpp
// For each agent_session in the loop:
agent_session->hsa_agent = get_hsa_agent(...);   // (A) guard now passes
hsa_ven_amd_pcs_create_from_id_fn(...);          // (B) ROCr session created — not yet!
parser->register_buffer_for_agent(...);
```

Between (A) and (B), `hsa_agent.has_value()` is `true` but the ROCr-level session does not exist yet. If the WriteInterceptor fires in this window the marker packet gets injected for a session ROCr doesn't know about. With multiple agents the problem is worse: once agent 0 passes (A), the guard returns `true` for agent 0 while agents 1..N are still at `nullopt`.

**Proposed fix**

Gate on `is_hsa_initialized()` instead — it is set to `true` only **after** the entire agent loop in `finish_configuration` completes, meaning all agents have `hsa_agent` set AND their ROCr sessions created:

```cpp
bool
is_pc_sample_service_configured(rocprofiler_agent_id_t agent_id)
{
    // Require HSA-level initialization to be complete (i.e. finish_configuration has run
    // for all agents) so that get_pcs_session_of() will succeed and the ROCr session
    // actually exists for every agent this function returns true for.
    if(!is_hsa_initialized().load()) return false;
    // If the agent_id is in the global sessions map,
    // then the PC sampling service is configured on this agent.
    return get_global_pc_sampling_sessions().rlock([agent_id](const auto& sessions) {
        return sessions.find(agent_id) != sessions.end();
    });
}
```

`is_hsa_initialized` is the only flag that precisely means "PC sampling is fully configured at the ROCr level for all agents" and it is already set in `post_hsa_init_start_active_service` right after the `finish_configuration` loop completes.

vlaindic (changes requested):
Pls check the inline comment.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/pc_sampling/service.cpp
meserve-amd: Resolving this comment as it's duplicated.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/pc_sampling/service.cpp
meserve-amd: Yes, this is correct, I updated the PR to use `is_hsa_initialized` instead

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/pc_sampling/service.cpp
```diff
@@ -231,11 +231,11 @@ configure_pc_sampling_service(context::context*                ctx,
...
 {
-    return get_global_pc_sampling_sessions().rlock([agent_id](const auto& sessions) {
-        // If the agent_id is in the global sessions map,
-        // then the PC sampling service is configured on this agent.
-        return sessions.find(agent_id) != sessions.end();
-    });
+    // Require HSA-level init to be complete so that get_pcs_session_of() will succeed
+    // and the ROCr session exists for every agent this function returns true for.
+    if(!is_hsa_initialized().load()) return false;
+    return get_global_pc_sampling_sessions().rlock(
```
vlaindic: I believe keeping the original comment would be useful for better understanding.

```suggestion   
     // If the agent_id is in the global sessions map,
     // then the PC sampling service is configured on this agent.
    return get_global_pc_sampling_sessions().rlock(
```

vlaindic (approved):
Thanks for accepting my comment. 

I'm approving but if you could, please accept the original comment that describes the original intent of the function. Should add some more clarity. Thanks!

PR #8220, opened 2026-07-07
[rocprof-compute] Adjustments to profiler interface architecture plan

svolkov-amd:
## Submission Checklist


projects/rocprofiler-compute/docs/design/profile_interface_architecture.md
```diff
@@ -456,10 +344,10 @@ flowchart LR
 
 ## Phase D: Analyze-Data Boundary
```
svolkov-amd, line 345: I believe this section is an overstretch and requires a separate design discussion. Part of which should be answering on the following questions:
- What format outputs do we want to support?
- What are the guidelines for each output format in terms of scope of data shown and data presentation?
- What components in the architecture should be data produces and which are data consumers. For example:
    - Should CSV output first generate analysis db and then convert it to actual CSV files? Benefit is that CSV and db will be consistent. Disadvantage is that it's performance cost implication.
    - Should metrics evaluation module generate output on the disk which then roofline consumes or should roofline invoke metrics evaluation module to produce required metrics calculations in memory?

And many others.

jamessiddeley-amd (approved):

PR #8221, opened 2026-07-07
refactor(kfdtest): [AILIKFD-40] modernize CMake dependency discovery

andmar-amd:
Modernize kfdtest CMakeLists.txt to follow rocr-debug-agent pattern (commit 3a3fb32060).

Changes:
- Use find_package(CONFIG) first for TheRock builds
- Fall back to manual find_library() for standalone builds
- Create imported targets (hsakmt::hsakmt, NUMA::NUMA) for clean dependency management
- Support multiple discovery methods:
  1. LIBHSAKMT_PATH CMake variable (TheRock)
  2. LIBHSAKMT_PATH environment variable
  3. pkg-config
  4. Standard system paths (/usr/lib/x86_64-linux-gnu, /opt/rocm/lib, etc.)
- Remove link_directories() in favor of target-based linking
- Add explicit dl library to target_link_libraries

Tested in both scenarios:
- TheRock build with LIBHSAKMT_PATH=/path/to/build
- Standalone build with system-installed libraries

Co-Authored-By: Claude Sonnet 4

## JIRA ID
JIRA ID: AILIKFD-40

## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/libhsakmt/tests/kfdtest/CMakeLists.txt
+++ b/projects/rocr-runtime/libhsakmt/tests/kfdtest/CMakeLists.txt
@@ -188,16 +188,8 @@ if(NOT hsakmt_FOUND)
     endif()
 endif()
 
-find_path( LIGHTNING_CMAKE_DIR NAMES LLVMConfig.cmake
-    PATHS $ENV{OUT_DIR}/llvm/lib/cmake/llvm NO_CACHE NO_DEFAULT_PATH)
-
-if ( DEFINED LIGHTNING_CMAKE_DIR AND EXISTS ${LIGHTNING_CMAKE_DIR} )
-    set ( LLVM_DIR ${LIGHTNING_CMAKE_DIR} )
-else()
-    message( STATUS "Couldn't find Lightning build in compute directory. "
-        "Searching LLVM_DIR then defaulting to system LLVM install if still not found..." )
-endif()
-
+# Find LLVM library
+# Uses LLVM_DIR (CMake/env variable) or CMAKE_PREFIX_PATH if set, otherwise system paths
 find_package( LLVM REQUIRED CONFIG )
 
 if( ${LLVM_PACKAGE_VERSION} VERSION_LESS "7.0" )
--- a/projects/rocr-runtime/libhsakmt/tests/kfdtest/CMakeLists.txt
+++ b/projects/rocr-runtime/libhsakmt/tests/kfdtest/CMakeLists.txt
@@ -24,7 +24,7 @@
 # If environment variable DRM_DIR or LIBHSAKMT_PATH is set, the script
 # will pick up the corresponding libraries from those pathes.
 
-cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.25 FATAL_ERROR)
 
 project(KFDTest)
 
@@ -93,9 +93,8 @@ set (CPACK_RPM_PACKAGE_REQUIRES "rocm-core")
 find_package(PkgConfig)
 
 # Prepend DRM_DIR to CMAKE_PREFIX_PATH if set
-# Note: list(PREPEND) was added in CMake 3.15, using INSERT for 3.12 compatibility
 if(DRM_DIR)
-    list(INSERT CMAKE_PREFIX_PATH 0 "${DRM_DIR}")
+    list(PREPEND CMAKE_PREFIX_PATH "${DRM_DIR}")
 endif()
 # The module name passed to pkg_check_modules() is determined by the
 # name of file *.pc
--- a/projects/rocr-runtime/libhsakmt/tests/kfdtest/CMakeLists.txt
+++ b/projects/rocr-runtime/libhsakmt/tests/kfdtest/CMakeLists.txt
@@ -24,7 +24,7 @@
 # If environment variable DRM_DIR or LIBHSAKMT_PATH is set, the script
 # will pick up the corresponding libraries from those pathes.
 
-cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
 
 project(KFDTest)
 
@@ -103,9 +103,7 @@ include_directories(${DRM_AMDGPU_INCLUDE_DIRS})
 # This allows NUMA_ROOT, hsakmt_ROOT, etc. to guide package discovery in both
 # TheRock (via CMake cache variables) and standalone builds (via environment).
 # Without this policy, find_package() ignores _ROOT hints, breaking custom paths.
-if ( POLICY CMP0074 )
-    cmake_policy( SET CMP0074 NEW )
-endif()
+cmake_policy( SET CMP0074 NEW )
 
 # Find NUMA library
 # Try CONFIG mode first (TheRock), fall back to manual discovery (standalone)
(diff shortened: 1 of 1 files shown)


cfreeamd (commented):

The CMake refactor in `libhsakmt/tests/kfdtest/CMakeLists.txt` looks correct. Key changes are sound:

- `CMP0074` is now set before the `find_package()` calls (correct ordering — the original placed it after the search block).
- All `find_library()` calls use `NAMES libhsakmt.a` only, correctly enforcing static linkage.
- `link_directories()` is removed in favor of imported targets (good CMake hygiene).
- `${CMAKE_DL_LIBS}` replaces the hardcoded `dl` dependency (more portable).
- `PRIVATE` keyword added to `target_link_libraries` (correct scoping).

Files in the ROCr perimeter that look correct: `projects/rocr-runtime/libhsakmt/tests/kfdtest/CMakeLists.txt`

cfreeamd (approved):

projects/rocr-runtime/libhsakmt/tests/kfdtest/CMakeLists.txt
```diff
@@ -99,32 +99,92 @@ pkg_check_modules(DRM REQUIRED libdrm)
 pkg_check_modules(DRM_AMDGPU REQUIRED libdrm_amdgpu)
 include_directories(${DRM_AMDGPU_INCLUDE_DIRS})
 
-if( DEFINED ENV{LIBHSAKMT_PATH} )
-    set ( LIBHSAKMT_PATH $ENV{LIBHSAKMT_PATH} )
-    message ( "LIBHSAKMT_PATH environment variable is set" )
+# Set CMP0074 before find_package() calls to honor _ROOT variables
```
lumachad: We might need a bit more explanation for this.

projects/rocr-runtime/libhsakmt/tests/kfdtest/CMakeLists.txt
lumachad, line 49: Just a heads up all of this CPACK code is likely dead. It isn't used because we don't package components individually anymore since moving to TheRock. This should be cleaned up by the component owners.

Also, ROCM_VERSION_FOR_PACKAGE is also dead. We should get the version from TheRock or git.

lumachad: The minimum version should be updated by the component owners. We've received the following warning recently:

```
CMake Deprecation Warning at CMakeLists.txt:63 (cmake_minimum_required):
Compatibility with CMake < 3.10 will be removed from a future version of
```

So clearly 3.5 is way too old.

projects/rocr-runtime/libhsakmt/tests/kfdtest/CMakeLists.txt
```diff
@@ -256,7 +311,18 @@ if( ${HSAKMT_CTX} )
     endif()
 endif()
 
-target_link_libraries(kfdtest ${HSAKMT_LIBRARIES} ${DRM_LDFLAGS} ${DRM_AMDGPU_LDFLAGS} ${llvm_libs} pthread m stdc++ rt numa)
+target_link_libraries(kfdtest
+    PRIVATE
+        hsakmt::hsakmt
+        NUMA::NUMA
+        ${DRM_LDFLAGS}
+        ${DRM_AMDGPU_LDFLAGS}
```
lumachad, line 315: I don't see this defined anywhere in the file. Is it still used? How is this passed? By the user? By the upper layers? Does it still make sense to have in the context of TheRock?

projects/rocr-runtime/libhsakmt/tests/kfdtest/CMakeLists.txt
```diff
@@ -256,7 +311,18 @@ if( ${HSAKMT_CTX} )
...
+        ${llvm_libs}
```
lumachad, line 316: Do we really need llvm_libs here? Don't we find those automatically in TheRock? I suppose we would also find those easily if we're looking for llvm libs from the system or ROCm stack.

lumachad (commented):
Left some comments. I'm assuming this has been tested standalone, TheRock with bundled sysdeps and TheRock without bundled sysdeps.

Standalone and TheRock with unbundled sysdeps needs to be tested manually.

As a general comment, it is really up to the component owners how far they want to cleanup the old code from the build system.

projects/rocr-runtime/libhsakmt/tests/kfdtest/CMakeLists.txt
andmar-amd: Greetings

This is just a warning and since rocr has to maintain it own set of compatible builds, it may be ok to ignore this.  But I am not able to make decision on this so relegating to @dayatsin-amd ?

projects/rocr-runtime/libhsakmt/tests/kfdtest/CMakeLists.txt
andmar-amd: I am not 100% sold on the last question here,  because even though TheRock gates this PR. this build has still work  on his own.  I (and Claude AI) will looking it though!

projects/rocr-runtime/libhsakmt/tests/kfdtest/CMakeLists.txt
andmar-amd: CPACK is actively used in standalone builds.  So let us keep this for now, unless someone more senior says otherwise.

projects/rocr-runtime/libhsakmt/tests/kfdtest/CMakeLists.txt
andmar-amd: - Line 98: pkg_check_modules(DRM REQUIRED libdrm) automatically creates DRM_LDFLAGS                                                                                     
    - Line 99: pkg_check_modules(DRM_AMDGPU REQUIRED libdrm_amdgpu) creates DRM_AMDGPU_LDFLAGS                                                                              
    - This is standard CMake pkg-config behavior                                                                                                                            
    - Still needed for standalone builds

projects/rocr-runtime/libhsakmt/tests/kfdtest/CMakeLists.txt
andmar-amd: - Still needed for standalone builds that use system/ROCm LLVM                                                                                                          
    - TheRock might provide this automatically, but standalone builds need it

projects/rocr-runtime/libhsakmt/tests/kfdtest/CMakeLists.txt
andmar-amd: upgraded to 3.12, but waiting from confirmation if we can go to 3.7?

lumachad: For context, TheRock requires 3.25.

andmar-amd: we going to stay at 3.12 for now!

projects/rocr-runtime/libhsakmt/tests/kfdtest/CMakeLists.txt
andmar-amd: I had tested we can still build standalone with just LLVM_DIR or  CMAKE_PREFIX_PATH set. So I Removed the Lightning-specific discovery code (lines 191-199)

kentrussell (approved):

PR #8225, opened 2026-07-07
refactor(rocm): unify AQL packet logging between debug log and hang a…

saleelk:
…nalysis

Introduce AqlLogSink enum with aqlFormat() and emitAqlLine() so every logAql* helper can emit to either ClPrint (debug log) or stderr (hang analysis) via a sink parameter, eliminating duplicate format strings.

Rewrite AnalyzeAqlQueue to delegate to the shared formatters instead of hand-rolling its own fprintf packet dumps. Hang analysis now prints the same detail as the debug log path, including the metadata packet blob with the full launch descriptor hex dump.

## JIRA ID

JIRA ID : AIRUNTIME-0

## Submission Checklist

### DIFF
--- a/projects/clr/rocclr/device/rocm/rocvirtual.cpp
+++ b/projects/clr/rocclr/device/rocm/rocvirtual.cpp
@@ -36,6 +36,7 @@
 #include <vector>
 #include <atomic>
 #include <cinttypes>
+#include <cstdarg>
 #include <mutex>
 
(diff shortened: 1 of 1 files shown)


saleelk:
> 
> 
> * ❌ Unit Test
> 

Not needed,

gandryey (approved):

PR #8226, opened 2026-07-07
Lazily parse dwarf information

ApoKalipse-V:
## Motivation

Avoid parsing dwarf information for code objects that were never entered.
This is a performance optimization.

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-sdk/source/include/rocprofiler-sdk/cxx/codeobj/code_printing.hpp
+++ b/projects/rocprofiler-sdk/source/include/rocprofiler-sdk/cxx/codeobj/code_printing.hpp
@@ -158,7 +158,7 @@ class CodeobjDecoderComponent
 
     void loadDebugLineInfo()
     {
-        if(!disassembly) throw std::exception();
+        if(!disassembly) throw std::runtime_error("No disassembly available!");
 
         ProtectedFd prot("");
         auto&       codeobj_data = disassembly->buffer;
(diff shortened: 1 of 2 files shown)


projects/rocprofiler-sdk/source/include/rocprofiler-sdk/cxx/codeobj/code_printing.hpp
```diff
@@ -374,6 +382,8 @@ class CodeobjDecoderComponent
         inst->faddr = faddr;
         inst->vaddr = vaddr;
 
+        std::call_once(m_debug_line_info_once, [this]() { loadDebugLineInfo(); });
```
Sushma1203: If loadDebugLineInfo() throws, call_once does not set the flag as done, so every subsequent call to disassemble_instruction() will retry and rethrow. Callers have no try/catch, so a persistent failure will std::terminate instead of gracefully skipping annotations.

// Instead of letting loadDebugLineInfo() throw through call_once,
// catch internally and record failure:
void loadDebugLineInfo()
{
    if(!disassembly) return;
    try {
        // ... existing DWARF parsing code ...
    } catch(...) {
        m_debug_line_info_failed = true;  // don't retry
    }
}

// In disassemble_instruction():
std::call_once(m_debug_line_info_once, [this]() { loadDebugLineInfo(); });

Sushma1203 (approved):

PR #8228, opened 2026-07-07
feat(rccl): CE Alltoallv for scaleup

nusislam:
## Motivation

This PR adds an AlltoAllv using SDMA engines. The Alltoallv use zero CUs and is enabled when the application uses symmetric memory window registration along with zero CTA policy.

## JIRA ID
AICOMRCCL-1481

## Test Plan

1. RCCL Test
2. Unit test (to be added)

## Test Result
Up to 5% improvement over CU-based alltoallv on MI350.

## Submission Checklist


nusislam:
%RCCL_TEST_CONFIG
mi300x_mellanox_ib.json‎

nusislam:
@copilot resolve the merge conflicts in this pull request

AbandiGa (approved):
LGTM

nusislam:
@nawrinsu - can you review this PR?

nusislam:
@copilot resolve the merge conflicts in this pull request

nusislam:
@copilot resolve the merge conflicts in this pull request

nusislam:
@copilot resolve the merge conflicts in this pull request

projects/rccl/src/collectives.cc
```diff
@@ -463,122 +409,172 @@ ncclResult_t ncclAlltoAllv_impl(const void* sendbuff, const size_t sendcounts[],
...
 #endif
 
-  Recorder::instance().skip(true);
-  NCCLCHECK(ncclGroupStart());
-  for (int r = 0; r < nRanks; r++) {
-    NCCLCHECK(ncclSend(((char*)sendbuff) + sdispls[r] * ncclTypeSize(datatype), sendcounts[r], datatype, r, comm,
-                       stream));
-    NCCLCHECK(ncclRecv(((char*)recvbuff) + rdispls[r] * ncclTypeSize(datatype), recvcounts[r], datatype, r, comm,
-                       stream));
+  if (comm->nNodes == 1 && (comm->config.CTAPolicy & NCCL_CTA_POLICY_ZERO)) {
```
nawrinsu: The guard only checks `nNodes==1 && CTAPolicy ZERO` but the CE path also requires symmetric support. If zero-CTA is set on non-symmetric buffers, we enqueue `ncclFuncAlltoAllv` with `count=0`. In `taskAppend`, it falls through to `collTaskAppend` instead of `ceCollTaskAppend`, which has no alltoallv kernel and ignores `info->sizes`. Please also gate on CE eligibility here.

nusislam: Good point. Fixed.

nawrinsu (approved):

nusislam:
@atulkulk - I assume the code coverage CI is still not fixed. Can you please confirm?

nusislam:
@copilot resolve the merge conflicts in this pull request

nusislam:
@sdarisala - can you review this PR? There is a timeout in the `precheckin` test but not related to this PR. Tagging you since you reviewed CE code before.

projects/rccl/src/collectives.cc
```diff
@@ -571,18 +578,51 @@ ncclResult_t ncclAlltoAllv_impl(const void* sendbuff, const size_t sendcounts[],
...
+    NCCLCHECK(ncclGroupStart());
+    for (int r = 0; r < nRanks; r++) {
+      void* recvPtr = (void*)((char*)comm->gatheredSizes + (size_t)r * nLocal * sizeof(size_t));
+      NCCLCHECK(ncclSend(comm->localSizes, nLocal, ncclUint64, r, comm, stream));
+      NCCLCHECK(ncclRecv(recvPtr, nLocal, ncclUint64, r, comm, stream));
+    }
+    NCCLCHECK(ncclGroupEnd());
+    CUDACHECK(cudaMemcpyAsync(gatheredSizes.data(), comm->gatheredSizes, nGather * sizeof(size_t),
+                              cudaMemcpyDeviceToHost, stream));
+    CUDACHECK(cudaStreamSynchronize(stream));
```
pvallem, line 623: This size exchange assumes the `ncclSend`/`ncclRecv` above have actually run by the time we copy `gatheredSizes` back to the host and `cudaStreamSynchronize`. But if the application calls `ncclAlltoAllv` inside its own `ncclGroupStart`/`ncclGroupEnd`, this internal `ncclGroupEnd` won't launch anything — a nested groupEnd just decrements the group depth, so the sends get deferred to the app's outer groupEnd. The stream sync then returns immediately and we read `gatheredSizes` before it's been filled: uninitialized on the first call, or the previous call's table afterwards. Those stale sizes/displacements drive the DMA copies, which can write outside the peer's window (silent corruption) or trip the `chunkBytes != peerRcvSize` check.

Could we make the size exchange independent of the caller's group depth — e.g. do it as a host-side exchange, or force the internal group to flush before the readback? That would also avoid the `cudaStreamSynchronize` being illegal under graph capture.

projects/rccl/test/CeAlltoAllvEligibilityTests.cpp
```diff
@@ -0,0 +1,158 @@
...
+                                 ncclFloat32,
+                                 ncclSymSendNonregRecvNonreg));
+    EXPECT_FALSE(ncclCeAvailable(mockComm_.get(),
+                                 ncclFuncAlltoAllv,
+                                 ncclDevSum,
+                                 ncclFloat32,
+                                 ncclSymSendRegRecvNonreg));
+}
+
+TEST_F(CeAlltoAllvEligibilityTest, LocalMetadataPackingMatchesGatheredLayout)
```
pvallem, line 159: These two tests (this one and `PeerMetadataIndexingMatchesCeCollLayout` below) only exercise the helper functions in `CeAlltoAllvTestHelpers.hpp` — they pack the matrix with the helpers and read it back with the same helpers, so they're effectively checking the helpers against themselves. The real layout arithmetic in `ce_coll.cc` (~715-717 and 752-753) and the host-side packing in `collectives.cc` (~527-530) is never called here, so if someone changed the production indexing (e.g. swapped recvSizes/recvDispls), these tests would still pass.

Since this layout is a two-sided contract, could we either factor the offset math into a single shared `inline` that both `ce_coll.cc` and the tests use, or drive the real `ncclCeAlltoAllv` path so that a drift actually fails a test?

projects/rccl/src/ce_coll.cc
```diff
@@ -703,6 +704,92 @@ ncclResult_t ncclCeAlltoAll(struct ncclComm* comm, struct ncclCeCollArgs* args,
...
+        batchOpsParams.sizes[batchOpsParams.numOps] = chunkBytes;
+        batchOpsParams.numOps++;
+      }
+    } else {
+      // Remote copy to other ranks: send to rank dstRank's receive buffer at position comm->rank
+      size_t* peerRecvSizes = args->sizes + 4 * comm->nRanks * dstRank + 2 * comm->nRanks;
+      size_t* peerRecvDispls = args->sizes + 4 * comm->nRanks * dstRank + 3 * comm->nRanks;
+      peerDispls = peerRecvDispls[comm->rank];
+      peerRcvSize = peerRecvSizes[comm->rank];
+      if (chunkBytes != peerRcvSize) {
```
pvallem: This size-consistency guard has no test coverage. The only test that calls `ncclCeAlltoAllv` (`CeInternalMPITests.cpp:621`) passes `sizes = nullptr` and returns at the early guard on line 710, so the whole data path below it — this `chunkBytes != peerRcvSize` check, the `winOff` peer-offset math, the batch launch, and the zero-chunk skip — is never exercised. This check in particular could be inverted or removed and every test would still pass.

Could we add a test that drives the real loop (e.g. with a mocked `ncclDevrGetLsaRankPtr`) or a small multi-rank correctness test? That would also let the layout tests above actually guard the production indexing.

projects/rccl/src/collectives.cc
```diff
@@ -571,18 +578,51 @@ ncclResult_t ncclAlltoAllv_impl(const void* sendbuff, const size_t sendcounts[],
     return ret;
   }
 #endif
+  struct ncclDevrWindow* sendWin = nullptr;
+  struct ncclDevrWindow* recvWin = nullptr;
+  NCCLCHECK(ncclDevrFindWindow(comm, sendbuff, &sendWin));
+  NCCLCHECK(ncclDevrFindWindow(comm, recvbuff, &recvWin));
+  ncclSymRegType_t winRegType;
+  NCCLCHECK(ncclGetSymRegType(sendWin, recvWin, &winRegType));
+  bool ceAlltoAllvEligible = (comm->config.CTAPolicy & NCCL_CTA_POLICY_ZERO) &&
```
pvallem: In `taskAppend` there's an extra condition on `hasSysmemSegment` ([R3775](https://github.com/ROCm/rocm-systems/pull/8228/changes#diff-d15a529829c95e3ce8d94113b0f348df5016198fbd127742e72493d6dc3fa244R3775)) which is not used here. If a symmetric window has a host-backed (sysmem) segment, this gate still passes and runs the full size-exchange, but `taskAppend` then skips the CE branches and falls through to `collTaskAppend` — and there is no AlltoAllv device kernel, so the kernel launch may fail. Please check.

projects/rccl/src/collectives.cc
pvallem, line 623: `cudaStreamSynchronize` is not allowed during HIP/CUDA graph capture, and this eligible path has no capture guard — unlike `taskAppend`, which sets `ceAvailable = !ceCapturing`. So capturing an AllToAllv with ZERO CTA + symmetric registration will fail here, even though the plain send/recv fallback is capture-safe. Could we also gate `ceAlltoAllvEligible` on `!capturing`? Please check.

projects/rccl/test/CeAlltoAllvEligibilityTests.cpp
```diff
@@ -0,0 +1,158 @@
...
+TEST_F(CeAlltoAllvEligibilityTest, CeImplementedReturnsTrueForAlltoAllvOnSupportedDriver)
+{
+    if (!isCeRuntimeDriverSupported())
+        GTEST_SKIP() << "CE driver not in supported range "
+                        "(need ROCm >= 7.12 or 7.0.2.x backport [70051831, 70060000))";
+
+    EXPECT_TRUE(ncclCeImplemented(ncclFuncAlltoAllv, ncclDevSum, ncclFloat32));
+}
+
+TEST_F(CeAlltoAllvEligibilityTest, CeAvailable_EligibleWithSymmetricSingleNode)
```
pvallem, line 47: None of these tests set `comm.config.CTAPolicy`, so the `(config.CTAPolicy & NCCL_CTA_POLICY_ZERO)` half of `ceAlltoAllvEligible` (collectives.cc:587) is never exercised — a regression that drops or inverts the zero-CTA condition wouldn't be caught. Could we add a case that toggles `CTAPolicy`? Please check.

projects/rccl/src/collectives.cc
nusislam: I have added a check on `ncclGroupDepth==0 && !capturing` to enable CE AlltoAllv. I had tried driving the size exchange with host buffers earlier. But that requires using `bootStrapAllGather()` and the performance of it is much worse than using `ncclSend()/ncclRecv()` with device buffers.

projects/rccl/src/collectives.cc
nusislam: @pvallem - addressed.

projects/rccl/src/collectives.cc
nusislam: @pvallem - addressed.

projects/rccl/test/CeAlltoAllvEligibilityTests.cpp
nusislam: addressed.

nusislam:
@pvallem - I have addressed your reviews. Can you please take a look at the PR one more time? This needs to go by 08/12 for the Meta drop.

nusislam:
code coverage CI failures are not related to this PR. Some tests for comm shrink/grow has timeout failure in the code coverage CI. cc: @pvallem

projects/rccl/src/ce_coll.cc
```diff
@@ -703,6 +714,96 @@ ncclResult_t ncclCeAlltoAll(struct ncclComm* comm, struct ncclCeCollArgs* args,
...
+        batchOpsParams.sizes[batchOpsParams.numOps] = chunkBytes;
+        batchOpsParams.numOps++;
+      }
+    } else {
+      // Remote copy to other ranks: send to rank dstRank's receive buffer at position comm->rank
+      size_t* peerRecvSizes = ncclAlltoAllvRecvSizes(args->sizes, dstRank, comm->nRanks);
+      size_t* peerRecvDispls = ncclAlltoAllvRecvDispls(args->sizes, dstRank, comm->nRanks);
+      peerDispls = peerRecvDispls[comm->rank];
+      peerRcvSize = peerRecvSizes[comm->rank];
+      NCCLCHECKGOTO(ncclAlltoAllvValidatePeerSendSize(chunkBytes, peerRcvSize, comm->rank, dstRank), ret, fail);
```
pvallem: This per-pair size check is done only by the sender, and on a mismatch we `goto fail` and skip the trailing `ncclMemOpSync` — so the one rank that spots the bad pair returns an error while every other rank blocks forever on the final barrier, turning a detected user error into a cross-rank hang.

Could we lift the check out of the loop and validate the whole gathered matrix once, before the first `ncclMemOpSync`, so every rank reaches the same accept/reject verdict? Something like, in `alltoallv_meta.h`:

```c
// verdict is a pure function of the gathered matrix — identical on every rank
static inline ncclResult_t ncclAlltoAllvValidateSizeMatrix(const size_t* gathered, int nRanks) {
  for (int src = 0; src < nRanks; ++src) {
    const size_t* s = ncclAlltoAllvSendSizes((size_t*)gathered, src, nRanks);
    for (int dst = 0; dst < nRanks; ++dst) {
      const size_t* r = ncclAlltoAllvRecvSizes((size_t*)gathered, dst, nRanks);
      if (s[dst] != r[src]) { return ncclInvalidUsage; }
    }
  }
  return ncclSuccess;
}
```

call it once at the top of `ncclCeAlltoAllv` (before the first barrier) and drop the per-pair check here. Failure stays symmetric (all ranks return `ncclInvalidUsage`, nobody is left waiting), and it's unit-testable without a GPU:

```cpp
TEST_F(CeAlltoAllvEligibilityTest, SizeMatrixVerdictIsRankIndependent) {
  constexpr int nRanks = 4;
  std::vector<size_t> g(4 * nRanks * nRanks, 0);
  for (int src = 0; src < nRanks; ++src) {
    size_t* s = ncclAlltoAllvSendSizes(g.data(), src, nRanks);
    for (int dst = 0; dst < nRanks; ++dst) { s[dst] = (src + 1) * (dst + 2); }
  }
  for (int dst = 0; dst < nRanks; ++dst) {
    size_t* r = ncclAlltoAllvRecvSizes(g.data(), dst, nRanks);
    for (int src = 0; src < nRanks; ++src) { r[src] = (src + 1) * (dst + 2); }
  }
  EXPECT_EQ(ncclAlltoAllvValidateSizeMatrix(g.data(), nRanks), ncclSuccess);
  ncclAlltoAllvSendSizes(g.data(), 0, nRanks)[2] = 99;  // break one ordered pair
  EXPECT_EQ(ncclAlltoAllvValidateSizeMatrix(g.data(), nRanks), ncclInvalidUsage);
}
```

projects/rccl/test/CeAlltoAllvEligibilityTests.cpp
```diff
@@ -0,0 +1,221 @@
...
+TEST_F(CeAlltoAllvEligibilityTest, FuncEnumValue)
+{
+    EXPECT_EQ(ncclFuncAlltoAllv, 18);
+    EXPECT_EQ(ncclNumFuncs, 19);
+}
+
+TEST_F(CeAlltoAllvEligibilityTest, CeImplementedReturnsFalseForUnsupportedCollectives)
+{
+    EXPECT_FALSE(ncclCeImplemented(ncclFuncAllReduce, ncclDevSum, ncclFloat32));
+    EXPECT_FALSE(ncclCeImplemented(ncclFuncBroadcast, ncclDevSum, ncclFloat32));
```
pvallem: `ncclCeImplemented` returns true for AllReduce on a supported driver (it's in the switch), so this `EXPECT_FALSE` fails on the gfx950/newer-ROCm nodes where CE actually runs — it only passes here because the driver is out of range and everything returns false. Broadcast is the genuinely-unsupported one, so let's drop AllReduce from this negative test.
```suggestion
    EXPECT_FALSE(ncclCeImplemented(ncclFuncBroadcast, ncclDevSum, ncclFloat32));
```

projects/rccl/src/collectives.cc
```diff
@@ -571,18 +581,53 @@ ncclResult_t ncclAlltoAllv_impl(const void* sendbuff, const size_t sendcounts[],
...
+  NCCLCHECK(ncclDevrFindWindow(comm, sendbuff, &sendWin));
+  NCCLCHECK(ncclDevrFindWindow(comm, recvbuff, &recvWin));
+  ncclSymRegType_t winRegType;
+  NCCLCHECK(ncclGetSymRegType(sendWin, recvWin, &winRegType));
+  bool hasSysmemSegment = ncclDevrWindowHasSysmemSegment(sendWin) || ncclDevrWindowHasSysmemSegment(recvWin);
+  struct ncclCudaGraph ceGraph;
+  NCCLCHECK(ncclCudaGetCapturingGraph(&ceGraph, stream, comm->config.graphUsageMode));
+  bool ceCapturing = ncclCudaGraphValid(ceGraph);
+
+  if (ncclCeAlltoAllvEligible(comm, datatype, winRegType, hasSysmemSegment, ceCapturing)) {
```
pvallem, line 609: We're not sure this actually engages for the full scale-up domain — since CE AlltoAllv is gated on `nNodes==1`, an 18×4 (72-GPU) deployment would only accelerate the intra-node 4-way exchange and fall back to send/recv for the cross-node AlltoAllv; is that the intended scope here? Please check.

projects/rccl/test/CeAlltoAllvEligibilityTests.cpp
```diff
@@ -0,0 +1,221 @@
...
+#include "common/CeAlltoAllvTestHelpers.hpp"
+
+#include "ce_coll.h"
+#include "collectives.h"
+#include "group.h"
+#include "gtest/gtest.h"
+#include "nccl_common.h"
+#include "sym_kernels.h"
+
+#include <vector>
```
pvallem:
```suggestion
#include "common/CeAlltoAllvTestHelpers.hpp"

#include <vector>

#include "gtest/gtest.h"

#include "ce_coll.h"
#include "collectives.h"
#include "group.h"
#include "nccl_common.h"
#include "sym_kernels.h"
```

projects/rccl/src/collectives.cc
nusislam: The current target platform is MI350. To run it on a POD, we have to enable `MNNVL=1`.

projects/rccl/src/ce_coll.cc
```diff
@@ -703,6 +714,96 @@ ncclResult_t ncclCeAlltoAll(struct ncclComm* comm, struct ncclCeCollArgs* args,
...
+  NCCLCHECKGOTO(ncclCeInitBatchOpsParams(&batchOpsParams, comm->nRanks), ret, fail);
+
+  // Ensure all ranks are ready before starting transfers
+  NCCLCHECKGOTO(ncclMemOpSync(comm, args, stream), ret, fail);
+
+  // Copy data to other ranks: send variable-sized chunk for each destination rank
+  for (int r = 0; r < comm->nRanks; r++) {
+    int dstRank = (comm->rank + r) % comm->nRanks;
+    const size_t chunkBytes = sendSizes[dstRank];
+    if (chunkBytes == 0) continue;
```
pvallem:
```suggestion
    if (chunkBytes == 0) {
      continue;
    }
```

projects/rccl/src/enqueue.cc
```diff
@@ -3430,7 +3431,13 @@ static ncclResult_t ceCollTaskAppend(struct ncclComm* comm, struct ncclInfo* inf
   t->collApiEventHandle = ncclProfilerApiState.collApiEventHandle;
   t->sendWin = sendWin;
   t->recvWin = recvWin;
-
+  t->sizes = nullptr;
+  if (t->func == ncclFuncAlltoAllv && info->sizes != nullptr) {
+    size_t nSizes = 4 * comm->nRanks * comm->nRanks;
+    t->sizes = ncclMemoryStackAlloc<size_t>(&comm->memScoped, nSizes);
+    memcpy(t->sizes, info->sizes, nSizes * sizeof(size_t));
+    for (int r = 0; r < comm->nRanks; r++) t->trafficBytes += t->sizes[comm->rank * 4 * comm->nRanks + r];
```
pvallem, line 3448:
```suggestion
    for (int r = 0; r < comm->nRanks; r++) {
      t->trafficBytes += t->sizes[comm->rank * 4 * comm->nRanks + r];
    }
```

projects/rccl/src/include/alltoallv_meta.h
```diff
@@ -0,0 +1,56 @@
...
+    sizes[nRanks + i] = sdispls[i];
+    sizes[2 * nRanks + i] = recvcounts[i];
+    sizes[3 * nRanks + i] = rdispls[i];
+  }
+}
+
+static inline size_t ncclAlltoAllvTrafficBytes(const size_t* gathered, int rank, int nRanks) {
+  size_t bytes = 0;
+  const size_t* sendSizes = ncclAlltoAllvSendSizes((size_t*)gathered, rank, nRanks);
+  for (int r = 0; r < nRanks; ++r) bytes += sendSizes[r];
```
pvallem:
```suggestion
  for (int r = 0; r < nRanks; ++r) {
    bytes += sendSizes[r];
  }
```

projects/rccl/test/common/CeAlltoAllvTestHelpers.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+namespace RcclUnitTesting
+{
+
+// Runtime driver-version gate mirroring ncclCeImplemented().
+inline bool isCeRuntimeDriverSupported()
+{
+    int driverVer = 0;
+    if (hipDriverGetVersion(&driverVer) != hipSuccess)
+        return false;
```
pvallem, line 27:
```suggestion
    if (hipDriverGetVersion(&driverVer) != hipSuccess)
    {
        return false;
    }
```

projects/rccl/src/include/comm.h
```diff
@@ -231,9 +231,9 @@ struct ncclTaskColl {
   // (an 8-bit signed field caps at 127).
   int32_t nMaxChannels:16;
 #endif
-#ifdef ENABLE_ROCSHMEM
+
   size_t* sizes;
-#endif
+
```
pvallem, line 236: Drop the stray blank left from removing the `#ifdef` so the field sits flush with its neighbors:
```suggestion
  size_t* sizes;
```

projects/rccl/src/include/info.h
```diff
@@ -30,10 +30,10 @@ struct ncclInfo {
   int chunkSteps;
   int sliceSteps;
   const void* acc;
-#ifdef ENABLE_ROCSHMEM
-  // Optional per-operation metadata for rocSHMEM collectives.
+
+  // Optional per-operation metadata (e.g., rocSHMEM collectives, CE AlltoAllv).
   size_t* sizes;
-#endif
+
```
pvallem, line 36: Same here — drop the trailing blank so the field sits flush (the comment above is good to keep):
```suggestion
  size_t* sizes;
```

projects/rccl/test/CeAlltoAllvEligibilityTests.cpp
```diff
@@ -0,0 +1,221 @@
...
+};
+
+TEST_F(CeAlltoAllvEligibilityTest, FuncToStringReturnsAlltoAllv)
+{
+    EXPECT_STREQ(ncclFuncToString(ncclFuncAlltoAllv), "AlltoAllv");
+}
+
+TEST_F(CeAlltoAllvEligibilityTest, FuncEnumValue)
+{
+    EXPECT_EQ(ncclFuncAlltoAllv, 18);
```
pvallem: These just mirror the enum literals in `nccl_common.h`, so they only trip on a benign renumber and exercise no dispatch — consider dropping this test or replacing it with a behavioral check (e.g. a `ncclFuncToString` round-trip).

projects/rccl/test/common/CeAlltoAllvTestHelpers.hpp
```diff
@@ -0,0 +1,50 @@
+/*************************************************************************
+ * Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * See LICENSE.txt for license information
+ ************************************************************************/
+
+#pragma once
```
pvallem:
```suggestion
#ifndef RCCL_TEST_COMMON_CE_ALLTOALLV_TEST_HELPERS_HPP
#define RCCL_TEST_COMMON_CE_ALLTOALLV_TEST_HELPERS_HPP
```

projects/rccl/test/common/CeAlltoAllvTestHelpers.hpp
```diff
@@ -0,0 +1,50 @@
...
+        comm.nRanks            = 4;
+        comm.rank              = 0;
+        comm.symmetricSupport  = true;
+        comm.config.CTAPolicy  = NCCL_CTA_POLICY_ZERO;
+    }
+
+    ncclComm* get() { return &comm; }
+};
+
+} // namespace RcclUnitTesting
```
pvallem, line 53: And close the guard at EOF:
```suggestion
} // namespace RcclUnitTesting

#endif  // RCCL_TEST_COMMON_CE_ALLTOALLV_TEST_HELPERS_HPP
```

projects/rccl/src/collectives.cc
```diff
@@ -514,22 +519,27 @@ ncclResult_t ncclAlltoAllv_impl(const void* sendbuff, const size_t sendcounts[],
...
+
   for (int i = 0; i < nRanks; i++) {
     sdispls1[i] = sdispls[i] * ncclTypeSize(datatype);
     rdispls1[i] = rdispls[i] * ncclTypeSize(datatype);
     sendcounts1[i] = sendcounts[i] * ncclTypeSize(datatype);
     recvcounts1[i] = recvcounts[i] * ncclTypeSize(datatype);
+    sizes[i] = sendcounts1[i];
+    sizes[nRanks + i] = sdispls1[i];
+    sizes[2 * nRanks + i] = recvcounts1[i];
+    sizes[3 * nRanks + i] = rdispls1[i];
```
pvallem, line 551: Now that this loop populates `sizes[]` unconditionally, the `for` loop inside the `ENABLE_ROCSHMEM` block below (~L543-548) re-assigns the identical values from the identical sources with no intervening change — it is dead and can be dropped.

pvallem (approved):

nusislam:
@copilot resolve the merge conflicts in this pull request

nusislam:
@copilot resolve the merge conflicts in this pull request

PR #6940, opened 2026-06-08
Revert "Fix SQ-block counter collection on GFX11+ GPUs with WGPs harvesting - [Re-upload]"

ApoKalipse-V:
Reverts ROCm/rocm-systems#5943

PR has caused broken counter collection on all gfx11 whenever SQ counters are involved.
This is caused by pmc_builder.h remapping WGPs without counters.cpp iterate_data knowing about it.

Reproducer by asking a SQ counter and any other counter. E.g:

 rocprofv3 --pmc SQ_WAVES SQ_CYCLES -d test -f csv

SQ_WAVES,1024

 rocprofv3 --pmc SQ_WAVES SQ_CYCLES -d test -f csv

SQ_WAVES,4301154
SQ_CYCLES,25794636

PR #5943, opened 2026-05-09
Fix SQ-block counter collection on GFX11+ GPUs with WGPs harvesting - [Re-upload]

vipangul:
## Summary

Fixes per-WGP performance counter aliasing on GFX11+ AMD GPUs that ship
with WGP harvesting (e.g. gfx1151 / Strix Halo).

This PR replaces the previously-closed draft #3900. Two commits:
1. The fix itself: introduce `aqlprofile_agent_info_v2_t` carrying
   physical CU topology, plumb the `cu_bitmap` from DRM into aqlprofile,
   and iterate only active WGPs in the GFX11 SQ-block read loop.
2. Review-feedback follow-up: derive the iteration bound from the
   bitmap alone (highest set bit), drop the redundant
   `cu_per_simd_array` field from the new V2 struct and from the
   aqlprofile-side plumbing.

## Background

On GFX11+ parts, SQ-block (WGP-level) counters are read by walking
`GRBM_GFX_INDEX` over physical WGP indices within each SE/SA. The
existing iteration assumes WGPs are densely packed in `[0, wgp_per_sa)`
where `wgp_per_sa` is derived from the active CU count. On parts with
WGP harvesting the active WGPs are sparse, so the dense iteration
queries harvested slots and the hardware aliases those reads onto active
slots — producing per-dispatch counter values that fluctuate between 0,
the architecturally-correct value, and 2× the correct value.

On gfx1151 (1 SE × 2 SA × 10 physical WGPs per SA, with harvesting),
this manifests as roughly:

| Reported `SQ_WAVES` | Frequency  | Meaning              |
|--------------------:|-----------:|----------------------|
|                 `0` |    ~17.9 % | counter missed       |
|        correct ×1   |    ~62.5 % | correct              |
|        correct ×2   |    ~18.8 % | aliased / doubled    |

## Fix

* New `AQLPROFILE_AGENT_VERSION_V2` and `aqlprofile_agent_info_v2_t`
  carrying `cu_bitmap[4][4]` (per-SE/SA active CU bitmap; matches the
  AMDGPU `drm_amdgpu_info_device` ABI).
* `core/pm4_factory.cpp` adds a V2 `RegisterAgent` overload that copies
  the bitmap into the internal `AgentInfo` cache, and a DRM fallback
  (`populate_cu_bitmap_from_drm`) for V0/V1 callers — it opens
  `/dev/dri/renderD*` and matches by
  `(cu_active_number, num_shader_engines, num_shader_arrays_per_engine)`
[...]

PR #8229, opened 2026-07-07
fix(aqlprofile): GFX11 harvested-WGP SQ aliasing + multi-counter desync

vipangul:
## Summary

Re-lands the GFX11+ harvested-WGP counter fix (#5943, reverted by #6940) with the
regression that caused the revert fixed. Addresses both:
1. Per-WGP counter aliasing on harvested parts (e.g. gfx1151/Strix Halo) —
   per-dispatch SQ counters fluctuated `{0, correct, 2x}`.
2. The multi-counter desync that #6940 reverted #5943 for: collecting an SQ
   counter alongside any other counter returned garbage
   (`rocprofv3 --pmc SQ_WAVES SQ_CYCLES` → SQ_WAVES = 4301154).

## Background

#5943 skipped harvested WGPs by remapping sparse active WGPs into dense output
slots, but the reader (`counters.cpp` iterate_data / `GetBlockSamplesCount`) still
assumed a fixed `GetNumWGPs() == wgp_per_sa` samples per SA. On harvested parts the
emitted count differed, so every block after the first SQ counter was read at the
wrong offset — hence #6940.

## Fix

- Keep the harvest-aware read (skip harvested WGPs via a per-SA `cu_bitmap` mask,
  no aliasing), but **stop remapping**: emit a uniform `wgp_slots_per_sa_` entries
  per SA, issuing real reads for active WGPs and zero-filling harvested slots.
- `GetNumWGPs()` returns that same bound so emitter, buffer sizing, and reader stay
  aligned (this is the piece #5943 was missing).
- Mirrors the existing GFX12 dense + zero-fill pattern; GFX12 unchanged.
- Retains #5943's V2 `cu_bitmap` plumbing (DRM → aqlprofile) since develop no longer
  has it post-revert.

## Test Plan / Result (gfx1151, Strix Halo, middle-harvested: active WGPs {0,1,2,4})

Deterministic workload (16384 threads → SQ_WAVES = Grid/32 = 512):
- Single SQ; multi-SQ (`SQ_WAVES SQ_BUSY_CYCLES [SQ_INSTS_VALU]`); SQ + non-SQ
  (`SQ_WAVES GRBM_COUNT`, `SQ_WAVES GL2C_HIT`) → **SQ_WAVES = 512** in every case;
  co-collected counters report sane non-zero values (GL2C_HIT=530, GRBM_COUNT=64164).
- Per-WGP JSON: sum = 512, layout `[32,32,32,0,32]/[32,32,32,0,32]/[0,32,32,32,32]/[32,32,32,32,0]`.
- 10x stability: all 512.
- `aqlprofile-pm4-tests` green (pre-existing `SqttBuilderTest.BufferStepCalculation`
  failure unrelated).

Reverts the revert: the #6940 reproducer (SQ + another counter) now returns the
correct SQ value.

## Submission Checklist


vipangul:
@ApoKalipse-V Please refer to this draft PR for code review and if we should cover any more test scenarios as part of this change.

projects/rocprofiler-sdk/source/lib/aqlprofile/pm4/pmc_builder.h
```diff
@@ -282,6 +293,62 @@ class GpuPmcBuilder
...
+    {
+        // Fits Navi31 (6 x 2), MI200 (8 x 1), and any future SE x SA-per-SE
+        // topology too large for a hardcoded kMaxSA.
+        sa_cu_mask_.assign(se_number_, std::vector<uint32_t>(sarrays_per_se_, 0u));
+
+        constexpr uint32_t kMaxWgpPerSa  = 16;
+        constexpr uint32_t bitmap_se_lim = std::extent_v<decltype(aqlprofile_cu_bitmap_t::bits), 0>;
+        constexpr uint32_t bitmap_sa_lim = std::extent_v<decltype(aqlprofile_cu_bitmap_t::bits), 1>;
+        // Matches the kernel's cu_bitmap_sh_mul (== SAs per SE: always 2 on GFX11).
+        const uint32_t cu_bitmap_sh_mul = sarrays_per_se_;
```
ApoKalipse-V: Do we need this variable?

projects/rocprofiler-sdk/source/lib/aqlprofile/pm4/pmc_builder.h
```diff
@@ -282,6 +293,62 @@ class GpuPmcBuilder
...
+                const uint32_t max_wgp    = max_cu_bit / 2u;
+                if(max_wgp > max_wgp_index) max_wgp_index = max_wgp;
+            }
+        }
+
+        // Grow wgp_per_sa_ to the physical span so [0, wgp_per_sa_) covers a
+        // middle-harvested WGP (gfx1151: active {0,1,2,4} -> 4 -> 5). GFX11
+        // only; GFX12's dense read path keeps the cu_num-derived count.
+        if constexpr(Primitives::GFXIP_LEVEL == 11)
+            if(max_wgp_index + 1 > wgp_per_sa_) wgp_per_sa_ = max_wgp_index + 1;
```
ApoKalipse-V:
```suggestion
        wgp_per_sa_ = std::max(wgp_per_sa_, max_wgp_index + 1);
```

projects/rocprofiler-sdk/source/lib/aqlprofile/pm4/pmc_builder.h
```diff
@@ -282,6 +293,62 @@ class GpuPmcBuilder
                    : instance_index;
     }
 
+    // Populate sa_cu_mask_[se][sa] and, on GFX11, grow wgp_per_sa_ to the
```
ApoKalipse-V: Can you improve the comments here? it's long yet doesn't seem to explain the logic properly

projects/rocprofiler-sdk/source/lib/aqlprofile/pm4/pmc_builder.h
```diff
@@ -154,6 +156,15 @@ class GpuPmcBuilder
     //       after CU mask support is added to agent_info
     bool asymmetric_cu_patch;
 
+    // WGP harvesting (GFX11+): per-(SE, SA) cu_bitmap mask (2 CU bits/WGP). The
```
ApoKalipse-V: Same here.

ApoKalipse-V (approved):
Functionality wise looks good, just needs simplifications and improved comments.
It felt like I understood things better by ignoring the comments.

ApoKalipse-V:
Hi @bwelton can you take a look since you also reviewed the original PR?

projects/rocprofiler-sdk/source/lib/aqlprofile/pm4/pmc_builder.h
vipangul: Updated, pl take a look.

projects/rocprofiler-sdk/source/lib/aqlprofile/pm4/pmc_builder.h
vipangul: Added this suggestion.

projects/rocprofiler-sdk/source/lib/aqlprofile/pm4/pmc_builder.h
vipangul: Removed it.

bwelton:
**Q: behavior on >4-SE parts (Navi31) and future expandability of `cu_bitmap`**

`aqlprofile_cu_bitmap_t` is fixed at `bits[4][4]` (`AQLPROFILE_DRM_CU_BITMAP_NUM_SE`/`_NUM_SA_PER_SE`), and `build_sa_cu_mask()` in `pmc_builder.h` folds logical SEs beyond the bitmap dimension into higher SA columns:

```cpp
const uint32_t raw_se = se % bitmap_se_lim;
const uint32_t raw_sa = sa + (se / bitmap_se_lim) * sarrays_per_se_;
```

with the comment that this "mirrors kernel `mqd_symmetrically_map_cu_mask`."

Two questions:

1. For a >4-SE part like Navi31 (6 SE × 2 SA), does this folding reconstruct the *same* (SE, SA) → bitmap-slot mapping the kernel actually uses to pack `drm_amdgpu_info_device.cu_bitmap`? If the kernel's packing convention differs from this reconstruction, harvested WGPs on those parts would be mis-detected (wrong read/zero-fill placement → wrong SQ counters), even though it stays memory-safe via the `.at()` bounds checks and the out-of-window fully-active fallback. The validation in the PR description is all gfx1151 (≤4 SE), so this folding path appears untested — can it be confirmed against the kernel source, or at least called out as untested for >4-SE parts?

2. Is the `4×4` layout intended to be expandable later if a future part needs more than 4 SE / 4 SA-per-SE genuinely represented (rather than folded)? The ABI static_asserts make widening a deliberate, build-breaking V2 bump, which is good — but is folding the long-term intent, or a stopgap until the bitmap dimensions can grow to match the kernel uAPI?

bwelton:
**Reduce duplication: the gfx1250/gfx94/gfx95 chip patch block is now verbatim in both `RegisterAgent` overloads**

In `core/pm4_factory.cpp`, the V1 `RegisterAgent(const aqlprofile_agent_info_v1_t*)` and the new V2 `RegisterAgent(const aqlprofile_agent_info_v2_t*)` contain an identical ~15-line block:

```cpp
if(!strncmp(int_agent_info.name, "gfx1250", 7))
{
    int_agent_info.cu_num      = agent_info->se_num * agent_info->shader_arrays_per_se * 9 * 2;
    int_agent_info.xcc_per_aid = 4;
}
else if(!strncmp(int_agent_info.name, "gfx94", 5) || !strncmp(int_agent_info.name, "gfx95", 5))
{
    int_agent_info.xcc_per_aid = 2;
}
else
{
    int_agent_info.xcc_per_aid = 1;
}
```

Could this be factored into a small helper (e.g. `apply_chip_specific_patch(AgentInfo&, const char* gfxip)`) called from both overloads? The `populate_agent_name()` extraction already sets the precedent. The risk of leaving it duplicated is that the next chip-specific patch (these are all marked `TODO: ... will remove after CU mask support`) has to be applied in two places and can silently drift between the V1 and V2 paths.

bwelton:
**Testing: can we add hardware-free unit coverage for the harvest logic?**

The new tests (`aql_profile_v2_tests.cpp`, `aql_profile_v2_c_test.c`) only assert the V2 **ABI** (size/offset/alignment). The actual harvest logic — which is exactly what was reverted once (#6940) for an emitter/reader mismatch — has no automated coverage; all functional evidence is manual gfx1151 runs.

Since `GpuPmcBuilder` takes a plain `AgentInfo` (which now carries `cu_bitmap`), we should be able to drive this entirely on CPU with no GPU present, by constructing an `AgentInfo` with a synthetic `cu_bitmap` and inspecting the built command stream / counts. Concretely, could we add `PmcBuilderTest` cases that:

1. **Emitter/reader invariant** (the #6940 regression guard): feed a harvested bitmap (e.g. gfx1151 active WGPs `{0,1,2,4}`) and assert the emitted per-SA sample count equals `GetNumWGPs()` — i.e. writer stride == reader stride. This is the single invariant whose violation caused the revert.
2. **`wgp_per_sa_` growth**: assert `GetNumWGPs()` grows to the physical span (5 for `{0,1,2,4}`) on GFX11 and is unchanged on GFX12.
3. **Zero-fill placement**: with a `buf` provided, assert harvested slots are zeroed at their physical index and active WGPs are read at their real slot (no compaction/remap).
4. **`build_sa_cu_mask` folding**: a synthetic 6-SE (Navi31-style) bitmap to lock down the SE→SA-column folding described above.
5. **UB guards**: `cu_bitmap == 0` (fully-active synthesize), `wgp_per_sa_ == 0`, and `wgp_per_sa_ >= kMaxWgpPerSa` (the `1u << 32` / `clz(0)` edge cases).

Is there an existing seam (mock `BuildCopyCounterDataPacket` / passing `buf == nullptr` to just count slots) that makes 1–3 feasible without hardware? If the read loop is hard to unit-test directly, would exposing `sa_cu_mask_` / the emitted count via a test hook be acceptable?

vipangul:
> **Q: behavior on >4-SE parts (Navi31) and future expandability of `cu_bitmap`**
> 
> `aqlprofile_cu_bitmap_t` is fixed at `bits[4][4]` (`AQLPROFILE_DRM_CU_BITMAP_NUM_SE`/`_NUM_SA_PER_SE`), and `build_sa_cu_mask()` in `pmc_builder.h` folds logical SEs beyond the bitmap dimension into higher SA columns:
> 
> ```c++
> const uint32_t raw_se = se % bitmap_se_lim;
> const uint32_t raw_sa = sa + (se / bitmap_se_lim) * sarrays_per_se_;
> ```
> 
> with the comment that this "mirrors kernel `mqd_symmetrically_map_cu_mask`."
> 
> Two questions:
> 
> 1. For a >4-SE part like Navi31 (6 SE × 2 SA), does this folding reconstruct the _same_ (SE, SA) → bitmap-slot mapping the kernel actually uses to pack `drm_amdgpu_info_device.cu_bitmap`? If the kernel's packing convention differs from this reconstruction, harvested WGPs on those parts would be mis-detected (wrong read/zero-fill placement → wrong SQ counters), even though it stays memory-safe via the `.at()` bounds checks and the out-of-window fully-active fallback. The validation in the PR description is all gfx1151 (≤4 SE), so this folding path appears untested — can it be confirmed against the kernel source, or at least called out as untested for >4-SE parts?
> 2. Is the `4×4` layout intended to be expandable later if a future part needs more than 4 SE / 4 SA-per-SE genuinely represented (rather than folded)? The ABI static_asserts make widening a deliberate, build-breaking V2 bump, which is good — but is folding the long-term intent, or a stopgap until the bitmap dimensions can grow to match the kernel uAPI?

@bwelton 

Fixed the citation: the fold now references the function it actually mirrors — kernel gfx_v11_0_get_cu_info, which packs cu_bitmap[i%4][j+(i/4)*sh_per_se] — instead of mqd_symmetrically_map_cu_mask. Correctness is validated two ways: on Navi31 (6 SE) where the SE→SA-column fold is exercised, and on gfx1151 (Strix Halo) hardware for the harvested-WGP path. Added a FoldingNavi31 unit test that asserts logical SE4/SE5 pull from the folded columns of raw rows 0/1 and that the out-of-window rows never leak. On the 4×4 width: that's the fixed DRM uAPI shape (locked by a static_assert on aqlprofile_cu_bitmap_t), and the fold is exactly how the kernel maps >4 SEs onto it — widening would be a deliberate V2 ABI bump, out of scope for this PR.

vipangul:
> **Reduce duplication: the gfx1250/gfx94/gfx95 chip patch block is now verbatim in both `RegisterAgent` overloads**
> 
> In `core/pm4_factory.cpp`, the V1 `RegisterAgent(const aqlprofile_agent_info_v1_t*)` and the new V2 `RegisterAgent(const aqlprofile_agent_info_v2_t*)` contain an identical ~15-line block:
> 
> ```c++
> if(!strncmp(int_agent_info.name, "gfx1250", 7))
> {
>     int_agent_info.cu_num      = agent_info->se_num * agent_info->shader_arrays_per_se * 9 * 2;
>     int_agent_info.xcc_per_aid = 4;
> }
> else if(!strncmp(int_agent_info.name, "gfx94", 5) || !strncmp(int_agent_info.name, "gfx95", 5))
> {
>     int_agent_info.xcc_per_aid = 2;
> }
> else
> {
>     int_agent_info.xcc_per_aid = 1;
> }
> ```
> 
> Could this be factored into a small helper (e.g. `apply_chip_specific_patch(AgentInfo&, const char* gfxip)`) called from both overloads? The `populate_agent_name()` extraction already sets the precedent. The risk of leaving it duplicated is that the next chip-specific patch (these are all marked `TODO: ... will remove after CU mask support`) has to be applied in two places and can silently drift between the V1 and V2 paths.

@bwelton 
Created a helper function `populate_arch_specific_info()`, called from both RegisterAgent overloads right after populate_agent_name().

vipangul:
> **Testing: can we add hardware-free unit coverage for the harvest logic?**
> 
> The new tests (`aql_profile_v2_tests.cpp`, `aql_profile_v2_c_test.c`) only assert the V2 **ABI** (size/offset/alignment). The actual harvest logic — which is exactly what was reverted once (#6940) for an emitter/reader mismatch — has no automated coverage; all functional evidence is manual gfx1151 runs.
> 
> Since `GpuPmcBuilder` takes a plain `AgentInfo` (which now carries `cu_bitmap`), we should be able to drive this entirely on CPU with no GPU present, by constructing an `AgentInfo` with a synthetic `cu_bitmap` and inspecting the built command stream / counts. Concretely, could we add `PmcBuilderTest` cases that:
> 
> 1. **Emitter/reader invariant** (the [Revert "Fix SQ-block counter collection on GFX11+ GPUs with WGPs harvesting - [Re-upload]" #6940](https://github.com/ROCm/rocm-systems/pull/6940) regression guard): feed a harvested bitmap (e.g. gfx1151 active WGPs `{0,1,2,4}`) and assert the emitted per-SA sample count equals `GetNumWGPs()` — i.e. writer stride == reader stride. This is the single invariant whose violation caused the revert.
> 2. **`wgp_per_sa_` growth**: assert `GetNumWGPs()` grows to the physical span (5 for `{0,1,2,4}`) on GFX11 and is unchanged on GFX12.
> 3. **Zero-fill placement**: with a `buf` provided, assert harvested slots are zeroed at their physical index and active WGPs are read at their real slot (no compaction/remap).
> 4. **`build_sa_cu_mask` folding**: a synthetic 6-SE (Navi31-style) bitmap to lock down the SE→SA-column folding described above.
> 5. **UB guards**: `cu_bitmap == 0` (fully-active synthesize), `wgp_per_sa_ == 0`, and `wgp_per_sa_ >= kMaxWgpPerSa` (the `1u << 32` / `clz(0)` edge cases).
> 
> Is there an existing seam (mock `BuildCopyCounterDataPacket` / passing `buf == nullptr` to just count slots) that makes 1–3 feasible without hardware? If the read loop is hard to unit-test directly, would exposing `sa_cu_mask_` / the emitted count via a test hook be acceptable?

Added pmc_builder_harvest_tests.cpp plus a small gfx12 companion. These are CPU-only — they run the real GpuPmcBuilder against a synthetic AgentInfo, so no GPU is needed (the offset table just falls back to sienna_cichlid).

**What they cover:**
the emitter/reader stride invariant that caused #6940 (a real Read() into an allocated buffer)
wgp_per_sa_ growth — gfx11 grows to span 5, gfx12 stays cu_num-derived
harvest zero-fill placement — harvested slots are zeroed in place, active WGPs keep their physical index
the 6-SE fold
the UB guards (clz(0), 1u << 32)
The harvest cases model the real gfx1151 (Strix Halo) {0,1,2,4} pattern.

6 Tests can be run using below command - 
./_build_test/bin/aqlprofile-pm4-tests --gtest_filter='PmcBuilderHarvest*'

bwelton (approved):

PR #8231, opened 2026-07-07
rocjitsu: route AMDGPU instruction VGPR access through RegisterAccess

newling:
## Summary

This PR introduces `RegisterAccess` as the instruction-facing boundary for
AMDGPU VGPR reads and writes. It routes SIMD helpers, matrix helpers, shared
AMDGPU helpers, and generated instruction code through that boundary.

Before this change, instruction-visible VGPR access used operand APIs, direct
CU physical VGPR reads/writes, `SimdAccess`, and raw VGPR storage views. After
this change, instruction emulators use:

- `Operand::read_*` / `Operand::write_*` for operand semantics;
- `RegisterAccess::read_operand()` for logical source operands in SIMD helpers;
- `RegisterAccess::write_operand()` for destination writes;
- `RegisterAccess::readwrite_operand()` when a destination is also an input;
- `RegisterAccess::{read,write,readwrite}_vgpr_region()` for physical VGPR
  region access.

Read and read-write acquisition reports VGPR reads before exposing storage.
Write-only access does not report reads.

## Impact

- SIMD source, destination, and accumulator views use `RegisterAccess`.
- MFMA and WMMA input and accumulator reads use `RegisterAccess`.
- Shared helpers such as DPP/SDWA, address calculation, and DS permute/swizzle
  use `RegisterAccess`.
- Generated AMDGPU ISA files are regenerated with RegisterAccess-backed physical
  VGPR access.
- VM-level CU register accessors are documented as VM/storage-level APIs.
- A temporary checker guards against reintroducing direct instruction-code VGPR
  access until a narrower C++ interface makes those paths unavailable at compile
  time.

`ExecutionPluginTest.MfmaReadObservationReportsRace` covers the main failure
mode: an outstanding VGPR load must be reported as a race when MFMA reads that
VGPR. Without the read hook, the race detector does not see the VGPR read.

## Notes

The current execution path still passes `Wavefront&` to instruction bodies. A
future cleanup can replace that with a narrower instruction context or capability
tokens for VM-only register operations, moving more of this boundary into C++
interface structure.

The final commits are generated-output commits:

1. non-gfx1250 AMDGPU ISA regeneration;
2. gfx1250 / MI450 ISA regeneration.

## Validation

- `check_register_access.py`
- `rocjitsu_tests`
- Focused ctest suite: `RegisterAccessTest`, `ExecutionPluginTest`,
  `SharedInfraTest`, `InstructionExecutionHarnessTest`, `MfmaExecTest`,
  `Gfx1250WmmaTest`, `SimdCorrectness`
- Changed-file pre-commit

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,6 +14,15 @@ exclude: |
     | projects/rocprofiler-compute/
 
 repos:
+  - repo: local
+    hooks:
+      - id: rocjitsu-register-access
+        name: rocjitsu register access boundary
+        entry: python3 emulation/rocjitsu/scripts/check_register_access.py
+        language: system
+        pass_filenames: false
+        files: ^emulation/rocjitsu/
+
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v5.0.0
     hooks:
--- a/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
+++ b/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
@@ -3418,6 +3418,24 @@ class CodeGenerator:
             L.append(f'  uint32_t offset = inst_.offset0 | (inst_.offset1 << 8);')
             L.append(f'  uint32_t lane_group_width = wf.wf_size();')
             L.append(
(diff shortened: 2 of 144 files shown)


newling:
Closing in favor of #8417.

#8417 keeps this `RegisterAccess` migration and replaces the temporary script/social boundary with C++ enforcement via `InstructionComputeUnitView`.

PR #8232, opened 2026-07-07
fix(rocjitsu): Harden interposer concurrency

atgutier:
## Motivation

Hardens and fixes several issues with interposer concurrency to make the KFD interposer and simulated driver safe under concurrent open/close/dup/dup2/dup3/fcntl and fork, and fix a shared-memory mmap fall-through that could silently detach a mapping from the daemon.

## Technical Details

Remote driver lifetime & fd tracking:
- Snapshot the active RemoteDriver through an atomic<shared_ptr> so
  ioctl/mmap/munmap dispatch holds a lifetime-extending reference rather
  than racing a concurrent teardown; handshake metadata (topology/drm/
  gpu_info) is immutable after open() so snapshots stay valid.
- Track dup/dup2/dup3/fcntl(F_DUPFD) KFD fds in a backend map
  (Local/Remote) with reserve-before-syscall then reconcile/rollback, so
  a dup keeps its backend alive after the primary closes and an
  overwriting dup2/dup3 invalidates the old primary identity.
- Serialize remote-primary invalidation with open under remote_mutex_:
  get_or_create_remote() returns the fd captured under the lock, and
  invalidate_overwritten_kfd_fd() compare-and-clears remote_kfd_fd_
  (with the reference drop and teardown) under the same lock, so a racing
  open("/dev/kfd") can no longer return -1 or a reused non-KFD fd.
- Route KFD/DRM ioctl and mmap dispatch strictly by recorded backend and
  by live remote() snapshot; drop the legacy local-driver fallback so a
  Remote-tagged dup is never misrouted to the local driver during a
  teardown window, and only route the late-ioctl safety net when the
  backend is positively known to be Remote.

Local driver:
- Make SimulatedKfd::fd_ atomic and CAS-publish the backing memfd once
  across racing open()/open_process(), closing the loser's fd to avoid a
  double memfd_create / fd leak.
- Serialize local-primary invalidation with open() under process_mutex_;
  invalidate_primary_fd() returns whether this call cleared the primary
  so the caller drops the open reference exactly once (no double-close).

Shared-memory mmap:
- Make RemoteDriver::find_memfd_for_addr() a tri-state (kNotFound/kFound/
  kDupFailed) with an overflow-safe range check. On a matched range whose
  dup or shared mmap fails, fail the mmap closed (preserving the real
  errno) instead of falling back to an anonymous MAP_FIXED mapping, which
  would silently detach the address from the daemon's shared memory.

Fork:
- Reset remote/guest state in a forked child via atomic store(nullptr)
  (not placement-new on the live atomic) and reinitialize mutexes.

Also: use real().fcntl / raw syscalls for internal memfd dups so the
interposer does not re-enter its own fcntl hook; take a single remote
snapshot when redirecting sysfs paths so topology and drm paths cannot
be combined from different RemoteDrivers.

Testing:
- Add LD_PRELOAD interposer_dup_test coverage for dup/dup2/dup3/fcntl KFD
  fd routing, reopen-after-primary-overwrite, and a threaded reopen-under-
  contention stress, plus daemon-backed (RemoteDriver) variants that
  exercise the remote reopen/re-mint and invalidation-vs-open paths the
  in-process harness cannot reach.

## JIRA ID

JIRA ID - N/A
PR #8232

## Test Plan

Run all CI tests

## Test Result

CI tests: PASSED

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
@@ -76,8 +76,8 @@ RJ_DIAGNOSTIC_POP
 #include <thread>
 #include <unistd.h>
 #include <unordered_map>
-#include <unordered_set>
 #include <utility>
+#include <vector>
 
 extern "C" rocjitsu::ExecutionPlugin *createKernelLoggingPlugin();
 extern "C" rocjitsu::ExecutionPlugin *createRaceDetectorPlugin();
@@ -172,6 +172,9 @@ __attribute__((constructor)) void rj_install_signal_handler() {
 /// @brief All mutable interposer state.
 class InterposerContext {
 public:
+  /// @brief Which backend holds the open reference for a tracked KFD dup fd.
+  enum class DupBackend : uint8_t { Local, Remote };
+
   static inline thread_local bool in_construction = false;
   static rocjitsu::LibcPassthrough &real() { return rocjitsu::libc_passthrough(); }
   static InterposerContext &ctx;
@@ -192,6 +195,26 @@ public:
     if (guest_driver_)
       guest_driver_->reset_after_fork();
     guest_driver_.reset();
+    // Drop the child's reference to the inherited RemoteDriver. Storing nullptr
+    // releases this shared_ptr; if it was the last reference in the child, the
+    // child's ~RemoteDriver runs and closes only the child's inherited (dup'd)
+    // fds — it does not send RPC_CLOSE and cannot invalidate the parent's live
+    // connection, so this is safe in a forked child.
+    //
+    // Unlike the plain mutexes below, remote_ is NOT reinitialized via
+    // placement-new: re-constructing a live object over itself ends the current
+    // object's lifetime without running its destructor, which is UB for a
+    // non-trivial type like std::atomic<std::shared_ptr<>> (it leaks/By-passes the
+    // control-block bookkeeping). A plain store(nullptr) is the correct way to
+    // release it.
+    //
+    // Caveat: std::atomic<std::shared_ptr<>> may serialize on a libstdc++-internal
+    // pool spinlock. If a parent thread was mid remote_.load()/store() at fork(),
+    // the child inherits that lock held-by-a-dead-thread and this store() could
+    // block. reset_after_fork() therefore assumes no in-flight remote_ atomic
+    // access at the fork point (the common fork-then-exec / single-threaded-fork
+    // case) — placement-new would not sidestep this deadlock either (still UB),
+    // and leaving remote_ non-null would wrongly reuse the parent's connection.
     remote_.store(nullptr, std::memory_order_relaxed);
     remote_kfd_fd_.store(-1, std::memory_order_relaxed);
     remote_open_refs_.store(0, std::memory_order_relaxed);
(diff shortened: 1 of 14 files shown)


bjacob (changes requested):
(Hey Tony, I'm not familiar with this code and the use of relaxed atomics, which are particularly hard to verify correctness of, made me switch right away in to "let's just ask Codex" mode rather than try to read this myself, for now.  So the below is entirely Codex's review. Maybe after 1-2 initial AI iterations, once we've narrowed remaining issues to something precise within mostly-correct code, we can swich to human discussion).

(Codex review begins)

I don’t think this is concurrency-correct yet.

First, the remote lifetime issue appears to be preexisting, but it is not fixed by this PR despite the commit message saying this switched to shared_ptr. `remote_` is still `std::atomic<RemoteDriver *>`, and callers of `remote_lookup()` / `remote()` dereference the returned raw pointer outside `remote_mutex_`. A racing close can enter `teardown_remote()`, store nullptr, and delete the `RemoteDriver` while another thread is between the load and `remote->ioctl()` / `remote->mmap()` / `remote->munmap()`. The atomic makes the pointer access data-race-free, but it does not protect object lifetime. This needs either a real shared_ptr snapshot, a lock held across the call, or a no-delete lifetime policy.

Second, the `rj_vm_` change introduces a new publication-ordering problem. `get_or_create()` now starts `rj_vm_run(tmp)` before publishing `tmp` to `rj_vm_`, and even clears `in_construction` before the release-store. Acquire readers of `rj_vm_` do not synchronize with work already done by the detached engine thread, so this no longer publishes a fully initialized/not-yet-running VM the way the old code effectively did. Please store `tmp` before starting the detached run thread, and clear `in_construction` only after the published state is coherent.

atgutier:
@copilot do another review

bjacob:
(Another Codex review!  Codex's words, not mine.  But we seem to be converging so I'll try to ramp up to become able to actually human-review this for the next iteration).

Thanks, this is much closer, but I think there is still a concurrency bug here.

`shared_ptr` prevents the `RemoteDriver` object storage from being freed while a caller holds a snapshot, but `teardown_remote()` still calls `active_remote->close()` while other snapshots may exist. `RemoteDriver::close()` mutates/clears `topology_path_`, `drm_path_`, `gpu_info_`, `has_gpu_info_`, `handle_memfds_`, `alloc_ranges_`, and `sock_`.

Some readers are not synchronized with those mutations. For example, `remote_topology_path()` and `remote_drm_path()` now copy strings from the `RemoteDriver` without holding `remote_mutex_` or `rpc_mutex_`, while `RemoteDriver::close()` clears those same strings. That is a real data race on the `std::string`s. This part looks like a regression in the latest revision: the old code held `remote_mutex_` while copying these paths, and teardown also held `remote_mutex_`.

There is also a remaining preexisting/incomplete-fix issue in `interposer_gpu_info()`: it returns a raw pointer into `RemoteDriver`’s `gpu_info_`, but the local `shared_ptr` snapshot is destroyed before the caller uses the pointer. A concurrent teardown can then clear or destroy the object while the DRM ioctl code is reading `gpu->...`.

So I think the fix needs to either:
- keep remote metadata immutable after `open()` and avoid clearing it in `close()`,
- return metadata by value / optional instead of raw internal pointers,
- or consistently protect metadata reads and `close()` with the same lock.

Separately, `get_or_create_remote()` still publishes `remote_` before publishing `remote_kfd_fd_` and `remote_open_refs_`, while `initialized()` still checks `remote_`. That leaves a small window where readers can observe “initialized” remote state before the fd/ref state is visible. This is lower priority than the data races above, but it contradicts the comment that only a fully-open driver is visible.

bjacob (changes requested):
(Left review as a PR comment).

atgutier:
@copilot do another review

bjacob:
One more round with Codex :-)

This looks much better: the original raw-pointer lifetime issue is addressed, the rj_vm publication order is fixed, and the remote metadata reads no longer race close().

I see one remaining concurrency issue. `RemoteDriver::ioctl()` still does `assert(sock_ >= 0)` before taking `rpc_mutex_`, but `RemoteDriver::close()` writes `sock_ = -1` while holding that mutex. With the new shared_ptr snapshot model, a thread can legitimately hold a RemoteDriver snapshot while another thread runs teardown/close, so this assert becomes an unsynchronized read of `sock_` racing with close()'s write in debug builds. Please either remove the assert or perform the disconnected check under `rpc_mutex_` with the rest of the socket accesses.

bjacob:
**Findings**

1. [interposer.cpp:375](/home/benoit/workspace/TheRock/rocm-systems/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp:375): `track_dup()` still guesses the dup backend from global state: `driver()` first, then remote. If local mode was created first and the daemon later becomes available, both `driver()` and `remote_` can be non-null. A `dup()` of the remote KFD fd would then retain/tag the dup as `Local`, so close releases the wrong backend. Root cause is preexisting, but this PR’s new `DupBackend` invariant makes it worth fixing here. `track_dup` probably needs to be told the source backend, rather than rediscovering it.

2. [remote_driver.cpp:115](/home/benoit/workspace/TheRock/rocm-systems/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/remote_driver.cpp:115), [interposer.cpp:1288](/home/benoit/workspace/TheRock/rocm-systems/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp:1288): `find_memfd_for_addr()` returns a raw memfd while holding `rpc_mutex_`, but the caller uses it after the lock is released. A concurrent last-close can enter `RemoteDriver::close()` and close `handle_memfds_` before the caller’s `ftruncate`/`fallocate`/`mmap`. This is preexisting, but not solved by the new `shared_ptr` lifetime fix. Returning a dup’d fd, or holding ownership through the mmap, would close that hole.

The previous issues I’d flagged look addressed in `e86ab99285`: metadata is no longer cleared during close, `interposer_gpu_info()` copies by value, remote publish order is sane, and the unlocked `sock_` assert is gone. I did not build; this was a review pass plus `git diff --check`, which is clean.

bjacob:
One more from Codex:

I re-reviewed latest PR head `0049971699`.

**Finding**

[interposer.cpp:1143](/home/benoit/workspace/TheRock/rocm-systems/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp:1143): `dup2`/`dup3` still do not reconcile KFD tracking for the overwritten `newfd`. They remove sysfs/DRM tracking, but if `newfd` was already a tracked KFD fd, `track_dup(rc, backend)` sees the old `kfd_dup_fds_` entry and leaves it in place. In a local/remote mixed case this can preserve the old backend tag, release the newly-retained backend, and later close the fd against the wrong backend. If `newfd` was a primary KFD fd, the stale primary numeric fd can continue to route to the old backend after `dup2` replaced it.

This is a preexisting `dup2`/`dup3` tracking bug, not a new regression in this push, but the PR’s new backend-tag invariant still needs it fixed. Suggested review wording:

```text
dup2/dup3 atomically close newfd before installing the duplicate, so after a successful call we also need to drop any KFD tracking/reference that belonged to the old newfd before tracking the replacement. Currently we only untrack sysfs/DRM. If newfd was a tracked KFD dup, track_dup() observes the stale map entry and leaves the old backend recorded; in local+remote mixed mode that releases/closes the wrong backend later. Primary KFD targets are worse because remote_kfd_fd_/driver_fd can keep classifying the reused fd number as KFD after dup2 replaced it. Can we add a target-KFD reconciliation path for dup2/dup3, covering both dup and primary targets?
```

The previous concerns I had do look addressed: `find_memfd_for_addr()` now returns a dup-owned fd, source backend is passed into `track_dup`, and I don’t see a remaining `RemoteDriver` lifetime/data-race issue in the shared_ptr/metadata/ioctl path. `git diff --check` is clean; I did not build.

kuhar (commented):
I had a few more comments but they largely overlapped with what @bjacob already reported; only one new finding

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
```diff
@@ -298,76 +347,151 @@ class InterposerContext {
...
+    // torn-down connection.
+    remote_open_refs_.store(0, std::memory_order_release);
     int fd = remote_kfd_fd_.exchange(-1, std::memory_order_acq_rel);
     if (fd >= 0)
       InterposerContext::real.close(fd);
-    std::lock_guard fd_lock(fd_mutex_);
-    kfd_dup_fds_.clear();
+    {
+      std::lock_guard fd_lock(fd_mutex_);
+      kfd_dup_fds_.clear();
```
kuhar: We should preserve `Local`-tagged dup entries when tearing down the remote backend. `kfd_dup_fds_` can now contain entries for both backends, but `teardown_remote_locked()` clears the entire map. If a local KFD dup is still live when the last remote reference closes, this drops its tracking without releasing its local open reference; its later close also cannot release that reference. Could we erase only `Remote` entries here, or leave the table alone if a zero remote refcount guarantees that no `Remote` entries remain?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
```diff
@@ -951,10 +1101,10 @@ RJ_INTERPOSER_EXPORT int ioctl(int fd, unsigned long request, ...) {
     return -1;
   }
 
-  if (auto *remote = InterposerContext::ctx.remote_lookup(fd))
+  if (auto remote = InterposerContext::ctx.remote_lookup(fd))
     return remote->ioctl(request, arg);
   if (InterposerContext::ctx.is_kfd_dup(fd)) {
```
kuhar: I think we should dispatch tracked KFD dups via `kfd_backend_of(fd)` in both the main and late-AMDKFD paths. The current `is_kfd_dup` branch always chooses the remote driver when any remote is live, so a `Local` dup silently changes backend in mixed mode. On current head, a local dup returned version `1.18` before remote startup but `-1/EPIPE` after the daemon was killed. The recorded `DupBackend` should govern ioctl dispatch as well as reference release.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
```diff
@@ -298,76 +347,158 @@ class InterposerContext {
...
+  /// tracked only if a reference was actually taken, and untrack_dup releases that
+  /// same backend regardless of which backend is active later. Passing the source
+  /// backend (rather than rediscovering it from global state) is what prevents
+  /// mis-tagging a remote-fd dup as Local when both backends are simultaneously
+  /// active.
+  void track_dup(int fd, DupBackend backend) {
     if (fd < 0 || is_kfd_primary(fd))
       return;
-    bool newly_tracked = false;
+    if (!retain_backend(backend))
```
kuhar: The backend needs to be retained before the real `dup`/`dup2`/`fcntl(F_DUPFD*)` syscall can be separated from tracking. Today the syscall succeeds first, then `kfd_backend_of()` and `track_dup()` run; another thread can close the last source reference in between, tear down the backend, and leave a valid new fd untracked. The local path has the same gap because `retain_local_open()` can do nothing after process removal while `retain_backend()` still returns true. Please reserve the source backend before the syscall and roll it back on failure, or serialize duplication with last-close so every successful KFD dup owns a backend reference.

kuhar (commented):
Maybe consider adding an interposer-level regression coverage for the fd/backend bookkeeping changed here, especially `dup2`/`dup3` overwriting local and remote primary KFD fds. The existing `KfdIoctlTest` exercises `SimulatedDriver` directly and would not catch this stale-backend bug; focused LD_PRELOAD tests are sufficient.

We should also reconcile primary KFD targets in `dup2`/`dup3` at `interposer.cpp:1148`. `untrack_dup(newfd)` cannot clear `driver_fd()` or `remote_kfd_fd_`, and `track_dup()` returns while `newfd` still matches either primary, so the replacement keeps the old backend identity. On current head, replacing the local primary with `/dev/null` and then closing it left the fd open (`F_GETFD` returned 0 instead of `EBADF`); the remote pipe repro similarly routed `FIONREAD` through the stale KFD backend and left the count at `-1` instead of `3`. Please invalidate and release the overwritten primary before classifying and tracking the replacement, with coverage for both backends.

atgutier:
@bjacob and @kuhar I had to rebase on top of some significant changes so you may want to do another in-depth review, though I think your original comments have been addressed.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
```diff
@@ -338,82 +396,218 @@ class InterposerContext {
...
+    // Remote primary?
+    if (fd == remote_kfd_fd_.load(std::memory_order_acquire)) {
+      // The number is being reused; it no longer names the synthetic remote KFD
+      // fd. Stop classifying it as the primary, then drop the primary's open
+      // reference (which triggers teardown on the last reference). If other refs
+      // keep the connection alive, remote_ stays non-null with remote_kfd_fd_
+      // == -1; that is coherent because routing uses ctx.remote() (not the fd
+      // number) and a later open("/dev/kfd") re-mints a fresh primary fd via
+      // get_or_create_remote() without reconnecting.
+      remote_kfd_fd_.store(-1, std::memory_order_release);
```
kuhar: We should serialize remote-primary invalidation with `open("/dev/kfd")`. `invalidate_overwritten_kfd_fd()` clears `remote_kfd_fd_` outside `remote_mutex_`, while `get_or_create_remote()` retains or reissues under that lock and the open hook reads the fd after the helper returns. A racing open can therefore return `-1` or a descriptor that `dup2` has already replaced. The two-thread stress reproduced this on the current build: one rerun returned `-1` after about 1.3M opens, and an earlier run returned a positive fd whose KFD version ioctl failed with `ENOTTY`. Could the helper return the fd snapshot while still holding `remote_mutex_`, with invalidation using the same lock?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.h
```diff
@@ -117,9 +120,23 @@ class SimulatedKfd : public LinuxKfd {
...
+  /// primary KFD fd number: the number no longer refers to this driver, so stop
+  /// classifying it as the local primary (fd() must no longer match it). The
+  /// caller separately releases the primary's open reference via close(). Only
+  /// clears if @p fd matches the current primary, so a stale call is a no-op.
+  /// @details fd_ is atomic: the interposer reads it lock-free from other threads
+  /// (driver_fd()/kfd_backend_of()/is_kfd_primary()) while this writes it.
+  void invalidate_primary_fd(int fd) override {
+    int expected = fd;
+    if (fd >= 0)
+      fd_.compare_exchange_strong(expected, -1, std::memory_order_acq_rel);
```
kuhar: This would be safer if local-primary invalidation and `SimulatedKfd::open()` used one `process_mutex_` transaction. The CAS at `simulated_kfd.h:137` is outside that lock, while `open()` selects or retains the process and reloads `fd_` in separate steps, so a racing `dup2` can make open return `-1` or the overwritten non-KFD descriptor. The current stress reproduced a positive fd whose KFD version ioctl failed with `ENOTTY` after about 47k opens. Please serialize fd selection, retain, invalidation, and the returned fd together.

kuhar (commented):
It would be good to add daemon-backed regression coverage for the new fd/backend state machine. The five `InterposerDupTest` cases run only in local, single-threaded mode, so the `Remote` path and mixed `Local`/`Remote` routing still have no coverage. Could we cover a remote dup after the primary closes, simultaneous local and remote dups routing to their recorded backends, and `dup2` over the remote primary followed by a working `open("/dev/kfd")`? Including a concurrent overwrite/open variant would also lock in the race fixes. `DaemonTest` already provides the daemon plus `LD_PRELOAD` harness.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
```diff
@@ -200,16 +200,17 @@ int SimulatedKfd::open() {
...
 
-  if (fd_ < 0) {
-    fd_ = static_cast<int>(syscall(SYS_memfd_create, "rocjitsu_kfd", 0));
-    if (fd_ < 0)
+  if (fd_.load(std::memory_order_acquire) < 0) {
+    int new_fd = static_cast<int>(syscall(SYS_memfd_create, "rocjitsu_kfd", 0));
+    if (new_fd < 0)
       return -1;
+    fd_.store(new_fd, std::memory_order_release);
   }
```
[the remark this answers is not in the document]
kuhar: The same load-create-store race is still present at `SimulatedKfd::open_process()` (`simulated_kfd.cpp:292-297`) on current head. Could we put both entry points behind one synchronized or once-only initialization path so concurrent callers cannot create and publish different memfds?

atgutier:
@copilot review again

bjacob (approved):
(Another Codex-generated review).

Thanks, this latest revision addresses the concurrency issues I was worried about in the previous rounds.

In particular, the remote driver lifetime now looks much healthier: publishing a fully-opened `RemoteDriver` through an atomic `shared_ptr`, taking snapshots before `ioctl`/`mmap`/sysfs routing, and keeping handshake metadata immutable after `close()` removes the raw-pointer/UAF and metadata-race concerns I had before. The `find_memfd_for_addr()` change also fixes the fd lifetime problem I was worried about: returning a caller-owned dup under `rpc_mutex_` means a concurrent remote teardown cannot close/reuse the stored memfd between lookup and mmap.

The dup tracking changes also seem to address the stale-fd regression we discussed. Reserving the source backend before `dup`/`dup2`/`dup3`/`F_DUPFD`, recording the backend on the dup, and explicitly invalidating the overwritten `dup2`/`dup3` target avoids both the “track after last-close teardown” race and the previous stale target entry problem. The `SimulatedKfd::fd_` atomic/CAS plus `process_mutex_` serialization also fixes the open-vs-primary-invalidation race.

One remaining caveat: `invalidate_overwritten_kfd_fd()` now relies on `LinuxKfd::invalidate_primary_fd()` to clear a local primary fd after `dup2`/`dup3` overwrites it, but `GuestKfd` does not override that hook. `GuestKfd::open()` returns an app-facing dup while keeping `real_kfd_fd_` hidden, so this is less obvious than the simulated primary fd case, but `real_kfd_fd_` is still a process fd. If some `dup2`/`dup3` target reuses that hidden fd number, `GuestKfd::fd()`/`owns_fd()` can keep treating the reused number as KFD and later route ioctl/mmap/close to whatever now lives there.

I’d classify that as a preexisting issue in `GuestKfd`’s hidden-real-fd model, but it is also an incomplete fix relative to this PR’s new primary-invalidation invariant. A naive `GuestKfd::invalidate_primary_fd()` that returns true may not be correct either, because the interposer calls `d->close()` after a successful local-primary invalidation, while `GuestKfd`’s hidden real fd is not represented in `open_refs_`; only app-facing fds are. So this probably needs either a GuestKfd-specific invalidation path or a slightly different contract for “cleared primary fd” vs “drop primary open reference”.

It would also be useful to add a GuestKfd/DBT-mode regression test for overwriting the hidden real fd if that mode is expected to be robust under arbitrary fd reuse.

kuhar (approved):

amd-eochoalo (approved):

emulation/rocjitsu/tests/CMakeLists.txt
```diff
@@ -614,11 +614,21 @@ if(RJ_HAS_GFX1201_GPU)
             --gtest_filter=GuestKfdMemoryTest.GuestAllocationMmapOffsetIsRejected
         WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
     )
+    add_test(
```
amd-eochoalo, line 617: Should we use rj_add_test_here?

PR #8234, opened 2026-07-08
fix(rocr): Bound section print buffer against attacker size

dayatsin-amd:
## Summary

Addresses Mythos security finding SEC-00386 (ROCM-26859 / SWSPLAT-28893): `AmdHsaCode::PrintRawData` used `alloca(section->size())` with an attacker-controlled ELF section size, which can trigger a stack clash when dumping crafted code objects.

- Replace stack `alloca()` with a heap-allocated `std::vector`
- Reject section data larger than 1 MiB before copying for display

## JIRA ID

JIRA ID : ROCM-26859

## Test plan

- [ ] Build libhsa-runtime64 on Linux
- [ ] Dump a normal code object and confirm section data still prints
- [ ] Dump a crafted object with an oversized section header and confirm it is rejected safely


projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_hsa_code.cpp
```diff
@@ -65,6 +66,15 @@ namespace rocr {
 namespace amd {
 namespace hsa {
 namespace code {
+namespace detail {
+
+constexpr size_t kMaxSectionPrintSize = 1024 * 1024;
+
```

**[Nit]** `IsWithinSectionPrintLimit` is a thin wrapper over a single comparison (`size <= kMaxSectionPrintSize`) used in exactly one place. The helper and `detail` namespace add a layer of indirection that makes readers jump to the definition for what amounts to a trivial bounds check. `section_size <= kMaxSectionPrintSize` reads naturally inline against the named constant. Having `kMaxSectionPrintSize` is the right call; the wrapper function itself is optional overhead.

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_hsa_code.cpp
```diff
@@ -1402,9 +1412,15 @@ namespace code {
...
-      unsigned char *sdata = (unsigned char*)alloca(section->size());
-      section->getData(0, sdata, section->size());
-      PrintRawData(out, sdata, section->size());
+      const size_t section_size = section->size();
+      if (!detail::IsWithinSectionPrintLimit(section_size)) {
+        out << "      (section data too large to display: " << section_size
+            << " bytes)" << std::endl;
+        return;
+      }
+      std::vector<unsigned char> sdata(section_size);
```
cfreeamd, line 1413: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** If libhsa-runtime64 is built with `-fno-exceptions`, a failed `std::vector<unsigned char>` constructor here (OOM with a near-1 MiB allocation) would call `std::terminate()` rather than throwing `std::bad_alloc`. An attacker who can exhaust system memory while supplying crafted ELF sections could trigger this. Is termination acceptable here, or should the code use a `nothrow` allocation path and fall back to the "too large to display" message on failure?

cfreeamd (commented):
`amd_hsa_code.cpp` is the only file changed; all other files look correct. Inline notes below are for @cfreeamd's review.

cfreeamd (approved):
LGTM, with optional change mentioned.

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_hsa_code.cpp
dayatsin-amd: Termination is acceptable here

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_hsa_code.cpp
dayatsin-amd: Done — dropped `IsWithinSectionPrintLimit` and compare `section_size` directly against `detail::kMaxSectionPrintSize` in eed9214a5e.

PR #8235, opened 2026-07-08
fix(rocr): Replace variable-length note _alloca with vector

dayatsin-amd:
## Summary

Addresses Mythos security finding SEC-00600 (ROCM-26860 / SWSPLAT-29300): AMD HSA note builders used `_alloca()` with sizes derived from variable-length note fields, which can overflow the stack when handling crafted or oversized note data.

- Replace `_alloca()` in `AddNoteIsa`, `AddNoteProducer`, and `AddNoteProducerOptions` with bounded `std::vector` buffers
- Reject note buffers larger than 4096 bytes before copying
- Fix `AddNoteIsa` to size its buffer with `sizeof(amdgpu_hsa_note_isa_t)` instead of the producer struct

## JIRA ID

JIRA ID : ROCM-26860

## Test plan

- [ ] Build libhsa-runtime64 on Linux
- [ ] Finalize and load a normal code object; confirm no regression
- [ ] Attempt to build a note with oversized producer/ISA strings and confirm it is rejected safely


projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_hsa_code.cpp
```diff
@@ -52,19 +52,25 @@
...
-#define _alloca alloca
-#endif
-
 namespace rocr {
 namespace amd {
 namespace hsa {
 namespace code {
+namespace detail {
+
+constexpr size_t kMaxAmdNoteBufferSize = 4096;
```
cfreeamd, line 69: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `kMaxAmdNoteBufferSize = 4096` is unexplained. Readers can't tell if 4096 was chosen as a page size, a round number, or based on realistic note size limits. A brief comment (e.g., `// max realistic size of any AMD HSA note field; values above this indicate malformed input`) would clarify intent.

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_hsa_code.cpp
```diff
@@ -52,19 +52,25 @@
...
+
+inline bool IsWithinAmdNoteBufferLimit(size_t size) {
```

**[Nit]** `IsWithinAmdNoteBufferLimit()` has a slightly misleading name — "within limit" in English typically implies `<`, but the implementation uses `<=`. Consider renaming to `FitsInAmdNoteBuffer()` or inlining the check as `if (size > kMaxAmdNoteBufferSize)`.

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_hsa_code.cpp
```diff
@@ -526,8 +532,13 @@ namespace code {
 
     void AmdHsaCode::AddNoteIsa(const std::string& vendor_name, const std::string& architecture_name, uint32_t major, uint32_t minor, uint32_t stepping)
     {
-      size_t size = sizeof(amdgpu_hsa_note_producer_t) + vendor_name.length() + architecture_name.length() + 1;
-      amdgpu_hsa_note_isa_t* desc = (amdgpu_hsa_note_isa_t*) _alloca(size);
+      size_t size = sizeof(amdgpu_hsa_note_isa_t) + vendor_name.length() + architecture_name.length() + 1;
+      if (!detail::IsWithinAmdNoteBufferLimit(size)) {
```

**[Important]** Silent failure when size limit is exceeded (same pattern in `AddNoteProducer` at line 829 and `AddNoteProducerOptions` at line 856).

All three `AddNote*` functions are `void`. When the 4096-byte guard triggers, the function writes to `this->out` and returns silently — the caller has no way to detect the note was not written, and the produced code object will be silently malformed (missing an ISA/producer/options note).

Suggestion: Either (a) change return type to `bool` and return `false` on limit breach so callers can propagate the error, or (b) add `assert(false)` / `abort()` to make the failure loud (these strings come from compiler metadata and should never exceed 4096 bytes in practice). At minimum, add a comment that callers cannot detect this failure path.

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_hsa_code.cpp
```diff
@@ -837,7 +853,13 @@ namespace code {
...
       size_t size = sizeof(amdgpu_hsa_note_producer_options_t) + options.length();
-      amdgpu_hsa_note_producer_options_t *desc = (amdgpu_hsa_note_producer_options_t*) _alloca(size);
+      if (!detail::IsWithinAmdNoteBufferLimit(size)) {
+        out << "Producer options note buffer size exceeds limit: " << size << std::endl;
+        return;
+      }
+      std::vector<char> buffer(size);
+      amdgpu_hsa_note_producer_options_t *desc =
+          reinterpret_cast<amdgpu_hsa_note_producer_options_t*>(buffer.data());
       desc->producer_options_size = options.length();
```
cfreeamd, line 864: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `AddNoteProducerOptions` does not call `memset(desc, 0, size)` before filling the struct, unlike `AddNoteIsa` and `AddNoteProducer` which both zero-initialize first. All struct fields are explicitly set so this isn't a correctness gap today, but the inconsistency is a maintenance hazard if new fields are added.

cfreeamd (commented):
The `sizeof(amdgpu_hsa_note_producer_t)` → `sizeof(amdgpu_hsa_note_isa_t)` fix in `AddNoteIsa` is the most impactful part — corrects a pre-existing 4-byte stack overwrite on every ISA note write. Size calculations and `memcpy` bounds check out. Note: `PrintRawData` at line 1427 still has an unbounded `alloca(section->size())` — pre-existing, outside this PR's scope, but worth a follow-up ticket given the SEC-00600 series context.

dayatsin-amd:
Thanks for the review, @cfreeamd.

Addressed the inline feedback in 86f2450683:
- Documented the 4096-byte note buffer limit
- Inlined the size check (`size > kMaxAmdNoteBufferSize`)
- Changed the three `AddNote*` builders to return `bool` on limit breach
- Added `memset` in `AddNoteProducerOptions`

On `PrintRawData` / unbounded `alloca`: agreed this is pre-existing and out of scope here. I'll track it as a follow-up under the SEC-00600 series (ROCM-26860 covers the note-builder `_alloca` path addressed in this PR).

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_hsa_code.cpp
dayatsin-amd: Addressed in 86f2450683 — added a comment explaining that 4096 is the upper bound for variable-length AMD HSA note payloads built from compiler metadata, and that larger values indicate malformed or attacker-controlled input.

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_hsa_code.cpp
dayatsin-amd: Addressed in 86f2450683 — dropped `IsWithinAmdNoteBufferLimit()` and inlined the check as `if (size > detail::kMaxAmdNoteBufferSize)`.

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_hsa_code.cpp
dayatsin-amd: Addressed in 86f2450683 — changed `AddNoteIsa`, `AddNoteProducer`, and both `AddNoteProducerOptions` overloads from `void` to `bool`. They log to `out` and return `false` when the limit is exceeded, so callers can detect a skipped note.

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_hsa_code.cpp
dayatsin-amd: Addressed in 86f2450683 — `AddNoteProducerOptions` now calls `memset(desc, 0, size)` before filling the struct, matching the other note builders.

cfreeamd (approved):

PR #6962, opened 2026-06-09
feat(rocjitsu): add functional CU dispatch pool

lialan:
## Summary

Add the core bounded host CPU worker pool for functional-mode CU execution.

- Use one pool per SoC, shared by its command processors, so the retained worker count stays bounded across a production topology.
- Let the command processor collect active CUs and execute at most one functional quantum per active CU in each batch.
- Keep CLOCKED mode on the existing event-driven path.
- Preserve deferred queue rescans, completion ordering, functional yields, plugin callback-serialization contracts, and byte-identical threaded execution.
- Honor declarative CU `functional_quantum` values and preserve BARRIER_AND/OR queue ordering even when the packet header barrier bit is clear.

## Concurrency and failure handling

- Queue registration and removal take an exclusive structural lock; dispatch holds the shared side while CU work temporarily releases the queue mutex.
- Worker exceptions finish batch bookkeeping before the first exception is rethrown on the engine thread.
- Pool construction and teardown use stop-aware waits so partial construction and parked workers cannot hang.
- Plugins can request serialized hot-hook callbacks without changing CU dispatch concurrency.

## Validation

- GCC 13 Release build of the standalone core layer.
- Core full `rocjitsu_tests`: 2,807 passed, 1 expected skip.
- Combined stack at #8965: 2,823 passed, 1 expected skip.
- The clear-header-bit barrier regression fails against the previous behavior and passes with this change.
- Repository pre-commit and GitHub release/sanitizer checks passed.

## PR stack

Review and merge in order:

1. #6962 — core functional CU dispatch pool
2. #10074 — public configuration and SoC-wide dispatch controls
3. #8965 — functional hot-path optimizations

The prerequisite concurrency and plugin-policy changes are already merged in #8702, #8703, #8705, and #8706.

ISSUE ID: #6447

PR #8236, opened 2026-07-08
fix(rocjitsu): support sanitizer interposition

lialan:
## Summary

Changes:
* Make the rocjitsu launcher composes `LD_PRELOAD` with existing entries instead of replacing them, deduplicates repeated libraries, and keeps the rocjitsu interposer ordered so HSA calls remain visible.
* The Linux KMD interposer now cooperates with sanitizer preload wrappers and can continue resolving the real HSA entry points behind those wrappers.
* Testings: a focused rocjitsu CLI test covers sanitizer runtime detection and preload-list construction so the launcher behavior is protected against regressions.

Rationale:
* To add sanitizer support, which is useful for debugging rocjitsu
* Sanitizer runtimes and rocjitsu both rely on dynamic interposition. Without explicit preload composition, one layer can hide the symbols the other layer needs, making sanitized rocjitsu runs unreliable.

## Validation

- pre-commit run --files emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp emulation/rocjitsu/tools/rocjitsu/main.cpp emulation/rocjitsu/tests/CMakeLists.txt emulation/rocjitsu/tests/rocjitsu_cli_test.cpp
- cmake -S emulation/rocjitsu -B build/rocjitsu -G Ninja -DCMAKE_BUILD_TYPE=Release
- cmake --build build/rocjitsu --target rocjitsu_tests -j 8
- build/rocjitsu/tests/rocjitsu_tests --gtest_filter='RocjitsuCliTest.*'

ISSUE ID : #6962

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
@@ -88,6 +88,16 @@ static int connect_to_daemon() {
 
 namespace {
 
+#if defined(__clang__) && defined(__has_attribute) &&                                              \
+    __has_attribute(disable_sanitizer_instrumentation)
+#define RJ_NO_SANITIZE_INTERPOSER __attribute__((disable_sanitizer_instrumentation))
+#elif defined(__GNUC__)
+#define RJ_NO_SANITIZE_INTERPOSER                                                                  \
+  __attribute__((no_sanitize_address, no_sanitize_thread, no_sanitize_undefined))
+#else
+#define RJ_NO_SANITIZE_INTERPOSER
+#endif
+
 void rj_sigsegv_handler(int, siginfo_t *, void *) {
   signal(SIGSEGV, SIG_DFL);
   raise(SIGSEGV);
@@ -126,7 +136,7 @@ public:
   ssize_t (*readlink_fn)(const char *, char *, size_t) = nullptr;
   pid_t (*fork)() = nullptr;
 
-  bool ready() const { return initialized_; }
+  RJ_NO_SANITIZE_INTERPOSER bool ready() const { return initialized_; }
 
   void resolve() {
     auto *handle = RTLD_NEXT;
@@ -546,7 +556,52 @@ alignas(16) uint8_t InterposerContext::storage_[sizeof(InterposerContext)];
 InterposerContext &InterposerContext::ctx =
     *reinterpret_cast<InterposerContext *>(InterposerContext::storage_);
 
+static void *(*real_dlsym_fn)(void *, const char *) = nullptr;
+
+RJ_NO_SANITIZE_INTERPOSER __attribute__((constructor(101))) static void resolve_real_dlsym() {
+  real_dlsym_fn =
+      reinterpret_cast<decltype(real_dlsym_fn)>(dlvsym(RTLD_NEXT, "dlsym", "GLIBC_2.34"));
+  if (!real_dlsym_fn)
+    real_dlsym_fn =
+        reinterpret_cast<decltype(real_dlsym_fn)>(dlvsym(RTLD_NEXT, "dlsym", "GLIBC_2.2.5"));
+  if (!real_dlsym_fn) {
+    fprintf(stderr, "rocjitsu: failed to resolve real dlsym\n");
+    abort();
+  }
+}
+
 __attribute__((constructor)) static void init_interposer() { InterposerContext::init(); }
+
+constexpr int DRM_NODE_PRIMARY = 0;
+constexpr int DRM_NODE_RENDER = 2;
+constexpr int DRM_BUS_PCI = 0;
+
+struct drmPciBusInfo {
+  uint16_t domain;
+  uint8_t bus;
+  uint8_t dev;
+  uint8_t func;
+};
+
+struct drmPciDeviceInfo {
+  uint16_t vendor_id;
+  uint16_t device_id;
+  uint16_t subvendor_id;
+  uint16_t subdevice_id;
+  uint8_t revision_id;
+};
+
+struct drmDevice {
+  char **nodes;
+  int available_nodes;
+  int bustype;
+  union {
+    drmPciBusInfo *pci;
+  } businfo;
+  union {
+    drmPciDeviceInfo *pci;
+  } deviceinfo;
+};
 
 } // namespace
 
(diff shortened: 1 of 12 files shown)


amd-eochoalo:
Do you know if any of these tests:

```
          ASAN_UBSAN_EXCLUDED_TESTS=(
            "HsaTest\.InitSucceeded"
            "HsaTest\.GpuAgentFound"
            "HipMemcpyTest\.RoundTripFloat"
            "HipMemcpyTest\.RoundTripInt"
            "HipMemcpyTest\.DeviceToDevice"
            "HipVectorAddTest\.CorrectResult"
            "DaemonTest\.HipVectorAdd"
            "DaemonTest\.HipMemcpyRoundTripFloat"
            "DaemonTest\.HipMemcpyRoundTripInt"
            "DaemonTest\.HipMemcpyDeviceToDevice"
            "DaemonTest\.TwoIndependentClients"
            "BfeRegressionTest\.ThreadIdxY"
            "LoggingTest\.dispatch_logged"
            "ProfiledPluginTest\.hook_profile_output"
            "CvtNarrow\.AllTests"
            "RaceTest\..*"
          )
  ```
  
  would now pass on CI?

**Comment by bjacob** [comment]
@lialan , could you elaborate on how this compares to HSA_TOOLS_LIB interposition (which I've been using in https://github.com/ROCm/rocm-systems/tree/users/bjacob/consan ) ?

File: emulation/rocjitsu/tests/rocjitsu_cli_test.cpp [hunk 1]
```diff
@@ -0,0 +1,94 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: MIT
+
+#define main rocjitsu_cli_main_for_test
+#include "../tools/rocjitsu/main.cpp"
```
kuhar, line 5: We shouldn't be including .cpp files

kuhar (changes requested):

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
```diff
@@ -1661,6 +1716,155 @@ RJ_INTERPOSER_EXPORT int __lxstat64(int ver, const char *path, struct stat64 *bu
...
+  if (!devices || max_devices <= 0)
+    return 0;
+  auto *gpu = interposer_gpu_info();
+  if (!gpu)
+    return 0;
+  devices[0] = alloc_drm_device(*gpu, 0);
+  return devices[0] ? 1 : 0;
+}
+
+RJ_INTERPOSER_EXPORT int drmGetDevices2(uint32_t /*flags*/, drmDevice **devices, int max_devices) {
```
Groverkss, line 1872: These changes are undoing what https://github.com/ROCm/rocm-systems/pull/7649 did, I don't think this is the right direction.

Groverkss (changes requested):

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
lialan: Ok, I will rethink this PR, I missed it.

atgutier (commented):
The agent had some comments:

 MAJOR

  1. Externally-supplied sanitizer lands after the interposer, breaking ASan-must-be-first —
  main.cpp:396 (build_ld_preload). The composed order is loaded_sanitizer_runtimes() → interposer →
  existing LD_PRELOAD appended verbatim. But loaded_sanitizer_runtimes() only inspects launcher-loaded
   libs + the compile-time RJ_*_RUNTIME_LIBRARY macros — it never scans the incoming LD_PRELOAD to
  hoist a sanitizer entry. So LD_PRELOAD=.../libclang_rt.asan.so rocjitsu -- ./app with a
  non-sanitizer launcher build puts ASan after the interposer → broken ASan init. The intended
  -DRJ_ENABLE_ASAN=ON workflow is correct; the gap is the external-sanitizer case. Fix: re-partition
  the existing entries so any is_sanitizer_runtime(basename) is moved ahead of the interposer (stable
  order), which also lets dedup collapse it against the configured one.
  2. Global dlsym interposition risks perturbing the sanitizer's own early init, and its real-fn cache
   is unsynchronized — interposer.cpp:1811. The new exported dlsym becomes process-wide; ASan/TSan
  call dlsym heavily during init, which (preloaded-first) runs before rocjitsu's constructors. The
  self-heal guard (if (!real_dlsym_fn) resolve_real_dlsym()) writes a plain non-atomic real_dlsym_fn
  and can be entered concurrently/re-entrantly, and some sanitizer runtimes special-case dlsym
  precisely because interposing it is fragile. The reviewer flagged this as needing an actual
  ASan-first runtime smoke test — static review can't fully settle init-order under a preloaded-first
  sanitizer. Fix: make real_dlsym_fn atomic / once-guarded; make "null symbol or !ready() → cheapest
  plain-forward" the very first branch; verify with a real ASan run and, if it perturbs init, restrict
   the interposition (e.g. only RTLD_NEXT/RTLD_DEFAULT handles, or gate behind ready()).

  MINOR

  - New drmDevice/drmPciBusInfo/drmPciDeviceInfo structs have no layout-guard test —
  interposer.cpp:614. They must be ABI-identical to libdrm's xf86drm.h (amdsmi reads their fields and
  calls the real drmFreeDevice on them). The reviewer verified the current layout matches exactly, but
   unlike drm_amdgpu_info_device (which has drm_info_layout_test.cpp) there's no static_assert/offset
  test to catch future libdrm drift. Add one.
  - real_dlsym_fn non-atomic (same mechanism as MAJOR #2, called out separately) — TSan (which this PR
   enables) will flag it; trivially fixable with an atomic or function-local-static accessor.
  - Dedup is exact-string only — main.cpp:333; the same lib named two ways isn't collapsed (harmless
  since the loader dedups by soname, but suboptimal). Use weakly_canonical, not basename (basename
  would wrongly drop distinct same-named libs).
  - execution_plugin_group.h:145 deliberately leaks the empty group to dodge the
  static-destruction-order fiasco with the winding-down engine thread — a legitimate teardown fix, but
   LSan will report it in a sanitizer PR. Add a suppression/comment, or confirm the interposer's new
  shutdown() engine-thread join ordering makes the leak unnecessary.

  NIT

  split_ld_preload can't represent space-containing paths (mirrors the loader's own ambiguity —
  document it); the new preload helpers (is_sanitizer_runtime/split_ld_preload/build_ld_preload) lack
  Doxygen for the subtle ASan-first ordering contract; test coverage gaps — no test for the
  external-sanitizer-first case (MAJOR #1 — which would currently fail, the point of adding it), the
  no-existing-preload path, or idempotent re-launch; missing #include <cstdlib> for the DRM alloc code
   (transitively satisfied).

PR #2669, opened 2026-01-19
rocr: Force XNACK-off for GFX 9.0.12 in ROCr runtime

HongleiHuang-amd:
## Motivation

Fix gfx90c issue

## Technical Details

Renoir/Green Sardine APUs (IP_VERSION 9.0.12) have retry enabled at the kernel layer (noretry=0), which is needed for per-queue-reset and page fault handling. However, official ROCm builds for these APUs are compiled with XNACK-off, causing shader loading failures when retry is default enabled in kernel. So for gfx 9.0.12, need to set it as xnack off mode.

This patch adds a userspace override in ROCr's KfdDriver::BindXnackMode() to force XNACK_DISABLE for IP_VERSION(9,0,12) only during runtime initialization. The GetXnackModeOverride() function detects Renoir/Green Sardine APUs by checking the EngineId and overrides the xnack mode regardless of user preferences or environment variables.

## JIRA ID

SWDEV-576070

## Test Plan

rocr
hipcatch

## Test Result

all passed

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #8237, opened 2026-07-08
fix(rocr): Disable XNACK for gfx9 client APUs (gfx902/gfx909/gfx90c)

HongleiHuang-amd:
A fix for xnack-off mode for gfx9 client APUs. Same fix like the previous one in: https://github.com/ROCm/rocm-systems/pull/2669
Alex request all the gfx9 client APUs need to be change to xanck-off mode, so that is why this PR here.

Some gfx9 client APUs have retry enabled at the kernel layer (noretry=0), which is needed for per queue reset and page fault handling. However, official ROCm builds for these APUs are compiled with XNACK-off, so when retry is default enabled in the kernel the ISA advertised by ROCr (xnack+) no longer matches the shipped code objects and shader loading fails. So for these gfx9 client APUs we need to force XNACK-off mode.

Change XNACK as unsupported in the ISA registry for the gfx9 *client* APUs so ROCr never advertises or selects an xnack-aware ISA for them. With xnack unsupported, GpuAgent skips the xnack enable/disable branch and always reports the neutral (xnack-off) target id, and comgr/HIP code-object selection likewise stops emitting/matching xnack+ variants. This makes xnack-off the effective default for these parts.

Affected APU changed into a single entry with xnack = unsupported, only gfx9 client APUs are changed:

  gfx902  (9,0,2)  Raven / Picasso
  gfx909  (9,0,9)  Raven2 (Dali / Pollock)
  gfx90c  (9,0,12) Renoir / Green Sardine / Cezanne / Barcelo / Mendocino

## Motivation

A fix for xnack-off mode for gfx9 client APUs. Same fix like the previous one in: https://github.com/ROCm/rocm-systems/pull/2669
Alex request all the gfx9 client APUs need to be change to xanck-off mode, so that is why this PR here.

## Technical Details

Change XNACK as unsupported in the ISA registry for the gfx9 *client* APUs so ROCr never advertises or selects an xnack-aware ISA for them. With xnack unsupported, GpuAgent skips the xnack enable/disable branch and always reports the neutral (xnack-off) target id, and comgr/HIP code-object selection likewise stops emitting/matching xnack+ variants. This makes xnack-off the effective default for these parts.

## JIRA ID

JIRA ID : SWDEV-576070

https://amd.atlassian.net/browse/SWDEV-576070?focusedCommentId=22886084&sourceType=mention

## Test Plan

kfd test passed, rocr passed, hip passed in local test.
## Test Result
kfd test passed, rocr passed, hip passed in local test.


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/isa.cpp
```diff
@@ -364,9 +364,7 @@ const IsaRegistry::IsaMap& IsaRegistry::GetSupportedIsas() {
   ISAREG_ENTRY_GEN("gfx900",                 9, 0, 0,  unsupported, any,         64, "gfx9-generic")
   ISAREG_ENTRY_GEN("gfx900:xnack-",          9, 0, 0,  unsupported, disabled,    64, "gfx9-generic:xnack-")
   ISAREG_ENTRY_GEN("gfx900:xnack+",          9, 0, 0,  unsupported, enabled,     64, "gfx9-generic:xnack+")
-  ISAREG_ENTRY_GEN("gfx902",                 9, 0, 2,  unsupported, any,         64, "gfx9-generic")
-  ISAREG_ENTRY_GEN("gfx902:xnack-",          9, 0, 2,  unsupported, disabled,    64, "gfx9-generic:xnack-")
-  ISAREG_ENTRY_GEN("gfx902:xnack+",          9, 0, 2,  unsupported, enabled,     64, "gfx9-generic:xnack+")
+  ISAREG_ENTRY_GEN("gfx902",                 9, 0, 2,  unsupported, unsupported, 64, "gfx9-generic")
```
cfreeamd, line 367: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** Removing `:xnack-` entries breaks code objects compiled with explicit XNACK-disabled targets

The `:xnack-` variant means "compiled knowing XNACK is off" — the correct compilation target for hardware that doesn't support XNACK. Code objects built with `clang++ -mcpu=gfx902 -mxnack-` (or explicitly targeting `gfx902:xnack-`) will now fail `IsaRegistry::GetSupportedIsas()` lookup since no matching entry exists. This is a silent backward-compatibility break.

Note also that `gfx904` (also `unsupported` XNACK, unchanged here) still retains its `:xnack-`/`:xnack+` entries — so this PR applies the cleanup inconsistently across client APUs.

Suggested fix: retain at minimum the `:xnack-` entries with `unsupported` XNACK for gfx902/909, or explicitly document that code objects targeting these variants are no longer supported.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/isa.cpp
```diff
@@ -388,9 +386,7 @@ const IsaRegistry::IsaMap& IsaRegistry::GetSupportedIsas() {
   ISAREG_ENTRY_GEN("gfx908:sramecc-:xnack+", 9, 0, 8,  disabled,    enabled,     64, "")
   ISAREG_ENTRY_GEN("gfx908:sramecc+:xnack-", 9, 0, 8,  enabled,     disabled,    64, "")
   ISAREG_ENTRY_GEN("gfx908:sramecc+:xnack+", 9, 0, 8,  enabled,     enabled,     64, "")
-  ISAREG_ENTRY_GEN("gfx909",                 9, 0, 9,  unsupported, any,         64, "gfx9-generic")
-  ISAREG_ENTRY_GEN("gfx909:xnack-",          9, 0, 9,  unsupported, disabled,    64, "gfx9-generic:xnack-")
-  ISAREG_ENTRY_GEN("gfx909:xnack+",          9, 0, 9,  unsupported, enabled,     64, "gfx9-generic:xnack+")
+  ISAREG_ENTRY_GEN("gfx909",                 9, 0, 9,  unsupported, unsupported, 64, "gfx9-generic")
```
cfreeamd, line 390: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** PR title says "gfx902/gfx909/gfx90c" but gfx90c is not modified in this diff. Was gfx90c already absent from the registry, was it intentionally omitted, or was it missed?

cfreeamd (commented):
One correctness concern before merge.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/isa.cpp
HongleiHuang-amd: Thanks for the review! Already fixed by the comments in latest version

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/isa.cpp
HongleiHuang-amd: Fixed

GreenShadows:
Why does AMD's solution involve breaking things instead of fixing them? 

The pull request below is a step in the right direction, unlike this one. 

https://github.com/ROCm/rocm-libraries/pull/9469

cfreeamd (commented):

The prior **[Important]** concern (removing `:xnack-` entries would break loading of code objects compiled for explicit XNACK-disabled targets) is resolved exactly as suggested: `gfx902:xnack-` and `gfx909:xnack-` are retained with the XNACK field set to `unsupported`, and only the `:xnack+` variants were dropped. This keeps named-lookup loads working (`GetIsaByName` → `GetSupportedIsas().find(full_name)`) and, since the field is now `unsupported`, `Isa::IsCompatible` no longer rejects them. Dropping `:xnack+` is correct because XNACK-enabled code cannot run on parts lacking XNACK support.

The prior **[Question]** about gfx90c is also resolved — a `gfx90c:xnack-` entry was added and the title now matches the diff.

Remaining non-blocking note: gfx904 (client APU, `unsupported` sramecc) still keeps `disabled`/`enabled` `:xnack-`/`:xnack+` entries, and gfx900 keeps its xnack variants, so the client-APU cleanup is applied inconsistently across gfx90x parts — a one-line rationale would help, but it is not a correctness issue. All Important+ items addressed.

cfreeamd (approved):

HongleiHuang-amd:
> Why does AMD's solution involve breaking things instead of fixing them?
> 
> The pull request below is a step in the right direction, unlike this one.
> 
> [ROCm/rocm-libraries#9469](https://github.com/ROCm/rocm-libraries/pull/9469)


Here is the original words from KFD architect:

XNACK (retry enabled) is only supported by ROCm on MI200 and later MI GPUs. On GFX9, shader code needs to be compiled differently for XNACK on or off. Official ROCm builds for non-MI GPUs are built for XNACK off. Those shaders will not load if XNACK is enabled. Therefore, enabling retry by default on non-MI GPUs or APUs is not currently useful, at least from a ROCm perspective.
As a workaround, you could try running your ROCm application with HSA_XNACK=0. That should tell ROCr to switch XNACK off on a per-process basis. It won't change the SH_MEM_CONFIG.RETRY_DISABLE on this GPU, but it will treat all page faults as fatal and allow XNACK-disabled code to run.

And the request of disable gfx9 APU xnack mode is from the Alex, not me, I am just trying close a JIRA assigned to me:
https://amd.atlassian.net/browse/SWDEV-576070?focusedCommentId=22886084&sourceType=mention

Please see the detail in this JIRA, it is very clear that the amdgpu architect requested that all the GFX9 APU need disable the xnack, so if you have some question about it please contact : Alex Deucher and Felix Kuehling. And this change only affects the APUs, not MI series, xnack on/ off  has  very close performance in thoses APUs.

Regards,
Honglei

PR #8242, opened 2026-07-08
feat(rocprofiler-compute): correlate ISA with kernel names

abchoudh-amd:
## Motivation

Record the SDK kernel-symbol registration mapping so PC sampling code-object ISA output can be correlated to kernel ids and display names without lossy name matching.

## Technical Details

- Add `kernel_t` and a `kernels` array to code object JSON output.
- Record `code_object_id -> kernel_t` mappings in the PC sampling collector and emit kernels during finalize.
- Kernel filtering and PC sampling use separate name forms: filter uses `.kd`-stripped demangled/truncated names, while PC sampling emits `.kd`-stripped truncated raw symbols.
- Thread `kernel_t` through `pc_sampling_feature_t`, `pc_sampling_collector_t`, and `code_object_writer_t` to match the symbol/instruction struct interfaces.
- Add writer, collector, feature, and callback unit coverage for kernel serialization, name-form handling, and forwarding.

JIRA ID : AIPROFCOMP-701

## Test Plan

- [x] Run ctests

## Test Result

- [x] Ctests pass

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-compute/src/lib/pc_sampling_collector/code_object_writer.cpp
+++ b/projects/rocprofiler-compute/src/lib/pc_sampling_collector/code_object_writer.cpp
@@ -32,13 +32,13 @@ void code_object_writer_json_t::end_code_obj()
     m_symbols = nlohmann::json::array();
 }
 
-void code_object_writer_json_t::write_kernel(uint64_t kernel_id, const std::string& name)
+void code_object_writer_json_t::write_kernel(const kernel_t& kernel)
 {
     Expects(m_code_object_closure_count != 0);
 
     m_kernels.push_back(nlohmann::json::object({
-        {"kernel_id", kernel_id},
-        {"name", name},
+        {"kernel_id", kernel.kernel_id},
+        {"name", kernel.name},
     }));
 }
 
--- a/projects/rocprofiler-compute/src/lib/pc_sampling_collector/code_object_writer.cpp
+++ b/projects/rocprofiler-compute/src/lib/pc_sampling_collector/code_object_writer.cpp
@@ -25,7 +25,20 @@ void code_object_writer_json_t::end_code_obj()
 
     m_code_objects.push_back(nlohmann::json::object({
         {"id", m_current_obj_id},
+        {"kernels", std::move(m_kernels)},
         {"symbols", std::move(m_symbols)},
+    }));
+    m_kernels = nlohmann::json::array();
+    m_symbols = nlohmann::json::array();
+}
+
+void code_object_writer_json_t::write_kernel(uint64_t kernel_id, const std::string& name)
+{
+    Expects(m_code_object_closure_count != 0);
+
+    m_kernels.push_back(nlohmann::json::object({
+        {"kernel_id", kernel_id},
+        {"name", name},
     }));
 }
 
@@ -50,6 +63,7 @@ void code_object_writer_json_t::end_symbol()
         {"size", m_current_symbol.size},
         {"instructions", std::move(m_instructions)},
     }));
+    m_instructions = nlohmann::json::array();
 }
 
 void code_object_writer_json_t::write_instruction(const instruction_t& inst)
--- a/projects/rocprofiler-compute/src/lib/pc_sampling_collector/pc_sampling_collector.h
+++ b/projects/rocprofiler-compute/src/lib/pc_sampling_collector/pc_sampling_collector.h
@@ -7,10 +7,14 @@
 
 #include <rocprofiler-sdk/rocprofiler.h>
 
+#include <cstddef>
 #include <filesystem>
+#include <map>
 #include <memory>
 #include <set>
+#include <string>
 #include <string_view>
+#include <vector>
 
 namespace rocprofiler_compute_tool
 {
@@ -30,8 +34,11 @@ public:
 
     virtual ~pc_sampling_collector_t() = default;
     virtual void on_code_object_load(const rocprofiler_callback_tracing_code_object_load_data_t& info) = 0;
-    virtual void                                   finalize(code_object_writer_t& writer) = 0;
-    virtual const std::set<std::filesystem::path>& get_source_paths() const               = 0;
+    virtual void on_kernel_symbol_register(size_t             code_object_id,
+                                           uint64_t           kernel_id,
+                                           const std::string& name)         = 0;
+    virtual void finalize(code_object_writer_t& writer)                     = 0;
+    virtual const std::set<std::filesystem::path>& get_source_paths() const = 0;
 };
 
 class pc_sampling_collector_impl_t : public pc_sampling_collector_t
@@ -40,6 +47,7 @@ public:
     pc_sampling_collector_impl_t(code_object_translator_t::ptr translator);
     pc_sampling_collector_impl_t(code_object_translator_t::ptr translator, sdk_wrapper_t::ptr sdk_wrapper);
     void on_code_object_load(const rocprofiler_callback_tracing_code_object_load_data_t& info) override;
+    void on_kernel_symbol_register(size_t code_object_id, uint64_t kernel_id, const std::string& name) override;
     void                                   finalize(code_object_writer_t& writer) override;
     const std::set<std::filesystem::path>& get_source_paths() const override;
 
@@ -50,8 +58,9 @@ private:
     void collect_source_paths_from_comment(std::string_view                 comment,
                                            std::set<std::filesystem::path>& source_paths) const;
 
-    code_object_translator_t::ptr   m_translator;
-    sdk_wrapper_t::ptr              m_sdk_wrapper;
-    std::set<std::filesystem::path> m_source_paths;
+    code_object_translator_t::ptr           m_translator;
+    sdk_wrapper_t::ptr                      m_sdk_wrapper;
+    std::set<std::filesystem::path>         m_source_paths;
+    std::map<size_t, std::vector<kernel_t>> m_kernels_by_obj;
 };
 }  // namespace rocprofiler_compute_tool
--- a/projects/rocprofiler-compute/src/lib/pc_sampling_collector/pc_sampling_collector.h
+++ b/projects/rocprofiler-compute/src/lib/pc_sampling_collector/pc_sampling_collector.h
@@ -34,11 +34,9 @@ public:
 
     virtual ~pc_sampling_collector_t() = default;
     virtual void on_code_object_load(const rocprofiler_callback_tracing_code_object_load_data_t& info) = 0;
-    virtual void on_kernel_symbol_register(size_t             code_object_id,
-                                           uint64_t           kernel_id,
-                                           const std::string& name)         = 0;
-    virtual void finalize(code_object_writer_t& writer)                     = 0;
-    virtual const std::set<std::filesystem::path>& get_source_paths() const = 0;
+    virtual void on_kernel_symbol_register(size_t code_object_id, const kernel_t& kernel) = 0;
+    virtual void finalize(code_object_writer_t& writer)                                   = 0;
+    virtual const std::set<std::filesystem::path>& get_source_paths() const               = 0;
 };
 
 class pc_sampling_collector_impl_t : public pc_sampling_collector_t
@@ -47,8 +45,8 @@ public:
     pc_sampling_collector_impl_t(code_object_translator_t::ptr translator);
     pc_sampling_collector_impl_t(code_object_translator_t::ptr translator, sdk_wrapper_t::ptr sdk_wrapper);
     void on_code_object_load(const rocprofiler_callback_tracing_code_object_load_data_t& info) override;
-    void on_kernel_symbol_register(size_t code_object_id, uint64_t kernel_id, const std::string& name) override;
-    void                                   finalize(code_object_writer_t& writer) override;
+    void on_kernel_symbol_register(size_t code_object_id, const kernel_t& kernel) override;
+    void finalize(code_object_writer_t& writer) override;
     const std::set<std::filesystem::path>& get_source_paths() const override;
 
 private:
--- a/projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_code_object_writer.cpp
+++ b/projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_code_object_writer.cpp
@@ -112,6 +112,18 @@ TEST_F(test_code_object_writer_t, ProvidedWriteInstructionAfterEndCodeObj_Throws
                  std::runtime_error);
 }
 
+TEST_F(test_code_object_writer_t, ProvidedWriteKernelWithoutAnyScope_Throws)
+{
+    EXPECT_THROW(m_writer.write_kernel(1, "kernel0"), std::runtime_error);
+}
+
+TEST_F(test_code_object_writer_t, ProvidedWriteKernelAfterEndCodeObj_Throws)
+{
+    m_writer.start_code_obj(0);
+    m_writer.end_code_obj();
+    EXPECT_THROW(m_writer.write_kernel(1, "kernel0"), std::runtime_error);
+}
+
 TEST_F(test_code_object_writer_t, ProvidedCodeObjDesc_SerializesIt)
 {
     constexpr uint32_t id0 = 10;
@@ -135,6 +147,76 @@ TEST_F(test_code_object_writer_t, ProvidedNoSymbols_SerializesEmptySymbolsArray)
     ASSERT_TRUE(json["code_objects"][0].contains("symbols"));
     EXPECT_TRUE(json["code_objects"][0]["symbols"].is_array());
     EXPECT_EQ(json["code_objects"][0]["symbols"].size(), 0);
+}
+
+TEST_F(test_code_object_writer_t, ProvidedNoKernels_SerializesEmptyKernelsArray)
+{
+    m_writer.start_code_obj(1);
+    m_writer.end_code_obj();
+
+    const auto json = nlohmann::json::parse(m_writer.get_result());
+    ASSERT_TRUE(json["code_objects"][0].contains("kernels"));
+    EXPECT_TRUE(json["code_objects"][0]["kernels"].is_array());
+    EXPECT_EQ(json["code_objects"][0]["kernels"].size(), 0);
+}
+
+TEST_F(test_code_object_writer_t, ProvidedKernel_SerializesItInsideCodeObj)
+{
+    constexpr uint64_t kernel_id = 42;
+    const std::string  name      = "kernel0";
+
+    m_writer.start_code_obj(1);
+    m_writer.write_kernel(kernel_id, name);
+    m_writer.end_code_obj();
+
+    const auto  json    = nlohmann::json::parse(m_writer.get_result());
+    const auto& kernels = json["code_objects"][0]["kernels"];
+    ASSERT_EQ(kernels.size(), 1);
+    EXPECT_EQ(kernels[0]["kernel_id"].get<uint64_t>(), kernel_id);
+    EXPECT_EQ(kernels[0]["name"].get<std::string>(), name);
+}
+
+TEST_F(test_code_object_writer_t, ProvidedMultipleKernels_SerializesAllInOrder)
+{
+    m_writer.start_code_obj(1);
+    m_writer.write_kernel(42, "kernel0");
+    m_writer.write_kernel(43, "kernel1");
+    m_writer.end_code_obj();
+
+    const auto  json    = nlohmann::json::parse(m_writer.get_result());
+    const auto& kernels = json["code_objects"][0]["kernels"];
+    ASSERT_EQ(kernels.size(), 2);
+    EXPECT_EQ(kernels[0]["kernel_id"].get<uint64_t>(), 42);
+    EXPECT_EQ(kernels[0]["name"].get<std::string>(), "kernel0");
+    EXPECT_EQ(kernels[1]["kernel_id"].get<uint64_t>(), 43);
+    EXPECT_EQ(kernels[1]["name"].get<std::string>(), "kernel1");
+}
+
+TEST_F(test_code_object_writer_t, ProvidedKernelsInDifferentCodeObjs_SerializesUnderRespectiveOwners)
+{
+    m_writer.start_code_obj(100);
+    m_writer.write_kernel(42, "kernel0");
+    m_writer.write_kernel(43, "kernel1");
+    m_writer.end_code_obj();
+
+    m_writer.start_code_obj(200);
+    m_writer.write_kernel(44, "kernel2");
+    m_writer.end_code_obj();
+
+    const auto json = nlohmann::json::parse(m_writer.get_result());
+    ASSERT_EQ(json["code_objects"].size(), 2u);
+
+    EXPECT_EQ(json["code_objects"][0]["id"], 100);
+    ASSERT_EQ(json["code_objects"][0]["kernels"].size(), 2);
+    EXPECT_EQ(json["code_objects"][0]["kernels"][0]["kernel_id"].get<uint64_t>(), 42);
+    EXPECT_EQ(json["code_objects"][0]["kernels"][0]["name"].get<std::string>(), "kernel0");
+    EXPECT_EQ(json["code_objects"][0]["kernels"][1]["kernel_id"].get<uint64_t>(), 43);
+    EXPECT_EQ(json["code_objects"][0]["kernels"][1]["name"].get<std::string>(), "kernel1");
+
+    EXPECT_EQ(json["code_objects"][1]["id"], 200);
+    ASSERT_EQ(json["code_objects"][1]["kernels"].size(), 1);
+    EXPECT_EQ(json["code_objects"][1]["kernels"][0]["kernel_id"].get<uint64_t>(), 44);
+    EXPECT_EQ(json["code_objects"][1]["kernels"][0]["name"].get<std::string>(), "kernel2");
 }
 
 TEST_F(test_code_object_writer_t, ProvidedSymbol_SerializesItInsideCodeObj)
@@ -219,6 +301,34 @@ TEST_F(test_code_object_writer_t, ProvidedInstruction_SerializesItInsideSymbol)
 
     const auto  json         = nlohmann::json::parse(m_writer.get_result());
     const auto& instructions = json["code_objects"][0]["symbols"][0]["instructions"];
+    ASSERT_EQ(instructions.size(), 1);
+    EXPECT_EQ(instructions[0]["name"], m_inst0.name);
+    EXPECT_EQ(instructions[0]["comment"], m_inst0.comment);
+    EXPECT_EQ(instructions[0]["virtual_address"], m_inst0.virtual_address);
+    EXPECT_EQ(instructions[0]["code_obj_offset"], m_inst0.code_obj_offset);
+    EXPECT_EQ(instructions[0]["size"], m_inst0.size);
+}
+
+TEST_F(test_code_object_writer_t, ProvidedKernelAndInstruction_PreservesInstructionOutput)
+{
+    m_writer.start_code_obj(1);
+    m_writer.write_kernel(42, "kernel0");
+    m_writer.start_symbol(m_symbol0);
+    m_writer.write_instruction(m_inst0);
+    m_writer.end_symbol();
+    m_writer.end_code_obj();
+
+    const auto  json    = nlohmann::json::parse(m_writer.get_result());
+    const auto& obj     = json["code_objects"][0];
+    const auto& symbols = obj["symbols"];
+    ASSERT_EQ(obj["kernels"].size(), 1);
+    ASSERT_EQ(symbols.size(), 1);
+    EXPECT_EQ(symbols[0]["name"], m_symbol0.name);
+    EXPECT_EQ(symbols[0]["code_object_offset"], m_symbol0.code_object_offset);
+    EXPECT_EQ(symbols[0]["virtual_address"], m_symbol0.virtual_address);
+    EXPECT_EQ(symbols[0]["size"], m_symbol0.size);
+
+    const auto& instructions = symbols[0]["instructions"];
     ASSERT_EQ(instructions.size(), 1);
     EXPECT_EQ(instructions[0]["name"], m_inst0.name);
     EXPECT_EQ(instructions[0]["comment"], m_inst0.comment);
--- a/projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_code_object_writer.cpp
+++ b/projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_code_object_writer.cpp
@@ -142,14 +142,16 @@ TEST_F(test_code_object_writer_t, ProvidedWriteInstructionAfterEndCodeObj_Throws
 
 TEST_F(test_code_object_writer_t, ProvidedWriteKernelWithoutAnyScope_Throws)
 {
-    expect_runtime_throw([&]() { m_writer.write_kernel(1, "kernel0"); });
+    expect_runtime_throw(
+        [&]() { m_writer.write_kernel(rocprofiler_compute_tool::kernel_t{1, "kernel0"}); });
 }
 
 TEST_F(test_code_object_writer_t, ProvidedWriteKernelAfterEndCodeObj_Throws)
 {
     m_writer.start_code_obj(0);
     m_writer.end_code_obj();
-    expect_runtime_throw([&]() { m_writer.write_kernel(1, "kernel0"); });
+    expect_runtime_throw(
+        [&]() { m_writer.write_kernel(rocprofiler_compute_tool::kernel_t{1, "kernel0"}); });
 }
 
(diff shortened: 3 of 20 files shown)


projects/rocprofiler-compute/src/lib/pc_sampling_collector/code_object_writer.cpp
```diff
@@ -25,8 +25,21 @@ void code_object_writer_json_t::end_code_obj()
 
     m_code_objects.push_back(nlohmann::json::object({
         {"id", m_current_obj_id},
+        {"kernels", std::move(m_kernels)},
         {"symbols", std::move(m_symbols)},
     }));
+    m_kernels = nlohmann::json::array();
```
svolkov-amd, line 31: Why are these two lines which assign `m_kernels` and `m_symbols` to empty array needed? Is it because `std::move` doesn't actually move the data? If so, is there a unit test added which reproduces such issue?

projects/rocprofiler-compute/src/lib/pc_sampling_collector/code_object_writer.cpp
```diff
@@ -50,6 +63,7 @@ void code_object_writer_json_t::end_symbol()
         {"size", m_current_symbol.size},
         {"instructions", std::move(m_instructions)},
     }));
+    m_instructions = nlohmann::json::array();
```
svolkov-amd, line 66: Same as question above on why is this required and if std::move doesn't work.

projects/rocprofiler-compute/src/lib/pc_sampling_collector/pc_sampling_collector.h
```diff
@@ -7,10 +7,14 @@
 
 #include <rocprofiler-sdk/rocprofiler.h>
 
+#include <cstddef>
 #include <filesystem>
+#include <map>
 #include <memory>
 #include <set>
+#include <string>
```
svolkov-amd, line 15: Why this include is here - don't see string used in the header.

projects/rocprofiler-compute/src/lib/pc_sampling_collector/pc_sampling_collector.h
```diff
@@ -7,10 +7,14 @@
 
 #include <rocprofiler-sdk/rocprofiler.h>
 
+#include <cstddef>
```
svolkov-amd, line 10: Why is this header included here?

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_code_object_writer.cpp
```diff
@@ -4,6 +4,34 @@
 
 #include "nlohmann/json.hpp"
 
+#include <string>
+#include <string_view>
+
+namespace
+{
+template<typename Operation>
+void expect_runtime_throw(Operation operation)
```
svolkov-amd, line 13: Could you please remove this wrapper? It doesn't shorten much but makes it harder to read.

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_code_object_writer.cpp
```diff
@@ -4,6 +4,34 @@
...
+{
+    EXPECT_THROW(operation(), std::runtime_error);
+}
+
+void expect_kernel_json(const nlohmann::json& kernel, uint64_t id, std::string_view name)
```
svolkov-amd, line 18: Declare it as a protected function of `test_code_object_writer_t` and move definition at the bottom of the file, so that tests themselves are at the top as the most important bit of information.

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_code_object_writer.cpp
```diff
@@ -4,6 +4,34 @@
...
+{
+    EXPECT_EQ(kernel["kernel_id"].get<uint64_t>(), id);
+    EXPECT_EQ(kernel["name"].get<std::string>(), std::string{name});
+}
+
+void expect_instruction_json(const nlohmann::json&                          instruction,
```
svolkov-amd, line 24: Same as for previous function: split on declaration and implementation. Declaration as protected member of `test_code_object_writer_t` and implementation is at the bottom of this file.

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_code_object_writer.cpp
```diff
@@ -137,6 +179,69 @@ TEST_F(test_code_object_writer_t, ProvidedNoSymbols_SerializesEmptySymbolsArray)
...
+    m_writer.start_code_obj(1);
+    m_writer.end_code_obj();
+
+    const auto json = nlohmann::json::parse(m_writer.get_result());
+    ASSERT_TRUE(json["code_objects"][0].contains("kernels"));
+    EXPECT_TRUE(json["code_objects"][0]["kernels"].is_array());
+    EXPECT_EQ(json["code_objects"][0]["kernels"].size(), 0);
+}
+
+TEST_F(test_code_object_writer_t, ProvidedKernel_SerializesItInsideCodeObj)
```
svolkov-amd, line 193: This test is a special case of `ProvidedMultipleKernels_SerializesAllInOrder`, could be removed.

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_code_object_writer.cpp
```diff
@@ -137,6 +179,69 @@ TEST_F(test_code_object_writer_t, ProvidedNoSymbols_SerializesEmptySymbolsArray)
...
+{
+    m_writer.start_code_obj(1);
+    m_writer.write_kernel(rocprofiler_compute_tool::kernel_t{42, "kernel0"});
+    m_writer.write_kernel(rocprofiler_compute_tool::kernel_t{43, "kernel1"});
+    m_writer.end_code_obj();
+
+    const auto  json    = nlohmann::json::parse(m_writer.get_result());
+    const auto& kernels = json["code_objects"][0]["kernels"];
+    ASSERT_EQ(kernels.size(), 2);
+    expect_kernel_json(kernels[0], 42, "kernel0");
```
svolkov-amd, line 217: Don't like duplicating magic values here, like `42` and `kernel0`. Needs to be defined once. Consider defining test kernel descriptions as `test_code_object_writer_t` protected members:
```
rocprofiler_compute_tool::kernel_t m_kernel_0{42, "kernel0"};
rocprofiler_compute_tool::kernel_t m_kernel_1{43, "kernel1"};
```

This will also reduce the size of other tests below.

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_code_object_writer.cpp
```diff
@@ -137,6 +179,69 @@ TEST_F(test_code_object_writer_t, ProvidedNoSymbols_SerializesEmptySymbolsArray)
...
+    m_writer.start_code_obj(200);
+    m_writer.write_kernel(rocprofiler_compute_tool::kernel_t{44, "kernel2"});
+    m_writer.end_code_obj();
+
+    const auto json = nlohmann::json::parse(m_writer.get_result());
+    ASSERT_EQ(json["code_objects"].size(), 2u);
+
+    EXPECT_EQ(json["code_objects"][0]["id"], 100);
+    ASSERT_EQ(json["code_objects"][0]["kernels"].size(), 2);
+    expect_kernel_json(json["code_objects"][0]["kernels"][0], 42, "kernel0");
```
svolkov-amd, line 237: Same here - large test, many duplications of magic values, which makes it harder to maintain.

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_code_object_writer.cpp
```diff
@@ -220,11 +325,31 @@ TEST_F(test_code_object_writer_t, ProvidedInstruction_SerializesItInsideSymbol)
...
     ASSERT_EQ(instructions.size(), 1);
-    EXPECT_EQ(instructions[0]["name"], m_inst0.name);
-    EXPECT_EQ(instructions[0]["comment"], m_inst0.comment);
-    EXPECT_EQ(instructions[0]["virtual_address"], m_inst0.virtual_address);
-    EXPECT_EQ(instructions[0]["code_obj_offset"], m_inst0.code_obj_offset);
-    EXPECT_EQ(instructions[0]["size"], m_inst0.size);
+    expect_instruction_json(instructions[0], m_inst0);
+}
+
+TEST_F(test_code_object_writer_t, ProvidedKernelAndInstruction_PreservesInstructionOutput)
```
svolkov-amd, line 331: Not very clear name of the test, not sure what Output is. Could you please rephrase it?

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_code_object_writer.cpp
```diff
@@ -220,11 +325,31 @@ TEST_F(test_code_object_writer_t, ProvidedInstruction_SerializesItInsideSymbol)
...
+    m_writer.write_instruction(m_inst0);
+    m_writer.end_symbol();
+    m_writer.end_code_obj();
+
+    const auto  json    = nlohmann::json::parse(m_writer.get_result());
+    const auto& obj     = json["code_objects"][0];
+    const auto& symbols = obj["symbols"];
+    ASSERT_EQ(obj["kernels"].size(), 1);
+    ASSERT_EQ(symbols.size(), 1);
+    EXPECT_EQ(symbols[0]["name"], m_symbol0.name);
```
svolkov-amd, line 345: Do you need to do these checks for symbol there? Would it be sufficient to check the symbols.size() only? Assuming that symbols serialization is checked at `ProvidedSymbol_SerializesItInsideCodeObj` test.

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_pc_sampling_collector.cpp
```diff
@@ -27,6 +27,34 @@ std::set<std::filesystem::path> collect_source_paths_from_comment(std::string co
     collector->finalize(*writer);
     return collector->get_source_paths();
 }
+
+void expect_kernel(const mock_code_object_writer_t::kernel_description_t& kernel,
```
svolkov-amd, line 31: Split on declaration and definition. Move declarate to test class, definition at the bottom of this file. Same for other functions in this file. Test file should start from tests and these functions are "implementation details".

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_pc_sampling_collector.cpp
```diff
@@ -247,3 +262,25 @@ void test_pc_sampling_collector_t::SetUp()
...
     m_file_info.load_size      = 0x2000;
 }
+
+INSTANTIATE_TEST_SUITE_P(
+    SourcePaths,
+    test_pc_sampling_collector_source_paths_t,
+    ::testing::Values(
+        source_path_test_params_t{"EmptyComment", "", {}},
+        source_path_test_params_t{"SingleFileLine", "kernel.cpp:42", {"kernel.cpp"}},
+        source_path_test_params_t{"MultipleFrames", "kernel.cpp:42 -> header.h:8", {"header.h", "kernel.cpp"}},
```
svolkov-amd, line 272: Just a sanity check: have you actually seen such examples on real workloads and checked that this parsing works?

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_pc_sampling_collector.cpp
```diff
@@ -247,3 +262,25 @@ void test_pc_sampling_collector_t::SetUp()
...
+        source_path_test_params_t{"InjectedFrameSeparator",
+                                  "kernel.cpp:42 | header.h:8",
+                                  {"header.h", "kernel.cpp"},
+                                  " | "},
```
svolkov-amd, line 276: Same as a previous comment: have we seen `|` separators used in real workloads? If no, why do we have a capability to define a separator?

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_pc_sampling_collector.cpp
```diff
@@ -247,3 +262,25 @@ void test_pc_sampling_collector_t::SetUp()
...
+        source_path_test_params_t{"UnknownLineToken", "kernel.cpp:?", {"kernel.cpp"}},
+        source_path_test_params_t{"NoColonFrame", "kernel.cpp", {"kernel.cpp"}},
+        source_path_test_params_t{"DuplicateFrames", "kernel.cpp:42 -> kernel.cpp:42", {"kernel.cpp"}},
+        source_path_test_params_t{"PathWithDirectories",
+                                  "/tmp/project/include/header.h:8",
+                                  {"/tmp/project/include/header.h"}},
+        source_path_test_params_t{"TrailingColon", "kernel.cpp:", {"kernel.cpp:"}},
```
svolkov-amd, line 283: This is obviously an incorrect processing. If you return `kernel.cpp:` as a file name then could could it be used to copy this file?

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_pc_sampling_collector.cpp
```diff
@@ -247,3 +262,25 @@ void test_pc_sampling_collector_t::SetUp()
...
+        source_path_test_params_t{"TrailingSeparator", "kernel.cpp:42 -> ", {"kernel.cpp"}},
+        source_path_test_params_t{"NonNumericColonSuffix", "kernel.cpp:label", {"kernel.cpp:label"}}),
```
svolkov-amd, line 285: Same as above, if you return `kernel.cpp:label` as an output how could it be used to copy the file?

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_pc_sampling_collector.cpp
```diff
@@ -247,3 +262,25 @@ void test_pc_sampling_collector_t::SetUp()
     m_file_info.load_base      = 0x1000;
     m_file_info.load_size      = 0x2000;
 }
+
+INSTANTIATE_TEST_SUITE_P(
+    SourcePaths,
+    test_pc_sampling_collector_source_paths_t,
+    ::testing::Values(
+        source_path_test_params_t{"EmptyComment", "", {}},
```
svolkov-amd, line 270: Please update test names to match other test names in this file, Ex, `ProvidedEmptyComment_ReturnsNoPaths`. Applies to other lines below.

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_pc_sampling_collector.cpp
```diff
@@ -27,6 +27,34 @@ std::set<std::filesystem::path> collect_source_paths_from_comment(std::string co
...
+                   size_t                                                 code_object_id,
+                   uint64_t                                               kernel_id,
+                   const std::string&                                     name)
+{
+    EXPECT_EQ(kernel.code_object_id, code_object_id);
+    EXPECT_EQ(kernel.kernel_id, kernel_id);
+    EXPECT_EQ(kernel.name, name);
+}
+
+struct source_path_test_params_t
```
svolkov-amd, line 41: Move to header please.

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_pc_sampling_collector.cpp
```diff
@@ -27,6 +27,34 @@ std::set<std::filesystem::path> collect_source_paths_from_comment(std::string co
...
+    std::set<std::filesystem::path> expected_paths;
+    std::string                     separator = " -> ";
+};
+
+class test_pc_sampling_collector_source_paths_t
+    : public ::testing::TestWithParam<source_path_test_params_t>
+{
+};
+
+std::string source_path_test_name(const ::testing::TestParamInfo<source_path_test_params_t>& info)
```
svolkov-amd, line 54: Move to header please

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_pc_sampling_collector.cpp
```diff
@@ -67,6 +95,62 @@ TEST_F(test_pc_sampling_collector_t, ProvidedCodeObjects_WritesTheirIds)
     EXPECT_EQ(m_writer->get_start_code_obj_ids()[1], m_mem_info.code_object_id);
 }
 
+TEST_F(test_pc_sampling_collector_t, ProvidedKernelSymbols_WritesThemUnderCodeObjects)
+{
+    m_pc_sampling_collector->on_code_object_load(m_file_info);
+    m_pc_sampling_collector->on_code_object_load(m_mem_info);
+    m_pc_sampling_collector->on_kernel_symbol_register(m_mem_info.code_object_id,
```
svolkov-amd, line 102: Declare kernel0, kernel1, kernel2 as private or protected members of a test class to eliminate duplication across tests.

projects/rocprofiler-compute/src/lib/pc_sampling_collector/tests/test_pc_sampling_collector.cpp
```diff
@@ -67,6 +95,62 @@ TEST_F(test_pc_sampling_collector_t, ProvidedCodeObjects_WritesTheirIds)
     EXPECT_EQ(m_writer->get_start_code_obj_ids()[1], m_mem_info.code_object_id);
 }
 
+TEST_F(test_pc_sampling_collector_t, ProvidedKernelSymbols_WritesThemUnderCodeObjects)
```
svolkov-amd, line 98: Seems like a special case of `ProvidedMultipleKernelSymbols_PreservesOrder`, could be deleted. Also `ProvidedMultipleKernelSymbols_PreservesOrder` would be nice to rename to `ProvidedMultipleKernelSymbols_WritesThemInOrderUnderCodeObjects`

projects/rocprofiler-compute/src/lib/rocprofiler_compute_tool/sdk_callbacks.cpp
```diff
@@ -341,14 +341,18 @@ void SdkCallbacksImpl::tool_tracing_callback(rocprofiler_callback_tracing_record
         // if matches store kernel id
         // Lock before modifying target_kernel_ids
         std::lock_guard<std::mutex> lock(tool->mut);
-        if (!tool->kernel_filter_include_regex.empty())
+        const bool                  pc_sampling_enabled = tool->pc_sampling.enabled();
```
svolkov-amd, line 344: No need to define variable there as `tool->pc_sampling.enabled()` is already self-explanatory and this variable is used a single time below.

projects/rocprofiler-compute/src/lib/rocprofiler_compute_tool/sdk_callbacks.cpp
```diff
@@ -341,14 +341,18 @@ void SdkCallbacksImpl::tool_tracing_callback(rocprofiler_callback_tracing_record
...
+        const bool                  filtering_by_name = !tool->kernel_filter_include_regex.empty();
```
svolkov-amd, line 345: Here variable definition is justified as it makes code more clear. The only nitpick I have to that it would be nice to make its name more clear, ex `is_filtering_by_name_enabled`.

projects/rocprofiler-compute/src/lib/rocprofiler_compute_tool/sdk_callbacks.cpp
```diff
@@ -341,14 +341,18 @@ void SdkCallbacksImpl::tool_tracing_callback(rocprofiler_callback_tracing_record
...
+
+        const char* raw_kernel_name = data->kernel_name == nullptr ? "" : data->kernel_name;
```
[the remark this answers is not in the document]
svolkov-amd: Agree with Copilot here. This is an old code so this check is missing here for a long time. Also to be more specific, throw an exception in case payload is nullptr and write a corresponding unit test which checks it.

projects/rocprofiler-compute/src/lib/rocprofiler_compute_tool/sdk_callbacks.cpp
```diff
@@ -341,14 +341,18 @@ void SdkCallbacksImpl::tool_tracing_callback(rocprofiler_callback_tracing_record
...
+
+        if (filtering_by_name)
         {
+            const auto stripped_kernel_name = strip_kd_suffix(raw_kernel_name);
```
svolkov-amd, line 351: Why is this strip of kd suffix required before a regex search? What is ".kd" suffix in a first place?

projects/rocprofiler-compute/src/lib/rocprofiler_compute_tool/sdk_callbacks.cpp
```diff
@@ -367,6 +371,13 @@ void SdkCallbacksImpl::tool_tracing_callback(rocprofiler_callback_tracing_record
         {
             tool->target_kernel_ids.insert(data->kernel_id);
         }
+
+        if (pc_sampling_enabled)
+        {
+            const auto kernel_name = truncate_name(strip_kd_suffix(raw_kernel_name));
```
svolkov-amd, line 377: Why do we don't need to do demangle here? Would `truncate_name` even work properly in this case?

projects/rocprofiler-compute/src/lib/rocprofiler_compute_tool/sdk_callbacks.cpp
```diff
@@ -367,6 +371,13 @@ void SdkCallbacksImpl::tool_tracing_callback(rocprofiler_callback_tracing_record
...
+            tool->pc_sampling.on_kernel_symbol_register(static_cast<size_t>(data->code_object_id),
```
svolkov-amd, line 378: Why do `static_cast` to size_t of the field which is already a size_t?

projects/rocprofiler-compute/src/lib/rocprofiler_compute_tool/tests/test_sdk_callbacks.cpp
```diff
@@ -7,6 +7,15 @@
 
 using namespace rocprofiler_compute_tool;
 
+namespace
+{
+std::string kernel_symbol_register_test_name(
+    const ::testing::TestParamInfo<kernel_symbol_register_test_params_t>& info)
+{
+    return fmt::format("KernelNameCase{}", info.index);
```
svolkov-amd, line 15: Not sure if it makes the output any bit more readable. Without it gtest will still write a test name + index. Do we need this block of code?

projects/rocprofiler-compute/src/lib/rocprofiler_compute_tool/tests/test_pc_sampling_feature.cpp
```diff
@@ -3,6 +3,7 @@
 #include "test_pc_sampling_feature.h"
 
 #include <set>
+#include <string>
```
svolkov-amd, line 6: Don't see string used below. Is this include required here?

projects/rocprofiler-compute/src/lib/rocprofiler_compute_tool/tests/test_sdk_callbacks.cpp
```diff
@@ -269,6 +408,13 @@ void TestSdkCallbacks::invoke_record_callback(uint64_t           counter_id,
 }
 
 void TestSdkCallbacks::invoke_tool_tracing_callback(uint64_t kernel_id, const std::string& kernel_name)
+{
+    invoke_kernel_symbol_register(0, kernel_id, kernel_name.c_str());
```
svolkov-amd, line 412: The confusing part there is that `invoke_kernel_symbol_register` call which is executed here doesn't guarantee to actually invoke `kernel_symbol_register` if pc_sampling is disabled. Proposal is to either define two overloaded `invoke_tool_tracing_callback` or define a single functino but make code_object_id a last parameter with default value set to zero.

projects/rocprofiler-compute/src/lib/rocprofiler_compute_tool/tests/test_sdk_callbacks.cpp
```diff
@@ -184,7 +187,131 @@ TEST_F(TestSdkCallbacks, ProvidedCodeObjectLoadWithPcSamplingEnabled_ForwardsToC
...
+    ASSERT_EQ(calls.size(), 2);
+    EXPECT_EQ(calls[0].code_object_id, matching_code_object_id);
+    EXPECT_EQ(calls[0].kernel_id, matching_kernel_id);
+    EXPECT_EQ(calls[0].name, "_Z9my_kernelv");
+    EXPECT_EQ(calls[1].code_object_id, nonmatching_code_object_id);
+    EXPECT_EQ(calls[1].kernel_id, nonmatching_kernel_id);
+    EXPECT_EQ(calls[1].name, "_Z12other_kernelv");
+}
+
+TEST_F(TestSdkCallbacks, ProvidedMangledKernelName_FilterMatchesDemangledPcSamplingKeepsMangled)
```
svolkov-amd, line 238: Doesn't seem to test filtering with demangled name. Because ".*jacobi_iteration.*" would match both mangled and demangled names.

projects/rocprofiler-compute/src/lib/rocprofiler_compute_tool/tests/test_sdk_callbacks.cpp
```diff
@@ -184,7 +187,131 @@ TEST_F(TestSdkCallbacks, ProvidedCodeObjectLoadWithPcSamplingEnabled_ForwardsToC
...
+{
+    constexpr size_t   matching_code_object_id    = 123;
+    constexpr size_t   nonmatching_code_object_id = 456;
+    constexpr uint64_t matching_kernel_id         = 10;
+    constexpr uint64_t nonmatching_kernel_id      = 20;
+
+    auto& collector                          = enable_pc_sampling();
+    m_tool_data->kernel_filter_include_regex = ".*my_kernel.*";
+
+    invoke_kernel_symbol_register(matching_code_object_id, matching_kernel_id, "_Z9my_kernelv");
```
svolkov-amd, line 222: These values are hardcoded below. Consider declaring `kernel_symbol_register_call_t` as members of `TestSdkCallbacks` and then passing them into `invoke_kernel_symbol_register` and comparing against in `EXPECT_EQ`.

projects/rocprofiler-compute/src/lib/rocprofiler_compute_tool/tests/test_sdk_callbacks.cpp
```diff
@@ -184,7 +187,131 @@ TEST_F(TestSdkCallbacks, ProvidedCodeObjectLoadWithPcSamplingEnabled_ForwardsToC
...
+
+TEST_F(TestSdkCallbacks, ProvidedMangledKernelName_FilterMatchesDemangledPcSamplingKeepsMangled)
+{
+    constexpr size_t   code_object_id = 123;
+    constexpr uint64_t kernel_id      = 10;
+
+    auto& collector                          = enable_pc_sampling();
+    m_tool_data->kernel_filter_include_regex = ".*jacobi_iteration.*";
+
+    invoke_kernel_symbol_register(code_object_id, kernel_id, "_Z16jacobi_iterationPfPKfi");
```
svolkov-amd, line 246: Same as above - consider removing duplication of strings. Also every test case there has different strings. I'm pretty sure it's sufficient to define 2-3 variants of kernel names and use them consistently across all the tests.

projects/rocprofiler-compute/src/lib/rocprofiler_compute_tool/tests/test_sdk_callbacks.cpp
```diff
@@ -184,7 +187,131 @@ TEST_F(TestSdkCallbacks, ProvidedCodeObjectLoadWithPcSamplingEnabled_ForwardsToC
...
+    EXPECT_TRUE(m_tool_data->target_kernel_ids.count(kernel_id));
+
+    const auto& calls = collector.get_kernel_symbol_register_calls();
+    ASSERT_EQ(calls.size(), 1);
+    EXPECT_EQ(calls[0].code_object_id, code_object_id);
+    EXPECT_EQ(calls[0].kernel_id, kernel_id);
+    EXPECT_EQ(calls[0].name, "_Z16jacobi_iterationPfPKfi");
+}
+
+TEST_F(TestSdkCallbacks, ProvidedKdSuffix_FilterAndPcSamplingUseStrippedName)
```
svolkov-amd, line 258: I believe you are checking it in `TestSdkCallbacksKernelSymbolRegister` parametric test. Also filtering is tested above. Don't see reason to repeat these checks.

projects/rocprofiler-compute/src/lib/rocprofiler_compute_tool/tests/test_sdk_callbacks.cpp
```diff
@@ -184,7 +187,131 @@ TEST_F(TestSdkCallbacks, ProvidedCodeObjectLoadWithPcSamplingEnabled_ForwardsToC
...
+    EXPECT_TRUE(m_tool_data->target_kernel_ids.count(kernel_id));
+
+    const auto& calls = collector.get_kernel_symbol_register_calls();
+    ASSERT_EQ(calls.size(), 1);
+    EXPECT_EQ(calls[0].code_object_id, code_object_id);
+    EXPECT_EQ(calls[0].kernel_id, kernel_id);
+    EXPECT_EQ(calls[0].name, "__amd_rocclr_streamOpsWrite");
+}
+
+TEST_F(TestSdkCallbacks, ProvidedInternalKd_FilterAndPcSamplingKeepInternalKd)
```
svolkov-amd, line 278: These checks are already done in `TestSdkCallbacksKernelSymbolRegister` parametric test. Therefore does this test required?

projects/rocprofiler-compute/src/lib/rocprofiler_compute_tool/tests/test_sdk_callbacks.cpp
```diff
@@ -184,7 +187,131 @@ TEST_F(TestSdkCallbacks, ProvidedCodeObjectLoadWithPcSamplingEnabled_ForwardsToC
...
+    EXPECT_TRUE(m_tool_data->target_kernel_ids.count(kernel_id));
+
+    const auto& calls = collector.get_kernel_symbol_register_calls();
+    ASSERT_EQ(calls.size(), 1);
+    EXPECT_EQ(calls[0].code_object_id, code_object_id);
+    EXPECT_EQ(calls[0].kernel_id, kernel_id);
+    EXPECT_EQ(calls[0].name, "prefix.kd.middle");
+}
+
+TEST_F(TestSdkCallbacks, ProvidedEmptyKernelName_FilterDoesNotTargetAndPcSamplingForwardsEmpty)
```
svolkov-amd, line 298: Just add this check into parametric `TestSdkCallbacksKernelSymbolRegister` test.

svolkov-amd (changes requested):

vedithal-amd (changes requested):
Context. [#8242](https://github.com/ROCm/rocm-systems/pull/8242) (This PR) (tracked by [AIPROFCOMP-701](https://amd-hub.atlassian.net/browse/AIPROFCOMP-701)) adds a kernels[] array to code_obj_info.json so collected ISA can be attributed to kernels. [#8395](https://github.com/ROCm/rocm-systems/pull/8395) is the consumer that ingests code_obj_info.json into the analysis DB and needs to tag each disassembled instruction with its kernel.

I logged AIPROFCOMP-701, and the premise was wrong: I overlooked that symbols[].name already carries the kernel name. Apologies for the confusion and the time spent on #8242.

The premise is wrong. The ticket assumes the artifact has "no link between a code object and its kernels." It already does: symbols[].name is the ELF symbol name (the kernel name, .kd-free), with each kernel's disassembly under that symbol.

The id mapping already exists. ps_file_results.json -> kernel_symbols[] gives kernel_id, code_object_id, and kernel_name per kernel. #8395 already reads this file (it pulls load_base from it), so kernel_id <-> code_object_id <-> name is available at ingestion.

Fix in #8395 instead. For each instruction under a symbols[] entry: take symbol.name + code_object_id, match kernel_symbols[] on code_object_id and kernel_name (.kd stripped) == symbol.name, use that kernel_id.

Decision: close/drop #8242; implement kernel attribution in #8395 via the method above.

Separation of concerns. Runtime facts (which code object loaded, its load_base, kernel_id, dispatch) belong in ps_file_results.json and #8395 (My code obj info ingestion PR) already sources them there (it records ISA only for code objects the ps-file shows as loaded); code_obj_info.json should stay pure disassembly, so #8242's (THIS PR) kernels[] wrongly mixes runtime linkage into the disassembly artifact.

Lets hold off on this PR or close it - since it does not add much value in the current state

PR #8246, opened 2026-07-08
fix(rocprofiler-compute): correct bandwidth unit label from Gb/s to GB/s

### LINKED ISSUE #8245: rocprofiler-compute: bandwidth unit label displays 'Gb/s' instead of 'GB/s'
## Description

The GUI and TUI display code in rocprofiler-compute uses "Gb/s" (gigabits per second) instead of "GB/s" (gigabytes per second) for bandwidth metrics in two places:

1. **`src/utils/gui.py:110`** — DataFrame filter uses `Unit == "Gb/s"`, but the YAML metric definitions define units as `"GB/s"`. This causes the filter to match zero rows, producing an empty HBM bandwidth bar chart in the web UI.

2. **`src/rocprof_compute_tui/widgets/charts.py:196`** — The x-axis label override for chart ID `1701.2` (L2 Cache SoL bandwidth) hardcodes `"Gb/s"` instead of `"GB/s"`.

## Steps to Reproduce

1. Profile a kernel with `rocprof-compute profile`
2. Run `rocprof-compute analyze -p <workload_dir>`
3. Navigate to L2 Cache section (table 1701) in the web GUI — the HBM bandwidth chart is empty
4. In the TUI, the bandwidth axis shows "Gb/s" instead of "GB/s"

## Expected Behavior

- The HBM bandwidth chart should display data (the filter should match `GB/s` rows)
- All bandwidth labels should read "GB/s" (gigabytes per second), matching the YAML definitions and the actual unit of the values

alexioslyrakis-amd:
Closes #8245

## Summary
- **Root cause**: `update_normal_unit_string()` in `expression.py` used `.capitalize()` to uppercase the first character of unit label strings, but `.capitalize()` also lowercases all subsequent characters — turning `"GB/s"` into `"Gb/s"` (gigabits) and `"Conflicts per Access"` into `"Conflicts per access"` for every metric in the CLI output.
- **Downstream symptoms** in the GUI/TUI display code that hardcoded `"Gb/s"` to match the (already mangled) data:
  - `gui.py`: DataFrame filter `Unit=="Gb/s"` failed to match YAML-defined `"GB/s"` rows, producing an empty HBM bandwidth chart.
  - `charts.py`: x-axis label for the SoL bandwidth chart (id=1701.2) displayed `"Gb/s"`.

## Changes
- `src/utils/metrics/expression.py`: Replace `.capitalize()` with first-char uppercasing that preserves the rest of the string
- `src/utils/gui.py`: Fix unit filter from `"Gb/s"` to `"GB/s"`
- `src/rocprof_compute_tui/widgets/charts.py`: Fix label from `"Gb/s"` to `"GB/s"`
- `tests/test_bandwidth_unit_label.py`: Regression tests for all three fixes

## Test plan
- [x] Root cause test: `update_normal_unit_string("GB/s", ...)` returns `"GB/s"` not `"Gb/s"` (parametrized for multiple units)
- [x] GUI test: `create_sol_charts` produces bandwidth chart with data (not empty)
- [x] TUI test: `px_simple_bar` uses `"GB/s"` axis label
- [x] All tests fail on unfixed code, pass on fixed code (verified locally)
- [x] Verify with `rocprof-compute analyze` CLI output on real profiling data


projects/rocprofiler-compute/tests/test_bandwidth_unit_label.py
vedithal-amd: TUI and GUI currently don't have any test cases and they are in deprecated mode. I think we missed to mark them as experimental in the argparser but the fact stands. That being said, adding test cases for TUI and GUI is a separate concern

projects/rocprofiler-compute/src/utils/metrics/expression.py
vedithal-amd: .capitalize() lowercases everything after the first char, so it corrupts acronyms in the unit strings: (IOPS + $normUnit) renders as Iops per wave, (FLOPs + ...) as Flops per wave. The yaml unit values already start capitalized (FLOPs, Cycles, Requests), so this call only  does harm. Please drop .capitalize() and just return the re.sub(...) result.

vedithal-amd (changes requested):
Please remove the testing and remove capitalize otherwise looks good

projects/rocprofiler-compute/src/utils/metrics/expression.py
alexioslyrakis-amd: Good call, dropped it entirely instead of replacing!

projects/rocprofiler-compute/tests/test_bandwidth_unit_label.py
alexioslyrakis-amd: Oh didn't know, ok removed the GUI and TUI test classes. Kept only a regression test for `update_normal_unit_string` that verifies that strings like "GB/s" pass through without case corruption

vedithal-amd: Folded the GB/s regression into TestExpression in test_metric_utils.py (replacing the stale capitalize test) and removed this standalone module. Made the change directly to save a round-trip. See cb13897ab06.

vedithal-amd (approved):
LGTM

vedithal-amd:
test_pc_sampling test which keeps getting timed out has been disabled temporarily - incorporating changes from develop branch to make the ci run green now

PR #8247, opened 2026-07-08
fix(rocjitsu): Add opsel support for vop3 16bit operands

lua1235:
## Motivation

For some vop3 instructions with true16 operands, Rocjitsu ignored OPSEL and always used the lower half of the VGPRs instead of selecting the encoded source and destination halves.

## Technical Details

- Add true16 source selection to vop3 scalar and SIMD paths where applicable.
- Add corresponding tests covering representative CDNA, RDNA, and gfx1250 use cases, to check source-half selection, destination half preservation, scalar/SIMD agreement.

## JIRA ID

N/A

## Test Plan

regenerate files, run rocjitsu tests, hip-fpsan tests 

## Test Result

Tests pass

## Submission Checklist


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/simd_glue.h
```diff
@@ -1598,19 +1621,85 @@ template <typename T, typename Inst, typename BinOp>
...
+  const auto a_bcast = r0 ? util::native<T>{} : util::broadcast<T>(inst.src0.read_lane(wf, 0));
+  const auto b_bcast = r1 ? util::native<T>{} : util::broadcast<T>(inst.src1.read_lane(wf, 0));
+  for (uint32_t base = 0; base < wf.wf_size(); base += static_cast<uint32_t>(W)) {
+    const uint64_t chunk = (exec >> base) & chunk_full;
+    if (chunk == 0)
+      continue;
+    const auto a = select_vop3_true16_src(simd_load_or<T>(r0, base, a_bcast), opsel, 0);
+    const auto b = select_vop3_true16_src(simd_load_or<T>(r1, base, b_bcast), opsel, 1);
+    const auto r = bin_op(a, b);
+    write_vop3_true16_simd_at(rd, inst.vdst, wf, base, r, chunk, opsel);
```
kuhar: We should split the true16 SIMD writeback from the shared zero-extending f16 VOP3 SIMD helpers. This helper and the unary/ternary fp16 siblings now always select and merge true16 halves, but many shared scalar callers still zero-extend the full dword on non-true16 arches such as CDNA4. The SIMD path preserves the old high 16 bits, so scalar and SIMD diverge even with `op_sel == 0`.

I reproduced this locally with:

```bash
build/tests/rocjitsu_tests --gtest_filter='Vop3BinarySimdCorrectness.F16_AllModifiers_*:Vop3Fp16UnarySimdCorrectness.*:Vop3Fp16TranscendentalSimdCorrectness.*:Vop3TernaryFpSimdCorrectness.*'
```

All 9 selected tests fail; for example `v_add_f16` lane 0 reports `scalar_out=0x00003c00` but `simd_out=0xcdcd3c00`. Please either keep the shared f16 SIMD probes zero-extending/full-dword, or route true16 and zero-extend f16 ops through distinct SIMD helpers. Once fixed, an `op_sel` scalar-vs-SIMD regression would pin both behaviors.

kuhar (commented):
For the record, this regenerated all the files properly -- I confirmed this locally

Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/simd_glue.h:1500`

We should use the true16 lane-read path for scalar broadcasts here, or bail out for scalar/inline f16 sources. The SIMD path broadcasts non-VGPR operands with `read_scalar()`, while the scalar true16 path goes through `read_vop3_true16_src(...)->read_lane()`.

That matters for inline f16 constants: encoding 242 resolves through `read_scalar()` as f32 `0x3f800000`, but the 16-bit lane resolver returns f16 `0x3c00`. After `select_vop3_true16_src`, the SIMD path can classify the wrong bits for inline f16 operands. The existing `CmpClassF16UsesSelectedMaskHalf` coverage uses VGPR sources, so it does not cover this broadcast case.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -6854,7 +6953,11 @@ def _write_shared_execute_templates(self) -> None:
                 f'inline void execute_{mnemonic}('
                 f'[[maybe_unused]] Inst &inst, [[maybe_unused]] Wavefront &wf) {{'
             )
-            probe = simd_probe_line(mnemonic)
+            true16_vop3_body = (
```
kuhar: I think it would help to carry an explicit true16 flag into the shared-body entries instead of deriving it from the generated C++ text. Here `true16_vop3_body` is set by searching `prefixed_body` for helper names, while the local-body path already computes `is_true16_vop3` from the instruction/profile model.

If a helper is renamed or a future true16 op uses a different helper, the shared probe could silently pick the zero-extend macro and reintroduce scalar-vs-SIMD divergence. Passing the model-derived boolean to `simd_probe_line` would keep both paths on the same source of truth.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -6031,6 +6115,7 @@ def _dpp_write_mask_lines(
                                 f'void {inst.fmt_name}::execute_impl'
                                 f'(amdgpu::Wavefront &wf) {{\n'
                                 f'{_dpp_preamble}'
+                                f'{_local_true16_probe_body}'
```
kuhar, line 6462: This would be safer if the local true16 SIMD probe did not return before the DPP/SDWA cleanup. `_local_true16_probe_body` is inserted after the DPP preamble but before `_dpp_cleanup` / `_sdwa_postamble`; generated gfx1250 `VRcpF16Vop3::execute_impl` has `ROCJITSU_TRY_SIMD_VOP3_UNARY_TRUE16_FP16(...)` before the `dpp_write_mask` restoration and `src0.clear_delegate()`.

If the probe succeeds, the macro returns early and skips that cleanup. Please either route these through the shared-template path so cleanup still runs, or disable the local probe when DPP/SDWA cleanup is present.

kuhar (commented):
A few comments

atgutier (approved):

kuhar (approved):

PR #8264, opened 2026-07-08
build(rocprofiler-sdk): Improve rocprofiler-sdk dependency discovery for clean ROCm installs

srihari-uttanur:
## Motivation

rocprofiler-sdk’s library dependencies for libdw, libdrm, libdrm_amdgpu and others fail in clean environments with just a rocm therock tarball installed. Manual intervation is necessary in these cases, patching the cmakelists.

rocprofiler-sdk previously located its libdw, libelf, libdrm and libdrm_amdgpu dependencies primarily through pkg-config (and, for libelf,a REQUIRED find-module). In clean ROCm / TheRock tarball environments the corresponding system development packages and their pkg-config (.pc) metadata are often not installed, so configuration failed and required
manually patching the CMake files.

This PR makes discovery resilient by introducing a layered fallback for each dependency and by teaching every lookup about the ROCm sysdeps layout, so the build configures against a bare ROCm/TheRock tarball without installing any additional system -dev packages.

Discovery strategy (libdw, libelf, libdrm, libdrm_amdgpu):
  1. Prefer an imported target from a CMake config package if available.
  2. Fall back to pkg-config when present (no longer treated as REQUIRED).
  3. Finally, perform a direct find_path/find_library search that also covers the ROCm install and its bundled sysdeps directory (lib/rocm_sysdeps) and the arch-specific multiarch libdir.

## Technical Details

**cmake/Modules/Findlibdw.cmake**
  - Short-circuit when libdw::libdw already exists or a libdw CMake config package is found; otherwise try pkg-config, then a manual search.
  - The manual search uses a shared hint list covering ${rocm_version_DIR}, ${ROCM_PATH}, $ENV{ROCM_PATH} and their lib/rocm_sysdeps subdirectories, and adds the multiarch suffix lib/<arch>-linux-gnu to find_library.

**cmake/Modules/Findlibelf.cmake**
  - Removed. libelf is only used internally (it is not part of the exported package interface), so its discovery is inlined in
    rocprofiler_config_interfaces.cmake rather than kept as a standalone find-module.

**cmake/rocprofiler_config_interfaces.cmake**
  - libelf: use find_package(LibElf CONFIG) and the elf::elf / libelf::libelf targets when available; otherwise discover inline via
    pkg-config, then a direct search that also covers the sysdeps location.
  - libdrm / libdrm_amdgpu: prefer config-package targets (libdrm::libdrm, drm::drm, libdrm and the amdgpu equivalents), then pkg-config (now made non-fatal instead of REQUIRED), then a direct search using the sysdeps-aware hint list. Removes the previous hard dependency on pkg-config being installed.

**cmake/rocprofiler_config_packaging.cmake**
  - Add ${rocm_version_DIR}/lib/rocm_sysdeps/lib to the Debian shlibdeps private-dirs so bundled sysdeps copies are treated as private.
  - Exclude the bundled sysdeps sonames from the RPM auto-requires via "%global __requires_exclude^librocm_sysdeps_.*" (set through CPACK_RPM_SPEC_MORE_DEFINE), preventing a public Requires on
    librocm_sysdeps_*.so for binaries that link the bundled copies.

**external/CMakeLists.txt**
  - Link the SQLite3 dependency through SQLite3::SQLite3 when that target exists, falling back to SQLite::SQLite3, to support both CMake FindSQLite3 target-name conventions.

## JIRA ID

AIPROFSDK-966

## Submission Checklist


ApoKalipse-V (approved):

projects/rocprofiler-sdk/cmake/rocprofiler_config_interfaces.cmake
```diff
@@ -267,10 +267,21 @@ rocprofiler_config_nolink_target(rocprofiler-sdk-hsakmt-nolink hsakmt::hsakmt)
...
+    $ENV{ROCM_PATH}
+    ${rocm_version_DIR}/lib/rocm_sysdeps
+    ${ROCM_PATH}/lib/rocm_sysdeps
+    $ENV{ROCM_PATH}/lib/rocm_sysdeps
+    /opt/amdgpu)
+
 if(PkgConfig_FOUND)
-    pkg_check_modules(DRM REQUIRED IMPORTED_TARGET libdrm)
-    pkg_check_modules(DRM_AMDGPU REQUIRED IMPORTED_TARGET libdrm_amdgpu)
+    pkg_check_modules(DRM IMPORTED_TARGET libdrm)
```
jrmadsen: No longer required?

srihari-uttanur: Earlier, the build logic said: “pkg-config must succeed for libdrm/libdrm_amdgpu, otherwise stop immediately.”

That was too strict for clean TheRock tarball environments, because those often have the actual headers and shared libraries, but not the pkg-config .pc files.

So changed it to:

1. Try pkg-config first (preferred path).
2. If pkg-config metadata is missing, do not fail yet.
3. Fall back to direct CMake searches for the real headers and libraries in ROCm/TheRock paths.
4. Fail only if those real files still cannot be found.

So nothing became optional at runtime. Only pkg-config metadata became optional. The actual libraries are still required.

lumachad:
Looks like this fails pre-commit (cmake formatting).

projects/rocprofiler-sdk/cmake/Modules/Findlibdw.cmake
```diff
@@ -24,6 +24,15 @@
...
 
+set(_libdw_ROOT_HINTS
+    ${libdw_ROOT}
+    ${rocm_version_DIR}
+    ${ROCM_PATH}
+    $ENV{ROCM_PATH}
+    ${rocm_version_DIR}/lib/rocm_sysdeps
+    ${ROCM_PATH}/lib/rocm_sysdeps
+    $ENV{ROCM_PATH}/lib/rocm_sysdeps)
+
```
lumachad: Do we need to set all of these to find libdw? As a counterexample (I'm not exactly sure what the situation is with rocprofiler-sdk), check rocr-debug-agent and how it looks up libdw both standalone and in TheRock:

https://github.com/ROCm/rocm-systems/blob/182bf6eedb4bbcc0b0bdd819787009f41dbdb6ed/projects/rocr-debug-agent/CMakeLists.txt#L85

projects/rocprofiler-sdk/cmake/Modules/Findlibelf.cmake
```diff
@@ -24,6 +24,15 @@
...
 
+set(_libelf_ROOT_HINTS
+    ${libelf_ROOT}
+    ${rocm_version_DIR}
+    ${ROCM_PATH}
+    $ENV{ROCM_PATH}
+    ${rocm_version_DIR}/lib/rocm_sysdeps
+    ${ROCM_PATH}/lib/rocm_sysdeps
+    $ENV{ROCM_PATH}/lib/rocm_sysdeps)
+
```
lumachad: Same comment as above. Check how we do it in rocr-debug-agent as a counterexample: https://github.com/ROCm/rocm-systems/blob/182bf6eedb4bbcc0b0bdd819787009f41dbdb6ed/projects/rocr-debug-agent/CMakeLists.txt#L133.

If that works for rocprofiler-sdk, might be a bit cleaner.

projects/rocprofiler-sdk/cmake/rocprofiler_config_interfaces.cmake
```diff
@@ -267,10 +267,21 @@ rocprofiler_config_nolink_target(rocprofiler-sdk-hsakmt-nolink hsakmt::hsakmt)
 
 find_package(PkgConfig)
 
+set(_rocprofiler_drm_ROOT_HINTS
```
lumachad, line 303: If we managed to get libdw/libelf lookups cleaner, I'd replicate that here.

projects/rocprofiler-sdk/cmake/rocprofiler_config_interfaces.cmake
```diff
@@ -279,37 +290,41 @@ else()
...
     find_library(
         drm_amdgpu_LIBRARY
         NAMES drm_amdgpu
-        HINTS ${rocm_version_DIR} ${ROCM_PATH} /opt/amdgpu
-        PATHS ${rocm_version_DIR} ${ROCM_PATH} /opt/amdgpu
-        PATH_SUFFIXES ${CMAKE_SYSTEM_PROCESSOR}-linux-gnu REQUIRED)
+        HINTS ${_rocprofiler_drm_ROOT_HINTS}
+        PATHS ${_rocprofiler_drm_ROOT_HINTS}
+        PATH_SUFFIXES lib lib64 lib/${CMAKE_SYSTEM_PROCESSOR}-linux-gnu
+                      ${CMAKE_SYSTEM_PROCESSOR}-linux-gnu REQUIRED)
```
lumachad: I expect all of these might become simpler if we simplify the lookups above. Again, take it with a grain of salt as I'm not sure about the inner workings of rocprofiler-sdk.

projects/rocprofiler-sdk/external/CMakeLists.txt
```diff
@@ -287,7 +287,11 @@ if(ROCPROFILER_BUILD_SQLITE3)
     add_dependencies(rocprofiler-sdk-sqlite3 rocprofiler-sdk-sqlite-build)
 else()
     find_package(SQLite3 REQUIRED)
-    target_link_libraries(rocprofiler-sdk-sqlite3 INTERFACE SQLite::SQLite3)
+    if(TARGET SQLite3::SQLite3)
+        target_link_libraries(rocprofiler-sdk-sqlite3 INTERFACE SQLite3::SQLite3)
+    else()
+        target_link_libraries(rocprofiler-sdk-sqlite3 INTERFACE SQLite::SQLite3)
+    endif()
```
lumachad, line 293: Cosmetic, but you could set the appropriate interface name in a variable and pass it to target_link_libraries unconditionally. Might be less code.

projects/rocprofiler-sdk/cmake/Modules/Findlibdw.cmake
```diff
@@ -22,64 +22,75 @@
 #   libdw::libdw
```
lumachad, line 22: You'll need some feedback from the profiler folks as I don't have a lot of context here, but why do we need these modules? If we're reworking the dependency system (like we did for rocr-debug-agent), you might want to drop these modules and go for an approach in the main CMakeLists.txt.

I'm assuming the modules help with standalone builds. TheRock won't use them.

projects/rocprofiler-sdk/cmake/Modules/Findlibelf.cmake
```diff
@@ -22,62 +22,80 @@
 # libelf::libelf
```
lumachad: Same comment for this modules as the libdw module.

projects/rocprofiler-sdk/cmake/rocprofiler_config_interfaces.cmake
```diff
@@ -201,9 +201,11 @@ target_link_libraries(rocprofiler-sdk-ptl INTERFACE PTL::ptl-static)
...
-find_package(LibElf QUIET)
-if(LibElf_FOUND)
+find_package(LibElf CONFIG QUIET)
+if(TARGET elf::elf)
     target_link_libraries(rocprofiler-sdk-elf INTERFACE elf::elf)
+elseif(TARGET libelf::libelf)
+    target_link_libraries(rocprofiler-sdk-elf INTERFACE libelf::libelf)
 else()
     find_package(libelf REQUIRED)
     target_link_libraries(rocprofiler-sdk-elf INTERFACE libelf::libelf)
```
lumachad: I'd advise you check here on how to find these dependencies in TheRock and adjust the standalone dependency lookup to match: https://github.com/ROCm/TheRock/blob/main/docs/development/dependencies.md#elfutils

projects/rocprofiler-sdk/cmake/Modules/Findlibdw.cmake
srihari-uttanur: Findlibdw.cmake is not just an internal helper — it's installed and exported as part of the rocprofiler-sdk CMake package, and consumed by downstream find_package(rocprofiler-sdk) users plus standalone samples/tests. So it can’t simply be dropped the way rocr-debug-agent did, because it’s part of rocprofiler-sdk’s exported package interface not just a standalone-build helper. 

libdw::libdw is a transitive link dependency of librocprofiler-sdk.so, so every downstream find_package(rocprofiler-sdk) needs this module

projects/rocprofiler-sdk/cmake/rocprofiler_config_interfaces.cmake
srihari-uttanur: rocr-debug-agent is a leaf project, it builds artifacts but doesn’t export a CMake package whose config re-resolves libdw for downstream consumers. rocprofiler-sdk is a library that ships a CMake package with libdw::libdw as a transitive dependency, so it must provide a Findlibdw mechanism at the install location. Inlining libdw discovery into the top-level CMakeLists.txt only would break: downstream find_package(rocprofiler-sdk), standalone samples, and standalone tests.

projects/rocprofiler-sdk/cmake/Modules/Findlibelf.cmake
srihari-uttanur: Findlibelf.cmake is build-time-only: it’s not in the install list and is only used inside rocprofiler_config_interfaces.cmake 
libelf is a private/internal dependency; libdw is public/transitive.

projects/rocprofiler-sdk/cmake/rocprofiler_config_packaging.cmake
```diff
@@ -226,6 +232,13 @@ set(CPACK_RPM_PACKAGE_AUTOREQ
     ON
     CACHE BOOL "") # auto-generate deps based on shared libs
 set(CPACK_RPM_PACKAGE_AUTOPROV ON) # generate list of shared libs provided by package
+
+# Keep bundled ROCm sysdeps libraries private to this package, matching the Debian
+# shlibdeps private-dirs behavior above. Without this, RPM auto-requires can emit a public
+# soname dependency on librocm_sysdeps_*.so for binaries that intentionally link the
+# bundled sysdeps copies.
+set(CPACK_RPM_REQUIRES_EXCLUDE_FROM "^librocm_sysdeps_.*")
```
lumachad, line 240: CPACK code is probably all dead code at this point, since we don't package things individually anymore. Up to the team to cleanup.

projects/rocprofiler-sdk/cmake/rocprofiler_config_packaging.cmake
```diff
@@ -196,8 +196,14 @@ set(CPACK_DEBIAN_PACKAGE_SHLIBDEPS_PRIVATE_DIRS
 if(rocm_version_DIR)
```
lumachad, line 196: Should use versioning scheme from TheRock. I think this is ${ROCM_MAJOR_VERSION}.${ROCM_MINOR_VERSION}.${ROCM_PATCH_VERSION} now.

projects/rocprofiler-sdk/cmake/Modules/Findlibdw.cmake
lumachad: Got it. Up to the team to decide if the new format is clean enough. I'm marking as resolved.

projects/rocprofiler-sdk/cmake/rocprofiler_config_interfaces.cmake
lumachad: Resolved from my side, but the team should confirm how much they want to cleanup the cmake files.

projects/rocprofiler-sdk/cmake/rocprofiler_config_packaging.cmake
meserve-amd, line 240: Major: Is anchoring the regex (`^`) correct here? What are we actually matching against? It seems more likely this should be a partial path (e.g. lib/rocm_sysdeps/librocm_sysdeps_.*)

projects/rocprofiler-sdk/cmake/rocprofiler_config_packaging.cmake
```diff
@@ -226,6 +232,13 @@ set(CPACK_RPM_PACKAGE_AUTOREQ
     ON
     CACHE BOOL "") # auto-generate deps based on shared libs
 set(CPACK_RPM_PACKAGE_AUTOPROV ON) # generate list of shared libs provided by package
+
+# Keep bundled ROCm sysdeps libraries private to this package, matching the Debian
+# shlibdeps private-dirs behavior above. Without this, RPM auto-requires can emit a public
+# soname dependency on librocm_sysdeps_*.so for binaries that intentionally link the
+# bundled sysdeps copies.
+set(CPACK_RPM_REQUIRES_EXCLUDE_FROM "^librocm_sysdeps_.*")
+set(CPACK_RPM_TESTS_PACKAGE_REQUIRES_EXCLUDE_FROM "^librocm_sysdeps_.*")
```
meserve-amd, line 241: Major: I don't think `CPACK_RPM_TESTS_PACKAGE_REQUIRES_EXCLUDE_FROM` exists, only the global version

projects/rocprofiler-sdk/cmake/rocprofiler_config_interfaces.cmake
```diff
@@ -201,12 +201,47 @@ target_link_libraries(rocprofiler-sdk-ptl INTERFACE PTL::ptl-static)
...
-else()
-    find_package(libelf REQUIRED)
+elseif(TARGET libelf::libelf)
     target_link_libraries(rocprofiler-sdk-elf INTERFACE libelf::libelf)
+else()
+    # libelf is only used internally by rocprofiler-sdk (it is not part of the exported
+    # package interface), so it is discovered inline here rather than through a standalone
+    # find-module: pkg-config first, then a direct search that also covers ROCm sysdeps
+    # locations.
+    set(_rocprofiler_elf_ROOT_HINTS
```
meserve-amd, line 214: Minor: Add `${libelf_ROOT}` to the beginning of hints (matches libdw)

meserve-amd (changes requested):
Some questions around how the cpack stuff is handled

Please also update the title of your PR, things are a bit out of order. I suggest:
build(rocprofiler-sdk): improve dependency discovery for clean ROCm installs
Under Conventional Commits, the commit message body is free form, so feel free to elaborate as much as you need to there.

projects/rocprofiler-sdk/cmake/rocprofiler_config_packaging.cmake
srihari-uttanur: Agree, But this is out of scope for this PR. WIll take this up for a new task.

projects/rocprofiler-sdk/cmake/rocprofiler_config_packaging.cmake
srihari-uttanur: This partly conflicts with above comment, will park this for now.

PR #8272, opened 2026-07-08
[ATT] Improved buffer management: per-agent shared trace buffers

minseobshin11:
## Motivation

Thread trace (ATT) allocated a separate GPU output buffer for every context on each agent (default 128 MB each), even though only one trace can be active per agent at a time. With multiple contexts on the same agent this wastes large amounts of device memory. This PR makes all contexts on an agent share a single output buffer sized to the largest requested `buffer_size`, and adds a per-agent lease so that at most one context traces a given agent at a time — a second, concurrent request runs untraced rather than writing into the shared buffer.

## Technical Details

- Adds a process-global per-agent buffer manager in `source/lib/rocprofiler-sdk/thread_trace/shared_trace_buffer.{hpp,cpp}`.
- `thread_trace::initialize()` runs a pre-pass that registers every context's per-agent `buffer_size` before any buffer is built, so the shared buffer is sized to the max; `finalize()` frees the buffers once.
- `TraceMemoryPool::Alloc` returns the shared device buffer (one per ring slot via `output_buffer_index`, so triple buffering keeps distinct slots); `TraceMemoryPool::Free` skips shared buffers to avoid a double-free across the contexts that reuse them.
- Adds a per-agent trace lease in `source/lib/rocprofiler-sdk/thread_trace/shared_trace_lease.{hpp,cpp}` enforcing the "one active trace per agent" invariant the shared buffer relies on: the first active trace on an agent atomically claims it (owner-keyed, re-entrant refcount) in `get_control()`, and a different context requesting a trace while the agent is held fails fast — `get_control()` returns null and that context runs untraced instead of concurrently writing the shared buffer. The lease is released when the agent's last in-flight trace ends (`iterate_data` / `stop` / destructor).
- The buffer and lease managers keep their process-global state in `common::static_object` rather than plain namespace-scope globals: the state is freed by `finalize()` (a `registration::finalize()` `atexit` handler), and on the attach path a plain global's destructor can run before `finalize()` — a use-after-free. `static_object` is destroyed by `destroy_static_objects()` (sequenced after `finalize()`), so the state outlives teardown without leaking.

## JIRA ID

JIRA ID : AIPROFSDK-102

## Test Plan

- Unit: `thread-trace-packet-test` (including a new `shared_buffer_reuse` case covering cross-context reuse, distinct ring slots, and max sizing) and `thread-trace-producer-consumer-test`.
- Integration: ATT single, multi, agent, large-buffer, before-hsa-init, and triple-buffer (consistency, hammer, slow, multiple-cmds) on MI300, ROCm 7.2.3.
- Flakiness: repeated reruns via `ctest --repeat until-fail:10`.

## Test Result

Unit tests pass reliably across repeated runs on MI300 (ROCm 7.2.3): `thread-trace-producer-consumer-test` 25/25 and `thread-trace-packet-test` 10/10. Integration tests single/multi/agent/large-buffer/before-hsa-init and triple-buffer slow/multiple-cmds pass 10/10 repeats each. The `triple-buffer-consistency` and `triple-buffer-hammer` tests can intermittently hit the pre-existing SQTT shutdown GPU hang tracked in AILIKFD-39; this was verified to reproduce on `develop` without this change (identical `reason :GPU Hang`), so it is unrelated to the buffer-sharing rework and is not addressed here.

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/aql_packet.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/aql_packet.cpp
@@ -168,7 +168,9 @@ TraceMemoryPool::Alloc(void** ptr, size_t size, desc_t flags, void* data)
     }
 
     // Device (SQTT output) buffer: the shared manager is the single source, handing out
-    // one buffer per ring slot (output_buffer_index), reused across contexts.
+    // one buffer per ring slot (output_buffer_index), reused across contexts. Marker and
+    // control packets allocate only host memory (above), so this device path is used
+    // solely for SQTT output slots and never aliases them with other allocations.
     const size_t index  = pool.output_buffer_index++;
     void*        shared = thread_trace::acquire_shared_buffer(pool, index, size);
     if(shared == nullptr) return HSA_STATUS_ERROR;
@@ -180,6 +182,8 @@ TraceMemoryPool::Alloc(void** ptr, size_t size, desc_t flags, void* data)
 void
 TraceMemoryPool::Free(void* ptr, void* data)
 {
+    if(ptr == nullptr) return;
+
     // Shared device buffers are owned by the manager (freed once in free_shared_buffers());
     // skip them here to avoid a double free. Only per-call host buffers are freed.
     if(thread_trace::is_shared_buffer(ptr)) return;
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/aql_packet.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/aql_packet.cpp
@@ -157,43 +157,31 @@ TraceMemoryPool::Alloc(void** ptr, size_t size, desc_t flags, void* data)
 
     if(!pool.allocate_fn || !pool.free_fn || !pool.allow_access_fn) return HSA_STATUS_ERROR;
 
-    hsa_status_t status = HSA_STATUS_ERROR;
     if(flags.host_access)
     {
-        status = pool.allocate_fn(pool.cpu_pool_, size, hsa_amd_memory_pool_executable_flag, ptr);
+        auto status =
+            pool.allocate_fn(pool.cpu_pool_, size, hsa_amd_memory_pool_executable_flag, ptr);
 
         if(status == HSA_STATUS_SUCCESS)
             status = pool.allow_access_fn(1, &pool.gpu_agent, nullptr, *ptr);
+        return status;
     }
-    else
-    {
-        // Device (SQTT output) buffer: reuse the per-agent shared buffer across
-        // contexts (one per ring slot via output_buffer_index), falling back to
-        // a private allocation if unavailable.
-        if(thread_trace::has_shared_buffer(pool.gpu_agent))
-        {
-            const size_t index  = pool.output_buffer_index++;
-            void*        shared = thread_trace::acquire_shared_buffer(pool, index);
-            if(shared != nullptr)
-            {
-                *ptr = shared;
-                return HSA_STATUS_SUCCESS;
-            }
-        }
 
-        // Return page aligned data to avoid cache flush overlap
-        status = pool.allocate_fn(
-            pool.gpu_pool_, size + 0x2000, hsa_amd_memory_pool_executable_flag, ptr);
-        *ptr = (void*) ((uintptr_t(*ptr) + 0xFFF) & ~0xFFFul);  // NOLINT(performance-no-int-to-ptr)
-    }
-    return status;
+    // Device (SQTT output) buffer: the shared manager is the single source, handing out
+    // one buffer per ring slot (output_buffer_index), reused across contexts.
+    const size_t index  = pool.output_buffer_index++;
+    void*        shared = thread_trace::acquire_shared_buffer(pool, index, size);
+    if(shared == nullptr) return HSA_STATUS_ERROR;
+
+    *ptr = shared;
+    return HSA_STATUS_SUCCESS;
 }
 
 void
 TraceMemoryPool::Free(void* ptr, void* data)
 {
-    // Shared buffers are owned by the manager and freed once at finalize(); skip
-    // them here to avoid a double-free across the contexts that reuse them.
+    // Shared device buffers are owned by the manager (freed once in free_shared_buffers());
+    // skip them here to avoid a double free. Only per-call host buffers are freed.
     if(thread_trace::is_shared_buffer(ptr)) return;
 
     assert(data);
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/aql_packet.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/aql_packet.cpp
@@ -26,6 +26,7 @@
 #include "lib/rocprofiler-sdk/spm/decode.hpp"
 #include "lib/rocprofiler-sdk/spm/interface.hpp"
 #include "lib/rocprofiler-sdk/thread_trace/dl.hpp"
+#include "lib/rocprofiler-sdk/thread_trace/shared_trace_buffer.hpp"
 
 #include <fmt/core.h>
 #include <cstddef>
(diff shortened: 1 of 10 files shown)


projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/aql_packet.hpp
```diff
@@ -190,6 +191,10 @@ struct TraceMemoryPool
     decltype(hsa_memory_copy)*              api_copy_fn;
 
     aqlprofile_handle_t handle;
+
+    // Ring-slot counter so shared buffers are handed out one per output buffer.
+    mutable uint32_t output_buffer_index = 0;
```
ApoKalipse-V: This mutable needs some explanation

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/shared_trace_buffer.cpp
```diff
@@ -0,0 +1,162 @@
...
+{
+    std::map<uint64_t, agent_buffers_t> agents      = {};       // keyed by hsa_agent_t.handle
+    std::unordered_set<void*>           shared_ptrs = {};       // aligned pointers handed out
+    decltype(hsa_amd_memory_pool_free)* free_fn     = nullptr;  // captured for shutdown free
+};
+
+std::mutex&
+get_mutex()
+{
+    static auto* _mutex = new std::mutex{};
```
ApoKalipse-V: why the static mutex pointer?

ApoKalipse-V (changes requested):
I think we shouldn't have two different code paths, one shared and one non-shared.
Ideally, there should be a single source of allocation. If that source cannot find an existing buffer satisfying the size requirements, then it can create a new shared one.
That way, we dont need to be checking if some pointer is shared.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/shared_trace_buffer.cpp
minseobshin11: was trying to use lazy allocation/construct on first use, but the codebase does have better convention of using static_object<T> - modified.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/aql_packet.hpp
minseobshin11: Thought it is needed since TraceMemoryPool is passed as const, but you are right that it is actually not needed - removed.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/aql_packet.hpp
```diff
@@ -190,6 +191,11 @@ struct TraceMemoryPool
     decltype(hsa_memory_copy)*              api_copy_fn;
 
     aqlprofile_handle_t handle;
+
+    // Ring-slot counter: each device Alloc hands out the next shared buffer slot
+    // (0 = primary, 1.. = double/triple-buffer slots).
+    uint32_t output_buffer_index = 0;
```
ApoKalipse-V, line 199: This should be a property of the buffer, right? Not the pool?

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/aql_packet.cpp
```diff
@@ -155,27 +156,33 @@ TraceMemoryPool::Alloc(void** ptr, size_t size, desc_t flags, void* data)
...
+    *ptr = shared;
+    return HSA_STATUS_SUCCESS;
 }
 
 void
 TraceMemoryPool::Free(void* ptr, void* data)
 {
+    // Shared device buffers are owned by the manager (freed once in free_shared_buffers());
+    // skip them here to avoid a double free. Only per-call host buffers are freed.
+    if(thread_trace::is_shared_buffer(ptr)) return;
```
ApoKalipse-V, line 189: If you need this, the model is probably not quite right.
See if you can simplify such that the shared (gpu_pool) vs non-shared (cpu_pool) memory never mix like this.

it's fine if you cant.

ApoKalipse-V (changes requested):

ApoKalipse-V (changes requested):
One more thing: We need a integration test that allocates hundreds of contexts (say with varying buffer size, target_cu, se_mask, simd_select, etc), runs each once and ensures it's all good. Note that not every target_cu is guaranteed to work, so the condition should be that "most" configs work for CUs in [0,1,2,3]
Allocating ~400 contexts of 1GB each would be impossible in the old implementation, so we can know for sure this works.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/aql_packet.hpp
minseobshin11: at the time of Alloc, the buffer doesn't exist yet and the pool is the aqlprofile's userdata (per-session context)

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/aql_packet.cpp
minseobshin11: I explored few approaches, but I dont think I can because both cpu and gpu pool buffers are allocated and freed through the same aqlrprofile callback pair. While the Alloc can distinguish the kind through passing in the flag, the free doesn't have a flag or anything to tell which kind to free. So i don't think it's possible to remove the check unless we modify aqlprofile itself

minseobshin11:
> One more thing: We need a integration test that allocates hundreds of contexts (say with varying buffer size, target_cu, se_mask, simd_select, etc), runs each once and ensures it's all good. Note that not every target_cu is guaranteed to work, so the condition should be that "most" configs work for CUs in [0,1,2,3] Allocating ~400 contexts of 1GB each would be impossible in the old implementation, so we can know for sure this works.

~400 contexts is not possible due to each context alloacating hsa queue, created a PR that addresses this issue. https://github.com/ROCm/rocm-systems/pull/8644

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/shared_trace_buffer.cpp
ApoKalipse-V: That is not what I meant - Is not affected by the global destructor in the same way.

minseobshin11: oh yea the previous approach did leak the new std::mutex and never destroyed it - I've changed it to static_object<T> which gets destructed after finalize()

ApoKalipse-V: its fine but unnecessary. The mutex does not need to be a pointer or a static_object , it can be just a regular object.

ApoKalipse-V (changes requested):
Changes looks good, but we are missing two things:
(1) A lock to ensure no two contexts can use the same buffer at the same time. Right now, there is nothing preventing that except context management. We need stronger enforcement.
We have edge cases such as "disable dispatch thread trace while a kernel is running, then enable device thread trace".
(2) The CPU buffers also need to be reused. This can be a distinct PR (quick scan was merged only recently), but ideally reusing the same locking mechanism.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/shared_trace_buffer.cpp
minseobshin11: keeping it static_object to outlive finalize() - regular namespace scoped mutex (Sychronized) can be destroyed before finalization and was failing attach-once-att

minseobshin11:
added a per-agent trace lease - the first active trace on an agent claims it and if a different context tries to trace the same agent, returns null.

PR #8274, opened 2026-07-08
feat(hotswap): cache retarget results per code object

suryajasper:
## Summary

Cache the result of `RetargetCodeObject` in `TryRetargetCodeObject` so that when the same code object is loaded for multiple GPUs of the same stepping, the expensive COMGR retarget (dominated by `decodeTextSection` disassembly, ~40s for RCCL's 212MB code object) runs once; subsequent loads return a shared reference to the cached result. Failed rewrites are also cached to skip known-failing decodes.

JIRA ID: DCGPUBU-1871

Note: DCGPUBU-1871 also reports SDMA timeout failures that occur without hotswap enabled and are a separate platform issue. This change addresses only the redundant LLVM disassembly bottleneck observed during multi-GPU runs with hotswap enabled.

### Context

4-GPU RCCL `all_reduce_perf` hangs during `ncclCommInitRank` because RCCL initializes GPUs sequentially on a single host thread, and each GPU triggers a full COMGR retarget of the same code object. With N GPUs, total retarget time = N × T(rco). For 4 GPUs with RCCL's large code objects, this exceeds the host's tolerance and wedges the machine (SSH unreachable, requires AC cycle). See [HSV-004](https://amd.atlassian.net/wiki/spaces/MLSE/pages/1778595000).

Note: an earlier version of this description incorrectly attributed the serialization to a HIP mutex. Analysis by @dayatsin-amd confirmed that no global mutex serializes the retarget across GPUs — the sequencing is application-level (RCCL's single-threaded init loop).

### Changes

| File | Change |
|------|--------|
| `projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp` | Add FNV-1a hash, `shared_ptr`-backed cache, and lookup/store logic in `TryRetargetCodeObject` |
| `projects/rocr-runtime/runtime/hsa-runtime/core/inc/hotswap.hpp` | Add `RetargetCacheSizeForTesting` and `ClearRetargetCacheForTesting` declarations |
| `projects/rocr-runtime/rocrtst/suites/functional/hotswap/hotswap_rewrite_test.cc` | Add `RetargetCacheServesSecondLoadFromCache` and `RetargetCacheClearResetsCacheSize` tests |

### Design

- **Cache location:** `TryRetargetCodeObject` in `rocr::hotswap`, before calling `RetargetCodeObject`
- **Cache key:** FNV-1a hash of (code object bytes, source ISA string, target ISA string, entry trampolines flag)
- **Cache value:** `shared_ptr<vector<uint8_t>>` for success (ref-counted, large copies outside mutex), or failure sentinel for deterministic COMGR errors
- **Cache lifetime:** Process-scoped static, no eviction
- **Thread safety:** Dedicated mutex for cache map; only map lookup/insert under lock, large `memcpy` performed after releasing the mutex via `shared_ptr` handle
- **Failure caching:** Only deterministic COMGR failures are cached; transient allocation failures are not, so later attempts can succeed

### Verified behavior

On H21-9 (4× gfx1250 A0) with RCCL `all_reduce_perf`:
- Code objects confirmed byte-identical across GPUs (diff of rewrite logs: IDENTICAL)
- Single-GPU: 29 rewrites, 0 cache hits (all misses on first load), rc=0 in 45s
- 4-GPU with cache: 1 rewrite + 3 cache hits on the first code object

## Test plan

- [x] Single-GPU `all_reduce_perf` passes with cache enabled (rc=0, correct results)
- [x] Cache hit/miss counts verified via `HSA_HOTSWAP_VERBOSE=1` logs
- [x] Code objects confirmed identical across GPUs
- [x] Unit tests: `RetargetCacheServesSecondLoadFromCache`, `RetargetCacheClearResetsCacheSize`
- [ ] 4-GPU `all_reduce_perf` — host wedge persists (root cause is IOMMU/platform, not retarget time; see HSV-004 open questions)


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -69,6 +70,37 @@ namespace {
...
+  hash ^= FnvHash(source_isa.data(), source_isa.size()) * 31;
+  hash ^= FnvHash(target_isa.data(), target_isa.size()) * 37;
+  hash ^= entry_trampolines ? 0x1ULL : 0x0ULL;
+  return hash;
+}
+
+struct CachedRetargetResult {
+  bool succeeded = false;
+  std::vector<uint8_t> elf_bytes;
+};
```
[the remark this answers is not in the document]
suryajasper: Addressed. Changed `CachedRetargetResult::elf_bytes` from `std::vector<uint8_t>` to `std::shared_ptr<std::vector<uint8_t>>`. On cache hit, we grab the shared_ptr under the lock (atomic ref increment), release the lock, then do the malloc + memcpy outside the critical section.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -454,11 +486,55 @@ bool TryRetargetCodeObject(const CodeObjectView& code_object, hsa_agent_t agent,
...
+          return true;
+        }
+      } else {
+        HOTSWAP_LOG("hotswap: cache hit (failed) src=%s tgt=%s entry_trampolines=%d "
+                    "in=%zu\n",
+                    decision->source_isa.c_str(), decision->target_isa.c_str(),
+                    decision->request_entry_trampolines, code_object.size);
+        return false;
+      }
+    }
```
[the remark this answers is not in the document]
suryajasper: Addressed — same fix as above. The cache-hit path now grabs a `shared_ptr` under the mutex and performs the large copy after releasing it.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -454,11 +486,55 @@ bool TryRetargetCodeObject(const CodeObjectView& code_object, hsa_agent_t agent,
...
+  {
+    std::scoped_lock lock(g_retarget_cache_mutex);
+    CachedRetargetResult entry;
+    entry.succeeded = rewritten;
+    if (rewritten) {
+      const auto* data = static_cast<const uint8_t*>((*out_elf_buffer).get());
+      entry.elf_bytes.assign(data, data + *out_elf_size);
+    }
+    g_retarget_cache.emplace(cache_key, std::move(entry));
+  }
```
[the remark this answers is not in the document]
suryajasper: Addressed. Only deterministic COMGR failures (where `RetargetCodeObject` returned false) are cached. The only transient failure path is `malloc` returning null in the cache-hit copy, which happens after the cache lookup and does not write to the cache — the function falls through to the uncached retarget path. Added a comment documenting this policy. The store path now uses `std::make_shared` and the lock is held only for the `emplace`.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -69,6 +70,37 @@ namespace {
...
+  return hash;
+}
+
+uint64_t ComputeRetargetCacheKey(const void* elf_data, size_t elf_size,
+                                 const std::string& source_isa,
+                                 const std::string& target_isa,
+                                 bool entry_trampolines) {
+  uint64_t hash = FnvHash(elf_data, elf_size);
+  hash ^= FnvHash(source_isa.data(), source_isa.size()) * 31;
```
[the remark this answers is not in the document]
suryajasper: Acknowledged. FNV-1a over 212MB is ~5ms, not microseconds. This is still an 8000x improvement over the 40s retarget it avoids. A pointer-based fast path was considered but deferred — pointer identity is unreliable across code object loads since HIP may mmap different copies of the same content. ELF build-id would be cheaper but not all code objects carry one.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -69,6 +70,37 @@ namespace {
...
+}
+
+struct CachedRetargetResult {
+  bool succeeded = false;
+  std::vector<uint8_t> elf_bytes;
+};
+
+std::mutex g_retarget_cache_mutex;
+std::unordered_map<uint64_t, CachedRetargetResult> g_retarget_cache;
+
```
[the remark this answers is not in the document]
suryajasper: Addressed. Added `RetargetCacheSizeForTesting()` and `ClearRetargetCacheForTesting()` hooks under `ROCR_HOTSWAP_TESTING` in both `hotswap.cpp` and `hotswap.hpp`. Added two tests in `hotswap_rewrite_test.cc`: `RetargetCacheServesSecondLoadFromCache` (loads same code object twice, asserts cache size doesn't grow on second load) and `RetargetCacheClearResetsCacheSize` (verifies cache clear hook).

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -454,11 +498,70 @@ bool TryRetargetCodeObject(const CodeObjectView& code_object, hsa_agent_t agent,
...
                          out_elf_buffer, out_elf_size,
                          decision->request_entry_trampolines);
+
+  // Cache the result. Only deterministic COMGR failures are cached;
+  // transient allocation failures in this function are not, so a
+  // later attempt with the same code object can still succeed.
+  {
+    std::scoped_lock lock(g_retarget_cache_mutex);
+    CachedRetargetResult entry;
+    entry.succeeded = rewritten;
```

**[Important]** Both `std::make_shared<std::vector<uint8_t>>(...)` here and `g_retarget_cache.emplace(...)` two lines below can throw `std::bad_alloc`. If either throws after `RetargetCodeObject` succeeds, the exception propagates through `TryRetargetCodeObject` — ROCr's C-API surface doesn't expect exceptions, so this would likely terminate the process under memory pressure, and the successful retarget result is lost.

Fix: wrap the entire store block in `try { ... } catch (...) {}` and skip caching on OOM while returning the already-computed `rewritten` result.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -69,6 +71,48 @@ namespace {
...
+                                 bool entry_trampolines) {
+  uint64_t hash = FnvHash(elf_data, elf_size);
+  hash ^= FnvHash(source_isa.data(), source_isa.size()) * 31;
+  hash ^= FnvHash(target_isa.data(), target_isa.size()) * 37;
+  hash ^= entry_trampolines ? 0x1ULL : 0x0ULL;
+  return hash;
+}
+
+struct CachedRetargetResult {
+  bool succeeded = false;
```
cfreeamd, line 112: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** `g_retarget_cache` has no eviction policy — it grows with every distinct code object loaded across the process lifetime. In applications with many unique code objects, cached ELF buffers (potentially MB each) accumulate indefinitely. A simple entry cap (`if (g_retarget_cache.size() >= MAX_ENTRIES) return;` before the emplace) would bound the growth without LRU complexity.

cfreeamd (commented):

`hotswap.hpp` and `hotswap_rewrite_test.cc` look correct. One issue in `hotswap.cpp` cache-store path noted inline.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
suryajasper: Addressed in dc12cbab. Wrapped the cache store block in `try { ... } catch (const std::bad_alloc&) {}`. If `make_shared` or `emplace` throws under OOM, the successfully retargeted code object in `out_elf_buffer` is still returned to the caller — we just skip caching it.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
suryajasper: Addressed in dc12cbab. Added `constexpr size_t kMaxRetargetCacheEntries = 256` and a size guard before the emplace. When the cache is full, new entries are silently skipped; existing cache hits still work. 256 is well above typical workloads (~29 for RCCL) while bounding memory for JIT-heavy applications.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -69,6 +71,50 @@ namespace {
...
+}
+
+uint64_t ComputeRetargetCacheKey(const void* elf_data, size_t elf_size,
+                                 const std::string& source_isa,
+                                 const std::string& target_isa,
+                                 bool entry_trampolines) {
+  uint64_t hash = FnvHash(elf_data, elf_size);
+  hash ^= FnvHash(source_isa.data(), source_isa.size()) * 31;
+  hash ^= FnvHash(target_isa.data(), target_isa.size()) * 37;
+  hash ^= entry_trampolines ? 0x1ULL : 0x0ULL;
```
shwetagkhatri: This will only flip the LSB of final hash. Even though collision probability is low, its better to change this to something like: 
hash ^= entry_trampolines ? 0xDEADBEEF12345678ULL : 0x0ULL;

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -454,11 +500,75 @@ bool TryRetargetCodeObject(const CodeObjectView& code_object, hsa_agent_t agent,
     return false;
   }
 
+  const uint64_t cache_key = ComputeRetargetCacheKey(
+      code_object.data, code_object.size, decision->source_isa,
+      decision->target_isa, decision->request_entry_trampolines);
+
+  // Cache lookup: grab a shared_ptr under the lock, then copy outside it.
+  {
```
shwetagkhatri, line 555: When two threads simultaneously load the same code object:
  1. Thread A: cache miss, starts COMGR
  2. Thread B: cache miss, starts COMGR
  3. Thread A: finishes, inserts to cache
  4. Thread B: finishes, emplace() is no-op (key exists)

Both perform expensive COMGR work. Thread B's result is discarded.
Its OK correctness wise but wastes CPU cycles. There can be optimization made here. But its not a blocker. Can be done in a new PR also.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -454,11 +500,75 @@ bool TryRetargetCodeObject(const CodeObjectView& code_object, hsa_agent_t agent,
...
                          decision->source_isa.c_str(), decision->target_isa.c_str(),
                          out_elf_buffer, out_elf_size,
                          decision->request_entry_trampolines);
+
+  // Cache the result. Only deterministic COMGR failures are cached;
+  // transient allocation failures in this function are not, so a
+  // later attempt with the same code object can still succeed.
+  try {
+    std::scoped_lock lock(g_retarget_cache_mutex);
+    if (g_retarget_cache.size() < kMaxRetargetCacheEntries) {
```
shwetagkhatri: With 256 entries of 212MB each (RCCL-sized code objects we have seen), theoretical max is 54GB. It looks like there is no eviction policy so cache will continue to grows monotonically until process exits and we can potentially run out of memory.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -454,11 +500,75 @@ bool TryRetargetCodeObject(const CodeObjectView& code_object, hsa_agent_t agent,
...
+      CachedRetargetResult entry;
+      entry.succeeded = rewritten;
+      if (rewritten) {
+        const auto* data = static_cast<const uint8_t*>((*out_elf_buffer).get());
+        entry.elf_bytes = std::make_shared<std::vector<uint8_t>>(data, data + *out_elf_size);
+      }
+      g_retarget_cache.emplace(cache_key, std::move(entry));
+    }
+  } catch (const std::bad_alloc&) {
+    // OOM during caching — the retarget result in out_elf_buffer is
```
shwetagkhatri: We should add a log here for debugging.

shwetagkhatri (changes requested):
Any new tests need to be added to -  rocrtst/suites/test_common/test_categories.yaml

These tests
  - TEST(HotswapRewrite, RetargetCacheServesSecondLoadFromCache)
  - TEST(HotswapRewrite, RetargetCacheClearResetsCacheSize)

are not registered in test_categories.yaml.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
dayatsin-amd: I think doing the XOR with 0x1 does not affect whether the keys are distinct. This part looks ok to me.

dayatsin-amd (approved):

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
dayatsin-amd: In the PR description, there this this mention "With HIP's code-object-loading mutex serializing across GPUs".
This seems to imply this code was written with the assumption that there is a mutex to serialize code object loading accross GPUs.

When using AI to analyze, it seems there is no mutex to protect this. So it seems this is a real concern and should be addressed. 

@suryajasper  can double check this on your side? 

This was the analysis summary from my side:

> Here's a short summary of everything I inspected to conclude no mutex serializes the hotswap retarget across GPUs.
> 
> HIP (projects/clr/hipamd/src)
> hip_code_object.cpp — StatCO::GetFunc: the static-fatbin path (RCCL). Its sclock_ lock wraps only DigestFatBinary (fatbin extraction) and is released before the per-device build/load that triggers the retarget.
> hip_global.cpp — Function::GetStatFunc: called outside sclock_; invokes BuildProgram(deviceId) → the actual retarget path.
> hip_fatbin.cpp — FatBinaryInfo::BuildProgram/AddDevProgram: each device gets its own amd::Program, so no shared lock; dev_programs_[id]->load() is what reaches ROCr.
> hip_code_object.hpp: StatCO::sclock_ and DynCO::dclock_ are the only relevant mutexes. dclock_ does wrap the retarget, but only for the dynamic hipModuleLoad path — not RCCL.
> CLR / ROCclr (projects/clr/rocclr)
> platform/program.cpp / program.hpp: Program::build/load take programLock_, but it's a per-amd::Program member (std::recursive_mutex), so distinct per device — no cross-GPU serialization.
> device/rocm/rocprogram.cpp: the Hsa::executable_load_agent_code_object call site — no lock.
> ROCr (projects/rocr-runtime/runtime/hsa-runtime)
> core/runtime/hsa.cpp — hsa_executable_load_agent_code_object: only TRY (exception guard) and IS_OPEN() (open-flag check) before calling LoadAgentCodeObjectWithHotswap — no mutex around the retarget.
> loader/executable.cpp / executable.hpp: ExecutableImpl::LoadCodeObject takes a ReaderWriterLock rw_lock_, but it's a per-ExecutableImpl member and runs after the retarget.
> COMGR / LLVM (/home/dayatsin/devel/llvm-project/amd/comgr/src)
> comgr.cpp: the two only mutexes — ComgrMutex (global "one LLVM at a time") is scoped entirely inside amd_comgr_do_action, and LlvmInitMutex guards one-time init. The hotswap APIs use neither.
> comgr-hotswap.cpp: amd_comgr_hotswap_rewrite[_with_options] bypass amd_comgr_do_action, calling hotswapRewrite → retargetCodeObject directly, unlocked.
> comgr-hotswap-llvm.cpp: only a magic-static (getAmdgcnTarget) for one-time target resolution; comments confirm no per-call mutex. The disassembly/raiser build per-call MC objects with no lock (grep across amd/comgr/src found no other mutexes).
> Conclusion: every lock in the load path is either per-object (per-Program/per-Executable) or scoped away from the retarget; the only global COMGR lock doesn't cover the hotswap entry points. So nothing serializes the retarget across GPUs — the observed sequencing comes from RCCL initializing GPUs sequentially on one host thread.
> 
>

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -69,6 +71,50 @@ namespace {
...
+// Only deterministic failures (COMGR returned an error) are cached; transient
+// allocation failures are not, so a later attempt can succeed.
+
+uint64_t FnvHash(const void* data, size_t size) {
+  constexpr uint64_t kFnvOffset = 14695981039346656037ULL;
+  constexpr uint64_t kFnvPrime = 1099511628211ULL;
+  uint64_t hash = kFnvOffset;
+  const auto* bytes = static_cast<const uint8_t*>(data);
+  for (size_t i = 0; i < size; ++i) {
+    hash ^= bytes[i];
```
scchan, line 93: I think this loop should compute 64-bit instead of 8-bit per iteration; otherwise, the XOR is only modifying the lower 8-bit and not taking advantage of the entire 64-bit.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -454,11 +500,75 @@ bool TryRetargetCodeObject(const CodeObjectView& code_object, hsa_agent_t agent,
     return false;
   }
 
+  const uint64_t cache_key = ComputeRetargetCacheKey(
+      code_object.data, code_object.size, decision->source_isa,
+      decision->target_isa, decision->request_entry_trampolines);
```
scchan: `target_isa` is the translated ISA?  Ideally, we want to be able to compute the hash key with only the source and to avoid the expensive translation if it's already in the cache.

scchan (changes requested):

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
suryajasper: Addressed in ddde27db89. Changed to `0xDEADBEEF12345678ULL` as suggested.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
suryajasper: Addressed in ddde27db89. Added `HOTSWAP_LOG` after successful cache store and in the `bad_alloc` catch block.

suryajasper:
@shwetagkhatri Tests registered in ddde27db89 — added `HotswapRewrite.RetargetCacheServesSecondLoadFromCache` and `HotswapRewrite.RetargetCacheClearResetsCacheSize` to all four tiers (quick, standard, comprehensive, full) in `test_categories.yaml`.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
suryajasper: The byte-by-byte XOR is the standard FNV-1a specification — the multiply by the 64-bit FNV prime (`0x00000100000001B3`) after each byte-XOR propagates bits across the full 64-bit accumulator via carry chains. This is how FNV-1a achieves its avalanche property: after processing just a few bytes, all 64 bits of the hash are influenced by every input byte. The fact that the XOR touches only the low 8 bits per iteration is by design, not a limitation.

That said, FNV-1a's throughput is limited to ~1 byte/cycle due to the loop-carried multiply dependency, which gives ~0.5-1 GB/s. For our 578 MB code objects that's ~5ms per hash — acceptable given it avoids a 103s retarget (8000x improvement), but not free.

If throughput becomes a concern, the principled upgrade would be to replace FNV-1a with XXH64, a hash specifically designed for bulk data (~10-20 GB/s, formally analyzed collision properties, BSD-2-licensed). This would require either vendoring the ~6000-line xxHash header or writing a ~65-line portable implementation from the [public spec](https://github.com/Cyan4973/xxHash/blob/dev/doc/xxhash_spec.md). Happy to do this as a follow-up if there's interest, but for this PR the FNV-1a cost is well within budget.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
suryajasper: `decision->target_isa` is derived by `DecideHotswapRewrite` from `GetAgentIsaName(agent)`. For the B0→A0 path it's always `WithGfx1250SteppingFeature(target_isa, kA0)`, so it is deterministic for all agents of the same stepping within a process.

However, removing it from the key wouldn't let us skip meaningful work earlier. The call sequence before the cache lookup is: `GetCodeObjectIsaName` (COMGR ELF parse) → `GetAgentIsaName` (agent query) → `DecideHotswapRewrite` (string comparisons) → cache lookup. The first two calls are needed regardless to determine eligibility, and `DecideHotswapRewrite` itself is just string comparisons (microseconds). Moving the cache check before `DecideHotswapRewrite` would save negligible time while removing a safety guard against hypothetical mixed-stepping configurations.

Prefer to keep `target_isa` in the key for defense-in-depth — the cost is one FNV-1a over a ~40-byte string. Happy to add a comment documenting why it's included.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
suryajasper: Agreed this is a real but non-blocking concern. For the primary use case (RCCL), GPUs are initialized sequentially on a single host thread, so duplicate retargets don't occur in practice. @dayatsin-amd's analysis below also confirmed that HIP doesn't hold a global mutex during the retarget — the serialization we observed is from RCCL's sequential init, not a runtime guarantee.

For frameworks that init GPUs on parallel threads, the fix would be a `std::shared_future` pending sentinel in the cache map — on miss, thread A inserts a "retarget in progress" marker; thread B finds it and waits instead of launching a redundant COMGR retarget. ~30-40 lines + an RAII guard to ensure the promise is always fulfilled. Will address in a follow-up PR.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
suryajasper: Valid concern. The 256-entry cap bounds count but not bytes — one 578 MB entry and one 30 KB entry count the same, so the theoretical max is 256 × 578 MB ≈ 148 GB.

The fix is to replace the entry-count cap with a byte-budget cap (e.g., 2 GB) tracking total cached ELF bytes, plus LRU eviction so stale entries get pushed out when the budget is full. Implementation is straightforward (~30-40 lines): a `std::list<uint64_t>` for LRU ordering alongside the existing `unordered_map`, `splice` to front on cache hit (O(1)), evict from back on insert until there's room. All under the existing mutex — no new concurrency concerns.

LRU eviction also layers cleanly with the disk-persistent cache tier in PR #8459 (which stacks on this PR). With the disk tier in place, the in-memory cache becomes a bounded hot tier: an entry evicted from in-memory by LRU isn't lost — it lives on disk and gets re-promoted on next access (~6s disk read vs 103s COMGR retarget). Without LRU, entries would stop being cached entirely once the budget fills, and every new distinct code object would pay full COMGR cost even if stale entries are occupying the budget.

Plan is to implement the byte-budget + LRU in #8459 alongside the disk tier, so both cache tiers land together with proper memory management. For this PR, the 256-entry cap provides a reasonable interim backstop — typical workloads (e.g., RCCL with ~29 distinct code objects) are well under 256.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
suryajasper: You're right — the PR description's claim about "HIP's code-object-loading mutex serializing across GPUs" is inaccurate. Your analysis confirms that `sclock_` is released before per-device build/load, and `programLock_` is per-`amd::Program`. The observed serialization comes from RCCL initializing GPUs sequentially on a single host thread, not a HIP mutex.

Updated the PR description to reflect this. The cache is correct and valuable regardless — it eliminates redundant retargets whether the serialization comes from a mutex or sequential application-level init.

scchan (approved):

shwetagkhatri (approved):
LGTM.

PR #8288, opened 2026-07-08
refactor(rocdbgapi): unify interface for make_gfx*_cwsr_record

lancesix:
## Motivation

In the current implementation, different architectures need to create CWSR record from different amount of data.  On gfx9, we have one dword of wave state, and one dword of common state (state common to multiple waves in a workgroup).  On gfx10 architectures, the common state is 2 dwords instead.  This lead to the introduction of two different functions: make_gfx9_cwsr_record and make_gfx1x_cwsr_record.

## Technical Details

This patch proposed to have a unified make_cwsr_record, which a const reference to a vector vector of dwords for shared state.  This way, the same interface can be used for all supported architectures.

Ideally, the interface should use a std::span, but this is c++20, so not available to us yet.  Instead the const ref to a vector is the closest approximation we have.

## JIRA ID

JIRA ID: AIROCGDB-1234

## Test Plan

Use ROCgdb's testsuite on gfx1031, and use CI to run on gfx940.

## Test Result

PASS for gfx1031, no observable change.

## Submission Checklist

### DIFF
--- a/projects/rocdbgapi/src/architecture.cpp
+++ b/projects/rocdbgapi/src/architecture.cpp
@@ -2496,13 +2496,14 @@ protected:
   };
 
   virtual std::unique_ptr<architecture_t::cwsr_record_t>
-  make_gfx9_cwsr_record (compute_queue_t &queue, uint32_t xcc_id,
-                         uint32_t compute_relaunch_wave,
-                         uint32_t compute_relaunch_state,
-                         agent_address_t context_save_address) const
+  make_cwsr_record (compute_queue_t &queue, uint32_t xcc_id,
+                    uint32_t compute_relaunch_wave,
+                    const std::vector<uint32_t> &compute_relaunch_state,
+                    agent_address_t context_save_address) const
   {
+    dbgapi_assert (compute_relaunch_state.size () == 1);
     return std::make_unique<cwsr_record_t> (
-      queue, xcc_id, compute_relaunch_wave, compute_relaunch_state,
+      queue, xcc_id, compute_relaunch_wave, compute_relaunch_state[0],
       context_save_address);
   }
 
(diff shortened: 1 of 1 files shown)


projects/rocdbgapi/src/architecture.cpp
```diff
@@ -2496,13 +2496,14 @@ class gfx9_architecture_t : public amdgcn_architecture_t
   };
 
   virtual std::unique_ptr<architecture_t::cwsr_record_t>
-  make_gfx9_cwsr_record (compute_queue_t &queue, uint32_t xcc_id,
```
aktemur, line 2499: In the commit message, there is a typo: "a vector vector of dwords"

aktemur (approved):

projects/rocdbgapi/src/architecture.cpp
sebdar, line 2404: Could make the cwsr record own a copy of the relaunch state vector of uint32_t to make it a bit more generic ?

sebdar, line 2452: And thus simply pass the vector to the ctor.

sebdar (approved):
This looks good. Just a minor comment below, I think we might be able to simplify the ctors / general layout further for cwsr records, but it's up to you

projects/rocdbgapi/src/architecture.cpp
lancesix: I thought about that, but in the end, it seemed to make more clutter than necessary.  For a given architecture, it's CWSR record has a fixed number of dword of data associated to it.  Holding this in a vector would mean holding a vector (with dynamic allocation, need to call the dtor and so on) where the plain fild works as well and is trivially destructible.

And for now, this also minimize the change as we keep what already exist that a gfx10_architecture_t inherits from a gfx9_architecture_t and its 1 dword state, just adds the next one.

But yes, if this becomes too cumbersome in the ctor, we could change this.

PR #8290, opened 2026-07-08
fix(rccl): IR - Fix WarpSpan named-barrier LDS in bitcode

PedramAlizadeh:
Making ncclCoopWarpSpan named-barrier LDS kernel-owned in bitcode:
ncclCoopSync on a multi-warp WarpSpan aborted with HSA_STATUS_ERROR_EXCEPTION (0x1016) when driven through librccl_device.bc. The AMD software named barrier kept its scratch in a function-local __shared__ (ncclCoopNamedBarrierState), and in the bitcode binding that LDS-using sync() is reached only indirectly via the ncclCoopAny vtable. AMDGPU sizes a kernel's group segment from the static call graph, so an indirect-only path leaves the launching kernel with group_seg_size=0 and the runtime LDS access faults. Let the launching kernel own the named-barrier scratch and pass it in:
- coop.h: ncclCoopWarpSpan gains an (AMD-only, defaulted) slots pointer; sync() uses it directly in the bitcode build (__clang_llvm_bitcode_lib__) and the function-local __shared__ is compiled out there, so the .bc emits no __shared__ (LDS, addrspace(3)). Native/in-tree builds keep the internal LDS via a fallback, so existing kernels are unchanged. Adds a size static_assert.
- nccl_device_wrapper{,__impl}.h: ncclCoopAnyInitWarpSpan takes a barrierLds pointer and threads it into the coop.
- IR_test.cpp: k_sync_warpspan declares the __shared__ slots and passes them. Verified: rebuilt bitcode has 0 __shared__ (LDS, addrspace(3) globals; B6 CoopSync passes with a clean log; host librccl.so still links.

## Motivation

The RCCL device-API bitcode binding (librccl_device.bc) exposes ncclCoopSync so kernels can synchronize a cooperative group without inlining NCCL's device headers. On AMD GPUs a multi-warp ncclCoopWarpSpan sync is emulated with a software named barrier, and any kernel that drove that sync through the bitcode aborted with HSA_STATUS_ERROR_EXCEPTION (code 0x1016). A hard hardware fault that crashed the launching kernel and made the sub-block WarpSpan barrier unusable via the binding. This blocked the coop-sync unit test (B6_CoopSync) and, more importantly, any downstream consumer that relies on the bitcode for WarpSpan/CTA-level synchronization, so the binding could not be trusted for real multi-warp workloads.

## Technical Details

The named barrier kept its state in a function-local __shared__ array (ncclCoopNamedBarrierState), i.e. a shared-memory/LDS (addrspace(3)) global baked into the bitcode. Shared memory is a per-workgroup resource that must be reserved at launch (the group_seg_size field), and AMDGPU determines the amount by walking the static call graph from each kernel to every LDS variable it can reach. In the binding, ncclCoopAny is type-erased, so the LDS-using sync() is reached only through an indirect vtable function pointer; the static walk can't follow that edge, so the kernel launched with group_seg_size=0 and the barrier's shared-memory access ran off unreserved LDS, triggering 0x1016. The fix makes the launching kernel own the scratch: ncclCoopWarpSpan gains an (AMD-only, defaulted) slots pointer, ncclCoopAnyInitWarpSpan takes a barrierLds argument, and in the bitcode build (__clang_llvm_bitcode_lib__) sync() dereferences that caller-provided pointer while the internal __shared__ is compiled out — so the .bc emits no LDS global at all and there is nothing for the sizing pass to miss. Native/in-tree builds keep the function-local __shared__ via a fallback path (the pointer defaults to the internal state), leaving existing kernels and the CUDA path byte-for-byte unchanged; a static_assert guarantees the added pointer still fits ncclCoopAny's 16-byte type-erased storage.

## JIRA ID

JIRA ID : AICOMRCCL-1513

## Test Plan

-Build RCCL with the LLVM IR bitcode enabled (-DEMIT_LLVM_IR=ON -DBITCODE_LIB_ARCH=gfx942) so librccl_device.bc and the hipify-staged headers exist.
-Establish the baseline (pre-fix) failure by running the IR device suite and confirming B6_CoopSync faults on the multi-warp WarpSpan case.
-Apply the fix, re-stage the nccl_device headers, and rebuild the bitcode.
-Inspect the rebuilt librccl_device.bc to confirm it no longer contains a __shared__/LDS (addrspace(3)) global.
-Re-run the full IR device suite (pytest) on an MI300X (gfx942) and confirm all cases pass, with the B6_CoopSync log clean of the hardware fault.
-Rebuild the host librccl.so to confirm the modified coop.h still compiles on the native/in-tree path (unchanged behavior for existing kernels + the new static_assert holds).

## Test Result

Test results (MI300X, gfx942, ROCm 7.2, branch on latest develop)

-Baseline (before fix): 3 passed, 1 failed. B6_CoopSync failed on k_sync_warpspan with HSA_STATUS_ERROR_EXCEPTION code: 0x1016 and group_seg_size=0.
-Bitcode after fix: addrspace(3) global count = 0 (the __shared__/LDS global is gone from the .bc).
-IR device suite after fix: 4 passed, 0 failed — test_get_peer_pointer_team, test_coop_init_and_accessors, test_coop_sync (B6), test_lsa_barrier_session all PASS; B6_CoopSync log is clean of 0x1016.
-Host library: librccl.so re-links successfully; host translation units that include coop.h recompile cleanly, confirming the native fallback path and the static_assert(sizeof(ncclCoopWarpSpan) <= 16) both hold.

## Submission Checklist

### DIFF
--- a/projects/rccl/bindings/ir/nccl_device_wrapper.h
+++ b/projects/rccl/bindings/ir/nccl_device_wrapper.h
@@ -114,7 +114,7 @@ struct ncclLsaBarrierSession_C {
 NCCL_IR_EXPORT void ncclCoopAnyInitThread(ncclCoopAny* coop);
 NCCL_IR_EXPORT void ncclCoopAnyInitWarp(ncclCoopAny* coop);
 NCCL_IR_EXPORT void ncclCoopAnyInitLanes(ncclCoopAny* coop, ncclCoopMask_t lane_mask);
-NCCL_IR_EXPORT void ncclCoopAnyInitWarpSpan(ncclCoopAny* coop, int warp0, int nWarps, int id);
+NCCL_IR_EXPORT void ncclCoopAnyInitWarpSpan(ncclCoopAny* coop, int warp0, int nWarps, int id, void* barrierLds);
 NCCL_IR_EXPORT void ncclCoopAnyInitCta(ncclCoopAny* coop);
 
 NCCL_IR_EXPORT int  ncclCoopThreadRank(const ncclCoopAny* coop);
--- a/projects/rccl/bindings/ir/nccl_device_wrapper__impl.h
+++ b/projects/rccl/bindings/ir/nccl_device_wrapper__impl.h
@@ -74,8 +74,16 @@ NCCL_IR_EXPORT void ncclCoopAnyInitWarp(ncclCoopAny* coop) {
 NCCL_IR_EXPORT void ncclCoopAnyInitLanes(ncclCoopAny* coop, ncclCoopMask_t lane_mask) {
   ::new (coop) ncclCoopAny(ncclCoopLanes(lane_mask));
 }
-NCCL_IR_EXPORT void ncclCoopAnyInitWarpSpan(ncclCoopAny* coop, int warp0, int nWarps, int id) {
+NCCL_IR_EXPORT void ncclCoopAnyInitWarpSpan(ncclCoopAny* coop, int warp0, int nWarps, int id, void* barrierLds) {
+#if defined(__HIP_PLATFORM_AMD__)
+  /* barrierLds is a kernel-owned __shared__ slot array; the bitcode build has no
+   * function-local LDS of its own (see coop.h), so the WarpSpan barrier uses this. */
+  ::new (coop) ncclCoopAny(ncclCoopWarpSpan(warp0, nWarps, id,
+                           static_cast<ncclCoopNamedBarrierSlot*>(barrierLds)));
+#else
+  (void)barrierLds;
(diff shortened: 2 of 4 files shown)


Kapil-Shyam-Pawar (approved):

PedramAlizadeh:
After applying the PR #8290 fix, the cooperative-group (ncclCoopAny) tests pass on the gfx950. Including test_coop_sync (B6: ncclCoopSync across all five coop types), which previously crashed on the multi-warp WarpSpan case with an LDS fault (HSA 0x1016). The fix makes the launching kernel own the named-barrier LDS and pass it into the bitcode thunk; the rebuilt librccl_device.bc now has 0 addrspace(3) (LDS) globals and B6 completes cleanly, bringing the IR-device suite to complete 9/9 passing.

PR #8291, opened 2026-07-08
feat(rocjitsu): add libvfio-user vfio-user GPU emulation for MI350P (…

### LINKED ISSUE #8218: [Feature]: rocjitsu: vfio-user GPU mode — expose rocjitsu as PCIe AMD GPUI via libvfio-user
### Summary

This PR adds `rocjitsu-vfu`, a new opt-in build target that exposes the rocjitsu GPU emulator as a (for the purposes of this initial work) PCIe AMD Instinct MI350P (GFX950/CDNA4) device inside a QEMU virtual machine, using the [libvfio-user](https://github.com/nutanix/libvfio-user) protocol. The guest `amdgpu` kernel driver sees a fully-formed PCIe device and can probe, initialize, and eventually run compute workloads against rocjitsu's simulated command processor.

### Motivation

rocjitsu today runs entirely in-process as a simulated KFD. Adding a vfio-user mode allows us to:
1. Exercise rocjitsu against the **real `amdgpu` driver** (not a userspace stub), validating driver-visible hardware behavior.
2. Enable **VM-based GPU workload testing** without physical hardware.
3. Develop and iterate on the GFX950/CDNA4 hardware model against an upstream driver consumer.

### What's added

**New library: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vfu/`**
- `pci_config.h/.cpp` — PCI identity (VID 0x1002, DID 0x75C8), BAR registration, PCIe/PM/MSI-X capabilities matching MI350P
- `bar0_vram.h/.cpp` — VRAM BAR (BAR0/1, memfd-backed): 256 MB default, 144 GB with `--rebar`
- `bar2_doorbell.h/.cpp` — Doorbell BAR (BAR2/3, 2 MB): traps guest writes and forwards to the CP
- `bar5_mmio.h/.cpp` — MMIO register model (BAR5, 256 KB): returns idle/version register stubs for amdgpu init
- `mmio_registers.h` — BAR5 register offset constants for GFX950
- `dma_mapper.h/.cpp` — Maps guest IOVA ranges (DMA_MAP/DMA_UNMAP from QEMU) into rocjitsu's GPU VA space
- `vfu_server.h/.cpp` — Top-level server: owns the libvfio-user context, VM handle, and all BAR handlers

**New binary: `emulation/rocjitsu/tools/rocjitsu-vfu/`**
- Standalone CLI (`--socket`, `--config`, `--rebar`) that starts the server and blocks until QEMU disconnects
- Documents the matching QEMU `-device vfio-user-pci` invocation

**New config: `emulation/rocjitsu/configs/gfx950_mi350p_kmd.json`**
- MI350P topology (1 AID, 4 XCDs, 144 GB HBM3E) for use with the vfu server

**Extensions to `SimulatedDriver`**
- `trigger_doorbell(process_id, doorbell_byte_offset, value)` — writes a doorbell slot so the CP picks up new AQL packets
- `register_guest_dma(process_id, iova, vaddr, length, map)` — map/unmap guest memory into GPU VA space

**Build integration (`CMakeLists.txt`)**
- New opt-in flag `RJ_ENABLE_VFIO_USER=ON`; all new targets are Linux-only and off by default
- External dep via `pkg-config: libvfio-user`

**Tests: `emulation/rocjitsu/tests/vfu/`**
- Shell smoke test, GTest integration test, and Python QEMU launch helper

### BAR layout (matches real MI350P)

| BAR | Size | Type | Purpose |
|-----|------|------|---------|
| BAR0/1 | 256 MB (or 144 GB with ReBAR) | 64-bit prefetchable | VRAM aperture |
| BAR2/3 | 2 MB | 64-bit prefetchable | Doorbell region |
| BAR5 | 256 KB | 32-bit non-prefetchable | MMIO registers |

### Known limitations / follow-on work

- BAR5 MMIO register stubs are minimal — will be filled in iteratively as `amdgpu` init probes reveal what's needed
- MI350P device ID (0x75C8) should be confirmed against the upstream `amdgpu_drv.c` GFX950 PCI ID table
- Doorbell BAR2 layout should be cross-checked against `aqua_vanjaram.c` before finalizing
- Full aqua_vanjaram IP init sequence (RLC, SDMA, GFX, GMC, IH, PSP) is out of scope for this PR

### Build & quick test

```bash
cmake -B build -G Ninja -DCMAKE_BUILD_TYPE=Debug -DRJ_ENABLE_VFIO_USER=ON
cmake --build build --target rocjitsu-vfu

# Launch server
./build/tools/rocjitsu-vfu/rocjitsu-vfu \
  --config emulation/rocjitsu/configs/gfx950_mi350p_kmd.json \
  --socket /tmp/rocjitsu-vfu-0.sock

# In another terminal, connect QEMU:
qemu-system-x86_64 -accel kvm -m 16G \
  -device '{"driver":"vfio-user-pci","socket":{"path":"/tmp/rocjitsu-vfu-0.sock","type":"unix"}}'


### Operating System

_No response_

### GPU

_No response_

### ROCm Component

rocJitsu

### PR DESCRIPTION
**Pull request by sbates130272**
…GFX950)

Adds rocjitsu-vfu, a libvfio-user server that exposes rocjitsu as a PCIe AMD GPU (device ID 0x75C8, MI350P/GFX950) inside a QEMU VM, and the amdgpu-dkms patch set needed to bring the guest driver to probe.

## rocjitsu-vfu server

A new tool (tools/rocjitsu-vfu/) starts a vfio-user server that QEMU connects to via a UNIX socket, presenting the emulated GPU over the vfio-user PCIe protocol:

  BAR0/1: 256 MB 64-bit prefetchable VRAM (memfd-backed, mmap-able)
  BAR2/3: 2 MB 64-bit prefetchable doorbell window
  BAR5:   256 KB 32-bit non-prefetchable MMIO register window

PCI class 0x120000 (PCI_CLASS_ACCELERATOR_PROCESSING) matches the amdgpu wildcard CHIP_IP_DISCOVERY entry, allowing driver probe.

## IP discovery

The IP discovery binary for GFX9.4.4 is written at VRAM offset (256 MB - DISCOVERY_TMR_OFFSET = 0x0FFF0000) at startup. BAR5 serves mmRCC_CONFIG_MEMSIZE = 256 so the driver computes the correct VRAM address. BAR5 also handles MM_INDEX/MM_INDEX_HI/MM_DATA indirect register reads, routing them to the VRAM memfd so the driver can read the discovery binary via amdgpu_device_mm_access() during early probe before the BAR0 aperture is ioremapped.

## amdgpu-dkms patches (patches/amdgpu-dkms/)

Two patches are required to make modprobe amdgpu succeed against the emulated device:

  0001-vfu-invalid-rreg-no-bug.patch
    Replaces BUG() with pr_warn_once() in amdgpu_invalid_rreg() so
    the driver survives unimplemented PCIE register accesses during
    probe (applies to 6.8 kernels; 6.17+ already handle this cleanly).

  0002-vfu-mm-access-skip-drm-dev-enter.patch
    Removes the drm_dev_enter() guard from amdgpu_device_mm_access()
    so VRAM reads via MM_INDEX/MM_DATA work during IP discovery, which
    runs before drm_dev_register() is called.

A rebuild-and-reload.sh helper script enables incremental kernel module builds and reloads without a full DKMS rebuild.

Fixes #8218

### CODE REVIEW

**Comment by sbates130272** [comment]
This is marked as DRAFT as is a Work In Progress. I wanted to get it into the system for early comments and feedback. Still work to do getting the (modified) amdgpu driver to bind to this but we are getting there.

**Review by atgutier — COMMENTED** [review]
Thanks for this! Really nice to see this coming together.

I went through this PR with an agent and had it provide some initial/immediate feedback based on the plan document I shared. Here are the suggestions directly from the agent:

1. Adopt the neutral simdojo::PciDevice layer

  A starting header is provided at lib/simdojo/include/simdojo/components/pci_device.h (PciId /
  BarSpec / DmaRegion / IrqSink / DmaEngine / PciDevice : Component / PciDeviceProvider). Please build
   the device model on this interface. The one rule that makes NIC and KVM reuse possible:
  <libvfio-user.h> and vfu_* should appear in exactly one place — the transport (VfioDeviceHost) — and
   nowhere else.

  2. Introduce VfioDeviceHost in kmd/linux/ and move all vfu_* into it

  Create lib/rocjitsu/src/rocjitsu/kmd/linux/vfio_device_host.{h,cpp} that owns the vfu_ctx_t +
  poll-fd loop on one thread, implements simdojo::IrqSink (over vfu_irq_trigger) and
  simdojo::DmaEngine (over vfu_addr_to_sgl), and builds the bus from a simdojo::PciDevice's pci_id() /
   bars() / msix_vectors(). The vfu_pci_* config programming currently in pci_config.cpp and the
  vfu_setup_region calls in the BAR classes move here.

  3. Turn the GPU specifics into a GpuPciDevice : simdojo::PciDevice

  This class holds a SoC* (composition, one instance per GPU), returns its PciId/BarSpec list from
  rj_vm_gpu_info, and implements bar_access / dma_map / dma_unmap. It reaches the guest only through
  the injected irq_ / dma_ — never through vfu_*.

  4. Collapse the three BAR classes into one BarSpec-driven primitive

  Bar0Vram and Bar2Doorbell duplicate the same memfd + ftruncate + seal sequence (and re-#define
  MFD_*). Extract a SealedMemfd RAII helper, and let the transport create regions from the device's
  bars() vector rather than from three bespoke classes. Keep MmioModel (BAR5) — just change its
  handler signature to the neutral bar_access(int bar, std::span<std::byte>, uint64_t off, bool write)
   instead of the vfu access_cb(vfu_ctx_t*, char*, size_t, loff_t, bool).

  5. Move the IP-discovery blob placement into the GPU device

  ip_discovery_blob.* is already transport-neutral and stays as-is, but its placement into VRAM should
   happen in GpuPciDevice, not inside Bar0Vram, so the memory-BAR primitive stays
  device-agnostic. While there, add an assert(pos <= buf.size()) in the Writer and a static_assert on
  the header-size constants so the layout can't drift silently, and move the file-scope static
  constexpr constants into the anonymous namespace / inline constexpr to match mmio_registers.h.

  6. Make dma_mapper the transport's DmaEngine, feeding the device via dma_map

  The dma_register/dma_unregister callbacks are libvfio-user signatures, so they belong in
  VfioDeviceHost. The device receives neutral dma_map(const DmaRegion&) / dma_unmap(...) calls. Please
   also guard the page-table update in register_guest_dma (simulated_driver.cpp:1937) with
  proc.alloc_mutex_, matching the other callers of map_to_gpu/unmap_from_gpu — this path runs on the
  vfu thread concurrently with the KFD/CP paths. (The longer-term home for this is the injectable
  MemoryAccessBackend seam on GpuMemory the plan describes, but the lock is worth adding now.)

 7. Wire the completion → MSI-X path through IrqSink

  MSI-X vectors are set up but nothing triggers them yet. When you wire it: have GpuPciDevice
  register a CommandProcessor::set_interrupt_callback that calls irq_->trigger(vector), and — since
  libvfio-user isn't thread-safe — marshal that from the CP/engine thread onto the poll thread via an
  eventfd the poll loop selects on (mirroring the existing schedule_event_now doorbell pattern) rather
   than calling vfu_irq_trigger directly.

  8. Convert the server to provider enumeration

  Add a GPU PciDeviceProvider that yields one GpuPciDevice per GPU.
  tools/rocjitsu-vfu/main.cpp then iterates provider.pci_devices() and stands up one VfioDeviceHost
  per device, never branching on device kind. This is the step that makes "add a NIC" == "register
  another provider."

  9. Build wiring

  Keep the whole front-end behind RJ_ENABLE_VFIO_USER (default OFF) as you have it. Two adjustments:
  pull libvfio-user via ExternalProject_Add (meson/ninja) as an IMPORTED target so the front-end is
  self-contained rather than requiring a pre-installed copy via pkg_check_modules(... REQUIRED); and
  link the libvfio-user dependency PRIVATE to the transport unit (the headers only forward-declare
  vfu_ctx_t), so it doesn't propagate to consumers of rocjitsu_vfu.

  Small style items (whenever the files are touched)

  - Prefer std::span over raw char* + size_t in the internal register/BAR read/write layer.
  - PciIdentity / BarSizes hold only static constexpr members — an inline constexpr set or a namespace
   reads better than an uninstantiated struct.
  - Add #include <algorithm> to dma_mapper.cpp (std::remove_if currently relies on a transitive
  include).

  Two litmus checks to aim at

  - NIC: a teammate adds a device by writing one simdojo::PciDevice subclass + a provider, touching
  zero GPU files and zero libvfio-user code.
  - KVM: a second …DeviceHost can drive the same PciDevice via KVM ioctls without touching the device,
   BARs, or MMIO handlers.

Here is a pci_device.h implementation it also drafted that may help get started:

```
// Copyright (c) 2026 Advanced Micro Devices, Inc.

/// @file pci_device.h
/// @brief Transport-agnostic PCI device model for the simulation graph.
///
/// @details A @ref simdojo::PciDevice is the neutral, transport-independent
/// object a PCI-attached front-end backs. It knows nothing about libvfio-user,
/// KVM, sockets, or any specific VMM: it declares its bus shape (identity, BARs,
/// interrupt count) and reacts to guest-driven events (BAR access, config
/// writes, DMA map/unmap, reset). A transport binds to it by injecting an
/// @ref simdojo::IrqSink and an @ref simdojo::DmaEngine, then translating its
/// own wire protocol into calls on the device's virtual hooks.
///
/// This inversion is what lets a single device model be driven by more than one
/// transport (e.g. a libvfio-user host today, an in-tree KVM path later) and
/// lets device kinds other than GPUs (e.g. a NIC) implement the same interface
/// with zero coupling to any one device family. The transport and the device
/// families live in higher layers; this header adds no dependency beyond the
/// simulation component base.

#ifndef SIMDOJO_COMPONENTS_PCI_DEVICE_H_
#define SIMDOJO_COMPONENTS_PCI_DEVICE_H_

#include "simdojo/sim/component.h"

#include <cstddef>
#include <cstdint>
#include <span>
#include <string>
#include <vector>

namespace simdojo {

/// @brief PCI configuration-space identity used to build the bus.
///
/// @details Populated by the device and read once by the transport to program
/// config space. @ref cls / @ref subcls / @ref prog_if follow the PCI class
/// code triple (e.g. 0x12/0x00/0x00 for a processing accelerator, 0x02 for a
/// network controller); a transport and an in-guest driver dispatch on the
/// class, so device kind is carried here rather than in any type name.
struct PciId {
  uint16_t vendor = 0;        ///< PCI vendor ID.
  uint16_t device = 0;        ///< PCI device ID.
  uint16_t subsys_vendor = 0; ///< Subsystem vendor ID.
  uint16_t subsys = 0;        ///< Subsystem ID.
  uint8_t cls = 0;            ///< PCI base class code.
  uint8_t subcls = 0;         ///< PCI subclass code.
  uint8_t prog_if = 0;        ///< Programming interface byte.
  uint8_t revision = 0;       ///< PCI revision ID.
};

/// @brief Neutral description of one Base Address Register.
///
/// @details The transport reads a device's @ref PciDevice::bars() and programs
/// each region. When @ref backing_fd is >= 0 the transport may expose the fd
/// directly to the guest for zero-copy mmap access; when it is < 0 every access
/// traps back to @ref PciDevice::bar_access(). No transport-specific types
/// appear here so the same spec drives any transport.
struct BarSpec {
  int index = 0;             ///< BAR index, 0..5.
  uint64_t size = 0;         ///< Region size in bytes.
  bool mem = true;           ///< True for a memory BAR, false for I/O.
  bool prefetch = false;     ///< Prefetchable hint.
  bool is_64bit = false;     ///< True if this BAR is the low half of a 64-bit pair.
  bool mmapable = false;     ///< True if the guest may mmap the backing fd.
  int backing_fd = -1;       ///< Backing fd for zero-copy access, or < 0 to trap.
  uint64_t fd_offset = 0;    ///< Offset into @ref backing_fd of the region base.
};

/// @brief A guest memory window shared with the device.
///
/// @details Delivered to the device via @ref PciDevice::dma_map when the guest
/// (through the transport) makes a region DMA-accessible, and withdrawn via
/// @ref PciDevice::dma_unmap. With no virtual IOMMU the guest-physical address
/// equals the I/O virtual address. @ref host_vaddr is non-null only when the
/// transport shared an mmap-able mapping; otherwise the device must reach the
/// window through the injected @ref DmaEngine.
struct DmaRegion {
  uint64_t guest_phys = 0;      ///< Guest-physical base (== IOVA).
  uint64_t length = 0;          ///< Window length in bytes.
  void *host_vaddr = nullptr;   ///< Host mapping, or nullptr if not shared.
  uint32_t prot = 0;            ///< Access protection flags (PROT_*).
};

/// @brief Sink through which a device raises interrupts to the guest.
///
/// @details Implemented by the transport and injected into the device so the
/// device can signal completion (e.g. MSI-X) without knowing how the interrupt
/// reaches the guest.
class IrqSink {
public:
  virtual ~IrqSink() = default;

  /// @brief Raise interrupt vector @p vector toward the guest.
  /// @param[in] vector Zero-based interrupt vector index.
  /// @retval true The interrupt was delivered to the transport.
  /// @retval false Delivery failed (e.g. no attached guest).
  [[nodiscard]] virtual bool trigger(uint32_t vector) = 0;
};

/// @brief Engine through which a device reaches guest RAM.
///
/// @details Implemented by the transport and injected into the device so the
/// device can read and write guest-physical memory (command buffers, results)
/// without depending on how the transport resolves and copies it.
class DmaEngine {
public:
  virtual ~DmaEngine() = default;

  /// @brief Read @p len bytes of guest memory at @p gpa into @p dst.
  /// @param[in] gpa Guest-physical source address.
  /// @param[out] dst Destination buffer, at least @p len bytes.
  /// @param[in] len Number of bytes to read.
  /// @retval true The full range was read.
  /// @retval false The access failed or fell outside a mapped window.
  [[nodiscard]] virtual bool read(uint64_t gpa, void *dst, size_t len) = 0;

  /// @brief Write @p len bytes from @p src to guest memory at @p gpa.
  /// @param[in] gpa Guest-physical destination address.
  /// @param[in] src Source buffer, at least @p len bytes.
  /// @param[in] len Number of bytes to write.
  /// @retval true The full range was written.
  /// @retval false The access failed or fell outside a mapped window.
  [[nodiscard]] virtual bool write(uint64_t gpa, const void *src, size_t len) = 0;
};

/// @brief Transport-agnostic PCI function in the simulation graph.
///
/// @details A PCI function is a node in the machine: it is a @ref Component so
/// it gets @ref Component::engine() (to post doorbell/interrupt events via the
/// engine's thread-safe stimulus path), lifecycle hooks, and topology
/// placement. It is a leaf — it does no @ref Component::step() work and reacts
/// only to transport-driven events.
///
/// A device declares its bus shape once (@ref pci_id, @ref bars,
/// @ref msix_vectors); the transport reads those to build the bus, injects its
/// @ref IrqSink and @ref DmaEngine, then delivers guest events to the virtual
/// hooks below. All guest-driven hooks are invoked on the transport thread;
/// the device is responsible for marshaling any cross-thread work onto the
/// engine (see @ref Component::engine()).
class PciDevice : public Component {
public:
  /// @brief Construct a PCI device with the given node name.
  /// @param[in] name Human-readable name for this function.
  explicit PciDevice(std::string name) : Component(std::move(name)) {}
  ~PciDevice() override = default;

  // --- Bus shape: declared by the device, read once by the transport. ---

  /// @brief Return the PCI configuration-space identity.
  /// @returns The device's @ref PciId.
  [[nodiscard]] virtual PciId pci_id() const = 0;

  /// @brief Return the device's BAR layout.
  /// @returns A vector of @ref BarSpec, one per populated BAR.
  [[nodiscard]] virtual std::vector<BarSpec> bars() const = 0;

  /// @brief Return the number of MSI-X vectors the device uses.
  /// @returns Vector count, or 0 if the device raises no interrupts.
  [[nodiscard]] virtual uint32_t msix_vectors() const { return 0; }

  // --- Transport wiring: the transport injects its sinks before attach. ---

  /// @brief Inject the interrupt sink the device raises IRQs through.
  /// @param[in] sink Transport-owned sink; must outlive the device's use of it.
  void set_irq_sink(IrqSink *sink) { irq_ = sink; }

  /// @brief Inject the DMA engine the device reaches guest RAM through.
  /// @param[in] engine Transport-owned engine; must outlive the device's use.
  void set_dma_engine(DmaEngine *engine) { dma_ = engine; }

  // --- Guest-driven events: invoked on the transport thread. ---

  /// @brief Service a guest read or write to a BAR.
  /// @param[in] bar BAR index the access targets.
  /// @param[in,out] buf On write, the bytes to store; on read, the destination.
  /// @param[in] off Byte offset within the BAR.
  /// @param[in] write True for a write access, false for a read.
  /// @returns Number of bytes serviced, or a negative value on error.
  [[nodiscard]] virtual ssize_t bar_access(int bar, std::span<std::byte> buf, uint64_t off,
                                           bool write) = 0;

  /// @brief Handle a guest write to PCI configuration space.
  /// @param[in] off Byte offset within config space.
  /// @param[in] buf Bytes written by the guest.
  /// @details Default is a no-op; devices override only if they track config
  /// state beyond what the transport programs from @ref pci_id.
  virtual void config_write(uint64_t off, std::span<const std::byte> buf) {}

  /// @brief Register a guest memory window as DMA-accessible.
  /// @param[in] region The mapped window.
  virtual void dma_map(const DmaRegion &region) = 0;

  /// @brief Withdraw a previously mapped guest memory window.
  /// @param[in] region The window being unmapped.
  virtual void dma_unmap(const DmaRegion &region) = 0;

  /// @brief Reset the device to its power-on state.
  /// @details Default is a no-op.
  virtual void reset() {}

  /// @brief Begin quiescing the device ahead of teardown.
  /// @retval 0 Quiesce completed synchronously.
  /// @retval -1 Quiesce is in progress; the transport should wait.
  [[nodiscard]] virtual int quiesce() { return 0; }

protected:
  IrqSink *irq_ = nullptr;   ///< Transport-injected interrupt sink.
  DmaEngine *dma_ = nullptr; ///< Transport-injected DMA engine.
};

/// @brief Enumeration seam that yields PCI functions to a transport server.
///
/// @details A server discovers the set of PCI functions by iterating providers
/// rather than by querying typed counters, so adding a new device kind is
/// "register another provider" and never a change to the server or to another
/// device family. Each provider contributes zero or more devices; the size of
/// the returned span is the count.
class PciDeviceProvider {
public:
  virtual ~PciDeviceProvider() = default;

  /// @brief Return the PCI functions this provider contributes.
  /// @returns Non-owning pointers to devices owned by the provider; the span
  ///          size is the device count.
  [[nodiscard]] virtual std::vector<PciDevice *> pci_devices() = 0;
};

} // namespace simdojo

#endif // SIMDOJO_COMPONENTS_PCI_DEVICE_H_
```

sbates130272:
@atgutier thanks for the feedback. I have some short-term priority issues I need to work on. Are you willing/able to run with this PR? If so, let me know. Otherwise it might take me a few weeks to get back to this depending on other deadlines I have coming up!

atgutier:
> @atgutier thanks for the feedback. I have some short-term priority issues I need to work on. Are you willing/able to run with this PR? If so, let me know. Otherwise it might take me a few weeks to get back to this depending on other deadlines I have coming up!

I don't have anytime this week and am in a similar boat. May have some time for this in a week or two. However, this is not urgent right now and would like to make sure we get this right as it is quite a substantial addition so taking our time is best.

gaoikawa:
@atgutier - Saw your other (related?) PR - https://github.com/ROCm/rocm-systems/pull/10729

Looking forward to this functionality!!!

atgutier:
> @atgutier - Saw your other (related?) PR - #10729
> 
> Looking forward to this functionality!!!

Yes, my pr stack supersedes this POC

sbates130272:
Closed. Superceded.

PR #8300, opened 2026-07-09
fix(libhsakmt): route primary-ctx render device init through model-aware w…

timhuang-amd:
## Motivation

In hsakmt_open_drm_render_device(), the primary-context path called libdrm's amdgpu_device_initialize() directly. This triggers a printing on ffm:

    _amdgpu_device_initialize: amdgpu_get_auth (1) failed (-1)

Use the existing model-aware wrapper hsakmt_amdgpu_device_initialize() instead.

## JIRA ID

N/A

## Test Plan

    kfdtest

### DIFF
--- a/projects/rocr-runtime/libhsakmt/src/fmm.c
+++ b/projects/rocr-runtime/libhsakmt/src/fmm.c
@@ -2403,7 +2403,7 @@ int hsakmt_open_drm_render_device(HsaKFDContext *ctx, int minor)
 	 *    its own VM space).
 	*/
 	if (ctx->hsakmt_is_primary_ctx) {
-		dev_init_ret = amdgpu_device_initialize(fd, &major_drm, &minor_drm, device_handle);
+		dev_init_ret = hsakmt_amdgpu_device_initialize(fd, &major_drm, &minor_drm, device_handle);
 	} else if (hsakmt_fn_amdgpu_device_initialize2) {
 		dev_init_ret = hsakmt_fn_amdgpu_device_initialize2(fd, false, &major_drm, &minor_drm,
 						    (HsaAMDGPUDeviceHandle *)device_handle);


projects/rocr-runtime/libhsakmt/src/fmm.c
```diff
@@ -2402,7 +2402,7 @@ int hsakmt_open_drm_render_device(HsaKFDContext *ctx, int minor)
 	 *    its own VM space).
 	*/
 	if (ctx->hsakmt_is_primary_ctx) {
-		dev_init_ret = amdgpu_device_initialize(fd, &major_drm, &minor_drm, device_handle);
+		dev_init_ret = hsakmt_amdgpu_device_initialize(fd, &major_drm, &minor_drm, device_handle);
```
cfreeamd, line 2406: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** Can you confirm that `hsakmt_amdgpu_device_initialize()` is a transparent passthrough to `amdgpu_device_initialize()` on real hardware — i.e., it only differs in behavior on FFM/simulation? If yes, the fix is unambiguously safe. If the wrapper has any behavioral differences on hardware (error propagation, return-value normalization, etc.), those should be documented.

cfreeamd (commented):
🤖 AI-assisted pre-review (for @cfreeamd) — fmm.c looks correct for the stated purpose; routing primary-ctx through the model-aware wrapper is consistent with how the else-if branch in the same function already handles non-primary paths.

amd-junhshen (commented):
I am not familiar with the functionality of the `haskmt model`. If `hsakmt_open_drm_render_device` is also used in the `haskmt model` path, this modification seems reasonable to me. 
Waiting for other reviewers' feedback.

projects/rocr-runtime/libhsakmt/src/fmm.c
timhuang-amd: On real hardware hsakmt_amdgpu_device_initialize() is a transparent passthrough with no behavioral difference from calling amdgpu_device_initialize() directly.

  The wrapper's entire model-specific logic is gated behind hsakmt_use_model:

  int hsakmt_amdgpu_device_initialize(int fd, uint32_t *major_version, uint32_t *minor_version,
                                       amdgpu_device_handle *device_handle)
  {
      if (hsakmt_use_model) {
          ... /* model / FFM path only */
      }
      return amdgpu_device_initialize(fd, major_version, minor_version, device_handle);
  }

  When hsakmt_use_model is false, it does nothing but forward all four arguments unchanged and return the value verbatim — no error propagation changes, no return-value normalization, no extra
  logging.

  hsakmt_use_model defaults to false and is set to true in exactly one place (model_init_env_vars()), only when the HSA_MODEL_TOPOLOGY env var is present. On real hardware that var is unset, so
  the model branch is never taken. The change is a genuine no-op on hardware and only alters the FFM/simulation path (suppressing the spurious amdgpu_get_auth (1) failed (-1) print), so there
  are no hardware behavioral differences to document.

cfreeamd (commented):
🤖 AI-assisted pre-review (for @cfreeamd) — follow-up after author response

Thanks @timhuang-amd — the prior [Question] on `hsakmt_amdgpu_device_initialize()` is resolved. I verified against `libhsakmt/src/hsakmtmodel.c`: when `hsakmt_use_model` is false (the hardware default, only set true when `HSA_MODEL_TOPOLOGY` is present), the wrapper forwards all four arguments and returns `amdgpu_device_initialize(...)` verbatim — a genuine no-op on hardware that only affects the FFM/model path. No remaining concerns from the AI pre-review.

cfreeamd (approved):

PR #8318, opened 2026-07-09
feat(rocr-runtime): Add DRM/libdrm backend for SVM (set/get attributes, prefetch)

amd-junhshen:
## Motivation

ROCr's SVM functionality (`hsa_amd_svm_attributes_set` / `_get` / `hsa_amd_svm_prefetch_async`) is currently implemented only against the KFD (`/dev/kfd`) SVM ioctls. This PR adds an alternative **DRM (libdrm/amdgpu) backend** so the same HSA-level SVM operations can run through the amdgpu DRM `svm_set_attr` / `svm_get_attr` ioctls, enabling SVM on the DRM path while preserving identical HSA semantics. The backend is gated behind the existing `enable_drm` flag and does not change default (KFD) behavior.

## Technical Details

- **Driver abstraction.** SVM entry points (`SvmSetAttr` / `SvmGetAttr` / `SvmPrefetch`) are virtual methods on `core::Driver`. `DrmDriver` derives from `KfdDriver` (via a new protected `KfdDriver(DriverType, devnode)` constructor), overrides only the SVM and DRM-specific queue paths, and inherits everything else — delegating back to KFD where DRM has no equivalent yet. GPU targets are identified by `core::Agent` rather than raw node ids: dispatch to the owning backend happens at the call site via `agent->driver()`, and each backend then reads only what it needs from the agent (KFD: `node_id()`; DRM: `device_type()` + `GpuAgent::libDrmDev()`), so the DRM backend never has to reverse-map a node id back to a device.

- **Per-device kernel model.** The amdgpu DRM SVM model maintains one independent context per device; an operation spanning "all devices" is expressed as one ioctl per participating GPU, issued on that GPU's fd (obtained from the agent's `GpuAgent::libDrmDev()`). The target device is chosen by which fd the ioctl is sent on, not by a node ID in the payload. Location fields are sentinels (SYSMEM=0, UNDEFINED=0xffffffff, any other value = "the device that received this ioctl"), so a fixed `kSvmLocationThisGpu = 1u` constant is used together with fd routing.

- **common vs per_device attributes.** `SvmSetAttr` splits attributes into range-wide (`common`) and device-specific (`per_device[agent]`, keyed by `core::Agent*`) buckets and fans them out per GPU agent (iterating `gpu_agents()`, reaching each context through `GpuAgent::libDrmDev()`). Boolean flags are accumulated as `set_mask`/`clr_mask` and materialized once. CPU accessibility maps to the range-wide HOST_ACCESS flag.

- **Attribute reset (count == 0).** HSA defines `hsa_amd_svm_attributes_set` with a zero-length attribute list as "restore this range to default attributes." The two backends realize this differently because their kernel models differ. KFD keeps no persistent per-range attribute state that the runtime must unwind, so `KfdDriver::SvmSetAttr` treats `count == 0` as a no-op that returns success. The amdgpu DRM model, by contrast, stores explicit attribute ranges per SVM context, so defaults must be re-established actively: `DrmDriver::SvmSetAttr` fans a dedicated `amdgpu_svm_reset_attr` ioctl out to every participating GPU (one call per fd, resolved via each GPU agent's `GpuAgent::libDrmDev()`), before any attribute parsing. On the kernel side the reset is deliberately lightweight — `amdgpu_svm_attr_reset` simply clears the attribute ranges over `[base, base+size)` rather than materializing default entries, relying on the invariant that "no attr node" is equivalent to "default attrs" for the GET, SET, and fault paths. Because it only touches attribute metadata (not GPU mappings), no mapping invalidation is issued. Consistent with the other set paths, the reset fails fast: any per-device ioctl error aborts the whole call with `HSA_STATUS_ERROR`.

- **Batched read-back.** `SvmGetAttr` reconstructs HSA answers in three phases (plan → collect one batched `amdgpu_svm_get_attr` per GPU agent → resolve), including ACCESS_QUERY rewriting the entry's attribute field to AGENT_ACCESSIBLE / _IN_PLACE / AGENT_NO_ACCESS. Preferred/prefetch locations resolved to a GPU are mapped back to the corresponding `core::Agent`.

- **Prefetch.** `SvmPrefetch` takes the destination `core::Agent&`. A GPU destination prefetches on that agent's fd only (`GpuAgent::libDrmDev()`, `device_type() == kAmdGpuDevice`); a CPU/sysmem destination fans a SYSMEM prefetch out to every GPU agent's fd, since each device's independent context must be updated.

- **Shared validation.** Per-call agent-handle conversion and de-duplication is factored into a short-lived `KfdDriver::SvmAttrParser` helper, reused by both backends and both set/get paths.

- **Error semantics.** `SvmSetAttr` and GPU-destination `SvmPrefetch` fail fast: any per-device ioctl error aborts the whole call with `HSA_STATUS_ERROR`. Sysmem-destination `SvmPrefetch` and get-collection are best-effort per device. Because the amdgpu SVM model has no cross-device atomic transaction and prefetch has no clean rollback, **any multi-device operation that fails partway leaves some contexts updated and others not — neither fail-fast nor best-effort can avoid this partial state.** The two paths therefore make a deliberate, workload-driven trade-off:

     - **GPU-destination prefetch** targets a single device, so fail-fast is unambiguous and surfaces errors directly (matching `SvmSetAttr`).
     - **Sysmem-destination prefetch** must fan out to *every* GPU context; the goal is to drive the range toward system memory as broadly as possible, so it attempts all devices rather than stopping at the first failure. Since prefetch is a migration *hint* (pages stay valid and are re-migrated on demand by the fault path), a partial sysmem prefetch is a performance concern, not a correctness/coherence violation — making best-effort the appropriate choice here.

A companion design document is included at:
-  [Design: DRM/libdrm Backend for ROCr SVM](https://amd.atlassian.net/wiki/spaces/ASLGP/pages/1743456343/Design+DRM+libdrm+Backend+for+ROCr+SVM)
- [Per-SVM-Per-Device Multi-GPU Attribute Consistency Analysis](https://amd.atlassian.net/wiki/spaces/ASLGP/pages/1798326999/Per-SVM-Per-Device+Multi-GPU+Attribute+Consistency+Analysis).
## Dependencies

This is part of a systematic, cross-component effort. This PR depends on:

- **Kernel** — the amd-unified-interface + DRM SVM branch: <https://gerrit-git.amd.com/plugins/gitiles/brahma/ec/linux/+/refs/heads/junhshen/drm-svm-on-unified-interface>
- **libdrm** — the amdgpu DRM SVM branch: <https://gerrit-git.amd.com/plugins/gitiles/brahma/ec/drm/+/refs/heads/junhshen/amdgpu-drm-svm>

## Test Plan

- rocrtst

## Test Result
- rocrtst `SvmMemory` subtests pass on gfx906 via the DRM backend.
- Note: current runs pin `ROCR_VISIBLE_DEVICES=0`, so validation is single-GPU; the multi-GPU fan-out paths should be exercised with ≥2 visible GPUs before merge.

## Note

This is a **draft** intended primarily to solicit feedback on:
- The **driver-interface design** (the `core::Driver` SVM abstraction and its KFD/DRM subclassing) and on whether the **runtime-layer integration** is reasonable.
- The implementation details still need further validation — in particular the split between **common (range-wide) attributes** and **per-device attributes**, and the model for **fanning these out into per-GPU ioctls**. Feedback on that model is especially welcome.

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/driver/drm/amd_drm_driver.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/driver/drm/amd_drm_driver.cpp
@@ -434,6 +434,18 @@ hsa_status_t DrmDriver::SvmSetAttr(void* base, size_t size,
   if (gpu_agents.empty())
     return HSA_STATUS_ERROR;
 
+  // if count is 0, reset all attributes to default values
+  if (count == 0) {
+    for (core::Agent* agent : gpu_agents) {
+      amdgpu_device_handle dev = static_cast<GpuAgent*>(agent)->libDrmDev();
+      if (dev == nullptr)
+        continue;
+      if (amdgpu_svm_reset_attr(dev, reinterpret_cast<uint64_t>(base), size))
+        return HSA_STATUS_ERROR;
+    }
+    return HSA_STATUS_SUCCESS;
+  }
+
   std::vector<drm_amdgpu_svm_attribute> common;
   std::unordered_map<core::Agent*, std::vector<drm_amdgpu_svm_attribute>> per_device;
   // Bitmasks of boolean DRM attributes to set (=1) or clear (=0), indexed by
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/driver/drm/amd_drm_driver.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/driver/drm/amd_drm_driver.cpp
@@ -7,10 +7,15 @@
 
 #include <cerrno>
 #include <cstring>
+#include <algorithm>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
 
 #include "core/inc/amd_drm_driver.h"
 #include "core/inc/amd_aql_queue.h"
 #include "core/inc/amd_gpu_agent.h"
+#include "core/inc/exceptions.h"
 
 namespace rocr {
 namespace AMD {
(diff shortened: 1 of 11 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/driver/drm/amd_drm_driver.cpp
```diff
@@ -390,6 +395,429 @@ hsa_status_t DrmDriver::DestroyQueue(HSA_QUEUEID queue_id) const {
...
+      common.push_back({attr, 0});
+  }
+
+  // Issue one ioctl per GPU with the shared entries plus that GPU's own ones.
+  for (uint32_t node : gpu_nodes) {
+    amdgpu_device_handle dev = LibDrmDevForNode(node);
+    if (dev == nullptr)
+      continue;
+
+    std::vector<drm_amdgpu_svm_attribute> atrrs = common;
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** Typo `atrrs` → `attrs` (repeated on the following lines that reference `atrrs`).

projects/rocr-runtime/runtime/hsa-runtime/core/driver/drm/amd_drm_driver.cpp
```diff
@@ -390,6 +395,429 @@ hsa_status_t DrmDriver::DestroyQueue(HSA_QUEUEID queue_id) const {
...
 
+namespace {
+
+// Boolean DRM SVM attributes occupy the contiguous enum range
+// [AMDGPU_SVM_ATTR_HOST_ACCESS, AMDGPU_SVM_ATTR_GPU_READ_MOSTLY]. Their enum
+// values double as bit positions in the set/clear masks used below.
+constexpr uint32_t kFirstDrmBoolAttr = AMDGPU_SVM_ATTR_HOST_ACCESS;
+constexpr uint32_t kLastDrmBoolAttr = AMDGPU_SVM_ATTR_GPU_READ_MOSTLY;
+
+constexpr uint32_t DrmFlagBit(uint32_t drm_attr) { return 1u << drm_attr; }
```
cfreeamd, line 413: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** `DrmFlagBit(x) = 1u << drm_attr` assumes every boolean `AMDGPU_SVM_ATTR_*` value stays `< 32`. If the DRM enum grows past 31 this is silent UB. A `static_assert(kLastDrmBoolAttr < 32, ...)` would guard it cheaply.

projects/rocr-runtime/runtime/hsa-runtime/core/driver/drm/amd_drm_driver.cpp
```diff
@@ -390,6 +395,429 @@ hsa_status_t DrmDriver::DestroyQueue(HSA_QUEUEID queue_id) const {
...
+    return true;
+  };
+
+  // Reconstruct the boolean-flag bitmask (indexed by DRM attribute enum value,
+  // see DrmFlagBit) from the representative device's cached values.
+  auto flags = [&]() -> uint32_t {
+    uint32_t f = 0;
+    uint32_t v = 0;
+    for (uint32_t drm_attr : kReportedDrmFlags)
+      if (node_val(gpu_nodes[0], drm_attr, v) && v)
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** `SvmGetAttr` reads boolean flags and granularity only from `gpu_nodes[0]` (representative device). In the one-SVM-per-GPU model each GPU context is independent. `SvmSetAttr` routes bool flags through `common` (all devices), so state stays uniform within this runtime — but is that guaranteed against prior/external state, or across GPUs where an ioctl partially failed? If not, the reported flags for a multi-GPU range could be wrong.

projects/rocr-runtime/runtime/hsa-runtime/core/driver/drm/amd_drm_driver.cpp
```diff
@@ -390,6 +395,429 @@ hsa_status_t DrmDriver::DestroyQueue(HSA_QUEUEID queue_id) const {
...
+        break;
+      }
+      case HSA_AMD_SVM_ATTRIB_PREFETCH_LOCATION: {
+        value = location_handle(AMDGPU_SVM_ATTR_PREFETCH_LOC);
+        break;
+      }
+      case HSA_AMD_SVM_ATTRIB_ACCESS_QUERY: {
+        core::Agent* agent = parser.Convert(value);
+        if (agent->device_type() == core::Agent::kAmdCpuDevice) {
+          attrib = (flags() & DrmFlagBit(AMDGPU_SVM_ATTR_HOST_ACCESS))
```
cfreeamd, line 770: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** ACCESS_QUERY path: when the per-node query fails or the device handle is null, `attrib` is left as `HSA_AMD_SVM_ATTRIB_ACCESS_QUERY` (unresolved). The KFD path `assert`s on bad data; here the caller silently receives an unrewritten query. Consider defaulting to `AGENT_NO_ACCESS` or surfacing an error.

projects/rocr-runtime/runtime/hsa-runtime/core/driver/drm/amd_drm_driver.cpp
```diff
@@ -390,6 +395,429 @@ hsa_status_t DrmDriver::DestroyQueue(HSA_QUEUEID queue_id) const {
...
+    for (uint32_t drm_attr : kReportedDrmFlags)
+      if (node_val(gpu_nodes[0], drm_attr, v) && v)
+        f |= DrmFlagBit(drm_attr);
+    return f;
+  };
+
+  // Resolve a preferred/prefetch location to a public agent handle (handle 0
+  // when the range has no or a mixed destination), scanning cached per-node
+  // values.
+  auto location_handle = [&](uint32_t drm_type) -> uint64_t {
```
cfreeamd, line 706: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** `location_handle`: SYSMEM maps unconditionally to node `0` — on multi-CPU/NUMA systems node 0 may not be the relevant CPU. Also, if two GPUs each report a local preferred location, the loop returns the first rather than reporting a mixed destination (handle 0), a minor semantic divergence from the unified KFD model.

projects/rocr-runtime/runtime/hsa-runtime/core/inc/runtime.h
```diff
@@ -468,6 +468,40 @@ class Runtime {
...
+  /// Returns the driver managing the GPU node @p node_id, or, when no id is
+  /// given (or the id is not a GPU node), the driver of the first GPU agent.
+  Driver& AgentDriver(uint32_t node_id = INVALID_NODEID) {
+    if (node_id != INVALID_NODEID) {
+      Agent* agent = agent_by_nodeid(node_id);
+      if (agent && agent->device_type() == Agent::kAmdGpuDevice)
+        return agent->driver();
+    }
+
+    if (gpu_agents_.empty())
```
cfreeamd, line 511: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** `AgentDriver()` now throws `hsa_exception` when `gpu_agents_` is empty; previously `SetSvmAttrib`/`GetSvmAttrib` called `hsaKmtSVMSetAttr` directly. Confirm every SVM entry point is only reachable with a GPU present, otherwise a previously-graceful no-GPU path becomes a thrown exception.

cfreeamd (commented):

The KFD-path refactor (moving inline `hsaKmtSVM*` calls into `KfdDriver::Svm*` behind a driver-dispatch abstraction) is faithful to the original logic. Findings are concentrated in the new `DrmDriver` SVM backend and the driver-dispatch seams — see inline comments. Two items worth author confirmation before merge that have no single line anchor: (1) whether the minimum libdrm version is bumped/guarded in the build config for the new `amdgpu_svm_{set,get}_attr` symbols and `AMDGPU_SVM_ATTR_*` enums, and (2) the representative-device flag-read model in `SvmGetAttr` for multi-GPU ranges.

projects/rocr-runtime/runtime/hsa-runtime/core/driver/drm/amd_drm_driver.cpp
amd-junhshen: Done.

projects/rocr-runtime/runtime/hsa-runtime/core/driver/drm/amd_drm_driver.cpp
```diff
@@ -0,0 +1,823 @@
...
+
+namespace {
+
+// Boolean DRM SVM attributes occupy the contiguous enum range
+// [AMDGPU_SVM_ATTR_HOST_ACCESS, AMDGPU_SVM_ATTR_GPU_READ_MOSTLY]. Their enum
+// values double as bit positions in the set/clear masks used below.
+constexpr uint32_t kFirstDrmBoolAttr = AMDGPU_SVM_ATTR_HOST_ACCESS;
+constexpr uint32_t kLastDrmBoolAttr = AMDGPU_SVM_ATTR_GPU_READ_MOSTLY;
+
+constexpr uint32_t DrmFlagBit(uint32_t drm_attr) { return 1u << drm_attr; }
```
cfreeamd, line 413: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** Follow-up: the "Done." reply suggests this was fixed, but the pushed HEAD (`1f435f8`) still defines `DrmFlagBit` as `1u << drm_attr` with no `static_assert(kLastDrmBoolAttr < 32, ...)`. `kLastDrmBoolAttr` is already defined just above, so the guard is a one-liner — looks like the change may not have been pushed.

projects/rocr-runtime/runtime/hsa-runtime/core/driver/drm/amd_drm_driver.cpp
```diff
@@ -0,0 +1,823 @@
...
+      common.push_back({attr, 0});
+  }
+
+  // Issue one ioctl per GPU with the shared entries plus that GPU's own ones.
+  for (uint32_t node : gpu_nodes) {
+    amdgpu_device_handle dev = LibDrmDevForNode(node);
+    if (dev == nullptr)
+      continue;
+
+    std::vector<drm_amdgpu_svm_attribute> atrrs = common;
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** Still present in the pushed HEAD: `atrrs` → `attrs` (and the following lines that reference `atrrs`).

cfreeamd (commented):
🤖 AI-assisted pre-review (for @cfreeamd) — follow-up after author reply

No new commits since the 2026-07-13 pre-review — HEAD is still `1f435f8`. The "Done." reply on the `DrmFlagBit` static_assert suggestion isn't reflected in the pushed branch: `1u << drm_attr` is still unguarded and the `atrrs` → `attrs` typo remains. The two prior [Question] items are also still open: (1) `SvmGetAttr` reads boolean flags/granularity only from the representative `gpu_nodes[0]`, and (2) minimum libdrm version gating for the new `amdgpu_svm_*` symbols. All items remain non-blocking; the KFD refactor still looks faithful.

projects/rocr-runtime/runtime/hsa-runtime/core/driver/drm/amd_drm_driver.cpp
```diff
@@ -393,6 +398,436 @@ hsa_status_t DrmDriver::DestroyQueue(HSA_QUEUEID queue_id) const {
...
+    return nullptr;
+  return static_cast<GpuAgent*>(agent)->libDrmDev();
+}
+
+}  // namespace
+
+hsa_status_t DrmDriver::SvmSetAttr(void* base, size_t size,
+                                   const hsa_amd_svm_attribute_pair_t* attribs, size_t count) {
+  if (count == 0)
+    return amdgpu_svm_reset_attr(LibDrmDevForNode(core::Runtime::runtime_singleton_->gpu_ids()[0]), reinterpret_cast<uint64_t>(base), size);
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** The count==0 reset calls `amdgpu_svm_reset_attr` on only `gpu_ids()[0]`. The non-zero path below issues one ioctl **per GPU** (loop over `gpu_nodes`) precisely because this is a one-SVM-context-per-GPU model. On a multi-GPU system the reset therefore clears GPU 0's context only, leaving residual attributes on GPUs 1..N — which defeats the leak-prevention this sentinel is documented to provide ("resets ALL SVM attributes ... atomically clearing any previously set attributes"). Consider looping over all `gpu_nodes` like the set path does.

**[Optional]** This branch also dereferences `gpu_ids()[0]` before the `if (gpu_nodes.empty())` guard a few lines down — count==0 with no GPU agents is an out-of-bounds access. And `LibDrmDevForNode(...)` can return nullptr; it's passed straight to `amdgpu_svm_reset_attr` without the null check the other DRM paths perform.

projects/rocr-runtime/runtime/hsa-runtime/core/driver/kfd/amd_kfd_driver.cpp
```diff
@@ -919,6 +922,319 @@ hsa_status_t KfdDriver::GetDeviceFd(uint32_t node_id, int *fd) const {
...
+  // Merge CPU access properties - grant access if any CPU needs access.
+  // Probably wrong.
+  if (set_flags & HSA_SVM_FLAG_HOST_ACCESS) clear_flags &= ~HSA_SVM_FLAG_HOST_ACCESS;
+
+  // Add flag updates
+  if (clear_flags) kmt_attribs.push_back(kmtPair(HSA_SVM_ATTR_CLR_FLAGS, clear_flags));
+  if (set_flags) kmt_attribs.push_back(kmtPair(HSA_SVM_ATTR_SET_FLAGS, set_flags));
+
+  HSAKMT_STATUS error = HSAKMT_CALL(
+      hsaKmtSVMSetAttr(base, size, kmt_attribs.size(), &kmt_attribs[0]));
```
cfreeamd, line 1075: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** `KfdDriver::SvmSetAttr` has no `count == 0` branch, so the new reset sentinel is not honored on the KFD path. With count==0 the loop is skipped, `kmt_attribs` stays empty, and this line calls `hsaKmtSVMSetAttr(base, size, 0, &kmt_attribs[0])`. Two problems: (a) `&kmt_attribs[0]` on an empty vector is UB (element-0 deref); (b) even if it survives, passing 0 attributes is a no-op — so the documented reset semantic silently does nothing on the mainline KFD path while the DRM path performs a real reset. Since CLR now deliberately invokes count==0 on free, please implement the reset here (or explicitly document KFD reset as unsupported and guard against the empty-vector deref).

cfreeamd (commented):
🤖 AI-assisted pre-review (for @cfreeamd) — follow-up after commit 3d90199e

The two prior actionable items are resolved at the new tip: the `atrrs`→`attrs` typo and the `static_assert(kLastDrmBoolAttr < 32)` guard on `DrmFlagBit` are both now present. Thanks.

The new `attribute_count == 0` reset sentinel raises two correctness concerns, noted inline: (1) the DRM path resets only the first GPU while the normal path is per-GPU, and (2) the KFD path has no count==0 handling. One non-inline build-config question: the new `amdgpu_svm_reset_attr` symbol (in addition to `amdgpu_svm_set_attr`/`get_attr`) — is the minimum libdrm version bumped/guarded so environments lacking it still build?

projects/rocr-runtime/runtime/hsa-runtime/core/driver/kfd/amd_kfd_driver.cpp
```diff
@@ -919,6 +922,323 @@ hsa_status_t KfdDriver::GetDeviceFd(uint32_t node_id, int *fd) const {
...
+hsa_status_t KfdDriver::SvmSetAttr(void* base, size_t size,
+                                   const hsa_amd_svm_attribute_pair_t* attribs, size_t count) {
+  SvmAttrParser parser("KfdDriver::SvmSetAttr");
+  std::vector<HSA_SVM_ATTRIBUTE> kmt_attribs;
+  kmt_attribs.reserve(count);
+  uint32_t set_flags = 0;
+  uint32_t clear_flags = 0;
+
+    //do not set any attributes if count is 0, just return success
+  if (count == 0)
```
cfreeamd, line 968: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** The early return fixes the prior empty-vector UB — thanks. But note the asymmetry: on the DRM path count==0 actively resets every GPU's SVM context (`amdgpu_svm_reset_attr` per node), while here it's a silent no-op. That's presumably correct because KFD clears a range's SVM attributes on free via svm_range teardown, whereas libdrm's per-GPU context persists and needs the explicit reset. Can you confirm that's the reasoning, and add a one-line comment explaining why KFD requires no explicit reset here? Otherwise a future reader may think the sentinel is unimplemented on KFD by accident.

projects/rocr-runtime/runtime/hsa-runtime/core/driver/drm/amd_drm_driver.cpp
```diff
@@ -393,6 +398,442 @@ hsa_status_t DrmDriver::DestroyQueue(HSA_QUEUEID queue_id) const {
...
+                                   const hsa_amd_svm_attribute_pair_t* attribs, size_t count) {
+  const auto& gpu_nodes = core::Runtime::runtime_singleton_->gpu_ids();
+
+  if (gpu_nodes.empty())
+    return HSA_STATUS_ERROR;
+
+  // if count is 0, reset all attributes to default values
+  if (count == 0) {
+    for (const auto& node : gpu_nodes) {
+      if (amdgpu_svm_reset_attr(LibDrmDevForNode(node), reinterpret_cast<uint64_t>(base), size))
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** The reset loop passes `LibDrmDevForNode(node)` straight into `amdgpu_svm_reset_attr` without the `if (dev == nullptr) continue;` guard the non-zero set path uses below. If a node has no libdrm device handle this hands a nullptr to the ioctl wrapper. Mirror the null check for consistency.

cfreeamd (commented):
🤖 AI-assisted pre-review (for @cfreeamd) — follow-up after re-push (tip cd14d1798a)

Both prior [Important] findings on the attribute_count==0 reset sentinel are resolved: the DRM reset now loops over all `gpu_nodes` (amd_drm_driver.cpp:455) instead of only `gpu_ids()[0]`, and the KFD path early-returns on count==0 (amd_kfd_driver.cpp:966), eliminating the empty-vector `&kmt_attribs[0]` UB. Remaining items are non-blocking and noted inline.

projects/rocr-runtime/runtime/hsa-runtime/core/inc/driver.h
```diff
@@ -408,6 +408,37 @@ class Driver {
...
+                                  hsa_amd_svm_attribute_pair_t* attribs, size_t count) {
+    return HSA_STATUS_ERROR_INVALID_AGENT;
+  }
+
+  /// @brief Prefetch (migrate) an SVM range to @p dst_node.
+  /// @param[in] base Start address of the range.
+  /// @param[in] size Size of the range in bytes.
+  /// @param[in] dst_node Target node id (a GPU, or a CPU/sysmem node).
+  /// @return HSA_STATUS_SUCCESS on success.
+  virtual hsa_status_t SvmPrefetch(void* base, size_t size, uint32_t dst_node) {
```
dayatsin-amd: Nit pick. The concept of node_id() is very KFD specific and long term, I want to avoid using the node_id as a mechanism to identify the GPU devices. 

Can you please change the function signature to take a core::Agent reference instead?

virtual hsa_status_t SvmPrefetch(**const core::Agent& dstAgent**, void* base, size_t size)

I have a task to change the node_id from the other functions to do the same later.

dayatsin-amd (approved):
One nit pick, but generally LGTM

projects/rocr-runtime/runtime/hsa-runtime/core/inc/driver.h
amd-junhshen: Good point. I've updated the SVM-related functions to take the `core::Agent` reference or pointer instead.

projects/rocr-runtime/runtime/hsa-runtime/core/driver/drm/amd_drm_driver.cpp
```diff
@@ -393,6 +398,428 @@ hsa_status_t DrmDriver::DestroyQueue(HSA_QUEUEID queue_id) const {
...
+  // Prefetch to system memory: every GPU's independent SVM context must be
+  // updated, so issue the ioctl on each device's fd.
+  drm_amdgpu_svm_attribute attr{AMDGPU_SVM_ATTR_PREFETCH_LOC, AMDGPU_SVM_LOCATION_SYSMEM};
+
+  for (core::Agent* agent : gpu_agents) {
+    amdgpu_device_handle dev = static_cast<GpuAgent*>(agent)->libDrmDev();
+    if (dev == nullptr)
+      continue;
+
+    amdgpu_svm_set_attr(dev, reinterpret_cast<uint64_t>(base), size, 1, &attr);
```
cfreeamd, line 836: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** The sysmem-destination prefetch loop ignores the return of `amdgpu_svm_set_attr` per device, whereas the GPU-destination path above returns `HSA_STATUS_ERROR` on failure. Since runtime.cpp's prefetch signal handler does `assert(error == HSA_STATUS_SUCCESS && "KFD Prefetch failed.")`, a genuine per-device failure on the sysmem path is silently swallowed and never surfaces even in debug builds. Consider checking the return and propagating an error (or documenting that sysmem prefetch is best-effort across GPUs).

cfreeamd (commented):
🤖 AI-assisted pre-review (for @cfreeamd) — follow-up after the node_id→core::Agent refactor (force-push 2026-07-28, amending both existing commits).

The Svm* driver surface now takes the raw attribute list (SvmSetAttr/SvmGetAttr) and a `const core::Agent&` (SvmPrefetch) instead of node_id, addressing @dayatsin-amd's driver.h nit; node_id usage is now correctly encapsulated inside KfdDriver. The KFD SetAttr refactor faithfully preserves the original runtime.cpp logic (CheckOnce bitmask, CPU host-access merge including the 'Probably wrong' clear-flags override, PREFERRED_LOC/ACCESS mapping). All prior findings remain resolved at this tip: KFD count==0 early-returns before &kmt_attribs[0] (no empty-vector UB); DRM count==0 reset loops per-GPU over gpu_agents with the empty() guard preceding it AND now includes the `if (dev == nullptr) continue;` guard previously flagged as [Optional]. runtime.cpp's GetSvmAttrib prefetch handling (resolve prefetch locally from prefetch_map_, forward the rest, copy results back including the in-place ACCESS_QUERY rewrite) looks correct. One small non-blocking note inline.

cfreeamd (approved):

gandryey (approved):

Issue #7891, opened 2026-06-26
[Task]: Add tests to rocjitsu CI


RattataKing:
### Suggestion Description

- [x] Add fpsan to CI
- [ ] Enable skipped corpus tests on CI as fixes come up
- [ ] Add OPT125 to CI
- [x] Add hipkitten gfx1250 to CI
- [ ] maintain rocjitsu CI
- [ ] Add tsan to CI test matrix
- [x] Add -fno-sanitize-recover=all to CI
- [x] Run tests in parallel
- [x] Rerun failed test
- [x] Move rocblas tests from rocjitsu unit tests to corpus test and enable it in CI
- [x] Add gfx1250 runtime semantic tests
- [x] Add llama tests
- [x] Add warning if passed test durations are close to the pytest timeout value.

PR #8325, opened 2026-07-09
fix(rocjitsu): fix ubsan related errors

amd-eochoalo:
## Motivation

Using ubsan with `-fno-sanitize-recover=all` revealed additional undefined behaviour. 

## Technical Details

Overflows in the autogenerated code and unaligned accesses on DBI/DBT code.

## JIRA ID

https://github.com/ROCm/rocm-systems/issues/7891

## Test Plan

Test on CI, add `-fno-sanitize-recover=all` to CI.

## Test Result

Test passes.

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/CMakeLists.txt
+++ b/emulation/rocjitsu/CMakeLists.txt
@@ -164,7 +164,6 @@ if(_rj_sanitizer_count GREATER 0)
     list(JOIN _rj_sanitizer_kinds "," _rj_sanitizer_arg)
     set(RJ_SANITIZER_COMPILE_OPTIONS)
     set(RJ_SANITIZER_LINK_OPTIONS)
-    list(APPEND RJ_SANITIZER_COMPILE_OPTIONS "-fno-sanitize-recover=all")
     message(STATUS "Sanitizer enabled: ${_rj_sanitizer_arg}")
     if(MSVC)
         # MSVC only supports AddressSanitizer; ubsan/tsan/msan are GCC/Clang-only.
@@ -177,6 +176,7 @@ if(_rj_sanitizer_count GREATER 0)
             message(FATAL_ERROR "Only RJ_ENABLE_ASAN is supported by MSVC.")
         endif()
     elseif(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang|AppleClang")
+        list(APPEND RJ_SANITIZER_COMPILE_OPTIONS "-fno-sanitize-recover=all")
         list(
             APPEND RJ_SANITIZER_COMPILE_OPTIONS
             -fsanitize=${_rj_sanitizer_arg}
--- a/emulation/rocjitsu/CMakeLists.txt
+++ b/emulation/rocjitsu/CMakeLists.txt
@@ -164,6 +164,7 @@ if(_rj_sanitizer_count GREATER 0)
     list(JOIN _rj_sanitizer_kinds "," _rj_sanitizer_arg)
     set(RJ_SANITIZER_COMPILE_OPTIONS)
     set(RJ_SANITIZER_LINK_OPTIONS)
+    list(APPEND RJ_SANITIZER_COMPILE_OPTIONS "-fno-sanitize-recover=all")
     message(STATUS "Sanitizer enabled: ${_rj_sanitizer_arg}")
     if(MSVC)
         # MSVC only supports AddressSanitizer; ubsan/tsan/msan are GCC/Clang-only.
--- a/emulation/rocjitsu/lib/python/amdisa/sema_derive.py
+++ b/emulation/rocjitsu/lib/python/amdisa/sema_derive.py
@@ -1206,6 +1206,22 @@ class _VectorTernary(_ScalarDeriver):
             body = _assign(_cast(_dst(0), ty), result)
             return SemaBlock(sem.name, ExecModel.VECTOR, body)
 
+        if op == 'mad' and dtype == 'u16':
+            result = SemaNode(
+                SemaNodeKind.CALL,
+                ty=ty,
+                call_name='mad_lo_u16',
+                children=(
+                    _id('mad_lo_u16'),
+                    _src(0),
+                    _src(1),
+                    _src(2),
+                ),
+            )
+            body = _assign(_cast(_dst(0), ty), result)
+            return SemaBlock(sem.name, ExecModel.VECTOR, body)
+
+
         if op in ('add_min', 'add_max') and dtype in ('i32', 'u32'):
             call_name = f'{op}_{dtype}'
             result = SemaNode(
--- a/emulation/rocjitsu/lib/python/amdisa/sema_derive.py
+++ b/emulation/rocjitsu/lib/python/amdisa/sema_derive.py
@@ -1221,7 +1221,6 @@ class _VectorTernary(_ScalarDeriver):
             body = _assign(_cast(_dst(0), ty), result)
             return SemaBlock(sem.name, ExecModel.VECTOR, body)
 
-
         if op in ('add_min', 'add_max') and dtype in ('i32', 'u32'):
             call_name = f'{op}_{dtype}'
             result = SemaNode(
--- a/emulation/rocjitsu/lib/python/amdisa/sema_derive.py
+++ b/emulation/rocjitsu/lib/python/amdisa/sema_derive.py
@@ -1163,6 +1163,18 @@ class _VectorBinop(_ScalarDeriver):
             body = _assign(_cast(_dst(0), ty), result)
             return SemaBlock(sem.name, ExecModel.VECTOR, body)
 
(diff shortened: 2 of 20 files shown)


emulation/rocjitsu/CMakeLists.txt
```diff
@@ -164,6 +164,7 @@ if(_rj_sanitizer_count GREATER 0)
     list(JOIN _rj_sanitizer_kinds "," _rj_sanitizer_arg)
     set(RJ_SANITIZER_COMPILE_OPTIONS)
     set(RJ_SANITIZER_LINK_OPTIONS)
+    list(APPEND RJ_SANITIZER_COMPILE_OPTIONS "-fno-sanitize-recover=all")
```
accauble: This passes it to all sanitizers, not just ubsan, right? Is that intentional (PR description just mentioned ubsan).

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
```diff
@@ -664,14 +666,16 @@ bool CodeObjectPatcher::redirect_kernel_entry(uint64_t descriptor_file_offset,
...
   const int64_t delta =
       static_cast<int64_t>(new_entry_text_offset) - static_cast<int64_t>(old_entry_text_offset);
-  const int64_t redirected = static_cast<int64_t>(desc->kernel_code_entry_byte_offset) + delta;
+  const int64_t redirected = static_cast<int64_t>(desc.kernel_code_entry_byte_offset) + delta;
   // The descriptor field is signed because the entry point may be before or
   // after the descriptor in virtual address order. Preserve that signed value
   // when applying the text-relative delta.
-  desc->kernel_code_entry_byte_offset = redirected;
+  desc.kernel_code_entry_byte_offset = redirected;
+  std::memcpy(image_.data() + descriptor_file_offset, &desc, sizeof(desc));
```
bjacob, line 678: `sizeof` here is the operator, so the `()` around `desc` isn't doing anything (but also doesn't hurt).

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/vop3_alu_2.cpp
```diff
@@ -413,14 +413,12 @@ void VAddNcI16Vop3::execute_impl(amdgpu::Wavefront &wf) {
...
-      uint32_t src_half =
-          static_cast<uint32_t>(static_cast<uint16_t>(static_cast<uint32_t>(static_cast<uint16_t>(
-              static_cast<int16_t>((static_cast<int16_t>(((amdgpu::vop3_opsel(inst_) & 0x1u) != 0
-                                                              ? (src0.read_lane(wf, lane) >> 16)
-                                                              : src0.read_lane(wf, lane))) +
-                                    static_cast<int16_t>(((amdgpu::vop3_opsel(inst_) & 0x2u) != 0
-                                                              ? (src1.read_lane(wf, lane) >> 16)
-                                                              : src1.read_lane(wf, lane)))))))));
+      uint32_t src_half = static_cast<uint32_t>(
+          static_cast<uint16_t>(static_cast<uint32_t>(static_cast<uint16_t>(static_cast<int16_t>(
```
bjacob, line 417: I don't understand why we have these redundant chains of casts. Is it an artifact of being python-generated?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/vop3_ternary.cpp
```diff
@@ -3938,16 +3938,17 @@ void VMadU16Vop3::execute_impl(amdgpu::Wavefront &wf) {
...
-          static_cast<uint16_t>(((static_cast<uint16_t>(((amdgpu::vop3_opsel(inst_) & 0x1u) != 0
-                                                             ? (src0.read_lane(wf, lane) >> 16)
-                                                             : src0.read_lane(wf, lane))) *
-                                  static_cast<uint16_t>(((amdgpu::vop3_opsel(inst_) & 0x2u) != 0
-                                                             ? (src1.read_lane(wf, lane) >> 16)
-                                                             : src1.read_lane(wf, lane)))) +
-                                 static_cast<uint16_t>(((amdgpu::vop3_opsel(inst_) & 0x4u) != 0
-                                                            ? (src2.read_lane(wf, lane) >> 16)
-                                                            : src2.read_lane(wf, lane))))))));
+      uint32_t src_half = static_cast<uint32_t>(static_cast<uint16_t>(
```
accauble, line 3941: The number of casts here boggles the mind lol

accauble (approved):
Thanks for the fixes. Only one comment making sure that you wanted to pass the flag to all sanitizers not just ubsan.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/vop3_alu_2.cpp
amd-eochoalo: Yes, I think the auto-generator assumes that operands will be needed in signed types, but then we cast them to unsigned, and then the result is also casted to unsigned. I think we could clean this up a bit. Let me give it a try.

emulation/rocjitsu/CMakeLists.txt
amd-eochoalo: This is intentional. The PR just mentions this for ubsan because ubsan is the one that triggered errors. But this will show errors for all sanitizers.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/vop3_alu_2.cpp
amd-eochoalo: @bjacob here are some changes that reduce the amount of casting. https://github.com/ROCm/rocm-systems/pull/8325/changes/6e2ce47bac14f87b616efdf6baad21ad4187c076 I just outlined the function. I think there are some other changes that are possible to do to prevent other redundant casting, but I think those are beyond the scope of this PR. cc: @accauble

kuhar (commented):
A few comments

emulation/rocjitsu/CMakeLists.txt
kuhar: nit: The risk here is that `-fno-sanitize-recover=all` is appended before the `if(MSVC)` branch, so future MSVC ASan builds will also pass that GCC/Clang-style `-f` option through `add_compile_options`. Could we move it into the GNU/Clang sanitizer path, or guard it by compiler, so Windows ASan does not get an unsupported flag?

emulation/rocjitsu/lib/python/amdisa/sema_derive.py
```diff
@@ -1163,6 +1163,18 @@ def derive(sem: InstructionSemantics) -> SemaBlock:
             body = _assign(_cast(_dst(0), ty), result)
             return SemaBlock(sem.name, ExecModel.VECTOR, body)
 
+        if op in ('add', 'sub', 'subrev', 'rsub') and ty.base == 'I':
```
kuhar, line 1166: It would be useful to add focused tests for the new UB-safe integer lowering paths. `sema_derive.py:1166` now special-cases signed vector `add`/`sub`/`subrev`/`rsub`, but `TestDeriveVectorBinop` still only covers `f32`, `u32`, `b32`, and `i24`, so a regression back to signed operands would not be caught. While touching this area, please also tighten `test_u16_mad_widens_before_multiply` at `test_sema_derive.py:1182`: it only proves `src0` is not multiplied while narrow, but should also pin down `src1` and the addend `src2` so all inputs are widened before arithmetic.

emulation/rocjitsu/CMakeLists.txt
amd-eochoalo: Thanks! https://github.com/ROCm/rocm-systems/pull/8325/commits/dce95cc7db91078c8c9eca6246b724ef5d384897

emulation/rocjitsu/lib/python/amdisa/sema_derive.py
amd-eochoalo: >  it only proves src0 is not multiplied while narrow, but should also pin down src1 and the addend src2 so all inputs are widened before arithmetic.

I think once I outlined the function, we cannot prove it directly, right?

https://github.com/ROCm/rocm-systems/pull/8325/commits/b73d5867e7e8cbe54691d2758115f4974add8dd8

PR #8334, opened 2026-07-09
feat(rocprofiler-compute): normalize PC-sampling analysis DB schema

vedithal-amd:
## Motivation

Replace the flat compute_pcsampling table with a normalized schema so PC-sampling data links cleanly to code objects and kernels in the analysis database. Sampled instructions are attributed per-kernel and carry typed instruction-class and stall-reason breakdowns. The change also tightens integrity across the existing analysis DB tables.

## Technical Details

- New compute_-prefixed tables: code_object_store, instruction_line, pc_sample_state, pc_sample_stall_reason(_type), instruction_sample(_type).
- instruction_line.kernel_uuid attributes each line to its kernel via pc_sample -> dispatch_id -> kernel_id correlation; it is non-nullable, so lines whose kernel is filtered out or unmapped are dropped rather than attributed to no kernel.
- Populated from ps_file_results.json (pid, load_base catalog).
- Add a compute_pc_sampling view flattening the normalized tables to the old column shape (stall_reason rebuilt as a JSON dict); CLI output is unchanged.
- CLI and DB share one load/aggregate/enrich path.
- Add unique constraints across the analysis DB tables and write the metadata row once per database.
- SCHEMA_VERSION bumped to 2.0.0 (breaking); schema/views .mmd and .png updated.

## JIRA ID

JIRA ID: AIPROFCOMP-540

## Test Plan

- ORM unit tests: median view, unique constraints, pc_sampling view, JSON sanitizer.
- PC-sampling unit tests: inst_type capture, shared aggregation, code-object tree.
- DB population unit tests: lookup dedup, no-tool-data no-op, per-kernel attribution, dropping lines whose kernel is filtered out.
- End-to-end analyze --output-format db asserting per-table row counts and child-table sums equal pc_sample_state totals.
- Validated CLI vs DB pc_sampling view row-for-row consistent on rocflop_pcsampling.

## Test Result

tests pass; ruff clean.

## Submission Checklist

- [ ] Look over the contributing guidelines


projects/rocprofiler-compute/src/utils/analysis_orm.py
```diff
@@ -146,27 +149,147 @@ class Kernel(Base):
...
 
 
-class PCsampling(Base):
-    __tablename__ = f"{PREFIX}pcsampling"
+class CodeObjectStore(Base):
+    __tablename__ = f"{PREFIX}code_object_store"
+    __table_args__ = (UniqueConstraint("workload_id", "pid", "code_object_id"),)
 
-    pc_sampling_uuid = Column(Integer, primary_key=True)
+    id = Column(Integer, primary_key=True)
```
vedithal-amd: shouldnt it be something like code_object_uuid for consistency - check this everywhere

vedithal-amd: Done in 3b91cfb0e50: renamed all new-table PKs/FKs to the <entity>_uuid convention (code_object_uuid, instruction_uuid, pc_sample_state_uuid, etc.).

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
```diff
@@ -399,42 +378,96 @@ def calc_roofline_ceilings(self) -> dict[str, dict[str, Any]]:
...
+            return
 
-        for workload_path in self._runs.keys():
-            pc_sampling_data = tool_data_per_workload.get(workload_path)
-            if pc_sampling_data is None:
-                console_warning(f"PC sampling data not found for {workload_path}.")
-                continue
+        pid = tool_data.get("metadata", {}).get("pid")
+        stall_reason_types: dict[str, orm.PCSampleStallReasonType] = {}
+        instruction_sample_types: dict[str, orm.InstructionSampleType] = {}
```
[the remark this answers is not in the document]
vedithal-amd: Fixed in 002a2c742d0. Moved dedup into Database.get_or_create_type with a session-scoped cache seeded in init(), so types dedup DB-wide across workloads. Added a test asserting a reused text after commit creates no duplicate.

projects/rocprofiler-compute/tests/conftest.py
```diff
@@ -214,6 +216,16 @@ def require_triton(*, gpu: bool = False) -> None:
...
 
+@pytest.fixture
+def db_session():
+    """An initialized in-memory analysis database, torn down after the test."""
+    Database.init(":memory:")
+    yield Database.get_session()
+    Database._session.close()
+    Database._session = None
+    Database._engine = None
+
```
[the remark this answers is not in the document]
vedithal-amd: Fixed in 002a2c742d0. The fixture now calls Database._engine.dispose() in teardown before dropping the references, so the StaticPool connection is released deterministically.

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
```diff
@@ -404,42 +383,74 @@ def calc_roofline_ceilings(self) -> dict[str, dict[str, Any]]:
...
+                self._add_instruction_line(line, code_object_store, kernel_objs)
+
+    @staticmethod
+    def _add_instruction_line(
+        line: InstructionLineRecord,
+        code_object_store: orm.CodeObjectStore,
+        kernel_objs: dict[str, orm.Kernel],
+    ) -> None:
+        """Insert one instruction line, its sample state, and child counts."""
+        instruction_line = orm.InstructionLine(
```
abchoudh-amd, line 422: With kernel filtering, instruction lines are attributed to a NULL kernel.

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
```diff
@@ -94,18 +97,15 @@ def pre_processing(self) -> None:
             )
 
         self._roofline_ceilings_per_workload = self.calc_roofline_ceilings()
-        pc_sampling_tool_data = (
+        self._pc_sampling_tool_data_per_workload = (
```
abchoudh-amd, line 100: Should the raw PC sampling data be persisted for the whole lifetime of the object? Could this be lazily loaded and deleted once the data is extracted to the database?

abchoudh-amd (changes requested):

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
vedithal-amd: The analysis DB design accumulates all workload data first, then populates the DB in a single pass. Freeing tool data per-workload would only help peak memory during multi-workload runs, and doing it properly means interleaving extraction and DB writes in one loop, which cuts against that core design. After add_pc_sampling_data there's only metric insertion, metadata, commit, and write before the process exits, so the tool data is reclaimed moments later regardless. CLI mode holds tool data just as long (via a local in the analysis loop). Leaving as-is; happy to revisit as a separate optimization if peak memory becomes a concern.

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
vedithal-amd: Fixed in 8a17aceb85. kernel_uuid is now non-nullable and _add_instruction_line skips any line whose kernel is absent from kernel_objs (filtered out or unmapped), so the analyze-time filter is honored and no NULL-kernel rows are inserted. Added a DB test for the drop path.

abchoudh-amd (approved):

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
```diff
@@ -37,16 +37,53 @@ erDiagram
...
+        int workload_id FK
+        int pid
+        int code_object_id
+        int load_base
+    }
+
+    compute_instruction_line {
+        int instruction_uuid PK
+        int code_object_uuid FK
         int kernel_uuid FK
```
svolkov-amd, line 51: Whi kernel_uuid is required for instruction when it's the same for the whole code_object. Should be part of code_object description in my opinion.

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
```diff
@@ -37,16 +37,53 @@ erDiagram
...
-        string source
-        string instruction
+        int code_object_offset
+        text comment
+        text instruction
+    }
+
+    compute_pc_sample_state {
```
svolkov-amd, line 57: The following fields are missing:
- dispatch_id
- active_threads_percent
- wave_occupancy_percent

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
```diff
@@ -37,16 +37,53 @@ erDiagram
...
+        int pc_sample_state_uuid PK
+        int instruction_uuid FK
+        int total_count
+        int issue_count
+        int stall_count
+    }
+
+    compute_pc_sample_stall_reason_type {
+        int pc_sample_stall_reason_type_uuid PK
+        string text
```
svolkov-amd, line 67: `description` field is missing.

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
```diff
@@ -37,16 +37,53 @@ erDiagram
...
+        int load_base
+    }
+
+    compute_instruction_line {
+        int instruction_uuid PK
+        int code_object_uuid FK
         int kernel_uuid FK
-        string source
-        string instruction
+        int code_object_offset
```
svolkov-amd, line 52: `instruction_type` is missing.

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
```diff
@@ -37,16 +37,53 @@ erDiagram
         int end_timestamp
     }
 
-    compute_pcsampling {
-        int pc_sampling_uuid PK
+    compute_code_object_store {
+        int code_object_uuid PK
+        int workload_id FK
+        int pid
```
svolkov-amd, line 43: `pid` should belong to workload descriptor. `workload_id` should belong to kernel descriptor. Instead it seems logical for `code_object_store` to reference only a kernel via `kernel_uuid`.

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
```diff
@@ -37,16 +37,53 @@ erDiagram
...
+    }
+
+    compute_pc_sample_stall_reason {
+        int pc_sample_stall_reason_uuid PK
+        int pc_sample_state_uuid FK
+        int pc_sample_stall_reason_type_uuid FK
+        int count
+    }
+
+    compute_instruction_sample_type {
```
svolkov-amd: What is `compute_instruction_sample_type` and `compute_instruction_sample` structures about? They seem duplicates of `compute_pc_sample_stall_reason_type` and `compute_pc_sample_stall_reason`.

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
svolkov-amd: Also based on code it seems like we have information about `dispatch_id` in pc sampling output.

projects/rocprofiler-compute/src/utils/pc_sampling_analysis.py
```diff
@@ -167,15 +194,57 @@ def enrich_with_metadata(
...
+    """Convert one aggregated dataframe row into an InstructionLineRecord."""
+    return InstructionLineRecord(
+        code_object_offset=int(row.code_object_offset),
+        kernel_name=row.kernel_name,
+        instruction=row.instruction,
+        comment=row.source_line,
+        total_count=int(row.count),
+        issue_count=None if pd.isna(row.count_issued) else int(row.count_issued),
+        stall_count=None if pd.isna(row.count_stalled) else int(row.count_stalled),
+        stall_reasons=row.stall_reason if isinstance(row.stall_reason, dict) else {},
```
svolkov-amd: Can we get rid of `isinstance`? Usage of `isinstance` is a very bad code pattern as it adds a coupling by data types. We should try realy-realy hard to avoid it as much as possible. Instead we should use a polimorphism, aka have an interface which is implemented differently for different data types.

projects/rocprofiler-compute/src/utils/pc_sampling_analysis.py
```diff
@@ -167,15 +194,57 @@ def enrich_with_metadata(
...
+                load_base=catalog_entry.get("load_base"),
+                instruction_lines=[
+                    _to_instruction_line_record(row) for row in group.itertuples()
+                ],
+            )
+        )
+    return records
+
+
+def _to_instruction_line_record(row: Any) -> InstructionLineRecord:  # noqa: ANN401
```
svolkov-amd, line 235: Can we define a concrete type instead of `Any`? Using `Any` basically makes any benefits from type hints useless. And also makes it harder to comprehend the code.

projects/rocprofiler-compute/src/utils/pc_sampling_analysis.py
```diff
@@ -204,3 +273,16 @@ def _aggregate_stall_reason(stall_reason: pd.Series) -> Optional[dict[str, int]]
...
     return valid.value_counts().to_dict()
+
+
+def _aggregate_inst_type(inst_type: pd.Series) -> Optional[dict[str, int]]:
+    """Count sampled instruction types as a descending {type: count} dict.
+
+    None when no inst_type info exists. The raw SDK enum prefix is stripped so
+    stored values read as VALU, FLAT, NO_INST, etc.
+    """
+    present = inst_type.dropna()
```
svolkov-amd, line 284: How the same instruction could have different instruction types?

svolkov-amd (changes requested):

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
```diff
@@ -37,16 +37,53 @@ erDiagram
         int end_timestamp
     }
 
-    compute_pcsampling {
-        int pc_sampling_uuid PK
+    compute_code_object_store {
+        int code_object_uuid PK
+        int workload_id FK
+        int pid
+        int code_object_id
```
svolkov-amd, line 44: Code object `uri` is missing. It should be present in native collector output.

projects/rocprofiler-compute/src/utils/pc_sampling_analysis.py
vedithal-amd: We PC-sampled the `rocflop` test workload (stochastic sampling, 71,436 samples) to check this.

The sampled instruction-type totals are: VALU 39367, MATRIX 16424, FLAT 13713, NO_INST 1589, SCALAR 232, BRANCH_TAKEN 98, BRANCH_NOT_TAKEN 13.

The key point: `inst_type` is not a static property of the instruction at a PC. It's a property of the **dynamic sample** taken at that PC, so the same instruction can be sampled with different types. Two causes, both present in the data:

**1. NO_INST (76 of 85 multi-type instructions).** `NO_INST` means the sample landed on the PC while no instruction was retiring in that slot (wave stalled / not issuing). The same PC is tagged with its real type when it issues and `NO_INST` when caught mid-stall:
- `v_pk_fma_f32 ...` -> VALU:93, NO_INST:19
- `v_mfma_f32_16x16x4_f32 ...` -> MATRIX:299, NO_INST:19
- `v_mul_lo_u32 ...` -> NO_INST:82, VALU:52 (higher-latency integer multiply, stalls more than it retires)

**2. Branch outcome (the `s_cbranch_scc0` instructions).** The type records the runtime branch result, so one static branch carries both:
- `s_cbranch_scc0` -> BRANCH_TAKEN:5, BRANCH_NOT_TAKEN:2

Collapsing to one type per PC would lose the issue/stall and taken/not-taken breakdown, so the `{type: count}` dict aggregation is the correct model here.

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
vedithal-amd: I PC-sampled the `rocflop` test workload to look into this.

A code object can hold many kernels: in `rocflop`, code_object 1 maps to 13 kernels and code_object 2 to 7. So the kernel isn't constant per code object, and can't be a property of `code_object_store`.

Instead each sampled instruction is attributed to the kernel of the dispatch that was executing when the sample fired (`dispatch_id -> kernel_id` from `kernel_dispatch` records). That's a per-instruction-line property, so `kernel_uuid` stays on `compute_instruction_line`.

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
vedithal-amd: I PC-sampled the `rocflop` test workload to look into this.

A code object can hold many kernels: in `rocflop`, code_object 1 maps to 13 kernels and code_object 2 to 7. So the kernel isn't constant per code object, and can't be a property of `code_object_store`.

Instead each sampled instruction is attributed to the kernel of the dispatch that was executing when the sample fired (`dispatch_id -> kernel_id` from `kernel_dispatch` records). That's a per-instruction-line property, so `kernel_uuid` stays on `compute_instruction_line`.

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
vedithal-amd: `compute_instruction_sample` and `compute_pc_sample_stall_reason` are per-state count tables, one row per (state, type) with a count.

`compute_instruction_sample` counts how many times an instruction was sampled as each instruction type. As noted on the earlier thread, instruction type is a dynamic property of the PC sample, not a static property of the instruction, so one instruction can be counted under several types. For example in `rocflop`, `v_mfma_f32_16x16x4_f32 v[12:15], v18, v18, v[12:15]` was counted as MATRIX 299 and NO_INST 19 (MATRIX when it issues, NO_INST when caught mid-stall).

`compute_pc_sample_stall_reason` similarly counts the different kinds of stalls sampled for an instruction. The same instruction above had ARBITER_NOT_WIN 293, NO_INSTRUCTION_AVAILABLE 19, OTHER_WAIT 6.

`compute_instruction_sample_type` and `compute_pc_sample_stall_reason_type` are their lookup tables, holding each distinct string once. They dedup different string sets, so they aren't duplicates.

Will rename the lookup tables to `*_lookup` to make the normalization clearer.

projects/rocprofiler-compute/src/utils/pc_sampling_analysis.py
vedithal-amd: `row` is a `df.itertuples()` namedtuple. The dataframe is built from `tool_data` (parsed from `ps_file_results.json`), and its cell values are genuinely dynamic (ints, floats, NaN, or dicts depending on the column), so `Any` is the honest annotation here. It matches the `dict[str, Any]` we already use for `tool_data` throughout this file (`load_pc_sample_records`, `enrich_with_metadata`, `load_aggregated_pc_sampling`); those don't need `# noqa` only because `Any` is nested in a concrete `dict`, whereas a bare parameter annotation trips ANN401.

projects/rocprofiler-compute/src/utils/pc_sampling_analysis.py
vedithal-amd: The `isinstance(..., dict)` was misleading here, sorry: it reads like several types are possible and we're dispatching on them, but the value is only ever the `{reason: count}` dict or NaN (`_aggregate_stall_reason`/`_aggregate_inst_type` return `None` for host_trap samples, which pandas stores as NaN). So there's no polymorphism, just a present-vs-missing case.

Switched to a `pd.isna` missing-value guard, which states that intent directly and matches the `pd.isna` we already use for `count_issued`/`count_stalled` two lines up:
```python
stall_reasons={} if pd.isna(row.stall_reason) else row.stall_reason,
inst_types={} if pd.isna(row.inst_type) else row.inst_type,
```

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
vedithal-amd: Thanks for flagging. `uri` here is binary provenance (`file:///.../rocflop#offset=...` or `memory://<pid>#...`), pointing at the compiled/JIT code object, not source. I don't surface it in the pc_sampling view or the per-kernel ISA view, and nothing else in the DB or CLI consumes it, so it's a code-object-internal detail with little bearing on analysis. The code object is already uniquely identified by `(workload_id, pid, code_object_id)`, so `uri` adds no identity either.

The provenance that helps analysis is source-of-ISA, from the `comment` field in the code-object info (sources snapshotted), which is scoped to a separate source-to-ISA mapping PR.

Unless I'm missing a use case you have in mind, I'd leave `uri` out here. Happy to add it if there's a consumer I haven't considered.

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
vedithal-amd: dispatch_id is in the raw records, but I use it only as a correlation key: `dispatch_id -> kernel_id`, to attribute each sample to its kernel, then it's dropped. The PC-sampling analysis and its view care about the kernel (kernel_id / kernel_name), not the dispatch. dispatch_id isn't something the analysis queries on.

Linking `compute_pc_sample_state` to `compute_dispatch` would also be awkward, because the relationship is many-to-many: one dispatch sprays samples across many instruction offsets (many sample-state rows), and one offset is sampled across a kernel's repeated dispatches (many dispatches per sample-state row). So there's no single dispatch to point at, and `compute_dispatch` is populated on its own path. Sample states already relate to dispatches transitively through the shared kernel, which is what matters for PC-sampling analysis.

Leaving dispatch_id off `pc_sample_state`.

vedithal-amd: For active_thread_percent and wave_occupancy_percent:

These are derivable per-instruction-line metrics: `active_threads_pct` from `exec_mask` popcount (÷ wave_size, both sampling modes) and `wave_occupancy_pct` from `wave_cnt` (÷ max_waves_per_cu, stochastic only; nullable/None for host-trap since the host-trap record has no `wave_cnt`).

Denominators come per-device from the machine spec, not literals. Both aggregate to `compute_pc_sample_state` as a simple per-sample mean: sampling is time-proportional (longer-running instructions accumulate more samples), so sample count already carries the time weight,

They will also surface in the `compute_pc_sampling` view.

Deferring to a separate PR to keep this one scoped to normalization; also noted on the JIRA.

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
vedithal-amd: instruction_type isn't a static property of the instruction, so it isn't a column on `compute_instruction_line`. As explained [on the earlier thread](https://github.com/ROCm/rocm-systems/pull/8334#discussion_r3606888888), it's dynamic sampling information: the same instruction is sampled as different types (e.g. VALU when it issues, NO_INST when caught mid-stall).

It's modeled in `compute_instruction_sample`, which points to a `compute_pc_sample_state` and carries a `{type: count}` breakdown, with the type string normalized via `compute_instruction_sample_lookup`. So a single instruction's per-type counts live there, not as one field on the instruction line.

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
vedithal-amd: The tool data has no stall-reason descriptions, only the raw enum name (e.g. `ARBITER_NOT_WIN`). The SDK exposes a `..._name` lookup, not a description.

The authoritative meanings live in the rocprofiler-sdk docs: [the "Stall reasons" table in cdna3-cdna4-pc-sampling.rst](https://github.com/ROCm/rocm-systems/blob/develop/projects/rocprofiler-sdk/source/docs/how-to/cdna3-cdna4-pc-sampling.rst#L16-L51).

I'd rather not hardcode an enum-to-description map here. Stall reasons can be added, removed, or redefined by the SDK, so a copied table drifts silently. Describing an enum isn't the schema's job; the table stores the authoritative `text` (the reason name).

The better home is our PC-sampling docs, linking to the SDK table. Deferring that to a separate PR; also noting it on the JIRA.

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
svolkov-amd: My concern with direct relation of ISA lines to kernel is that there could be cases where this will stop being 1-to-1 relation. I've outlined this in HLD in this [section](https://amd.atlassian.net/wiki/spaces/AGSRCIT/pages/1608843848/HLD+PC+Sampling+in+rocprof-compute#Analysis%3A-Analysis-database-schema-for-PC-sampling-data):
```
“kernel<->code_object” is “one to many” relationship. Each kernel could potentially have several code objects. For example, due to compilation into different versions. This is the case for graphics; it could be or become the case for AI/HPC workloads eventually. The main reasons for several versions is different states or runtime optimization of kernel - [New in D3D12 – background shader optimizations - DirectX Developer Blog](https://devblogs.microsoft.com/directx/background-shader-optimizations/).
```

Therefore, I would prefer us to have some "intermediate" between a kernel and its source. Because in general case a single kernel could have different ISA representation. We could call this "intermediate" a `compute_kernel_binary` instead of `compute_code_object_store` with the following structure:
```
compute_kernel_binary {
    int compute_kernel_binary_uuid PK
    int kernel_uuid FK
    int code_object_id
    int load_base
}
```

So for a code_object with 13 kernels there will be 13 individual `compute_kernel_binary` tables. If the same kernel is compiled into different code objects, then Optiq will be allow the user to select different versions of it.
`pid` still should belong to some other table - `workload` or a separate `process` table. `workload_id` could be removed this way as it's a part of `compute_kernel` description.

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
svolkov-amd: Agree, we could omit this.

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
svolkov-amd: Resolving this as duplication because discussion on this decision is continued in a previous comment.

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
svolkov-amd: Regarding `dispatch_id`:
- We should start adding per-dispatch analysis types in rocprof-compute. PC-sampling is an easy target to start doing so, as it's collected via a single run. The reason for this is because performance profile will be different for different dispatches, due to differences in states, data sizes and locality. Therefore, we need to aggregate `pc_sample_state` per dispatch. And Optiq should allow to filter out PC sampling data for concrete kernel+dispatch. 
- Regarding the statement that `Linking compute_pc_sample_state to compute_dispatch would also be awkward, because the relationship is many-to-many: one dispatch sprays samples across many instruction offsets` - not sure it makes sense to me. `compute_pc_sample_state` already defines a state for a concrete instruction, so it can't "spray across many instruction offsets". It however aggregates across multiple dispatches for this instruction. The ask is to not aggregate across dispatches and put separate entires for this instruction per-dispatch.

svolkov-amd: Regarding active_thread_percent and wave_occupancy_percent: totally fine deferring it into a separate PR, see your updates in the JIRA ticket.

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
svolkov-amd: Someone will have to add these descriptions - either we or Optiq, so that they could be shown inside the tool. Showing explanations to stall reasons in tooltip inside UI is much more convenient that forcing the user to search this in the docs. I would rather prefer us to manage this information, then Optiq. Fine to defer this into a separate PR. Updated corresponding item in JIRA to reflect this.

projects/rocprofiler-compute/src/utils/pc_sampling_analysis.py
svolkov-amd: Thanks for the explanation, now I understand where the confusion comes from. SDK calls it `inst_type` but in fact it's not as it's missing the actual instruction type (VALU, MATRIX, BRANCH) and execution state (NO_INST, BRANCH_TAKEN, BRANCH_NOT_TAKEN). Also, it defines this for sampled instructions only. However, each and every instruction (even non-sampled) has a static type defined during compilation phase. Optiq is supposed to use it to color the instructions per each type differently. 

Therefore, we still need to add `instruction_type` field per each instruction and write code which defines it by instruction name (`v_mfma` -> Matrix, `v_` -> VALU, `s_cbranch` -> Branch, `s_` -> Scalar).

At the same time instruction state information at PC sampling time is also useful but I would rename the corresponding fields/structure to `instruction_state` or similar. And would rename `VALU`, `MATRIX` and similar fields to `ISSUED_TO_VALU`, `ISSUED_TO_MATRIX`, etc. To a certain extent, this level of data analysis and processing is an expected outcome of our expertise as a team in order to make profiling easier for the end-user.

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
svolkov-amd: Resolving this conversation as duplication of conversation in `pc_sampling_analysis.py`.

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
svolkov-amd: Added this to JIRA as well.

projects/rocprofiler-compute/src/utils/pc_sampling_analysis.py
svolkov-amd: Added to JIRA as well

svolkov-amd (commented):
There are still things which needs to be resolved. But we've logged them in https://amd-hub.atlassian.net/browse/AIPROFCOMP-540, so conditionaly approving.

svolkov-amd (approved):

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
vedithal-amd: Regarding `dispatch_id`

We need to start honoring --dispatch option in PC Sampling analysis, currently we don't.

I am not sure what is the use case of showing dispatch information in pc sampling analysis and even if there is use case how do we show that - since one instruction or pc sampling state can have thousands of dispatches which is not feasible to show. If we ever want to do this- we don't need to change schema and can link the following tables pc_sampling_state → instruction_line → kernel → dispatch

Noted this in JIRA and will defer to future PR and marking this conversation as resolved

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
vedithal-amd: @svolkov-amd  I hope you have logged ticket with SDK team to expose stall reason enum to full text description in form of a public API which we can use. I had rather not define this map in rocprofiler-compute since SDK seems source of truth for this.

When we incorporate pc sampling into our native tool - then we can leverage this api to dump the mapping in the profiling output and analysis can populate this in analysis database.

Marking this as resolved

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
vedithal-amd: We already have `instruction_line` table containing foreign keys to `code_object_id` and `kernel_id` to correlate many to many mapping between code objects and kernels. I don't think we need to change schema for this. But we could add `kernel_binary` table for clarity.

Also kernel <-> code object seems like many to many mapping not one to many.

Resolving.

projects/rocprofiler-compute/src/utils/pc_sampling_analysis.py
vedithal-amd: Resolving

projects/rocprofiler-compute/src/utils/pc_sampling_analysis.py
vedithal-amd: Good call and agreed (will defer to another PR as noted in JIRA)

But static instruction type is responsibility of ISA collection in native tool and we need to check if disassembly API provides this static instruction type per instruction. This should be present in code object info JSON.

Resolving

PR #7173, opened 2026-06-12
(fix) [rhdc] Fix command injection via LIBDIR_MAX_DEPTH

superm1:
check_libraries() interpolated the LIBDIR_MAX_DEPTH environment variable into a find command run with shell=True, allowing arbitrary command execution as root via sudo -E. Validate the value as a non-negative integer and invoke find with an argv list and shell=False.

JIRA ID: SWSPLAT-24266

PR #8339, opened 2026-07-09
fix: [rdhc] Fix command injection via LIBDIR_MAX_DEPTH

superm1:
check_libraries() interpolated the LIBDIR_MAX_DEPTH environment variable into a find command run with shell=True, allowing arbitrary command execution as root via sudo -E. Validate the value as a non-negative integer and invoke find with an argv list and shell=False.

JIRA ID: SWSPLAT-24266


arvindcheru (commented):
Re-reviewed as follow-up to #7173. The LIBDIR_MAX_DEPTH hardening matches prior verification: injection/negative values rejected, valid depths normalized, find uses argv + shell=False.

#8339 improvements over #7173:

Whitespace-only LIBDIR_MAX_DEPTH treated as unset (unit test added)
ldd replaced with readelf -d -- (non-executing; safer under sudo -E)
Tests moved to tests/test_rdhc.py

solaiys:
## Review: command-injection hardening + `ldd` → `readelf` rewrite

The core security fix is solid — no remaining `shell=True` in the changed paths, `find`/`readelf` use argv lists, `--` guards the filename, and `LIBDIR_MAX_DEPTH` validation is correct (normalizes non-ASCII digits / whitespace, treats whitespace-only as unset). Both prior Copilot comments are already addressed.

My remaining concern is that replacing `ldd` with a hand-rolled `readelf` + manual path search **reimplements the dynamic linker imperfectly**, which introduces functional regressions in the check itself.

### Potential issues

- 🟠 **Transitive dependencies are no longer checked (coverage regression)** — `rdhc.py` (`_check_rocm_libs_dependency`, NEEDED parsing)
  `ldd` reported the full *recursive* dependency closure; `readelf -d` only lists direct `(NEEDED)` entries. A missing *transitive* library (a dependency of a dependency) will now silently pass. For a dependency-health tool this is the main detection capability being reduced — likely an accepted security tradeoff, but worth confirming it's intended.

- 🟠 **RPATH/RUNPATH is not actually parsed → false "not found" (false FAIL)** — `_get_library_search_paths` (`# 1. RPATH/RUNPATH ... (we'll add lib_dir as proxy)`)
  The search path uses `lib_dir` as a "proxy" for RPATH/RUNPATH. ROCm libraries very commonly resolve deps via `$ORIGIN`-relative rpaths (e.g. `$ORIGIN/../lib`, `$ORIGIN/../../lib64/llvm/lib`) that are **not** `lib_dir`, `LD_LIBRARY_PATH`, `<rocm>/lib(64)`, or the hardcoded system dirs. Those deps will be reported as `=> not found` and produce a spurious FAIL even on a healthy install. Consider reading the actual `DT_RPATH`/`DT_RUNPATH` (already in `readelf -d` output; expand `$ORIGIN`) before falling back.

- 🟠 **ldconfig cache / non-x86_64 multiarch dirs ignored → false "not found"** — `_get_library_search_paths` (`# 4. System library paths`)
  The system path list is hardcoded and x86_64-only. Libraries resolvable via `/etc/ld.so.cache` (`ldconfig`), `/usr/local/lib`, other `/etc/ld.so.conf.d/*` dirs, or aarch64 multiarch (`/usr/lib/aarch64-linux-gnu`) will be flagged as missing. Parsing `ldconfig -p` (read-only, non-executing) would restore parity with `ldd`'s resolution.

- 🟡 **Search-order mismatch makes "outside ROCm path" warnings unreliable** — `_get_library_search_paths` / `_find_library_in_paths`
  `lib_dir` is searched first, but the real loader order is `DT_RPATH → LD_LIBRARY_PATH → DT_RUNPATH → ld.so.cache → defaults`. The path this code resolves a dep to can differ from what the loader actually uses, causing the wrong-path warning to be suppressed or reported against the wrong location. Affects warning accuracy, not FAIL/PASS.

- ⚪ **Dead branch (minor)** — readelf error handling
  `readelf -d` on a file with no dynamic section prints "There are no dynamic sections in this file." to **stdout with exit code 0**, so the `"no dynamic section" in stderr` check inside the `ret_code != 0` block never fires. Harmless (empty `needed_libs` handles it), but misleading.

### Summary

The command-injection hardening (the PR's stated goal) is correct and well-tested. The risk lives in the `ldd → readelf` rewrite: it can produce **false failures** on legitimate installs (missing `$ORIGIN`/rpath and ldconfig resolution) and **misses transitive deps**. None reintroduce a security hole, but the false-positive cases are real behavioral regressions for a diagnostic tool and are worth resolving (or explicitly accepting) before merge. The new tests cover validation and shell-free invocation well, but not dependency *resolution* correctness — adding a case with an `$ORIGIN` rpath dep would catch the biggest risk.

PR #8342, opened 2026-07-09
fix(amdsmi): clean up static profile and mem-carveout CLI output

marifamd:
## Motivation

`amd-smi static --profile` rendered its available-profiles list mangled in human-readable output (items outdented below the following field), and every `amd-smi static` run printed a `MEM_CARVEOUT: N/A` line on hardware without VRAM carveout support. Tidy both.

## Technical Details

**Human-readable logger** (`amdsmi_cli/amdsmi_logger.py`):
- Render scalar list items nested under their key instead of as outdented `- item` bullets, fixing the root cause of the mangled `--profile` output

**static subcommand** (`amdsmi_cli/subcommands/static.py`):
- Drop the profile-specific human-readable workaround; profile now uses one plain list across human-readable, JSON, and CSV
- Omit `MEM_CARVEOUT` from default output on carveout-unsupported hardware; show it only with `-m`/`--mem-carveout` and report a plain `N/A`

**Parser** (`amdsmi_cli/amdsmi_parser.py`):
- Note Linux-Baremetal-only availability in the `--profile` help text

**Tests** (`tests/python_unittest/logger_format_unit_test.py`):
- Add a hardware-free regression test for the scalar-list formatter

## JIRA ID

JIRA ID: AILITOOLS-281

## Test Plan

- `python3 tests/python_unittest/logger_format_unit_test.py` (hardware-free)
- Manual `amd-smi static --profile` / `--json` / `--csv`, and `static` / `static -m`, on an MI210 + Navi31 host
- `pre-commit run` on all changed files

## Test Result

- New unit test: 3 passed
- Human-readable `--profile` list nests correctly; JSON/CSV unchanged
- `MEM_CARVEOUT` absent from default output; `-m` shows plain `N/A`; pre-commit clean

## Submission Checklist

### DIFF
--- a/projects/amdsmi/CHANGELOG.md
+++ b/projects/amdsmi/CHANGELOG.md
@@ -17,6 +17,9 @@ Full documentation for amd_smi_lib is available at [https://rocm.docs.amd.com/pr
 - **`amd-smi static` now omits `MEM_CARVEOUT` from default human-readable output on hardware without VRAM carveout support**.  
   - On a default run the human-readable output drops the section entirely instead of printing a verbose reason string; JSON and CSV keep a stable `N/A` key. An explicit `-m`/`--mem-carveout` always reports `N/A` in every format. VRAM carveout is only exposed on carveout-capable APUs.
 
+- **Expanded `amd-smi static --mem-carveout` CSV output**.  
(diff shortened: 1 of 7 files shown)


yalmusaf (approved):
Tested on APU systems, output looks good.

adam360x (approved):

gabrpham (changes requested):
Unit test is in the wrong directory? My understanding is that the common directory is meant to be a place for common test functions which are used across many or all the other actual tests and therefore CI and/or the testing suite wouldn't pick up any tests that are in the common directory.

PR #8344, opened 2026-07-09
fix(rocjitsu): D16 Loads Count as Partial Defs

theSK2005:
## Motivation
D16 memory loads are 16 bit writes into VGPRs. These operations currently are seen as full defs by liveness analysis since VDST is only a dst and not a src and has a 32 bit operand size so PR #7979 won't cover it.

## Technical Details

Added check in _generator.py's _dst_is_also_source to cover d16 load operations and have codegen write VDST as both a dst and a src.

Fixed existing semantics.py bug that didn't use _BUFFER_FORMAT_MAP for derive_mubuf which left many instructions unimplemented and not properly flagged.

## Test Plan

- Verify InstDefUse on decoded D16 instructions sees VDST as a src and a dst.
- Verify all other tests pass.

## Test Result
All tests pass

## Submission Checklist


emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -1744,15 +1748,26 @@ def _dst_is_also_source(self, inst: Instruction) -> bool:
...
-        return (
+        if (
             sem.operation in self._READS_DST_OPS
             or sem.semantic_class in self._READS_DST_CLASSES
+        ):
+            return True
+        return (
+            (sem.d16_hi or sem.d16_lo)
+            and sem.num_elems == 1
+            and sem.semantic_class in self._D16_LOAD_CLASSES
```
kuhar, line 2924: I think we should make `_dst_is_also_source()` tolerate semantic objects that do not define the new D16 fields, or update the existing test doubles at the same time. `amdisa/tests/test_generator_profile_gates.py::test_bf16_mad_mix_half_updates_read_destination_operand` now fails with `AttributeError: 'types.SimpleNamespace' object has no attribute 'd16_hi'` before it can validate the non-D16 read-dst cases.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/cdna1/flat.cpp
```diff
@@ -708,9 +708,10 @@ FlatLoadUbyteD16Flat::FlatLoadUbyteD16Flat(const MachineInst *inst)
       vdst(32, OperandType::OPR_VGPR, reinterpret_cast<const OpEncoding *>(inst)->vdst),
       addr(64, OperandType::OPR_VGPR, reinterpret_cast<const OpEncoding *>(inst)->addr),
       saddr(0, OperandType::OPR_SREG, 0) {
+  src_operands_[0] = &vdst;
```
kuhar: We should avoid adding the preserved destination half to the printable source operand list. `Instruction::disassemble()` prints every `src_operands_` entry after the destinations, so generated D16 loads like `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/cdna1/flat.cpp:711` now expose the old `vdst` as an extra assembly operand even though it is only an implicit semantic use. It would be safer to model this through `implicit_uses()` or another def/use-only path so decoded syntax does not change.

emulation/rocjitsu/tests/analysis/liveness_test.cpp
```diff
@@ -749,5 +749,54 @@ TEST(InstDefUse, Predicated) {
...
+// so v5 is distinct from the address source. FLAT can be a 3-dword encoding, so
+// the buffer is zero-padded to avoid out-of-bounds reads during decode.
+std::unique_ptr<Instruction> decode_rdna4(std::initializer_list<uint32_t> words) {
+  std::array<uint32_t, 4> buf{};
+  std::copy(words.begin(), words.end(), buf.begin());
+  auto decoder = Decoder::create(ROCJITSU_CODE_ARCH_RDNA4);
+  return std::unique_ptr<Instruction>(decoder ? decoder->decode(buf.data()) : nullptr);
+}
+
+TEST(GeneratedInstDefUse, D16HiLoadReadsDestination) {
```
kuhar, line 4842: It would be useful to add one representative test for each non-FLAT path this semantic change is meant to cover. The current decode tests exercise only RDNA4 FLAT loads, so they miss the buffer `vdata` naming case and do not lock in DS behavior. After the buffer path is fixed, a `buffer_load_d16_*` assertion that `vdata` is a use, a `ds_*_d16` read assertion, and a D16 store negative case would catch the main ways this logic can regress.

kuhar (commented):
The risk here is that `_D16_LOAD_CLASSES` advertises buffer D16 coverage, but the generated source operands still do not include the preserved destination for MUBUF/VBUFFER loads. `_dst_is_also_source()` can return true for `buffer_load_d16_u8`, yet the emission gates at `emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py:1810` and `emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py:5215` only accept output operands named `vdst` or `sdst`; buffer loads use `vdata`. The generated `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/rdna4/vbuffer.cpp:867` still has `vdata` only in `dst_operands_`, with sources starting at `vaddr`, so liveness still treats those D16 buffer loads as full defs. If buffer loads are in scope, please include the `vdata` path and regenerate the affected files; otherwise remove `buffer_load`/`tbuffer_load` from the D16 class set so the contract matches the generated behavior.

kuhar:
There are also some pytest failures

accauble:
@theSK2005 could you rebase please?

newling:
Just to check I understand the description 

> D16 memory loads are 16 bit writes into VGPRs. These operations currently are seen as full defs by liveness analysis since VDST is only a dst and not a src and has a 32 bit operand size so PR https://github.com/ROCm/rocm-systems/pull/7979 won't cover it.

My understanding is that before this PR: 
- the destination register is seen as having 32-bits written, and 0-bits read. 

and then after this PR, it is either option A or B below?
(A) the destination register is seen as having 16-bits written, and other 16-bits read 
(B) (conservative) the destination register is seen as having 32-bits read, then 32-bits written 

This PR caught my attention because I've encountered a semantically problem in a parallel place in rocjitsu https://github.com/ROCm/rocm-systems/pull/8628.

theSK2005:
> Just to check I understand the description
> 
> > D16 memory loads are 16 bit writes into VGPRs. These operations currently are seen as full defs by liveness analysis since VDST is only a dst and not a src and has a 32 bit operand size so PR #7979 won't cover it.
> 
> My understanding is that before this PR:
> 
> * the destination register is seen as having 32-bits written, and 0-bits read.
> 
> and then after this PR, it is either option A or B below? (A) the destination register is seen as having 16-bits written, and other 16-bits read (B) (conservative) the destination register is seen as having 32-bits read, then 32-bits written
> 
> This PR caught my attention because I've encountered a semantically problem in a parallel place in rocjitsu #8628.

This PR doesn't change the operand sizes, so it's still seen as a 32-bit write and 0-bit read. What this PR changes is that the dst register is counted as an implicit use so that they're surfaced to liveness analysis. If you look at the regenerated ISA on this PR, it's all implicit use functions.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -2260,10 +2260,11 @@ def _op_sel_hi_2_expr(self, enc_name: str) -> str:
             'mad_mixhi_bf16',
         }
     )
-    # D16 load semantic classes. Stores share the d16_hi/d16_lo flags, so the
-    # class gate is what restricts the partial-def treatment to loads.
-    # global/scratch loads use the 'flat_load' class.
-    _D16_LOAD_CLASSES = frozenset({'flat_load', 'buffer_load', 'tbuffer_load', 'ds_read'})
+    # D16 load classes. Stores share the d16 flags, so the class gate restricts
+    # the partial-def treatment to loads; global/scratch loads use 'flat_load'.
+    _D16_LOAD_CLASSES = frozenset(
```
accauble, line 2871: Nit: Is there a way to protect this somehow? Something parsable in the ISA to make sure that we don't miss when new ones are introduced?

accauble (commented):
I have one minor concern but otherwise LGTM.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
theSK2005: the d16 fields already exist on develop and I don't see this issue any more

kuhar (commented):
Original anchor: `emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py:2784`

Can we make the FORMAT D16 cases consistent before this lands? The current predicate only admits `num_elems == 1`, and the generated output shows two holes in that model: `tbuffer_load_format_d16_x` gets an `implicit_uses()` override, but `buffer_load_format_d16_x` / RDNA4 `buffer_load_d16_format_x` do not, while odd forms like `tbuffer_load_format_d16_xyz` write 3 halfwords into a 64-bit `vdata` and leave the upper half of the second VGPR preserved. This means liveness can still treat real FORMAT D16 loads as full defs. I think we should either model those preserved reads (for example, `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/cdna3/mubuf.cpp:246`, `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/rdna4/vbuffer.cpp:228`, and `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/cdna3/mtbuf.cpp:443`) and add decode tests for `*_format_d16_x` and `*_format_d16_xyz`, or explicitly document/test why these formats are out of scope.

At `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/vflat.cpp:316`, the D16 preserved-destination read is emitted only as `implicit_uses()`. On gfx1250, `InstDefUse` resolves VGPR implicit reads with `implicit_use_operands()` and then drops `RegClass::VGPR` from the flat `implicit_uses()` result, so this `vdst` dependency is lost for banked VGPR liveness. We should emit the operand-backed `implicit_use_operands()` twin for D16 load destinations on VGPR-MSB profiles, and add a gfx1250 decode/liveness test for this path.

newling:
Agent review at https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr8344.md highlights 2 actionable items

> gfx1250’s bank-aware liveness path does not receive the new use, and legacy MUBUF format-D16 loads are not recognized.

Related to earlier discussion (from 2 weeks ago... apologies for my glacial response time on this) longer term I wonder if we should be modeling this at the byte level rather than the dword (4-byte) level. Depending on what tools (consan?) use this, it might be necessary. For future consideration

newling:
Agent review says there is just 1 outstanding issue from a previous review (not sure whose), please take a look and either address it or say if it's not valid: https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr8344_followup.md#1-preserve-d16-destination-reads-in-gfx1250-bank-aware-liveness

theSK2005:
> Agent review says there is just 1 outstanding issue from a previous review (not sure whose), please take a look and either address it or say if it's not valid: https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr8344_followup.md#1-preserve-d16-destination-reads-in-gfx1250-bank-aware-liveness

yep, I'm working on fixing that atm

newling:
> yep, I'm working on fixing that atm

Thanks. LGTM now, no more comments/questions on my side. I'll let one of the main reviewers listed make any final reviews and stamp when ready

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -6670,6 +6694,33 @@ def gen_insts(self) -> None:
...
+                    # register, so _partial_def_outputs does not catch them).
+                    d16_implicit_use_opnd = None
+                    # Offset (in 32-bit registers) of the one partially-written
+                    # register within the destination: only the last register of
+                    # an odd-count FORMAT load (e.g. format_d16_xyz) is partial.
+                    d16_partial_reg_offset = 0
+                    if self._d16_load_reads_dst(inst):
+                        # FLAT/GLOBAL/SCRATCH/DS name the dest 'vdst'; MUBUF/MTBUF
+                        # (buffer/tbuffer) name it 'vdata'.
+                        d16_implicit_use_opnd = next(
```
kuhar, line 6821: I think it would help to make this invariant fail loudly. `_d16_load_reads_dst(inst)` can return true while the destination lookup still falls through to `None`; in that case both the declaration and definition branches are skipped, so a future renamed D16 destination would silently lose the preserved-destination read in liveness. Can we assert after the lookup that a matching output operand was found, with the instruction name in the message?

emulation/rocjitsu/tests/analysis/liveness_test.cpp
```diff
@@ -4096,5 +4133,135 @@ TEST(GeneratedInstDefUse, Vop3SdstEncDppPartialRowMaskReadsOnlyVgprResult) {
   EXPECT_FALSE(idu.uses.contains({RegClass::SGPR, 8, 2}));
 }
 
+// --- D16 memory loads: sub-dword loads preserve half of vdst (real decode) ---
+//
+// A D16(_HI) load writes one 16-bit half of the destination VGPR and preserves
+// the other, so it reads the old vdst -- a read-modify-write. The generator
+// emits an implicit_uses override for these (see _d16_load_reads_dst), so the
+// decoded instruction reports vdst in both defs and uses while keeping it out
+// of the printed operand list. A regular (non-D16) load fully overwrites vdst
```
kuhar, line 4836: It would be useful to add two small regression checks around properties the current def/use assertions do not cover. First, this comment says the preserved destination stays out of the printed operand list, but the tests would still pass if a future generator change put it back in `src_operands_` and made `Instruction::disassemble()` print it as an extra source. Second, GLOBAL and SCRATCH D16 loads share `semantic_class == 'flat_load'` with FLAT, but they are emitted into separate generated sources (`vglobal.cpp` / `vscratch.cpp`), so one real decode case for each would cover the remaining `flat_load` overrides.

kuhar (approved):
LGTM with two minor comments

atgutier (commented):
[MAJOR] /tmp/pr8344/emulation/rocjitsu/lib/python/amdisa/semantics.py:2090
  — D16 FORMAT classification activates incorrect load/store execution

  Reasoning: Adding _BUFFER_FORMAT_MAP to MUBUF derivation changes
  previously unimplemented BUFFER FORMAT operations into executable memory
  operations. The execution pipeline does not implement the packed layout—
  two 16-bit components per VGPR—that the new semantics describe.

  Evidence:

  - /tmp/pr8344/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/
    memory_pipeline.cpp:137 uses floor division for the destination VGPR
    count despite its “rounded up” comment.

  - Lines 169–194 copy only two bytes at each four-byte VGPR offset and
    preserve the high half of every visited register.

  - Consequently, D16 XY writes only X and preserves the old high half
    instead of writing Y. D16 XYZ calculates one destination VGPR for six
    bytes and never writes vdata+1, despite its generated two-register
    definition and preserve-read of the last register at /tmp/pr8344/
    emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/rdna4/
    vbuffer.cpp:442.

  - Stores are also wrong: /tmp/pr8344/emulation/rocjitsu/lib/rocjitsu/src/
    rocjitsu/isa/arch/amdgpu/rdna4/vbuffer.cpp:558 reads the low half of the
    same VGPR for both XY components. XYZ and XYZW similarly repeat low
    halves rather than extracting packed components.

  - The added tests validate semantic metadata and def/use sets, but never
    execute multi-component D16 loads or stores.

  Suggested fix: Either keep BUFFER FORMAT execution unclassified while
  deriving partial-definition metadata separately, or implement packed-D16
  execution comprehensively: ceil the VGPR count, pack/unpack two components
  per dword, and preserve only the unused half of the final register for odd
  component counts. Cover normal and OOB completion, plus XY/XYZ/XYZW load
  and store execution tests using distinct halfword values.

  Confidence: High.

[MAJOR] /tmp/pr8344/emulation/rocjitsu/lib/python/amdisa/15:42:07 [43/761]
  — RDNA4 typed D16 loads remain missing from the fix

  Reasoning: RDNA4 combines untyped and typed buffer instructions under
  ENC_VBUFFER, but VBUFFER is routed only to _derive_mubuf, which recognizes
  BUFFER_*, not TBUFFER_*. Therefore typed D16 X and XYZ loads receive no
  preserved-destination use, leaving DBT liveness unsound.

  Evidence:

  - _derive_mubuf accepts only BUFFER prefixes at /tmp/pr8344/emulation/
    rocjitsu/lib/python/amdisa/semantics.py:2087.

  - An exact-head semantic probe classified both TBUFFER_LOAD_D16_FORMAT_X
    and ...XYZ under ENC_VBUFFER as nop.

  - RDNA4 X at /tmp/pr8344/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/
    arch/amdgpu/rdna4/vbuffer.cpp:3394 defines vdata but has no
    implicit_uses() override.

  - XYZ at /tmp/pr8344/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/
    arch/amdgpu/rdna4/vbuffer.cpp:3442 defines vdata:vdata+1 but does not
    preserve-read the unused half of the final register.

  - These are the unified-encoding equivalents of the pre-RDNA4 TBUFFER
    instructions that this PR does handle.

  Suggested fix: Give VBUFFER derivation support for both BUFFER and TBUFFER
  families, or detect partial definitions independently of execution
  classification. Regenerate RDNA4 and add real-decode def/use tests for
  typed D16 X, XY, XYZ, and XYZW.

  Confidence: High.

[MINOR] /tmp/pr8344/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/
  amdgpu/cdna1/mubuf.cpp:1470 — LDS-destination loads falsely keep vdata
  live

  Reasoning: On older MUBUF encodings, inst_.lds redirects the loaded value
  exclusively to LDS. The new hook nevertheless unconditionally reports the
  encoded vdata VGPR as a preserved-destination read.

  Evidence:

  - The implicit use is unconditional.
  - The LDS path at /tmp/pr8344/emulation/rocjitsu/lib/rocjitsu/src/
    rocjitsu/isa/arch/amdgpu/cdna1/mubuf.cpp:1477 sets lds_dst, does not
    establish a VGPR destination, and returns.

  - LDS completion is selected at /tmp/pr8344/emulation/rocjitsu/lib/
    rocjitsu/src/rocjitsu/vm/amdgpu/memory_pipeline.cpp:126.

  - This cannot under-approximate liveness, but it creates a false live
    range and can unnecessarily exclude a DBT scratch register.

  Suggested fix: For encodings with an LDS field, emit the D16 implicit use
  only when !inst_.lds. Add def/use tests for the same opcode with LDS clear
  and set.

  Confidence: High.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -4559,6 +4592,16 @@ def _semantic_reg_class(opnd: Operand) -> RegClass:
         if cls in ('buffer_store', 'tbuffer_store'):
             return self._gen_buffer_store(dst_ops, src_ops, sem, cls)
 
+        if cls in ('buffer_load_format_d16', 'buffer_store_format_d16'):
```
kuhar, line 4685: It would be useful to add one regression test for the non-executable D16 FORMAT contract. The current def/use tests validate the metadata, but they do not fail if `buffer_load_format_d16` or `buffer_store_format_d16` is accidentally routed back to an executable `_gen_buffer_*` body. A small generator-level check that the generated `BufferLoadD16FormatX*::execute_impl` body throws `util::UnimplementedInst`, or a runtime `EXPECT_THROW` on a decoded `buffer_load_d16_format_x`, would keep Tony's execution fix from regressing.

emulation/rocjitsu/lib/python/amdisa/semantics.py
```diff
@@ -2028,18 +2028,80 @@ def _derive_flat(name: str) -> InstructionSemantics | None:
...
+    if not info:
+        return None
+    esz, ne, se = info
+    if fmt_info is not None and esz == 2:  # D16 FORMAT: packed, metadata-only
+        cls = 'buffer_store_format_d16' if is_store else 'buffer_load_format_d16'
+    elif typed and not typed_format_executable:
+        return None  # non-D16 typed FORMAT under VBUFFER: leave as nop
+    else:
+        base = 'tbuffer' if typed else 'buffer'
+        cls = f'{base}_store' if is_store else f'{base}_load'
```
kuhar, line 2095: The issue I am worried about is the extra execution surface for untyped non-D16 FORMAT loads/stores. On the base revision, `BUFFER_LOAD_FORMAT_X*` under MUBUF was `nop` and generated `BufferLoadFormatXVbuffer::execute_impl` threw `UnimplementedInst`; this branch now classifies those ops as `buffer_load`/`buffer_store`, so the generated RDNA4 VBUFFER code executes raw dword transfers and sets `MEMORY_OP`. If that is intentional, please document that non-D16 FORMAT is modeled without DFMT/NFMT conversion; otherwise keeping the untyped FORMAT path unclassified would keep this PR focused on the D16 liveness fix.

kuhar (commented):
The risk here is that D16 FORMAT loads/stores decode with GPUMEM operands but are not marked as memory ops. At `emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py:7116`, `_MEM_CLASSES` includes `buffer_load` and `buffer_store` but not the new `buffer_load_format_d16` or `buffer_store_format_d16` classes; the generated `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/rdna4/vbuffer.cpp:348` constructors set `gpumem` but never `flags_ |= MEMORY_OP`. We should either add both metadata-only classes to the memory flag contract or leave an explicit test/comment if `Instruction::is_memory_op()` is meant to return false for these memory instructions.

It would help to tighten this D16 FORMAT liveness test block in a couple of targeted spots. `emulation/rocjitsu/tests/analysis/liveness_test.cpp:4224` only checks `!uses.contains({VGPR, 5, 2})`; since `RegisterSet::contains` requires every lane in the range, a regression that reads only v5 or only v6 would still pass, so please assert both single registers are absent. Also, the new VBUFFER typed path is covered only by `tbuffer_load_d16_format_x` at `emulation/rocjitsu/tests/analysis/liveness_test.cpp:4280`; adding one `tbuffer_load_d16_format_xyz` decode would lock in the odd-count typed case where only the last VGPR is preserved.

atgutier (approved):
Approving conditioned on fixing this minor issue and @kuhar's recent comments.

emulation/rocjitsu/lib/python/amdisa/tests/test_generator_profile_gates.py
```diff
@@ -3219,3 +3219,44 @@ def test_only_gfx1250_generates_implicit_use_operands_overrides(
...
+    GLOBAL/SCRATCH share semantic_class 'flat_load' with FLAT but emit to
+    separate files, so cover all three. Listing vdst as a source would print it
+    twice in disassembly; pin that a future generator change doing so fails here.
+    """
+    cpp = (gfx1250_generated_root / src_file).read_text()
+    ctor = _generated_constructor_body(cpp, class_name)
+
+    # vdst is a destination, never a source (kept out of the printed operands).
+    assert 'dst_operands_[0] = &vdst;' in ctor
+    assert 'src_operands_[0] = &vdst;' not in ctor
```
kuhar: Could this check reject `&vdst` in any `src_operands_` slot instead of only slots 0 and 1? The test is documenting that the preserved destination stays out of printed operands, but `ScratchLoadD16U8Vscratch` already has three source operands. A regression that assigned `src_operands_[2] = &vdst` would still pass, so matching any `src_operands_` left-hand side would make the guard cover the property it names.

emulation/rocjitsu/tests/instruction_execution_harness_test.cpp
```diff
@@ -5767,4 +5767,49 @@ TEST(HwregTest, SetregImm32PartialBitfield) {
...
+  }};
+
+  // The exec body throws before touching wavefront state, so a single wf can be
+  // reused across cases without any reset.
+  for (const auto &c : cases) {
+    SCOPED_TRACE(c.mnemonic);
+    std::unique_ptr<Instruction> inst(decoder->decode(c.words.data()));
+    ASSERT_NE(inst, nullptr);
+    ASSERT_EQ(std::string_view(inst->mnemonic()), c.mnemonic);
+    EXPECT_THROW(cu->execute_instruction(inst.get(), *wf), util::UnimplementedInst);
```
kuhar, line 6072: I think it would help to assert the memory-op flag in this same loop. Adding `buffer_load_format_d16` and `buffer_store_format_d16` to `_MEM_CLASSES` makes these decoded instructions surface `is_memory_op()` / `RJ_CODE_INST_MEMORY_OP`, but the current test only locks in the `UnimplementedInst` throw. Since each instruction is already decoded here, an `EXPECT_TRUE(inst->is_memory_op()) << c.mnemonic;` before the throw would catch a future regen that drops the metadata. If the comment is meant to exercise the packed two-components-per-VGPR layout specifically, using an `_xy` case would also make that intent clearer than the current single-component `_x` cases.

kuhar (approved):
LGTM % two comments

PR #6332, opened 2026-05-20
[rocprofiler-sdk] add check to keep loads of nameless kernels from producing a warning

samcrater-amd:
## Motivation
This PR addresses AIPROFSDK-829, which is about nightly builds crashing when profiling a particular HIP application.

## Technical Details
A HIP application may load a kernel via hipModuleGetFunction without providing a kernel name. Currently, this causes rocprof to generate a warning when recording the event due to the value field being empty. In CI builds, warning logs are overridden to be fatal, thus causing the nightlies to crash.

This commit extends the check in iterate_args_callback to omit recording the argument if the value string is empty.

## JIRA ID
Addresses AIPROFSDK-829.

## Test Plan
Ran unit test suite and reran target application with CI defines set and unset.

## Test Result
All unit tests run and pass. [testOutput.txt](https://github.com/user-attachments/files/28071933/testOutput.txt)
Target application in ticket now runs without crashing, regardless of CI flag status.

## Submission Checklist
- [X] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #8346, opened 2026-07-09
fix(rocprofiler-sdk): stringify hip const char* args to avoid later segfault

itrowbri:
## Motivation

Fix segfault when tracing hip functions.

## Technical Details

Add `common::get_string_entry(val ? val : "")->c_str();` to store hip string arg instead of raw pointer to avoid later crash.

## JIRA ID

JIRA ID : AIPROFSDK-965

## Submission Checklist


itrowbri:
Added a test due to Copilot and PR bot requesting I do so

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hip/hip.cpp
```diff
@@ -122,6 +123,10 @@ convert_arg_type(Tp&& val)
     {
         return rocprofiler_dim3_t{val.x, val.y, val.z};
     }
+    else if constexpr(std::is_same<data_type, const char*>::value)
+    {
+        return common::get_string_entry(val ? val : "")->c_str();
```
jrmadsen: I am pretty sure it would make more sense to just fix `common::get_string_entry` to handle `nullptr`.

jrmadsen (changes requested):

projects/rocprofiler-sdk/source/lib/common/string_entry.cpp
```diff
@@ -71,6 +71,12 @@ get_string_entry(std::string_view name)
         .first->second.get();
 }
 
+const std::string*
+get_string_entry(const char* name)
+{
+    return get_string_entry(std::string_view{name ? name : ""});
```
MythreyaK: what about 

```suggestion
    return get_string_entry(std::string_view{name ? name : "(null)"});
```

instead?  Using `""` for a null pointer means the trace output can't distinguish "the app passed nullptr" from "the app passed an empty string". 

If the application has a bug where it's passing `nullptr` to an API that doesn't expect it, we'd be silently masking that in the trace rather than making it visible to the user debugging with our tool.

`"(null)"` probably preserves that information? Is returning `nullptr` problematic? That way `null` stays `null` in the buffer record and the serialization layer decides how to present it. Not sure if that causes bad interactions with say #6332.

MythreyaK (approved):

projects/rocprofiler-sdk/source/lib/common/string_entry.cpp
MythreyaK: > Is returning `nullptr` problematic? That way null stays null in the buffer record and the serialization layer decides how to present it.

Could you check with a senior too?

projects/rocprofiler-sdk/source/lib/common/string_entry.hpp
```diff
@@ -33,6 +33,9 @@ namespace common
 const std::string*
 get_string_entry(std::string_view name);
 
+const std::string*
+get_string_entry(const char* name);
```
ApoKalipse-V, line 37: NI: I don't know about the overload here. One could accidentally invoke the string_view version that doesn't have the proper check.

projects/rocprofiler-sdk/tests/CMakeLists.txt
```diff
@@ -90,6 +90,7 @@ add_subdirectory(rocjpeg)
 add_subdirectory(hipfile)
 add_subdirectory(hip-host-tracing)
 add_subdirectory(code-object-multi-threaded)
+add_subdirectory(hip-buffered-dangling-arg)
```
ApoKalipse-V: Suggestion: Do we need a new entry here for this specific test case? Can it not be part of a larger set of hip based tests; or a local unit test?

ApoKalipse-V (changes requested):

projects/rocprofiler-sdk/tests/CMakeLists.txt
itrowbri: I only added a unit test since therock-pr-bot was insisting on it. I'll remove it and add a small unit test instead

projects/rocprofiler-sdk/source/lib/common/string_entry.hpp
itrowbri: I think this is fine? Passing in `std::string_view{nullptr}` would be undefined behavior and be the responsibility of the caller to check make sure they construct the string_view object correctly. Several callers pass `std::string` objects which are converted to string_view objects, and I don't think removing the `string_view` method would be appropriate here. his const char* overload just lets us null-check the raw-pointer case, which is what caused the bug in this PR. If you want me to add a check to the string_view function, I can do that?

ApoKalipse-V (approved):

jrmadsen (approved):

PR #8349, opened 2026-07-09
fix(rccl): [ROCM-26926] Use nChannelsMax for single-node asymmetric P2P patterns

prasanna-amd:
## Summary

Fixes ~20-38% bandwidth regression on MI355X DPX+NPS2 (16 GPUs) for gather and scatter collectives at 1GB message size (ROCM-26926).

JIRA ID : ROCM-26926

## Root Cause

In `addP2pToPlan()`, the per-peer P2P channel count starts from `nChannelsMin`, which is computed as:

```
nChannelsMin = nChannelsMax;
while (nChannelsMin * nRanks > p2pnChannels) nChannelsMin /= 2;
```

On DPX+NPS2 with 16 logical GPUs and `p2pnChannels=64`, `nChannelsMax=8` (gfx950 PATH_NVL), this yields `nChannelsMin=4` (since 8×16=128 > 64). The channel formula then starts from 4 and cannot double further (partSize equals maxPartSize with the default 512KB chunk size), limiting each peer to 4 channels.

This under-utilizes XGMI bandwidth for root-based collectives like gather/scatter where only one rank is the traffic hub — there is no contention problem justifying the throttle.

## Fix

Detect asymmetric vs symmetric traffic patterns using `planTotalTasks[]` (already populated before channel computation):

- **Gather/scatter** (asymmetric): every rank has a zero-task direction — root only receives (`planTotalTasks[1]==0`), leaves only send (`planTotalTasks[0]==0`)
- **AllToAll** (symmetric): `planTotalTasks` is `[16,16]` — both directions have many tasks

The predicate `planTotalTasks[0] == 0 || planTotalTasks[1] == 0` precisely captures gather/scatter patterns. Using `== 0` rather than `<= 1` avoids misclassifying small symmetric exchanges (e.g., 2-rank alltoall with `planTotalTasks=[1,1]`) as asymmetric.

For single-node (`comm->nNodes <= 1`) asymmetric patterns, start from `nChannelsMax` instead of `nChannelsMin`. Symmetric patterns retain `nChannelsMin` to avoid XGMI link contention when all ranks send simultaneously.

The change is applied at both channel computation sites in `addP2pToPlan()` (pre-registration and post-registration paths).

## Test Results

Tested on MI355X DPX+NPS2, 16 GPUs, 1GB message size, 5 iterations (`-N 5`):

| Collective | System 7.14 (GB/s) | With Fix (GB/s) | Delta |
|---|---|---|---|
| gather | 305.0 | 422.3 | **+38.5%** |
| scatter | 222.0 | 225.5 | +1.6% |
| alltoall | 112.3 | 111.3 | -0.9% (noise) |
| alltoallv | 73.3 | 73.0 | -0.5% (noise) |

AllToAll and AllToAllV are unaffected (within noise) because they correctly detect as symmetric and use the unchanged `nChannelsMin` path.

## Unit Tests

Added `P2pChannelScalingTests.cpp` with large-message correctness tests for gather, scatter, alltoall, and alltoallv that exercise the channel selection path in `addP2pToPlan()`.

## Env Var Workaround (partial, for reference)

`NCCL_P2P_NVL_CHUNKSIZE=262144` recovers gather (305 → 358 GB/s) by changing the part size arithmetic to trigger the doubling loop, but does not help alltoall/scatter/alltoallv. The code fix is more targeted and effective.

## Test Plan

- [x] Tested gather_perf, scatter_perf, alltoall_perf, alltoallv_perf at 1GB on MI355X DPX+NPS2 (16 GPUs)
- [x] Verified alltoall/alltoallv are unaffected (symmetric detection works correctly)
- [x] Verified with `-N 5` iterations to rule out outlier noise
- [x] Added P2pChannelScaling unit tests (gather, scatter, alltoall, alltoallv large message)
- [ ] Needs QA validation against 7.13-RC2 baseline (364 GB/s gather) on DPX+NPS2
- [ ] Needs verification that SPX mode (8 GPUs) is unaffected
- [ ] Needs multi-node regression check

### DIFF
--- a/projects/rccl/src/enqueue.cc
+++ b/projects/rccl/src/enqueue.cc
@@ -1171,7 +1171,7 @@ static ncclResult_t addP2pToPlan(
       // Single-node asymmetric patterns (gather/scatter): use nChannelsMax to
       // fully utilize XGMI bandwidth when only one rank is the traffic hub.
       // Symmetric patterns (alltoall): stay at nChannelsMin to avoid contention.
-      bool asymmetric = planTotalTasks[0] <= 1 || planTotalTasks[1] <= 1;
+      bool asymmetric = planTotalTasks[0] == 0 || planTotalTasks[1] == 0;
       int nChStart = (comm->nNodes <= 1 && asymmetric) ? nChannelsMax : nChannelsMin;
       nChannels[dir] = std::min<int>(nChStart, divUp(bytes[dir], minPartSize));
       size_t partSize = std::max(minPartSize, divUp(bytes[dir], nChannels[dir]));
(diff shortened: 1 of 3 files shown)


wenkaidu (approved):

prasanna-amd:
therock-pr-bot doesn't recognize Unit Test file name pattern in rccl

PR #7419, opened 2026-06-17
feat(hipfile): Gate fastpath-only tests only to systems that support the AIS fastpath

john00003:
## Motivation

We don't want test failures on systems that don't support the fastpath if the user is aware of this limitation.

## Technical Details

Some of our tests previously forced the use of the fastpath, and manifested as test failures. These tests can instead be skipped when we detect the system does not support the fastpath.

We duplicate the logic in the `ais-check` tool to provide a C++ helper that can be used from the tests. The helper also does 0-sized I/O to gate fastpath availability, and attempts to report a reason why it may not fail. 0-sized I/O require ROCm >= 7.14, so we use preprocessor macros to gate these tests to systems that have the latest version of ROCm available.

Note, all of the output that indicates to the user that they can use a specific flag to skip the tests instead of fail, or reports why the tests got skipped, are only output when -V is used with `ctest`.

## JIRA ID

AIHIPFILE-202

## Test Plan

Test on a system that does not support the fastpath. Verify that the following tests are reported as skipped when the CMake option `HIPFILE_ALLOW_SKIP_FASTPATH_TESTS` is enabled when building:
```
77 - HipFileIoHipInit.spawnedThreadReadRunsWithoutSegfault (Skipped)
78 - HipFileIoHipInit.spawnedThreadWriteRunsWithoutSegfault (Skipped)
216 - HipFileIo.ReadToUnregisteredBufferAtOffset/Fastpath  # GetParam() = 40-byte object <00-00 00-00 00-00 00-00 E8-3B 44-2D 00-00 00-00 08-00 00-00 00-00 00-00 46-61 73-74 70-61 74-68 00-00 00-00 00-00 00-00> (Skipped)
218 - HipFileIo.ReadToUnregisteredBufferAtOffsetReturnsErrorIfOverflow/Fastpath  # GetParam() = 40-byte object <00-00 00-00 00-00 00-00 E8-3B 44-2D 00-00 00-00 08-00 00-00 00-00 00-00 46-61 73-74 70-61 74-68 00-00 00-00 00-00 00-00> (Skipped)

```

Verify that `ctest . -V` shows the skip message `SKIP: fastpath not available in this environment` for the above tests.

Verify that when not built with the CMake option `HIPFILE_ALLOW_SKIP_FASTPATH_TESTS`, these tests output a message to indicate the option is available for users that are aware the fastpath is not available.

## Test Result

[...]

PR #8354, opened 2026-07-09
feat(hipfile): ais-check, add kfd alternative check in p2pdma check on ais-check

akolliasAMD:
## Technical Details

added the method of checking the p2pdma by checking the `capability` property in /sys/class/kfd/kfd/topology/nodes/X/properties and checking the correct 

## JIRA ID
Jira ID: AIHIPFILE-210
## Test Plan
N/A
## Test Result
N/A
## Submission Checklist

### DIFF
--- a/projects/hipfile/tools/ais-check/ais-check
+++ b/projects/hipfile/tools/ais-check/ais-check
@@ -138,8 +138,9 @@ def amdgpu_dkms_version():
 def topology_reports_ais():
     """
     True if any KFD topology GPU node reports AIS P2P-DMA initialized via its
-    "capability" bit (capability & KFD_CAPABILITY_AIS). CPU nodes omit the field
-    and are simply skipped.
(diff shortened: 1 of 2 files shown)


projects/hipfile/tools/ais-check/ais-check
```diff
@@ -66,9 +67,109 @@ PSEUDO_FSTYPES = {
...
+                        capability = int(value.strip())
+                    except ValueError:
+                        break
+                    if capability & KFD_CAPABILITY_AIS:
+                        return True
+                    break
+        except OSError:
+            continue
+
+    return False
```
john00003, line 173: Should we gate the capability bit check to only be for GPUs, and should we only return True if all the GPUs have the capability bit? 

https://github.com/ROCm/rocm-systems/pull/7419/changes#diff-6a207e14d94343a911bd6e36287beceb68b6085270245ab05785031b643ae5eaR46-R65

john00003 (approved):

Issue #8353, opened 2026-07-09
fix(rocjitsu): return the libc ioctl(2) errno contract from interposed ioctls


amd-arosa:
rocm-dbgapi drives the emulated KFD entirely
through `ioctl(2)`, and it decodes failures as `ret < 0 ? -errno : ret`. The
interposer returned the driver's raw kernel-style `-errno`, so dbgapi read a
stale `errno` and intermittently misread `-EAGAIN` from `QUERY_DEBUG_EVENT` as
success - then acted on uninitialized event fields and aborted the debug
session with "could not find os_agent_id 0".

PR #8357, opened 2026-07-09
fix(rocjitsu): return the libc ioctl(2) errno contract from interposed ioctls

amd-arosa:
## Motivation

Convert a kernel-style driver ioctl result into the libc ioctl(2) return/`errno` contract.

this fixes dbgapi read a
stale errno and intermittently misread -EAGAIN from QUERY_DEBUG_EVENT as
success - then acted on uninitialized event fields and aborted the debug
session with "could not find os_agent_id 0".

## Technical Details

add
```
/// @brief Convert a kernel-style driver ioctl result into the libc ioctl(2)
/// return/`errno` contract.
///
/// @param r Driver result: `>= 0` on success, `-errno` on failure.
/// @returns @p r unchanged when non-negative; otherwise `-1` with `errno` set to
///          `-r`.
inline int kfd_ioctl_ret(int r) {
  if (r < 0) {
    errno = -r;
    return -1;
  }
  return r;
}
```

and use where relevent.

## JIRA ID

#8353

## Test Plan

tests pass

## Test Result

tests pass locally

## Submission Checklist


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
```diff
@@ -88,6 +88,20 @@ static int connect_to_daemon() {
 
 namespace {
 
+/// @brief Convert a kernel-style driver ioctl result into the libc ioctl(2)
+/// return/`errno` contract.
+///
+/// @param r Driver result: `>= 0` on success, `-errno` on failure.
+/// @returns @p r unchanged when non-negative; otherwise `-1` with `errno` set to
+///          `-r`.
+inline int kfd_ioctl_ret(int r) {
```
atgutier: inline is unnecessary here.

atgutier (approved):
You can remove inline. Otherwise LGTM.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
amd-arosa: un-inlined

PR #8359, opened 2026-07-09
fix(cuid): corrected CPU fingerprint formulation

gabrpham:
## Motivation
CPU fingerprints were being generated incorrectly at the core level rather than at the package level which created fingerprints that were highly simplistic.

## Technical Details
Redesigned finger print formulation to correctly represent CPUs at the package level rather than at the logical core level.

## JIRA ID
JIRA ID : AILITOOLS-230

## Test Plan
Ran unit tests to ensure no regressions and confirmed that fingerprints were being created at the package level rather than logical core level.

## Test Result
Unit tests:
image

### DIFF
--- a/projects/cuid/docs/conceptual/cuid_file.rst
+++ b/projects/cuid/docs/conceptual/cuid_file.rst
@@ -36,7 +36,7 @@ Here's a sample CUID file for your reference:
   family=0019
   model=0074
   unit_id=0004
-  package_core_id=0:0
+  package_id=0
   last_update=6980d761
 
   [NIC:0]
@@ -59,7 +59,7 @@ Depending on the device type, different device attributes (present in the CUID f
 
 - **GPU:** The ``device_node`` links the derived CUID to the hardware.
 
-- **CPU:** The ``package_core_id`` consists of two components: physical ID and core ID, as ``physical_id:core_id``. This helps to identify CPU cores in multi-CPU systems.
+- **CPU:** The ``package_id`` identifies the physical CPU package, which is akin to a socket ID. CPUs will also display a ``device_node`` which represents the first logical CPU core of the physical CPU package. The derived CUID is associated with the physical CPU package, not the logical CPU core.
 
 - **NIC:** The ``device_node`` links the derived CUID to the network device.
 
--- a/projects/cuid/docs/conceptual/cuid_file.rst
+++ b/projects/cuid/docs/conceptual/cuid_file.rst
@@ -59,7 +59,7 @@ Depending on the device type, different device attributes (present in the CUID f
 
 - **GPU:** The ``device_node`` links the derived CUID to the hardware.
 
- - **CPU:** The ``package_id`` identifies the physical CPU package, which is akin to a socket ID. CPUs will also display a ``device_node`` which represents the representative logical CPU for that package (currently selected as the lowest APIC ID seen for the socket). The derived CUID is associated with the physical CPU package, not the logical CPU core.
+- **CPU:** The ``package_id`` identifies the physical CPU package, which is akin to a socket ID. CPUs will also display a ``device_node`` which represents the representative logical CPU for that package (currently selected as the lowest APIC ID seen for the socket). The derived CUID is associated with the physical CPU package, not the logical CPU core.
 
 - **NIC:** The ``device_node`` links the derived CUID to the network device.
 
--- a/projects/cuid/docs/conceptual/cuid_file.rst
+++ b/projects/cuid/docs/conceptual/cuid_file.rst
@@ -59,7 +59,7 @@ Depending on the device type, different device attributes (present in the CUID f
 
 - **GPU:** The ``device_node`` links the derived CUID to the hardware.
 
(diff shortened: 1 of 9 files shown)


projects/cuid/lib/src/cuid_file.cc
```diff
@@ -456,8 +457,12 @@ amdcuid_status_t CuidFile::load() {
           current_entry.derived_cuid = string_to_cuid(value);
         } else if (key == "device_node") {
           current_entry.device_node = value;
-        } else if (key == "package_core_id") {
-          current_entry.package_core_id = value;
+        } else if (key == "package_id") {
+          current_entry.package_id =
+              static_cast<uint16_t>(std::stoul(value, nullptr, 16));
+        } else if (key == "core_id") {
```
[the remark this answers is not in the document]
gabrpham: We're not yet concerned about backwards compatibility since we haven't officially moved to 1.0 version.

bill-shuzhou-liu, line 463: The save path now writes `package_id=0x0001` (with `"0x"` prefix), but `stoul(..., nullptr, 16)` with explicit base 16 does **not** accept a `0x` prefix — it will throw `std::invalid_argument`. Any CUID file written by this version will fail to reload?

Same issue applies to every other numeric field now saved with `0x` (vendor_id, device_id, revision_id, family, model, pci_class, unit_id, hardware_fingerprint, core_id).

projects/cuid/lib/src/cuid_device.cc
```diff
@@ -174,7 +166,7 @@ amdcuid_status_t CuidDevice::get_derived_cuid(amdcuid_derived_id &id,
   // check the temporary bit in the primary CUID to determine whether to use the
   // real HMAC key or the temp key for derived CUID generation
   bool temp = primary.raw_bits[14] &
-              0x04; // check the temp indicator bit in the reserved bits
+              0x20; // check the temp indicator bit in the reserved bits
   if (temp) {
```
bill-shuzhou-liu, line 170: temp indicator bit changed from `0x04` → `0x20`. We may need a CUID file format version bump if this is already released to production.

projects/cuid/lib/src/cuid_util.cc
```diff
@@ -388,20 +388,14 @@ CuidUtilities::make_fallback_fingerprint(const std::string &id,
...
-      if (hostname_file.is_open())
-        std::getline(hostname_file, system_id);
-    }
+  std::ifstream machine_id_file("/etc/machine-id");
+  if (machine_id_file.is_open())
+    std::getline(machine_id_file, system_id);
+
+  if (system_id.empty()) {
+    std::ifstream hostname_file("/etc/hostname");
+    if (hostname_file.is_open())
```
bill-shuzhou-liu, line 397: If both `/etc/machine-id` and `/etc/hostname` are absent (minimal containers, initramfs), `system_id` stays empty and `combined` becomes just the hex-filtered digits of the socket `id` argument. The function returns `AMDCUID_STATUS_SUCCESS` with a fingerprint that isn't stable across machines or reboots.

maybe log a warning?

projects/cuid/docs/conceptual/cuid_file.rst
```diff
@@ -59,7 +59,7 @@ Depending on the device type, different device attributes (present in the CUID f
 
 - **GPU:** The ``device_node`` links the derived CUID to the hardware.
 
-- **CPU:** The ``package_core_id`` consists of two components: physical ID and core ID, as ``physical_id:core_id``. This helps to identify CPU cores in multi-CPU systems.
+ - **CPU:** The ``package_id`` identifies the physical CPU package, which is akin to a socket ID. CPUs will also display a ``device_node`` which represents the representative logical CPU for that package (currently selected as the lowest APIC ID seen for the socket). The derived CUID is associated with the physical CPU package, not the logical CPU core.
```
bill-shuzhou-liu, line 63: extra leading space may break RST list alignment:
This line starts with ` - ` (space-dash-space) while the surrounding `- **GPU:**` and `- **NIC:**` entries start at column 0.

gabrpham: corrected

projects/cuid/lib/src/cuid_file.cc
gabrpham: I'm seeing in public docs that stoul and its variants are able to convert the prefix since c++11, so this shouldn't be an issue: https://en.cppreference.com/cpp/string/basic_string/stoul

projects/cuid/lib/src/cuid_device.cc
gabrpham: we're currently tracking CUID library version but not CUID file format version. I'm considering our integration into theRock as our official 1.0 release, so I don't think it warrants any major notifications, but I will grant that a patch version bump for the library would be warranted with this.

projects/cuid/lib/src/cuid_util.cc
gabrpham: If we've reached this point in the code, then we've reached the temp CUID creation stage and we no longer guarantee stability across machines or reboots and we're simply seeking to output a CUID of some sort without crashing downstream applications.

bill-shuzhou-liu (approved):

PR #8361, opened 2026-07-09
feat(rocprofiler-sdk): resolve rocattach register symbols from target ELF

mcao59:
## Motivation

This PR improves rocprofv3 attach support for containerized target processes. 

The current rocattach symbol resolution assumes the host and target `librocprofiler-register.so` have identical ELF layouts by calculating a symbol offset from the host library and applying it to the target mapping. This is fragile when attaching from the host to a process inside a container, where the target may use a different library path or binary layout.

The goal is to resolve attach entry points from the target process’s own mapped ELF and validate target-visible tool paths before injection.

### `symbol_lookup.cpp` (Main changes)
The current symbol_lookup.cpp flow is:
```
find_symbol(pid, addr, library, symbol)
  -> select_unique_mapped_object(pid, library)
  -> resolve_target_symbol(pid, mapped_object, symbol)
       -> open_target_elf(pid, mapped_object)
       -> parse ELF headers and load segments with ELFIO
       -> calculate load bias from the mapped object
       -> look up the symbol through SHT_DYNSYM
          or fall back to PT_DYNAMIC for section-headerless ELF files
       -> address = load_bias + sym.st_value
       -> validate address is inside an executable mapping
```

- Resolve `rocprofiler_register_attach` / `rocprofiler_register_detach` from the target process’s mapped `librocprofiler-register.so` instead of using host-computed symbol offsets. 
- Parse `/proc/<pid>/maps` to identify mapped register libraries by device/inode. 
- Open the target mapped ELF through `/proc/<pid>/map_files`, with `/proc/<pid>/root` fallback. 
- Parse ELF dynamic symbol information w/ ELFIO, retain a bounded PT_DYNAMIC fallback for section-headerless ELF files, supporting both GNU and SysV hash metadata.
- Calculate target runtime addresses using target ELF load bias and symbol values. 
- Reject ambiguous matching libraries instead of selecting the first match.
- Validate resolved symbols against executable target mappings. 
- Add bounds checks and limits for ELF parsing to handle malformed inputs more safely.

### `rocattach.cpp`
- Validate absolute `ROCPROF_ATTACH_TOOL_LIBRARY` paths from the target process’s mount namespace using `/proc/<pid>/root`. 
- Fail early when the target-visible tool library path is missing or is not a regular file. 

### `tests/rocattach` 
- Add fixture shared libraries for normal, GNU-hash, SysV-hash, shifted-layout, and ambiguous-library test cases. 
- Add symbol lookup tests for target ELF resolution. 
- Add coverage for shifted symbol layout where the target attach symbol offset differs. 
- Add tests for ambiguous mapped library rejection. 
- Add malformed ELF failure handling tests. 
- Add target tool path validation tests for missing and non-regular paths.

This change removes the identical ELF-layout requirement between host and target `librocprofiler-register.so`, **while still assuming the same rocprofiler-register attach ABI.**

Requirements and Caveat:
- [ ] Document host-to-container attach requirements and ABI compatibility caveat.

## JIRA ID

JIRA ID : AIPROFSDK-907

## Test Plan

- Added rocattach symbol lookup tests with fixture shared libraries.
- Tested normal dynamic symbol lookup.
- Tested GNU-hash and SysV-hash symbol resolution.
- Tested shifted-layout fixture where the target attach symbol offset differs.
- Tested ambiguous mapped library rejection.
- Tested malformed mapped ELF failure handling.
- Tested absolute target tool path validation for missing and non-regular paths.

## Test Result

Existing and new tests pass

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/rocattach.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/rocattach.cpp
@@ -238,13 +238,8 @@ resolve_attach_tid(pid_t pid)
 }
 
 rocattach_status_t
-validate_target_tool_path(pid_t pid, std::string_view tool_lib_path)
+validate_target_absolute_tool_path(pid_t pid, const std::filesystem::path& tool_path)
 {
-    auto tool_path = std::filesystem::path{tool_lib_path};
-    if(tool_path.empty()) return ROCATTACH_STATUS_SUCCESS;
-
-    if(!tool_path.is_absolute()) return ROCATTACH_STATUS_SUCCESS;
-
     auto target_path =
         std::filesystem::path{fmt::format("/proc/{}/root", pid)} / tool_path.relative_path();
     std::error_code ec;
@@ -256,12 +251,12 @@ validate_target_tool_path(pid_t pid, std::string_view tool_lib_path)
             // restrictions, keep attachment best-effort and let target-side dlopen
             // report the final load failure.
             ROCP_WARNING << "[rocprofiler-sdk-rocattach] Could not validate tool library path '"
-                         << tool_lib_path << "' at " << target_path.string()
+                         << tool_path.string() << "' at " << target_path.string()
                          << " from the target process mount namespace: " << ec.message();
             return ROCATTACH_STATUS_SUCCESS;
         }
 
-        ROCP_ERROR << "[rocprofiler-sdk-rocattach] Tool library path '" << tool_lib_path
+        ROCP_ERROR << "[rocprofiler-sdk-rocattach] Tool library path '" << tool_path.string()
                    << "' is absolute but is not visible at " << target_path.string()
                    << " from the target process mount namespace. Attachment requires "
                       "ROCPROF_ATTACH_TOOL_LIBRARY to name a library path that target-side "
@@ -277,12 +272,12 @@ validate_target_tool_path(pid_t pid, std::string_view tool_lib_path)
             // restrictions, keep attachment best-effort and let target-side dlopen
             // report the final load failure.
             ROCP_WARNING << "[rocprofiler-sdk-rocattach] Could not validate tool library path '"
-                         << tool_lib_path << "' at " << target_path.string()
+                         << tool_path.string() << "' at " << target_path.string()
                          << " from the target process mount namespace: " << ec.message();
             return ROCATTACH_STATUS_SUCCESS;
         }
 
-        ROCP_ERROR << "[rocprofiler-sdk-rocattach] Tool library path '" << tool_lib_path
+        ROCP_ERROR << "[rocprofiler-sdk-rocattach] Tool library path '" << tool_path.string()
                    << "' is absolute but does not refer to a regular file at "
                    << target_path.string() << " from the target process mount namespace.";
         return ROCATTACH_STATUS_ERROR_INVALID_ARGUMENT;
@@ -325,10 +320,23 @@ setup(int pid)
     }
     const char* tool_lib_path = tool_lib_path_env.c_str();
     ROCP_TRACE << "[rocprofiler-sdk-rocattach] Tool library path: " << tool_lib_path;
-    status = validate_target_tool_path(pid, tool_lib_path_env);
-    if(status != ROCATTACH_STATUS_SUCCESS)
+
+    auto tool_path = std::filesystem::path{tool_lib_path_env};
+    if(tool_path.empty())
     {
-        return status;
+        ROCP_ERROR << "[rocprofiler-sdk-rocattach] Tool library path must not be empty.";
+        return ROCATTACH_STATUS_ERROR_INVALID_ARGUMENT;
+    }
+
+    // Bare or relative library names must be resolved by dlopen in the target
+    // process using the target's loader search path and working directory.
+    if(tool_path.is_absolute())
+    {
+        status = validate_target_absolute_tool_path(pid, tool_path);
+        if(status != ROCATTACH_STATUS_SUCCESS)
+        {
+            return status;
+        }
     }
 
     auto*      sessions = CHECK_NOTNULL(get_sessions());
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/rocattach.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/rocattach.cpp
@@ -238,7 +238,7 @@ resolve_attach_tid(pid_t pid)
 }
 
 rocattach_status_t
-validate_target_tool_path(pid_t pid, const std::string& tool_lib_path)
+validate_target_tool_path(pid_t pid, std::string_view tool_lib_path)
 {
     auto tool_path = std::filesystem::path{tool_lib_path};
     if(tool_path.empty()) return ROCATTACH_STATUS_SUCCESS;
@@ -261,11 +261,11 @@ validate_target_tool_path(pid_t pid, const std::string& tool_lib_path)
             return ROCATTACH_STATUS_SUCCESS;
         }
 
-        ROCP_WARNING << "[rocprofiler-sdk-rocattach] Tool library path '" << tool_lib_path
-                     << "' is absolute but is not visible at " << target_path.string()
-                     << " from the target process mount namespace. Attachment requires "
-                        "ROCPROF_ATTACH_TOOL_LIBRARY to name a library path that target-side "
-                        "dlopen can resolve.";
+        ROCP_ERROR << "[rocprofiler-sdk-rocattach] Tool library path '" << tool_lib_path
+                   << "' is absolute but is not visible at " << target_path.string()
+                   << " from the target process mount namespace. Attachment requires "
+                      "ROCPROF_ATTACH_TOOL_LIBRARY to name a library path that target-side "
+                      "dlopen can resolve.";
         return ROCATTACH_STATUS_ERROR_INVALID_ARGUMENT;
     }
 
@@ -282,9 +282,9 @@ validate_target_tool_path(pid_t pid, const std::string& tool_lib_path)
             return ROCATTACH_STATUS_SUCCESS;
         }
 
-        ROCP_WARNING << "[rocprofiler-sdk-rocattach] Tool library path '" << tool_lib_path
-                     << "' is absolute but does not refer to a regular file at "
-                     << target_path.string() << " from the target process mount namespace.";
+        ROCP_ERROR << "[rocprofiler-sdk-rocattach] Tool library path '" << tool_lib_path
+                   << "' is absolute but does not refer to a regular file at "
+                   << target_path.string() << " from the target process mount namespace.";
         return ROCATTACH_STATUS_ERROR_INVALID_ARGUMENT;
     }
 
@@ -306,10 +306,23 @@ setup(int pid)
     // ROCPROF_ATTACH_TOOL_LIBRARY override in a secure-execution context (setuid/setgid,
     // file capabilities, etc.), so an unprivileged user cannot cause a privileged attach
     // helper to inject an arbitrary library.
-    auto        tool_lib_path_env = rocprofiler::common::is_at_secure()
-                                        ? std::string{"librocprofiler-sdk-tool.so"}
-                                        : rocprofiler::common::get_env("ROCPROF_ATTACH_TOOL_LIBRARY",
-                                                                "librocprofiler-sdk-tool.so");
+    constexpr auto default_tool_lib_path = std::string_view{"librocprofiler-sdk-tool.so"};
+    auto           tool_lib_path_env     = std::string{default_tool_lib_path};
+    auto           tool_lib_path_override =
+        rocprofiler::common::get_env_optional("ROCPROF_ATTACH_TOOL_LIBRARY");
+    if(rocprofiler::common::is_at_secure())
+    {
+        if(tool_lib_path_override)
+        {
+            ROCP_WARNING << "[rocprofiler-sdk-rocattach] Ignoring ROCPROF_ATTACH_TOOL_LIBRARY "
+                            "override in secure-execution context; using generic tool library "
+                         << default_tool_lib_path;
+        }
+    }
+    else if(tool_lib_path_override)
+    {
+        tool_lib_path_env = *tool_lib_path_override;
+    }
     const char* tool_lib_path     = tool_lib_path_env.c_str();
     ROCP_TRACE << "[rocprofiler-sdk-rocattach] Tool library path: " << tool_lib_path;
     status = validate_target_tool_path(pid, tool_lib_path_env);
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/rocattach.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/rocattach.cpp
@@ -299,6 +299,25 @@ setup(int pid)
                   "for pid "
                << pid;
 
+    auto status = ROCATTACH_STATUS_SUCCESS;
+
+    // Build the tool library path before ptrace setup so invalid absolute paths
+    // fail before modifying target process state. Do not honor the user-controllable
+    // ROCPROF_ATTACH_TOOL_LIBRARY override in a secure-execution context (setuid/setgid,
+    // file capabilities, etc.), so an unprivileged user cannot cause a privileged attach
+    // helper to inject an arbitrary library.
+    auto        tool_lib_path_env = rocprofiler::common::is_at_secure()
+                                        ? std::string{"librocprofiler-sdk-tool.so"}
+                                        : rocprofiler::common::get_env("ROCPROF_ATTACH_TOOL_LIBRARY",
+                                                                "librocprofiler-sdk-tool.so");
+    const char* tool_lib_path     = tool_lib_path_env.c_str();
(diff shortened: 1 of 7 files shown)


projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/rocattach.cpp
```diff
@@ -237,6 +237,60 @@ resolve_attach_tid(pid_t pid)
...
+            // If host-side validation is blocked by procfs permissions or namespace
+            // restrictions, keep attachment best-effort and let target-side dlopen
+            // report the final load failure.
+            ROCP_WARNING << "[rocprofiler-sdk-rocattach] Could not validate tool library path '"
+                         << tool_lib_path << "' at " << target_path.string()
+                         << " from the target process mount namespace: " << ec.message();
+            return ROCATTACH_STATUS_SUCCESS;
+        }
+
+        ROCP_WARNING << "[rocprofiler-sdk-rocattach] Tool library path '" << tool_lib_path
```
meserve-amd: Nit: Promote to error log

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/rocattach.cpp
```diff
@@ -237,6 +237,60 @@ resolve_attach_tid(pid_t pid)
...
```
meserve-amd: Nit: Promote to error log

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/rocattach.cpp
```diff
@@ -245,6 +299,25 @@ setup(int pid)
...
                << pid;
 
+    auto status = ROCATTACH_STATUS_SUCCESS;
+
+    // Build the tool library path before ptrace setup so invalid absolute paths
+    // fail before modifying target process state. Do not honor the user-controllable
+    // ROCPROF_ATTACH_TOOL_LIBRARY override in a secure-execution context (setuid/setgid,
+    // file capabilities, etc.), so an unprivileged user cannot cause a privileged attach
+    // helper to inject an arbitrary library.
+    auto        tool_lib_path_env = rocprofiler::common::is_at_secure()
```
meserve-amd: Minor: Add a warning log indicating the path has been overridden, IF one has been specified AND we select the generic tool instead. (I realize this is copy/pasted from below, but we should add it anyways)

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/rocattach.cpp
```diff
@@ -237,6 +237,60 @@ resolve_attach_tid(pid_t pid)
     return -1;
 }
 
+rocattach_status_t
+validate_target_tool_path(pid_t pid, const std::string& tool_lib_path)
```
meserve-amd: Nit: tool_lib_path could be a string_view

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/symbol_lookup.cpp
meserve-amd, line 1: Major: This is a rather large block of code with no comments. Please add comments explaining code flow and what functions are trying to do. Elaborate with what the ELF or maps file should contain when we are parsing.

meserve-amd, line 1: Minor: Design consideration, this file seems to have two separate sets of functions that could be broken into different files. It's probably fine as-is, but could be split for clarity.
select_unique_mapped_object is the root of one tree of functions and it could be made into its own file, resolve_target_symbol is the root of the other

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/symbol_lookup.cpp
```diff
@@ -22,245 +22,900 @@
...
+find_mapped_objects(pid_t pid, const std::string& library)
+{
+    auto objects = std::unordered_map<std::string, mapped_object>{};
+
+    for(const auto& mapping : parse_maps(pid))
+    {
+        if(mapping.inode == 0 || !library_name_matches(mapping, library)) continue;
+
+        auto key =
+            fmt::format("{}:{}:{}", mapping.device_major, mapping.device_minor, mapping.inode);
```
meserve-amd: Minor: Using a string as a key is slower than just using the values themselves. We have hashers available in common to do it quickly

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/symbol_lookup.cpp
```diff
@@ -22,245 +22,900 @@
...
+{
+    auto        iss = std::istringstream{line};
+    std::string range;
+    std::string offset;
+    std::string device;
+    auto        mapping = memory_mapping{};
 
-    // try to load with the given path
-    if(!_lib_handle)
+    if(!(iss >> range >> mapping.permissions >> offset >> device >> mapping.inode))
```
meserve-amd, line 243: Nit: the maps file is practically designed for use with sscanf, and some of the rigmarole can be cut out from here using it. That said, I can see it either way and it comes down to personal taste.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/symbol_lookup.cpp
```diff
@@ -22,245 +22,900 @@
...
-
-    // uses "maps" file to find where library has been loaded in target process
-    // does not require this process to be attached
-    std::stringstream filename;
-    filename << "/proc/" << inpid << "/maps";
-    std::ifstream maps(filename.str().c_str());
+std::optional<target_elf>
+open_target_elf(pid_t pid, const mapped_object& object)
+{
+    for(const auto& mapping : object.mappings)
```
meserve-amd: Minor: Is it necessary to try every mapping? I think they will either all succeed or all fail

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/symbol_lookup.cpp
```diff
@@ -22,245 +22,900 @@
...
     }
 
-    // NOLINTNEXTLINE(performance-no-int-to-ptr)
-    addr = reinterpret_cast<void*>(std::stoull(line, nullptr, 16));
-    //  target_library_addrs[searchname.str()] = addr;
     return true;
 }
 
+std::optional<uint64_t>
+calculate_load_bias(const target_elf& elf, const mapped_object& object)
```
meserve-amd, line 538: Major: if I remember how this works, I believe you only need one offset calculation here.

I asked Claude and it remarks that the bias will always be `first_mapping_start − align_down(first_PT_LOAD.p_vaddr, page_size)`. You'd need to sort the PT_LOAD to determine the first vaddr, but it's more straightforward.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/symbol_lookup.cpp
```diff
@@ -22,245 +22,900 @@
...
+
+struct target_elf
+{
+    std::string             path          = {};
+    std::vector<uint8_t>    data          = {};
+    ELFIO::elfio            reader        = {};
+    std::vector<Elf64_Phdr> load_segments = {};
+};
+
+class unique_fd
```
meserve-amd: Minor: Because this is used inside one function, it could be replaced by a scope_destructor (rocprof common utility)

projects/rocprofiler-sdk/tests/rocattach/symbol_lookup_test.cpp
```diff
@@ -0,0 +1,267 @@
...
+}
+
+loaded_library
+create_and_load_sectionless_copy(const loaded_library& source, std::string_view label)
+{
+    auto path =
+        std::filesystem::temp_directory_path() /
+        ("librocprofiler-register.so.rocattach-sectionless-" + std::string{label} + "-XXXXXX");
+    auto path_buffer = path.string();
+    auto fd          = mkstemp(path_buffer.data());
```
meserve-amd, line 165: Minor: cleanup this file, as in expect_malformed_mapped_elf_fails

projects/rocprofiler-sdk/tests/rocattach/CMakeLists.txt
```diff
@@ -47,6 +47,85 @@ target_link_libraries(
...
+add_dependencies(
+    rocattach-symbol-lookup-test
+    rocattach-symbol-lookup-fixture-normal
+    rocattach-symbol-lookup-fixture-gnu
+    rocattach-symbol-lookup-fixture-sysv
+    rocattach-symbol-lookup-fixture-shifted
+    rocattach-symbol-lookup-fixture-ambiguous-a
+    rocattach-symbol-lookup-fixture-ambiguous-b)
+
+rocprofiler_add_integration_execute_test(
```
meserve-amd: Minor: missing ENVIRONMENT

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/symbol_lookup.cpp
```diff
@@ -22,245 +22,900 @@
...
+std::vector<memory_mapping>
+parse_maps(pid_t pid)
+{
+    auto filename = fmt::format("/proc/{}/maps", pid);
+    auto maps     = std::ifstream{filename};
+    auto result   = std::vector<memory_mapping>{};
+
+    if(!maps)
+    {
+        ROCP_ERROR << "[rocprofiler-sdk-rocattach] Couldn't open " << filename;
```
meserve-amd: Minor: strerror(errno) could be useful here

meserve-amd (changes requested):

projects/rocprofiler-sdk/tests/rocattach/CMakeLists.txt
mcao59: Good catch, added in [fda5daf](https://github.com/ROCm/rocm-systems/pull/8361/commits/fda5dafb219146a4b54829b85f331f20dc274ea7)

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/rocattach.cpp
mcao59: Added in [adabe4a](https://github.com/ROCm/rocm-systems/pull/8361/commits/adabe4a66d905fcf06c101e2a86152e49ef6de2f)

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/rocattach.cpp
mcao59: addressed in [adabe4a](https://github.com/ROCm/rocm-systems/pull/8361/commits/adabe4a66d905fcf06c101e2a86152e49ef6de2f)

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/rocattach.cpp
mcao59: addressed in [adabe4a](https://github.com/ROCm/rocm-systems/pull/8361/commits/adabe4a66d905fcf06c101e2a86152e49ef6de2f)

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/rocattach.cpp
mcao59: addressed in [adabe4a](https://github.com/ROCm/rocm-systems/pull/8361/commits/adabe4a66d905fcf06c101e2a86152e49ef6de2f)

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/symbol_lookup.cpp
mcao59: Addressed in [24946ce](https://github.com/ROCm/rocm-systems/pull/8361/commits/24946ced7b0577fced3a89578636e2d6ced067bc)

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/symbol_lookup.cpp
mcao59: Addressed in [24946ce](https://github.com/ROCm/rocm-systems/pull/8361/commits/24946ced7b0577fced3a89578636e2d6ced067bc)

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/symbol_lookup.cpp
mcao59: Simplified calculate_load_bias to the suggested approach. 
Previous code additionally protects against unusual mappings, which are likely outside rocattach’s practical contract for the known librocprofiler-register.so. Removed the unnecessary handlings.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/symbol_lookup.cpp
mcao59: Sure! Added more comments in the file

mcao59: I prefer keeping it together for this PR to avoid extra file/build churn while the focus is the symbol resolution logic design. I added comments around the two flows, hopefully making the current structure easier to follow, and we can split it later if this code grows further.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/symbol_lookup.cpp
mcao59: yeah, seems unnecessary, removed in [a844d3e](https://github.com/ROCm/rocm-systems/pull/8361/commits/a844d3eb00e35d15f7d51f1d6a2bc5a1847fb9da)

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/symbol_lookup.cpp
mcao59: sure, addressed in [a844d3e](https://github.com/ROCm/rocm-systems/pull/8361/commits/a844d3eb00e35d15f7d51f1d6a2bc5a1847fb9da)

projects/rocprofiler-sdk/tests/rocattach/symbol_lookup_test.cpp
mcao59: Addressed in [74e8f0f](https://github.com/ROCm/rocm-systems/pull/8361/commits/74e8f0f12214e36a8b42f1865bb72cd389a8df5b)

meserve-amd (approved):

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/rocattach.cpp
```diff
@@ -237,6 +237,60 @@ resolve_attach_tid(pid_t pid)
     return -1;
 }
 
+rocattach_status_t
+validate_target_tool_path(pid_t pid, std::string_view tool_lib_path)
+{
+    auto tool_path = std::filesystem::path{tool_lib_path};
+    if(tool_path.empty()) return ROCATTACH_STATUS_SUCCESS;
```
jrmadsen: Why return success here?

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/rocattach.cpp
```diff
@@ -237,6 +237,60 @@ resolve_attach_tid(pid_t pid)
...
+
+    if(!tool_path.is_absolute()) return ROCATTACH_STATUS_SUCCESS;
```
jrmadsen: Also maybe strange to return success here

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/rocattach.cpp
mcao59: Good point. An empty path should not be treated as valid, so I’ll reject it as an invalid argument. Fixed in [39a3848](https://github.com/ROCm/rocm-systems/pull/8361/commits/39a38485bf5fee14ff4c2b6634969c2f6b915e63)

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-rocattach/rocattach.cpp
mcao59: Relative paths remain supported because they are resolved by dlopen using the target process’s loader search path. I moved the absolute-path check to the caller and rename the helper so it is clear that only absolute paths are host-validated, please see [39a3848](https://github.com/ROCm/rocm-systems/pull/8361/commits/39a38485bf5fee14ff4c2b6634969c2f6b915e63)

jrmadsen (approved):

PR #8382, opened 2026-07-10
fix(rocdbgapi): accept smaller entry_size from older KFD snapshots

akondrat-amd:
## Motivation

Per kfd_ioctl.h, entry_size is an IN/OUT field — the caller passes the size of its struct, and KFD writes back the number of bytes it actually populated. An older KFD driver will return a smaller value (it has a smaller struct), which is perfectly valid. The != check incorrectly treated that as an error.

## Technical Details

Fix: Changed != to > in both locations. This correctly accepts an older KFD returning fewer bytes (the unread fields in the caller's zero-initialized vector stay zero), and only errors if KFD somehow returned a larger entry size than the struct we know about.

JIRA ID : AIROCGDB-620

### DIFF
--- a/projects/rocdbgapi/src/os_driver_kfd.cpp
+++ b/projects/rocdbgapi/src/os_driver_kfd.cpp
@@ -1443,7 +1443,7 @@ kfd_driver_t::kfd_agent_snapshot (kfd_dbg_device_info_entry *agents_infos,
   if (err == -ESRCH)
     return AMD_DBGAPI_STATUS_ERROR_PROCESS_EXITED;
   else if (args.device_snapshot.entry_size
-             != sizeof (kfd_dbg_device_info_entry)
+             > sizeof (kfd_dbg_device_info_entry)
            || err < 0)
     return AMD_DBGAPI_STATUS_ERROR;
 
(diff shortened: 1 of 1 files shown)


lumachad:
Let's add the JIRA ID to the description.

lumachad:
Could we please also have a more meaningful commit message with the description of the change? For rocdbgapi we do rebase and merge and not squash and merge, so we don't use the PR description/Title wheh merging.

projects/rocdbgapi/src/os_driver_kfd.cpp
```diff
@@ -1443,7 +1443,7 @@ kfd_driver_t::kfd_agent_snapshot (kfd_dbg_device_info_entry *agents_infos,
   if (err == -ESRCH)
     return AMD_DBGAPI_STATUS_ERROR_PROCESS_EXITED;
   else if (args.device_snapshot.entry_size
-             != sizeof (kfd_dbg_device_info_entry)
+             > sizeof (kfd_dbg_device_info_entry)
```
lumachad, line 1446: Now that we accept smaller/older entries, are there any features that might not work if we get a too old KFD? Do we need to guard those?

lancesix: There is a version we can query from the driver.

Once we start supporting a new version of the interface (where we grow one of such structs), we need to check the version before accessing the new field.  I do not think we use this for now, but if kfd_dbg_device_info_entry grows, we'll need to have something like this in the implementation:

```c++
amd_dbgapi_status_t
kfd_driver_base_t::agent_snapshot (
  os_agent_info_t *snapshots, size_t snapshot_count, size_t *agent_count,
  os_exception_mask_t exceptions_cleared) const
{
   ...

   agent_info.name = marketing_name (
      agent_info.vendor_id, agent_info.device_id, agent_info.revision_id);
    if (get_kfd_version () > {1, 16 /* or whatever  */})
      agent_info.?? = entry.new_field;

  ...
}
```

so if the driver did not fill in the entire entry as we know it, we stop reading after the last field a given kfd version knows.

akondrat-amd: We fill the record with zeroes on initialization and we use the extra space for capabilities flags that do not exist in the previous architectures, if KFD can follow this pattern, the we should be ok.

lancesix (commented):
The diff seems ok, but as already mentioned, please provide a full message in your commit.

lumachad (approved):
OK if CI is good.

lancesix (approved):

PR #8385, opened 2026-07-10
[clr] Allow 32-bit warp-sync masks on wave32 architectures

iassiour:
## Motivation
Allow 32-bit warp-sync masks on wave32 AMD targets. Code such as:
__shfl_sync(0xffffffff, value, 0, width)
currently fails to compile because HIP requires every explicit warp-sync mask to be a 64-bit integer, even on wave32 architectures where a 32-bit mask is the natural width.

## Technical Details

Replaces the duplicated sizeof(MaskT) == 8 checks in amd_warp_sync_functions.h with a shared mask-type predicate.
Allows 32-bit integer masks for wave32-only targets covered by this change while keeping the existing 64-bit requirement for other device targets.
Leaves the existing runtime behavior intact: wave32 masks still ignore the upper half via __hip_adjust_mask_for_wave32.
Adds Unit_hipShflSync_32BitMask to validate that __shfl_sync accepts a 32-bit mask and broadcasts lane 0 correctly.

## JIRA ID
Resolves ROCM-27352.

## Test Plan
Compile raw 32-bit-mask repro for gfx1250/gfx942/gfx950.
Compile the new guarded unit test for gfx1250/gfx942/gfx950.
Run existing CI.

## Test Result
Raw 32-bit mask compiles for wave32 targets covered by this change.
Raw 32-bit mask remains rejected on gfx942/gfx950 wave64 targets.
New Unit_hipShflSync_32BitMask syntax-checks cleanly for gfx1250/gfx942/gfx950.

## Submission Checklist

### DIFF
--- a/projects/clr/hipamd/include/hip/amd_detail/amd_warp_sync_functions.h
+++ b/projects/clr/hipamd/include/hip/amd_detail/amd_warp_sync_functions.h
@@ -82,7 +82,7 @@ template <typename T> __device__ inline T __hip_readfirstlane(T val) {
   union {
     unsigned long long l;
     T d;
-  } u;
+  } u = {};
   u.d = val;
   // NOTE: The builtin returns int, so we first cast it to unsigned int and only
   // then extend it to 64 bits.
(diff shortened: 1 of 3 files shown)


g-h-c:
If we do this, for consistency, all the __reduce_sync operations should also allow a 32-bit mask

ssahasra:
Keeping the mask argument independent of wavesize was a deliberate decision. The mask is always 64-bit so that a single HIP source will compile correctly to any target. Perhaps we need a deeper understanding of the need for a 32-bit mask. Is it just convenience? Or a real issue? Even if it is just convenience, how important is it?

iassiour:
> If we do this, for consistency, all the __reduce_sync operations should also allow a 32-bit mask

I think this is covered by the changes in reduce_op_sync https://github.com/ROCm/rocm-systems/pull/8385/changes#diff-c885872acc0fe466252f4a87941bfc669032ba494f1107cbd020e94e2847ac4bR327

But modified the  test to cover the case as well

iassiour:
closing based on the comments on https://amd-hub.atlassian.net/browse/ROCM-27352

PR #8386, opened 2026-07-10
feat(rccl-tests): Add AllGatherV grouped-broadcast mode

rahulvaidya20:
## Motivation

`broadcast_perf` issues one root per call and never triggers AllGatherV fusion (AICOMRCCL-467). No existing rccl-tests binary exercises the grouped-broadcast path or validates receiver buffers on every rank.

## Technical Details

Adds grouped-broadcast mode to `broadcast_perf`, enabled via `RCCL_TESTS_ALLGATHERV_ENABLE=1`. When set:
- `RunColl` issues `ncclGroupStart` + N `ncclBroadcast` calls (root=k for k in 0..nranks-1) + `ncclGroupEnd` — fuses into the AllGatherV ring kernel
- Buffer layout: `sendcount=chunkElems`, `recvcount=chunkElems×nranks`, aligned via `& -(16/eltSize)` matching `all_gather.cu`
- In-place layout follows AllGather convention: `sendInplaceOffset=chunkElems`, `recvInplaceOffset=0` — each rank's send chunk at `rank*chunkElems` in the recv buffer
- `InitData` seeds root k with `seed=k+1`; expected recv slot k uses same seed — validates every rank's every recv chunk, including in-place runs
- BW reporting uses allgather bus factor `(nranks-1)/nranks`
- Control path: `NCCL_ALLGATHERV_ENABLE=0` disables fusion, runs N independent broadcast kernels

## JIRA ID

JIRA ID: AICOMRCCL-1476

## Test Plan

Single-node 8-GPU and multi-node (2n/4n/8n/16n) on gfx942, single-node 8-GPU on gfx950. Fused vs unfused (`NCCL_ALLGATHERV_ENABLE=0`), `-N 10` stress, 64K–16G.

## Test Result

All sizes pass validation (`#wrong=0`) on gfx942 and gfx950, up to 16N×128 ranks×16G. Performance:
- Single-node: fused wins at ≥1M, peak +67% at 2–16M
- Multi-node: unfused faster below ~8M/broadcast; fused wins above, peak +50–52% at large sizes
- Crossover stable at ~8M bytes/broadcast across all multi-node configs (2n–16n)

## Submission Checklist

### DIFF
--- a/projects/rccl-tests/src/broadcast.cu
+++ b/projects/rccl-tests/src/broadcast.cu
@@ -8,15 +8,56 @@
 #include "cuda_runtime.h"
 #include "common.h"
 
+// Set RCCL_TESTS_ALLGATHERV_ENABLE=1 to run grouped-broadcast (AllGatherV fusion) mode
+static int broadcast_grouped = [] {
+  const char* s = getenv("RCCL_TESTS_ALLGATHERV_ENABLE");
+  return s != nullptr && s[0] == '1';
+}();
+
 void BroadcastGetCollByteCount(size_t *sendcount, size_t *recvcount, size_t *paramcount, size_t *sendInplaceOffset, size_t *recvInplaceOffset, size_t count, size_t eltSize, int nranks) {
-  *sendcount = count;
-  *recvcount = count;
-  *sendInplaceOffset = 0;
-  *recvInplaceOffset = 0;
-  *paramcount = *sendcount;
+  if (broadcast_grouped) {
+    if (count < (size_t)nranks) {
+      *sendcount = *recvcount = *paramcount = *sendInplaceOffset = *recvInplaceOffset = 0;
+      return;
+    }
+    size_t chunkElems = (count / nranks) & -(16 / eltSize);
+    if (chunkElems == 0) {
+      *sendcount = *recvcount = *paramcount = *sendInplaceOffset = *recvInplaceOffset = 0;
+      return;
+    }
+    *sendcount  = chunkElems;
+    *recvcount  = chunkElems * nranks;
+    *paramcount = chunkElems;
+    *sendInplaceOffset = chunkElems;
+    *recvInplaceOffset = 0;
+  } else {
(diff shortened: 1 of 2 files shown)


alex-breslow-amd:
I'm a little confused here.  If we need this functionality, why wouldn't we just implement it directly in RCCL.  Can you comment @rahulvaidya20?

rahulvaidya20:
> I'm a little confused here. If we need this functionality, why wouldn't we just implement it directly in RCCL. Can you comment @rahulvaidya20?

@alex-breslow-amd PR #7991 does the RCCL AGV enablement. This is to test the AGV enablement.

thomas-huber:
Not sure if rccl-tests DOCs include the env vars, but if so, maybe we should have NCCL_ALLGATHERV_ENABLE=0? (I left similar comment about doc'ing this in rccl itself, and i saw you already did some of that)

thomas-huber:
I see the ticket calls for registering the binary in Makefile and CMakeLists.txt, @rahulvaidya20 should we do that?

thomas-huber (approved):

rahulvaidya20:
@thomas-huber  added the functionality directly in existing (bcast) collective. It can now be enabled by setting an env var. Will document in readme.

projects/rccl-tests/src/broadcast.cu
```diff
@@ -8,15 +8,56 @@
 #include "cuda_runtime.h"
 #include "common.h"
 
+// Set RCCL_TESTS_ALLGATHERV_ENABLE=1 to run grouped-broadcast (AllGatherV fusion) mode
+static int broadcast_grouped = [] {
+  const char* s = getenv("RCCL_TESTS_ALLGATHERV_ENABLE");
+  return s ? atoi(s) : 0;
```
alex-breslow-amd, line 14: I think you can do something simpler than this, which is just check that s != nullptr && s[0] == '1'.

garzaran (approved):

PR #8387, opened 2026-07-10
feat(rocr): enforce required gfx1250 hotswap rewrites

harsh-amd:
## Summary
- Add `HSA_HOTSWAP_STRICT_MODE`, default off, to request COMGR `AMD_COMGR_HOTSWAP_REWRITE_FLAG_STRICT_MODE` for non-A0 `gfx1250` HotSwap rewrites.
- Keep A0 retargeting outside ROCR strict mode; A0 continues to use the normal B0-to-A0 hotswap path and does not set the COMGR strict flag.
- Treat required rewrites as non-fallback operations: A0 B0-to-A0 retargeting and opt-in B0 strict rewrites return `HSA_STATUS_ERROR_INVALID_CODE_OBJECT` if COMGR cannot produce the rewrite or the rewritten code object cannot be loaded.
- Keep entry-trampoline selection independent from strict-mode mask workarounds, so `AMD_COMGR_HOTSWAP_ENTRY_TRAMPOLINES=0` only disables entry trampolines.
- Preserve the original fallback behavior for optional rewrites and no-decision paths.
- Document `HSA_HOTSWAP_STRICT_MODE` in the runtime environment variable table.
- Depends on ROCm/llvm-project#3312 for COMGR B0 strict mask workaround support.

## JIRA ID
JIRA ID : ROCM-27304
JIRA ID : PLAT-204339

## Testing
- `cmake --build build-rocrtst-hotswap-review --target hotswap_rewrite -j 8`
- `ctest --test-dir build-rocrtst-hotswap-review -R hotswap_rewrite --output-on-failure`
- Result: `hotswap_rewrite` 30/30 passed.

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
@@ -322,7 +322,7 @@ std::optional<RewriteDecision> DecideHotswapRewrite(
   RewriteDecision decision{source_isa, source_isa,
                            request_entry_trampolines,
                            request_strict_mode};
-  if (request_strict_mode) {
+  if (source_gfx == kGfx1250) {
     decision.source_isa =
         WithGfx1250SteppingFeature(source_isa, Gfx1250Stepping::kB0);
     decision.target_isa =
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
@@ -96,15 +96,10 @@ bool IsEnvFlagEnabled(const char* name) {
 
 bool IsHotswapDisabledByEnv() { return IsEnvFlagEnabled("HSA_HOTSWAP_DISABLE"); }
 
-bool AreEntryTrampolinesRequested() {
-  // Entry-trampoline rewriting is opt-in: disabled unless the caller sets
-  // AMD_COMGR_HOTSWAP_ENTRY_TRAMPOLINES to a truthy value. Unset falls back to
-  // the compiled-in default; false-like values keep it disabled.
+bool IsGfx12_5RewriteRequested() {
   constexpr char kEnvName[] = "AMD_COMGR_HOTSWAP_ENTRY_TRAMPOLINES";
-  if (!os::IsEnvVarSet(kEnvName)) {
-    return kDefaultEntryTrampolinesEnabled;
-  }
-  return IsEnvFlagEnabled(kEnvName);
+  // Default-on policy owned by ROCR: only literal "0" opts out.
+  return !os::IsEnvVarSet(kEnvName) || os::GetEnvVar(kEnvName) != "0";
 }
 
 bool IsVerboseLoggingEnabled() {
@@ -131,6 +126,7 @@ struct ComgrHotswapRewriteOptions {
 constexpr int kComgrStatusSuccess = 0;
 constexpr int kComgrDataKindExecutable = 0x8;
 constexpr uint64_t kComgrHotswapRewriteFlagEntryTrampolines = 0x1;
+constexpr uint64_t kComgrHotswapRewriteFlagStrictMode = 0x2;
 
 struct ComgrApi {
   os::LibHandle lib = nullptr;
@@ -282,8 +278,9 @@ std::string WithGfx1250SteppingFeature(const std::string& isa_name,
 
 bool HasCandidateHotswapRewrite(const AgentGfxRevision& gfx,
                                 const RewriteOptions& options) {
-  return IsHotswapSupportedGfxRevision(gfx) ||
-      (options.entry_trampolines_enabled && IsGfx12_5Target(gfx.gfx_target));
+  return (options.strict_mode_enabled && gfx.gfx_target == kGfx1250) ||
+         (options.gfx12_5_rewrite_enabled &&
+          IsGfx12_5Target(gfx.gfx_target));
 }
 
 std::optional<RewriteDecision> DecideHotswapRewrite(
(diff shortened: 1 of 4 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -293,18 +302,31 @@ std::optional<RewriteDecision> DecideHotswapRewrite(
   const std::string target_gfx = ExtractGfxTarget(target_isa);
   if (IsHotswapSupportedGfxRevision(gfx) && source_gfx == kGfx1250 &&
```
shwetagkhatri, line 305: Can we please add code comments here explaining why A0 path uses legacy API -  hotswap_rewrite() without the strict mode flag, while B0 strict uses hotswap_rewrite_with_options()? 
I see legacy API already handles B0 to A0 instruction patching internally and doesn't need the strict mode flag.  But it will help clarify this to  future maintainers in ROCr.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -68,6 +68,9 @@ namespace {
 
 std::mutex g_retained_rewritten_elf_buffers_mutex;
 std::unordered_map<uint64_t, std::vector<OwnedElfBuffer>> g_retained_rewritten_elf_buffers;
+#ifdef ROCR_HOTSWAP_TESTING
+bool g_force_retarget_code_object_failure_for_testing = false;
```
shwetagkhatri: This g_force_retarget_code_object_failure_for_testing  should be std::atomic<bool> type for thread safety.

shwetagkhatri (changes requested):
New tests (HotswapRewriteDecision.*, HotswapRewrite.RuntimeLoad*Required*) are not registered in rocrtst/suites/test_common/test_categories.yaml.  Please take a note before adding any tests to rocrtsts next time.

powderluv:
Do you want the strict mode to work even in the B0 -> A0 path ?

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
powderluv: I think even A0 needs to use the same path as B0 and provide the strict mode since the issue would exist for both variants

harsh-amd: Added a comment explaining why the A0 path intentionally uses the legacy rewrite API, while B0 strict rewrites use the options API.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
harsh-amd: Updated the test-only flag to `std::atomic<bool>` and used relaxed load/store since it does not synchronize any other state.

harsh-amd:
These tests are built as the standalone `hotswap_rewrite` target and already have the `quick`, `standard`, `pr`, and `pre-commit` CTest labels. Since `test_categories.yaml` currently only applies to `rocrtst64`, adding these test names there would not affect the standalone target. Are the existing CTest labels sufficient, or should standalone targets also be added to the YAML?

harsh-amd:
> Do you want the strict mode to work even in the B0 -> A0 path ?

So far we haven't found a need for this. The D# WA is always required on A0, while it can be avoided on B0. Hence currently we only have strict mode for B0 -> B0.

shwetagkhatri:
> These tests are built as the standalone `hotswap_rewrite` target and already have the `quick`, `standard`, `pr`, and `pre-commit` CTest labels. Since `test_categories.yaml` currently only applies to `rocrtst64`, adding these test names there would not affect the standalone target. Are the existing CTest labels sufficient, or should standalone targets also be added to the YAML?

I see it now. Yes  the existing CTest labels are the correct for standalone tests.

shwetagkhatri (approved):
LGTM.

dayatsin-amd (approved):

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -456,25 +496,43 @@ bool TryRetargetCodeObject(const CodeObjectView& code_object, hsa_agent_t agent,
...
-              rewritten ? *out_elf_size : 0, rewritten ? 1 : 0);
-  return rewritten;
+              decision->request_entry_trampolines, decision->request_strict_mode,
+              decision->rewrite_required, code_object.size, rewritten ? *out_elf_size : 0,
+              rewritten ? 1 : 0);
+  if (rewritten) {
+    return {RetargetCodeObjectStatus::kRewritten,
+            decision->rewrite_required};
+  }
+  if (decision->rewrite_required) {
```
cfreeamd, line 525: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** The `} else` at the `#endif` boundary is correct — in non-testing builds the preprocessor removes the entire `if`/`else`, leaving just the standalone `{ rewritten = RetargetCodeObject(...); }` block — but the `else` keyword straddling a `#endif` boundary can surprise readers. Consider inverting the condition so the `#ifdef` block doesn't straddle an `else`:
```cpp
#ifdef ROCR_HOTSWAP_TESTING
  if (!g_force_retarget_code_object_failure_for_testing.load(std::memory_order_relaxed))
#endif
    rewritten = RetargetCodeObject(...);
```
or hoisting to a boolean flag before the branch.

cfreeamd (commented):

The ROCr files look correct overall — `hotswap.hpp`, `hotswap.cpp`, and `env_variables.rst` all look consistent with the new `rewrite_required` enforcement semantics.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
harsh-amd: Agreed this would read cleaner. Since this is readability-only and CI is already deep into the matrix, I’ll handle
this in a follow-up PR unless we need to respin this PR for another reason.

harsh-amd: Agreed this reads cleaner. I opened a stacked follow-up PR with the cleanup so this PR does not need another CI restart: https://github.com/harsh-amd/rocm-systems/pull/3

PR #8389, opened 2026-07-10
fix(rocprof-systems): Fix transpose to work under MPI

sputhala-amd:
## Motivation

Under MPI, transpose only ran GPU work on ranks mapped 1:1 to a distinct device, and worker threads never selected their stream's device — so ranks bound to a non-zero GPU crashed with "invalid resource handle" and ranks beyond the device count did no GPU work at all.

## Technical Details

- **Per-thread device selection**: added `hipSetDevice(devid)` at the start of `run()`. HIP tracks the active device per host thread, and freshly spawned `std::thread` workers default to device 0; without this they launched kernels on a stream owned by a different device and failed with "invalid resource handle" whenever a rank was mapped to a non-zero device (e.g. rank 1 → device 1).
- **Round-robin GPU sharing**: replaced the restrictive guard `if(rank == devid && rank < ndevice)` with `if(ndevice > 0)`, so ranks map round-robin via `devid = rank % ndevice` and ranks beyond the device count share a GPU (e.g. 2 ranks on 1 GPU both use device 0) instead of skipping GPU work.
- Removed the stale "temporary workaround in rocprof-sys when HIP + MPI is enabled" comment; the underlying limitation no longer applies — verified that multiple ranks sharing a single GPU each collect full HW counter data.
 - Thread wiring: the device id is now threaded through to every worker `(run(rank, i, devid, stream, ...))` and the main-thread invocation, and the device-assignment + GPU-work blocks are merged into a single `if(ndevice > 0)` block

## JIRA ID

JIRA ID : AIPROFSYST-622

## Test Plan

Manual and CI test.

## Test Result

Pass.

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-systems/examples/transpose/transpose.cpp
+++ b/projects/rocprofiler-systems/examples/transpose/transpose.cpp
@@ -107,8 +107,14 @@ size_t nsync    = 10;
 }  // namespace
 
 void
-run(int rank, int tid, hipStream_t stream, int argc, char** argv)
+run(int rank, int tid, int devid, hipStream_t stream, int argc, char** argv)
 {
+    // HIP tracks the active device per host thread, so each worker thread must
+    // select the same device its stream was created on. Otherwise kernel
+    // launches fail with "invalid resource handle" whenever a rank is mapped to
+    // a non-zero device (e.g. rank 1 -> device 1 under MPI).
+    HIP_API_CALL(hipSetDevice(devid));
+
     unsigned int M = 4960 * 2;
     unsigned int N = 4960 * 2;
     if(argc > 2) nitr = atoll(argv[2]);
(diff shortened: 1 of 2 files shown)


kcossett-amd:
I had to rerun the gfx950 workflow, some odd GPU hangs occurred:
```
2026-07-10T02:55:18.4245615Z HW Exception by GPU node-6 (Agent handle: 0x5c8feb1f4aa0) reason :GPU Hang
```
I don't believe these are due to your changes.

kcossett-amd:
On another note, this time to myself, the `sampling` variants of `transpose` are still using more than 1 GPU. I suspect its because `base_environment` sets `"ROCPROFSYS_SAMPLING_GPUS": "all",`.

Inside our code, `all` maps to "enumerate everything" which is done though `AMD-SMI`, which most likely bypasses `HIP_VISIBLE_DEVICES/ROCR_VISIBLE_DEVICES`... which is not good because that is how the OSCII MIXXX runners enforce isolation to 1 GPU.

I'll have to investigate this quickly.

dgaliffiAMD (approved):

Issue #8364, opened 2026-07-09
feat(rocjitsu): authorize DBG_TRAP by real ptrace


amd-arosa:
Running ROCgdb against the emulator requires the daemon to route and authorize the debugger's `DBG_TRAP` ioctls against a *different* client process (the inferior). Two gaps blocked that: the cross-process authorization was circular (it compared against a debugger pid recorded only during `ENABLE`), and the `DBG_TRAP` embedded output pointers were not marshaled across the RPC. Fixing both lets a debugger connected to the daemon enable a session on, and enumerate the agents of, a separate inferior.

## Scope
Authorize `DBG_TRAP` by the real OS ptrace relationship and marshal it over the daemon.

- Replace the stored-pid check with `tracer_pid_of()` reading `/proc/<pid>/status` TracerPid and comparing to the caller's client pid.
- Implement `SET_EXCEPTIONS_ENABLED` (records the forwarded-exception mask).
- Marshal the two attach-time output pointers (`ENABLE.rinfo_ptr`, `GET_DEVICE_SNAPSHOT.snapshot_buf_ptr`) over the RPC: the client reserves a trailing buffer and copies it back; the server repoints them per `args->op`.

Only the two attach-time pointers are marshaled; ops added later extend the same per-op switch.

## Proof of concept
https://github.com/ROCm/rocm-systems/commit/106c163e2d5f6642484c84584f6d382b3491b5de

amd-arosa:
```c++
  // Real kernel: EPERM unless target == p or ptrace_parent(target) == current,
  // re-checked on every call. We have no live ptrace relationship to consult,
  // so we approximate it with "a prior ENABLE from this same caller succeeded".
  // That approximation can never be satisfied by a genuine external debugger's
  // *first* ENABLE (sess.debugger_pid starts at 0), so only self-debug is
  // reachable today; cross-process attach needs real ptrace-state tracking.
  if (!self_debug && args->op != KFD_IOC_DBG_TRAP_DISABLE &&
      sess.debugger_pid != caller.client_pid())
    return -EPERM;
```

Issue #8371, opened 2026-07-10
feat(rocjitsu): drive GPU wave debugging end-to-end with ROCgdb


amd-arosa:
The prior issues stop a wave, serialize it into the CWSR area, and raise the event. This closes the loop so real ROCgdb, through `mirage run`, sets a breakpoint on a GPU kernel, stops the trapping wave, reads its registers/PC/disassembly, and continues to a correct result - the first fully working GPU-kernel debug session on the emulator, with a CI harness that proves it.

## Scope
Drive GPU wave debugging end-to-end and add the CI harness.

- Implement the ROCr `RUNTIME_ENABLE` debugger handshake (raise `EC_PROCESS_RUNTIME`, wait bounded for the debugger's ack, return the saved `kfd_runtime_info` so dbgapi installs `r_brk`). Pass the debugger's `dbg_fd` notifier pipe to the daemon out-of-band (`SCM_RIGHTS` on `ENABLE`) so a wave stop can wake it.
- Add `deserialize_queue_cwsr` + `apply_cwsr_to_wave`: `RESUME_QUEUES` reloads a wave from its CWSR area (picking up the debugger's edits) and unhalts it. Marshal `QUERY_EXCEPTION_INFO`.
- Add the `RocgdbDebug` CTest (`tests/rocgdb/debug_kernel.sh` + `add_one.hip`): compiles a HIP kernel with device debug info and drives real ROCgdb through `mirage run`. Skips cleanly when hipcc/rocgdb/mirage are unavailable.

Deterministic single-step and displaced stepping are the next issues; `continue` past the entry breakpoint already works here.

## Proof of concept
https://github.com/ROCm/rocm-systems/commit/19781fb990ca4a14d9cdc5bfb1441b34e52c5f4f

PR #8393, opened 2026-07-10
feat(rocjitsu): route AMDKFD_IOC_DBG_TRAP into a debug-session dispatcher 

### LINKED ISSUE #8281: feat(rocjitsu): route AMDKFD_IOC_DBG_TRAP into a debug-session dispatcher
rocm-dbgapi drives every debugger action through the `AMDKFD_IOC_DBG_TRAP` ioctl. rocjitsu's KFD emulation dropped that ioctl into the silent `default: return 0` arm, so a debugger saw every operation "succeed" with no effect. This routes it into a real per-process dispatcher that enforces the driver's validation contract - the foundation the rest of the ROCgdb bring-up builds on.

## Scope
Add a `DBG_TRAP` dispatcher that mirrors the kernel's `kfd_ioctl_set_debug_trap()` validation ladder before its op switch.

- Add `KfdProcess::DebugSession` (enabled, runtime_state, exception mask, dbg_fd, debugger_pid).
- `SimulatedDriver::debug_trap_ioctl` enforces the pre-switch ladder: unknown pid -> `ESRCH`; cross-process op without a debugger -> `EPERM`; op without an enabled session -> `EINVAL`; a HW-op before runtime-enable -> `EPERM`; address-watch op with a bad gpu_id -> `ENODEV`.
- `ENABLE` allocates the session and returns `kfd_runtime_info`; `DISABLE` tears it down. The remaining ops return `-ENOSYS` until later issues wire them up.

Authorization here is by-pid only; the real ptrace check and the daemon transport come in #8364.

## Proof of concept
https://github.com/ROCm/rocm-systems/commit/86235532fad783abab205971e1959019dcee9b47

amd-arosa:
## Motivation

rocm-dbgapi drives every debugger action through the `AMDKFD_IOC_DBG_TRAP` ioctl.
rocjitsu's KFD emulation dropped that ioctl into the silent `default: return 0`
arm, so a debugger saw every operation "succeed" with no effect. This routes it
into a real per-process debug-session dispatcher that enforces the driver's
validation contract, and — for daemon-served workloads (`mirage`,
`rocjitsu --daemon`) — carries the debugger's notifier fd across the process
boundary so a later wave-stop can actually wake the debugger.

Resolves #8281.

### Debug-session dispatcher

Adds a `DBG_TRAP` dispatcher that mirrors the kernel's
`kfd_ioctl_set_debug_trap()` validation ladder (`amd/amdkfd/kfd_chardev.c`)
before its op switch.

- Add `KfdProcess::DebugSession` (`enabled`, `runtime_state`,
  `exception_enable_mask`, `dbg_fd`, `owned_dbg_fd`, `debugger_pid`) guarded by a
  per-process `debug_mutex_`; each field is documented against its `kfd_priv.h`
  counterpart.
- Record the caller's Linux pid at `open()` so `DBG_TRAP` can resolve a
  self-debug target (`args->pid == getpid()`), mirroring the kernel's
  `kfd_lookup_process_by_pid()`.
- `SimulatedKfd::debug_trap_ioctl` enforces the pre-switch ladder: unknown pid →
  `ESRCH`; a foreign client driving a session it does not own → `EPERM`; an op
  without an enabled session → `EINVAL`; a HW-op before runtime-enable → `EPERM`;
  an address-watch op with a bad `gpu_id` → `ENODEV`; a second `ENABLE` →
  `EALREADY`.
- `ENABLE` validates `dbg_fd` (`fcntl(F_GETFD)` → `EBADF` for an unusable
  descriptor, matching the kernel's `fget()`), allocates the session, and copies
  `kfd_runtime_info` (`r_debug`, `ttmp_setup`, `runtime_state`) back to the
  debugger — a truncating `min(user_size, sizeof)` copy that still reports the
  full struct size, as the kernel does. `ENABLE` now snapshots the runtime state
  under a single `runtime_mutex_` acquisition (lock order `debug_mutex_` →
  `runtime_mutex_`) so the marshaled `runtime_state` cannot disagree with
  `r_debug`/`ttmp_setup`.
- `DISABLE` tears the session down. Recognized-but-unwired ops
  (`QUERY_DEBUG_EVENT`, `SET_EXCEPTIONS_ENABLED`, `GET_DEVICE_SNAPSHOT`, …)
  return `-ENOSYS` so a debugger can tell a stubbed-but-valid op apart from a
  genuinely unknown one, which still returns `-EINVAL`. Each op graduates out of
  the `ENOSYS` group as its handler lands.

### Daemon transport — notifier fd via `SCM_RIGHTS`

In daemon mode the debugger's `dbg_fd` is a number in the client's fd table and
is meaningless to the daemon. The client hands the real descriptor over
out-of-band as `SCM_RIGHTS` ancillary data on `DBG_TRAP ENABLE`; the daemon
receives it in its own fd space and the `rj_vm_execute_as()` glue substitutes it
into the ioctl's `dbg_fd` — the same fd the real kernel would receive through the
ioctl.

- Wired end-to-end through `RemoteDriver` (client), the C++ daemon
  (`tools/rocjitsu`), and the mirage Rust daemon (`rocjitsu/src/daemon.rs`), plus
  the `DBG_TRAP` embedded-pointer marshalling for the runtime-info and
  device-snapshot buffers.
- `UniqueFd` — a small move-only RAII fd owner. In daemon mode the session owns
  the transferred descriptor and releases it deterministically on `DISABLE` or
  process teardown (no manual close, no leak on abnormal teardown). Local mode
  borrows the debugger's own fd and never closes it.
- The daemon header read requires the full RPC header, dropping a truncated
  header (and reclaiming any fd that arrived with it) rather than acting on a
  partial read.

### Driver logging

Adds a dedicated driver logger group and routes the `DBG_TRAP` handler's
diagnostics through it.

### Notes

- Cross-process attach is intentionally closed by the by-pid gate until #8364:
  `debugger_pid` is only set by a successful `ENABLE`, so a first cross-process
  `ENABLE` can never pass the check. Self-debug bypasses this, so local mode —
  the only supported path today — is unaffected.
- `DBG_TRAP` self-debug caches the caller pid at `open()`, so it requires a
  post-fork re-open. The interposer's `reset_after_fork()` drops the driver and
  forces a re-open, so this is satisfied in practice; documented on the handler.
- The `kfd_ioctl.h` doc block lists `EACCES` for a HW-op before runtime-enable,
  but the kernel code path returns `EPERM`; this follows the code (what
  rocm-dbgapi observes).
- Authorization is by-pid only; the real ptrace check is future work.
- `kfd_process.h` now includes `<sys/types.h>` for `pid_t` (previously compiled
  only transitively).

## JIRA ID

Resolves #8281.

## Test Plan

18 `DBG_TRAP` tests (`rocjitsu_tests`):

- **Validation ladder (local self-debug):** `DbgTrapUnknownPidReturnsESRCH`,
  `DbgTrapOpBeforeEnableReturnsEINVAL`, `DbgTrapBareDisableReturnsEINVAL`,
  `DbgTrapEnablePopulatesRuntimeInfoThenDisable`,
  `DbgTrapDoubleEnableReturnsEALREADY`, `DbgTrapEnableBadFdReturnsEBADF`,
  `DbgTrapHwOpWithoutRuntimeReturnsEPERM`, `DbgTrapWatchBadGpuReturnsENODEV`,
  `DbgTrapAdmittedOpNotYetImplemented` (recognized op → `ENOSYS`).
- **Concurrency:** `DbgTrapConcurrentEnableDisableIsRaceFree` hammers
  `ENABLE`/`DISABLE` on one session from 8 threads × 250 iterations to exercise
  `debug_mutex_` under ThreadSanitizer; every call returns a well-defined code
  (`0`/`EALREADY`/`EINVAL`), never a race or torn state.
- **Runtime-info marshalling / fd ownership:**
  `DbgTrapLocalDisableLeavesDebuggerFdOpen`,
  `DbgTrapEnableUndersizedRuntimeInfoTruncates`,
  `DbgTrapEnableOversizedRuntimeInfoPreservesTailInDaemonMode`.
- **Daemon `SCM_RIGHTS` notifier transfer (`DbgTrapDaemonTest`, real
  `rj_vm_execute_as` path):** `EnableAdoptsTransferredNotifierFd`,
  `DisableClosesAdoptedNotifierFd`, `EnableWithoutTransferredFdReturnsEbadf`,
  `EnableSendsNotifierFdOverScmRights`, `ForeignClientCannotDriveAnothersSession`.

## Test Result

`ctest -R "DbgTrap"` → 18/18 pass. Clean build.

## Submission Checklist


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
```diff
@@ -1833,6 +1838,115 @@ int SimulatedDriver::runtime_enable_ioctl(KfdProcess &proc, void *arg) {
...
+      return -ENODEV;
+  }
+
+  // https://github.com/torvalds/linux/blob/a635d6748234582ea287c5ffeae28b9b23f91c7e/drivers/gpu/drm/amd/amdkfd/kfd_chardev.c#L3158
+  switch (args->op) {
+  // https://github.com/torvalds/linux/blob/a635d6748234582ea287c5ffeae28b9b23f91c7e/drivers/gpu/drm/amd/amdkfd/kfd_debug.c#L788-L847
+  case KFD_IOC_DBG_TRAP_ENABLE: {
+    if (sess.enabled)
+      return -EALREADY; // target process is already debug enabled
+    sess.enabled = true;
```
kuhar, line 1853: We should validate `dbg_fd` before setting `sess.enabled` and return `-EBADF` for an unusable descriptor, matching `kfd_dbg_trap_enable()`. The current handler accepts every value; four new tests even enable successfully with `KFD_INVALID_FD`. Please use a real eventfd in the success tests and add an invalid-fd case so the stored notification target cannot be unusable from the start.

kuhar (commented):
A few comments

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
```diff
@@ -1833,6 +1838,115 @@ int SimulatedDriver::runtime_enable_ioctl(KfdProcess &proc, void *arg) {
...
+    {
+      std::lock_guard<std::mutex> rlk(target->runtime_mutex_);
+      info.r_debug = target->runtime_state_.r_debug;
+      info.ttmp_setup =
+          (target->runtime_state_.mode_mask & KFD_RUNTIME_ENABLE_MODE_TTMP_SAVE_MASK) ? 1u : 0u;
+    }
+    info.runtime_state = sess.runtime_state;
+    size_t copy_size = std::min(static_cast<size_t>(args->enable.rinfo_size), sizeof(info));
+    if (args->enable.rinfo_ptr != 0 && copy_size > 0)
+      std::memcpy(reinterpret_cast<void *>(static_cast<uintptr_t>(args->enable.rinfo_ptr)), &info,
```
kuhar, line 1880: We should either marshal the `DBG_TRAP` runtime-info buffer through the RPC layer here or reject `ENABLE` in daemon mode until #8364 lands. `RemoteDriver::has_embedded_pointers()` does not include this ioctl, so this `memcpy` dereferences a client address inside the daemon. In a current-build repro, `rinfo_size == 0` succeeds, while a 16-byte client stack buffer makes the RPC fail and terminates the daemon.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
amd-arosa: 833d32b68b0f98bd4388953600adec14ab2a0a0c

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
amd-arosa: 801c0178cc228bc68763f35916d58c81e419eb49

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/remote_driver.cpp
```diff
@@ -385,6 +402,24 @@ int RemoteDriver::send_ioctl(unsigned long request, void *arg) {
       buf.resize(buf.size() + aperture_args->num_of_nodes * sizeof(kfd_process_device_apertures));
       break;
     }
+    case AMDKFD_IOC_DBG_TRAP: {
+      auto *dbg = reinterpret_cast<kfd_ioctl_dbg_trap_args *>(args_base);
+      size_t inline_size = 0;
+      switch (dbg->op) {
+      case KFD_IOC_DBG_TRAP_ENABLE:
+        inline_size = dbg->enable.rinfo_size;
```
kuhar: We should bound the DBG_TRAP ENABLE inline buffer and response copy to the bytes the handler can produce. `rinfo_size` is caller capacity, but the handler writes only `min(input_size, sizeof(kfd_runtime_info))` and reports the struct size; allocating and copying the full capacity overwrites data beyond the returned result. In a 48-byte repro, local mode preserved the 32-byte tail while daemon mode overwrote it, even though both reported 16 bytes. Please marshal at most `min(input_capacity, sizeof(kfd_runtime_info))`, then copy back at most the original capacity, returned `rinfo_size`, and actual inline bytes. A daemon-mode oversized-buffer test should check the tail, returned info, and daemon liveness, which would also lock in the original pointer-marshalling crash fix.

kuhar (commented):
A few comments

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
kuhar: We should make daemon-mode ENABLE reject or correctly transport `dbg_fd` before storing it. The `!daemon_mode_` guard skips validation, and the RPC sends only the client fd number, which does not refer to the client eventfd in the daemon. At `801c0178cc`, `KFD_INVALID_FD` returns `-EBADF` locally but succeeds and enables the session through the daemon. Please either transfer the eventfd with `SCM_RIGHTS`, keep notification delivery client-side, or reject daemon-mode ENABLE until fd transport exists, and cover the chosen behavior with a daemon-mode test.

amd-arosa: SCM_RIGHTS was originally planned for #8371
moving it to here.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
```diff
@@ -1707,6 +1710,120 @@ int SimulatedKfd::runtime_enable_ioctl(KfdProcess &proc, void *arg) {
...
+  for (auto &[id, proc] : processes_)
+    if (proc->client_pid() == pid)
+      return proc;
+  return nullptr;
+}
+
+// in real kernel, amd/amdkfd/kfd_chardev.c kfd_ioctl_set_debug_trap
+int SimulatedKfd::debug_trap_ioctl(KfdProcess &caller, void *arg) {
+  auto *args = static_cast<kfd_ioctl_dbg_trap_args *>(arg);
+  util::Logger::cp("DBG_TRAP pid=", args->pid, " op=", args->op);
```
atgutier: We should probably add a new logger group for the driver.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/remote_driver.cpp
amd-arosa: 9aede935ba4c06cca8e04fc7267025053266f613

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
```diff
@@ -1707,6 +1710,120 @@ int SimulatedKfd::runtime_enable_ioctl(KfdProcess &proc, void *arg) {
...
+      std::memcpy(reinterpret_cast<void *>(static_cast<uintptr_t>(args->enable.rinfo_ptr)), &info,
+                  copy_size);
+    args->enable.rinfo_size = sizeof(info);
+    return 0;
+  }
+  case KFD_IOC_DBG_TRAP_DISABLE:
+    sess = KfdProcess::DebugSession{};
+    return 0;
+  default:
+    return -ENOSYS;
```
atgutier, line 1910: In kfd_chardev.c the real kernel returns -EINVAL here so we should respect that because SW above may have expectations about the errnos.

atgutier (commented):
Additional review comments from an agent:

- Cross-process ENABLE is structurally impossible, not just "deferred" —
  simulated_kfd.cpp:1744-1746. The kernel's cross-process guard is a live ptrace check that admits the
   first cross-process ENABLE (the ptrace relationship predates any KFD session). The PR replaces it
  with sess.debugger_pid != caller.client_pid(), but debugger_pid is only set by a successful ENABLE
  (line 1796) — so on a first attach debugger_pid==0, the check returns EPERM, and it can never
  succeed. Self-debug bypasses this, so local mode (the only supported path now) is unaffected. But
  the PR comment "authorization is by-pid only" understates that cross-process is fully closed. Fix:
  exempt ENABLE from the by-pid check, or clearly document that cross-process attach is intentionally
  closed until #8364 — otherwise a future author will assume the skeleton partially works.
 - runtime_mutex_ dropped and reacquired within one ENABLE — simulated_kfd.cpp:1753-1756 /
  1806-1810. runtime_enabled is read in one critical section, then r_debug/mode_mask in a second, with
   runtime_mutex_ released in between (only debug_mutex_ held). A concurrent RUNTIME_ENABLE/DISABLE
  can change state between them, so the marshaled runtime_state can disagree with r_debug/ttmp_setup.
  Benign today but latent. Fix: hold runtime_mutex_ once across the whole ENABLE marshal (ordering is
  already debug_mutex_ → runtime_mutex_, so widening is safe).
 - Test coverage gaps — the 7 tests assert exact errnos per rung and check marshaled values (good),
  but nothing covers: self-vs-other pid resolution (ESRCH vs success), two-thread ENABLE/DISABLE race
  (the entire point of debug_mutex_), disable-without-enable → EINVAL, or fork staleness. Add at
  minimum the bare-DISABLE→EINVAL test and a threaded enable/disable smoke test under TSan.
  - fork() leaves client_pid_ stale in the child — cached at open() via getpid(); a self-debug
  DBG_TRAP on an inherited fd would ESRCH. Masked today because the interposer's reset_after_fork()
  nulls the driver so the child re-opens and re-caches. Optionally assert/document that DBG_TRAP
  requires a post-fork re-open.
  - pid_t used in kfd_process.h without #include <sys/types.h> — compiles only transitively;
  pre-existing, not introduced here.
  - The debugger_pid vs client_pid() comparison reads like "is the caller the debugger" — a one-line
  comment noting it only admits re-entrant ops from an already-attached debugger would prevent
  confusion (ties to MINOR #1).

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
amd-arosa: fe2b0ac6fd04d520fe852b0fd18ac1b4b87d0aba

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
amd-arosa: 27d2449d2746265f4ef9de9dee4ad420126d8bf8

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/remote_driver.cpp
```diff
@@ -460,6 +527,27 @@ int RemoteDriver::send_ioctl(unsigned long request, void *arg) {
...
+        size_t copy_len = 0;
+        switch (dbg->op) {
+        case KFD_IOC_DBG_TRAP_ENABLE:
+          dst = reinterpret_cast<void *>(saved_dbg_rinfo_ptr);
+          copy_len = std::min(static_cast<size_t>(saved_dbg_rinfo_size),
+                              std::min(static_cast<size_t>(dbg->enable.rinfo_size), extra));
+          break;
+        case KFD_IOC_DBG_TRAP_GET_DEVICE_SNAPSHOT:
+          dst = reinterpret_cast<void *>(saved_dbg_snapshot_ptr);
+          copy_len = extra;
```
kuhar: We should avoid copying a `GET_DEVICE_SNAPSHOT` payload back when the ioctl failed, and clamp any successful copy to the original `num_devices * entry_size` capacity. This branch currently copies `extra` bytes regardless of `resp.result`; in a current-build repro, both paths returned `ENOSYS`, but local mode preserved an 8-byte sentinel buffer while daemon mode zeroed it. An unsupported operation should not mutate caller memory or dereference the saved output pointer.

emulation/mirage/rocjitsu_sys/src/lib.rs
```diff
@@ -77,6 +77,8 @@ pub struct RjVmCmd {
     pub result: i32,
     /// `[out]` Backing handle for shareable allocations, or -1.
     pub shared_handle: RjHandle,
+    /// `[in]` Client-provided fd (e.g. debugger notifier), or -1.
+    pub in_handle: RjHandle,
```
kuhar, line 83: We should fix both the stale ABI assertion and the compatibility of this public FFI struct. `cargo test -p rocjitsu_sys struct_sizes_match_c_abi` now fails (40 bytes versus 32), and `rocjitsu_sys` runtime-loads `librocjitsu`: an older 32-byte caller can be paired with the new library, which reads `cmd->in_handle` past its object during daemon `DBG_TRAP`. Please update the layout test and pass this fd through a size-tagged or versioned contract rather than silently extending `rj_vm_cmd_t`.

emulation/mirage/rocjitsu/src/daemon.rs
```diff
@@ -560,6 +574,64 @@ fn recv_exact(fd: RawFd, buf: &mut [u8]) -> bool {
...
+    let mut msg: libc::msghdr = unsafe { std::mem::zeroed() };
+    let cmsg_space = unsafe { libc::CMSG_SPACE(std::mem::size_of::<RawFd>() as u32) } as usize;
+    let mut cmsg_buf = vec![0u8; cmsg_space];
+    msg.msg_iov = &mut iov;
+    msg.msg_iovlen = 1;
+    msg.msg_control = cmsg_buf.as_mut_ptr() as *mut c_void;
+    msg.msg_controllen = cmsg_space as _;
+
+    let n = loop {
+        let r = unsafe { libc::recvmsg(fd, &mut msg, 0) };
```
kuhar: This would be safer if the Rust receiver used `MSG_CMSG_CLOEXEC` for `recvmsg`, matching `rpc_recv_msg()` in the C++ daemon. As written, the transferred notifier is inheritable until explicitly closed and can leak through a later `exec`. Please add the flag and assert `FD_CLOEXEC` in the SCM_RIGHTS test.

kuhar (commented):
A few comments

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
kuhar: We should preserve `EBADF` when SCM_RIGHTS rejects a closed positive notifier fd. `rpc_send_msg()` returns `-1` with `errno == EBADF`, but `RemoteDriver::send_ioctl()` forwards plain `-1`, which the interposer reports as `EPERM`. The current-build repro returns `EBADF` locally and `EPERM` in daemon mode. Please translate the transport failure or validate the fd before sending, and cover this case in the daemon test.

emulation/mirage/rocjitsu_sys/src/lib.rs
amd-arosa: i plan to add a rj_daemon_start and rj_daemon_stopfunction

atgutier (commented):
The headline: the PR's framing overstates what landed

  The expanded PR appears to wire up cross-process attach, but it does not. What actually landed is
  (a) the SCM_RIGHTS transport for the debugger notifier fd, and (b) self-debug routed through the
  daemon dispatch path. The cross-process authorization gate is unchanged in substance — the first
  cross-process ENABLE is still structurally rejected with -EPERM. So this is a solid incremental
  step, but the description implying cross-process attach now works is inaccurate. That mismatch
  drives the top finding.

  Prior findings — status

  Prior finding: #2 runtime_mutex_ torn view in ENABLE
  Status: ✅ Resolved — marshal now snapshots enabled/r_debug/mode_mask under one critical section
  ────────────────────────────────────────
  Prior finding: #3 default: -ENOSYS vs kernel EINVAL
  Status: ✅ Resolved — out-of-range ops now -EINVAL; -ENOSYS reserved for
  recognized-but-unimplemented
  ────────────────────────────────────────
  Prior finding: #4 test gaps
  Status: 🟡 Partial — now covers bare-disable, 8-thread concurrency, transport round-trip (real
    socketpair + sentinel through the transferred fd), foreign-client EPERM, rinfo over/undersize.
    Still missing: a positive cross-process attach test (impossible today), daemon-mode concurrent
    ENABLE/DISABLE with fd adoption, fork staleness
  ────────────────────────────────────────
  Prior finding: #1 cross-process ENABLE impossible
  Status: ⚠️ Not resolved (by design) — still EPERMs first attach; correctly documented in-code as
    deferred to #8364, but the PR description doesn't say so
  ────────────────────────────────────────
  Prior finding: #5 fork client_pid_ staleness
  Status: ⚠️ Not resolved (documented, latent) — masked by interposer re-open

  BLOCKER

  Stale ABI-pin test will fail cargo test — mirage/rocjitsu_sys/src/lib.rs:489. The PR adds in_handle:
   RjHandle to RjVmCmd, growing it 32→40 bytes (field-for-field matching the new rj_handle_t in_handle
   in rj_vm_cmd_t), but the ABI-pinning assertion still says assert_eq!(size_of::<RjVmCmd>(), 32).
  This is exactly the test meant to catch ABI drift; it now fails. Fix: bump to 40 and update the
  stale layout comment. (The FFI itself is correct — both sides are 40 bytes and identical.)

 MAJOR

  1. Cross-process authorization substitute is absent — simulated_kfd.cpp:1759. The current gate only
  re-admits an already-attached debugger; there's no positive authorization for a first attach. The
  code is safe today only because the debugger_pid == 0 structural block makes first attach
  unreachable — a fragile property. When #8364 flips this on, the only remaining check is "does a
  KfdProcess with client_pid == args->pid exist" (find_process_by_client_pid scans the whole table
  with no relationship check), so any daemon client could attach as debugger to any other client's
  process. The kernel gates this on ptrace parentage. Before graduating cross-process, land an
  explicit substitute (SO_PEERCRED-based ptrace-permitted check on the daemon side) and document it as
   the security model — or if the daemon fully trusts its clients, state that explicitly.
  2. recvmsg missing MSG_CMSG_CLOEXEC — mirage/rocjitsu/src/daemon.rs:591. The received notifier fd
  lacks FD_CLOEXEC, leaking the debugger's pipe across any fork/exec in the daemon. Diverges from the
  C rpc_recv_msg (which uses MSG_CMSG_CLOEXEC) and from the CLOEXEC convention everywhere else in this
   file. Fix: pass libc::MSG_CMSG_CLOEXEC.

  MINOR

  - caller.client_pid() read without synchronization — simulated_kfd.cpp:1735/1760/1814; client_pid_
  is non-atomic, written under process_mutex_, read lock-free in the dispatcher. No real race today
  (set-once before discoverable) but TSan-flaggable. Make it std::atomic<pid_t> or document the
  invariant.
  - Daemon fcntl(dbg_fd, F_GETFD) EBADF path is implicit across 3 files — rj_vm.cpp:256 + dispatcher
  :1806. The missing-transfer→EBADF guarantee relies on the client sending exactly KFD_INVALID_FD; a
  bogus small fd number with no SCM_RIGHTS transfer would validate an unrelated daemon fd and wrongly
  adopt it. Don't trust the payload dbg_fd in daemon mode — key EBADF on the absence of an adopted
  in_handle.
  - Duplicated DBG_TRAP embedded-pointer/inline-size ladder — the op→field mapping (ENABLE→rinfo_ptr,
  GET_DEVICE_SNAPSHOT→snapshot_buf_ptr) is spelled out ~4 times across remote_driver.cpp and
  rj_vm.cpp; they must stay consistent or the round-trip silently corrupts. Factor into one shared
  helper.
  - Single-fd SCM_RIGHTS extraction without overflow-close — daemon.rs:604; keeps only the last fd per
   cmsg and doesn't close surplus (the C ref counts and closes extras). Limited by the one-fd-sized
  control buffer (MSG_CTRUNC closes the rest), so unlikely to leak in the expected case, but not
  robust.
  - No MSG_CTRUNC handling/logging — daemon.rs:591; a truncated ancillary payload silently drops the
  notifier fd (ENABLE loses it) with no trace. At least tracing::warn!.

  NIT

  No test asserts -EINVAL for a genuinely out-of-range op (all negative op tests use recognized ops);
  DebugSession doxygen in kfd_process.h cites AMD-internal kernel field names + a driver source path
  rather than the public UAPI contract (confidentiality-hygiene); main.cpp fd-ownership "safety net"
  is correct but the one-owner invariant is spread across 3 points — a one-line comment would prevent
  a double-close regression; close() return ignored on fd reclaim (matches surrounding style).

emulation/mirage/rocjitsu/src/daemon.rs
kuhar: It would be useful to add a focused Rust test for the close-on-exec contract. `recv_header_with_fd()` now passes `MSG_CMSG_CLOEXEC`, but the Rust daemon tests do not exercise SCM_RIGHTS receipt; only the C++ path asserts `FD_CLOEXEC` (`kfd_ioctl_test.cpp:877-880`). A socketpair/sendmsg unit test around `recv_header_with_fd()` could verify that the returned descriptor has `FD_CLOEXEC` set.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/remote_driver.cpp
kuhar: We should also gate the `ENABLE` copy-back on `resp->result == 0`, as this thread now does for `GET_DEVICE_SNAPSHOT`. On the current head, a daemon stub that returns `-EBADF` leaves the normal zero-filled inline payload; `RemoteDriver` returns `-EBADF` but still overwrites a 16-byte sentinel `rinfo_ptr`, while the local path leaves it untouched. Please add the same success check for `ENABLE` and a daemon-path bad-fd regression.

emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/vm/rj_vm.h
```diff
@@ -228,6 +234,27 @@ RJ_API_EXPORT rj_status_t rj_vm_execute(rj_vm_t *vm, rj_vm_cmd_t *cmd);
...
+/// ENABLE) alongside @p cmd instead of embedding it in the descriptor. Passing
+/// the fd through the call signature — rather than growing rj_vm_cmd_t — keeps
+/// the descriptor ABI frozen so a caller built against an older header can never
+/// have this fd read past the end of its shorter object. The symbol's presence
+/// is the capability tag: a caller that runtime-loads the library resolves it
+/// optionally and falls back to rj_vm_execute_as (no fd transfer) when absent.
+/// @param[in] vm VM handle.
+/// @param[in] process_id The target process ID.
+/// @param[in,out] cmd Command descriptor.
+/// @param[in,out] in_handle Points to the fd to transfer (or -1). On adoption
```
atgutier: If an invalid FD is allowed here (-1) then why do we need the extended version?

If we need this new function, can you give it a more descriptive name than "extended".

Also consider just adding to the rj_vm_cmd_t if this is general enough. We're still in beta so I'd rather get the ABI/API right and have it be nice and clean for 1.0.0 release vs. trying to keep back compat while we're still in beta.

atgutier (commented):
Mostly looks good. Other than my API/ABI ask here is an issue highlighted by an agent:

Fix A — daemon DBG_TRAP ENABLE must not adopt an untransferred fd

  File: emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/rj_vm.cpp
  Function: execute_impl

  The bug: when a client sends DBG_TRAP ENABLE in daemon mode with no SCM_RIGHTS fd, the substitution
  block is skipped, so dbg->enable.dbg_fd keeps the raw client-supplied number. The dispatcher then
  runs fcntl(dbg_fd, F_GETFD) on that number against the daemon's fd table — if it happens to be a
  live daemon fd (socket, memfd), the check passes and the daemon adopts and later closes its own fd.

  Find this block:
```    
bool adopting_notifier = false;
    if (driver->daemon_mode() && cmd->cmd == AMDKFD_IOC_DBG_TRAP && in_handle && *in_handle >= 0) {
      auto *dbg = static_cast<kfd_ioctl_dbg_trap_args *>(cmd->buf);
      if (dbg->op == KFD_IOC_DBG_TRAP_ENABLE) {
        dbg->enable.dbg_fd = static_cast<uint32_t>(*in_handle);
        adopting_notifier = true;
      }
    }
```

  Replace with:
```
    bool adopting_notifier = false;
    if (driver->daemon_mode() && cmd->cmd == AMDKFD_IOC_DBG_TRAP) {
      auto *dbg = static_cast<kfd_ioctl_dbg_trap_args *>(cmd->buf);
      if (dbg->op == KFD_IOC_DBG_TRAP_ENABLE) {
        if (in_handle && *in_handle >= 0) {
          // A notifier fd was transferred over SCM_RIGHTS (already in the daemon's
          // fd space); substitute it for the client-side number in the payload.
          dbg->enable.dbg_fd = static_cast<uint32_t>(*in_handle);
          adopting_notifier = true;
        } else {
          // No fd was transferred. The client-supplied dbg_fd is a number in the
          // client's fd table and must never be validated/adopted against the
          // daemon's own fds. Force it invalid so the dispatcher's
          // fcntl(dbg_fd, F_GETFD) check returns -EBADF.
          dbg->enable.dbg_fd = static_cast<uint32_t>(-1);
        }
      }
    }
```

  Why this is complete: the dispatcher (debug_trap_ioctl in simulated_kfd.cpp, case
  KFD_IOC_DBG_TRAP_ENABLE) already does if (fcntl(dbg_fd, F_GETFD) == -1) return -EBADF;. Forcing
  dbg_fd = -1 when nothing was transferred makes that existing check fail cleanly — no dispatcher
  change needed. The whole block is guarded by driver->daemon_mode(), and local mode never routes
  through execute_impl, so local debugging is untouched.
 Regression test (add to tests/kfd_ioctl_test.cpp): daemon-mode rj_vm_execute_as_ex with a DBG_TRAP
  ENABLE whose args->enable.dbg_fd is a valid daemon fd number (e.g. dup() of any open fd) and
  in_handle == nullptr; assert the result is -EBADF and that the dup'd fd is still open afterward.
  (The existing EnableWithoutTransferredFdReturnsEbadf only passes dbg_fd = -1, which fails fcntl by
  coincidence rather than by this guard.)

emulation/mirage/rocjitsu/src/daemon.rs
amd-arosa: we will be removing this rust code in next pr.

amd-arosa:
> Mostly looks good. Other than my API/ABI ask here is an issue highlighted by an agent:
> 
> Fix A — daemon DBG_TRAP ENABLE must not adopt an untransferred fd
> 
> File: emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/rj_vm.cpp Function: execute_impl
> 
> The bug: when a client sends DBG_TRAP ENABLE in daemon mode with no SCM_RIGHTS fd, the substitution block is skipped, so dbg->enable.dbg_fd keeps the raw client-supplied number. The dispatcher then runs fcntl(dbg_fd, F_GETFD) on that number against the daemon's fd table — if it happens to be a live daemon fd (socket, memfd), the check passes and the daemon adopts and later closes its own fd.
> 
> Find this block:
> 
> ```
> bool adopting_notifier = false;
>     if (driver->daemon_mode() && cmd->cmd == AMDKFD_IOC_DBG_TRAP && in_handle && *in_handle >= 0) {
>       auto *dbg = static_cast<kfd_ioctl_dbg_trap_args *>(cmd->buf);
>       if (dbg->op == KFD_IOC_DBG_TRAP_ENABLE) {
>         dbg->enable.dbg_fd = static_cast<uint32_t>(*in_handle);
>         adopting_notifier = true;
>       }
>     }
> ```
> 
> Replace with:
> 
> ```
>     bool adopting_notifier = false;
>     if (driver->daemon_mode() && cmd->cmd == AMDKFD_IOC_DBG_TRAP) {
>       auto *dbg = static_cast<kfd_ioctl_dbg_trap_args *>(cmd->buf);
>       if (dbg->op == KFD_IOC_DBG_TRAP_ENABLE) {
>         if (in_handle && *in_handle >= 0) {
>           // A notifier fd was transferred over SCM_RIGHTS (already in the daemon's
>           // fd space); substitute it for the client-side number in the payload.
>           dbg->enable.dbg_fd = static_cast<uint32_t>(*in_handle);
>           adopting_notifier = true;
>         } else {
>           // No fd was transferred. The client-supplied dbg_fd is a number in the
>           // client's fd table and must never be validated/adopted against the
>           // daemon's own fds. Force it invalid so the dispatcher's
>           // fcntl(dbg_fd, F_GETFD) check returns -EBADF.
>           dbg->enable.dbg_fd = static_cast<uint32_t>(-1);
>         }
>       }
>     }
> ```
> 
> Why this is complete: the dispatcher (debug_trap_ioctl in simulated_kfd.cpp, case KFD_IOC_DBG_TRAP_ENABLE) already does if (fcntl(dbg_fd, F_GETFD) == -1) return -EBADF;. Forcing dbg_fd = -1 when nothing was transferred makes that existing check fail cleanly — no dispatcher change needed. The whole block is guarded by driver->daemon_mode(), and local mode never routes through execute_impl, so local debugging is untouched. Regression test (add to tests/kfd_ioctl_test.cpp): daemon-mode rj_vm_execute_as_ex with a DBG_TRAP ENABLE whose args->enable.dbg_fd is a valid daemon fd number (e.g. dup() of any open fd) and in_handle == nullptr; assert the result is -EBADF and that the dup'd fd is still open afterward. (The existing EnableWithoutTransferredFdReturnsEbadf only passes dbg_fd = -1, which fails fcntl by coincidence rather than by this guard.)

that was already fixed in b5e8caf25b6c3a18696f498ed5f1ca85f953d9d4

emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/vm/rj_vm.h
amd-arosa: fixing.

amd-arosa: fixed in 9c5d55b6e63558bbc9692bae247deb05d2ac2136

amd-arosa (commented):
lgtm

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/remote_driver.cpp
amd-arosa: fixed in 442c1efa9e95a7b2ee9e7d8408bc6a35825d1cb0

atgutier (approved):

PR #8394, opened 2026-07-10
refactor(rocr): Unify SDMA submission API

saleelk:
- Restructure SubmitBodies: both fused (gfx125+) and classic paths now use a single ring reservation for all entries (one doorbell per engine group). Classic path assumes platform atomics and decrements out_signal directly per entry—no per-body signals needed.

- Simplify SubmitPrologue/SubmitEpilogue: remove atomic-decrement branches where the signal value is known (prologue signal is always 1, epilogue polls then fences to 0). Remove body_signals parameter and drain_bodies_individually logic from the epilogue.

- Eliminate heap allocation in classic body path: copy packets are formed in-place in the ring buffer instead of staging through a temporary std::vector + memcpy.

- Consolidate linearB2B paths into DmaCopyFanOutOp with coord_engine and max_engines parameters. Drop engine_mask; existing topology resolution handles engine selection, capped by max_engines.

- Add SubmitFusedCoordinator for gfx125+: packs prologue + body entries + epilogue into a single ring reservation (one doorbell) for the coordinator engine, reducing doorbell overhead in multi-engine fan-out submissions.

- Remove dead code from SubmitFusedCoordinator: strip pre-gfx125+ fallback branches (split-word polls, 32-bit fences, atomic decrements) that are unreachable since the function is exclusively called on gfx125+ instances. Remove vestigial epilogue_body_signals parameter.

- Gather SubmitBodies entries from master lists: take the caller's master dst/src/size arrays plus an index vector and gather each engine's entries internally, instead of receiving pre-built per-engine vectors. Removes four std::vector allocations per engine group on every fan-out submission. Collapse sizes_a/sizes_b into a single size_list.

- **Rotate SDMA engines within a single copy, not across API calls:** remove the persistent per-agent round-robin counter (`sdma_rr_index_`) and `PickSdmaEngine`. Engine assignment now uses `NthSdmaEngine(mask, k)` driven by the per-entry index, so a fan-out spreads across engines within the copy while the engine set stays deterministic per call and no longer marches across successive copies.

- **gfx125+ fused-coordinator epilogue fast-path:** when the completion signal has no event mailbox and GCR/profiling are off, skip the epilogue poll+fence entirely. Every body packet's SIGNAL stage already atomically subtracts 1 after its copy (SDMA is in-order per engine), so the last chunk's sub is the ordered completion edge on its own; the signal target is shifted down by one so that final sub lands on the done value. Collapses the coordinator submission to the same single-body packet the worker engines emit.

- **Revert gfx1250 multicast crossover 8MB → 256KB:** a single-engine multicast serialises the writes to all destinations and becomes bandwidth-bound well before 8MB, so multi-destination copies above 256KB now fan out across engines for higher aggregate bandwidth. Fixes a large-copy regression in broadcast copies. The `HSA_SDMA_MULTICAST` override is unchanged.

- Tune linearB2B serialization threshold: lower kB2BMaxSize from 256KB to 64KB to eliminate latency blips at ~2MB sizes in multi-entry batch copies.

- **Serialize the ReleaseWrite debug log:** move it into `UpdateWriteAndDoorbellRegister` under the doorbell serialization and log the written wptr plus the live rptr (instead of the old→new wptr), so it can't report a wptr ordering that never happened under concurrent submission.

## Motivation

Reduce SDMA submission overhead (doorbell rings, heap allocations, branch complexity) for both single-engine and multi-engine copy paths. The fused coordinator path on gfx125+ saves 1-2 doorbells per fan-out submission by coalescing prologue, body copies, and epilogue into one contiguous ring write. The multicast crossover revert and within-copy engine rotation restore multi-destination bandwidth at large sizes.

## Technical Details

- `SubmitFusedCoordinator`: single AcquireWriteAddress/ReleaseWriteAddress pair emits all coordinator packets; the epilogue is dropped on the fast path when no interrupt mailbox / GCR / profiling is needed.
- `DmaCopyFanOutOp`: `if (fused)` / `else` branches -- fused path calls SubmitFusedCoordinator for the coordinator group and SubmitBodies for others; classic path retains separate SubmitPrologue/SubmitBodies/SubmitEpilogue calls.
- Engine selection: `NthSdmaEngine(mask, k)` replaces the global round-robin `PickSdmaEngine`; `sdma_rr_index_` removed.
- `kMulticastMaxSize` restored to 256 KB (gfx1250); copies above it fan out across engines.

## JIRA ID

JIRA ID: AIRUNTIME-0

## Test Plan

- MPI-based allgather batch test (`hip-allgather-batch`, 8 GPUs, 23 size tiers 1KB-4GB) on gfx942 (MI300X) -- correctness and performance.
- gfx125+ (MI450) validation still required for the fused-coordinator epilogue fast-path, within-copy rotation, and the multicast crossover revert (these paths are gfx125+-only and cannot be exercised on gfx942).

## Test Result

All 23 size tiers pass cleanly on gfx942 (MI300X), no hangs or errors. Smooth latency curve with no anomalous blips at the 2MB boundary. gfx125+ validation pending.

## Submission Checklist

- Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_blit_sdma.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_blit_sdma.cpp
@@ -1289,18 +1289,17 @@ hsa_status_t BlitSdma<useGCR, scopeFields>::SubmitFusedCoordinator(
 template <bool useGCR, bool scopeFields>
 hsa_status_t BlitSdma<useGCR, scopeFields>::SubmitBodies(
     hsa_amd_memory_copy_op_type_t op,
-    const std::vector<void*>& dsts,
-    const std::vector<const void*>& srcs,
-    const std::vector<size_t>& sizes_a,
-    const std::vector<size_t>& sizes_b,
+    void* const* dst_list,
+    const void* const* src_list,
+    const size_t* size_list,
+    const std::vector<uint32_t>& indices,
     bool indirect_src, bool indirect_dst,
     const std::vector<core::Signal*>& dep_signals,
     core::Signal& out_signal,
     core::Signal* body_signal) {
 
-  const size_t num_entries = dsts.size();
-  if (num_entries == 0 || srcs.size() != num_entries || sizes_a.size() != num_entries)
-    return HSA_STATUS_ERROR_INVALID_ARGUMENT;
+  const size_t num_entries = indices.size();
+  if (num_entries == 0) return HSA_STATUS_ERROR_INVALID_ARGUMENT;
 
   const bool is_swap = (op == HSA_AMD_MEMORY_COPY_OP_LINEAR_SWAP);
   const bool is_indirect = (indirect_src || indirect_dst);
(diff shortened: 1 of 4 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_blit_sdma.cpp
dayatsin-amd, line 899: While I expect the fence command to also work when platform atomic is supported, I wonder if this will cause some performance regression on platform where atomics are supported. I would expect the atomic operation to be generally faster.

saleelk: let me add the support, i didnt make those change yet since we talked last time

PR #8395, opened 2026-07-10
feat(rocprofiler-compute): Ingest and attribute code-object ISA in analysis DB

vedithal-amd:
## Motivation

Add the disassembly emitted by the native PC sampling collector to the analysis database, including instructions that were never PC-sampled, and attribute every line to its kernel. This gives the DB complete ISA context per kernel alongside the sampled counts landed in the normalized schema PR. Stacked on #8334; no schema change and no user-visible output change.

## Technical Details

- New `src/pc_sampling/code_object_analysis.py`: a pure parser (code-object disassembly dict -> per-object instruction records) plus a thin discovery loader that finds the per-process disassembly artifacts under a workload.
- Reconcile disassembly instructions to the offset the PC-sampling rows use, measured from the code object's load address: `code_object_offset = virtual_address - load_base` (the disassembly's own offset is into the ELF file, a different axis).
- Attribute each line to its kernel by joining `code_obj_info` symbols to `kernel_symbols` on the mangled ELF name (strip `.kd`), then mapping to the demangled name the kernel table is keyed by.
- Store ISA only for dispatched kernels: a never-dispatched symbol has no kernel and no possible sample, so its lines and its code-object store are dropped rather than stored un-attributed.
- Scope ingestion to the surviving `ps_file_results.json` pid: `load_base` is known only for that pid and `code_object_id` collides across pids, so a non-surviving pid's ISA would otherwise be offset with the wrong `load_base`.
- Skip any offset already inserted by the sampled path so no duplicate instruction lines are added.
- Code objects with no known `load_base` are skipped (the offset cannot be derived).
- Relocate `pc_sampling_analysis.py` from `src/utils` to the `src/pc_sampling` package (mechanical, no behavior change).
- Reconcile the ENABLE_COVERAGE test list, which was missing several registered Python tests.

## JIRA ID

JIRA ID: AIPROFCOMP-540

## Test Plan

- Parser unit tests against a sample workload's disassembly artifact (per-object grouping, symbol-instruction flattening, field capture) and discovery-loader tests (pid parse, malformed JSON, missing pid prefix).
- DB unit tests: ISA of a dispatched kernel is attributed via the mangled-name join, an already-sampled offset inserts no duplicate row, a dispatched-but-unsampled code object gets a store, un-dispatched ISA is dropped, a non-surviving pid's ISA is skipped, and skip-on-null-load_base.
- End-to-end analyze --output-format db on the PC-sampling workload, asserting every line is kernel-attributed, sampled lines carry a sample state, and no (code object, offset) is duplicated.

## Test Result

Python unit tests pass; ruff clean.

## Submission Checklist

- [ ] Look over the contributing guidelines


projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
```diff
@@ -400,52 +403,111 @@ def add_pc_sampling_data(
...
-            kernel=kernel_objs.get(line.kernel_name),
-        )
-        Database.get_session().add(instruction_line)
+        """Add the full code-object disassembly as instruction lines,
+        skipping any offset already present."""
+        tool_data = self._pc_sampling_tool_data_per_workload.get(workload_path)
+        load_base_by_id = {
+            code_object["code_object_id"]: code_object.get("load_base")
+            for code_object in (tool_data or {}).get("code_objects", [])
+        }
```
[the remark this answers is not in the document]
vedithal-amd: Real hazard, but it is a symptom of the known multi-process results-file clobbering (out of scope here): load_base exists only for the single surviving ps_file pid, so non-survivor pids can pick up the wrong load_base by code_object_id collision. This only affects ISA-only rows of already-broken pids. Documented as a known limitation at the load_base_by_id map in 14a7712; whoever fixes the clobbering fixes this mapping in the same change.

projects/rocprofiler-compute/src/pc_sampling/code_object_analysis.py
```diff
@@ -0,0 +1,91 @@
...
+    for json_path in sorted(Path(workload_path).glob(CODE_OBJ_INFO_GLOB)):
+        pid = _parse_pid(json_path)
+        if pid is None:
+            continue
+        try:
+            data = json.loads(json_path.read_text(encoding="utf-8"))
+        except json.JSONDecodeError as error:
+            console_warning(f"Code object info: failed to parse {json_path}: {error}")
+            continue
+        disassemblies_per_pid[pid] = parse_code_object_info(data)
```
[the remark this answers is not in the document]
vedithal-amd: Declining: <pid>_code_obj_info.json is written by our own native collector with a fixed schema, an internal artifact rather than an external boundary. A schema mismatch would mean the collector is broken, which we want to fail loudly on, not silently skip. Real-world corruption (a partial write) surfaces as JSONDecodeError, which is already caught.

abchoudh-amd (approved):
LGTM.

Note: As discussed offline, we should do away with using workloads for testing.

projects/rocprofiler-compute/src/pc_sampling/code_object_analysis.py
```diff
@@ -0,0 +1,95 @@
...
+                _to_instruction(instruction, symbol.get("name"))
+                for symbol in code_object.get("symbols", [])
+                for instruction in symbol.get("instructions", [])
+            ],
+        )
+        for code_object in data.get("code_objects", [])
+    ]
+
+
+def load_code_object_disassemblies(
```
svolkov-amd, line 55: Like a lot that the parsing code is separated into its own files.
But would suggest to rename this file to `code_object_data_reader.py` to better reflect what it specific "action" it does.
Also would suggest to organize it as a class and segregate functions on public (it seems there will be a single on - `load_code_object_disassemblies`) and internal implementation details (aka other functions a private, meaning with `__` prefix).

svolkov-amd: Same comments for `pc_sampling_analysis.py`

projects/rocprofiler-compute/src/pc_sampling/code_object_analysis.py
```diff
@@ -0,0 +1,95 @@
...
+CODE_OBJ_INFO_GLOB = "**/*_code_obj_info.json"
+
+
+class CodeObjectInstruction(NamedTuple):
+    """One disassembled instruction within a code object."""
+
+    virtual_address: int
+    instruction: str
+    comment: str
+    kernel_name: str
```
svolkov-amd, line 26: This is not to be fixed in this PR. But just to emphasize a similar comment in a previous PR. It's not efficient in terms of memory, output size and database read performance to attribute every single ISA code line to kernel. Instead we need to use an intermediate between a kernel and ISA lines, which we should call `compute_kernel_symbol`. Json from native collector already does this breakdown.

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
```diff
@@ -405,57 +408,144 @@ def add_pc_sampling_data(
...
+        # The disassembly carries the mangled ELF symbol; kernel_objs is keyed by
+        # the demangled name. kernel_symbols bridges the two per code object.
+        kernel_by_symbol = {
+            (
+                symbol["code_object_id"],
+                symbol["kernel_name"].removesuffix(".kd"),
```
svolkov-amd, line 474: Don't understand this part. `symbol["kernel_name"].removesuffix(".kd")` and `symbol["formatted_kernel_name"]` seem to be equal if judge by ps_file_result.json included in the test. So why can't we simply use `symbol["formatted_kernel_name"]` here and below and reduce complexity of this code?

svolkov-amd: Note: this is for further discussion with Vignesh. PR can be merged with it.

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
```diff
@@ -405,57 +408,144 @@ def add_pc_sampling_data(
...
-        kernel = kernel_objs.get(line.kernel_name)
-        if kernel is None:
-            # Drop lines whose kernel was filtered out or never mapped.
-            return
+        """Add dispatched kernels' disassembly as instruction lines,
+        skipping any offset already present."""
+        tool_data = self._pc_sampling_tool_data_per_workload.get(workload_path)
+        # load_base is known only for the surviving ps_file pid, and
+        # code_object_id collides across pids, so scope ISA to that pid.
+        surviving_pid = (tool_data or {}).get("metadata", {}).get("pid")
```
svolkov-amd, line 464: Don't understand this part - what is a surviving_pid?
Also `(tool_data or {})` is an awful way to do error handling. If tool_data is None then exception shall be thrown.

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
```diff
@@ -405,57 +408,144 @@ def add_pc_sampling_data(
...
+                        orm.InstructionSample(
+                            pc_sample_state=sample_state,
+                            instruction_sample_lookup=Database.get_or_create_type(
+                                orm.InstructionSampleLookup, text
+                            ),
+                            count=count,
+                        )
+                    )
+
+    def add_code_object_isa(
```
svolkov-amd, line 453: Too large function, needs to be broken down into smaller well-named functions which allow to understand the high-level execution flow at a glance.

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
```diff
@@ -405,57 +408,144 @@ def add_pc_sampling_data(
...
+        # the demangled name. kernel_symbols bridges the two per code object.
+        kernel_by_symbol = {
+            (
+                symbol["code_object_id"],
+                symbol["kernel_name"].removesuffix(".kd"),
+            ): kernel
+            for symbol in (tool_data or {}).get("kernel_symbols", [])
+            if (kernel := kernel_objs.get(symbol["formatted_kernel_name"])) is not None
+        }
+        # A code object is worth storing only if one of its symbols was dispatched.
```
svolkov-amd, line 479: Comment is misattributed - the check for invocation and storage of code object is not done here. The better way is to wrap corresponding code into a well-named function, aka `is_code_object_invoked`.

projects/rocprofiler-compute/tests/test_analysis_db.py
```diff
@@ -937,3 +950,219 @@ def test_add_pc_sampling_data_drops_lines_without_kernel(db_session):
...
+def test_add_code_object_isa_creates_store_for_unsampled_code_object(db_session):
+    """A dispatched-but-never-sampled code object present only in code_obj_info
+    gets a new store, using the load_base from the results' code_objects list,
+    and its ISA is attributed to the dispatched kernel."""
+    workload_path = common.get_output_dir()
+    Path(workload_path).mkdir(parents=True, exist_ok=True)
+    try:
+        tool_data = make_pc_sampling_tool_data()
+        # code object 9 has a load_base and a dispatched kernel, but no samples.
+        tool_data["code_objects"].append({"code_object_id": 9, "load_base": 0x2000})
```
svolkov-amd, line 1084: Too much duplication in tests. And test quality matters. Better to organize them as class and move the common part into test class members and into setup/teardown functions.

svolkov-amd (approved):
Left some comments related to the code quality. But functionally this PR looks correct. Therefore it could be merged and I would ask Vignesh to review comments when he's back.

projects/rocprofiler-compute/tests/test_analysis_db.py
abchoudh-amd: Resolving to unblock merge.

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
abchoudh-amd: Resolving to unblock merge.

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
abchoudh-amd: Resolving to unblock merge.

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
abchoudh-amd: Resolving to unblock merge.

projects/rocprofiler-compute/src/pc_sampling/code_object_analysis.py
abchoudh-amd: Resolving to unblock merge.

projects/rocprofiler-compute/src/pc_sampling/code_object_analysis.py
abchoudh-amd: Resolving to unblock merge.

projects/rocprofiler-compute/src/pc_sampling/code_object_analysis.py
vedithal-amd: @svolkov-amd We can add a table for kernel symbol (extracted from code object info json). It will relate to a code object (one code object can have many symbols) and it will relate to an isa (one symbol can have many ISA) and it will relate to a kernel (one symbol has one kernel). This way we can map code objects and isa to kernels.

But the problem is this will not give you the memory and compute efficiency you are seeking since you would still have to maintain symbol foreign key per ISA which is similar to the current approach of maintain a kernel foreign key per ISA.

In the current approach we abstract away the symbol information since I don't think it is of much value as compared to relating the kernel directly with ISA - in terms of overall pc analysis. We can add the symbol redirection since that is theoretically correct but i dont see the practical benefit as compared to current approach

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
vedithal-amd: @svolkov-amd  you mean the complexity of `removesuffix`? Last time I checked `symbol_name` actually has the .kd suffix

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
vedithal-amd: Currently the ps_file_result.json is cloberred among pids - so we have to be sure we only use code_object_info json of the pid which actually survived the cloberring. Once the cloberring stops this code will be improved.

As you know we need to correlate code object info json with ps file result json to ensure we are only adding ISA for symbols which were actually executed so that they can provide context for sampled instructions. There is no point adding ISA for symbols which were never actually sampled since they wont provide any benefit in terms of pc sampling analysis

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
vedithal-amd: Agreed

projects/rocprofiler-compute/tests/test_analysis_db.py
vedithal-amd: Agreed, we can extract out some module level helpers and organize test functions into classes per their responsibility

PR #8399, opened 2026-07-10
fix(rocprof-sys-python): restore -a annotations dropped by cache replay

marantic-amd:
## Motivation

- `rocprof-sys-python --help` advertises `-a`/`--annotate-trace` ("Add detailed annotations to the trace"), but passing it silently produced a Perfetto trace that was byte-for-byte equivalent to a run without it. No error, no warning, just a documented flag doing nothing.

## Technical Details

- The Perfetto trace is no longer written by the live instrumentation path; it's rebuilt during finalization from cached region_sample` records (`cache_manager::post_process_bulk`). The `-a` flag passes a `rocprofsys_annotation_t*` array into `rocprofsys_push_category_region`, and `regions.cpp`'s `invoke_category_region_start` turned that array into a Perfetto `EventContext` lambda for the *live* path only, hence, it was never serialized into the trace-cache args wire format, so cache replay reconstructed each region with empty args regardless of `-a`.

- PR #7618 fixed the equivalent problem for gotcha args (e.g. `pthread_create`) by routing them through `append_cache_args`, but left the annotation-array path which is used by `rocprof-sys-python -a`, and any other C-API caller passing annotations.

- This PR adds a small converter in `regions.cpp` that turns the annotation array into the same `function_args_t` wire format and routes it through the existing `append_cache_args` channel, so `perfetto_processor_t::handle(region_sample)` and `rocpd_processor_t::handle(region_sample)`, which already re-emit `args_str`, pick it up on replay. 

## JIRA ID

Resolves AIPROFSYST-579

## Test Plan

- Strengthened `tests/pytest/test_python.py::TestPython::test_builtin` so the `annotated=True`/`False` parametrization actually asserts something: the six per-function debug keys (`file`, `line`, `lasti`, `argcount`, `nlocals`, `stacksize`) are present with the expected count when `-a` is passed, and absent (count 0) otherwise — for both the Perfetto trace and `rocpd.db`.
- Added `tests/rocpd-validation-rules/python/python-builtin-annotated-rules.json` validating the same six annotations land in `rocpd.db`'s `region_args` table with correct types.
- Verified fail-before/pass-after: stashed the fix, reran the strengthened tests, confirmed they fail with "0 entries" where 895 were expected; restored the fix and confirmed they pass.
- Manually reproduced the original report's exact repro workload and inspected the resulting `.proto` with Perfetto's `trace_processor`.
- Ran the full `rocprof-sys-unit-tests` binary and the full `test_python.py`/`test_annotate.py` pytest suites to check for regressions.

## Test Result

- Manual repro: annotated run now has 8 distinct `debug.*` keys (`begin_ns`, `corr_id`, `file`, `line`, `lasti`, `argcount`, `nlocals`, `stacksize`) vs. 2 before the fix; non-annotated run unchanged at 2 keys.
- `rocprof-sys-unit-tests`: 1062/1062 passed (unchanged from baseline).
- `test_python.py`: 9/9 tests, 22/22 subtests passed.
- `test_annotate.py` (native-tool annotation path): unaffected, all passing.

## Before/After

- Before fix when annotation was set we would still get the output that looked like this:
image

- After applying the fix the annotations are clearly visible in the slice:
image

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-systems/source/lib/rocprof-sys/regions.cpp
+++ b/projects/rocprofiler-systems/source/lib/rocprof-sys/regions.cpp
@@ -25,20 +25,18 @@ namespace
 // must handle arbitrary caller types) these can be dispatched explicitly instead
 // of relying on a generic fmt::is_formattable/fmt::streamed fallback.
 template <typename Tp>
-std::string
+[[nodiscard]] std::string
 annotation_arg_type_name()
 {
     using value_type = std::decay_t<Tp>;
     if constexpr(concepts::is_string_type<value_type>::value)
         return "string";
-    else if constexpr(std::is_pointer<value_type>::value)
-        return "pointer";
     else
         return utility::demangle<value_type>();
 }
 
 template <typename Tp>
-std::string
+[[nodiscard]] std::string
 annotation_arg_value_string(Tp&& _val)
 {
     using value_type = std::decay_t<Tp>;
@@ -46,6 +44,8 @@ annotation_arg_value_string(Tp&& _val)
         return std::string{ std::string_view{ std::forward<Tp>(_val) } };
     else if constexpr(std::is_pointer<value_type>::value)
         return fmt::format("{:#x}", reinterpret_cast<std::uintptr_t>(_val));
+    else if constexpr(std::is_integral<value_type>::value)
+        return fmt::format_int{ _val }.str();
     else
         return fmt::format("{}", std::forward<Tp>(_val));
 }
@@ -64,13 +64,13 @@ append_annotation_arg(function_args_t& _args, const rocprofsys_annotation_t& _an
         using type = tracing::annotation_value_type_t<Idx>;
         if constexpr(std::is_pointer<type>::value)
         {
-            auto _value = reinterpret_cast<type>(_annotation.value);
+            const auto _value = reinterpret_cast<type>(_annotation.value);
             _args.push_back({ _arg_number, annotation_arg_type_name<type>(),
                               _annotation.name, annotation_arg_value_string(_value) });
         }
         else
         {
-            auto* _value = reinterpret_cast<type*>(_annotation.value);
+            const auto* _value = reinterpret_cast<type*>(_annotation.value);
             _args.push_back({ _arg_number, annotation_arg_type_name<type>(),
                               _annotation.name, annotation_arg_value_string(*_value) });
         }
@@ -86,7 +86,7 @@ append_annotation_arg(function_args_t& _args, const rocprofsys_annotation_t& _an
 // rocprofsys_push_category_region) into the trace-cache args wire format, so
 // annotations attached to a region survive cache-replay (Perfetto + rocpd)
 // instead of only appearing on the live-instrumentation path.
-function_args_t
+[[nodiscard]] function_args_t
 annotations_to_function_args(const rocprofsys_annotation_t* _annotations, size_t _count)
 {
     function_args_t _args;
@@ -96,8 +96,8 @@ annotations_to_function_args(const rocprofsys_annotation_t* _annotations, size_t
     for(size_t i = 0; i < _count; ++i)
     {
         const auto& _annotation = _annotations[i];
-        if(_annotation.name == nullptr || _annotation.type == ROCPROFSYS_VALUE_NONE ||
-           _annotation.value == nullptr)
+        if(_annotation.name == nullptr || _annotation.type <= ROCPROFSYS_VALUE_NONE ||
+           _annotation.type >= ROCPROFSYS_VALUE_LAST || _annotation.value == nullptr)
             continue;
 
         append_annotation_arg(
--- a/projects/rocprofiler-systems/source/lib/rocprof-sys/regions.cpp
+++ b/projects/rocprofiler-systems/source/lib/rocprof-sys/regions.cpp
@@ -6,6 +6,7 @@
 #include "core/config.hpp"
 #include "library/components/category_region.hpp"
 #include "library/tracing.hpp"
+#include <cstdint>
 
 #if defined(__GNUC__) && (__GNUC__ == 7)
 #    pragma GCC diagnostic push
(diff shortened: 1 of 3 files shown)


kcossett-amd:
> * PR [feat(rocprofiler-systems): Allow trace cache to handle gotcha args #7618](https://github.com/ROCm/rocm-systems/pull/7618) fixed the equivalent problem for gotcha args (e.g. `pthread_create`) by routing them through `append_cache_args`, but left the annotation-array path which is used by `rocprof-sys-python -a`, and any other C-API caller passing annotations.

I had put it in originally, but was told the API was deprecated?
See https://github.com/ROCm/rocm-systems/pull/7618#discussion_r3457434195
(Just want clarification for my future understanding)

projects/rocprofiler-systems/source/lib/rocprof-sys/regions.cpp
kcossett-amd, line 158: Are the return args not going to be skipped here? Please see `https://github.com/ROCm/rocm-systems/pull/7618/changes/eaa97bde73e2d34fc970f2959602eb6e806abf69#diff-3c6c3fc87b2fe4f0b37209fa68feae0d630cf5213e80a39a0ca8702632134c39R79` (from #7618 removed handling of this)

Edit: Github doing some weird things to the link, so I had to put it in a code block

marantic-amd:
> > * PR [feat(rocprofiler-systems): Allow trace cache to handle gotcha args #7618](https://github.com/ROCm/rocm-systems/pull/7618) fixed the equivalent problem for gotcha args (e.g. `pthread_create`) by routing them through `append_cache_args`, but left the annotation-array path which is used by `rocprof-sys-python -a`, and any other C-API caller passing annotations.
> 
> I had put it in originally, but was told the API was deprecated? See [#7618 (comment)](https://github.com/ROCm/rocm-systems/pull/7618#discussion_r3457434195) (Just want clarification for my future understanding)

I think that deprecation is aimed toward `user API` but I ma not sure. We still need this mechanism for python

projects/rocprofiler-systems/source/lib/rocprof-sys/regions.cpp
```diff
@@ -18,6 +19,94 @@ namespace impl
...
+    static_assert(Idx > ROCPROFSYS_VALUE_NONE && Idx < ROCPROFSYS_VALUE_LAST,
+                  "Error! index sequence should only contain values which are greater "
+                  "than ROCPROFSYS_VALUE_NONE and less than ROCPROFSYS_VALUE_LAST");
+
+    if(_annotation.type == Idx)
+    {
+        using type = tracing::annotation_value_type_t<Idx>;
+        if constexpr(std::is_pointer<type>::value)
+        {
+            auto _value = reinterpret_cast<type>(_annotation.value);
```
dgaliffiAMD: nit-pick: could be `const`.

```suggestion
            const auto _value = reinterpret_cast<type>(_annotation.value);
```

projects/rocprofiler-systems/source/lib/rocprof-sys/regions.cpp
```diff
@@ -18,6 +19,94 @@ namespace impl
...
+            _args.push_back({ _arg_number, annotation_arg_type_name<type>(),
+                              _annotation.name, annotation_arg_value_string(_value) });
+        }
+        else
+        {
+            auto* _value = reinterpret_cast<type*>(_annotation.value);
```
dgaliffiAMD: nit-pick: could be a `const`.

```suggestion
            const auto* _value = reinterpret_cast<type*>(_annotation.value);
```

projects/rocprofiler-systems/source/lib/rocprof-sys/regions.cpp
```diff
@@ -18,6 +19,94 @@ namespace impl
 {
 namespace
 {
+// The annotation type enum (ROCPROFSYS_ANNOTATION_TYPE) only ever resolves to the
+// fixed scalar set registered via ROCPROFSYS_DEFINE_ANNOTATION_TYPE in
+// annotation.hpp, so (unlike category_region's gotcha-arg serialization, which
+// must handle arbitrary caller types) these can be dispatched explicitly instead
+// of relying on a generic fmt::is_formattable/fmt::streamed fallback.
+template <typename Tp>
+std::string
```
dgaliffiAMD: nit-pick:

```suggestion
[[nodiscard]] std::string
```

projects/rocprofiler-systems/source/lib/rocprof-sys/regions.cpp
```diff
@@ -18,6 +19,94 @@ namespace impl
...
+    if constexpr(concepts::is_string_type<value_type>::value)
+        return "string";
+    else if constexpr(std::is_pointer<value_type>::value)
+        return "pointer";
+    else
+        return utility::demangle<value_type>();
+}
+
+template <typename Tp>
+std::string
```
dgaliffiAMD: nit-pick

```suggestion
[[nodiscard]] std::string
```

projects/rocprofiler-systems/source/lib/rocprof-sys/regions.cpp
```diff
@@ -18,6 +19,94 @@ namespace impl
...
+// of relying on a generic fmt::is_formattable/fmt::streamed fallback.
+template <typename Tp>
+std::string
+annotation_arg_type_name()
+{
+    using value_type = std::decay_t<Tp>;
+    if constexpr(concepts::is_string_type<value_type>::value)
+        return "string";
+    else if constexpr(std::is_pointer<value_type>::value)
+        return "pointer";
```
dgaliffiAMD: Not an issue, but I just wanted to mention, for consistency, that in [category_region](https://github.com/ROCm/rocm-systems/blob/7ab563d4d5d7fa6459c99143a439deaf3f6cbc71/projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/category_region.hpp#L150) we do not have a special case for a pointer.

projects/rocprofiler-systems/source/lib/rocprof-sys/regions.cpp
```diff
@@ -18,6 +19,94 @@ namespace impl
...
+        append_annotation_arg(_args, _annotation, _arg_number,
+                              std::index_sequence<Tail...>{});
+    }
+}
+
+// Converts a rocprofsys_annotation_t array (as passed to
+// rocprofsys_push_category_region) into the trace-cache args wire format, so
+// annotations attached to a region survive cache-replay (Perfetto + rocpd)
+// instead of only appearing on the live-instrumentation path.
+function_args_t
```
dgaliffiAMD: nit-pick:

```suggestion
[[nodiscard]] function_args_t
```

projects/rocprofiler-systems/source/lib/rocprof-sys/regions.cpp
```diff
@@ -18,6 +19,94 @@ namespace impl
...
+{
+    function_args_t _args;
+    if(_annotations == nullptr || _count == 0) return _args;
+
+    _args.reserve(_count);
+    for(size_t i = 0; i < _count; ++i)
+    {
+        const auto& _annotation = _annotations[i];
+        if(_annotation.name == nullptr || _annotation.type == ROCPROFSYS_VALUE_NONE ||
+           _annotation.value == nullptr)
```
dgaliffiAMD: static-analysis: checking range of `_annotation.type`.

```suggestion
        if(_annotation.name == nullptr || _annotation.type <= ROCPROFSYS_VALUE_NONE ||
           _annotation.type >= ROCPROFSYS_VALUE_LAST || _annotation.value == nullptr)
```

projects/rocprofiler-systems/source/lib/rocprof-sys/regions.cpp
```diff
@@ -18,6 +19,94 @@ namespace impl
...
+
+template <typename Tp>
+std::string
+annotation_arg_value_string(Tp&& _val)
+{
+    using value_type = std::decay_t<Tp>;
+    if constexpr(concepts::is_string_type<value_type>::value)
+        return std::string{ std::string_view{ std::forward<Tp>(_val) } };
+    else if constexpr(std::is_pointer<value_type>::value)
+        return fmt::format("{:#x}", reinterpret_cast<std::uintptr_t>(_val));
```
dgaliffiAMD, line 46: Optional: micro-optimization, but since the majority of annotations in your example are integers, it may be worth it.

```suggestion
    else if constexpr(std::is_integral<value_type>::value)
        return fmt::format_int{ _val }.str();
```

dgaliffiAMD (approved):

projects/rocprofiler-systems/source/lib/rocprof-sys/regions.cpp
marantic-amd: You are right, I think that category_region fallbacks to `void*` for those cases. I will align the implementation to mathc with `category_region.cpp`

projects/rocprofiler-systems/source/lib/rocprof-sys/regions.cpp
marantic-amd: This isn't a hot-path so I didn't really look for maximum optimization, but this is no brainer fix, which doesn't require I think any new includes

dgaliffiAMD (approved):

PR #6799, opened 2026-06-04
feat(clr): add DFS stream assignment to segment scheduling path

anugodavar:
## Summary
JIRA ID: AIRUNTIME-2294

This PR modernizes HIP graph stream assignment by bringing DFS-based scheduling into the segment path, making segment scheduling the default on Linux/ROCm, and splitting graph execution into classic (PAL/Windows) and segmented (Linux/ROCm) implementations.

## Segment scheduling improvements (Linux/ROCm)

- Add DFSStreamAssignment() — iterative DFS over the segment DAG (same algorithm as the former classic ScheduleOneNode, but at segment granularity). Linear segment chains stay on one stream; branches rotate to a new stream at each leaf to minimize cross-stream barriers. 
- Add SelectStreamAssignment() — chooses DFS vs round-robin at instantiation based on graph complexity (≥16 segments with average length <8 nodes → DFS; otherwise round-robin).
- Rename PrecomputeStreamAssignment() → RoundRobinStreamAssignment() for clarity.
- Remove hipErrorNotReady fallback from ScheduleNodesIntoBatches(); complexity handling is now inside SelectStreamAssignment().
- Consolidate initialization — ScheduleNodesIntoBatches() and SetKernelArgManager() moved into GraphExecSegmented::Init() so ihipGraphInstantiate only does clone() + Init() for both paths.
- Fix captureDeviceId_ — derive from segment metadata in FindStreamsReqPerDevForSegments() (with empty-graph fallback); call unconditionally in Init() to avoid SIGSEGV in BuildSyncPlan().

## GraphExec architecture split
Introduce a clean class hierarchy to support PAL (no AQL) and ROCm (segment + AQL) without unsafe casts:

- GraphExecBase — shared state, virtual Init()/Run(), RecycleLaunchSignals(), no-op virtuals for AQL update methods.
- GraphExecClassic — PAL/Windows path: topological order, per-node ScheduleOneNode/RunOneNode loop, no segments or AQL capture.
- GraphExecSegmented — Linux/ROCm path: existing segment scheduling, AQL packet capture, and multi-stream machinery.
- ihipGraphInstantiate selects classic vs segmented via GPU_ENABLE_PAL || DEBUG_HIP_GRAPH_CLASSIC_PATH. Remove the palsettings.cpp hack that zeroed segment scheduling; PAL uses GraphExecClassic directly.

## Debug flags
- DEBUG_HIP_GRAPH_SEGMENT_SCHEDULING (segmented path only):
[...]

PR #8400, opened 2026-07-10
fix(clr): remove kernelNamesOwner to fix graph UAF (ROCM-27628)

anugodavar:
## Summary

Removes `kernelNamesOwner_` from `AccumulateCommand`, eliminating the heap-use-after-free in `Unit_hipGraphAddChildGraphNode_CmplxGrph_SchedLogic` (ROCM-27628).

The previous code pre-updated lastEnqueueCommand_ before releasing the old command to guard against re-entrant queue access: releasing the old AccumulateCommand could trigger kernelNamesOwner_->release() -> ~GraphExecBase
-> finish() back into the queue while lastEnqueueCommand_ still pointed to the old command. That re-entrant path no longer exists since this PR removes kernelNamesOwner_ from AccumulateCommand entirely, so the workaround is
removed and the order restored to the straightforward release-then-retain.

## Motivation

`AccumulateCommand` retained `GraphExecBase*` via `kernelNamesOwner_` to keep borrowed kernel-name strings alive for `ReportActivity()`. When the command destructed on the ROCr async-events thread, it released `GraphExecBase`, triggering cascading graph destruction while `updateCommandsState` was still walking the command batch — heap-use-after-free.

Regression introduced by df88b76dad8 (DFS stream assignment, #6799).

## Technical Details

- Remove `kernelNamesOwner_`, `kernelNamesRef_`, and the borrowing API from `AccumulateCommand`
- Replace with `kernelNames_` (`vector<const char*>`): stable pointers into `Kernel::getDemangledName()`, resolved at dispatch time from `KernelMap()`
- `kernelNames_` and `timestamps_` are parallel vectors walked by the same index in `ReportActivity`
- `~AccumulateCommand` no longer touches `GraphExecBase`; the UAF race is structurally eliminated
- Remove `kernelNames` param from `dispatchAqlPacketBatchFlat`; extract kernel names from flat AQL packets via `KernelMap` at dispatch time

## JIRA ID
ROCM-27628

## Test Plan
- `Unit_hipGraphAddChildGraphNode_CmplxGrph_SchedLogic` — UAF fixed

## Test Result
- 'Unit_hipGraphAddChildGraphNode_CmplxGrph_SchedLogic' - passed without UAF

### DIFF
--- a/projects/clr/rocclr/device/rocm/rocvirtual.cpp
+++ b/projects/clr/rocclr/device/rocm/rocvirtual.cpp
@@ -1562,6 +1562,10 @@ bool VirtualGPU::dispatchAqlPacketBatchFlat(const std::vector<uint8_t>& flatPack
                                             amd::AccumulateCommand* vcmd, bool attach_signal,
                                             bool pre_patched, bool blocking,
                                             const std::vector<uint8_t>* flatMetadataData) {
+  // Both base and ext kernel dispatch packets place kernel_object at the same offset.
+  static_assert(offsetof(hsa_kernel_dispatch_packet_t, kernel_object) ==
+                    offsetof(hsa_amd_ext_kernel_dispatch_packet_t, kernel_object),
+                "kernel_object offset mismatch between base and ext dispatch packets");
   if (vcmd == nullptr || flatPacketData.empty() || validFullHeaders.empty()) {
     return false;
   }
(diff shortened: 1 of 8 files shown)


projects/clr/rocclr/device/rocm/rocvirtual.cpp
```diff
@@ -1575,8 +1566,41 @@ bool VirtualGPU::dispatchAqlPacketBatchFlat(const std::vector<uint8_t>& flatPack
...
-  if (kernelNames != nullptr) {
-    vcmd->setKernelNamesRef(kernelNames);
+  // Resolve kernel names at dispatch time from KernelMap — one entry per kernel
+  // dispatch slot (base HSA_PACKET_TYPE_KERNEL_DISPATCH and AMD ext variants),
+  // parallel to timestamps_ populated at signal completion. Stable const char*
+  // into Kernel objects that live for the device lifetime. No copies, no
+  // GraphExec coupling. Non-dispatch slots (barriers, memcpy) are skipped so
+  // kernelNames_ and timestamps_ always have the same cardinality.
+  {
+    static constexpr size_t kPacketSize = 64;
```
jaydeeppatel1111: This kernel-name resolution now runs unconditionally on every graph launch — an offsetof memcpy + KernelMap().find() per dispatch packet. The previous setKernelNamesRef path was the deliberate fast path for "no profiler active, names never read." This would cause perf penalty.

projects/clr/rocclr/device/rocm/rocvirtual.cpp
```diff
@@ -1575,8 +1574,39 @@ bool VirtualGPU::dispatchAqlPacketBatchFlat(const std::vector<uint8_t>& flatPack
...
-    vcmd->setKernelNamesRef(kernelNames);
+  // Resolve kernel names only when profiling is active. Gated to preserve the
+  // fast path for non-profiled launches (no KernelMap lookup, no memcpy).
+  // One entry per kernel dispatch slot (base and AMD ext variants), parallel to
+  // timestamps_ populated at signal completion. Non-dispatch slots are skipped.
+  if (vcmd->profilingInfo().enabled_) {
+    static constexpr size_t kPacketSize = 64;
+    static constexpr size_t kBaseKernelObjectOffset =
+        offsetof(hsa_kernel_dispatch_packet_t, kernel_object);
+    static constexpr size_t kExtKernelObjectOffset =
```
saleelk: umm is it at the same spot? if not you use a separte inline method to get it only based on the packet.

projects/clr/rocclr/device/rocm/rocvirtual.cpp
```diff
@@ -1575,8 +1574,39 @@ bool VirtualGPU::dispatchAqlPacketBatchFlat(const std::vector<uint8_t>& flatPack
...
+    const auto& kernel_map = dev().KernelMap();
+    for (size_t i = 0; i < numPackets; ++i) {
+      uint16_t header = static_cast<uint16_t>(validFullHeaders[i]);
+      uint8_t pkt_type = extractAqlBits(header, HSA_PACKET_HEADER_TYPE,
+                                        HSA_PACKET_HEADER_WIDTH_TYPE);
+      uint8_t amd_format = static_cast<uint8_t>((validFullHeaders[i] >> 16) & 0xFF);
+      const bool is_base_dispatch = (pkt_type == HSA_PACKET_TYPE_KERNEL_DISPATCH);
+      const bool is_ext_dispatch  = (pkt_type == HSA_PACKET_TYPE_VENDOR_SPECIFIC &&
+                                     amd_format == HSA_AMD_PACKET_TYPE_EXT_KERNEL_DISPATCH);
+      if (is_base_dispatch || is_ext_dispatch) {
```
saleelk: do I not check this already is base_dispatch or ext?

saleelk (changes requested):

anugodavar:
We fix asan failures for unit test. So testcase is not required for this PR.

saleelk (approved):

PR #8401, opened 2026-07-10
feat(rocr/clr): Add support for ext dispatch packet 4

kjayapra-amd:
## Motivation

Add support for packet 4 format, which supports launch descriptor in the AQL packet.

## Technical Details

Use packet 4 format, when using launch descriptor whenever possible.

## JIRA ID
N/A

## Test Plan
CI Tests and tests targeting dynamic dispatch, verified packet 4 usage using logs.

## Test Result
No drop in performance when using packet 4.

## Submission Checklist


projects/rocr-runtime/runtime/hsa-runtime/inc/hsa_ext_amd.h
```diff
@@ -306,17 +317,34 @@ typedef struct hsa_amd_ext_kernel_dispatch_packet_s {
...
+   *   - Bits [1:0]: setup - kernel dispatch parameters (number of dimensions)
+   *   - Bit  [2]:   dispatch_ahead_programmatic - Enable dispatch-ahead feature
+   *   - Bits [7:3]: reserved - Must be 0
    */
-  uint8_t setup;
+  union {
+    uint8_t setup;  // Format 3: full 8-bit setup field
+    struct {        // Format 4: bitfield with dispatch_ahead
+      uint8_t setup_dimensions : 2;
+      uint8_t dispatch_ahead_programmatic : 1;
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** This `setup` byte is read by the CP/FW, so its bit layout is part of the hardware contract — but C bitfield allocation order is implementation-defined. It matches the documented layout (dimensions bits[1:0], `dispatch_ahead_programmatic` bit[2]) only on little-endian GCC/Clang, which is all AMD ships, yet the rest of HSA deliberately encodes such fields with explicit shift/mask constants (e.g. `HSA_KERNEL_DISPATCH_PACKET_SETUP_DIMENSIONS`) rather than bitfields. Consider a mask/shift accessor or a `static_assert` locking the layout. Note too that the anonymous struct/union requires C11+ from every consumer of this public header.

projects/rocr-runtime/runtime/hsa-runtime/inc/hsa_ext_amd.h
```diff
@@ -410,16 +438,25 @@ typedef struct hsa_amd_ext_kernel_dispatch_packet_s {
...
+   * Format 4 (HSA_AMD_PACKET_TYPE_EXT_KERNEL_DISPATCH1):
+   *   - launch_descriptor: Pointer to amd_launch_descriptor_t structure
+   *     providing advanced scheduling control (CU enable, wave limiter,
+   *     L2 prefetch, etc.). A NULL pointer indicates no launch descriptor
+   *     is used. Dependencies must be handled via barrier packets.
    */
-  hsa_signal_t dep_signal;
+  union {
+    hsa_signal_t dep_signal;                      // Format 3: signal-based dependency
+    amd_launch_descriptor_t* launch_descriptor;   // Format 4: pointer to descriptor
```
cfreeamd, line 443: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** These 8 bytes are now either a `hsa_signal_t` (Format 3) or an `amd_launch_descriptor_t*` (Format 4), reinterpreted by the CP based on `amd_format`. The new doc says Format 4 dependencies "must be handled via barrier packets," so any emitter that stamps a Format-4 packet while still writing `dep_signal` would have the CP treat a signal handle as a descriptor pointer. Can you confirm every Format-4 emit path sets only `launch_descriptor` and never `dep_signal`? A one-line caveat in this doc block ("do not set dep_signal for Format 4") would make the union harder to misuse.

cfreeamd (commented):

Only reviewing the rocr-runtime file (hsa_ext_amd.h). The change is ABI-safe: `amd_launch_descriptor_t` resolves via the existing `#include "amd_launch_descriptor.h"`, both new unions preserve the original field sizes/offsets (1 byte and 8 bytes on LP64), the 64-byte packet size is unchanged, and existing `.setup`/`.dep_signal` accesses still compile through the anonymous unions. Two non-blocking notes inline.

projects/rocr-runtime/runtime/hsa-runtime/inc/hsa_ext_amd.h
```diff
@@ -133,6 +133,12 @@ typedef enum {
    */
   HSA_AMD_PACKET_TYPE_EXT_KERNEL_DISPATCH = 3,
 
+  /**
+   * Extended kernel dispatch packet with spatial locality computing support.
+   * Supports clusters, launch descriptor for advanced scheduling, and L2 prefetch.
+   */
+  HSA_AMD_PACKET_TYPE_EXT_KERNEL_DISPATCH1 = 4,
```
saleelk: we should call it HSA_AMD_PACKET_TYPE_EXT_KERNEL_DISPATCH_LD

projects/clr/rocclr/device/rocm/rocvirtual.cpp
saleelk: Did we miss a spot here?

projects/clr/rocclr/device/rocm/rocvirtual.cpp
```diff
@@ -1539,7 +1567,8 @@ bool VirtualGPU::dispatchAqlPacket(AqlPacket* packet, uint16_t header,
     }
     return true;
   } else {
-    dispatchBlockingWait(reinterpret_cast<hsa_kernel_dispatch_packet_t*>(packet));
+    hsa_amd_packet_type8_t amd_format = static_cast<hsa_amd_packet_type8_t>(rest & 0xFF);
```
saleelk: This is a bug, you cannot do this for a regular packet., this needs a ext packet detect guard. rest && 0xff will give you dimensions for a regular packet. for a dim3 kenrel it means we think now its type3

saleelk (changes requested):

projects/clr/rocclr/device/rocm/rocvirtual.cpp
```diff
@@ -1419,6 +1424,24 @@ bool VirtualGPU::dispatchGenericAqlPacket(AqlPacket* packet, uint16_t header, ui
   *aql_loc = *packet;
 
   metadata_preloader_.Set(packet, header, index & queueMask);
+
+  // For Format 4 (DISPATCH1), set the launch_descriptor pointer in the AQL packet
```
saleelk: this could be done inside the Set function btw . Also fix the comment

projects/clr/rocclr/device/rocm/rocvirtual.cpp
```diff
@@ -1538,7 +1567,12 @@ bool VirtualGPU::dispatchAqlPacket(AqlPacket* packet, uint16_t header,
     }
     return true;
   } else {
-    dispatchBlockingWait(reinterpret_cast<hsa_kernel_dispatch_packet_t*>(packet));
+    // rest is setup (dimensions) for regular dispatch; amd_format lives in rest only for ext packets.
+    hsa_amd_packet_type8_t amd_format = 0;
+    if constexpr (std::is_same_v<AqlPacket, hsa_amd_ext_kernel_dispatch_packet_t>) {
```
saleelk, line 1557: do you need to check for hsa_amd_ext_*_ld too here?

saleelk (changes requested):
its mising the change in queue.h 

  bool __forceinline isExtDispatch() const {
    return (type(packet.header) == HSA_PACKET_TYPE_VENDOR_SPECIFIC &&
            amd_vendor.format == HSA_AMD_PACKET_TYPE_EXT_KERNEL_DISPATCH);
  }

dayatsin-amd:
Please rename PR description to match PR bot format:

feat(rocr/clr): Add support for ext dispatch packet 4

cfreeamd (approved):

projects/clr/rocclr/device/rocm/rocsettings.hpp
```diff
@@ -38,7 +38,7 @@ class Settings : public device::Settings {
       uint blocking_blit_ : 1;         //!< Blit ops can be blocking on CPU
       uint queue_pipe_dist_ : 1;       //!< gfx94x queue pipe distribution
       uint ext_dispatch_packet_ : 1;   //!< Uses new ext dispatch packet for all launches
-      uint aql_barrier_opt_ : 1;       //!< Per-stream barrier-bit optimization
+      uint format4_supported_ : 1;     //!< Format 4 (DISPATCH_LD) supported by CP ucode
```
saleelk: we should call it ext_dispatch_packet_ld

projects/clr/rocclr/device/rocm/rocvirtual.hpp
```diff
@@ -362,12 +353,28 @@ class VirtualGPU : public device::VirtualDevice {
...
+
+        // For Format 4 (DISPATCH_LD), set the launch_descriptor pointer in the AQL packet
+        // to point to the launch_descriptor structure in the metadata packet
+        if constexpr (std::is_same_v<AqlPacket, hsa_amd_ext_kernel_dispatch_packet_t>) {
+          hsa_amd_packet_type8_t amd_format = static_cast<hsa_amd_packet_type8_t>(rest & 0xFF);
+          if (amd_format == HSA_AMD_PACKET_TYPE_EXT_KERNEL_DISPATCH_LD) {
+            auto* metadata = GetMetadataPacket(index);
+            if (metadata != nullptr && aql_loc != nullptr) {
+              // Set the pointer in both the local packet copy and the queue location
+              auto* launch_desc = const_cast<amd_launch_descriptor_t*>(&metadata->launch_descriptor);
```
saleelk, line 372: the whole idea of the Type4 packet is that the launch descriptor in the metadata packet is purely optional, Now this points to the metadata packet, which it shouldn't . we may have to think of keeping persistent launch descriptors, for now its fine here, but this design has to change

projects/clr/rocclr/device/rocm/rocvirtual.cpp
```diff
@@ -2350,7 +2369,7 @@ bool VirtualGPU::create() {
                                        dedicated_queue_));
   if (!gpu_queue_) return false;
 
-  if (dev().isa().versionMajor() == 12 && dev().isa().versionMinor() >= 5) {
+  if (dev().settings().ext_dispatch_packet_ld_) {
```
mangupta: 🤖 AI-assisted pre-review (for @mangupta)

**[BLOCKING]** Launch-descriptor version gate replaces the gfx12.5 ISA check with a bare CP ucode-version comparison

`VirtualGPU::create()` used to enable `SetLaunchDescriptorVersion(AMD_LAUNCH_DESCRIPTOR_VERSION_GFX1250)` only for `isa().versionMajor() == 12 && isa().versionMinor() >= 5`. It now keys solely off `dev().settings().ext_dispatch_packet_ld_`, which `Device::create()` computes as `ucodeVersion >= 352` with no architecture qualification. CP/MEC ucode version numbers are per-ASIC counters, not a globally ordered capability scale, so a non-gfx1250 agent (gfx90a/gfx942/gfx11xx) that reports a ucode version at or above 352 now silently enables the GFX1250 launch-descriptor layout, and a gfx1250 agent running older firmware silently loses it.

**Suggested fix:** Keep the architecture predicate and add the firmware check as an additional condition, e.g. set `ext_dispatch_packet_ld_ = isa_is_gfx1250_or_later && ucodeStatus == HSA_STATUS_SUCCESS && ucodeVersion >= kMinUcodeForFormat4`, or derive the minimum ucode threshold per ISA rather than from one global constant. Also document where 352 comes from.

projects/clr/rocclr/device/rocm/rocvirtual.hpp
```diff
@@ -353,12 +353,28 @@ class VirtualGPU : public device::VirtualDevice {
...
+              packet->launch_descriptor = launch_desc;
+              aql_loc->launch_descriptor = launch_desc;
```
mangupta: 🤖 AI-assisted pre-review (for @mangupta)

**[BLOCKING]** Format-4 launch_descriptor is never populated: SetMetadata's new `rest`/`aql_loc` parameters are defaulted at every call site

The Format-4 plumbing was moved into `SetMetadata()` behind two new defaulted parameters (`uint32_t rest = 0, AqlPacket* aql_loc = nullptr`), but no caller in this change passes them. The wrapper actually used on the dispatch path, `metadata_preloader_.Set(packet, header, index & queueMask)`, is unchanged and has no way to forward them. With `rest == 0`, `amd_format` computes to 0, never equals `HSA_AMD_PACKET_TYPE_EXT_KERNEL_DISPATCH_LD`, and the branch is dead; `aql_loc == nullptr` would block it a second time. Every Format-4 packet therefore reaches the CP with the `dep_signal`/`launch_descriptor` union holding whatever the dispatch packet was left with — at best a NULL descriptor (feature silently inert), at worst a stale `hsa_signal_t` handle reinterpreted by the packet processor as a descriptor pointer.

**Suggested fix:** Pass `rest` and the queue slot pointer from the dispatch path into `Set()`/`SetMetadata()` (or derive `amd_format` from the packet's own `amd_format` field instead of `rest`), drop the default arguments so a missing argument is a compile error, and explicitly initialize `launch_descriptor` to `nullptr` for Format-4 packets that carry no descriptor.

mangupta (commented):

projects/rocr-runtime/runtime/hsa-runtime/inc/hsa_ext_amd.h
```diff
@@ -134,6 +134,12 @@ typedef enum {
    */
   HSA_AMD_PACKET_TYPE_EXT_KERNEL_DISPATCH = 3,
 
+  /**
+   * Extended kernel dispatch packet with spatial locality computing support.
+   * Supports clusters, launch descriptor for advanced scheduling, and L2 prefetch.
+   */
+  HSA_AMD_PACKET_TYPE_EXT_KERNEL_DISPATCH_LD = 4,
```
mangupta: 🤖 AI-assisted pre-review (for @mangupta)

**[IMPORTANT]** ROCr's `isExtDispatch()` predicate is not extended to the new Format 4, leaving format-3-only packet classification

This change adds `HSA_AMD_PACKET_TYPE_EXT_KERNEL_DISPATCH_LD = 4` and updates the CLR-side classifiers (`ApplyHwEventPatches`, `AnalyzeAqlQueue`, both `dispatchAqlPacketBatchFlat` sites) to accept it, but the ROCr-side predicate that tests the same vendor format is still `amd_vendor.format == HSA_AMD_PACKET_TYPE_EXT_KERNEL_DISPATCH` only. Format-4 packets are therefore not recognized as kernel dispatches by ROCr's queue bookkeeping/error-handling path, so queue-error reporting, packet inspection, and any dispatch accounting keyed on that predicate misclassify every Format-4 dispatch.

**Suggested fix:** Extend `isExtDispatch()` in `runtime/hsa-runtime/core/inc/queue.h` (and any other `== HSA_AMD_PACKET_TYPE_EXT_KERNEL_DISPATCH` comparison) to also accept `HSA_AMD_PACKET_TYPE_EXT_KERNEL_DISPATCH_LD`, ideally via a single shared helper so future formats only need one update.

projects/clr/rocclr/device/rocm/rocvirtual.cpp
```diff
@@ -1479,6 +1480,11 @@ bool VirtualGPU::dispatchGenericAqlPacket(AqlPacket* packet, uint16_t header, ui
   bool ring_doorbell = IS_LINUX || dev().IsPm4Emulation() || blocking ||
                        (skippedDispatches_ >= skip_limit) || ring_for_non_profiler_signal;
   if (ring_doorbell) {
+    if constexpr (std::is_same_v<AqlPacket, hsa_amd_ext_kernel_dispatch_packet_t>) {
+      auto* ext_aql = reinterpret_cast<hsa_amd_ext_kernel_dispatch_packet_t*>(aql_loc);
+      ClPrint(amd::LOG_DEBUG, amd::LOG_AQL, "LD_DEBUG: amd_format=%d, rest=0x%x, rest_low_byte=0x%x",
+              static_cast<int>(ext_aql->amd_format), rest, (rest & 0xFF));
+    }
```
mangupta: 🤖 AI-assisted pre-review (for @mangupta)

**[IMPORTANT]** Leftover `LD_DEBUG` trace left in the per-dispatch doorbell path

A development trace (`ClPrint(amd::LOG_DEBUG, amd::LOG_AQL, "LD_DEBUG: amd_format=%d, rest=0x%x, rest_low_byte=0x%x", ...)`) is added inside `dispatchGenericAqlPacket`'s `ring_doorbell` branch, which runs for every ext-dispatch kernel submission. It is debugging scaffolding — the `LD_DEBUG:` tag carries no actionable handle/device/queue context, it duplicates data the existing AQL packet logging already dumps, and it adds an extra argument-evaluation and redundant `reinterpret_cast` of `aql_loc` on the hottest submission path.

**Suggested fix:** Remove the `LD_DEBUG` block before merge, or fold the format into the existing `logAqlDispatchPacketExtended` diagnostics with a descriptive message and no redundant cast.

PR #8403, opened 2026-07-10
test(rocprofiler-compute): Relocate analyze unit tests to module homes

vedithal-amd:
## Motivation

test_analyze_commands.py is a CLI/exit-code suite, but a number of pure unit tests had accumulated inside it. Move each to the test module that mirrors the code under test so the CLI suite stays focused and the unit tests follow their host modules' conventions.

## Technical Details

- Relocate unit tests to the module that mirrors the code under test:
  - `create_df_kernel_top_stats` -> new `test_file_io.py`
  - `OmniAnalyze_Base.join_prof` -> new `test_analysis_base.py`
  - `apply_filters`, `apply_kernel_filter` -> `test_parser.py`
  - `load_pc_sampling_data`, `load_pc_sample_records` -> `test_pc_sampling.py`
  - time-unit conversion / `has_time_data` -> `test_tty.py`
- Fold the one non-duplicate `MetricEvaluator` NaN case into `test_metric_utils.py` and drop the redundant division-by-zero class.
- Adapt moved tests to host conventions: no pytest markers, top-level imports (no lazy imports), class grouping, type hints.
- Deregister now-unused markers: `time_unit_conversion`, `time_unit_integration`, `time_unit_edge_cases`, `division_by_zero`.

## JIRA ID

N/A

## Test Plan

- Run the touched unit modules: `pytest tests/test_file_io.py tests/test_analysis_base.py tests/test_parser.py tests/test_tty.py tests/test_metric_utils.py tests/test_pc_sampling.py::TestLoadPcSamplingData`.

## Test Result

- [x] All touched unit modules pass (138 tests).

## Submission Checklist


xuchen-amd (commented):
## Defensive patching nit — `test_analysis_base.py`

`join_prof` calls three console helpers: `console_debug`, `console_warning`, and `console_error` (the last on the invalid-`join_type` path at `analysis_base.py:473`). The current patch line only covers two of them:

```python
common.patch_console(monkeypatch, MODULE, "debug", "warning")
```

The existing test never hits `console_error` because the YAML fixture sets `join_type: kernel`, so this isn't a bug today. But if someone later adds a test case with an invalid join type, `console_error` will call the real `sys.exit(1)` and the test will die instead of asserting cleanly.

**Suggested fix** — add `"error"` to the same call:

```python
common.patch_console(monkeypatch, MODULE, "debug", "warning", "error")
```

No behavior change for the existing test; purely defensive for future extensibility.

xuchen-amd (approved):
added a non-blocking comment, pending CI.

vedithal-amd:
@xuchen-amd This PR only relocates existing tests, no new coverage, so wiring up `console_error` is out of scope. There's no invalid-`join-type` test today, and the fixture pins `join_type: kernel`, so the `console_error` branch in `join_prof` is never hit. Whoever adds that case should add `"error"` then (they'll need the mock to assert it fires).

abchoudh-amd (approved):

PR #7444, opened 2026-06-18
Revert "[rocJPEG/rocDecode] Error on GPU UUID mismatch instead of silently using wrong render node"

amd-chiranjeevi:
Reverts ROCm/rocm-systems#7349
fixes: https://github.com/ROCm/rocm-systems/issues/7442
```
[0, Critical] rocjpeg_decoder.cpp:93: InitializeDecoder(): Failed to initialize the VA-API JPEG decoder!
rocJpegCreate(rocjpeg_backend, device_id, &rocjpeg_handle) returned ROCJPEG_STATUS_INVALID_PARAMETER at ...jpegdecode.cpp:72

0% tests passed, 22 tests failed out of 22
Total Test time (real) =  36.49 sec
```

Issue #7442, opened 2026-06-18
[Issue] Linux gfx94X-dcgpu: rocjpeg tests all fail — InitializeDecoder ROCJPEG_STATUS_INVALID_PARAMETER after GPU UUID sysfs mismatch


TriveniTadapaneni:
## Summary

All 22 rocjpeg CTest tests fail on gfx94X-dcgpu after the rocm-systems bump to `c59673f`. `InitializeDecoder()` returns `ROCJPEG_STATUS_INVALID_PARAMETER` and aborts before opening the VA-API JPEG decoder.

## Context

- **Workflow:** Linux::release / Test gfx94X-dcgpu / Test rocjpeg
- **Platform:** Linux
- **GPU arch:** gfx94X-dcgpu (AMD Instinct MI300X, gfx942)
- **Failing run:** https://github.com/ROCm/TheRock/actions/runs/27729666118/job/82101305308
- **PR that introduced the regression:** https://github.com/ROCm/TheRock/pull/5946 (Bump rocm-systems from `7ced1a6` to `c59673f`)

## Error

```text
[0, Critical] rocjpeg_decoder.cpp:93: InitializeDecoder(): Failed to initialize the VA-API JPEG decoder!
rocJpegCreate(rocjpeg_backend, device_id, &rocjpeg_handle) returned ROCJPEG_STATUS_INVALID_PARAMETER at ...jpegdecode.cpp:72

0% tests passed, 22 tests failed out of 22
Total Test time (real) =  36.49 sec
```

## Suspected Root Cause

Commit [`f1752333b7`](https://github.com/ROCm/rocm-systems/commit/f1752333b7a01331871a81a7cd8699490f5a4829) (`[rocJPEG/rocDecode] Error on GPU UUID mismatch`) replaced the silent fallback to renderD128 with a hard `return ROCJPEG_STATUS_INVALID_PARAMETER` when the GPU UUID from HIP is not found in the sysfs render-nodes map.

The bug is that `rocjpeg_decoder.cpp:90` constructs the `gpu_uuid` as raw binary bytes (`std::string gpu_uuid(hip_dev_prop_.uuid.bytes, ...)`), while `GetGpuUuids()` reads the sysfs `unique_id` as an ASCII hex string. These two representations never compare equal, so the map lookup always fails and the new hard-error path is always taken. Before this commit the fallback to renderD128 masked this mismatch.

amd-chiranjeevi:
[...]

PR #8404, opened 2026-07-10
fix(rocjpeg): bind VA-API/VCN decode to the requested GPU by PCI BDF

Xinyu302:
## Summary
On architectures where the HIP device UUID (`hipDeviceProp_t.uuid`) and the amdgpu sysfs `unique_id` come from different namespaces (observed on **gfx950 / MI355X**), `RocJpegVappiDecoder::InitializeDecoder` never finds a match in `gpu_uuids_to_render_nodes_map_` (the map is keyed by `unique_id` but looked up by the HIP UUID) and hard-falls-back to `renderD128`. As a result **every `rocJpegCreate(HARDWARE, device_id)`, for any `device_id`, binds the VCN/JPEG decoder to `renderD128` = GPU 0** — all hardware JPEG decode is funneled onto a single GPU.

## Root cause
- `rocjpeg_decoder.cpp` builds the lookup key from the HIP UUID: `std::string gpu_uuid(hip_dev_prop_.uuid.bytes, ...)`.
- `rocjpeg_vaapi_decoder.cpp` `GetGpuUuids()` keys the map by the sysfs `unique_id`, and `InitializeDecoder()` does `render_node_id = map.count(gpu_uuid) ? map[gpu_uuid] : 128`.
- On gfx950 the two ID namespaces are disjoint, so the lookup always misses → `renderD128` (GPU 0).

## Fix
Select the DRM render node by matching the device's **PCI bus ID (BDF)** via `hipDeviceGetPCIBusId` &harr; `/sys/class/drm/renderD*/device`, which is consistent between the HIP runtime and the amdgpu sysfs nodes. The `unique_id` match is kept as a fallback (no behavior change where it already works, e.g. gfx942), and when nothing matches the first available render node is used instead of hard-coding `renderD128` (this also lets single-GPU containers that expose a non-default render node work).

## Relationship to prior work
This is the proper resolution of the issue #7349 tried to address by erroring out (`Error on GPU UUID mismatch instead of silently using wrong render node`), which was reverted in #7444 (issue #7442) because erroring broke `rocJpegCreate` for affected users. Matching by BDF binds to the **correct** GPU without erroring — it neither silently uses the wrong node nor fails initialization.

## Testing (8x AMD Instinct MI355X, gfx950, ROCm 7.2)
Built the `projects/rocjpeg` library (rocJPEG 1.6.0) from this branch and validated with per-GPU `amdgpu_fence_info` `jpeg_dec_*` counters (co-tenant-immune):
- Before (unmodified develop): decode requested on `cuda:0` / `cuda:4` / `cuda:7` &rarr; JPEG-ring activity all on **GPU 0**.
- After (this change): `cuda:0` &rarr; GPU 0, `cuda:4` &rarr; GPU 4, `cuda:7` &rarr; GPU 7; 8 concurrent decoders spread across all 8 GPUs.
- Output is byte-identical across devices and matches the CPU reference; single-GPU (`cuda:0`) behavior unchanged.

## Notes
- No public API/ABI change (the modified `InitializeDecoder` is the internal VA-API decoder method).
- The same fix applies to rocDecode, which shares this render-node selection pattern.

### DIFF
--- a/projects/rocjpeg/src/rocjpeg_decoder.cpp
+++ b/projects/rocjpeg/src/rocjpeg_decoder.cpp
@@ -88,7 +88,10 @@ RocJpegStatus RocJpegDecoder::InitializeDecoder() {
     }
     if (backend_ == ROCJPEG_BACKEND_HARDWARE) {
         std::string gpu_uuid(hip_dev_prop_.uuid.bytes, sizeof(hip_dev_prop_.uuid.bytes));
-        rocjpeg_status = jpeg_vaapi_decoder_.InitializeDecoder(hip_dev_prop_.name, device_id_, gpu_uuid);
+        char pci_bus_id[64] = {0};
+        CHECK_HIP(hipDeviceGetPCIBusId(pci_bus_id, sizeof(pci_bus_id), device_id_));
+        std::string gpu_pci_bdf(pci_bus_id);
+        rocjpeg_status = jpeg_vaapi_decoder_.InitializeDecoder(hip_dev_prop_.name, device_id_, gpu_uuid, gpu_pci_bdf);
         if (rocjpeg_status != ROCJPEG_STATUS_SUCCESS) {
             CriticalLog(g_rocjpeg_logger, "Failed to initialize the VA-API JPEG decoder!");
             FunctionExitLog(g_rocjpeg_logger);
--- a/projects/rocjpeg/src/rocjpeg_vaapi_decoder.cpp
+++ b/projects/rocjpeg/src/rocjpeg_vaapi_decoder.cpp
@@ -22,6 +22,10 @@ THE SOFTWARE.
 
 #include "rocjpeg_vaapi_decoder.h"
 
+#include <algorithm>
+#include <cctype>
+#include <stdlib.h>
+
 /**
  * @brief Default constructor for RocJpegVaapiMemoryPool class.
  *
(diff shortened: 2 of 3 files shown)


projects/rocjpeg/src/rocjpeg_vaapi_decoder.cpp
```diff
@@ -22,6 +22,11 @@ THE SOFTWARE.
 
 #include "rocjpeg_vaapi_decoder.h"
 
+#include <algorithm>
+#include <cctype>
+#include <climits>
+#include <cstdlib>
```
rrawther: suggest replacing it with stdlib

projects/rocjpeg/src/rocjpeg_vaapi_decoder.cpp
```diff
@@ -389,19 +394,59 @@ RocJpegVappiDecoder::~RocJpegVappiDecoder() {
...
     device_id_ = device_id;
     std::vector<int> visible_devices;
     GetVisibleDevices(visible_devices);
     GetGpuUuids();
 
+    // Normalize the PCI bus ID (BDF) to lowercase so it matches the sysfs-derived keys.
+    std::string gpu_bdf_lower = gpu_bdf;
+    std::transform(gpu_bdf_lower.begin(), gpu_bdf_lower.end(), gpu_bdf_lower.begin(),
+                   [](unsigned char c) { return std::tolower(c); });
+
```
AryanSalmanpour, line 412: Add the code below to strip the `.function` portion from the BDF (for example, converting `"0000:8e:00.0"` to `"0000:8e:00"`). We should apply the same logic in` GetRenderNodeBusId` before returning `bus_id`, since that value is later used as the key in `gpu_bdf_to_render_nodes_map_.`

```
size_t dot_pos = gpu_bdf_lower.find_last_of('.');
if (dot_pos != std::string::npos) {
    gpu_bdf_lower = gpu_bdf_lower.substr(0, dot_pos);
}
```

This change is required to correctly support non-SPX partition modes on MI3XX GPUs. The `realpath()` call in `GetRenderNodeBusId` always resolves child nodes to a BDF with function .0 (for example, both `renderD144` and `renderD145` resolve to `"0000:8e:00.0"`). In contrast, HIP reports distinct BDFs for child nodes, including non-zero function numbers (for example, "`0000:8e:00.0`" for `renderD144` and `"0000:8e:00.1"` for `renderD145`).

As a result, if the `.function` portion is preserved, the HIP-reported BDF may not match any entry in `gpu_bdf_to_render_nodes_map_`, causing render node lookup to fail and incorrectly fall back to `renderD128`. Normalizing both paths by removing the `.function` component ensures consistent BDF matching and correct render node resolution later in the `GetDrmNodeOffset` call.

projects/rocjpeg/src/rocjpeg_vaapi_decoder.cpp
```diff
@@ -1124,4 +1190,63 @@ void RocJpegVappiDecoder::GetGpuUuids() {
...
+    }
+    std::string path(resolved);
+    size_t pos = path.find_last_of('/');
+    std::string bus_id = (pos == std::string::npos) ? path : path.substr(pos + 1);
+    // A PCI BDF has the form domain:bus:device.function; reject anything that isn't one.
+    if (bus_id.find(':') == std::string::npos || bus_id.find('.') == std::string::npos) {
+        return "";
+    }
+    std::transform(bus_id.begin(), bus_id.end(), bus_id.begin(),
+                   [](unsigned char c) { return std::tolower(c); });
```
AryanSalmanpour, line 1194: Add the code below to strip the `.function` portion from the BDF (for example, converting "0000:8e:00.0" to "0000:8e:00") for the same reason that I explained above.

   ```
size_t dot_pos = bus_id.find_last_of('.');
    if (dot_pos != std::string::npos) {
        bus_id = bus_id.substr(0, dot_pos);
    }
```

**Review by AryanSalmanpour — CHANGES_REQUESTED** [review]
Hi @Xinyu302, Thank you for creating this PR and for your contribution. We appreciate your help.

I reviewed the changes and found an issue that causes the implementation to work only in the default partition mode (SPX) on MI3XX GPUs. In non-SPX partition modes, such as DPX, QPX, and CPX, the render node lookup can fail and incorrectly fall back to the default renderD128 node.

I have explained the root cause in my review comments below, along with a proposed fix that ensures the change works correctly across all MI3XX partition modes. Please take a look at the suggestions and let me know if anything is unclear or if you would like to discuss the details further.

File: projects/rocjpeg/src/rocjpeg_decoder.cpp [hunk 4]
```diff
@@ -88,7 +88,10 @@ RocJpegStatus RocJpegDecoder::InitializeDecoder() {
     }
     if (backend_ == ROCJPEG_BACKEND_HARDWARE) {
         std::string gpu_uuid(hip_dev_prop_.uuid.bytes, sizeof(hip_dev_prop_.uuid.bytes));
-        rocjpeg_status = jpeg_vaapi_decoder_.InitializeDecoder(hip_dev_prop_.name, device_id_, gpu_uuid);
+        char pci_bus_id[64] = {0};
+        CHECK_HIP(hipDeviceGetPCIBusId(pci_bus_id, sizeof(pci_bus_id), device_id_));
+        std::string gpu_bdf(pci_bus_id);
```
rrawther: what does better to rename gpu_bdf to more specific like gpu_pci_bdf.

rrawther (commented):
Can be merged after resolving all the comments

Xinyu302:
Thanks @AryanSalmanpour and @rrawther for the review. Pushed an update addressing all comments:

- **Partition modes (@AryanSalmanpour):** strip the PCI `.function` suffix in both `GetRenderNodeBusId` and `InitializeDecoder`, so BDF matching also works in MI3XX non-SPX modes (DPX/QPX/CPX); the specific partition node is selected by `GetDrmNodeOffset`.
- **@rrawther:** switched to `<stdlib.h>` for `realpath`, and renamed `gpu_bdf` -> `gpu_pci_bdf`.
- **Copilot:** use `realpath(path, nullptr)` + `free()` (no more `PATH_MAX`); made `device_name`/`gpu_uuid`/`gpu_pci_bdf` `const std::string&`; the `gpu_uuid` doc no longer implies the sysfs `unique_id` namespace.

Rebuilt rocJPEG 1.6.0 from this branch and verified on 8x MI355X (gfx950): a decode requested on `cuda:N` now uses **GPU N**'s VCN/JPEG engine (per-GPU `amdgpu_fence_info` `jpeg_dec` counters), vs all landing on GPU 0 before; decoded output is unchanged.

AryanSalmanpour (approved):

rrawther (approved):

PR #8405, opened 2026-07-10
fix(rocprofiler-systems): Use GFX generation macros in transferBench GetHwId

sputhala-amd:
## Motivation

Building the `transferBench` example for `gfx1152` fails with invalid hardware register: not supported on this GPU, because the `GetHwId` macro's hard-coded arch list omits `gfx1152` and falls through to the CDNA-only `HW_REG_HW_ID` register that RDNA doesn't have. Switching to compiler-defined GFX generation macros fixes the build and keeps future GPU targets working without list maintenance.

## Technical Details

- Replaced the enumerated `__gfx11xx__`/`__gfx12xx__` list in `GetHwId` with generation-level macros `__GFX9__` / `__GFX10__` / `__GFX11__` / `__GFX12__`, matching upstream ROCm/TransferBench project.
- GFX9 (CDNA) keeps reading `HW_REG_HW_ID`; GFX10/11/12 (RDNA) now read the valid `HW_REG_HW_ID1` register instead of returning 0, so the CU id is actually meaningful on RDNA.
- Renamed the gpu_connect test to `test_transferbench.py`, in order to match the policy bot restriction.

## JIRA ID

JIRA ID : AIPROFSYST-602

## Test Plan

ctest -R transferBench
[gfx942, ]

## Test Result

Pass

## Submission Checklist


projects/rocprofiler-systems/examples/transferBench/TransferBench.hpp
```diff
@@ -3287,8 +3289,8 @@ RunNicExecutor(int const iteration, ConfigOptions const& cfg, int const exeIndex
 //========================================================================================
 
 // Converts register value to a CU/SM index
-static uint32_t
-GetId(uint32_t hwId)
+static std::uint32_t
+GetId(std::uint32_t hwId)
 {
 #if defined(__NVCC_)
```
dgaliffiAMD: Pre-existing typo: 

```suggestion
#if defined(__NVCC__)
```

dgaliffiAMD (approved):

PR #8406, opened 2026-07-10
[rocjitsu] DBT changes to make Qwen bf16 models work

Groverkss:
## Motivation

rocjitsu emulates AMD GPUs by intercepting the KMD/driver and syscall/ioctl layer. To run a CDNA4 (gfx950) workload on a CDNA3 (gfx942) or RDNA host, the guest kernels' machine code must be translated at load time — a dynamic binary translation (DBT) problem. The concrete driver for this work is running Qwen bf16 models end-to-end, which exercises three things the translator did not previously handle correctly:

1. Instruction-level ISA gaps. Real transformer kernels (Triton/HIP GEMM, flash-attention, HipKittens matmul) emit CDNA4 opcodes with no 1:1 CDNA3/RDNA equivalent — 64-bit address arithmetic (v_lshl_add_u64), bf16 conversions, v_dot2, and permlane/swap forms — which must be lowered to multi-instruction target sequences that preserve exact semantics,
  including carry, liveness, and lane layout.

2. LDS capacity mismatch. A CDNA4 kernel can request more LDS (static or dynamic) than a CDNA3/RDNA host physically provides. Rather than reject those kernels, we virtualize the overflow into a global-memory backing buffer addressed through a per-dispatch sidecar descriptor, so LDS-heavy kernels still run.

3. Correctness under the emulator's real threading and ELF-relocation model. The translator rewrites .text (compaction, expansion, per-kernel placement) and appends descriptors, so any relocation or symbol whose value it does not remap must be detected and rejected rather than silently miscompiled. The dispatch path runs across the doorbell/scanner thread,
  ROCr intercept callbacks, and HSA client threads, so the hook state must be race- and lifetime-correct.

This PR makes those three areas work and hardens every failure path to fail closed (reject the load or abort with a diagnostic) instead of fail open (miscompile, or submit a host-faulting launch).

What's in the change

124 files, of which 35 are auto-generated ISA/DBT tables regenerated from the amdisa Python code-generation pipeline (no hand edits to generated code). The remaining changes are the DBT engine, the virtual-LDS mechanism, the HSA hook layer, the code-object patcher, and tests. New end-to-end kernel assets (virtual_lds_smoke.hip, HipKittens bf16 matmul.s, cvt_pk_bf16_f32.hip) exercise the real workloads.

## Technical details

1. Semantic instruction lowering (code/dbt/semantic/)

Each unsupported CDNA4 instruction is lowered to a target sequence via per-ISA rule tables shared through cdna4_to_rdna_common. Notable lowerings:

- v_lshl_add_u64 → shift + 64-bit carry-add. Computes D = (S0 << S1[5:0]) + S2 as v_lshlrev_b64 followed by v_add_co_u32/v_add_co_ci_u32. The carry destination is a dead SGPR pair obtained from the liveness analysis (find_free_sgpr_pair + monotonic require_sgprs), never VCC — VCC is not liveness-tracked and the source instruction defines no carry, so clobbering it would be a silent correctness bug. When the destination aliases the VGPR addend the shift lands in a scratch VGPR pair so the addend survives to the add. Non-register addends (inline constants, literals, and v255 — whose high half selector 512 cannot encode a VGPR) are rejected rather than miscompiled.
- bf16 / v_dot2 / cvt paths and DPP/DPP8 on read/write-output swaps. The generator's DPP preamble assumed src0 at operand index 0; for the swap classes (vector_swap, vector_permlane16/32_swap) index 0 is the read/write output, so applying DPP there would shuffle vdst. The generator now emits throw UnimplementedInst for a DPP/DPP8 source on those classes across all affected ISAs (regenerated), rather than producing wrong code.

2. Virtual LDS via sidecar descriptors (code/dbt/virtual_lds*, code/patch/sidecar_metadata*, hooks/virtual_lds*)

When a kernel's requested LDS exceeds the host limit, the translator emits a sidecar descriptor variant that redirects LDS traffic to a dense global backing buffer, plus a kernarg wrapper carrying a runtime-state block (backing pointer + per-dimension strides). At dispatch, the hook layer plans the rewrite (plan_virtual_lds_dispatch: KeepNormal / UseSidecar /
  Reject), allocates the backing/kernarg buffers, writes the wrapper, and republishes the AQL packet on the sidecar descriptor under a transient INVALID header. Metadata is a versioned wire format (RJVLDS1/RJSIDE1) with bounds-checked parsing.

Fail-closed dispatch state machine. A nullopt from prepare_virtual_lds_dispatch now means strictly "the normal descriptor is safe" (no plan and packet fits host LDS, or below threshold). Any case where the normal descriptor would be oversized — an unregistered kernel whose group segment exceeds the host limit, or any preparation failure after the planner selects a sidecar (geometry reject, missing runtime-state ABI, null kernarg, wrapper-layout mismatch, backing/kernarg
  OOM, wrapper-write failure) — calls a [[noreturn]] abort_virtual_lds_dispatch with a diagnostic. Falling back to the
  oversized normal descriptor would fault the GPU, and a persistent INVALID header would stall the queue forever; abort is strictly ordered before any INVALID publish so no slot is left half-written.

3. ELF / code-object patching safety (code/patch/code_object_patcher.*, code/amdgpu_elf.h)

Because DBT moves .text without remapping symbol st_value or relocation addends, the patcher fails the translation closed when it sees anything it cannot relocate:

- In-.text relocation places (has_relocations_within_text) — handles both ET_DYN (r_offset = vaddr) and ET_REL (r_offset section-relative via sh_info).
- Relocations resolving into .text (has_relocation_to_text_symbol) — rejects any symbol defined in .text regardless of type (STT_FUNC,  STT_NOTYPE labels, STT_SECTION+addend), keyed on st_shndx; and rejects a symbol-less R_AMDGPU_RELATIVE64 whose r_addend lands in the source .text interval (the loader forms its value from load bias + addend, which DBT does not remap). The reloc-type constant is pinned to the AMDGPU ELF ABI value (13) with a static_assert in the tests so it cannot drift.

4. HSA hook thread-safety (hooks/rj_hsa_dbt_hooks.cpp,hooks/sidecar_registry.*)

- Scanner jthread lifecycle: on teardown the thread is moved to a local under the lifecycle mutex and joined with no lock held, avoiding the self-deadlock (the scanner body takes the queue mutex).
- HSA runtime calls made under the dispatch-queue mutex are the non-re-entrant saved originals (captured at OnLoad), documented as a load-bearing invariant so a future change cannot route them through the wrappers and deadlock.
- Consistent cross-registry lock ordering (DispatchQueue → Runtime → Sidecar), copy-on-return lookups, and correct release/acquire ordering on the 64-bit AQL header via atomic_ref.

5. Supporting infrastructure

Liveness analysis extensions (restrict_live_before_to_instructions, VGPR/SGPR free-run search), kernarg extension layout, kernel text layout, spill manager, and the amdisa generator changes that produce the regenerated ISA tree.

## Issue Tracking

#8406

## Test Plan

- Focused unit tests for every lowering and guard: v_lshl_add_u64 (carry-uses-SGPR-not-VCC, v255 reject, inline-constant reject, dest-aliasing scratch), text-symbol and RELATIVE64 relocation rejection (patcher + E2E, in-text vs out-of-text addend), sidecar metadata parse guards (version, truncation, oversized string/record-count, non-zero reserved).
- E2E DBT translation of real Triton/HipKittens/flash-attention kernels.
- Run CI tests

## Test Result

All CI tests: PASSED

## Submission Checklist


Groverkss:
I need to regenerate gfx1250 files btw, but will do that once reviewed.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/hooks/rj_hsa_dbt_hooks.cpp
```diff
@@ -1852,6 +2105,658 @@ void map_pointer_info_accessible_agents(uint32_t *num_agents_accessible, hsa_age
...
+          metadata.static_lds_bytes, packet.group_segment_size,
+          static_cast<unsigned long long>(rewrite.requested_lds));
+      return std::nullopt;
+    }
+    if (rewrite.decision == VirtualLdsDispatchDecision::Reject) {
+      trace_virtual_lds_dispatch(
+          "virtual-LDS plan rejected delivery=%s packet=%llu kernel=%s requested=%llu reason=%s",
+          delivery, static_cast<unsigned long long>(packet_id), metadata.kernel_name.c_str(),
+          static_cast<unsigned long long>(rewrite.requested_lds), rewrite.diagnostic.c_str());
+      return std::nullopt;
```
kuhar: We should distinguish "keep the normal descriptor" from a failed virtual-LDS rewrite and reject the latter explicitly. `plan_virtual_lds_dispatch` already returns `Reject` for invalid geometry, missing workgroup-ID SGPRs, stride/backing overflow, and private-size overflow, but `prepare_virtual_lds_dispatch` turns that into `nullopt`; both delivery adapters then forward the original oversized packet. The same fallback happens after wrapper or allocation failures. Please return a tri-state result and publish a safe error or trap outcome instead of dispatching the unusable normal descriptor.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/hooks/rj_hsa_dbt_hooks.cpp
```diff
@@ -1919,8 +2824,47 @@ hsa_status_t HSA_API rj_queue_create(hsa_agent_t agent, uint32_t size, hsa_queue
...
-                  queue);
+
+  hsa_status_t status = HSA_STATUS_ERROR;
+  auto *intercept_create = layer().amd_queue_intercept_create();
+  auto *intercept_register = layer().amd_queue_intercept_register();
+  if (intercept_create != nullptr && intercept_register != nullptr && size >= 3) {
+    status = intercept_create(mapped, size, type, callback, data, private_segment_size,
+                              group_segment_size, queue);
+    if (status == HSA_STATUS_SUCCESS && queue != nullptr) {
+      VirtualLdsDispatchQueueRegistry::instance().record_queue(*queue, mapped, true);
```
kuhar: This would be safer if we only recorded an intercept-managed queue after `hsa_amd_queue_intercept_register` succeeds. The helper currently drops the registration status, so `rj_queue_create` returns success with `uses_packet_interceptor=true`; that disables both doorbell rewriting and polling fallback even though no callback is installed. Please propagate the status and either keep a fallback rewrite path or fail and recreate the queue.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/binary_translator.cpp
```diff
@@ -525,321 +855,590 @@ TranslatedCodeObject BinaryTranslator::translate(const AmdGpuCodeObject &obj) {
...
+      return false;
     }
-    if (!found_entry_delta) {
-      append_error(result.diagnostics, DiagnosticKind::KernelDescriptor,
-                   "kernel descriptor entry offset is not present in the reachable body",
-                   layout.source_entry);
-      return leave_unchanged();
+
+    for (KdTranslation &translation : descriptor_translations) {
+      // Normal hardware-LDS and virtual-LDS sidecar descriptors share the same
```
kuhar, line 1063: We should not keep the normal descriptor when a required virtual-LDS sidecar fails. `SkipFailedVirtualLdsSidecarKeepsNormalDescriptor` sets static LDS to 105600 bytes for CDNA3, whose limit is 64 KiB, then passes while leaving that oversized descriptor live and emitting no virtual-LDS metadata. That normal path cannot dispatch on the target. Please distinguish required sidecars from dynamic-overflow fallbacks; if a required sidecar is skipped, redirect the matching normal descriptor to the skipped-kernel trap while preserving independent kernels.

emulation/rocjitsu/tests/kernels/CMakeLists.txt
```diff
@@ -73,14 +75,21 @@ function(rj_add_triton_asm_fixture name target)
...
+rj_add_asm_fixture(triton_cdna3_matmul_dynamic_32x32x64 gfx942)
+rj_add_asm_fixture(triton_cdna3_matmul_buffer_async_1024 gfx942)
+rj_add_asm_fixture(triton_cdna3_flash_attention_no_async_1024 gfx942)
+rj_add_asm_fixture(triton_cdna3_flash_attention_buffer_async_1024 gfx942)
+
+# HipKittens-generated assembly fixtures.  The upstream sources build as Python
+# extension modules and pull in the full HipKittens/PyTorch/pybind stack, so the
+# checked-in .s files preserve the generated gfx950 code object in a readable,
+# reassemblable form for DBT tests.
+rj_add_asm_fixture(hipkittens_bf16fp32_256_256_64_32_with16x32 gfx950)
```
kuhar, line 91: We should install the four new fixtures used by installed CTest: `virtual_lds_smoke.o`, `cvt_pk_bf16_f32.o`, and both HipKittens `.hsaco` files. I reproduced this from an install-tree CTest run: all four corresponding `Cdna4ToCdna3DispatchTest` cases fail at `Executable::is_valid()` because the files are absent from `RJ_DEVICE_KERNEL_FIXTURES`. Please also rewrite `ROCJITSU_HIPKITTENS_KERNEL_DIR` to the install-tree kernel directory instead of retaining the build-tree path.

kuhar (commented):
A few comments

emulation/rocjitsu/tests/kernels/CMakeLists.txt
kuhar: Or alternatively, move these tests to the test corpus?

newling:
It is easier to review PRs that have 1 fix/feature, rather than a single PR that gets an integration working. Can you at least in the PR description, include a summary of what fixes/features are added to get the integration test working?

Groverkss:
> It is easier to review PRs that have 1 fix/feature, rather than a single PR that gets an integration working. Can you at least in the PR description, include a summary of what fixes/features are added to get the integration test working?

I mentioned this offline, but it's really hard to split up this PR. This PR lets us run some pytorch models which is the minimum validation we need. Changes after this will be smaller, but it's almost impossible for me to split this up and catch bugs without having some real validation through a model. This sets the design for DBT and further changes will be modifications to the design, not a redesign.

I can include a summary of features in the summary.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/hooks/rj_hsa_dbt_hooks.cpp
```diff
@@ -1852,6 +2105,658 @@ void map_pointer_info_accessible_agents(uint32_t *num_agents_accessible, hsa_age
...
+                                   uint64_t end_packet_id, bool advance_contiguous_cursor) {
+    // The HSA queue can contain invalid/no-op holes before later ready dispatches.
+    // Keep scanning the visible range so an oversized-LDS kernel is not missed behind
+    // such a hole, but only advance the contiguous cursor across packets that were
+    // actually ready when observed.
+    for (uint64_t packet_id = begin_packet_id; packet_id < end_packet_id; ++packet_id) {
+      const bool ready = rewrite_packet(state, packet_id);
+      if (advance_contiguous_cursor)
+        note_packet_ready(state, packet_id, ready);
+    }
```
amd-arosa, line 2784: `retired_buffers` is only swept in `prepare_virtual_lds_dispatch` and `retire_destroyed_completion_signal`. A queue that issues one virtual-LDS dispatch and then only normal dispatches (never destroying that completion signal) keeps the backing allocation — potentially a whole grid worth of LDS — alive until queue destruction. Consider calling `release_completed_retired_buffers(state)` once here per scan/doorbell pass so completed buffers are reclaimed even when no new virtual-LDS dispatch is prepared.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/hooks/rj_hsa_dbt_hooks.cpp
```diff
@@ -1852,6 +2105,658 @@ void map_pointer_info_accessible_agents(uint32_t *num_agents_accessible, hsa_age
...
+        static std::atomic<uint32_t> scan_trace_count{0};
+        if (scan_trace_count.fetch_add(1, std::memory_order_relaxed) < 128) {
+          trace_virtual_lds_dispatch("scan queue=%p next=%llu write=%llu",
+                                     static_cast<void *>(state.queue),
+                                     static_cast<unsigned long long>(state.next_packet_id),
+                                     static_cast<unsigned long long>(write_index));
+        }
+      }
+      const uint64_t tail_packets = std::min(write_index, kVirtualLdsScannerTailPackets);
+      rewrite_packet_range(state, write_index - tail_packets, write_index, false);
```
amd-arosa: This rescans `[write_index - tail, write_index)` from the producer write index and can set a slot header to INVALID and mutate fields while the CP may already be consuming that slot. It is safe only because the doorbell hook performs the first rewrite before the CP sees the packet (so the scanner only re-confirms idempotently) and it never starts for intercept queues (the primary path). Worth a one-line comment stating that invariant so the residual CP race on the no-intercept fallback is explicit.

emulation/rocjitsu/tests/dbt/sidecar_registry_test.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+  EXPECT_FALSE(registry.kernel_name_for_object(kNormalObject).has_value());
+  EXPECT_FALSE(registry.find_by_kernel_object(kNormalObject).has_value());
+
+  registry.note_kernel_object(10, kNormalObject + 0x10000, 64);
+  EXPECT_EQ(registry.kernel_name_for_object(kNormalObject + 0x10000), "new");
+  EXPECT_EQ(registry.private_segment_size_for_object(kNormalObject + 0x10000), 64u);
+}
+
+TEST_F(SidecarRegistryTest, RejectsVariantAddressOverflow) {
```
amd-arosa, line 108: Thread-safety is the core contract of this registry, but every test here is single-threaded. Consider a multithreaded stress test driving `record_load` / `record_symbol` / `note_kernel_object` / `find_by_kernel_object` / `erase_executable` from several threads (ThreadSanitizer-friendly) to lock in the mutex invariant and the dispatch-vs-signal-destroy ordering.

emulation/rocjitsu/tests/dbt/virtual_lds_metadata_test.cpp
```diff
@@ -0,0 +1,44 @@
...
+  ASSERT_EQ(parsed->size(), 1u);
+  const auto &record = parsed->front();
+  EXPECT_EQ(record.kernel_name, input.front().kernel_name);
+  EXPECT_EQ(record.sidecar_variant_name, input.front().sidecar_variant_name);
+  EXPECT_EQ(record.static_lds_bytes, input.front().static_lds_bytes);
+  EXPECT_EQ(record.normal_private_segment_size, input.front().normal_private_segment_size);
+  EXPECT_EQ(record.virtual_private_segment_size, input.front().virtual_private_segment_size);
+  EXPECT_EQ(record.virtual_lds_base_sgpr, input.front().virtual_lds_base_sgpr);
+  EXPECT_EQ(record.flags, input.front().flags);
+}
```
amd-arosa, line 41: This covers only the happy-path round-trip. `parse_virtual_lds_metadata` has dedicated bounds/version/reserved checks; please add negative cases (truncated header, `record_count` larger than the buffer, wrong magic/version, non-zero reserved) that assert it returns `nullopt`.

emulation/rocjitsu/tests/patch/sidecar_metadata_test.cpp
```diff
@@ -0,0 +1,68 @@
...
+TEST(SidecarMetadata, RejectsUnknownWireVersion) {
+  const std::vector<SidecarVariantMetadata> input = {{.kernel_name = "k", .variant_name = "v"}};
+  auto bytes = serialize_sidecar_metadata(input);
+  ASSERT_GE(bytes.size(), 12u);
+  const uint32_t unknown_version = 2;
+  std::memcpy(bytes.data() + 8, &unknown_version, sizeof(unknown_version));
+
+  EXPECT_FALSE(parse_sidecar_metadata(bytes).has_value());
+}
+
```
amd-arosa, line 66: Good to see version rejection covered. Consider also asserting rejection of a truncated buffer (bytes cut mid-record, or `string_bytes` exceeding the remaining size) so the offset/bounds checks in `parse_sidecar_metadata` are exercised, not just the version gate.

emulation/rocjitsu/tests/dbt/waitcnt_translator_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+  for (uint32_t word : words) {
+    const uint8_t op = (word >> 16) & 0x7F;
+    has_loadcnt |= op == rdna4::kSWaitLoadcntSopp;
+    has_storecnt_dscnt |= op == rdna4::kSWaitStorecntDscntSopp;
+  }
+  EXPECT_TRUE(has_loadcnt);
+  EXPECT_TRUE(has_storecnt_dscnt);
+}
+
+} // namespace
```
amd-arosa, line 74: These two cases assert only a word count (`>= 3`) and boolean opcode presence. Since the encoding is deterministic, consider asserting the exact emitted words / opcode fields so a regression in the encoded simm16 or opcode ordering is caught.

amd-arosa (commented):
Overall looks pretty good to me.
But it breaks the build for multiple reasons, and that needs to be resolved.
Also Have a couple comments that I would like you to resolve.
In future this should have been like 3-6 prs to make reviewing easier.

atgutier (commented):
From the agent:

 MAJOR — must fix (domain correctness / silent corruption)

  1. BF16 narrowing turns NaN into Inf — cdna4_to_cdna3.cpp:157 (emit_cdna3_f32_to_bf16_rne, used by
  v_cvt_pk_bf16_f32). The emitted sequence unconditionally applies the RNE rounding bias (bits +
  0x7fff + ((bits>>16)&1)) >> 16 to every input including exp==0xFF. The authoritative guest reference
   util::f32_to_bf16_rne (data_types.h:278) special-cases NaN/Inf (no bias-add; sets bit 16 to keep
  NaN a NaN). So a NaN activation like 0x7F800001 becomes +Inf — a real value change, not the "~1ulp
  drift" the PR accepts. Concrete, in-scope, fixable: add the exp==0xFF guard before the bias add. The
   existing test can't catch it — its oracle copies the emulation's own truncated (NaN-incorrect)
  formula and restricts to finite inputs.
  2. v_permlane16_swap_b32: no correct oracle — the translation (cdna4_to_cdna3.cpp:625) swaps all
  four 16-lane groups (0-15↔16-31 and 32-47↔48-63, per its own doc comment and public ISA knowledge),
  but the guest simulator (cdna4/vop1.cpp:10956, vop3.cpp:2102) only swaps 0-15↔16-31, leaving lanes
  32-63 untouched. The translation is very likely the hardware-faithful side (simulator is buggy), but
   the disagreement means there's no correct in-tree oracle and any execute test would validate
  against the wrong behavior. Reconcile — fix the simulator loop to cover all groups, then add a
  wave64 execute test covering lanes 32-63. (Reviewer flagged it couldn't confirm 4-vs-2-group ground
  truth without quoting the confidential ISA spec — please confirm before relying on either side.)
  3. v_dot2_f32_bf16: simulator widens as f16, not bf16 — the guest execute_v_dot2_f32_bf16_vop3p
  (execute_shared.h:10033) uses util::f16_to_f32 (5-bit exponent) but the instruction operates on bf16
   (8-bit exponent); the translation correctly widens as bf16. Again the translation is right and the
  simulator is the buggy oracle, so the dot2 translate-and-execute path is unvalidated for numerics
  (the translate_test checks are structural-only). Fix the guest to use util::bf16_to_f32 and add an
  execute-vs-guest test.
  4. Non-atomic set-once g_runtime_api global — hooks/virtual_lds.cpp:23. A struct of 9 pointers
  written by set_virtual_lds_runtime_api and read on every dispatch/allocate path across app threads +
   the scanner jthread. Race-free only because it's set once in OnLoad before threads exist — an
  undocumented, fragile invariant (any reload or moved call is an immediate data race). Fix: document
  + assert set-once, publish behind std::atomic<const T*> with release/acquire, and clear it in
  uninstall().
  5. Unbounded retired_buffers growth — rj_hsa_dbt_hooks.cpp:2318. Virtual-LDS dispatches whose
  completion signal is never pending (borrowed handle 0 with no signal_create, or a signal already ≤0
  at enqueue) move their backing+kernarg buffers to retired_buffers, never reclaimed until queue
  destruction. A long-lived queue repeatedly dispatching such kernels leaks for the queue's life. Fix:
   bound the list (flush by polling) or force an internal completion signal for every rewritten
  dispatch.

  The systemic finding (ties the translation MAJORs together)

 The correctness tests for the new bf16/gfx950-only sequences are structural (opcode-sequence)
  checks, and the one execute test deliberately uses finite inputs and validates against a helper that
   copies the emulation's own NaN-incorrect formula. No new sequence has a translate-and-execute test
  comparing against the true guest semantics on edge inputs (NaN/Inf/denormal/±0, full-wave lane
  coverage for cross-lane ops), and two of the guest simulators that should be the oracle are
  themselves buggy (permlane16, dot2). For a PR whose entire justification is numeric fidelity for
  bf16 inference, this is the gap that matters most: the layer-by-layer "~1ulp drift" validation was
  done externally, but the in-tree test suite can't catch a regression in any of these sequences.

  MINOR (across subsystems)

  - kernel_text_layout.cpp (+799) and code_object_patcher.cpp have no dedicated unit tests — the
  branch-relocation/long-branch/island-chain code most able to produce a wild jump is exercised only
  indirectly, while the easier serialization helpers all got tests. Add a kernel_text_layout_test
  (in-range/out-of-range branch resolution decoded and followed to target; kernarg-preload dual-stub
  256-byte spacing; body-entry residue preservation).
  - Silent verbatim copy of untranslated encodings — binary_translator.cpp:1815: a gfx950-only opcode
  with no rule and no legalization entry is copied through as-is (may be a valid-but-different CDNA3
  instruction). The Expand-required and matched-but-failed paths fail loudly, but the "no entry at
  all" case doesn't — and the VOP3 forms of the permlane swaps exist in the decoder but aren't in the
  rule table, so they'd hit this path. Add the VOP3 rule entries + a gfx950-only deny-list forcing a
  loud diagnostic.
  - Unguarded kernarg memcpy bound (kernarg_extension.cpp:141 — original_kernarg_size not checked
  against wrapper_size before the copy; safe for in-tree callers, latent for a hand-built layout);
  polling-scanner races the AQL producer on non-intercept queues (inherent to the fallback design,
  documented, intercept path is primary); find_free_run doesn't assert power-of-two alignment
  (liveness.cpp:283); no concurrency test for the (correctly-locked) registries under TSan; duplicated
   scalar-add opcode constants (instruction_builder); g_runtime_api not cleared on unload; dot2 clamp
  v_med3_f32 vs guest std::clamp differ on NaN.

  NIT

  VgprForbiddenRange hand-unrolled padding idiom (use std::span); manual kSrcLiteral64=254 magic
  constant that should be generated; LivenessAnalysisOptions tri-state bool+span; SGPR-write-mask
  silent clamp on wide refs; the launch-stub residue invariant documented only in code comments.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/binary_translator.h
```diff
@@ -110,6 +110,15 @@ struct BinaryTranslatorOptions {
   /// because the partially translated text is only useful for finding failures,
   /// not for execution.
   bool debug_continue_after_failure = false;
+
+  /// @brief Preserve a failed kernel body and keep translating independent kernels.
```
kuhar, line 114: We should reject skipped-kernel artifacts from executable output paths, or mark and enforce them as non-dispatchable. skip_failed_kernels records only a warning, so TranslatedCodeObject::ok() remains true and rj_dbt_translate --output-mode code-object writes the ELF and exits 0. The HSA hook already rejects the same KernelSkipped diagnostic because its s_trap; s_endpgm stub can complete normally without a trap handler; callers of the public translator API and the CLI have no equivalent guard.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/kernarg_extension.cpp
```diff
@@ -0,0 +1,287 @@
...
+  for (const ExtensionRecord &record : extension_records)
+    append_pod(out, record);
+  for (const PayloadRecord &record : payload_records)
+    append_pod(out, record);
+  out.insert(out.end(), strings.begin(), strings.end());
+  return out;
+}
+
+std::optional<std::vector<KernargExtensionMetadata>>
+parse_kernarg_extension_metadata(std::span<const uint8_t> bytes) {
```
kuhar, line 228: It would be useful to add focused negative-path coverage for the kernarg extension helpers. The parser currently has only a round trip even though malformed header, count, string, and alignment fields feed the wrapper layout; write_kernarg_extension_wrapper also has no tests for mismatched sizes or counts and null non-empty inputs, and the layout overflow guards are untested. A compact table of corrupt metadata plus a few EXPECT_FALSE writer and max-size layout cases would cover the fail-closed contracts already exercised for the sibling metadata formats.

kuhar (commented):
A few comments

kuhar (approved):
Approving but it would be nice to address at least https://github.com/ROCm/rocm-systems/pull/8406#discussion_r3619243644 before landing

PR #8407, opened 2026-07-10
feat(ci): Enable MI455 tests as non-blocking

geomin12:
Enable MI455 tests for gfx125X-dcgpu builds by removing the hardcoded disable flag.

Added a custom build and test step for mi455, so builds are blocking but tests are not (due to limited capacity)


dan-garvey (approved):
LGTM as long as tests pass to expectation

PR #8408, opened 2026-07-10
[rocjitsu] Use generated instruction builders

Groverkss:
In DBT we currently have a number of places that reinvent instruction fields through isa facts. Generate instruction builders so we have a canonical way of building instructions.


kuhar (commented):
Very nice, thanks for cleaning this up. gfx1250 seems missing?

kuhar:
Seems like you also need to regenerate cdna1/2 files from xml. Other than that looks good!

RattataKing:
The corpus CI failure is a likely a false positive, the failed test `cts.gfx942.fpsan.fpsan_mfma_cdna3_f64_test` passed locally with 9s on my side

kuhar (approved):
I regenerated gfx1250 and cdna1/2

PR #8416, opened 2026-07-10
fix(rocr): Return all SDMA engines for gfx125+ in DmaPreferredEngine

dayatsin-amd:
## Summary

On gfx125+ platforms all SDMA engines are equivalent and there are no dedicated XGMI engines. `DmaPreferredEngine` previously split recommendations by transfer direction—first two engines for H2D/D2H and the remainder for P2P—which unnecessarily restricted engine choice.

This change returns the full SDMA engine mask for gfx125+ regardless of copy direction, allowing load balancing to use any available engine.

## JIRA ID

JIRA ID : PLAT-203739

## Test plan

- [ ] Verify SDMA copy on gfx1250 for H2D, D2H, and P2P transfers
- [ ] Confirm `DmaPreferredEngine` returns all engines on gfx125+
- [ ] Confirm non-gfx125+ engine selection behavior is unchanged


dayatsin-amd:
I will try to find a system and run TransferBench A2A to compare perf

dayatsin-amd:
Closing per author request.

PR #8417, opened 2026-07-10
feat(rocjitsu): Ensure that register accesses are plugin-observed

newling:
## Summary

This PR supersedes #8075 and #8231.

Those PRs fixed an existing bug (register read through backdoor didn't fire plugin hook) and generalized VGPR read observation in instruction emulation. In particular, #8231 introduced `RegisterAccess` as the intended way for instruction code to read and write registers, but that boundary was still mostly conventional: generated and hand-written instruction code was expected to use `RegisterAccess`, while raw register APIs were still reachable through `Wavefront::cu()` and public `Operand` accessors.

This PR turns that convention into an API boundary.

Instruction code now sees a narrow `InstructionComputeUnitView` from `Wavefront::cu()` instead of a raw `ComputeUnitCore&`, and logical operand value access goes through `amdgpu::RegisterAccess`. The goal is to prevent ordinary instruction implementations from bypassing the observed access path by accident.

## Review Structure

The branch has two commits:

1. `feat(rocjitsu): enforce register access boundary`

   This is the main review target. It contains the hand-written source changes, generator source changes, and tests.

2. `chore(rocjitsu): refresh generated AMDGPU ISA files`

   This is generated C++ output only.

## What Changed

The main new type is `amdgpu::RegisterAccess`. It is the instruction-facing facade for:

- logical operand reads and writes;
- SIMD operand read/write/read-write views;
- physical VGPR region reads and writes;
- physical SGPR reads and writes.

The distinction between logical operand access and physical register access is intentional. Most instruction bodies should deal in operands. Some matrix, memory, and address helpers already operate on physical register indices, so they use the physical-register side of the same facade.

`Operand` still owns the ISA-specific resolution logic: inline constants, SGPR operands, VGPR operands, DPP/SDWA delegates, packed 16-bit operands, and so on. Its value/chunk methods are now private backend hooks rather than instruction-facing APIs. `RegisterAccess` is the public path into those hooks.

`InstructionComputeUnitView` is the narrow CU service view exposed to instruction emulation. It deliberately omits raw VGPR/SGPR storage APIs. Raw CU access remains available to VM/storage code and to the AMDGPU operand backend where the emulator needs it to implement operand resolution and zero-copy SIMD views.

## Design Notes

This is a boundary-tightening PR, not a wholesale rewrite of operand resolution. The AMDGPU operand backend already lived in the generic operand layer before this PR; this change keeps that structure, but makes the backend private and routes instruction-visible access through `RegisterAccess`.

The view objects returned by `RegisterAccess` are instruction-scoped. They expose observed spans or pointers into register storage for hot SIMD/matrix paths, and callers should not cache them across instruction executions.

One subtle point is that SIMD operand views are lane-oriented. If a SIMD view falls back to broadcasting a scalar-backed operand, it must broadcast the value that `read_lane` would have produced, not necessarily the value from `read_scalar`. That matters for true16/f16 inline constants.

## Testing

The test coverage has three broad layers.

First, direct `RegisterAccess` unit tests pin the facade contract: VGPR region reads are observed, write-only access does not report reads, read-write access reports the read side, 64-bit SGPR/VGPR reads observe both registers, operand chunk reads observe the correct lane window, and scalar-backed SIMD views use lane semantics.

Second, integration tests exercise real instruction/plugin behavior. These cover VALU, VOP3, MFMA, WMMA, race-detector reporting, and generated instruction paths that should now flow through `RegisterAccess`.

Third, codegen tests check that generated instruction bodies emit `RegisterAccess` calls rather than direct operand/raw-register access.

A few tests also cover pre-existing behavioral bugs that this work exposed or protects:

- MFMA fast-path source reads now report plugin-visible VGPR reads.
- Public operand chunk reads now observe their VGPR lane window.
- Scalar-backed SIMD operand fallback now uses lane semantics, which matters for true16/f16 inline constants.
- CDNA `v_mac_f16` true16 low-destination writes preserve scalar/SIMD equivalence for high-half clearing behavior.

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/register_access.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/register_access.h
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: MIT
 
 /// @file register_access.h
-/// @brief Instruction-facing register access facade for observed VGPR regions.
+/// @brief Instruction-facing facade for observed AMDGPU register access.
 ///
 /// @details AMDGPU instruction emulation has two competing needs. Most code
 /// reads and writes logical operands, while hot SIMD and matrix paths also need
@@ -11,12 +11,12 @@
 /// by ComputeUnitCore.
 ///
 /// Reads acquired through RegisterAccess notify the execution plugin before
-/// exposing scalar values, SIMD storage, or physical VGPR regions. Write-only
+/// exposing scalar values, SIMD storage, or physical register regions. Write-only
 /// accessors do not report reads. Read-write accessors report the read part at
 /// acquisition time and then allow the caller to update the same instruction-
 /// scoped storage. VM/storage code may still use raw register storage for tasks
 /// such as memory completion, but instruction emulators should use this facade
-/// for operand and physical VGPR access.
+/// for operand and physical register access.
 
 #ifndef ROCJITSU_VM_AMDGPU_REGISTER_ACCESS_H_
 #define ROCJITSU_VM_AMDGPU_REGISTER_ACCESS_H_
@@ -38,11 +38,11 @@
 
 namespace rocjitsu::amdgpu {
 
-/// @brief Facade for instruction-visible VGPR reads and writes.
+/// @brief Facade for instruction-visible register reads and writes.
 ///
 /// @details This class centralizes the observation contract for
-/// instruction-visible VGPR access. Callers should not pair raw storage access
-/// with ad hoc plugin notifications. Instead, they acquire one of two forms of
+/// instruction-visible register access. Callers should not pair raw storage access
+/// with ad hoc plugin notifications. Instead, they acquire one of these forms of
 /// access here:
 ///
 /// - Logical operand access, including scalar/lane/chunk reads and SIMD views
@@ -50,6 +50,8 @@ namespace rocjitsu::amdgpu {
 ///   contiguous VGPR lane storage.
 /// - Physical VGPR regions, used by matrix, memory-address, and other helpers
 ///   that already operate on physical register indices.
+/// - Physical SGPR access, used by address, descriptor, and generated
+///   SGPR-relative helpers.
 ///
 /// API selection guide:
 /// - Use read_scalar(), read_lane(), and the 64-bit variants for
@@ -81,8 +83,8 @@ namespace rocjitsu::amdgpu {
 /// here. They should be acquired during a single instruction's emulation and
 /// not cached across instructions.
 ///
-/// Logical operand APIs require construction from a Wavefront. Physical VGPR
-/// region APIs may be constructed from a ComputeUnitCore; write access requires
+/// Logical operand APIs require construction from a Wavefront. Physical register
+/// APIs may be constructed from a ComputeUnitCore; write access requires
 /// a mutable ComputeUnitCore.
 class RegisterAccess {
   template <typename T>
@@ -333,7 +335,7 @@ public:
     OperandReadWrite64View(const Operand &op, Wavefront &wf, VgprStoragePair64 storage)
         : op_(&op), wf_(&wf), storage_(storage) {
       if (!storage_.lo)
-        scalar_fallback_.emplace(op.read_scalar64(wf));
+        scalar_fallback_.emplace(op.read_lane64(wf, 0));
     }
 
     uint64_t scalar_fallback() const {
@@ -647,17 +649,24 @@ public:
     return OperandReadPair32View(op, wf, storage);
   }
 
-  OperandWriteView write_operand(const Operand &op, uint64_t /*lane_mask*/) const {
+  OperandWriteView write_operand(const Operand &op, uint64_t lane_mask) const {
+    // Kept at acquisition time so future write-observation plugins can validate
+    // the instruction's destination lane set before writable storage is exposed.
+    (void)lane_mask;
     Wavefront &wf = mutable_wavefront();
     return OperandWriteView(op, wf, op.simd_vgpr_storage_mut(wf));
   }
 
-  OperandWrite64View write_operand64(const Operand &op, uint64_t /*lane_mask*/) const {
+  OperandWrite64View write_operand64(const Operand &op, uint64_t lane_mask) const {
+    // See write_operand().
+    (void)lane_mask;
     Wavefront &wf = mutable_wavefront();
     return OperandWrite64View(op, wf, op.simd_vgpr_storage64_mut(wf));
   }
 
-  OperandWritePair32View write_operand_pair32(const Operand &op, uint64_t /*lane_mask*/) const {
+  OperandWritePair32View write_operand_pair32(const Operand &op, uint64_t lane_mask) const {
+    // See write_operand().
+    (void)lane_mask;
     Wavefront &wf = mutable_wavefront();
     return OperandWritePair32View(op, wf, op.simd_vgpr_storage64_mut(wf));
   }
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/register_access.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/register_access.h
@@ -1,0 +1,776 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: MIT
+
+/// @file register_access.h
+/// @brief Instruction-facing register access facade for observed VGPR regions.
+///
+/// @details AMDGPU instruction emulation has two competing needs. Most code
+/// reads and writes logical operands, while hot SIMD and matrix paths also need
+/// direct lane spans over VGPR storage. This file provides the boundary between
+/// those instruction-visible accesses and the lower-level register files owned
+/// by ComputeUnitCore.
+///
+/// Reads acquired through RegisterAccess notify the execution plugin before
+/// exposing scalar values, SIMD storage, or physical VGPR regions. Write-only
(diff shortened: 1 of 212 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/register_access.h
```diff
@@ -0,0 +1,776 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: MIT
+
+/// @file register_access.h
+/// @brief Instruction-facing register access facade for observed VGPR regions.
```
accauble: Nit: The comments in this files seem to imply VGPR-only but there is some sgpr access (`read_sgpr`).

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/register_access.h
```diff
@@ -0,0 +1,776 @@
...
+          op_->write_lane64(*wf_, lane_base + static_cast<uint32_t>(i), buf[i]);
+    }
+
+  private:
+    friend class RegisterAccess;
+
+    OperandReadWrite64View(const Operand &op, Wavefront &wf, VgprStoragePair64 storage)
+        : op_(&op), wf_(&wf), storage_(storage) {
+      if (!storage_.lo)
+        scalar_fallback_.emplace(op.read_scalar64(wf));
```
accauble: The PR description said "If a SIMD view falls back to broadcasting a scalar-backed operand, it must broadcast the value that read_lane would have produced" and the other ones use `read_lane`. Why is `read_scalar` used here?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/register_access.h
```diff
@@ -0,0 +1,776 @@
...
+  OperandReadPair32View read_operand_pair32(const Operand &op, uint64_t lane_mask,
+                                            uint8_t byte_mask = 0xF) const {
+    const Wavefront &wf = wavefront();
+    ConstVgprStoragePair64 storage = op.simd_vgpr_storage64(wf);
+    if (storage.lo)
+      op.simd_notify_read64(wf, lane_mask, byte_mask);
+    return OperandReadPair32View(op, wf, storage);
+  }
+
+  OperandWriteView write_operand(const Operand &op, uint64_t /*lane_mask*/) const {
```
accauble: What is the `lane_mask` parameter used for here?

accauble (commented):
I think this PR looks good, but my confidence level is medium-low/medium. The PR description really helped...motivated well and described the design and testing well. The overall design changes look right to me. My only concern seems to be that the enforcement of certain polices (like don't access the raw storage) relies heavily on other devs reading your comments. So it seemed more like a "don't accidentally access the raw storage?"

atgutier (commented):
LGTM, but the agents had a few minor issues:

 MINOR (worth addressing, none blocking)

  1. RDNA addr_calc.cpp observes per-lane inside the loop — rdna1/2/3/3_5/4 call
  RegisterAccess(cu).read_vgpr(vbase, lane) inside the lane loop (one plugin notify per lane, up to
  64× per instruction), whereas gfx1250 and the shared headers hoist a single read_vgpr_region(vbase,
  reg_count, exec) before the loop. Functionally correct but non-uniform and materially less efficient
   on the RDNA path. Fix: apply the hoisting pattern to the 5 RDNA files.
  2. Missing [[nodiscard]] throughout register_access.h — every value/view-returning read method lacks
   it, yet a discarded read still fires the observation hook (reports a phantom access, drops the
  data) — a latent bug the compiler could catch, and inconsistent with operand.h in the same PR.
  3. 64-bit DPP-delegate notify asymmetry — simd_notify_read64* delegate while the 32-bit hooks
  intentionally don't; a 64-bit read of a DPP-delegated operand would notify nothing.
  Narrow/likely-unreachable (DPP is 32-bit) but a trap; make the 64-bit hooks non-delegating to match,
   or document the difference.
  4. read_operand default byte_mask = 0xF over-reports sub-dword operands — a 16-bit source touches 2
  bytes but reports 4 unless the caller overrides; conservative for the race detector (no missed
  races) but can produce false-positive conflicts on the untouched half. Consider deriving the default
   from op.size_bits().
  5. Bare RegisterAccess (unqualified) in the DS-permute generator template — compiles only because DS
   ops route to the shared header (inside namespace amdgpu); would fail to resolve if ever emitted
  into a per-arch .cpp. Emit amdgpu::RegisterAccess for robustness.
  6. sema_lower.py D-tag read now references inst.dst{idx} — an intentional latent semantic
  improvement (currently unexercised — zero occurrences in generated output); flagged so it's
  understood as deliberate, not accidental.

  NIT

  OperandReadWrite64View scalar fallback uses read_scalar64 while every sibling view uses read_lane*
  (functionally identical at 64-bit today, but the odd one out); missing type annotation on
  SIMD_VOP2_TERNARY_ACCUMULATE; raw C-arrays vs std::array in the SIMD blit helpers.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/register_access.h
newling: It will be used when I eventually add write-related race detection. The idea is that the lane_mask will be used to choose which lane's registers to check for outstanding operations. For example, if all even lanes do a read, and then all odd lanes do a write, there is no potential race as the sets of lanes do not intersect. 

So currently it's not used for anything, but it will be

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/register_access.h
newling: I'm updating this. Initially the PR was VGPR-only, and the docs are lagging behind the code.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/register_access.h
newling: Good spot, I'm updating this. The PR description is correct, it should be read_lane. (The problem with read_scalar is that it ignores the 'size_bits_' of an operand, and so if 'size_bits_=16' it returns the incorrect (32-bit) value stored. read_lane does not have this issue. It's not great needing to document this, I'll consider a follow-up that makes it easier to just use the correct API without needing to think

newling:
> My only concern seems to be that the enforcement of certain polices (like don't access the raw storage) relies heavily on other devs reading your comments. So it seemed more like a "don't accidentally access the raw storage?"

@accauble the main backdoor that I was concerned about is closed with this PR: a dev implementing the emulation for a new instruction cannot read/write a register without notifying the plugin, unless they change class friendships, member private/public status, or do some weird caching of pointers on the side. The only place (I hope) that reads/writes can be done without notifying the plugin are in more privileged places (within the Operand implementation, or within the core compute unit implementation) but those parts of the code should be touched infrequently. 

I am reviewing the docs now to make this clearer, I think probably I (and agents) have over-documented what is safe/not in places where the compiler already enforces the desired API usage through language constructs.

newling:
> LGTM, but the agents had a few minor issues:

(I had the agent address these in https://github.com/ROCm/rocm-systems/pull/8417/changes/ea94917eb57e08b8d3bc537f3fa8dad1376e4caf, agent's summary below)

 Addressed with code changes:
  - RDNA addr_calc per-lane observation: updated the RDNA1/2/3/3.5/4 addr_calc paths to hoist VGPR reads through `read_vgpr_region(...)` before the lane loop, matching the GFX/shared pattern.
  - Missing `[[nodiscard]]`: added `[[nodiscard]]` to the value/view-returning `RegisterAccess` APIs and view accessors. The one intentional observe-only discard in `mma_exec.h` is now explicit with `(void)`.
  - 64-bit DPP notify asymmetry: removed the delegate forwarding from `simd_notify_read64(...)` so it matches the 32-bit notify behavior.
  - Bare `RegisterAccess` in the DS-permute generator: updated the generator to emit fully-qualified `::rocjitsu::amdgpu::RegisterAccess`; regenerated output is in the generated refresh commit.
  - `OperandReadWrite64View` fallback nit: changed the fallback from `read_scalar64(wf)` to `read_lane64(wf, 0)` in the earlier review-fix commit.

  Addressed with code comments:
  - `read_operand` default byte mask: kept the conservative `0xF` default and documented why. `size_bits()` alone does not tell us high-vs-low sub-dword selection, e.g. true16 op_sel, so deriving a narrower default there could under-report and miss conflicts. Callers with precise byte windows can still pass a
  narrower mask.
  - `sema_lower.py` D-tag read: added a comment clarifying that a D-tagged `instoperand` is intentionally a read of the destination operand’s old value for accumulator/update forms.

  Unaddressed:
  - Missing type annotation on `SIMD_VOP2_TERNARY_ACCUMULATE`: left this out because it is a Python style cleanup unrelated to the register-access boundary.
  - Raw C-arrays vs `std::array` in SIMD blit helpers: left this out because it is a broader style cleanup and not needed for the correctness/performance issues raised here.

accauble (approved):
I believe you sufficiently addressed mine and @atgutier's comments. Thanks!

newling:
Failure in 

The following tests FAILED:
	1414 - Cdna4ToCdna3SimulatedDispatchTest.TritonDynamicMatmulDispatchAndRun (Failed)
	
I will rerun CI to see if this is a flake, and if so create an issue for it

newling:
Rerun with 
 gh run rerun 29593182424 --repo ROCm/rocm-systems --failed
 resulted in a clean run. So it was a flake. Possibly related to https://github.com/ROCm/rocm-systems/issues/8773

PR #8419, opened 2026-07-10
fix(amdsmi): report VRAM total from KFD topology

### LINKED ISSUE #97: [Issue]: amd-smi reports different "size" value for XCD memory size different from the docs on MI300A at TPX and CPX
### Problem Description

Hello
I hope this finds you well.

On `MI300A` in `TPX` and `CPX` modes, when running `amd-smi` command it correctly identifies the number of GPUs and everything is fine.

The problem comes to the `size` column (attribute) that is used to represent the memory size for each XCD (GPU).

[AMD documentation](https://instinct.docs.amd.com/projects/amdgpu-docs/en/latest/gpu-partitioning/mi300a/overview.html) show that in `TPX` mode each XCD (GPU) gets `32GB` but the `amd-smi` command shows `42.66GB`
The same for `CPX` mode, the docs mention `16GB` for each XCD, but the `amd-smi` tool show `21.33GB`

In the case of `SPX` mode, everything matches as it is.

Any idea/help what could be the reason for these differences and why?

Appreciate your precious time.

### Operating System

Linux RHE

### CPU

N/A

### GPU

MI300A

### ROCm Version

6.2.4

### ROCm Component

amdsmi

### Steps to Reproduce

_No response_

_No response_

### Additional Information

_No response_

koushikbillakanti-amd:
## Motivation

Fixes ROCm/amdsmi#97. Multi-partition modes (CPX/DPX/TPX/QPX): sysfs `mem_info_vram_total` splits whole-die memory evenly and ignores reserved memory, disagreeing with the driver's per-partition allocation.
APUs (e.g. gfx1151 / Strix Halo): sysfs reports only the small BIOS VRAM carveout instead of the unified pool the GPU addresses.
Also folds in the APU fix from ROCm/rocm-systems#8475.

## Technical Details

`rsmi_dev_memory_total_get` (VRAM) now sources the total from KFD topology (`mem_banks` `size_in_bytes`) via a pure helper `vram_total_prefer_kfd` (new internal header `rocm_smi/rocm_smi_vram.h`, shared with the test).
It prefers KFD when sysfs is unusable (0/read failure), in a multi-partition mode, or on an APU (KFD total > sysfs, since no APU marker exists — tracked in ROCm/rocm-systems#8476); discrete and SPX keep the sysfs value.
`KFDNode::get_total_memory` caches its `mem_banks` sum, and unit tests in `vram_total_prefer_kfd_test.cc` cover every branch, plus a CHANGELOG entry.

## JIRA ID

JIRA ID: SWDEV-536184

## Test Plan

Built libamd_smi clean under -Wall -Wextra -Wconversion.
Ran on MI300A via ctypes calling amdsmi_get_gpu_memory_total across all partitions, cross-checked against KFD mem_banks/0 size_in_bytes.
Ran the GpuUnit.VramTotal* unit tests covering all branches.

## Test Result

CPX GPUs (0-5): 16,852,725,760 B (~16 GB), matching KFD node size_in_bytes.
SPX GPUs (6-8): 101,116,366,848 B (~94 GB), unchanged/no regression.
GpuUnit.VramTotal*: 7 tests passed.

## Submission Checklist

### DIFF
--- a/projects/amdsmi/CHANGELOG.md
+++ b/projects/amdsmi/CHANGELOG.md
@@ -107,6 +107,10 @@ Full documentation for amd_smi_lib is available at [https://rocm.docs.amd.com/pr
   - `amdsmi_get_gpu_device_cuid` has been added as an API for this upcoming change but will remain disabled until full support from the amdgpu driver is available.
   - The CLI `list` output and GPU selection now report the CUID in place of the UUID when a CUID is available, and fall back to the UUID otherwise.
 
+- **Fixed VRAM total reporting incorrect values in CPX/DPX/TPX/QPX compute-partition modes and on APUs**.  
+  - In multi-partition modes, `amdsmi_get_gpu_memory_total()` / `rsmi_dev_memory_total_get()` reported the whole-device VRAM split evenly across partitions instead of the driver's actual per-partition allocation; on APUs (for example gfx1151 / Strix Halo) they reported only the small BIOS VRAM carveout instead of the unified pool the GPU addresses.
+  - The VRAM total is now sourced from the KFD topology (`mem_banks`) in multi-partition modes, when the sysfs read is unusable, or on APUs where sysfs under-reports the carveout. Discrete and SPX GPUs are unaffected.
+
 ## amd_smi_lib for ROCm 7.14.0
 
 ### Added
--- a/projects/amdsmi/CHANGELOG.md
+++ b/projects/amdsmi/CHANGELOG.md
@@ -123,6 +123,10 @@ Full documentation for amd_smi_lib is available at [https://rocm.docs.amd.com/pr
   - `rev_id` was also assigned from the device-info structure before the `AMDGPU_INFO_DEV_INFO` query populated it. Every such path already returned an error, so this was not observable through a checked return, but the value no longer contradicts the status.
   - `amdsmi_asic_info_t` is now reset through one shared initializer used by every backend, so a field a backend cannot supply keeps its not-supported value rather than a plausible zero.
 
+- **Fixed VRAM total reporting incorrect values in CPX/DPX/TPX/QPX compute-partition modes and on APUs**.  
+  - In multi-partition modes, `amdsmi_get_gpu_memory_total()` / `rsmi_dev_memory_total_get()` reported the whole-device VRAM split evenly across partitions instead of the driver's actual per-partition allocation; on APUs (for example gfx1151 / Strix Halo) they reported only the small BIOS VRAM carveout instead of the unified pool the GPU addresses.
+  - The VRAM total is now sourced from the KFD topology (`mem_banks`) in multi-partition modes, when the sysfs read is unusable, or on APUs where sysfs under-reports the carveout. Discrete and SPX GPUs are unaffected.
+
 ### Upcoming Changes
 
 - **UUIDs will be replaced by CUIDs in an upcoming version**.  
--- a/projects/amdsmi/CMakeLists.txt
+++ b/projects/amdsmi/CMakeLists.txt
@@ -377,6 +377,7 @@ set(CMN_INC_LIST
     "${ROCM_INC_DIR}/rocm_smi_binary_parser.h"
     "${ROCM_INC_DIR}/rocm_smi_board_temp.h"
     "${ROCM_INC_DIR}/rocm_smi_npm.h"
+    "${ROCM_INC_DIR}/rocm_smi_vram.h"
 )
 
 add_subdirectory("rocm_smi")
--- a/projects/amdsmi/CMakeLists.txt
+++ b/projects/amdsmi/CMakeLists.txt
@@ -391,6 +391,7 @@ set(CMN_INC_LIST
     "${ROCM_INC_DIR}/rocm_smi_binary_parser.h"
     "${ROCM_INC_DIR}/rocm_smi_board_temp.h"
     "${ROCM_INC_DIR}/rocm_smi_npm.h"
+    "${ROCM_INC_DIR}/rocm_smi_vram.h"
 )
 
 add_subdirectory("rocm_smi")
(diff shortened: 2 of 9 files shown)


Lafunamor:
This PR fixes the case where sysfs **over**-reports the VRAM total. There is a **mirror-image case in the same function** that the current condition does not cover: on APUs, sysfs **under**-reports.

I initially opened #8475 as a standalone fix against `develop`, but carrying it separately would mean a second decision helper sitting next to `vram_total_prefer_kfd()` — the wrong shape, since both cases are really one decision. I've closed it in favour of folding the APU case in here. Details below in case they're useful.

### The APU case

On a Ryzen AI MAX+ 395 (Radeon 8060S, **gfx1151** / Strix Halo), ROCm 7.2.4, amd_smi 26.2.2, 128 GB unified memory:

| source | total |
|---|---|
| `amdsmi_get_gpu_memory_total(VRAM)` | **536,870,912** (0.50 GiB) |
| `amdsmi_get_gpu_memory_total(GTT)` | 118,111,600,640 (110.00 GiB) |
| KFD `mem_banks/0` `size_in_bytes` | 118,111,600,640 (110.00 GiB) |
| HIP `totalGlobalMem` | 118,111,600,640 (110.00 GiB) |

sysfs `mem_info_vram_total` reports the **512 MiB BIOS VRAM carveout**, while KFD `mem_banks` already reports the true unified size — the same value HIP reports as the device total. Consumers that size a memory budget from the VRAM total therefore see 0.5 GiB on a 110 GiB device (this is how I hit it: vLLM's ROCm platform calls `amdsmi_get_gpu_memory_total(..., VRAM)` in `get_device_total_memory()`).

Tracing an APU through the new helper:

- `sysfs_read_ok` = `true`
- `sysfs_total` = `536870912` (not 0 — a "valid" carveout)
- `compute_partition` = not CPX/DPX/TPX/QPX (APUs have no compute partitions)

→ returns `false`, so KFD is never consulted and the carveout is kept.

### Suggested extension

Both cases reduce to the same principle — *KFD topology is the driver's truth for VRAM total* — so one extra clause covers both:

```cpp
bool vram_total_prefer_kfd(bool sysfs_read_ok, uint64_t sysfs_total,
                           const std::string& compute_partition,
                           uint64_t kfd_total) {
  if (!sysfs_read_ok || sysfs_total == 0) {
    return true;
  }
  if (compute_partition == "CPX" || compute_partition == "DPX" ||
      compute_partition == "TPX" || compute_partition == "QPX") {
    return true;
  }
  // APUs (e.g. gfx1151 / Strix Halo): sysfs mem_info_vram_total exposes only the
  // small dedicated BIOS VRAM carveout, while the GPU addresses the unified pool
  // that KFD mem_banks reports. KFD is authoritative when it reports more.
  return kfd_total > 0 && kfd_total > sysfs_total;
}
```

Discrete GPUs are unaffected: the two sources agree, so the new clause is false and behavior is unchanged. The partition case is unaffected too, since there KFD reports *less* than sysfs.

### One implementation gotcha

To evaluate the new clause you need `kfd_total` **before** the decision, i.e. the KFD lookup moves ahead of the `if`. Note that `GET_DEV_AND_KFDNODE_FROM_INDX` returns `RSMI_INITIALIZATION_ERROR` when no KFD node exists — harmless today because it only runs *after* sysfs has already failed, but once it runs on every VRAM query it would discard an otherwise valid sysfs value. Looking the node up directly avoids that:

```cpp
GET_DEV_FROM_INDX
auto& kfd_nodes = smi.kfd_node_map();
auto kfd_it = kfd_nodes.find(dev->kfd_gpu_id());
uint64_t kfd_total = 0;
if (kfd_it != kfd_nodes.end()) {
  kfd_it->second->get_total_memory(&kfd_total);
}
```

### Test cases

These slot into `memory_partition_total_test.cc` alongside the existing ones:

```cpp
// APU: sysfs reports only the small BIOS VRAM carveout; KFD reports the unified
// pool and must win.
TEST(AmdSmiVramTotalPreferKfdTest, ApuCarveoutPrefersKfd) {
  constexpr uint64_t kCarveout = 512ULL * 1024 * 1024;          // 512 MiB
  constexpr uint64_t kUnified  = 110ULL * 1024 * 1024 * 1024;   // 110 GiB
  EXPECT_TRUE(amd::smi::vram_total_prefer_kfd(true, kCarveout, "", kUnified));
  EXPECT_TRUE(amd::smi::vram_total_prefer_kfd(true, kCarveout, "SPX", kUnified));
}

// Discrete GPU: sysfs and KFD agree, so the sysfs value is kept.
TEST(AmdSmiVramTotalPreferKfdTest, DiscreteAgreeingSourcesKeepsSysfs) {
  EXPECT_FALSE(amd::smi::vram_total_prefer_kfd(true, kSampleVramTotal, "SPX", kSampleVramTotal));
}
```

Related: #8476 (APUs are not identifiable through amdsmi — `processor_type` returns `AMD_GPU`, never `AMD_APU`, which is why this has to be inferred from the memory totals at all).

*Disclosure: this analysis was prepared with AI assistance (Claude, Anthropic); the measurements are from my own hardware and I've reviewed the content.*

gabrpham (approved):

yalmusaf (approved):

adam360x (approved):

adam360x (approved):

marifamd (changes requested):
need more information on when we fallback to sysfs and if transitioning to using KFD entirely is correct

adam360x:
@koushikbillakanti-amd most of this PR is outdated and already in develop. Only **prefer-KFD** logic for multi-partition (CPX/DPX/TPX/QPX) and APU cases is still unique. Please update.

adam360x (approved):

AndreasKaratzas:
@koushikbillakanti-amd can we accelerate the merging of this PR? It is critical for CI clusters that are slowly transposing to DPX.

gabrpham:
Converted to draft for now since we're still discussing other avenues for solutions for the time being.

PR #8422, opened 2026-07-10
test(rocprofiler-compute): Enable test_pc_sampling with streaming output

vedithal-amd:
## Motivation

Enable `test_pc_sampling` in TheRock CI and make it debuggable when a live-GPU run stalls on a runner. The test drives the profiler as a subprocess through the fixture, and `subprocess.run` only surfaces captured output after the child exits. If collection hangs, the child never exits, so nothing streams and CTest reports `***Timeout` with zero output and no clue where it stuck. Streaming the profiler output line by line makes the last stage before a stall visible.

## Technical Details

- Add `common.run_subprocess`, a `subprocess.run`-style helper that can optionally tee the child's stdout and stderr live (one reader thread per pipe) while still capturing them separately, returning a `subprocess.CompletedProcess`.
- Route the profile fixture's binary and non-binary subprocess paths through the helper and add a `stream` flag to the fixture.
- Enable `stream=True` on the `test_pc_sampling` profiling calls so `-VVV` output streams under `pytest -s`.
- Re-enable `test_pc_sampling` by removing it from `therock_ci_exclude` in all test categories.

## JIRA ID

JIRA ID: TEST-1234

## Test Plan

Run `test_pc_sampling` locally on a live-GPU host and confirm the profiler `-VVV` output streams live line by line; on a stall the last streamed line pinpoints the stuck stage.

## Test Result

Ran `test_pc_sampling` locally on gfx94X; passes with full live `-VVV` output streaming.

## Submission Checklist


projects/rocprofiler-compute/tests/common.py
```diff
@@ -238,6 +239,59 @@ def strip_ansi(s: str) -> str:
...
+) -> subprocess.CompletedProcess:
+    """Run command and return a CompletedProcess.
+
+    capture_output: capture stdout and stderr onto the returned object.
+    stream: echo output line by line as the child produces it (requires
+        capture_output); otherwise captured output is printed once at the end.
+    text: decode output as text rather than bytes.
+    """
+    if not capture_output:
+        return subprocess.run(command, text=text)
```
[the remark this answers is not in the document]
vedithal-amd: Removed the text flag; run_subprocess is now text-only, so the bytes-mode issue is gone. The stream/capture_output coupling is not a silent loss: with capture_output=False there is no pipe, so the child writes to the terminal live already and stream has nothing to defer. It only affects presentation of captured output.

projects/rocprofiler-compute/tests/conftest.py
```diff
@@ -262,6 +263,7 @@ def _handler(
         workload_dir_type="output_directory",
         num_ranks=1,
         capture_output=False,
+        stream=False,
```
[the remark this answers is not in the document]
vedithal-amd: Done, added a stream entry to the _handler docstring noting it requires capture_output.

projects/rocprofiler-compute/tests/common.py
```diff
@@ -238,6 +239,59 @@ def strip_ansi(s: str) -> str:
     return ansi_escape.sub("", s)
 
 
+def _tee(pipe, sink, out) -> None:
+    """Echo each line from pipe to sink while accumulating it in out."""
+    with pipe:
+        for line in pipe:
+            print(line, end="", file=sink)
```
abchoudh-amd: CTest runs `pytest -s` through a non-TTY pipe, so `print(line, end="", file=sink)` remains block-buffered. If profiling hangs, the last buffered diagnostics may never reach CTest before it kills pytest; adding `flush=True` makes it immediately visible

projects/rocprofiler-compute/tests/test_pc_sampling.py
```diff
@@ -246,6 +250,7 @@ def test_pc_sampling_profile_then_analyze(
         options,
         check_success=False,
         capture_output=True,
+        stream=True,
```
abchoudh-amd: The test request capsys. Fixture capture takes precedence even with `pytest -s`, so _tee writes into pytest’s in-memory capture objects rather than CTest-visible streams.

Please wrap profiling calls `with capsys.disabled`

projects/rocprofiler-compute/tests/test_pc_sampling.py
```diff
@@ -339,6 +344,7 @@ def test_pc_sampling_with_sol_block(
...
```
abchoudh-amd: Same comment as above

abchoudh-amd (changes requested):

vedithal-amd:
Temporarily disabled streaming in test_pc_sampling to see if that was the root cause of Timeout

projects/rocprofiler-compute/tests/common.py
vedithal-amd: Good catch, this will ensure we dont loose the tail end of the logs in ctest stdout near the hang

vedithal-amd (commented):
I have enabled stream=True since stream=False wasn't the culprit causing the test to hang it was some underlying ROCr issue

projects/rocprofiler-compute/tests/common.py
vedithal-amd: Good catch. Added flush=True in _tee so the tail-end lines near a hang reach ctest stdout instead of dying in the block buffer when pytest is killed on timeout.

projects/rocprofiler-compute/tests/test_pc_sampling.py
vedithal-amd: Fixed at the fixture level instead of per-test: run_subprocess now tees to dup(1)/dup(2) rather than sys.stdout/sys.stderr, so it writes past capsys CaptureIO and streams live even when the test requests capsys for the analyze assertions. No capsys.disabled() needed in the test body.

projects/rocprofiler-compute/tests/test_pc_sampling.py
vedithal-amd: Same fix as above: run_subprocess tees to the real fds via dup(1)/dup(2), so streaming is unaffected by capsys being active in this test.

abchoudh-amd (approved):

PR #8426, opened 2026-07-10
fix(rocjitsu): Enable Field-less Operands in amdisa

accauble:
## Motivation

The amdisa parser did not include fieldless operands (ones that are implicit or hardcoded into the instruction). The generated C++ still accounted for the hidden side effects, (see `set_vcc()`) but missing operands makes it difficult for rocjitsu analyses to use them (e.g. liveness analysis).

## Technical Details

This PR, adds these fieldless operands to the operand set so that we can start to use them in rocjitsu. I tried my best to contain the changes so this does not change the hardcoded effects like setting VCC. The exception to this is the OPR_SIMM32 operand, for which parser had some hacks to generate the pertinent code (in one case it even just added it in as an Operand).

Follow-Up PRs would include:
- Implement VCC/EXEC/SCC/M0/PC side-effects from operand presence rather than hard-coding them
- Include field-less operands in cross-ISA structural signatures
- Include field-less (explicit) operands in disassembly
- This is described in more detail now in Issue #8625.

Unfortunately, this still includes a lot of changes to C++. I can try to split it up more to make this easier to review. Some options:
- The OPR_SIMM32 operand was added "manually" before this change. I could revert this and just exclude that from the `operands` structure until later.
- Could probably do the same with the OPR_VGPR which adds quite a few `vaddr` Operands to the C++ code.
- Open to other suggestions!

## Test Plan

Added tests for the functionality I added.

## Test Result

All tests pass.

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/isa_operand_simd_inl.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/isa_operand_simd_inl.h
@@ -39,6 +39,13 @@ std::optional<uint32_t> resolved_vgpr_offset_for_operand(const amdgpu::Wavefront
 template <typename Isa> bool AmdgpuIsaOperand<Isa>::simd_capable() const {
   if (this->delegate())
     return this->delegate()->simd_capable();
+  // Placeholder/metadata field-less operands (e.g. image vaddr) are inert and
+  // must not inherit VGPR/immediate SIMD-capability from their operand type;
+  // value-bearing field-less operands are exempt. This is a hand-written mirror
+  // of CodeGenerator._VALUE_BEARING_FIELDLESS_TYPES -- keep the exempt type(s)
+  // here in sync with that set (a pytest golden-locks it).
+  if (this->is_field_less() && this->opr_type_ != Isa::OperandType::OPR_SIMM32)
+    return false;
   return Isa::simd_capable_value(this->opr_type_, this->encoding_value_);
 }
 
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/operand.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/operand.h
@@ -84,7 +84,12 @@ public:
   /// @brief Whether this operand references a VGPR or AccVGPR.
   /// @details Classified at construction time by ISA-specific subclasses using
   /// the auto-generated is_vgpr_operand_type() from operand_types.h.
-  [[nodiscard]] bool is_vgpr() const { return is_vgpr_; }
+  /// TODO: field-less operands are currently inert. Update this as they
+  /// switch over to being used. (Stay in sync with
+  /// CodeGenerator._VALUE_BEARING_FIELDLESS_TYPES). Currently, the only
+  /// operands in this are non-VGPR so gating on field_less_ alone is
+  /// sufficient here.
+  [[nodiscard]] bool is_vgpr() const { return !field_less_ && is_vgpr_; }
 
   /// @brief Whether this operand is field-less
   /// @details Field-less operands (has no encoding field in MR ISA) are
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/operand.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/operand.h
@@ -86,6 +86,15 @@ public:
   /// the auto-generated is_vgpr_operand_type() from operand_types.h.
   [[nodiscard]] bool is_vgpr() const { return is_vgpr_; }
 
+  /// @brief Whether this operand is field-less
+  /// @details Field-less operands (has no encoding field in MR ISA) are
+  /// constructed from a fixed canonical encoding value rather than a decoded
+  /// field.
+  [[nodiscard]] bool is_field_less() const { return field_less_; }
+
+  /// @brief Mark this operand as field-less. Set by generated constructors.
+  void set_field_less(bool value = true) { field_less_ = value; }
+
   /// @brief Assign the GFX12 VGPR high-bank role for this operand.
   void set_vgpr_msb_role(amdgpu::VgprMsbRole role) { vgpr_msb_role_ = role; }
 
@@ -207,6 +216,9 @@ public:
   int size_bits_ = 0;
   int encoding_value_ = 0;
   bool is_vgpr_ = false;
+  /// @brief True if this operand has no MR ISA encoding field (see
+  /// is_field_less()). Suppressed from disassembly.
+  bool field_less_ = false;
   amdgpu::VgprMsbRole vgpr_msb_role_ = amdgpu::VgprMsbRole::None;
 
 private:
--- a/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
+++ b/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
@@ -473,18 +473,52 @@ class CodeGenerator:
         ordered.extend(op for op in src_operands if op.name not in ordered_names)
         return ordered
 
-    @staticmethod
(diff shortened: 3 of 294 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/operand.h
```diff
@@ -86,6 +86,15 @@ class Operand {
   /// the auto-generated is_vgpr_operand_type() from operand_types.h.
   [[nodiscard]] bool is_vgpr() const { return is_vgpr_; }
```
kuhar: This would be safer if field-less operands did not retain register and SIMD behavior from their operand type by accident. For example, generated RDNA4 image instructions construct field-less `vaddr` as `OPR_VGPR` with canonical value 0: `to_register_ref()` returns `nullopt`, but `is_vgpr()`, `simd_capable()`, and `read_lane()` still classify or read it as `v0`. Please make the semantic split explicit: placeholder or metadata operands such as image `vaddr` should be inert across all register and SIMD APIs, while hardwired architectural operands such as VCC, EXEC, and SCC can use a distinct policy when special-register def/use is supported. A focused test for the field-less `vaddr` queries would lock this down.

kuhar (commented):
Just one comment

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/operand.h
accauble: Fair point. To keep with the rest of the PR (as few changes to C++ as possible), I changed these to be inert for all but the OPR_SIMM32 operand which is not actually inert since it was too hacky to keep the manual OPR_SIMM32 updates. I have put comments to remind future devs (presumably me) that they need to update these functions when make the field-less operands actually do something.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/isa_operand_simd_inl.h
```diff
@@ -39,6 +39,13 @@ std::optional<uint32_t> resolved_vgpr_offset_for_operand(const amdgpu::Wavefront
 template <typename Isa> bool AmdgpuIsaOperand<Isa>::simd_capable() const {
   if (this->delegate())
     return this->delegate()->simd_capable();
+  // Placeholder/metadata field-less operands (e.g. image vaddr) are inert and
+  // must not inherit VGPR/immediate SIMD-capability from their operand type;
+  // value-bearing field-less operands are exempt. This is a hand-written mirror
+  // of CodeGenerator._VALUE_BEARING_FIELDLESS_TYPES -- keep the exempt type(s)
+  // here in sync with that set (a pytest golden-locks it).
+  if (this->is_field_less() && this->opr_type_ != Isa::OperandType::OPR_SIMM32)
```
kuhar, line 47: This might be easier to maintain if the hand-written `simd_capable()` and `is_vgpr()` guards were generated from `_fieldless_inert_cpp_cond()` or checked against it. The current golden tests validate only the Python set and generated condition; they never inspect these C++ mirrors, so adding another value-bearing type could update generated reads while leaving the hard-coded `!= OPR_SIMM32` guards stale. Please derive these guards from the same source or add a test that locks the committed C++ expressions to it.

kuhar (commented):
A few comments

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/operand.h
kuhar: This would be safer if every read, write, and storage path applied the same field-less guard before resolving a register. On the current head, field-less RDNA4 `vaddr` still reaches `write_lane()` and the chunk helpers as `v0`, while field-less EXEC destinations still reach `write_scalar*()`, even though `is_vgpr()` and `simd_capable()` reject them. Please emit the shared guard into the write accessors, apply it in the common chunk/storage helpers, and extend `PlaceholderVaddrInertButSimm32StaysValueBearing` to assert guarded reads return zero and representative write/chunk calls leave seeded `v0` and EXEC values unchanged.

atgutier (commented):
From the agent:

 BLOCKER — must fix

  s_setreg_imm32_b32 on gfx1250 now writes the constant 0 instead of the instruction's 32-bit
  immediate — gfx1250/sopk.cpp:238. The execute body was migrated from reading literal_ (the real
  literal) to a new Operand literal(32, OPR_SIMM32, 0) that is never repopulated from literal_ and
  never marked field-less — so read_scalar returns encoding_value_ = 0, and the register write applies
   its mask over a zero source. Every other ISA generates the correct pattern (populate from literal_
  + set_field_less(true), e.g. cdna4 sopk.cpp:410/422); only the gfx1250 generator branch diverges
  (operand named literal, populate step omitted). This directly contradicts the behavior-preservation
  claim and must be fixed in the generator (not hand-edited) then all ISAs re-run, with a
  decode+execute test asserting a non-zero immediate is written.

  MAJOR

  Literal-at-src_operands_[1] breaks the positional SDWA/DPP src1 invariant for
  v_fmaak/v_fmamk/v_madak/v_madmk — both reviewers found this independently (generator side:
  _generator.py SDWA/DPP block; emitted side: cdna4 vop2.cpp:3296, rdna4 vop2.cpp:4466, etc.). For
  these inline-literal VOP2s the field-less simm32 is inserted at src_operands_[1], pushing the real
  vsrc1 to [2]. But the SDWA/DPP boilerplate still hard-codes src_operands_[1] as "the second VGPR
  source" — so if that path runs it computes a VGPR base from the literal value (garbage/out-of-range
  read) and builds the src1 delegate from the wrong operand. The PR carefully de-positionalized the
  execute body and the delegate targeting, but left this one raw-array index. Latent, likely
  unreachable (FMAAK/FMAMK consume the literal dword, so an SDWA/DPP-encoded form is generally not a
  legal encoding) — which is why it's MAJOR not BLOCKER, but it's exactly the bug class the PR set out
   to avoid, and it becomes a BLOCKER if the decoder can ever set src0 == SRC_SDWA/SRC_DPP for these
  opcodes (including on malformed input — please confirm reachability). Fix in the generator: either
  append literals at the end of the src list (preserving the "implicits appended" invariant that holds
   everywhere else), or have the SDWA/DPP block reference src1 by named member / a generated index
  constant instead of [1]. A build-time assert ("no field-less input before a DPP/SDWA-permutable
  vector source") would have caught both this and gated future regressions.

  MINOR

  - OPR_SIMM32 value-bearing exception is triplicated as hand-coded != OPR_SIMM32 checks
  (operand.h:88, isa_operand_simd_inl.h:46, per-ISA read_*) coupled to _generator.py's
  _VALUE_BEARING_FIELDLESS_TYPES only by a comment + golden test. A future field-less immediate of a
  different type silently reads 0. Single-source it (generated constant/list consumed by both C++ and
  the test).
  - Write paths aren't gated on field_less_ (operand.h read paths are inert-guarded, write_* aren't) —
   inert today (no execute body writes through a field-less operand) but an asymmetric latent hazard;
  mirror the read guard or assert(!field_less_).
  - New Instruction query properties (implicit_operands/src_operands/etc. in gpuisa.py) have no
  in-tree consumer yet (staged for #8625) and src_operands/dst_operands include field-less operands
  (unlike the execute-visible set) — a foot-gun; defer or document.
  - Lazy _fieldless_canon_cache not declared in __init__ (style).

 NIT

  Generated member exec shadows the common uint64_t exec = wf.exec() local in cmpx bodies (harmless,
  -Wshadow not enabled, but a generated-code smell — and cmpx now models EXEC twice via exec +
  sdst_exec, both inert; consider de-duplicating).

accauble:
Thanks for the reviews, @atgutier and @kuhar . I found the gfx1250 s_setreg_imm32_b32 bug I introduced while completely refactoring the inertness gating. I originally intended to make that refactor a follow-up PR since I thought that would be easier to review. But, given that this branch still has problems and the refactor combined with this code is so much cleaner, I am closing this this and opening a new PR.

PR #8427, opened 2026-07-10
fix(amdsmi): fall back across libdrm_amdgpu sonames

marifamd:
## Motivation

On ROCm tarball / TheRock installs the versioned `libdrm_amdgpu.so.1` may be absent, and libdrm is vendored under a renamed soname (`librocm_sysdeps_drm_amdgpu.so.1`). amd-smi loaded libdrm via a single `dlopen` of the build-time `LIBDRM_AMDGPU_SONAME` with no fallback, so the load failed and the DRM query path degraded: `asic`/`vbios`/`driver` info returned `NOT_SUPPORTED` and the BDF rendered as garbage, with a `Fail to open libdrm_amdgpu.so.1` warning on stderr.

## Technical Details

**Loader** (`include/amd_smi/impl/amd_smi_lib_loader.h`, `src/amd_smi/amd_smi_lib_loader.cc`):
- Add a `load(const std::vector<const char*>&)` overload that tries each candidate in order and returns success on the first that opens; the single-name `load()` delegates to it
- Add a centralized `libdrm_amdgpu_sonames()` helper: build-time soname, unversioned `libdrm_amdgpu.so`, then the vendored `librocm_sysdeps_drm_amdgpu.so.1`
- Log the failure only once all candidates fail, so a successful fallback stays silent
- Keep the `RTLD_NOLOAD` handle when a candidate is already loaded so `load_symbol()` resolves and `unload()` balances the refcount

**Call sites** (`src/amd_smi/amd_smi.cc`, `amd_smi_drm.cc`, `amd_smi_utils.cc`):
- Switch every libdrm load to the candidate list via `libdrm_amdgpu_sonames()`

**Tests** (`tests/amd_smi_test/functional/lib_loader_read.cc`):
- New GPU-free loader tests: first-candidate-wins, fallback-on-missing-primary, all-missing-fails, and already-loaded-resolves-symbol

## JIRA ID

JIRA ID: AILITOOLS-290

## Test Plan

- C++ loader unit tests (`amdsmitst --gtest_filter='LibLoaderFallback.*'`)
- Hardware verification on a Helios MI455X box (ROCm 7.15): moved the system `libdrm_amdgpu.so.1` out of all ldconfig-scanned dirs to reproduce, then compared stock vs patched library across `asic`/`vbios`/`driver`/`bdf` fields

## Test Result

- All 4 loader unit tests pass; pre-commit (clang-format) green
- With libdrm removed: stock library returns `NOT_SUPPORTED` for asic/vbios/driver and a garbled BDF plus the stderr warning; the patched library resolves all fields correctly and silently via the vendored `libdrm_amdgpu.so` fallback

## Submission Checklist

- [x] Code builds
- [x] Unit tests added and passing
- [x] Verified on hardware


dmitrii-galantsev (approved):

projects/amdsmi/src/amd_smi/amd_smi.cc
```diff
@@ -1658,7 +1658,7 @@ amdsmi_status_t amdsmi_get_gpu_vram_usage(amdsmi_processor_handle processor_hand
   }
 
   amd::smi::AMDSmiLibraryLoader libdrm;
-  amdsmi_status_t status = libdrm.load(LIBDRM_AMDGPU_SONAME);
+  amdsmi_status_t status = libdrm.load(amd::smi::libdrm_amdgpu_sonames());
   if (status != AMDSMI_STATUS_SUCCESS) {
     libdrm.unload();
     ss << __PRETTY_FUNCTION__ << " | Failed to load " LIBDRM_AMDGPU_SONAME ": " << strerror(errno)
```
gabrpham: LIBDRM_AMDGPU_SONAME here is incorrect now since we get the name from amd::smi::libdrm_amdgpu_sonames() rather than from the defines in amd_smi_config.h. Also, dlopen doesn't reliably set errno, so we should instead use dlerror.

gabrpham (changes requested):
Comment below also applies to all the other error output lines in that same file.

projects/amdsmi/src/amd_smi/amd_smi.cc
marifamd: fixed

gabrpham (approved):
Comment has been addressed. Approved.

PR #8429, opened 2026-07-10
[rocprofiler-sdk] address several hangs on bad input parameters

samcrater-amd:
## Motivation

Several different invalid parameter configurations would cause rocprof to abort and hang instead of exiting with an error message.

## Technical Details

The signal handler uses a call_once(). If the signal handler ends up recursively calling itself, the second instance will cause call_once() to hang, since the thread will be waiting on the previous instance to complete (which is itself). To avoid this, the signal handler checks if the current thread is already executing the signal handler and calls _exit() if so.
Additionally, added a variant of the ROCPROFILER_CALL macro that logs an error message and cleanly exits when the error is a bad parameter, added some error messages in the parameter checking code, and added a couple ATT parameter tests.

## JIRA ID

JIRA ID : AIPROFSDK-912.

## Test Plan

Profiled test app with parameter configurations that previously caused hangs.

## Test Result

Profiler exited cleanly with an error message.

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/helper.hpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/helper.hpp
@@ -1,6 +1,6 @@
 // MIT License
 //
-// Copyright (c) 2023-2025 Advanced Micro Devices, Inc. All rights reserved.
+// Copyright (c) 2023-2026 Advanced Micro Devices, Inc. All rights reserved.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/helper.hpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/helper.hpp
@@ -81,27 +81,20 @@
         }                                                                                          \
     }
 
-// Like ROCPROFILER_CALL, but intended for service-configuration calls that are
-// driven directly by user-supplied command-line options (e.g. ATT and
-// PC-sampling parameters). A failure here is almost always caused by an invalid
-// user parameter, so it must be reported as a clean, user-readable error with a
-// non-zero exit instead of aborting via a fatal log. Aborting would raise
-// SIGABRT and route through the process error/signal handler, which is both
-// hostile to users (no actionable message) and has historically deadlocked the
-// tool. ROCPROFILER_STATUS_ERROR_INVALID_ARGUMENT is treated as a user error;
-// any other unexpected status is still escalated as fatal for diagnosis.
+// Like ROCPROFILER_CALL, but reports bad arguments with ERROR instead of FATAL and informs the user
+// that there is a bad parameter.
 #define ROCPROFILER_CHECK_USER_CONFIG(result, msg)                                                 \
     {                                                                                              \
         rocprofiler_status_t ROCPROFILER_VARIABLE(USERCFGSTATUS, __LINE__) = result;               \
-        if(ROCPROFILER_VARIABLE(USERCFGSTATUS, __LINE__) ==                                         \
-           ROCPROFILER_STATUS_ERROR_INVALID_ARGUMENT)                                               \
+        if(ROCPROFILER_VARIABLE(USERCFGSTATUS, __LINE__) ==                                        \
+           ROCPROFILER_STATUS_ERROR_INVALID_ARGUMENT)                                              \
         {                                                                                          \
-            ROCP_ERROR << msg                                                                       \
+            ROCP_ERROR << msg                                                                      \
                        << " failed: one or more parameters are invalid. Please check the "         \
                           "corresponding command-line options and try again.";                     \
-            ::std::exit(EXIT_FAILURE);                                                              \
+            ::std::exit(EXIT_FAILURE);                                                             \
         }                                                                                          \
-        else if(ROCPROFILER_VARIABLE(USERCFGSTATUS, __LINE__) != ROCPROFILER_STATUS_SUCCESS)        \
+        else if(ROCPROFILER_VARIABLE(USERCFGSTATUS, __LINE__) != ROCPROFILER_STATUS_SUCCESS)       \
         {                                                                                          \
             std::string status_msg =                                                               \
                 rocprofiler_get_status_string(ROCPROFILER_VARIABLE(USERCFGSTATUS, __LINE__));      \
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/helper.hpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/helper.hpp
@@ -57,6 +57,7 @@
 #include <sys/types.h>
 #include <unistd.h>
 #include <cstdint>
+#include <cstdlib>
 #include <fstream>
 #include <iostream>
 #include <map>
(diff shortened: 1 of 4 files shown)


projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
```diff
@@ -102,24 +102,41 @@ thread_trace_callback(uint32_t shader, void* buffer, uint64_t size, void* callba
...
 thread_trace_parameter_pack::are_params_valid() const
 {
-    // Guard against the most common misconfigurations before touching HSA
-    // state so we can fail early with a descriptive message.
+    // Guard against the most common misconfigurations before touching HSA state
+    // so we can fail early with a descriptive message.
     if(shader_cb_fn == nullptr)
     {
-        ROCP_CI_LOG(WARNING) << "Callback cannot be null!";
+        ROCP_ERROR << "Thread trace shader callback cannot be null";
```
vlaindic: I'm not familiar whether this is recoverable or not. If recoverable, ROCP_WARNING might be more appropriate I guess.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/helper.hpp
```diff
@@ -80,6 +81,29 @@
...
+#define ROCPROFILER_CHECK_USER_CONFIG(result, msg)                                                 \
+    {                                                                                              \
+        rocprofiler_status_t ROCPROFILER_VARIABLE(USERCFGSTATUS, __LINE__) = result;               \
+        if(ROCPROFILER_VARIABLE(USERCFGSTATUS, __LINE__) ==                                        \
+           ROCPROFILER_STATUS_ERROR_INVALID_ARGUMENT)                                              \
+        {                                                                                          \
+            ROCP_ERROR << msg                                                                      \
+                       << " failed: one or more parameters are invalid. Please check the "         \
+                          "corresponding command-line options and try again.";                     \
+            ::std::exit(EXIT_FAILURE);                                                             \
```
vlaindic, line 95: I don't mind this this, but just want to confirm that even though we implemented a guard for checking which thread executes the signal handler, this is still required, correct?

vlaindic (changes requested):
Couple minor comments to better understand all edge cases + AI generated review comments:

### Finding 1 — `ROCP_ERROR` severity too high in library-layer `are_params_valid()` [`thread_trace/core.cpp:109`]

**[IMPORTANT]**

`are_params_valid()` lives in the library layer (`source/lib/rocprofiler-sdk/`). Its caller (`thread_trace/service.cpp`) handles the `false` return and propagates `ROCPROFILER_STATUS_ERROR_INVALID_ARGUMENT` back to the client — this is a recoverable, API-boundary condition, not an internal error.

`ROCP_ERROR` here is too loud for third-party library clients that handle errors themselves or probe configs speculatively. The user-facing message belongs at the tool layer (`ROCPROFILER_CHECK_USER_CONFIG` already does this). The library should use warning severity:

```cpp
ROCP_CI_LOG(WARNING) << "Thread trace shader callback cannot be null";
```

Apply the same change to all five checks in this function.

---

### Finding 2 — `_owner_tid` stored inside `call_once` leaves a gap before the guard activates [`tool.cpp:3926`]

**[IMPORTANT]**

The re-entrancy guard checks `_owner_tid` at the top of the handler, but `_owner_tid` is only stored inside the `call_once` lambda. The `ROCP_WARNING` / `fmt::format` call runs *before* `call_once` starts — before `_owner_tid` is set. A fault on the same thread during that logging call would not be caught by the guard (`_owner_tid` still `-1`), and `call_once` would deadlock as before.

Fix: move `_owner_tid.store(_self_tid)` to just before the `ROCP_WARNING`, outside `call_once`:

```cpp
// Store BEFORE any non-async-signal-safe work
_owner_tid.store(_self_tid, std::memory_order_release);

ROCP_WARNING << fmt::format(...);

std::call_once(_once, [&]() {
    // remove the store from here
    ...
});
```

Note: this guard covers same-thread re-entry only. A different thread receiving the same signal while the first thread is inside `call_once` can still block — known residual, worth a follow-up ticket rather than expanding this PR's scope.

---

### Finding 3 — No automated regression test for the four fixed hang scenarios

**[IMPORTANT]**

Test Plan covers manual verification only. For a P1 fix with four distinct repro cases, there should be an automated guard. All four are testable: invoke `rocprofv3` with each bad config, assert non-zero exit and expected stderr substring. The existing `tests/rocprofv3/` ATT / PC-sampling suite already has scaffolding for this (CTest `WILL_FAIL ON`, stderr matching).

At minimum, extend that suite with the two ATT cases (`--att-target-cu 99999`, `--att-buffer-size 4096`).

---

### Finding 4 — Copyright year stale in two substantially modified files

**[IMPORTANT]**

- `helper.hpp`: reads `2023-2025` — this PR adds 24 lines. Bump to `2023-2026`.
- `thread_trace/core.cpp`: reads `2024-2025` — this PR rewrites `are_params_valid()`. Bump to `2024-2026`.

(`tool.cpp` already reads `2023-2026` — correct.)

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/helper.hpp
samcrater-amd: Right, the signal handler fix just prevents hanging. This is still needed to inform the user and gracefully exit.

ApoKalipse-V (commented):
I'm not sure about these changes in thread trace. We have CI_LOG(WARNING) to ensure these are not tripped in CI and later accidentally ignored in our tests.
Are we sure removing these guards wont cause other tests to accidentally ignore errors?

The tool.cpp changes LGTM

ApoKalipse-V (approved):
LGTM for thread trace but I'd like someone else to sign off on it.
I'm not too familiar with the policies sending warnings on invalid user input from within the SDK.
A better long term approach might be having API return codes specific for these errors.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/helper.hpp
vlaindic: I think it might be worth mentioning why not ROCP_FATAL here, something like

```
  Because ROCP_FATAL likely calls abort() or otherwise raises a signal (SIGABRT), which would re-enter rocprofv3_error_signal_handler — the exact hang this
  PR was fixing. ROCP_ERROR + std::exit() sidesteps that loop entirely.
```

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/tool.cpp
```diff
@@ -4216,7 +4235,13 @@ rocprofv3_error_signal_handler(int signo, siginfo_t* info, void* ucontext)
 
     // below is for testing purposes. re-raising the signal causes CTest to ignore WILL_FAIL ON
     if(signal_handler_exit) ::quick_exit(signo);
+
+    // restore the default disposition before re-raising so that, even if a
+    // chained handler re-installed this handler during cleanup, the re-raise
+    // terminates the process instead of re-entering and looping.
+    ::signal(signo, SIG_DFL);
     ::raise(signo);
+    ::_exit(128 + signo);
```
vlaindic, line 4244: One exotic edge case that we might track for the future.

**Residual race in `_owner_tid` re-entrancy guard [`tool.cpp`]** — the current guard has a window where same-thread re-entry is not detected if a second thread has overwritten `_owner_tid` in between. Full scenario:

   1. Thread A enters handler, stores its TID into `_owner_tid`.
   2. Thread B receives the same signal concurrently, overwrites `_owner_tid` with B's TID, enters `call_once`.
   3. Thread A receives another signal (re-entry), re-enters the handler. Loads `_owner_tid` → sees B's TID, not A's → guard does not fire → thread A proceeds toward `call_once` → **deadlocks on thread B**.

   The guard only protects against re-entry when no other thread has raced in between. In the common case (single-thread re-entry via `ROCP_FATAL` → `abort()` → SIGABRT) it works correctly. The exotic two-thread race above is a residual.

   **Proper fix:** replace the unconditional `store` with a `compare_exchange_strong` — only write `_owner_tid` if it currently holds `-1`. If the CAS fails, another thread is already in the handler; proceed to `call_once` as usual (blocking is acceptable for that case). This eliminates the overwrite window entirely.

   ```cpp
   long expected = -1;
   _owner_tid.compare_exchange_strong(expected, _self_tid,
                                      std::memory_order_acq_rel,
                                      std::memory_order_acquire);
   // re-entry check stays the same: load == _self_tid → guard fires
   ```

   Scope: follow-up ticket rather than expanding this PR.

projects/rocprofiler-sdk/tests/rocprofv3/advanced-thread-trace/CMakeLists.txt
vlaindic, line 1: I think for the future, we might want to add the following cases:

- PC sampling bad-param case
- Missing one of `--att-perfcounters` / `--att-perfcounter-ctrl` (only one flag set)

vlaindic (approved):
Overall, LGTM, so I'll approve. I added couple more non-blocking comments that might be worth addressing or at least tracking for the future. 

As my knowledge about error reporting is limited, I agree with Gio about asking someone else (Jonathon, Ben, Raksha, or Gopesh) to quickly take a glance about the error reporting.

PR #8430, opened 2026-07-10
feat(rocr): Changes to allow FreeBSD compilation

Yohello1:
## Motivation

So far I've been submitting the header changes file by file, which will take a while. So I batched all the simple/easy ones together. This PR is full of all the short/simple header changes to allow FreeBSD compilation & running.

## Technical Details

I made included the correct headers for FreeBSD

## JIRA ID

ISSUE ID: #9999999

## Test Plan

I ran the rocrtst suite, and the results were the same before and after these changes.

My Fluid simulation runs with the same performance. 

## Test Result

They all passed

## Submission Checklist

Note: this PR is also related; [https://github.com/ROCm/rocm-systems/pull/7917](https://github.com/ROCm/rocm-systems/pull/7917)

Sponsored By: The FreeBSD Foundation

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/driver/kfd/amd_kfd_driver.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/driver/kfd/amd_kfd_driver.cpp
@@ -47,6 +47,9 @@
 #include <link.h>
 #include <sys/ioctl.h>
 #include <fcntl.h>
+#elif defined(__FreeBSD__)
+#include <link.h>
+#include <sys/ioctl.h>
 #endif
 
 #include "hsakmt/hsakmt.h"
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/driver/kfd/amd_kfd_driver.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/driver/kfd/amd_kfd_driver.cpp
@@ -42,14 +42,13 @@
 
 #include "core/inc/amd_kfd_driver.h"
 
+#if defined(__linux__) || defined(__FreeBSD__)
+    #include <sys/ioctl.h>
+    #include <link.h>
+#endif
+
 #if defined(__linux__)
-#include <amdgpu_drm.h>
-#include <link.h>
-#include <sys/ioctl.h>
-#include <fcntl.h>
-#elif defined(__FreeBSD__)
-#include <link.h>
-#include <sys/ioctl.h>
+    #include <amdgpu_drm.h>
 #endif
 
 #include "hsakmt/hsakmt.h"
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
@@ -674,28 +674,19 @@ void AqlQueue::CloseRingBufferFD(const char* ring_buf_shm_path, int fd) const {
 
 int AqlQueue::CreateRingBufferFD(const char* ring_buf_shm_path,
                                  uint32_t ring_buf_phys_size_bytes) const {
-
-#ifdef __linux__
+#if defined(__linux__) || defined(__FreeBSD__)
   int fd;
-#ifdef HAVE_MEMFD_CREATE
-  fd = memfd_create(ring_buf_shm_path, 0);
-#else
-  fd = syscall(__NR_memfd_create, ring_buf_shm_path, 0);
-#endif
-  if (fd == -1) return -1;
-#elif defined(__FreeBSD__)
-  int fd;
-  fd = memfd_create(ring_buf_shm_path, 0);
-  if (fd == -1) return -1;
-#endif
 
-  if (fd == -1) return -1;
+  fd = memfd_create(ring_buf_shm_path, 0);
 
-  if (ftruncate(fd, ring_buf_phys_size_bytes) == -1) {
-    CloseRingBufferFD(ring_buf_shm_path, fd);
+  if (fd != -1) {
+    if (ftruncate(fd, ring_buf_phys_size_bytes) == 0) {
+      return fd;
+    }
+    close(fd);
     return -1;
   }
-#else
+
   fd = shm_open(ring_buf_shm_path, O_CREAT | O_RDWR | O_EXCL, S_IRUSR | S_IWUSR);
 
   if (fd == -1) return -1;
@@ -704,10 +695,10 @@ int AqlQueue::CreateRingBufferFD(const char* ring_buf_shm_path,
     CloseRingBufferFD(ring_buf_shm_path, fd);
     return -1;
   }
-#endif
+
   return fd;
 #else
-  assert(false && "Function only needed on Linux.");
+  assert(false && "Function only needed on Linux and FreeBSD.");
   return -1;
 #endif
 }
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
@@ -674,13 +674,19 @@ void AqlQueue::CloseRingBufferFD(const char* ring_buf_shm_path, int fd) const {
 
 int AqlQueue::CreateRingBufferFD(const char* ring_buf_shm_path,
                                  uint32_t ring_buf_phys_size_bytes) const {
+
 #ifdef __linux__
   int fd;
 #ifdef HAVE_MEMFD_CREATE
-#ifdef __FreeBSD__
   fd = memfd_create(ring_buf_shm_path, 0);
 #else
   fd = syscall(__NR_memfd_create, ring_buf_shm_path, 0);
+#endif
+  if (fd == -1) return -1;
+#elif defined(__FreeBSD__)
+  int fd;
+  fd = memfd_create(ring_buf_shm_path, 0);
+  if (fd == -1) return -1;
 #endif
 
   if (fd == -1) return -1;
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
@@ -677,7 +677,11 @@ int AqlQueue::CreateRingBufferFD(const char* ring_buf_shm_path,
 #ifdef __linux__
   int fd;
 #ifdef HAVE_MEMFD_CREATE
+#ifdef __FreeBSD__
+  fd = memfd_create(ring_buf_shm_path, 0);
+#else
   fd = syscall(__NR_memfd_create, ring_buf_shm_path, 0);
+#endif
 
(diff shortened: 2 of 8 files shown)


Yohello1:
@dayatsin-amd I just batched most of the simple header changes

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
```diff
@@ -677,7 +677,11 @@ int AqlQueue::CreateRingBufferFD(const char* ring_buf_shm_path,
 #ifdef __linux__
   int fd;
 #ifdef HAVE_MEMFD_CREATE
+#ifdef __FreeBSD__
+  fd = memfd_create(ring_buf_shm_path, 0);
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** Dead code — `#ifdef __FreeBSD__` is nested inside `#ifdef __linux__`, and those two macros are mutually exclusive. This block can never compile on FreeBSD. On FreeBSD the ring buffer FD will fall through to the `shm_open` path; the intended `memfd_create` fast path is silently skipped.

Fix: move the FreeBSD `memfd_create` block to its own `#elif defined(__FreeBSD__)` branch at the same level as `#ifdef __linux__`:
```cpp
#ifdef __linux__
  fd = syscall(__NR_memfd_create, ring_buf_shm_path, 0);
#elif defined(__FreeBSD__)
  fd = memfd_create(ring_buf_shm_path, 0);
#endif
```

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/lnx/amd_core_dump.cpp
```diff
@@ -64,6 +64,11 @@
 #include "core/inc/amd_gpu_agent.h"
 #include "core/inc/amd_aql_queue.h"
```
cfreeamd, line 67: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** `thr_self()` on FreeBSD has signature `void thr_self(long *id)` — it stores the TID through a pointer and returns void. Passing `0` (null) as the argument is a null-pointer dereference (undefined behavior).

Use `pthread_getthreadid_np()` instead, which returns the TID directly:
```cpp
#include <pthread_np.h>
#define SYS_gettid pthread_getthreadid_np()
```
Alternatively, check how `SYS_gettid` is used downstream in this file — if only used to initialise a variable, a local `long tid; thr_self(&tid);` approach works, but that cannot be wrapped in a simple macro.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_blit_kernel.cpp
```diff
@@ -47,6 +47,11 @@
 #include <sstream>
 #include <string>
 
+#ifdef __FreeBSD__
+#include <stdatomic.h>
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `<stdatomic.h>` is a C11 header. Including it alongside `<atomic>` in a C++ translation unit is redundant and can cause `_Atomic` keyword conflicts in some C++ compilers. Since ROCr is C++, only `<atomic>` is needed here.

cfreeamd (commented):

ROCr-only review of the FreeBSD porting changes. The `amd_topology.cpp` and `memory.h` guard extensions look correct. Two important issues in `amd_aql_queue.cpp` and `amd_core_dump.cpp`, plus a minor nit in `amd_blit_kernel.cpp`.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
Yohello1: I  rewrote the section to this;
```c++
 #ifdef __linux__
   int fd;
 #ifdef HAVE_MEMFD_CREATE
-#ifdef __FreeBSD__
   fd = memfd_create(ring_buf_shm_path, 0);
 #else
   fd = syscall(__NR_memfd_create, ring_buf_shm_path, 0);
 #endif
+  if (fd == -1) return -1;
+#elif defined(__FreeBSD__)
+  int fd;
+  fd = memfd_create(ring_buf_shm_path, 0);
+  if (fd == -1) return -1;
+#endif
```

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_blit_kernel.cpp
Yohello1: We didn't need header gaurd there, so I just removed it + stdatomic.h

```c++
-
-#ifdef __FreeBSD__
-#include <stdatomic.h>
 #include <atomic>
-#endif
 
```

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/lnx/amd_core_dump.cpp
Yohello1: Change was made;

```c++
 #ifdef __FreeBSD__
-#include <sys/thr.h>
-#define SYS_gettid thr_self(0)
+#include <pthread_np.h>
+#define SYS_gettid pthread_getthreadid_np()
+#elif defined(__linux__)
+#include <sys/syscall.h>
+#define SYS_gettid syscall(SYS_gettid)
 #endif
 
```

Yohello1 (commented):
I've made the requested changes

projects/rocr-runtime/runtime/hsa-runtime/core/driver/kfd/amd_kfd_driver.cpp
```diff
@@ -47,6 +47,9 @@
 #include <link.h>
 #include <sys/ioctl.h>
 #include <fcntl.h>
+#elif defined(__FreeBSD__)
+#include <link.h>
+#include <sys/ioctl.h>
```
dayatsin-amd: We already include sys/ioctl.h at line 48, can you remove the one at line 52.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
dayatsin-amd: Can you try this:

#if defined(__linux__) || defined(__FreeBSD__)
int fd;
#if defined(HAVE_MEMFD_CREATE) || defined(__FreeBSD__)
    fd = memfd_create(ring_buf_shm_path, 0);
#else
   fd = syscall(__NR_memfd_create, ring_buf_shm_path, 0);
 #endif
if (fd == -1) return -1;
#endif

dayatsin-amd (changes requested):

projects/rocr-runtime/runtime/hsa-runtime/core/driver/kfd/amd_kfd_driver.cpp
Yohello1: One on line 48 is gaurded by a __linux__ block, and placing outside both would break windows support

I can do this instead;
```c++
#if defined(__linux__) || defined(__FreeBSD__)
    #include <sys/ioctl.h>
    #include <link.h>
#endif

#if defined(__linux__)
    #include <amdgpu_drm.h>
#endif
```

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
Yohello1: @dayatsin-amd It might be prudent to just remove the syscall entirely and rely on memfd_create, it's been supported since 2019 by both glibc, and the linux kernel, and as far as I can tell ROCm doesn't really support things that old. If you're ok with switching to memfd_create we can do this instead which is a lot easier to read;
```c++
int AqlQueue::CreateRingBufferFD(const char* ring_buf_shm_path,
                                 uint32_t ring_buf_phys_size_bytes) const {
#if defined(__linux__) || defined(__FreeBSD__)
  int fd;

  fd = memfd_create(ring_buf_shm_path, 0);

  if (fd != -1) {
    if (ftruncate(fd, ring_buf_phys_size_bytes) == 0) {
      return fd;
    }
    close(fd);
    return -1;
  }

  fd = shm_open(ring_buf_shm_path, O_CREAT | O_RDWR | O_EXCL, S_IRUSR | S_IWUSR);

  if (fd == -1) return -1;

  if (posix_fallocate(fd, 0, ring_buf_phys_size_bytes) != 0) {
    CloseRingBufferFD(ring_buf_shm_path, fd);
    return -1;
  }

  return fd;
#else
  assert(false && "Function only needed on Linux and FreeBSD.");
  return -1;
#endif
}```

projects/rocr-runtime/runtime/hsa-runtime/core/driver/kfd/amd_kfd_driver.cpp
dayatsin-amd: Sure this works. Thanks

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
dayatsin-amd: ok. agree. Lets remove the syscall path. Thank you.

projects/rocr-runtime/runtime/hsa-runtime/core/driver/kfd/amd_kfd_driver.cpp
Yohello1: Fixed, and pushed

Yohello1:
Made the changes, should be good

dayatsin-amd (approved):
Thank you for your patience! Approved.

Yohello1:
@dayatsin-amd Can I merged [https://github.com/ROCm/rocm-systems/pull/7917](https://github.com/ROCm/rocm-systems/pull/7917) this into this pr?

dayatsin-amd:
> @dayatsin-amd Can I merged #7917 this into this pr?

Yes, sure. Go ahead.

Yohello1:
@dayatsin-amd Let me know when this pr gets merged in so I can open https://github.com/ROCm/rocm-systems/pull/9042, it has the build system changes to allow FreeBSD. The pull request after that will have the FreeBSD source files
Thank you!

dayatsin-amd:
I've cherry-picked the commits from this PR into a new branch and fixed the Linux build break in CI.

The last commit here (`Update hsakmttypes.h to also include freebsd include`) had a typo in the `#elif` guard — `defind(__FreeBSD__)` instead of `defined(__FreeBSD__)`. That caused a `missing binary operator before token "("` error and skipped the entire Linux/FreeBSD typedef block, leaving `HSAKMTAPI` and the `HSAuintXX`/`HSAintXX` types undefined and cascading into hundreds of errors across `libhsakmt`.

The fix (`defind` → `defined`) plus all commits from this PR are in the new PR: #9057. Once that one is green we can proceed there.

Yohello1:
Sounds good, I will close this

PR #8431, opened 2026-07-10
[rocprofiler-sdk] fix rejection of maximum valid stochastic sampling interval

samcrater-amd:
## Motivation

Some PCS interval values were not being accepted/rejected properly.

## Technical Details

Ensured we were only reporting valid interval ranges, and added log telling user how to find valid ranges on a config failure.

## JIRA ID

JIRA ID : AIPROFSDK-913.

## Test Plan

Ran profiler with intervals that were just below, at, and above the maximum supported interval value.

## Test Result

Profiler accepted maximum interval and rejected others.

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/tool.cpp
+++ b/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/tool.cpp
@@ -2270,7 +2270,34 @@ configure_pc_sampling_on_all_agents(uint64_t                        buffer_size,
     }
     if(!config_match_found)
     {
-        ROCP_ERROR << "Given PC sampling configuration is not supported on any of the agents";
+        // If unsuccessful, get the supported interval range and output it
+        uint64_t supported_min = std::numeric_limits<uint64_t>::max();
+        uint64_t supported_max = 0;
+        for(auto& itr : agent_ptr_vec)
+        {
+            for(const auto& config :
+                CHECK_NOTNULL(tool_metadata)->get_pc_sample_config_info(itr->id))
+            {
+                if(config.method == method && config.unit == unit)
+                {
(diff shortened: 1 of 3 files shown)


vlaindic:
Thanks for the PR: @samcrater-amd . Please check the comment I added into the Jira.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/tool.cpp
```diff
@@ -2266,7 +2266,34 @@ configure_pc_sampling_on_all_agents(uint64_t                        buffer_size,
     }
     if(!config_match_found)
     {
-        ROCP_ERROR << "Given PC sampling configuration is not supported on any of the agents";
+        // If unsuccessful, get the supported interval range and output it
```
ApoKalipse-V: Would it not be simpler to ask them to run rocprofv3 --list-avail?

vlaindic: I agree, I think this might be overly complicated. We could simply refer them to query the available configurations via `rocprofv3-avail info --pc-sampling` and check the output against the arguments they inputted. 

The reason for mismatch might be:
- sampling interval out of range
- sampling interval in range, but not power of 2 for stochastic PC sampling
- wrong combination of unit, interval and method

I think one generic message and referral to the `avail` sound more reasonable.

vlaindic (changes requested):
I'm seeing this in the testing plan: `Ran profiler with intervals that were just below, at, and above the maximum supported interval value.` but I'm not seeing any automation tests. Should we add them? We could do it as a separate PR (tracked in another JIRA/issue)

vlaindic (approved):
LGTM, thanks!

PR #8438, opened 2026-07-11
ci(rocjitsu): Add extra failure handling steps in workflow

### LINKED ISSUE #8558: [CI] Add rerun in rocjitsu CI workflow
Update test corpus commit reference and implement the rerun logics.

RattataKing:
## Motivation

Some corpus tests have inconsistent execution time, they may occasionally exceed 15s timeout and cause false positive ci failures.
We want collect more info from test results to tell if a failed test is either hanging / too slow / affected by timing fluctuation.

## Technical Details

- Updated corpus ref.
- Changed pytest command to also report test durations.
- Added extra failure handling steps in workflow.
  - If pytest reports any regular, non-timeout failures, CI fails as usual.
  - If pytest results only have timeout type failures, and total failures <=5, rerun them with timeout=30s instead of 15s.
  - If retry passes, raise warning message and treat test as passed.

## JIRA ID

Closes: https://github.com/ROCm/rocm-systems/issues/8558

## Test Plan

- Temporally enable some tests in `gfx1201_skip_tests.json` and run CI. (Reverted)
- 2/3 tests needs >15s to pass, 1/3 tests fails after 15s, so these can be used to verify if:
  - CI reruns timeout tests correctly.
  - CI raises error status if tests fails again with 30s.

## Test Result

CI failed as expected: https://github.com/ROCm/rocm-systems/actions/runs/29135988995/job/86501278754?pr=8438
 
## Submission Checklist


.github/workflows/rocjitsu-corpus-tests.yml
```diff
@@ -102,8 +103,8 @@ jobs:
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         with:
           repository: ${{ github.event.inputs.corpus_repository || 'ROCm/rocjitsu-test-corpus' }}
-          # Last updated on 2026/07/07.
-          ref: 3278883f548e1353ffbda1a8090219d866eca035
+          # Last updated on 2026/07/10.
+          ref: 4691a2ce389ab9f721259c53bff502786c5a91b5
```
amd-eochoalo: This will need to be updated once the branch in rocjitsu-test-corpus is merged.

amd-eochoalo (changes requested):
I'm just blocking this PR until we finish review of the PR on the rocjitsu-test-corpus repo https://github.com/ROCm/rocjitsu-test-corpus/pull/13. Let's have the conversation there to avoid spamming other co-workers while we finalize this.

.github/workflows/rocjitsu-corpus-tests.yml
```diff
@@ -24,8 +24,8 @@ on:
         required: true
       corpus_ref:
         description: "Branch, tag, or SHA to check out from the corpus repository"
-        # Last updated on 2026/07/07.
-        default: "3278883f548e1353ffbda1a8090219d866eca035"
+        # Last updated on 2026/07/14
```
kuhar:
```suggestion
        # Last updated on 2026/07/14.
```

.github/workflows/rocjitsu-corpus-tests.yml
```diff
@@ -222,19 +222,86 @@ jobs:
           ROCM_PATH="$(rocm-sdk path --root)"
           export ROCM_PATH
           corpus_status=0
+          corpus_results=""
+          finish_target() {
```
kuhar: This whole step should be extracted into a script -- it's too complicated to have it inline in a github action that can only be tested in CI

kuhar (changes requested):

.github/workflows/rocjitsu-corpus-tests.yml
```diff
@@ -222,19 +222,86 @@ jobs:
...
-              --target "${name}" \
-              --suite iree,kernels,cts \
-              --skip-tests-config "${ROCJITSU_SOURCE_DIR}/tests/corpus/${skip_tests_config}" \
-              --timeout 15; then
-              corpus_status=1
+              "${common_pytest_args[@]}" \
+              --run-tests-config "${timeout_failures_config}" \
+              --timeout "${LONG_TIMEOUT}"; then
+              echo "::warning::Retried pytest-timeout failures passed."
+              target_status=0
```
amd-eochoalo: We should probably have this not as a warning but as a hard error. Otherwise, we will ignore it. It is good to know that a test passed when given more time to execute, but a difference of 15 seconds to 60 seconds per test is too much. If 5 tests change from 15 seconds to 5 minutes that's something we should flag.

RattataKing: The warning here is more about telling people that even though all tests passes at the end, some suspicious timeout occurred in the middle. We can also change 60s to 30s if the latter value makes more sense to consider as a pass.

amd-eochoalo: I looked a little bit more into this. What I was thinking was that we should have an red circle on CI, but still allow merging. However, this is done through the repository settings and not through YAML.

It is fine then. 

Just for the record, I wanted something like:

"It is red, let me see what happened. Ah, just a timeout, I can still merge" or alternatively once we have timeouts for each test "let me increase the timeout so that we don't have this again". But as I mentioned, this appears to not be possible from the YAML file. We can only have continue-on-error: true which keeps the circle green and allows people to just say. "It worked, I don't have to look into this, and I don't even know that a timeout happened".

So, in other words, I think switching to 30s is sensible enough. Thanks!

.github/workflows/rocjitsu-corpus-tests.yml
```diff
@@ -222,19 +222,86 @@ jobs:
...
+          corpus_results=""
+          finish_target() {
+            local target_status="$1"
+            if [[ "${target_status}" -ne 0 ]]; then
+              corpus_status=1
+            fi
+            corpus_results+="${name}: $([[ "${target_status}" -eq 0 ]] && echo PASS || echo FAIL)"$'\n'
+            echo "::endgroup::"
+          }
+          SHORT_TIMEOUT=15
```
newling: I think names that describe the policy rather than the mechanism would be better. 

  SOFT_TIMEOUT_SECONDS=15
  HARD_TIMEOUT_SECONDS=30 # or is it 60? 
  MAX_SOFT_TIMEOUT_RETRIES=5

newling (commented):
Is the 'repeat this 5 times' policy because some tests are hanging, rather than just being slow? Do we have an idea of the distribution of execution times? I wonder if we can track the times, it might be useful to see the emulation performance through time.

.github/workflows/rocjitsu-corpus-tests.yml
RattataKing: It's not `repeat this 5 times`, it will only rerun with longer timeout if there are at most 5 timeout cases. The current corpus tests are all created and expected to pass within 15s, but we sometimes get ci errors due to time noise, such as one test normally pass with 10s, it can occasionally pass with 15s, if there are too many timeout cases happened at the same time, this can indicate to the performance issues and CI would raise errors instead.

RattataKing: We will add expected durations for each test case in future

amd-eochoalo (approved):

RattataKing:
Implemented with a different approach: https://github.com/ROCm/rocm-systems/pull/8654

PR #8446, opened 2026-07-11
fix(rocr): Use recursive mutex for agent memory lock

dayatsin-amd:
## Summary

Replace the deferred-free re-entrancy workaround (PR #7492) with a `std::recursive_mutex` on `agent_memory_lock_`. When AddressSanitizer quarantine is enabled, an allocate can evict an older GPU block and re-enter `MemoryRegion::Free` on the same thread; a non-recursive mutex deadlocks. A recursive mutex allows the re-entrant free without the thread-local deferred-free queue.

Validated on s72-2 (gfx942) with TheRock 7.14 (`/home/amd/dayatsin/therock`):

- `Unit_Warp_Shfl_Down_Positive_Basic` (10 variants): **passed** with `ASAN_OPTIONS=quarantine_size_mb=600` and `HSA_XNACK=1`
- Full hip catch suite (recursive-mutex ROCr, no ASAN): **3371 passed / 6 failed / 1218 skipped** (99%)
- Full hip catch suite (recursive-mutex ROCr, ASAN + `HSA_XNACK=1`): **3357 passed / 36 failed / 1196 skipped** (99%)
- ASAN baseline (stock TheRock ROCr, same env): **3355 passed / 38 failed / 1196 skipped** (99%) — on par with recursive mutex (+2 pass / −2 fail); 28 common failures (managed-memory, prefetch, stream-attach, IPC); remaining diff is run-to-run timeout/flake noise, not a new failure class

## JIRA ID

JIRA ID : ROCM-26090

## Test plan

- [x] Built ROCr against TheRock `/home/amd/dayatsin/therock` on s72-2
- [x] `Unit_Warp_Shfl_Down_Positive_Basic` (10 variants) with `ASAN_OPTIONS=quarantine_size_mb=600`, `HSA_XNACK=1`
- [x] Full hip catch suite without ASAN/`HSA_XNACK` (3371 passed / 6 failed / 1218 skipped)
- [x] Full hip catch suite with `ASAN_OPTIONS=quarantine_size_mb=600`, `HSA_XNACK=1` (3357 passed / 36 failed / 1196 skipped)
- [x] ASAN baseline vs stock TheRock ROCr (3355 passed / 38 failed / 1196 skipped)

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/inc/agent.h
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/inc/agent.h
@@ -433,7 +433,7 @@ protected:
   // Serial memory operations are needed to ensure, among other things, that allocation failures are
   // due to true OOM conditions and per region caching (Trim and Allocate must be serial and
   // exclusive to ensure this).
-  std::mutex agent_memory_lock_;
+  std::recursive_mutex agent_memory_lock_;
 
   // Forbid copying and moving of this object
   DISALLOW_COPY_AND_ASSIGN(Agent);
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_memory_region.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_memory_region.cpp
@@ -48,9 +48,7 @@
 #include "core/inc/amd_memory_region.h"
 
 #include <algorithm>
-#include <cstdio>
 #include <mutex>
-#include <vector>
 #include <shared_mutex>
 
 #include "core/inc/runtime.h"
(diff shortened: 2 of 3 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_memory_region.cpp
```diff
@@ -230,13 +164,7 @@ hsa_status_t MemoryRegion::AllocateImpl(size_t& size, AllocateFlags alloc_flags,
...
 
 hsa_status_t MemoryRegion::Free(void* address, size_t size) const {
-#if defined(SANITIZER_AMDGPU)
-  // If re-entrant (under ASAN), defer; otherwise take lock and free normally.
-  if (DeferFreeIfLockHeld(address, size)) return HSA_STATUS_SUCCESS;
-  ScopedAgentMemoryLock lock(owner()->agent_memory_lock_);
-#else
-  std::lock_guard<std::mutex> lock(owner()->agent_memory_lock_);
-#endif
+  std::lock_guard<std::recursive_mutex> lock(owner()->agent_memory_lock_);
```
cfreeamd, line 167: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** With `recursive_mutex`, a re-entrant `Free` (ASAN quarantine eviction during an allocate on the same thread) now runs `FreeImpl` **immediately, nested** inside the outer `AllocateImpl`/`FreeImpl` still on the stack. The previous deferred-free design specifically avoided that — it queued the re-entrant free and drained it only after the *outermost* critical section exited. Recursive mutex solves the deadlock, but it also changes ordering: the inner free can now mutate the region's block/fragment allocator state mid-operation. Was the deferral's post-critical-section ordering load-bearing for allocator-state integrity, or purely a deadlock-avoidance mechanism? The ASAN quarantine validation is reassuring; just want to confirm the outer op holds no iterator / cached free-list state across the nested-free point.

projects/rocr-runtime/runtime/hsa-runtime/core/inc/agent.h
```diff
@@ -433,7 +433,7 @@ class Agent : public Checked<0xF6BC25EB17E6F917> {
   // Serial memory operations are needed to ensure, among other things, that allocation failures are
   // due to true OOM conditions and per region caching (Trim and Allocate must be serial and
   // exclusive to ensure this).
-  std::mutex agent_memory_lock_;
+  std::recursive_mutex agent_memory_lock_;
```
cfreeamd, line 436: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** This makes the lock a `recursive_mutex` in *all* builds, not just `SANITIZER_AMDGPU`. A recursive mutex silently permits accidental same-thread re-locking in normal production paths that a plain `std::mutex` would otherwise surface as a deadlock during testing. The single-code-path simplicity is a fair trade, but worth making it a conscious decision vs. gating `recursive_mutex` to `SANITIZER_AMDGPU` and keeping `std::mutex` elsewhere.

cfreeamd (commented):

Mechanical simplification looks clean — the three call sites (`Allocate`, `Free`, `IPCFragmentExport`) consistently switch to `lock_guard<std::recursive_mutex>`, and the removed `<cstdio>`/`<vector>` includes were only used by the deleted deferred-free code. Two questions left inline about re-entrant-free ordering and unconditional recursive-mutex use.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_memory_region.cpp
dayatsin-amd: It was a deadlock-avoidance mechanism

projects/rocr-runtime/runtime/hsa-runtime/core/inc/agent.h
dayatsin-amd: It is a conscious decision. We want the mutex to protect against concurrent API calls from CLR, but allow calls from inside ROCr/ASAN library to go through.

cfreeamd (approved):

PR #8447, opened 2026-07-11
fix(rocr): shader-blit fallback for DmaCopyBatch multi-entry LINEAR

dayatsin-amd:
## Summary

Under `HSA_ENABLE_SDMA=0`, multi-entry LINEAR batch copies (`hipMemcpyBatchAsync` H2D/D2H and `hsa_amd_memory_async_batch_copy`) routed through `DmaCopyMulti` → `DmaCopyFanOutOp` fail with `HSA_STATUS_ERROR_INVALID_ARGUMENT` because no SDMA engine is available. Single-entry LINEAR copies already fall back to the BlitDevToDev shader via `DmaCopyOnEngine`, but the multi-entry batch path had no equivalent.

Add `DmaCopyBatchFallback`, invoked at the top of `DmaCopyBatch` when SDMA is disabled:
- **LINEAR**: per-entry synchronous `DmaCopy` (BlitDevToDev shader), covering both multi-entry batches and single scalar ops
- **BROADCAST / SWAP / INDIRECT**: no shader equivalent yet; explicit reject (same as SDMA fan-out would)

### Review follow-ups (addressed)
- Completion signal is decremented with `SubRelease(1)` (not `SubRelaxed`) so a consumer waiting with `scacquire` is guaranteed to observe the copied bytes.
- The completion signal is resolved only on success; on a per-entry failure the helper returns the error without decrementing, matching the normal `DmaCopyBatch` switch and its caller (callee resolves only on success).
- Dropped the `default:` case so the switch stays exhaustive over `hsa_amd_memory_copy_op_t` and a newly added op type triggers a compiler warning instead of being silently rejected.
- Documented the BlitDevToDev shader-selection rationale (SDMA=0 forces `BlitDevToDev` in `DmaCopyOnEngine` anyway) and the global-SDMA probe assumption.

## JIRA ID

JIRA ID : ROCM-27858

## Test plan

Validated on b8-3 (gfx1250) against a TheRock `/opt/rocm` toolchain, comparing stock vs patched `libhsa-runtime64` with an 8-entry LINEAR `hipMemcpyBatchAsync` repro (H2D/D2H/D2D, with data verification):

- [x] Stock runtime, `HSA_ENABLE_SDMA=0`: H2D/D2H fail (silent corruption, ret=0), D2D passes — bug reproduced
- [x] Stock runtime, default SDMA: all pass
- [x] Patched runtime, `HSA_ENABLE_SDMA=0`: H2D/D2H/D2D all pass — fix validated
- [x] Patched runtime, default SDMA: all pass — no regression

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
@@ -1965,6 +1965,44 @@ hsa_status_t GpuAgent::DmaCopyIndirect(
                          op.dst_list, op.dst_agent_list, op.size_list);
 }
 
+hsa_status_t GpuAgent::DmaCopyBatchFallback(
+    const hsa_amd_memory_copy_op_t& op,
+    std::vector<core::Signal*>& dep_signals) {
+  core::Signal& out_signal = *core::Signal::Convert(op.completion_signal);
+
+  switch (op.type) {
+  case HSA_AMD_MEMORY_COPY_OP_LINEAR: {
+    // BlitDevToDev linear copy shader, one entry at a time. Covers both the
+    // multi-entry batch (hipMemcpyBatchAsync H2D/D2H) and the single scalar op.
+    for (core::Signal* sig : dep_signals)
+      sig->WaitRelaxed(HSA_SIGNAL_CONDITION_EQ, 0, UINT64_MAX,
+                       HSA_WAIT_STATE_BLOCKED);
+    hsa_status_t status = HSA_STATUS_SUCCESS;
+    if (op.num_entries > 0) {
+      for (uint16_t d = 0; d < op.num_entries; ++d) {
+        status = DmaCopy(op.dst_list[d], op.src_list[d], op.size_list[d]);
+        if (status != HSA_STATUS_SUCCESS) break;
+      }
+    } else {
+      status = DmaCopy(op.dst, op.src, op.size);
+    }
+    out_signal.SubRelaxed(1);
+    return status;
+  }
+  case HSA_AMD_MEMORY_COPY_OP_LINEAR_BROADCAST:
+  case HSA_AMD_MEMORY_COPY_OP_LINEAR_SWAP:
+  case HSA_AMD_MEMORY_COPY_OP_LINEAR_INDIRECT_SRC:
+  case HSA_AMD_MEMORY_COPY_OP_LINEAR_INDIRECT_DST:
+  case HSA_AMD_MEMORY_COPY_OP_LINEAR_INDIRECT_SRCDST:
+    // No shader-blit equivalent for broadcast/swap/indirect yet; these are the
+    // slots for the 1-to-N / swap / indirect blit shaders once added. Until
+    // then, reject under SDMA=0 (same as the SDMA fan-out path would).
+    return HSA_STATUS_ERROR_INVALID_ARGUMENT;
+  default:
+    return HSA_STATUS_ERROR_INVALID_ARGUMENT;
+  }
+}
+
 hsa_status_t GpuAgent::DmaCopyBatch(const hsa_amd_memory_copy_op_t* ops,
                                     uint32_t num_ops,
                                     std::vector<core::Signal*>& dep_signals) {
@@ -1978,6 +2016,17 @@ hsa_status_t GpuAgent::DmaCopyBatch(const hsa_amd_memory_copy_op_t* ops,
     core::Signal& out_signal = *out_signal_obj;
 
     hsa_status_t status;
+
+    // SDMA disabled: route the op through the shader-blit fallback helper,
+    // which selects the appropriate blit shader per op type (or rejects ops
+    // with no shader equivalent yet). Done before the switch so all SDMA=0
+    // shader-selection lives in one place.
+    if (!GetBlitObject(BlitHostToDev)->isSDMA()) {
+      status = DmaCopyBatchFallback(op, dep_signals);
+      if (status != HSA_STATUS_SUCCESS)
+        return status;
+      continue;
+    }
 
     switch (op.type) {
     case HSA_AMD_MEMORY_COPY_OP_LINEAR: {
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
@@ -1974,20 +1974,32 @@ hsa_status_t GpuAgent::DmaCopyBatchFallback(
   case HSA_AMD_MEMORY_COPY_OP_LINEAR: {
     // BlitDevToDev linear copy shader, one entry at a time. Covers both the
     // multi-entry batch (hipMemcpyBatchAsync H2D/D2H) and the single scalar op.
+    // The 3-arg DmaCopy issues blits_[BlitDevToDev] synchronously; under SDMA=0
+    // this is the same shader the single-entry LINEAR path lands on, since
+    // DmaCopyOnEngine forces engine_offset = BlitDevToDev when SDMA is disabled
+    // (covering local H2D/D2H as well as peer entries the copy agent can map).
     for (core::Signal* sig : dep_signals)
       sig->WaitRelaxed(HSA_SIGNAL_CONDITION_EQ, 0, UINT64_MAX,
                        HSA_WAIT_STATE_BLOCKED);
-    hsa_status_t status = HSA_STATUS_SUCCESS;
     if (op.num_entries > 0) {
       for (uint16_t d = 0; d < op.num_entries; ++d) {
-        status = DmaCopy(op.dst_list[d], op.src_list[d], op.size_list[d]);
-        if (status != HSA_STATUS_SUCCESS) break;
+        hsa_status_t status =
+            DmaCopy(op.dst_list[d], op.src_list[d], op.size_list[d]);
+        // On error, leave the completion signal untouched and return, matching
+        // the normal DmaCopyBatch switch (callee resolves the signal only on
+        // success; the caller propagates the error). Decrementing here would
+        // signal completion for a copy that did not happen.
+        if (status != HSA_STATUS_SUCCESS) return status;
       }
     } else {
-      status = DmaCopy(op.dst, op.src, op.size);
+      hsa_status_t status = DmaCopy(op.dst, op.src, op.size);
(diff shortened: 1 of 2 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -2048,6 +2048,44 @@ hsa_status_t GpuAgent::DmaCopyIndirect(
...
+    hsa_status_t status = HSA_STATUS_SUCCESS;
+    if (op.num_entries > 0) {
+      for (uint16_t d = 0; d < op.num_entries; ++d) {
+        status = DmaCopy(op.dst_list[d], op.src_list[d], op.size_list[d]);
+        if (status != HSA_STATUS_SUCCESS) break;
+      }
+    } else {
+      status = DmaCopy(op.dst, op.src, op.size);
+    }
+    out_signal.SubRelaxed(1);
```

**[Important]** `out_signal.SubRelaxed(1)` provides no release edge. The consumer waiting on `op.completion_signal` (via `hsa_signal_wait_scacquire`) needs a release/acquire pair to be guaranteed to observe the copied data. The nearly identical synchronous-copy-then-signal pattern in `amd_cpu_agent.cpp:471` (wait deps → `memcpy` → decrement) uses `SubRelease(1)` for exactly this reason. As written, a consumer could observe completion before the copied bytes are visible — this should be `SubRelease(1)`.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -2048,6 +2048,44 @@ hsa_status_t GpuAgent::DmaCopyIndirect(
...
+  case HSA_AMD_MEMORY_COPY_OP_LINEAR: {
+    // BlitDevToDev linear copy shader, one entry at a time. Covers both the
+    // multi-entry batch (hipMemcpyBatchAsync H2D/D2H) and the single scalar op.
+    for (core::Signal* sig : dep_signals)
+      sig->WaitRelaxed(HSA_SIGNAL_CONDITION_EQ, 0, UINT64_MAX,
+                       HSA_WAIT_STATE_BLOCKED);
+    hsa_status_t status = HSA_STATUS_SUCCESS;
+    if (op.num_entries > 0) {
+      for (uint16_t d = 0; d < op.num_entries; ++d) {
+        status = DmaCopy(op.dst_list[d], op.src_list[d], op.size_list[d]);
```

**[Question]** This uses the 3-arg `DmaCopy`, which hardcodes `blits_[BlitDevToDev]` on *this* agent and ignores the per-entry `op.dst_agent_list`/`op.src_agent_list`. The normal path threads `dst_agent`/`src_agent` through `DmaCopyMulti`/`DmaCopyOnEngine` and selects `BlitHostToDev`/`BlitDevToHost` by direction. Under SDMA=0 those are all shader blits, so this is likely fine when this agent can access both pointers — but can you confirm it's correct for cross-agent / peer entries within a batch, not just local H2D/D2H?

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -2048,6 +2048,44 @@ hsa_status_t GpuAgent::DmaCopyIndirect(
...
+  }
+  case HSA_AMD_MEMORY_COPY_OP_LINEAR_BROADCAST:
+  case HSA_AMD_MEMORY_COPY_OP_LINEAR_SWAP:
+  case HSA_AMD_MEMORY_COPY_OP_LINEAR_INDIRECT_SRC:
+  case HSA_AMD_MEMORY_COPY_OP_LINEAR_INDIRECT_DST:
+  case HSA_AMD_MEMORY_COPY_OP_LINEAR_INDIRECT_SRCDST:
+    // No shader-blit equivalent for broadcast/swap/indirect yet; these are the
+    // slots for the 1-to-N / swap / indirect blit shaders once added. Until
+    // then, reject under SDMA=0 (same as the SDMA fan-out path would).
+    return HSA_STATUS_ERROR_INVALID_ARGUMENT;
```

**[Question]** The LINEAR path always decrements the completion signal (even on a mid-loop per-entry failure), but this reject path returns `HSA_STATUS_ERROR_INVALID_ARGUMENT` without touching `out_signal`. A consumer waiting on that op's completion signal would hang. The two paths in this function should agree on the error contract (caller resolves vs. callee always resolves), and match how the normal `DmaCopyBatch` switch leaves completion signals on error.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -2062,6 +2100,17 @@ hsa_status_t GpuAgent::DmaCopyBatch(const hsa_amd_memory_copy_op_t* ops,
 
     hsa_status_t status;
 
+    // SDMA disabled: route the op through the shader-blit fallback helper,
+    // which selects the appropriate blit shader per op type (or rejects ops
+    // with no shader equivalent yet). Done before the switch so all SDMA=0
+    // shader-selection lives in one place.
+    if (!GetBlitObject(BlitHostToDev)->isSDMA()) {
```
cfreeamd, line 2045: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** This gates the entire batch (any direction) on the `BlitHostToDev` blit's SDMA state. Fine for the global `HSA_ENABLE_SDMA=0` case, but if per-direction SDMA availability can diverge this could misroute D2H/D2D entries. `DmaCopyRect` picks the blit for the actual direction; worth noting the global assumption here.

cfreeamd (commented):

The fallback structure reads clearly and the 3-arg `DmaCopy` path is confirmed synchronous (`BlitKernel::SubmitLinearCopyCommand` waits with `scacquire` internally), so the manual completion decrement is timing-correct. Main inline point is a completion-signal memory-ordering issue (`SubRelaxed` vs `SubRelease`); a few consistency questions vs. the normal `DmaCopyBatch` path are noted inline too.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -2048,6 +2048,44 @@ hsa_status_t GpuAgent::DmaCopyIndirect(
...
+  case HSA_AMD_MEMORY_COPY_OP_LINEAR_BROADCAST:
+  case HSA_AMD_MEMORY_COPY_OP_LINEAR_SWAP:
+  case HSA_AMD_MEMORY_COPY_OP_LINEAR_INDIRECT_SRC:
+  case HSA_AMD_MEMORY_COPY_OP_LINEAR_INDIRECT_DST:
+  case HSA_AMD_MEMORY_COPY_OP_LINEAR_INDIRECT_SRCDST:
+    // No shader-blit equivalent for broadcast/swap/indirect yet; these are the
+    // slots for the 1-to-N / swap / indirect blit shaders once added. Until
+    // then, reject under SDMA=0 (same as the SDMA fan-out path would).
+    return HSA_STATUS_ERROR_INVALID_ARGUMENT;
+  default:
```
cfreeamd: Optional: Consider not having a "default:" and just returning HSA_STATUS_ERROR_INVALID_ARGUMENT for the cases above. The thinking on that is that you are currently covering all the cases so it should work fine. If another enum is added, we may forget to handle that case here. Without the default, at least the compiler should give a warning that we are not handling the new case (though, admittedly, I think we have several warnings on a normal build that might make a new, legit warning hard to notice).

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
dayatsin-amd: Good catch — fixed in 4fe1563188. The decrement is now `SubRelease(1)`, mirroring the `CpuAgent::DmaCopy` synchronous copy-then-signal pattern so a consumer waiting with `scacquire` is guaranteed to observe the copied bytes.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
dayatsin-amd: Confirmed correct for peer/cross-agent entries under SDMA=0. When SDMA is disabled, `DmaCopyOnEngine` itself forces `engine_offset = BlitDevToDev` (the `enable_sdma()==SDMA_DISABLE` branch), so the single-entry LINEAR path already lands on the exact `blits_[BlitDevToDev]` shader the 3-arg `DmaCopy` uses. Direction-based `BlitHostToDev`/`BlitDevToHost` selection only matters for picking an SDMA *engine*, not a shader; peer entries rely on the same peer-mapping requirement as the single-copy SDMA=0 path. Documented this in 4fe1563188. (Caveat: this path does not set `async_copy_agent`/collect timestamps, so profiled batch copies under SDMA=0 lack copy timestamps — acceptable for the fallback.)

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
dayatsin-amd: Fixed in 4fe1563188. The helper now resolves the completion signal only on success: on a per-entry failure it returns the error immediately without decrementing `out_signal`, matching both the reject path and the normal `DmaCopyBatch` switch + its caller in `hsa_ext_amd.cpp` (callee resolves only on success; the caller propagates the error and never decrements on failure).

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
dayatsin-amd: Acknowledged — intentional for the `HSA_ENABLE_SDMA=0` case this fallback targets, where SDMA is enabled/disabled globally and all blits share one state. Added a comment in 4fe1563188 noting the global-SDMA probe assumption and that the probe would need to move per-entry if per-direction SDMA availability ever diverges.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
dayatsin-amd: Done in 4fe1563188. Removed the `default:` case and added a trailing `return HSA_STATUS_ERROR_INVALID_ARGUMENT;` after the switch, so it stays exhaustive over `hsa_amd_memory_copy_op_t` (`-Wswitch` flags a newly added op type) while still safely handling any out-of-enum value.

cfreeamd (approved):

Issue #8448, opened 2026-07-11
[Issue]: Resolve BO handles per-agent for AIE support


ypapadop-amd:
### Problem Description

Memory-handle resolution in assumes a single driver/agent context, so when multiple drivers are involved (KFD + AIE), handles could be resolved or created against the wrong agent's driver.

Memory sharing/import across heterogeneous agents (GPU and AIE/NPU) could silently use the wrong driver context, causing incorrect handle resolution or masked errors.

### Operating System

Ubuntu 24.04

### CPU

AMD Ryzen 9 7940HS

### GPU

Radeon 780M

### ROCm Version

ROCm 7.2

### ROCm Component

ROCR-Runtime

### (Optional for Linux users) Output of /opt/rocm/bin/rocminfo --support

<details>
<summary>rocminfo --support output</summary>

```
Paste output here
```

</details>

PR #8449, opened 2026-07-11
fix (rocr):  resolve BO handles per-agent / driver

ypapadop-amd:
## Motivation

Memory-handle resolution in rocr-runtime assumed a single driver/agent context, so when multiple drivers are involved (GPU + XDNA), handles could be resolved or created against the wrong agent's driver.

Solves #8448 

## Technical Details

- Refactored memory management interfaces to pass DriverMemoryHandle by reference instead of by value/raw handle, so allocation metadata (handle + size) flows through consistently instead of being re-derived.
- Runtime now resolve driver memory handles per requesting agent, not per allocation-owner alone - needed because a BO can be allocated by one agent (e.g., GPU) but accessed/mapped through another's driver (e.g., AIE).

## JIRA ID

NA

## Test Plan

Added AIE tests to VirtMemoryBasic.

Additional tests run from https://github.com/ROCm/rocm-systems/tree/users/ypapadop-amd/aie-tests

## Test Result

Success.

## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
@@ -40,14 +40,13 @@
 //
 ////////////////////////////////////////////////////////////////////////////////
 
-#include <cassert>
 #include <algorithm>
+#include <cassert>
 #include <chrono>
 #include <thread>
 #include <cstring>
 #include <regex>
 #include <string>
-#include <algorithm>
 #if defined(__linux__)
 #include <link.h>
 #include <dlfcn.h>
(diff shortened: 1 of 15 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
```diff
@@ -3883,32 +3946,36 @@ hsa_status_t Runtime::VMemoryHandleCreate(const MemoryRegion* region, size_t siz
 
   std::lock_guard<std::shared_mutex> lock(memory_lock_);
   void *mem;
```
dayatsin-amd: Minor. It seems we do not need `mem` anymore, can you remove it?

Otherwise LGTM.

cfreeamd (commented):

`amd_kfd_driver.cpp`, `amd_kfd_virtio_driver.cpp`, `amd_xdna_driver.cpp`, `runtime.cpp`, `amd_aie_aql_queue.cpp`, `amd_gpu_agent.cpp`, `amd_memory_region.cpp`, and `virtual_memory.cc` all look correct. The `rocrtst_yaml_missing` checklist flag appears to be a false positive: `ImportedShareableHandleSetAccessAfterFdClose` is an existing test that the PR extends, not a new `TEST()` registration.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
ypapadop-amd: Done in https://github.com/ROCm/rocm-systems/pull/8449/changes/105f8918f08aaa352de5d7c824018639ca783f2f

The side-effect of this is that we populate `handle.vaddr` for KFD as well (same as the handle since handle and mapped memory pointer are one and the same). Let me know if that's acceptable.

dayatsin-amd (approved):

PR #8450, opened 2026-07-11
fix(log): order ROCr logs with HIP and match thread ids

saleelk:
Route ROCr's log output through CLR's async ring buffer via a cookie sink so ROCr and HIP lines share one ordered stream instead of ROCr's direct fflush racing ahead. Also render ROCr's thread id with the same masked std::thread::id hash as CLR so ids match across the two layers.

## JIRA ID

JIRA ID : AIRUNTIME-0

## Submission Checklist

### DIFF
--- a/projects/clr/rocclr/utils/debug.cpp
+++ b/projects/clr/rocclr/utils/debug.cpp
@@ -4,6 +4,12 @@
  * SPDX-License-Identifier: MIT
  */
 
+// _GNU_SOURCE must be defined before any header so glibc declares fopencookie(),
+// which backs the ordered log sink handed to ROCr (see GetRocrLogSink below).
+#if defined(__linux__) && !defined(_GNU_SOURCE)
+#define _GNU_SOURCE
+#endif
+
 #include "top.hpp"
 #include "utils/debug.hpp"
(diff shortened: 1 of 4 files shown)


projects/clr/rocclr/utils/debug.cpp
```diff
@@ -157,6 +165,32 @@ class AsyncLogger {
         std::hash<std::thread::id>{}(std::this_thread::get_id()) & 0xFFFFF);
     entry.duration = duration;
     entry.hasDuration = hasDuration;
+    entry.isRaw = false;
+    entry.valid.store(true, std::memory_order_release);
+  }
+
+  // Enqueue a pre-formatted line (e.g. from ROCr) so it is emitted verbatim and
+  // in generation order relative to the structured entries from this logger.
+  void logRaw(std::string line) {
```
chrispaquot, line 174: Better to pass a const&

cfreeamd (commented):

The ROCr file (`core/util/flag.cpp`) looks correct — thread-id hash and snprintf format match HIP's rendering cleanly.

cfreeamd (approved):

PR #8456, opened 2026-07-12
refactor(rocr): Unify and optimize SDMA fan-out submission

saleelk:
Unify the SDMA copy submission paths and optimize multi-engine fan-out.

Submission API
- Restructure SubmitBodies: both fused (gfx125+) and classic paths use a single ring reservation for all entries (one doorbell per engine group). Classic path assumes platform atomics and decrements out_signal directly per entry -- no per-body signals needed.
- Simplify SubmitPrologue/SubmitEpilogue: remove atomic-decrement branches where the signal value is known (prologue signal is always 1, epilogue polls then fences to 0). Remove body_signals and drain_bodies_individually logic.
- Eliminate heap allocation in the classic body path: copy packets are formed in-place in the ring buffer instead of staging through a temporary vector.
- Consolidate linearB2B paths into DmaCopyFanOutOp with coord_engine and max_engines; drop engine_mask (topology resolution handles engine selection, capped by max_engines).
- Add SubmitFusedCoordinator (gfx125+): pack prologue + body entries + epilogue into a single ring reservation (one doorbell) for the coordinator engine, and strip pre-gfx125+ dead branches unreachable behind IsGfx125Plus().
- Gather SubmitBodies entries from master lists (index vector) instead of pre-built per-engine vectors, removing four std::vector allocations per engine group per fan-out; collapse sizes_a/sizes_b into a single size_list.

Fan-out optimizations
- Rotate SDMA engines within a single copy, not across API calls: remove the persistent round-robin counter (sdma_rr_index_) and PickSdmaEngine; engine assignment now uses NthSdmaEngine(mask, k) driven by the per-entry index.
- gfx125+ fused-coordinator epilogue fast-path: when the completion signal has no event mailbox and GCR/profiling are off, skip the epilogue poll+fence -- the per-body SIGNAL subs already provide the ordered completion edge; the signal target is shifted down by one so the last sub lands on the done value.
- Revert the gfx1250 multicast crossover from 8MB back to 256KB: single-engine multicast serialises writes to all destinations and becomes bandwidth-bound well before 8MB, so multi-destination copies above 256KB now fan out across engines. HSA_SDMA_MULTICAST override unchanged.
- Tune linearB2B threshold: lower kB2BMaxSize from 256KB to 64KB to eliminate latency blips at ~2MB sizes in multi-entry batch copies.
- Move the ReleaseWrite debug log into UpdateWriteAndDoorbellRegister under the doorbell serialization, logging the written wptr plus the live rptr.

## JIRA ID

JIRA ID : AIRUNTIME-2416

## Submission Checklist


cfreeamd (commented):

`amd_blit_sdma.h` interface changes look correct and well-documented. `NthSdmaEngine` replacement is consistent. Signal arithmetic for fused and classic paths is correct. Fan-out engine group logic in `amd_gpu_agent.cpp` is correct.

**[Important] Potential profiling regression in `amd_gpu_agent.cpp`:** The `!profiling_enabled()` gate was removed from the gfx125+ single-copy fast path. Now when profiling is enabled, single gfx125+ copies still enter `SubmitLinearCopyBodyWaitSignal`, which never emits SDMA timestamp packets — start/end timestamps will be zero/stale. The fan-out path (`SubmitFusedCoordinator`) handles profiling correctly, but the single-copy path at this call site does not. Suggest: restore the `!profiling_enabled()` guard, or add timestamp emission to `SubmitLinearCopyBodyWaitSignal`.

**[Important] Null dereference in `SubmitBodies` classic path (non-gfx125+) when `dep_signals` is empty:** `prologue_sig` is set to `nullptr` when `dep_signals.empty()`, but `prologue_sig->ValueLocation()` is unconditionally called below. All current internal callers supply a prologue signal, so the bug is latent, but `SubmitBodies` is a public virtual function — external callers with an empty `dep_signals` on non-gfx125+ hardware will crash. Suggest: assert `!dep_signals.empty()` in the classic-path branch and document the requirement.

saleelk:
> **[Important] Null dereference in `SubmitBodies` classic path (non-gfx125+) when `dep_signals` is empty:** `prologue_sig` is set to `nullptr` when `dep_signals.empty()`, but `prologue_sig->ValueLocation()` is unconditionally called below. All current internal callers supply a prologue signal, so the bug is latent, but `SubmitBodies` is a public virtual function — external callers with an empty `dep_signals` on non-gfx125+ hardware will crash. Suggest: assert `!dep_signals.empty()` in the classic-path branch and document the requirement.

ACK. I forgot to add TS packets in the linearwaitsignal path.

cfreeamd (approved):

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_blit_sdma.cpp
```diff
@@ -696,109 +683,114 @@ hsa_status_t BlitSdma<useGCR, scopeFields>::SubmitPrologue(
...
-    }
+  if (emit_prologue) {
+    BuildGCRCommand(command_addr, true);
+    command_addr += gcr_command_size();
+    bytes_written_[wrapped_index] = prior_bytes;
+    wrapped_index += gcr_command_size();
+  }
+
+  for (uint32_t i = 1; i < dep_signals.size(); ++i) {
+    BuildPoll64bCommand(command_addr, dep_signals[i]->ValueLocation(), 0);
```
dayatsin-amd, line 743: Does this need a  "if (is_gfx1250_plus)" and then use buildPollCommand otherwise, buildPoll64bCommand. It seems this could try to use 64-bit poll commands on older asics.

saleelk: No thee are dependent signals that are passed as dep signal to the ROCr profiling APIs, but the new packet only has space for 1 wait. So I pass that to the packet and the others are waited on outside the packet

dayatsin-amd (approved):

PR #8457, opened 2026-07-12
fix(amdsmi): open UALoE/IFoE session lazily to prevent amdsmi_init hang

marifamd:
## Summary

On UALink systems, `AMDSmiGPUDevice` opened the per-GPU **IFoE/UALoE generic-netlink** session in its **constructor**. When the Broadcom Thor Ultra `ifoe` driver is wedged (its firmware/MCDI not responding, holding the global generic-netlink lock), that `genl_ctrl_resolve("ifoe-cfg")` handshake blocks in an **uninterruptible (`D`-state) wait**. As a result `amdsmi_init(AMDSMI_INIT_AMD_GPUS)` — and therefore every `amd-smi` CLI verb — hung indefinitely, even for queries that never use fabric data. `SIGKILL`/`timeout` cannot recover it.

This was root-caused where `amd-smi version`, `amd-smi list`, and a pure `amdsmi_init(INIT_AMD_GPUS)` all wedged in `genl_rcv_msg`, while GPU sysfs/`rocminfo`/KFD were healthy.

## Fix

Open the UALoE session **lazily** on the first fabric query via `get_ualoe_handle()` (guarded by `std::once_flag`) instead of in the constructor. Initialization and non-fabric queries no longer touch the `ifoe` driver. Only an explicit UALink fabric query (`amdsmi_get_gpu_fabric_info` UALoE path / fabric telemetry) attempts the genl handshake, and only that query could stall on a wedged NIC.

A socket receive/send timeout was considered and rejected: the block is an uninterruptible in-kernel genl-mutex wait, which userspace socket timeouts cannot break. Not touching the driver unless asked is the correct fix.

- `src/amd_smi/amd_smi_gpu_device.cc`: remove `ualoe_open` from both ctors; add `open_ualoe_session()` and lazy `get_ualoe_handle()`.
- `include/amd_smi/impl/amd_smi_gpu_device.h`: lazy non-const `get_ualoe_handle()`, `std::once_flag`, `<mutex>`.
- `CHANGELOG.md`: Resolved Issues entry.

## JIRA ID

JIRA ID : AILITOOLS-284

## How tested

- Builds clean (`libamd_smi.so.26.5.0` links).
- **Verified on the affected hardware**: with the rebuilt library, `amdsmi_init(AMDSMI_INIT_AMD_GPUS)` returns `ret=0` in `0.00s` and `amdsmi_shut_down()` succeeds, versus a permanent `D`-state hang with the stock library.
- Not unit-testable in CI: reproduction requires UALink hardware with a wedged `ifoe` driver.

### DIFF
--- a/projects/amdsmi/CHANGELOG.md
+++ b/projects/amdsmi/CHANGELOG.md
@@ -109,6 +109,10 @@ Full documentation for amd_smi_lib is available at [https://rocm.docs.amd.com/pr
 
 ### Resolved Issues
 
+- **Fixed `amd-smi` hanging in `amdsmi_init()` on UALink systems when the IFoE driver is unresponsive**.  
+  - `AMDSmiGPUDevice` opened the per-GPU IFoE/UALoE generic-netlink session in its constructor, so `amdsmi_init(AMDSMI_INIT_AMD_GPUS)` (and every CLI verb) blocked in an uninterruptible netlink wait when the Broadcom IFoE driver was wedged, even for queries that never use fabric data.
+  - The UALoE session is now opened lazily on the first fabric query via `get_ualoe_handle()`, so initialization and non-fabric queries no longer touch the IFoE driver.
+
 - **Fixed `amd-smi set --power-cap` rejecting the minimum allowed value**.  
   - The lower bound is now inclusive, so setting the power cap to the exact minimum of the reported range (e.g. `210` when the range is 210-300W) succeeds instead of failing validation, matching the inclusive range shown in the error message.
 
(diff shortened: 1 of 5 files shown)


dmitrii-galantsev (changes requested):
test_fabric_lazy_init.py should be under tests/python_unittest

marifamd:
> test_fabric_lazy_init.py should be under tests/python_unittest

refactored in #7651

dmitrii-galantsev (approved):

gabrpham (approved):

PR #8459, opened 2026-07-13
[do not merge; reference] feat(hotswap): persist retarget cache to disk across processes

xintin:
## Summary
Stacks on #8274. Adds a disk-persistent tier to the hotswap retarget cache so repeat launches (any process) skip the COMGR B0->A0 retarget entirely.
- On 4x gfx1250 A0 with rccl-tests alltoall, warm launches drop from ~7 min to ~12 s, with 0 COMGR retargets and peak RSS ~90 GB -> ~11.5 GB.
- Keyed by the existing FNV-1a content hash; entries namespaced by a COMGR-identity salt (lib path, size, mtime, format version) so a toolchain change auto-invalidates.
- Atomic temp-file + rename publish; validated header on read; POSIX-only (no-op on Windows).
- Controls: HSA_HOTSWAP_CACHE_DIR, HSA_HOTSWAP_DISK_CACHE=0 to disable.
Also folds in the #8274 review fixes (entry-trampoline hash mix, bad_alloc cache-store log, RetargetCache test registration).
Depends on #8274.
## Test plan
- Cold run populates the cache (1 COMGR retarget per object + disk store).
- Warm run: 0 retargets, served from disk; results validated (0 OOB).
- Unit tests HotswapRewrite.RetargetCache* pass.


cfreeamd (commented):

`hotswap_rewrite_test.cc`, `hotswap.hpp`, and `test_categories.yaml` look correct on their own; one concern in `hotswap.cpp` around disk cache salt ordering.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -454,11 +671,129 @@ bool TryRetargetCodeObject(const CodeObjectView& code_object, hsa_agent_t agent,
     return false;
   }
 
+  const uint64_t cache_key = ComputeRetargetCacheKey(
+      code_object.data, code_object.size, decision->source_isa,
+      decision->target_isa, decision->request_entry_trampolines);
+
+#if !defined(_WIN32) && !defined(_WIN64)
+  const std::string disk_cache_dir = GetDiskCacheDir();
+  const uint64_t disk_cache_salt = disk_cache_dir.empty() ? 0ULL : ComgrIdentitySalt();
```

**[Important]** `disk_cache_salt` is computed here via `ComgrIdentitySalt()`, which reads `g_comgr_lib_path`. That global is only set when `GetComgrApi()` first resolves the COMGR library — inside `RetargetCodeObject()`, which is called *after* this point.

If `DecideHotswapRewrite()` (which runs before this line) does not itself call `GetComgrApi()`, then on the first call per process `g_comgr_lib_path` is empty and `ComgrIdentitySalt()` returns a constant (`kDiskCacheFormatVersion * 1000003ULL`). The disk entry is written with this wrong salt. On the second call COMGR is now loaded, the correct salt is computed, the disk read misses (salt mismatch), and COMGR is invoked again — only the entry written on the second call is ever a cache hit.

Fix options:
- Move salt computation to after `RetargetCodeObject()` returns (use it only in the write path, not the read path — or recompute for the write).
- Add an explicit `GetComgrApi()` call before computing `disk_cache_salt` to ensure COMGR is loaded.
- If `DecideHotswapRewrite()` always calls `GetComgrApi()` internally, this is a no-op — please confirm.

projects/rocr-runtime/rocrtst/suites/functional/hotswap/hotswap_rewrite_test.cc
```diff
@@ -553,4 +553,65 @@ TEST(HotswapRewrite, RuntimeLoadRewrittenLoadFailureFallsBackToOriginal) {
       rocr::hotswap::RetainedRewrittenElfBufferCountForTesting(executable), 0u);
 }
 
+TEST(HotswapRewrite, RetargetCacheServesSecondLoadFromCache) {
+  ResetRuntimeTestEnv();
+  if (!NewComgrHotswapApiAvailable()) return;
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** `ClearRetargetCacheForTesting()` only clears the in-memory `g_retarget_cache`; it does not remove disk entries. If a previous test run (or CI job) left a disk entry for `MakeRealCodeObjectView()`, the "First load: cache miss" comment below is wrong — it would be a disk cache hit instead. The assertions still pass since `LoadPath::kRewritten` is returned either way, but the test no longer exercises what it claims.

Consider one of: (a) expose `ClearDiskRetargetCacheForTesting()` and call it here, (b) set `HSA_HOTSWAP_DISK_CACHE=0` in the test environment, or (c) document that the disk tier is not exercised under rocrtst.

projects/rocr-runtime/rocrtst/suites/test_common/test_categories.yaml
```diff
@@ -108,6 +110,8 @@ test_categories:
       - "rocrtstPerf.AQL_Dispatch_Time_Single_Interrupt"
       - "rocrtstPerf.AQL_Dispatch_Time_Multi_SpinWait"
       - "rocrtstPerf.AQL_Dispatch_Time_Multi_Interrupt"
+      - "HotswapRewrite.RetargetCacheServesSecondLoadFromCache"
```
cfreeamd, line 111: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** Both new tests appear in all four category groups (lines 33–34, 113–114, 203–204, 295–296). `RetargetCacheServesSecondLoadFromCache` exercises a full COMGR retarget on the first load — running it four times across CI tiers adds cost. Is all-four intentional, or should it be restricted to a subset (e.g. Core + Extended only)? Other `HotswapRewrite.*` tests can be used as a reference for the expected category coverage.

xintin:
Thanks for prefetching it @cfreeamd.
This is a draft but your initial review helps in incorporating the changes. Please feel free to add any further suggestions. 
Thanks.

xintin:
Handled in #8649.

PR #8460, opened 2026-07-13
feat(amdsmi): implement unified amdsmi_get_link_topology on BM

koushikbillakanti-amd:
## Motivation

BM lacked the unified `amdsmi_get_link_topology` API that already exists on the host. This PR adds it to BM with the identical signature and `amdsmi_link_topology_t` return struct, giving callers consistent GPU-to-GPU link topology (type, weight, hops, status, FB sharing) across host and BM.

## Technical Details

- Added the `amdsmi_link_topology_t` struct and `amdsmi_get_link_topology()` in `amdsmi.h`/`amd_smi.cc`, wrapping the `rsmi_topo_*` link type/weight calls plus `rsmi_is_P2P_accessible` for FB sharing.
- Added Python bindings (`amdsmi_wrapper.py`, `amdsmi_interface.py`, `__init__.py`) with an ABI-matching 64-byte ctypes struct.
- Documented the API in `amdsmi-py-api.md` and added a hardware-independent unit test in `test_link_topology.py`.

## JIRA ID

JIRA ID : AILITOOLS-276

## Test Plan

- Built the library (`cmake --build ... --target amd_smi`) on a 2-GPU host (Navi31 + Navi48).
- Ran a runtime check calling `amdsmi_get_link_topology` between the GPUs.
- Ran the new unit test via the amdsmi python unit-test runner: `unit_tests.py -k link_topology -v`.

## Test Result

- API returned correct values: PCIE link type, `num_hops=2`, `weight=40`, `fb_sharing=1`, `sizeof(amdsmi_link_topology_t)==64` matching the host struct.
- Unit test: 7/7 passed.

## Submission Checklist

### DIFF
--- a/projects/amdsmi/CHANGELOG.md
+++ b/projects/amdsmi/CHANGELOG.md
@@ -134,6 +134,7 @@ GPU: 0
 - **Added unified `amdsmi_get_link_topology()` API (baremetal and host)**.  
   - New C API `amdsmi_get_link_topology()` returns a single `amdsmi_link_topology_t` aggregating link weight, link status, link type, hop count, and framebuffer-sharing capability between two GPUs.
   - New Python API `amdsmi_get_link_topology()` exposes the same data, mirroring the host interface for API parity.
+  - Behavior note for callers migrating from the component APIs: for a self-pair (identical source and destination), the call short-circuits to `link_type = AMDSMI_LINK_TYPE_INTERNAL`, `num_hops = 0`, `weight = 0`, `fb_sharing = 1`, and `link_status = AMDSMI_LINK_STATUS_ENABLED`. This is intentionally different from the underlying calls, where `amdsmi_topo_get_link_type(j, j)` reports PCIe with 2 hops and `amdsmi_topo_get_link_weight(j, j)` reports twice the NUMA node weight (typically 40).
 
 ### Changed
 
--- a/projects/amdsmi/CHANGELOG.md
+++ b/projects/amdsmi/CHANGELOG.md
@@ -131,6 +131,10 @@ GPU: 0
   - Both are UALoE-backed and report `AMDSMI_STATUS_NOT_SUPPORTED` (or the `UINT32_MAX` sentinel for `physical_acc_id`) on systems without an active UALoE session.
   - CLI: `amd-smi static --asic` now includes `PHYSICAL_ACC_ID`; new `amd-smi node --tray`/`-T` flag prints tray type and accelerator count.
 
+- **Added unified `amdsmi_get_link_topology()` API (baremetal and host)**.  
+  - New C API `amdsmi_get_link_topology()` returns a single `amdsmi_link_topology_t` aggregating link weight, link status, link type, hop count, and framebuffer-sharing capability between two GPUs.
+  - New Python API `amdsmi_get_link_topology()` exposes the same data, mirroring the host interface for API parity.
+
 ### Changed
 
 - **Bumped the library major version to 27.0.0** (breaking).  
--- a/projects/amdsmi/include/amd_smi/amdsmi.h
+++ b/projects/amdsmi/include/amd_smi/amdsmi.h
@@ -6894,9 +6894,11 @@ amdsmi_status_t amdsmi_topo_get_link_type(amdsmi_processor_handle processor_hand
  *  ::amdsmi_topo_get_link_type, not the number of physical xGMI links.
  *
  *  @note On @platform{gpu_bm_linux}, @p link_status is derived from the resolved
- *  link type, not live hardware state. The source only resolves a concrete type
- *  on success, so a successful call always reports ::AMDSMI_LINK_STATUS_ENABLED;
- *  the DISABLED, INACTIVE, and ERROR states are not produced on baremetal.
+ *  link type, not live hardware state. Every successful query resolves either
+ *  xGMI or PCIe, so a successful call always reports ::AMDSMI_LINK_STATUS_ENABLED.
+ *  ::AMDSMI_LINK_STATUS_DISABLED is only the initialized default returned with a
+ *  failure status; ::AMDSMI_LINK_STATUS_INACTIVE and ::AMDSMI_LINK_STATUS_ERROR
+ *  are host-only and never produced on baremetal.
  *
  *  @note @p fb_sharing is best-effort: a P2P query that cannot complete is
  *  reported as 0, indistinguishable from a genuine "not shared" result.
--- a/projects/amdsmi/include/amd_smi/amdsmi.h
+++ b/projects/amdsmi/include/amd_smi/amdsmi.h
@@ -1276,6 +1276,20 @@ typedef enum {
 } amdsmi_link_status_t;
 
 /**
+ * @brief Unified link topology information between two processors
+ *
+ * @cond @tag{gpu_bm_linux} @tag{host} @endcond
+ */
+typedef struct {
+  uint64_t weight;                   //!< link weight
+  amdsmi_link_status_t link_status;  //!< link status, derived from link type (not live HW status)
+  amdsmi_link_type_t link_type;      //!< type of the link
+  uint8_t num_hops;                  //!< number of hops
+  uint8_t fb_sharing;                //!< 1 if P2P framebuffer access is available, 0 otherwise
+  uint32_t reserved[10];
+} amdsmi_link_topology_t;
+
+/**
  * @brief Link Metrics
  *
  * @cond @tag{gpu_bm_linux} @tag{host} @endcond
@@ -6863,6 +6877,51 @@ amdsmi_status_t amdsmi_get_minmax_bandwidth_between_processors(
 amdsmi_status_t amdsmi_topo_get_link_type(amdsmi_processor_handle processor_handle_src,
                                           amdsmi_processor_handle processor_handle_dst,
                                           uint64_t* hops, amdsmi_link_type_t* type);
+
+/**
+ *  @brief Retrieve the unified link topology information between 2 GPUs
+ *
+ *  @ingroup tagHWTopology
+ *
+ *  @platform{gpu_bm_linux} @platform{host}
+ *
+ *  @details Writes the link weight, status, type, abstracted hop count, and
+ *  framebuffer sharing flag for the connection between @p processor_handle_src
+ *  and @p processor_handle_dst into @p topology_info.
+ *
+ *  This is the unified topology query that aggregates the information
+ *  otherwise obtained from ::amdsmi_topo_get_link_type and
+ *  ::amdsmi_topo_get_link_weight into a single ::amdsmi_link_topology_t
+ *  structure, matching the host interface.
+ *
+ *  @note The @p num_hops value is the abstracted topology step count reported
+ *  by ::amdsmi_topo_get_link_type, not the number of physical xGMI links.
+ *
+ *  @note On @platform{gpu_bm_linux}, the @p link_status is derived from link-type
+ *  resolvability rather than live hardware state: it is reported as
+ *  ::AMDSMI_LINK_STATUS_ENABLED when the connection resolves to a concrete link
+ *  type and ::AMDSMI_LINK_STATUS_DISABLED otherwise. The
+ *  ::AMDSMI_LINK_STATUS_INACTIVE and ::AMDSMI_LINK_STATUS_ERROR states are not
+ *  produced on baremetal.
+ *
+ *  @note The @p fb_sharing flag reflects whether the two devices can directly
+ *  access each other's framebuffer over P2P. It is best-effort: if the
+ *  underlying P2P-accessibility query cannot be completed @p fb_sharing is
+ *  reported as 0 (not shared), which is indistinguishable from a genuine
+ *  "not shared" result.
+ *
+ *  @param[in] processor_handle_src the source processor handle
+ *
+ *  @param[in] processor_handle_dst the destination processor handle
+ *
+ *  @param[out] topology_info A pointer to an ::amdsmi_link_topology_t to
+ *  which the link topology information should be written.
+ *
+ *  @return ::amdsmi_status_t | ::AMDSMI_STATUS_SUCCESS on success, non-zero on fail
+ */
+amdsmi_status_t amdsmi_get_link_topology(amdsmi_processor_handle processor_handle_src,
+                                         amdsmi_processor_handle processor_handle_dst,
+                                         amdsmi_link_topology_t* topology_info);
 
 /**
  *  @brief Retrieve the set of GPUs that are nearest to a given device
--- a/projects/amdsmi/include/amd_smi/amdsmi.h
+++ b/projects/amdsmi/include/amd_smi/amdsmi.h
@@ -6885,35 +6885,24 @@ amdsmi_status_t amdsmi_topo_get_link_type(amdsmi_processor_handle processor_hand
  *
  *  @platform{gpu_bm_linux} @platform{host}
  *
- *  @details Writes the link weight, status, type, abstracted hop count, and
- *  framebuffer sharing flag for the connection between @p processor_handle_src
- *  and @p processor_handle_dst into @p topology_info.
+ *  @details Aggregates the link weight, status, type, abstracted hop count, and
+ *  framebuffer-sharing flag for the connection between @p processor_handle_src
+ *  and @p processor_handle_dst into a single ::amdsmi_link_topology_t, mirroring
+ *  the host interface.
  *
- *  This is the unified topology query that aggregates the information
- *  otherwise obtained from ::amdsmi_topo_get_link_type and
- *  ::amdsmi_topo_get_link_weight into a single ::amdsmi_link_topology_t
- *  structure, matching the host interface.
+ *  @note @p num_hops is the abstracted topology step count from
+ *  ::amdsmi_topo_get_link_type, not the number of physical xGMI links.
  *
- *  @note The @p num_hops value is the abstracted topology step count reported
- *  by ::amdsmi_topo_get_link_type, not the number of physical xGMI links.
+ *  @note On @platform{gpu_bm_linux}, @p link_status is derived from the resolved
+ *  link type, not live hardware state. The source only resolves a concrete type
+ *  on success, so a successful call always reports ::AMDSMI_LINK_STATUS_ENABLED;
+ *  the DISABLED, INACTIVE, and ERROR states are not produced on baremetal.
  *
- *  @note On @platform{gpu_bm_linux}, the @p link_status is derived from link-type
- *  resolvability rather than live hardware state: it is reported as
- *  ::AMDSMI_LINK_STATUS_ENABLED when the connection resolves to a concrete link
- *  type and ::AMDSMI_LINK_STATUS_DISABLED otherwise. The
- *  ::AMDSMI_LINK_STATUS_INACTIVE and ::AMDSMI_LINK_STATUS_ERROR states are not
- *  produced on baremetal.
+ *  @note @p fb_sharing is best-effort: a P2P query that cannot complete is
+ *  reported as 0, indistinguishable from a genuine "not shared" result.
  *
- *  @note The @p fb_sharing flag reflects whether the two devices can directly
- *  access each other's framebuffer over P2P. It is best-effort: if the
- *  underlying P2P-accessibility query cannot be completed @p fb_sharing is
- *  reported as 0 (not shared), which is indistinguishable from a genuine
- *  "not shared" result.
- *
- *  @note The link type, weight, and framebuffer-sharing values are queried
- *  sequentially and not atomically, so a concurrent topology change between the
- *  individual queries may leave the returned fields reflecting slightly
- *  different points in time.
+ *  @note link type, weight, and fb_sharing are queried sequentially, not
+ *  atomically, so a concurrent topology change may skew the returned fields.
  *
  *  @param[in] processor_handle_src the source processor handle
  *
--- a/projects/amdsmi/include/amd_smi/amdsmi.h
+++ b/projects/amdsmi/include/amd_smi/amdsmi.h
@@ -6910,6 +6910,11 @@ amdsmi_status_t amdsmi_topo_get_link_type(amdsmi_processor_handle processor_hand
  *  reported as 0 (not shared), which is indistinguishable from a genuine
  *  "not shared" result.
  *
+ *  @note The link type, weight, and framebuffer-sharing values are queried
+ *  sequentially and not atomically, so a concurrent topology change between the
+ *  individual queries may leave the returned fields reflecting slightly
+ *  different points in time.
+ *
  *  @param[in] processor_handle_src the source processor handle
  *
  *  @param[in] processor_handle_dst the destination processor handle
--- a/projects/amdsmi/src/amd_smi/amd_smi.cc
+++ b/projects/amdsmi/src/amd_smi/amd_smi.cc
@@ -3276,6 +3276,23 @@ amdsmi_status_t amdsmi_get_link_topology(amdsmi_processor_handle processor_handl
   uint32_t src_id = src_device->get_gpu_id();
   uint32_t dst_id = dst_device->get_gpu_id();
 
+#ifdef ENABLE_WSL_BACKEND
+  // WSL exposes a single WDDM adapter with no xGMI fabric, so the native sysfs
+  // topology calls below do not apply. Mirror amdsmi_topo_get_link_type here: a
+  // self-pair is INTERNAL, any peer is a single PCIe hop.
+  if (src_device->backend() || dst_device->backend()) {
+    if (src_device == dst_device) {
+      topology_info->link_type = AMDSMI_LINK_TYPE_INTERNAL;
+      topology_info->fb_sharing = 1;
+    } else {
+      topology_info->link_type = AMDSMI_LINK_TYPE_PCIE;
+      topology_info->num_hops = 1;
+    }
+    topology_info->link_status = AMDSMI_LINK_STATUS_ENABLED;
+    return AMDSMI_STATUS_SUCCESS;
+  }
+#endif
+
   // A device has no link to itself; report a trivial internal, shared result.
   if (src_id == dst_id) {
     topology_info->link_type = AMDSMI_LINK_TYPE_INTERNAL;
--- a/projects/amdsmi/src/amd_smi/amd_smi.cc
+++ b/projects/amdsmi/src/amd_smi/amd_smi.cc
@@ -154,6 +154,24 @@ static_assert(offsetof(rsmi_gpu_metrics_t, apu_metrics) ==
                   offsetof(amdsmi_gpu_metrics_t, apu_metrics),
               "GPU metrics: apu_metrics (last field) offset mismatch");
 
+// amdsmi_link_topology_t crosses the baremetal/host ABI boundary and is consumed
+// by the generated Python bindings, so its 64-byte layout must stay in lockstep
+// with the host definition; guard the size and field offsets at compile time.
+static_assert(sizeof(amdsmi_link_topology_t) == 64,
+              "amdsmi_link_topology_t must remain 64 bytes to match the published ABI");
+static_assert(offsetof(amdsmi_link_topology_t, weight) == 0,
+              "amdsmi_link_topology_t: weight offset mismatch");
+static_assert(offsetof(amdsmi_link_topology_t, link_status) == 8,
+              "amdsmi_link_topology_t: link_status offset mismatch");
+static_assert(offsetof(amdsmi_link_topology_t, link_type) == 12,
+              "amdsmi_link_topology_t: link_type offset mismatch");
+static_assert(offsetof(amdsmi_link_topology_t, num_hops) == 16,
+              "amdsmi_link_topology_t: num_hops offset mismatch");
+static_assert(offsetof(amdsmi_link_topology_t, fb_sharing) == 17,
+              "amdsmi_link_topology_t: fb_sharing offset mismatch");
+static_assert(offsetof(amdsmi_link_topology_t, reserved) == 20,
+              "amdsmi_link_topology_t: reserved offset mismatch");
+
 static void copy_rsmi_gpu_metrics_to_amdsmi(const rsmi_gpu_metrics_t& rsmi_metrics,
                                             amdsmi_gpu_metrics_t* amdsmi_metrics) {
   if (amdsmi_metrics == nullptr) return;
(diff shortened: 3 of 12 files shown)


projects/amdsmi/include/amd_smi/amdsmi.h
```diff
@@ -1282,6 +1282,20 @@ typedef enum {
...
+ * @brief Unified link topology information between two processors
+ *
+ * @cond @tag{gpu_bm_linux} @tag{host} @endcond
+ */
+typedef struct {
+  uint64_t weight;                   //!< link weight
+  amdsmi_link_status_t link_status;  //!< HW status of the link
+  amdsmi_link_type_t link_type;      //!< type of the link
+  uint8_t num_hops;                  //!< number of hops
+  uint8_t fb_sharing;                //!< framebuffer sharing flag
```
gabrpham: Can you elaborate on the comment here, maybe saying something like "//!< 1 if P2P framebuffer access is available, 0 otherwise"?

projects/amdsmi/src/amd_smi/amd_smi.cc
```diff
@@ -3052,6 +3052,66 @@ amdsmi_status_t amdsmi_topo_get_p2p_status(amdsmi_processor_handle processor_han
...
+  amdsmi_status_t status = amd::smi::rsmi_to_amdsmi_status(rsmi_topo_get_link_type(
+      src_id, dst_id, &hops, reinterpret_cast<RSMI_IO_LINK_TYPE*>(&link_type)));
+  if (status != AMDSMI_STATUS_SUCCESS) return status;
+  topology_info->link_type = link_type;
+  topology_info->num_hops = static_cast<uint8_t>(hops);
+
+  // Link weight.
+  uint64_t weight = 0;
+  status = amd::smi::rsmi_to_amdsmi_status(rsmi_topo_get_link_weight(src_id, dst_id, &weight));
+  if (status != AMDSMI_STATUS_SUCCESS) return status;
```
gabrpham, line 3161: If we have a failure here, I think we also want to ensure that everything we return does so sensibly and here we have a partially filled out topology_info. If the user uses the structure for some reason after this failure, that could result in undefined behavior.

gabrpham (changes requested):
See comments below.

adam360x (approved):

projects/amdsmi/include/amd_smi/amdsmi.h
koushikbillakanti-amd: Done - reworded to your exact suggestion. The field comment now reads `//!< 1 if P2P framebuffer access is available, 0 otherwise` (`amdsmi.h` L1295). Addressed in commit 5e314b6b73.

projects/amdsmi/src/amd_smi/amd_smi.cc
koushikbillakanti-amd: Resolved - `topology_info` is now zero-initialized up front (`*topology_info = {}` plus `link_type = AMDSMI_LINK_TYPE_UNKNOWN` and `link_status = AMDSMI_LINK_STATUS_DISABLED`, `amd_smi.cc` L3083-3085) before any fallible call, and the result fields are written only at the end (L3122-3133) after every rsmi call succeeds. So on this failure return - and every early-return path - the caller now receives a fully well-defined, zeroed struct rather than a partial one, avoiding the UB you flagged. Addressed in commit 5e314b6b73.

projects/amdsmi/tests/amd_smi_test/functional/system/hw_topology_read_test.cc
```diff
@@ -196,6 +196,35 @@ void TestHWTopologyRead::Run(void) {
...
+            return;
+          } else {
+            CHK_ERR_ASRT(err)
+          }
+        } else {
+          // The unified API must agree with the component queries it aggregates.
+          const gpu_link_t& link = gpu_links[dv_ind_src][dv_ind_dst];
+          EXPECT_EQ(topology.weight, link.weight);
+          EXPECT_EQ(topology.num_hops, link.hops > 255 ? 255 : static_cast<uint8_t>(link.hops));
+          EXPECT_EQ(topology.fb_sharing, link.accessible ? 1 : 0);
```
oliveiradan, line 224: The test never validates `link_type`, after the weight assertion.

projects/amdsmi/src/amd_smi/amd_smi.cc
```diff
@@ -3102,6 +3120,72 @@ amdsmi_status_t amdsmi_topo_get_p2p_status(amdsmi_processor_handle processor_han
...
+
+  // Framebuffer sharing: two GPUs can share framebuffer memory when they are
+  // P2P-accessible (e.g. same xGMI hive). Best-effort: a missing P2P result
+  // must not fail the query.
+  uint8_t fb_sharing = 0;
+  bool accessible = false;
+  if (amd::smi::rsmi_to_amdsmi_status(rsmi_is_P2P_accessible(src_id, dst_id, &accessible)) ==
+      AMDSMI_STATUS_SUCCESS) {
+    fb_sharing = accessible ? 1 : 0;
+  }
```
oliveiradan, line 3173: Note: As we call `rsmi_topo_get_link_type()`, `rsmi_topo_get_link_weight()`, and `rsmi_is_P2P_accessible()` in sequence, a concurrent topology change could affect some of the info/results "mid-flight". It is worth a one-line `@note` about it.

projects/amdsmi/src/amd_smi/amd_smi.cc
```diff
@@ -3102,6 +3120,72 @@ amdsmi_status_t amdsmi_topo_get_p2p_status(amdsmi_processor_handle processor_han
...
+  uint64_t weight = 0;
+  status = amd::smi::rsmi_to_amdsmi_status(rsmi_topo_get_link_weight(src_id, dst_id, &weight));
+  if (status != AMDSMI_STATUS_SUCCESS) return status;
+
+  // Framebuffer sharing: two GPUs can share framebuffer memory when they are
+  // P2P-accessible (e.g. same xGMI hive). Best-effort: a missing P2P result
+  // must not fail the query.
+  uint8_t fb_sharing = 0;
+  bool accessible = false;
+  if (amd::smi::rsmi_to_amdsmi_status(rsmi_is_P2P_accessible(src_id, dst_id, &accessible)) ==
```
oliveiradan, line 3170: How do we distinguish "not shared" from "P2P query failed" here?

oliveiradan (changes requested):
@koushikbillakanti-amd , I added a few more comments for you. Also, the Rust wrapper needs to get updated.

projects/amdsmi/tests/amd_smi_test/functional/system/hw_topology_read_test.cc
koushikbillakanti-amd: Added an amdsmi_link_type_t link_type field to gpu_link_t, capture the raw enum from amdsmi_topo_get_link_type before p2p_status overwrites it, and now assert EXPECT_EQ(topology.link_type, link.link_type); right after the weight assertion.

projects/amdsmi/src/amd_smi/amd_smi.cc
koushikbillakanti-amd: Added an @note documenting that link type, weight, and framebuffer-sharing are queried sequentially and not atomically, so a concurrent topology change between the individual queries may leave the returned fields reflecting slightly different points in time.

projects/amdsmi/src/amd_smi/amd_smi.cc
koushikbillakanti-amd: Clarified in the comment that a failed or unsupported P2P query is intentionally reported the same as "not shared" (fb_sharing = 0); callers that must tell the two apart should call amdsmi_is_P2P_accessible() directly (also noted in the header).

oliveiradan (changes requested):
It doesn't seem we ran `/amdsmi-review-pr` on this PR. There are a few `should fix` when looking at the summary report.

projects/amdsmi/include/amd_smi/amdsmi.h
```diff
@@ -6830,6 +6844,56 @@ amdsmi_status_t amdsmi_topo_get_link_type(amdsmi_processor_handle processor_hand
...
+ *  This is the unified topology query that aggregates the information
+ *  otherwise obtained from ::amdsmi_topo_get_link_type and
+ *  ::amdsmi_topo_get_link_weight into a single ::amdsmi_link_topology_t
+ *  structure, matching the host interface.
+ *
+ *  @note The @p num_hops value is the abstracted topology step count reported
+ *  by ::amdsmi_topo_get_link_type, not the number of physical xGMI links.
+ *
+ *  @note On @platform{gpu_bm_linux}, the @p link_status is derived from link-type
+ *  resolvability rather than live hardware state: it is reported as
```
oliveiradan: Nit: I think we probably need to clarify the contract here, for example: 
```cpp
...
 *  @note On @platform{gpu_bm_linux}, @p link_status is derived from the resolved
 *  link type, not live hardware state. The baremetal source resolves only PCIe or
 *  xGMI on a successful query, so a successful call always reports
 *  ::AMDSMI_LINK_STATUS_ENABLED; the other states are not produced.
...
```

projects/amdsmi/tests/amd_smi_test/functional/system/hw_topology_read_test.cc
```diff
@@ -196,6 +199,36 @@ void TestHWTopologyRead::Run(void) {
...
+            CHK_ERR_ASRT(err)
+          }
+        } else {
+          // The unified API must agree with the component queries it aggregates.
+          const gpu_link_t& link = gpu_links[dv_ind_src][dv_ind_dst];
+          EXPECT_EQ(topology.weight, link.weight);
+          EXPECT_EQ(topology.link_type, link.link_type);
+          EXPECT_EQ(topology.num_hops, link.hops > 255 ? 255 : static_cast<uint8_t>(link.hops));
+          EXPECT_EQ(topology.fb_sharing, link.accessible ? 1 : 0);
+          if (topology.link_type == AMDSMI_LINK_TYPE_NOT_APPLICABLE ||
```
oliveiradan: The if/else mirrors the implementation's condition, and the `DISABLED` branch never runs; it is like we have two-option branch coverage, but only deliver one.  Maybe we could try something like this? 

```cpp
...
    EXPECT_TRUE((topology.link_type == AMDSMI_LINK_TYPE_PCIE) || 
                (topology.link_type == AMDSMI_LINK_TYPE_XGMI));
    EXPECT_EQ(topology.link_status, AMDSMI_LINK_STATUS_ENABLED);
...
```

The `EXPECT_TRUE` is where we try to catch unexpected state/error. If the code ever starts resolving `NOT_APPLICABLE` or `UNKNOWN` **on success**, this fires and points whoever hits it to the `@note` that then needs updating. Without it, the `EXPECT_EQ` would fail with no indication of why

projects/amdsmi/include/amd_smi/amdsmi.h
```diff
@@ -6830,6 +6844,56 @@ amdsmi_status_t amdsmi_topo_get_link_type(amdsmi_processor_handle processor_hand
...
+ *  ::AMDSMI_LINK_STATUS_ENABLED when the connection resolves to a concrete link
+ *  type and ::AMDSMI_LINK_STATUS_DISABLED otherwise. The
```
oliveiradan: Question: Do we have any cases where either `DISABLED` or `INACTIVE` is valid for real links? I don't think we do, but if that is the case, then just adjusting the documentation won't fix the problem, which then becomes a missing implementation, right?

projects/amdsmi/src/amd_smi/amd_smi.cc
```diff
@@ -3102,6 +3120,74 @@ amdsmi_status_t amdsmi_topo_get_p2p_status(amdsmi_processor_handle processor_han
   return amd::smi::rsmi_to_amdsmi_status(rstatus);
 }
 
+amdsmi_status_t amdsmi_get_link_topology(amdsmi_processor_handle processor_handle_src,
+                                         amdsmi_processor_handle processor_handle_dst,
+                                         amdsmi_link_topology_t* topology_info) {
```
oliveiradan, line 3257: What happens here if `source` and `destination` are the same?

projects/amdsmi/src/amd_smi/amd_smi.cc
```diff
@@ -3102,6 +3120,74 @@ amdsmi_status_t amdsmi_topo_get_p2p_status(amdsmi_processor_handle processor_han
...
+
+  // rsmi_topo_get_link_type writes an RSMI_IO_LINK_TYPE through the cast below, so
+  // amdsmi_link_type_t must be at least as wide to avoid an out-of-bounds write.
+  static_assert(sizeof(amdsmi_link_type_t) >= sizeof(RSMI_IO_LINK_TYPE),
+                "amdsmi_link_type_t must be at least as wide as RSMI_IO_LINK_TYPE");
+
+  uint64_t hops = 0;
+  amdsmi_link_type_t link_type = AMDSMI_LINK_TYPE_UNKNOWN;
+  amdsmi_status_t status = amd::smi::rsmi_to_amdsmi_status(rsmi_topo_get_link_type(
+      src_id, dst_id, &hops, reinterpret_cast<RSMI_IO_LINK_TYPE*>(&link_type)));
```
oliveiradan: It is important to understand that we are guarding only the `type size`, not enumerator correspondence. 
In fact just by looking at it, 3/5 `rsmi` enums map to different `amdsmi` meanings.

oliveiradan (changes requested):
@koushikbillakanti-amd I just left a few comments for you to take a look.

projects/amdsmi/include/amd_smi/amdsmi.h
koushikbillakanti-amd: Good call. I tightened the note so it states the contract directly: on baremetal a successful call always reports ENABLED (the source only resolves PCIe or xGMI), and DISABLED, INACTIVE, and ERROR are called out as not produced there.

projects/amdsmi/include/amd_smi/amdsmi.h
koushikbillakanti-amd: Right, they are not valid on this baremetal path. The rocm-smi source only resolves PCIe or xGMI on success, so those states never occur here. It is not a missing implementation, just not exposed by this source, and the note now says so.

projects/amdsmi/tests/amd_smi_test/functional/system/hw_topology_read_test.cc
koushikbillakanti-amd: Done, I went with your version: EXPECT_TRUE that the type is PCIe or xGMI, then EXPECT_EQ the status to ENABLED. It now catches an unexpected type on success instead of just mirroring the implementation.

projects/amdsmi/src/amd_smi/amd_smi.cc
koushikbillakanti-amd: Added an explicit self check at the top: when source and destination match it returns INTERNAL, ENABLED, fb_sharing 1, and 0 hops. That matches the WSL path and avoids leaning on undefined rocm-smi behavior for self.

projects/amdsmi/src/amd_smi/amd_smi.cc
koushikbillakanti-amd: Agreed, the size guard was misleading. I dropped the reinterpret_cast and the assert, and now map explicitly (PCIEXPRESS to PCIE, XGMI to XGMI, everything else to UNKNOWN) so mismatched enumerators cannot slip through.

projects/amdsmi/src/amd_smi/amd_smi.cc
```diff
@@ -3235,6 +3252,82 @@ amdsmi_status_t amdsmi_topo_get_p2p_status(amdsmi_processor_handle processor_han
...
+  amdsmi_status_t r = get_gpu_device_from_handle(processor_handle_src, &src_device);
+  if (r != AMDSMI_STATUS_SUCCESS) return r;
+  r = get_gpu_device_from_handle(processor_handle_dst, &dst_device);
+  if (r != AMDSMI_STATUS_SUCCESS) return r;
+
+  uint32_t src_id = src_device->get_gpu_id();
+  uint32_t dst_id = dst_device->get_gpu_id();
+
+  // A device has no link to itself; report a trivial internal, shared result.
+  if (src_id == dst_id) {
```
oliveiradan, line 3297: Document this behavior somewhere (`CHANGELOG` or similar). A caller migrating from another component API to this one will get a different answer for the same pair. We are short-circuiting  with `INTERNAL/0/0/fb=1`. `rsmi_topo_get_link_type(j, j)` for the same case returns: `SUCCESS` with `PCIEXPRESS/hops=2`. `rsmi_topo_get_link_weight(k, k)` for the same situation returns `2 * numa_node_weight`, usually `40`.
@bill-shuzhou-liu, what do you think?

projects/amdsmi/tests/amd_smi_test/functional/system/hw_topology_read_test.cc
```diff
@@ -106,6 +107,7 @@ void TestHWTopologyRead::Run(void) {
     for (uint32_t dv_ind_dst = 0; dv_ind_dst < num_devices; dv_ind_dst++) {
       if (dv_ind_src == dv_ind_dst) {
         gpu_links[dv_ind_src][dv_ind_dst].type = "X";
+        gpu_links[dv_ind_src][dv_ind_dst].link_type = AMDSMI_LINK_TYPE_UNKNOWN;
```
oliveiradan: I am not sure this conditional or the else for that matter ever calls `amdsmi_get_link_topology()`. 
Also, we preload it with `AMDSMI_LINK_TYPE_UNKNOWN` while the implementation returns `INTERNAL`?

projects/amdsmi/tests/python/unit/system/test_link_topology.py
```diff
@@ -0,0 +1,115 @@
...
+            set(result), {"weight", "link_status", "link_type", "num_hops", "fb_sharing"}
+        )
+        self.assertEqual(result["weight"], 42)
+        self.assertEqual(result["link_status"], 0)
+        self.assertEqual(result["link_type"], 2)
+        self.assertEqual(result["num_hops"], 3)
+        self.assertEqual(result["fb_sharing"], 1)
+
+    def test_disabled_link_maps_status_and_capped_hops(self):
+        # Complements the enabled case: disabled mapping and capped (255) hops.
```
oliveiradan: Does this execute anything production-related that `test_success_path_returns_mapped_dict()` doesn't already?

projects/amdsmi/include/amd_smi/amdsmi.h
```diff
@@ -6864,6 +6878,45 @@ amdsmi_status_t amdsmi_topo_get_link_type(amdsmi_processor_handle processor_hand
...
+ *  and @p processor_handle_dst into a single ::amdsmi_link_topology_t, mirroring
+ *  the host interface.
+ *
+ *  @note @p num_hops is the abstracted topology step count from
+ *  ::amdsmi_topo_get_link_type, not the number of physical xGMI links.
+ *
+ *  @note On @platform{gpu_bm_linux}, @p link_status is derived from the resolved
+ *  link type, not live hardware state. The source only resolves a concrete type
+ *  on success, so a successful call always reports ::AMDSMI_LINK_STATUS_ENABLED;
+ *  the DISABLED, INACTIVE, and ERROR states are not produced on baremetal.
```
oliveiradan: This contradicts `py-api.md`. "the DISABLED, INACTIVE, and ERROR states are not produced on baremetal"
`py-api.md`: "only ENABLED (0) and DISABLED (1) are produced."
Based on ` rocm_smi.cc`, it looks more like: "every SUCCESS path writes XGMI or PCIEXPRESS, so DISABLED is unreachable on success"

oliveiradan (changes requested):
@koushikbillakanti-amd , please take a look at my last notes/questions.
@bill-shuzhou-liu, please check if I am missing anything else.

projects/amdsmi/src/amd_smi/amd_smi.cc
koushikbillakanti-amd: Good call. I documented this in the CHANGELOG so anyone migrating off the component APIs sees the self pair now returns INTERNAL with 0 hops and 0 weight, unlike the old PCIe with 2 hops and weight near 40.

projects/amdsmi/tests/amd_smi_test/functional/system/hw_topology_read_test.cc
koushikbillakanti-amd: Good catches. The diagonal now actually calls amdsmi_get_link_topology and asserts INTERNAL with ENABLED, and I changed the preload from UNKNOWN to INTERNAL to match the implementation. The off diagonal branch already exercises the same API.

projects/amdsmi/tests/python/unit/system/test_link_topology.py
koushikbillakanti-amd: It did not add anything the success path test was not already covering, since the Python binding is just a field copy. So I removed it.

projects/amdsmi/include/amd_smi/amdsmi.h
koushikbillakanti-amd: Agreed, and your phrasing is clearer. I updated both the header note and py-api.md to say every successful query resolves xGMI or PCIe, so success always reports ENABLED and DISABLED is only the initialized default returned with a failure status.

projects/amdsmi/src/amd_smi/amd_smi.cc
```diff
@@ -3235,6 +3252,82 @@ amdsmi_status_t amdsmi_topo_get_p2p_status(amdsmi_processor_handle processor_han
...
+
+  if (topology_info == nullptr) {
+    return AMDSMI_STATUS_INVAL;
+  }
+
+  // Well-defined defaults so every error path leaves a coherent result.
+  *topology_info = {};
+  topology_info->link_type = AMDSMI_LINK_TYPE_UNKNOWN;
+  topology_info->link_status = AMDSMI_LINK_STATUS_DISABLED;
+
```
bill-shuzhou-liu, line 3268: we may need to add the WSL guard to handle the WSL use case (https://amd.atlassian.net/wiki/spaces/AMDSMI/pages/1807502880/amdsmi+in+WSL)
#ifdef ENABLE_WSL_BACKEND
... ...
#endif

koushikbillakanti-amd: Good catch. I added the ENABLE_WSL_BACKEND guard so on WSL it skips the native sysfs topology calls and mirrors amdsmi_topo_get_link_type, a self pair returns INTERNAL and any peer returns a single PCIe hop.

projects/amdsmi/include/amd_smi/amdsmi.h
```diff
@@ -6864,6 +6878,47 @@ amdsmi_status_t amdsmi_topo_get_link_type(amdsmi_processor_handle processor_hand
...
+ *  @platform{gpu_bm_linux} @platform{host}
+ *
+ *  @details Aggregates the link weight, status, type, abstracted hop count, and
+ *  framebuffer-sharing flag for the connection between @p processor_handle_src
+ *  and @p processor_handle_dst into a single ::amdsmi_link_topology_t, mirroring
+ *  the host interface.
+ *
+ *  @note @p num_hops is the abstracted topology step count from
+ *  ::amdsmi_topo_get_link_type, not the number of physical xGMI links.
+ *
```
bill-shuzhou-liu, line 6895: The `@note` here says link_status is derived from link_type and "every successful query resolves either xGMI or PCIe" — but the self-pair branch below (`amd_smi.cc:3297-3302`) resolves to `AMDSMI_LINK_TYPE_INTERNAL`, not xGMI/PCIe, and sets `link_status = ENABLED` directly rather than via the derivation this note describes. 

Also, on MI4XX, the link type could be IFoE (pending to implement).

projects/amdsmi/tests/amd_smi_test/functional/system/hw_topology_read_test.cc
```diff
@@ -106,11 +107,34 @@ void TestHWTopologyRead::Run(void) {
...
+
+        // Self-pair: the unified API short-circuits a device to itself.
+        amdsmi_link_topology_t self_topology = {};
+        DISPLAY_AMDSMI_API("amdsmi_get_link_topology",
+                           "gpu=" + std::to_string(dv_ind_src) + "," + std::to_string(dv_ind_dst),
+                           VERB(STANDARD));
+        err = amdsmi_get_link_topology(processor_handles_[dv_ind_src],
+                                       processor_handles_[dv_ind_dst], &self_topology);
+        DISPLAY_AMDSMI_STATUS(VERB(STANDARD), __FILE__, __LINE__, err, AMDSMI_STATUS_SUCCESS);
+        if (err != AMDSMI_STATUS_SUCCESS) {
```
bill-shuzhou-liu, line 127: if `amdsmi_get_link_topology` reports NOT_SUPPORTED on a platform where the pre-existing component APIs (`amdsmi_topo_get_link_type`, `amdsmi_topo_get_link_weight`, `amdsmi_topo_get_p2p_status`, etc.) still work, the entire rest of `Run()` — all pre-existing topology coverage — is silently skipped, and the test still reports pass (GTest doesn't distinguish an early return from full completion). Consider `continue` instead of `return`, or probing support once before the loops.

projects/amdsmi/tests/amd_smi_test/functional/system/hw_topology_read_test.cc
```diff
@@ -106,11 +107,34 @@ void TestHWTopologyRead::Run(void) {
...
         gpu_links[dv_ind_src][dv_ind_dst].weight = 0;
         gpu_links[dv_ind_src][dv_ind_dst].accessible = true;
         gpu_links[dv_ind_src][dv_ind_dst].cap = {UINT8_MAX, UINT8_MAX, UINT8_MAX, UINT8_MAX,
                                                  UINT8_MAX};
+
+        // Self-pair: the unified API short-circuits a device to itself.
+        amdsmi_link_topology_t self_topology = {};
+        DISPLAY_AMDSMI_API("amdsmi_get_link_topology",
+                           "gpu=" + std::to_string(dv_ind_src) + "," + std::to_string(dv_ind_dst),
+                           VERB(STANDARD));
```
bill-shuzhou-liu, line 123: No test exercises `amdsmi_get_link_topology(..., nullptr)` for the `topology_info` output pointer, even though `amd_smi.cc:3260` explicitly guards against it (`if (topology_info == nullptr) return AMDSMI_STATUS_INVAL;`).

projects/amdsmi/tests/python/unit/system/test_link_topology.py
```diff
@@ -0,0 +1,90 @@
...
+            result = amdsmi.amdsmi_interface.amdsmi_get_link_topology(src, dst)
+
+        self.assertEqual(
+            set(result), {"weight", "link_status", "link_type", "num_hops", "fb_sharing"}
+        )
+        self.assertEqual(result["weight"], 42)
+        self.assertEqual(result["link_status"], 0)
+        self.assertEqual(result["link_type"], 2)
+        self.assertEqual(result["num_hops"], 3)
+        self.assertEqual(result["fb_sharing"], 1)
```
bill-shuzhou-liu, line 90: No Python/ctypes-layer unit test locks in the self-pair (`src == dst`) short-circuit behavior — only the hardware-dependent C++ GTest exercises it (which most CI/dev sandboxes can't run without a GPU).

Detail: if the ctypes struct mapping or dict-key derivation regresses specifically for the self-pair branch, nothing at this layer would catch it. Suggest adding a `test_success_path_self_pair` mock test mirroring `test_success_path_returns_mapped_dict` above, asserting `link_type=INTERNAL`, `num_hops=0`, `weight=0`, `fb_sharing=1`.

projects/amdsmi/src/amd_smi/amd_smi.cc
```diff
@@ -3235,6 +3252,99 @@ amdsmi_status_t amdsmi_topo_get_p2p_status(amdsmi_processor_handle processor_han
...
+  bool accessible = false;
+  if (amd::smi::rsmi_to_amdsmi_status(rsmi_is_P2P_accessible(src_id, dst_id, &accessible)) ==
+      AMDSMI_STATUS_SUCCESS) {
+    fb_sharing = accessible ? 1 : 0;
+  }
+
+  topology_info->link_type = link_type;
+  topology_info->num_hops = hops > 255 ? 255 : static_cast<uint8_t>(hops);  // clamp to uint8_t
+  topology_info->weight = weight;
+  // Derived, not read from hardware: a concrete type is ENABLED, UNKNOWN is DISABLED.
```
bill-shuzhou-liu: On baremetal, `link_status` here is a pure function of `link_type` (`UNKNOWN → DISABLED`, else `ENABLED`) and carries no information beyond what `link_type` already encodes. It also shares its name with the CLI's pre-existing `link_status` JSON field (`amdsmi_cli/subcommands/topology.py`), which is derived from `amdsmi_is_P2P_accessible` instead — a different signal.

Detail: a pair that's xGMI/PCIe-reachable but not P2P-accessible would report `ENABLED` from this new API and `DISABLED` from the existing CLI field for what reads as the same concept. Not currently conflated in any single code path, but worth an explicit doc note that this is a host-parity artifact distinct from the CLI's existing predicate, to avoid future confusion.

projects/amdsmi/src/amd_smi/amd_smi.cc
```diff
@@ -3235,6 +3252,99 @@ amdsmi_status_t amdsmi_topo_get_p2p_status(amdsmi_processor_handle processor_han
...
+    } else {
+      topology_info->link_type = AMDSMI_LINK_TYPE_PCIE;
+      topology_info->num_hops = 1;
+    }
+    topology_info->link_status = AMDSMI_LINK_STATUS_ENABLED;
+    return AMDSMI_STATUS_SUCCESS;
+  }
+#endif
+
+  // A device has no link to itself; report a trivial internal, shared result.
```
bill-shuzhou-liu: **[SUGGESTION]** The WSL branch just above (L3283) checks pointer identity (`src_device == dst_device`) for "same processor", while this native branch checks `src_id == dst_id` (post-`get_gpu_id()` values) — two different equality bases for the same concept within one function. Currently equivalent in practice (one GPU ID per device, no aliasing found), and this split mirrors the pre-existing sibling `amdsmi_topo_get_link_type`, so not a new defect — but worth unifying on one basis if compute-partition sub-handles sharing a physical GPU ID are ever introduced.

projects/amdsmi/include/amd_smi/amdsmi.h
koushikbillakanti-amd: Good catch. The docs now explicitly cover the INTERNAL/ENABLED self pair and distinguish it from current PCIe/xGMI peers, with MI4XX IFoE reporting marked as not yet implemented.

projects/amdsmi/tests/amd_smi_test/functional/system/hw_topology_read_test.cc
koushikbillakanti-amd: You're right, that could hide the remaining coverage. Both unified-query paths now use continue on NOT_SUPPORTED so the existing topology tests keep running.

projects/amdsmi/tests/amd_smi_test/functional/system/hw_topology_read_test.cc
koushikbillakanti-amd: Added an explicit nullptr output check expecting AMDSMI_STATUS_INVAL before the topology loops. It uses EXPECT_EQ so a failure is recorded without stopping the other checks.

projects/amdsmi/tests/python/unit/system/test_link_topology.py
koushikbillakanti-amd: Added the self-pair mock test with the same handle passed twice, checking INTERNAL/ENABLED, zero weight and hops, and fb_sharing=1. All seven binding tests pass locally.

projects/amdsmi/src/amd_smi/amd_smi.cc
koushikbillakanti-amd: Agreed, those are different signals. The C and Python docs now distinguish host-parity link_status from the topology CLI's P2P check and explain why their values can differ.

bill-shuzhou-liu (approved):

oliveiradan (approved):

projects/amdsmi/src/amd_smi/amd_smi.cc
```diff
@@ -154,6 +154,23 @@ static_assert(offsetof(rsmi_gpu_metrics_t, apu_metrics) ==
...
+static_assert(offsetof(amdsmi_link_topology_t, link_status) == 8,
+              "amdsmi_link_topology_t: link_status offset mismatch");
+static_assert(offsetof(amdsmi_link_topology_t, link_type) == 12,
+              "amdsmi_link_topology_t: link_type offset mismatch");
+static_assert(offsetof(amdsmi_link_topology_t, num_hops) == 16,
+              "amdsmi_link_topology_t: num_hops offset mismatch");
+static_assert(offsetof(amdsmi_link_topology_t, fb_sharing) == 17,
+              "amdsmi_link_topology_t: fb_sharing offset mismatch");
+static_assert(offsetof(amdsmi_link_topology_t, reserved) == 20,
+              "amdsmi_link_topology_t: reserved offset mismatch");
```
marifamd, line 172: Why do we have this?

marifamd: Compile time check coverage is fine, but this isn't the place for it

projects/amdsmi/CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ Full documentation for amd_smi_lib is available at [https://rocm.docs.amd.com/pr
 
 ### Added
 
+- **Added unified `amdsmi_get_link_topology()` API on baremetal**.  
+  - New C API `amdsmi_get_link_topology()` returns a single `amdsmi_link_topology_t` aggregating link weight, link status, link type, hop count, and framebuffer-sharing capability between two GPUs, matching the existing host interface.
+  - New Python and Rust bindings expose the same topology data.
+  - Behavior note for callers migrating from the component APIs: for a self-pair (identical source and destination), the call short-circuits to `link_type = AMDSMI_LINK_TYPE_INTERNAL`, `num_hops = 0`, `weight = 0`, `fb_sharing = 1`, and `link_status = AMDSMI_LINK_STATUS_ENABLED`. This is intentionally different from the underlying calls, where `amdsmi_topo_get_link_type(j, j)` reports PCIe with 2 hops and `amdsmi_topo_get_link_weight(j, j)` reports twice the NUMA node weight (typically 40).
```
marifamd, line 14: make this bullet more readable and understandable

marifamd (changes requested):

projects/amdsmi/src/amd_smi/amd_smi.cc
koushikbillakanti-amd: Moved the ABI checks out of the library implementation and into the topology test. All seven checks are unchanged.

projects/amdsmi/CHANGELOG.md
koushikbillakanti-amd: Split the changelog note into short bullets, with the self-pair return values and migration differences listed separately.

PR #8465, opened 2026-07-13
[rocSHMEM] Support symmetric user-buffer registration on the GDA backend

avinashkethineedi:
## Motivation
- `rocshmem_buffer_register_symmetric` previously worked only on the IPC (single-node) backend. This PR extends symmetric registration of user-allocated VMM buffers to the GDA (NIC) backend so registered buffers are reachable over the network, while node-local peers still take an IPC fast path. It also factors the shared registration machinery out of the IPC backend so both backends use a single implementation.

## Technical Details

- **Shared base class:** Move the symmetric-registration machinery (VMM alias create/teardown, IPC handle export/exchange, peer-mapping publication, and the collective consensus helpers `all_pes_agree`/`all_pes_succeeded`/
  `symm_lengths_match`) into `Backend`. The IPC backend is reduced to delegate to `register_ipc_symm_region`/`unregister_ipc_symm_region`; behavior unchanged.
- **GDA host registration:** `buffer_register_symmetric`/  `buffer_unregister_symmetric` (collective, VMM-only) register the rocSHMEM-owned alias with every NIC via the dmabuf path (retaining the VMM backing handle and dmabuf fd for the MR's lifetime), exchange each PE's alias base and per-PE/per-NIC keys, and publish them to the device, with rollback on partial failure. When IPC is available, the buffer is also exposed to node-local peers over IPC (mixed IPC/NIC, disable via `ROCSHMEM_DISABLE_MIXED_IPC`). Concurrent registrations are bounded by `ROCSHMEM_MAX_SYMM_REGIONS`.
- **Device lookup:** `gda_symm_table.hpp` defines `QpSymmEntry`, a self-contained
  record pre-specialized to one `(dest_pe, nic_idx)`. The backend builds a flat table (`num_pes * num_nics * capacity`); each `QueuePair` holds the slice for its own `(dest_pe, nic_idx)` plus a shared `symm_count` and a cached `remote_heap_base`. `translate_remote`/`local_lkey` resolve heap addresses as pure register arithmetic and registered buffers by scanning the QP's slice (no pointer chasing). AMOs and alltoall route through these resolvers; heap behavior is unchanged.
- **IPC fast path:** device `ipc_peer_ptr` (via `IpcImpl::ipcPeerPtr`) routes node-local peers through IPC in `p`/`g`, falling back to the NIC when unavailable.
- **RO backend:** override the registration entrypoints to warn and return `ROCSHMEM_ERROR`, since symmetric registration is unsupported there.
- **Docs:** update CHANGELOG and the memory management API docs.

## JIRA ID
- AIROCSHMEM-411

## Test Plan
- Build `ipc_single -DUSE_HEAP_DEVICE_VMM_POSIX=ON -DUSE_HEAP_DEVICE_FINEGRAIN=OFF` (IPC, single-node) and run `rocshmem_user_mem_symm_test` across ranks.
- Build `gda_mlx5 -DUSE_IPC=ON -DUSE_HEAP_DEVICE_VMM_POSIX=ON -DUSE_HEAP_DEVICE_FINEGRAIN=OFF` (GDA + mixed IPC) and run the same test.
- Exercise non-power-of-two rank counts to cover the registration collective and the NIC/IPC dispatch paths.

## Test Result
- IPC single-node: `rocshmem_user_mem_symm_test` (VMM) PASS on all PEs at np = 3, 5, 6, 7, 8.
- GDA (mlx5, mixed IPC):

## Submission Checklist

### DIFF
--- a/projects/rocshmem/docs/api/memory_management.rst
+++ b/projects/rocshmem/docs/api/memory_management.rst
@@ -175,7 +175,7 @@ Backend behavior:
 
 * **IPC backend:** peer buffers are mapped into the local address space, so
   registered buffers are reached directly through inter-process copies.
-* **GDA backend:** registered buffers are always registered with each NIC's
+* **GDA backend:** registered buffers are registered with each NIC's
   protection domain and are reachable over the NIC by every peer. When the IPC
   fast path is additionally available for node-local peers (see
   ``ROCSHMEM_DISABLE_MIXED_IPC``), the buffer is also exposed to those peers
--- a/projects/rocshmem/docs/api/memory_management.rst
+++ b/projects/rocshmem/docs/api/memory_management.rst
@@ -181,6 +181,8 @@ Backend behavior:
   ``ROCSHMEM_DISABLE_MIXED_IPC``), the buffer is also exposed to those peers
   over IPC, and their accesses use the IPC path in preference to the NIC; the
   NIC registration remains in place and is used for all other peers.
+* **RO (reverse offload) backend:** symmetric user-buffer registration is not
+  supported; this routine returns ``NULL``.
 
 ROCSHMEM_BUFFER_UNREGISTER_SYMMETRIC
 ------------------------------------
--- a/projects/rocshmem/docs/api/memory_management.rst
+++ b/projects/rocshmem/docs/api/memory_management.rst
@@ -168,6 +168,19 @@ Notes:
   the job.
 * The underlying buffer ranges must not overlap a region that is already
   registered.
+* The maximum number of concurrently registered symmetric buffers is bounded
+  by the ``ROCSHMEM_MAX_SYMM_REGIONS`` environment variable.
+
+Backend behavior:
+
+* **IPC backend:** peer buffers are mapped into the local address space, so
+  registered buffers are reached directly through inter-process copies.
+* **GDA backend:** registered buffers are always registered with each NIC's
+  protection domain and are reachable over the NIC by every peer. When the IPC
+  fast path is additionally available for node-local peers (see
+  ``ROCSHMEM_DISABLE_MIXED_IPC``), the buffer is also exposed to those peers
+  over IPC, and their accesses use the IPC path in preference to the NIC; the
+  NIC registration remains in place and is used for all other peers.
 
 ROCSHMEM_BUFFER_UNREGISTER_SYMMETRIC
 ------------------------------------
--- a/projects/rocshmem/examples/rocshmem_user_mem_symm_test.cc
+++ b/projects/rocshmem/examples/rocshmem_user_mem_symm_test.cc
@@ -215,8 +215,11 @@ static bool run_symm_variant(const char *label, void *dest_buf,
 
   size_t nbytes = static_cast<size_t>(nelem) * npes * sizeof(int);
 
-  /* Initialize the destination to -1 (0xff bytes) before any remote writes. */
+  /*
+   * Initialize the destination to -1 (0xff bytes) before any remote writes.
+   */
   CHECK_HIP(hipMemset(dest, 0xff, nbytes));
+  CHECK_HIP(hipDeviceSynchronize());
   rocshmem_barrier_all();
 
   user_mem_symm_test<<<dim3(1), dim3(256), 0, 0>>>(source, dest, nelem, team);
--- a/projects/rocshmem/src/backend_bc.hpp
+++ b/projects/rocshmem/src/backend_bc.hpp
@@ -539,28 +539,22 @@ class Backend {
   /**
    * @brief Create the rocSHMEM-managed symmetric address for a registration.
    *
-   * The default implementation maps the user's buffer to a fresh
-   * rocSHMEM-owned virtual address (an "alias") via the heap allocator, so the
-   * returned symmetric address is distinct from and owned independently of the
-   * user's pointer. This is required by the IPC backend, which maps peer
-   * buffers into this address space.
-   *
-   * Backends may override this: the GDA backend reaches registered buffers
-   * over the NIC and uses the user's pointer directly, which avoids a
-   * secondary VMM mapping (HIP cannot reliably unmap such a mapping once its
-   * physical allocation has been dmabuf-imported by the NIC).
+   * Maps the user's buffer to a rocSHMEM-owned virtual address (an "alias")
+   * via the heap allocator, so the returned symmetric address is
+   * distinct from and owned independently of the user's pointer. Both the IPC
+   * and GDA backends use this alias as the symmetric address.
    *
    * @param[in]  addr   User's buffer to register.
    * @param[in]  length Length in bytes.
    * @param[out] alias  Filled with the symmetric address to return to caller.
    */
-  virtual hipError_t create_symm_alias(void *addr, size_t length, void **alias);
+  hipError_t create_symm_alias(void *addr, size_t length, void **alias);
 
   /**
    * @brief Release a symmetric address previously produced by
-   *        create_symm_alias(). Default unmaps the alias mapping.
+   *        create_symm_alias() (unmaps the alias mapping).
    */
-  virtual void destroy_symm_alias(void *alias, size_t length);
+  void destroy_symm_alias(void *alias, size_t length);
 
   /**
    * @brief Dumps derived class statistics. Default is a no-op.
--- a/projects/rocshmem/src/backend_bc.hpp
+++ b/projects/rocshmem/src/backend_bc.hpp
@@ -33,6 +33,8 @@
  * It is the top-level interface for these resources.
  */
 
+#include <hip/hip_runtime_api.h>
+
 #include <cstdint>
 #include <map>
 #include <vector>
@@ -387,6 +389,125 @@ class Backend {
   void symm_allgather(void *inout, size_t bytes_per_pe);
 
   /**
+   * @brief Collective agreement across all PEs on a per-PE value.
+   *
+   * Symmetric registration is collective and exchanges data via
+   * symm_allgather. If a PE returned early on a local error it would drop out
+   * of those collectives and either deadlock the remaining PEs or leave a
+   * region published on some PEs but not others. To keep all PEs on the same
+   * path, each fallible step contributes a per-PE @p local_value, all-gathers
+   * the values here, and proceeds only if predicate @p ok holds for every PE's
+   * contribution. Every PE inspects the same gathered vector, so agreement is
+   * unanimous and all PEs continue or unwind together.
+   *
+   * @tparam T       Trivially-copyable per-PE value type.
+   * @param[in] local_value This PE's contribution.
+   * @param[in] ok          Predicate applied to each PE's gathered value.
+   * @return true only if @p ok holds for every PE's value.
+   */
+  template <typename T, typename Predicate>
+  bool all_pes_agree(T local_value, Predicate ok) {
+    std::vector<T> values(num_pes, T{});
+    values[my_pe] = local_value;
+    symm_allgather(values.data(), sizeof(T));
+    for (int i = 0; i < num_pes; i++) {
+      if (!ok(values[i])) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  /**
+   * @brief Collective agreement that every PE's fallible step succeeded.
+   *
+   * @param[in] local_success Non-zero if this PE's step succeeded.
+   * @return true only if every PE reported success.
+   */
+  bool all_pes_succeeded(int local_success) {
+    return all_pes_agree(local_success, [](int v) { return v != 0; });
+  }
+
+  /**
+   * @brief Collective check that every PE registered the same length.
+   *
+   * @param[in] length This PE's registered length in bytes.
+   * @return true if all PEs registered @p length.
+   */
+  bool symm_lengths_match(size_t length) {
+    return all_pes_agree(length, [length](size_t v) { return v == length; });
+  }
+
+  /**
+   * @brief IPC-specific per-registration state, shared by backends that expose
+   * registered buffers over IPC (the IPC backend for all peers, the GDA
+   * backend for its node-local peers).
+   *
+   * The common base/length bookkeeping lives in Backend::symm_buffer_regions;
+   * this holds the IPC handle/mapping state needed to tear a registration
+   * down. Peer arrays are indexed by the backend's peer index (global PE for
+   * the IPC backend, node-local shm index for GDA), matching how the owning
+   * backend calls ipcPeerPtr().
+   */
+  struct IpcSymmRecord {
+    int slot{-1};                       // index into ipcImpl.symm_table
+    int num_peers{0};                   // size of the peer arrays
+    int self_index{-1};                 // this PE's own peer index
+    char** dev_peer_bases{nullptr};     // device array[num_peers] (published)
+    std::vector<char*> peer_bases{};    // host copy[num_peers] (for CloseIpcHandle)
+    std::vector<char> local_handle{};   // exported IPC handle (for cleanup)
+  };
+
+  /**
+   * @brief Host-side map of IPC symmetric registration state, keyed by the
+   * symmetric address returned to the caller.
+   */
+  std::map<uintptr_t, IpcSymmRecord> ipc_symm_records_{};
+
+  /**
+   * @brief Allocate the device-visible IPC symmetric-registration table
+   * (ipcImpl.symm_table), sized to ROCSHMEM_MAX_SYMM_REGIONS. Shared by the
+   * IPC and GDA backends.
+   */
+  void alloc_ipc_symm_table();
+
+  /**
+   * @brief Free ipcImpl.symm_table (regions array and header).
+   */
+  void free_ipc_symm_table();
+
+  /**
+   * @brief Register a symmetric buffer over IPC and publish it into
+   * ipcImpl.symm_table (collective).
+   *
+   * Exports this PE's IPC handle for @p export_addr, all-gathers handles across
+   * all PEs, opens the handles of the peers listed in @p peer_global, and
+   * publishes the resulting peer-base array (indexed by @p peer_global's order)
+   * so ipcPeerPtr() can translate symmetric addresses to peer mappings. The
+   * calling PE's own slot (@p self_index) is filled with @p symm_addr.
+   *
+   * @param[in] symm_addr   Symmetric address (this PE's local base / self peer)
+   * @param[in] export_addr Buffer whose IPC handle is exported (== symm_addr,
+   *                        except the IPC backend which exports the original
+   *                        buffer behind a distinct alias)
+   * @param[in] length      Length in bytes
+   * @param[in] peer_global Peer index -> global PE whose handle to open
+   * @param[in] self_index  This PE's index within @p peer_global
+   * @return ROCSHMEM_SUCCESS or ROCSHMEM_ERROR (agreed across all PEs)
+   */
+  int register_ipc_symm_region(void* symm_addr, void* export_addr,
+                               size_t length,
+                               const std::vector<int>& peer_global,
+                               int self_index);
+
+  /**
+   * @brief Tear down an IPC symmetric registration published by
+   * register_ipc_symm_region(): close peer mappings, release the exported
+   * handle, compact ipcImpl.symm_table, and drop the record.
+   */
+  int unregister_ipc_symm_region(void* symm_addr);
+
+  /**
    * @brief Required to support static inheritance for device calls.
    *
    * The Context DISPATCH implementation requires this member.
@@ -414,6 +535,32 @@ class Backend {
    *        handled separately.  Default is a no-op.
    */
   virtual void accumulate_default_host_ctx_stats() {}
+
+  /**
+   * @brief Create the rocSHMEM-managed symmetric address for a registration.
+   *
+   * The default implementation maps the user's buffer to a fresh
+   * rocSHMEM-owned virtual address (an "alias") via the heap allocator, so the
+   * returned symmetric address is distinct from and owned independently of the
+   * user's pointer. This is required by the IPC backend, which maps peer
+   * buffers into this address space.
+   *
+   * Backends may override this: the GDA backend reaches registered buffers
+   * over the NIC and uses the user's pointer directly, which avoids a
+   * secondary VMM mapping (HIP cannot reliably unmap such a mapping once its
+   * physical allocation has been dmabuf-imported by the NIC).
+   *
+   * @param[in]  addr   User's buffer to register.
+   * @param[in]  length Length in bytes.
+   * @param[out] alias  Filled with the symmetric address to return to caller.
+   */
+  virtual hipError_t create_symm_alias(void *addr, size_t length, void **alias);
+
+  /**
+   * @brief Release a symmetric address previously produced by
+   *        create_symm_alias(). Default unmaps the alias mapping.
+   */
+  virtual void destroy_symm_alias(void *alias, size_t length);
 
   /**
    * @brief Dumps derived class statistics. Default is a no-op.
--- a/projects/rocshmem/src/gda/backend_gda.cpp
+++ b/projects/rocshmem/src/gda/backend_gda.cpp
@@ -941,11 +941,11 @@ int GDABackend::buffer_register_symmetric([[maybe_unused]] void *addr,
 #endif
 }
 
-void GDABackend::gda_nic_unregister([[maybe_unused]] uintptr_t key) {
+int GDABackend::gda_nic_unregister([[maybe_unused]] uintptr_t key) {
 #if HIP_VERSION >= 70000000
   auto it = gda_symm_records_.find(key);
   if (it == gda_symm_records_.end()) {
-    return;
+    return ROCSHMEM_ERROR;
   }
 
   int slot = it->second.slot;
@@ -1012,6 +1012,9 @@ void GDABackend::gda_nic_unregister([[maybe_unused]] uintptr_t key) {
   }
 
   gda_symm_records_.erase(it);
+  return ROCSHMEM_SUCCESS;
+#else
+  return ROCSHMEM_ERROR;
 #endif
 }
 
@@ -1022,22 +1025,25 @@ int GDABackend::buffer_unregister_symmetric([[maybe_unused]] void *addr) {
   }
 
   uintptr_t key = reinterpret_cast<uintptr_t>(addr);
-  if (gda_symm_records_.find(key) == gda_symm_records_.end()) {
-    return ROCSHMEM_ERROR;
-  }
 
   /*
-   * Tear down the IPC-side registration first (if this buffer was also exposed
-   * over IPC), then the NIC-side state, then the common bookkeeping.
+   * Unregister in the reverse order of registration, unwinding the most
+   * specialized state first:
+   *   1. IPC exposure (only present if this buffer was also mapped for
+   *      node-local peers via the mixed IPC path),
+   *   2. NIC state (device table entry, per-NIC MRs, dmabuf fds, VMM handle),
+   *   3. common host-side bookkeeping (alias unmap + region maps) in the base
+   *      class.
    */
   if (ipcImpl.symm_table != nullptr &&
       ipc_symm_records_.find(key) != ipc_symm_records_.end()) {
     (void)unregister_ipc_symm_region(addr);
   }
 
-  gda_nic_unregister(key);
+  if (gda_nic_unregister(key) != ROCSHMEM_SUCCESS) {
+    return ROCSHMEM_ERROR;
+  }
 
-  /* Common host-side bookkeeping (alias unmap + region maps). */
   return Backend::buffer_unregister_symmetric(addr);
 #else
   return ROCSHMEM_ERROR;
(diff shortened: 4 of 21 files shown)


projects/rocshmem/src/gda/queue_pair.hpp
```diff
@@ -291,9 +302,117 @@ class QueuePair {
...
+    if (table == nullptr) {
+      return nullptr;
+    }
+    int n = table->count;
+    for (int i = 0; i < n; ++i) {
+      GDASymmRegion &r = table->regions[i];
+      if (addr >= r.local_base && addr < r.local_base + r.length) {
+        return &r;
+      }
+    }
```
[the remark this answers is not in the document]
avinashkethineedi: commit: https://github.com/ROCm/rocm-systems/pull/8465/changes/616f87badc1eb0c777808d54a2ffccd184ffc88c

edgargabriel (approved):
LGTM

projects/rocshmem/docs/api/memory_management.rst
```diff
@@ -168,6 +168,21 @@ Notes:
...
 * The underlying buffer ranges must not overlap a region that is already
   registered.
+* The maximum number of concurrently registered symmetric buffers is bounded
+  by the ``ROCSHMEM_MAX_SYMM_REGIONS`` environment variable.
+
+Backend behavior:
+
+* **IPC backend:** peer buffers are mapped into the local address space, so
+  registered buffers are reached directly through inter-process copies.
+* **GDA backend:** registered buffers are always registered with each NIC's
```
abouteiller:
```suggestion
* **GDA backend:** registered buffers are registered with each NIC's
```

projects/rocshmem/src/gda/backend_gda.cpp
```diff
@@ -642,6 +657,393 @@ void GDABackend::buffer_unregister_all() {
...
+  return ibv.reg_mr(pd, iova, length, access);
+}
+#endif
+
+int GDABackend::buffer_register_symmetric([[maybe_unused]] void *addr,
+                                          [[maybe_unused]] size_t length,
+                                          [[maybe_unused]] void **registered_addr) {
+#if HIP_VERSION >= 70000000
+  if (registered_addr == nullptr || symm_entries_ == nullptr) {
+    return ROCSHMEM_ERROR;
```
abouteiller, line 744: should we return a more informative error code?

projects/rocshmem/src/gda/backend_gda.cpp
```diff
@@ -642,6 +657,393 @@ void GDABackend::buffer_unregister_all() {
...
+      gda_nic_unregister(key);
+      Backend::buffer_unregister_symmetric(alias);
+      return ROCSHMEM_ERROR;
+    }
+  }
+
+  *registered_addr = alias;
+  return ROCSHMEM_SUCCESS;
+#else
+  return ROCSHMEM_ERROR;
```
abouteiller, line 929: ROCSHMEM_NOT_IMPLEMENTED?

projects/rocshmem/src/gda/backend_gda.cpp
```diff
@@ -642,6 +657,393 @@ void GDABackend::buffer_unregister_all() {
   Backend::buffer_unregister_all();
 }
 
+void GDABackend::setup_symm_registration() {
+#if HIP_VERSION >= 70000000
```
omor1, line 661: Not necessarily a blocker for this PR since it needs to be cleaned up elsewhere (e.g. throughout the VMM APIs), but it would be better I think to have a CMake configuration feature e.g. `USE_GDA_USER_HEAP` (or something like that), defaulting to `ON`, but disabled for ROCm < 7.0.

projects/rocshmem/src/gda/context_gda_device.hpp
```diff
@@ -395,6 +397,20 @@ class GDAContext : public Context {
...
+   *
+   * Delegates to the shared IpcImpl::ipcPeerPtr resolver, indexed by this PE's
+   * shm rank and the node-local peer index (as returned by isIpcAvailable).
+   * Handles both heap objects and IPC-registered user buffers; returns nullptr
+   * when the address is not reachable over IPC (e.g. an NIC-only registered
+   * buffer or IPC disabled), so callers fall back to the NIC path.
+   */
+  __device__ __forceinline__ char *ipc_peer_ptr(const void *sym_addr,
+                                                 int local_pe) {
+    return ipcImpl_.ipcPeerPtr(sym_addr, ipcImpl_.shm_rank, local_pe);
```
abouteiller, line 453: double check if constmem.shm_rank should be used instead (PR for that change may not have landed yet).

projects/rocshmem/src/gda/backend_gda.hpp
```diff
@@ -86,6 +89,56 @@ class GDABackend : public Backend {
...
+   * @brief GDA-specific per-registration state kept on the host.
+   *
+   * The common alias/length bookkeeping lives in Backend::symm_buffer_regions;
+   * this holds the transport-specific resources needed to tear a registration
+   * down, keyed by the registered alias base address.
+   */
+  struct GdaSymmRecord {
+    int slot{-1};                       // registration slot in each QP slice
+    std::vector<struct ibv_mr*> mrs{};  // per-NIC MRs for the buffer
+    std::vector<int> mr_fds{};          // per-NIC dmabuf fds (kept open for MR lifetime)
```
omor1, line 132: _Do_ the fds need to be kept open for the MR lifetime? Doesn't the kernel internally retain a reference to it (increment the refcount) so that we can close the fd immediately?

projects/rocshmem/src/gda/context_gda_tmpl_device.hpp
```diff
@@ -94,16 +95,17 @@ __device__ void GDAContext::get_nbi(T *dest, const T *source, size_t nelems, int
 template <typename T>
 __device__ void GDAContext::amo_add(void *dst, T value, int pe) {
   if constexpr (sizeof(T) != 8) { LOGD_ERROR_ABORT("gda::amo_add not implemented for non-64bit types"); }//TODO:support for non-uint64t
-  uint64_t L_offset = reinterpret_cast<char *>(dst) - base_heap[constmem.my_pe];
   ActiveWFInfo wf_info(pe);
   int qp_index = get_qp_index(pe, wf_info);
+  uint32_t rkey;
+  uintptr_t raddr = qps[qp_index].translate_remote(dst, &rkey);
```
abouteiller: Can we have translate_remote remain in context_gda instead of going in qp? dealing with base_heap is a rocshmem specialty, we are trying to sever the hard link between qp and rocshmem.

projects/rocshmem/examples/rocshmem_user_mem_symm_test.cc
```diff
@@ -215,8 +215,11 @@ static bool run_symm_variant(const char *label, void *dest_buf,
 
   size_t nbytes = static_cast<size_t>(nelem) * npes * sizeof(int);
 
-  /* Initialize the destination to -1 (0xff bytes) before any remote writes. */
+  /*
+   * Initialize the destination to -1 (0xff bytes) before any remote writes.
+   */
   CHECK_HIP(hipMemset(dest, 0xff, nbytes));
+  CHECK_HIP(hipDeviceSynchronize());
```
omor1, line 222: Is this necessary? Isn't `hipMemset` synchronous anyway?

projects/rocshmem/src/gda/queue_pair.cpp
abouteiller, line 320: this is not using translate_remote, this is going to crash on symm buffers?

projects/rocshmem/src/gda/queue_pair.hpp
```diff
@@ -514,6 +640,22 @@ class QueuePair {
   uint32_t rkey{0};
   uint32_t lkey{0};
 
+  /**
+   * @brief Index of the NIC backing this QP.
+   *
+   * Used at setup to select this QP's registration slice (see QpSymmEntry),
+   * mirroring how the symmetric heap picks heap_rkey[pe * num_nics + nic_idx].
+   */
+  int nic_idx{0};
```
omor1: Unless I've missed something, this is a write-only variable? Can we remove it?

projects/rocshmem/src/gda/queue_pair.hpp
```diff
@@ -514,6 +640,22 @@ class QueuePair {
...
+
+  /**
+   * @brief Remote PE this QP is connected to (conn_num % num_pes).
+   *
+   * Set once at setup. The resolvers use it to index the per-PE arrays (heap
+   * bases and registered-region remote bases/keys).
+   */
+  int dest_pe{0};
```
omor1: Same as for `nic_idx`, I don't think this is read anywhere.

projects/rocshmem/src/gda/queue_pair.hpp
```diff
@@ -161,14 +162,19 @@ class QueuePair {
...
+   * internally (heap or registered buffer), so callers pass symmetric addresses
+   * only. Small transfers are inlined (no source key).
+   *
+   * @param[in] dest    Symmetric destination address (local address space).
+   * @param[in] source  Local source address.
+   * @param[in] nelems  Size in bytes of data transmission.
    * @param[in] wf_info Wavefront information.
    */
-  __device__ void put_nbi(void *dest, const void *source, size_t length,
+  __device__ void put_nbi(void *dest, const void *source, size_t nelems,
```
abouteiller: The sizes are in bytes here not in nelems, the change to lenght was intentional to mark this, we should also do it to the rest of the code later (nelem when we need to later multiply by sizeof(<T>), length if in bytes.

projects/rocshmem/src/gda/queue_pair.hpp
```diff
@@ -291,9 +302,124 @@ class QueuePair {
...
+   * @brief Explicit remote-key atomic variants.
+   *
+   * Used when the destination is a symmetrically-registered user buffer whose
+   * remote key differs from the QP's default heap key. Mirror the default-key
+   * versions above but forward the caller-supplied @p rkey.
+   */
+  __device__ int64_t atomic_fetch(void *dest, uint32_t rkey, int64_t value,
+      int64_t cond, ActiveWFInfo &wf_info);
+
+  __device__ void atomic_nofetch(void *dest, uint32_t rkey, int64_t value,
```
abouteiller, line 298: I think I already added that under a different name

projects/rocshmem/src/gda/queue_pair.hpp
```diff
@@ -291,9 +302,124 @@ class QueuePair {
...
   uintptr_t base_heap = 0;
   size_t base_heap_size = 0;
 
+  /**
+   * @brief Cached remote heap base for this QP's connected peer (@ref dest_pe).
+   *
+   * Equals heap_bases[dest_pe], captured once at setup. Lets the (common) heap
+   * translation run as pure register arithmetic with no memory load.
+   */
+  uintptr_t remote_heap_base{0};
```
abouteiller, line 326: double check if there is a constmem for this (I have preped a PR don't remember if it landed already)

projects/rocshmem/src/gda/queue_pair.hpp
```diff
@@ -291,9 +302,124 @@ class QueuePair {
...
+  }
+
+  /**
+   * @brief Resolve the local key for a locally-sourced symmetric buffer.
+   *
+   * Heap buffers use this QP's default heap key; registered buffers use their
+   * per-NIC local key (from this QP's slice); anything else falls back to the
+   * user-buffer lookup.
+   */
+  __device__ __forceinline__ uint32_t local_lkey(const void *local_addr) {
```
abouteiller: isn't a lkey always local by definition? :)

projects/rocshmem/src/gda/queue_pair.hpp
```diff
@@ -291,9 +302,124 @@ class QueuePair {
...
+      return lkey;
+    }
+    int n = symm_count ? *symm_count : 0;
+    for (int i = 0; i < n; ++i) {
+      QpSymmEntry e = symm_entries[i];
+      if ((addr - e.local_base) < e.length) {
+        return e.lkey;
+      }
+    }
+    return get_lkey(addr);
```
abouteiller: What is different from what get_lkey here?

projects/rocshmem/src/reverse_offload/backend_ro.cpp
```diff
@@ -299,6 +299,20 @@ void ROBackend::ctx_destroy(Context *ctx) {
   delete ro_net_host_ctx;
 }
 
+int ROBackend::buffer_register_symmetric([[maybe_unused]] void *addr,
+                                         [[maybe_unused]] size_t length,
+                                         [[maybe_unused]] void **registered_addr) {
+  LOG_WARN("rocshmem_buffer_register_symmetric is not supported by the RO "
+           "backend");
+  return ROCSHMEM_ERROR;
```
abouteiller, line 307: check if we have a more informative error code to return

projects/rocshmem/src/backend_bc.hpp
```diff
@@ -386,6 +388,125 @@ class Backend {
...
+   * contribution. Every PE inspects the same gathered vector, so agreement is
+   * unanimous and all PEs continue or unwind together.
+   *
+   * @tparam T       Trivially-copyable per-PE value type.
+   * @param[in] local_value This PE's contribution.
+   * @param[in] ok          Predicate applied to each PE's gathered value.
+   * @return true only if @p ok holds for every PE's value.
+   */
+  template <typename T, typename Predicate>
+  bool all_pes_agree(T local_value, Predicate ok) {
```
abouteiller, line 409: there are a lot of calls to this collective during the initialization, at small scale this is fine but for Frontier scale jobs that could be very slow. Ideally we would reduce the number of symm_allgather to 1 per registration.

projects/rocshmem/src/gda/queue_pair.hpp
```diff
@@ -291,9 +302,124 @@ class QueuePair {
...
+  /**
+   * @brief Resolve the local key for a locally-sourced symmetric buffer.
+   *
+   * Heap buffers use this QP's default heap key; registered buffers use their
+   * per-NIC local key (from this QP's slice); anything else falls back to the
+   * user-buffer lookup.
+   */
+  __device__ __forceinline__ uint32_t local_lkey(const void *local_addr) {
+    uintptr_t addr = reinterpret_cast<uintptr_t>(local_addr);
+    if ((addr - base_heap) < base_heap_size) {
```
omor1: Can we use the `is_ptr_in_range` helper?

projects/rocshmem/src/gda/queue_pair.hpp
omor1: Can this entire function be integrated with `get_lkey`, instead of there being two separate functions that do essentially the same thing?

```HIP
__device__ uint32_t QueuePair::get_lkey(uintptr_t laddr) {
  /* Check if in heap */
  if (is_ptr_in_range(base_heap, base_heap_size, laddr)) [[likely]] {
    return lkey;
  }

  /* Get the correct lkey for the user buffer */
  for (size_t i = 0; i < num_user_buffers; i++) {
    if (is_ptr_in_range(buffer_info[i].addr, buffer_info[i].length, laddr)) {
      return buffer_info[i].lkey;
    }
  }

  /* Get the correct lkey for the user symmetric buffer */
  for (size_t i = 0; i < num_user_symmetric_buffers; i++) {
    if (is_ptr_in_range(symm_entries[i].local_base, symm_entries[i].length, laddr)) {
      return symm_entries[i].lkey;
    }
  }

  LOGD_ERROR_ABORT("Valid lkey for address %p not found", reinterpret_cast<void*>(laddr));
  return std::numeric_limits<uint32_t>::max();
}
```

projects/rocshmem/src/gda/queue_pair.hpp
```diff
@@ -291,9 +302,124 @@ class QueuePair {
...
+   * symmetric-heap address, translated from the cached remote heap base using
+   * this QP's default heap key. Otherwise the address is matched against this
+   * QP's registration slice, which supplies the peer base and per-NIC key
+   * directly.
+   *
+   * @param[in]  sym_addr  Symmetric address (valid in the local address space)
+   * @param[out] rkey_out  Filled with the remote key to use
+   * @return Remote address in the connected peer's address space
+   */
+  __device__ __forceinline__ uintptr_t translate_remote(const void *sym_addr,
```
omor1: I'd prefer calling this `get_raddr` and having it return `std::pair<uintptr_t, uint32_t>`, `std::tuple<uintptr_t, uint32_t>`, or e.g. `struct raddr_info { uintptr_t raddr; uint32_t rkey; };`.

`std::pair` and `std::tuple` are entirely `constexpr`, so will work fine in device code. (`constexpr` functions are implicitly `__host__ __device__`)

projects/rocshmem/src/gda/queue_pair.hpp
```diff
@@ -291,9 +302,124 @@ class QueuePair {
...
+   */
+  QpSymmEntry *symm_entries{nullptr};
+
+  /**
+   * @brief Shared registration count (number of live entries per slice).
+   *
+   * Single device int shared by all QPs; register/unregister publish updates
+   * here. One load per non-heap lookup (null when registration unavailable).
+   */
+  const int *symm_count{nullptr};
```
omor1, line 344: Add this to `constmem_t` instead? I don't think we permit registering/unregistering buffers while a kernel is running so this should be better.

projects/rocshmem/src/gda/backend_gda.cpp
```diff
@@ -642,6 +657,393 @@ void GDABackend::buffer_unregister_all() {
...
+}
+
+int GDABackend::buffer_unregister_symmetric([[maybe_unused]] void *addr) {
+#if HIP_VERSION >= 70000000
+  if (addr == nullptr || symm_entries_ == nullptr) {
+    return ROCSHMEM_ERROR;
+  }
+
+  uintptr_t key = reinterpret_cast<uintptr_t>(addr);
+  if (gda_symm_records_.find(key) == gda_symm_records_.end()) {
```
omor1: We end up doing this lookup twice, once here and again in `gda_nic_unregister`. Can we just check the return value from `gda_nic_unregister` after unregistering the IPC symmetric region?

projects/rocshmem/src/gda/backend_gda.cpp
```diff
@@ -642,6 +657,393 @@ void GDABackend::buffer_unregister_all() {
...
+   * entry table itself is freed later in cleanup_gpu_qps().
+   */
+  std::vector<uintptr_t> addrs;
+  addrs.reserve(gda_symm_records_.size());
+  for (auto &kv : gda_symm_records_) {
+    addrs.push_back(kv.first);
+  }
+  for (auto a : addrs) {
+    buffer_unregister_symmetric(reinterpret_cast<void *>(a));
+  }
```
omor1: I don't especially like this code and would _prefer_ to have something like the below, but it's not a blocker and could be done as a follow-up.
```C++
void GDABackend::symmetric_buffer_unregister_all() {
  for (auto&& [addrint, symm_record] : gda_symm_records_) {
    // unregister symm_record while not removing it from gda_symm_records_
  }
  gda_symm_records_.clear();
}
```

projects/rocshmem/src/gda/context_gda_tmpl_device.hpp
```diff
@@ -950,9 +959,23 @@ __device__ void GDAContext::alltoall_linear_thread_puts(rocshmem_team_t team,
...
+     * default keys exactly.
+     */
+    size_t nbytes = nelems * sizeof(T);
+    uint32_t dst_rkey;
+    uintptr_t dst_raddr = qps[dest_pe].translate_remote(
+        &dst[my_pe_in_team * nelems], &dst_rkey);
+    uint32_t src_lkey =
+        (static_cast<int32_t>(nbytes) <=
+         static_cast<int32_t>(qps[dest_pe].inline_threshold))
+            ? 0 : qps[dest_pe].local_lkey(&src[j * nelems]);
```
omor1: This particular check is already done inside `put_nbi_single`, isn't it? If it's not, then it _should_ be.

omor1 (commented):
Has this been validated on ROCm 7.0.x? I know that there are current build issues due to some prior VMM-related changes and I want to ensure that this PR doesn't compound those issues.

If this doesn't cause additional failures on ROCm 7.0.x then it's alright, but I'd like a followup to address some of the comments.

projects/rocshmem/src/gda/queue_pair.hpp
abouteiller: I think same but can be done later

projects/rocshmem/docs/api/memory_management.rst
avinashkethineedi: commit: https://github.com/ROCm/rocm-systems/pull/8465/changes/250a032b95bc5110ac1d7329315d9af920322571

projects/rocshmem/src/gda/queue_pair.hpp
avinashkethineedi: Removed in 250a032 — it was only used host-side to pick the slice/rkey, dead on the QP.

projects/rocshmem/src/gda/queue_pair.hpp
avinashkethineedi: Removed in 250a032

projects/rocshmem/src/gda/queue_pair.hpp
avinashkethineedi: Done in 250a032 — the merged get_lkey uses is_ptr_in_range for the symm-entry range check.

projects/rocshmem/src/gda/queue_pair.hpp
avinashkethineedi: commit: 250a032

projects/rocshmem/src/gda/queue_pair.hpp
avinashkethineedi: merged into a single get_lkey in 250a032.

projects/rocshmem/src/gda/queue_pair.hpp
avinashkethineedi: Renamed the byte-valued param to length on put_nbi/get_nbi in 250a032.

projects/rocshmem/src/gda/backend_gda.cpp
avinashkethineedi: Done in 250a032 — gda_nic_unregister now returns ROCSHMEM_SUCCESS/ERROR and buffer_unregister_symmetric propagates it, dropping the up-front find.

projects/rocshmem/src/gda/queue_pair.hpp
avinashkethineedi: Removed the redundant explicit-rkey atomic_add/atomic_add_single in 6951d205 — the used atomic_nofetch family already covers it.

projects/rocshmem/src/gda/queue_pair.hpp
avinashkethineedi: Done in 6951d205 — folded local_lkey into a single get_lkey (heap → registered user buffers → symm entries, via is_ptr_in_range).

projects/rocshmem/src/gda/queue_pair.hpp
avinashkethineedi: Done — the resolver now returns a raddr_info {raddr, rkey} struct instead of an out-param (e3549b83), renamed to get_raddr_info (2b316631); all callers use structured bindings.

projects/rocshmem/src/gda/queue_pair.hpp
```diff
@@ -291,9 +285,112 @@ class QueuePair {
...
+   *
+   * @param[in] sym_addr Symmetric address (valid in the local address space)
+   * @return Remote address and remote key. Returns {0, 0} for a non-symmetric
+   *         pointer (undefined in SHMEM) so the NIC rejects the transfer,
+   *         rather than fabricating a heap offset that could silently corrupt
+   *         peer memory.
+   */
+  __device__ __forceinline__ raddr_info get_raddr_info(const void *sym_addr) {
+    uintptr_t addr = reinterpret_cast<uintptr_t>(sym_addr);
+    if ((addr - base_heap) < base_heap_size) {
```
omor1: Can we use `is_ptr_in_range` here too? It would also be nice for `get_lkey` and `get_raddr_info` to be implemented next to each-other in the same file, but that's not a blocker.

projects/rocshmem/src/gda/backend_gda.cpp
avinashkethineedi: will be followed up in a separate PR

projects/rocshmem/src/gda/queue_pair.hpp
avinashkethineedi: commit: https://github.com/ROCm/rocm-systems/pull/8465/changes/2b316631aff9a4a2e0c02c6c68a2835bd6a78679

projects/rocshmem/src/gda/backend_gda.cpp
avinashkethineedi: commit: https://github.com/ROCm/rocm-systems/pull/8465/changes/c8339b2b76fc836fe297cde5ffdcedcbbbb4e885

projects/rocshmem/src/backend_bc.hpp
avinashkethineedi: will address it in a follow-up PR

projects/rocshmem/src/gda/queue_pair.hpp
avinashkethineedi: No constmem for this yet

projects/rocshmem/src/gda/context_gda_tmpl_device.hpp
avinashkethineedi: follow-up PR

projects/rocshmem/src/gda/context_gda_device.hpp
avinashkethineedi: No constmem.shm_rank yet

projects/rocshmem/src/gda/backend_gda.cpp
```diff
@@ -987,34 +976,77 @@ int GDABackend::gda_nic_unregister([[maybe_unused]] uintptr_t key) {
...
+  /*
+   * Now tear down each registration's transport resources and host bookkeeping,
+   * mirroring buffer_unregister_symmetric (IPC exposure, NIC resources, common
+   * base-class bookkeeping).
+   */
+  for (auto &[key, rec] : gda_symm_records_) {
+    void *addr = reinterpret_cast<void *>(key);
+    if (ipcImpl.symm_table != nullptr &&
+        ipc_symm_records_.find(key) != ipc_symm_records_.end()) {
+      (void)unregister_ipc_symm_region(addr);
```
omor1, line 1043: Why the cast to void? If `unregister_ipc_symm_region` has a return value, we should check it and at _least_ warn on error.

Do as a follow up to get the PR through now.

projects/rocshmem/src/gda/backend_gda.cpp
```diff
@@ -987,34 +976,77 @@ int GDABackend::gda_nic_unregister([[maybe_unused]] uintptr_t key) {
...
+    }
+    release_symm_record_nic_resources(rec);
+    (void)Backend::buffer_unregister_symmetric(addr);
```
omor1, line 1046: Ditto.

projects/rocshmem/examples/rocshmem_user_mem_symm_test.cc
omor1: Errors occur without it; leave for now, try to figure out the root cause later.

projects/rocshmem/src/gda/backend_gda.cpp
```diff
@@ -987,34 +976,77 @@ int GDABackend::gda_nic_unregister([[maybe_unused]] uintptr_t key) {
...
+    (void)hipMemRelease(rec.gen_handle);
+  }
+#endif
+}
+
+void GDABackend::symmetric_buffer_unregister_all() {
+#if HIP_VERSION >= 70000000
+  if (gda_symm_records_.empty()) {
+    return;
+  }
```
omor1, line 1021: I think this check is extraneous; the remaining parts of the code work perfectly fine even if `gda_symm_records_` is empty. Just a comment, no change needed.

projects/rocshmem/src/gda/backend_gda.cpp
```diff
@@ -987,34 +976,77 @@ int GDABackend::gda_nic_unregister([[maybe_unused]] uintptr_t key) {
...
-    if (n < it->second.mr_fds.size() && it->second.mr_fds[n] != -1) {
-      close(it->second.mr_fds[n]);
+    if (n < rec.mr_fds.size() && rec.mr_fds[n] != -1) {
+      close(rec.mr_fds[n]);
     }
   }
 
+  /* Release the backing VMM handle retained at registration. */
+  if (rec.has_gen_handle) {
+    (void)hipMemRelease(rec.gen_handle);
```
omor1, line 1012: See other error-checking comments. To be done as a follow up.

omor1 (approved):
lgtm, I have some comments about checking return values but those can (and should) be done as a follow up so the PR doesn't get delayed further.

PR #8469, opened 2026-07-13
fix(rocprofiler-system): bypass pthread wrap at thread limit

anujshuk-amd:
## Motivation

When profiling thread-heavy apps (PyTorch CIFAR-100), rocprof-sys would hang after the number of threads exceeded the compile-time limit ROCPROFSYS_MAX_THREADS. It was noticed on aac6.
**What was the issue :** When profiling thread-heavy applications such as PyTorch training, rocprofiler-systems would hang once the number of threads exceeded the compile-time limit ROCPROFSYS_MAX_THREADS. Thread profiling data is stored in fixed-size arrays indexed by an internal thread ID, but the old code kept wrapping pthread_create, growing storage, and calling stable_vector. That led to, heavy lock and instrumentation overhead as thousands of threads were still launched, and the application appearing frozen because rocprof-sys never stopped profiling and fell back to normal thread creation. The fix adds an early bypass at the thread cap, safe handling for wrapping pthread_create, growing storage, and calling stable_vector.

**Additional info : Checked by testing ROCPROFSYS_SAMPLING_DELAY at 1, 2, 3, 4, 5, and 10 seconds on the PyTorch CIFAR-100 workload (ROCm 7.13 and Mainline, rocprof-sys v1.8.0). Increasing the sampling delay did not resolve the hang/deadlock.**

## Technical Details

- When the app tries to create a new thread and the profiler has already reached its maximum allowed thread count, stop intercepting that thread creation and let the operating system handle it normally.
- Show a single warning that the thread limit was reached, and do not try to record profiling data for any threads created after that point.
- Before using a thread’s slot in the profiler’s internal storage table, check that the slot number is within the built-in limit; if it is not, skip profiling for that thread instead of accessing invalid memory.
- Stop expanding the profiler’s thread storage tables once the maximum size is reached, so memory and locks are not kept growing without bound.
- Track how many threads have actually been initialized for profiling, and use that count as an additional signal to stop wrapping new threads.
- Raise the default built-in thread limit at compile time from a very small (128) value to 2048 on typical machines.
- Added automated tests that verify coverage for half/at/double/8 times(load test)scenarios.

## JIRA ID

JIRA ID : ROCM-25389

## Test Plan

Ran impacted tests. Added new coverage.

## Test Result

All impacted tests are passing locally.

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-systems/CMakeLists.txt
+++ b/projects/rocprofiler-systems/CMakeLists.txt
@@ -312,8 +312,8 @@ endif()
 include(ProcessorCount)
 ProcessorCount(ROCPROFSYS_PROCESSOR_COUNT)
 
-if(ROCPROFSYS_PROCESSOR_COUNT LESS 8)
-    set(ROCPROFSYS_THREAD_COUNT 128)
+if(ROCPROFSYS_PROCESSOR_COUNT LESS 128)
+    set(ROCPROFSYS_THREAD_COUNT 2048)
 else()
     math(EXPR ROCPROFSYS_THREAD_COUNT "16 * ${ROCPROFSYS_PROCESSOR_COUNT}")
     compute_pow2_ceil(ROCPROFSYS_THREAD_COUNT "16 * ${ROCPROFSYS_PROCESSOR_COUNT}")
@@ -335,7 +335,7 @@ set(ROCPROFSYS_MAX_THREADS
 )
 rocprofiler_systems_add_feature(
     ROCPROFSYS_MAX_THREADS
-    "Maximum number of total threads supported in the host application (default: max of 128 or 16 * nproc)"
+    "Maximum number of total threads supported in the host application (default: max of 2048 or 16 * nproc)"
 )
 
 compute_pow2_ceil(_MAX_THREADS "${ROCPROFSYS_MAX_THREADS}")
--- a/projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/pthread_create_gotcha.cpp
+++ b/projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/pthread_create_gotcha.cpp
@@ -130,6 +130,21 @@ using native_handle_set_t = std::set<pthread_create_gotcha::native_handle_t>;
 auto native_handles          = native_handle_set_t{};
 auto internal_native_handles = native_handle_set_t{};
 auto native_handles_mutex    = locking::atomic_mutex{};
+
+// Shared across the parent interceptor and the child wrapper so the thread-limit
+// warning is emitted exactly once per process, regardless of which gate trips first.
+std::once_flag thread_limit_warning_flag;
+
+inline void
+warn_thread_limit_once()
+{
+    std::call_once(thread_limit_warning_flag, []() {
+        LOG_WARNING(
+            "Maximum allowed thread limit ({}) reached. Further thread creation and "
+            "profiling will be disabled to prevent resource exhaustion.",
+            static_cast<size_t>(ROCPROFSYS_MAX_THREADS));
+    });
+}
 }  // namespace
 
 //--------------------------------------------------------------------------------------//
@@ -145,14 +160,20 @@ void*
 pthread_create_gotcha::wrapper::operator()() const
 {
     using thread_bundle_data_t = thread_data<thread_bundle_t>;
-
+    auto _child_internal_tid   = utility::get_thread_index();
+    if(static_cast<size_t>(_child_internal_tid) >= ROCPROFSYS_MAX_THREADS ||
+       thread_info::get_initialized_thread() >= ROCPROFSYS_MAX_THREADS)
+    {
+        warn_thread_limit_once();
+        if(m_config.promise) m_config.promise->set_value();
+        return m_routine(m_arg);
+    }
     if(is_shutdown && *is_shutdown)
     {
         if(m_config.promise) m_config.promise->set_value();
         // execute the original function
         return m_routine(m_arg);
     }
-
     push_thread_state(ThreadState::Internal);
 
     std::int64_t _tid         = -1;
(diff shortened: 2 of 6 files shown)


projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/pthread_create_gotcha.cpp
```diff
@@ -130,6 +130,21 @@ using native_handle_set_t = std::set<pthread_create_gotcha::native_handle_t>;
...
+// Shared across the parent interceptor and the child wrapper so the thread-limit
+// warning is emitted exactly once per process, regardless of which gate trips first.
+std::once_flag thread_limit_warning_flag;
+
+inline void
+warn_thread_limit_once()
+{
+    std::call_once(thread_limit_warning_flag, []() {
+        LOG_WARNING(
+            "Maximum allowed thread limit ({}) reached. Further thread creation and "
```
sputhala-amd: Is the thread creation disabled? I think only profiling of further threads are disabled?

projects/rocprofiler-systems/source/lib/rocprof-sys/library/thread_info.cpp
```diff
@@ -105,6 +124,13 @@ grow_data(std::int64_t _tid)
...
     {};
 
+    // Do not grow when _tid is greater than or equal to max_supported_threads
+    if(_tid >= static_cast<std::int64_t>(max_supported_threads))
+    {
+        // if thread limit exceeded, return current peak number of threads
+        return peak_num_threads;
+    }
+
     if(_tid >= peak_num_threads)
```
sputhala-amd, line 136: After this new gating `grow_data` becomes a no-op, there is no dynamic growth logic that will trigger, since `max_supported_threads` and `peak_num_threads` are same.

projects/rocprofiler-systems/CMakeLists.txt
```diff
@@ -312,8 +312,8 @@ endif()
 include(ProcessorCount)
 ProcessorCount(ROCPROFSYS_PROCESSOR_COUNT)
 
-if(ROCPROFSYS_PROCESSOR_COUNT LESS 8)
-    set(ROCPROFSYS_THREAD_COUNT 128)
+if(ROCPROFSYS_PROCESSOR_COUNT LESS 128)
+    set(ROCPROFSYS_THREAD_COUNT 2048)
 else()
```
sputhala-amd, line 317: We should reconsider this logic for a future PR to drop the `nproc` scaling, since it's misleading for packaged builds. A big build host can inflate the constant for all package consumers.

sputhala-amd (commented):
`ROCPROFSYS_MAX_THREADS` is purely a self-imposed limit on the size of rocprof-sys's own `thread_data` slot arrays, exceeding it only means "**no profiling slot for this thread**." It's a compile-time constant baked into the data-structure sizing.

Ideally we are saying rocprof-sys will profile upto 2048 threads, in this case its not simultaneous live threads but a cumulative cap. Hence increasing this number to 2048 seems reasonable. However in future it would be good to re-enable dynamic growth and possibly introduce recycle logic for thread slots.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/thread_info.cpp
anujshuk-amd: True, there is no further dynamic growth.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/pthread_create_gotcha.cpp
anujshuk-amd: Thread creation and profiling both are disabled, once thread limit is reached.

projects/rocprofiler-systems/tests/pytest/test_thread_limit.py
```diff
@@ -92,15 +136,36 @@ def test(self, mode, thread_count, thread_limit_env):
...
-            fail_value = thread_limit + 1
+        pass_value = get_expected_pass_value(thread_count, thread_limit)
+        fail_value = get_expected_fail_value(thread_count, thread_limit)
 
         self.assert_regex(
             result,
             mode,
             pass_regex=[f"\\|{pass_value}>>>"],
             fail_regex=[f"\\|{fail_value}>>>"],
         )
```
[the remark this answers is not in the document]
anujshuk-amd: Covered warning in diffrent test.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/components/pthread_create_gotcha.cpp
sputhala-amd: I dont think we limit thread creation. Applications can create as many threads as it want, we just stop profiling additional threads as there is no storage for profiling data for those additional threads.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/thread_info.cpp
sputhala-amd: Please mention this in a comment somewhere. That this function is now a no-op.

anujshuk-amd: Added.

sputhala-amd (approved):

projects/rocprofiler-systems/CMakeLists.txt
```diff
@@ -312,8 +312,8 @@ endif()
 include(ProcessorCount)
 ProcessorCount(ROCPROFSYS_PROCESSOR_COUNT)
 
-if(ROCPROFSYS_PROCESSOR_COUNT LESS 8)
-    set(ROCPROFSYS_THREAD_COUNT 128)
+if(ROCPROFSYS_PROCESSOR_COUNT LESS 128)
```
kcossett-amd, line 315: Could you increase this to 256 instead of 128?

The machines that run our tests on TheRock have 192 cores.

kcossett-amd (approved):

tools/systems_pr_bot/policy.yml
```diff
@@ -152,10 +152,13 @@ diff:
     - '**/*.asc'
 
 checks:
-  # External checks that must be GREEN before PR is "ready for review"
```
sputhala-amd, line 155: This change can be removed from the scope of this PR and created as a separate PR to be reviewed by the rocm maintainer?

projects/rocprofiler-systems/CMakeLists.txt
anujshuk-amd: if 192 will come it will automatically calculate to new  thread numbers which will be more than 2048. Since took limit 2048 thats why limit adjusted to the 128.

tools/systems_pr_bot/policy.yml
anujshuk-amd: Removed.

projects/rocprofiler-systems/CMakeLists.txt
kcossett-amd: The reason why is I'd like to re-enable the `thread-limit` tests on TheRock.
I had to disable them at the time as the computed `ROCPROFSYS_MAX_THREADS` on the CI machine would be higher than the version generated on the build machine, making one of the tests always fail, IIRC it was `thread-limit-above`. 

If we have it set to `256`, I'd reckon the test could be re-enabled on TheRock CI.

anujshuk-amd: As discussed, After fix of this PR and https://github.com/ROCm/rocm-systems/pull/8602 the rock CI should be passed. Thefore marking resolved.

PR #8474, opened 2026-07-13
Users/lloginov amd/tim-rem-fs

lloginov-amd:
## Submission Checklist

### DIFF
--- a/projects/rocprofiler-systems/source/lib/common/path.hpp
+++ b/projects/rocprofiler-systems/source/lib/common/path.hpp
@@ -7,8 +7,11 @@
 #include "common/delimit.hpp"
 #include "common/env_vars.hpp"
 #include "common/environment.hpp"
+
 #include <spdlog/fmt/fmt.h>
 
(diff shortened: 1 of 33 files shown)


projects/rocprofiler-systems/source/lib/common/path.hpp
```diff
@@ -253,6 +212,15 @@ is_link(const std::string& _path)
     return false;
 }
 
+bool
+is_directory(const std::string& _path)
```
mradosav-amd, line 216: https://en.cppreference.com/cpp/filesystem/is_directory

projects/rocprofiler-systems/source/lib/common/path.hpp
```diff
@@ -281,6 +249,140 @@ readlink(const std::string& _path)
     return _path;
 }
 
+std::string
+get_cwd()
```
mradosav-amd, line 253: https://en.cppreference.com/cpp/filesystem/current_path

PR #8480, opened 2026-07-13
fix(rocprofiler-compute): improve HBM roofline benchmark accuracy

alexioslyrakis-amd:
JIRA ID: SILOTIGER-778
JIRA ID: AIPROFCOMP-728

## Summary

- Replace the copy-based HBM bandwidth kernel (`dst[gid] = src[gid]` with `double`) with a read-only kernel using non-temporal loads (`__builtin_nontemporal_load`), `__uint128_t` (16-byte) accesses, and an unroll factor of 16 — matching TransferBench's approach
- Use a fixed 4 GB dataset with `128 * CUs` workgroups to ensure full HBM saturation regardless of CU count
- No external dependencies needed — the built-in kernel now achieves equivalent results to TransferBench

## Results

Tested on real hardware:

| GPU | CUs | Old kernel | New kernel | TransferBench | New/TB |
|-----|-----|-----------|-----------|---------------|--------|
| MI308XHF (gfx942) | 64 | ~157 GB/s | 4073 GB/s | 4027 GB/s | 101% |
| MI350X (gfx950) | 256 | ~5600 GB/s | 6914 GB/s | 7200 GB/s | 96% |

## Test plan

- [x] Validated on MI308XHF (64 CU gfx942) — matches TransferBench within ~1%
- [x] Validated on MI350X (256 CU gfx950) — achieves ~96% of TransferBench
- [x] Minimal diff: only `benchmark_base.py` (launch logic) and `gfx9_base.py` (kernel source) changed


vedithal-amd (changes requested):
This PR should be rebased on rocprofiler-compute-develop branch and the PR target should be updated to the same branch.

rocprofiler-compute-develop branch is the active development branch for the team for shipping features. And features and merged to develop every 3 weeks (next merge on July 24)

projects/rocprofiler-compute/src/argparser.py
```diff
@@ -550,6 +550,22 @@ def omniarg_parser(
             " (Default: 0)"
         ),
     )
+    roofline_group.add_argument(
```
vedithal-amd: I dont think this option is necessary - folks should be able to just use the built in kernels for empirical benchmark

projects/rocprofiler-compute/src/roofline/benchmark/benchmark_base.py
```diff
@@ -471,6 +492,59 @@ def hbm_bw_benchmark(self, device: int) -> PerfMetrics:
 
         return perf_metrics
 
+    # HBM bandwidth benchmark via TransferBench
+    def hbm_bw_transferbench(self, device: int) -> PerfMetrics:
```
vedithal-amd: I dont think adding third party dependency such as TransferBench for empirical benchmarks would be acceptable. We should improve our benchmarking kernel rather than depending on third party is better.

projects/rocprofiler-compute/src/roofline/benchmark/gfx9/benchmark_gfx9_base.py
```diff
@@ -34,7 +36,7 @@ def __init__(self, device_id: int, cache_sizes: dict) -> None:
         }
 
         self.tests = {
-            "HBM": super().hbm_bw_benchmark,
+            "HBM": self._select_hbm_benchmark(),
             "MALL": super().mall_bw_bench,
             "L2": super().l2_bw_bench,
```
vedithal-amd, line 39: It should delegate to super().hbm_benchmark which would use hbm_source to enable transfer bench - we dont need _select_hbm_benchmark() here

projects/rocprofiler-compute/src/roofline/benchmark/gfx11/benchmark_gfx11_base.py
```diff
@@ -43,7 +45,7 @@ def __init__(self, device_id: int, cache_sizes: dict) -> None:
         self.matrix_ops = {}
 
         self.tests = {
-            "HBM": super().hbm_bw_benchmark,
+            "HBM": self._select_hbm_benchmark(),
```
vedithal-amd: It should delegate to super().hbm_benchmark which would use hbm_source to enable transfer bench - we dont need _select_hbm_benchmark() here

projects/rocprofiler-compute/src/roofline/benchmark/benchmark_base.py
vedithal-amd: If we really do want to support TransferBench it has to be a pure python submodule which has to be vendored per https://github.com/ROCm/rocm-systems/blob/develop/projects/rocprofiler-compute/CONTRIBUTING.md#vendoring-external-dependencies

projects/rocprofiler-compute/src/roofline/benchmark/benchmark_base.py
```diff
@@ -422,21 +429,35 @@ def hbm_bw_benchmark(self, device: int) -> PerfMetrics:
...
+        prog = self.Program(self.hbm_bw_src)
+        func = prog.get_kernel("HBM_bw")
 
         workgroup_size = DEFAULT_WORKGROUP_SIZE
         workgroups_per_cu = 20 * 1024
         workgroups = cus * workgroups_per_cu
+        num_iters = 10
+
+        # Each thread processes one float4 (16 bytes) per iteration
+        float4_size = sizeof(c_float) * 4
```
vedithal-amd: I tried this benchmark on a MI300X with large number of CUs (304 as opposed to 64 CU in MI308X) and the benchmark ran for 12 straight minutes and then faults due to memory issue. I think this is due to the drastic increase in the amount of work in HBM benchmark kernel.

Note that even pre-PR the HBM benchmark was giving correct estimate of 4166 GB/s on MI 300X and in this PR it just crashes which is not tolerable

I think the problem is that this kernel is not able to saturate the HBM on MI308X due to its lower CU count - but there should be a better fix which does not break MI300X. This is because we couple the dataset_entries based on workgroups which depend on - that is `workgroups = cus * 20*1024`

The minimal solution here just needs to decouple dataset entries with cu count using a fixed dataset size. No need to strided iterations - no need for Transfer Bench - no need to any --hbm-source-data option

```
dataset_bytes = 4 * 1024**3 # 4 GB
dataset_entries = dataset_bytes // sizeof(c_double)
workgroups = dataset_entries // workgroup_size
```

This will also make the HBM Bandwidth benchmark kernel much faster

projects/rocprofiler-compute/tests/test_roofline_hbm_benchmark.py
```diff
@@ -0,0 +1,250 @@
+# Copyright (c) Advanced Micro Devices, Inc.
```
vedithal-amd: I dont think we have a MI308 machine in our CI need to check with Dev Ops - but either way asserting hbm bandwidth values in a CI environement with multiple profiling tests running in background is bound to give wrong values.

We can update the existing `test_metric_validation.py` which is already excluded from CI to add capability to assert values from roofline.csv against theoretical calculated values across MI 100, MI 200, MI 300A, MI 300X, MI 308 (need three expected values for MI 300 series). This way someone can manually run our ctest suite on a desired machine to validate hbm benchmark kernel

I dont think we need to re-invent the wheel by adding this `test_roofline_hbm_benchmark.py` just update `test_metric_validation.py` and add a new dict for mi 200, 300, 250 series with multiple expected values for mi 300 series

vedithal-amd (changes requested):
Created JIRA AIPROFCOMP-728 for this

projects/rocprofiler-compute/src/argparser.py
alexioslyrakis-amd: Agreed, the only reason I've added it was because built-in kernel was much slower than TransferBench. But I refactored it to improve its performance so now it matches the bw we achieve we TransferBench. Since we're talking about just a single kernel, I believe that's the simplest option indeed.

projects/rocprofiler-compute/src/roofline/benchmark/benchmark_base.py
alexioslyrakis-amd: Let's go for the simplest approach which is to improve the benchmarking for now. See https://github.com/ROCm/rocm-systems/pull/8480#discussion_r3605299388

projects/rocprofiler-compute/src/roofline/benchmark/gfx9/benchmark_gfx9_base.py
alexioslyrakis-amd: Removed, kept original

projects/rocprofiler-compute/src/roofline/benchmark/gfx11/benchmark_gfx11_base.py
alexioslyrakis-amd: Removed, kept original

projects/rocprofiler-compute/src/roofline/benchmark/benchmark_base.py
alexioslyrakis-amd: Agreed on the fixed dataset size. The new version uses a fixed 4 GB dataset with `128 * CUs` workgroups. However, simply using the old double copy kernel with a fixed size (as suggested) still only achieves ~46% of peak on MI308XHF and ~79% on MI325 because the copy pattern (read+write) forces read-modify-write of cache lines on the write side. The key improvement is the kernel itself: switching to a read-only kernel with non-temporal loads and 16-byte accesses. Results on real hardware now almost match TransferBench (97% match). Runtime is fast (fixed 4 GB dataset), no crashes.

projects/rocprofiler-compute/tests/test_roofline_hbm_benchmark.py
alexioslyrakis-amd: The old `test_roofline_hbm_benchmark.py` is removed. The rebased PR includes `test_benchmark_base.py` with deterministic checks only  (kernel source properties and dataset sizing math). No GPU required for that.\

I added this test just to satisfy CI gate.

projects/rocprofiler-compute/src/roofline/benchmark/benchmark_base.py
```diff
@@ -419,21 +419,33 @@ def hbm_bw_benchmark(self, device: int) -> PerfMetrics:
...
-        prog = self.Program(self.hbm_bw_src, ["HBM_bw<double>"])
-        func = prog.get_kernel("HBM_bw<double>")
+        prog = self.Program(self.hbm_bw_src)
+        func = prog.get_kernel("HBM_bw")
 
         workgroup_size = DEFAULT_WORKGROUP_SIZE
-        workgroups_per_cu = 20 * 1024
-        workgroups = cus * workgroups_per_cu
-        dataset_entries = workgroups * workgroup_size
+        UNROLL = 16
```
vedithal-amd: this should be lower case local variable

alexioslyrakis-amd: Right, changed this to lower case local var

vedithal-amd:
Pushed a small follow-up (f361c4dbf4) on the read-only peak point.

The empirical HBM peak now comes from a read-only kernel, but the plotted `HBM Bandwidth` value still counts read+write traffic, so a read-heavy kernel can appear to beat its own peak. Read-only is the correct ceiling since HBM's bus is half-duplex (single-direction is fastest). I added a clause to the HBM Bandwidth description across the gfx9 arches noting the peak is read-only and write-heavy kernels fall below it, then synced panel -> per-arch -> docs and refreshed the hash DB.

vedithal-amd:
Pushed 1d0bce3641: removed test_benchmark_base.py and added an hbm_bandwidth case to test_metric_validation.py. It profiles memcopy with roofline and asserts the empirical HBM peak (metric 4.1.9, Peak (Empirical)) against a measured MI300X value (4916, 10% tolerance). Passed locally on MI300X. It's meant to be run manually via ctest, not in CI, since the CI environment isn't stable enough for bandwidth validation. MI300-only for now; other SKUs (MI300A/MI308) will likely need their own expected values.

vedithal-amd (approved):
LGTM. Review comments addressed and the read-only HBM benchmark verified on MI300X.

cc @cfallows-amd please review as well before this can be merged.

projects/rocprofiler-compute/src/roofline/benchmark/benchmark_base.py
vedithal-amd: resolving

vedithal-amd:
I validated the HBM bandwidth kernel on MI300X and MI308X and updated the values in test_metric_validation.py - Note that this is skipped in CI but would be part of ctest suite when run automatically. Reason being that values from profiling output in heavy overloaded CI machines is unstable leading to flaky tests.

vedithal-amd:
@alexioslyrakis-amd we have obtained feedback from HPC team that a read-only benchmark should be provided as a separate roofline peak. I will have to put this in limbo until @cfallows-amd implements that work.

If you would still want to get something merged for MI308X then we would have to reduce the scope of the PR to not touch the kernel behavior in-terms of read/write and only allow the number of workgroups change to better saturate low CU count GPUs like MI308X. OR let's wait for @cfallows-amd to implement the separate roofline peak for this improved kernel.

vedithal-amd (changes requested):
Putting in limbo until we get agreement with HPC team @gsitaram

vedithal-amd (approved):
From further discussion in `Profilers Discussion` teams channel it seems like HPC team is fine with read-only kernel being true upper bound for hbm bandwidth

PR #8481, opened 2026-07-13
fix(rocprofiler-compute): Resolve intermittent tarball issues

jbonnell-amd:
## Motivation

We have been noticing that, occasionally, the version of TheRock installed on the gfx1151 runners is outdated by weeks. This issue does not appear to impact gfx94X and gfx950 runs.

## Technical Details

- Updated `.github/workflows/rocprofiler-compute-continuous-integration.yml` Docker options
  - Added `--pull always` option to ensure that we always pull from remote instead of referencing from local cache, which may contain an older version of the daily image which contains a previous version of TheRock

## JIRA ID

JIRA ID : AIPROFCOMP-717

## Test Plan

- Ensure that TheRock tarball version is consistent amongst every runner, including gfx1151
- No regression issues from change

## Test Result

- Consistent tarball versions across all instances
  - https://github.com/ROCm/rocm-systems/actions/runs/29266522150

## Submission Checklist

### DIFF
--- a/.github/workflows/rocprofiler-compute-continuous-integration.yml
+++ b/.github/workflows/rocprofiler-compute-continuous-integration.yml
@@ -103,6 +103,7 @@ jobs:
         --device /dev/dri
         --cap-add CAP_SYS_ADMIN
         --env-file /etc/podinfo/gha-gpu-isolation-settings
+        --pull always
 
     steps:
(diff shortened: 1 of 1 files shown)


dgaliffiAMD (approved):
Please confirm that the run logs show that the issue is resolved.

PR #8487, opened 2026-07-13
[rocjitsu] Preserve vector memory transactions

kuhar:
## Motivation

Keep dwordxN accesses as one per-lane transaction through the functional memory hierarchy and make sleep instructions yield the functional quantum so communication kernels cannot monopolize a CU.

Preserve VMID translation priority at page boundaries, keep generated sleep behavior sourced from amdisa, and align transaction metrics and regression coverage with the coalesced path.

This is to fix timeouts on RCCL tests.

## Technical Details

RCCL tests using the default protocol were unusually slow and intermittently timed out under RocJITsu. Setting `NCCL_PROTO` avoided the problematic path, but that also reduced coverage of the behavior users receive by default.

The slowdown came from two simulator behaviors:
1. Vector memory operations were decomposed first into individual elements and then into individual bytes, greatly amplifying backing-memory traffic.
2. `s_sleep` was treated as a no-op, allowing polling and wait loops to consume an entire functional execution quantum instead of yielding to work on peer compute units.

This change fixes those simulator behaviors directly, so the default RCCL protocol can remain covered.

## JIRA ID

N/A

## Test Plan

Run unit tests.

## Test Result

RCCL tests take <9s each now and don't time out.

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
@@ -12,11 +12,13 @@
 #include "simdojo/sim/component.h"
 #include "util/log.h"
 
+#include <algorithm>
 #include <cstring>
 #include <format>
 #include <memory>
 #include <mutex>
 #include <shared_mutex>
+#include <span>
 #include <string>
 #include <sys/uio.h>
 #include <unordered_map>
@@ -41,11 +43,9 @@ public:
       auto &hdr = msg->header();
       auto *data = reinterpret_cast<uint8_t *>(msg->payload());
       if (hdr.op == simdojo::MessageOp::READ) {
-        for (uint32_t i = 0; i < hdr.size_bytes; ++i)
-          data[i] = read8(hdr.addr + i, hdr.vmid);
+        read_block(hdr.addr, std::span<uint8_t>(data, hdr.size_bytes), hdr.vmid);
       } else if (hdr.op == simdojo::MessageOp::WRITE) {
-        for (uint32_t i = 0; i < hdr.size_bytes; ++i)
-          write8(hdr.addr + i, data[i], hdr.vmid);
+        write_block(hdr.addr, std::span<const uint8_t>(data, hdr.size_bytes), hdr.vmid);
       }
       hdr.op = simdojo::MessageOp::RESPONSE;
     });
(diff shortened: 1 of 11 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l2_cache.cpp
```diff
@@ -16,18 +16,20 @@ namespace amdgpu {
 
 void L2Cache::send_backing(uint64_t addr, uint8_t *data, uint32_t size, simdojo::MessageOp op,
                            uint32_t vmid) {
+  if (op == simdojo::MessageOp::WRITE)
+    ++backing_write_transactions_;
```
atgutier: Does this need to be atomic?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l2_cache.cpp
```diff
@@ -16,18 +16,20 @@ namespace amdgpu {
...
+  else
+    ++backing_read_transactions_;
```
atgutier: Same: does this need to be atomic?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
```diff
@@ -107,6 +106,54 @@ class GpuMemory : public simdojo::SparseMemory {
 
   uint32_t fetch32(uint64_t addr, uint32_t vmid = 0) const { return read32(addr, vmid); }
 
+  /// @brief Read a contiguous range from simulated GPU memory.
+  /// @details Handles each page through mapped host memory, client memory, or
+  /// sparse backing memory.
+  void read_block(uint64_t addr, void *dst, size_t len, uint32_t vmid = 0) const {
```
atgutier: Could you use std::span instead of raw void* here?

Also, this and the write variant are nearly identical. Can they be unified a bit to reduce the code dupe? Maybe factor out the copy loop into a helper?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l2_cache.h
```diff
@@ -202,6 +205,8 @@ class L2Cache : public simdojo::Component {
   std::array<std::mutex, ATOMIC_STRIPE_COUNT> atomic_stripes_;
   std::vector<simdojo::Port *> cpl_ports_;
   uint64_t write_count_ = 0; ///< Debug: total L2 writes (for trace).
+  uint64_t backing_read_transactions_ = 0;
```
atgutier: I think these may need to be atomic.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l1_vector_cache.cpp
```diff
@@ -154,11 +154,8 @@ void L1VectorCache::load(const uint64_t *addrs, uint64_t lane_mask, uint32_t ele
     uint32_t lane = std::countr_zero(remaining);
     remaining &= remaining - 1;
     uint64_t base = addrs[lane];
-    for (uint32_t e = 0; e < num_elems; ++e) {
-      uint64_t ea = base + e * elem_size;
-      read_bytes(ea, dst + lane * stride + e * elem_size, elem_size, mtype, non_temporal,
-                 request_l1_bypass, vmid);
-    }
+    // Keep all elements from one vector-memory lane in the same transaction.
+    read_bytes(base, dst + lane * stride, stride, mtype, non_temporal, request_l1_bypass, vmid);
```
Groverkss: I wonder if we should implement coalescing also.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l1_vector_cache.cpp
```diff
@@ -15,6 +15,34 @@
...
+uint32_t for_each_coalesced_lane_run(const uint64_t *addrs, uint64_t lane_mask, uint32_t stride,
+                                     F &&fn) {
+  uint32_t run_count = 0;
+  uint64_t remaining = lane_mask;
+  while (remaining) {
+    const uint32_t first_lane = std::countr_zero(remaining);
+    remaining &= ~(uint64_t{1} << first_lane);
+
+    uint32_t last_lane = first_lane;
+    while (last_lane < 63) {
```
Groverkss: 63 looks like a hardcoded constant. Is this intended?

Groverkss (approved):
lgtm

atgutier (approved):

PR #8489, opened 2026-07-13
fix (ROCM-27881): disable net-ib level fault injection by default

VadimKutovoi:
## Motivation

This PR disables net-ib fault-injection test layer by default.

## JIRA ID

ROCM-27881
AIRDEL-40


mch:
What's the actual motivation for disabling this by default? Is it causing a problem of some kind?

marbre (changes requested):
This isn't marked as P1, thus is does not qualify for a cherry-picking.

wenkaidu (approved):

alexander-sannikov (approved):

ddurnov (approved):

astyrrian1 (approved):

marbre (approved):
Passed for gfx94x, build for gfx950 still running but looks fine.

PR #8492, opened 2026-07-13
fix(rccl) [JIRA ID ROCM-25479]  LL128: generate separate kernels for UBR

wenkaidu:
## Motivation

LL128: generate separate registered / non-registered user-buffer kernels

## Technical Details

    Registered user buffers must be accessed with system-scope cache-bypassing
    loads/stores for cross-GPU coherence, but that path forces the compiler into
    heavy AGPR usage and spilling. Non-registered buffers only need the cheaper
    plain/non-temporal path. Selecting between them at runtime kept both code paths
    (and the AGPR-spilling one) live in every LL128 kernel, hurting occupancy.

    Make user-buffer registration a compile-time dimension for LL128 so each launch
    runs a single, clean code path:

    - device.h / generate.py: add a UserRegMode kernel dimension (1=registered,
      2=non-registered) and emit two LL128 variants per collective (ring + tree for
      AllReduce, ring for AllGather/Broadcast/Reduce/ReduceScatter). Kernel key
      packing, device/host tables and the total function count are updated
      accordingly.
    - enqueue.cc: pick the registered vs non-registered LL128 function id at launch
      time from work->regUsed / netRegUsed.
    - primitives.h / prims_ll128.h: thread UserRegMode through Primitives; the
      registered specialization keeps the system-scope cache-bypass path, the
      non-registered specialization compiles only the plain/non-temporal path (no
      dead cache-bypass code, no AGPR spill).
    - common.h / add_unroll.sh: forward UserRegMode through RunWorkBatch/RunWorkColl
      and the runRing/runTreeSplit dispatch helpers.

    Result on gfx950: the common non-registered LL128 kernels drop from AGPR=13-32
    to AGPR=1 and shed ~30 VGPR (e.g. AllReduce ring f32 197->165, AllGather i8
    245->181), improving occupancy, while the registered specialization retains the
    required system-scope path in isolation.

    LL (ProtoLL) is intentionally left at the f2af4e8 baseline: it does not benefit
    from the split and the non-temporal FIFO / runtime-split experiments regressed
    it, so its user-buffer and FIFO accesses are unchanged.

## JIRA ID

ROCM-25479

## Test Plan

    Adds LL128RegVariantKernelTests validating that reg-variant LL128
    collectives (AllReduce, AllGather, Broadcast) generate two distinct
    device functions (registered vs non-registered user buffer) with
    unique function IDs, that ncclDevFuncIsLL128RegVariant() agrees with
    the reg-variant collective set, and that non-reg-variant collectives
    emit a single kernel (reg=0).

    The suite is added to the Debug-only fixtures target because the
    ncclDevFuncNameToId lookup table is a local symbol in release builds
    and only gains default visibility under Debug.

## Test Result

Build: ./install.sh --debug-fast -l -t with ONLY_FUNCS="AllReduce|AllGather|Broadcast|ReduceScatter" completed successfully. Both test targets linked (rccl-UnitTestsFixtures and rccl-UnitTestsFixturesDebug), and LL128RegVariantKernelTests.cpp.o compiled with no errors.

Direct confirmation of the split during codegen — the two reg variants were emitted as separate device objects (suffix _1 = non-registered, _2 = registered):

specialized_broadcast_ring_ll128_sum_i8_0_0_1_1.o
specialized_broadcast_ring_ll128_sum_i8_0_0_1_2.o
specialized_allgather_ring_ll128_sum_i8_0_0_1_1.o
specialized_allgather_ring_ll128_sum_i8_0_0_1_2.o
Test run (LL128RegVariantKernelTests.*):

[  PASSED  ] 3 tests.
  - TwoDistinctLL128KernelsPerRegVariantCollective   OK
  - ClassificationMatchesRegVariantSet               OK
  - NonRegVariantEmitsSingleKernel                   OK
So the new test verifies, against the actual generated host dispatch table, that:

Each reg-variant LL128 collective (AllReduce, AllGather, Broadcast) resolves to two distinct function IDs (registered vs non-registered user buffer).
ncclDevFuncIsLL128RegVariant() agrees with the reg-variant collective set.
Non-reg-variant collectives still produce a single kernel (reg=0).
The only warning in the log (llvm-readobj --offloading at 98%) is an unrelated offload-bundle inspection step, not a compile/link error for the test.

Note on placement: the suite lives in the Debug-only rccl-UnitTestsFixturesDebug target because ncclDevFuncNameToId is a local symbol (b) in release builds and only gets default visibility (linkable) under Debug — so this test can't be run from a release build by design.

## Submission Checklist

### DIFF
--- a/projects/rccl/cmake/scripts/add_unroll.sh
+++ b/projects/rccl/cmake/scripts/add_unroll.sh
@@ -37,6 +37,34 @@ if [[ "$HIP_FILE" =~ .*/src/device/.*\.h ]]; then
   perl -pi -e 's/(runRing<T, RedOp, Proto, USE_ACC, COLL_UNROLL.*?)>/\1, Pipeline>/' "$HIP_FILE"
   perl -pi -e 's/(runTreeSplit<T, RedOp, Proto, USE_ACC, COLL_UNROLL.*?)>/\1, Pipeline>/' "$HIP_FILE"
   perl -pi -e 's/(runTreeUpDown<T, RedOp, Proto, USE_ACC, COLL_UNROLL.*?)>/\1, Pipeline>/' "$HIP_FILE"
-  sed -i "s/\\(struct RunWorkBatch<ncclFunc[^>]*\\)>*/\\1, USE_ACC, COLL_UNROLL, Pipeline>/" "$HIP_FILE"
-  sed -i "s/\\(RunWorkColl<[^,]*,[^,]*,[^,]*,[^,]*,[^>]*\\)>/\\1, USE_ACC, COLL_UNROLL, Pipeline>/" "$HIP_FILE"
+  sed -i "s/\\(struct RunWorkBatch<ncclFunc[^>]*\\)>*/\\1, USE_ACC, COLL_UNROLL, Pipeline, UserRegMode>/" "$HIP_FILE"
+  sed -i "s/\\(RunWorkColl<[^,]*,[^,]*,[^,]*,[^,]*,[^>]*\\)>/\\1, USE_ACC, COLL_UNROLL, Pipeline, UserRegMode>/" "$HIP_FILE"
+
+  # Declare the UserRegMode template parameter on RunWorkColl / RunWorkBatch
+  # partial specialization headers (those immediately followed by a
+  # `struct RunWork...`). Partial specializations may not carry default template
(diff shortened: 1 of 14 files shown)


wenkaidu:
replaced with https://github.com/ROCm/rocm-systems/pull/9347

PR #8493, opened 2026-07-13
fix(clr): multi-GPU fine-grain SVM peer maps VA.

jaydeeppatel1111:
## Motivation
The multi-GPU `MemoryTest2.exe` failures that surfaced after
[`c7a59f7` (#6308)](https://github.com/ROCm/rocm-systems/commit/c7a59f77c1) share this root cause.
That commit added a file-scope `__managed__` global to `hipPointerGetAttribute.cc`, which is
allocated at module load (`ihipMallocManaged ptr=0x304000000`) and pins offset 0 of the shared
fine-grain SVM chunk. That desyncs the per-device free-lists, so later `hipHostAlloc`/managed
allocations pick divergent per-GPU offsets and collide (`Memobj map already has an entry ...`),
failing tests like `Unit_hipHostAlloc_DataValidation`. #6308 is a valid test exposing this
pre-existing bug — this PR fixes it (no revert needed).

On a multi-GPU system (observed on 2-GPU Windows/PAL), fine-grain SVM allocations return one canonical pointer that must resolve to the same virtual address on every device. Two allocation types hit this path:
hipHostAlloc with the coherent/fine-grain flag combos — specifically hipHostMallocPortable, hipHostMallocWriteCombined, and Portable | WriteCombined (these map to CL_MEM_ALLOC_HOST_PTR | CL_MEM_SVM_FINE_GRAIN_BUFFER), and __managed__ variables / hipMallocManaged, which use the same flags.
The situation: Each device suballocates fine-grain SVM from its own buddy allocator inside a shared 64 MB reserved-VA chunk. The canonical VA is chunkBase + intra-chunk offset.

The persistent __managed__ buffer on dev0 lands inside that SVM chunk at offset 0. On dev1 it instead took the pinned/dedicated-VA path (not SVM), so dev1's SVM chunk was later lazily created empty by the first hipHostAlloc.
Result: the buddy free-lists are asymmetric across devices → each device picks a different intra-chunk offset → the canonical VA diverges. dev1's mapping for the shared pointer either doesn't exist (GPU fault) or aliases another live allocation in the global MemObjMap (silent corruption).

Concretely: a kernel on dev1 dereferencing a __managed__ symbol faulted (HIP error 719), and multi-GPU hipHostAlloc buffers collided.

The fix: for multi-GPU fine-grain SVM buffers, stop taking the pinned-host path on peer devices and fall through to the reserved-VA SVM path, so every device reserves + maps the one canonical VA the leader (dev0) chose — exactly like a regular fine-grain hipHostAlloc peer. This removes the free-list asymmetry (the __managed__ buffer now sits in the SVM chunk on every device), which fixes both the dev1 __managed__ fault and the downstream hipHostAlloc offset divergence.

Five candidate solutions evaluated (this PR is C):
A — force offset: add a PAL AllocateAtOffset so peers mirror the leader's exact intra-chunk offset (true root cause, keeps packing; requires a vendored-PAL/upstream change).
B — dedicated chunk: give each multi-GPU fine-grain SVM alloc its own reserved-VA chunk at offset 0 (robust, rocclr-only; loses suballocation packing).
C — peer reserve+map (this PR): route multi-GPU fine-grain SVM peers off the pinned path onto the reserved-VA SVM path so all devices map the same canonical VA and free-lists stay symmetric.
D — allocation symmetry: keep per-device allocation sequences lock-step so buddy allocators converge on the same offset, guarded by a drift assert (rocclr-only; fragile).
E — shared backing / allocate-once: model one canonical VA mapped by every device (mechanically coincides with B at the rocclr level).

## Technical Details
In pal::Device::createBuffer, detect a multi-GPU fine-grain SVM peer allocation: CL_MEM_ALLOC_HOST_PTR && CL_MEM_SVM_FINE_GRAIN_BUFFER && owner.getSvmPtr() != nullptr && context.devices().size() > 1.
For such buffers, skip the Resource::Pinned branch (add !mgpuFineGrainSvm to its guard). Control then falls through to the general create() path with params.svmBase_ = owner.svmBase(), dispatching to Resource::CreateSvm.
In CreateSvm, the peer (svmPtr != 0) already takes the reserved-VA branch (useReservedGpuVa = true, pReservedGpuVaOwner = svmBase_->iMem()), so it maps its own per-device IGpuMemory at dev0's canonical VA and installs that translation into dev1's page tables.
Scope is intentionally limited to devices().size() > 1; single-GPU keeps the existing pinned path (no peer exists, so the invariant is trivially satisfied and behavior is unchanged).

## JIRA ID

ROCM-27508

## Test Plan
Failing reported tests.
	935 - Unit_hipMemcpy_Positive_Basic (Failed)
        967 - Unit_hipMemcpyDtoH_Positive_Basic (Failed)
        970 - Unit_hipMemcpyHtoD_Positive_Basic (Failed)
        973 - Unit_hipMemcpyDtoD_Positive_Basic (Failed)
        976 - Unit_hipMemcpyAsync_Positive_Basic (Failed)
        992 - Unit_hipMemcpyDtoHAsync_Positive_Basic (Failed)
        994 - Unit_hipMemcpyHtoDAsync_Positive_Basic (Failed)
        996 - Unit_hipMemcpyDtoDAsync_Positive_Basic (Failed)
        1275 - Unit_hipHostAlloc_DataValidation (Failed)
        2790 - Unit_HRR_VMMRoundtrip (Failed)

## Test Result

Failing reported tests pass.

## Submission Checklist

### DIFF
--- a/projects/clr/rocclr/device/pal/paldevice.cpp
+++ b/projects/clr/rocclr/device/pal/paldevice.cpp
@@ -1582,8 +1582,22 @@ pal::Memory* Device::createBuffer(amd::Memory& owner, bool directAccess) const {
         }
         remoteAlloc = true;
       }
+      // Multi-GPU fine-grain SVM (e.g. __managed__ / hipMallocManaged, which use
+      // CL_MEM_ALLOC_HOST_PTR | CL_MEM_SVM_FINE_GRAIN_BUFFER) must be addressable
+      // at the SAME canonical VA on every device. The pinned path maps the shared
+      // host pages at a device-local Default-range VA that differs from the owner's
+      // SVM VA, so a peer kernel using the canonical pointer faults or reads the
+      // wrong memory (observed as a dev1 GPU fault accessing a __managed__ var).
+      // For these, skip pinning on the peer and fall through to the reserved-VA SVM
+      // path so the peer reserves the canonical VA (same as a regular fine-grain
+      // hipHostAlloc peer). This also keeps the per-device suballocator free-lists
+      // symmetric, which is required for consistent intra-chunk offsets.
+      const bool mgpuFineGrainSvm =
+          (owner.getMemFlags() & CL_MEM_ALLOC_HOST_PTR) &&
+          (owner.getMemFlags() & CL_MEM_SVM_FINE_GRAIN_BUFFER) &&
+          (owner.getSvmPtr() != nullptr) && (owner.getContext().devices().size() > 1);
       // Make sure owner has a valid hostmem pointer and it's not COPY
-      if (!remoteAlloc && (owner.getHostMem() != nullptr)) {
+      if (!remoteAlloc && !mgpuFineGrainSvm && (owner.getHostMem() != nullptr)) {
         Resource::PinnedParams params;
         params.owner_ = &owner;
         params.gpu_ = reinterpret_cast<VirtualGPU*>(owner.getVirtualDevice());
--- a/projects/hip-tests/catch/config/configs/unit/memory.yaml
+++ b/projects/hip-tests/catch/config/configs/unit/memory.yaml
@@ -506,6 +506,9 @@ memory:
   Unit_hipManagedMultiGpuVaConsistency_MultiDevice:
     <<: *level_2
     tags: [alloc, multigpu]
+  Unit_hipManagedMultiGpuVaConsistency_HostAllocNoDivergence:
+    <<: *level_2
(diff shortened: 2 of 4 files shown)


jaydeeppatel1111:
Hello @gandryey, @saleelk , Can you pls review this PR? thanks!

saleelk (approved):

gandryey (approved):

PR #8499, opened 2026-07-13
[rocjitsu] Support RDNA WGP mode and fix wave32 EXEC semantics

kuhar:
## Motivation

This is to allow running gfx1250->rdna4 DBT over the simulated rdna4 target.

1. Make valid RDNA WGP-mode kernels run correctly in the simulator instead of being rejected for exceeding one CU’s LDS capacity or being placed without enough sibling-CU resources.
2. Prevent silent corruption in wave32 kernels that use EXEC_HI as scalar storage, including across checkpoint and restore.
3. Ensure these correctness properties are consistent across the RDNA targets that support the same execution modes.

## Technical Details

- Reserve compute and shared-LDS capacity for the complete WGP or cluster before dispatch. Roll back incomplete placement and release reservations with workgroup lifetime.
- Route LDS operations through the selected placement so WGP-mode kernels use the sibling pair’s shared backing.
- Preserve the complete 64-bit EXEC register pair while exposing a separate lane-masked view for vector execution.
- Allow WGP mode with rdna1, rdna2, rdna3, rdna3_5, and rdna4. Keep gfx1250 in CU mode because WGP_MODE is unavailable.

## JIRA ID

N/A

## Test Plan

Run unit tests.

## Test Result

All good.

## Submission Checklist


newling (commented):
Is this logically 2 changes
1) WGP mode support
2) wave32 EXEC fix? 

or are they connected in a way that makes sense for them to be delivered together?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/isa_traits.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include "rocjitsu/isa/arch/amdgpu/gfx1250/isa.h"
+#include "rocjitsu/isa/arch/amdgpu/rdna1/isa.h"
+#include "rocjitsu/isa/arch/amdgpu/rdna2/isa.h"
+#include "rocjitsu/isa/arch/amdgpu/rdna3/isa.h"
+#include "rocjitsu/isa/arch/amdgpu/rdna3_5/isa.h"
+#include "rocjitsu/isa/arch/amdgpu/rdna4/isa.h"
+
+namespace rocjitsu {
+
+bool arch_supports_wgp_mode(rj_code_arch_t arch) {
```
Groverkss: I don't think we should use a .cpp file for these. Using includes will inline this which is better.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -8353,9 +8130,9 @@ def gen_operand(self) -> None:
                 else '  if (ev == 124)\n' '    return wf.m0();\n'
             )
             + '  if (ev == 126)\n'
-            '    return static_cast<uint32_t>(wf.exec());\n'
+            '    return static_cast<uint32_t>(wf.exec_raw());\n'
```
Groverkss: any reason for this rename?

Groverkss: This is just causing a lot of churn

kuhar:
> Is this logically 2 changes
> 
> 1. WGP mode support
> 2. wave32 EXEC fix?
> 
> or are they connected in a way that makes sense for them to be delivered together?

Yes, these could be split up into two separate PRs but I wanted to land them together as I discovered it as a gap in rdan4 support that needs to be closed to enable using the simulator as the gfx1250->rdna4 dbt target.

Groverkss (approved):
lgtm

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/spi.h
```diff
@@ -157,11 +190,106 @@ class ShaderProcessorInput {
...
+    if (wgps_.empty())
+      return std::nullopt;
+    const uint32_t aligned = util::align_up(entry.group_segment_fixed_size, 256u);
+    for (size_t attempt = 0; attempt < wgps_.size(); ++attempt) {
+      size_t wgp_index = (next_wgp_ + attempt) % wgps_.size();
+      auto &wgp = *wgps_[wgp_index];
+      wgp.cu0->retire_halted_wfs();
+      wgp.cu1->retire_halted_wfs();
+
+      // A WGP allocation cannot overlap CU-mode residents or cluster-pinned
```
newling, line 212: I gather from this PR that there are no architectures with both workgroup clusters and WGP? So is this check just defensive? 

I'm assuming that 'wgp.cu0->lds_allocation_pinned()' is for gfx1250 only, currently.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/isa_properties.h
```diff
@@ -0,0 +1,47 @@
...
+  case ROCJITSU_CODE_ARCH_RDNA3:
+    return {true};
+  case ROCJITSU_CODE_ARCH_RDNA3_5:
+    return {true};
+  case ROCJITSU_CODE_ARCH_RDNA4:
+    return {true};
+  case ROCJITSU_CODE_ARCH_GFX1250:
+    return {false};
+  default:
+    return {};
```
newling, line 41: Some sort of assertion/exception that you don't fall through to default here? I don't think a default set of isa properties makes sense

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
```diff
@@ -737,6 +742,7 @@ uint32_t CommandProcessor::dispatch_workgroups(DispatchEntry &entry) {
     uint32_t global_wg_id = local_wg_id + entry.workgroup_id_offset;
 
     if (entry.has_workgroup_clusters()) {
+      assert(!entry.wgp_mode && "workgroup clusters are gfx1250-only and use CU mode");
```
newling, line 745: Just to confirm what this is asserting: there are no archs with cluster workgroups and WGP (contiguous LDS's can be used by a single workgroup)?

newling (approved):
LGTM. I haven't checked all the tests, but the code makes sense to me. 

Only minor comments/questions, none blocking.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/spi.h
kuhar: yes, there's an assertion for this somewhere in the code

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
kuhar: yes

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/isa_properties.h
kuhar: this will be a compiler warning for unhandled switch case

PR #8501, opened 2026-07-13
fix(clr): ihipLaunchKernel catch all errors, except InvalidSymbol

jiabaxie:
## Motivation

When KFD Doorbell exhausts, we do not return an error and rather assume the symbol pointer to be correct. We should tell the user that the symbol they queried for is invalid because there is a doorbell exhaustion.

## Technical Details

ihipLaunchKernel currently catches 3 known errors. If any are these 3, it will propagate the error up and stop. If not, then it assumes the pointer to be valid even if it cannot validate it. This feature I believe is for backwards compatibility. However, it makes more sense to by default catch all errors except hipInvalidSymbol, as other hipErrors are also valid to propagate. hipInvalidSymbol is for when you cannot find the symbol in the dynamic library, which is for the backwards compatibility case. All other errors, including the one described in the ticket hipErrorNoBinaryForGpu for kfd doorbell exhaustion, should indicate real invalid symbols. 

## JIRA ID
JIRA ID 27216

ROCM-27216

## Test Plan

Testing was done on gfx942 machine by attempting to ring doorbell from 300 threads
## Test Result

 Before patch: segmentation fault on some threads. After: no segmentation fault.

## Submission Checklist


chrispaquot (approved):

litvaOo:
The only issue is with the rocgdb-cpu job, overriding

PR #8502, opened 2026-07-13
docs(rocprofiler-compute): HLD for analysis config YAML redesign

xuchen-amd:
## Motivation

- Adds `docs/design/hld-analysis-config-redesign.md` — high-level design proposal for redesigning the analysis config YAML system into a three-layer architecture (Counter Registry, Metric Library, Display/View)
- Adds `docs/design/hld-metric-analysis-2026-07-13.md` — static analysis of the existing 122 YAML files supporting the schema design (field inventory, consistency matrix, architecture coverage)

## Technical Details

N/A

## JIRA ID

AIPROFCOMP-665

## Test Plan

N/A

## Test Result

N/A

## Submission Checklist


projects/rocprofiler-compute/docs/design/analysis-config-redesign/hld-analysis-config-redesign.md
```diff
@@ -0,0 +1,546 @@
...
+### [PS1] No clear requirements constrain what analysis configs should provide
+
+Without defined requirements, new additions increase maintenance cost, and manual modifications introduce further divergence. The design cannot be evaluated against criteria that don't exist.
+
+The two known users — tool developers and tool users — have different needs:
+
+- **Developers** need easy maintenance for metric modifications and architecture additions, and a design that integrates well with the source code.
+- **Users** need to profile metrics relevant to their workload, view results in a structured way, and understand what the tool is showing them.
+
+### [PS2] Per-architecture duplication makes metric maintenance costly and error-prone
```
vedithal-amd, line 120: Note that we would still like to have flexibility of per gfx metric definitions in some cases - reason being that some counters have slightly different functionality in different gfx arch (e.g. TCC_BUBBLE for MI 300 vs MI 350) - moreover there are architectural nuances which will require separate metric definitions

Agreed that in most cases a single metric definition can be shared among gfx archs if they support it

projects/rocprofiler-compute/docs/design/analysis-config-redesign/hld-analysis-config-redesign.md
```diff
@@ -0,0 +1,546 @@
...
+
+#### [FR2] — A metric and its full specification should be defined exactly once
+- Any other appearance of that metric should be a reference to the single definition, not a copy.
+- **[FR2.1]** The attributes that make up a complete metric definition must be explicitly declared. Based on static analysis of the current 122 YAML files (see [`metric-analysis-2026-07-13.md`](metric-analysis-2026-07-13.md)), the required and optional fields are:
+    - **Required:** `id`, `name`, `unit`, aggregation type (`avg/min/max` or `value`), `archs` or `implementations`
+    - **Optional:** `description`, `peak`, `pct_of_peak`, `coll_level`
+    - **Typed variants** (not general metric fields): `xfer`, `coherency`, `expr`, `type`, `transaction` — these appear in specific table types only and require a separate schema treatment
+- **[FR2.2]** Metrics must be uniquely identifiable by a stable, explicitly declared id, not by position in a file or directory. A positional reference (e.g., the 3rd metric in file `1100_*.yaml`) breaks whenever metrics are reordered or files are renamed. An explicit `id` field survives both.
```
vedithal-amd, line 187: This is one of the most important contributions of this design.
Unique metric ids should be specified where metric formula is defined.

Right now, we derive metric id based on the position/ordering in the table, which hurts us a lot.

This will improve things like baseline comparison and reduce the maintenance cost of change the hardcoded is in sets yaml and other places when we add/delete a metric

For baseline comparison, we can the implement comparison across unique metric ids. So, if a metric id is not supported in one the gfx - the baseline comparison will drop it when doing comparison

Note that we should maintain the invariant that same metric id semantically means the same metric across all gfx arch. It should never be the case that metric id X.Y.Z means VALU FLOPs in one gfx arch and Matrix FLOPS in the other - this might not be the best example but you get the point. In essence same metric id should be comparable across gfx arch

projects/rocprofiler-compute/docs/design/analysis-config-redesign/hld-analysis-config-redesign.md
```diff
@@ -0,0 +1,546 @@
...
+- **[FR2.3]** A metric's architecture availability must be declared within its single definition to avoid redundancy.
+- **[FR2.4]** Metrics with table-type-specific fields (e.g., fabric stall breakdown metrics with `xfer`, `coherency`, `expr` fields) must be handled as typed variants. The general metric definition schema does not need to accommodate these fields.
+- **[Measured by]** The number of full metric definitions for any given metric is exactly 1. With YAML inheritance, the family base file is the single definition; per-arch override files are not redefinitions — they modify specific fields while inheriting the rest.
+- **[Assumption]** Description alone is not a reliable equality criterion — ambiguous cases require manual classification. See [OQ3](#oq3--39-metrics-with-both-description-and-formula-drift-require-manual-classification) and [OQ4](#oq4--43-same-description-different-name-cases-require-canonical-name-selection).
```
vedithal-amd, line 193: Currently we have duplicate metric names across different tables in the same yaml.
But we have only one metric description blob at the end of the yaml.
This makes it impossible to ascertain which metric the description targets.

I would metric descriptions should be linked with metric ids to avoid these problems in the future

projects/rocprofiler-compute/docs/design/analysis-config-redesign/hld-analysis-config-redesign.md
```diff
@@ -0,0 +1,546 @@
...
+### Non-functional requirements
+
+#### [NFR1] — The redesigned config structure should be adoptable incrementally
+It should be possible to migrate one concern at a time without requiring a coordinated change across all concerns simultaneously.
+
+
+## Design
+
+> **Assumptions**
+> - This redesign prioritizes the developer experience. The problems identified in [[PS2]](#ps2-per-architecture-duplication-makes-metric-maintenance-costly-and-error-prone) (duplication) and [[PS3]](#ps3-analysis-config-files-mix-unrelated-concerns-making-changes-harder-and-riskier-than-they-need-to-be) (mixed concerns) are specific to developers. There is no sufficient evidence that users regularly interact with the analysis config files. The analysis configs therefore do not need to be readable or self-contained for users. Design trade-offs in this document are resolved in favor of developers.
```
vedithal-amd, line 210: I agree that users done interact with analysis yaml as frequently as we would think. I dont even know how useful is the --config option for the end users but I am not saying we should drop it.

At any cost the analysis yaml file should be readable (that is why it is yaml - if we dont want it to be readable then JSON is better). Its ok to separate out the concerns across files but it should still be readable.

projects/rocprofiler-compute/docs/design/analysis-config-redesign/hld-analysis-config-redesign.md
```diff
@@ -0,0 +1,546 @@
...
+      metric:
+        MFMA FLOPs (F8):
+          unit: GFLOPs
+```
+
+This 7-line file replaces a 300+ line duplicate. The deep merge replaces only the `unit` field within `MFMA FLOPs (F8)`, preserving all other content from gfx940.
+
+### How the layers connect
+
+The diagram below shows how `mem.l2_hit_rate` flows from counters through the metric library into two family-specific views, with the formula resolved at analysis time based on the profiled architecture.
```
vedithal-amd, line 318: I think we should also add a layer for placeholder variables - think about METRIC_PER_XCD metrics which are referenced as $METRIC_PER_XCD in many metric formulas. Where METRIC_PER_XCD = METRIC / $NUM_XCD

Placeholder variables are currently mixed in the code which is not good - as part of this design can we also split them into a separate layer so that they are driven by configuration.

projects/rocprofiler-compute/docs/design/analysis-config-redesign/hld-analysis-config-redesign.md
```diff
@@ -0,0 +1,546 @@
...
+render: mem_bw_chart
+metrics:
+  - mem.hbm_rd_bw
+  - mem.hbm_wr_bw
+  - ...
+```
+
+#### What changes when something needs to be updated
+
+Each change touches only the layer that owns the concern — satisfying the [[FR3]](#fr3) measurement criterion.
```
vedithal-amd, line 474: I would say metric id should be of the form X.Y.Z where X, Y, X are integers encoding group number, subgroup number, metric related number. They should not be strings like group_name.subgroup_name.metric_short_name

The reason being that we need to support -b block filter to filter out metric by their ids during profiling and analysis

projects/rocprofiler-compute/docs/design/analysis-config-redesign/hld-analysis-config-redesign.md
```diff
@@ -0,0 +1,546 @@
...
+
+
+## Open questions
+
+*Ordered by impact on implementation — higher priority items block earlier stages.*
+
+### OQ1 — Layer 3 view definition requires customer input
+*(blocks Stages 5 and 7)*
+
+What views exist, what metrics each view contains, and how views are organized are decisions that reflect how tool users think about GPU performance analysis. These cannot be determined by developers alone. The current analysis config structure (roofline, memory chart, IP block panels) is a starting point, but must be validated against actual user workflows before the Layer 3 schema is finalized.
```
vedithal-amd, line 533: Agreed, the current grouping - subgrouping - views are the way they are currently due to a good reason.
Even though i like your design - changing the views and grouping is a bug customer impact item and we should at least communicate to them before making any changes to prevent breaking them.

I dont see a reason why your design has to absolutely change the views and grouping - in general try not to change the existing functionality of the software too much when refactoring things. Unless there is a very good reason to do so which is communicated to our customers.

vedithal-amd, line 533: In addition to this HLD, we also need a phased wise implementation plan (think of it as LLD) where each phase is a self-consistent PR (it should ship with its tests, documentations) which maintains CI green. I know that this will cause some extra redundant work but that is the cost of phased wise implementation - it also makes blaming and rolling back easier if something goes wrong.

Each phased wise implementation can be its own markdown file which exactly nails down what changes to make and what tests and documentation to ship.

The HLD can just have one section containing one liners linking to each LLD.

I am emphasizing on the need of LLD because this HLD still reads somewhat vaguely in the sense that I dont know what exact change are you proposing to our software.

projects/rocprofiler-compute/docs/design/analysis-config-redesign/hld-analysis-config-redesign.md
```diff
@@ -0,0 +1,546 @@
...
+```
+
+
+## Requirements
+
+> **Assumption:** There are no pre-existing defined requirements the design must follow.
+>
+> **Constraints:** The design should not break existing working conditions unless specified.
+
+### Functional requirements
```
vedithal-amd, line 173: I think the metric sync mechanism in our tooling (tools dir) needs to adjust for this design. We have tooling for hash consistency check, yaml formatting, sets metric id validation, metric formula styling (I think there was some discussion about it) and finally the sync mechanism to sync the plain and rst metric descriptions to our documentation (docs).

While this is not a functional requirement it should be still be accounted for somewhere

projects/rocprofiler-compute/docs/design/analysis-config-redesign/hld-analysis-config-redesign.md
```diff
@@ -0,0 +1,546 @@
...
+
+```
+  "VALU Utilization" defined 7 times (3 shown):
+
+  gfx908/  ──►  VALU Utilization { formula, unit, description, peak }
+  gfx941/  ──►  VALU Utilization { ... unit="GFLOPs" }  (unit drifted)
+  gfx115x/ ──►  VALU Utilization { ... }  (different formula — different arch family)
+```
+
+### [PS3] Analysis config files mix unrelated concerns, making changes harder and riskier than they need to be
```
vedithal-amd, line 145: I would say that we should first strengthen our test_metric_valudation.py to use a mega kernel on a given gfx to validate all the metric values. This need not be run on the CI since this has to be a standalone profile without much background processes which needs exclusive GPU access which is tricky to get in CI. This can be run manually.

While this design does not propose formula changes - we have seen in the past that even unrelated things can cause serious breakages. Think about mistakes in refactoring or typos.

Overall metric values breakage is a risk which should be documented in this proposal and we should ensure we have a manually runnable test after every phase of implementation to make sure we dont break metrics which is a core service of our product

vedithal-amd (commented):
Provided my thoughts on the design

projects/rocprofiler-compute/docs/design/analysis-config-redesign/hld-analysis-config-redesign.md
xuchen-amd: The design supports modifications of values, so distinguish the architectural nuances is not a problem. A question that extends from yours, is, what makes two metrics the same metrics? My thinking is that for any two metrics with the same definition would be the same metric, and any negligible architectural nuances (with respective to profiling) should be omitted from the definition. If the nuances should not be negilited, then the two metrics are different. What do you think?

projects/rocprofiler-compute/docs/design/analysis-config-redesign/hld-analysis-config-redesign.md
xuchen-amd: I've added the LLD that includes the actual schema to resolve the issue of unlinked descriptions vs metrics: https://github.com/ROCm/rocm-systems/blob/a8ddb7622bee8da2da5c9374de79735e81ba27ea/projects/rocprofiler-compute/docs/design/analysis-config-redesign/lld-phase1-metric-library.md

projects/rocprofiler-compute/docs/design/analysis-config-redesign/hld-analysis-config-redesign.md
xuchen-amd: I believe we have a consensus here that YAML is still the formatting to use so that the analysis configs are readable, but not all aspects of the analysis configs need to be covered in the same file.

projects/rocprofiler-compute/docs/design/analysis-config-redesign/hld-analysis-config-redesign.md
xuchen-amd: please see the newly added [LLD](https://github.com/ROCm/rocm-systems/blob/a8ddb7622bee8da2da5c9374de79735e81ba27ea/projects/rocprofiler-compute/docs/design/analysis-config-redesign/lld-phase1-metric-library.md) where `variables.yaml` could be introduced to improve the design around placeholder variables.

projects/rocprofiler-compute/docs/design/analysis-config-redesign/hld-analysis-config-redesign.md
xuchen-amd: String IDs are self-identifiable, `mem.l2_cache_hit` tells you exactly what the metric is, while `17.1.3` requires looking up a table. For filtering, an average user benefits more from `-b mem.l2_cache_hit` than `-b 17.1.3` where the numbers carry no meaning. That said, numeric metric IDs will remain fully supported until string IDs mature and stabilize in production. Deprecation of numeric IDs will only happen after proper customer communication and a transition period.

projects/rocprofiler-compute/docs/design/analysis-config-redesign/hld-analysis-config-redesign.md
xuchen-amd: accounted for this in LLD.

projects/rocprofiler-compute/docs/design/analysis-config-redesign/hld-analysis-config-redesign.md
```diff
@@ -0,0 +1,555 @@
...
+            peak: "($max_sclk * $cu_per_gpu * 4096) / 1000"  # metric specification
+  metrics_description:
+    VALU Utilization: >-                # documentation
+      The percentage of time the VALU was busy...
+```
+
+
+## Problem statement
+
+### [PS1] No clear requirements constrain what analysis configs should provide
```
feizheng10, line 111: Could you point me where does PS1 has been addressed in this design?

vedithal-amd (approved):

svolkov-amd (approved):

PR #8503, opened 2026-07-13
feat(rocshmem): add _av targeted-ordering variants

joalsop:
Adds _av-suffix variants of rocshmem operations that relax MakeAvailable/ MakeVisible ordering semantics. When enabled, rocshmem operations still establish hb ordering, but they don't establish availability/visibility as defined in the AMD GPU LLVM memory model
(https://llvm.org/docs/AMDGPUMemoryModel.html#availability-and-visibility). Practically, this means that ordering semantics like fences are lowered to a blocking waitcnt rather than a waitcnt plus cache actions (L1/L2 flush or invalidate), but it also requires that accesses that are ordered by these semantics use cache-bypassing available/visible semantics to avoid data races.

- fence_av() / fence_av(int pe): GDA context fence with relaxed ordering
- putmem_nbi_wave_av / put_nbi_wave_av: wave-level NBI put variants
- putmem_wg_av: WG-level blocking put with targeted ordering
- ipcAMOSet_av: IPC AMO set with relaxed store semantics
- wait_until_av: P2P sync wait with targeted ordering
- ROCSHMEM_BYPASS_LANE_STORES env var for large memcpy_lane puts
- queue_pair_mlx5: cache-bypassing WQE store path for _av puts

## JIRA ID

JIRA ID: <AIROCSHMEM-#>

## Submission Checklist

### DIFF
--- a/projects/rocshmem/include/rocshmem/rocshmem.hpp
+++ b/projects/rocshmem/include/rocshmem/rocshmem.hpp
@@ -882,6 +882,20 @@ __device__ ATTR_NO_INLINE void rocshmem_ctx_fence(rocshmem_ctx_t ctx, int pe);
 __device__ ATTR_NO_INLINE void rocshmem_fence(int pe);
 
 /**
+ * @brief Targeted fence variant: orders outstanding memory operations to the
+ * remote PE using fence_targeted() (s_waitcnt vmcnt(0)) without a full
+ * system-scope cache flush.  Use in place of rocshmem_fence() when the caller
+ * can guarantee that only the target PE needs to observe the write ordering.
+ */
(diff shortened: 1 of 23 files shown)


omor1:
@joalsop for the GDA portions, can you take a look at the Queue Pair posting options mechanism from #7217? I think that these AV targeted-ordering variants could be a good target for implementation via these options.

I've discussed with @abouteiller having options to specify the scope to acquire/release the internal QP thread safety mechanisms (if e.g. the user knows that the usage pattern ensures that other threads/waves/workgroups don't access the same QP it doesn't need to go all the way to the device/agent scope), but that's something we're pushing until later.

PR #8506, opened 2026-07-13
rocclr: report CPU accessor in hipMemRangeGetAttribute(AccessedBy)

sourabhuday:
## Problem
`hipMemRangeGetAttribute(AccessedBy)` never reports the CPU, even after
`hipMemAdvise(SetAccessedBy, hipCpuDeviceId)`. Fails `AccessedBy_Basic` and
`AccessedBy_Partial_Range`.

## Cause
KFD/ROCr track CPU access only via the always-on `HSA_SVM_FLAG_HOST_ACCESS` flag,
which can't be distinguished from managed memory's default host access. The CPU
comes back as plain `AGENT_ACCESSIBLE` and is dropped, while GPUs work because
they use a distinct per-node `ACCESS_IN_PLACE` attribute.

## Fix
Track the explicit CPU advice in ROCclr (per page). Set/Unset update it, a fresh
allocation clears it, and the query reports the CPU only when the advice covers
the whole queried range (matching KFD's range-uniform behavior for GPUs).

## Test
- `AccessedBy_Basic`, `AccessedBy_Partial_Range`: pass (were failing)
- No regressions: `*hipMemRangeGetAttribute*` (41/41), `*hipMemAdvise*` (4113/4113)

### DIFF
--- a/projects/clr/rocclr/device/rocm/rocdevice.cpp
+++ b/projects/clr/rocclr/device/rocm/rocdevice.cpp
@@ -2804,12 +2804,17 @@ void UpdateCpuAccessedBy(const void* ptr, size_t count, bool set) {
   const uintptr_t base = SvmPageBase(ptr);
   const uintptr_t end = SvmPageEnd(ptr, count);
   std::lock_guard<std::mutex> lock(g_cpuAccessedByLock);
-  for (uintptr_t p = base; p < end; p += kSvmPageSize) {
-    if (set) {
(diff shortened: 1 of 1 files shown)


AlexXAmd (changes requested):
The performance impact of this commit may outweigh the benefit of the issue it is intended to fix.

A global lock and a global set:set per page for memory allocation, both are performance hit and considerable extra memory usage.

PR #8507, opened 2026-07-13
feat(rocprofiler-systems): Cache test configuration to be used across CTest run

kcossett-amd:
## Motivation

GFX950 workflows would intermittently skip tests. With #8478, I traced the skips to "No Valid GPU Available"... a failing `rocminfo` call. Because this is a shared machine and every PR runs this workflow (each CTest runs pytest, which runs `rocminfo`), concurrent `rocminfo` invocations likely raced and intermittently failed.

The fix: write a JSON file to `/tmp` that stores the pytest environment (GPU, capabilities, config) once, so subsequent tests reuse it rather than each calling `rocminfo` again.

This also has the obvious bonus that other runs will not have to recompute certain variables, and can avoid expensive subprocess calls, reducing overall test time.

## Technical Details

**One JSON file per build tree, shared by every process.**
The first process to need a value computes and writes it; every later process
reads it back.

- **Lifecycle:** the `rocprofiler-systems-pytest-config` setup fixture *clears and fully regenerates* the cache at the start of a `ctest` run; the cleanup fixture deletes the file at the end. 
 - If you end a test run early, the next run always regenerates cache file.
 - The size of the cache file is not very large. For example, with ~500 tests run, the file reaches ~10000 bytes (10 KiB)

- **File location:** `$TMPDIR/$USER/rocprofsys-syscache-<uid>-<discriminator>.tmp`.
  `<discriminator>` is a short SHA-256 of the build/install tree
  (`ROCPROFSYS_BUILD_DIR` / `ROCPROFSYS_INSTALL_DIR`, falling back to the package
  path).
- **`get_shared_cache()`** returns the process-wide cache (memoized, so the file is read at most once per process), or `None` when caching is disabled.
- A lock file also exists at the same level of the cache file. 
   - `_atomic_write` swaps the cache file's inode via os.replace and flock locks the inode, so locking the cache file would not serialize concurrent writers 

**Typed JSON codec (`encode` / `decode`).**
JSON natively handles `None`/`bool`/`int`/`float`/`str`/`list`/`dict`. Other types are stored as tagged objects (`{"__type__": ..., "value": ...}`):

**Failure semantics.**

- **Reads and I/O are fail-open:** a corrupt/stale entry (`DeserializationError`) or any I/O / lock error silently falls back to recomputing
- Note that `serialization` errors will force a fail as it means the value could never be cached (else it would silently re-probe every process)

See `DEVELOPMENT.md` on short note for how to use decorators.

 - Centralize `rocminfo` calls to `run_rocminfo`.
   - Raises an error only if `rocminfo` fails due to permission of HSA failure/race. Failure due to GPU-less device is not raised and is considered to be a GPU-less system.

## JIRA ID

Jira ID: AIPROFSYST-628

## Test Plan

Local + Workflows (I expect only a set of gfx950 tests to skip).

The set of `GFX950` tests that will be skipped (with reason) are:
```
jacobi-hip-sys-run             Test requires atleast 2 GPUs but system has 1
julia-vecadd-sys-run           Julia not available
scratch-memory-baseline        Test cannot run inside a Docker container
scratch-memory-binary-rewrite  Test cannot run inside a Docker container
scratch-memory-sampling        Test cannot run inside a Docker container
scratch-memory-sys-run         Test cannot run inside a Docker container
sdma-usage-sys-run             amdgpu 6.16.6 < required 6.19.14
transferbench-sys-run          Test requires atleast 2 GPUs but system has 1
video-decode-sampling          videodecode binary not found
video-decode-sys-run           videodecode binary not found
```

## Test Result

You will not see `[syscache]` messages by default. They require `ROCPROFSYS_TEST_CACHE_DEBUG=1`

### Local Testing 
(`ROCPROFSYS_TEST_CACHE_DEBUG=1 ctest --test-dir rocprof-sys-build -L quick`)

With `ROCPROFSYS_TEST_CACHE_DEBUG=1`, (new test only debug print var), we see `rocprofiler-systems-pytest-config` sets up the cache (it says overwriting as I had an existing cache file):
```
Overwriting old cache file: /tmp/rocm/rocprofsys-syscache-1000-4406d4561e44d378.tmp
[syscache] MISS: config.discover_build_config:(python_root_dirs=None,python_versions=None,rocm_optional=False)
[syscache] MISS: gpu.detect_gpu:(/opt/rocm-7.1.1)
[syscache] MISS: gpu.get_offload_extractor:(/opt/rocm-7.1.1)
[syscache] MISS: gpu.get_xnack_support:(/opt/rocm-7.1.1)
[syscache] MISS: cap.rocprofsys.config.RocprofsysConfig.llvm_lib_paths
[syscache] MISS: cap.rocprofsys.capabilities.SystemCapabilities.mpi_implementation
[...] # Other capabilities
[syscache] MISS: cap.rocprofsys.capabilities.SystemCapabilities.amdsmi_version
[syscache] MISS: cap.rocprofsys.capabilities.SystemCapabilities.amdgpu_version
[syscache] HIT: gpu.detect_gpu:(/opt/rocm-7.1.1)
[syscache] HIT: gpu.get_offload_extractor:(/opt/rocm-7.1.1)
[syscache] HIT: gpu.get_xnack_support:(/opt/rocm-7.1.1)
```
The 3 hits are from generating the header, as it looks up GPU information.
The information is then written to the cache file and in subsequent tests (for example `transpose-sampling`):
```
============================= test session starts ==============================
collecting ... [syscache] HIT: config.discover_build_config:(python_root_dirs=None,python_versions=None,rocm_optional=False)
[syscache] HIT: cap.rocprofsys.capabilities.SystemCapabilities.mpiexec_exec
[syscache] HIT: cap.target_support_mpi:(/home/rocm/rocm-systems-branches/pytest-caching/projects/rocprofiler-systems/rocprof-sys-build/transpose)
[syscache] HIT: gpu.detect_gpu:(/opt/rocm-7.1.1)
collected 1 item

../../transpose-sampling [syscache] HIT: cap.rocprofsys.capabilities.SystemCapabilities.num_procs
[syscache] HIT: cap.rocprofsys.config.RocprofsysConfig.llvm_lib_paths
[syscache] HIT: gpu.get_target_gpu_arch:(/opt/rocm-7.1.1,/home/rocm/rocm-systems-branches/pytest-caching/projects/rocprofiler-systems/rocprof-sys-build/transpose)
SUBPASSED[Regex validation]
```
The reason the `[syscache]` messages are seemingly placed "randomly" is because `pytest` checks certain conditions at different times. 

**Note: You will NOT see Hit/Miss messages without setting `ROCPROFSYS_TEST_CACHE_DEBUG=1`**

### Workflow

Currently running

## Submission Checklist


projects/rocprofiler-systems/tests/pytest/test_julia.py
```diff
@@ -21,11 +23,13 @@
 # =============================================================================
 
 
-@lru_cache(maxsize=1)
+@persistent_cache("julia.lib_paths")
```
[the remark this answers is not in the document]
kcossett-amd: I'll correct the docstring. `lru_cache` is meaningless as we use CTest to drive pytest, and CTest spawns a new pytest process for each test it runs.

projects/rocprofiler-systems/tests/pytest/rocprofsys/cache.py
```diff
@@ -0,0 +1,329 @@
...
+the CTest-spawned processes share. The first process that needs a value pays for
+it once and writes it; every later process reads it back.
+
+Set ``ROCPROFSYS_DISABLE_CAPCACHE=1`` to disable the cache entirely.
+"""
+
+from __future__ import annotations
+
+import fcntl
+import functools
```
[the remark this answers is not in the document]
kcossett-amd: Tests don't run on windows, so this is fine.

projects/rocprofiler-systems/tests/pytest/rocprofsys/cache.py
```diff
@@ -0,0 +1,329 @@
...
+    """Return the process-wide cache, or ``None`` if disabled/unavailable.
+
+    The result is memoized so the file is read at most once per process.
+    """
+    if os.environ.get(_DISABLE_ENV, "").strip() in ("1", "ON", "on", "true", "True"):
+        return None
+    try:
+        return PersistentCache(_cache_file_path(_cache_discriminator()))
+    except Exception:
+        return None  # fail-open: caching disabled, callers compute normally
```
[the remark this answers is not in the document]
kcossett-amd: Tests don't run on windows.

projects/rocprofiler-systems/tests/pytest/rocprofsys/cache.py
```diff
@@ -0,0 +1,329 @@
...
+            self._entries = raw_entries
+
+    def get(self, key: str) -> tuple[bool, Any]:
+        """Return ``(found, value)``. ``found`` is False on miss or any error."""
+        if key not in self._entries:
+            return False, None
+        try:
+            return True, decode(self._entries[key])
+        except (ValueError, TypeError, KeyError):
+            return False, None
```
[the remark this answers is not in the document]
kcossett-amd: Skipped intentionally. `rocprofiler-systems-pytest-config` fixture fully generates the cache before any test process starts (`FIXTURE_REQUIRED`) so every test process already loads a complete cache.

Furthermore, even if this were a problem, most subprocess calls are blocked on a "need to know" basis.

projects/rocprofiler-systems/tests/pytest/rocprofsys/cache.py
```diff
@@ -0,0 +1,329 @@
...
+        try:
+            self.path.parent.mkdir(parents=True, exist_ok=True)
+            lock_fd = os.open(self.path, os.O_RDWR | os.O_CREAT, 0o600)
+            try:
+                fcntl.flock(lock_fd, fcntl.LOCK_EX)
+                data = {"entries": {}}
+                tmp = f"{self.path}.{os.getpid()}.tmp"
+                with open(tmp, "w") as f:
+                    json.dump(data, f)
+                os.replace(tmp, self.path)
```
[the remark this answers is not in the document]
kcossett-amd: Will do

projects/rocprofiler-systems/tests/pytest/rocprofsys/cache.py
```diff
@@ -0,0 +1,329 @@
...
+                    data["entries"] = existing["entries"]
+            except (OSError, ValueError):
+                pass  # missing/corrupt -> start fresh
+
+            data["entries"][key] = encoded
+
+            tmp = f"{self.path}.{os.getpid()}.tmp"
+            with open(tmp, "w") as f:
+                json.dump(data, f)
+            os.replace(tmp, self.path)
```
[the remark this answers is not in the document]
kcossett-amd: Will do

projects/rocprofiler-systems/tests/pytest/rocprofsys/cache.py
```diff
@@ -0,0 +1,329 @@
...
+            inst_dict = instance.__dict__
+        except AttributeError:  # e.g. __slots__ without __dict__
+            return self.func(instance)
+
+        if self.attrname in inst_dict:
+            return inst_dict[self.attrname]
+
+        cache_key = f"cap.{self.attrname}"
+        store = get_shared_cache()
+        if store is not None:
```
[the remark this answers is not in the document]
kcossett-amd: I'll switch it to `cap.<ClassQualName>.<attr>` (e.g. `cap.SystemCapabilties.mpi_implementation`)

projects/rocprofiler-systems/tests/pytest/rocprofsys/cache.py
```diff
@@ -0,0 +1,442 @@
...
+_proc_cache_disabled = False
+
+# Debug aid, prints HIT/MISS lines for cache lookup attempts.
+_DEBUG_ENV = "ROCPROFSYS_TEST_CACHE_DEBUG"
+
+
+def _debug_log(event: str, cache_key: str) -> None:
+    """Print a ``[syscache] HIT/MISS: <key>`` line when debugging is enabled."""
+    if os.environ.get(_DEBUG_ENV, "").strip() in ("1", "ON", "on", "true", "True"):
+        print(f"[syscache] {event}: {cache_key}")
```
mradosav-amd: why don't we use `logging` module for this kind of output instead of print?

projects/rocprofiler-systems/tests/pytest/rocprofsys/cache.py
```diff
@@ -0,0 +1,442 @@
...
+# Wrap other types as tagged objects:
+# {"__type__": "<kind>", "<value>": <payload>}
+#
+# Dataclasses are handled automatically: any dataclass instance is stored by
+# value (its ``init=True`` fields, each recursively encoded) tagged with the
+# defining ``module``/``qualname``, and decode() re-imports that class to rebuild
+# it.
+# ---------------------------------------------------------------------------
+def encode(obj: Any) -> Any:
+    if obj is None or isinstance(obj, (bool, int, float, str)):
```
mradosav-amd, line 89: maybe we can replace `if isinstance` pattern with table 

```python
_ENCODERS = {
Path:  lambda o: {"__type__": "path",  "value": str(o)},
# etc
}
# ...
handler = _ENCODERS.get(type(obj))
if handler:
        return handler(obj)
# ...
```

We can do it for encode and decode as well.

projects/rocprofiler-systems/tests/pytest/rocprofsys/cache.py
```diff
@@ -0,0 +1,442 @@
...
+                    existing = json.load(f)
+                if isinstance(existing, dict) and isinstance(
+                    existing.get("entries"), dict
+                ):
+                    data["entries"] = existing["entries"]
+            except (OSError, ValueError):
+                pass  # missing/corrupt -> start fresh
+
+            data["entries"][key] = encoded
+            self._atomic_write(data)
```
mradosav-amd: We can maybe split, to multiple functions. Will be more cleaner.

`_read_disk_state` - only read file(`_load` can reuse this)
`_flush` - only write file
`_persist` - coordinate 

```python
def _persist(self, key: str, encoded: Any) -> None
    with self._exclusive_lock():
        data = self._read_disk_state()
        data["entries"][key] = encoded
        self._flush(data)
```

mradosav-amd (approved):
Looks good. Thanks!
Few suggestions for better readability and easier maintaining in future.

projects/rocprofiler-systems/tests/pytest/rocprofsys/cache.py
```diff
@@ -0,0 +1,442 @@
...
+    )
+    return hashlib.sha256(root.encode("utf-8")).hexdigest()[:16]
+
+
+def _cache_file_path(discriminator: str) -> Path:
+    # Resolve the tmp base exactly like the suite's cleanup step
+    # (ROCPROFSYS_TMPDIR -> TMPDIR -> /tmp)
+    base = os.environ.get("ROCPROFSYS_TMPDIR", os.environ.get("TMPDIR", "/tmp"))
+    try:
+        uid = os.getuid()
```
marantic-amd, line 317: Does it make sense to extract this into a helper, something like this:

```python
def resolve_username() -> str:
    try:
        return getpass.getuser()
    except Exception:
        try:
            return str(os.getuid())
        except Exception:
            return "0"
```
Since this is repeating pattern in another 3-4 places in `config.py` na `conftest.py`

projects/rocprofiler-systems/tests/pytest/rocprofsys/config.py
```diff
@@ -515,6 +516,7 @@ def discover_install_config(
     )
 
 
+@persistent_cache("config.discover_build_config")
```
marantic-amd, line 502: can we add a small comment about `build_dir` always being env-defaulted value, so that we are aware that this is a intended behaviour

projects/rocprofiler-systems/tests/pytest/rocprofsys/cache.py
```diff
@@ -0,0 +1,442 @@
...
+        self.path = path
+        # Lock a separate sidecar file, not the cache file itself: _atomic_write
+        # swaps the cache file's inode via os.replace and flock locks the inode,
+        # so locking the cache file would not serialize concurrent writers
+        # (lost writes in CTest parallel mode).
+        self.lock_path = path.with_name(f"{path.stem}.lock{path.suffix}")
+        self._entries: dict[str, Any] = {}
+        self._load()
+
+    def _load(self) -> None:
```
marantic-amd: Could we maybe extract the logic for this inbto a common helper like `_read_entries_from_disk(self) -> dict[str, Any]`, which would allow us that `_load` will just become something like
```python
self._entries = self._read_entries_from_disk()`
``` 
Or it will become obsolete.

projects/rocprofiler-systems/tests/pytest/rocprofsys/cache.py
```diff
@@ -0,0 +1,442 @@
...
+
+    def _atomic_write(self, data: dict[str, Any]) -> None:
+        """Replace the data file with ``data`` atomically (caller holds the lock)."""
+        tmp = f"{self.path}.{os.getpid()}.tmp"
+        fd = os.open(tmp, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
+        with os.fdopen(fd, "w") as f:
+            json.dump(data, f)
+        os.replace(tmp, self.path)
+
+    def _write_locked(self, key: str, encoded: Any) -> None:
```
marantic-amd: This can also reuse the helper from https://github.com/ROCm/rocm-systems/pull/8507/changes#r3621770600

projects/rocprofiler-systems/tests/pytest/rocprofsys/cache.py
marantic-amd: I would also push on this split. Makes code more understandable and less error prone

projects/rocprofiler-systems/tests/pytest/rocprofsys/cache.py
```diff
@@ -0,0 +1,442 @@
...
+        try:
+            fcntl.flock(lock_fd, fcntl.LOCK_EX)
+            yield
+        finally:
+            fcntl.flock(lock_fd, fcntl.LOCK_UN)
+            os.close(lock_fd)
+
+    def _atomic_write(self, data: dict[str, Any]) -> None:
+        """Replace the data file with ``data`` atomically (caller holds the lock)."""
+        tmp = f"{self.path}.{os.getpid()}.tmp"
```
marantic-amd: Could we use `tempfile.mkstemp` here. I think that with that method the permissions are set automatically.

projects/rocprofiler-systems/tests/pytest/conftest.py
```diff
@@ -1280,6 +1286,85 @@ def _ctest_generate_tests(
...
+            seen.add(name)
+            try:
+                getattr(obj, name)
+            except SerializationError:
+                raise
+            except Exception:
+                pass
+
+
+def _seed_capability_cache() -> None:
```
marantic-amd: This can also be split into 3 logical units  
`reset_cache_for_regeneration`
`warm_gpu_probes`
`warm_capabilities`

projects/rocprofiler-systems/tests/pytest/rocprofsys/cache.py
```diff
@@ -0,0 +1,442 @@
...
+    return obj
+
+
+# ---------------------------------------------------------------------------
+# PersistentCache
+# ---------------------------------------------------------------------------
+class PersistentCache:
+    """A JSON key/value store shared across processes for one build tree."""
+
+    def __init__(self, path: Path):
```
marantic-amd:
```suggestion
    def __init__(self, path: Path) -> None:
```

projects/rocprofiler-systems/tests/pytest/rocprofsys/cache.py
```diff
@@ -0,0 +1,442 @@
...
+class persistent_cached_property:
+    """``functools.cached_property`` that also persists across processes.
+
+    When reading, it looks up the value in the following order:
+     - instance ``__dict__`` (this process)
+     - shared file cache (another process)
+     - compute (if necessary, then written to shared file)
+    """
+
+    def __init__(self, func: Callable):
```
marantic-amd:
```suggestion
    def __init__(self, func: Callable) -> None:
```

projects/rocprofiler-systems/tests/pytest/test_pytest_impl.py
```diff
@@ -201,3 +264,237 @@ def validate_rocpd_database(db_path: Path) -> ValidationResult:
...
+@pytest.mark.class_name("cache")
+class TestCache(RocprofsysTest):
+    """Unit tests for the persistent cross-process cache (``rocprofsys.cache``).
+
+    One test per component to keep the count small; each walks that component's
+    behaviours in a single flow. The decorator/property tests use the
+    ``cache_store`` fixture so they never touch the real shared cache file.
+    """
+
+    def test_codec_round_trips_supported_types_and_rejects_others(self):
```
marantic-amd:
```suggestion
    def test_codec_round_trips_supported_types_and_rejects_others(self) -> None:
```
This goes for all newly added tests here

projects/rocprofiler-systems/tests/pytest/rocprofsys/cache.py
kcossett-amd: For `decode`, sure. 

For encode though, a `dict` keyed on a type does exact-type matching, whereas encode needs subclass matching (e.g. `Path` is really a `PosixPath`, `bool` is an `int`), so a table doesn't work there.

projects/rocprofiler-systems/tests/pytest/rocprofsys/cache.py
kcossett-amd: Sounds good. I'll follow the convention then.

projects/rocprofiler-systems/tests/pytest/rocprofsys/config.py
kcossett-amd: Sure, let me also remove some unused arguments from both `discover_*_config` methods.

projects/rocprofiler-systems/tests/pytest/rocprofsys/cache.py
kcossett-amd: Good call, thanks, will use.

projects/rocprofiler-systems/tests/pytest/conftest.py
kcossett-amd: I'll merge `config+capabilities` into just `_warm_cached_properties` and take the rest of your suggestion.

dgaliffiAMD (approved):

PR #8510, opened 2026-07-13
[CI] [Docs] Fix an edge case missed by update-docs.yml

alexxu-amd:
## Motivation

To resolve a "project not found" error from update-docs.yml: https://github.com/ROCm/rocm-systems/actions/runs/29004064811/job/86071296796

## Technical Details

The pipeline hardcoded project_slug uses `advanced-micro-device-` + project folder name. Works for most of the projects. However, in some edge cases (i.e. rocdbgapi), the formula fails.

To resolve this, added a lookup loop to extract `project_slug` through a pre-defined lookup table.

## JIRA ID

JIRA ID: AIROCDOC-4084

## Test Plan

Locally ran the added lookup code, check if output matches true project_slugs.

## Test Result

image

## Submission Checklist

### DIFF
--- a/.github/docs-config.json
+++ b/.github/docs-config.json
@@ -84,6 +84,11 @@
       "folder": "hipfile",
       "rtd_slug": "advanced-micro-devices-hipfile",
       "rtd_alias": "hipFile"
+    },
+    {
+      "folder": "rocdbgapi",
+      "rtd_slug": "advanced-micro-devices-rocdbgapi-docs",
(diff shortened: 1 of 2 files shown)


lumachad (commented):
Looks reasonable to me to a point, but I'm not used to this code. I can approve it but would appreciate some other eyes more familiar with this workflow.

neon60 (approved):

PR #5264, opened 2026-04-21
feat(rdc): JIRA ID #218 RDC CI and RDC Test

yalmusaf:
## Motivation

Add project-specific GitHub Actions coverage for RDC build, packaging, and tests in the rocm-systems monorepo while RDC integration with TheRock CI is completed.

TheRock remains the repository-wide source of truth. The standalone `.github/workflows/rdc-ci.yml` is temporary and is limited to RDC packaging and test coverage. The legacy nested workflow at `projects/rdc/.github/workflows/main.yml` is left unchanged, and `.github/workflows/rdc-ci.yml` is listed as a skippable path in `therock_configure_ci.py` so workflow-only changes do not trigger TheRock.

## Technical Details

- Adds a three-job RDC pipeline: `build`, non-GPU `test`, and opt-in `test-gpu`.
- Uses `rocm/dev-ubuntu-22.04:7.2` for build and test jobs to avoid cross-version ABI risk.
- Builds gRPC 1.67.1 from source into `/opt/grpc` using only the required CMake provider submodules.
- Builds AMD SMI from `projects/amdsmi` and produces DEB packages only.
- Builds RDC with configurable `GPU_TARGETS` using the repository variable with a `gfx942` fallback.
- Installs generated packages with `apt-get` so dependency failures are not masked.
- Adds explicit token permissions, concurrency cancellation, and job timeouts.
- Adds pytest coverage for installation, CLI behavior, Python bindings, integration assets, negative paths, parser helpers, daemon resilience, performance, and the packaged GTest runner.
- Hardens GPU detection, subprocess cleanup, unauthenticated argument handling, and negative-path assertions.

## JIRA ID

JIRA ID: AILITOOLS-218

## Test Plan

- Parse both RDC workflow YAML files.
- Run the 26 `therock_configure_ci.py` unit tests.
- Build gRPC 1.67.1, AMD SMI, and RDC in `rocm/dev-ubuntu-22.04:7.2`.
- Install the generated AMD SMI and RDC DEBs in a fresh `rocm/dev-ubuntu-22.04:7.2` container.
- Run all non-GPU workflow test commands: 57 tests passed locally.
- Run the optional GPU suite when `RDC_GPU_TESTS_ENABLED=true` on a GPU runner.

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #8511, opened 2026-07-13
ci: invert .github/workflows path filter for TheRock CI skip

### LINKED ISSUE #7849: CI Path filtering issue is not updated , better solution is to invert the path selection
https://github.com/ROCm/rocm-systems/blob/5ab64a49b946c84eafe9f5e4adb94202fda87079/.github/scripts/therock_configure_ci.py#L119-L143 has a partial list of files in .github/workflows/ that are treated as "skippable". We can either extend that list to include the new files or (my preference) invert the check for .github/workflows/ and only run "TheRock CI" if a file known to be related to it is modified, like in https://github.com/ROCm/TheRock/blob/8484067b044134dc2a4b1f3fe45c415b2a631a17/build_tools/github_actions/configure_ci_path_filters.py#L166-L232
super repo for rocm systems projects. Contribute to ROCm/rocm-systems development by creating an account on GitHub.
 
Looking through https://github.com/ROCm/rocm-systems/tree/develop/.github/workflows I think all workflows are prefixed with therock, so could use regex or string contains. We have a more sophisticated check in TheRock itself but I don't think that's necessarily needed here: https://github.com/ROCm/TheRock/blob/8484067b044134dc2a4b1f3fe45c415b2a631a17/build_tools/github_actions/tests/configure_ci_path_filters_test.py#L59-L85
super repo for rocm systems projects. Contribute to ROCm/rocm-systems development by creating an account on GitHub.
 
Elaborating on why I don't think overriding is appropriate here:
We've now seen several instances of changes that can't possibly affect CI running CI and then blocking. Fixing now will likely help later.
We'd like for overrides to be rare events for situations outside the norm - flaky hardware, flaky tests, flaky network downloads, etc. Here while there are test timeouts (which would be good to fix for PRs where that does matter), developers do have enough tools already to get through this: adjust the CI triggers so this blocking check doesn't run at all 

Another instance - https://teams.microsoft.com/l/message/19:mts01nzrUNtAfxYZqYpo-l4OOfyHQxdNiOHUikvwBF41@thread.tacv2/1782407874650?tenantId=3dd8961f-e488-4e60-8e11-a82d994e183d&groupId=69e7275a-4d49-495d-8bd5-04fb519c8e9c&parentMessageId=1782407874650&teamName=AIG%20ROCm&channelName=DevOps%20-%20Support%20or%20Help&ngc=true&allowXTenantAccess=true

marifamd:
## Motivation

New non-TheRock workflows added under `.github/workflows/` (e.g. `rdc-ci.yml`) incorrectly trigger TheRock CI and block PRs, because CI-skip logic uses a hand-maintained blocklist that must be extended for every new workflow. This has been reported repeatedly. Inverting the check (only run CI for known-related workflow files) fixes the class of problem permanently.

## Technical Details

**CI path filter** (`.github/scripts/therock_configure_ci.py`):
- `is_path_skippable()` now treats any `.github/workflows/` file as skippable unless it matches the existing CI patterns (`therock*`), reusing `is_path_workflow_file_related_to_ci()`
- Removed the redundant `labeler.yml`, `amdsmi-manylinux-build.yml`, and `rocjitsu-corpus-tests.yml` entries from `SKIPPABLE_PATH_PATTERNS`
- Scope limited to `.github/workflows/` only; `.github/scripts/` behavior is unchanged so CI-gating scripts still trigger CI

**Tests** (`.github/scripts/tests/therock_configure_ci_test.py`):
- Added cases asserting unrelated workflows are skippable and TheRock CI workflow/script files are non-skippable

## JIRA ID

Closes #7849

## Test Plan

- `python3 -m pytest .github/scripts/tests/therock_configure_ci_test.py`

## Test Result

- 28 passed

### DIFF
--- a/.github/scripts/tests/therock_configure_ci_test.py
+++ b/.github/scripts/tests/therock_configure_ci_test.py
@@ -123,6 +123,22 @@ class ConfigureCITest(unittest.TestCase):
             therock_configure_ci.is_path_skippable(".github/workflows/labeler.yml")
         )
 
+        # Workflow/script files unrelated to TheRock CI are skippable without
+        # needing to be enumerated (issue #7849).
+        self.assertTrue(
+            therock_configure_ci.is_path_skippable(".github/workflows/rdc-ci.yml")
+        )
+        self.assertTrue(
+            therock_configure_ci.is_path_skippable(
+                ".github/workflows/amdsmi-manylinux-build.yml"
+            )
+        )
+        self.assertTrue(
+            therock_configure_ci.is_path_skippable(
+                ".github/workflows/rocjitsu-corpus-tests.yml"
+            )
+        )
+
         # Test non-skippable patterns
         self.assertFalse(
             therock_configure_ci.is_path_skippable("projects/rocminfo/src/main.cpp")
@@ -130,6 +146,17 @@ class ConfigureCITest(unittest.TestCase):
         self.assertFalse(therock_configure_ci.is_path_skippable("CMakeLists.txt"))
         self.assertFalse(
             therock_configure_ci.is_path_skippable("projects/rocminfo/test/test.cpp")
+        )
+
+        # TheRock CI workflow and script files are non-skippable so changes to
+        # them still trigger CI.
+        self.assertFalse(
+            therock_configure_ci.is_path_skippable(".github/workflows/therock-ci.yml")
+        )
+        self.assertFalse(
+            therock_configure_ci.is_path_skippable(
+                ".github/scripts/therock_configure_ci.py"
+            )
         )
 
     def test_check_for_non_skippable_path(self):
--- a/.github/scripts/therock_configure_ci.py
+++ b/.github/scripts/therock_configure_ci.py
@@ -144,14 +144,17 @@ SKIPPABLE_PATH_PATTERNS = [
     "experimental/python/perfxpert/*",
     ".github/CODEOWNERS",
     ".github/label*.yml",
-    ".github/workflows/labeler.yml",
-    ".github/workflows/amdsmi-manylinux-build.yml",
-    ".github/workflows/rocjitsu-corpus-tests.yml",
 ]
 
 
 def is_path_skippable(path: str) -> bool:
     """Determines if a given relative path to a file matches any skippable patterns."""
+    # A .github/workflows/ file only affects TheRock CI when it matches the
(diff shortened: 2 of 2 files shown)


ScottTodd (approved):
Thanks, the logic and unit tests here look correct to me

.github/scripts/tests/therock_configure_ci_test.py
```diff
@@ -123,6 +123,22 @@ def test_is_path_skippable(self):
             therock_configure_ci.is_path_skippable(".github/workflows/labeler.yml")
         )
 
+        # Workflow/script files unrelated to TheRock CI are skippable without
+        # needing to be enumerated (issue #7849).
+        self.assertTrue(
+            therock_configure_ci.is_path_skippable(".github/workflows/rdc-ci.yml")
```
yalmusaf, line 129: @marifamd I have the exact same lines added in the rdc-ci pr should i remove it?

.github/scripts/therock_configure_ci.py
```diff
@@ -144,14 +144,17 @@ def check_trigger_windows_ci_for_subtree(subtree: str) -> bool:
     "experimental/python/perfxpert/*",
     ".github/CODEOWNERS",
     ".github/label*.yml",
-    ".github/workflows/labeler.yml",
-    ".github/workflows/amdsmi-manylinux-build.yml",
-    ".github/workflows/rocjitsu-corpus-tests.yml",
 ]
```
yalmusaf, line 148: I added the rdc-ci to the SKIPPABLE_PATH_PATTERNS here as well: https://github.com/ROCm/rocm-systems/pull/5264/changes#diff-799b835c8c3eb43f09573a4e8745d19d8deb421adab173762c5379f4b19cf2d1 should that be removed?

geomin12 (approved):

yalmusaf (approved):

.github/scripts/tests/therock_configure_ci_test.py
marifamd: when this merges, you can rebase your PR ontop of develop and resolve any conflicts

.github/scripts/therock_configure_ci.py
marifamd: when this merges, you can rebase your PR ontop of develop and resolve any conflicts

PR #8521, opened 2026-07-14
rocr: Preserve metadata through intercept queues

bwelton:
## Summary
Preserve AMD metadata-prefetch packets when HSA queues are wrapped by `InterceptQueue`, so profiler/interceptor paths do not silently drop metadata needed by supported hardware.

## Changes

- `projects/rocr-runtime/runtime/hsa-runtime/core/inc/intercept_queue.h`
  - Add paired AQL/metadata storage for overflow and pre-interceptor snapshots.
  - Add proxy metadata-ring state and the host-only test construction seam.
- `projects/rocr-runtime/runtime/hsa-runtime/core/runtime/intercept_queue.cpp`
  - Reuse the public `AqlMetadataPrefetchPacket` representation.
  - Redirect metadata-ring queries to the intercept queue's proxy ring.
  - Match rewritten packets to original metadata using exact completion-signal matching followed by signal-agnostic fallback.
  - Preserve metadata through insertion, deletion, reorder, replacement, overflow, and retry-barrier paths.
  - Update event IDs for standard and AMD vendor dispatch/barrier packets.
  - Reset consumed proxy metadata slots to prevent stale metadata after ring reuse.
  - Forward metadata version and queue-property queries.
- `projects/rocr-runtime/rocrtst/suites/functional/intercept_queue_metadata_test.cc`
  - Add 23 host-only GTest cases covering metadata preservation, packet transformations, event IDs, vendor packets, wraparound, stale-slot reset, overflow/retry, unsupported metadata, null signals, and intercept markers.
- `projects/rocr-runtime/rocrtst/suites/test_common/CMakeLists.txt`
  - Add the standalone test target, exclude it from the aggregate source list, and register/install it with CTest.

## Design

- No new public HSA API or ABI-table changes.
- Uses the current public `core::AqlMetadataPrefetchPacket` type; no duplicate metadata definitions are added.
- Metadata is written before the wrapped AQL packet header is published.
- Unmatched inserted/replaced packets receive four-invalid-header no-op metadata.

## Testing

- `clang-format -style=file --dry-run --Werror` on all changed C++ files.
- Production syntax check for `intercept_queue.cpp` with `clang++ -fsyntax-only`.
- Standalone `intercept_queue_metadata` target build with `/usr/bin/cmake --build`.
- `ctest -R '^intercept_queue_metadata$' -j1 --output-on-failure`.
- Direct GTest execution: 23/23 tests passed.

## Limitations

Full ROCR runtime build/install and GPU-backed validation were not available in the development environment. This draft is intended for review of the implementation and host-side coverage before device validation.

### DIFF
--- a/projects/rocr-runtime/rocrtst/suites/functional/intercept_queue_metadata_test.cc
+++ b/projects/rocr-runtime/rocrtst/suites/functional/intercept_queue_metadata_test.cc
@@ -1,0 +1,839 @@
+//===- intercept_queue_metadata_test.cc - InterceptQueue metadata tests --===//
+//
+// SPDX-License-Identifier: NCSA
+//
+//===----------------------------------------------------------------------===//
+
+#include <cstdlib>
+#include <cstring>
+#include <memory>
+#include <vector>
+
+#include "core/inc/intercept_queue.h"
+#include "core/util/atomic_helpers.h"
+#include "gtest/gtest.h"
+#include "inc/amd_hsa_signal.h"
+
+namespace rocr {
+namespace core {
+
+// The host-only target constructs DoorbellSignal without the runtime signal
+// implementation. It never exposes or destroys that signal through HSA.
+Signal::~Signal() {}
+void Signal::registerIpc() {}
+void SharedSignalPool_t::clear() {}
+SharedSignal* SharedSignalPool_t::alloc() { return nullptr; }
+void SharedSignalPool_t::free(SharedSignal*) {}
+
+}  // namespace core
+}  // namespace rocr
+
+namespace {
+
+using AqlPacket = rocr::core::AqlPacket;
+using MetadataPacket = rocr::core::AqlMetadataPrefetchPacket;
+using Queue = rocr::core::Queue;
+using SharedQueue = rocr::core::SharedQueue;
+
+constexpr uint32_t kQueueSize = 8;
+
+class QueueStorage {
+ protected:
+  SharedQueue storage_{};
+};
+
+class FakeQueue final : private QueueStorage, public Queue {
+ public:
+  explicit FakeQueue(bool metadata_supported = true)
+      : Queue(&storage_, 0, nullptr), metadata_supported_(metadata_supported) {
+    ring_.resize(kQueueSize);
+    metadata_ring_.resize(kQueueSize);
+    amd_queue_.hsa_queue.size = kQueueSize;
+    amd_queue_.hsa_queue.base_address = ring_.data();
+    for (auto& metadata : metadata_ring_) metadata = NoopMetadata();
+  }
+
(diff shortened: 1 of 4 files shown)


projects/rocr-runtime/rocrtst/suites/functional/intercept_queue_metadata_test.cc
```diff
@@ -0,0 +1,839 @@
+//===- intercept_queue_metadata_test.cc - InterceptQueue metadata tests --===//
+//
+// SPDX-License-Identifier: NCSA
```
cfreeamd, line 3: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `SPDX-License-Identifier: NCSA` — the rest of ROCr uses `SPDX-License-Identifier: MIT`. NCSA (University of Illinois/NCSA Open Source License) and MIT are compatible but are distinct licenses. Verify this is intentional or change to MIT for consistency.

cfreeamd (commented):

ROCr review of intercept queue metadata preservation. The `intercept_queue.h`/`.cpp` implementation looks well-structured — the metadata write ordering, matching heuristic, and test-seam design are all correct. One SPDX license nit on the new test file. Note: the `rocrtst_yaml_missing` checklist flag is a false positive — the test is correctly built as a standalone CTest binary, not part of rocrtst.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/intercept_queue.cpp
```diff
@@ -200,17 +336,48 @@ void InterceptQueue::Submit(const void* pkts, uint64_t pkt_count, uint64_t user_
   InterceptQueue* queue = reinterpret_cast<InterceptQueue*>(data);
   const AqlPacket* packets = (const AqlPacket*)pkts;
 
+  const MetadataPacket* metadata = nullptr;
+  if (queue->metadata_ring_buf_) {
+    queue->metadata_scratch_.resize(pkt_count);
+    for (uint64_t i = 0; i < pkt_count; ++i) {
+      OriginalPacketInfo* match = nullptr;
+      for (auto& original : queue->originals_) {
```
saleelk, line 344: Output packet i almost always came from original packet i. You can try positional matching first and only fall back to the scan for mismatches, 
Also 
If the positional fast-path misses, you could build a hash map keyed on the packet content, that might help graphs, it could become O(m+n) avg case instead of always O(m*n)

this could help graphs?

saleelk (commented):
Change looks good to me

bwelton:
Superseded by #8587 (compliant branch name, Conventional Commits title, JIRA tracking, and a follow-up fix deriving the masked-packet-match size instead of a hardcoded literal). Closing this PR in favor of the replacement.

PR #8524, opened 2026-07-14
feat(rocprof-sys): Add hipFile GPU-direct storage I/O telemetry collection

sputhala-amd:
## Motivation

ROCm Systems Profiler should be able to provide storage telemetry — letting users correlate file/buffer registrations, read/write throughput, fast-path vs. fallback usage, and I/O errors with GPU and application activity in a single Perfetto trace or RocPD database. This is increasingly important for GPU-direct storage workloads where the storage path is a common performance bottleneck that was previously invisible in the profiler.

Depends on:

- [ ]  https://github.com/ROCm/rocm-systems/pull/8252
- [ ] https://github.com/ROCm/rocm-systems/pull/9584

Todo:

- [ ] Ensure version gating for hipFile public API on 

## Technical Details

- **New optional hipfile backend** `(source/lib/backends/hipfile/`) that links libhipfile via the hip::hipfile imported target (consistent with how amd_smi and other ROCm libraries are consumed) and wraps hipFile's in-process hipFileGetStatsL3 query.
- **New PMC collector** (`pmc/collectors/hipfile/collector.hpp`) modeled on the AMD-SMI GPU collector. Because hipFile exposes a single process-wide call (rather than per-device enumeration), the collector snapshots hipFileGetStatsL3 once per sampling interval and emits per-GPU counters.
- **Metrics reported per GPU** (as `hipFile GPU<N> <metric>`tracks): read/write bytes, read/write bandwidth, total/fastpath/fallback/unaligned read & write op counts, and read/write errors; plus two process-global counters (file and buffer registrations) reported on GPU 0.
- **Per-interval deltas, not cumulative totals** — the collector keeps the previous snapshot and reports the change per interval (with reset-guarding); bandwidth is recomputed each interval as delta(bytes) / delta(duration) rather than delta-ing hipFile's lifetime average.
- Output via the existing `pmc_event_with_sample` path — no new trace-cache sample type or processor vtable changes. Added a hipfile category and a corresponding entry in the perfetto processor's PMC_TRACK_MAP so counters land in both Perfetto counter tracks and RocPD PMC tables.
- **Build-time gating**: new `ROCPROFSYS_USE_HIPFILE` CMake option (default OFF); find_package(hipfile) is optional, so builds against ROCm versions without hipFile degrade gracefully (no hard dependency). `ROCPROFSYS_BUILD_HIPFILE` drives conditional compilation.
- **Runtime gating**: new `ROCPROFSYS_USE_HIPFILE` setting (default disabled) drives the collector on the process-sampling thread; the collector auto-sets HIPFILE_STATS_LEVEL=1 in the target (without overriding a user-set value). Zero overhead when disabled or when the target doesn't use hipFile.
- **Automation test**: added a hipfile-test example workload (looped registration + read/write, adapted from hipFile's examples) plus tests/pytest/test_hipfile.py and RocPD validation rules, registered as the ctest hipfile-telemetry-sys-run; validates the counters in both RocPD and Perfetto (ctest -R hipfile).
- **Docs + changelog**: new how-to `docs/how-to/hipfile-telemetry.rs`t (wired into both TOCs) and a CHANGELOG.md entry under the unreleased 1.8.0 section.

## JIRA ID

Jira ID: 

## Test Result

image

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-systems/CHANGELOG.md
+++ b/projects/rocprofiler-systems/CHANGELOG.md
@@ -48,6 +48,21 @@ Full documentation for ROCm Systems Profiler is available at [https://rocm.docs.
   package installations automatically include this support. A `rocshmem` example
   demonstrating two-PE usage of all nine APIs is included under `examples/rocshmem`.
 
+### Added
+
+- hipFile GPU-direct storage I/O telemetry. A background sampler queries
+  hipFile's in-process statistics API and reports per-GPU I/O counters (read/write
+  bytes, total/fastpath/fallback/unaligned operation counts, errors, and
+  read/write bandwidth) in both Perfetto and RocPD outputs. The counters are
+  cumulative and the bandwidths are normalized to wall-clock time over the
+  sampling interval, matching the conventions of the existing AMD SMI PCIe, XGMI,
+  and AI NIC metrics. Select metrics with `ROCPROFSYS_HIPFILE_METRICS` and GPUs
+  with `ROCPROFSYS_SAMPLING_GPUS`. Build with `-DROCPROFSYS_USE_HIPFILE=ON`
+  (optional; disabled automatically when hipFile is not found) and enable at run
+  time with `ROCPROFSYS_USE_HIPFILE=ON` (requires a target application that uses
+  hipFile). See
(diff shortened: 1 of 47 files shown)


sputhala-amd:
Superseded by https://github.com/ROCm/rocm-systems/pull/10630, hence closing this PR.

PR #8533, opened 2026-07-14
[rocr][hotswap] Add opt-in HSA_HOTSWAP_PROFILE rewrite profiler

rahjain-amd:
### What this is
Introduces a profiler for the gfx1250 B0→A0 HotSwap rewrite. Every code object COMGR rewrites at load time goes through a pipeline (parse → decode → apply patch strategies → grow ELF), and until now that cost was a black box. This makes it measurable, so we can see where time goes and drive timing/perf optimization.

### How it works
Lightweight scoped timers around each pipeline stage and each patch strategy feed a single process-wide accumulator. Results aggregate across all rewritten code objects and print once to stderr at process exit. It reports three families:

- `phase:*` — pipeline stages (elf_parse, initLLVM, decode, b0a0_dispatch, grow_elf…), summing to `phase:rewrite_total`.
- `strat:*` — per patch strategy (s_clause, trampoline, wmma_*, scratch, vop3px2), with trampoline sub-rules nested (`ds_2addr`, `tensor_tdm`, `addtid`).
- `jump:*` — trampoline placement outcomes (short / long / declined / nop-sled).

### How to enable
Two gates, so a normal build has zero overhead:

```bash
# 1. compile it in (off by default → no-op stubs)
cmake -DENABLE_HOTSWAP_PROFILE=ON ...

# 2. turn it on at runtime
HOTSWAP_PROFILE=1 ./your_workload
```

When the macro is off the profiler is compiled out entirely; when on but the env var is unset it stays dormant.

### Sample output
```text
=== HotSwap COMGR rewrite profile (HOTSWAP_PROFILE) ===
name                       calls   total_us    avg_us  patches
phase:rewrite_total           11   830439.8  75494.5        0
phase:decode                  11   375898.2  34172.5        0
phase:b0a0_dispatch           11   137352.0  12486.5        0
strat:trampoline              11    83253.6   7568.5      456
  ds_2addr                   456    79747.6    174.8      456
strat:inplace_s_clause        11     3561.3    323.7      581
jump:short_s_branch          ...        -         -        -
======================================================
```

### Why it's useful
One run tells you which phase dominates and which rewrite strategy is expensive — the starting point for any HotSwap load-time optimization. It ships safely: off by default, no cost unless explicitly built and enabled.

### Test
- Builds clean with the macro **off** (no-op stubs; profiler absent from the binary) and **on**.
- A real gfx1250 B0→A0 run prints the tables above at exit.
- No linter errors.

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
@@ -45,6 +45,7 @@
 #include <algorithm>
 #include <atomic>
 #include <cctype>
+#include <chrono>
 #include <cstdio>
 #include <cstdint>
 #include <mutex>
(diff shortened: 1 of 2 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -124,6 +125,119 @@ bool IsVerboseLoggingEnabled() {
...
+
+  HotswapProfiler() : enabled_(IsHotswapProfilingEnabled()) {}
+
+  void Dump() {
+    if (!enabled_ || order_.empty()) return;
+    fprintf(stderr, "\n=== HotSwap ROCR rewrite profile (HSA_HOTSWAP_PROFILE) ===\n");
+    fprintf(stderr, "%-24s %8s %12s %11s %11s %11s\n", "name", "calls", "total_us",
+            "avg_us", "min_us", "max_us");
+    for (const std::string& name : order_) {
+      const Entry& e = entries_[name];
```
cfreeamd, line 189: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `entries_[name]` uses `operator[]`, which inserts a default `Entry{}` if the key is absent. Since every key in `order_` is inserted into `entries_` at the same time in `Record()`, this is harmless in practice. `entries_.at(name)` would make the invariant explicit and surface any future logic error as a thrown exception rather than silently producing zero stats.

cfreeamd (commented):

`projects/rocr-runtime/runtime/hsa-runtime/CMakeLists.txt` and `projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp` look correct — cmake option is properly opt-in (OFF by default), no-op stubs compile away all overhead when disabled, mutex coverage in `Record()` is adequate for the profiling use case.

PR #8534, opened 2026-07-14
fix(clr): drain ROCr async thread in hip::Device::~Device() to prevent ASAN mempool UAF

anugodavar:
## Summary

Fixes ASAN use-after-free failures in hipGraphMemPool tests.

### Root cause

Graph completion callbacks run on the ROCr async-events thread and can cascade into `GraphMemAllocNode -> MemoryPool/Heap::DecrementRefCount`. At process exit, `hip::Device` is released (via `RuntimeTearDown`'s external-object loop) *before* `amd::roc::Device` is destroyed — so the drain in `roc::Device::~Device()` runs too late to protect `graph_mem_pool_`.

### JIRA ID
ROCM-27628

### Fix

Add `WaitForHsaAsyncHandlersIdle()` at the top of `hip::Device::~Device()` to guarantee the ROCr async thread is idle before any HIP-layer pools are freed.

The drain in `roc::Device::~Device()` is intentionally kept as an independent guard for backend-level resources and OpenCL builds (`CLR_BUILD_OCL`).

## Related PRs
- #8254
- #8400


projects/clr/hipamd/src/hip_device.cpp
```diff
@@ -399,6 +399,19 @@ const ResourceMeta* Device::lookupResource(uint32_t resId) {
...
+  // node destruction that touches this device's memory pools (e.g.
+  // GraphMemAllocNode -> MemoryPool/Heap::DecrementRefCount). At process exit the
+  // hip::Device is released (via RuntimeTearDown's external-object loop) *before*
+  // the backend amd::roc::Device is destroyed, so the drain that lives in
+  // amd::roc::Device::~Device() runs too late to protect graph_mem_pool_ below.
+  // Draining here guarantees the async thread is idle before we free those pools.
+  const auto& backendDevices = devices();
+  if (!backendDevices.empty() && backendDevices[0] != nullptr) {
+    backendDevices[0]->WaitForHsaAsyncHandlersIdle();
+  }
```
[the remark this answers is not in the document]
anugodavar: it is only a single device .

projects/clr/hipamd/src/hip_context.cpp
```diff
@@ -63,6 +63,18 @@ void init(bool* status) {
     amd::RuntimeTearDown::RegisterObject(device);
   }
 
+  // Drain the ROCr async thread before any hip::Device is released so that
+  // in-flight graph completion callbacks don't race hip::Device memory pools.
+  amd::RuntimeTearDown::RegisterTearDownCallback(
```
jaydeeppatel1111: The fix looks correct for the reported cause — draining before hip::Device frees graph_mem_pool_ (and the other HIP-layer pools) closes the ASAN UAF, and doing it for all devices is right. This is also better than the first commit that drained inside hip::Device::~Device().

Regarding the placement of the drain (not a blocker): since it's a global, unconditional precondition for teardown — not subsystem-specific cleanup like the profiler or fatbin cleanup — would it be possible to do it as the first step of ~RuntimeTearDown(), ahead of the callback loop, instead of as a teardown callback? Unless there's a blocker I'm missing.

With the callback approach, it has to always be the last-registered callback in order to run first (reverse-order execution), so any future callback must be registered with that constraint in mind.

jaydeeppatel1111: It's also the same all-device WaitForHsaAsyncHandlersIdle the profiler already runs in ProfilerAtExit, so duplicates if profiler enabled.

anugodavar: Enabled DrainAllDevices irrespective of profiler enabled. So DrainAllDevices should make sure no pending callbacks before device is destroyed .

projects/clr/hipamd/src/profiler/hip_clr_profiler.cpp
```diff
@@ -2388,10 +2388,10 @@ static std::string AddPidToPath(const std::string& path) {
 // Drains all queues when profiling was active; writes trace if GPU_CLR_PROFILE_OUTPUT is set.
 static void ProfilerAtExit() {
 #if !defined(_WIN32)
-  // Drain in-flight GPU work whenever profiling was active (env var or API),
-  // so all ReportActivity callbacks arrive before we serialise records.
+  // Drain all in-flight GPU work and async handlers unconditionally so that
+  // all ReportActivity callbacks complete before we serialise records.
   // Skipped on Windows where KFD streams may already be partially torn down.
-  if (IsProfilingActive()) DrainAllDevices();
+  DrainAllDevices();
```
jaydeeppatel1111, line 2394: I believe, This is HIP only path..right? OCL does not run this.

projects/clr/rocclr/device/rocm/rocdevice.cpp
```diff
@@ -199,8 +199,6 @@ void Device::checkAtomicSupport() {
 }
 
 Device::~Device() {
-  WaitForHsaAsyncHandlersIdle();
```
jaydeeppatel1111, line 202: I believe, This runs on both HIP and OCL so if async handler drain is required for OCL, It may break by removing it??

saleelk (approved):

projects/clr/rocclr/device/rocm/rocdevice.cpp
anugodavar: done.

anugodavar:
PR fixed the Asan failures. unittest is not required.

PR #8540, opened 2026-07-14
Add IFH null-submit and gfx1250 ASIC-revision masquerade for hardware-free hotswap profiling

rahjain-amd:
## What is it
Two opt-in, env-gated runtime knobs that let you exercise and profile the gfx1250 hotswap
code-object rewrite paths (including the B0 entry-trampoline path) on A0 hardware, with no B0
silicon:

- **IFH (In-Function Hardware) null-submit** — instead of ringing the GPU doorbell, the host
  completes each AQL/SDMA packet's `completion_signal` itself (generic offset-56 read, so it also
  covers the PM4 IB packet used during queue setup) and advances the read index. Code objects load,
  parse, and rewrite to completion without any GPU execution.
- **ASIC masquerade** — overrides the reported gfx1250 stepping / gfx target so the hotswap decision
  selects the desired rewrite path (A0 b0a0-downgrade vs B0 entry-trampoline).

Both are off by default: no behavior change unless explicitly enabled.

## How to build
Standard rocr-runtime build; produces `libhsa-runtime64.so`:

```bash
cmake -G Ninja -S projects/rocr-runtime -B build \
  -DCMAKE_BUILD_TYPE=Release \
  -DCMAKE_PREFIX_PATH="<rocm-llvm-dist>/lib/llvm;<rocm-llvm-dist>" \
  -DBUILD_SHARED_LIBS=ON
ninja -C build
```

Then point the app at the built library (overlay via `LD_LIBRARY_PATH`, ahead of the system ROCr):

```bash
export LD_LIBRARY_PATH=<build>/lib:$LD_LIBRARY_PATH
```

## How to use
Enable via environment variables:

| Env var | Values | Purpose |
|---|---|---|
| `HSA_ENABLE_IFH` | `1` | Complete fences on the host; never submit to the GPU |
| `HSA_HOTSWAP_FORCE_ASIC_REVISION` | `0`=A0, `1`=B0 | Fake the reported stepping (selects the rewrite path) |
| `HSA_HOTSWAP_FORCE_GFX` | e.g. `gfx1250` | Force the reported gfx target |

Example — profile the B0 entry-trampoline rewrite on an A0 box:

```bash
HSA_ENABLE_IFH=1 \
HSA_HOTSWAP_FORCE_ASIC_REVISION=1 HSA_HOTSWAP_FORCE_GFX=gfx1250 \
AMD_COMGR_HOTSWAP_ENTRY_TRAMPOLINES=1 \
<your workload>
```

**Notes:** IFH runs no kernels, so numeric results are invalid (correctness tests fail as expected)
and workloads that spin on device-computed data will hang — this is for load/parse/rewrite
profiling, not execution. Masquerade only affects gfx1250.

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
@@ -479,7 +479,57 @@ uint64_t AqlQueue::AddWriteIndexRelease(uint64_t value) {
                      std::memory_order_release);
 }
 
+void AqlQueue::CompleteSubmittedPacketsIFH() {
+  core::AqlPacket* ring =
+      reinterpret_cast<core::AqlPacket*>(amd_queue_.hsa_queue.base_address);
+  const uint32_t size = amd_queue_.hsa_queue.size;
+  if (ring == nullptr || size == 0) {
+    return;
+  }
+  const uint32_t mask = size - 1;
+
+  const uint64_t write = atomic::Load(&amd_queue_.write_dispatch_id, std::memory_order_acquire);
+  uint64_t read = atomic::Load(&amd_queue_.read_dispatch_id, std::memory_order_relaxed);
+
+  for (; read < write; ++read) {
+    const core::AqlPacket& pkt = ring[read & mask];
+    const uint16_t header = pkt.packet.header;
+    if (!core::AqlPacket::IsValid(header)) {
+      continue;
+    }
+
+    // Every 64-byte AQL packet the runtime places on the ring carries its
+    // completion_signal in the last 8 bytes (offset 56): kernel dispatch,
+    // barrier-and/or, agent dispatch, ext kernel dispatch, and AMD
+    // vendor-specific packets such as the PM4 IB packet emitted by
+    // ExecutePM4() during queue/scratch setup. Read it generically -- missing
+    // the PM4 IB packet leaves ExecutePM4's ACTIVE wait spinning forever, which
+    // hangs queue initialization under IFH. dispatch.completion_signal aliases
+    // that offset for all of these packet layouts.
+    const hsa_signal_t completion = pkt.dispatch.completion_signal;
+
+    if (completion.handle != 0) {
+      core::Signal* signal = core::Signal::Convert(completion);
+      if (signal != nullptr) {
+        signal->SubRelease(1);
+      }
+    }
+  }
+
+  // Mark every submitted packet as consumed so the queue looks drained, exactly
+  // as the GPU would advance the read index after executing them.
+  atomic::Store(&amd_queue_.read_dispatch_id, write, std::memory_order_release);
+}
+
 void AqlQueue::StoreRelaxed(hsa_signal_value_t value) {
+  if (core::Runtime::runtime_singleton_->flag().enable_ifh()) {
+    // IFH null-submit: skip the doorbell entirely and complete the just-enqueued
+    // packets on the host so waiters unblock without any GPU execution. This is
+    // a host-side profiling aid, not a correctness path.
+    CompleteSubmittedPacketsIFH();
+    return;
+  }
+
   if (core::Runtime::runtime_singleton_->thunkLoader()->IsDTIF() ||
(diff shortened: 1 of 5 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
```diff
@@ -479,7 +479,57 @@ uint64_t AqlQueue::AddWriteIndexRelease(uint64_t value) {
...
+      reinterpret_cast<core::AqlPacket*>(amd_queue_.hsa_queue.base_address);
+  const uint32_t size = amd_queue_.hsa_queue.size;
+  if (ring == nullptr || size == 0) {
+    return;
+  }
+  const uint32_t mask = size - 1;
+
+  const uint64_t write = atomic::Load(&amd_queue_.write_dispatch_id, std::memory_order_acquire);
+  uint64_t read = atomic::Load(&amd_queue_.read_dispatch_id, std::memory_order_relaxed);
+
```
cfreeamd, line 493: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** Race in concurrent IFH dispatch: `read_dispatch_id` is loaded with `relaxed` and then stored unconditionally — no CAS or lock serializes concurrent callers. If two threads each call `AddWriteIndexAcqRel` (advancing `write_dispatch_id` to 1 and 2), then both call `StoreRelaxed` → `CompleteSubmittedPacketsIFH()` concurrently, both can load `read=0` before either stores the updated value. Both threads then iterate the overlapping range and call `SubRelease(1)` on the same completion signals twice. The double-decrement underflows the signal value, leaving any waiter polling for value `== 0` spinning indefinitely.

For a profiling-only path a `std::mutex` in `CompleteSubmittedPacketsIFH()` or a comment scoping use to single-threaded dispatch would suffice.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -124,6 +125,119 @@ bool IsVerboseLoggingEnabled() {
...
+    fprintf(stderr, "%-24s %8s %12s %11s %11s %11s\n", "name", "calls", "total_us",
+            "avg_us", "min_us", "max_us");
+    for (const std::string& name : order_) {
+      const Entry& e = entries_[name];
+      const double total_us = e.total_nanos / 1000.0;
+      const double avg_us = e.calls ? total_us / e.calls : 0.0;
+      const double min_us = e.min_nanos == UINT64_MAX ? 0.0 : e.min_nanos / 1000.0;
+      const double max_us = e.max_nanos / 1000.0;
+      fprintf(stderr, "%-24s %8llu %12.1f %11.3f %11.3f %11.3f\n", name.c_str(),
+              static_cast<unsigned long long>(e.calls), total_us, avg_us, min_us, max_us);
```

**[Nit]** `fprintf` uses `%llu` with `static_cast<unsigned long long>` for `uint64_t`. On Linux x86-64, `uint64_t` is `unsigned long`, not `unsigned long long`, so the cast avoids UB. Consider `PRIu64` from `<cinttypes>` — removes the cast and is portable across all targets.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap_gfx_query.cpp
```diff
@@ -45,14 +45,48 @@
...
 #include <unordered_map>
 
 #include "core/inc/hsa_internal.h"
+#include "core/util/os.h"
 
 namespace rocr {
 namespace hotswap {
 
+namespace {
+
```
cfreeamd, line 60: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `ApplyMasqueradeOverrides` calls `os::IsEnvVarSet(...)` then `os::GetEnvVar(...)` for the same key — two separate env-var lookups. A single `GetEnvVar` call plus an emptiness check achieves the same result with one lookup.

cfreeamd (commented):
🤖 AI-assisted pre-review (for @cfreeamd) — IFH null-submit paths (AQL queue, SDMA) and hotswap profiling instrumentation look structurally sound. Flag.h integration is clean. One concurrency concern on the IFH doorbell path flagged inline.

PR #8544, opened 2026-07-14
Revert "fix(rocprofiler-systems): make MPI auto-detection not CWD-dependent"

amd-chiranjeevi:
Reverts ROCm/rocm-systems#8415


kcossett-amd:
@amd-chiranjeevi 
I just checked the CI for the original PR, and I noticed `roctx-runtime-instrument` is hardcoded to not run. Seems to be using the command from the (what I though was deprecated) `test_rocprofiler_systems.py` script in TheRock.

https://github.com/ROCm/rocm-systems/actions/runs/29250953808/job/86846251057?pr=8415#step:11:138
So I don't think this PR will be able to check if it works or not.

PR #8546, opened 2026-07-14
fix(rocprofiler-systems): prevent trace-cache mutex self-deadlock

mradosav-amd:
## Motivation

pthread mutex/rwlock gotcha tracing intercepts every lock/unlock call, but rocprofsys's own internal locking (e.g. the trace cache's `buffer_storage` mutex) is also a pthread mutex. Acquiring that internal lock therefore triggers a trace event that itself tries to acquire the same lock, self-relocking a non-recursive `std::mutex` and deadlocking `rocprof-sys-run` whenever `-I mutex-locks` is enabled.

## Technical Details

The codebase already has a mechanism that fixes this issue: `ThreadState::Internal`, set via `ROCPROFSYS_SCOPED_THREAD_STATE`, is checked by `pthread_mutex_gotcha_policy.hpp`'s `is_disabled_check()` (`get_thread_state() != ThreadState::Enabled`) and is already used throughout the project to mark internal work as not-to-be-traced.  `core/trace_cache/buffer_storage.hpp`: `store()` and `flush()` now wrap their `m_mutex` critical sections in `ROCPROFSYS_SCOPED_THREAD_STATE(ThreadState::Internal)`, so lock/unlock activity on the trace cache's own mutex is invisible to pthread mutex/rwlock tracing while it's held.

## Test Plan

- [x] Added `test_mutex_locks` to `tests/pytest/test_transpose.py`: runs the `transpose` example under `rocprof-sys-run -I mutex-locks` and asserts it completes instead of hanging, as a regression test for the self-deadlock.
- [x] Test case from the JIRA ticket no longer hangs.

---
JIRA ID : AIPROFSYST-574


projects/rocprofiler-systems/source/lib/core/trace_cache/buffer_storage.hpp
```diff
@@ -244,7 +245,7 @@ class buffer_storage
 
     std::shared_ptr<typename WorkerFactory::worker_t> m_worker;
 
-    std::mutex                      m_mutex;
+    self_suppressing_mutex          m_mutex;
```
sputhala-amd: This could also be fixed by guarding the `m_mutex` critical sections in `store()`/`flush()` with `ROCPROFSYS_SCOPED_THREAD_STATE(ThreadState::Internal)` instead of introducing a new mutex type. The macro pushes the thread to `Internal` for the scope, and since `is_disabled_check()` returns true for any `state != Enabled`, the gotcha skips tracing the lock. This reuses the existing idiom already used throughout `sampling.cpp`, `perf.cpp`, `causal/*`, etc.

sputhala-amd (commented):
The re-entrancy/nesting is already handled with `ThreadState`, so may be we can try to re-use that mechanism.

projects/rocprofiler-systems/source/lib/common/trace_suppression.hpp
```diff
@@ -0,0 +1,28 @@
...
+{
+/// Tracks, per-thread, whether the calling thread is currently inside code
+/// whose pthread mutex/rwlock usage must not be traced (e.g. rocprofsys's
+/// own internal locking). A depth counter rather than a bool so nested
+/// enter()/exit() pairs on the same thread compose correctly.
+class trace_suppression
+{
+public:
+    static void enter() noexcept { ++s_depth; }
+    static void exit() noexcept { --s_depth; }
```
dgaliffiAMD: We should add a guard against decrementing `s_depth` when it is already `0`.

```suggestion
    static void exit() noexcept
    {
        assert(s_depth > 0 && "trace_suppression::exit() without matching enter()");
        --s_depth;
    }
```

or a clamp: 

```suggestion
    static void exit() noexcept
    {
        assert(s_depth > 0 && "trace_suppression::exit() without matching enter()");
        if(s_depth > 0) --s_depth;
    }
```

And an associated test. 

```
#ifdef NDEBUG
// With asserts compiled out (release builds), an unmatched exit() must clamp
// rather than wrap the depth counter around to a huge value, which would leave
// the thread permanently suppressed.
TEST(TraceSuppressionTest, ExitWithoutEnterClampsAtZero)
{
    EXPECT_FALSE(trace_suppression::is_active());

    trace_suppression::exit();
    EXPECT_FALSE(trace_suppression::is_active());

    trace_suppression::enter();
    EXPECT_TRUE(trace_suppression::is_active());
    trace_suppression::exit();
    EXPECT_FALSE(trace_suppression::is_active());
}
#else
// With asserts enabled (debug builds), an unmatched exit() trips the guard.
TEST(TraceSuppressionDeathTest, ExitWithoutEnterAsserts)
{
    EXPECT_FALSE(trace_suppression::is_active());
    EXPECT_DEATH({ trace_suppression::exit(); }, "");
}
#endif
```

projects/rocprofiler-systems/source/lib/core/trace_cache/buffer_storage.hpp
mradosav-amd: Thanks. I'll switch to that approach. Not introducing new one.

projects/rocprofiler-systems/source/lib/common/trace_suppression.hpp
mradosav-amd: I'll probably abandon this approach (`TraceSuppression` and `SupressionMutex`), in favor of `ROCPROFSYS_SCOPED_THREAD_STATE(ThreadState::Internal)`

projects/rocprofiler-systems/source/lib/core/trace_cache/buffer_storage.hpp
mradosav-amd: Fixed.

projects/rocprofiler-systems/source/lib/common/trace_suppression.hpp
mradosav-amd: This part of code is removed.

marantic-amd (approved):
LGTM.

sputhala-amd (approved):
LGTM. Thanks.

Issue #8340, opened 2026-07-09
[rocr-runtime][rocdxg][WSL] error when linking librocdxg.so: `undefined reference to rocr::log_printf(char const*, int, char const*, ...)'`


dysu-amd:
### Description
All recent ROCR Runtime WSL CI checks are failing to link rocdxg:
https://github.com/ROCm/rocm-systems/actions/workflows/rocr-runtime-wsl.yml

examples:
https://github.com/ROCm/rocm-systems/actions/runs/29034141422/job/86174452532#step:10:405
https://github.com/ROCm/TheRock/actions/runs/29027555678/job/86156984352?pr=6447#step:20:576

### Error message
```
[30/31] Linking CXX shared library librocdxg.so.1.1.0
FAILED: librocdxg.so.1.1.0 
[...]

PR #8550, opened 2026-07-14
feat(ci): Move rocr runtime WSL build and nvidia build into component-ci

litvaOo:
## Motivation
The ROCr runtime WSL Test is not currently a hard requirement for PRs impacting CLR or ROCr runtime projects. This allows for WSL path breaking code to [be merged into develop](https://github.com/ROCm/rocm-systems/issues/8340). With inclusion of the WSL ROCDXG in TheRock, we want to make it required

## Technical Details
Adds a new component-ci pipeline, which will include current path-based workflows (ROCr runtime WSL Test and HIP NVidia build) as its internal steps.
After merging, component-CI will replace NVidia pipeline as a required one

## JIRA ID
NA

## Test Plan
HIP Nvidia and ROCr Runtime WSL Test should run as steps within a single Component CI job and pass as expected.

## Test Result
Waiting for PSDB

## Submission Checklist

### DIFF
--- a/.github/scripts/configure_component_ci.py
+++ b/.github/scripts/configure_component_ci.py
@@ -1,0 +1,89 @@
+"""
+Determines which component CI jobs to run based on the GitHub event type
+and the files changed in the pull request or push.
+"""
+
+import logging
+import os
+import re
+import subprocess
+from typing import Mapping
+
+logging.basicConfig(level=logging.INFO)
+
+
+def set_github_output(d: Mapping[str, str]):
+    """Sets GITHUB_OUTPUT values.
+    See https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/passing-information-between-jobs
+    """
+    logging.info(f"Setting github output:\n{d}")
+    step_output_file = os.environ.get("GITHUB_OUTPUT", "")
+    if not step_output_file:
+        logging.warning(
+            "Warning: GITHUB_OUTPUT env var not set, can't set github outputs"
+        )
+        return
+    with open(step_output_file, "a") as f:
+        f.writelines(f"{k}={v}" + "\n" for k, v in d.items())
+
+
+def get_changed_files(base_ref: str) -> list[str]:
+    result = subprocess.run(
+        ["git", "diff", "--name-only", base_ref],
+        capture_output=True,
+        text=True,
+        check=True,
+    )
+    return result.stdout.splitlines()
+
+
+def matches_any(files: list[str], pattern: str) -> bool:
+    regex = re.compile(pattern)
+    return any(regex.search(f) for f in files)
+
+
+def main():
+    event_name = os.environ.get("GITHUB_EVENT_NAME", "")
+    base_ref = os.environ.get("BASE_REF", "HEAD^")
+
+    nvidia = False
+    wsl = False
+    hip_contract_tests = False
+
+    if event_name == "workflow_dispatch":
+        nvidia = True
+        wsl = True
+        hip_contract_tests = True
+    else:
+        changed_files = get_changed_files(base_ref)
+        logging.info("Changed files:\n" + "\n".join(changed_files))
+
+        if matches_any(
+            changed_files,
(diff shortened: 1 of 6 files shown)


.github/workflows/component-ci.yml
```diff
@@ -0,0 +1,105 @@
+name: Component CI
```
geomin12, line 1: can we specify which component? this naming is pretty generic

.github/workflows/component-ci.yml
```diff
@@ -0,0 +1,105 @@
...
+
+on:
+  workflow_dispatch:
+  pull_request:
+    types:
```
geomin12, line 11: should we only trigger for specific paths? does this need to run on every PR on all changes?

.github/workflows/component-ci.yml
```diff
@@ -0,0 +1,105 @@
...
+              WSL=true
+            fi
+
+            if echo "${CHANGED_FILES}" | grep -qE '^projects/hip/include/hip/|^projects/hip-tests/catch/(contract|tools)/|^projects/hip-tests/catch/config/configs/contract\.yaml|^projects/hip-tests/catch/TEST_PLAN\.md|^\.github/workflows/(component-ci|hip-contract-coverage)\.yml'; then
+              HIP_CONTRACT_TESTS=true
+            fi
+          fi
+
+          echo "nvidia=${NVIDIA}" >> "${GITHUB_OUTPUT}"
+          echo "wsl=${WSL}" >> "${GITHUB_OUTPUT}"
```
geomin12: perhaps a python file to handle this? will be cleaner / easier to read / easier to unit test

.github/workflows/component-ci.yml
```diff
@@ -0,0 +1,105 @@
...
+          echo "hip-contract-tests=${HIP_CONTRACT_TESTS}" >> "${GITHUB_OUTPUT}"
+
+  nvidia:
```
geomin12: can we add a better job name

.github/workflows/component-ci.yml
litvaOo: This is a generic workflow for different unrelated component checks we would like to have as required. We want to use this workflow as a single required entry-point, which will then select jobs based on changed paths.

.github/workflows/component-ci.yml
litvaOo: Yes, we need it on every PR, because that's the only way to mark job as a required. The job will skip checks for unrelated paths and go straight to the successful summary.

.github/scripts/configure_component_ci.py
```diff
@@ -0,0 +1,89 @@
...
+
+def main():
+    event_name = os.environ.get("GITHUB_EVENT_NAME", "")
+    base_ref = os.environ.get("BASE_REF", "HEAD^")
+
+    nvidia = False
+    wsl = False
+    hip_contract_tests = False
+
+    if event_name == "workflow_dispatch":
```
geomin12, line 84: hm. what if for workflow_dispatch, we can specify which tests to run? seems expensive. what if i just wanted wsl?

.github/scripts/configure_component_ci.py
```diff
@@ -0,0 +1,89 @@
...
+
+        if matches_any(
+            changed_files,
+            r"^projects/(rocr-runtime|clr|hip|hip-tests)/|^\.github/workflows/(component-ci|rocr-runtime-wsl)\.yml|^shared/amdgpu-windows-interop/wkmi",
+        ):
+            wsl = True
+
+        if matches_any(
+            changed_files,
+            r"^projects/hip/include/hip/|^projects/hip-tests/catch/(contract|tools)/|^projects/hip-tests/catch/config/configs/contract\.yaml|^projects/hip-tests/catch/TEST_PLAN\.md|^\.github/workflows/(component-ci|hip-contract-coverage)\.yml",
```
geomin12: is it possible to re-work this so it's multi-line? so it's both easier to read and write to?

.github/scripts/configure_component_ci.py
geomin12, line 1: perhaps a few unit tests to test this? what if someone adds a regex but it's a bad regex. seems there is little validation

PR #8552, opened 2026-07-14
refactor(rocprofiler-compute): Refactoring roctx recordfn module

ggottipa-amd:
## Motivation

Rename roctx_recordfn to torch_trace_collector. Split it into focused units (wire format, snapshot store, marker stack, process/install state), with RecordFunction callbacks in the collector .cpp and a separate pybind module.

## Technical Details

**Pre-existing**

- RecordFunction callbacks emit ROCTX ranges on every thread.
- Snapshot store joins backward to forward by `seqNr`. TLS overlay seeds empty autograd workers.
- Loader could compile a helper under `~/.cache/rocprofiler-compute/`.

**Modified in this PR**

- Snapshot key is `(seqNr, thread id)`.
- Prebuilt `torch_trace_collector-<Torch_VERSION>.so` for Torch 2.13 & 2.14. Loader matches workload `torch.__version__`. No cache compile.
- Design notes: `docs/design/hld-torch-trace-collector.md` and `docs/design/lld-torch-trace-collector.md`.

## Dependencies

Requires ROCm/TheRock#7546 (renames the gtest in the RPATH post-hook) to land first, plus the TheRock ref bump that picks it up. Supersedes and closes #10437.

## JIRA ID

AIPROFCOMP-716

## Test Plan

`ctest -R 'test_profile_torch_trace|test-torch-trace-collector|test_torch_cpp_loader|test_torch_trace_coverage_utils|test_torch_trace_kernel_names' --output-on-failure`

## Test Result

PASS

## Submission Checklist


projects/rocprofiler-compute/src/lib/roctx_recordfn/CMakeLists.txt
```diff
@@ -1,10 +1,11 @@
 ## Copyright (c) Advanced Micro Devices, Inc.
 ## SPDX-License-Identifier:  MIT
 ##
-## Builds roctx_recordfn-<tag>.so for --torch-trace.
-##   A. parent add_subdirectory() under ENABLE_TESTS (CI fixture).
-##   B. standalone (runtime cmake-tier from
-##      utils/inject_roctx/_backends/torch_cpp_loader.py).
+## Builds the roctx-recordfn library and the roctx_recordfn-<tag>.so pybind
```
svolkov-amd: 1. Are there reasons why it's not compiled in CI and installed along with rocprof-compute?
2. Why is so complicated way to search for PyTorch libraries? That should be as simple as:
```cmake
  find_package(Torch REQUIRED)

  add_library(myext SHARED myext.cpp)
  target_link_libraries(myext PRIVATE ${TORCH_LIBRARIES})
  target_include_directories(myext PRIVATE ${TORCH_INCLUDE_DIRS})
```

projects/rocprofiler-compute/src/utils/inject_roctx/_backends/torch_cpp_loader.py
```diff
@@ -36,7 +35,20 @@
 
 C_TIER_NAMES = frozenset((TIER_PREBUILT, TIER_JIT))
 
-_FINGERPRINT_INPUTS = (_SO_SOURCE, _SO_BUILDFILE)
+
```
svolkov-amd, line 44: Why do we add a new way to compile C++ sources while we already have compilation defined in `native_tool_filder.py`. If you make your code a part of native tool root cmake it will be properly compiled.

svolkov-amd (commented):
There are several major architectural decisions with not fully clear intention:
- Why we have a separate C++ code compliation path in torch_cpp_loader.py?
- Why only JIT compilation is used instead of compliation and installation via CI? It looks like AT callbacks don't guarantee ABI and API stability. That means that this code may not guarantee to compile with newer PyTorch versions. Most likely the only way for us is to explicitly define what PyTorch versions we support and build separate PyTorch tracing libraries, while extracting the common part into a separate reusable library.
- Also code heavily uses global variables which is a terrible practice. And manually defines mutexes instead of `utils/syncronized` which would have been also easier if C++ code would have been a part of a single C++ project along the native collector.
- Also it's nice to have some breakdown onto files but it's done in a way where some files contain a small fraction of the code which don't have a well defined responcibility. While some concepts are still clumped together, ex callbacks installation and callback data processing.

svolkov-amd (dismissed):

projects/rocprofiler-compute/src/lib/torch_trace_collector/user_scope.h
```diff
@@ -0,0 +1,158 @@
...
+            wire_string += backend;
+        }
+        roctxRangePushA(wire_string.c_str());
+        auto roctx_rollback = make_scope_guard([] { roctxRangePop(); });
+
+        g_capture.capture(wire_string);
+        inc(g_stats.user_scope_pushes);
+        inc(g_stats.pushes);
+
+        roctx_rollback.dismiss();
```
svolkov-amd: Why these guards are dismissed right after their creation? I don't understand the logic here. This would mean that guards are not doing anything, right?

projects/rocprofiler-compute/docs/how-to/profile/mode.rst
```diff
@@ -953,6 +952,25 @@ these wraps. ``ROCPROFCOMPUTE_ROCTX_DEEP_TENSOR_WRAPS`` is enabled by default.
 
    $ ROCPROFCOMPUTE_ROCTX_DEEP_TENSOR_WRAPS=0 rocprof-compute profile --experimental --torch-trace --name mnist_torch -- python train.py
 
+Native collector
```
vedithal-amd: `Native collector` is confusing here since it conflicts with our native collector tool. I would recommend calling it something else like `Torch trace collector`

vedithal-amd (changes requested):

vedithal-amd:
Pushed build changes directly (29017b4) to save a round trip. What and why:

**Consistency with the other native targets**
- Root `CMakeLists.txt` registers the gtest and installs the extension, as it does for `test-rocprofiler-compute-tool`, `test-pc-sampling-collector` and `test-compression`. The subdirectory just publishes `ROCPROF_TEST_LD_LIBRARY_PATH` and `TORCH_TRACE_COLLECTOR_TARGET`. This is the pattern from ed52b066ee.
- That let `include(GNUInstallDirs)` come back out of `src/lib/CMakeLists.txt`.
- Added both tests to `test_categories.yaml`. The gtest was missing under either name. Left it out of the `sanitizer` lane, which is documented as "no GPU, no workload"; this binary links libtorch and runs real ATen ops.

**Bugfixes**
- `install(DIRECTORY src/lib/)` was unconditional, shipping C++ sources in every package. It honors `SKIP_NATIVE_TOOL_BUILD` again: libraries when OFF, sources when ON.
- The artifact tag had two derivations. CMake used `Torch_VERSION`, the loader used `torch.__version__`. They disagree on dev and nightly wheels, so `--target torch_trace_collector-<loader-tag>` named a target that did not exist. `torch_trace_fingerprint.py` now owns `artifact_tag()`, the loader passes `-DTORCH_TRACE_ARTIFACT_TAG`, and a build-time configure runs the module. Removes the CMake reimplementation added in 3a869bac. Verified locally against a `2.15.0a0` wheel, which is exactly the case that used to mismatch.
- The DebugInfoKind probe set `CMAKE_REQUIRED_FLAGS`, which `try_compile` overrides by appending its own `-std=` last. Now goes through `CMAKE_CXX_STANDARD`. Supersedes #10437, which I am closing.

**Smaller**
- Dropped the pinned `_GLIBCXX_USE_CXX11_ABI=1`. The toolchain default already matches ROCm libtorch, and the tag has no ABI field to distinguish a build that does not.
- Replaced `from __future__ import annotations` with real 3.8 syntax, per `PYTHON_CODING_STYLE.md`.
- `NativeToolFinder` keeps its new kwargs, which the loader needs. Shorter docstring, consistent `_` prefixes.
- Split the fingerprint tests into `test_torch_trace_fingerprint.py` and covered `required_input_paths`. One of the new tests caught a shadowing bug: running the fingerprint module as a script puts `_backends/` on `sys.path`, where `torch.py` hides the real torch package.
- Restored the two LD_PRELOAD rationale comments in `profiler_rocprofiler_sdk.py`.
- Three lines of unrelated gersemi reformatting at the top of `CMakeLists.txt`. Pre-existing on the branch, and the pinned hook refuses to commit the file without it.

Docs updated for all of the above. `ctest` gtest and GPU runs not exercised locally; ruff, gersemi and the unit suites pass.

vedithal-amd (approved):
Do not merge until ROCm/TheRock#7546 lands and the TheRock ref here is bumped to pick it up, otherwise the renamed gtest silently loses its RPATH origin.

projects/rocprofiler-compute/docs/how-to/profile/mode.rst
```diff
@@ -953,6 +952,30 @@ these wraps. ``ROCPROFCOMPUTE_ROCTX_DEEP_TENSOR_WRAPS`` is enabled by default.
 
    $ ROCPROFCOMPUTE_ROCTX_DEEP_TENSOR_WRAPS=0 rocprof-compute profile --experimental --torch-trace --name mnist_torch -- python train.py
 
+Torch trace collector
+---------------------
+
+Torch trace is collected by ``torch_trace_collector``, a C++ extension built for one
```
svolkov-amd: What combinations are supported?

projects/rocprofiler-compute/docs/how-to/profile/mode.rst
```diff
@@ -953,6 +952,30 @@ these wraps. ``ROCPROFCOMPUTE_ROCTX_DEEP_TENSOR_WRAPS`` is enabled by default.
...
+combination of Python and PyTorch versions. The copy installed with ROCm Compute
+Profiler is used when it matches the interpreter that runs the workload.
+
+Otherwise the extension is built on first use, which requires an installation that
```
svolkov-amd: External users should not know or care about it. For external users who don't clone rocprof-compute sources and just install ROCm package, compilation shall never happen.

projects/rocprofiler-compute/src/lib/torch_trace_collector/marker_stack.h
```diff
@@ -0,0 +1,58 @@
...
+    std::size_t              common = 0;
+    for (; common < maxc; ++common)
+    {
+        if (chain[common].marker != stack[common].marker || chain[common].context != stack[common].context)
+        {
+            break;
+        }
+    }
+    std::size_t pushed = 0;
+    for (std::size_t i = common; i < chain.size(); ++i)
```
abchoudh-amd: Could you use std::copy instead of a loop?

projects/rocprofiler-compute/src/lib/torch_trace_collector/marker_stack.h
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// Pushes `chain` onto the thread stack, skipping any leading prefix that is
+// already present. Returns the number of frames pushed.
+inline std::size_t push_with_prefix_dedup(const std::vector<StackEntry>& chain)
+{
+    std::vector<StackEntry>& stack  = thread_state().stack;
+    const std::size_t        maxc   = std::min(chain.size(), stack.size());
+    std::size_t              common = 0;
+    for (; common < maxc; ++common)
```
abchoudh-amd: Rewriting this with inverse of `if` statements condition in a `while` loop would be more readbale.

projects/rocprofiler-compute/src/lib/torch_trace_collector/record_function_callback.h
```diff
@@ -0,0 +1,168 @@
...
+// Records what start_cb pushed so end_cb can unwind it.
+struct RoctxObserverContext : public at::ObserverContext
+{
+    bool        pushed_roctx_range     = false;
+    bool        pushed_leaf            = false;
+    std::size_t pushed_snapshot_frames = 0;
+};
+
+// The RecordFunction tier instruments PyTorch ATen operators.
+inline constexpr const char* kRecordFnBackend = "torch";
```
abchoudh-amd: Are the `inline`'s intended to bypass the One Definition Rule? If not, could you remove them?

projects/rocprofiler-compute/src/lib/torch_trace_collector/record_function_callback.h
```diff
@@ -0,0 +1,168 @@
...
+        stack.pop_back();
+    }
+    for (std::size_t i = 0; i < observer_ctx.pushed_snapshot_frames && !stack.empty(); ++i)
+    {
+        stack.pop_back();
+    }
+}
+
+inline std::unique_ptr<at::ObserverContext> start_cb(const at::RecordFunction& record_fn)
+{
```
abchoudh-amd: Could you split this function?

projects/rocprofiler-compute/src/lib/torch_trace_collector/scope_guard.h
```diff
@@ -0,0 +1,54 @@
...
+    ~ScopeGuard()
+    {
+        if (active_)
+        {
+            try
+            {
+                fn_();
+            }
+            catch (...)
+            {
```
abchoudh-amd: The action's error are being silently ignored. Could you add a log message?

projects/rocprofiler-compute/src/lib/torch_trace_collector/snapshot_store.h
```diff
@@ -0,0 +1,201 @@
...
+
+    bool operator==(const SnapshotKey& other) const noexcept
+    {
+        return seq_nr == other.seq_nr && thread_id == other.thread_id;
+    }
+};
+
+// Sequence numbers are dense and thread ids are small, so both fields are mixed
+// to spread keys across buckets and shards.
+inline std::size_t hash_snapshot_key(const SnapshotKey& key) noexcept
```
abchoudh-amd: Could you specialize std::hash and use bit mixing; instead of writing custom hash function?

ex:
```
namespace std {
    template <>
    struct hash<SnapshotKey> {
        std::size_t operator()(const SnapshotKey& u) const {
            // Compute individual hash values
            std::size_t h1 = std::hash<std::int64_t>{}(u.seq_nr);
            std::size_t h2 = std::hash<std::uint64_t>{}(u.thread_id);

            // Combine hashes using a classic bit-mixing formula
            return h1 ^ (h2 << 1); 
        }
    };
}
```

projects/rocprofiler-compute/src/lib/torch_trace_collector/snapshot_store.h
```diff
@@ -0,0 +1,201 @@
...
+
+using rocprofiler_compute_tool::common::synchronized_t;
+
+// Sharded store mapping an autograd node to a forward-stack snapshot, with
+// per-shard LRU eviction.
+class SnapshotStore
+{
+public:
+    static constexpr std::size_t kNumShards    = 64;
+    static constexpr std::size_t kShardSoftCap = 10000;
```
abchoudh-amd: Could you disambiguate the variable naming? It seems like there are 64 shards, which is a constant, but there is again a soft cap on number of shards at 10000.

projects/rocprofiler-compute/src/lib/torch_trace_collector/snapshot_store.h
```diff
@@ -0,0 +1,201 @@
...
+
+    explicit SnapshotStore(Stats& stats)
+        : stats_(stats)
+    {
+    }
+
+    static std::size_t shard_index(const SnapshotKey& key) noexcept
```
abchoudh-amd: There is a high chance for index collision. Could you rewrite the index calculation?

projects/rocprofiler-compute/src/lib/torch_trace_collector/snapshot_store.h
```diff
@@ -0,0 +1,201 @@
...
+
+namespace torch_trace_collector::detail
+{
+
+using rocprofiler_compute_tool::common::synchronized_t;
+
+// Sharded store mapping an autograd node to a forward-stack snapshot, with
+// per-shard LRU eviction.
+class SnapshotStore
+{
```
abchoudh-amd, line 65: Most of the work done inside SnapshotStore's functions should be delegated to Shard.

projects/rocprofiler-compute/src/lib/torch_trace_collector/stats.h
```diff
@@ -0,0 +1,32 @@
...
+#pragma once
+
+#include <atomic>
+#include <cstdint>
+
+namespace torch_trace_collector::detail
+{
+
+// Runtime counters exposed through dump_stats().
+struct Stats
```
abchoudh-amd, line 13: Decompose the stats into multiple specialized stats structs.

abchoudh-amd (changes requested):
1. Most of the C++ code resides in the header files. Could you move them to the concerned cpp files?
2. There is a generous use of `inline`, which don't look like are to bypass ODR. Could you leave the optimization to the compiler?

vedithal-amd:
> Do not merge until [ROCm/TheRock#7546](https://github.com/ROCm/TheRock/pull/7546) lands and the TheRock ref here is bumped to pick it up, otherwise the renamed gtest silently loses its RPATH origin.

@ggottipa-amd TheRock side PR has been merged to accomodate the rename of gtest this PR is doing https://github.com/ROCm/rocm-systems/pull/10537

Please incorporate changes from latest rocprofiler-compute-develop branch to get the CI passing here again

projects/rocprofiler-compute/src/lib/torch_trace_collector/snapshot_store.h
```diff
@@ -0,0 +1,102 @@
...
+public:
+    static constexpr std::size_t kNumShards          = 64;
+    static constexpr std::size_t kMaxEntriesPerShard = 10000;
```
abchoudh-amd: This is a property of the shard not the store. Could you move it to shard?

projects/rocprofiler-compute/src/lib/torch_trace_collector/snapshot_store.h
```diff
@@ -0,0 +1,102 @@
...
+
+private:
+    struct Shard
+    {
+        std::unordered_map<SnapshotKey, std::vector<StackEntry>>          snapshots;
+        std::list<SnapshotKey>                                            lru_order;
+        std::unordered_map<SnapshotKey, std::list<SnapshotKey>::iterator> lru_idx;
+    };
+
+    synchronized_t<Shard>& shard_for(const SnapshotKey& key);
```
abchoudh-amd, line 94: Could you make the LRU a standalone template class, so that it can be reused?

projects/rocprofiler-compute/src/lib/torch_trace_collector/torch_trace_collector.cpp
```diff
@@ -0,0 +1,387 @@
...
+        std::string wire_string = build_marker_string(stack);
+        wire_string += '|';
+        wire_string += kRecordFnBackend;
+        roctxRangePushA(wire_string.c_str());
+        observer_ctx->pushed_roctx_range = true;
+        state.stats.pushes.fetch_add(1, std::memory_order_relaxed);
+
+        return observer_ctx;
+    }
+    catch (...)
```
abchoudh-amd, line 228: Broad exception catching and handling only a specific case could hide bugs and make debugging difficult. Could you catch specific exceptions or rethrow the ones not handled in this block?

projects/rocprofiler-compute/src/lib/torch_trace_collector/torch_trace_collector.cpp
```diff
@@ -0,0 +1,387 @@
...
+namespace torch_trace_collector::detail
+{
+
+ProcessState& process_state()
+{
+    static ProcessState state;
+    return state;
+}
+
+ThreadState& thread_state()
```
abchoudh-amd: Could you convert this to a thread local singleton?

projects/rocprofiler-compute/src/lib/torch_trace_collector/snapshot_store.h
ggottipa-amd: Moved.

projects/rocprofiler-compute/src/lib/torch_trace_collector/torch_trace_collector.cpp
ggottipa-amd: catch-all is required so callbacks never throw into libtorch; we can log std::exception::what()

projects/rocprofiler-compute/src/lib/torch_trace_collector/snapshot_store.h
ggottipa-amd: Since there is no second caller for LRU, will leave this as follow-up.

abchoudh-amd (approved):

vedithal-amd (changes requested):
Will review by EOD

vedithal-amd (approved):
LGTM.

Pushed a few review changes directly to the branch, all on top of your work. Summary of what changed and why:

**Collector CMake**
- Removed `BUILD_TORCH_TRACE_COLLECTOR`. Nothing set it, including TheRock, and `AUTO` was the only reachable behavior.
- Replaced `find_package(Torch)` with a read of `TORCH_VERSION_MAJOR`/`MINOR` from the PyTorch headers. `TorchConfig` pulls in Caffe2 and LoadHIP, whose errors abort our configure and which `QUIET` does not suppress. This also drops the cmake 3.25+ requirement that chain imposed.
- Artifacts are now keyed on PyTorch major.minor. `Torch_VERSION` truncates, reporting `2.15.0` for a `2.15.0a0` wheel, so the filename and the loader key could never agree on a prerelease build.
- The filename now carries `Python3_SOABI`. The loader parses it but does not enforce it, so honoring it or rebuilding at runtime is your call.
- roctx via `find_package(rocprofiler-sdk-roctx)` and its imported target instead of hand-rolled `find_library`/`find_path`.
- Fixed target name plus versioned `OUTPUT_NAME`, which removes the `TORCH_TRACE_COLLECTOR_TARGET` cache variable used to pass the name to the parent install rule.
- Explicit `cxx_std_20` on the collector, replacing the C++17 it inherited from `src/lib`. PyTorch 2.15 headers require C++20 and torch's own config sets `CXX_STANDARD` on its imported target, which does not propagate to consumers. This raises the floor unconditionally rather than falling back, so there is also a guard that skips the collector when the compiler lacks C++20, since `target_compile_features` would otherwise fail the whole configure.

**gtest**
- Dropped `--allow-shlib-undefined` and gave the linker the rocm-openblas directory instead. The flag was hiding a real missing search path: `libtorch_cpu` needs rocm-openblas and its RUNPATH points at the wheel build container. Kept `--wrap=roctxRangePushA` with a comment.
- Removed the absolute `INSTALL_RPATH` from the gtest. It listed only the torch and roctx lib dirs, not rocm-openblas, which `libtorch_cpu` also needs, so an installed test always relied on the ctest `LD_LIBRARY_PATH` anyway. That path carries all three.

**Loader**
- Version key comes from `Version(torch.__version__).release[:2]`, with `Version` imported from `torch.torch_version` so it tracks whatever parser torch vendors.
- Split "no collector was built" from "collector exists but not for your version". The first case previously printed an empty supported-versions list. Both are still fatal, as you designed.

**NativeToolFinder**
- Removed the constructor options added for the runtime-compile path: artifact name, build target, build path, configure options, cmake executable, and the search and reuse flags. None had callers once that path was removed.
- Dropped the `CMAKE_BINARY_DIR` search location, which nothing sets for a profiled process.

**Tests and docs**
- Folded the kernel-name helpers into `tests/common.py` and dropped the single-consumer module and its test target.
- Guarded the stats report in `launch.py` so a failure there cannot replace the workload's own exception.
- Supported versions stated once in the profile guide; changelog reduced to the user-facing line. Added a note that PyTorch must come from the TheRock index alongside ROCm, linking their install docs.

Verified locally on gfx942 with a source build installed to a scratch prefix and `TEST_FROM_INSTALL=ON`: the collector gtest, the loader and finder unit tests, `test_profile_torch_trace`, and `test_torch_trace_coverage` all pass.

PR #8556, opened 2026-07-14
AIRUNTIME-2074: bound WinDXG signal waits

victzhan:
## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/interrupt_signal.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/interrupt_signal.cpp
@@ -194,6 +194,12 @@ hsa_signal_value_t InterruptSignal::WaitRelaxed(hsa_signal_condition_t condition
       static_cast<uint32_t>(signal_abort_timeout ? signal_abort_timeout * 1000 : 0xFFFFFFFFUL)
     );
 
+    // WinDXG completion events can miss a GPU wake. Recheck the signal at a bounded interval
+    // instead of waiting for the thunk's six-second safety timeout.
+    if (core::Runtime::runtime_singleton_->thunkLoader()->IsWinDxg()) {
+      wait_ms = std::min<uint32_t>(wait_ms, 1);
+    }
+
     HSAKMT_CALL(hsaKmtWaitOnEvent_Ext(event_, wait_ms, &event_age));
   }
 }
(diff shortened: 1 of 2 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
```diff
@@ -1868,7 +1868,7 @@ void Runtime::AsyncEventsLoop(void* _eventsInfo) {
   // KFD will move this thread into sleep, until any event from the list is complete or
   // if ROCR can wake it up with hsaKmtSetEvent()
   auto WaitForInterrupt = [&]() {
-    constexpr uint32_t wait_ms = 0xFFFFFFFEu;
+    const uint32_t wait_ms = runtime_singleton_->thunkLoader()->IsWinDxg() ? 1 : 0xFFFFFFFEu;
```
cfreeamd, line 1871: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]**

`AsyncEventsLoop` runs for the lifetime of the ROCr runtime. With `wait_ms = 1` on WinDXG, this thread wakes ~1000×/sec even when the system is otherwise idle — converting from interrupt-driven to tight polling. Is the CPU/power overhead acceptable for the WinDXG target platform? The 6-second stall being avoided is clearly more severe; just confirming this tradeoff was considered for power-managed Windows devices.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/interrupt_signal.cpp
```diff
@@ -194,6 +194,12 @@ hsa_signal_value_t InterruptSignal::WaitRelaxed(hsa_signal_condition_t condition
       static_cast<uint32_t>(signal_abort_timeout ? signal_abort_timeout * 1000 : 0xFFFFFFFFUL)
     );
 
+    // WinDXG completion events can miss a GPU wake. Recheck the signal at a bounded interval
+    // instead of waiting for the thunk's six-second safety timeout.
+    if (core::Runtime::runtime_singleton_->thunkLoader()->IsWinDxg()) {
```
cfreeamd, line 199: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]**

`runtime_singleton_->thunkLoader()->IsWinDxg()` traverses two pointer dereferences + a virtual dispatch on each loop iteration. Since `IsWinDxg()` is fixed for the lifetime of the runtime, caching it once before the loop (or as a member) avoids the chain per iteration. Not a correctness issue — the branch predictor will learn it — but consistent with how similar platform flags are typically handled in ROCr.

cfreeamd (commented):

`projects/rocr-runtime/` only. Two files touched: `interrupt_signal.cpp` and `runtime.cpp`. The 1ms cap correctly addresses the missed-event problem. One question about CPU overhead and one nit left as inline comments.

victzhan:
Thanks for the feedback. The 1 ms wait was useful as a diagnostic workaround, but I agree it should not be the production fix. The AIRUNTIME-2074 trace now shows 38 cases where the HSA signal reached completion within 250 ms while the matching WinDXG Windows event wait returned WAIT_TIMEOUT at six seconds; 37 of those cases had no competing waiter. This isolates the failure to GPU mailbox/trap-to-KMD event signaling or ordering, so I am closing this PR and handing the trace to the driver owner instead of adding permanent runtime polling.

PR #8559, opened 2026-07-14
fix(rocr): Fix exported hsakmt NUMA dependency

victzhan:
## Motivation

This PR updates libhsakmt’s exported CMake package configuration so downstream consumers correctly recreate the NUMA imported target when the build used an imported target for libnuma, preventing failures when loading hsakmtTargets.cmake.

## Technical Details
Add conditional find_dependency(NUMA) to hsakmt-config.cmake.in to recreate the NUMA imported target before including exported targets.
Detect (during build configuration) whether NUMA_LIBRARIES contains an imported target and propagate that decision into the installed package config via HSAKMT_FIND_NUMA_DEPENDENCY.

## JIRA ID
ROCM-21395

## Submission Checklist


cfreeamd (commented):

`projects/rocr-runtime/libhsakmt/` only. The CMake changes in CMakeLists.txt and hsakmt-config.cmake.in look correct: `if(TARGET)` is the right idiom for detecting imported vs. raw-path NUMA entries, the `BUILD_SHARED_LIBS` guard matches when LINK_ONLY export applies, and the template expansion in the config file handles both ON/OFF cleanly. No concerns on the ROCr/libhsakmt side.

dayatsin-amd (approved):

PR #8560, opened 2026-07-14
Make DDA LL kernels HIP graph-capture safe

nawrinsu:
## Motivation

Move LL flag to a device-resident, per-block epoch array, so the flag and bank are derived on-device and advance monotonically across graph replays.

## Submission Checklist

### DIFF
--- a/projects/rccl/src/include/algorithms/all_gather/all_gather_dda_fabric_ll.h
+++ b/projects/rccl/src/include/algorithms/all_gather/all_gather_dda_fabric_ll.h
@@ -52,8 +52,8 @@ __global__ void ddaAllGatherFabricLL(
     size_t perRankBytes,                   // per-rank payload; multiple of 16
     int selfRank,
     int nRanksRt,
-    uint32_t flag,                         // == epoch; never 0
-    size_t bankOffsetPkts) {               // (epoch & 1) * nRanks * slotStride
+    uint32_t* __restrict__ epochDev,       // per-block LL epoch cells
+    int epochLen) {                        // number of cells in epochDev
 
   const int nRanks = NRANKS_CT ? NRANKS_CT : nRanksRt;
   const int peer = blockIdx.x;             // grid.x == nRanks: one column/peer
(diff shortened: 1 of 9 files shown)


projects/rccl/src/include/algorithms/all_gather/all_gather_dda_fabric_ll.h
```diff
@@ -65,14 +65,29 @@ __global__ void ddaAllGatherFabricLL(
...
 
+  // Flat block id + total launched blocks. tid 0 reads our own epoch cell (all
+  // cells hold the same value) and derives this launch's flag on the device, so
+  // nothing is baked into a HIP graph capture. bank = flag & 1.
+  const int flatBlockId = blockIdx.x * gridDim.y + blockIdx.y;
+  const int total = gridDim.x * gridDim.y;
+  __shared__ uint32_t s_flag;
+  if (tid == 0) {
+    uint32_t f = epochDev[flatBlockId] + 1u;
+    if (f == 0u) f = 2u;                   // skip 0 sentinel; keep bank parity
```
srirajpaul, line 76: when does f get a value 0?  epochDev seems memset to 0. so f should atleast be 1?
is it something related overflow? if so we could have kept it as size_t. As long as it works, it is fine.

projects/rccl/src/include/algorithms/all_gather/all_gather_dda_fabric_ll.h
```diff
@@ -65,14 +65,29 @@ __global__ void ddaAllGatherFabricLL(
...
+    s_flag = f;
+  }
+  __syncthreads();
+  const uint32_t flag = s_flag;
```
srirajpaul, line 80: i think every thread could just do `const uint32_t flag =  epochDev[flatBlockId] + 1u;`
and there is no need to use shared memory to get the flag. Not sure which one is better for performance.

projects/rccl/src/include/algorithms/all_gather/all_gather_dda_fabric_ll.h
```diff
@@ -95,34 +110,39 @@ __global__ void ddaAllGatherFabricLL(
...
-  for (size_t pk = pkBegin + tid; pk < pkEnd; pk += nthreads) {
-    uint32_t d0, f0, d1, f1;
-    do {
-      ddaLLLoadLineB128(
-          reinterpret_cast<const uint32_t*>(const_cast<LLPacket16*>(&src[pk])),
-          d0, f0, d1, f1);
-    } while (f0 != flag || f1 != flag);
-    out[2 * pk] = d0;
-    out[2 * pk + 1] = d1;
+  if (tid == 0) {
```
srirajpaul, line 142: all threads in block could update different locations

srirajpaul (approved):
Good to go.
All comments are just different options. none is better than other. in case you have already tested and the code is working, we dont need to change.

projects/rccl/src/include/algorithms/all_gather/all_gather_dda_fabric_ll.h
nawrinsu: Yes, this is handling overflow.

projects/rccl/src/include/algorithms/all_gather/all_gather_dda_fabric_ll.h
nawrinsu: Yes, either way works. I evaluated the performance with shared flag. I'll try the other way when I get the machine to see if there is any perf diff.

PR #8561, opened 2026-07-14
fix(amdsmi):ionic_rdma blacklisted

oliveiradan:
Fix the error (when running amd-smi):
'Drivers not loaded (amdgpu, amd_hsmp, ionic, rdma drivers not found in modules)'

Code changes related to the following:
  * API Check for rdma

## Motivation

Running `amd-smi` on a system where the RDMA (ionic) driver is not loaded fails with:
  `Drivers not loaded (amdgpu, amd_hsmp, ionic, rdma drivers not found in modules)`.

RDMA device info is optional for AI-NIC reporting, so its absence should not abort population of the rest of the NIC data

## Technical Details

In populate_amd_ainic_device() (src/amd_smi/amd_smi_system.cc), the call to `smi_get_nic_rdma_dev_info()` was guarded by `CHK_AMDNIC_RET`, which treats any `non-SMI_NIC_STATUS_SUCCESS` return as fatal: 
- It logs, converts the NIC status to an `amdsmi_status_t` error via `ainic_to_amdsmi_status()`, and returns early, aborting population of the device

## JIRA ID

JIRA ID : ROCM-26318

## Test Plan

  - Built `libamd_smi.so` and ran `amd-smi static` / `amd-smi list` on a host with the ionic/RDMA driver **not** loaded
  - Repeated on a host **with** the RDMA driver loaded to confirm no regression in RDMA device reporting

## Test Result

  - Without RDMA driver: `amd-smi` completes and reports NIC info: the "Drivers not loaded … rdma …" error no longer appears
  - With RDMA driver: RDMA device fields populated as before, no regression.

## Submission Checklist


oliveiradan:
!verify

projects/amdsmi/src/amd_smi/amd_smi_system.cc
```diff
@@ -437,7 +437,9 @@ static amdsmi_status_t populate_amd_ainic_device(const smi_nic_ctx_t& ctx, uint6
   static_assert(sizeof(smi_nic_rdma_devices_info_t) == sizeof(ai_nic_info.rdma_dev));
   status = smi_get_nic_rdma_dev_info(
       ctx, bdf_int, reinterpret_cast<smi_nic_rdma_devices_info_t*>(&ai_nic_info.rdma_dev));
-  CHK_AMDNIC_RET(status)
+  if (nic_rdma_status_is_fatal(status)) {
+    CHK_AMDNIC_RET(status);
```
bill-shuzhou-liu, line 441: If the `ionic` RDMA driver fails to load on a live NIC, there is currently no log line. Consider logging at DEBUG level before skipping, e.g. inside `nic_rdma_status_is_fatal` handling or right after this check.

bill-shuzhou-liu: **[WARNING]** The fix only special-cases the RDMA call; the five other `CHK_AMDNIC_RET` calls earlier in this function (bus/driver/asic/numa/port, lines ~413-435) still treat `SMI_NIC_STATUS_DRIVER_NOT_LOADED` as fatal.

Detail: `smi_get_nic_driver_info`, `smi_get_nic_asic_info`, `smi_get_nic_bus_info`, and `smi_get_nic_numa_info` in `smi_nic_interface.cpp` all return `SMI_NIC_STATUS_DRIVER_NOT_LOADED` when the base `ionic` driver isn't loaded on port 0. If a card is discovered but the base `ionic` driver hasn't bound yet (plausibly at least as likely as the RDMA-specific sub-driver being absent), `populate_amd_ainic_device` still aborts early on one of those five checks before ever reaching the RDMA fix — the whole NIC gets silently dropped, the same class of bug this PR fixes for RDMA. Worth confirming whether "RDMA is optional" is meant to be uniquely true among these six calls, or whether the same driver-not-loaded tolerance should extend to the others.

projects/amdsmi/tests/amd_smi_test/unit/nic/rdma_status_test.cc
```diff
@@ -0,0 +1,48 @@
...
+
+#include <gtest/gtest.h>
+
+#include "amd_smi/impl/amd_smi_common.h"
+
+namespace {
+
+using amd::smi::nic_rdma_status_is_fatal;
+
+TEST(NicUnit, RdmaOptionalStatusesAreNotFatal) {
```
bill-shuzhou-liu: **[WARNING]** This test only validates the standalone `nic_rdma_status_is_fatal()` predicate in isolation; it doesn't exercise `populate_amd_ainic_device()`, which is where the actual regression lived.

Detail: The bug being fixed is in `populate_amd_ainic_device()`'s control flow — it previously called `CHK_AMDNIC_RET(status)` unconditionally on the RDMA status, discarding bus/driver/asic/numa/port data already gathered. No existing test (the `functional/nic/{identity,discovery}` tests are placeholders) proves `populate_amd_ainic_device()` now returns `AMDSMI_STATUS_SUCCESS` with the rest of `ai_nic_info` intact when RDMA lookup fails. A future edit that reintroduces the unconditional `CHK_AMDNIC_RET(status)` (dropping the `if`) would still pass this test suite undetected.

bill-shuzhou-liu (approved):

projects/amdsmi/src/amd_smi/amd_smi_system.cc
oliveiradan: Logging inside `nic_rdma_status_is_fatal()` would break it as a pure predicate (a pure inline function). The unit test calls it directly and would emit log lines. I will add it right after the check.  Thanks for catching that.

marifamd (approved):

PR #8563, opened 2026-07-14
fix(rocjitsu): Fix 32 bit literal -> 64bit operand

lua1235:
## Motivation

32-bit literals should supply the high 32 bits of a 64-bit fp operand, rather than being zero extended. This was one of the issues exposed by the rocRAND test suite. 

## Technical Details

- Fix FP64 literal operand materialization so 32-bit literal extension words are placed in the high half of the 64-bit source value.

## JIRA ID

N/A

## Test Plan

Run ctests, rocjitsu-test-corpus

## Test Result

Tests pass

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
+++ b/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
@@ -615,7 +615,7 @@ class CodeGenerator:
         if operand_type not in ('OPR_SIMM16', 'OPR_SIMM32'):
             return None
         literal_expr = f'reinterpret_cast<const {lit_struct} *>(inst)->simm32'
-        if operand_type == 'OPR_SIMM16' or opnd.size == 16:
+        if operand_type == 'OPR_SIMM16':
             literal_expr = f'({literal_expr} & 0xFFFFu)'
         operand_size = size_expr or opnd.size
         if CodeGenerator._literal_operand_uses_f64_high_bits(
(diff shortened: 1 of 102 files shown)


kuhar (commented):
A few comments

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -5991,10 +6038,17 @@ def gen_insts(self) -> None:
                                 opnd.name in _lit_fields
                                 and opnd.name in enc_field_names
                             ):
+                                fixup = self._literal_operand_fixup_stmt(
```
kuhar, line 6056: We should preserve the full 32-bit extension word for dynamically encoded 16-bit operands. This path now goes through `_literal_operand_fixup_stmt`, whose `opnd.size == 16` branch masks off bits 31:16 before constructing the operand. `read_vop3_true16_src` later shifts by 16 when `op_sel` selects the high half, so a literal such as `0x80000000` becomes zero instead of `0x8000`. The existing `B16BitwiseLiteralPreservesInputSignBit` test only selects the low half. Please keep the full SIMM32 value on this dynamic path and add a high-half `op_sel` regression.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -605,18 +605,46 @@ def _literal_encoding_info(
...
 
+    @staticmethod
+    def _literal_operand_uses_f64_high_bits(
+        opnd: Operand, inst_sem: InstructionSemantics | None, operand_type: str
+    ) -> bool:
+        if (
+            operand_type != 'OPR_SIMM32'
+            or not opnd.is_input
+            or opnd.size != 64
+            or not inst_sem
```
kuhar: Literal placement should come from the operand or instruction signature rather than the availability of executable semantics. `_literal_operand_uses_f64_high_bits()` returns false when `inst_sem.data_type` is absent, so generated constructors still put the extension word in the low half for `V_TRIG_PREOP_F64` and gfx1250 `V_FMAMK_F64`/`V_FMAAK_F64`. For `V_TRIG_PREOP_F64`, the same `0xaf123456` extension must decode as `0xaf12345600000000` for 64-bit `src0` while remaining `0xaf123456` for 32-bit `src1`. Please classify this per operand and add a mixed-width regression.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
kuhar: The current fix should distinguish raw true16 storage from declared 16-bit literals and disassembly. Dynamically retyped true16 sources need the full extension word for `op_sel`, but declared `OPR_SIMM32` operands still need low-half masking: CDNA3 `v_madak_f16` with `simm32 = 0xdead3e00` now prints `0x-2152c200` instead of `0x3e00`. The same raw-storage change makes the two RDNA4 regressions disassemble as `0x-8000` and `0x-80000000` even though execution passes, while `test_literal_codegen.py:215` pins only the unmasked helper output. Could we thread the dynamic-true16 distinction through `_literal_operand_fixup_stmt()`, keep the selected 16-bit display value separate from the raw extension word, and assert both the declared-literal decode and true16 disassembly?

kuhar (approved):

Issue #5071, opened 2026-04-15
[Issue]: Segfault in AsyncEventsLoop when queuing many small kernels without synchronization ("fixed" with `HSA_TOOLS_DISABLE_REGISTER=1`)


fxmarty-amd:
### Problem Description

Hi,

This is ported from https://github.com/pytorch/pytorch/issues/180341 as it looks to be neither a torch nor vLLM issue.

I am seeing segfaults (with various errors detailed in the above issue) when running workload stacking many small kernels together. Eventually getting segfaults in `rocr::core::Runtime::AsyncEventsLoop`.

This issue does not occur on Nvidia devices, neither does it occur when using `HIP_LAUNCH_BLOCKING=1` or `AMD_SERIALIZE_KERNEL=3`.

**This issue can be reproduced with:**
- `torch==2.12.0.dev20260413+rocm7.2` (torch nightly), in a docker container with `ROCm version: 7.2.1 | amdgpu version: 6.16.13`
- `rocm/vllm-dev:nightly_main_20260413` that uses `torch==2.10.0+git8514f05`, `ROCm version: 7.2.1 | amdgpu version: 6.16.13`

**This issue can be reproduced as well with:**
```dockerfile
FROM rocm/dev-ubuntu-24.04:7.2.2-complete

RUN apt update && apt install -y git wget nano

ENV PATH="/root/miniforge3/bin:${PATH}"
ARG PATH="/root/miniforge3/bin:${PATH}"

RUN wget https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Linux-x86_64.sh && \
    bash Miniforge3-Linux-x86_64.sh -b && \
    conda init

ENV PYTORCH_ROCM_ARCH=gfx942
ENV GPU_ARCHS=gfx942
ENV NUM_JOBS=96

RUN pip3 install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/rocm7.2
```

with `torch==2.12.0.dev20260413+rocm7.2` and `ROCm version: 7.2.2 | amdgpu version: 6.16.13`.

**This issue does NOT occur in:**
- `rocm/pytorch:rocm7.2.2_ubuntu24.04_py3.12_pytorch_release_2.10.0` that uses  `torch==2.10.0+rocm7.2.2.lw.git40d237bf`, `ROCm version: 7.2.2 | amdgpu version: 6.16.13`
[...]

PR #8564, opened 2026-07-14
fix(rocr): fix out-of-bounds access in hsa_amd_signal_wait_any

bwelton:
JIRA ID : ROCM-27290

## Summary
Fixes an out-of-bounds read/write in `hsa_amd_signal_wait_any` that occurs whenever the satisfying signal in a multi-signal batch is at any index other than 0, plus two related condition/value-compaction and wait-any value-slot contract bugs found by review in the same code path.

## Root Cause
`hsa_amd_signal_wait_any` allocated `satisfying_value_vec` with a fixed size of `1`:

```cpp
std::vector<hsa_signal_value_t> satisfying_value_vec(1);
uint32_t satisfying_signal_idx =
    core::Signal::WaitMultiple(valid_signals.size(), valid_signals.data(), conds, values,
                               timeout_hint, wait_hint, satisfying_value_vec, false);
...
if (satisfying_value) *satisfying_value = satisfying_value_vec.at(0);
```

But `core::Signal::WaitMultiple()` (called with `wait_on_all=false`, i.e. "wait for any") writes `satisfying_values[i] = value` and returns `i` directly, where `i` can be any index in `[0, valid_signals.size())` — whichever signal is satisfied first. Whenever `i != 0`, `WaitMultiple()` writes out of bounds into the 1-element vector, and the subsequent `.at(0)` read (even if the OOB write somehow didn't corrupt anything) returns the wrong value regardless of which signal actually satisfied.

This also missed handling `WaitMultiple()`'s `uint32_t(-1)` timeout/invalid-signal sentinel return, which would separately index `valid_signal_ids` out of bounds when mapping the result back to the caller's original array position.

Additionally, two related bugs were found by review in the same area:
1. `hsa_amd_signal_wait_any()`/`hsa_amd_signal_wait_all()` built a compacted `valid_signals`/`valid_signal_ids` array by skipping NULL/invalid signals, but passed the caller's original, unfiltered `conds`/`values` pointers straight through to `WaitMultiple()`, which indexes them in the COMPACTED position — misaligning every valid signal after a skipped one.
2. `core::Signal::WaitMultiple()`'s wait-any branch violated its own documented contract (`satisfying_values` should contain only the first satisfying value at slot 0) by writing `satisfying_values[i]` at the local satisfying index instead — a separate real OOB write hit by `runtime.cpp`'s `AsyncEventsLoop`.

## Fix
- Size `satisfying_value_vec` to `valid_signals.size()` (matching `WaitMultiple()`'s local index space), not a fixed `1`.
- Return early on the `uint32_t(-1)` sentinel before doing any indexing.
- Read the satisfying value using `WaitMultiple()`'s *local* index into `satisfying_value_vec`, then map the *index returned to the caller* through `valid_signal_ids` back to the caller's original array position — these two index spaces are different and must not be conflated (the fix separates them explicitly with comments).
- Compact `conds`/`values` alongside `valid_signals` in both `hsa_amd_signal_wait_any` and `hsa_amd_signal_wait_all`.
- Make `Signal::WaitMultiple()`'s wait-any branch write `satisfying_values[0]` per its documented contract, letting `hsa_amd_signal_wait_any()` read a 1-element vector directly.

Sanity-checked `hsa_amd_signal_wait_all`'s analogous code path; it already sized its output vector correctly (`valid_signal_count`), so no equivalent fix was needed for that specific issue.

## How this was found
Discovered while prototyping a batched-signal-wait completion mechanism for rocprofiler-sdk's Queue Interposition feature (tracked separately, ROCM-27290 investigation), which calls `hsa_amd_signal_wait_any` with a multi-signal batch where the satisfying index is very commonly greater than 0 (a dedicated "wake" signal is always appended at the end of the batch). Any existing caller of `hsa_amd_signal_wait_any` with more than one signal in its batch is exposed to this bug today, independent of that follow-on work.

## Testing
- Added a new gtest regression suite, `SignalWaitMultiTest`, in `rocrtst/common/utils_test/signal_wait_multi_test.cpp`:
  - `WaitAnyReportsNonzeroSatisfyingIndex` — regresses the `satisfying_value_vec` OOB fix (satisfying signal at index 1 of 3).
  - `WaitAnyCompactsConditionsAndValues` — regresses the `conds`/`values` compaction fix (NULL signal interleaved before valid ones with distinct condition/value pairs).
  - `WaitAllReportsSatisfyingValues` — non-regression check for `hsa_amd_signal_wait_all`.
  - Confirmed these fail against the pre-fix library (`WaitAnyReportsNonzeroSatisfyingIndex` returns `satisfying_value=0` instead of the expected `41`) and pass 3/3 against a freshly rebuilt library containing this PR's fix.
- Exercised indirectly via a downstream rocprofiler-sdk prototype that calls `hsa_amd_signal_wait_any` with multi-signal batches where the satisfying index is >0 in the common case; before this fix, that prototype produced incorrect/garbage wake-signal values, consistent with this bug. After the fix: correct behavior confirmed across `ctest` runs and repeated stress runs.

## Related
Related to ROCM-27290 (rocprofiler-sdk Queue Interposition performance investigation) but is a standalone, independently-valid bug fix.
</content>


bwelton:
@cfreeamd this is a re-creation of https://github.com/ROCm/rocm-systems/pull/8094 to resolve the complaints from the bot on PR formatting / branch naming. Sorry about the churn here.

projects/rocr-runtime/rocrtst/common/utils_test/signal_wait_multi_test.cpp
```diff
@@ -0,0 +1,126 @@
...
+      3, signals, conds, values, timeout_hint_, HSA_WAIT_STATE_ACTIVE, &satisfying_value);
+
+  ASSERT_EQ(1u, satisfying_index);
+  EXPECT_EQ(signals[1].handle, signals[satisfying_index].handle);
+  EXPECT_EQ(41, satisfying_value);
+}
+
+TEST_F(SignalWaitMultiTest, WaitAnyCompactsConditionsAndValues) {
+  hsa_signal_t signals[3] = {};
+  ASSERT_EQ(HSA_STATUS_SUCCESS, CreateSignal(0, &signals[0]));
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `signals[1]` is intentionally left as `handle=0` (zero-initialized by `= {}`) to exercise the null/invalid skip logic — this is the exact failure scenario from the bug. A one-line comment like `// signals[1] intentionally null (handle=0) to exercise the null-skip path` would make the intent obvious to future readers rather than looking like a missing CreateSignal call.

projects/rocr-runtime/rocrtst/common/utils_test/signal_wait_multi_test.cpp
```diff
@@ -0,0 +1,126 @@
...
+  const uint32_t satisfying_index = hsa_amd_signal_wait_any(
+      3, signals, conds, values, timeout_hint_, HSA_WAIT_STATE_ACTIVE, &satisfying_value);
+
+  ASSERT_EQ(1u, satisfying_index);
+  EXPECT_EQ(signals[1].handle, signals[satisfying_index].handle);
+  EXPECT_EQ(41, satisfying_value);
+}
+
+TEST_F(SignalWaitMultiTest, WaitAnyCompactsConditionsAndValues) {
+  hsa_signal_t signals[3] = {};
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** The new test is wired into `rocrtstUtilsTestSrcs` (the `utils_test` binary), not the main `rocrtst` suite. Is `utils_test` run in CI on real hardware? If not, these tests only catch regressions in developer/local runs and the regression coverage for this fix would be manual only.

cfreeamd (commented):

Checked `hsa_ext_amd.cpp`, `signal.cpp`, and the new `signal_wait_multi_test.cpp`. The three-bug fix (conds/values misalignment, OOB-on-timeout in wait-any, wrong satisfying_value slot in signal.cpp) is correct and the tests directly reproduce the failure scenarios. One nit and one open question inline.

projects/rocr-runtime/rocrtst/common/utils_test/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@
 set (rocrtstUtilsTestSrcs utils_timer_gtest.cpp)
 set (rocrtstUtilsTestSrcs ${rocrtstUtilsTestSrcs} utils_timer_test.cpp)
 set (rocrtstUtilsTestSrcs ${rocrtstUtilsTestSrcs} utils_cpp11_gtest.cpp)
+set (rocrtstUtilsTestSrcs ${rocrtstUtilsTestSrcs} signal_wait_multi_test.cpp)
```
atgutier: Does this actually add the test to our active tests so that it's built/run?

atgutier (approved):
LGTM, nice find.

projects/rocr-runtime/rocrtst/common/utils_test/CMakeLists.txt
bwelton: My understanding is yes (at least it is actively run locally).

cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** Following up on this thread: "actively run locally" confirms developer/local coverage, but does the `utils_test` binary get picked up by a CI gtest filter group (Core/Extended)? If it isn't wired into a CI-invoked filter, this regression test won't guard against future breakage in CI — only local runs would catch a regression. Worth confirming before merge.

cfreeamd (commented):

Re-reviewed after @atgutier's approval. The three-bug fix (conds/values misalignment, OOB-on-timeout in wait-any, and the wrong satisfying-value slot in signal.cpp) still looks correct. Only lingering item is the CI-coverage question below — non-blocking.

bwelton:
## Correction: @atgutier and @cfreeamd were right — the test was not wired up

@atgutier asked whether this actually adds the test to our active tests so it's built/run, and I answered "My understanding is yes (at least it is actively run locally)." **That answer was wrong.** @cfreeamd's follow-up about CI filter groups was pointing at the same real problem. Apologies for the bad information — I've verified it properly and fixed it in 83c0458.

### What was actually wrong

`rocrtst/common/utils_test/` is dead CMake. The test was never built or run — not in CI, and not locally either.

1. **Nothing `add_subdirectory()`s it.** There is no top-level `rocrtst/CMakeLists.txt`; the rocrtst CMake project root is `rocrtst/suites/test_common/CMakeLists.txt` (`project(rocrtst64)`), and the only `add_subdirectory` it does is `gtest`. Grepping every `CMakeLists.txt`/`*.cmake` under `projects/rocr-runtime` returns zero references to `utils_test`, `rocrtstUtilsTestSrcs`, or `ROCRTST_UTIL_TEST_NAME` outside that one orphaned file.

2. **`rocrtst64`'s source list can't reach it.** `suites/test_common/CMakeLists.txt` uses `aux_source_directory(${ROCRTST_ROOT}/common common_srcs)`, and `aux_source_directory` is **not recursive**, so `common/utils_test/*.cpp` is excluded. There's no `file(GLOB_RECURSE)` anywhere in rocrtst.

3. **Proof by contradiction — that file can never have been processed.** `ROCRTST_UTIL_TEST_NAME` and `ROCRTST_LIBS` are both undefined in its scope, so it is a hard configure error. Force-including it:

   ```
   CMake Error at .../rocrtst/common/utils_test/CMakeLists.txt:24 (target_link_libraries):
     Cannot specify link libraries for target "elf" which is not built by this project.
   ```

   (`add_executable(${ROCRTST_UTIL_TEST_NAME} ...)` with an empty variable shifts the arguments, naming the target `utils_timer_gtest.cpp`; `target_link_libraries` then binds to `elf`.) Since rocrtst configures fine today, that file is definitively never processed.

4. `common/utils_test/` hasn't been touched since the pre-monorepo refactor.

### The fix (83c0458)

Reimplemented the tests as a `TestBase`-derived suite at `rocrtst/suites/functional/signal_wait_multi.{h,cc}`, which `aux_source_directory` **does** sweep into `rocrtst64` — the binary `test/therock/test_rocrtst.py` actually invokes under `test:rocrtst`. No CMake change needed. Registered in `suites/test_common/main.cc` under the `rocrtstFunc` convention and added to all four `test_categories.yaml` tiers (`quick`/`standard`/`comprehensive`/`full`) so they carry ctest labels:

- `rocrtstFunc.Signal_Wait_Any_Nonzero_Index`
- `rocrtstFunc.Signal_Wait_Any_Compacted_Conds_Values`
- `rocrtstFunc.Signal_Wait_All_Satisfying_Values`

The orphaned `.cpp` is deleted and the `utils_test/CMakeLists.txt` line is reverted (that file is now byte-identical to `develop`).

Also addressed @cfreeamd's nit — the intentionally-null `signals[1]` now carries an explanatory comment.

### Verified on hardware (gfx1201)

Built `rocrtst64`, then swapped **only** `libhsa-runtime64` via `LD_LIBRARY_PATH` (the binary uses `RUNPATH`, not `RPATH`, so the override genuinely applies — confirmed with `ldd` in both directions).

Against a **pre-fix** runtime, each test run in isolation:

| Test | Result |
|---|---|
| `Signal_Wait_Any_Nonzero_Index` | exit 1 — `Actual: 0 / Expected: 41` |
| `Signal_Wait_Any_Compacted_Conds_Values` | **exit 139, SIGSEGV** — the OOB write actually crashes |
| `Signal_Wait_All_Satisfying_Values` | exit 0, passes (correct; non-regression check) |

Against a runtime built from this tree: **3/3 pass**. Broader sanity `rocrtstFunc.Signal_*:rocrtst.Test_Example`: **8/8 pass**. `--gtest_list_tests` lists all three under `rocrtstFunc`, and `signal_wait_multi.cc.o` is present in the build tree. Rebuilding both changed TUs produced zero warnings.

### Suggested follow-up

`rocrtst/common/utils_test/` should just be deleted. It contains `utils_timer_gtest.cpp`, `utils_timer_test.cpp`, and `utils_cpp11_gtest.cpp` behind a `CMakeLists.txt` that cannot configure, so none of it has been compiled — let alone run — for as long as the current build structure has existed. It is untested dead code that actively misleads (it looked like a legitimate place to add a test, which is exactly how this PR went wrong). I've left it alone here to keep this PR scoped to the bug fix, but I'm happy to send a separate PR dropping the directory if there are no objections.

cc @atgutier @cfreeamd — since this adds a commit on top of the approved head, this probably needs a fresh look.

projects/rocr-runtime/rocrtst/common/utils_test/CMakeLists.txt
bwelton: You were right and my earlier answer was wrong — apologies.

`rocrtst/common/utils_test/` is dead CMake. Nothing `add_subdirectory()`s it, and `rocrtst64` gathers sources with `aux_source_directory(${ROCRTST_ROOT}/common ...)`, which is **not recursive**, so `common/utils_test/*.cpp` can never be compiled into the binary. `ROCRTST_UTIL_TEST_NAME` is also undefined everywhere, so force-including that `CMakeLists.txt` is a hard configure error:

```
CMake Error at .../common/utils_test/CMakeLists.txt:24 (target_link_libraries):
  Cannot specify link libraries for target "elf" which is not built by this project.
```

Since rocrtst configures fine today, that file is definitively never processed. The test was not built or run — not in CI, and not locally either. My "actively run locally" claim did not hold up.

Fixed in 83c0458: the tests are now a `TestBase` suite under `suites/functional/`, which `aux_source_directory` does sweep into `rocrtst64`, registered in `main.cc` as `rocrtstFunc.Signal_Wait_*` and added to all four `test_categories.yaml` tiers. This line is reverted and the orphaned file deleted.

Verified failing-first on gfx1201: against a pre-fix runtime `Signal_Wait_Any_Nonzero_Index` fails (`Actual: 0 / Expected: 41`) and `Signal_Wait_Any_Compacted_Conds_Values` segfaults; against a runtime built from this tree, 3/3 pass. Details in https://github.com/ROCm/rocm-systems/pull/8564#issuecomment-5132714521.

bwelton: Confirmed — it was not picked up by any CI filter group, and in fact was not picked up by anything at all. `utils_test` is never `add_subdirectory()`d, so the `utils_test` binary is never even configured, let alone run.

For CI coverage the relevant path is: source must live in a directory that `suites/test_common/CMakeLists.txt` sweeps with `aux_source_directory` (i.e. land inside `rocrtst64`), because `test/therock/test_rocrtst.py` runs `./rocrtst64` directly under `test:rocrtst`. Note that no workflow currently invokes `ctest` for rocrtst at all, so the standalone `add_test` targets (`intercept_queue_logic_test`, `hotswap_*`) get ctest labels but no TheRock CI execution — a standalone binary would not have solved this.

83c0458 moves the tests into `suites/functional/` so they are inside `rocrtst64`, and adds them to `test_categories.yaml` in all four tiers for the label-driven runs. Also verified they are not filtered out by `config/platform_config.yaml`: `REAL_HARDWARE` uses `allowed_groups: ["*"]`, so ungrouped tests run.

projects/rocr-runtime/rocrtst/common/utils_test/signal_wait_multi_test.cpp
bwelton: Answering the original question now that it is resolved: no, `utils_test` was never run in CI — and it turned out it was never built anywhere, including locally. See https://github.com/ROCm/rocm-systems/pull/8564#discussion_r3683932218. The tests now live in `rocrtst64` via `suites/functional/`, which CI does execute.

projects/rocr-runtime/rocrtst/common/utils_test/signal_wait_multi_test.cpp
bwelton: Addressed in 83c0458 — the test moved to `rocrtst/suites/functional/signal_wait_multi.cc`, and the intentionally-null signal now carries:

```cpp
hsa_signal_t signals[3] = {};
// signals[1] is intentionally left null (handle == 0) to exercise the
// NULL/invalid-signal skip and the conds/values compaction path.
```

cfreeamd (commented):

Both prior points are resolved in 83c0458:

- **CI coverage [Question]**: confirmed the old `utils_test` path was never built/run anywhere. The three regression tests now live in `rocrtst/suites/functional/signal_wait_multi.cc`, are registered in `test_common/main.cc` (`rocrtstFunc.Signal_Wait_Any_Nonzero_Index`, `Signal_Wait_Any_Compacted_Conds_Values`, `Signal_Wait_All_Satisfying_Values`), and are wired into `test_categories.yaml` filter groups — so they now get real rocrtst64 CI coverage.
- **[Nit]**: `signals[1]` now carries the explicit comment explaining it is intentionally null to exercise the NULL/invalid-skip and conds/values compaction path.

Thanks for chasing down the dead-CMake investigation.

dayatsin-amd (approved):

bwelton:
> Waiting on code owner review from cfreeamd and/or shwetagkhatri.

@cfreeamd / @shwetagkhatri can one of you stamp this change?

atgutier (approved):

jtb20:
FYI: I hit this bug and mistakenly attributed it to the same cause as https://github.com/ROCm/rocm-systems/issues/5071.

So, the unrelated bug has been fixed and closed, but not the wait_any crash. I thought I'd create a new PR so it didn't get forgotten, but then I found this one instead. Even better!

There's a test case attached to my comment in the above issue which might be helpful.

HTH!

jtb20:
Claude suggested a follow up on "might be helpful": excuse his robotic enthusiasm...

I built current `develop` (dfd3f592c6) and this PR's head (8a8f0526) and ran the
standalone reproducer from #5071 against each — same machine, same build config,
40 chained barrier-AND dispatches per run:

| runtime build | clean runs |
| --- | --- |
| `develop` @ dfd3f592c6 | 10 / 150 |
| `develop` + this PR | 60 / 60 |
| `develop` + the `signal.cpp` one-liner alone | 150 / 150 |

So roughly 93% of runs abort on `develop` as it stands. The failure is usually
`free(): invalid pointer` (SIGABRT), occasionally SIGSEGV in the waiting thread —
the heap metadata is smashed long before anything notices.

Environment: gfx942, ROCm 7.13.0 (`libhsa-runtime64.so.1.21.0`), Release build,
`BUILD_SHARED_LIBS=ON`.

## On the reproducer vs. the new unit tests

I think they're complementary rather than redundant, and it's worth being explicit
about why.

`TestWaitAnyNonzeroSatisfyingIndex` does fail deterministically on unfixed code, but
through the *value* assertion rather than the overflow: the old code writes
`satisfying_values[1]`, then `hsa_amd_signal_wait_any` reads `.at(0)`, which is still
zero, so `EXPECT_EQ(41, satisfying_value)` fails while `ASSERT_EQ(1u, satisfying_index)`
passes. The out-of-bounds write itself is never detected — on a 1-element
`std::vector<int64_t>`, element `[1]` lands inside glibc's minimum usable chunk, so
nothing actually gets corrupted.

That asymmetry is, I think, why this survived to production. The damage scales with the
satisfying index, and in `AsyncEventsLoop` the in-flight set is large: an index of ~30
writes ~240 bytes past an 8-byte allocation. The reproducer reaches that regime by
having one observer thread sit in `hsa_amd_signal_wait_any` over a growing set of
in-flight completion signals while the main thread keeps enqueuing without
synchronising.

Either of these would close the gap:

- run the new suite under ASan in CI, which catches the write directly rather than its
  downstream symptom; or
- add the reproducer as a stress test.

It's HSA-only (`dlopen`s `libhsa-runtime64`, no HIP or torch), single file, and exits
non-zero on failure, so it drops in as-is. [...]

itrowbri:
Just following up here: what is needed to get the approval for this PR to be merged? Do we need to add another test/reproducer @jtb20?

cfreeamd (approved):

cfreeamd:
> > Waiting on code owner review from cfreeamd and/or shwetagkhatri.
> 
> @cfreeamd / @shwetagkhatri can one of you stamp this change?

Sorry for the delay. I was on PTO for a couple of weeks and missed this ping.

shwetagkhatri (approved):

jtb20:
> Just following up here: what is needed to get the approval for this PR to be merged? Do we need to add another test/reproducer @jtb20?

The Claude-generated test might be useful to add so this doesn't regress but it shouldn't be a blocker, IMO. This is a serious bug, and we should prioritise getting the fix committed.

Thanks!

PR #8565, opened 2026-07-14
feat(rocr): round-robin SDMA copies across PCIe and xGMI engines

chun-wan:
## Motivation

gfx942, gfx950 and newer (CDNA, ISA major 9 minor >= 4) expose two
general-purpose "PCIe/paging" SDMA copy engines for host<->device transfers.
The stock runtime pins every process' host-to-device blit queue to the same
engine, so concurrent processes/streams serialize on one engine.

`HSA_SDMA_ROUND_ROBIN` (already in this PR) spreads those copies across the
two PCIe/paging engines using a `getpid()` seed. On these parts KFD reports
`num_sdma_engines == 2` and **both live on the first XCD/XCC**, so that flag
never leaves XCC0 -- `rec_eng` is always 0 or 1 (this was the reviewer's
observation, and it is correct).

This change adds a second, independent opt-in flag,
`HSA_SDMA_ROUND_ROBIN_PCIE_XGMI`, that widens the round-robin to **all** SDMA
engines: the `NumSdmaEngines` PCIe/paging engines plus the
`NumSdmaXgmiEngines` xGMI engines. On gfx942/gfx950 the engines are grouped
two-per-XCD/XCC (2 PCIe/paging on XCC0 + xGMI engines on the remaining XCCs),
so widening the modulus lets concurrent processes reach engines on other XCCs
and use each XCC's own path to host memory.

- **SPX (single partition):** the device spans all XCDs; spreading copies over
  engines on other XCCs raises aggregate host<->device bandwidth.
- **CPX (core-partitioned):** a partition is a single XCD with no xGMI engines
  (`NumSdmaXgmiEngines == 0`), so the widened modulus collapses back to the two
  PCIe engines -- degrades safely to the existing behavior, a no-op.

Existing `HSA_SDMA_ROUND_ROBIN` behavior is unchanged. When both flags are
set, the wider `PCIE_XGMI` behavior takes precedence.

## Recommended configuration

For anyone deploying this, the tested best-practice env combinations are:

**1) General multi-process concurrency (recommended, robust, general).** Spread
host-to-device over every engine for aggregate bandwidth, and confine
device-to-host to the near-AID engines so a single stream never drops onto a
far-AID engine (~20 GB/s):

```
export HSA_ENABLE_SDMA=1
export HSA_SDMA_ROUND_ROBIN_PCIE_XGMI=1
export HSA_SDMA_D2H_ENGINE_LIMIT=<near-AID engine count>   # = half the total engines: 8 on an 8-XCC part, 4 on a 4-XCC part
```

**2) With a launcher (torchrun / MPI / SLURM), also add** a distinct offset per
rank -- it removes PID-aliasing collisions and gives the most stable
multi-process D2H (or just rely on the built-in `LOCAL_RANK` /
`OMPI_COMM_WORLD_LOCAL_RANK` / `SLURM_LOCALID` auto-detection):

```
export HSA_SDMA_ENGINE_ID_OFFSET=${LOCAL_RANK}
```

**How to choose `HSA_SDMA_D2H_ENGINE_LIMIT`.** Set it to the near-AID engine
count, i.e. half of the total engines `(NumSdmaEngines + NumSdmaXgmiEngines) / 2`
(engine index `e` maps to AID `floor(e/2)`; the near AIDs are the lower half).
On an 8-XCC / 16-engine part that is `8`; on a 4-XCC / 8-engine part it is `4`.
When unsure, use half.

**Ceiling caveat.** host<->device aggregate bandwidth is bounded by the single
host PCIe Gen5 x16 link (~55-68 GB/s). Spreading across XCCs / AIDs helps you
*reach* that ceiling at lower concurrency and avoids the far-AID D2H penalty; it
does not raise the PCIe ceiling itself.

### Full configuration x concurrency matrix (8-XCC / 16-engine SPX part)

Aggregate GB/s, 4 MiB/stream, median of 3, measured on an 8-XCC / 8-AID SPX part
(16 SDMA engines). Scenarios:

- **S0** baseline (all round-robin flags off)
- **S1** `HSA_SDMA_ROUND_ROBIN=1` (2 near PCIe/paging engines)
- **S2** `HSA_SDMA_ROUND_ROBIN_PCIE_XGMI=1` (all 16 engines across 8 AIDs)
- **S3** S2 + `HSA_SDMA_D2H_ENGINE_LIMIT=8` (D2H clamped to near engines 0-7) -- **recommended**
- **S4** S2 + per-process `HSA_SDMA_ENGINE_ID_OFFSET=<rank>` (clean, collision-free) -- **recommended with a launcher**

| config | H2D N=1 | H2D N=2 | H2D N=4 | H2D N=8 | H2D N=16 | D2H N=1 | D2H N=2 | D2H N=4 | D2H N=8 | D2H N=16 |
|---|---|---|---|---|---|---|---|---|---|---|
| S0 | 54.2 | 53.6 | 52.6 | 52.1 | 64.2 | 48.6 | 46.7 | 46.5 | 47.2 | 65.3 |
| S1 | 54.2 | 55.1 | 55.7 | 56.4 | 60.8 | 48.7 | 52.6 | 53.2 | 55.7 | 59.5 |
| S2 | 54.2 | 58.6 | 59.6 | 59.4 | 61.2 | **19.5** | 54.1 | 56.2 | 67.5 | 65.9 |
| S3 | 54.8 | 58.4 | 58.8 | 60.3 | 62.1 | 47.5 | 55.3 | 67.4 | 62.8 | 63.6 |
| S4 | 54.5 | 60.2 | 57.9 | 58.9 | 64.0 | 42.9 | 55.1 | 66.8 | 66.7 | 68.5 |

Key reading: S2 at N=1 collapses to **19.5 GB/s** because a lone D2H stream can
land on a far-AID engine; S3 fixes that (47.5) by clamping D2H to the near AIDs
while H2D still spreads. Under real concurrency S3/S4 give the best D2H and
match or beat S2 on H2D. All absolute values are bounded by the single host
PCIe x16 link, so the win is "reach the ceiling sooner + no far-AID cliff",
not a higher ceiling.

## Feasibility (verified on hardware)

On a gfx942 SPX part (`num_sdma_engines=2`, `num_sdma_xgmi_engines=14`,
`num_xcc=8`; 16 engines total, 2 per XCC) we confirmed that host<->device SDMA
queues can be created on **every** engine id 0-15 (including the xGMI engines)
via KFD `SDMA_BY_ENG_ID`, and that they successfully move host (system) memory
-- there is no driver restriction limiting xGMI engines to GPU<->GPU traffic.

Single-stream bandwidth per engine (one process), showing AID/XCC locality:

| engine (-> XCC / AID) | H2D GB/s | D2H GB/s |
|---|---|---|
| 0,1 (XCC0 / AID0, near) | ~44 | ~42-49 |
| 2,3 (XCC1 / AID1, near) | ~44 | ~42-49 |
| 4,5 (XCC2 / AID2, near) | ~44 | ~42-49 |
| 6,7 (XCC3 / AID3, near) | ~44 | ~42-49 |
| 8,9 (XCC4 / AID4, remote) | ~44 | ~19.6 |
| 10,11 (XCC5 / AID5, remote) | ~44 | ~19.6 |
| 12,13 (XCC6 / AID6, remote) | ~44 | ~19.6 |
| 14,15 (XCC7 / AID7, remote) | ~44 | ~19.6 |

H2D is uniform (~44 GB/s) across all 16 engines and is insensitive to engine
position; single-stream D2H on the remote AIDs (engines 8-15) is ~2.2x slower
(~19.6 vs ~42-49 GB/s near) because it is cross-die latency bound, so a single
stream that happens to land on a remote engine can regress -- hence the wider
flag is opt-in and D2H can be clamped to the near AIDs.

## Alternatives evaluated

- **Atomic counter** (increment per queue): rejected -- counters are
  per-process (separate address spaces), so every process would start at 0 and
  collide on the same engine; it does not spread across processes without shared
  IPC state.
- **PID -> XCC mapping**: map the seed to an XCC and keep a process' H2D/D2H on
  that XCC's engines, preserving H2D/D2H direction separation. This can give a
  more consistent spread at moderate concurrency, but it bakes in XCC-layout
  assumptions, so it is not included here. The explicit engine-id offset plus the
  near-AID D2H limit reach the same goals (collision-free spreading, protected
  D2H) without hardcoding a layout, and remain a reasonable base for a future
  topology-aware refinement.

## Explicit engine-id offset (HSA_SDMA_ENGINE_ID_OFFSET)

The round-robin seed defaulted to `getpid()`. Because the engine is chosen by
`(seed + rec_eng) % engine_count`, two processes whose PIDs are congruent modulo
`engine_count` alias onto the same engine and collide, defeating the spread. Per
reviewer feedback (David), the seed can now be supplied explicitly:

- **`HSA_SDMA_ENGINE_ID_OFFSET=<n>`** -- a valid non-negative integer replaces
  the `getpid()` seed: `rec_eng = (n + rec_eng) % engine_count`. A launcher can
  hand each process a distinct offset (e.g. its local rank) so processes
  deterministically land on different engines with no collisions.
- The offset applies to **both** flags; the modulus is unchanged
  (`NumSdmaEngines` for `HSA_SDMA_ROUND_ROBIN`, `NumSdmaEngines +
  NumSdmaXgmiEngines` for `HSA_SDMA_ROUND_ROBIN_PCIE_XGMI`).
- `0` is a valid explicit offset (distinct from unset). Unset or malformed input
  (empty, negative, non-numeric, out of range) falls back to the `getpid()`
  seed, preserving current ease of use.

Reviewer preference is to always pass an explicit offset; the PID fallback is
retained only for convenience and can be dropped if the project prefers to
require an explicit offset.

Verified on a gfx942 SPX part via runtime instrumentation (engine_count 8 in
this instrumented run):

| env | seed | engine (rec_eng) |
|---|---|---|
| `HSA_SDMA_ENGINE_ID_OFFSET=0` | 0 (offset) | 0 |
| `HSA_SDMA_ENGINE_ID_OFFSET=3` | 3 (offset) | 3 |
| `HSA_SDMA_ENGINE_ID_OFFSET=5` | 5 (offset) | 5 |
| unset | 12054 (pid) | 6 (= pid % 8) |

Example (one distinct offset per local rank):

    export HSA_ENABLE_SDMA=1
    export HSA_SDMA_ROUND_ROBIN_PCIE_XGMI=1
    export HSA_SDMA_ENGINE_ID_OFFSET=${LOCAL_RANK}

Unit tests cover parsing of unset / 0 / 5 / non-numeric / negative /
out-of-range (all pass).

## Launcher local-rank auto-detection (default seed)

Building on `HSA_SDMA_ENGINE_ID_OFFSET`, the round-robin seed is now resolved
in priority order so a launcher no longer has to set the offset explicitly:

1. `HSA_SDMA_ENGINE_ID_OFFSET` (explicit, includes 0) -- highest priority.
2. otherwise the first launcher local-rank env var that parses to a valid
   non-negative integer, tried in this order: `LOCAL_RANK`,
   `OMPI_COMM_WORLD_LOCAL_RANK`, `MPI_LOCALRANKID`, `PMI_LOCAL_RANK`,
   `MV2_COMM_WORLD_LOCAL_RANK`, `SLURM_LOCALID`.
3. otherwise `getpid()` (unchanged fallback).

Only the seed default changes; the modulus and flag semantics are unchanged.
Because torchrun / Open MPI / MPICH / MVAPICH2 / SLURM already export a
per-process local rank, co-located ranks now land on distinct engines with zero
extra configuration, while a manual `HSA_SDMA_ENGINE_ID_OFFSET` still overrides
everything.

Example (no offset needed; the rank is supplied by the launcher):

    export HSA_ENABLE_SDMA=1
    export HSA_SDMA_ROUND_ROBIN_PCIE_XGMI=1
    # LOCAL_RANK / OMPI_COMM_WORLD_LOCAL_RANK / SLURM_LOCALID etc. are picked up automatically

Verified on a gfx942 SPX part via runtime instrumentation (H2D blit, base
`rec_eng` 0, `HSA_SDMA_ROUND_ROBIN`, modulus = `NumSdmaEngines` = 2):

| env | seed | source | rec_eng |
|---|---|---|---|
| `OMPI_COMM_WORLD_LOCAL_RANK=3` | 3 | `OMPI_COMM_WORLD_LOCAL_RANK` | 1 (= (3+0)%2) |
| `HSA_SDMA_ENGINE_ID_OFFSET=1` + `OMPI_COMM_WORLD_LOCAL_RANK=3` | 1 | `HSA_SDMA_ENGINE_ID_OFFSET` | 1 (offset overrides rank) |
| none set | 12936 | `getpid()` | 0 (= pid%2) |

Unit tests cover offset-only, offset-overrides-rank, single rank var, rank-var
priority order, invalid-rank-skipped, and none->pid (all pass).

## Opt-in near-AID limit for D2H (HSA_SDMA_D2H_ENGINE_LIMIT)

On gfx942 / gfx950 the SDMA engines are physically grouped per AID; host PCIe
egress lives on the near AID, so a device-to-host copy driven from a far-AID
engine is bandwidth-limited (the feasibility table above shows remote-AID
single-stream D2H at ~20 GB/s vs ~42-49 GB/s near). H2D is uniform across engines
and is unaffected.

`HSA_SDMA_D2H_ENGINE_LIMIT=N` (opt-in, default off) caps **D2H** round-robin
selection to the first `N` SDMA engines (the near-AID subset), computed as
`rec_eng = (seed + rec_eng) % min(N, engine_count)`. H2D is never limited and
continues to spread over all engines. `0` / unset / malformed keeps the stock
behavior of spreading D2H across all engines.

Why an explicit knob rather than auto-detection: the runtime has **no
per-engine->AID topology field** -- KFD node properties expose only
`NumSdmaEngines`, `NumSdmaXgmiEngines` and `NumXcc`, with no mapping from an SDMA
engine index to its AID or to the host PCIe-egress AID. The physical
engine->AID layout therefore cannot be derived reliably at runtime, so a
deployment sets `N` to its near-AID engine count rather than having the runtime
guess (a wrong guess would be worse than the status quo).

Verified on a gfx942 SPX part via runtime instrumentation, D2H blit (base
`rec_eng` 1) with `HSA_SDMA_ROUND_ROBIN_PCIE_XGMI=1`, seed 3 (engine_count 8 in
this instrumented run):

| env | D2H rec_eng | note |
|---|---|---|
| (no limit) | 4 (= (3+1)%8) | D2H may land on a far-AID engine |
| `HSA_SDMA_D2H_ENGINE_LIMIT=1` | 0 (= (3+1)%1) | D2H pinned to near-AID engine 0 |
| `HSA_SDMA_D2H_ENGINE_LIMIT=2` | 0 (= (3+1)%2) | D2H within near-AID subset {0,1} |

In the same runs H2D was confirmed **unaffected** by the limit (the direction
guard reports `limit_d2h=0` for H2D even with the knob set). Unit tests cover
unset / 0 / valid / large-value (clamped) / malformed (all pass).

## Scaling to other XCC counts

The implementation reads the engine counts dynamically from KFD
(`NumSdmaEngines` + `NumSdmaXgmiEngines`) and never hardcodes 2/4/8/16. The
round-robin path uses only the integer engine index `rec_eng` (passed to KFD as
`sdma_engine_id` via `SDMA_BY_ENG_ID`); it computes no bitmasks itself. This was
validated on an 8-XCC SPX part that reports 2 PCIe/paging + 14 xGMI = 16 engines,
where `rec_eng` naturally spans 0..15 across all 8 XCCs / 8 AIDs with no code
change. The public `HSA_AMD_SDMA_ENGINE_*` enum is defined through `_15`
(16 engines) and all engine masks are `uint32_t` (bit 0..31), so the 16-engine
case is fully within range. Parts with fewer XCCs (and therefore fewer engines)
are handled by the same count-agnostic path: the modulus simply shrinks to the
reported engine count, and on a single-XCD partition (`NumSdmaXgmiEngines == 0`)
the wider flag collapses back to the two PCIe/paging engines (a no-op).

## JIRA ID

JIRA ID: ROCM-28024

## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
@@ -55,7 +55,11 @@
 #include <utility>
 #include <iomanip>
 #include <cmath>
-#include <unistd.h>
+#ifdef _WIN32
+#include <process.h>  // _getpid
+#else
+#include <unistd.h>  // getpid
+#endif
 
 #include "core/inc/amd_aql_queue.h"
 #include "core/inc/amd_blit_kernel.h"
@@ -78,6 +82,18 @@
 // Generated header
 #include "amd_trap_handler_v2.h"
 #include "amd_blit_shaders_v2.h"
+
+namespace {
+// Portable process id used only as the SDMA round-robin seed fallback.
+// getpid() is POSIX; MSVC provides _getpid() via <process.h>.
+inline int PortableGetpid() {
+#ifdef _WIN32
+  return _getpid();
+#else
+  return getpid();
+#endif
+}
+}  // namespace
 
 #if defined(__linux__)
 // libdrm headers
@@ -964,7 +980,7 @@ void GpuAgent::InitDma() {
         const Flag& rt_flag = core::Runtime::runtime_singleton_->flag();
         const int64_t sdma_seed_off = rt_flag.sdma_seed_offset();
         const uint32_t sdma_seed = (sdma_seed_off >= 0) ? static_cast<uint32_t>(sdma_seed_off)
-                                                        : static_cast<uint32_t>(getpid());
+                                                        : static_cast<uint32_t>(PortableGetpid());
         const char* sdma_seed_src =
             (sdma_seed_off >= 0) ? rt_flag.sdma_seed_source().c_str() : "getpid()";
 
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
@@ -929,24 +929,37 @@ void GpuAgent::InitDma() {
       }
 
       // gfx942, gfx950 and newer (CDNA in major 9, minor >= 4) expose two
-      // general-purpose SDMA copy engines for host<->device transfers, and
-      // the default reverses
-      // SDMA0/1 for those copies. With ROCR_SDMA_ROUND_ROBIN=1, spread queue
-      // creation across all available SDMA engines using a pid seed so
-      // concurrent processes/streams land on different engines instead of
-      // contending on a single one.
+      // general-purpose (PCIe/paging) SDMA copy engines for host<->device
+      // transfers, and the default reverses SDMA0/1 for those copies.
       //
-      // The benefit depends on the partition mode. In CPX (core-partitioned)
-      // mode each partition is a single XCD that can enable only two SDMA
-      // engines -- one dedicated to H2D and one to D2H -- so there is no spare
-      // engine to round-robin onto and this flag has no effect. In SPX (single
-      // partition) mode the one device spans all XCDs and their SDMA engines,
-      // so host<->device copies can use SDMA engines on otherwise-idle XCDs,
-      // raising aggregate bandwidth. The NumSdmaEngines / NumSdmaXgmiEngines
-      // guards naturally no-op the flag when there is no spare engine.
+      // ROCR_SDMA_ROUND_ROBIN=1 spreads queue creation across the two
+      // PCIe/paging engines (NumSdmaEngines) using a pid seed so concurrent
+      // processes/streams land on different engines instead of contending on a
+      // single one. On these parts both PCIe/paging engines live on the first
+      // XCD/XCC, so this stays within that XCC.
+      //
+      // ROCR_SDMA_ROUND_ROBIN_PCIE_XGMI=1 widens the round-robin to ALL SDMA
+      // engines (NumSdmaEngines PCIe/paging + NumSdmaXgmiEngines xGMI). The
+      // engines are grouped two-per-XCD/XCC, so spreading over the full set
+      // lets concurrent processes/streams reach engines on other XCCs and use
+      // each XCC's own path to host memory, raising aggregate bandwidth in SPX
+      // (single-partition) mode. In CPX (core-partitioned) mode a partition is
+      // a single XCD with no xGMI engines (NumSdmaXgmiEngines == 0), so it
+      // degrades to the two-engine behavior and is a no-op. When both flags are
+      // set the wider PCIE_XGMI behavior takes precedence. The downstream
+      // NumSdmaXgmiEngines guard still forces the default engine pick when no
+      // xGMI engine exists.
       if (!prefer_xgmi && supported_isas()[0]->GetMajorVersion() == 9 && supported_isas()[0]->GetMinorVersion() >= 4 &&
           properties_.NumSdmaEngines > 0) {
-        if (core::Runtime::runtime_singleton_->flag().sdma_round_robin() == Flag::SDMA_ENABLE) {
+        if (core::Runtime::runtime_singleton_->flag().sdma_round_robin_pcie_xgmi() ==
+            Flag::SDMA_ENABLE) {
+          const uint32_t total_eng =
+              properties_.NumSdmaEngines + properties_.NumSdmaXgmiEngines;
+          rec_eng = (static_cast<uint32_t>(getpid()) + rec_eng) % total_eng;
+          debug_print("ROCR_SDMA_ROUND_ROBIN_PCIE_XGMI: queue rec_eng=%u of %u SDMA engines\n",
+                      rec_eng, total_eng);
+        } else if (core::Runtime::runtime_singleton_->flag().sdma_round_robin() ==
+                   Flag::SDMA_ENABLE) {
           rec_eng = (static_cast<uint32_t>(getpid()) + rec_eng) % properties_.NumSdmaEngines;
           debug_print("ROCR_SDMA_ROUND_ROBIN: queue rec_eng=%u of %u SDMA engines\n",
                       rec_eng, properties_.NumSdmaEngines);
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
@@ -951,8 +951,8 @@ void GpuAgent::InitDma() {
       // behavior takes precedence. The downstream
       // NumSdmaXgmiEngines guard still forces the default engine pick when no
       // xGMI engine exists.
-      if (!prefer_xgmi && supported_isas()[0]->GetMajorVersion() == 9 && supported_isas()[0]->GetMinorVersion() >= 4 &&
-          properties_.NumSdmaEngines > 0) {
+      if (!prefer_xgmi && supported_isas()[0]->GetMajorVersion() == 9 &&
+          supported_isas()[0]->GetMinorVersion() >= 4 && properties_.NumSdmaEngines > 0) {
         // Method A: the round-robin seed comes from Flag, which resolves it in
         // priority order: HSA_SDMA_ENGINE_ID_OFFSET (explicit, includes 0),
         // then a launcher local-rank env var (LOCAL_RANK,
@@ -963,9 +963,8 @@ void GpuAgent::InitDma() {
         // cause (two pids reducing to the same engine).
         const Flag& rt_flag = core::Runtime::runtime_singleton_->flag();
         const int64_t sdma_seed_off = rt_flag.sdma_seed_offset();
-        const uint32_t sdma_seed =
-            (sdma_seed_off >= 0) ? static_cast<uint32_t>(sdma_seed_off)
-                                 : static_cast<uint32_t>(getpid());
+        const uint32_t sdma_seed = (sdma_seed_off >= 0) ? static_cast<uint32_t>(sdma_seed_off)
+                                                        : static_cast<uint32_t>(getpid());
         const char* sdma_seed_src =
             (sdma_seed_off >= 0) ? rt_flag.sdma_seed_source().c_str() : "getpid()";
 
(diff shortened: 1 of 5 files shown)


cfreeamd (commented):

Checked `amd_gpu_agent.cpp` and `flag.h`. The env-var addition follows the existing `SDMA_OVERRIDE` / `Flag` class pattern correctly. `getpid()` seeding gives per-process (and per-queue within a process) distribution. The `NumSdmaEngines > 0` guard defensively prevents a latent divide-by-zero in the existing else branch. Two minor comment nits inline.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -815,9 +816,32 @@ void GpuAgent::InitDma() {
         *blits_[BlitHostToDev];
       }
 
-      // gfx94x is more efficient with reverse order of SDMA0/1 for host<->device copies
-      if (!use_xgmi && isa_->GetMajorVersion() == 9 && isa_->GetMinorVersion() >= 4)
-        rec_eng = (rec_eng + 1) % properties_.NumSdmaEngines;
+      // gfx942, gfx950 and newer (CDNA in major 9, minor >= 4; e.g.
+      // MI300/MI308/MI325 and MI350/MI355) expose two general-purpose SDMA
```

**[Nit]** The comment says "gfx942, gfx950 and newer" but the condition `isa_->GetMinorVersion() >= 4` also covers gfx940/941 (MI300A variants). Consider "gfx940 and newer (CDNA, major 9, minor >= 4)" to match the actual condition bounds.

projects/rocr-runtime/runtime/hsa-runtime/core/util/flag.h
```diff
@@ -106,6 +106,19 @@ class Flag {
     enable_sdma_recommended_eng_ = (var == "0") ? SDMA_DISABLE :
                                    ((var == "1") ? SDMA_ENABLE : SDMA_DEFAULT);
```
cfreeamd, line 128: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** The CPX/SPX explanation here is identical to the block comment in `amd_gpu_agent.cpp`. Both locations are fine, but the duplication means two places to update if partition-mode behavior changes. A cross-reference comment in one of them (e.g., `// see also GpuAgent::InitDma() comment`) would make the duplication intentional rather than accidental.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
dayatsin-amd: MI-300A is also gfx942

projects/rocr-runtime/runtime/hsa-runtime/core/util/flag.h
chun-wan: Good catch. Reworded the comment to match the actual ISA guard (major == 9 && GetMinorVersion() >= 4, i.e. gfx940/941/942/950 and newer) so it no longer implies only gfx942/gfx950. Fixed in the latest commit.

chun-wan: Agreed. Added a cross-reference comment so the duplication is intentional (the flag.h block and the GpuAgent::InitDma() block now point at each other). Fixed in the latest commit.

saleelk:
This policy is already implemented in HIP\ROCr just after 7.2 for exactly avoiding the serialization issue, I would rather prefer a back port of it rather than a new change

chun-wan:
@saleelk thanks, I wasn't aware of this update.

If the cross-process case isn't already covered by #8456, I'm happy to rework this on top of the new NthSdmaEngine architecture — keeping only the cross-process env controls and dropping the round-robin machinery your change already replaced, rather than introducing a separate mechanism. Does that sound reasonable to you?

PR #8569, opened 2026-07-14
test(amdsmi): [AILITOOLS-227] add monitor consistency tests

amd-josnarlo:
## Motivation

The amd-smi monitor CLI had no automated check that its reported telemetry is self-consistent, agrees with the metric command, or actually responds to GPU load. Add functional tests so regressions in monitor's power, thermal, clock, activity, and VRAM reporting are caught on real hardware.

## Technical Details

CLI tests

- Add a shared monitor_args covering power, base-board/GPU temps, gfx/mem activity, encoder/decoder, ECC, VRAM usage, and PCIe
- Add a multiprocessing harness (_multiprocess_commands / _worker) that launches commands against a shared high-precision start time so readings are captured near-simultaneously
- Add _get_monitor_metric_data to normalize monitor and metric [--json] output into per-GPU [[value, other, diff, unit]] tuples (selecting GTT vs VRAM by larger capacity)
- Add _compare_monitor_metric_data / _print_monitor_results for tolerance-based comparison and pass/fail reporting 

- test_monitor_monitor: two concurrent monitor runs must agree within tolerance 
- test_monitor_metric: monitor and metric readings must agree within tolerance
- test_monitor_with_workload: monitor values must shift under an rvs workload versus an idle two-sample baseline; skips cleanly when rvs or its config files are absent

## JIRA ID

JIRA ID : AILITOOLS-270

## Test Plan

Run python cli tests, cli_unit_test.py -v -k test_monitor

## Test Result

Unit passed as expected

### DIFF
--- a/projects/amdsmi/tests/python/cli/test_monitor.py
+++ b/projects/amdsmi/tests/python/cli/test_monitor.py
@@ -2,12 +2,71 @@
 # Copyright Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
 
-"""CLI leaf test: monitor command."""
+"""CLI leaf test: monitor command and its agreement with metric."""
+
+import json
+import os
+import signal
+import subprocess
+import time
+import unittest
+from collections import namedtuple
+
+import common.common as common
 
 from cli.base import TestCliBase
 
+# Live counters. Two samples of these legitimately differ, so only their
+# presence and units are cross-checked, never their values.
+VOLATILE_KEYS = (
+    "power_usage",
+    "hotspot_temperature",
+    "memory_temperature",
+    "gfx_clk",
+    "gfx",
+    "mem",
+)
+# Memory accounting. Both commands read it from the same driver call.
+STABLE_KEYS = ("vram_used", "vram_free", "vram_total")
+
+# Monitor column -> path into a metric gpu_data entry. The vram keys are
+# resolved separately because metric reports both the VRAM and GTT pools.
+METRIC_PATHS = {
+    "power_usage": ("power", "socket_power"),
+    "hotspot_temperature": ("temperature", "hotspot"),
+    "memory_temperature": ("temperature", "mem"),
+    "gfx_clk": ("clock", "gfx_0", "clk"),
+    "gfx": ("usage", "gfx_activity"),
+    "mem": ("usage", "umc_activity"),
+}
+
+Field = namedtuple("Field", "unit value")
+Result = namedtuple("Result", "gpu key passed message")
+
+NA = Field("N/A", None)
+
+
+def _field(container, *path):
+    """Resolve a {value, unit} node, yielding NA if any level is absent or "N/A"."""
+    node = container
+    for key in path:
+        if not isinstance(node, dict) or key not in node:
+            return NA
+        node = node[key]
+    if not isinstance(node, dict) or "value" not in node:
+        return NA
+    return Field(node.get("unit", ""), node["value"])
+
 
 class TestMonitor(TestCliBase):
+    # Only the columns the comparisons read; test_command covers the rest.
+    monitor_args = "--power-usage --temperature --gfx --mem --vram-usage"
+    tolerance = 0.1
+    # Counters a compute/memory workload is expected to move.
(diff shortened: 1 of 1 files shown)


projects/amdsmi/tests/python/cli/test_monitor.py
```diff
@@ -38,3 +46,392 @@ def test_command(self):
...
+        self.common.print(msg)
+
+        self.rvs_exe = "rvs"
+        self.rvs_config_folder = "/opt/rocm/share/rocm-validation-suite/conf"
+        # name, options, ramp_time
+        # rvs_configs is a list of tuples containing:
+        # configuration file name, additional options, the ramp time for the workload
+        self.rvs_configs = []
+        self.rvs_configs.append(("iet_single.conf", "", 5.0))
+        self.rvs_configs.append(("mem.conf", "--numTimes 3", 1.0))
```
marifamd, line 335: I would say split this behavior out into python functional test folder since we are workload testing

marifamd (changes requested):

projects/amdsmi/tests/python/cli/test_monitor.py
```diff
@@ -38,3 +46,392 @@ def test_command(self):
...
+                        elif key == "hotspot_temperature":
+                            data2 = int(metric["gpu_data"][i]["temperature"]["hotspot"]["value"])
+                        elif key == "memory_temperature":
+                            data2 = int(metric["gpu_data"][i]["temperature"]["mem"]["value"])
+                        elif key == "gfx_clk":
+                            data2 = int(metric["gpu_data"][i]["clock"]["gfx_0"]["clk"]["value"])
+                        elif key == "gfx":
+                            data2 = int(metric["gpu_data"][i]["usage"]["gfx_activity"]["value"])
+                        elif key == "mem":
+                            data2 = int(metric["gpu_data"][i]["usage"]["mm_activity"]["value"])
```
gabrpham: --mem option in the monitor command returns umc_activity, not mm_activity. This should be umc_activity instead.

projects/amdsmi/tests/python/cli/test_monitor.py
```diff
@@ -38,3 +46,392 @@ def test_command(self):
...
+                data[i]["vram_free"] = None
+                data[i]["vram_total"] = None
+            # Find the larger of the two amounts
+            if metric is not None:
+                total_gtt = int(metric["gpu_data"][i]["mem_usage"]["total_gtt"]["value"])
+                total_vram = int(metric["gpu_data"][i]["mem_usage"]["total_vram"]["value"])
+                if total_gtt > total_vram:
+                    vram_key = "gtt"
+                else:
+                    vram_key = "vram"
```
gabrpham: This comparison should only apply to APU devices. Discrete GPUs should default to VRAM.

projects/amdsmi/tests/python/cli/test_monitor.py
```diff
@@ -38,3 +46,392 @@ def test_command(self):
         )
         self.RunCmds(cmds)
         return
+
+    def _get_monitor_metric_data(self, monitor1, monitor2, metric, exclude=False):
```
gabrpham: I assume that this is supposed to get all the metric data that monitor gathers, but it doesn't seem to also get things from the encoder/decoder option, the ecc option, the pcie option, and others that are present in the monitor args. I assume they should align?

gabrpham (changes requested):

projects/amdsmi/tests/python/cli/test_monitor.py
amd-josnarlo: Workload test has been disabled but remain in place when reliable workload can be enabled.
A stand-alone monitor with workload is being developed.

projects/amdsmi/tests/python/cli/test_monitor.py
amd-josnarlo: Changed

projects/amdsmi/tests/python/cli/test_monitor.py
amd-josnarlo: They should align and can be added at a later date

PR #8573, opened 2026-07-14
fix(rocjitsu): Fix high-output word selector for v_pk_mov_b32

lua1235:
## Motivation

This updates the packed move generator so the high 32-bit output is selected from `op_sel[1]`, matching the hardware-observed behavior for `v_pk_mov_b32` with the assembler default `op_sel_hi = 3`. The generated shared execution body now uses `op_sel` for both output dword selectors instead of mixing `op_sel` and `op_sel_hi`. This issue was exposed by the rocRAND test suite.

Despite being in the VOP3P family, `v_pk_mov_b32` does not use op_sel_hi to select input halves to compute the high output half. Instead, cdna3/4 ISA states that "The source operands are treated as 64 bit and are subject to alignment restrictions for both SGPR and VGPR. For two VGPR inputs this opcode can be used as an arbitrary gather by using OP_SEL to select either the even VGPR specified or the next odd VGPR." 
ISA example (op_sel determines the source for both the low and high half of output):
  `v_pk_mov_b32 v0, v2, v4 op_sel:[0,1] // evaluates v0 <- v2 and v1 <- v5`

## Technical Details

-_PK_MOV_B32` now treats `op_sel` as the selector for both output dwords: `op_sel[0]` selects `src0.lo/src0.hi` for the low result, and `op_sel[1]` selects `src1.lo/src1.hi` for the high result. `op_sel_hi` remains constrained to the assembler default value for the SIMD fast path.
- Widened SIMD implementation from default `op_sel == 0` to all four `op_sel` combinations.

## JIRA ID

N/A

## Test Plan

ctest and rocjitsu-test-suite

## Test Result

all pass

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/tests/simd_correctness/vop3p_mov_b32_simd_correctness_test.cpp
+++ b/emulation/rocjitsu/tests/simd_correctness/vop3p_mov_b32_simd_correctness_test.cpp
@@ -8,7 +8,8 @@
 /// op_sel[0] selects the low output dword from src0.{lo,hi}, and op_sel[1]
 /// selects the high output dword from src1.{lo,hi}. Each case runs twice in
 /// the same process -- once forcing the scalar body, once allowing the SIMD
-/// fast path -- and both paths must match the same golden 64-bit result.
+/// fast path -- and both paths must match the same golden 64-bit result. A
+/// mixed SGPR-pair/VGPR-pair case covers the legal scalar source encoding.
 /// In-process inactive lanes must keep the sentinel.
 
 #include "util/simd_test_hooks.h"
@@ -43,6 +44,10 @@ constexpr uint32_t SGPRS_PER_WF = 106;
 constexpr uint32_t VGPRS_PER_WF = 256;
 constexpr uint32_t kDstVgpr = 8; // pair occupies kDstVgpr..kDstVgpr+1
 constexpr uint32_t DST_SENTINEL = 0xCDCDCDCDu;
+constexpr uint32_t kMixedSgprLo = 0x11111111u;
+constexpr uint32_t kMixedSgprHi = 0x22222222u;
+constexpr uint32_t kMixedVgprLo = 0x55555555u;
+constexpr uint32_t kMixedVgprHi = 0x66666666u;
 
 struct ArchCase {
   rj_code_arch_t arch;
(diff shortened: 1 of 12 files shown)


emulation/rocjitsu/tests/simd_correctness/vop3p_mov_b32_simd_correctness_test.cpp
```diff
@@ -120,39 +171,72 @@ struct ForceScalarGuard {
...
+                      uint32_t op_sel) -> std::array<uint64_t, WF_SIZE> {
     util::set_force_scalar_for_testing(force_scalar);
-    Fixture fx;
+    Fixture fx(arch);
     EXPECT_NE(fx.cu, nullptr);
     EXPECT_NE(fx.wf, nullptr);
     uint32_t words[2] = {0u, 0u};
     // src0 = VGPR 256 (pair v0:v1), src1 = VGPR 258 (pair v2:v3).
-    vop3p_encode(/*op=*/51, kDstVgpr, /*src0=*/256, /*src1=*/258, words);
+    vop3p_encode(/*op=*/51, kDstVgpr, /*src0=*/256, /*src1=*/258, op_sel, words);
```
kuhar, line 215: I think we should cover the legal SGPR-pair source form before treating these selector cases as the instruction contract. `VPkMovB32Vop3p` is generated with `OPR_SRC_VGPR_OR_ACCVGPR`, so an SGPR encoding such as `src0=0` resolves as VGPR0. A local mixed-source repro with `s0:s1 = 0x11111111:0x22222222`, `v4 = 0x55555555`, and `op_sel:[1,0]` produced `0x5555555500000000` instead of `0x5555555522222222`; LLVM accepts this form on gfx90a, gfx942, and gfx950. Please use the legal SGPR-or-VGPR operand class and add a mixed SGPR/VGPR golden case.

kuhar (commented):
Looks good overall, just one comment

kuhar (commented):
I think it would help to keep one end-to-end SIMD dispatch case in `emulation/rocjitsu/tests/simd_correctness/vop3p_mov_b32_simd_correctness_test.cpp:143`. `run_simd_probe` calls `try_execute_vop3p_mov_b32_simd` directly, while only forced-scalar mode goes through `execute_instruction`, so the suite would still pass if the production `ROCJITSU_TRY_SIMD_VOP3P_MOV_B32` hook were disconnected. Please run at least one `force_scalar=false` case through `execute_instruction` and compare it with the golden result.

kuhar (approved):

PR #8579, opened 2026-07-14
feat(rocshmem): [scripts] add --variant-dir option to run_perf_compare.sh

omor1:
## Motivation

The baseline and comparison branches could be input as pre-built directories to the `run_perf_compare.sh` script, but additional variants could not. Adding a `--variant-dir` option adds consistency and simplifies some workflows.

## Technical Details

* Add `--variant-dir SPEC` option to `run_perf_compare.sh`.
  Each `SPEC` is in the form `NAME:ENV1=V1,...:PATH`, where
  * `NAME` is the label used in plots
  * `ENV1=V1,...` is a list of comma-separate environment variables to set at runtime
  * PATH is the directory path containing the pre-built variant
* Fix handling of the `-h|--help` option to only print out the script's help text.

## Test Result

Testing using #7217 with a pre-built variant that builds only a single GDA provider.

## Submission Checklist

### DIFF
--- a/projects/rocshmem/scripts/functional_tests/run_perf_compare.sh
+++ b/projects/rocshmem/scripts/functional_tests/run_perf_compare.sh
@@ -16,9 +16,9 @@
 #   --cmake-args   ARGS   Extra cmake args for all builds (quoted string)
 #   --branch-args  ARGS   Extra cmake args for the branch build only (quoted string)
 #   --variant-args SPEC   Additional named variant: NAME:ENV1=V1,...:cmake-args
-#                         NAME       - label used in plots and directory name
-#                         ENV1=V1,…  - comma-separated env vars set at runtime
-#                         cmake-args - extra cmake args for this variant's build
+#                         NAME       - Label used in plots and directory name
+#                         ENV1=V1,…  - Comma-separated environment variables to set at runtime
+#                         cmake-args - Extra CMake arguments for this variant's build
 #                         May be repeated for multiple variants.
 #                         Example: --variant-args "sdma-on:ROCSHMEM_SDMA_ENABLED=1:-DUSE_SDMA=ON"
 #   --pr           NUM    Compare GitHub PR NUM against origin/develop baseline.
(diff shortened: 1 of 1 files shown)


projects/rocshmem/scripts/functional_tests/run_perf_compare.sh
```diff
@@ -76,13 +83,20 @@ while [[ $# -gt 0 ]]; do
...
     --branch-dir)    BRANCH_DIR="$2";                      shift 2 ;;
+    --variant-dir)   VARIANT_DIR_SPECS+=("$2");            shift 2 ;;
     --skip-build)    SKIP_BUILD=1;                         shift ;;
     --skip-baseline) SKIP_BASELINE=1;                      shift ;;
     --skip-develop)  SKIP_BASELINE=1;                      shift ;;
     --skip-branch)   SKIP_BRANCH=1;                        shift ;;
     --outdir)        OUTDIR="$2";                          shift 2 ;;
     -h|--help)
-      sed -n '2,/^###$/p' "$0" | head -n -1
+      # /^##\+$/,/^##\+$/ : match lines between those filled with two or more #s
```
abouteiller: what was wrong with the simpler expression beside the head?

abouteiller (approved):

projects/rocshmem/scripts/functional_tests/run_perf_compare.sh
omor1: It didn't work. `2,/^###$/` matches all lines between line 2 and the line consisting of exactly 3 ###, which there are none in the file, so it would print out the entire file. It also doesn't strip off the leading `'#'` or `'# '`.

`sed -n '3,/^##*#$/s/^# \?//p' "$0" | head -n -1` would work, but it's somewhat fragile in assuming that the first line in the help text starts on line 3. Probably not a huge problem though. I can use that if preferred.

abouteiller: If it works its good enough :)

projects/rocshmem/scripts/functional_tests/run_perf_compare.sh
```diff
@@ -9,29 +9,36 @@
...
 #   --baseline-dir PATH   Use PATH as pre-built baseline (skips baseline build).
 #                         Use with --skip-baseline to also skip baseline test runs.
-#   --branch-dir PATH     Use PATH as pre-built branch build (skips branch build).
+#   --branch-dir   PATH   Use PATH as pre-built branch build (skips branch build).
+#                         Use with --skip-branch to also skip branch test runs.
+#   --variant-dir  SPEC   Additional named variant: NAME:ENV1=V1,...:PATH
+#                         NAME       - Label used in plots
+#                         ENV1=V1,…  - Comma-separated environment variables to set at runtime
+#                         PATH       - Use PATH as pre-built variant build (skips variant build)
+#                         May be repeated for multiple variants.
```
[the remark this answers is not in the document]
omor1: This is the same as existing practice for `--variant-args`, @abouteiller is this a concern?

projects/rocshmem/scripts/functional_tests/run_perf_compare.sh
```diff
@@ -345,6 +371,16 @@ for spec in "${VARIANT_SPECS[@]}"; do
...
 
+for dir_spec in "${VARIANT_DIR_SPECS[@]}"; do
+  IFS=':' read -r vname venv vdir <<< "$dir_spec"
+  env_prefix=""
+  IFS=',' read -ra pairs <<< "$venv"
+  for pair in "${pairs[@]}"; do
+    [[ -n "$pair" ]] && env_prefix+="export $pair; "
+  done
+  run_iterations "$vdir" "logs-${SUITE}" "$vname" "$env_prefix"
+done
```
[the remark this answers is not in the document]
omor1: @abouteiller is this a concern, I don't think path existence is checked for `--baseline-dir` or `--branch-dir` either?

abouteiller (approved):

PR #8584, opened 2026-07-14
feat(hotswap): own the retargeted artifact via a loader prepare API

xintin:
Adds a loader prepare API so the retargeted (B0->A0) code object is owned by the caller and consumed at the load boundary, replacing the init-time prewarm model.
- Loader vendor table 1_04: `hsa_ven_amd_loader_code_object_reader_prepare` retargets an eligible object and returns a prepared code-object reader that owns the artifact. When no retarget is needed it returns a null prepared handle and the source loads unchanged. A prepared reader loads directly, bypassing onLoad interception so it is never re-analyzed or re-rewritten.
- Consumed in `roc::Program` (HIP); the existing onLoad interception stays as the fallback for direct-HSA users and runtimes without the 1_04 entry.
- `FatBinaryInfo` starts prepare on a background thread at fat-binary digest time, so the one-time retarget can overlap the window before the first launch. In-flight preparations are joined before the source bytes are released.

The persistent retarget cache is tracked separately in #8659 and is intentionally not part of this PR.

## JIRA ID
https://amd-hub.atlassian.net/browse/ROCM-28061

## Results
Prepare-only; perf/memory neutral standalone. Establishes the owned-artifact load boundary required by the persistent cache (#8659), and replaces the init-time prewarm/global-cache design.

## Test Result
- Builds clean (ROCr + CLR).
- Unit tests pass: `HotswapRewrite.*` and `HotswapPrepare.*` (34/34).

## Submission Checklist


projects/clr/hipamd/src/hip_fatbin.cpp
```diff
@@ -20,7 +26,53 @@
...
+
 namespace hip {
+
+namespace {
+// Init-time hotswap prewarm is on by default; no env var needs to be set to
+// enable it. It can be explicitly disabled by setting HIP_HOTSWAP_PREWARM to a
+// false-like value (0/off/false/no/n/f).
+bool HotswapPrewarmEnabled() {
+  static const bool enabled = []() {
+    const char* v = getenv("HIP_HOTSWAP_PREWARM");
```
harsh-amd: Why don't we want this always enabled?

projects/rocr-runtime/rocrtst/suites/functional/hotswap/hotswap_rewrite_test.cc
```diff
@@ -777,4 +861,237 @@ TEST(HotswapRewrite, RuntimeLoadRequiredStrictRewrittenLoadFailureReturnsError)
...
+size_t CountCacheFilesRecursive(const std::string& dir);
+
+}  // namespace
+
+// A file-backed code object is keyed by (uri, file mtime, file size, object
+// size) without hashing the object, so a second load of the same file is
+// served from cache without growing it.
+TEST(HotswapRewrite, FileUriKeyServesSecondLoadFromCache) {
+  ResetRuntimeTestEnv();
+  if (!NewComgrHotswapApiAvailable()) return;
```
harsh-amd: These four tests do not compile because `NewComgrHotswapApiAvailable()` is not defined. Should these use `ComgrHotswapOptionsApiAvailable()`?

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -73,6 +87,396 @@ std::unordered_map<uint64_t, std::vector<OwnedElfBuffer>> g_retained_rewritten_e
...
+struct ComgrApi;
+ComgrApi* GetComgrApi();
+
+// Resolve the COMGR path for the disk-cache salt. Force the (cheap, memoized)
+// COMGR resolution so g_comgr_lib_path is set regardless of load ordering;
+// otherwise a salt computed before COMGR loads differs from one computed after,
+// and warm runs miss the entry the cold run wrote. The retarget still only runs
+// on a genuine cold miss.
+std::string ResolveComgrPathForSalt() {
+  (void)GetComgrApi();
```
harsh-amd: This calls `GetComgrApi()` before probing the disk cache, so a warm hit still dlopens COMGR. Does that defeat the no-COMGR warm path? Can the salt come from an ABI or package identity available without loading COMGR?

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -400,6 +824,146 @@ void LogRequiredRewrittenLoadFailure(hsa_status_t status) {
...
+#endif
+
+  // In-memory tier. Only deterministic COMGR failures are cached; transient
+  // allocation failures are not, so a later attempt can still succeed.
+  try {
+    std::scoped_lock lock(g_retarget_cache_mutex);
+    if (g_retarget_cache.find(cache_key) == g_retarget_cache.end() &&
+        g_retarget_cache.size() < kMaxRetargetCacheEntries) {
+      CachedRetargetResult entry;
+      entry.succeeded = rewritten;
```
harsh-amd: `rewritten == false` also covers `create_data`, `set_data`, `get_data`, and allocation failures. That makes a transient failure permanent for the process. Can we distinguish deterministic and transient failures and only negative-cache the former?

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -499,33 +1152,108 @@ RetargetCodeObjectResult TryRetargetCodeObject(const CodeObjectView& code_object
...
+        inflight = std::make_shared<InflightRetarget>();
+        g_inflight.emplace(cache_key, inflight);
+        winner = true;
+      } else {
+        inflight = it->second;
+      }
+    }
+
+    if (winner) {
+      const bool rewritten = PerformRetargetAndCache(
```
harsh-amd: Can the winner completion be RAII? If anything throws after the in-flight entry is published, the waiting threads are never notified.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -73,6 +87,396 @@ std::unordered_map<uint64_t, std::vector<OwnedElfBuffer>> g_retained_rewritten_e
...
+#endif
+}
+
+// Compute the retarget cache key. For file-backed code objects (the common
+// case: HIP modules, RCCL device libs) this uses (uri, file mtime, file size,
+// object size) and never scans the object, avoiding a full hash of a large
+// (~578 MB) ELF on every load. Memory-backed objects (or an unstatable path)
+// fall back to a memoized content hash. HSA_HOTSWAP_CACHE_CONTENT_HASH=1 forces
+// the content hash everywhere.
+uint64_t ComputeRetargetCacheKeyFast(const CodeObjectView& code_object,
```
harsh-amd: This key does not seem safe for an executable cache. A reused allocation can keep the same first/last 4 KiB while changing in the middle, and a same-size file can change within the one-second mtime window. Both cases reuse the old rewritten ELF. Can we compute a strong digest once at ingestion and carry a stable object identity instead?

projects/clr/hipamd/src/hip_fatbin.cpp
```diff
@@ -20,7 +26,53 @@
 #include <rocm_kpack/kpack.h>
 #endif
 
+// Optional ROCr hotswap prewarm entry point. Weak so HIP still loads against an
+// older runtime that does not export it (the pointer is then null). Declared
+// with C linkage and int return (hsa_status_t is an int-sized enum) to avoid
+// pulling HSA headers into this TU.
+extern "C" __attribute__((weak)) int hsa_amd_hotswap_prewarm_code_object(
```
harsh-amd: This breaks the Windows build, and the declaration also drops the `HSA_API` calling convention. Since this needs to remain optional with older ROCr, can we resolve the exact API type dynamically and keep the header/export contract consistent on both platforms?

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -73,6 +87,396 @@ std::unordered_map<uint64_t, std::vector<OwnedElfBuffer>> g_retained_rewritten_e
...
+}
+
+std::string DiskCachePath(const std::string& dir, uint64_t key, uint64_t salt) {
+  char buf[32];
+  std::snprintf(buf, sizeof(buf), "/%016llx.co", static_cast<unsigned long long>(key));
+  return DiskCacheSubdir(dir, salt) + buf;
+}
+
+// Returns true iff the entry was atomically published to disk.
+bool WriteDiskCache(const std::string& dir, uint64_t key, uint64_t salt, const void* data,
```
harsh-amd: This cache is default-on and entries can be hundreds of MB, but there is no byte limit or eviction. Also, a corrupt entry is never invalidated, so strict mode can keep failing indefinitely. Can we add a quota plus invalidation and recomputation on a bad cache hit?

projects/clr/hipamd/src/hip_fatbin.cpp
```diff
@@ -20,7 +26,53 @@
...
+// valid for the retarget's duration.
+void MaybePrewarmHotswap(const void* binary_image, size_t binary_size) {
+  using PrewarmFn = int (*)(const void*, size_t);
+  PrewarmFn fn = hsa_amd_hotswap_prewarm_code_object;  // null if runtime lacks it
+  if (fn == nullptr || !HotswapPrewarmEnabled() || binary_image == nullptr ||
+      binary_size == 0) {
+    return;
+  }
+  try {
+    std::thread([fn, binary_image, binary_size]() { fn(binary_image, binary_size); }).detach();
```
harsh-amd: This detached thread keeps only a raw pointer, but `FatBinaryInfo` can free or unmap it during unload or teardown. Can we use a managed once-per-object worker with shared ownership of the existing storage and join it during teardown? That would also avoid one detached thread per device.

projects/clr/hipamd/src/hip_fatbin.cpp
xintin: It is always enabled. This is a debug var that I was using. 
It will not be in the final code.

harsh-amd:
Could we simplify this by making the prepared object something `FatBinaryInfo` owns, instead of prewarming a global ROCr cache and looking it up again during the real load?

Roughly, could `FatBinaryInfo` retain one preparation future per exact rewrite key?

```cpp
struct RewriteKey {
  Digest source;
  Isa target;
  RewriteOptions options;
  RewriterVersion version;
};

struct PreparedArtifact {
  std::shared_ptr<const MappedFile> mapping;
  CodeObjectView view() const;
};

std::unordered_map<RewriteKey,
                   std::shared_future<PrepareResult>,
                   RewriteKeyHash>
    prepared_;
```

Preparation could use a single persistent artifact store with cross-process locking:

```cpp
PrepareResult ArtifactStore::GetOrCreate(const RewriteKey& key,
                                         CodeObjectView source) {
  FileLock lock(LockPath(key));

  if (auto artifact = OpenAndValidate(ArtifactPath(key), key))
    return artifact;

  auto temporary = RewriteToTemporaryFile(source, key);
  AtomicRename(temporary, ArtifactPath(key));
  return MmapArtifact(ArtifactPath(key));
}
```

Then the normal program construction would select the prepared mapping directly, with the original object as fallback:

```cpp
CodeObjectView image = original.view();
if (auto artifact = GetPreparedArtifact(device))
  image = artifact->view();

program->addDeviceProgram(device, image.data, image.size, ...);
```

This seems like it would remove the prewarm export, weak-symbol handling, detached per-device threads, global payload/negative caches, and the second cache lookup during `onLoad`. Is there a reason the prepared object cannot be bound directly to `FatBinaryInfo` this way?

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
xintin: Fixed by threading a RetargetFailureKind { kNone, kDeterministic, kTransient } out of RetargetCodeObject and only negative-caching kDeterministic. Transient failures now fall through uncached so a later load retries. 
Added a test covering both paths.

xintin:
> Could we simplify this by making the prepared object something `FatBinaryInfo` owns, instead of prewarming a global ROCr cache and looking it up again during the real load?
> 

I think binding the prepared artifact to the object (and feeding it straight into the load) is reasonable. I want to confirm a few things, because they materially change where the code lands:

1. Engine placement / ROCr surface. The retarget engine is currently ROCr-only: amd_comgr_hotswap_rewrite[_with_options] is resolved only in hotswap.cpp (not in CLR's comgrctx), and eligibility (asic_revision == 0) + ELF source-ISA parsing live in ROCr. To "bind to FatBinaryInfo" we either:

(a) move the rewrite + eligibility + a disk artifact store fully into CLR and delete the ROCr hotswap path entirely. This also removes the weak-symbol/dlsym handling by construction (no cross-component symbol at all); or
(b) keep a stateless hsa_amd_hotswap_prepare_artifact(...) export in ROCr (returns an fd/mapping) and move only the ownership/future/bind into CLR; smaller diff, but retains one ROCr export and the same "older runtime lacks the symbol" version-coupling that the current weak symbol exists to tolerate.
**Do you prefer** (a) full move into CLR, or (b) a stateless ROCr prepare API?

2. Coverage. Today the ROCr interception (hsa_executable_load_agent_code_object) transparently covers HIP and OpenCL (both go through rocprogram.cpp), plus any direct-HSA loader. Binding to FatBinaryInfo scopes hotswap to HIP fat binaries only. I can instead put the bind one layer down in roc::Program so HIP + OpenCL both keep it (with FatBinaryInfo still owning the async prewarm futures). Is HIP-only acceptable, or should OpenCL/direct-HSA stay covered?

harsh-amd:
I think these are two separate decisions. For engine placement, I prefer keeping eligibility, rewrite policy, COMGR, and the persistent artifact store in ROCr. Moving everything into CLR would either make this HIP-only or duplicate the hotswap implementation.

For coverage, I do not think we should remove the existing OpenCL/direct-HSA behavior. `FatBinaryInfo` can own the early HIP preparation future, but the prepared result should be consumable at a common loader boundary.

Could the result be represented as an opaque prepared code-object reader instead of returning a raw FD or mapping?

```cpp
hsa_status_t hsa_amd_code_object_reader_prepare(
    hsa_code_object_reader_t source,
    hsa_agent_t agent,
    const RewriteOptions* options,
    hsa_code_object_reader_t* prepared);
```

The caller could manage the asynchronous work, while ROCr owns the persistent mapping and reader lifetime. HIP could start this early from `FatBinaryInfo`, OpenCL could use it through `roc::Program`, and the existing `onLoad` interception could remain as the fallback for direct-HSA users and older runtimes.

The target and options should be explicit rather than selecting the first eligible agent. Would this fit the existing code-object reader ownership model?

xintin:
Thanks. That sounds feasible. 

A couple follow ups:

1. Options shape
For the prepare API, how would you like the rewrite options expressed, a versioned C struct (e.g. hsa_amd_rewrite_options_t with fields like entry_trampolines, strict) or a flags bitmask? Do you have a preference, and are there options beyond entry-trampolines/strict-mode you'd want represented up front?

2. API home + export surface
Where should hsa_amd_code_object_reader_prepare live? Two options I see: (a) the AMD extension surface (hsa_ext_amd.h, exported via hsacore.so.def), or (b) the loader vendor table (hsa_ven_amd_loader_1_xx_pfn_t). This determines where the symbol and its export entry go.

-----

nvm. went ahead with the implementation.

projects/rocr-runtime/rocrtst/suites/functional/hotswap/hotswap_rewrite_test.cc
```diff
@@ -773,8 +858,208 @@ TEST(HotswapRewrite, RuntimeLoadRequiredStrictRewrittenLoadFailureReturnsError)
...
+  rocr::hotswap::ClearRetargetCacheForTesting();
+}
+
+namespace {
+
+size_t CountCacheFilesRecursive(const std::string& dir) {
+  // Cheap recursive count via the shell; the test only needs "at least one".
+  const std::string cmd = "find '" + dir + "' -type f -name '*.co' 2>/dev/null | wc -l";
+  std::FILE* p = ::popen(cmd.c_str(), "r");
+  if (!p) return 0;
```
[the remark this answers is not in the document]
xintin: CountCacheFilesRecursive() using an opendir/readdir walk (added <dirent.h>, <ftw.h>, <sys/stat.h>).

projects/rocr-runtime/rocrtst/suites/functional/hotswap/hotswap_rewrite_test.cc
```diff
@@ -773,8 +858,208 @@ TEST(HotswapRewrite, RuntimeLoadRequiredStrictRewrittenLoadFailureReturnsError)
...
+    ASSERT_EQ(load.calls.size(), 1u);
+    EXPECT_EQ(load.calls[0].path, LoadPath::kRewritten);
+    EXPECT_GT(load.calls[0].code_object_size, 0u);
+    rocr::hotswap::ReleaseRetainedRewrittenElfBuffers(executable);
+  }
+
+  rocr::hotswap::ClearRetargetCacheForTesting();
+  // Best-effort cleanup of the temp cache tree.
+  std::string cmd = "rm -rf '" + cache_dir + "'";
+  (void)std::system(cmd.c_str());
```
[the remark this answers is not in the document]
xintin: in-process RemoveDirRecursive() using nftw(..., FTW_DEPTH | FTW_PHYS)

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa.cpp
```diff
@@ -2310,79 +2198,79 @@ hsa_status_t hsa_executable_load_code_object(
...
+    hsa_code_object_t code_object = {reinterpret_cast<uint64_t>(reader->GetCodeObjectMemory())};
+    return exec->LoadCodeObject(agent, code_object, reader->GetCodeObjectSize(), options,
+                                reader->GetUri(), loaded_code_object);
+  }
+
   hotswap::CodeObjectView code_object;
   code_object.data = reader->GetCodeObjectMemory();
   code_object.size = reader->GetCodeObjectSize();
   code_object.uri = reader->GetUri();
+  code_object.content_digest = reader->GetContentDigest();
```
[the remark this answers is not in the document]
xintin: CodeObjectView now carries a std::function<uint64_t()> content_digest_provider instead of an eagerly-computed uint64_t. The provider is invoked only inside ComputeRetargetCacheKey, which runs after the disabled / eligibility / decision gates (hotswap.cpp). So a disabled or ineligible load no longer pays the O(n) hash, while the compute-once-per-reader caching (multi-GPU) is preserved.

projects/clr/rocclr/device/rocm/rocprogram.cpp
```diff
@@ -219,6 +227,104 @@ bool Program::createKernels(void* binary, size_t binSize, bool useUniformWorkGro
...
+  // Only worth spawning a thread when HotSwap preparation is actually available
+  // and enabled; otherwise setKernels() will create the reader inline for free.
+  if (!device().isOnline()) {
+    return;
+  }
+  auto* prepare_code_object =
+      Device::loaderExtensionTable().hsa_ven_amd_loader_code_object_reader_prepare;
+  if (prepare_code_object == nullptr || !amd::hotswap::Enabled()) {
+    return;
+  }
```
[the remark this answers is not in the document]
xintin: prepareCodeObjectAsync() now returns early if prepareFuture_.valid(), so a second call won't reassign the future and block in the old future's destructor.

projects/rocr-runtime/rocrtst/suites/test_common/test_categories.yaml
```diff
@@ -108,6 +108,16 @@ test_categories:
       - "rocrtstPerf.AQL_Dispatch_Time_Single_Interrupt"
       - "rocrtstPerf.AQL_Dispatch_Time_Multi_SpinWait"
       - "rocrtstPerf.AQL_Dispatch_Time_Multi_Interrupt"
+      - "HotswapRewrite.RetargetCacheServesSecondLoadFromCache"
```
cfreeamd, line 116: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** These new hotswap entries don't line up with the tests actually defined in `hotswap_rewrite_test.cc`, so CI coverage will be wrong on both ends:

Listed here but **not defined** in the test file: `FileUriKeyServesSecondLoadFromCache`, `WarmHotswapCachePopulatesCacheForLoad`, `HotswapPrewarm.ExportPopulatesRetargetCache`, `HotswapPrewarm.ExportRejectsInvalidArgs` (these appear to be the renamed/removed tests referenced in the earlier `NewComgrHotswapApiAvailable()` review thread — the yaml wasn't reconciled).

Defined in the test file but **missing here**, so they never run under any CI filter group: `HotswapRewrite.TransientRetargetFailureIsNotNegativeCached`, `HotswapRewrite.ContentDigestKeysRetargetCache`, `HotswapPrepare.PreparesEligibleObject`, `HotswapPrepare.SecondPrepareServedFromCache`.

Please reconcile both lists (both the Core and Extended blocks) so every added `TEST(...)` has an entry and every entry maps to a real test.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -73,6 +88,411 @@ std::unordered_map<uint64_t, std::vector<OwnedElfBuffer>> g_retained_rewritten_e
...
+
+struct DiskCacheHeader {
+  char magic[8];
+  uint32_t format_version;
+  uint32_t reserved;
+  uint64_t comgr_salt;
+  uint64_t payload_size;
+};
+
+bool IsDiskCacheDisabledByEnv() {
```

**[Question]** The disk cache is default-on (only an explicit false-like `HSA_HOTSWAP_DISK_CACHE` opts out), and `GetDiskCacheDir()` falls back to `$XDG_CACHE_HOME/rocm/hotswap` or `$HOME/.cache/rocm/hotswap`. So any process that loads an eligible (gfx1250) code object now silently writes retargeted ELFs (each potentially hundreds of MB, up to the 20 GiB budget) under the user's home. Is default-on into `$HOME/.cache` the intended out-of-box behavior, or should the persistent tier require an explicit opt-in / an explicit `HSA_HOTSWAP_CACHE_DIR` until this stabilizes? The in-memory tier alone already coalesces the per-GPU loads within a process.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -470,67 +980,269 @@ bool RetargetCodeObject(const void* elf_data, size_t elf_size, const char* sourc
...
+    if (decision->rewrite_required) {
+      return {RetargetCodeObjectStatus::kRequiredRewriteFailed, true};
+    }
+    return {};
+  };
+
+  // Probe the cache, then single-flight the miss so concurrent loads of the
+  // same object (e.g. one per GPU, or a prewarm thread) run COMGR only once.
+  // Losers wait and re-probe the now-populated cache.
+  for (int attempt = 0; attempt < 2; ++attempt) {
```

**[Optional]** The forward-progress fallback after the 2-attempt single-flight loop runs `PerformRetargetAndCache` directly, outside the in-flight coordination. That's correct for liveness, but when the in-memory cache is saturated (`kMaxRetargetCacheEntries == 256`) the winner's result is never cacheable, so every concurrent loser drops through to a full COMGR retarget of the same object — a thundering herd precisely in the many-distinct-objects case the cache is meant to help. Since the disk tier is unbounded by that 256 cap, consider having the fallback re-probe the disk tier once (or let losers coalesce on the disk write) before re-running COMGR.

cfreeamd (commented):

ROCr-scoped pass (Opus). The runtime side of the loader prepare API + two-tier retarget cache reads cleanly: the single-flight coordination (InflightCompletion RAII, predicate-guarded cv.wait), full-content FNV cache key via the lazy content_digest_provider, transient-vs-deterministic negative caching, disk-tier atomic publish + mmap zero-copy read + quota eviction + corrupt-entry invalidation all look correct, and most earlier reviewer concerns appear resolved in the current revision. amd_hsa_loader.cpp (SetOwnedMemory), amd_hsa_loader.hpp (GetContentDigest/IsPrepared), the v1_04 pfn table wiring in hsa.cpp, and hsa_ven_amd_loader.cpp prepare glue look correct. One inline item on test registration is worth fixing before merge; two others are non-blocking.

projects/rocr-runtime/rocrtst/suites/test_common/test_categories.yaml
xintin: Thanks. reconciled both the Core (standard/pr) and Extended (full/weekly) blocks so every entry maps to a real TEST(...) and every added test is registered.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
xintin: Fair point. The in-memory tier already coalesces the per GPU loads within a process, so the disk tier only adds cross process reuse and isn't needed for the multi-GPU. 
This code is part of the persistent cache commit, so it belongs to #8659, not this stacked PR. I'll make the disk tier opt in there (off unless an explicit cache dir is set, in memory coalescing stays on) and follow up on that PR.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
xintin: You're right that the 256 cap only gates the in memory map. PerformRetargetAndCache still writes to the disk tier unconditionally, so the winner's bytes are on disk even when memory is full. So the fallback can do one disk re probe before re-running COMGR and the misses will mmap instead of each retargeting.
Same cache engine code from #8659, so we'll fix it there.

shwetagkhatri (approved):
LGTM

projects/rocr-runtime/rocrtst/suites/functional/hotswap/hotswap_rewrite_test.cc
```diff
@@ -839,4 +839,45 @@ TEST(HotswapRewrite, RuntimeLoadRequiredStrictRewrittenLoadFailureReturnsError)
       rocr::hotswap::RetainedRewrittenElfBufferCountForTesting(executable), 0u);
 }
 
+// PrepareRetargetedCodeObject runs the retarget on the calling thread using
+// explicit options and returns the prepared bytes. This is the engine behind
+// the loader prepare API that FatBinaryInfo/roc::Program drive early.
+TEST(HotswapPrepare, PreparesEligibleObject) {
```
cfreeamd, line 849: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** The two new tests added here — `HotswapPrepare.PreparesEligibleObject` and `HotswapPrepare.SecondPrepareServedFromCache` — are not registered in `test_categories.yaml`, so they won't run in CI under any filter group. The yaml reconcile you described on the earlier revision ("every added test is registered") isn't present on the current head; it looks like it was dropped in the 2026-07-17 squash along with the disk-cache tests that moved to #8659. These two prepare tests are the direct coverage for this PR's headline API, so please re-add them the same way `HotswapRewrite.RetargetCacheServesSecondLoadFromCache` / `RetargetCacheClearResetsCacheSize` are registered.

projects/rocr-runtime/runtime/hsa-runtime/core/inc/amd_hsa_loader.hpp
```diff
@@ -142,16 +143,63 @@ struct CodeObjectReaderImpl final {
...
+  /// already-prepared bytes would be re-analyzed (and, for fixed-stepping
+  /// targets, re-rewritten) since the ELF target-id does not encode the
+  /// stepping feature.
+  bool IsPrepared() const { return is_prepared; }
+
+  /// @brief Full-content digest of this code object, computed once on first
+  /// use and cached. Provides a stable identity that survives across the
+  /// per-agent loads of the same reader (used to key the hotswap retarget
+  /// cache). Returns 0 if the object has no backing memory.
+  uint64_t GetContentDigest() const {
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** `GetContentDigest()` and the `content_digest` atomic member are added here, but nothing on this branch calls them — the consumer (the retarget cache keyed on a full-content digest, via `content_digest_provider` / `ComputeRetargetCacheKey`) lives in the stacked #8659. As landed, this is dead code in #8584. Consider moving it to #8659 alongside its consumer, or confirm it's intentionally landed early. Non-blocking.

(Also: the `SetOwnedMemory` doc mentions the backing deleter carrying "heap free or mmap release" — in this PR `OwnedElfBuffer` is `unique_ptr<void, decltype(&std::free)>`, so only the heap-free deleter is reachable; the munmap variant is part of #8659.)

cfreeamd (commented):
🤖 AI-assisted pre-review (for @cfreeamd) — follow-up after the 2026-07-17 squash

Thanks for the rework. The PR is now a clean, much smaller change: the loader v1_04 prepare API + shared retarget engine refactor + prepared-reader ownership, with the disk cache / single-flight / content-digest keying moved to stacked PR #8659.

Resolved for this PR:
- **Disk cache default-on into $HOME** — the disk-tier code is entirely absent from this diff; validly deferred to #8659.
- **Forward-progress thundering-herd** — same, deferred to #8659.

The prepare-API code reads well: the ownership transfer in `hsa_ven_amd_loader_code_object_reader_prepare` is correct (unique_ptr → shared_ptr backing with the captured deleter, clean failure paths), and `hsa_ven_amd_loader_1_04_pfn_t` is a proper superset-append of 1_03 wired through `get_extension_table_length` and `hsa_system_get_major_extension_table`.

One prior item looks like it regressed in the squash (test_categories.yaml) — see inline. Non-blocking observation on `GetContentDigest()` also inline.

projects/rocr-runtime/runtime/hsa-runtime/inc/hsa_ven_amd_loader.h
```diff
@@ -527,10 +527,83 @@ hsa_ven_amd_loader_iterate_executables(
...
+ * @retval ::HSA_STATUS_ERROR_INVALID_CODE_OBJECT A required rewrite failed or
+ * the prepared object is invalid.
+ *
+ * @retval ::HSA_STATUS_ERROR_INVALID_ARGUMENT @p prepared_code_object_reader is
+ * NULL.
+ *
+ * @retval ::HSA_STATUS_ERROR_OUT_OF_RESOURCES The HSA runtime failed to
+ * allocate the required resources.
+ */
+hsa_status_t hsa_ven_amd_loader_code_object_reader_prepare(
```
atgutier, line 597: If this is advertised on the public API, it'll need to support all agents. At the moment it seems like a caller would have to use trial and error to see if the agent supports this feature by checking an error?

Typically, we have agent attributes to understand specialized behavior, but if we add that, this starts to creep to more places on the API surface.

Is there any way you can achieve the desired outcome without creating a new public API so this is hidden and done only for B0/A0 stepping?

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa_ven_amd_loader.cpp
```diff
@@ -301,4 +305,76 @@ hsa_ven_amd_loader_iterate_executables(
...
+
+    hotswap::OwnedElfBuffer prepared_bytes(nullptr, &std::free);
+    size_t prepared_size = 0;
+    const hotswap::PrepareStatus status =
+        hotswap::PrepareRetargetedCodeObject(view, agent, options, &prepared_bytes, &prepared_size);
+
+    switch (status) {
+      case hotswap::PrepareStatus::kNotNeeded:
+        // No preparation required; caller loads the source reader as-is.
+        return HSA_STATUS_SUCCESS;
```
atgutier, line 344: This leaves the capabilities of this API ambiguous. A caller will see success but will have no effect for some agents, which seems to be desired, but the semantics are not very clear here. A user will not have an understanding of when or why to use this and which agents can be supported. Likely would have follow-on or unexpected effects in the future.

See my prior comments about the need for agent attribute.

atgutier (commented):
Most of the implementation looks ok here, but I have concerns about the public API and how this will be supported overall for arbitrary agents.

PR #8585, opened 2026-07-14
[rocjitsu] Allow CDNA4 DBT guests on simulated CDNA3 target

kuhar:
## Motivation

This is so that we can develop and test DBT on systems without the target HW. It's also surprisingly good for testing both the DBT (since the simulator is often less forgiving than the real hw) and the simulator (since DBT often exercises rare ASM patterns).

Add a simulator-backed execution mode for DBT guests, allowing CDNA4 (gfx950) code to be translated to CDNA3 (gfx942) and executed without physical CDNA3 hardware.

## Technical Details

- Add hardware and simulator DBT execution backends, including simulator-config validation and relative path resolution.
- Connect GuestKfd to SimulatedKfd, exposing a synthetic gfx950 guest while executing translated code on simulated gfx942.
- Correctly manage the simulator’s bootstrap KFD open across guest discovery, application opens, closes, and teardown.
- Add clean failure handling for missing simulator configurations.
- Register simulator-backed CDNA4->CDNA3 tests by default while retaining hardware-backed tests when a CDNA3 GPU is detected.
- Add end-to-end coverage through the public DBT guest path for dynamic copy, Triton matmul, and flash-attention kernels

## JIRA ID

N/A

## Test Plan

New unit and e2e tests. I also tested this on many kernel libraries and iree on the gx1250->rdna4 dbt path.

## Test Result

All good.

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/configs/guest_gfx950_on_simulated_gfx942.json
+++ b/emulation/rocjitsu/configs/guest_gfx950_on_simulated_gfx942.json
@@ -24,7 +24,7 @@
       "wave_front_size": 64,
       "max_slots_scratch_cu": 32,
       "local_mem_size": 309237645312,
-      "lds_size_kb": 160,
+      "lds_size_kb": 64,
       "mem_width": 8192,
       "mem_clk_max": 1600,
       "l1_size_kb": 32,
--- a/emulation/rocjitsu/configs/guest_gfx950_on_simulated_gfx942.json
+++ b/emulation/rocjitsu/configs/guest_gfx950_on_simulated_gfx942.json
@@ -1,0 +1,42 @@
+{
+  "dbt_guest": {
+    "enabled": true,
+    "guest_isa": "gfx950",
+    "host_isa": "gfx942",
+    "host_gpu_id": 0,
+    "execution_backend": "simulator",
+    "simulator_config": "gfx942_cdna3_kmd.json",
+    "guest_device": {
+      "gpu_id": 38144,
+      "gfx_target_version": 90500,
+      "vendor_id": 4098,
+      "device_id": 30112,
+      "family_id": 160,
+      "unique_id": 5929628898254127105,
+      "marketing_name": "AMD Instinct MI350X",
+      "drm_render_minor": 191,
+      "simd_count": 64,
+      "max_waves_per_simd": 8,
+      "num_shader_engines": 4,
+      "num_shader_arrays_per_engine": 2,
+      "num_cu_per_sh": 4,
+      "simd_per_cu": 4,
+      "wave_front_size": 64,
+      "max_slots_scratch_cu": 32,
+      "local_mem_size": 309237645312,
+      "lds_size_kb": 160,
+      "mem_width": 8192,
+      "mem_clk_max": 1600,
+      "l1_size_kb": 32,
+      "l1_line_size": 128,
+      "l1_assoc": 4,
+      "l2_size_kb": 4096,
+      "l2_line_size": 128,
+      "l2_assoc": 16,
+      "num_sdma_engines": 5,
+      "num_sdma_xgmi_engines": 12,
+      "num_cp_queues": 128,
+      "max_engine_clk_fcompute": 2400
+    }
+  }
+}
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/dbt_guest_config.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/dbt_guest_config.cpp
@@ -21,16 +21,14 @@ namespace rocjitsu {
 namespace config {
 namespace {
 
-DbtExecutionBackend parse_execution_backend(const fb::DbtGuestConfig *guest) {
-  if (!guest->execution_backend())
+DbtExecutionBackend execution_backend_from_fb(fb::DbtExecutionBackend backend) {
+  switch (backend) {
+  case fb::DbtExecutionBackend_hardware:
     return DbtExecutionBackend::Hardware;
-  const std::string value = guest->execution_backend()->str();
-  if (value == "hardware")
-    return DbtExecutionBackend::Hardware;
-  if (value == "simulator")
+  case fb::DbtExecutionBackend_simulator:
     return DbtExecutionBackend::Simulator;
-  throw std::runtime_error("dbt_guest.execution_backend must be 'hardware' or 'simulator', got '" +
-                           value + "'");
+  }
+  throw std::runtime_error("dbt_guest.execution_backend is invalid");
 }
 
 void validate_guest_device_geometry(const KfdDeviceConfig &device) {
@@ -53,27 +51,9 @@ void validate_guest_device_geometry(const KfdDeviceConfig &device) {
 
 } // namespace
 
-const char *dbt_execution_backend_name(DbtExecutionBackend backend) {
-  switch (backend) {
-  case DbtExecutionBackend::Hardware:
-    return "hardware";
-  case DbtExecutionBackend::Simulator:
-    return "simulator";
-  }
-  return "unknown";
-}
-
-std::string resolve_dbt_simulator_config_path(const std::string &dbt_config_path,
-                                              const std::string &simulator_config) {
-  std::filesystem::path resolved(simulator_config);
-  if (resolved.is_relative())
-    resolved = std::filesystem::path(dbt_config_path).parent_path() / resolved;
-  return resolved.lexically_normal().string();
-}
-
 void validate_dbt_simulator_device_limits(const DbtGuestConfig &guest,
                                           const KfdDeviceConfig &simulator_device) {
-  if (!guest.enabled || guest.execution_backend != DbtExecutionBackend::Simulator)
+  if (!guest.enabled || guest.host.backend != DbtExecutionBackend::Simulator)
     return;
   if (!guest.guest_device.present || !simulator_device.present)
     throw std::runtime_error("simulator-backed dbt_guest requires guest and simulator devices");
@@ -107,23 +87,31 @@ DbtGuestConfig dbt_guest_from_fb(const fb::DbtGuestConfig *guest) {
   if (guest->guest_isa())
     config.guest_isa = guest->guest_isa()->str();
   if (guest->host_isa())
-    config.host_isa = guest->host_isa()->str();
-  config.host_gpu_id = guest->host_gpu_id();
-  config.execution_backend = parse_execution_backend(guest);
+    config.host.isa = guest->host_isa()->str();
+  config.host.gpu_id = guest->host_gpu_id();
+  config.host.backend = execution_backend_from_fb(guest->execution_backend());
   if (guest->simulator_config())
-    config.simulator_config = guest->simulator_config()->str();
+    config.host.simulator_config_path = guest->simulator_config()->str();
   config.log_level = guest->log_level();
   config.signal_backtrace = guest->signal_backtrace();
   config.guest_device = kfd_device_from_fb(guest->guest_device());
   validate_guest_device_geometry(config.guest_device);
-  if (config.enabled && config.execution_backend == DbtExecutionBackend::Simulator &&
-      config.simulator_config.empty())
-    throw std::runtime_error(
-        "dbt_guest.simulator_config is required when execution_backend is 'simulator'");
-  if (config.enabled && config.execution_backend == DbtExecutionBackend::Hardware &&
-      !config.simulator_config.empty())
-    throw std::runtime_error("dbt_guest.simulator_config requires execution_backend 'simulator'");
+  if (config.enabled && config.host.backend == DbtExecutionBackend::Hardware &&
+      !config.host.simulator_config_path.empty())
+    throw std::runtime_error("dbt_guest.simulator_config requires execution_backend=\"simulator\"");
   return config;
+}
+
+std::string resolve_dbt_host_config_path(const std::string &dbt_config_path,
+                                         const std::string &host_config_path) {
+  const std::filesystem::path dbt_path(dbt_config_path);
+  if (host_config_path.empty())
+    return dbt_path.lexically_normal().string();
+
+  const std::filesystem::path host_path(host_config_path);
+  if (host_path.is_absolute())
+    return host_path.lexically_normal().string();
+  return (dbt_path.parent_path() / host_path).lexically_normal().string();
 }
 
 DbtGuestConfig load_dbt_guest_config_from_file(const std::string &path) {
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/dbt_guest_config.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/dbt_guest_config.cpp
@@ -11,6 +11,7 @@
 #include "simulation_config_generated.h"
 
 #include <cstdlib>
+#include <filesystem>
 #include <fstream>
 #include <optional>
 #include <stdexcept>
@@ -19,6 +20,17 @@
 namespace rocjitsu {
 namespace config {
 namespace {
+
+DbtExecutionBackend parse_execution_backend(const fb::DbtGuestConfig *guest) {
+  const std::string value =
+      guest->execution_backend() ? guest->execution_backend()->str() : std::string("hardware");
+  if (value.empty() || value == "hardware")
+    return DbtExecutionBackend::Hardware;
+  if (value == "simulator")
+    return DbtExecutionBackend::Simulator;
+  throw std::runtime_error("dbt_guest.execution_backend must be 'hardware' or 'simulator', got '" +
+                           value + "'");
+}
 
 void validate_guest_device_geometry(const KfdDeviceConfig &device) {
   if (!device.present || device.simd_count == 0)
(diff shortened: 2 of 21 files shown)


emulation/rocjitsu/configs/guest_gfx950_on_nonexistent_simulator_config.json
```diff
@@ -0,0 +1,9 @@
+{
```
atgutier: It's not immediately obviously to me why this is called "missing".

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/dbt_guest_config.cpp
```diff
@@ -20,6 +21,17 @@ namespace rocjitsu {
 namespace config {
 namespace {
 
+DbtExecutionBackend parse_execution_backend(const fb::DbtGuestConfig *guest) {
+  const std::string value =
+      guest->execution_backend() ? guest->execution_backend()->str() : std::string("hardware");
+  if (value.empty() || value == "hardware")
```
atgutier: Should we allow empty to fallback or should we fault?

atgutier (commented):
A few questions from me and some issues from the agent:

 BLOCKER

  - Use-after-free on the discovery-failure path — interposer.cpp:806-811. For the simulator backend,
  execution_driver points into the VM (rj_vm_->vm->driver()). When prepare_for_discovery() fails, the
  code calls rj_vm_destroy(rj_vm_) (deletes the VM → deletes the SimulatedKfd) and then return nullptr
   destroys the local guest_driver, whose ~GuestKfd sees the flag true and calls
  execution_driver_->close() on the already-freed SimulatedKfd. Reachable whenever create_local_vm
  succeeds but discovery fails (e.g. host_isa absent from the sim topology, or overlay generation
  failure) — and untested (MissingSimulatorConfigFailsCleanly fails earlier, before GuestKfd is
  constructed). Fix: guest_driver.reset() before rj_vm_destroy so ~GuestKfd closes while the VM is
  still alive; and set owns_execution_driver_open_ only on a real acquired open.

  MAJOR

  - dup2 over the hidden fd tears down simulator process state + desyncs the ownership flag —
  guest_kfd.cpp:676-702. For the sim backend, GuestKfd's hidden real_kfd_fd_ is the same fd number as
  execution_driver_->fd(). When an app dup2/dup3s over it, invalidate_primary_fd calls
  execution_driver_->close(), dropping the sim's only KfdProcess reference → the process is destroyed
  (allocations freed, queues unregistered). Two consequences: (1) the next ensure_ready() re-opens the
   sim with a fresh empty KfdProcess, silently discarding all guest execution state — the "dups keep
  working transparently" contract is true for real /dev/kfd but false for the in-process sim; (2)
  owns_execution_driver_open_ is left true, so the later real close() calls execution_driver_->close()
   a second time — currently non-fatal only because SimulatedKfd::close no-ops on a missing pid, a
  fragile accident. Untested (the fd-reuse test requires real /dev/kfd and skips otherwise). Fix: for
  the sim backend, don't release the sim ref in invalidate_primary_fd (treat the sim fd like
  GuestKfd's own hidden fd, kept alive by app dups); keep the ownership flag balanced. Add a
  sim-backed dup2 test.

 - DBT-guest e2e tests assert only finiteness, not correctness —
  cdna4_to_cdna3_dispatch_test.cpp:1015-1016, 1037-1038, 1059-1060. This is the crux: the PR's
  headline feature is executing translated CDNA4 kernels on a simulated CDNA3 target through the
  public DBT-guest path, but the guest flash-attention and buffer-async-matmul tests only check
  std::isfinite(v) && v != -12345.0f — a translation/dispatch bug producing plausible-but-wrong finite
   numbers passes silently. The native-path siblings (Cdna4ToCdna3DispatchTest.*) do proper
  element-wise comparison (expect_float_vectors_near(..., 0.02f)). Only the dynamic-matmul guest case
  actually compares against a CPU reference. Fix: give the guest tests the same bar — compare to the
  CPU reference (extend reference_* to flash-attention) or run the native kernel and compare
  guest-vs-native with tolerance.

  MINOR

  - owns_execution_driver_open_ asserted at construction rather than on acquire/release (the root
  cause above) — tightening this single invariant resolves the MINOR and removes the enabler for both
  the BLOCKER and MAJOR.
  - Misspelled execution_backend in JSON silently defaults to hardware (skip_unexpected_fields_in_json
   = true) — pre-existing parser behavior, but the new required-field coupling makes the failure more
  surprising; consider louder validation.
 - Unvalidated guest-vs-sim LDS mismatch — guest_gfx950_on_simulated_gfx942.json:26. Guest
  advertises lds_size_kb: 160 (gfx950) but the execution target is 64 KiB (gfx942). ROCr admits
  dispatches against the guest's 160 KiB, so a translated kernel using 65-160 KiB LDS is admitted but
  can't fit on the target → mis-dispatch/corruption with no diagnostic. The test author knows (a
  harness-side ASSERT_LE(shared_bytes, 65536) + comment), but nothing in production validates it. Fix:
   when the sim backend is selected, validate guest execution-relevant limits (LDS, scratch, VGPR) ≤
  the resolved simulator target's capacity, failing config load with a clear message.
- Unbounded open() retry loop — guest_kfd.cpp:633-665. for(;;) re-calls ensure_ready() until ready_
   && real_kfd_fd_ >= 0; a concurrent close()/invalidate flipping ready_ in the window re-runs the
  loop with no iteration cap and no failure exit. Converges in practice but no progress guarantee.
  Fix: bound the retry (return -1/EAGAIN) or make ready+dup atomic under a single lock acquisition.

emulation/rocjitsu/configs/guest_gfx950_on_nonexistent_simulator_config.json
kuhar: /s/missing/nonexistent/

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/dbt_guest_config.cpp
kuhar: I made it default to hardware when not specified but not allow empty strings as values.

kuhar:
@atgutier should be addressed now

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/dbt_guest_config.h
```diff
@@ -19,22 +19,51 @@ struct DbtGuestConfig;
...
+/// selected hardware or simulated host agent.
 struct DbtGuestConfig {
-  bool enabled = false;          ///< True when GuestKfd mode is active.
-  std::string guest_isa;         ///< Guest ISA advertised by the synthetic agent.
-  std::string host_isa;          ///< Host ISA used for actual ROCR execution.
-  uint32_t host_gpu_id = 0;      ///< Host KFD topology gpu_id; 0 matches topology to host_isa.
-  int log_level = 0;             ///< DBT hook logging level loaded from the config file.
-  bool signal_backtrace = false; ///< Install a best-effort HSA-hook crash backtrace handler.
-  KfdDeviceConfig guest_device;  ///< Synthetic guest device appended to KFD topology.
+  bool enabled = false;     ///< True when GuestKfd mode is active.
```
Groverkss: nit: these look a bit unaligned, but ok if this is what clang-format wants.

emulation/rocjitsu/tests/dbt/cdna4_to_cdna3_dispatch_test.cpp
```diff
@@ -882,19 +1023,120 @@ TEST(Cdna4ToCdna3DispatchTest, DynamicCopyLoopDispatchAndRun) {
...
+  std::vector<float> observed;
+  std::vector<float> expected;
+  ASSERT_NO_FATAL_FAILURE(run_flash_attention_triton(
+      guest_elf, guest, "flash_attention_fwd_async_buffer_kernel.kd", 65536, &observed, &expected));
+  expect_float_vectors_near(observed, expected, 0.02f, "guest DBT flash attention buffer-async");
+}
+
 TEST(Cdna4ToCdna3DispatchTest, TritonDynamicMatmulDispatchAndRun) {
-  const std::array<TritonMatmulCase, 11> cases = {{
+  const std::array<TritonMatmulCase, 4> cases = {{
```
Groverkss: are we remoiving a bunch of these cases?

Groverkss: are we remoiving a bunch of these cases?

emulation/rocjitsu/tests/dbt/cdna4_to_cdna3_dispatch_test.cpp
```diff
@@ -752,9 +877,25 @@ void run_flash_attention_triton(const std::vector<uint8_t> &elf_bytes, const Cdn
   std::vector<uint16_t> v_host(kKVElements);
   std::vector<float> o_init(kOElements, -12345.0f);
   std::vector<float> o_host(kOElements);
-  fill_seeded_half_inputs(q_host, 0xF1A5'0001u);
-  fill_seeded_half_inputs(k_host, 0xF1A5'0002u);
-  fill_seeded_half_inputs(v_host, 0xF1A5'0003u);
+  if (reference) {
+    // Zero Q/K makes every attention score equal. Repeating one V row across
```
Groverkss: I don't think we should use zero Q/K. This is not a good representative input.

Groverkss: I don't think we should use zero Q/K. This is not a good representative input. I have had cases where my agent also tried this and didnt catch a bunch of bugs.

emulation/rocjitsu/tests/dbt/cdna4_to_cdna3_dispatch_test.cpp
Groverkss: The higher dimension cases are important and do catch some bugs.

Groverkss (commented):
I think this needs a docs/ update also in the guest mode document.

emulation/rocjitsu/tests/dbt/cdna4_to_cdna3_dispatch_test.cpp
kuhar: They take a bit long to execute on the simulator, not sure if they meaningfully contribute to the test coverage. We could bifurcate if you want to keep them for the hw path.

kuhar: Discussed offline, I added a new testcase with k=128 and kept the other ones dropped

kuhar: See https://github.com/ROCm/rocm-systems/pull/8585#discussion_r3588554131

emulation/rocjitsu/configs/guest_gfx950_on_simulated_gfx942.json
amd-arosa, line 1: could you add support for mirage for this?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/dbt_guest_config.h
```diff
@@ -19,22 +19,51 @@ struct DbtGuestConfig;
...
+  std::string guest_isa;    ///< Guest ISA advertised by the synthetic agent.
+  std::string host_isa;     ///< Host ISA used for actual ROCR execution.
```
amd-arosa: that is host config.

amd-arosa: move to a DbtHostConfig struct?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/guest_kfd.cpp
```diff
@@ -361,8 +369,8 @@ class GuestKfd::TopologyOverlay {
   /// @brief Copy a host sysfs subtree into the generated overlay.
   bool copy_tree(const std::string &src, const std::string &dst);
 
-  /// @brief Copy the first available real KFD topology root into the overlay.
-  bool copy_host_topology();
+  /// @brief Copy a simulated topology or the first available real KFD topology.
+  bool copy_host_topology(const std::string &host_topology);
```
amd-arosa: this is a bit unclear what this arg should have. is it path or json?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/guest_kfd.h
```diff
@@ -99,8 +104,10 @@ class GuestKfd : public LinuxKfd {
   /// close are not routed to whatever now occupies the number. The real fd is NOT
   /// counted in open_refs_ (only app-facing dups are), so a match returns
   /// kClearedKeepRefs — the interposer must clear the classification WITHOUT
-  /// dropping an open reference. Returns kNotPrimary if @p fd is not the current
-  /// hidden real fd.
+  /// dropping an open reference. For simulator execution, the separately owned
```
amd-arosa, line 107: added a helper that stores a fd and auto closes when dropped. might be useful here?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/linux_kfd.h
```diff
@@ -33,9 +33,10 @@ inline constexpr uint64_t kfd_mmap_gpu_id(uint32_t gpu_id) {
 ///
 /// @details The base Driver interface handles /dev/kfd calls. The Linux shim
 /// also needs sysfs redirection, DRM render-node ownership, and fd tracking.
-/// SimulatedKfd and GuestKfd both implement this surface, but with
-/// different ownership rules: simulation owns all visible GPU discovery, while
-/// GuestKfd owns only the appended guest GPU.
+/// SimulatedKfd and GuestKfd both implement this surface, but with different
+/// ownership rules: simulation owns all visible GPU discovery, while GuestKfd
+/// owns the appended guest and delegates host operations to either hardware or
+/// a SimulatedKfd execution backend.
```
amd-arosa, line 39: nit: only edit the line that changed?

emulation/rocjitsu/schemas/simulation_config.fbs
```diff
@@ -133,6 +133,9 @@ table DbtGuestConfig {
   log_level: int32 = 0;           // DBT hook logging: 0 off, 1 info, 2 verbose, 3 debug.
   signal_backtrace: bool = false; // Install a best-effort crash backtrace handler in the HSA hook.
   guest_device: KfdDeviceInfo;    // Synthetic guest KFD/sysfs identity appended to host topology.
+  // Keep new fields after the original table members to preserve FlatBuffers field IDs.
```
amd-arosa: this comment is silly

emulation/rocjitsu/schemas/simulation_config.fbs
```diff
@@ -133,6 +133,9 @@ table DbtGuestConfig {
...
+  execution_backend: string;      // "hardware" or "simulator"; omitted means hardware.
```
amd-arosa: does this have to be a string? can it be enum?

emulation/rocjitsu/schemas/simulation_config.fbs
```diff
@@ -133,6 +133,9 @@ table DbtGuestConfig {
...
+  simulator_config: string;       // Simulator VM config, relative to this file when not absolute.
```
amd-arosa: i would prefer the files be flat and verbose.
canonical even.
like not requiring external files.
that way it deterministic and unambiguous.
mirage cli handles creating the verbose files

amd-arosa (changes requested):
really cool!
poke me on teams if you want me to review again.

edit: didn't mean to block.

emulation/rocjitsu/schemas/simulation_config.fbs
amd-arosa: this is comment is for line 139.

emulation/rocjitsu/configs/guest_gfx950_on_simulated_gfx942.json
amd-arosa: lmk if you want me to do the mirage code.

amd-arosa (commented):
very good. two changes that are not nits:
1. make config work as single file.
2. add mirage cli support (i can do if you want)

emulation/rocjitsu/configs/guest_gfx950_on_simulated_gfx942.json
kuhar: I think I'd want to land that separately in another PR

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/guest_kfd.h
kuhar: I kept it as-is, those descriptors must be released through `LinuxKfd::close()` to maintain simulated-driver reference bookkeeping, so a raw auto-closing fd wrapper would be incorrect there.

emulation/rocjitsu/schemas/simulation_config.fbs
kuhar: I allowed both since I don't want simulator configs to diverge... I'd take that extra upfront complexity for less future maintanance.

emulation/rocjitsu/schemas/simulation_config.fbs
```diff
@@ -133,6 +138,8 @@ table DbtGuestConfig {
   log_level: int32 = 0;           // DBT hook logging: 0 off, 1 info, 2 verbose, 3 debug.
   signal_backtrace: bool = false; // Install a best-effort crash backtrace handler in the HSA hook.
   guest_device: KfdDeviceInfo;    // Synthetic guest KFD/sysfs identity appended to host topology.
+  execution_backend: DbtExecutionBackend = hardware;
```
amd-arosa, line 141: document that if this is set that it overrides simulator_config
or throw error if both or set.

amd-arosa (approved):
LGTM

atgutier (approved):

PR #8587, opened 2026-07-14
feat(rocr): preserve metadata through intercept queues

bwelton:
## Summary
Preserve AMD metadata-prefetch packets when HSA queues are wrapped by `InterceptQueue`, so profiler/interceptor paths do not silently drop metadata needed by supported hardware.

## Changes

- `projects/rocr-runtime/runtime/hsa-runtime/core/inc/intercept_queue.h`
  - Add paired AQL/metadata storage for overflow and pre-interceptor snapshots.
  - Add proxy metadata-ring state and the host-only test construction seam.
- `projects/rocr-runtime/runtime/hsa-runtime/core/runtime/intercept_queue.cpp`
  - Reuse the public `AqlMetadataPrefetchPacket` representation.
  - Redirect metadata-ring queries to the intercept queue's proxy ring.
  - Match rewritten packets to original metadata using exact completion-signal matching followed by signal-agnostic fallback.
  - Preserve metadata through insertion, deletion, reorder, replacement, overflow, and retry-barrier paths.
  - Update event IDs for standard and AMD vendor dispatch/barrier packets.
  - Reset consumed proxy metadata slots to prevent stale metadata after ring reuse.
  - Forward metadata version and queue-property queries.
  - Derive the masked-packet-match body length from `sizeof(AqlPacket)` instead of a hardcoded literal, with a `static_assert` to catch future layout changes.
- `projects/rocr-runtime/rocrtst/suites/functional/intercept_queue_metadata_test.cc`
  - Add 23 host-only GTest cases covering metadata preservation, packet transformations, event IDs, vendor packets, wraparound, stale-slot reset, overflow/retry, unsupported metadata, null signals, and intercept markers.
- `projects/rocr-runtime/rocrtst/suites/test_common/CMakeLists.txt`
  - Add the standalone test target, exclude it from the aggregate source list, and register/install it with CTest.

## Design

- No new public HSA API or ABI-table changes.
- Uses the current public `core::AqlMetadataPrefetchPacket` type; no duplicate metadata definitions are added.
- Metadata is written before the wrapped AQL packet header is published.
- Unmatched inserted/replaced packets receive four-invalid-header no-op metadata.

## Testing

- `clang-format -style=file --dry-run --Werror` on all changed C++ files: clean.
- Standalone `intercept_queue_metadata` target configured and built with `cmake`/`cmake --build`.
- `ctest -R '^intercept_queue_metadata$' -j1 --output-on-failure`: 1/1 passed.
- Direct GTest execution: 23/23 tests passed.
- Rebased onto current `origin/develop`; `git diff origin/develop...HEAD --check` is clean.

## Limitations

- Full ROCR runtime build/install and GPU-backed validation were not available in the development environment. This PR is intended for review of the implementation and host-side coverage before device validation.
- `BARRIER_AND`/`BARRIER_OR` event-ID propagation and the ambiguous-duplicate-packet matching path are not yet covered by a dedicated regression test; tracked as follow-up work.
- Tests are single-threaded; concurrent `StoreRelaxed` interaction with the new proxy-ring state is covered by production locking but not exercised by this test suite.

## Tracking
JIRA ID : AIPROFSDK-976

### DIFF
--- a/projects/rocr-runtime/rocrtst/suites/functional/intercept_queue_metadata_test.cc
+++ b/projects/rocr-runtime/rocrtst/suites/functional/intercept_queue_metadata_test.cc
@@ -1,0 +1,839 @@
+//===- intercept_queue_metadata_test.cc - InterceptQueue metadata tests --===//
+//
+// SPDX-License-Identifier: NCSA
+//
+//===----------------------------------------------------------------------===//
+
+#include <cstdlib>
+#include <cstring>
+#include <memory>
+#include <vector>
+
+#include "core/inc/intercept_queue.h"
+#include "core/util/atomic_helpers.h"
+#include "gtest/gtest.h"
+#include "inc/amd_hsa_signal.h"
+
+namespace rocr {
+namespace core {
+
+// The host-only target constructs DoorbellSignal without the runtime signal
+// implementation. It never exposes or destroys that signal through HSA.
+Signal::~Signal() {}
+void Signal::registerIpc() {}
+void SharedSignalPool_t::clear() {}
+SharedSignal* SharedSignalPool_t::alloc() { return nullptr; }
+void SharedSignalPool_t::free(SharedSignal*) {}
+
+}  // namespace core
+}  // namespace rocr
+
+namespace {
+
+using AqlPacket = rocr::core::AqlPacket;
+using MetadataPacket = rocr::core::AqlMetadataPrefetchPacket;
+using Queue = rocr::core::Queue;
+using SharedQueue = rocr::core::SharedQueue;
+
+constexpr uint32_t kQueueSize = 8;
+
+class QueueStorage {
+ protected:
+  SharedQueue storage_{};
+};
+
+class FakeQueue final : private QueueStorage, public Queue {
+ public:
+  explicit FakeQueue(bool metadata_supported = true)
+      : Queue(&storage_, 0, nullptr), metadata_supported_(metadata_supported) {
+    ring_.resize(kQueueSize);
+    metadata_ring_.resize(kQueueSize);
+    amd_queue_.hsa_queue.size = kQueueSize;
+    amd_queue_.hsa_queue.base_address = ring_.data();
+    for (auto& metadata : metadata_ring_) metadata = NoopMetadata();
+  }
+
+  hsa_status_t Inactivate() override { return HSA_STATUS_SUCCESS; }
+  hsa_status_t SetPriority(rocr::HSA::hsa_amd_queue_priority_internal_t) override {
+    return HSA_STATUS_SUCCESS;
+  }
+  uint64_t LoadReadIndexAcquire() override { return read_index_; }
+  uint64_t LoadReadIndexRelaxed() override { return read_index_; }
+  uint64_t LoadWriteIndexAcquire() override { return write_index_; }
+  uint64_t LoadWriteIndexRelaxed() override { return write_index_; }
+  void StoreReadIndexRelaxed(uint64_t value) override { read_index_ = value; }
+  void StoreReadIndexRelease(uint64_t value) override { read_index_ = value; }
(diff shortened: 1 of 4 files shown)


bwelton:
Replaces #8521 (closed: non-compliant branch name, non-Conventional-Commits title, and missing tracking reference).

projects/rocr-runtime/rocrtst/suites/functional/intercept_queue_metadata_test.cc
```diff
@@ -0,0 +1,839 @@
+//===- intercept_queue_metadata_test.cc - InterceptQueue metadata tests --===//
+//
+// SPDX-License-Identifier: NCSA
```
cfreeamd, line 3: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** Wrong SPDX identifier — `NCSA` should be `MIT`.

The University of Illinois/NCSA Open Source License is distinct from MIT — it adds a non-endorsement clause. Every other file in `rocrtst/suites/functional/` carries AMD's standard:
```cpp
//
```
Needs to match before merge for IP compliance.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/intercept_queue.cpp
```diff
@@ -200,17 +339,48 @@ void InterceptQueue::Submit(const void* pkts, uint64_t pkt_count, uint64_t user_
...
   const AqlPacket* packets = (const AqlPacket*)pkts;
 
+  const MetadataPacket* metadata = nullptr;
+  if (queue->metadata_ring_buf_) {
+    queue->metadata_scratch_.resize(pkt_count);
+    for (uint64_t i = 0; i < pkt_count; ++i) {
+      OriginalPacketInfo* match = nullptr;
+      for (auto& original : queue->originals_) {
+        if (!original.matched && MaskedPacketMatch(packets[i], original.aql) &&
+            CompletionSignalMatch(packets[i], original.aql)) {
```
cfreeamd, line 349: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** O(n²) matching is fine for current use, but worth documenting.

The double loop (`for i in submitted_count: for original in originals_`) is O(submitted × originals). Both are bounded by the queue ring size, so this isn't a practical concern. A short comment — e.g., `// O(queue_size²) is acceptable: both vectors are bounded by the ring size` — would prevent future reviewers from treating this as a performance gap.

projects/rocr-runtime/rocrtst/suites/functional/intercept_queue_metadata_test.cc
```diff
@@ -0,0 +1,839 @@
...
+}
+
+void ExpectNoopMetadata(const MetadataPacket& metadata) {
+  ExpectMetadataEq(metadata, NoopMetadata());
+}
+
+AqlPacket* ProxyRing(TestInterceptQueue& queue) {
+  return reinterpret_cast<AqlPacket*>(queue.amd_queue_.hsa_queue.base_address);
+}
+
```
cfreeamd, line 253: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** Module-level mutable globals for marker state.

`marker_callback_fired` and `marker_packet_index` are module-level globals reset by `MarkerPacketsDoNotConsumeMetadataSlots` before use. If gtest runs tests in shuffled order or future tests fire markers without resetting, stale state can cause false passes. Consider making these `InterceptQueueMetadataTest` members or a local struct passed via the interceptor `data` pointer.

cfreeamd (commented):
🤖 AI-assisted pre-review (for @cfreeamd) — `intercept_queue.h` / `intercept_queue.cpp` changes look structurally sound; the metadata opt-in path and `ROCR_INTERCEPT_QUEUE_TESTING` seams are well-isolated. CMakeLists.txt registration is correct (standalone binary, not in the rocrtst yaml system).

PR #6942, opened 2026-06-08
fix(rocr): release GWS queue in ReleaseResources to fix shutdown UAF

ihhethan:
## Motivation
Fix a use-after-free SIGSEGV at process exit during rocprofv3 counter collection.

## Technical Details
The GWS-access AQL queue (gws_queue_.queue_) is a GpuAgent member released only by
~GpuAgent's automatic destruction, which runs after SharedSignalPool has been cleared.
Its ~AqlQueue then stores to an already-freed queue_inactive_signal, a use-after-free
at process exit. Counter collection keeps this queue alive until shutdown, exposing the
bug. Fix: release it in GpuAgent::ReleaseResources(), while the signal pool and async
handler are still alive, so ~AqlQueue tears down safely.

## JIRA ID
JIRA ID : ROCM-10675

## Test Plan
Rebuilt libhsa-runtime64 and ran `rocprofv3 --pmc SQ_WAVES -d out --kernel-trace -- <workload>`,
stock vs. fixed library.

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #8588, opened 2026-07-14
fix(rocr): Destroy GWS queue before agent resource teardown

tsrw2048:
## Summary
Destroy the persistent GWS/cooperative queue during `GpuAgent::ReleaseResources()` instead of leaving it to be destroyed later during agent teardown.

This prevents a shutdown-time use-after-free when a cooperative queue survives until `hsa_shut_down()`. In that case, `Runtime::Unload()` may clear shared signal resources before `Runtime::DestroyAgents()` destroys the GpuAgent, causing `AqlQueue::~AqlQueue()` to store to a freed queue_inactive_signal.

## Motivation (Where this was found)
This was found while enabling a JAX/XLA function with cooperative kernels in ROCm. The computation completes successfully, but the process segfaults later during Python shutdown.

A simple standalone HIP program that launches a cooperative kernel did not reproduce the issue. That appears to be because the simple HIP application follows a different teardown path and does not leave the persistent cooperative/GWS queue alive long enough until ROCR shutdown.

The JAX/XLA plugin path is different: XLA owns ROCm executors, streams, and plugin-global state, and those objects are torn down during Python/module shutdown. That lifetime shape can leave the runtime-owned cooperative/GWS queue alive until hsa_shut_down(), exposing a bad ordering inside ROCR shutdown.

The invalid access itself occurs entirely inside ROCR/HIP shutdown, not inside JAX/XLA.

Backtrace:
```
Thread 1 "python3" received signal SIGSEGV, Segmentation fault.
#0  rocr::HSA::hsa_signal_store_screlease(hsa_signal_s, long)
#1  rocr::AMD::AqlQueue::~AqlQueue()
#2  rocr::AMD::AqlQueue::~AqlQueue()
#3  rocr::AMD::GpuAgent::~GpuAgent()
#4  rocr::AMD::GpuAgent::~GpuAgent()
#5  rocr::core::Runtime::DestroyAgents()
#6  rocr::core::Runtime::Unload()
#7  rocr::core::Runtime::Release()
#8  rocr::HSA::hsa_shut_down()
#9  libamdhip64.so.7
#10 libamdhip64.so.7
#11 __run_exit_handlers
```

## Details
A cooperative/GWS queue is owned by `GpuAgent::gws_queue_` and can remain alive until runtime shutdown. Currently, if it is still present when agents are destroyed, its AqlQueue destructor runs from the GpuAgent destructor path.

```
Runtime::Unload()
  ...
  SharedSignalPool.clear()
  EventPool.clear()
  system_regions_fine_.clear()
  system_regions_coarse_.clear()
  DestroyAgents()
    GpuAgent::~GpuAgent()
      AqlQueue::~AqlQueue()
```
`AqlQueue::~AqlQueue()` stores to `amd_queue_.queue_inactive_signal`. If the shared signal backing has already been cleared, this can segfault during process exit.

This change resets `gws_queue_.queue_` from `GpuAgent::ReleaseResources()`, before `Runtime::Unload()` clears shared signal resources. The queue destructor therefore runs while the resources it depends on are still valid.

The GWS queue lock is held while resetting queue_ and ref_ct_, matching the existing synchronization used by the cooperative queue acquire/release path.

## Testing
A stand alone unit test to address this is very difficult as the failure depends on full process/runtime shutdown ordering: HIP/ROCR atexit handling, cooperative/GWS queue lifetime, async signal handler teardown, shared signal pool teardown, and agent destruction during `hsa_shut_down()`. The issue is not reproducible as a small in-process unit test. Instead, it needs to be invoked through a cooperative kernel launched through a larger framework like JAX/XLA (that is also how I verified that the patch in this PR works).

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
@@ -1038,6 +1038,15 @@ void GpuAgent::ReleaseResources() {
   if (this->Enabled()) {
     this->Disable();
 
+    // Destroy the persistent GWS/cooperative queue before tearing down
+    // resources. Otherwise ~AqlQueue may run later and store to a freed
+    // queue_inactive_signal.
+    {
+      std::lock_guard<std::mutex> lock(gws_queue_.lock_);
+      gws_queue_.queue_.reset();
+      gws_queue_.ref_ct_ = 0;
+    }
+
     // Remove all shared hardware queues from pool
     queue_pool_.Cleanup();
 


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -1038,6 +1038,15 @@ void GpuAgent::ReleaseResources() {
   if (this->Enabled()) {
     this->Disable();
 
+    // Destroy the persistent GWS/cooperative queue before tearing down
+    // resources. Otherwise ~AqlQueue may run later and store to a freed
+    // queue_inactive_signal.
+    {
+      std::lock_guard<std::mutex> lock(gws_queue_.lock_);
+      gws_queue_.queue_.reset();
```
cfreeamd, line 1046: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** `~AqlQueue` destructor runs while `gws_queue_.lock_` is held.

`queue_.reset()` destroys the `AqlQueue` object synchronously while `gws_queue_.lock_` is held. `~AqlQueue` is not short — it signals the error-handler loop and polls for `ERROR_HANDLER_DONE`. If any path reached from the error-handler thread (or from `~AqlQueue` itself) tries to acquire `gws_queue_.lock_`, the result is deadlock. Holding a mutex across a blocking async wait is fragile for future maintenance.

**Suggestion:** Steal the owned pointer under the lock, then destroy it after releasing:
```cpp
std::unique_ptr<core::Queue> to_destroy;
{
  std::lock_guard<std::mutex> lock(gws_queue_.lock_);
  to_destroy = std::move(gws_queue_.queue_);
  gws_queue_.ref_ct_ = 0;
}
// ~AqlQueue runs here, after the lock is released
```
The existing code is not wrong for the current shutdown scenario, but this pattern is safer for long-term maintenance.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -1038,6 +1038,15 @@ void GpuAgent::ReleaseResources() {
...
+      gws_queue_.ref_ct_ = 0;
```
cfreeamd, line 1047: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** `ref_ct_ = 0` ignores outstanding references — needs a clarifying comment.

If `ref_ct_ > 0` at the moment `ReleaseResources()` runs (a caller still holds a cooperative queue handle), the code forcibly deletes the queue and zeros `ref_ct_`. A subsequent `hsa_queue_destroy()` by that caller would invoke `Destroy()` on a deleted object — another UAF. This is safe in practice because `IS_OPEN()` rejects `hsa_queue_destroy()` after `hsa_shut_down()` returns. A brief comment confirming this invariant would make the reasoning explicit:
```cpp
// IS_OPEN() prevents hsa_queue_destroy() calls after hsa_shut_down(), so
// leaked cooperative queue refs are safe to discard here.
```

cfreeamd (commented):

`amd_gpu_agent.cpp` — fix structure and placement in `ReleaseResources()` look correct. See inline comments for two implementation concerns to address before merge. Also: this PR appears to fix the same shutdown UAF as the still-open PR #6942 — coordination or closure of one would help.

tsrw2048:
Closing this PR as https://github.com/ROCm/rocm-systems/pull/6942 does the same thing and has been merged.

PR #8589, opened 2026-07-14
ci(rocjitsu): normalize sanitizer linking behaviour

amd-eochoalo:
## Motivation

GCC and Clang differ when using sanitizers. Clang defaults to statically linking sanitizers, while GCC defaults to dynamically linking them. If the ASan library is dynamically linked but was not the first library mapped by the dynamic linker, then it will throw the error "ASan runtime does not come first in initial library list". 

## Technical Details

When using the interposer, the interposer must come second after libasan. Furthermore, the rocjitsu binary can only use sanitizers when dynamically linked against sanitizer libraries. CTests can be linked statically with sanitizers as they include the startup code needed by the sanitizers.

This PR also:
- [X] includes a small fix related to a use after free revealed by GCC's ASan configuration on CI.
- [X] Modifies the execve script to prevent modifying environment variables.
- [x] Changes CI and CMake scripts to default to dynamically linking sanitizers
- [x] Enables TSan on CI

## JIRA ID

https://github.com/ROCm/rocm-systems/issues/7891

## Test Plan

Run on CI.

## Test Result

Test on CI passes.

## Submission Checklist


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/execution_plugin_group.h
```diff
@@ -143,8 +143,9 @@ class ExecutionPluginGroup {
   }
 
   static std::shared_ptr<ExecutionPluginGroup> empty_group() {
-    static auto instance = std::make_shared<ExecutionPluginGroup>();
-    return instance;
+    static const auto *instance =
+        new std::shared_ptr<ExecutionPluginGroup>(std::make_shared<ExecutionPluginGroup>());
+    return *instance;
```
amd-eochoalo: This is the change that I'm not sure of and that was discovered by GCC with Asan. I believe it is a difference between Clang and GCC's destructor ordering but I'm not yet too familiar with this code to say. I reverted the other vibe coded changes I had.

emulation/rocjitsu/CMakeLists.txt
```diff
@@ -93,6 +94,204 @@ option(
...
+#       SHARED <ON|OFF>
+#       [SHARED_LIBRARIES <path>...]
+#       SANITIZERS <address|undefined|thread|memory>...
+#   )
+#
+# The caller is responsible for adding the common `-fsanitize=<kinds>` option.
+# This helper only selects whether sanitizer runtime libraries are linked
+# dynamically or statically. When shared runtime paths are available, it also
+# adds rpaths so build-tree tools can run without setting LD_LIBRARY_PATH.
+function(rj_sanitizer_runtime_link_options out_var)
```
kuhar: I think it would help to move the two sanitizer helpers into a focused cmake/rj_sanitizers.cmake module and include it here. They add about 180 lines of reusable, compiler-specific logic, while the existing rj_* helpers live under cmake; keeping the same boundary will make this logic easier to test and maintain.

emulation/rocjitsu/tools/rocjitsu/main.cpp
```diff
@@ -358,6 +368,27 @@ void prepend_env_path(const char *name, const std::string &value) {
...
+  const std::filesystem::path runtime_path(info.dli_fname);
+  if (runtime_path.filename().string().find("asan") == std::string::npos)
+    return {};
+  return canonical_existing_path(runtime_path);
+#else
+  return {};
+#endif
+}
+
+void prepend_launch_preloads(const std::string &interposer_path) {
```
kuhar: It would be useful to add a focused unit test for the LD_PRELOAD composition here. The GCC shared-ASan integration tests pass, but they only prove that the child starts; they do not directly pin the required asan:interposer:existing order or the no-ASan branch. Extracting the pure composition step and testing those cases would cover the new correctness invariant and satisfy the current no-unit-test policy failure.

kuhar (commented):
A few comments

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/execution_plugin_group.h
kuhar: I think we should add a short comment explaining that this process-lifetime allocation is intentional. It prevents a late empty_group() call during static teardown from using a destroyed function-local shared_ptr; the GCC ASan interposer tests exposed that ordering. Without the rationale, a future cleanup could restore the original lifetime bug.

emulation/rocjitsu/tools/rocjitsu/launch_preload.cpp
```diff
@@ -0,0 +1,68 @@
...
+} // namespace
+
+std::string find_loaded_asan_runtime() {
+#if defined(RJ_BUILT_WITH_ASAN)
+  Dl_info info{};
+  if (!dladdr(reinterpret_cast<void *>(&__asan_init), &info) || !info.dli_fname)
+    return {};
+
+  const std::filesystem::path runtime_path(info.dli_fname);
+  if (runtime_path.filename().string().find("asan") == std::string::npos)
```
kuhar: We should reject the main executable here rather than accepting any loaded object whose basename contains `asan`. In a static-ASan build, `dladdr(&__asan_init)` resolves to the executable; when it is invoked through an `asan`-named symlink, this check treats the executable as the runtime and adds it to `LD_PRELOAD`. I reproduced this with `/tmp/launch-asan-preload-test`: `LaunchPreloadTest.NoAsanPrependsInterposerBeforeExistingPreload` fails because the helper returns the test executable. Please identify an actual ASan shared object (or explicitly reject the main executable) and add this alias case to the test.

kuhar (commented):
Just one comment

emulation/rocjitsu/tools/rocjitsu/launch_preload.cpp
kuhar: Also, do we need it for tsan?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/execution_plugin_group.h
amd-eochoalo: https://github.com/ROCm/rocm-systems/pull/8589/commits/828813c7a9cd462b094d88ef2e14d0ec9e955f80

emulation/rocjitsu/tools/rocjitsu/launch_preload.cpp
amd-eochoalo: https://github.com/ROCm/rocm-systems/pull/8589/commits/8ff99facc46b26e1abe83418c0fd97964e432095 and yes. It is needed for TSan.

emulation/rocjitsu/tools/rocjitsu/launch_preload.cpp
```diff
@@ -0,0 +1,78 @@
...
+  std::string runtime = canonical_existing_path(runtime_path);
+  if (runtime.empty() || runtime == current_executable_path())
+    return {};
+  return runtime;
+#else
+  return {};
+#endif
+}
+
+void prepend_launch_preloads(const std::string &interposer_path) {
```
amd-arosa: please document this.

also now that you are touching this im not a fan of editing env variables while running. esp not LD_PRELOAD as it is confusing.

if you are extracting a helper can you do a helper to make launches?

amd-arosa (commented):
looks pretty good.

emulation/rocjitsu/tools/rocjitsu/launch_preload.cpp
amd-eochoalo: https://github.com/ROCm/rocm-systems/pull/8589/commits/56a29d3bb63e444de7f03d65daa08ea5c5762460

emulation/rocjitsu/tools/rocjitsu/launch_preload.cpp
amd-eochoalo: https://github.com/ROCm/rocm-systems/pull/8589/commits/52f1fbf3b85b914402f3f984b407f90201e5faa7

kuhar (commented):
We should finish the TSan launcher path with both a single-runtime test value and CI coverage. At emulation/rocjitsu/tests/CMakeLists.txt:467, RJ_EXPECTED_SHARED_TSAN_RUNTIME receives the full RJ_SANITIZER_SHARED_LIBRARIES list; a shared TSan+UBSan build therefore passes a semicolon-joined value to canonical_existing_path(), which cannot resolve it. Please expose a dedicated RJ_TSAN_SHARED_LIBRARY like the ASan value and use it here. Also add a .github/workflows/rocjitsu-corpus-tests.yml matrix entry with enable_tsan: 1, ideally with sanitize_shared: 1: every current entry has TSan disabled, so SharedTsanPrependsTsanBeforeInterposerAndExistingPreload is always skipped and this wiring is not compiled in CI.

Can we use execvpe(file, argv, environment.envp()) at emulation/rocjitsu/tools/rocjitsu/launch_preload.cpp:188 instead of maintaining a separate execvp implementation? The helper reimplements PATH walking, EACCES precedence, and the ENOEXEC to /bin/sh fallback, but launch_preload_test.cpp does not exercise any of those branches. If portability requires keeping the custom implementation, please add focused fork/wait tests for a bare name on a custom PATH, EACCES precedence, the ENOEXEC fallback, and an empty file argument.

This would be safer if LaunchEnvironment were non-copyable, or if copying rebuilt its pointer cache. At emulation/rocjitsu/tools/rocjitsu/launch_preload.h:21, the implicit copy operations copy envp_ raw pointers after envp() materializes them, so they still point into the source entries_ while envp_dirty_ is copied as false. Modifying or destroying the source then leaves copy.envp() with dangling pointers. The launcher only needs return-by-value and move semantics, so deleting copy construction and copy assignment while retaining move support is the smaller contract.

amd-eochoalo:
> Can we use execvpe(file, argv, environment.envp()) at emulation/rocjitsu/tools/rocjitsu/launch_preload.cpp:188 instead of maintaining a separate execvp implementation? 

execvpe is a _GNU_SOURCE extension. We do have another target that compiles with _GNU_SOURCE extensions

```
(env) eochoalo@shark-mkm-2:~/code/rocm-systems/emulation/rocjitsu$ git grep GNU_SOUR
CMakeLists.txt:    target_compile_definitions(rocjitsu_shared PRIVATE _GNU_SOURCE)
```

Given that we are depending already on _GNU_SOURCE for rocjitsu_shared (uses include: RTLD_NEXT, memfd_create, dup3). I think we can use execvpe here and we can look into making it more portable later on / as a separate ticket.

amd-eochoalo:
Please note that the error should be fixed by https://github.com/ROCm/rocm-systems/pull/8775

lialan:
Made a little bit of update to this PR:
* change `RJ_SANITIZE_SHARED` to `RJ_SANITIZER_RUNTIME=SHARED` (which is new), because shared does not apply to MSAN.

.github/workflows/rocjitsu-corpus-tests.yml
```diff
@@ -94,9 +97,26 @@ jobs:
...
             enable_expensive_checks: 1
-            sanitize_shared: 1
+            sanitizer_runtime: SHARED
+            run_tests: 1
             c_compiler: gcc
             cxx_compiler: g++
+          - name: msan-auto
+            allow_failure: false
+            # MSan requires instrumented runtime dependencies. Keep this gate
+            # to configuration and final linkage without running the binary.
```
kuhar: let's not add msan in this PR

.github/workflows/rocjitsu-corpus-tests.yml
```diff
@@ -166,10 +180,17 @@ jobs:
...
             "RaceTest\..*"
           )
+          TSAN_EXCLUDED_TESTS=(
+            "${ASAN_UBSAN_EXCLUDED_TESTS[@]}"
+            "MatmulStressTest\.TiledAllCU"
+            "MatmulStressTest\.MfmaAllCUs"
+            "VectorAddStressTest\.AllCUsGoldenReference"
+            "Gfx1250SimulationTest\.*"
+            "Gfx1250ExecutionTest\.*"
+            "InterposerDupTest.SerializedReopenUnderContentionStaysRoutable"
```
kuhar, line 190: Can we have a tracking issue that explains why these are disabled and an inline comment pointing to it?

kuhar (commented):
It would be useful to add a focused test for the custom LaunchEnvironment envp cache. At emulation/rocjitsu/tools/rocjitsu/launch_preload.cpp:124-133, materialize envp(), verify its entries and null terminator, mutate the environment with set(), and call envp() again to cover the dirty-cache rebuild. This is the buffer passed to execvpe, while the current launch_preload tests only inspect values through get().

We should keep libclang-rt-dev in the dependency list at .github/workflows/rocjitsu-corpus-tests.yml:124, or stop selecting RJ_SANITIZER_RUNTIME=SHARED for the Clang jobs. This workflow installs with --no-install-recommends; on Ubuntu 24.04, clang-18 depends on libclang-common-18-dev, but that package only recommends libclang-rt-18-dev. After commit 60e58f52c9 removed the explicit package, fresh asan-ubsan and tsan jobs cannot find the Clang shared sanitizer runtimes and CMake fails before testing. Please restore the dependency.

.github/workflows/rocjitsu-corpus-tests.yml
amd-eochoalo: I created this issue here: https://github.com/ROCm/rocm-systems/issues/8922

.github/workflows/rocjitsu-corpus-tests.yml
```diff
@@ -166,10 +175,17 @@ jobs:
             "LoggingTest\.dispatch_logged"
             "ProfiledPluginTest\.hook_profile_output"
             "CvtNarrow\.AllTests"
-            "Interposer.*Test\..*"
-            "GuestKfd.*\..*"
             "RaceTest\..*"
           )
+          TSAN_EXCLUDED_TESTS=(
```
amd-eochoalo, line 183:
```suggestion
          # https://github.com/ROCm/rocm-systems/issues/8922
          TSAN_EXCLUDED_TESTS=(
```

amd-eochoalo: @lialan

kuhar (commented):
It would be useful to add one focused case at emulation/rocjitsu/tests/launch_preload_test.cpp:79 that runs prepend_launch_preloads with LD_PRELOAD unset, and ideally with an empty value, then checks that the result is exactly the interposer path. Every current ordering test seeds an existing preload first, so the common no-existing-value branch is not guarded against an accidental leading or trailing colon.

I think this ASan alias regression should be exercised by at least one static-ASan CI configuration. The shared sanitizer jobs skip SanitizerNamedExecutableAliasDoesNotPreloadExecutable, while unsanitized builds compile out the runtime lookup; the current static-TSan job therefore covers only the TSan half of the self-exclusion guard. Since the ASan-named executable was the original failure case, please either add static-ASan coverage or refactor the test so that branch is exercised without a separate matrix entry.

lialan:
Added exact unset/empty LD_PRELOAD coverage, a blocking static-ASan LaunchPreloadTest job, and the #8922 annotation in ee6687b8ea.

.github/workflows/rocjitsu-corpus-tests.yml
```diff
@@ -73,6 +73,16 @@ jobs:
...
+          - name: asan-static-launch-preload
+            allow_failure: false
+            build_target: launch_preload_test
+            test_regex: '^LaunchPreloadTest\.'
+            build_type: RelWithDebInfo
+            enable_asan: 1
+            enable_ubsan: 0
+            enable_tsan: 0
+            enable_expensive_checks: 0
+            sanitizer_runtime: STATIC
```
kuhar: I don't want to further expand the configuration matrix if we can avoid it

lialan: folded the static-ASan launcher test into the existing blocking sanitizer job

kuhar (commented):
We should prepend the selected shared sanitizer runtime to the direct preload environment at emulation/rocjitsu/tests/CMakeLists.txt:649. With RJ_SANITIZER_RUNTIME=AUTO, LD_PRELOAD currently starts with librocjitsu.so; ctest --test-dir build-clang-23-asan-ubsan -R ^RocminfoTest.ExitsSuccessfully$ --output-on-failure aborts before the test starts because the ASan runtime is not first. The launcher tests do not cover this path, and CI excludes RocminfoTest. It would also help to reuse one sanitizer-first prefix for this environment and the early-mmap and daemon direct-preload paths.

lialan:
> We should prepend the selected shared sanitizer runtime to the direct preload environment at emulation/rocjitsu/tests/CMakeLists.txt:649. With RJ_SANITIZER_RUNTIME=AUTO, LD_PRELOAD currently starts with librocjitsu.so; ctest --test-dir build-clang-23-asan-ubsan -R ^RocminfoTest.ExitsSuccessfully$ --output-on-failure aborts before the test starts because the ASan runtime is not first. The launcher tests do not cover this path, and CI excludes RocminfoTest. It would also help to reuse one sanitizer-first prefix for this environment and the early-mmap and daemon direct-preload paths.

Latest changes reuse a sanitizer-first preload prefix for Rocminfo, early-mmap, and daemon tests.

emulation/rocjitsu/cmake/rj_sanitizers.cmake
```diff
@@ -0,0 +1,209 @@
...
+        if(NOT _runtime_library)
+            message(
+                FATAL_ERROR
+                "RJ_SANITIZER_RUNTIME=SHARED requires the ${_sanitizer} "
+                "sanitizer shared runtime, but CMake could not locate "
+                "${_runtime_name} with ${RJ_SANI_LIB_COMPILER}."
+            )
+        endif()
+
+        list(APPEND _sanitizer_shared_libraries "${_runtime_library}")
```
kuhar, line 198: We should build the direct-preload prefix from only RJ_ASAN_SHARED_LIBRARY or RJ_TSAN_SHARED_LIBRARY, not from the aggregate RJ_SANITIZER_SHARED_LIBRARIES inventory. At emulation/rocjitsu/tests/CMakeLists.txt:711, Clang ASan+UBSan currently becomes asan:ubsan; timeout 5 with that LD_PRELOAD on /bin/true returns 124, while ASan alone returns 0. The same combined prefix will hang the non-excluded EarlyMmap test before main in the Clang ASan+UBSan job.

kuhar (commented):
A couple of comments, the other one is a reply here: https://github.com/ROCm/rocm-systems/pull/8589#discussion_r3631527666

.github/workflows/rocjitsu-corpus-tests.yml
kuhar: I think we should make the fold explicit by gating this step on a matrix field such as run_static_asan_launch instead of matrix.name == tsan-sparse-memory. As written, renaming or removing that unrelated TSan entry silently drops the static-ASan launcher coverage. A boolean keeps the job count unchanged while making the dependency survive entry renames.

lialan: Made the fold explicit with `run_static_asan_launch`, so the static-ASan step no longer depends on the host matrix row name, in 8fd034aa3a.

emulation/rocjitsu/cmake/rj_sanitizers.cmake
lialan: Restricted the direct-preload prefix to the selected ASan or TSan runtime while leaving the aggregate sanitizer inventory unchanged,.

kuhar:
@lialan we have test failures

kuhar (commented):
We should restore libclang-rt-dev to the dependency list at .github/workflows/rocjitsu-corpus-tests.yml:120, or stop selecting RJ_SANITIZER_RUNTIME=SHARED for the Clang ASan and TSan rows. This install uses --no-install-recommends, and on Ubuntu 24.04 clang-18 reaches libclang-rt-18-dev only through a recommendation from libclang-common-18-dev. Without the explicit package, fresh jobs lack the shared ASan and TSan runtimes and fail in rj_find_sanitizer_shared_libraries before running tests.

.github/workflows/rocjitsu-corpus-tests.yml
```diff
@@ -219,6 +241,32 @@ jobs:
...
+            -DRJ_ENABLE_ASAN=1 \
+            -DRJ_ENABLE_UBSAN=0 \
+            -DRJ_ENABLE_TSAN=0 \
+            -DRJ_ENABLE_EXPENSIVE_CHECKS=0 \
+            -DRJ_SANITIZER_RUNTIME=STATIC
+
+          WORKERS="$(( ($(nproc) + 1) / 2 ))"
+          cmake --build "${STATIC_ASAN_BUILD_DIR}" \
+            --target launch_preload_test \
+            -j "${WORKERS}"
```
kuhar: There's no need to set explicit parallelism level with ninja, we should drop this variable and the `-j N` flag.

kuhar (approved):
LGTM % one comment

PR #8591, opened 2026-07-14
ROCM-1028 - Toggle L2 Cache of pinned host buffer on device.

TomSang:
In PAL,
  Add gl2Uncached in PinnedGpuMemoryCreateInfo.
  When hipExtHostRegisterCoarseGrained is set in hipHostRegister(),
    gl2Uncached = 0,
    m_mtype = MType::Default(0) in PAL,
    KMD will map MType to MEMORY_CACHE_POLICY__NC, which enables L2
    cache on device for pinned host (i.e., coarse grain).
  Otherwise,
    gl2Uncached = 1,
    m_mtype = MType::Uncached in PAL,
    KMD will map MType to MEMORY_CACHE_POLICY__UC, which disables L2
    cache on device for pinned host (i.e., fine grain).

In Rocr,
  Replace HsaMemMapFlags with HsaMemFlags in some functions where
  HsaMemMapFlags is not refereneced, so that we can transfer CoarseGrain bit.
  Remove HsaMemMapFlags as it isn't referenced anymore.

Without this change, PAL/Rocr in Windows will always enable L2 cache with or without hipExtHostRegisterCoarseGrained in hipHostRegister().  This isn't expected.

## Motivation

Fix an unexpected behavior of hipHostRegister().

## Technical Details

Add gl2Uncached in PinnedGpuMemoryCreateInfo to transfer L2 cache info from clr to pal.
hipHostRegister(hipExtHostRegisterCoarseGrained) should follow coarse grain behavior.
hipHostRegister(Non hipExtHostRegisterCoarseGrained) should follow fine grain behavior.

## JIRA ID
ROCM-1028

## Test Plan
hipHostRegister test should pass.

## Test Result
all pass

## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/libhsakmt/include/hsakmt/hsakmt.h
+++ b/projects/rocr-runtime/libhsakmt/include/hsakmt/hsakmt.h
@@ -784,7 +784,7 @@ hsaKmtMapMemoryToGPUNodes(
     void*           MemoryAddress,         //IN (page-aligned)
     HSAuint64       MemorySizeInBytes,     //IN (page-aligned)
     HSAuint64*      AlternateVAGPU,        //OUT (page-aligned)
-    HsaMemMapFlags  MemMapFlags,           //IN
+    HsaMemFlags     MemFlags,              //IN
     HSAuint64       NumberOfNodes,         //IN
     HSAuint32*      NodeArray              //IN
     );
--- a/projects/rocr-runtime/libhsakmt/src/dxg/memory.cpp
+++ b/projects/rocr-runtime/libhsakmt/src/dxg/memory.cpp
@@ -809,15 +809,17 @@ HSAKMT_STATUS HSAKMTAPI hsaKmtMapMemoryToGPU(void *MemoryAddress,
 
   HSAuint64 NumberOfNodes = 1;
   HSAuint32 NodeArray[] = {dxg_runtime->default_node};
-  HsaMemMapFlags MemMapFlags;
-  MemMapFlags.Value = 0;
+  HsaMemFlags MemFlags;
+  MemFlags.Value = 0;
+  MemFlags.ui32.CoarseGrain = 1;
 
   return hsaKmtMapMemoryToGPUNodes(MemoryAddress, MemorySizeInBytes, AlternateVAGPU,
-    MemMapFlags, NumberOfNodes, NodeArray);
+    MemFlags, NumberOfNodes, NodeArray);
 }
+
 HSAKMT_STATUS HSAKMTAPI hsaKmtMapMemoryToGPUNodes(
     void *MemoryAddress, HSAuint64 MemorySizeInBytes, HSAuint64 *AlternateVAGPU,
-    HsaMemMapFlags MemMapFlags, HSAuint64 NumberOfNodes, HSAuint32 *NodeArray) {
+    HsaMemFlags MemFlags, HSAuint64 NumberOfNodes, HSAuint32 *NodeArray) {
   CHECK_DXG_OPEN();
 
   if (!MemoryAddress || !AlternateVAGPU) {
@@ -900,7 +902,7 @@ HSAKMT_STATUS HSAKMTAPI hsaKmtMapMemoryToGPUNodes(
   create_info.size = aligned_size;
   create_info.user_ptr = aligned_ptr;
   // create_info.mem_flags = 0 means coarse grain by default
-  if (MemMapFlags.ui32.CachePolicy == HSA_CACHING_NONCACHED) {
+  if (!MemFlags.ui32.CoarseGrain) {
     create_info.mem_flags = Wkmi::kFineGrain;
   }
   auto code = dev->CreateGpuMemory(create_info, &gpu_mem);
--- a/projects/rocr-runtime/libhsakmt/src/dxg/memory.cpp
+++ b/projects/rocr-runtime/libhsakmt/src/dxg/memory.cpp
@@ -899,7 +899,10 @@ HSAKMT_STATUS HSAKMTAPI hsaKmtMapMemoryToGPUNodes(
   create_info.domain = Wkmi::kUserMemory;
   create_info.size = aligned_size;
   create_info.user_ptr = aligned_ptr;
-
+  // create_info.mem_flags = 0 means coarse grain by default
+  if (MemMapFlags.ui32.CachePolicy == HSA_CACHING_NONCACHED) {
+    create_info.mem_flags = Wkmi::kFineGrain;
+  }
   auto code = dev->CreateGpuMemory(create_info, &gpu_mem);
   if (code == ErrorCode::Success) {
     addr = gpu_mem->GpuAddress();
--- a/projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDEvictTest.cpp
+++ b/projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDEvictTest.cpp
@@ -63,10 +63,8 @@ void KFDEvictTest::AllocBuffers(bool m_IsParent, HSAuint32 defaultGPUNode, HSAui
               << totalMB << ")MB VRAM in KFD" << std::endl;
     }
 
-    HsaMemMapFlags mapFlags = {0};
-    HSAKMT_STATUS ret;
+    HSAKMT_STATUS ret = 0;
     HSAuint32 retry = 0;
-
     m_Flags.Value = 0;
     m_Flags.ui32.PageSize = HSA_PAGE_SIZE_4KB;
     m_Flags.ui32.HostAccess = 0;
@@ -77,7 +75,7 @@ void KFDEvictTest::AllocBuffers(bool m_IsParent, HSAuint32 defaultGPUNode, HSAui
         if (ret == HSAKMT_STATUS_SUCCESS) {
             if (hsakmt_is_dgpu()) {
                 if (HSAKMT_CALL(hsaKmtMapMemoryToGPUNodes, m_hsakmt_current_ctx, m_pBuf, vramBufSize, NULL,
-                       mapFlags, 1, reinterpret_cast<HSAuint32 *>(&defaultGPUNode)) == HSAKMT_STATUS_ERROR) {
+                       m_Flags, 1, reinterpret_cast<HSAuint32 *>(&defaultGPUNode)) == HSAKMT_STATUS_ERROR) {
                     EXPECT_SUCCESS(HSAKMT_CALL(hsaKmtFreeMemory, m_hsakmt_current_ctx, m_pBuf, vramBufSize));
                     LOG() << "Map failed for " << i << "/" << count << " buffer. Retrying allocation" << std::endl;
                     goto retry;
--- a/projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDIPCTest.cpp
+++ b/projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDIPCTest.cpp
@@ -78,7 +78,6 @@ void KFDIPCTest::BasicTestChildProcess(int defaultGPUNode, int *pipefd, HsaMemFl
     HSAuint64 size = PAGE_SIZE, sharedSize;
     HsaMemoryBuffer tempSysBuffer(size, defaultGPUNode, false);
     HSAuint32 *sharedLocalBuffer = NULL;
-    HsaMemMapFlags mapFlags = {0};
 
     /* Read from Pipe the shared Handle. Import shared Local Memory */
     ASSERT_GE(read(pipefd[0], reinterpret_cast<void*>(&sharedHandleLM), sizeof(sharedHandleLM)), 0);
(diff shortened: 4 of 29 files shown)


projects/rocr-runtime/libhsakmt/src/dxg/memory.cpp
```diff
@@ -894,6 +894,8 @@ HSAKMT_STATUS HSAKMTAPI hsaKmtMapMemoryToGPUNodes(
   create_info.domain = Wkmi::kUserMemory;
   create_info.size = aligned_size;
   create_info.user_ptr = aligned_ptr;
+  if (MemMapFlags.ui32.CachePolicy == HSA_CACHING_NONCACHED)
+    create_info.mem_flags = Wkmi::kFineGrain;
```
cfreeamd, line 906: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** This assignment uses `=` not `|=`:

```cpp
create_info.mem_flags = Wkmi::kFineGrain;
```

If `Wkmi::GpuMemoryCreateInfo::mem_flags` is zero-initialized before this point and `kFineGrain` is the only relevant flag here, `=` is correct. But if `mem_flags` is a bitmask and could have other flags set before line 897 (or in future code), this silently clears them — `|=` would be safer. Is `mem_flags` a mutually-exclusive enum or a bitmask?

Also: missing braces around the single-statement `if` body (nit — prefer braces).

cfreeamd (commented):

ROCr-only review of `libhsakmt/src/dxg/memory.cpp` and `amd_memory_region.cpp`. The `amd_memory_region.cpp` change looks clean. One question on the `memory.cpp` assignment — see inline comment.

projects/rocr-runtime/libhsakmt/src/dxg/memory.cpp
TomSang: mem_flags isn't a bitmask. Before line 897, it is 0 which means coarse grain.   There is no explicity definition of Wkmi::kCoarseGrain.

projects/rocr-runtime/libhsakmt/src/dxg/memory.cpp
```diff
@@ -899,7 +899,10 @@ HSAKMT_STATUS HSAKMTAPI hsaKmtMapMemoryToGPUNodes(
   create_info.domain = Wkmi::kUserMemory;
   create_info.size = aligned_size;
   create_info.user_ptr = aligned_ptr;
-
+  // create_info.mem_flags = 0 means coarse grain by default
+  if (MemMapFlags.ui32.CachePolicy == HSA_CACHING_NONCACHED) {
```
dayatsin-amd: Can you use:
if (!MemMapFlags.ui32.CoarseGrain) {
   create_info.mem_flags = Wkmi::kFineGrain;
}

fcui-amd (approved):

cfreeamd (approved):

dayatsin-amd (changes requested):

projects/rocr-runtime/libhsakmt/src/dxg/memory.cpp
TomSang: Yes, I want to use !MemMapFlags.ui32.CoarseGrain, but  CoarseGrain is not available  in MemMapFlags.    So I have to use MemMapFlags.ui32.CachePolicy .
Note that the related code is different in Linux and Windows paths.

The PR won't affect Rocr in Linux . I traced and tested in Linux, everything is good

TomSang: using MemMapFlags.ui32.CoarseGrain instead of MemMapFlags.ui32.CachePolicy == HSA_CACHING_NONCACHED leading to a big refactoring change in https://github.com/ROCm/rocm-systems/pull/8591/changes/0b2e283c8a9e2ae1d305cefdc79c462eb95a5f07

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDIPCTest.cpp
```diff
@@ -127,12 +126,11 @@ void KFDIPCTest::BasicTestParentProcess(int defaultGPUNode, pid_t cpid, int *pip
     HsaMemoryBuffer tempSysBuffer(PAGE_SIZE, defaultGPUNode, false);
     SDMAQueue sdmaQueue;
     HsaSharedMemoryHandle sharedHandleLM;
-    HsaMemMapFlags mapFlags = {0};
 
     ASSERT_SUCCESS(HSAKMT_CALL(hsaKmtAllocMemory, m_hsakmt_current_ctx, defaultGPUNode, size, mflags, &toShareLocalBuffer));
     /* Fill a Local Buffer with a pattern */
     ASSERT_SUCCESS(HSAKMT_CALL(hsaKmtMapMemoryToGPUNodes, m_hsakmt_current_ctx, toShareLocalBuffer, size, &AlternateVAGPU,
-                       mapFlags, 1, reinterpret_cast<HSAuint32 *>(&defaultGPUNode)));
+                       mFlags, 1, reinterpret_cast<HSAuint32 *>(&defaultGPUNode)));
```
dayatsin-amd: `mFlags` is undeclared here (the parameter is `mflags`), so this does not compile.
```suggestion
                       mflags, 1, reinterpret_cast<HSAuint32 *>(&defaultGPUNode)));
```

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDMemoryTest.cpp
```diff
@@ -3123,9 +3119,10 @@ void KFDMemoryTest::ExportDMABufTest(int gpuNode) {
     buf = reinterpret_cast<HSAuint32 *>(info.MemoryAddress);
     ASSERT_EQ_GPU(info.SizeInBytes, PAGE_SIZE, gpuNode);
 
-    HsaMemMapFlags mapFlags = {0};
+    HsaMemFlags memFlags = {0};
```
dayatsin-amd: `memFlags` is already declared earlier in this function (`HsaMemFlags memFlags = GetHsaMemFlags();`), so this redeclaration fails to compile. Reset the existing variable instead.
```suggestion
    memFlags.Value = 0;
```

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDMemoryTest.cpp
```diff
@@ -3174,7 +3171,8 @@ void KFDMemoryTest::VA_VRAM_Only_AllocTest(int gpuNode) {
     memFlags.ui32.NonPaged = 1;
     memFlags.ui32.HostAccess = 0;
 
-    HsaMemMapFlags mapFlags = {0};
+    HsaMemFlags memFlags = {0};
```
dayatsin-amd: Same issue: `memFlags` is already declared above in this function, so this redeclaration does not compile. Reset the existing variable instead.
```suggestion
    memFlags.Value = 0;
```

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDEvictTest.cpp
```diff
@@ -63,10 +63,6 @@ void KFDEvictTest::AllocBuffers(bool m_IsParent, HSAuint32 defaultGPUNode, HSAui
               << totalMB << ")MB VRAM in KFD" << std::endl;
     }
 
-    HsaMemMapFlags mapFlags = {0};
-    HSAKMT_STATUS ret;
-    HSAuint32 retry = 0;
-
     m_Flags.Value = 0;
```
dayatsin-amd, line 68: `ret` and `retry` were removed but are still used below (`ret = ...`, `retry++`, `retry = 0`), so `AllocBuffers` no longer compiles. Re-add their declarations.
```suggestion
    HSAKMT_STATUS ret;
    HSAuint32 retry = 0;

    m_Flags.Value = 0;
```

dayatsin-amd (commented):
Reviewed the branch and found that kfdtest currently fails to compile due to a few issues introduced in the `HsaMemMapFlags` -> `HsaMemFlags` migration. The inline suggestions below fix them.

### Verified
Each of the six touched kfdtest sources passes `g++ -std=c++17 -fsyntax-only` after these fixes.

### Follow-up (optional, not inline-suggestible)
With this migration, `HsaMemMapFlags` is no longer referenced by any consumer. It can be removed entirely, but the remaining references live in files **not** part of this PR's diff, so they can't be posted as click-apply suggestions:
- `libhsakmt/include/hsakmt/hsakmttypes.h` - delete the now-unused `_HsaMemMapFlags` typedef.
- `runtime/hsa-runtime/core/inc/amd_memory_region.h` - remove the orphaned `map_flags()` accessor and `HsaMemMapFlags map_flag_;` member (no longer read after this PR switches to `mem_flag_`).
- `runtime/hsa-runtime/core/runtime/amd_memory_region.cpp` - drop the corresponding `map_flag_.Value = 0;` init.
- `libhsakmt/src/virtio/hsakmt_virtio_memory.c` + `hsakmt_virtio_proto.h` - the virtio path still takes `HsaMemMapFlags`; note the declaration in `hsakmt_virtio.h` was already switched to `HsaMemFlags` in this PR, so the definition/prototype currently mismatch and virtio won't build. Switching the field/param to `HsaMemFlags` (same 32-bit size, so the `VHSAKMT_STATIC_ASSERT_SIZE` still holds) fixes both.

dayatsin-amd:
With the `HsaMemMapFlags` -> `HsaMemFlags` migration in this PR, `HsaMemMapFlags` is no longer referenced by any consumer (verified: 0 remaining references across rocr-runtime after the kfdtest/driver/virtio updates). The typedef in `hsakmttypes.h` can therefore be removed.

I couldn't attach this as a click-apply suggestion because `hsakmttypes.h` isn't part of this PR's diff (GitHub returns `422 Path could not be resolved`). Here is the change to apply:

```diff
--- a/projects/rocr-runtime/libhsakmt/include/hsakmt/hsakmttypes.h
+++ b/projects/rocr-runtime/libhsakmt/include/hsakmt/hsakmttypes.h
@@ -608,6 +608,4 @@
 } HsaMemFlags;
 
-typedef struct _HsaMemMapFlags
-{
-    union
-    {
-        struct
-        {
-            unsigned int Reserved1      :  1; //
-            unsigned int CachePolicy    :  2; // see HSA_CACHING_TYPE
-            unsigned int ReadOnly       :  1; // memory is not modified while mapped
-            	    	    	    	      // allows migration scale-out
-	    unsigned int PageSize	    :  2; // see HSA_PAGE_SIZE, hint to use
-					  // this page size if possible and
-					  // smaller than default
-	    unsigned int HostAccess     :  1; // default = 0: GPU access only
-	    unsigned int Migrate        :  1; // Hint: Allows migration to local mem
-						  // of mapped GPU(s), instead of mapping
-						  // physical location
-            unsigned int Probe          :  1;     // default = 0: Indicates that a range
-                                                  // will be mapped by the process soon,
-						  // but does not initiate a map operation
-						  // may trigger eviction of nonessential
-						  // data from the memory, reduces latency
-						  // “cleanup hint” only, may be ignored
-            unsigned int Reserved       : 23;
-        } ui32;
-        HSAuint32 Value;
-    };
-} HsaMemMapFlags;
-
 typedef struct _HsaGraphicsResourceInfo {
```

Note: this also requires removing the now-orphaned `map_flags()`/`map_flag_` in `amd_memory_region.{h,cpp}` and switching the virtio path (`hsakmt_virtio_memory.c` / `hsakmt_virtio_proto.h`) from `HsaMemMapFlags` to `HsaMemFlags` (same 32-bit size), as noted in my review above.

dayatsin-amd (approved):

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDMemoryTest.cpp
```diff
@@ -3123,9 +3119,10 @@ void KFDMemoryTest::ExportDMABufTest(int gpuNode) {
     buf = reinterpret_cast<HSAuint32 *>(info.MemoryAddress);
     ASSERT_EQ_GPU(info.SizeInBytes, PAGE_SIZE, gpuNode);
 
-    HsaMemMapFlags mapFlags = {0};
+    memFlags.value = 0;
```
yxsamliu: Looks like this should be `memFlags.Value = 0`; `HsaMemFlags` has no lowercase `value`, so this will fail to compile.

TomSang: right

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDMemoryTest.cpp
```diff
@@ -2524,6 +2519,7 @@ void KFDMemoryTest::VramCacheCoherenceWithCPU(int gpuNode) {
 
     unsigned int *buffer;
     HsaMemFlags memFlags = {0};
+    memFlags.ui32.CoarseGrain = 1;
```
yxsamliu: This test says it allocates fine-grained memory, but now sets `CoarseGrain = 1`. Is that intended? Otherwise, this assignment should be removed.

TomSang: fixed

projects/rocr-runtime/libhsakmt/include/hsakmt/hsakmt.h
```diff
@@ -784,7 +784,7 @@ hsaKmtMapMemoryToGPUNodes(
     void*           MemoryAddress,         //IN (page-aligned)
     HSAuint64       MemorySizeInBytes,     //IN (page-aligned)
     HSAuint64*      AlternateVAGPU,        //OUT (page-aligned)
-    HsaMemMapFlags  MemMapFlags,           //IN
+    HsaMemFlags     MemFlags,              //IN
```
yxsamliu, line 787: Could we keep the existing API and add an `Ext` or `V2` version for `CoarseGrain`? hsakmt uses this pattern for other APIs, and GitHub search found an outside caller using the current `HsaMemMapFlags` type.

TomSang: Looks like HsaMemMapFlags is only used by AMD internally.     And it is not referenced even in rocr in Linux .

gandryey (approved):

kentrussell (approved):

yxsamliu (approved):
LGTM

PR #8593, opened 2026-07-15
fix(rocr): Force legacy IPC under Valgrind; Auto-detect Valgrind

ApurvMishra-amd:
## Motivation

dma-buf IPC passes the buffer FD across processes via SCM_RIGHTS, which
Valgrind cannot reproduce, causing `hsa_amd_ipc_memory_attach` to fail with
error 4097.

## Technical Details

Fall back to the legacy IPC path (no FD passing) when running
under Valgrind. No effect outside Valgrind.

## JIRA ID

ROCM-27031

## Test Plan

Tested successfully with: `valgrind --leak-check=full /opt/rocm-7.15.0/bin/rocrtst64 --gtest_filter=rocrtstFunc.IPC`

## Additional Changes

Detect Valgrind at runtime via valgrind.h's `RUNNING_ON_VALGRIND` so the
`HSA_RUNNING_UNDER_VALGRIND` flag activates without requiring the env var to
be set manually. The env var is still honored as an explicit override.

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/util/flag.h
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/util/flag.h
@@ -130,8 +130,13 @@ class Flag {
     visible_gpus_ = os::GetEnvVar("ROCR_VISIBLE_DEVICES");
     filter_visible_gpus_ = os::IsEnvVarSet("ROCR_VISIBLE_DEVICES");
 
-    var = os::GetEnvVar("HSA_RUNNING_UNDER_VALGRIND");
-    running_valgrind_ = (var == "1") ? true : false;
+    // Honor the env var as an explicit override; otherwise auto-detect Valgrind.
+    if (os::IsEnvVarSet("HSA_RUNNING_UNDER_VALGRIND")) {
+      var = os::GetEnvVar("HSA_RUNNING_UNDER_VALGRIND");
+      running_valgrind_ = (var == "1") ? true : false;
+    } else {
+      running_valgrind_ = DetectRunningUnderValgrind();
+    }
 
     var = os::GetEnvVar("HSA_SDMA_WAIT_IDLE");
     sdma_wait_idle_ = (var == "1") ? true : false;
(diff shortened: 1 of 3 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/util/flag.h
```diff
@@ -52,6 +52,14 @@
 #include "core/util/os.h"
 #include "core/util/utils.h"
 
+// Auto-detect Valgrind via its standalone header
+#if defined(__has_include) && __has_include(<valgrind/valgrind.h>)
+#  include <valgrind/valgrind.h>
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** `flag.h` is a widely-transitive utility header. On any build system where valgrind is installed, this `__has_include` guard fires and every TU that includes `flag.h` (directly or transitively) pulls in Valgrind's client-request macros — unexpectedly in production builds where a developer happened to install valgrind for unrelated work.

The valgrind macros themselves are unlikely to collide (they're mostly `VALGRIND_`-prefixed), but the transitive include is surprising for a lean utility header and increases include-chain depth across the runtime.

Suggested fix: move the `#include <valgrind/valgrind.h>` into a `flag.cpp` (even a thin one created specifically to house this), or detect via CMake and pass a compile definition. That isolates the dependency to a single TU.

Copilot flagged the same concern with the same suggestion.

cfreeamd (commented):

`runtime.cpp`: IPC dma-buf disable under Valgrind looks correct — `force_legacy_ipc` is computed once and applied consistently in both assignment paths.

`flag.h`: One concern on the include strategy (see inline comment).

projects/rocr-runtime/runtime/hsa-runtime/core/util/flag.h
ApurvMishra-amd: Please, take a look at the updated changes @cfreeamd. I moved the `#include <valgrind/valgrind.h>` into a _flag.cpp_.

cfreeamd:

The prior [Important] finding is resolved. `<valgrind/valgrind.h>` is no longer pulled into the widely-transitive `flag.h`: it now lives in `flag.cpp` behind a static `Flag::DetectRunningUnderValgrind()` (declared in the header, defined in the .cpp), keeping the `__has_include` guard and the `RUNNING_ON_VALGRIND 0` fallback. The env-var override behavior (including an explicit `HSA_RUNNING_UNDER_VALGRIND=0`) is preserved, and `flag.cpp` was already a compiled TU so no CMake change is needed. The `runtime.cpp` force-legacy-IPC logic is unchanged and still correct (`force_legacy_ipc` computed once, applied consistently in both assignment paths). No remaining concerns from the AI pre-review.

cfreeamd (approved):

PR #8594, opened 2026-07-15
perf(rocjitsu): Optimize MFMA/WMMA execution

atgutier:
## Motivation

Improve performance of key matmul operations

## Technical Details

Eliminate heap allocs and redundant VGPR reads for MFMA/WMMA instruction execution to drastically speedup execution time.

Replace std::vector<Result> with stack-allocated arrays and pre-extract A/B matrix operands into flat stack-local buffers before the M×N compute loop in all mma_exec.h functions. This reduces read_vgpr virtual dispatch calls from O(M×N×K×B) to O((M+N)×K×B) and eliminates per-instruction malloc/free overhead.

## JIRA ID

PR #8594

## Test Plan

Run all CI tests.

## Test Result

CI tests: PASSED

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/mma_exec.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/mma_exec.h
@@ -32,9 +32,11 @@
 
 #include <algorithm>
 #include <bit>
+#include <cassert>
 #include <cmath>
 #include <cstdint>
 #include <limits>
+#include <string>
 #include <vector>
 
 namespace rocjitsu {
@@ -105,6 +107,24 @@ inline float apply_wmma_c_modifier(float acc, uint32_t c_modifier) {
   if (c_modifier & 0x1u)
     acc = -acc;
   return acc;
+}
+
+constexpr uint32_t MAX_MMA_ELEMS = 4096;
+constexpr uint32_t MAX_MMA_OUTPUTS = 2048;
+constexpr uint32_t MAX_PACKED16_DST_WORDS = 1024;
+
+/// @brief Fail safe (in debug AND release) if a matrix shape would overflow one
+/// of the fixed-size stack staging buffers below.
+/// @details The exec_* helpers stage results/operands into fixed C-arrays sized
+/// by the MAX_MMA_* limits, which are large enough for every shape the decoders
+/// can produce for the ISAs this file serves. A bare assert() would be compiled
+/// out under NDEBUG and let an unexpected (e.g. future) shape write out of
+/// bounds; throw instead so it fails deterministically rather than corrupting
+/// the stack. @p needed and @p limit are compared as size_t (callers pass
+/// products already widened to size_t) so the check itself cannot overflow.
+inline void require_mma_capacity(size_t needed, size_t limit, const char *what) {
+  if (needed > limit)
+    throw util::ConfigError(std::string("MMA shape exceeds ") + what + " capacity");
 }
 
 /// Resolve the accumulator source (src2) for MFMA instructions.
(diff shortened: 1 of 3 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/mma_exec.h
```diff
@@ -107,6 +109,24 @@ inline float apply_wmma_c_modifier(float acc, uint32_t c_modifier) {
   return acc;
 }
 
+constexpr uint32_t MAX_MMA_ELEMS = 4096;
+constexpr uint32_t MAX_MMA_OUTPUTS = 2048;
```
kuhar, line 113: Can you derive this from the mma sizes instead of hardcoding the results? I mean something like `128 * 16 * ...`

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/mma_exec.h
```diff
@@ -983,34 +1003,41 @@ void exec_f32_mixed(amdgpu::ComputeUnitCore &cu, uint32_t M, uint32_t N, uint32_
     uint32_t lane;
     uint32_t val;
   };
-  std::vector<Result> results;
-  results.reserve(M * N * B);
+  require_mma_capacity(static_cast<size_t>(M) * N * B, MAX_MMA_OUTPUTS, "output");
+  Result results[MAX_MMA_OUTPUTS];
```
kuhar, line 1007: This reminds me I should land my `util::inline_vector` stuff...

emulation/rocjitsu/tests/mfma_benchmark.cpp
```diff
@@ -0,0 +1,192 @@
...
+TEST(MfmaBenchmark, F32_32x32x8_F16) {
+  MfmaBenchFixture fx;
+  ASSERT_NE(fx.cu, nullptr);
+  ASSERT_NE(fx.wf, nullptr);
+  constexpr uint64_t SEED = 0xAF3A'3232'0008ULL;
+  fx.seed_f16(SEED, 4, 4, 16);
+  uint32_t dst = fx.vb + DST_OFF, s0 = fx.vb + S0_OFF, s1 = fx.vb + S1_OFF, s2 = fx.vb + DST_OFF;
+  auto s = time_mfma(
+      [&] {
+        amdgpu::exec_f32(*fx.cu, 32, 32, 8, 1, 16, dst, s0, s1, s2, amdgpu::extract_f16,
```
kuhar, line 136: I think we should benchmark the actual CDNA4 instruction paths so these MIPS labels measure what they claim. The f16 cases call generic `exec_f32`, while CDNA4 dispatches the 32x32x8 and 16x16x32 shapes through `exec_f32_mfma_f16_spec`; the 4x4x4 opcode uses `B=16`, not `B=4`; and CDNA4 has no 32x32x8 i8 decoder (its 32x32 i8 shapes use K=16 or 32). Using the decoder path or the exact specialization/shape tuples would make these results meaningful.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/shared/mma_exec.h
kuhar, line 1007: This would be safer if the stack budget covered the full helper frame. In the current O2 object, `exec_f32_mixed` reserves `0xd200` bytes (about 52.5 KiB), and its scalar lambda adds `0x8028` bytes (about 32 KiB) when called, so the scalar path uses roughly 85 KiB before callers. The 48 KiB `static_assert` counts only the SIMD scratch arrays and misses `results` plus the scalar operand buffers. Could we size storage from the active shape or share scratch storage, then enforce the intended whole-frame limit?

kuhar (commented):
Can we also remove the remaining per-call allocations in `mma_exec.h:1433` (`exec_gfx11_wmma_f32_mixed`) and `mma_exec.h:1478` (`exec_gfx11_wmma_packed16`)? Both still use `std::vector<Result>`; the packed16 helper now uses fixed `words`/`masks`, so it still performs a heap allocation for result staging. The commit says the conversion covers all `mma_exec.h` functions, and these paths have the same bounded `M * N` output shape as the converted GFX12 helpers.

kuhar (commented):
I think we should benchmark and tune the fixed capacities separately for MFMA and WMMA before treating this as a uniform win. On a Threadripper PRO 9995WX, I built the recorded base (`0522418d`) and this head (`070723e0`) with Clang 23 RelWithDebInfo, pinned the unchanged non-specialized benchmark suites to one core, and took the median of three 4,000-iteration samples. Normal SIMD execution improved by 1.18x geometric mean across eight MFMA cases, but only 1.00x across ten WMMA/SWMMAC cases.

The code shape helps explain the asymmetry: WMMA already hoists A/B/C into about 28 KiB of fixed SIMD scratch, so this change mainly removes one `M * N` result allocation. At the same time, `Result[MAX_MMA_OUTPUTS]` grows the generic WMMA frames from about 29 KiB to 53 KiB (packed16 to about 62 KiB). Since gfx1250 already has the largest WMMA tiles, its current call sites need at most 512 outputs, and most need 256, versus the shared capacity of 2048; packed16 also clears fixed 4 KiB `words` and 1 KiB `masks` arrays. A reduced-capacity ablation lowered those frames to 35-37 KiB but did not improve the overall WMMA geometric mean, so the oversized storage is a pressure point rather than the whole explanation. Could we derive/split these capacities from the actual ISA shape bounds and profile/rebenchmark WMMA separately?

newling:
> This reduces read_vgpr virtual dispatch calls from O(M×N×K×B) to O((M+N)×K×B) and 

That sounds like an uncontroversial win, nice. There is some fast path specialization for mfma already (from @lialan) I assumed it did this already but perhaps not in all cases and perhaps it only improves the inner-loop SIMD. 

If I land https://github.com/ROCm/rocm-systems/pull/8417 this PR will not rebase cleanly and might need to semantic changes. Is that ok Tony?

PR #8597, opened 2026-07-15
fix(rocprofiler-systems): Fix AMDSMI version detection in CMake files

ajanicijamd:
## Motivation

* In TheRock builds, `ROCPROFSYS_BUILD_AINIC` was always `OFF`, silently disabling AI NIC profiling support even when `ROCPROFSYS_USE_AINIC=ON`.
* The root cause was in `projects/rocprofiler-systems/cmake/Packages.cmake`: the AMD SMI version detection constructed a hardcoded path `${ROCM_PATH}/include/amd_smi/amdsmi.h`. TheRock explicitly unsets `ROCM_PATH` for all subproject builds to prevent accidental use of system-installed ROCm, so the path resolved incorrectly, the header was never found, version detection returned empty strings, and the `>= 26.3` version check failed.

## Technical Details

**CMake detection (`cmake/Packages.cmake`)**
* `find_package(amd_smi REQUIRED)` runs immediately before the version-detection block and always succeeds; the config-mode package already reports its version in the `amd_smi_VERSION` variable.
* Read the version directly from `amd_smi_VERSION` instead of locating and regex-parsing `amdsmi.h`. This has no dependency on `ROCM_PATH` (or any environment variable) and no header-path/regex fragility, so it works in all build environments (TheRock, standalone, `/opt/rocm`). When the package reports no version, fall back to `0.0` (AI NIC unsupported) with a visible status message rather than failing silently.
* Gate `ROCPROFSYS_BUILD_AINIC` on `amd_smi_VERSION VERSION_GREATER_EQUAL 26.3` together with `ROCPROFSYS_USE_AINIC`.

**Expose compiled-in support as a runtime signal**
* Promote `ROCPROFSYS_BUILD_AINIC=1` to the shared `rocprofiler-systems-compile-definitions` INTERFACE target so it reaches core/avail, not just the pmc/amd_smi backends; drop the now-redundant per-target define on the amd_smi backend.
* Register the `ROCPROFSYS_USE_AINIC` and `ROCPROFSYS_SAMPLING_AINICS` settings only when `ROCPROFSYS_BUILD_AINIC=ON`, so their presence in `rocprof-sys-avail --settings` is a direct, hardware-independent indicator of compile-time support. Add `get_ainic_supported()` and macro-guard `get_sampling_ainics()` for the disabled case.

**Tests (`tests/pytest/test_ainic.py`)**
* Merge the compile-time check and the RDMA performance-track check into a single `TestAINIC` class (replacing `test_ainic_compiled.py` and `test_ainic_perf.py`). The compile-time check asserts that `rocprof-sys-avail --settings` lists the AI NIC settings — no `nm` symbol probing or dependency on AMD SMI symbol names. The standalone `check_ainic_support.sh` is removed.
* Split the `ainic` marker into `ainic` (informational label) and `ainic_required` (skips when no AI NIC hardware is present). CTest names: `ainic-settings-present` (hardware-independent, gated on AMD SMI >= 26.3) and `ainic-performance-tracks` (`network`-labeled, hardware-gated).

## Issue Tracking

JIRA ID: ROCM-28005

## Test Plan

**Compile-time support check (no hardware required)**

```
ctest -V -R ainic-settings-present
```

Passes on an AI NIC-enabled build; skips (rather than fails) on builds against AMD SMI < 26.3 via `@pytest.mark.amdsmi_min_version("26.3")`.

**Manual equivalent**

```
rocprof-sys-avail --settings | grep ROCPROFSYS_USE_AINIC
```

Lists the setting when AI NIC support is compiled in; produces no output otherwise.

**Performance tracks (requires AI NIC hardware)**

```
ctest -V -R ainic-performance-tracks
```

Skipped automatically when no AI NIC devices are present; excluded from CI via `-LE "network|gpu"`.

**Configure log check (TheRock build)**

```
-- AMD SMI version detected: <major>.<minor>
-- AINIC support enabled (AMD SMI >= 26.3)
```

And in `CMakeCache.txt`: `ROCPROFSYS_BUILD_AINIC:INTERNAL=ON`.

## Test Result

Verified on a build with AMD SMI 26.5:

```
-- AMD SMI version detected: 26.5
-- AINIC support enabled (AMD SMI >= 26.3)
```

* `CMakeCache.txt` confirms `ROCPROFSYS_BUILD_AINIC:INTERNAL=ON`.
* `ctest -V -R ainic-settings-present` passes; `rocprof-sys-avail --settings` lists `ROCPROFSYS_USE_AINIC` and `ROCPROFSYS_SAMPLING_AINICS`.
* On an AI NIC-off build (`-DROCPROFSYS_USE_AINIC=OFF`), `rocprof-sys-avail --settings` lists no AI NIC settings, confirming the negative case.

## Submission Checklist


projects/rocprofiler-systems/tests/pytest/test_ainic_compiled.py
```diff
@@ -0,0 +1,85 @@
+# Copyright (c) Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""
+Tests that AI NIC support was compiled into librocprof-sys.so.
+
+This test does NOT require NIC hardware to be present — it checks the
+binary artifacts for compile-time indicators. It will fail if the library
+was built with ROCPROFSYS_BUILD_AINIC=OFF (e.g. due to ROCM_PATH not
+being set during the build, as in ROCM-28005).
```
dgaliffiAMD:
```suggestion
was built with ROCPROFSYS_BUILD_AINIC=OFF.
```
Removing reference of JIRA ticket from comment.

projects/rocprofiler-systems/tests/pytest/test_ainic_compiled.py
```diff
@@ -0,0 +1,85 @@
...
+@pytest.fixture(scope="module")
+def librocprof_sys(rocprof_config: RocprofsysConfig) -> Path:
+    lib = _find_librocprof_sys(rocprof_config.rocprofsys_lib_dir)
+    return lib
+
+
+class TestAinicCompiledIn:
+    """Verify AI NIC support was compiled into librocprof-sys.so."""
+
+    def test_ainic_dynamic_symbols_present(self, librocprof_sys: Path):
```
dgaliffiAMD:
```suggestion
class TestAiNic:
    """Verify AI NIC support was compiled into librocprof-sys.so."""

    def test_symbols_present(self, librocprof_sys: Path):
```

This will produce a more readable test name: `ainic-symbols-present` rather than `ainiccompiledin-ainic-dynamic-symbols-present`.

projects/rocprofiler-systems/tests/pytest/test_ainic_compiled.py
```diff
@@ -0,0 +1,85 @@
...
+    "amdsmi_get_nic_port_info",
+]
+
+# String literals embedded only when ROCPROFSYS_BUILD_AINIC=ON.
+_AINIC_STRINGS = [
+    "ROCPROFSYS_USE_AINIC",
+    "ainic",
+]
+
+
```
dgaliffiAMD:
```suggestion
```
`_AINIC_STRINGS ` does not appear to be used elsewhere.

projects/rocprofiler-systems/docs/how-to/nic-profiling.rst
```diff
@@ -178,6 +178,76 @@ AI NIC interfaces support the Remote Direct Memory Access (RDMA) standard. RDMA
 
    -D ROCPROFSYS_USE_AINIC=OFF
 
+Verifying AI NIC compile-time support
```
dgaliffiAMD, line 187: I don't think this is the right approach for checking if AI NIC is enabled in their build. We can't ask the general user to `nm` to search for symbols in our libraries. The `rocprof-sys-avail` tool was built for these types of queries; that is what we should use for this test and documentation instead.

dgaliffiAMD: It turns out `rocprof-sys-avail` had a short coming. It would advertise the AI-NIC related settings whether or not the build actually supported it. I would consider this a bug. 
Please see on my test branch (https://github.com/ROCm/rocm-systems/commits/users/dgaliffi/review/pr-8597/):
- https://github.com/ROCm/rocm-systems/commit/1ddb489ff882a4f66ac1778ba0a98d9736a0248d - expose compile-in AI-NIC support via config settings. 
- https://github.com/ROCm/rocm-systems/commit/a7d876d736548941be10daae0241a3480ddae1da - updated test and docs.

dgaliffiAMD (changes requested):

projects/rocprofiler-systems/tests/pytest/test_ainic_compiled.py
kcossett-amd, line 1: I don't like that we have `test_ainic_perf.py` and `test_ainic_compiled.py` that each contain a single class.
I'd rather we combine both of these into a single `test_ainic.py` file. 

You would just needto change the name of the `ainic` functional marker would to `require_ainic_card` (or something), that way `ainic` marker can be used as a module level marker.

However, this is not a blocker, I can always do it down the line

projects/rocprofiler-systems/tests/pytest/test_ainic_compiled.py
```diff
@@ -0,0 +1,85 @@
...
+        These symbols are undefined references (U) in librocprof-sys.so —
+        they are resolved at runtime from libamd_smi.so. Their presence
+        proves that the NIC collector code was compiled in.
+        """
+        result = subprocess.run(
+            ["nm", "-D", str(librocprof_sys)],
+            capture_output=True,
+            text=True,
+        )
```
kcossett-amd:
```suggestion
        result = subprocess.run(
            ["nm", "-D", str(librocprof_sys)],
            timeout=5,
            capture_output=True,
            text=True,
        )
```
Adding a timeout

projects/rocprofiler-systems/tests/check_ainic_support.sh
kcossett-amd, line 1: Does this script need to be installed (i.e. distributed as part of TheRock)?

projects/rocprofiler-systems/tests/pytest/test_ainic_compiled.py
```diff
@@ -0,0 +1,85 @@
...
+    raise FileNotFoundError(f"librocprof-sys.so not found under {lib_dir}")
+
+
+@pytest.fixture(scope="module")
+def librocprof_sys(rocprof_config: RocprofsysConfig) -> Path:
+    lib = _find_librocprof_sys(rocprof_config.rocprofsys_lib_dir)
+    return lib
+
+
+class TestAinicCompiledIn:
```
kcossett-amd: I'm not sure how long these tests take to run, but the default given is `300` seconds, which seems too long.
Mark the class with `@pytest.mark.timeout(<shorter time in seconds>)`.

dgaliffiAMD (approved):

dgaliffiAMD:
Thank you for handling the "cli-help-topic-sampling" test failure.

PR #8602, opened 2026-07-15
fix(rocprofiler-systems): use cgroup CPU affinity for pytest num_procs

anujshuk-amd:
## Motivation
The fix makes pytest report the same CPU count the build and runtime actually use, instead of the whole machine's CPU count.

## Technical Details
On ppac, a Slurm job with `--cpus-per-task=16` was getting two different CPU counts.

- CMake (`ProcessorCount` at configure time) with 16 (CPUs in cgroup) compute `ROCPROFSYS_MAX_THREADS = 2048`
- `os.cpu_count()` calculate with 192 (full node) which expect 4096

Reason for test failure : `os.cpu_count()` returns the machine’s logical CPU count (192 on ppac), not what your job is allowed to use. Tests fails because CMake and the binary used allocated CPU count. The pytest  expected full node CPU count.

The fix aligns pytest CPU detection with Slurm job allocation and CMake configure-time behavior, so `thread-limit` tests expect same counts as Cmake configuration.

## JIRA ID

JIRA ID : AIPROFSYST-589

## Test Plan

Ran on aac6 and other linux machines

## Test Result

Passed locally on aac6 and other linux machines

## Submission Checklist


projects/rocprofiler-systems/tests/pytest/rocprofsys/capabilities.py
```diff
@@ -274,11 +274,13 @@ def ucx_availability(self) -> bool:
...
-        if num_procs_real is None:
-            return 2
-        return num_procs_real
+        """Number of processors available to this process.
+
+        Uses sched_getaffinity so Slurm/cgroup/taskset limits match CMake
+        ProcessorCount and runtime thread-pool sizing.
+        """
+        count = len(os.sched_getaffinity(0))
+        return count if count > 0 else 2
```
sputhala-amd, line 288: As copilot suggested lets have a fallback path.

sputhala-amd (commented):
This fix probably will not be helpful if the project is built on one machine and the tests are run on a different machine. However in a cluster node, with the slurm setting you mention, this fix will be needed.

projects/rocprofiler-systems/tests/pytest/rocprofsys/capabilities.py
anujshuk-amd: Added for safer side.

dgaliffiAMD:
Please include in the PR description which test failures this is addressing? ie., the "thread-limit" family of tests.

dgaliffiAMD:
> Please include in the PR description which test failures this is addressing? ie., the "thread-limit" family of tests.

Disregard. It was there, I just didn't see it. I modified the description slightly to have it more obvious.

dgaliffiAMD (approved):

kcossett-amd (approved):

PR #8604, opened 2026-07-15
perf(rocprofiler-sdk): cache perfetto event names for static interning.

srihari-uttanur:
## Motivation

The pftrace file generated from rocprofv3 is smaller in size than the one generated by rocpd. After investigation found out that the difference in the pftrace files is because the one generated from rocpd uses dynamic string copy which creates string deduplication whereas the one generated from rocprofv3 uses static copies.

## Technical Details

rocpd2pftrace was emitting DB-derived event names as Perfetto DynamicString values. DynamicString writes the full event name into each TrackEvent packet, while the direct rocprofv3 Perfetto path uses StaticString and lets Perfetto emit interned name IDs.

Add a local static_string_cache in the rocpd Perfetto writer so event names loaded from the database have stable storage for the lifetime of trace generation. The converter can then pass those cached names as StaticString values, allowing Perfetto to intern repeated names safely.

Apply the cache to region, sample, memory-copy, and kernel-dispatch event names. This reduces repeated string payload in generated pftrace files without changing event timing, tracks, annotations, categories, or other trace metadata.

## JIRA ID

ROCM-21204

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-sdk/source/lib/python/rocpd/source/perfetto.cpp
+++ b/projects/rocprofiler-sdk/source/lib/python/rocpd/source/perfetto.cpp
@@ -134,10 +134,10 @@ PerfettoSession::PerfettoSession(const tool::output_config& output_cfg, sqlite3*
 ::perfetto::StaticString
 PerfettoSession::get_static_event_name(const std::string& value) const
 {
-    // insert() is a no-op when the name is already cached; the returned iterator
+    // emplace() is a no-op when the name is already cached; the returned iterator
     // points to storage that stays valid for the whole PerfettoSession lifetime,
     // which is what perfetto::StaticString interning requires.
-    return ::perfetto::StaticString{static_event_names.insert(value).first->c_str()};
+    return ::perfetto::StaticString{static_event_names.emplace(value).first->c_str()};
 }
 
 PerfettoSession::~PerfettoSession()
--- a/projects/rocprofiler-sdk/source/lib/python/rocpd/source/perfetto.cpp
+++ b/projects/rocprofiler-sdk/source/lib/python/rocpd/source/perfetto.cpp
@@ -43,6 +43,10 @@
 #include <atomic>
 #include <future>
 #include <mutex>
+#include <set>
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
 #include <variant>
 #include <vector>
 
@@ -71,6 +75,7 @@ get_hash_id(Tp&& _val)
     else
         return get_hash_id(*_val);
 }
+
 }  // namespace
 
 PerfettoSession::PerfettoSession(const tool::output_config& output_cfg, sqlite3* conn)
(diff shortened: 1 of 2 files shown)


projects/rocprofiler-sdk/source/lib/python/rocpd/source/perfetto.cpp
```diff
@@ -126,6 +131,15 @@ PerfettoSession::PerfettoSession(const tool::output_config& output_cfg, sqlite3*
     tracing_session->StartBlocking();
 }
 
+::perfetto::StaticString
+PerfettoSession::get_static_event_name(const std::string& value) const
```
bgopesh, line 135: Add a brief comment  stating it is not thread-safe and relies on single-threaded write_perfetto invocation,

projects/rocprofiler-sdk/source/lib/python/rocpd/source/perfetto.hpp
```diff
@@ -45,9 +47,15 @@ namespace tool = ::rocprofiler::tool;
 
 struct PerfettoSession
 {
+private:
```
bgopesh: As a rule in rocprofiler-sdk, "Headers should declare simple structs + standalone functions, not classes with public/protected/private  members."

Consider keeping the struct's members public (consistent with                                 
  tracing_session/config/connection right below it) and just adding the method, or moving the cache behind a pimpl/opaque handle if hiding is     
  desired. At minimum, the mixed private/public split here is inconsistent with the rest of the struct.

bgopesh (commented):
some nitpicks but fine overall

bgopesh (approved):

projects/rocprofiler-sdk/source/lib/python/rocpd/source/perfetto.cpp
```diff
@@ -126,6 +131,15 @@ PerfettoSession::PerfettoSession(const tool::output_config& output_cfg, sqlite3*
...
+{
+    // insert() is a no-op when the name is already cached; the returned iterator
+    // points to storage that stays valid for the whole PerfettoSession lifetime,
+    // which is what perfetto::StaticString interning requires.
+    return ::perfetto::StaticString{static_event_names.insert(value).first->c_str()};
```
jrmadsen: use `emplace` instead of `insert`

projects/rocprofiler-sdk/source/lib/python/rocpd/source/perfetto.hpp
```diff
@@ -48,6 +50,11 @@ struct PerfettoSession
     PerfettoSession(const tool::output_config&, sqlite3* connection);
     ~PerfettoSession();
 
+    // NOTE: not thread-safe; assumes write_perfetto() is invoked single-threaded per session.
+    ::perfetto::StaticString get_static_event_name(const std::string& value) const;
+
+    mutable std::unordered_set<std::string> static_event_names = {};
```
jrmadsen: Move to end of the struct with the other members.

jrmadsen (dismissed):

projects/rocprofiler-sdk/source/lib/python/rocpd/source/perfetto.cpp
```diff
@@ -43,6 +43,10 @@
 #include <atomic>
 #include <future>
 #include <mutex>
+#include <set>
+#include <string>
+#include <unordered_map>
```
sauverma93, line 48: unordered_map may not be needed. Also would be good idea to check if other headers added here are needed

sauverma93 (approved):
Hi @srihari-uttanur, approving the changes. Just make sure to address the comments left by "therock-pr-bot". Thanks!

projects/rocprofiler-sdk/source/lib/python/rocpd/source/perfetto.cpp
srihari-uttanur: previously these headers were not strictly required to compile as they were transitively getting included and this PR included them during the refactor. As per the IWYU principle I have kept these headers. If in the future transitive header inclusion is broken then this file should not be affected. Hence keeping them.

PR #5344, opened 2026-04-23
rocprofv3: add PC sampling attach/detach integration tests

julijose-amd:
## Motivation

PC sampling was recently extended to support the rocprofv3 attach/detach workflow (#3921 / commit d86e9a3097: "Enable PC sampling with attach/detach"). That code path currently has no integration tests. This PR adds them.

## Technical Details

 Two new ctest cases under `tests/rocprofv3/attachment/attach-pc-sampling/`:
  - `attach-pc-sampling-host-trap`  — 5s attach, host_trap method
  - `attach-pc-sampling-stochastic` — 5s attach, stochastic method
  
Each test launches `exec-mask-manipulation` in the background, attaches rocprofv3 with `--pc-sampling-*` flags and `--attach-duration-msec 5000`, detaches, and validates the resulting CSV.
  
**Validation.** In attach mode the first captured dispatch surfaces as `Correlation_Id == 1` rather than its true wrap-mode index. The new `validate.py`:
  1. Drops the `CID == 0` prefix (unresolved samples emitted before profiling actually started tracking).
  2. Infers `offset = popcount(Exec_Mask @ CID==1) - 1` (relies on the loop kernels' block-size-equals-popcount design property).
  3. Shifts `Correlation_Id` and `Dispatch_Id` by that offset.
  4. Reuses the existing wrap-mode helpers from `tests/pytest-packages/pc_sampling/exec_mask_manipulation/csv.py` verbatim on the shifted DataFrame.

**App change.** `exec_mask_manipulation.cpp` now takes an optional `iter_num` in `argv[1]` (default: the existing `ITER_NUM` macro), so the app can be kept alive long enough to cover the 5s attach window. Wrap-mode invocations without arguments pick up the default and are unaffected.

## Test Result

- [x] `ctest -R attach-pc-sampling` — host-trap and stochastic both pass
- [x]  Reproduced the attach workflow end-to-end with standalone scripts and manually verified the CSV output against the validator logic before integrating the checks into ctest.  

## Follow-ups (tracked, not in this PR)
  - Tune `iter_num` so the attach window reliably captures kernel3, enabling the `v_rcp_f64`/`v_rcp_f32` decoding checks.
  -  Add PC sampling reattach coverage (multiple attach/detach cycles on the same process, mirror of `attach-twice`).
  - Add stochastic-specific gfx9 CSV validation (`Wave_Issued_Instruction`, `Instruction_Type`, `Stall_Reason`, `Wave_Count`).
[...]

PR #8606, opened 2026-07-15
test(rocprofv3): add PC sampling attach/detach integration tests

julijose-amd:
## Motivation

rocprofv3 supports PC sampling in attach mode, but that path had no integration coverage. This PR aims to add it for both host_trap and stochastic methods. Supersedes #5344 (closed), addressing its review feedback.

## Technical Details

New tests under `tests/rocprofv3/attachment/attach-pc-sampling/` (`host-trap/` + `stochastic/`), driven by a shared `run_attach_pc_sampling_test.sh`: launch `exec-mask-manipulation` in the background, wait for it to become attach-ready, attach rocprofv3 with `--pc-sampling-*` for a fixed window, detach, and validate the output.

- The `exec-mask-manipulation` app takes an optional iteration count so it stays alive across the attach window; the default is unchanged, so existing tests are unaffected.
- The shared `exec_mask_manipulation_validate_csv` is extended with `wave_size`/`num_kernels` parameters — it reads the wave size from `out_agent_info.csv` and handles partial captures, replacing the hardcoded 33/65-kernel assumption.
- Attach output is normalized before the shared validators run: drop the unresolved `Correlation_Id == 0` prefix, then shift `Correlation_Id`/`Dispatch_Id` by the attach offset back into wrap-mode alignment.
- CSV validation: after normalization, reuses the shared exec-mask checks — instruction comments, correlation/dispatch relation, and per-kernel exec-mask patterns (popcount == correlation id); plus the gfx9 stochastic-specific checks (wave count, instruction types, arbiter state) for the stochastic method.
- JSON validation scope: JSON is checked via CSV↔JSON parity, per-record checks (inst_index, correlation/dispatch consistency, exec_mask), and the gfx9 stochastic checks. We intentionally don't run the full validate_json_exec_mask_manipulation — it assumes a complete run and un-normalized IDs, so applying it to attach data would mean a complete re-do for offset normalization on the nested JSON. Currently the parity check plus the CSV validation cover the same records.

**Note:** the test normalizes attach-mode correlation IDs to match their current behavior, as it starts from 1 at the attach point, preceded by an unresolved Correlation_Id == 0 prefix. If this behavior changes in the tool (for example, absolute dispatch IDs or a different handling of the prefix), `normalize_attach_csv` will need to be updated accordingly.

## JIRA ID

JIRA ID: AIROCVAL-54

## Test Plan

`ctest -R attach-pc-sampling`

## Test Result

All 8 tests pass on MI300x (gfx942)

### DIFF
--- a/projects/rocprofiler-sdk/tests/pytest-packages/pc_sampling/attach/csv.py
+++ b/projects/rocprofiler-sdk/tests/pytest-packages/pc_sampling/attach/csv.py
@@ -1,0 +1,63 @@
+# MIT License
+#
+# Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from __future__ import absolute_import
+
+
+def normalize_attach_csv(df):
+    # In attach mode the profiler joins mid-run. Early samples that were not yet mapped
+    # to a dispatch carry Correlation_Id == 0 -- drop that prefix. The first mapped
+    # dispatch is numbered Correlation_Id == 1 instead of its true index, so shift
+    # Correlation_Id and Dispatch_Id by that offset so the shared exec-mask validators apply unchanged.
+    assert not df.empty, "attach-mode PC sampling CSV is empty"
+
+    df = df.sort_values("Sample_Timestamp").reset_index(drop=True)
+    resolved = df[df["Correlation_Id"] != 0]
+    assert not resolved.empty, "no resolved samples (all Correlation_Id == 0)"
+    first_resolved_idx = int(resolved.index.min())
+    # the dropped prefix must be genuinely unresolved (Dispatch_Id == 0)
+    prefix = df.iloc[:first_resolved_idx]
+    assert (
+        prefix["Dispatch_Id"] == 0
+    ).all(), "unresolved prefix has a non-zero Dispatch_Id"
+    df = df.iloc[first_resolved_idx:].copy()
+    assert (
+        df["Correlation_Id"] != 0
+    ).all(), "Correlation_Id == 0 samples found after the first resolved sample"
+
+    cid1 = df[df["Correlation_Id"] == 1]
+    assert not cid1.empty, (
+        "no Correlation_Id == 1 samples; the attach window missed the first captured "
+        "dispatch and the offset cannot be inferred"
+    )
+    # loop kernel i launches i active threads, so popcount(Exec_Mask) == true index
+    offset = int(cid1["Exec_Mask"].apply(lambda m: bin(int(m)).count("1")).mode()[0]) - 1
+    df["Correlation_Id"] = df["Correlation_Id"].astype(int) + offset
+    df["Dispatch_Id"] = df["Dispatch_Id"].astype(int) + offset
+
+    # a gap in the dispatch range means a lost correlation-id mapping during attach
+    lo, hi = int(df["Dispatch_Id"].min()), int(df["Dispatch_Id"].max())
+    assert df["Dispatch_Id"].nunique() == hi - lo + 1, (
+        f"gap in Dispatch_Id range [{lo}, {hi}]: "
+        f"{df['Dispatch_Id'].nunique()} unique, expected {hi - lo + 1}"
+    )
+    return df, hi
--- a/projects/rocprofiler-sdk/tests/pytest-packages/pc_sampling/exec_mask_manipulation/csv.py
+++ b/projects/rocprofiler-sdk/tests/pytest-packages/pc_sampling/exec_mask_manipulation/csv.py
@@ -149,7 +149,9 @@ def validate_exec_mask_based_on_correlation_id(df):
     ).all(), "Exec_Mask does not match expected mask derived from Correlation_Id for all samples"
 
 
-def exec_mask_manipulation_validate_csv(df, all_sampled=False):
+def exec_mask_manipulation_validate_csv(
+    df, all_sampled=False, wave_size=None, num_kernels=None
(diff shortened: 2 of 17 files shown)


projects/rocprofiler-sdk/tests/pytest-packages/pc_sampling/attach/csv.py
```diff
@@ -0,0 +1,63 @@
...
+    ).all(), "unresolved prefix has a non-zero Dispatch_Id"
+    df = df.iloc[first_resolved_idx:].copy()
+    assert (
+        df["Correlation_Id"] != 0
+    ).all(), "Correlation_Id == 0 samples found after the first resolved sample"
+
+    cid1 = df[df["Correlation_Id"] == 1]
+    assert not cid1.empty, (
+        "no Correlation_Id == 1 samples; the attach window missed the first captured "
+        "dispatch and the offset cannot be inferred"
```
vlaindic, line 50:
```suggestion
        "no Correlation_Id == 1 samples; the attach window missed any kernel dispatch "
        "and the offset cannot be inferred"
```

projects/rocprofiler-sdk/tests/pytest-packages/pc_sampling/attach/csv.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+    # loop kernel i launches i active threads, so popcount(Exec_Mask) == true index
```
vlaindic, line 52: How about a bit more explicit comment?


```suggestion
  # kernel i launches i active threads, so popcount(Exec_Mask) gives the original
  # (pre-attach) dispatch index. Correlation_Id == 1 in attach view, so
  # offset = original_index - 1
```

projects/rocprofiler-sdk/tests/pytest-packages/pc_sampling/exec_mask_manipulation/csv.py
```diff
@@ -158,40 +160,55 @@ def exec_mask_manipulation_validate_csv(df, all_sampled=False):
...
-            == unique_kernels_num
+            == num_kernels
         )
 
+    is_full_capture = num_kernels == wave_size + 1
+
+    # The last kernel (kernel3, the v_rcp kernel) is only present on a full capture.
+    # On a partial attach capture, validate the exec masks on the captured kernels.
+    if not is_full_capture:
+        validate_exec_mask_based_on_correlation_id(samples_cid_non_zero_df.copy())
```
vlaindic, line 195: Are we sure that this guard from the case where we don't capture all kernels, but do capture the last one? In that case, v_rcp kernel will be part of the samples_cid_non_zero_df and we'll apply validate_exec_mask_based_on_correlation_id on it, which might fail.

projects/rocprofiler-sdk/tests/pytest-packages/pc_sampling/exec_mask_manipulation/csv.py
```diff
@@ -158,40 +160,55 @@ def exec_mask_manipulation_validate_csv(df, all_sampled=False):
...
             len(samples_cid_non_zero_df["Correlation_Id"].astype(int).unique())
-            == unique_kernels_num
+            == num_kernels
         )
 
+    is_full_capture = num_kernels == wave_size + 1
+
+    # The last kernel (kernel3, the v_rcp kernel) is only present on a full capture.
+    # On a partial attach capture, validate the exec masks on the captured kernels.
+    if not is_full_capture:
```
vlaindic, line 194: Somehow, I feel that even without this guard, the would work as it will separate kernels into the two groups:
* kernels before v_rcp that will be passed to the `validate_exec_mask_based_on_correlation_id`
* v_rcp kernel that will be checked separately.

vlaindic (changes requested):
Thanks for the PR. I agree with you that overriding the existing JSON test would be to complicated to support attach, so the minimal checks you implemented is sufficient, considering that the stochastic checks apply to the full JSON. 

Overall, the PR look good to me, but there are couple nits worth addressing + plus AI review:

## IMPORTANT (2)

### 1. `run_kernel(size_t)` passes to kernels taking `const int` — silent narrowing + unhandled exception

`run_kernel` was changed to take `size_t iter_num`, but `kernel1/2/3` still take `const int iter_num`. A user-supplied value larger than `INT_MAX` parsed via `std::stoull` truncates silently when passed to the GPU kernels. Also, `std::stoull` throws `std::invalid_argument`/`std::out_of_range` on bad input — unhandled here, causing an opaque crash rather than a diagnostic.

Two viable fixes:

**Option A — keep everything `int`** (minimal change, matches the kernel signatures; `1048576` fits comfortably in `int`):
```cpp
void run_kernel(int iter_num) { ... }

int main(int argc, char* argv[])
{
    int iter_num = ITER_NUM;
    if(argc > 1)
    {
        try { iter_num = std::stoi(argv[1]); }
        catch(const std::exception& e) {
            std::cerr << "invalid iteration count '" << argv[1] << "': " << e.what() << "\n";
            return 1;
        }
    }
    run_kernel(iter_num);
    return 0;
}
```

**Option B — change everything to `size_t`** (semantically more correct — iteration counts are non-negative, removes type mismatch entirely, but ripples through three GPU kernel signatures):
```cpp
__global__ void kernel1(const float c, const size_t iter_num) { ... }
__global__ void kernel2(const float c, const size_t iter_num) { ... }
__global__ void kernel3(const float c, const size_t iter_num) { ... }
void run_kernel(size_t iter_num) { ... }
```

Option A is lower diff and lower risk for a test binary. Either way, the unhandled exception from `std::stoi`/`std::stoull` on bad input needs the try/catch.

---

### 2. `<string>` not explicitly included despite using `std::stoull`

`std::stoull` lives in `<string>` — the file pulls it in only via transitive includes (likely HIP or iostream). This is fragile and can silently break with a toolchain update.

**Fix**: Add `#include <string>` (and `#include <stdexcept>` if catching parse exceptions) to `exec_mask_manipulation.cpp`.

---

## SUGGESTIONS (3)

### 3. `wave_size` fixture skips without calling `_check_skip` — inconsistent ptrace skip reason

In both `conftest.py` files, `wave_size` does not call `_check_skip(request)`. If a test's first-used fixture is `wave_size`, the ptrace-disabled scenario skips with `"agent info unavailable"` instead of `"attach tests unavailable (insufficient ptrace permissions)"`.

**Fix**: Add `_check_skip(request)` as the first statement in `wave_size` in both `conftest.py` files.

---

### 4. Local imports inside test functions lack explanation

`stochastic/validate.py:63-65` and `76-78` defer imports of `validate_stochastic_samples_csv` / `validate_stochastic_samples_json` inside the test functions. If intentional (module only importable on certain configs, e.g. gfx9-only), add a one-line comment explaining why; otherwise move to module-level imports.

---

### 5. Magic positional args to shell script are opaque in CMakeLists

`host-trap/CMakeLists.txt:57` and `stochastic/CMakeLists.txt:57` pass args like `warning out host_trap time 1000 5000 1048576` positionally. A reader must cross-reference the script header to map each position to its name. The `1048576` iteration count also appears in three places (C++ default and both CMake invocations).

**Suggestion**: Add a brief comment above the `COMMAND` in each CMakeLists mapping each positional arg to its name.

---

## Future Improvements (non-blocking)

### Conftest fixtures are copy-pasted between host-trap and stochastic

`host-trap/conftest.py` and `stochastic/conftest.py` are nearly identical. `pytest_addoption`, `_check_skip`, `input_json`, and `wave_size` are byte-for-byte copies. Only `input_csv` differs (dtype map) and stochastic adds `input_agent_info_csv`. Any future fix to the skip logic or wave-size parsing must be made in two places.

**Note on approach**: A shared `conftest.py` placed in the parent `attach-pc-sampling/` directory will not work — each subdirectory has its own `pytest.ini`, which sets the rootdir, and pytest stops conftest discovery at the rootdir.

**Suggested approach**: Export shared helpers as a plain Python module (not a conftest) from `tests/pytest-packages/pc_sampling/attach/` — e.g. a `fixtures.py` with factory functions that return the fixture objects. Each leaf `conftest.py` imports and re-registers them, parametrizing only the dtype for `input_csv`. The `wave_size` and `_check_skip` functions are the most important to consolidate.

vlaindic:
Hi @julijose-amd , do we still need it or we decided to discard the work?

PR #8607, opened 2026-07-15
feat(rocprofiler-systems): wire in profiler-hub/rocpdsna writer API

adjordje-amd:
## Motivation

Introduce the new profiler-hub/rocpdsna writer API as a replacement backend for trace persistence. This is PR1 of 2: it wires in the new backend additively while keeping the existing SQLite3/rocpd backend fully intact and buildable. PR2 (follow-up)
will remove the now-dead old backend once this lands.

JIRA ID : AIPROFSYST-52, AIPROFSYST-25

## Technical Details

Adds `cmake/ProfilerHub.cmake` (resolves profiler-hub via installed package, else sparse git checkout fallback) and links the new `rocprofiler-systems-profiler-hub` interface target into the core library alongside the existing `rocprofiler-systems-sqlite3` link. Rewrites `trace_cache/rocpd_processor` to use the new writer API and adds `rocpd_helpers.hpp`. No existing SQLite3/rocpd CMake wiring, options, or source files are removed in this PR.

Follow-up commit: dropped the in-tree monorepo fallback tier from ProfilerHub.cmake's resolution order. It silently picked up whatever local `profilers/profiler-hub` checkout a developer had lying around, which masked a CMake export bug (separate `fmt` export fix, tracked on its own branch) in local dev builds while it still failed on CI. Resolution is now strictly find_package, else sparse-checkout-and-build, so behavior is deterministic in every environment.

## Test Plan

Configured and built with both `ROCPROFSYS_BUILD_TESTING=ON` and the old `rocprofiler-sdk-rocpd`/SQLite3 path active alongside the new profiler-hub target, confirming both backends coexist and compile. Ran the new `rocpd_write_read_test`/`make_agent_uid_test`/`make_trace_env_test`/`make_event_test` suites via `rocprof-sys-unit-tests`.

## Test Result

Full build succeeded (706/706 targets). 32/32 new profiler-hub unit tests passed.

## Checklist

- [x] Old SQLite3/rocpd backend untouched and still builds
- [x] New profiler-hub backend builds and links
- [x] Unit tests added and passing
- [x] No unrelated changes


projects/rocprofiler-systems/cmake/ProfilerHub.cmake
```diff
@@ -0,0 +1,186 @@
...
+# ------------------------------------------------------------------------------
+
+set(ROCPROFSYS_PROFILER_HUB_GIT_REPOSITORY
+    "https://github.com/ROCm/rocm-systems.git"
+    CACHE STRING
+    "Git repository URL for profiler-hub fallback sparse checkout"
+)
+
+set(ROCPROFSYS_PROFILER_HUB_GIT_TAG
+    "users/adjordje-amd/profiler-hub-fix-fmt-export"
```
adjordje-amd: this is a temporarily, workaround. The final version should target `develop`

projects/rocprofiler-systems/source/lib/core/trace_cache/rocpd_helpers.hpp
```diff
@@ -0,0 +1,115 @@
...
+#include <string_view>
+#include <unordered_map>
+#include <utility>
+
+namespace rocprofsys
+{
+namespace trace_cache
+{
+namespace rocpd_helpers
+{
```
mradosav-amd:
```suggestion
namespace rocprofsys::trace_cache::rocpd_helpers
{


```

projects/rocprofiler-systems/source/lib/core/trace_cache/rocpd_helpers.hpp
```diff
@@ -0,0 +1,115 @@
...
+
```
mradosav-amd, line 19: maybe to move all of methods to class with static methods.

projects/rocprofiler-systems/source/lib/core/trace_cache/rocpd_processor.hpp
```diff
@@ -33,25 +36,21 @@ class rocpd_processor_t : public processor_t<rocpd_processor_t>
...
+    void post_process_metadata();
 
-    std::shared_ptr<metadata_registry>     m_metadata;
-    std::shared_ptr<agent_manager>         m_agent_manager;
-    std::shared_ptr<rocpd::data_processor> m_data_processor;
-    output_file_registry&                  m_output_registry;
-    std::string                            m_db_output_path;
+    std::shared_ptr<metadata_registry>      m_metadata;
+    std::shared_ptr<agent_manager>          m_agent_manager;
+    std::unique_ptr<profiler_hub::writer_t> m_writer;
```
mradosav-amd, line 51: can we make it not directly dependent on `writer_t`? so we can mock it, and have that kind of tests.

projects/rocprofiler-systems/source/lib/core/trace_cache/tests/test_rocpd_processor.cpp
```diff
@@ -0,0 +1,1791 @@
...
+    {
+        m_temp_dir = std::filesystem::temp_directory_path() /
+                     ("rocpd_test_" + std::to_string(::getpid()) + "_" +
+                      std::to_string(test_counter_++));
+        std::filesystem::create_directories(m_temp_dir);
+
+        m_db_path    = (m_temp_dir / "test.rocpd").string();
+        m_uuid       = "test123";
+        auto storage = std::make_unique<profiler_hub::storage_t>(m_db_path, m_uuid);
+        m_writer     = std::make_unique<profiler_hub::writer_t>(std::move(storage));
```
mradosav-amd, line 240: I think we don't want to create database file for unit test.

projects/rocprofiler-systems/source/lib/core/trace_cache/rocpd_helpers.hpp
adjordje-amd: Done

projects/rocprofiler-systems/source/lib/core/trace_cache/rocpd_processor.hpp
adjordje-amd: Will be done in a follow up PR

projects/rocprofiler-systems/source/lib/core/trace_cache/rocpd_helpers.hpp
adjordje-amd: as we already discussed offline, there are no difference between a **namespace + free functions** and a **class + static methods** approach

projects/rocprofiler-systems/source/lib/core/trace_cache/tests/test_rocpd_processor.cpp
adjordje-amd: Will be addressed in a follow up PR

projects/rocprofiler-systems/cmake/Packages.cmake
```diff
@@ -405,6 +405,8 @@ if(NOT ROCPROFSYS_USE_ROCPD_LIBRARY)
     rocprofsys_configure_rocpd_schema_files()
 endif()
 
+include(ProfilerHub)
```
dgaliffiAMD, line 414:
```suggestion

# ----------------------------------------------------------------------------------------#
#
# Profiler Hub
#
# ----------------------------------------------------------------------------------------#

include(ProfilerHub)
```

nit-pick: Adding section header.

projects/rocprofiler-systems/source/lib/core/trace_cache/rocpd_processor.cpp
```diff
@@ -874,126 +1045,141 @@ rocpd_processor_t::post_process_metadata()
...
+                  pmc_agent_uid.type_index, pmc_info.name);
+
+        profiler_hub::writer_types::pmc_info_t           pmc_info_data;
+        profiler_hub::writer_types::pmc_info_unique_id_t uid;
+        uid.name                = pmc_info.name;
+        uid.agent_id            = pmc_agent_uid;
+        pmc_info_data.unique_id = uid;
+        pmc_info_data.target_arch =
+            is_cpu_gpu_agent ? std::make_optional<std::string>(pmc_info.target_arch)
+                             : std::nullopt;
```
dgaliffiAMD:
```suggestion
    pmc_info_data.target_arch =
        is_cpu_gpu_agent
            ? std::optional<std::string_view>{ pmc_info.target_arch }
            : std::nullopt;
```

 **Use-after-free: dangling `string_view` on `target_arch`** :warning:

`pmc_info_data.target_arch` is `std::optional<std::string_view>` (a non-owning view — see the lifetime contract documented in `writer_types.hpp` lines 19-37). The RHS here builds a **temporary `std::string`** via `std::make_optional<std::string>(...)`, and the field stores a view into it. That temporary is destroyed at the end of this statement, so by the time `m_writer->register_pmc_info(pmc_info_data)` runs (line 1181) the view dangles.

Any CPU/GPU PMC descriptor with a non-empty `target_arch` (e.g. `"gfx942"`) will read freed heap — garbage written to the DB, or a crash under ASAN.

Bind the view directly to the alive source instead. `pmc_info` is a `const&` element of the list being iterated and outlives the `register_pmc_info` call, so its `target_arch` member is safe to view:

  ```cpp
  pmc_info_data.target_arch =
      is_cpu_gpu_agent
          ? std::optional<std::string_view>{ pmc_info.target_arch }
          : std::nullopt;
  ```

projects/rocprofiler-systems/source/lib/core/trace_cache/rocpd_processor.cpp
```diff
@@ -618,139 +685,207 @@ rocpd_processor_t::handle([[maybe_unused]] const cpu_pmc_sample& _cpu_pmc_sample
...
+
+            const auto core_load_samples = deserialize_loads(cpu_pmc_smpl.loads);
+            for(const auto& core : core_load_samples)
+            {
+                auto track_name = get_load_track_name(core.id);
+                insert_event_and_sample(trait::name<category::cpu_load>::value,
+                                        track_name.c_str(),
+                                        static_cast<double>(core.value));
+            }
+        }
```
dgaliffiAMD: Did you mean to pull the CPU frequency and load telemetry into the `if(is_process_owner)` block? This is a change in behaviour not mentioned in the description.

projects/rocprofiler-systems/source/lib/core/trace_cache/rocpd_processor.cpp
```diff
@@ -618,139 +685,207 @@ rocpd_processor_t::handle([[maybe_unused]] const cpu_pmc_sample& _cpu_pmc_sample
...
-    }
+            if(enabled_m.bits.kernel_time)
+            {
+                insert_event_and_sample(
+                    trait::name<category::process_kernel_mode_time>::value,
+                    trait::name<category::process_kernel_mode_time>::value,
+                    static_cast<double>(cpu_pmc_smpl.process_data.kernel_mode_time) /
+                        units::sec);
+            }
+        }
```
dgaliffiAMD, line 781:
```suggestion
        if(enabled_m.bits.user_time)
        {
            insert_event_and_sample(
                trait::name<category::process_user_mode_time>::value,
                trait::name<category::process_user_mode_time>::value,
                static_cast<double>(cpu_pmc_smpl.process_data.user_mode_time) /
                    units::sec);
        }

        if(enabled_m.bits.kernel_time)
        {
            insert_event_and_sample(
                trait::name<category::process_kernel_mode_time>::value,
                trait::name<category::process_kernel_mode_time>::value,
                static_cast<double>(cpu_pmc_smpl.process_data.kernel_mode_time) /
                    units::sec);
        }
```

This looked like a typo - `kernel_time` was nested in `user_time`. Please confirm if it was intended.

projects/rocprofiler-systems/source/lib/core/trace_cache/rocpd_processor.cpp
```diff
@@ -764,8 +899,11 @@ rocpd_processor_t::rocpd_processor_t(const std::shared_ptr<metadata_registry>& m
 , m_output_registry(output_registry)
 , m_db_output_path(generate_db_output_path(pid))
 {
-    m_data_processor = std::make_shared<rocpd::data_processor>(
-        std::make_shared<rocpd::data_storage::database>(pid, ppid, m_db_output_path));
+    auto n_info = node_info::get_instance();
+    auto uuid   = common::md5sum{ n_info.id, pid, ppid }.hexdigest();
+
+    auto storage = std::make_unique<profiler_hub::storage_t>(m_db_output_path, uuid);
+    m_writer     = std::make_unique<profiler_hub::writer_t>(std::move(storage));
```
dgaliffiAMD, line 902: Would you consider adding "writer_t" as an input parameter to facilitate dependency injection and full testing of the "handle()" functions. This would have caught the changes / regressions mentioned in the previous comments. 
This can be taken to a subsequent pull request.

projects/rocprofiler-systems/cmake/ProfilerHub.cmake
```diff
@@ -0,0 +1,206 @@
...
+# ------------------------------------------------------------------------------
+
+set(ROCPROFSYS_PROFILER_HUB_GIT_REPOSITORY
+    "https://github.com/ROCm/rocm-systems.git"
+    CACHE STRING
+    "Git repository URL for profiler-hub fallback sparse checkout"
+)
+
+set(ROCPROFSYS_PROFILER_HUB_GIT_TAG
+    "develop"
```
dgaliffiAMD: Pinning the sparse checkout to the HEAD of develop would break the "hermetic" build. Can we pin this "fall-back" path to a certain SHA or tag?

projects/rocprofiler-systems/cmake/ProfilerHub.cmake
```diff
@@ -0,0 +1,206 @@
...
+    if(EXISTS "${_PROFILER_HUB_STAMP}")
+        file(READ "${_PROFILER_HUB_STAMP}" _stamp_contents)
+        if(
+            _stamp_contents
+                STREQUAL
+                "${ROCPROFSYS_PROFILER_HUB_GIT_REPOSITORY}@${ROCPROFSYS_PROFILER_HUB_GIT_TAG}:${ROCPROFSYS_PROFILER_HUB_GIT_SUBDIR}"
+        )
+            set(_PROFILER_HUB_NEEDS_CHECKOUT FALSE)
+        endif()
+    endif()
```
dgaliffiAMD, line 73: **If you use a pinned commit, then you can ignore this observation**: 
It looks like this is checking the existence of a stamp, but not that the SHA is the latest. So, if `develop` advances, this won't fetch the latest update.

projects/rocprofiler-systems/source/lib/core/trace_cache/rocpd_processor.cpp
dgaliffiAMD: Sorry, I saw Milan's comment. Disregard.

projects/rocprofiler-systems/source/lib/core/trace_cache/rocpd_processor.cpp
```diff
@@ -50,310 +65,325 @@ get_handle_from_code_object(
...
+
+    m_writer->insert_region_data(region, env);
 }
 
 void
-rocpd_processor_t::handle(const in_time_sample& _its)
+rocpd_processor_t::handle(const in_time_sample& its)
 {
-    auto track_primary_key = m_data_processor->insert_string(_its.track_name.c_str());
+    auto event    = make_event(its.stack_id, its.parent_stack_id, its.correlation_id,
```
dgaliffiAMD, line 288: This might be out-of-scope for this PR, but there are a number of `its` members that are just dropped. It doesn't look like producers are ever populating them; does it make sense to remove them from the struct? 

In scope: maybe adding a comment stating that they were intentionally left out.

projects/rocprofiler-systems/cmake/ProfilerHub.cmake
adjordje-amd: I think that we should pin to a specific commit until we finish the TheRock integration. Changes in profiler-hub can introduce regressions in rocprof-sys because we don't have a defensive mechanism in place. In other words, the profiler-hub CI/CD pipeline doesn't validate the rocprof-sys integration. What do you think, should we use the latest develop or leave it as it is?

projects/rocprofiler-systems/source/lib/core/trace_cache/rocpd_processor.cpp
adjordje-amd: This issue and the one mentioned in this [comment ]( https://github.com/ROCm/rocm-systems/pull/8607#discussion_r3623853369)share the same root cause: a mismatched parenthesis. Addressed

dgaliffiAMD (approved):

dgaliffiAMD:
Also, please update CHANGELOGS before merging.

projects/rocprofiler-systems/source/lib/core/trace_cache/rocpd_processor.cpp
adjordje-amd: Checked this closely: `category_enum_id` is actually used, `perfetto_processor.cpp`'s `dispatch_in_time_sample()` reads it to pick the right annotation category, it's just not needed by this rocpd/profiler-hub handler since we don't do per-category dispatch here. `call_stack` and `line_info` are genuinely unused though `kokkosp.cpp` (the only producer) hardcodes them to "{}" and no consumer reads them back. Leaving the struct as-is for this PR to keep scope narrow; I'll note in a follow-up that `call_stack`/`line_info` are vestigial placeholders rather than removing them here.

PR #5267, opened 2026-04-22
[rocprofiler-sdk] [rocpd] Update rocprofiler-sdk-rocpd api to support versioning

yhuiYH:
## Motivation

Put in the ground work for supporting rocpd schema versions

## Technical Details

Extend the API to support:
- requesting particular rocpd schema versions
- retrieving list of supported schemas
- adding API test for C API as well as Python binding

On Rocprof-systems side:
- had to update rocprof-sys due to failed TheRock integration tests
- Make rocprof-sys check the rocprofiler-sdk-rocpd version, use new API if >= 1.3.2
- If < 1.3.2, fall back to use the legacy API
- Also, if newer rocprof-sys is installed onto older ROCm (ie, runtime < 1.3.2), then provided fall-back path to call load_schema() using older API.

## JIRA ID

AIPROFSDK-72 and ROCPDSNA-67

## Test Plan

Added tests that:
1) Use the API to get all supported schema versions
2) Loop over all schema version and use the API to request each of the schema version
3) Validate the Tables and Views are found

Can run it with: `ctest -R rocpd-api`

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #8610, opened 2026-07-15
fix(profiler-hub): resolve export-set, and schema-API build failures

### LINKED ISSUE #8609: profiler-hub: install(EXPORT) fails - vendored fmt leaks into static export set
## Description

`cmake --install` on `profilers/profiler-hub` fails with:

```
CMake Error: install(EXPORT "profiler-hub" ...) includes target "profiler-hub" which
requires target "fmt" that is not in any export set.
```

This only surfaces when `find_package(fmt 11.2.0)` does not find a matching system
package, so `cmake/fmt.cmake` falls back to vendoring fmt via `FetchContent` +
`add_library(fmt::fmt ALIAS fmt)`. It does not surface in the existing
`profiler-hub-install.yml` CI job because that job installs `libfmt-dev` on the
runner beforehand, so the system-fmt branch is always taken there.

## Root cause

`profiler-hub-static` (STATIC) is built from `$<TARGET_OBJECTS:profiler-hub-objects>`
and separately relinks its private dependencies (including fmt) a second time in
`profilers/profiler-hub/CMakeLists.txt`. CMake automatically copies a STATIC
library's `PRIVATE` link dependencies into its own `INTERFACE_LINK_LIBRARIES`
(a static archive cannot resolve its own undefined symbols, so consumers must supply
them at final link time). The existing `$<BUILD_INTERFACE:fmt::fmt>` guard does not
stop `install(EXPORT ...)`'s export-set validation from resolving and flagging the
underlying vendored `fmt` target, since that in-build-tree alias is never itself
installed/exported.

## Discovery context

Found while building rocprofiler-systems PR #8607, which is the first change to
drive an in-tree `add_subdirectory(profilers/profiler-hub)` build of profiler-hub
without a pre-installed `profiler-hub` package already sitting on the CMake package
search path (which previously masked this failure).

## Fix

Tracked in a follow-up PR scoped only to `profilers/profiler-hub/`.

adjordje-amd:
## Motivation

Started as a single fmt export-set bug (fixes #8609), found while building rocprofiler-systems PR #8607, the first change to drive an in-tree `add_subdirectory(profilers/profiler-hub)` build without a pre-installed profiler-hub package on the search path. Each subsequent verification pass (debian/rhel CI, TheRock's superbuild) exposed further build-portability bugs in profiler-hub's own CMake and one C++ API call site, all fixed here as they were found.

## Technical Details

- `install(EXPORT "profiler-hub" ...)` rejected `profiler-hub-static`'s vendored fmt: CMake copies a STATIC target's PRIVATE deps into its own `INTERFACE_LINK_LIBRARIES`; fixed by explicitly overriding that property post-link to keep only genuinely `find_package()`-imported deps.
- nlohmann_json's `find_package()` had no minimum version, silently accepting an ancient RHEL 8.10 EPEL package missing a header added upstream in a later release; then switched to always vendoring it via FetchContent instead of `find_package()` at all, since TheRock (ROCm's superbuild) rejects any `find_package()` call for a dependency not pre-declared for this subproject.
- fmt built SHARED instead of its usual static default: `rocprofiler-systems` sets `BUILD_SHARED_LIBS ON` project-wide and `add_subdirectory()`s profiler-hub, leaking that cache variable into profiler-hub's FetchContent'd fmt with no guard; fixed by forcing `BUILD_SHARED_LIBS OFF` around the fetch, matching the existing pattern already used for spdlog.
- `fmt`/`spdlog`/`nlohmann_json`'s install-interface logic was keyed off each package's `_FOUND` variable, which can flip true later in the same directory scope (e.g. TheRock redirects an unrelated, unversioned `find_package(fmt)` elsewhere to its own real package) without recreating the target if one already exists under that name - so a previously-vendored, non-`IMPORTED` target could get incorrectly exported. Fixed by checking the target's actual `IMPORTED` property instead.
- `rocpd_sql_load_schema` call site was written against an older API: rocprofiler-sdk-rocpd PR #5267 added a `schema_version` parameter, growing the signature from 8 to 9 arguments. This path only compiles when `rocprofiler-sdk-rocpd` is genuinely found, which only happens in TheRock's build. Fixed by passing the documented "use latest schema" sentinel, matching the pattern rocprofiler-systems itself already uses for the same API.

## Test Plan

Each fix verified against the real `dgaliffiamd/rocprofiler-systems:ci-rocm-6.4-debian-12` CI container (not host approximation) covering both the vendored-dependency and real-system-package paths, plus real GitHub Actions CI runs on PR #8607 (which pulls this branch) for the TheRock-specific and rocprofiler-sdk-rocpd-specific paths that can't be reproduced locally.

## Test Result

- Vendored-fmt and real-system-fmt configure/build/`cmake --install` both succeed cleanly; installed `profiler-hub-targets.cmake` has no stray entries.
- `ldd` on installed binaries resolves all dependencies with no "not found" entries.
- Unit suite passes in every verified context.
- TheRock's build now passes the CMake configure stage that previously failed on both the fmt export-set and nlohmann_json declaration issues.

## Checklist

- [x] Scoped only to `profilers/profiler-hub/`
- [x] Pre-commit hooks pass
- [x] Tests pass in all verified contexts
- [x] No unrelated changes


profilers/profiler-hub/CMakeLists.txt
```diff
@@ -270,6 +257,34 @@ foreach(_target profiler-hub profiler-hub-static)
...
+# STATIC libraries copy PRIVATE link deps into INTERFACE_LINK_LIBRARIES so
+# consumers can resolve undefined symbols, but install(EXPORT) then
+# validates every target referenced there, even inside $<BUILD_INTERFACE:>.
+# Rebuild the property with only the deps that are genuinely
+# find_package()-imported and thus resolvable by real consumers; queries
+# and sqlite3 are always in-tree-only and never belong in it.
+# profiler-hub (SHARED) needs no such fix: PRIVATE deps of a shared library
+# are never copied into INTERFACE_LINK_LIBRARIES.
+set(profiler_hub_static_iface_libs "")
+foreach(_dep fmt spdlog nlohmann_json)
```
adanalis-amd: The rebuilt static interface adds `$<INSTALL_INTERFACE:spdlog::spdlog>`, `nlohmann_json::nlohmann_json`, and `rocprofiler-sdk-rocpd` (when system-found), but `profiler-hub-config.cmake.in` only emits `find_dependency(fmt)`. A consumer that does `find_package(profiler-hub)` and links `profiler-hub::profiler-hub-static` fails at generate time ("target spdlog::spdlog not found") whenever those were system packages. The smoke test only links the shared target, so it misses this. Fix: emit matching `find_dependency(spdlog)` / `find_dependency(nlohmann_json)` (and rocpd) in the config template, gated on the same `*_FOUND` conditions. Note `nlohmann_json` is header-only (INTERFACE) and a PRIVATE dep not used in the installed public headers, so it carries no link symbols a static consumer needs - dropping it from this list is an alternative that both narrows the interface and shrinks the `find_dependency` surface.

profilers/profiler-hub/source/data_storage/backends/sqlite_backend_impl.hpp
```diff
@@ -18,6 +18,11 @@
     USE_SCHEMA_FROM_ROCPROFILER_SDK_ROCPD > 0
 #    include <rocprofiler-sdk-rocpd/sql.h>
 #    include <rocprofiler-sdk-rocpd/types.h>
+#    include <rocprofiler-sdk-rocpd/version.h>
+// rocpd_sql_load_schema() and its callback gained a schema_version parameter in
+// rocprofiler-sdk-rocpd 1.3.2 (upstream PR #5267). Older installs (e.g. ROCm <= 7.2,
+// version 1.3.1) still use the 8-argument signature without it.
+#    define PROFILER_HUB_ROCPD_SQL_HAS_SCHEMA_VERSION (ROCPD_VERSION >= 10302)
```
dgaliffiAMD, line 25:
```suggestion
#    define PROFILER_HUB_ROCPD_SQL_HAS_SCHEMA_VERSION (defined(ROCPD_VERSION) && (ROCPD_VERSION >= 10302))
```

nit-pick: Checking if `ROCPD_VERSION` is defined before checking it's value. This is non-blocking, slightly defensive, mostly stylistic.

dgaliffiAMD (approved):

PR #8611, opened 2026-07-15
Enable segmented cross-device graph launches

vstojilj:
## Motivation

Enable HIP graph launches where a graph is captured or instantiated on one GPU but launched from a stream on another GPU. The current segmented graph path rejected or mishandled this case after the classic graph execution path was removed, which blocked multidevice graph workflows such as hipGraphUpload with device context changes.

## Technical Details

This change implements cross-device launch support in GraphExecSegmented by replaying segmented AQL on an internal stream from the graph capture device, rather than falling back to per-node classic execution or running graph packets on the foreign launch stream.

For cross-device launches, the runtime does next:

- Builds streams_ from capture-device internal streams via UpdateStreams(nullptr).
- Creates the full internal capture-device stream pool during graph exec initialization.
- Uses marker pairs to preserve ordering between the user launch stream and the capture-device graph stream before replay, and to make graph completion visible back on the user launch stream after replay.
- Keeps same-device launch behavior unchanged by exposing one fewer internal stream when the user launch stream occupies stream slot 0.
- Cleans up segmented internal streams in ~GraphExecSegmented.
- Enables relevant graph tests and adds focused cross-device graph launch coverage.

## JIRA ID

AIRUNTIME-572

## Test Plan
 Unit_hipGraphUpload_Functional_multidevice_test
 Unit_hipGraphCrossDeviceLaunch_CrossStreamDependencyOrdering
 Unit_hipGraphCrossDeviceLaunch_ExplicitNodes
 Unit_hipGraphLaunch_Functional_multidevice_test
 Unit_hipGraphInstantiateWithFlags_DependencyGraphDeviceCtxtChg
 Unit_hipGraphExecUpdate_Negative_MultiDevice_Context_Changed
 Unit_hipDeviceGetGraphMemAttribute_Functional_Multi_Device
 Unit_hipGraphMem_Alloc_Free_NodeGetParams_Functional_MultiDevice
 Unit_hipGraphMultiDevice

## Test Result

Tests are passing on MI300.

## Submission Checklist

### DIFF
--- a/projects/clr/hipamd/src/hip_graph_internal.cpp
+++ b/projects/clr/hipamd/src/hip_graph_internal.cpp
@@ -1077,11 +1077,11 @@ hipError_t GraphExecBase::CreateStreams(uint32_t num_streams, int devId) {
     return hipSuccess;
   }
 
-  // num_streams is already capped by Init() but guard here defensively.
-  // For the instantiation device one slot is occupied by the launch stream,
-  // so create one fewer extra stream. Other devices use all slots as parallel streams.
-  uint32_t capped = std::min(num_streams, DEBUG_HIP_FORCE_GRAPH_QUEUES);
-  uint32_t max_streams = (devId == captureDeviceId_ && capped > 0) ? capped - 1 : capped;
+  // num_streams is already capped by Init() but guard here defensively. Create
+  // the full requested internal pool: cross-device launches do not have a
+  // launch stream on captureDeviceId_, so streams_[0] must be an internal
+  // capture-device stream.
+  uint32_t max_streams = std::min(num_streams, DEBUG_HIP_FORCE_GRAPH_QUEUES);
   if (max_streams == 0) {
     return hipSuccess;
   }
@@ -1107,14 +1107,21 @@ hipError_t GraphExecBase::CreateStreams(uint32_t num_streams, int devId) {
       return hipErrorOutOfMemory;
     }
 
-    // Pin the queue so dynamic queue management won't release it between launches
-    stream->vdev()->PinQueue();
-    // Acquire a queue that doesn't collide with previously created internal streams.
-    // On the first stream (used_qids empty) this is a normal acquisition.
-    if (!used_qids.empty()) {
-      stream->vdev()->ReacquireQueueExcluding(used_qids);
+    // Same-device launches supply stream slot 0 with the user launch stream, so
+    // keep the last capture-device stream as an unpinned reserve. It does not
+    // acquire a HW queue unless a cross-device launch needs the full internal
+    // pool. All other streams stay pinned for the graph lifetime.
+    const bool cross_device_reserve =
+        devId == captureDeviceId_ && (i + 1) == max_streams;
+    if (!cross_device_reserve) {
+      stream->vdev()->PinQueue();
+      // Acquire a queue that doesn't collide with previously created internal streams.
+      // On the first stream (used_qids empty) this is a normal acquisition.
+      if (!used_qids.empty()) {
+        stream->vdev()->ReacquireQueueExcluding(used_qids);
+      }
+      used_qids.insert(stream->getQueueID());
     }
-    used_qids.insert(stream->getQueueID());
 
     parallel_streams_[devId].push_back(stream);
   }
@@ -1221,9 +1228,9 @@ hipError_t GraphExecSegmented::FindStreamsReqPerDevForSegments() {
 // ================================================================================================
 void GraphExecSegmented::RoundRobinStreamAssignment() {
   // max_streams_dev_ holds the raw parallelism count per device as computed by
-  // FindStreamsReqPerDevForSegments() and capped in Init(). CreateStreams() handles
-  // the -1 adjustment for the instantiation device internally, so the value here
-  // represents the total stream pool size for every device uniformly.
+  // FindStreamsReqPerDevForSegments() and capped in Init(). It represents the
+  // total stream pool size uniformly; UpdateStreams() substitutes the user
+  // launch stream for the capture-device reserve on same-device launches.
   auto getPoolSize = [&](int dev_id) -> size_t {
     auto it = max_streams_dev_.find(dev_id);
     return (it != max_streams_dev_.end() && it->second > 0)
@@ -2800,20 +2807,34 @@ hipError_t GraphExecSegmented::EnqueueSegment(const Segment& segment, hip::Strea
 
 // ================================================================================================
 void GraphExecBase::UpdateStreams(hip::Stream* launch_stream) {
-  int devId = launch_stream->vdev()->device().index();
+  int devId = (launch_stream != nullptr) ? launch_stream->vdev()->device().index()
+                                         : captureDeviceId_;
   streams_.clear();
-  streams_.push_back(launch_stream);
+  if (launch_stream != nullptr) {
+    streams_.push_back(launch_stream);
+  }
   if (parallel_streams_.find(devId) == parallel_streams_.end()) {
-    // No parallel streams were created for this device
+    if (launch_stream == nullptr) {
+      LogPrintfError("UpdateStreams failed for capture device id:%d", devId);
+    }
     return;
   }
   auto& parallel_streams = parallel_streams_[devId];
 
-  // Collect queue IDs already in use, starting with the launch stream
+  // Collect queue IDs already in use, starting with the launch stream if present.
   std::unordered_set<uint64_t> used_qids;
-  used_qids.insert(launch_stream->getQueueID());
+  hip::Stream* skipped_stream = nullptr;
+  if (launch_stream != nullptr) {
+    used_qids.insert(launch_stream->getQueueID());
+    // Same-device launches use the user launch stream as stream slot 0, so
+    // leave the unpinned reserve unused.
+    skipped_stream = parallel_streams.back();
+  }
 
-  for (auto stream : parallel_streams) {
+  for (auto* stream : parallel_streams) {
+    if (stream == skipped_stream) {
+      continue;
+    }
     uint64_t qid = stream->getQueueID();
     if (used_qids.count(qid) > 0) {
       // Collision: this stream shares a HW queue with the launch stream or another
@@ -3091,9 +3112,24 @@ hipError_t GraphExecSegmented::Run(hip::Stream* launch_stream) {
   // try to re-acquire the same one it used last time.
   // Then run collision detection to ensure graph-internal streams don't share
   // a HW queue with the launch stream.
+  const bool cross_device_launch = (captureDeviceId_ != launch_stream->DeviceId());
   launch_stream->vdev()->SetPreferredQueue();
   launch_stream->vdev()->AcquireQueueWithPreference();
-  UpdateStreams(launch_stream);
+  if (cross_device_launch) {
+    // The last capture-device stream is kept unpinned and queue-less while the
+    // graph is used only on its capture device. Pin the full pool before a
+    // cross-device launch; UpdateStreams() will acquire a distinct queue for
+    // the reserve when it builds streams_.
+    auto it = parallel_streams_.find(captureDeviceId_);
+    if (it != parallel_streams_.end()) {
+      for (auto* stream : it->second) {
+        if (stream != nullptr) {
+          stream->vdev()->PinQueue();
+        }
+      }
+    }
+  }
+  UpdateStreams(cross_device_launch ? nullptr : launch_stream);
 
   // Signals borrowed from the per-graph pool for this launch (segmented path
   // only); handed to the completion callback to re-arm and return to the pool.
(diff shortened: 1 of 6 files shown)


projects/clr/hipamd/src/hip_graph_internal.hpp
```diff
@@ -1104,6 +1105,18 @@ class GraphExecSegmented : public GraphExecBase {
       signalManager_ = nullptr;
     }
 
+    for (auto& streams : parallel_streams_) {
```
jaydeeppatel1111: I think, This is already moved to GraphExecBase with https://github.com/ROCm/rocm-systems/pull/8254

vstojilj: I will rebase, thanks.

projects/clr/hipamd/src/hip_graph_internal.cpp
```diff
@@ -2698,20 +2698,54 @@ hipError_t GraphExecSegmented::EnqueueSegment(const Segment& segment, hip::Strea
 
 // ================================================================================================
 void GraphExecBase::UpdateStreams(hip::Stream* launch_stream) {
-  int devId = launch_stream->vdev()->device().index();
+  int devId = (launch_stream != nullptr) ? launch_stream->vdev()->device().index()
+                                         : captureDeviceId_;
   streams_.clear();
-  streams_.push_back(launch_stream);
+  if (launch_stream != nullptr) {
```
AlexeySachkov, line 2813: I've seen this a lot in the CLR codebase and I'm wondering why don't we simply write `if (!launch_stream)`? Implicit conversion of pointer to boolean seems to be a pretty widespread idiom in C++. Is it a part of some coding guideline that we are avoiding it?

projects/clr/hipamd/src/hip_graph_internal.cpp
```diff
@@ -3001,39 +3037,80 @@ hipError_t GraphExecSegmented::Run(hip::Stream* launch_stream) {
...
-      last_cmd = EnqueueSegmentedGraph(launch_stream, streams_, &status, &launch_signal_set);
-    } else {
-      // Multi-device: pass empty vector, will use parallel_streams_ internally
-      last_cmd = EnqueueSegmentedGraph(launch_stream, {}, &status, &launch_signal_set);
-    }
+  if (cross_device_launch && streams_.empty()) {
+    ClPrint(amd::LOG_ERROR, amd::LOG_CODE,
+            "[hipGraph] GraphExecSegmented::Run: no capture-device stream for device %d",
+            captureDeviceId_);
+    status = hipErrorInvalidDevice;
```
AlexeySachkov: Why don't we do an early exit here, returning the `hipErrorInvalidDevice` immediately, so that we don't need to guard everything below with `if (status == hipSuccess)`?

projects/clr/hipamd/src/hip_graph_internal.cpp
vstojilj: That's a preference for clr code.

projects/clr/hipamd/src/hip_graph_internal.cpp
vstojilj: Agreed, fixed.

projects/clr/hipamd/src/hip_graph_internal.cpp
```diff
@@ -1077,11 +1077,11 @@ hipError_t GraphExecBase::CreateStreams(uint32_t num_streams, int devId) {
...
-  // num_streams is already capped by Init() but guard here defensively.
-  // For the instantiation device one slot is occupied by the launch stream,
-  // so create one fewer extra stream. Other devices use all slots as parallel streams.
-  uint32_t capped = std::min(num_streams, DEBUG_HIP_FORCE_GRAPH_QUEUES);
-  uint32_t max_streams = (devId == captureDeviceId_ && capped > 0) ? capped - 1 : capped;
+  // num_streams is already capped by Init() but guard here defensively. Create
+  // the full requested internal pool: cross-device launches do not have a
+  // launch stream on captureDeviceId_, so streams_[0] must be an internal
+  // capture-device stream.
+  uint32_t max_streams = std::min(num_streams, DEBUG_HIP_FORCE_GRAPH_QUEUES);
```
iassiour, line 1084: I think this will create an extra stream even for a same-device launch that is pinned for the graph lifetime. Wondering if this is sth that can be avoided?

anugodavar: We dont know the launch stream/device during instantiate so this cannot be avoided to support multi device or cross device launch.

projects/clr/hipamd/src/hip_graph_internal.cpp
```diff
@@ -2800,20 +2800,54 @@ hipError_t GraphExecSegmented::EnqueueSegment(const Segment& segment, hip::Strea
...
   std::unordered_set<uint64_t> used_qids;
-  used_qids.insert(launch_stream->getQueueID());
+  if (launch_stream != nullptr) {
+    used_qids.insert(launch_stream->getQueueID());
+  }
+
+  // Same-device launches use the user launch stream as stream slot 0, so expose
+  // one fewer internal stream. Prefer dropping an internal stream that already
+  // shares the launch stream's queue; otherwise drop the last internal slot.
+  const bool same_device_launch = (launch_stream != nullptr && devId == captureDeviceId_);
```
anugodavar: can you move this under 2819 check? no need to determine again.

projects/clr/hipamd/src/hip_graph_internal.cpp
```diff
@@ -2800,20 +2800,54 @@ hipError_t GraphExecSegmented::EnqueueSegment(const Segment& segment, hip::Strea
...
+        break;
+      }
+    }
+    if (skipped_stream == nullptr) {
+      skipped_stream = parallel_streams.back();
+    }
+  }
 
-  for (auto stream : parallel_streams) {
+  size_t exposed_streams = 0;
```
anugodavar: exposed_streams and the break are redundant — the loop will naturally exhaust parallel_streams after skipping exactly one stream

projects/hip-tests/catch/unit/graph/hipGraphCrossDeviceLaunch.cc
```diff
@@ -0,0 +1,186 @@
+/*
+ * Copyright (c) Advanced Micro Devices, Inc., or its affiliates.
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+/**
+ * Tests for hipGraphLaunch when the launch stream's device differs from the
+ * device on which the graph was instantiated (cross-device launch). This
+ * exercises the Path B branch in GraphExec::Run, including dependency waits
```
jaydeeppatel1111: " Path B branch" - What does Path B for?

projects/clr/hipamd/src/hip_graph_internal.cpp
```diff
@@ -3103,39 +3139,77 @@ hipError_t GraphExecSegmented::Run(hip::Stream* launch_stream) {
...
-  if (captureDeviceId_ == launch_stream->DeviceId()) {
-    // If the graph has kernels that does device side allocation,  during packet capture, heap is
-    // allocated because heap pointer has to be added to the AQL packet, and initialized during
-    // graph launch.
-    // Todo: Hidden heap initialization is done only for single device graph
-    if (HasHiddenHeap() && hiddenHeapInitializedDevices_.insert(launch_stream->DeviceId()).second) {
-      launch_stream->vdev()->HiddenHeapInit();
-    }
-    amd::Command* last_cmd = nullptr;
+  if (cross_device_launch && streams_.empty()) {
```
anugodavar: streams_.empty() check in the cross_device_launch path is dead code — stream creation failures are caught during hipGraphInstantiate and return an error there, so Run() is never reached with an empty streams_. Removed the check.

projects/clr/hipamd/src/hip_graph_internal.cpp
```diff
@@ -3103,39 +3139,77 @@ hipError_t GraphExecSegmented::Run(hip::Stream* launch_stream) {
...
+  } else {
+    // Cross-device launch: replay segmented AQL on a capture-device stream.
+    // The marker pair before replay preserves prior launch-stream ordering;
+    // the marker pair after replay makes completion visible on the user's
+    // foreign launch stream without falling back to per-node execution.
+    hip::Stream* graph_stream = streams_[0];
+    auto* launch_start = new amd::Marker(*launch_stream, kMarkerDisableFlush, {});
+    launch_start->enqueue();
+
+    amd::Command::EventWaitList launch_wait_list;
```
anugodavar: Avoid enqueuing an unnecessary marker on the launch stream for cross-device synchronization. Instead, use getLastQueuedCommand(kRetainCommand) to get the last queued command on the launch stream and only enqueue a marker on the graph stream if that command is non-null. If nothing has been queued on the launch stream yet, no synchronization is needed and the marker is skipped entirely.

projects/clr/hipamd/src/hip_graph_internal.cpp
```diff
@@ -3103,39 +3139,77 @@ hipError_t GraphExecSegmented::Run(hip::Stream* launch_stream) {
...
+    launch_wait_list.push_back(launch_start);
+    auto* graph_start = new amd::Marker(*graph_stream, true, launch_wait_list);
+    graph_start->enqueue();
+    graph_start->release();
+    launch_start->release();
+
+    last_cmd = EnqueueSegmentedGraph(graph_stream, streams_, &status, &launch_signal_set);
+    if (status == hipSuccess && last_cmd != nullptr) {
+      auto* graph_done = new amd::Marker(*graph_stream, kMarkerDisableFlush, {});
+      graph_done->enqueue();
```
anugodavar: last_cmd is already the last command on graph_stream, so graph_done marker is redundant. Use last_cmd directly in the completion_wait_list for launch_done, eliminating an unnecessary marker enqueue on the graph stream.

projects/clr/hipamd/src/hip_graph_internal.cpp
anugodavar: For cross-device launches, the callback command is incorrectly placed on launch_stream (the user-visible stream on a different device). It should be on graph_stream (the capture-device stream) so the completion callback fires only after the graph work on the capture device has actually finished.

anugodavar (changes requested):

projects/clr/hipamd/src/hip_graph_internal.cpp
vstojilj: I changed it so that we still create the full stream-object pool at instantiation, but the last capture-device stream remains unpinned and without a HW queue for same-device launches. Same-device launches skip this reserve and cross-device launch pins it and acquires a queue only when needed.

projects/clr/hipamd/src/hip_graph_internal.cpp
vstojilj: Done

projects/clr/hipamd/src/hip_graph_internal.cpp
vstojilj: Done

projects/hip-tests/catch/unit/graph/hipGraphCrossDeviceLaunch.cc
vstojilj: Updated, left over from previous version

projects/clr/hipamd/src/hip_graph_internal.cpp
vstojilj: Removed the check

projects/clr/hipamd/src/hip_graph_internal.cpp
vstojilj: Done. The code now retrieves the foreign launch stream’s last queued command and only creates a graph-stream marker when prior work exists. No marker is added when the launch stream is empty

projects/clr/hipamd/src/hip_graph_internal.cpp
vstojilj: Done. launch_done now waits directly on the graph’s last_cmd.

projects/clr/hipamd/src/hip_graph_internal.cpp
vstojilj: Done. last_cmd remains the capture-device completion command that drives OnLaunchComplete. launch_done is used only to expose completion to the foreign launch stream. Fallback callback markers and callback-failure draining now also use graph_launch_stream.

anugodavar (approved):

saleelk (approved):

projects/clr/hipamd/src/hip_graph_internal.cpp
```diff
@@ -3103,39 +3139,67 @@ hipError_t GraphExecSegmented::Run(hip::Stream* launch_stream) {
...
+    if (status == hipSuccess && last_cmd != nullptr) {
+      amd::Command::EventWaitList completion_wait_list;
+      completion_wait_list.push_back(last_cmd);
+
+      auto* launch_done = new amd::Marker(*launch_stream, kMarkerDisableFlush, completion_wait_list);
+      launch_done->enqueue();
+      // The launch stream queue owns this marker now. Keep last_cmd as the
+      // capture-device completion command that drives graph resource cleanup.
+      launch_done->release();
+    } else if (status != hipSuccess) {
```
jungx098, line 3187: Only for Cross-device launch? We don't need this for single device?

vstojilj: Single device launches enqueue accumulate command directly on the user launch stream, so queue ordering is sufficient, we don't need additional marker.

PR #8615, opened 2026-07-15
refactor(rocprof-sys): Adapt state to be unit testable

mradosav-amd:
## Motivation

The process/thread state machine in `core/state.hpp` was implemented as free functions, a module-global `std::atomic`, and a `ROCPROFSYS_SCOPED_THREAD_STATE(...)` macro that mutated hidden global state directly, so none of the state-transition rules (monotonic process-state advancement, thread-state push/pop stacking, no-op behavior once a thread is `Completed`) could be exercised in isolation without pulling in the rest of the runtime. This PR restructures the state module so its behavior can be unit tested directly, and adds the first unit tests for it.

## Technical Details

`core/state.hpp` is rewritten from free functions/macros into two static-method classes, `rocprofsys::state::process` and `rocprofsys::state::thread`; `core/state.cpp` is deleted and its logic (atomic storage for process state, thread-local storage plus a per-thread history vector for thread state) now lives inline in the header. `process::set` validates monotonic transitions and throws `std::runtime_error` on a regression, while `process::reset` bypasses validation for re-attach scenarios. The `ROCPROFSYS_SCOPED_THREAD_STATE(...)` macro is replaced by a `[[nodiscard]]` `thread::scoped(...)` factory returning an RAII `scoped_guard`. `buffer_storage` (`core/trace_cache/buffer_storage.hpp`) now takes an injectable `ThreadStatePolicy` template parameter, constrained by a new `type_traits::thread_state_policy` concept, instead of calling the macro directly, so tests can substitute a policy that observes/controls thread-state transitions during `store()`/flush. The enums were renamed for clarity — `State` to `process_lifecycle`, `ThreadState` to `thread_lifecycle`, `Mode` to `mode::process`, `CausalBackend` to `backend::causal`, `CausalMode` to `mode::process_causal` — with call sites updated across roughly 40 files (`library.cpp`, the causal/, components/, pmc/, rocprofiler-sdk/ subtrees, etc.). New tests: `core/tests/test_state.cpp` covers `process` (default state, `set` advancing and returning the previous value, monotonic-violation throwing, `reset` bypassing validation) and `thread` (default state, `set` return value, push/pop round-trip, pop-on-empty no-op, push/pop no-op once `Completed`); `trace_cache/tests/test_buffer_storage.cpp` covers thread-state transitions during buffer storage operations. Both are wired into `core/tests/CMakeLists.txt` and `tests/CMakeLists.txt` as new `state-tests` / updated `buffer-storage-tests` objects. Per the CHANGELOG entry, there is no user-visible behavior change.

## JIRA ID
JIRA ID : AIPROFSYST-574

## Test Plan

- [x] `core/tests/test_state.cpp`: process state default/set/monotonic-violation-throws/reset, thread state default/set/push-pop/pop-on-empty/no-op-once-completed
- [x] `trace_cache/tests/test_buffer_storage.cpp`: thread-state transitions during buffer store/flush via the injected `ThreadStatePolicy`
- [x] Full project build with `ROCPROFSYS_BUILD_TESTING` to confirm the renamed enums and refactored state classes compile across all updated call sites

## Test Result

Pending CI; new unit tests above are expected to pass, and the existing suite is unaffected per the CHANGELOG's no-behavior-change note.

## Submission Checklist


projects/rocprofiler-systems/source/lib/core/state.hpp
```diff
@@ -20,87 +27,277 @@ enum class State : unsigned short
...
-#define ROCPROFSYS_SCOPED_THREAD_STATE(STATE)                                            \
-    ::rocprofsys::scoped_thread_state ROCPROFSYS_VARIABLE(_scoped_thread_state_,         \
-                                                          __LINE__)                      \
-    {                                                                                    \
-        ::rocprofsys::STATE                                                              \
+    [[gnu::cold]] static State reset()
+    {
+        auto last_state = get();
+        storage().store(PreInit, std::memory_order_relaxed);
+        LOG_DEBUG("Resetting state :: {} -> PreInit", get());
```
marantic-amd:
```suggestion
        LOG_DEBUG("Resetting state :: {} -> PreInit", last_state );
```

projects/rocprofiler-systems/source/lib/core/trace_cache/buffer_storage.hpp
```diff
@@ -81,7 +82,17 @@ struct flush_worker_factory_t
     }
 };
 
-template <typename WorkerFactory, typename TypeIdentifierEnum>
+namespace type_traits
+{
+template <typename T>
+concept thread_state_policy = requires(state::thread::State state_to_set) {
+    { T::scoped(state_to_set) };
```
marantic-amd:
```suggestion
    { T::scoped(state_to_set) } -> std::destructible;
```
Maybe we can add this here to enforce a policy with a stronger constraint?

mradosav-amd: fixed.

projects/rocprofiler-systems/source/lib/core/state.hpp
mradosav-amd: fixed.

marantic-amd (approved):

projects/rocprofiler-systems/source/lib/core/trace_cache/buffer_storage.hpp
```diff
@@ -150,7 +161,7 @@ class buffer_storage
         // for position management; extending the critical section to cover the
         // actual memcpy closes the window that TSan (correctly) flags.
         //
-        ROCPROFSYS_SCOPED_THREAD_STATE(ThreadState::Internal);
+        auto _thread_state_guard = ThreadStatePolicy::scoped(ThreadStatePolicy::Internal);
```
adjordje-amd: I think `_` is against the existing clang-tidy rules

projects/rocprofiler-systems/source/lib/core/tests/CMakeLists.txt
```diff
@@ -69,3 +69,20 @@ target_include_directories(
...
+add_library(state-tests OBJECT test_state.cpp)
+
+target_link_libraries(
+    state-tests
+    PRIVATE
+        rocprofiler-systems-googletest-library
+        rocprofiler-systems-core-library
+        rocprofiler-systems-logger
+        rocprofiler-systems-headers
+        rocprofiler-systems-timemory
```
adjordje-amd, line 82: why are we linking against timemory here?

projects/rocprofiler-systems/source/lib/core/trace_cache/buffer_storage.hpp
mradosav-amd: Done

projects/rocprofiler-systems/source/lib/core/tests/CMakeLists.txt
mradosav-amd: We have to because it's using some stuff from timemory. When timemory is removed, we will get rid of it. It's out of scope for this PR.

adjordje-amd (approved):

PR #8619, opened 2026-07-15
test(kfdtest): add AMD Infinity Storage (AIS) test

amd-willpalacek:
Add a kfdtest component test for AMD Infinity Storage (AIS), which transfers data directly between a PCIe-backed file (e.g. NVMe) and GPU VRAM via `hsaKmtAisReadWriteFile`, bypassing system-memory bounce buffers.

## Motivation

AIS enables direct storage-to-VRAM transfers without staging through system-memory bounce buffers. Prior to this change there was no kfdtest coverage for the `hsaKmtAisReadWriteFile` thunk path, so functional regressions in AIS read/write, the address-range guard, or concurrent access could land undetected.

## Technical Details

The suite has five tests, all gated on `hsakmt_is_dgpu()` since AIS is only supported on dGPUs:

- **AISGracefulFailureTest**: exercises the API on /tmp and checks it either succeeds (NVMe-backed) or fails without crashing (non-NVMe).
- **AISUnregisteredMemoryTest**: deterministic negative case that passes an unregistered host pointer and expects the thunk address-range guard in `ais.c` to reject it without crashing. Runs on any dGPU.
- **AISReadTest / AISWriteTest**: read a file into VRAM and write VRAM back to a file, validating the full buffer contents with `memcmp`.
- **AISMultiThreadedTest**: threads each transfer a distinct file section to a separate VRAM buffer concurrently, then write them back. Thread count is configurable via `--ais_threads` (default 8) and the total workload via `--ais_size_mb` (default 1024), clamped to free disk and 1/4 of VRAM so it never overcommits on any machine.

Host buffers used with O_DIRECT `read()`/`write()` are page-aligned via `posix_memalign` (a `std::vector` is only ~16-byte aligned and fails EINVAL on NVMe). `createTestFile` falls back to buffered I/O when O_DIRECT is rejected (e.g. tmpfs /tmp). Temp files are removed in TearDown.

## JIRA ID

JIRA ID : AILIKFD-831

## Test Plan

Build libhsakmt + kfdtest and run the AIS suite on a dGPU with an NVMe-backed test directory:

```
./kfdtest --gtest_filter='*AIS*'
./kfdtest --ais_threads 4 --ais_size_mb 256 --gtest_filter='*AISMultiThreadedTest*'
```

## Test Result

Verified on MI308X (gfx942), /tmp on ext4/NVMe, ROCm 7.13.0: all 5 AIS tests pass 3 consecutive runs.

```
[  PASSED  ] KFDAISTest.AISGracefulFailureTest
[  PASSED  ] KFDAISTest.AISUnregisteredMemoryTest
[  PASSED  ] KFDAISTest.AISReadTest
[  PASSED  ] KFDAISTest.AISWriteTest
[  PASSED  ] KFDAISTest.AISMultiThreadedTest
[  PASSED  ] 5 tests.
```

## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDAISTest.cpp
+++ b/projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDAISTest.cpp
@@ -1,0 +1,805 @@
+// Copyright © Advanced Micro Devices, Inc., or its affiliates.
+// SPDX-License-Identifier: MIT
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <sys/sysmacros.h>
+#include <sys/statvfs.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <cstdlib>
+#include <cstring>
+#include <cerrno>
+#include <climits>
+#include <algorithm>
+#include <memory>
+
+#include "KFDAISTest.hpp"
+
+/*
+ * O_DIRECT read()/write() require a page-aligned buffer; std::vector is only
+ * ~16-byte aligned and fails EINVAL on NVMe. Return a self-freeing buffer
+ * aligned to MT_ALIGN, or nullptr on failure.
+ */
+static std::unique_ptr<void, decltype(&free)> allocAligned(size_t size) {
+    void *p = nullptr;
+    if (posix_memalign(&p, KFDAISTest::MT_ALIGN, size) != 0)
+        p = nullptr;
+    return std::unique_ptr<void, decltype(&free)>(p, free);
+}
+
+void KFDAISTest::SetUp() {
+    ROUTINE_START
+
+    KFDBaseComponentTest::SetUp();
+
+    m_gpuNode = m_NodeInfo.HsaDefaultGPUNode();
+    m_fd = -1;
+    m_fdOut = -1;
+    m_isOnNVME = false;
+    m_bufSize = BUFFER_SIZE;
+    m_pBufs.clear();
+    m_patterns.clear();
+
+    ROUTINE_END
+}
+
+void KFDAISTest::TearDown() {
+    ROUTINE_START
+
+    deleteTestFiles();
+    freeVRAMBuffers(m_pBufs, m_bufSize);
+    m_pBufs.clear();
+    m_patterns.clear();
+
+    KFDBaseComponentTest::TearDown();
+
+    ROUTINE_END
+}
+
+bool KFDAISTest::isAISSupported(int node) {
+    const HsaNodeProperties *props = m_NodeInfo.GetNodeProperties(node);
+
+    if (props == NULL)
+        return false;
+
+    /*
+     * The KFD currently advertises "AIS is initialized for this device" by
+     * temporarily reusing the VA_LIMIT capability bit (see the comment on
+     * HSA_CAP_VA_LIMIT in kfd_topology.c). Keep that knowledge in this one
+     * helper so there is a single edit site once a dedicated capability bit
+     * is allocated.
+     *
+     * TODO (AILIKFD-831): move to a dedicated HSA_CAPABILITY2 bit; the
+     * VA_LIMIT reuse is temporary.
+     */
+    return props->Capability.ui32.VALimit;
+}
+
+int KFDAISTest::checkIfFileIsOnNVME(int fd) {
+    struct stat st;
+    char sysPath[PATH_MAX];
+    char linkTarget[PATH_MAX];
+
+    if (fstat(fd, &st) == -1) {
+        LOG() << "Failed to stat file descriptor " << fd << std::endl;
+        return -1;
+    }
+
+    snprintf(sysPath, sizeof(sysPath), "/sys/dev/block/%u:%u",
+             major(st.st_dev), minor(st.st_dev));
+
+    ssize_t len = readlink(sysPath, linkTarget, sizeof(linkTarget) - 1);
+    if (len == -1) {
+        LOG() << "File not backed by block device (may be tmpfs)" << std::endl;
+        return -1;
+    }
+
+    linkTarget[len] = '\0';
+
+    if (strstr(linkTarget, "nvme") == NULL) {
+        LOG() << "File is not on NVME device (path: " << linkTarget << ")" << std::endl;
+        return -1;
+    }
+
+    LOG() << "File is on NVME device (path: " << linkTarget << ")" << std::endl;
+    return 0;
+}
+
+bool KFDAISTest::findNVMEPath(std::string &outPath) {
+    const char* candidates[] = {
+        "/mnt/nvme/",
+        "/data/nvme/",
+        "/scratch/nvme/",
+        "/nvme/",
+        "/mnt/scratch/",
+        "/data/",
+        "/scratch/",
+        /*
+         * Try /tmp/ itself before giving up: on many single-drive dev/test
+         * boxes /tmp is just a directory on the NVMe-backed root fs, not a
+         * separate mount or tmpfs. Checking it here (instead of only as the
+         * unconditional non-NVME fallback below) lets checkIfFileIsOnNVME()
+         * actually detect that case. Confirmed needed on real MI300X test
+         * hardware (AILIKFD-831, 2026-07-27): root fs was nvme0n1p2 and
+         * every candidate above this line was absent.
+         */
+        "/tmp/"
+    };
+
+    for (const char* path : candidates) {
+        struct stat st;
+        if (stat(path, &st) == 0 && S_ISDIR(st.st_mode)) {
+            std::string tmpl = std::string(path) + "kfdtest_ais_probe.XXXXXX";
+            std::vector<char> tmp(tmpl.begin(), tmpl.end());
+            tmp.push_back(0);
+            int fd = mkstemp(tmp.data());
+            std::string testPath(tmp.data());
+            if (fd >= 0) {
+                bool isNVME = (checkIfFileIsOnNVME(fd) == 0);
+                close(fd);
+                unlink(testPath.c_str());
+                if (isNVME) {
+                    outPath = path;
+                    LOG() << "Found NVME path: " << path << std::endl;
+                    return true;
+                }
+            }
+        }
+    }
+
+    outPath = "/tmp/";
+    LOG() << "No NVME path found, using /tmp/ (AIS tests will detect non-NVME)" << std::endl;
+    return false;
+}
+
+int KFDAISTest::createTestFile(const std::string &dir, const char *templateName,
+                               size_t size, std::string &outPath) {
+    std::string tmpl = dir + templateName;
+    std::vector<char> nameBuf(tmpl.begin(), tmpl.end());
+    nameBuf.push_back('\0');
+
+    /*
+     * mkstemp() gives a unique, 0600, caller-owned file. That removes the
+     * truncate/symlink-clobber risk of a fixed name in a world-writable
+     * directory and lets concurrent kfdtest instances coexist.
+     */
+    int fd = mkstemp(nameBuf.data());
+    if (fd < 0) {
+        LOG() << "Failed to create temp file in " << dir
+              << " (" << strerror(errno) << ")" << std::endl;
+        return -1;
+    }
+    outPath = nameBuf.data();
+
+    /*
+     * mkstemp() cannot request O_DIRECT, so reopen the file we just created.
+     * Some filesystems (e.g. tmpfs-backed /dev/shm) reject O_DIRECT with
+     * EINVAL; fall back to the buffered descriptor so tests that only need a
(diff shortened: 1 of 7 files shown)


projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDAISTest.cpp
```diff
@@ -0,0 +1,690 @@
...
+            if (fd >= 0) {
+                bool isNVME = (checkIfFileIsOnNVME(fd) == 0);
+                close(fd);
+                unlink(testPath.c_str());
+                if (isNVME) {
+                    outPath = path;
+                    LOG() << "Found NVME path: " << path << std::endl;
+                    return true;
+                }
+            }
```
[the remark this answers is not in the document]
amd-willpalacek: Fixed in 832330d5. findNVMEPath() now uses mkstemp() with a unique kfdtest_ais_probe.XXXXXX template instead of a fixed-name file opened with O_CREAT | O_TRUNC. This eliminates both the truncation of any pre-existing file and the symlink/clobber risk in world-writable candidate directories. The probe file is still unlinked immediately after the NVMe check, and mkstemp creates it mode 0600, matching the previous S_IRUSR | S_IWUSR.

amd-willpalacek:
> 
> 
> * ❌ Unit Test
> 

False positive, confirmed with @kentrussell . KFDAISTest.cpp is the test file (follows the suite's KFD<Feature>Test.cpp convention, registered in CMakeLists.txt). No extra **Unit Test** required.

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDAISTest.cpp
```diff
@@ -0,0 +1,690 @@
...
+        ret = hsaKmtAllocMemory(gpuNode, bufferSize, flags, &bufs[i]);
+        if (ret != HSAKMT_STATUS_SUCCESS) {
+            LOG() << "Failed to allocate VRAM buffer " << i << std::endl;
+            freeVRAMBuffers(bufs, bufferSize);
+            return -1;
+        }
+
+        if (hsakmt_is_dgpu()) {
+            ret = hsaKmtMapMemoryToGPUNodes(bufs[i], bufferSize, NULL,
+                       mapFlags, 1, reinterpret_cast<HSAuint32*>(&gpuNode));
```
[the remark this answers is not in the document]
amd-willpalacek: Fixed in 832330d5. allocVRAMBuffers() now copies the node id into a properly typed HSAuint32 node = static_cast<HSAuint32>(gpuNode); and passes &node to hsaKmtMapMemoryToGPUNodes(), removing the int<->uint32_t type-punning and signedness assumption.

amd-willpalacek:
`allocVRAMBuffers()` uses raw hsaKmtAllocMemory hsaKmtMapMemoryToGPUNodes instead of the HsaMemoryBuffer helper because AIS requires host-accessible VRAM (HostAccess=1, NonPaged=1), the buffers are validated on the CPU with memcmp in `checkBuffersWithPatterns()`. HsaMemoryBuffer's local-memory path forces HostAccess=0, so it can't produce that flag combination.

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDAISTest.cpp
```diff
@@ -0,0 +1,694 @@
+/*
+ * Copyright (C) 2026 Advanced Micro Devices, Inc. All Rights Reserved.
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** New files should carry the SPDX short-form license header rather than the legacy full MIT text block:
```
//
```
Same applies to `KFDAISTest.hpp`. Non-blocking.

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDAISTest.cpp
```diff
@@ -0,0 +1,694 @@
...
+    TEST_START(TESTPROFILE_RUNALL);
+
+    if (!hsakmt_is_dgpu()) {
+        LOG() << "Skipping test: AIS is only supported on dGPU" << std::endl;
+        return;
+    }
+
+    ASSERT_GE(m_gpuNode, 0) << "Failed to get default GPU node";
+
+    m_testDir = "/tmp/";
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** `AISGracefulFailureTest` hardcodes `/tmp/` and then branches on `checkIfFileIsOnNVME`. On machines with an NVMe root filesystem, `/tmp` is itself NVMe-backed, so `m_isOnNVME` is true and the test silently asserts *success* — the intended "fail gracefully on non-NVMe" path is never exercised. Consider forcing a guaranteed non-NVMe location (e.g. `/dev/shm`/tmpfs) for this test, or at least LOG clearly when the failure branch was skipped so CI logs show the negative case wasn't covered.

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDAISTest.cpp
```diff
@@ -0,0 +1,694 @@
...
+    if (!m_testFilePathOut.empty()) {
+        unlink(m_testFilePathOut.c_str());
+        m_testFilePathOut.clear();
+    }
+}
+
+int KFDAISTest::fillFileWithPatterns(int fd, size_t fileSize, size_t chunkSize,
+                                     std::vector<uint32_t> &patterns) {
+    const uint32_t START_PATTERN = 0x11111111;
+    int numBuffers = fileSize / chunkSize;
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `int numBuffers = fileSize / chunkSize;` narrows `size_t → int`. Harmless at current sizes, but the multi-threaded test drives `fileSize` up to VRAM/4 (multi-GB); using `size_t` here would avoid a latent narrowing as buffer sizes grow.

cfreeamd (commented):

Reviewed the rocr-runtime kfdtest changes. The `--ais_threads`/`--ais_size_mb` CLI plumbing (KFDTestFlags.hpp, KFDTestMain.cpp, LinuxOSWrapper.cpp, OSWrapper.hpp) looks correct — the new `getopt_long` `switch(option_index)` cases 9/10 line up with their positions in `long_options[]`. The multi-threaded read/write path is safe: `hsaKmtAisReadWriteFile` uses positioned I/O (`file_offset` into the ioctl), so concurrent same-fd access at distinct offsets does not race. Copilot's two earlier catches (fixed-name probe file, int→HSAuint32 type-punning) are already fixed in this head. Only low-severity inline notes below.

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDAISTest.cpp
amd-willpalacek: Done in f7162d0. Both KFDAISTest.cpp and KFDAISTest.hpp now use the SPDX short-form header matching KFDQMTest.cpp:
// SPDX-License-Identifier:  MIT

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDAISTest.cpp
amd-willpalacek: Good catch - on a box where /tmp is NVMe-backed, m_isOnNVME came back true and the test silently took the success branch, so the graceful-failure path was never exercised. Fixed in f7162d0: AISGracefulFailureTest now uses /dev/shm (guaranteed tmpfs -> non-block-device), so checkIfFileIsOnNVME always returns non-NVMe and the failure branch always runs. createTestFile already falls back to buffered I/O when O_DIRECT is rejected on tmpfs, so file creation still succeeds. Verified on MI308X (gfx942, /tmp on ext4/NVMe): the test now logs File is NOT on NVME - AIS should fail gracefully + SUCCESS: AIS failed gracefully without crashing (thunk returns status=0xffffffed), passing 3 consecutive runs.

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDAISTest.cpp
amd-willpalacek: Fixed in f7162d0: numBuffers is now size_t and the fill loop uses a size_t index (with an explicit static_cast<uint32_t> on the pattern arithmetic), so there's no narrowing as the multi-threaded fileSize grows toward VRAM/4. Builds warning-clean.

cfreeamd (approved):

kentrussell:
Will this work on all ASICs from Hawaii up to GFX12? If there are GFX restrictions (FAMILY_AI etc), we should have that in the tests too.

amd-willpalacek:
> Will this work on all ASICs from Hawaii up to GFX12? If there are GFX restrictions (FAMILY_AI etc), we should have that in the tests too.

I'll sort that out.

amd-willpalacek:
> Will this work on all ASICs from Hawaii up to GFX12? If there are GFX restrictions (FAMILY_AI etc), we should have that in the tests too.

@kentrussell 

AIS isn't family-gated, the kernel decides per device in
  kfd_ais_init() (needs CONFIG_PCI_P2PDMA, large BAR, not
  SRIOV/APU/XGMI-to-CPU). So the tests query the capability the
  kernel already exports, plus an -ENODEV check on the first
  transfer. Anything unsupported skips rather than fails, so
  Hawaii→GFX12 is safe in CI either way.

  Verified on gfx942 (MI300X, ROCm 7.3) and gfx950 (MI350X, ROCm
  10.1) - 5/5 with real NVMe↔VRAM transfers.

  Happy to layer explicit FAMILY_* checks on top if you'd prefer; I
  just didn't want two sources of truth for a policy the kernel
  already owns.

amd-willpalacek:
@kentrussell Can you please review?

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDAISTest.cpp
```diff
@@ -0,0 +1,805 @@
...
+}
+
+bool KFDAISTest::isAISSupported(int node) {
+    const HsaNodeProperties *props = m_NodeInfo.GetNodeProperties(node);
+
+    if (props == NULL)
+        return false;
+
+    /*
+     * The KFD currently advertises "AIS is initialized for this device" by
```
kentrussell, line 67: Good call

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDAISTest.cpp
```diff
@@ -0,0 +1,805 @@
...
+/*
+ * Test 1: AISGracefulFailureTest
+ *
+ * Exercises the AIS path against a tmpfs-backed file (/dev/shm). tmpfs has
+ * no backing block or network device, so kfd_ais_rw_file()'s storage-type
+ * dispatch falls through to the AIS_STORAGE_UNKNOWN case ("AIS: unsupported
+ * storage type") and returns -ENODEV -- the same errno userspace sees when
+ * AIS isn't initialized at all for the device (kfd_chardev.c), so this is
+ * reported via the same AIS_UNSUPPORTED/SKIP path as isAISSupported()==false.
+ * Confirmed against the shipped amd-staging-dkms-7.1.1 kfd_ais.c on real
```
kentrussell, line 400: Not sure if we need to mention amd-staging-dkms-7.1.1 since that's not publicly available. Neither is AILIKFD-831. Not a blocker, but something to change if we have to make any other changes

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDAISTest.cpp
```diff
@@ -0,0 +1,805 @@
...
+
+    ASSERT_EQ(fillFileWithPatterns(m_fd, FILE_SIZE_BASIC, BUFFER_SIZE, m_patterns), 0);
+
+    m_bufSize = BUFFER_SIZE;
+    ASSERT_EQ(allocVRAMBuffers(m_gpuNode, 1, BUFFER_SIZE, m_pBufs), 0);
+
+    int result = aisReadWrite(m_pBufs[0], BUFFER_SIZE, m_fd, 0, HSA_AIS_READ, NULL);
+
+    if (result == AIS_UNSUPPORTED) {
+        LOG() << "SKIP: kernel reports AIS unavailable (ENODEV) on node "
```
kentrussell, line 441: We need to consolidate test-skip output so that it can be parsed by people who wish to do such things. Again, not a blocker

kentrussell (approved):
Not worth blocking on 2 comments. Something for some future cleanup

PR #8620, opened 2026-07-15
fix(rccl): Enable fabric handles in ncclMemAlloc_impl

AbandiGa:
## Motivation

In this PR, we enable the Fabric handles by default in ncclMemAlloc_impl. On devices that do not support Fabric, the implementation automatically falls back to the POSIX file descriptor handle type. This approach is required because HIP does not support a combined handle-type bitmask like CUDA.

## JIRA ID

AICOMRCCL-928

## Test Plan

The changes got tested in MI355 to make sure it's functional.

## Submission Checklist


AbandiGa:
Changes were merged in PR #8945

PR #8626, opened 2026-07-15
fix(rocshmem): install test binaries to CMAKE_INSTALL_BINDIR

edgargabriel:
## Motivation
Install rocshmem_functional_tests and rocshmem_unit_tests to
CMAKE_INSTALL_BINDIR instead of CMAKE_INSTALL_DATADIR/rocshmem so that
executables land in the standard binary directory and are found by the
monorepo CI CTest pass.

Also probe for Python3 and PyYAML before register_all_*_tests() in both
functional and unit CMakeLists. Without PyYAML, test_categories.yaml
cannot be parsed and tier labels (quick/standard/comprehensive/full) are
not applied, making tests invisible to `ctest -L <tier>` in CI. If
either is absent, emit a warning and skip registration entirely.

## JIRA ID

JIRA ID: AIROCSHMEM-340

## Submission Checklist


amirsojoodi (approved):

abouteiller:
Shouldn't they go to libexec or is the-rock recommendation to set them differently?

edgargabriel:
> Shouldn't they go to libexec or is the-rock recommendation to set them differently?

we tried that, it didn't work, the executable was not found when theRock was running the CI. This is the only solution that seemed to work. It is weird, I agree.

PR #8627, opened 2026-07-15
fix(rocr): [AILIKFD-55] Detect when KFD reports zero L2 cache line size

andmar-amd:
Cache line size affects alignment requirements for kernargs and other data structures, so accurate reporting is critical. 

Changes:
- Add kfdtest that validates L2 cache line size is reasonable
- WARNS (doesn't fail) if value is 0 - provides visibility for bring-up
- FAILS if non-zero value is invalid (not power of 2, or out of 64-256 range)
- Runtime reports exactly what KFD provides (no defaults)
- Runtime emits fprintf warning if value is 0 (visible in all build types)
- Warning also visible if amdgpu driver reloaded with debug enabled

This detects potential issues during bring-up/testing while being tolerant of legacy hardware or special cases where zero might be valid.

Co-Authored-By: Claude Sonnet 4

## JIRA ID

JIRA ID: AILIKFD-55

## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDTopologyTest.cpp
+++ b/projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDTopologyTest.cpp
@@ -218,6 +218,73 @@ TEST_F(KFDTopologyTest, GetNodeCacheProperties) {
     TEST_END
 }
 
+// Test that L2 cache line size is reported with a valid value
+// This is critical for newer ASICs which have 128 or 256 byte cache lines.
+// Cache line size affects alignment requirements for kernargs and other data structures.
+TEST_F(KFDTopologyTest, ValidateL2CacheLineSize) {
+    TEST_START(TESTPROFILE_RUNALL)
+
+    const HsaNodeProperties *pNodeProperties;
+    bool foundGpuWithL2 = false;
+
+    for (unsigned node = 0; node < m_SystemProperties.NumNodes; node++) {
+        pNodeProperties = m_NodeInfo.GetNodeProperties(node);
+        ASSERT_NE(pNodeProperties, nullptr) << "Failed to get properties for node " << node;
+
+        if (pNodeProperties->NumFComputeCores > 0) {  // GPU node
+            HsaCacheProperties *cacheProperties = new HsaCacheProperties[pNodeProperties->NumCaches];
+            auto status = HSAKMT_CALL(hsaKmtGetNodeCacheProperties, g_baseTest->m_hsakmt_current_ctx,
+                           node, pNodeProperties->CComputeIdLo,
+                           pNodeProperties->NumCaches, cacheProperties);
+            EXPECT_SUCCESS(status);
+            if (status != HSAKMT_STATUS_SUCCESS) {
+                delete [] cacheProperties;
+                continue;
+            }
+
+            for (unsigned n = 0; n < pNodeProperties->NumCaches; n++) {
+                if (cacheProperties[n].CacheLevel == 2) {
+                    foundGpuWithL2 = true;
+                    HSAuint32 size = cacheProperties[n].CacheLineSize;
+
+                    LOG() << "GPU Node " << node << " L2 cache line size: "
+                          << size << " bytes" << std::endl;
+
+                    // Cache line size must not be zero - it is not valid
+                    EXPECT_NE(size, 0UL)
+                        << "GPU Node " << node << " L2 cache line size is 0. "
+                        << "This is not valid - KFD topology must be correctly configured. "
+                        << "Newer ASICs typically have 128 or 256 byte cache lines.";
+
+                    // For non-zero values, validate they are reasonable
+                    if (size != 0) {
+                        // Must be a power of 2
+                        EXPECT_EQ(size & (size - 1), 0UL)
+                            << "GPU Node " << node << " L2 cache line size " << size
+                            << " is not a power of 2";
+
+                        // Should be in reasonable range for modern GPUs
+                        // Typical values: 64, 128, 256 bytes
+                        EXPECT_GE(size, 64UL)
+                            << "GPU Node " << node << " L2 cache line size " << size
+                            << " is unexpectedly small (< 64 bytes)";
+                        EXPECT_LE(size, 256UL)
+                            << "GPU Node " << node << " L2 cache line size " << size
+                            << " is unexpectedly large (> 256 bytes)";
+                    }
+                }
+            }
+            delete [] cacheProperties;
+        }
+    }
+
+    if (!foundGpuWithL2) {
+        LOG() << "No GPU nodes with L2 cache found, skipping validation" << std::endl;
+    }
+
+    TEST_END
+}
+
 // Test that we can get NodeIoLink property successfully per node
 // TODO: Check validity of values returned
 // GetNodeIoLinkProperties is disabled for now, test fails due to bug in BIOS


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -2367,13 +2367,24 @@ hsa_status_t GpuAgent::GetInfo(hsa_agent_info_t attribute, void* value) const {
...
+              warned = true;
+            }
+          }
           return HSA_STATUS_SUCCESS;
         }
       }
-      // Fallback for when KFD is returning zero.
-      *((uint32_t*)value) = 64;
+      // No L2 cache found - shouldn't happen on GPU
+      *((uint32_t*)value) = 0;
```

**[Important]** This removes the previous `*value = 64` fallback and now returns **0** for `HSA_AMD_AGENT_INFO_CACHELINE_SIZE` when no (valid) L2 entry is found. The old fallback guaranteed callers always got a usable, non-zero value. A cache line size of 0 is not a valid alignment/quantum — any consumer using this for alignment math (`align_up`, division, modulo) could hit divide-by-zero or invalid alignment (the new test's own comment notes cache line size drives kernarg alignment).

**[Question]** Which components consume this attribute today, and do they tolerate 0? Consider keeping the warning but still returning a safe non-zero fallback (e.g. 64) so downstream alignment stays valid, or document that 0 is now a possible return after auditing consumers.

cfreeamd (commented):

The new kfdtest `ValidateL2CacheLineSize` structure looks reasonable (power-of-2 + range checks gated on non-zero, non-fatal warning otherwise). Comments below focus on the `amd_gpu_agent.cpp` behavior change for `HSA_AMD_AGENT_INFO_CACHELINE_SIZE`.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -2367,13 +2367,24 @@ hsa_status_t GpuAgent::GetInfo(hsa_agent_info_t attribute, void* value) const {
       break;
     case HSA_AMD_AGENT_INFO_CACHELINE_SIZE:
       for (auto& cache : cache_props_) {
-        if ((cache.CacheLevel == 2) && (cache.CacheLineSize != 0)) {
+        if (cache.CacheLevel == 2) {
```

**[Important]** The old condition `(CacheLevel == 2) && (CacheLineSize != 0)` skipped zero-line-size L2 entries and kept scanning, so a valid later L2 entry could still be returned. This new `if (cache.CacheLevel == 2)` returns on the **first** L2 entry unconditionally, so a leading zero entry short-circuits and returns 0, masking a valid subsequent entry. Moot for single-L2 topologies, but a behavior regression for multi-entry cases.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -2367,13 +2367,24 @@ hsa_status_t GpuAgent::GetInfo(hsa_agent_info_t attribute, void* value) const {
...
+          // Warn if cache line size is zero - use fprintf for visibility in all builds
+          // Can also be detected by reloading amdgpu driver with debug enabled to push to system log
+          if (cache.CacheLineSize == 0) {
+            static bool warned = false;
+            if (!warned) {
+              fprintf(stderr, "ROCr Warning: L2 cache line size is reported as 0 for GPU node %u. "
+                      "Newer ASICs typically have 128 or 256 byte cache lines. "
+                      "Please verify KFD topology is correctly configured.\n", node_id());
+              warned = true;
+            }
```

**[Nit]** `GetInfo` is `const` and can be called concurrently from multiple threads; the unsynchronized read/write of `static bool warned` is a benign but real data race. Consider `std::atomic<bool>` with `exchange`, or `std::call_once`.

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDTopologyTest.cpp
```diff
@@ -218,6 +218,67 @@ TEST_F(KFDTopologyTest, GetNodeCacheProperties) {
...
+    const HsaNodeProperties *pNodeProperties;
+    bool foundGpuWithL2 = false;
+
+    for (unsigned node = 0; node < m_SystemProperties.NumNodes; node++) {
+        pNodeProperties = m_NodeInfo.GetNodeProperties(node);
+        if (pNodeProperties != NULL && pNodeProperties->NumCPUCores == 0) {  // GPU node
+            HsaCacheProperties *cacheProperties = new HsaCacheProperties[pNodeProperties->NumCaches];
+            EXPECT_SUCCESS(HSAKMT_CALL(hsaKmtGetNodeCacheProperties, g_baseTest->m_hsakmt_current_ctx,
+                           node, pNodeProperties->CComputeIdLo,
+                           pNodeProperties->NumCaches, cacheProperties));
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `EXPECT_SUCCESS` does not abort on failure. If `hsaKmtGetNodeCacheProperties` fails, the loop below reads uninitialized `cacheProperties[n]` (`new[]` without value-init) — UB. Prefer `ASSERT_SUCCESS` (or `continue`/`delete[]` on failure), and consider a value-initialized `std::vector<HsaCacheProperties>` for leak/exception safety.

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDTopologyTest.cpp
```diff
@@ -218,6 +218,67 @@ TEST_F(KFDTopologyTest, GetNodeCacheProperties) {
...
+
+            for (unsigned n = 0; n < pNodeProperties->NumCaches; n++) {
+                if (cacheProperties[n].CacheLevel == 2) {
+                    foundGpuWithL2 = true;
+                    HSAuint32 size = cacheProperties[n].CacheLineSize;
+
+                    LOG() << "GPU Node " << node << " L2 cache line size: "
+                          << size << " bytes" << std::endl;
+
+                    // Warn if cache line size is zero (may be legitimate for some ASICs)
```
dayatsin-amd: The test should fail if cacheline is 0. It is not valid to have a cacheline size of 0.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -2367,13 +2367,24 @@ hsa_status_t GpuAgent::GetInfo(hsa_agent_info_t attribute, void* value) const {
       break;
     case HSA_AMD_AGENT_INFO_CACHELINE_SIZE:
       for (auto& cache : cache_props_) {
-        if ((cache.CacheLevel == 2) && (cache.CacheLineSize != 0)) {
+        if (cache.CacheLevel == 2) {
           *((uint32_t*)value) = cache.CacheLineSize;
+          // Warn if cache line size is zero - use fprintf for visibility in all builds
```
dayatsin-amd: We do not need the changes to amd_gpu_agent. We want to be aware that KFD is reporting a cacheline of zero, but we do not want this to print each time an application runs.

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDTopologyTest.cpp
andmar-amd: ok, will fix that!

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
andmar-amd: ok, got it!

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDTopologyTest.cpp
```diff
@@ -218,6 +218,71 @@ TEST_F(KFDTopologyTest, GetNodeCacheProperties) {
...
+// Cache line size affects alignment requirements for kernargs and other data structures.
+TEST_F(KFDTopologyTest, ValidateL2CacheLineSize) {
+    TEST_START(TESTPROFILE_RUNALL)
+
+    const HsaNodeProperties *pNodeProperties;
+    bool foundGpuWithL2 = false;
+
+    for (unsigned node = 0; node < m_SystemProperties.NumNodes; node++) {
+        pNodeProperties = m_NodeInfo.GetNodeProperties(node);
+        if (pNodeProperties != NULL && pNodeProperties->NumCPUCores == 0) {  // GPU node
```
dayatsin-amd: pNodeProperties == null should be a test failure.

Check for NumGPUCores  > 0 instead

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDTopologyTest.cpp
```diff
@@ -218,6 +218,71 @@ TEST_F(KFDTopologyTest, GetNodeCacheProperties) {
...
+            for (unsigned n = 0; n < pNodeProperties->NumCaches; n++) {
+                if (cacheProperties[n].CacheLevel == 2) {
+                    foundGpuWithL2 = true;
+                    HSAuint32 size = cacheProperties[n].CacheLineSize;
+
+                    LOG() << "GPU Node " << node << " L2 cache line size: "
+                          << size << " bytes" << std::endl;
+
+                    // Cache line size must not be zero - it is not valid
+                    EXPECT_NE(size, (HSAuint32)0)
```
dayatsin-amd: Try to use 0UL instead to declare an unsigned long constant

dayatsin-amd (dismissed):

kentrussell (approved):

PR #8628, opened 2026-07-15
fix(rocjitsu): avoid plugin false-firing on D16 load completion

newling:
### Motivation 

Fixes a false positive in rocjitsu race detection for D16 vector memory loads (gfx1151)

D16 loads merge the loaded half into the destination VGPR while preserving the other half. Rocjitsu was reading the preserved half through the plugin-observed VGPR read path during memory completion, so the race detector could report a same-instruction self-conflict after the load event had already been registered.

This changes D16 memory completion to read the preserved destination half as internal storage state rather than as an instruction-visible VGPR read.

 ### Tests
 
New tests added


lialan (approved):
LGTM.

PR #8635, opened 2026-07-15
fix(rocprofiler-sdk): rocpd modules should be backwards compatible

yhuiYH:
## Motivation

With the bump to rocpd schema v3.0.2, the converting scripts only parse 3.0.2 schema.  We need to fix that to support older schemas like 3.0.1.  The script should detect the schema and convert seamlessly.

Since the rebase, also included backward compatibility support for schema 3.0.3 as well (and the accompanying tests).

## Technical Details

Check the schema version of the database being opened
Apply the correct sql queries and don't try to access missing tables/views/fields.

## JIRA ID

JIRA ID: ROCPDSNA-72

## Test Plan

Added rocpd-back-compat test. Do "ctest -R rocpd-back-compat" to run them.

This new set of tests will:
1) Identify & fail if a new schema version is added, but backwards compatibility tests are not added
2) Small make_db.py script is added to make minimal databases that have fake data. It will make a small fake database for each schema version it retrieves from the API
3) Using that small fake DB, it will run convert tests for CSV, pftrace and OTF2 and ensure each succeeds
4) If there are new fields/tables added, the  validate.py tests should be extended to check for the new schema version & check for the new data.

## Test Result

rocpd-back-compat tests all passing

## Submission Checklist


projects/rocprofiler-sdk/source/lib/python/rocpd/libpyrocpd.cpp
```diff
@@ -492,7 +498,43 @@ PYBIND11_MODULE(libpyrocpd, pyrocpd)
...
+                    }
+                }
+                for(const auto& part : rocprofiler::sdk::parse::tokenize(ver_str, "."))
+                {
+                    parts.emplace_back(part);
+                }
+                if(!parts.empty()) version.major = static_cast<uint32_t>(std::stoul(parts.at(0)));
+                if(parts.size() > 1) version.minor = static_cast<uint32_t>(std::stoul(parts.at(1)));
+                if(parts.size() > 2) version.patch = static_cast<uint32_t>(std::stoul(parts.at(2)));
+                return version;
```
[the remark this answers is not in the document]
yhuiYH: Sure, changed.

projects/rocprofiler-sdk/tests/rocpd-back-compat/conftest.py
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env python3
+
+# MIT License
+#
+# Copyright (c) 2023-2025 Advanced Micro Devices, Inc. All rights reserved.
```
mcao59: Minor: year should be 2026? (same for other new files in this dir)

projects/rocprofiler-sdk/source/lib/python/rocpd/libpyrocpd.cpp
```diff
@@ -492,7 +498,57 @@ PYBIND11_MODULE(libpyrocpd, pyrocpd)
...
+            auto get_db_schema_version = [&data](sqlite3* c) -> rocpd_version_triplet_t {
+                auto        version = rocpd_version_triplet_t{0, 0, 0};
+                auto* const q       = const_cast<char*>(
+                    "SELECT value FROM rocpd_metadata WHERE tag='schema_version'");
+                auto ver_str = data.schema_version_str;
+                auto parts   = std::vector<std::string>{};
+                parts.reserve(3);
+
+                // if the schema_version_str is not set, query the database for the schema_version
+                if(ver_str == "0.0.0")
```
mcao59:
```
if(ver_str.empty() || ver_str == "0.0.0")
{
    // read schema_version from rocpd_metadata
}
```
Minor: schema_version_str defaults to an empty string, while the metadata fallback recognizes only "0.0.0". A caller using libpyrocpd directly could create a valid native object with an empty version. So may be worth checking ver_str.empty() as well

mcao59:
It looks like #8972 is addressing overlapping schema-version handling
Should these PRs be coordinated so we have one implementation?

yhuiYH:
> It looks like #8972 is addressing overlapping schema-version handling Should these PRs be coordinated so we have one implementation?

Yup, I saw it.  I'll try to incorporate Jonathan's partial change into this PR (they're pretty similar anyway).  His PR only addressed CSVs. But this one, I tried to address CSVs, PFtrace and OTF2, and added the tests.  So I'll work on improving this one.

projects/rocprofiler-sdk/source/lib/python/rocpd/otf2.py
```diff
@@ -174,6 +175,10 @@ def add_graph_attributes(
                     kernel_rename = getattr(config, "kernel_rename")
                     agent_index_value = getattr(config, "agent_index_value")
 
+                    hip_graph_fields = []
+                    if _schema_version >= (3, 0, 2):
```
bgopesh: Magic schema version (3,0,2) duplicated across 6 files — no single source of truth                                                                                                                                                                                                            
  - csv.py:226,284,311, otf2.py:179,294,433, libpyrocpd.cpp:527, make_db.py:211,1609, validate.py, CMakeLists.txt:958

bgopesh (approved):
Nothing Major. Just one comment

projects/rocprofiler-sdk/source/lib/python/rocpd/csv.py
```diff
@@ -213,22 +213,26 @@ def build_agent_id_string(agent_index_value, prefix=""):
...
     if config.kernel_rename:
         kernel_name = "region"
     else:
         kernel_name = "name"
 
+    # check schema version and add new fields:
+    hip_graph_fields = ()
+    if get_schema_version(importData) >= (3, 0, 2):
+        hip_graph_fields = ("graph_exec_id", "graph_node_id")
+
```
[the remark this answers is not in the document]
yhuiYH: This was intentional clean up.  csv.py will **auto apply** the title case, so no need to explicitly add the alias.

projects/rocprofiler-sdk/source/lib/python/rocpd/csv.py
```diff
@@ -271,14 +279,21 @@ def write_memory_copy_csv(importData, config) -> None:
     src_agent_id = build_agent_id_string(config.agent_index_value, "src")
     dst_agent_id = build_agent_id_string(config.agent_index_value, "dst")
 
+    # check schema version and add new fields:
+    hip_graph_fields = []
+    if get_schema_version(importData) >= (3, 0, 2):
+        hip_graph_fields = [
+            "graph_exec_id",
+            "graph_node_id",
+        ]
```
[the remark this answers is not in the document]
yhuiYH: This was intentional clean up.  csv.py will **auto apply** the title case, so no need to explicitly add the alias.

projects/rocprofiler-sdk/tests/rocpd-back-compat/conftest.py
yhuiYH: Fixed. Thx!

projects/rocprofiler-sdk/source/lib/python/rocpd/libpyrocpd.cpp
yhuiYH: I removed the ver_str part when moving to the similar model as #8972 so this one is no longer a concern, I believe.

projects/rocprofiler-sdk/source/lib/python/rocpd/otf2.py
yhuiYH: Yes, unfortunately, we probably need to re-design the modules better so they can handle these schema-deltas better.  Right now, the short term fix, is the scattered checks.  
Longer term, need a cleaner, more concise strategy, otherwise the code will be a mess.  Thanks for calling it out.  I acknowledge the pain and welcome any suggestions on improving the maintainability of these schema specific items.

mcao59 (approved):

projects/rocprofiler-sdk/source/lib/python/rocpd/otf2.py
yhuiYH: @bgopesh , after thinking more about it last evening, I ended up adding a single source of truth for the python modules & C++.  Haven't gotten around to the tests relying on it yet, but will eventually add it.  Thx for the comment/feedback.

bgopesh: Thanks Young

projects/rocprofiler-sdk/source/lib/python/rocpd/source/perfetto.cpp
```diff
@@ -194,12 +194,13 @@ write_perfetto(
...
     const std::unordered_map<uint64_t, std::pair<rocpd::types::agent, tool::agent_index>>&
                                                      agent_data,
+    rocpd_version_triplet_t                          schema_version,
     const tool::generator<types::thread>&            thread_gen,
     const tool::generator<types::region>&            region_gen,
     const tool::generator<types::sample>&            sample_gen,
     const tool::generator<types::kernel_dispatch>&   kernel_dispatch_gen,
     const tool::generator<types::memory_copies>&     memory_copy_gen,
-    const tool::generator<types::graph_launch>&      graph_launch_gen,
+    const tool::generator<types::graph_launch>*      graph_launch_gen,
```
jrmadsen: Why must this be a pointer? Can't this just be a generator that returns no results?

yhuiYH: OK, I created an empty generator template that can be used to pass in whenever there's newly introduced generators that a schema doesn't support yet.  That way, no need for changing it to a pointer, and removed the null-check.

projects/rocprofiler-sdk/source/lib/python/rocpd/csv.py
```diff
@@ -213,22 +213,26 @@ def build_agent_id_string(agent_index_value, prefix=""):
...
+
     select_columns = [
         "guid",
         "'KERNEL_DISPATCH' AS Kind",
         f"{agent_id} AS Agent_Id",
         "queue_id",
         "stream_id",
-        "graph_exec_id AS Graph_Exec_Id",
-        "graph_node_id AS Graph_Node_Id",
+        *hip_graph_fields,
```
jrmadsen, line 235: I have a feeling people are going to complain about this. We might need to allow the user to specify whether they want this data in the CSV.

projects/rocprofiler-sdk/source/lib/python/rocpd/libpyrocpd.cpp
```diff
@@ -354,16 +359,28 @@ PYBIND11_MODULE(libpyrocpd, pyrocpd)
...
 
     py::class_<rocpd::RocpdImportData>(pyrocpd, "RocpdImportData", "RocPD database(s) instances")
         .def(py::init<>())
         .def(py::init<rocpd::RocpdImportData>())
         .def(py::init<py::object, std::vector<std::string>>())
         .def_readonly("connection", &rocpd::RocpdImportData::connection)
-        .def_readonly("databases", &rocpd::RocpdImportData::databases);
+        .def_readonly("databases", &rocpd::RocpdImportData::databases)
+        .def_readwrite("schema_version", &rocpd::RocpdImportData::schema_version)
+        .def_readwrite("supported_features", &rocpd::RocpdImportData::supported_features);
```
jrmadsen: I'm not sure these should be readwrite. My feeling is that we don't want to allow users to directly manipulate these values.

projects/rocprofiler-sdk/source/lib/python/rocpd/csv.py
yhuiYH: Can you elaborate a bit here?  What do you mean by allow the user to specify if they want this data in the CSV?  Do you mean provide an option like `--csv-graph-launch` (not sure how sustainable that would be) or `--csv-include-fields="graph_exec_id, graph_node_id"`, but would have to rely on user knowing what fields are called in the schema?

projects/rocprofiler-sdk/source/lib/python/rocpd/libpyrocpd.cpp
yhuiYH: Initially, they needed to be readwrite because the python side was populating it.  I made the change to make them readonly and the population of the values are done now in C++ at construction time.  Then, they can be readonly and should not be manipulated at runtime.

jrmadsen (approved):

PR #8636, opened 2026-07-15
fix(rocDecode): Fix fake CTest passes.

jeffqjiangNew:
## Motivation

This PR corrects fake CTest passes.

## Technical Details

Fake rocDecode CTest passes are observed where CTest reports pass even when the decoder encounters critical errors.
 - The reason is the return value of the sample apps is still set 0 when critical error occurs.
 - rocDecode CTest currently uses the default check rule of checking app exit code.
 - Fixed sample apps return values when there are critical errors: return non-zero. 

## JIRA ID

AICV-191

## Test Plan

Create a controlled test case of critical decode error, for example, create decoder fails. Run CTest.

## Test Result

CTest properly reports failure with this PR.

## Submission Checklist


projects/rocdecode/samples/videoToSequence/videotosequence.cpp
```diff
@@ -481,7 +482,8 @@ int main(int argc, char **argv) {
                 std::right << std::hex << pci_domain_id << "." << pci_device_id << std::dec << std::endl;
             }
             if (!v_dec_info[thread_idx]->viddec->CodecSupported(v_dec_info[thread_idx]->dec_device_id, v_dec_info[thread_idx]->rocdec_codec_id, v_dec_info[thread_idx]->bit_depth)) {
-                std::cerr << "Codec not supported on GPU, skipping this file!" << std::endl;
+                std::cerr << "Error: Codec not supported on GPU for " << input_file_names[j] << "!" << std::endl;
+                b_unsupported_codec = true;
                 continue;
             }
```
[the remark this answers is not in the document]
jeffqjiangNew: Addressed.

AryanSalmanpour (approved):

kiritigowda (approved):

projects/rocdecode/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Full documentation for rocDecode is available at [https://rocm.docs.amd.com/proj
 ### Resolved issues
 
 * Fixed decode errors of some AVC interlaced container streams by adding support for the picture data packet from the demuxer which contains multiple pictures.
+* Corrected fake CTest passes.
```
spolifroni-amd, line 14: I know this has already been merged in, but I think this should be "false positive" ctest passes or something like that.

spolifroni-amd (commented):
The changelog wording should be changed at some point before release.

projects/rocdecode/CHANGELOG.md
jeffqjiangNew: I think "false negative" is the correct one since the issue is that some critical errors could not be reported by the CTest (it passed when it shouldn't).

PR #8641, opened 2026-07-15
fix(rccl): fix ddaPeerPtrsHost heap overflow in DDA fabric path

Nikhil-Nunna:
## Motivation

Fix a heap buffer overflow in the DDA fabric path that corrupts the ncclComm object on multi-node runs with more than 8 total ranks.

## Technical Details

ncclDdaFabricCommInit() wrote up to comm->nRanks pointers into ddaPeerPtrsHost, a fixed 8-element array left over from the single-node IPC DDA path. For nRanks > 8 this overflowed past the array into endMagic and beyond the ncclComm heap allocation.

ddaPeerPtrsHost is now heap-allocated and sized correctly per active path (8 entries for the IPC path, comm->nRanks for the fabric path), freed on every error/teardown path. Also added a fallback so the fabric path is skipped when a communicator spans more than one fabric clique, since the handle exchange only works within one clique.

## JIRA ID

AICOMRCCL-1541

## Test Plan

Incremental build of the changed files. rccl-tests on 8 GPUs (gfx942) exercising the shared IPC DDA code path across AllReduce, AllGather, ReduceScatter, and AllToAll.

## Test Result

Build clean, no warnings or errors. rccl-tests correctness passed on all four collectives. Multi-node gfx1250 hardware validation of the original crash repro is still outstanding.

## Submission Checklist

### DIFF
--- a/projects/rccl/src/include/comm.h
+++ b/projects/rccl/src/include/comm.h
@@ -950,7 +950,14 @@ struct ncclComm {
   // Temporary Buffer [RCCL]
   void* tempBuff;
 
-  void* ddaPeerPtrsHost[nccl_dda_detail::kDdaNranks];
+  // Host mirror of the per-rank DDA scratch base pointers. Heap-allocated once
+  // per comm and sized to the number of ranks actually served by the active DDA
+  // path (kDdaNranks for the IPC path, comm->nRanks for the fabric path), not a
+  // fixed in-struct array: the fabric path admits up to kDdaMaxNranks ranks, so
+  // a fixed kDdaNranks-element array here would overflow into endMagic and past
+  // the end of the ncclComm allocation. Allocated in the *CommInit routines and
+  // freed in the matching *CommFini; nullptr when no DDA path is active.
+  void** ddaPeerPtrsHost;
 
   uint64_t endMagic;
 };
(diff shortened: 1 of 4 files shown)


projects/rccl/src/include/comm.h
```diff
@@ -950,7 +950,14 @@ struct ncclComm {
...
 
-  void* ddaPeerPtrsHost[nccl_dda_detail::kDdaNranks];
+  // Host mirror of the per-rank DDA scratch base pointers. Heap-allocated once
+  // per comm and sized to the number of ranks actually served by the active DDA
+  // path (kDdaNranks for the IPC path, comm->nRanks for the fabric path), not a
+  // fixed in-struct array: the fabric path admits up to kDdaMaxNranks ranks, so
+  // a fixed kDdaNranks-element array here would overflow into endMagic and past
+  // the end of the ncclComm allocation. Allocated in the *CommInit routines and
+  // freed in the matching *CommFini; nullptr when no DDA path is active.
+  void** ddaPeerPtrsHost;
```
atulkulk: nitpick: Pls restrict the comment size. In the other file as well.

Nikhil-Nunna: done

nawrinsu (approved):

nileshnegi (approved):

Nikhil-Nunna:
Validated on multi-node MI450 hardware — 2 nodes, 8 GPUs — with the new heap-allocated `ddaPeerPtrsHost` path and clique-scoping guard both exercised at runtime, and it passed cleanly.

PR #8644, opened 2026-07-15
feat(rocprofiler-sdk): share per-agent ATT resources across contexts

minseobshin11:
## Motivation

ATT allocates heavyweight resources per context and per agent, even though only one thread trace can use an agent at a time:

- A separate GPU output-buffer set per context wastes device memory. At ~1 GiB each, a few hundred contexts exceed a single device's HBM.
- A separate HSA submission queue per context exhausts the agent's queue budget. `HSA_AGENT_INFO_QUEUES_MAX` is 128 on `gfx942` and is shared with the application's own queues; exceeding it aborts.

This PR gives each GPU agent a single resource owner. Contexts targeting the same agent reuse its output slots, submission queue, and CPU staging buffers.

## Technical Details

- Adds `shared_trace_resources.{hpp,cpp}` with one `AgentTraceResources` per agent, owning the submission queue, CPU staging buffers, and GPU output slots.
- Tracers, AQLProfile packets, and producer/consumer workers hold the owner by `shared_ptr`, so a buffer cannot be released while anything still references it.
- Collects each context's buffer and staging sizes before the first owner is constructed. The output buffer is allocated from the AQLProfile allocation callback during the first tracer's construction, so its size must be final by then.
- Sizes output and staging slots per slot index. Slot `i` is sized to the largest context that reaches it, so a context requesting many small buffers does not inherit an unrelated context's larger buffer size.
- Maps AQLProfile output allocations by packet-local slot index, so slot `i` resolves to the same buffer for every context on an agent. Host-visible allocations stay per packet.
- Skips the AQLProfile free callback for borrowed output buffers, which are released once with the agent's resources rather than by the first context to be destroyed.
- Passes each trace's own buffer size to the producer thread, since shared staging buffers are sized to the largest context on the agent.
- Records the owning context on its first trace and fails loudly if a second context tries to trace the same agent concurrently.
- Waits for a trace's stop packet to retire before releasing the agent's resources.

## JIRA ID

AIPROFSDK-102

## Test Plan

- ~400 contexts per agent in both device and dispatch mode, each requesting a 1 GiB output buffer, then one trace run to confirm collection still works. This is over 3x the 128-queue limit, and ~400 GiB of output buffers if they were still per context.
- New unit coverage for shared output-slot reuse across contexts, per-slot staging sizing, and sequential lease handoff between contexts.
- Existing producer/consumer, standard ATT, pre-HSA-init, large-buffer, and triple-buffer tests.
- Hardware: AMD Instinct MI300X (`gfx942`).

## Test Result

- `thread-trace-packet-test`: 11/11 passed.
- `thread-trace-producer-consumer-test`: 10/10 passed.
- ATT integration suite: all 12 enabled tests passed (2 disabled upstream).
- Unit tests also pass under `--gtest_shuffle` across repeated runs.

## Submission Checklist


projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
```diff
@@ -418,6 +420,18 @@ DispatchThreadTracer::register_shared_buffer_sizes()
     }
 }
 
+void
+DispatchThreadTracer::register_shared_queue_sizes()
```
ApoKalipse-V: This is duplicated. Maybe join with above function?

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
```diff
@@ -592,6 +606,18 @@ DeviceThreadTracer::register_shared_buffer_sizes()
     }
 }
 
+void
+DeviceThreadTracer::register_shared_queue_sizes()
```
ApoKalipse-V: Same deduplication

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
```diff
@@ -654,12 +680,20 @@ initialize(HsaApiTable* table)
...
 
-    // Register every context's buffer sizes before resource_init builds buffers,
-    // so the shared per-agent buffers are sized to the max requested size.
+    // Register every context's buffer and queue sizes before resource_init builds
+    // them, so the shared per-agent buffer and queue are sized to the max requested.
     for(auto& ctx : context::get_registered_contexts())
     {
-        if(ctx->device_thread_trace) ctx->device_thread_trace->register_shared_buffer_sizes();
-        if(ctx->dispatch_thread_trace) ctx->dispatch_thread_trace->register_shared_buffer_sizes();
+        if(ctx->device_thread_trace)
```
ApoKalipse-V: You can revert this change if they are joined

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/shared_trace_queue.cpp
```diff
@@ -0,0 +1,97 @@
...
+namespace thread_trace
+{
+namespace
+{
+struct agent_queue_t
+{
+    // Max triple-buffer staging size requested by any context on this agent; the
+    // shared queue is created with this size so every context's flips fit.
+    uint64_t        max_triple_buffer_size = 0;
+    att_queue_ptr_t queue                  = {};
```
ApoKalipse-V: Maybe join with the buffer struct? They have the same semantics.
(not mandatory)

projects/rocprofiler-sdk/tests/tools/thread-trace-many-contexts-tool.cpp
```diff
@@ -0,0 +1,318 @@
...
+    static const size_t n = std::getenv("ATT_NUM_CONTEXTS")
+                                ? std::strtoul(std::getenv("ATT_NUM_CONTEXTS"), nullptr, 10)
+                                : 400;
+    return n;
+}
+
+// Keep each context active across a few dispatches so it reliably captures one.
+constexpr uint64_t CAPTURE_WINDOW = 2;
+
+// Repro mode for the AILIKFD-39 teardown hang: start one context and never stop it,
```
ApoKalipse-V: Why are we reproing it here?

projects/rocprofiler-sdk/tests/tools/thread-trace-many-contexts-tool.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+void
+tool_fini(void* /* tool_data */)
+{
+    if(state == nullptr) return;
+
+    if(leave_active())
+    {
+        // Deliberately do NOT stop the trace context: leave SQTT active so rocprofiler
+        // teardown destroys the still-active per-agent tracers one-by-one (AILIKFD-39).
```
ApoKalipse-V: same comment
Also, this is unlikely to work in dispatch mode

projects/rocprofiler-sdk/tests/tools/thread-trace-many-contexts-tool.cpp
```diff
@@ -0,0 +1,318 @@
...
+    // Every context must have configured and allocated its (shared) buffer, which is
+    // the memory-sharing proof: this would OOM with a per-context buffer.
+    assert(configured == num_contexts() && "not all contexts configured/allocated");
+
+    // Every context should have had a turn.
+    assert(ran == configured && "not every context was run (need more dispatches)");
+
+    // Not every target_cu is guaranteed to have active waves, so require that MOST
+    // configs produced data rather than all.
+    assert(captured * 4 >= configured * 3 && "most configs should capture data");
```
ApoKalipse-V: This is not not going to fail even if the profiler gave you the same header-only buffer every time. You need to verify the waves are (mostly) valid in the capture window and that they are all in the correct SIMD/CUs you selected in the configuration.

projects/rocprofiler-sdk/tests/tools/thread-trace-many-contexts-tool.cpp
ApoKalipse-V, line 1: We also need to cover device thread trace

ApoKalipse-V (changes requested):

projects/rocprofiler-sdk/tests/tools/thread-trace-many-contexts-tool.cpp
ApoKalipse-V: Note that SIMD selection is different for gfx >= 10, you need to take SEL%4 while for gfx9 you take the bitmask.
Then in the validation: Verify you don't get wave_t from simd/cus you didnt ask for, and that you do get (most) of them for the ones you did. To verify waves are valid, you can check the instructions have a valid PC address (except for TRAP and for the first few in case of gfx1250).

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
minseobshin11: done

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
minseobshin11: fixed

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
minseobshin11: true-- changed

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/shared_trace_queue.cpp
minseobshin11: left them separate because their contents diverged after changes

projects/rocprofiler-sdk/tests/tools/thread-trace-many-contexts-tool.cpp
minseobshin11: removed -

projects/rocprofiler-sdk/tests/tools/thread-trace-many-contexts-tool.cpp
minseobshin11: removed -

projects/rocprofiler-sdk/tests/tools/thread-trace-many-contexts-tool.cpp
minseobshin11: fixed so that the test now decodes every captured trace and asserts real content

minseobshin11: I believe gfx9 emits all the wave records for all four SIMDs regardless of the requested simd_select - is there a way to validate on gfx10+

projects/rocprofiler-sdk/tests/tools/thread-trace-many-contexts-tool.cpp
minseobshin11: added more tests to cover

projects/rocprofiler-sdk/tests/thread-trace/CMakeLists.txt
```diff
@@ -74,6 +74,32 @@ rocprofiler_add_integration_execute_test(
...
+    PRELOAD "$<TARGET_FILE:thread-trace-many-contexts-tool>"
+    # The app emits 2 dispatches/iteration; device rotation advances ~1 context every 2-3
+    # dispatches, so keep well above ~3x num_contexts of dispatches for margin.
+    ENVIRONMENT "ATT_NUM_CONTEXTS=400;ATT_APP_LOOPCOUNT=1600;ATT_TRACE_MODE=device"
+    DISABLED ${IS_DISABLED})
+
+rocprofiler_add_integration_execute_test(
+    thread-trace-api-many-contexts-dispatch-test
+    TARGET thread-trace-test-app
+    TIMEOUT 300
```
ApoKalipse-V: Too long timeout

projects/rocprofiler-sdk/tests/thread-trace/CMakeLists.txt
```diff
@@ -74,6 +74,32 @@ rocprofiler_add_integration_execute_test(
...
+# Shared-buffer/queue stress: allocate hundreds of ~1GB contexts with varying params and
+# run each once, decoding each trace to verify it holds real waves with valid PCs from the
+# shader engines the context selected. Only fits in memory / the HSA queue limit because
+# all contexts on an agent share one buffer and one queue; a per-context buffer/queue
+# would need ~400GB and exhaust the per-agent queue limit. Exercised in device and
+# dispatch modes.
+rocprofiler_add_integration_execute_test(
+    thread-trace-api-many-contexts-test
+    TARGET thread-trace-test-app
+    TIMEOUT 300
```
ApoKalipse-V: Too long timeout

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/core.cpp
```diff
@@ -219,6 +232,7 @@ ThreadTracerAgent::iterate_data(aqlprofile_handle_t handle, rocprofiler_user_dat
         ROCP_CI_LOG(ERROR) << "Failed to iterate ATT data: " << status;
 
     active_traces.fetch_sub(1);
+    resources->end_trace(params.context_id);
```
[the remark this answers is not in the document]
minseobshin11: Fixed. Moved `trace_resources_mut` so the lease release is serialized against load_codeobj's check and submit.

PR #8646, opened 2026-07-15
refactor(rocprofiler-compute): extract shared mem_chart helpers 

xuchen-amd:
## Motivation

`mem_chart_gfx11.py` contained general-purpose formatting functions that are not specific to the RDNA3.5 architecture. Moving them to a shared module avoids duplication when adding chart renderers for other architectures.

Prep work for #8648 

## Technical Details

- New file `src/utils/mem_chart_common.py` — contains common helper functions
- `src/utils/mem_chart_gfx11.py` — removed the inlined helpers and imports them from `mem_chart_common`. Replaced 10 direct `format_bw_human_readable` call sites with `format_value(..., "Bytes/s", ...)` to consolidate bandwidth formatting through a single code path.
- `tests/test_mem_chart.py` — updated existing tests to reference `mem_chart_common` instead of `mem_chart_gfx11`. Added new tests for `safe_float_sum` and `fmt_edge`. Fixed a fragile test import that accessed `format_bw_human_readable` through a re-export instead of its canonical module.

## JIRA ID

AIPROFCOMP-116

## Test Plan

- All test_mem_chart.py tests pass (gfx11 + common helpers)

## Test Result

- [x] All test_mem_chart.py tests pass (gfx11 + common helpers)

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-compute/src/utils/mem_chart_gfx11.py
+++ b/projects/rocprofiler-compute/src/utils/mem_chart_gfx11.py
@@ -42,14 +42,12 @@ from rich.text import Text
 
 from utils.mem_chart_common import (
     COLORS,
-    _fmt_edge,
-    _safe_float_sum,
     bar,
-    format_sci,
+    fmt_edge,
     format_value,
     metric_line,
+    safe_float_sum,
 )
-from utils.utils_analysis import format_bw_human_readable
 
 # ---------------------------------------------------------------------------
 # Constants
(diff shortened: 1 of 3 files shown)


projects/rocprofiler-compute/src/utils/mem_chart_gfx11.py
```diff
@@ -40,7 +40,14 @@
 from rich.table import Table
 from rich.text import Text
 
-from utils.utils_analysis import format_bw_human_readable
+from utils.mem_chart_common import (
```
vedithal-amd, line 43: Well I dont see mem_chart_gfx9.py so how do we justify the existence of mem_chart_common.py if only mem_chart_gfx11.py is using it

Common modules with only one consumer does not make much sense since they are not really common module then

xuchen-amd: this PR is created to separate scope of changes, https://github.com/ROCm/rocm-systems/pull/8648 will reference the common modules.

vedithal-amd: I think PR should be logical self-consistent - if something is being extracted to a common utility - then all consumers should use that. I think separating this concern into https://github.com/ROCm/rocm-systems/pull/8648 is not the ideal way of splitting PRs.

Could we reduce the scope of https://github.com/ROCm/rocm-systems/pull/8648 and fold the scope into this PR so its read as logically self-consistent?

vedithal-amd: Actually you can in this PR still have mem chart gfx 9 use the common module right?
All we have to change is the import and https://github.com/ROCm/rocm-systems/pull/8648 will have to adjust based on what this PR lands

xuchen-amd:
Closing.
This work has been merged into #8648, which combines the common helper extraction with the full gfx9 Rich rewrite into a single PR. All commits from this branch (a89e2a3f4d, 00b7d2ced9) are included in the squashed commit f6fbe6009b.

PR #8647, opened 2026-07-15
ci(rocjitsu): Enable passing corpus tests in CI

RattataKing:
## Motivation

Some previously skipped corpus tests now pass after recent rocjitsu fixes.

## Technical Details

Removed some tests from the skip list.

## JIRA ID
Issue ID: https://github.com/ROCm/rocm-systems/issues/7891

## Test Plan

Run CI.

## Test Result

CI passed with the updated test config json.

## Submission Checklist


emulation/rocjitsu/tests/corpus/gfx1100_skip_tests.json
```diff
@@ -1,27 +1,21 @@
 {
   "iree": [
-    "iree.gfx1100.e2e.pack_i8.static_pack_vnni_lhs_large"
```
amd-eochoalo, line 3: I'd keep this one. I think it is stochastic.

amd-eochoalo (approved):
Thank you Amily!

kuhar (commented):
LGTM overall, just one comment

emulation/rocjitsu/tests/corpus/gfx1100_skip_tests.json
kuhar: I think we should remove this gfx1100 entry and the matching gfx1201 entry. At the CI-pinned corpus revision (`3278883f548e1353ffbda1a8090219d866eca035`), both target configs list `static_pack_vnni_lhs_large` in `skip_compile_tests`, so collection excludes the cases before these manifests are applied; direct selector matching leaves both entries unmatched. The gfx1250 entry still matches because that target uses `skip_run_tests`, so retaining it there is meaningful.

PR #8648, opened 2026-07-15
feat(rocprofiler-compute): redesign gfx9 mem chart with rich layout

xuchen-amd:
## Motivation

The gfx9 (CDNA) memory chart used plotille, a canvas-based renderer that drew boxes with coordinate math. The gfx11 (RDNA3.5) chart already used Rich panels and tables, which are easier to maintain, extend, and look better in terminals with color support. This PR rewrites the gfx9 chart to use Rich, unifies shared code between the two renderers, and adds new metrics to the memory chart.

## Technical Details

- Rewrite `mem_chart_gfx9.py` from plotille canvas to Rich panels, tables, and text
- Extract shared helpers into new `mem_chart_common.py`
- Refactor `mem_chart_gfx11.py` to use the shared helpers
- Add new metrics to gfx908–gfx950 YAML configs
- Gate xGMI/PCIe panels to gfx950 only; earlier archs show traffic percentages in the Data Fabric panel

### BEFORE (same rendering for all gfx9 archs, `tests/workloads/vcopy`)
image

#### gfx942
image

#### gfx950
image

## Issue Tracking

JIRA ID: AIPROFCOMP-776
JIRA ID: AIPROFCOMP-797

## Test Plan

- test_mem_chart.py and test_tty.py
- Verified chart output against real profiling data on MI200/MI300/MI350 hardware
- CI passes
## Test Result

- [x] test_mem_chart.py and test_tty.py
- [x] Verified chart output against real profiling data on MI200/MI300/MI350 hardware
- [x] CI passes

## Submission Checklist


projects/rocprofiler-compute/src/utils/mem_chart_gfx9.py
```diff
@@ -1,1270 +1,805 @@
...
+        metrics.get("vl1_l2_atomic_bytes"), vl1_peak, COLORS["atomic"]
+    )
+    arrow_left = arrows["left"]
+    arrow_right = arrows["right"]
+
+    vl1_rd_bw = format_value(metrics["vl1_l2_rd_bytes"], "Bytes/s", 1)
+    vl1_wr_bw = format_value(metrics["vl1_l2_wr_bytes"], "Bytes/s", 1)
+    vl1_at_bw = format_value(metrics["vl1_l2_atomic_bytes"], "Bytes/s", 1)
+    sl1d_rd_bw = format_value(metrics["sl1d_l2_rd_bytes"], "Bytes/s", 1)
+    il1_rd_bw = format_value(metrics["il1_l2_rd_bytes"], "Bytes/s", 1)
```
jamessiddeley-amd: Here these metrics are formatted as bytes/s, but as it stands in the codebase its actually bytes per kernel?
The counter counts requests per normalization unit, not per second:

```
         analysis_configs/gfx940/0300_memory_chart.yaml
         ...
        VL1_L2 Rd:
          value: ROUND(SUM(TCP_TCC_READ_REQ_sum) / SUM($denom), 0)
        VL1_L2 Wr:
          value: ROUND(SUM(TCP_TCC_WRITE_REQ_sum) / SUM($denom), 0)
         ...
```

The code then multiplies by the cacheline size (128 in this case below) to turn requests into bytes
This line in mem_chart_gfx9.py
`metrics["vl1_l2_rd_bytes"] = scale_or_none(get("VL1_L2 Rd"), 128)` => bytes moved per kernel
we convert requests to bytes

This is fine. But then it prints those bytes with a /s suffix, under Read BW / Write BW / Atomic BW labels with a time unit that sort of comes out of nowhere

If it _was_ "per second", the number shouldnt change if you ask the tool to report per-kernel vs per-wave since its a rate either way

I ran vcopy twice, changing only that setting and it does change when it shouldnt
L1->L2 "Read BW"
8.4 MB/s with per_kernel
512 B/s with per_wave

In your images in your PR description this shows there as well actually. 
Every "after" bandwidth is the "before" count times the cacheline size exactly
For ex `4,007,065 × 128 B = 512.9 MB`

I think we need to pick one label, either show request counts meaning drop the x128 and label it Rd, like the old chart did
or show bytes moved meaning keep the x128 but label it Read Bytes, with no /s

Let me know if I am missing something in my trace or if there was a design decision made to include the /s

projects/rocprofiler-compute/src/utils/mem_chart_gfx9.py
```diff
@@ -1,1270 +1,805 @@
...
+    arrows: dict[str, str],
+    peak_bw: Optional[PeakBandwidths] = None,
+) -> Text:
+    """L2→Fabric edges: Read BW and Write/Atomic BW."""
+    l2_peak = peak_bw.l2 if peak_bw else None
+    return build_bw_edge_column(
+        [
+            (
+                "Read BW",
+                format_value(metrics["l2_fabric_read_bw"], "Bytes/s", 1),
```
jamessiddeley-amd: Image

In the after memchart there is a calculation error in which we are dividing the bandwidths by 1e9 when we shouldnt be. Your after screenshot of the after shows 0.4 B/s, but it should be 0.4 GB/s.

This one looks like the opposite problem to the L1 -> L2 one above, these values come out 1e9 too small.
If we trace this from the yamls
The timestamps are in nanoseconds, so this formula gives gb/s:

``` 1700_l2_cache.yaml
        L2-Fabric Read BW:
          value: SUM(128 * TCC_BUBBLE_sum + 64 * (TCC_EA0_RDREQ_sum - TCC_BUBBLE_sum - TCC_EA0_RDREQ_32B_sum) + 32 * TCC_EA0_RDREQ_32B_sum) / SUM(End_Timestamp - Start_Timestamp)
          unit: GB/s
```

So the L2 panel prints this metric under GB/s

But in this line here that ive commented on, the chart gives that same value to `format_value(..., "Bytes/s")`, which reads it as bytes per second and scales it down moving forward throughout the entire memchart.

Thats why the entire after screenshot looks impossibly slow, like we're getting 1.5 KB/s L2-Fabric Read BW shown in the after memchart when it should be 1543.58 GB/s! Which is what it is in the L2 Panel block 17.

So I believe we should change bytes/s here to gb/s, and also update the six 0300_memory_chart.yaml files, change the unit Bytes/s to GB/s on the copied bw metrics which will then match 1700_l2_cache.yaml.

projects/rocprofiler-compute/src/utils/mem_chart_gfx9.py
```diff
@@ -1,1270 +1,805 @@
...
+# Public API helpers
+# ---------------------------------------------------------------------------
+
+
+def normalize_mem_chart_metrics(metric_dict: dict[str, Any]) -> dict[str, Any]:
+    """Filter/reorder input to panel key order; missing keys become None."""
+    return {k: metric_dict.get(k) for k in MEM_CHART_PANEL_METRIC_KEYS}
+
+
+def compute_peak_bw(sys_info_row: dict[str, Any]) -> PeakBandwidths:
```
jamessiddeley-amd: is `compute_peak_bw` meant to be wired up in this PR? Right now the only caller is test_mem_chart.py and peak_bw is never passed at either call site

projects/rocprofiler-compute/src/utils/mem_chart_gfx9.py
```diff
@@ -1,1270 +1,805 @@
...
+    ("L2 Rd", 300),
+    ("L2 Wr", 52),
+    ("L2 Atomic", 12),
+    ("L2 Hit", 85),
+    ("L2 Rd Lat", 220),
+    ("L2 Wr Lat", 180),
+    ("Fabric_L2 Rd", 45),
+    ("Fabric_L2 Wr", 8),
+    ("Fabric_L2 Atomic", 1),
+    ("L2-Fabric Read BW", 45e9),
```
jamessiddeley-amd, line 102: I think these sample values are about 1e9 off from what the metric actually produces.

45e9 formatted as Bytes/s comes out as 45.0 GB/s which is good and fine for tests. So the chart looks right in every test and in the main() CLI demo, but real data shows 1.5 KB/s

See my comment about bottlenecking the memchart to bytes/s instead of gb/s

jamessiddeley-amd (changes requested):

jamessiddeley-amd:
Couple nits:

In `cli.rst` : "Visualized memory chart requires the width of the terminal output to be greater than or equal to 234 to display the whole chart properly."

I think the new memchart is not 234 columns anymore (less which is great) so that should be updated.

projects/rocprofiler-compute/src/utils/mem_chart_common.py
```diff
@@ -0,0 +1,409 @@
...
+        "both": "<" + "-" * (length - 2) + ">",
+    }
+
+
+def strip_ansi(text: str) -> str:
+    """Remove ANSI escape sequences from *text*."""
+    return re.sub(r"\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])", "", text)
+
+
+def format_mem_chart_heading(
```
jamessiddeley-amd, line 165: format_mem_chart_heading  here and _mem_chart_heading in tty.py have the exact same output, 

Worth to delete one and reuse code

projects/rocprofiler-compute/src/utils/mem_chart_common.py
```diff
@@ -0,0 +1,409 @@
...
+def make_arrows(length: int) -> dict[str, str]:
+    """Build arrow dict with left/right/both keys of given length."""
+    return {
+        "left": "<" + "-" * (length - 1),
+        "right": "-" * (length - 1) + ">",
+        "both": "<" + "-" * (length - 2) + ">",
+    }
+
+
+def strip_ansi(text: str) -> str:
```
jamessiddeley-amd, line 149: Same helper already in tests/common.py, we can have that test just import this one now

projects/rocprofiler-compute/src/utils/mem_chart_gfx11.py
```diff
@@ -243,20 +168,6 @@ def _print_mem_chart_scope_bar(console: Console) -> None:
...
-
-
-def get_sample_metrics() -> dict[str, Any]:
-    """Return sample metrics (flat panel order) for testing or demos."""
-    return normalize_mem_chart_metrics(DEFAULT_SAMPLE_METRICS.copy())
-
-
 # ---------------------------------------------------------------------------
 # Diagram construction: _extract_metrics, _build_kernel_and_l0,
 #   _build_cache_columns, _build_memory_columns, create_mem_chart_diagram
```
jamessiddeley-amd: Remove stale comments, I believe these dont exist anymore

projects/rocprofiler-compute/src/utils/mem_chart_gfx9.py
```diff
@@ -1,1270 +1,805 @@
...
+        width=24,
+        height=3,
+    )
+    xgmi_layout = Table.grid(padding=0)
+    xgmi_layout.add_column(width=97)
+    xgmi_layout.add_column()
+    xgmi_layout.add_row("", xgmi_panel)
+    console.print(xgmi_layout)
+
+    pad = " " * 100
```
jamessiddeley-amd: Not sure we should be hardcoding spacing, between this, `add_column(width=97)` and `width=87` in the PCIe row, these blocks are positioned by three hardcoded numbers to stay in sync

Could `Table.grid` maybe carry these so we consolidate layout logic? Feel free to push back

projects/rocprofiler-compute/src/utils/mem_chart_common.py
```diff
@@ -0,0 +1,353 @@
...
+        height=80,
+    )
+    create_fn(flat, console, show_debug=False, **diagram_kwargs)
+    return buf.getvalue()
+
+
+def mem_chart_cli_main(
+    description: str,
+    create_fn: Callable[..., None],
+    normalize_fn: Callable[[dict[str, Any]], dict[str, Any]],
```
ggottipa-amd, line 328: normalize_fn is not being used in this function.

projects/rocprofiler-compute/src/utils/mem_chart_gfx9.py
```diff
@@ -1,1270 +1,750 @@
...
-
-    if key is not None:
-        key_format = make_format_spec(key_step_prec_leftalign, key_align)
-        key_str = f"{key:{key_format}}" if isinstance(key, (int, float)) else str(key)
-        result_str_no_unit = f"{key_str}{mark_between}{value_str}"
+    """Build Rich markup for the HBM panel (gfx950 BW metrics)."""
+    color_read = COLORS["read"]
+    color_write = COLORS["write"]
+    color_atomic = COLORS["atomic"]
+    color_rd = color_read
```
ggottipa-amd: Six bindings for three constant values. Can we use `color_read` / `color_write` / `color_atomic` directly?

projects/rocprofiler-compute/src/utils/mem_chart_common.py
```diff
@@ -0,0 +1,353 @@
...
+    *,
+    panel_id: int = 300,
+    section_label: str = "Memory Chart",
+) -> str:
+    """Build heading: '{panel_id//100}. {label} (Normalization: {unit})'."""
+    section = max(0, int(panel_id)) // 100
+    return f"{section}. {section_label} (Normalization: {normal_unit})"
+
+
+_LEGEND_ENTRIES: tuple[tuple[str, str, str], ...] = (
```
ggottipa-amd: Can we move `_LEGEND_ENTRIES` and `_STALL_ENTRY` to the top of the file, next to COLORS?

projects/rocprofiler-compute/src/utils/mem_chart_gfx9.py
```diff
@@ -1,1270 +1,750 @@
...
+    pcie_layout.add_column(width=fabric_col - (pcie_width - 24) // 2)
+    pcie_layout.add_column()
+    pcie_layout.add_row("", pcie_panel)
+    console.print(pcie_layout)
+
+
+def _measure_grid(grid: Table, console_width: int = 240) -> tuple[int, int]:
+    """Return (total_width, fabric_col) by rendering the grid to a buffer."""
+    buf = StringIO()
+    tmp = Console(file=buf, force_terminal=False, width=console_width, height=5)
```
ggottipa-amd: Use of `tmp` goes against the python style guide. Can we rename this variable?

projects/rocprofiler-compute/src/utils/mem_chart_common.py
```diff
@@ -0,0 +1,353 @@
...
+    "hit": "yellow",
+    "stall": "indian_red",
+    "bw": "bright_cyan",
+}
+
+
+def format_value(
+    value: Union[int, float, str, None], unit: str = "", precision: int = 1
+) -> str:
+    """Format a metric value with unit. Returns 'N/A' for None/NaN/invalid."""
```
ggottipa-amd, line 50: Please check if there is a chance of `format_value `returning `NaN` by delegating to `format_bw_human_readable`. Wondering if `(NaN, "Bytes/s")` is a valid input and returns `NaN`.

projects/rocprofiler-compute/src/utils/mem_chart_gfx9.py
xuchen-amd: good catch! fixed in 1e8ea629 and a258dce5 to read time-based formulas.

projects/rocprofiler-compute/src/utils/mem_chart_gfx9.py
xuchen-amd: fixed in a258dce5 by rewriting BW formulas to produce genuine `Bytes/s` and changed `unit:` from `GB/s` to `Bytes/s`

projects/rocprofiler-compute/src/utils/mem_chart_gfx9.py
xuchen-amd: removed in ad992add

projects/rocprofiler-compute/src/utils/mem_chart_gfx9.py
xuchen-amd: fixed in a258dce5

projects/rocprofiler-compute/src/utils/mem_chart_common.py
xuchen-amd: removed `_mem_chart_heading()` in de2afefb and imports `format_mem_chart_heading()` when needed

projects/rocprofiler-compute/src/utils/mem_chart_common.py
xuchen-amd: removed duplicate in de2afefb

projects/rocprofiler-compute/src/utils/mem_chart_gfx11.py
xuchen-amd: removed in 00faad31

projects/rocprofiler-compute/src/utils/mem_chart_common.py
xuchen-amd: fixed in 00faad31

projects/rocprofiler-compute/src/utils/mem_chart_common.py
xuchen-amd: fixed in 00faad31

projects/rocprofiler-compute/src/utils/mem_chart_gfx9.py
xuchen-amd: fixed in 00faad31

projects/rocprofiler-compute/src/utils/mem_chart_common.py
xuchen-amd: fixed in 00faad31

projects/rocprofiler-compute/src/utils/mem_chart_gfx9.py
xuchen-amd: addressed the hardcoded paddings by replacing with named constants and a computed formula in 00faad31.
 
moving away from manually drawn chart is underway and in discussion with Optiq, good suggestion!

projects/rocprofiler-compute/src/utils/tty.py
```diff
@@ -821,30 +814,30 @@ def format_table_output(
...
-                        int(table_config["id"]),
+                    chart_title=format_mem_chart_heading(
                         args.normal_unit,
+                        panel_id=int(table_config["id"]),
                     ),
                 )
                 + "\n"
             )
-        elif is_gfx115x(gpu_arch):
+        else:
```
ggottipa-amd: Can removing the positive test allow gfx1250 here?

projects/rocprofiler-compute/tests/common.py
```diff
@@ -236,11 +237,6 @@ def get_num_pmc_file(output_dir):
     ])
 
 
-def strip_ansi(s: str) -> str:
```
ggottipa-amd, line 126: This has a reference in test_gpu_specs.py.
https://github.com/ROCm/rocm-systems/blob/f68615debd87cf007ce4cdfce8d0acee8e3c379d/projects/rocprofiler-compute/tests/test_gpu_specs.py#L147

xuchen-amd: reverted in ed425b0

projects/rocprofiler-compute/src/utils/tty.py
xuchen-amd: restored in f9a337c1802c0bad28650c2c0a26f80f3a70e6db, good catch!

feizheng10 (changes requested):
Thanks for catching up on this!!

Here are some of my suggestions:
1. Interface consistency
plot_mem_chart API is inconsistent between gfx9 and gfx11
gfx9
plot_mem_chart(normal_unit, metric_dict, *, chart_title=None, gpu_arch=None)
gfx11
plot_mem_chart(metric_dict, *, chart_title: str)
2. MALL should not show up in old archs <= gfx90a. And the front-end layout can be improved a bit.
3. We should remove plotille in requirements.txt.

We can chat offline for some of the details.

xuchen-amd:
> Thanks for catching up on this!!
> 
> Here are some of my suggestions:
> 
> 1. Interface consistency
>    plot_mem_chart API is inconsistent between gfx9 and gfx11
>    gfx9
>    plot_mem_chart(normal_unit, metric_dict, *, chart_title=None, gpu_arch=None)
>    gfx11
>    plot_mem_chart(metric_dict, *, chart_title: str)
> 2. MALL should not show up in old archs <= gfx90a. And the front-end layout can be improved a bit.
> 3. We should remove plotille in requirements.txt.
> 
> We can chat offline for some of the details.

fixed 1 and 2 in a432fad8d1126cdcfeea5185edbc0db2be7dbc35, will take rest offline.

feizheng10 (approved):
Please go ahead after you double check:
* CI passed.
* Formula issues raised by James have been confirmed.
* From the snapshot, I don't see Util rate and Bank conflict for LDS. Please make sure those show up as expected with the related test case.

jamessiddeley-amd (approved):
My blockers were addressed, LGTM. 

One nit/ comment though, the chart dropped a lot of data, and the changelog isnt updated for this.

46 of the 78 metrics in the gfx950 panel config no longer affect the output at all. Gone are all six latency figures, wave occupancy, wave life, coalescing rate, stall cycles, the whole instruction mix (VALU, SALU, VMEM, LDS, branch, matrix), register and scratch allocation, workgroup counts, and every request count at L1, L2, fabric, and HBM. I rendered the old chart from the base branch on the same workload to confirm it displayed basically all of them.

They're still reachable via --view table so its not unrecoverable, is this a design decision?

vedithal-amd (changes requested):
will review

projects/rocprofiler-compute/src/rocprof_compute_soc/analysis_configs/gfx908/0300_memory_chart.yaml
```diff
@@ -111,6 +134,12 @@ Panel Config:
           value: ROUND(SUM(TCC_EA0_WRREQ_sum) / SUM($denom), 0)
         Fabric_L2 Atomic:
           value: ROUND(SUM(TCC_EA0_ATOMIC_sum) / SUM($denom), 0)
+        L2-Fabric Read BW:
+          value: SUM(TCC_EA0_RDREQ_32B_sum * 32 + (TCC_EA0_RDREQ_sum - TCC_EA0_RDREQ_32B_sum) * 64) / (SUM(End_Timestamp - Start_Timestamp) / 1e9)
+          unit: Bytes/s
+        L2-Fabric Write and Atomic BW:
+          value: SUM(TCC_EA0_WRREQ_64B_sum * 64 + (TCC_EA0_WRREQ_sum - TCC_EA0_WRREQ_64B_sum) * 32) / (SUM(End_Timestamp - Start_Timestamp) / 1e9)
+          unit: Bytes/s
```
vedithal-amd, line 142: We should use noise clamping here - for all gfx arch

projects/rocprofiler-compute/src/rocprof_compute_soc/analysis_configs/gfx908/0300_memory_chart.yaml
```diff
@@ -121,6 +150,18 @@ Panel Config:
...
+          unit: Percent
+        HBM Write and Atomic Traffic:
+          value: ROUND(100 * SUM(TCC_EA0_WRREQ_DRAM_sum) / SUM(TCC_EA0_WRREQ_sum), 0)
+          unit: Percent
+        Remote Read Traffic:
+          value: ROUND(100 * SUM(TCC_EA0_RDREQ_sum - TCC_EA0_RDREQ_DRAM_sum) / SUM(TCC_EA0_RDREQ_sum), 0)
+          unit: Percent
+        Remote Write and Atomic Traffic:
+          value: ROUND(100 * SUM(TCC_EA0_WRREQ_sum - TCC_EA0_WRREQ_DRAM_sum) / SUM(TCC_EA0_WRREQ_sum), 0)
+          unit: Percent
```
vedithal-amd, line 164: we should use noise clamping here - for all gfx arch

projects/rocprofiler-compute/src/rocprof_compute_soc/analysis_configs/gfx908/0300_memory_chart.yaml
```diff
@@ -47,6 +43,18 @@ Panel Config:
           value: ROUND(AVG(SPI_CSN_WAVE), 0)
         Workgroups:
           value: ROUND(AVG(SPI_CSN_NUM_THREADGROUPS), 0)
+        Flat Read:
```
vedithal-amd, line 46: - When changing metrics yaml - please always run the full sync pipeline in tools config management.

- We need to sync panel yaml -> per arch yaml -> docs yaml also re-generate hash map and validate against template.

- Ensure that rst metric descriptions for new metrics have been added in per arch yaml and they are the same as plain metric description - use rst spefici hyperlink refernces and formatting as required.

- Finally ensure the pre commit is passing since that does yaml check.

projects/rocprofiler-compute/src/utils/mem_chart_gfx11.py
```diff
@@ -1,45 +1,43 @@
...
-(If a future gfx target adds ``0300_memory_chart.yaml``, keep keys aligned there.)
-
-Bandwidth values are **Bytes/s**, matching the YAML ``unit: Bytes/s`` rows.
-
-RDNA3.5 MEMORY HIERARCHY (GCEA = Graphics Core Efficiency Arbiter):
-   Kernel -> GL0 (TCP Cache)      -> GL1 Cache -> GL2 Cache -> GCEA -> System Memory
-         -> SQC (ICache/DCache)   -> GL1 Cache -> GL2 Cache -> GCEA -> System Memory
-         -> LDS (Local Data Share) [stays on CU, no GL1 Cache connection]
+Metric keys must match ``gfx115x/0300_memory_chart.yaml``.
+BW values are Bytes/s.
```
vedithal-amd: Drop this line - this is not the responsibility of this file

projects/rocprofiler-compute/src/utils/mem_chart_common.py
```diff
@@ -0,0 +1,395 @@
+# Copyright (c) Advanced Micro Devices, Inc.
+# SPDX-License-Identifier:  MIT
+
+"""Shared helpers for memory chart renderers (gfx9, gfx11)."""
```
vedithal-amd, line 4: There are no unit tests for this file - please add them

projects/rocprofiler-compute/tests/unit/utils/test_mem_chart_gfx11.py
```diff
@@ -1,20 +1,16 @@
 # Copyright (c) Advanced Micro Devices, Inc.
 # SPDX-License-Identifier:  MIT
 
-"""
-Unit tests for the GFX11 memory-chart renderer.
-
-Covers:
-- mem_chart_gfx11.py - RDNA3.5 (Rich-based) memory architecture visualization
-"""
+"""Unit tests for the gfx11 memory-chart renderer and shared helpers."""
```
vedithal-amd: mem_chart_common.py should not be unit tested in this file - it deserves its own

projects/rocprofiler-compute/tests/unit/utils/test_mem_chart_gfx9.py
```diff
@@ -13,6 +12,7 @@
 import yaml
 
 from utils import mem_chart_gfx9
+from utils.mem_chart_common import strip_ansi
```
vedithal-amd, line 15: mem_chart_common.py should not be unit tested in this file - it deserves its own
that however does not mean we cannot import mem_chart_common.py here - but testing should be done in its own unit file

xuchen-amd:
> My blockers were addressed, LGTM.
> 
> One nit/ comment though, the chart dropped a lot of data, and the changelog isnt updated for this.
> 
> 46 of the 78 metrics in the gfx950 panel config no longer affect the output at all. Gone are all six latency figures, wave occupancy, wave life, coalescing rate, stall cycles, the whole instruction mix (VALU, SALU, VMEM, LDS, branch, matrix), register and scratch allocation, workgroup counts, and every request count at L1, L2, fabric, and HBM. I rendered the old chart from the base branch on the same workload to confirm it displayed basically all of them.
> 
> They're still reachable via --view table so its not unrecoverable, is this a design decision?

these metrics are used in `--gui` and will discuss plan to remove them.

projects/rocprofiler-compute/tests/unit/utils/test_mem_chart_gfx11.py
vedithal-amd: that however does not mean we cannot import mem_chart_common.py here - but testing should be done in its own unit file

projects/rocprofiler-compute/tests/common.py
```diff
@@ -19,6 +19,7 @@
 
 # Imported after sys.path is extended, since it lives under src/.
 from utils import csv_compression  # noqa: E402
+from utils.mem_chart_common import strip_ansi as strip_ansi  # noqa: E402
```
vedithal-amd: strip_ansi is not being used here - lets drop it

Note `common.strip_ansi` is still consumed by `test_num_xcds.py` and five call sites in `test_mem_chart_gfx11.py` - please repoint those at `utils.mem_chart_common` directly rather than just deleting the re-export.

projects/rocprofiler-compute/CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ Full documentation for ROCm Compute Profiler is available at [https://rocm.docs.
   * Each edge now reports the traffic measured at the interface it represents.
   * Updated arrows, labels, and the legend in the memory chart to better represent their meaning.
 
+* Redesigned the CDNA (gfx9) Memory Chart with a new Rich-based layout that improves readability in the terminal. Added Flat/Buffer request breakdowns (Read/Write/Atomic wavefronts) and L2-Fabric bandwidth metrics across all CDNA architectures.
```
vedithal-amd: I dont think we are showing Flat metrics in the memory chart - at least i dont see that.
If that is the case, can we correct this CHANGELOG line

projects/rocprofiler-compute/LICENSE.md
```diff
@@ -32,8 +32,8 @@ Nuitka specific runtime code: Apache 2.0 license
 numpy python library: BSD
 pandas python library: BSD
 plotext python library: MIT
-plotille python library: MIT
 pyyaml python library: MIT
+rich python library: MIT
```
vedithal-amd: we get `rich` module transitively from `textual` since latter pulls in the former - for that reason we dont need to mention `rich` explicitly here

vedithal-amd (changes requested):

projects/rocprofiler-compute/src/utils/mem_chart_gfx9.py
```diff
@@ -1,1270 +1,699 @@
...
+def _build_l1_stack(metrics: dict[str, Any]) -> Group:
+    """Build vertically stacked L1 cache panels: VL1D, LDS, sL1D, L1I."""
+
+    vl1_panel = build_cache_panel(
+        "VL1D",
+        [("Hit", metrics["vl1_hit"], "%", COLORS["hit"])],
+        width=_IP_BLOCK_W,
+        height=_VL1D_H,
+    )
+    lds_panel = build_ip_block("LDS", _IP_BLOCK_W, _LDS_H)
```
vedithal-amd: The LDS panel is built with no content. `LDS Util` is defined in every gfx9 `0300_memory_chart.yaml` but `_extract_metrics()` never reads it. Please surface it here the way `_build_kernel_and_l0()` does in `mem_chart_gfx11.py`.

projects/rocprofiler-compute/src/utils/mem_chart_gfx9.py
```diff
@@ -1,1270 +1,699 @@
...
+    show_debug: bool = False,
+    chart_title: str = "",
+    gpu_arch: Optional[str] = None,
+) -> None:
+    """Create the CDNA memory diagram matching the reference PNG layout."""
+    metrics = _extract_metrics(metric_dict)
+    kernel_arrows = make_arrows(_KERNEL_ARROW_LEN)
+    std_arrows = make_arrows(_STD_ARROW_LEN)
+    is_gfx950 = gpu_arch is not None and gpu_arch.startswith("gfx950")
+    has_mall = gpu_arch is not None and gpu_arch >= "gfx940"
```
vedithal-amd: `gpu_arch >= "gfx940"` orders arch names lexicographically; nothing else in the tree compares archs that way, and the `is_gfx950` line right above it uses prefix matching. Precedent for capability gating is explicit membership: `is_gfx9()` / `is_gfx115x()` in `utils_common.py`, and `get_matrix_ops_type()` in `utils_analysis.py`.

projects/rocprofiler-compute/tests/unit/utils/test_mem_chart_gfx9.py
```diff
@@ -38,88 +38,71 @@
...
-    "Fabric Wr Lat": 53,
-    "Fabric Atomic Lat": 54,
-    "HBM Rd": 55,
-    "HBM Wr": 56,
+    k: v if v is not None else 1
+    for k, v in mem_chart_gfx9.DEFAULT_SAMPLE_METRICS.items()
 }
 
 
 def render_gfx9_chart(metrics, chart_title=DEFAULT_TITLE):
```
vedithal-amd: This never passes `gpu_arch`, so every test in the file runs the `gpu_arch=None` path. `_build_hbm_content()`, `_build_io_row()` (xGMI/PCIe), and the MALL panel have no coverage.

projects/rocprofiler-compute/tests/unit/utils/test_mem_chart_gfx9.py
```diff
@@ -130,605 +113,124 @@ def panel_yaml_metric_keys(architecture: str) -> frozenset[str]:
...
-
         assert reference_keys - panel_keys == (
             expected_architecture_missing_metric_keys(architecture)
         )
-        assert panel_keys - reference_keys == frozenset()
+        assert panel_keys - reference_keys == (
+            expected_architecture_extra_metric_keys(architecture)
+        )
 
     def test_panel_yaml_metrics_render_same_line_count_across_architectures(self):
```
vedithal-amd: All six archs render down the same `gpu_arch=None` path here, so the equal-line-count assertion cannot fail on an arch-specific layout regression. Parametrize on arch.

projects/rocprofiler-compute/src/utils/mem_chart_gfx11.py
```diff
@@ -637,45 +458,28 @@ def create_mem_chart_diagram(
     console: Console,
     show_debug: bool = False,
     chart_title: str = "",
+    gpu_arch: Optional[str] = None,  # noqa: ARG001
```
vedithal-amd: `# noqa: ARG001` is a no-op; `ARG` isn't in ruff's `select` in `pyproject.toml`. Drop it.

vedithal-amd (changes requested):
The gfx9 LDS panel renders empty even though `LDS Util` is defined in every gfx9 panel YAML, and the MALL gate orders arch strings lexicographically instead of using the explicit-membership pattern the rest of the tree uses.

The gfx9 tests never pass `gpu_arch`, so the gfx950 HBM/xGMI/PCIe panels and MALL - the headline of this PR - have no coverage at all, and the cross-arch line-count test cannot fail as written.

projects/rocprofiler-compute/src/rocprof_compute_soc/analysis_configs/gfx908/0300_memory_chart.yaml
xuchen-amd: applied for all gfx arch in: 145db3a

projects/rocprofiler-compute/src/rocprof_compute_soc/analysis_configs/gfx908/0300_memory_chart.yaml
xuchen-amd: applied for all gfx arch in: 145db3a

projects/rocprofiler-compute/src/rocprof_compute_soc/analysis_configs/gfx908/0300_memory_chart.yaml
xuchen-amd: ran in: 145db3a

projects/rocprofiler-compute/src/utils/mem_chart_gfx11.py
xuchen-amd: updated in: 0cdc451

projects/rocprofiler-compute/src/utils/mem_chart_common.py
xuchen-amd: added in: 60cf27c

projects/rocprofiler-compute/tests/unit/utils/test_mem_chart_gfx11.py
xuchen-amd: created its own test file: 60cf27c

projects/rocprofiler-compute/tests/unit/utils/test_mem_chart_gfx9.py
xuchen-amd: created its own test file: 60cf27c

projects/rocprofiler-compute/tests/common.py
xuchen-amd: fixed in: 60cf27c

projects/rocprofiler-compute/CHANGELOG.md
xuchen-amd: good catch, fixed in: 0cdc451

projects/rocprofiler-compute/LICENSE.md
xuchen-amd: fixed in: 0cdc451

projects/rocprofiler-compute/src/utils/mem_chart_gfx9.py
xuchen-amd: added in: 0cdc451

projects/rocprofiler-compute/src/utils/mem_chart_gfx9.py
xuchen-amd: fixed in: 0cdc451

projects/rocprofiler-compute/tests/unit/utils/test_mem_chart_gfx9.py
xuchen-amd: fixed in: 60cf27c

projects/rocprofiler-compute/tests/unit/utils/test_mem_chart_gfx9.py
xuchen-amd: fixed in: 60cf27c

projects/rocprofiler-compute/src/utils/mem_chart_gfx11.py
xuchen-amd: dropped in: 0cdc451

vedithal-amd (approved):
LGTM. All five of my comments are addressed: `LDS Util` now renders in the gfx9 LDS panel, MALL gating uses explicit membership via `_MALL_ARCHS`, the gfx9 tests pass `gpu_arch` and pin the gfx908/gfx940/gfx950 branches, the line-count test excludes gfx950 correctly, and the dead `noqa: ARG001` is gone.

Verified locally on `0cdc451`: 1293 unit tests pass, ruff check and format clean, and the config hash DB is up to date.

PR #8649, opened 2026-07-15
[hotswap] Add env var to force entry trampolines for all agents

lmoriche:
Introduce AMD_COMGR_HOTSWAP_FORCE_ENTRY_TRAMPOLINES: when enabled, the loader requests the entry-trampoline rewrite for every agent whose loaded code object's ISA matches, instead of only gfx12.5. This replaces any per-target hardcoding with a single, general override useful for bring-up and validation on any target COMGR supports (currently gfx1250, gfx1030, gfx1200).

The override is best-effort: the decision reuses the source ISA as the target and leaves rewrite_required false, so COMGR gates actual per-target support and a rewrite/rewritten-load failure falls back to the original code object. A target COMGR cannot handle simply loads unrewritten.

- hotswap.hpp: add RewriteOptions::force_all_entry_trampolines.
- hotswap.cpp: read AMD_COMGR_HOTSWAP_FORCE_ENTRY_TRAMPOLINES into the option; admit any non-empty gfx target in HasCandidateHotswapRewrite(); and OR a force-all clause into DecideHotswapRewrite()'s request_entry_trampolines. The force-all path still requires the code object's gfx target to equal the agent's, preserving COMGR's source==target contract.

Behavior notes:
- Default (no env) is unchanged: nothing is forced; gfx12.5 keeps its existing AMD_COMGR_HOTSWAP_ENTRY_TRAMPOLINES opt-in.
- HSA_HOTSWAP_DISABLE still disables the rewrite globally (checked first).

Tests (hotswap_rewrite_test.cc): gfx1030/gfx1200 are not rewritten by default; force-all routes gfx1030, gfx1200, and (best-effort) an unsupported target; and force-all still requires the code object ISA to match the agent.

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
@@ -108,6 +108,12 @@ bool AreEntryTrampolinesRequested() {
   return IsEnvFlagEnabled(kEnvName);
 }
 
+// Opt-in override to force the entry-trampoline rewrite on every agent (not
+// only gfx12.5). Best-effort: COMGR gates actual support per target.
+bool AreEntryTrampolinesForcedForAll() {
+  return IsEnvFlagEnabled("AMD_COMGR_HOTSWAP_FORCE_ENTRY_TRAMPOLINES");
+}
+
 bool IsStrictModeRequested() {
   return IsEnvFlagEnabled("HSA_HOTSWAP_STRICT_MODE");
 }
@@ -290,7 +296,8 @@ bool HasCandidateHotswapRewrite(const AgentGfxRevision& gfx,
                                 const RewriteOptions& options) {
   return IsHotswapSupportedGfxRevision(gfx) ||
       (options.strict_mode_enabled && gfx.gfx_target == kGfx1250) ||
-      (options.entry_trampolines_enabled && IsGfx12_5Target(gfx.gfx_target));
+      (options.entry_trampolines_enabled && IsGfx12_5Target(gfx.gfx_target)) ||
+      (options.force_all_entry_trampolines && !gfx.gfx_target.empty());
 }
 
 std::optional<RewriteDecision> DecideHotswapRewrite(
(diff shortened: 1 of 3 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -316,8 +323,14 @@ std::optional<RewriteDecision> DecideHotswapRewrite(
     return decision;
   }
 
-  const bool request_entry_trampolines = options.entry_trampolines_enabled &&
-      IsGfx12_5Target(gfx.gfx_target) && IsGfx12_5Target(source_gfx);
+  // Force-all: request entry trampolines for any agent whose loaded code
+  // object's ISA matches, independent of the gfx12.5 opt-in. Best-effort --
+  // COMGR gates actual per-target support and the load falls back on error.
+  const bool force_all_trampolines = options.force_all_entry_trampolines &&
+      !source_gfx.empty() && source_gfx == gfx.gfx_target;
```
cfreeamd, line 330: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** Force-all uses exact string equality here (`source_gfx == gfx.gfx_target`), whereas the gfx12.5 opt-in path uses the `IsGfx12_5Target()` helper. Exact match means a *generic* code object (e.g. `gfx10-3-generic`, `gfx12-generic`) loaded on a specific agent extracts a non-matching `source_gfx` and is skipped. Is excluding generic source ISAs intentional for force-all? For a best-effort debug knob this is defensible, but generic code objects are common in ROCr load paths, so worth confirming. Non-blocking.

cfreeamd (commented):

Reviewed the ROCr hotswap change. `core/inc/hotswap.hpp`, `core/runtime/hotswap.cpp`, and `hotswap_rewrite_test.cc` all look correct — the force-all path is best-effort (`rewrite_required = false`) so an unsupported target safely falls back to the original code object, the env-var name matches the existing `AMD_COMGR_HOTSWAP_ENTRY_TRAMPOLINES` sibling, and the new tests cover default-off, force-routing gfx10.3/gfx12.0, unsupported-target best-effort, and the matching-ISA requirement. One non-blocking question inline about generic source ISAs.

PR #8650, opened 2026-07-15
rocr: DRM user-queue fixes for UKI (user_queue=1, gfx1200)

cfreeamd:
## Summary

Five ROCr fixes for the DRM user-queue (UKI) path on gfx1200 with `amdgpu.user_queue=1`, taking a DRM AQL compute queue from crash/hang to executing dispatches.

- **libhsakmt,hsa-runtime: DRM userq fixes from Jesse Zhang** — export DRM-allocated BOs via `amdgpu_bo_export` for dma-buf (fixes `hsa_amd_vmem_map` `OUT_OF_RESOURCES` in the DRM path); force polling signals (`g_use_interrupt_wait=false`) in DRM mode until the KGD userq event framework lands.
- **fix `AqlQueue::Inactivate` to use the DRM queue ID** — it was destroying via the KFD `queue_id_` (which is 0 in DRM mode), leaving the kernel userq alive → `amdgpu_userq_evict` NULL deref when EOP/CWSR buffers were freed.
- **fix the constructor `QueueGuard`** — same KFD-vs-DRM queue-id bug in the error-cleanup scope guard.
- **create the DRM AQL queue with `AQL_COMPUTE`, not `SECURE`** — the queue was created with `AMDGPU_USERQ_CREATE_FLAGS_QUEUE_SECURE` (1<<2) but never `..._AQL_COMPUTE` (1<<3), so MES/CP configured it as non-AQL (PM4) + TMZ while ROCr wrote AQL packets → the CP never executed them → dispatch hang → MODE1 reset. **This is the core dispatch-hang fix.**
- **place the AQL queue descriptor+ring in device memory in DRM mode** — the descriptor (rptr/wptr) and ring defaulted to system memory (KFD VM), invisible to the DRM userq's CPF (page fault). Route them through the device-memory path when `enable_drm()` + LargeBAR. Eliminates the CPF descriptor fault. Workaround pending KFD/DRM VM unification (see the commit note — the descriptor still resolves to system memory per `umr`; the fault is gone but VRAM placement is unconfirmed).

## Test plan

gfx1200, `amdgpu.user_queue=1`, kernel including the AILIKGD-530 fix:

- [x] `rocrtst.Test_Example` (trivial dispatch): passes (previously hung).
- [x] `rocrtstFunc.VirtMemory_Access_Test` — CPU-to-GPU subtests pass (previously hung at first dispatch).
- Known residual, tracked separately in **AILIKGD-547**: the GPU-to-GPU subtest hits a CPF read permission fault (map/unmap eviction race on a dispatch-referenced buffer).

## Related

- AILIKGD-531 — DRM user-queue AQL dispatch hang (this PR)
- AILIKGD-530 — kernel `amdgpu_userq_evict_all` NULL deref (prerequisite)
- AILIKGD-547 — GPU-to-GPU CPF permission fault (residual, follow-up)
- AILIKGD-528 — original user-queue hang

### DIFF
--- a/projects/rocr-runtime/libhsakmt/src/fmm.c
+++ b/projects/rocr-runtime/libhsakmt/src/fmm.c
@@ -4892,10 +4892,16 @@ HSAKMT_STATUS hsakmt_fmm_export_dma_buf_fd(HsaKFDContext *ctx,
 		if (offset + MemorySizeInBytes <= obj->size) {
 			/* DRM-allocated BOs have no valid KFD handle for the KFD
 			 * export ioctl; export them through libdrm instead so the
-			 * unified interface (hsakmt_enable_drm) works.
+			 * unified interface (hsakmt_enable_drm) works. Guard on a
+			 * non-NULL DRM handle: some VRAM-domain BOs fall back to the
+			 * KFD alloc path (e.g. contiguous VRAM, which the kernel
+			 * rejects from userspace DRM BOs) and thus have no DRM handle;
+			 * those must use the KFD export ioctl below to avoid passing
+			 * NULL to amdgpu_bo_export().
 			 */
 			if (hsakmt_enable_drm &&
-			    is_supported_on_drm(obj->alloc_flags)) {
+			    is_supported_on_drm(obj->alloc_flags) &&
+			    obj->handles[0].drm) {
 				export_use_drm = true;
 				export_drm_bo = obj->handles[0].drm;
 			} else {
--- a/projects/rocr-runtime/libhsakmt/src/fmm.c
+++ b/projects/rocr-runtime/libhsakmt/src/fmm.c
@@ -4872,6 +4872,8 @@ HSAKMT_STATUS hsakmt_fmm_export_dma_buf_fd(HsaKFDContext *ctx,
 	HsaApertureInfo ApeInfo;
 	vm_object_t *obj;
 	HSAuint64 offset;
+	bool export_use_drm = false;
+	amdgpu_bo_handle export_drm_bo = NULL;
 	int r;
 	struct hsa_kfd_fmm_context *fmm_ctx = ctx->fmm_context;
 
@@ -4884,9 +4886,19 @@ HSAKMT_STATUS hsakmt_fmm_export_dma_buf_fd(HsaKFDContext *ctx,
 	if (obj) {
 		offset = VOID_PTRS_SUB(MemoryAddress, obj->start);
 		if (offset + MemorySizeInBytes <= obj->size) {
-			exportArgs.handle = obj->handles[0].kfd;
-			exportArgs.flags = O_CLOEXEC;
-			exportArgs.dmabuf_fd = 0;
+			/* DRM-allocated BOs have no valid KFD handle for the KFD
+			 * export ioctl; export them through libdrm instead so the
+			 * unified interface (hsakmt_enable_drm) works.
+			 */
+			if (hsakmt_enable_drm &&
+			    is_supported_on_drm(obj->alloc_flags)) {
+				export_use_drm = true;
+				export_drm_bo = obj->handles[0].drm;
+			} else {
+				exportArgs.handle = obj->handles[0].kfd;
+				exportArgs.flags = O_CLOEXEC;
+				exportArgs.dmabuf_fd = 0;
+			}
 		} else {
 			obj = NULL;
 		}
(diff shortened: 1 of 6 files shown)


cfreeamd (commented):

The flag fix in `amd_drm_driver.cpp` (SECURE→AQL_COMPUTE), the paired PM4 IB finegrain/system allocator paths in `amd_aql_queue.cpp`, and the DRM queue destroy in both the constructor guard and `Inactivate()` look correct. Interop type check update in `runtime.cpp` also looks correct. One Important inline comment on `fmm.c`.

projects/rocr-runtime/libhsakmt/src/fmm.c
```diff
@@ -4884,9 +4890,19 @@ HSAKMT_STATUS hsakmt_fmm_export_dma_buf_fd(HsaKFDContext *ctx,
...
 		if (offset + MemorySizeInBytes <= obj->size) {
-			exportArgs.handle = obj->handles[0].kfd;
-			exportArgs.flags = O_CLOEXEC;
-			exportArgs.dmabuf_fd = 0;
+			/* DRM-allocated BOs have no valid KFD handle for the KFD
+			 * export ioctl; export them through libdrm instead so the
+			 * unified interface (hsakmt_enable_drm) works.
+			 */
+			if (hsakmt_enable_drm &&
+			    is_supported_on_drm(obj->alloc_flags)) {
```

**[Important]** The allocation path in `fmm_allocate_memory_object` (earlier in this PR) gates the DRM path with `!alloc_flags.mflags.ui32.Contiguous`, but this export check only uses `is_supported_on_drm(obj->alloc_flags)` — missing the Contiguous guard. If `obj` was allocated with Contiguous=true (→ went through KFD, so `handles[0].drm` is NULL/unset), the code below would call `amdgpu_bo_export(NULL, amdgpu_bo_handle_type_dma_buf_fd, ...)` — undefined behavior / crash. The guard here should match the allocation check:
```c
if (hsakmt_enable_drm &&
    is_supported_on_drm(obj->alloc_flags) &&
    !obj->alloc_flags.mflags.ui32.Contiguous) {
```

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -2692,6 +2692,19 @@ hsa_status_t GpuAgent::QueueCreate(size_t size, hsa_queue_type32_t queue_type, u
...
   queues_[QueueUtility].touch();
 
+  // DRM/UKI mode: the DRM user queue's CPF runs in the DRM VM, but the queue
+  // descriptor (amd_queue_v2_t rptr/wptr) and AQL ring default to system memory,
+  // which lands in the KFD VM and page-faults the CPF. Route them through the
+  // device-memory path so they are DRM-allocated/mapped. Requires LargeBAR
+  // (CPU-accessible VRAM); gfx1200 has it.
+  // NOTE(UKI): umr shows the descriptor still resolves to system memory even with
+  // these flags set — the CPF descriptor fault is gone, but the intended VRAM
+  // placement is unconfirmed. Workaround pending KFD/DRM VM unification.
```

**[Question]** The `NOTE(UKI)` comment above says the descriptor still resolves to system memory despite the `DEVICE_MEM_QUEUE_DESCRIPTOR` flag being set, and mentions "Workaround pending KFD/DRM VM unification." Is this tracked in a Jira/OnTrack ticket? If the expected VRAM placement is unconfirmed, a ticket ensures this doesn't get lost when the descriptor memory issue is eventually diagnosed.

projects/rocr-runtime/libhsakmt/src/fmm.c
cfreeamd: Good catch — fixed in `656cf3d062`. I guarded the DRM export branch on a non-NULL DRM handle (`obj->handles[0].drm`) rather than `!Contiguous`: it directly prevents passing NULL to `amdgpu_bo_export()` and is robust to any allocation path that yields a KFD-backed BO, not just the contiguous case.

One clarification on reachability: a contiguous BO can't actually reach `amdgpu_bo_export(NULL)` today, because KFD-allocated objects are created with `alloc_flags.domain = ALLOC_DOMAIN_INVALID` (`vm_create_and_init_object`) — only the DRM alloc path sets `domain = VRAM`. So `is_supported_on_drm()` already returns false for a KFD contiguous BO and it takes the KFD export path. The guard makes that safety explicit rather than relying on the implicit domain-init invariant.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
cfreeamd: Filed **AILIKGD-549** and referenced it in the comment (`656cf3d062`). The device-mem flag does route the descriptor through `finegrain_allocator()` with `preferred_heap = VRAM` and it's correctly DRM-mapped into the userq VM (CPF descriptor fault is gone), but `umr` shows the physical backing is still system/GTT (`S=1`) — `preferred_heap` is only a hint and the kernel satisfies these small `CPU_ACCESS_REQUIRED` BOs from GTT. Functionally fine (GTT UC-coherent is a valid home for a descriptor), but the intended VRAM placement is unconfirmed, so it's tracked there pending the broader KFD/DRM VM work.

PR #8651, opened 2026-07-15
fix(rccl): [AICOMRCCL-1539] Fix gather/scatter P2P channel mismatch

prasanna-amd:
## Summary

- Fixes Gather.OutOfPlace deadlock on gfx1100 (and potentially MI355X) caused by PR #8349's asymmetric P2P channel detection
- Root cause: `planTotalTasks` differs per rank for gather/scatter — root has both send+recv tasks (`asymmetric=false`, uses `nChannelsMin`), non-root has only send tasks (`asymmetric=true`, uses `nChannelsMax`), causing a sender/receiver channel count mismatch that deadlocks
- Fix: replace per-rank `planTotalTasks` check with per-collective `collAPI` check (`ncclFuncGather`/`ncclFuncScatter`), which evaluates identically on all ranks

## Root Cause Detail

PR #8349 added this asymmetric detection in `addP2pToPlan()`:
```c
bool asymmetric = planTotalTasks[0] == 0 || planTotalTasks[1] == 0;
int nChStart = (comm->nNodes <= 1 && asymmetric) ? nChannelsMax : nChannelsMin;
```

For gather with root=1, 2 ranks on gfx1100 (`nChannelsMax=2, nChannelsMin=1`):
- **Rank 0** (non-root): `planTotalTasks=[0,1]` → `asymmetric=true` → `nChStart=2` → splits send across 2 channels
- **Rank 1** (root): `planTotalTasks=[2,1]` → `asymmetric=false` → `nChStart=1` → listens on 1 channel

Data sent on channel 1 is never consumed → both ranks wait forever → deadlock.

## Fix

Replace the per-rank task count check with a per-collective type check:
```c
ncclFunc_t collAPI = p2pTasks[dir] ? p2pTasks[dir]->collAPI : (ncclFunc_t)0;
bool asymmetric = (collAPI == ncclFuncGather || collAPI == ncclFuncScatter);
int nChStart = (comm->nNodes <= 1 && asymmetric) ? nChannelsMax : nChannelsMin;
```

Each P2P task already carries the originating collective type via `collAPI`, which is a property of the collective (not the rank), ensuring both sides of each P2P pair agree on channel count.

## Test plan

- [x] Gather.OutOfPlace passes on gfx1100 2-GPU (previously deadlocked 5/5 runs)
- [x] All 5 P2pChannelScaling tests pass (GatherLargeMessage, ScatterLargeMessage, AllToAllLargeMessage, AllToAllVLargeMessage, GatherNonZeroRoot)
- [x] All Scatter tests pass (OutOfPlace, OutOfPlaceGraph, InPlace, InPlaceGraph, InPlaceRoot0, InPlaceRoot0Graph)
- [ ] CI pipeline

### DIFF
--- a/projects/rccl/src/enqueue.cc
+++ b/projects/rccl/src/enqueue.cc
@@ -1168,10 +1168,10 @@ static ncclResult_t addP2pToPlan(
     else {
       ssize_t minPartSize = comm->nNodes > 1 ? stepSize[dir]/2 : stepSize[dir]/8;
       ssize_t maxPartSize = comm->nNodes > 1 ? stepSize[dir]   : stepSize[dir]*32;
-      // Single-node asymmetric patterns (gather/scatter): use nChannelsMax to
-      // fully utilize XGMI bandwidth when only one rank is the traffic hub.
-      // Symmetric patterns (alltoall): stay at nChannelsMin to avoid contention.
-      bool asymmetric = planTotalTasks[0] == 0 || planTotalTasks[1] == 0;
+      // Use collAPI (collective type) rather than per-rank task counts so both
+      // sides of a P2P pair agree on nChStart — planTotalTasks varies per rank.
+      bool asymmetric = p2pTasks[dir] &&
+        (p2pTasks[dir]->collAPI == ncclFuncGather || p2pTasks[dir]->collAPI == ncclFuncScatter);
       int nChStart = (comm->nNodes <= 1 && asymmetric) ? nChannelsMax : nChannelsMin;
       nChannels[dir] = std::min<int>(nChStart, divUp(bytes[dir], minPartSize));
       size_t partSize = std::max(minPartSize, divUp(bytes[dir], nChannels[dir]));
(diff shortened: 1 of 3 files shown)


projects/rccl/test/gather_channel_test.cpp
```diff
@@ -0,0 +1,48 @@
+/*************************************************************************
+ * Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * See LICENSE.txt for license information
+ ************************************************************************/
+
```
nileshnegi: why not add these cases to the existing Gather/Scatter tests?
https://github.com/ROCm/rocm-systems/blob/users/prasanna-amd/fix-rccl-gather-deadlock/projects/rccl/test/GatherTests.cpp
https://github.com/ROCm/rocm-systems/blob/users/prasanna-amd/fix-rccl-gather-deadlock/projects/rccl/test/ScatterTests.cpp

prasanna-amd: Good point — moving them to GatherTests.cpp and ScatterTests.cpp. The separate file was only to satisfy the therock-pr-bot unit test naming policy (`test_*` / `*_test.*`), since the existing test files don't match that pattern. Will update.

prasanna-amd: Done — regression tests moved to GatherTests.cpp and ScatterTests.cpp, gather_channel_test.cpp deleted, CMakeLists.txt reverted. Force-pushed.

nileshnegi (approved):

mch (approved):

projects/rccl/src/enqueue.cc
```diff
@@ -1247,10 +1247,10 @@ static ncclResult_t addP2pToPlan(
     else {
       ssize_t minPartSize = comm->nNodes > 1 ? stepSize[dir]/2 : stepSize[dir]/8;
       ssize_t maxPartSize = comm->nNodes > 1 ? stepSize[dir]   : stepSize[dir]*32;
-      // Single-node asymmetric patterns (gather/scatter): use nChannelsMax to
-      // fully utilize XGMI bandwidth when only one rank is the traffic hub.
-      // Symmetric patterns (alltoall): stay at nChannelsMin to avoid contention.
-      bool asymmetric = planTotalTasks[0] == 0 || planTotalTasks[1] == 0;
+      // Use collAPI (collective type) rather than per-rank task counts so both
```
mch, line 1250: Do you happen to know what is different between this block and the one above it?

prasanna-amd: The two sites handle different transport protocols. The first block (around line 1168) runs for the default path — it picks the channel count before IPC is considered. The second block (around line 1250) runs when an IPC-capable protocol is selected instead (the `useIpc` branch). Both need to agree on the same `nChStart` logic so that sender and receiver get matching channel counts regardless of which transport path each side takes. The pre-existing code already had them as near-identical blocks; this change just replaces the asymmetry detection in both.

PR #8652, opened 2026-07-15
fix(rocjitsu): smem don't scale ioffset

### LINKED ISSUE #8539: rocjitsu (gfx1250): SIGSEGV in L2Cache::send_backing during multi-tensor (foreach) Adam/AdamW
## Summary

On an emulated **gfx1250 (MI450X)** device under **rocjitsu** (functional mode, via mirage), a torch training/inference step crashes with a native **SIGSEGV inside the emulated memory subsystem** when torch uses its default **multi-tensor (`foreach`) Adam/AdamW** optimizer.

Simpler work is unaffected: elementwise ops, GEMM, `scaled_dot_product_attention`, a full transformer-encoder-layer forward, backward, **SGD**, and **single-tensor `Adam(foreach=False)`** all run fine. Only the multi-tensor/`foreach` optimizer path triggers the crash, which is why every torch training/inference workload fails while trivial GPU smoke tests pass.

## Environment

- rocjitsu functional emulator launched via `mirage run` (mirage built from `emulation/mirage`).
- Emulated agent: MI450X (gfx1250); pure software emulation, **no physical GPU**.
- Container torch built for gfx1250: `torch.cuda.is_available() == True`, `torch.cuda.get_arch_list() == ['gfx1250']`.

## Steps to reproduce

1. Build mirage and create a gfx1250 profile:

```bash
# from ROCm/rocm-systems
cd emulation/mirage
./scripts/mirage-docker-build.sh
export MIRAGE_BIN="$PWD/build/manylinux/bin/mirage"

"$MIRAGE_BIN" profile create mi450x \
  --emulator rocjitsu --agent MI450X \
  --num-nodes 1 --gpus-per-node 1 --no-input
```

2. Save `repro.py`:

```python
import torch, torch.nn as nn

d = torch.device("cuda:0")
assert torch.cuda.is_available()
print("device:", torch.cuda.get_device_name(0), "arch:", torch.cuda.get_arch_list())

m = nn.Linear(64, 64).to(d)
# foreach=True (torch default) -> crashes;  foreach=False -> runs fine
opt = torch.optim.Adam(m.parameters(), lr=1e-3, foreach=True)
x = torch.randn(8, 64, device=d)

for i in range(5):
    opt.zero_grad()
    m(x).pow(2).mean().backward()
    opt.step()
    torch.cuda.synchronize()
    print(f"step {i} ok", flush=True)
print("DONE")
```

3. Run it under mirage, using a container image whose torch is built for gfx1250:

```bash
"$MIRAGE_BIN" run --in-process --profile mi450x \
  --image <ROCm image with a gfx1250 torch> \
  --mount "$PWD/repro.py:/repro.py:ro" \
  -- python3 /repro.py
```

**Observed:** prints `step 0 ok`, then **`Segmentation fault (core dumped)`** on the next step (exit 139).
**Expected:** the loop completes and prints `DONE` (it does with `foreach=False`).

## Native backtrace

`librocjitsu.so` (not stripped); the fault is on a simulation worker thread servicing a global vector load:

```
Thread received signal SIGSEGV, Segmentation fault.
#0  rocjitsu::amdgpu::L2Cache::send_backing(unsigned long, unsigned char*, unsigned int, simdojo::MessageOp, unsigned int) [clone .constprop.0]
#1  rocjitsu::amdgpu::L2Cache::ensure_line(unsigned long, unsigned int)
#2  rocjitsu::amdgpu::L2Cache::fetch_line(unsigned long, unsigned char*, unsigned int)
#3  rocjitsu::amdgpu::L1VectorCache::ensure_line(unsigned long, unsigned int)
#4  rocjitsu::amdgpu::L1VectorCache::read_bytes(...)
#5  rocjitsu::amdgpu::L1VectorCache::load(...)
#6  rocjitsu::amdgpu::GlobalMemPipeline::initiate_access(rocjitsu::Instruction&, rocjitsu::amdgpu::Wavefront&)
#7  rocjitsu::amdgpu::MemoryPipeline::issue(rocjitsu::Instruction*, rocjitsu::amdgpu::Wavefront&)
#8  rocjitsu::amdgpu::ComputeUnitCore::issue_instruction(rocjitsu::amdgpu::Wavefront*)
#9  simdojo::SimulationEngine::worker_loop(unsigned int)
#10 simdojo::SimulationEngine::run()
#11 rj_vm_run()
```

## Additional observations

- **`foreach=False` (or SGD) is a working workaround** — the same loop runs 200 steps cleanly.
- State-dependent: a single `Adam(foreach=True)` step in a fresh process usually succeeds; the crash appears reliably on the **second** step (or after other GPU work has accumulated), which points at a memory-safety issue (out-of-bounds / bad address) in the emulated global-load cache path rather than a specific unsupported instruction.
- Individual `torch._foreach_*` primitives (`add_`, `addcmul_`, `sqrt`, `lerp_`, `div_`, `addcdiv_`) succeed in isolation; the fault only appears once enough operations accumulate.

## Notes

gfx1250 simulation is listed as Experimental in the rocjitsu support matrix, so this may be a known-in-progress area — filing with a minimal reproducer in case it's useful.

amd-arosa:
## Motivation

was causing problems.

## Technical Details

gfx1250 smem offset should not.

so delete `off *= scale;`

## JIRA ID

resolves #8539

## Test Plan

ran tests

## Test Result

tests pass

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/tests/gfx1250_sim_test.cpp
+++ b/emulation/rocjitsu/tests/gfx1250_sim_test.cpp
@@ -3635,7 +3635,7 @@ TEST(Gfx1250SimulationTest, DispatchPreloadsKernargWhenDescriptorSizeIsUnknown) 
   EXPECT_EQ(sim.cu()->read_sgpr(sbase + 3), args[2]);
 }
 
-TEST(Gfx1250SimulationTest, SLoadB32ScalesImmediateOffset) {
+TEST(Gfx1250SimulationTest, SLoadB32DoesNotScaleImmediateOffset) {
   using namespace rocr::llvm::amdhsa;
 
   constexpr uint64_t kKernelAddr = 0x10000;
@@ -3643,7 +3643,8 @@ TEST(Gfx1250SimulationTest, SLoadB32ScalesImmediateOffset) {
   constexpr uint32_t kExpected = 0x12345678u;
 
   std::vector<uint32_t> code;
-  append_instruction(code, make_s_load_b32_scaled_imm(4, 0, 1));
+  // s_load_b32 s4, s[0:1], 0x4 scale_offset
+  append_instruction(code, make_s_load_b32_scaled_imm(4, 0, 4));
   append_instruction(code, S_WAIT_KMCNT_0_GFX12);
   append_instruction(code, S_ENDPGM_GFX12);
 
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/addr_calc.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/addr_calc.cpp
@@ -154,7 +154,6 @@ uint64_t smem_calculate_address(const SmemMachineInst &inst, amdgpu::Wavefront &
   uint64_t base = (static_cast<uint64_t>(cu.read_sgpr(sbase + 1)) << 32) | cu.read_sgpr(sbase);
   int64_t off = static_cast<int64_t>(static_cast<int32_t>(inst.ioffset << 8) >> 8);
   uint32_t scale = inst.scale_offset ? access_size_bytes : 1;
-  off *= scale;
   if (has_smem_offset(inst.soffset))
     off += static_cast<int64_t>(read_sreg_m0_operand(wf, inst.soffset)) * scale;
   uint64_t addr = base + off;


newling (commented):
LGTM.

emulation/rocjitsu/tests/gfx1250_sim_test.cpp
```diff
@@ -3635,15 +3635,16 @@ TEST(Gfx1250SimulationTest, DispatchPreloadsKernargWhenDescriptorSizeIsUnknown)
...
   using namespace rocr::llvm::amdhsa;
 
   constexpr uint64_t kKernelAddr = 0x10000;
   constexpr uint64_t kKernargAddr = 0x400000;
   constexpr uint32_t kExpected = 0x12345678u;
 
   std::vector<uint32_t> code;
-  append_instruction(code, make_s_load_b32_scaled_imm(4, 0, 1));
+  // s_load_b32 s4, s[0:1], 0x4 scale_offset
+  append_instruction(code, make_s_load_b32_scaled_imm(4, 0, 4));
```
kuhar, line 3647: We should consider adding a companion case with a real SGPR `soffset` and `scale_offset` set. The new test hard-codes `OPR_SMEM_OFFSET_NULL`, so it only covers the unscaled immediate and would still pass if the `soffset * access_size_bytes` branch at `addr_calc.cpp:158` stopped scaling the register offset. We can assert tht the effective address is `base + soffset * access_size_bytes`.

kuhar (approved):
Just one comment, the change itself LGTM

PR #8654, opened 2026-07-15
ci(rocjitsu): Run corpus test in parallel and rerun failures

RattataKing:
## Motivation

The test corpus repo now supports running tests in parallel, rocjitsu CI can set `-n` to pytest to reduce runtime.

## Technical Details

- Updated corpus ref
- Extracted test logic to a separated bash
- Add a rerun round with longer timeout value (15 -> 40s)
- Excluded cpp test building time from measurements
- Cleaned up the report messages
- Set `-j nproc/2` to cmake and ctest, reduced rocjitsu building time 5 -> 2.5 mins
- Enabled run pytest in parallel, corpus runtime reduced 17.5 -> 4.5 mins
- Total CI runtime: 24 -> 9 mins

## JIRA ID

ISSUE ID: https://github.com/ROCm/rocm-systems/issues/7891

## Test Plan

Run CI.

## Test Result

CI passed.

## Submission Checklist


emulation/rocjitsu/tests/corpus/run-corpus-tests.sh
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+: "${ROCM_PATH:?ROCM_PATH must be set}"
+: "${ROCJITSU_SOURCE_DIR:?ROCJITSU_SOURCE_DIR must be set}"
+
+WORKER_COUNT="${WORKER_COUNT:-8}"
+SOFT_TIMEOUT_SECONDS="${SOFT_TIMEOUT_SECONDS:-30}"
+HARD_TIMEOUT_SECONDS="${HARD_TIMEOUT_SECONDS:-60}"
+
```
amd-eochoalo, line 35: I would pass this as positional parameters instead of via the environment variables. I would also pass this the rocjitsu-test-corpus path as a parameter. It is implied from the environment that the script is to be used in that directory only.

amd-eochoalo (approved):
I'll approve since it will improve the CI experience a lot but consider making the recommended changes in this or a subsequent PR.

emulation/rocjitsu/tests/corpus/run-corpus-tests.sh
```diff
@@ -0,0 +1,63 @@
...
+    --durations=0
+    -vv
+    -o "cache_dir=${cache_dir}"
+    --tb=no
+  )
+
+  if "${pytest_cmd[@]}" -n "${WORKER_COUNT}" -o "timeout_func_only=true" --timeout "${SOFT_TIMEOUT_SECONDS}"; then
+    status_message="All (${name}) tests passed."
+  elif "${pytest_cmd[@]}" --last-failed --timeout "${HARD_TIMEOUT_SECONDS}"; then
+    status_message="Retried (${name}) tests passed."
```
lialan: I mean, should we even retry here ...  We should just fail and call it a day?

emulation/rocjitsu/tests/corpus/run-corpus-tests.sh
```diff
@@ -0,0 +1,63 @@
...
+    --artifact-directory "${artifact_dir}"
+    --durations=0
+    -vv
+    -o "cache_dir=${cache_dir}"
+    --tb=no
+  )
+
+  if "${pytest_cmd[@]}" -n "${WORKER_COUNT}" -o "timeout_func_only=true" --timeout "${SOFT_TIMEOUT_SECONDS}"; then
+    status_message="All (${name}) tests passed."
+  elif "${pytest_cmd[@]}" --last-failed --timeout "${HARD_TIMEOUT_SECONDS}"; then
```
kuhar: The retry (`elif`) branch changes more than just the timeout relative to the soft run: it also drops `-n "${WORKER_COUNT}"` (so the retry is serial) and drops `-o "timeout_func_only=true"` (so on retry the timeout now covers fixtures/setup, not just the test body). The serial retry is a reasonable way to shed contention-induced flakes, but the silent change in timeout scope means a test whose slowness is in a fixture is measured differently on the two passes. Worth making that intentional with a one-line comment, or keeping `timeout_func_only=true` on both so the only variable between passes is parallelism + timeout budget.

One edge to keep in mind: `--last-failed` reads the lastfailed cache from the first run. If the first (xdist) run exits non-zero without recording per-test failures (e.g. a worker crash), pytest's default `--last-failed` behavior is to run *everything*, so the "retry" could quietly become a full serial re-run. Setting `--last-failed-no-failures none` would make that case a no-op instead.

.github/workflows/rocjitsu-corpus-tests.yml
```diff
@@ -182,11 +182,13 @@ jobs:
             -DRJ_ENABLE_TSAN=${{ matrix.enable_tsan }} \
             -DRJ_ENABLE_MSAN=${{ matrix.enable_msan }}
 
-          cmake --build "${ROCJITSU_BUILD_DIR}"
+          WORKERS="$(( ($(nproc) + 1) / 2 ))"
```
kuhar, line 186: This halves build and ctest parallelism (`(nproc+1)/2`) for every matrix entry, including the plain `release` build that has no sanitizer. If the motivation is avoiding OOM under ASAN/UBSAN, it would be worth either scoping the reduction to the sanitizer configs or leaving a short comment on why half the cores was chosen -- otherwise a future maintainer is likely to "optimize" this back to full parallelism and reintroduce whatever problem it was working around. Documenting the intent here keeps the trade-off legible.

emulation/rocjitsu/tests/corpus/run-corpus-tests.sh
kuhar: **Retry can silently re-run the entire suite and mask a crash.**

The retry relies on `--last-failed` reading the `lastfailed` cache written by the
first (parallel) pass. But `--last-failed` falls back to running *everything* when
the cache has no recorded failures. I verified this on pytest 9.1.1:

- empty `cache_dir` + `--last-failed` => all 3 tests ran (`3 passed`), not zero.

That fallback is exactly what happens when the first pass fails *without*
populating `lastfailed`: an xdist worker crash (segfault / OOM / sanitizer abort),
a collection error, or an internal error. In those cases the "retry" is not a
retry of the failed tests at all -- it re-runs the full corpus serially. If that
serial run happens to pass (because the crash was a one-off under `-n
${WORKER_COUNT}` contention), the target is reported as `Retried (${name}) tests
passed.` and the job goes green, hiding a real crash.

Two consequences worth guarding against:
1. Distinguish "no failures recorded" from "some tests failed": check the
   `lastfailed` cache is non-empty before treating the elif as a targeted retry,
   or use `--last-failed --last-failed-no-failures none` so an empty cache runs
   nothing instead of everything.
2. A first-pass crash should not be convertible to green by a full serial rerun.

kuhar (commented):
Original anchor: `.github/workflows/rocjitsu-corpus-tests.yml:217`

The GPU target list is now duplicated: it lives here in the workflow and again as `default_targets` in `run-corpus-tests.sh` (lines 11-17). Whenever a target is added/removed (e.g. the extra configs like `gfx1151.json`, `gfx950_cdna4_kmd*.json`, `guest_*` that already exist under `configs/`), both lists must be kept in sync or they silently drift.

Since the script already carries a sensible `default_targets`, one clean option is to let the script own the list and invoke it here with no positional args (the `(( ${#targets[@]} )) || targets=(...)` fallback already handles that). That keeps a single source of truth and makes the script the obvious place to extend the target matrix later.

Nice foundational refactor -- pulling the corpus loop out of the workflow into `run-corpus-tests.sh` with defaulted targets, timeouts, and artifact/cache dirs is exactly the kind of change that pays off later. The header comment already notes this is a stopgap that will "likely be added to TheRock tests," and this shape makes that migration much easier: TheRock (or a developer running locally) can call the script directly instead of copy-pasting the workflow's inline loop.

Two forward-looking thoughts, neither blocking:

1. The script currently reads the corpus location implicitly via CWD + `ROCJITSU_SOURCE_DIR` env. This overlaps with the reviewer suggestion on threading the corpus path through explicitly -- accepting the corpus dir as an argument/env with a clear name would make the script self-contained and trivially reusable outside this one workflow (which is the direction TheRock integration will want anyway).

2. The retry + `.pytest-artifacts/` + lastfailed-summary plumbing is a good base for surfacing structured results. If you later emit a small machine-readable summary (per-target pass/retry/fail counts) it would compose well with a job-summary step or dashboards, without the retry logic having to change. Worth keeping the door open for as a follow-up rather than something to build now.

Happy to review the TheRock-migration follow-up when it lands.

emulation/rocjitsu/tests/corpus/run-corpus-tests.sh
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+: "${ROCM_PATH:?ROCM_PATH must be set}"
+: "${ROCJITSU_SOURCE_DIR:?ROCJITSU_SOURCE_DIR must be set}"
+
+WORKER_COUNT="${WORKER_COUNT:-8}"
```
kuhar: **`WORKER_COUNT` default of 8 is not aligned with the runner and is never set by
the workflow.**

The build step in `rocjitsu-corpus-tests.yml` deliberately sizes parallelism to
the host (`WORKERS="$(( ($(nproc) + 1) / 2 ))"` for `cmake --build` / `ctest`),
but the corpus step invokes this script without exporting `WORKER_COUNT`, so
pytest runs with the hardcoded default of 8 regardless of `nproc`.

On a runner with fewer than ~8 usable cores this oversubscribes, which directly
increases the odds of the SOFT-timeout (25s) tripping under contention -- feeding
the retry path and the flakiness concern raised elsewhere in this file. Recommend
deriving the default from `nproc` the same way the build step does (e.g.
`WORKER_COUNT="${WORKER_COUNT:-$(( ($(nproc) + 1) / 2 ))}"`), or have the workflow
pass an explicit `WORKER_COUNT` next to the timeouts it already exports.

emulation/rocjitsu/tests/corpus/run-corpus-tests.sh
```diff
@@ -0,0 +1,63 @@
...
+  pytest_cmd=(
+    rocjitsu --config "${rocjitsu_config_path}" -- pytest tests/test_corpus.py
+    --target "${name}"
+    --suite iree,kernels,cts
+    --skip-tests-config "${skip_tests_config_path}"
+    --artifact-directory "${artifact_dir}"
+    --durations=0
+    -vv
+    -o "cache_dir=${cache_dir}"
+    --tb=no
```
kuhar: **`--tb=no` plus discarded artifacts makes CI failures hard to diagnose.**

The previous invocation showed full output (`-vv`); this one adds `--tb=no`, so a
genuine failure produces no traceback at all. The only failure detail left is the
`--cache-show="cache/lastfailed"` dump, which prints failing test *names* but not
*why* they failed. For a workflow whose stated purpose is to let developers
"quickly check that their changes didn't break emulation," dropping tracebacks
removes the first thing a developer needs.

Compounding this: the run writes per-target output to `--artifact-directory
.pytest-artifacts/${name}`, but nothing uploads it. There is no
`actions/upload-artifact` step in `rocjitsu-corpus-tests.yml`, so those artifacts
die with the runner and are never accessible.

So on failure the developer gets neither a traceback nor the artifacts. Suggest
either `--tb=short` (or `--tb=line`) for a compact failure reason in-log, and/or an
`upload-artifact` step for `.pytest-artifacts/` gated on failure.

emulation/rocjitsu/tests/corpus/run-corpus-tests.sh
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env bash
```
kuhar, line 1: Please commit this as an executable file (mode 100755). Every other first-party shell script under `emulation/rocjitsu` is executable, and keeping this one at 100644 means the natural `./tests/corpus/run-corpus-tests.sh` invocation fails even though the file has a shebang.

emulation/rocjitsu/tests/corpus/run-corpus-tests.sh
```diff
@@ -0,0 +1,63 @@
...
+set -euo pipefail
```
kuhar, line 22: For consistency with the other first-party RocJITsu shell scripts, please add the standard 2026 AMD copyright and MIT SPDX header after the shebang; see `scripts/rocjitsu-docker-build.sh:1-4` for the local template. This is currently the only such script in this subtree without that header.

emulation/rocjitsu/tests/corpus/run-corpus-tests.sh
kuhar: Seconding the doubt on retrying, from a test-signal standpoint. The retry does not
re-run the failed tests under the same conditions -- it changes three variables at
once relative to the first pass:

- drops `-n ${WORKER_COUNT}` (so the retry runs serially, no parallel contention),
- drops `-o timeout_func_only=true` (so the timeout now also covers fixture setup,
  not just the test body),
- raises the timeout from SOFT (25s in CI) to HARD (40s).

So `Retried (...) tests passed.` is not evidence the test is healthy -- it only
tells us the test passes when run alone with a laxer timeout. A test that fails
under parallel load at 25s but passes solo at 40s is a genuine flaky/perf
regression, and this logic converts it to green with only a log-line difference.
That signal is exactly what this smoke workflow exists to surface.

If the goal is just to absorb timeouts caused by parallel oversubscription, I'd
make that explicit (retry only on timeout, keep `-n` and `timeout_func_only`
identical, only bump the timeout) so a retry can't hide a real failure. Otherwise
I agree with the doubt raised here: failing outright is more honest.

RattataKing: From the past experience time noisy did happen a few times, people end up clicking rerun button most of the time to get CI passed. I’ll make this an optional feature controlled by flags, but keep it enabled for now

emulation/rocjitsu/tests/corpus/run-corpus-tests.sh
RattataKing: Will change to flags

emulation/rocjitsu/tests/corpus/run-corpus-tests.sh
RattataKing: Changed to make CI fails as if the first run fails, but will still rerun to collect more info

.github/workflows/rocjitsu-corpus-tests.yml
```diff
@@ -212,30 +215,21 @@ jobs:
...
-            echo "::endgroup::"
-          done
-          exit "${corpus_status}"
+          export ROCJITSU_SOURCE_DIR
+
+          corpus_test_status=0
+
+          # Limit pytest to 8 workers to avoid oversubscription.
+          WORKERS="$(( ($(nproc) + 1) / 2 ))"
+          PYTEST_WORKERS="$(( WORKERS < 8 ? WORKERS : 8 ))"
```
lialan: Any reason we limit  it to only maximum 8?

lialan (approved):
LGTM, except the hard cap 8 workers. 

My opinion is that we shouldn't set a physical number here, it should scale with CI machine.

.github/workflows/rocjitsu-corpus-tests.yml
RattataKing: 8 was the sweet spot for perf vs. # of cores, I worried too many cores may oversubscribe the system and slow down pytest, however on this machine 8->32 cores looks like can still reduce runtime further and doesnt seem to cause extra overhead, I will remove the limit.

PR #8661, opened 2026-07-16
fix(rocprof-systems): Build rccl-tests natively and fix alltoall comm tracing

sputhala-amd:
## Motivation

On `gfx1151` the RCCL binary_rewrite tests failed because the in-tree `rccl-tests` example was silently skipped (its arch wasn't in a hard-coded list), so the test harness fell back to a multi-arch, rocm_kpack-processed RCCL binary from PATH that Dyninst cannot rewrite (the instrumented binary segfaulted in the loader). 
Separately, the RCCL sys_run validation failed for alltoall because rocprof-sys recorded no "RCCL Comm" data for that collective. This PR makes the rccl-tests build first-class and self-contained, and fixes the RCCL API tracing gap.

## Technical Details

- Replaced the wrapper that shelled out to the vendored rccl-tests Makefiles with a native CMake build. The build now targets the detected arch, so the fat binary stays small and Dyninst-rewritable — and the harness resolves the in-tree binary instead of the kpack'd PATH copy.
- Trimmed the built collectives to the minimal set covering every distinct RCCL API entry point. ; dropped scatter/gather/sendrecv (redundant ncclSend/ncclRecv coverage) and removed their now-unused sources. Updated test_rccl.py targets and the example README.
- **Fix the handling of ncclAllToAll** -  RCCL's deprecated `ncclAllToAll` forwards to `ncclAlltoAll` (Note the different case of letter T), the SDK reported the op as `ncclAlltoAll`, which fell through to default and recorded 0 bytes, and RCCL Comm validation was failing in the traces. Added a `ncclAlltoAll` case (guarded by `#if RCCL_API_TRACE_VERSION_PATCH >= 3`) alongside the existing `ncclAllToAll`. 
- **Fix logging command output** - The binary_rewrite runner only reported the run-phase command, so the rocprof-sys-instrument invocation that produces the .inst was invisible under ctest -V

## JIRA ID

Jira ID: AIPROFSYST-631

## Test Plan

 ctest -R rccl

## Test Result

All 30 rccl tests pass.

## Submission Checklist


sputhala-amd:
`gfx1151` workflows on this branch passes the rccl tests- https://github.com/ROCm/rocm-systems/actions/runs/29608075569/job/87976248213

projects/rocprofiler-systems/examples/rccl/CMakeLists.txt
```diff
@@ -12,164 +12,246 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
...
-    foreach(arch IN LISTS ROCPROFSYS_GFX_TARGETS)
-        if(arch IN_LIST RCCL_SUPPORTED_TARGETS)
-            list(APPEND RCCL_GPU_TARGETS ${arch})
-        endif()
-    endforeach()
-    if(RCCL_GPU_TARGETS STREQUAL "")
-        message(AUTHOR_WARNING "rccl-tests skipped. No supported GPU targets found.")
-        return()
-    endif()
+if(
```
dgaliffiAMD: This block is unreachable. The early `return()` at _lines 68-71_ guarantees `HIPCC_EXECUTABLE` is non-empty, so `NOT HIPCC_EXECUTABLE` in the second AND-group is always `false` → the whole condition is always `false` and the warning/`return()` can never fire. (`NOT hip_FOUND` on _line 85_ is also always `false`, guaranteed by the `return()` at _line 24_).

projects/rocprofiler-systems/examples/rccl/README.md
dgaliffiAMD:
```suggestion
This example builds the rccl-tests performance test suite, which benchmarks RCCL (ROCm Communication Collectives Library) operations across multiple GPUs. It includes performance tests for all major collectives - AllGather, AllReduce, AllToAll, Broadcast, Reduce, and ReduceScatter - measuring bandwidth and latency for inter-GPU communication. This is useful for profiling GPU-to-GPU collective communication patterns and PCIe/Infinity Fabric interconnect performance.
```
nit-pick: This line still includes mentions of the tests that you removed.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/rocprofiler-sdk/rccl_internal.hpp
```diff
@@ -3,6 +3,9 @@
 
 #pragma once
 
+// Defines RCCL_API_TRACE_VERSION_PATCH, which gates the newer RCCL argument
+#include <rocprofiler-sdk/rccl/details/api_trace.h>
```
dgaliffiAMD, line 14: If I am reading this correctly, `api_trace.h` needs to be included before `api_trace.h`, because it defines `RCCL_API_TRACE_VERSION_PATCH`? 
I would call that our more clearly in the above comment, so it is not accidentally move (for example, to the `cpp` file) or removed.

projects/rocprofiler-systems/examples/rccl/CMakeLists.txt
```diff
@@ -12,164 +12,246 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
...
 
-        # Create imported executable target
-        add_executable(rccl-tests::${_EXE_NAME} IMPORTED GLOBAL)
-        set_property(
-            TARGET rccl-tests::${_EXE_NAME}
-            PROPERTY IMPORTED_LOCATION ${_EXE_DEST_PATH}
+    if(
+        CMAKE_CXX_COMPILER_ID MATCHES "Clang"
+        AND NOT CMAKE_CXX_COMPILER_IS_HIPCC
+        AND NOT HIPCC_EXECUTABLE
```
dgaliffiAMD: At this point, `HIPCC_EXECUTABLE` should always be true (thanks to the return on _line 68_). Meaning this block is now "unreachable".

projects/rocprofiler-systems/source/lib/rocprof-sys/library/rocprofiler-sdk/rccl.cpp
```diff
@@ -127,6 +127,14 @@ rccl_get_event_info_impl(
             set_event(info, false, payload.args.ncclAllToAll.count,
                       payload.args.ncclAllToAll.datatype, payload.args.ncclAllToAll.comm);
             break;
+#if RCCL_API_TRACE_VERSION_PATCH >= 3
+        // ncclAllToAll was renamed to ncclAlltoAll; the deprecated entry point
+        // forwards to this one, so the SDK reports collectives under this id.
```
dgaliffiAMD: Can we make the comment clearer that `T` was replaced with `t`? I thought you had just written `ncclAllToAll` twice until I eventually saw the difference in cases. Maybe it's time to increase my font size. 😆

projects/rocprofiler-systems/source/lib/rocprof-sys/library/rocprofiler-sdk/rccl_internal.hpp
dgaliffiAMD: It is a shame that this has to be so brittle. 
Could we add something like this in the `cpp` file, or would that break legacy ROCm support. 

```
  #ifndef RCCL_API_TRACE_VERSION_PATCH
  #    error "RCCL_API_TRACE_VERSION_PATCH is not defined here — <rocprofiler-sdk/rccl/details/api_trace.h> must be included before api_args.h / api_id.h (see rccl.hpp / rccl_internal.hpp)"
  #endif
```

dgaliffiAMD (approved):
Thank you. I just have a couple of nit-picks. It looks good overall.

projects/rocprofiler-systems/examples/rccl/CMakeLists.txt
```diff
@@ -12,164 +12,246 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
...
-                COMPONENT rocprofiler-systems-examples
-            )
-        endif()
-    endforeach()
+# ----------------------------------------------------------------------------
+# One executable per collective, linking the shared object library.
+#
+foreach(_test IN LISTS _rccl_tests)
+    set(_target "${_test}_perf")
+    add_executable(${_target} ${_rccl_hipify_dir}/${_test}.cpp)
```
kcossett-amd, line 219: Now that these executables are placed under `build-dir` as opposed to `build-dir/examples/rccl`, some modifications to pytest discover will have to be done. 

1. https://github.com/ROCm/rocm-systems/blob/develop/projects/rocprofiler-systems/tests/pytest/rocprofsys/config.py#L236 this is no longer needed
2. https://github.com/ROCm/rocm-systems/blob/develop/projects/rocprofiler-systems/tests/pytest/rocprofsys/config.py#L256 no longer needed
3. https://github.com/ROCm/rocm-systems/blob/develop/projects/rocprofiler-systems/tests/pytest/rocprofsys/config.py#L246 this was likely the cause of all your grief. Why did I put this here in the first place? Because I thought I'd use this for generic executables as well, but I never did. So remove this as well. 

That should be it (off the top of my head)...

kcossett-amd (commented):
Thank you for also displaying rewrite command during test. Completely forgot about that 😅

projects/rocprofiler-systems/source/lib/rocprof-sys/library/rocprofiler-sdk/rccl_internal.hpp
sputhala-amd: We need to add this to the hpp and the rccl_internal.hpp because this is needed also by the unit tests. Unfortunately, I could not come up with any better way to resolve the macro `RCCL_API_TRACE_VERSION_PATCH`.

projects/rocprofiler-systems/examples/rccl/CMakeLists.txt
sputhala-amd: @kcossett-amd - I prefer making these changes in a different PR for 2 reason, one to not expand the scope of this PR and two to ensure I don't break anything else with those changes. I hope that approach is fine with you. PR: https://github.com/ROCm/rocm-systems/pull/8921

PR #8664, opened 2026-07-16
feat(rocshmem/gda/ionic) Latency optimization for inline data on ionic

abouteiller:
single store to set the inline value

## Motivation

Using 32B inline write to WQE causes a latency spike for message sizes 8, 16, 32, most pronounced for 16 and 32. This is due to using multiple memory transactions to write to the wqe inline field. From analysis there appear to be very little benefit to using inlining for sizes>8B. 

## Technical Details

* Do not override the max inline
* Use a single 8B write to set the inline wqe field (single memory transaction). Overwrite for sizes 1-7 is harmless. 

## JIRA ID

JIRA ID AIROCSHMEM-453

## Test Plan

Performance evaluated on Pollara 400G

## Test Result

Latency curve flat for sizes 1-128

## Submission Checklist

### DIFF
--- a/projects/rocshmem/src/gda/ionic/queue_pair_ionic.cpp
+++ b/projects/rocshmem/src/gda/ionic/queue_pair_ionic.cpp
@@ -22,6 +22,8 @@
  * IN THE SOFTWARE.
  *****************************************************************************/
 
+#include <cassert>
+
 #include "gda/queue_pair.hpp"
 #include "gda/endian.hpp"
 #include "util.hpp"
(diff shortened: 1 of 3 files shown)


projects/rocshmem/src/gda/ionic/queue_pair_ionic.cpp
```diff
@@ -327,13 +329,18 @@ __device__ void QueuePair::ionic_post_wqe_rma(int32_t length,
...
       wqe->base.num_sge_key = 0;
       if (!laddr) {
         // TODO why is this needed?
         wqe->common.pld.data[0] = 1;
       } else {
-        memcpy(wqe->common.pld.data, reinterpret_cast<const void*>(laddr), length);
+        // Always read/write 8B to emit a single load+store instruction;
+        // overread is safe (heap page), NIC uses wqe length for actual size.
+        uint64_t val;
+        __builtin_memcpy(&val, reinterpret_cast<const void*>(laddr), sizeof(val));
```
omor1, line 342: Not sure whether this is possible, but suppose `laddr` points near the end of the page, e.g. it points to the last byte of a page. Wouldn't this result in errors at runtime?

I think something like the following is valid:
```C++
uint64_t* wqe_pld64 = reinterpret_cast<uint64_t*>(&wqe->common.pld.data);
if constexpr (__builtin_constant_p(__builtin_is_aligned(laddr, alignof(uint64_t))) &&
              __builtin_is_aligned(laddr, alignof(uint64_t))) {
  /* we know at compile time that laddr is aligned to (at least) 8 bytes */
  *wqe_pld64 = *reinterpret_cast<const uint64_t*>(laddr);
} else {
  /* we either don't know the alignment of laddr or its alignment is less than 8 bytes */
  uint64_t val = 0;
  if constexpr (__builtin_constant_p(length)) {
    /* but we do know the length at compile time, so can use __builtin_memcpy_inline */
    __builtin_memcpy_inline(&val, reinterpret_cast<const void*>(laddr), length);
  } else {
    __builtin_memcpy(&val, reinterpret_cast<const void*>(laddr), length);
  }
  *wqe_pld64 = val;
}
```
Depending on what the compiler does for `__builtin_memcpy_inline`, it may be extraneous to check for the alignment.

omor1 (commented):
Let's discuss the post-page-read issue when we meet later.

PR #8665, opened 2026-07-16
fix(rocshmem/api) LOGD in broadcast_wave use nelem, all shmem apis should use nelems

abouteiller:
## Motivation

* Code does not compile with BUILD_DEBUG_TRACE_DEVICE
* shmem convention is to use `nelems` for the buffer size, even for xxxmem procedures. We had an inconsistent mix and match.

## Technical Details

All code uses the same conventional `nelems`, not variants `nelements`, `nelem`, `size`, reducing inconsistency and chances to use the wrong spelling. 

## JIRA ID 

JIRA ID AIROCSHMEM-450

## Test Plan
compiles with debug logs enabled. 

## Submission Checklist


edgargabriel (approved):
LGTM

avinashkethineedi (approved):

PR #8672, opened 2026-07-16
fix(rocjitsu): Harden the rocjitsu KFD, CP, and multi-process runtime

### LINKED ISSUE #8818: [rocjitsu] Recycled doorbell can lose or falsely complete replacement queue work
# Recycled rocJITsu doorbells can lose or falsely complete the first dispatch

## Tracking

- Requested priority: P0. The defect can both hang callers and report work
  complete without executing it.
- Upstream tracker: <https://github.com/ROCm/rocm-systems/issues/8818>.
- Complete source and one-command reproducer:
  <https://gist.github.com/benvanik/3850371ed5eec87363117a1e8cbf3dce>

## Environment

- ROCm SDK: `7.15.0a20260718`.
- rocJITsu source baseline: PR #8815 head
  `03ba8c6b8632ed34ffd800650d3747ab9f73b941`, merged to `develop` as
  `3caf8d08053e0bf4acdc6bd4aa03a3430843802d`.
- Current `develop` when this report was prepared: the same merge commit. The
  queue-creation, queue-destruction, and command-processor sources are byte-for-byte
  identical to the tested PR head.
- rocJITsu configuration: `gfx942_cdna3_kmd.json`.
- Physical control: real gfx942 with ROCm 7.13.0. The matching July 18 SDK is not
  installed on the available physical host; the public HSA queue lifecycle and
  gfx942 code object are otherwise identical.

## Problem

rocJITsu recycles a destroyed HSA queue's doorbell offset without resetting the
shared doorbell slot. A replacement queue initializes its private
`last_doorbell` state to `UINT64_MAX`, while the recycled shared slot still
contains packet index zero from the previous queue. The command-processor poller
can consume that stale `0 != UINT64_MAX` transition before ROCr publishes the
replacement queue's first packet. ROCr then rings packet zero, which is no longer
a value transition.

The result is intermittent but severe. The replacement queue's first dispatch
either never runs, or its completion signal reaches zero while the kernel's 64
stores never occur. Both outcomes have been observed from the same standalone
program.

## Expected behavior

Destroying a queue and creating another queue in the same HSA lifetime must not
transfer submission state between those queues. The first packet on every new
queue must execute exactly once and decrement only its own completion signal.

## Standalone reproducer

`repro.c` uses only public ROCm SDK HSA APIs. Under one `hsa_init` lifetime it
creates a gfx942 code-object reader, executable, output allocation, kernarg
allocation, single-producer queue, and completion signal; submits a trivial
scalar-store kernel; verifies all 64 outputs; destroys every object; and repeats
the entire queue generation. It reports the queue address, ring address,
doorbell handle, completion handle, and packet index for every generation.

The kernel writes `32.0f` once per lane. This deliberately removes
flash-attention, DBT, and complex ISA simulation from the progress question.
The completion wait is infinite because submitted HSA work has no valid local
deadline. The batch script owns the outer process deadline and gives every
process a private rocJITsu runtime directory under `/tmp`; source, binaries, and
result logs remain outside `/tmp`.

Build and run a high-probability rocJITsu batch:

```bash
export ROCM_PATH=/path/to/rocm-7.15.0a20260718
export ROCJITSU_PATH=/path/to/source-built-rocjitsu-install

CONCURRENCY=32 BATCHES=10 ./run-batch.sh
```

`run-batch.sh` builds both files with the selected SDK and runs two queue
generations with one packet per queue by default. Its useful controls are:

```bash
# Keep one queue and submit two consecutive packet indices.
QUEUE_GENERATION_COUNT=1 PACKETS_PER_QUEUE=2 \
  CONCURRENCY=32 BATCHES=10 ./run-batch.sh

# Destroy the queue before its already-completed signal.
CLEANUP_ORDER=queue-first CONCURRENCY=32 BATCHES=10 ./run-batch.sh
```

Run the direct-HSA physical control on gfx942 without rocJITsu:

```bash
export ROCM_PATH=/path/to/rocm-with-gfx942-support
./run-hardware-control.sh
```

The hardware script defaults to 100 complete queue generations in one HSA
lifetime and requires the public agent ISA name to contain `gfx942`.

## Observed results

The unmodified merged source fails with both cleanup orders:

| Path | Correct runs | False completions | Hangs |
| --- | ---: | ---: | ---: |
| Two queue generations, signal destroyed first | 314/320 | 5 | 1 |
| Two queue generations, queue destroyed first | 314/320 | 1 | 5 |
| One retained queue, two packets | 320/320 | 0 | 0 |

The retained-queue control is important: resetting the signal and advancing
from packet zero to packet one on the same live queue is clean. Queue
destruction and recreation are necessary. Swapping signal/queue destruction
order changes the distribution of symptoms but does not remove the defect.

A failing false-completion instance is:

```text
dispatch=1 completion=0
dispatch=1 verified=64 mismatches=0
dispatch=2 completion=0
dispatch=2 verified=64 mismatches=64
exit_status=2
```

A failing progress instance reaches only:

```text
queue_generation=1 packet_number=1 verified=64 mismatches=0
queue_generation=2 packet_number=1 submitted queue=0x1000029000
  ring=0x1000c00000 doorbell=68725767168
  completion=68725766784 packet_index=0
exit_status=124
```

The queue, ring, doorbell, and completion handles are all reused across those
two generations. The second completion signal remains at one until the outer
harness terminates the process.

On real gfx942, the direct-HSA control completed 100/100 queue generations with
zero mismatches. Generations 2 through 100 reused the same ring, doorbell, and
completion handles on every iteration:

```text
gpu_isa=amdgcn-amd-amdhsa--gfx942:sramecc+:xnack-
queue_generation=100 packet_number=1 completion_value=0
queue_generation=100 packet_number=1 verified=64 mismatches=0
summary queue_generations=100 packets_per_queue=1 workgroups=1
  cleanup=signal-first result=0
```

## Source diagnosis

The initial doorbell page is correctly filled with `0xFF`, matching
`HwQueue::last_doorbell == UINT64_MAX`. Queue destruction later pushes the
queue's offset into `free_doorbell_offsets` but leaves the last packet index in
the shared page. Queue creation pops that offset and publishes a new `HwQueue`
whose private sentinel no longer matches the shared slot.

The stale transition can be observed before ROCr advances and populates the
replacement ring. An early fetch then either consumes the only wakeup before
the new packet exists, producing the hang, or races the reused AQL slot while it
still contains stale/partially replaced state, producing the observed completion
without the new kernel's stores.

## Candidate fix validation

The minimal invariant-level fix resets a recycled shared doorbell slot to
`UINT64_MAX` under the existing KFD allocation lock, before the replacement
queue is registered with a command processor. The store is atomic with release
ordering, matching the poller's acquire load. A deterministic KFD test writes
zero to the first queue's doorbell, destroys that queue, creates its replacement,
confirms the same offset was reused, and verifies the shared slot was restored
to the unrung sentinel.

Results with only that runtime fix:

| Validation | Result |
| --- | ---: |
| Standalone, signal-first, 10 batches x 32 processes | 320/320 pass |
| Standalone, queue-first, 10 batches x 32 processes | 320/320 pass |
| Original flash-attention regression, 10 batches x 32 processes | 320/320 pass |
| Focused KFD/DBT/simulated-dispatch suite | 12/12 pass |
| Complete source corpus, sealed July 18 SDK | 1,551 pass, 0 fail, 1 intentional skip |

The complete corpus includes all five RCCL collectives. The host has software
RoCE interfaces, so the suite's intended socket-only RCCL path also requires
`NCCL_IB_DISABLE=1`; without it RCCL fails synthetic-vs-physical PCI topology
construction before reaching this queue behavior.

## Source-test lifetime correction

The upstream flash-attention helper used a routine 10-second HSA wait followed
by a fatal assertion before its queue, signal, allocations, executable, and
reader cleanup. A lost dispatch therefore appeared as a later exit-time
segmentation fault. That segfault is a test cleanup consequence, not the root
queue-progress symptom.

A separate test-only repair gives every dispatch helper one RAII owner, destroys
the queue before any packet-referenced resource, and uses an infinite HSA wait.
The existing 300-second CTest timeout remains the outer hang detector. All 12
affected simulated-dispatch, DBT-guest, and KFD tests pass with that ownership
repair.

## Impact

Any process that destroys and recreates HSA queues can intermittently lose the
replacement queue's first submission or receive a false completion. That is a
general runtime correctness failure, not a workload-specific simulator fidelity
gap: callers may hang forever, or consume untouched output as if GPU work
completed successfully.

atgutier:
## Motivation

rocjitsu runs unmodified ROCr/HIP applications against an emulated AMD GPU by intercepting /dev/kfd + DRM render-node ioctls (LD_PRELOAD interposer + SimulatedKfd driver) and executing kernels on the simdojo simulation engine. As coverage grew to multi-process workloads (RCCL collectives via a shared daemon), real Triton/HIP dispatch, and multi-GPU topologies, several classes of latent defect surfaced — process-teardown races, doorbell/dispatch-lifecycle bugs, GEM/fd leaks, and a severe RCCL performance regression. This PR hardens the KFD driver, command processor (CP), dispatch lifecycle, and the multi-process/daemon runtime so the emulator is correct and deterministic under concurrency and load.

## Technical Details

Command processor & dispatch lifecycle

  - Hardware-accurate wavefront termination. s_endpgm/halt() now frees a wave's SGPR/VGPR immediately and notifies the CP as one action (matching real HW),
  replacing the old lazy two-phase retire_halted_wfs*. num_wfs() now counts resident (non-halted) waves. This removes an event-ordering nondeterminism where
   a stray empty CU tick could free halted waves and make dispatch-count assertions flaky in full-suite runs.
  - Lost-doorbell fix on recycled queue slots. Doorbell offsets recycled via free_doorbell_offsets retained the prior queue's last-rung write index;
  combined with the CP's last_doorbell = ~0 init, a poll-thread scan in the register→ring window could latch a stale value and drop the next submission —
  hanging the host in hsa_signal_wait. Recycled slots are now reset to the ~0 sentinel at queue creation.
  - Cluster dispatch is now all-or-nothing. A clustered (gfx1250) dispatch that fails partway (dispatch_wf null or a malformed-kernarg/TTMP throw) now rolls
   back already-committed peers (abort_workgroup + erase_cluster_workgroup) instead of orphaning WG refcounts and LDS pins.
  - Inline CU dispatch continuation on CU-idle (avoids a per-step tick of latency across the many tiny kernels of a collective), plus a schedule_work()
  guard that never wakes an idle CU.

  RCCL / daemon performance

  - Eliminated a busy-wait on external dependency stalls. Barrier/dependency/SDMA-wait stalls were rescheduled on the MAIN (simulated-time) event queue via
  now+1, spinning millions of ticks per collective while real wall-clock RPC latency elapsed (~7–8× RCCL slowdown). arm_stall_recheck() now routes
  host-accessible (KFD) queues — which have a doorbell poll thread — to a paced 100 µs re-check via stall_pending_, keeping the MAIN queue for
  simulated-world timing only; internal test queues (no poll thread) keep the now+1 fallback.

  Multi-process / daemon runtime

  - Per-PID runtime directories for sockets and config handoff, inherited by descendants via $ROCJITSU_INVOCATION_DIR, eliminating stale-socket/config
  collisions across concurrent invocations; stale dead-PID dirs are reaped with a kill(pid,0) liveness check and symlink-safe teardown.
  - Daemon teardown wakes parked waiters. A client thread blocked in an infinite-timeout WAIT_EVENTS ioctl cannot be interrupted by stop_token or
  shutdown(fd); teardown now calls rj_vm_close_all_devices → close_all_processes, which drains each process's open-refcount to fire notify_closing() and
  unblock the join.
  - Daemon startup ordering (bind/listen before spawning the engine thread so error paths have nothing to join) and atomic client-fd close+erase under
  client_threads_mutex.
  - Local-mode engine thread is joined at exit. The interposer's local VM engine thread was detached and never stopped, so process exit() ran librocjitsu's
  static destructors while the engine was still in CommandProcessor::startup() — a use-after-free SIGSEGV that flaked under ctest -jN. The thread is now
  joinable and stopped/joined from an early (destructor(101)) hook before library teardown.

  GEM / DRM memory & fd lifetime

  - AMDGPU_GEM_VA MAP/REPLACE/UNMAP/CLEAR with interval-overlap matching for vmem GPU page-table mapping; PRIME_FD_TO_HANDLE mints stable monotonic handles
  and dup()s the dmabuf fd; GEM handles are reaped at DRM-file close and dup2/dup3 overwrite.
  - fork safety: reset_after_fork closes owned dmabuf dups before clearing GEM tables; async-signal-safety of the atfork child path documented.
  - Fixed premature munmap on close() via host_ptr ownership tracking, and a pending_gem_flags_ leak on dup2/dup3 target overwrite.

Config handoff

  - Three-tier config-path resolution ($ROCJITSU_INVOCATION_DIR → per-PID → well-known), applied consistently in the interposer, the DBT-guest loader, and
  the test helper; empty env vars treated as unset; bare-LD_PRELOAD clients fall back to the well-known location.

  simdojo engine

  - First-event self-scheduling moved from initialize() to startup() (so it runs after topology is frozen/partitioned); lock-free async-event drain via a
  pending double-checked flag; spin-based idle wait with no lost-wakeup between request_exit() and wait entry; improved run/dispatch flow.
  - Multi-GPU naming fix: each SoC is wrapped in a uniquely-named gpuN node so multi-GPU topologies no longer collide on the shared "soc" name (which had
  routed both GPUs' links to the same CUs).

  Tests & docs

  - New HaltSnapshotPlugin records wavefront register/exec/TID state at halt (reading live registers before they free), letting value/distribution tests
  observe dispatch results deterministically without reading freed slots; broad test migration to the new resident-num_wfs() semantics.
  - New interposer GEM/dup fd-lifetime tests; widened DBT dispatch completion-signal wait into a hang guard; updated vm-design.md / simdojo.md.

  Testing

  - Full serial Release ctest suite passes (daemon/RCCL tests are timing-sensitive and are run serially).
  - The previously-flaky CI hang (Cdna4ToCdna3SimulatedDispatchTest.TritonDynamicMatmulDispatchAndRun) is fixed and verified 40/40 under 2-core
  CPU-starvation repro (was a deterministic hang before).
  - The -jN interposer SIGSEGV is fixed and verified 0 failures across -j16×12 and -j32×15 over the Interposer/Daemon/GuestKfd suites (≈50% crash rate
  before).

## JIRA ID

N/A

PR #8672

## Test Plan

Run all CI tests
Run all rocjitsu test corpus tests

## Test Result

All CI tests: PASSED
All rocjitsu test corpus tests: PASSED

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/docs/vm-design.md
+++ b/emulation/rocjitsu/docs/vm-design.md
@@ -73,26 +73,33 @@ compute units in round-robin order.
 The CP is event-driven. During `startup()`, it schedules a doorbell event
 for any pre-loaded packets. Each doorbell event triggers `step()`, which
 processes one dispatch packet: for each workgroup, it calls `dispatch_wf()`
-on the next CU in round-robin order, then calls `activate()` on that CU.
+on the next CU in round-robin order. `dispatch_wf()` self-schedules the CU's
+tick via `schedule_work()`; there is no separate `activate()` call.
 
 - `enqueue(packet)` - pre-loads packets without triggering dispatch (used
   by config loader and tests)
 - `submit(packet)` - thread-safe; appends packet and schedules an async
   doorbell event via `schedule_event_now()`
-- `activate()` - registers the CP as a primary component and begins
-  processing doorbell events
 
 Each CU runs its dispatched wavefronts independently. The CU is
-self-driving: `dispatch_wf()` schedules a tick event that calls
-`execute_quantum()`, which executes up to `functional_quantum`
-instructions before yielding back to the event loop. The tick event
-reschedules itself at `now + quantum` if work remains. Since functional
-mode is 1 CPI, ticks advance proportionally to instruction count. The
-quantum allows CU events to interleave, guaranteeing forward progress
-for inter-CU synchronization patterns such as spin-locks or semaphore
-acquire/release on global memory. When a CU becomes idle, it stops
-scheduling and fires its `on_idle` callback. When all CUs are idle
-and no packets remain, the CP signals completion via
+self-driving: `dispatch_wf()` calls `schedule_work()`, which schedules a
+tick event (only when the CU has runnable wavefronts) that calls
+`execute_quantum()`. A quantum executes up to `kFunctionalQuantum`
+instructions, but may yield early when a wavefront requests it (e.g.
+`s_sleep`, a vendor-dependency retry). The tick reschedules itself at
+`now + max(1, last_quantum_executed_)` — i.e. by the work actually
+executed, so an early yield resumes promptly instead of leaping a full
+quantum, while `max(1, ...)` keeps the event strictly in the future.
+Since functional mode is 1 CPI, ticks advance proportionally to
+instruction count. The quantum allows CU events to interleave,
+guaranteeing forward progress for inter-CU synchronization patterns such
+as spin-locks or semaphore acquire/release on global memory.
+
+A wavefront that reaches `s_endpgm` halts: it frees its SGPR/VGPR
+resources immediately and notifies the CP of workgroup completion (there
+is no separate lazy retirement pass). When a CU has no resident
+wavefronts it stops scheduling and fires its `on_idle` callback. When all
+CUs are idle and no packets remain, the CP signals completion via
 `engine()->primary_release()`.
 
 ---
@@ -108,7 +115,7 @@ Config loader / Driver::submit()
                 cu->dispatch_wf(wg_id, pc, sgprs, vgprs)
                   └── find idle slot, allocate SGPR/VGPR blocks
                       initialize wavefront state (pc, wg_id)
-                      schedule tick event if CU was idle (self-driving)
+                      schedule_work() -> tick event (self-driving)
 ```
 
 A `DispatchPacket` specifies:
--- a/emulation/rocjitsu/docs/vm-design.md
+++ b/emulation/rocjitsu/docs/vm-design.md
@@ -82,13 +82,16 @@ on the next CU in round-robin order, then calls `activate()` on that CU.
 - `activate()` - registers the CP as a primary component and begins
   processing doorbell events
 
-Each CU runs its dispatched wavefronts independently. In functional mode,
-a work event calls `advance()`, which executes up to `functional_quantum`
-instructions (or all remaining if quantum is 0) before yielding back to
-the event loop. A non-zero quantum allows CU events to interleave,
-guaranteeing forward progress for inter-CU synchronization patterns such
-as spin-locks or semaphore acquire/release on global memory. When a CU
-becomes idle, it fires its `on_idle` callback. When all CUs are idle
+Each CU runs its dispatched wavefronts independently. The CU is
+self-driving: `dispatch_wf()` schedules a tick event that calls
+`execute_quantum()`, which executes up to `functional_quantum`
+instructions before yielding back to the event loop. The tick event
+reschedules itself at `now + quantum` if work remains. Since functional
+mode is 1 CPI, ticks advance proportionally to instruction count. The
+quantum allows CU events to interleave, guaranteeing forward progress
+for inter-CU synchronization patterns such as spin-locks or semaphore
+acquire/release on global memory. When a CU becomes idle, it stops
+scheduling and fires its `on_idle` callback. When all CUs are idle
 and no packets remain, the CP signals completion via
 `engine()->primary_release()`.
 
@@ -105,8 +108,7 @@ Config loader / Driver::submit()
                 cu->dispatch_wf(wg_id, pc, sgprs, vgprs)
                   └── find idle slot, allocate SGPR/VGPR blocks
                       initialize wavefront state (pc, wg_id)
-                cu->activate()
-                  └── schedule work event (functional) or resume clock (clocked)
+                      schedule tick event if CU was idle (self-driving)
 ```
 
 A `DispatchPacket` specifies:
@@ -208,7 +210,7 @@ The SoC plugs into simdojo's `SimulationEngine` directly. The C API layer
    config and creates a `SoC` with engine configuration. The C API sets the
    SoC as the topology root via `engine.topology().set_root(std::move(soc))`.
 
-2. **`build()`** - Partitions topology, initializes all components, and
+2. **`create()`** - Partitions topology, initializes all components, and
    sets up the engine.
 
 3. **`run()`** - Starts all components and drains events continuously.
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
@@ -90,12 +90,30 @@ using rocjitsu::Sysfs;
 namespace {
 
 /// @brief Attempt to connect @p sock to the AF_UNIX socket at @p path.
-/// @returns 0 on success; -1 with errno set on failure (socket left open).
-int try_connect(int sock, const std::string &path) {
+/// @returns A connected socket fd on success; -1 with errno set on failure.
+/// @details Creates a FRESH socket for each attempt: POSIX leaves a stream
+/// socket's state unspecified after a failed connect(), so a socket must not be
+/// reused for a second connect(). Fails with ENAMETOOLONG rather than silently
+/// truncating a path that does not fit sun_path, so a too-long runtime dir cannot
+/// connect to an unintended (truncated) socket endpoint.
+int try_connect(const std::string &path) {
   sockaddr_un addr{};
   addr.sun_family = AF_UNIX;
-  path.copy(addr.sun_path, sizeof(addr.sun_path) - 1);
-  return connect(sock, reinterpret_cast<sockaddr *>(&addr), sizeof(addr));
+  if (path.size() >= sizeof(addr.sun_path)) {
+    errno = ENAMETOOLONG;
+    return -1;
+  }
+  std::memcpy(addr.sun_path, path.c_str(), path.size());
+  int sock = socket(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0);
+  if (sock < 0)
+    return -1;
+  if (connect(sock, reinterpret_cast<sockaddr *>(&addr), sizeof(addr)) != 0) {
+    int saved = errno;
+    rocjitsu::libc_passthrough().close(sock);
+    errno = saved;
+    return -1;
+  }
+  return sock;
 }
 
 /// @brief Connect to the daemon for this invocation's per-PID runtime directory.
@@ -106,21 +124,21 @@ int try_connect(int sock, const std::string &path) {
 /// app is never silently cross-connected to an unrelated daemon at the shared
 /// well-known socket if its own daemon's socket is transiently unavailable.
 int connect_to_daemon(const std::string &runtime_dir) {
-  int sock = socket(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0);
-  if (sock < 0)
-    return -1;
-
-  if (try_connect(sock, runtime_dir + "/daemon.sock") == 0)
+  int sock = try_connect(runtime_dir + "/daemon.sock");
+  if (sock >= 0)
     return sock;
 
   // Fall back to the well-known socket only for invocations that never created a
   // per-PID directory (attach / daemon-only). access() goes through the real libc
-  // so it does not re-enter this interposer's own path hooks.
-  if (rocjitsu::libc_passthrough().access(runtime_dir.c_str(), F_OK) != 0 &&
-      try_connect(sock, rocjitsu::rpc_default_socket_path()) == 0)
-    return sock;
-
-  rocjitsu::libc_passthrough().close(sock);
+  // so it does not re-enter this interposer's own path hooks. Gate strictly on the
+  // dir genuinely not existing (ENOENT / a non-directory component, ENOTDIR): any
+  // other error (EACCES/EPERM, transient IO) must NOT trigger the fallback, or a
+  // daemon-mode client whose own dir is momentarily inaccessible could be silently
+  // cross-connected to an unrelated daemon at the shared socket.
+  const bool dir_absent = rocjitsu::libc_passthrough().access(runtime_dir.c_str(), F_OK) != 0 &&
+                          (errno == ENOENT || errno == ENOTDIR);
+  if (dir_absent)
+    return try_connect(rocjitsu::rpc_default_socket_path());
   return -1;
 }
 
@@ -263,11 +281,18 @@ public:
     sysfs_fds_.clear();
     drm_fds_.clear();
     kfd_dup_fds_.clear();
+    // Drop GEM bookkeeping WITHOUT munmapping cpu_ptr or unmapping PTEs: those host
+    // pages and the parent's page table belong to the parent process and the child
+    // must not touch them. The child re-initializes a fresh driver; any GEM mappings
+    // it needs are re-created via EXPORT/PRIME/GEM_VA. (Deliberate, like the remote_
+    // and mutex handling above — not an oversight.)
     gem_entries_.clear();
     in_construction = false;
   }
 
   LinuxKfd *driver() { return active_driver_.load(std::memory_order_acquire); }
+  /// @brief True if the active driver is the local SimulatedKfd (not remote/guest).
+  bool driver_is_simulated() { return dynamic_cast<SimulatedKfd *>(driver()) != nullptr; }
   int driver_fd() {
     auto *d = driver();
     return d ? d->fd() : -1;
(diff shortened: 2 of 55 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
```diff
@@ -1108,6 +1274,23 @@ RJ_INTERPOSER_EXPORT int close(int fd) {
...
+  {
+    // Tear down any GPU page-table ranges still installed from this dmabuf BO
+    // BEFORE munmapping its lazy host mapping: the page table holds host
+    // pointers into cpu_ptr, so unmapping the pages without clearing the PTEs
+    // would leave the simulated GPU dereferencing freed memory. gem_va_unmap
+    // needs the local simulated driver; if none, the ranges are moot (the page
+    // table went away with the driver).
+    void *gem_cpu_ptr = nullptr;
+    uint64_t gem_size = 0;
+    auto gem_ranges = InterposerContext::ctx.untrack_gem(fd, &gem_cpu_ptr, &gem_size);
```
kuhar: We should keep imported GEM state alive until the GEM handle is closed, not until the temporary dma-buf fd is closed. ROCr closes that export fd immediately after `VMemorySetAccessPerHandle()` returns; the current `close()` path then calls `untrack_gem()`, removes the installed VA ranges, and unmaps `cpu_ptr`. Please track this state by DRM file description/GEM handle and release it on `DRM_IOCTL_GEM_CLOSE`, otherwise the mapping disappears before the caller can use it.

emulation/rocjitsu/docs/vm-design.md
```diff
@@ -82,13 +82,16 @@ on the next CU in round-robin order, then calls `activate()` on that CU.
...
-instructions (or all remaining if quantum is 0) before yielding back to
-the event loop. A non-zero quantum allows CU events to interleave,
-guaranteeing forward progress for inter-CU synchronization patterns such
-as spin-locks or semaphore acquire/release on global memory. When a CU
-becomes idle, it fires its `on_idle` callback. When all CUs are idle
+Each CU runs its dispatched wavefronts independently. The CU is
+self-driving: `dispatch_wf()` schedules a tick event that calls
+`execute_quantum()`, which executes up to `functional_quantum`
+instructions before yielding back to the event loop. The tick event
+reschedules itself at `now + quantum` if work remains. Since functional
```
kuhar: It would be good to align this design section with the implementation. `dispatch_wf()` now self-schedules through `schedule_work()` with no CU `activate()` call, and remaining work is rescheduled at `now + max(1, last_quantum_executed_)`, not at a fixed quantum. The current text preserves the removed lifecycle and misstates the early-yield timing.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
```diff
@@ -1275,6 +1468,38 @@ RJ_INTERPOSER_EXPORT int ioctl(int fd, unsigned long request, ...) {
...
+        return 0;
+
+      int dmabuf_fd = InterposerContext::ctx.gem_fd_for_handle(va->handle);
+      if (va->operation == AMDGPU_VA_OP_MAP || va->operation == AMDGPU_VA_OP_REPLACE) {
+        void *host = nullptr;
+        uint32_t alloc_flags = 0;
+        // gem_prepare_map does the lazy mmap, bounds check, and VA-range tracking
+        // under fd_mutex_, returning the host pointer by value so no GEM entry
+        // pointer escapes the lock (avoids a race/UAF against a concurrent close).
+        if (InterposerContext::ctx.gem_prepare_map(dmabuf_fd, va->va_address, va->offset_in_bo,
```
kuhar: GEM_VA should report failure when it does not perform the requested operation. A missing local driver, unknown or stale handle, out-of-bounds range, failed `mmap()`, and unknown operation all currently reach `return 0`. Please return an appropriate error so userspace does not record a nonexistent mapping and fail later with a GPU page-table fault.

kuhar (commented):
We should add targeted regression coverage for the behavioral fixes in this PR. The test diff only renames `advance()` to `execute_quantum()`; it does not exercise event-loop early-yield timing (`compute_unit.h:602`), ownership-specific teardown (`simulated_kfd.cpp:534`), multi-GPU `gpuN` routing (`virtual_machine.cpp:45`), stale-directory symlink rejection (`main.cpp:423`), or the GEM_VA lifetime/error paths (`interposer.cpp:1286,1491`). Please cover these distinct failure modes so the fixes cannot regress.

We should get the configured RCCL regression suite green before merging. On the same machine and configuration, `RcclDaemonTest.AllReduce` passes at base `72b967429114` in 6.3 seconds, but fails at head `185215006e9d` after 91.8 seconds with both ranks reporting HIP `invalid device function`; all five registered RCCL cases failed the same way in the reviewer run. This isolates a patch regression rather than an unsupported installation. Please fix it, or gate and document the intended support boundary.

emulation/rocjitsu/tools/rocjitsu/main.cpp
```diff
@@ -396,11 +396,38 @@ bool write_config_file(const std::string &config_path) {
...
+  // Advance the iterator with an error_code (not the throwing operator++): another
+  // launcher may remove_all an entry concurrently, and a throw here would abort the
+  // launcher before exec. Best-effort — any filesystem error just ends the scan.
+  std::error_code ec;
+  std::filesystem::directory_iterator it(rpc_default_runtime_dir(), ec);
+  const std::filesystem::directory_iterator end;
+  for (; !ec && it != end; it.increment(ec)) {
+    const std::string name = it->path().filename().string();
+    if (!std::all_of(name.begin(), name.end(), [](unsigned char c) { return std::isdigit(c); }))
+      continue;
```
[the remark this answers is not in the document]
kuhar: We should reject symlinks before this `remove_all()`. `directory_entry::is_directory(error_code&)` follows the target, so a numeric symlink to a directory still passes this filter despite the new comment. Please use `symlink_status()` plus `file_type::directory`, or explicitly reject symlinks.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/compute_unit.cpp
```diff
@@ -109,14 +109,14 @@ Wavefront *ComputeUnitCore::dispatch_wf(uint32_t wg_id, uint64_t pc, uint32_t sg
       break;
     }
   }
-  if (slot >= config_.num_wf_slots)
-    return nullptr;
 
-  int32_t sgpr_base = sgpr_file_.allocate(sgprs);
+  assert(slot < config_.num_wf_slots && "tried to dispatch to an invalid wavefront slot");
+
+  int32_t sgpr_base = sgpr_file_.allocate(num_sgprs);
```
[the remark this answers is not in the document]
kuhar: This would be safer with a release-build guard in the callee. Both current dispatch sites call `can_accept_workgroup()` first, but if that invariant drifts the assertion disappears and `wfs_[slot]` becomes an out-of-bounds access; the header also still promises `nullptr` when no slot is free. Please keep a release-safe early return or shared slot check, and add slot-exhaustion coverage so the API contract cannot drift.

benvanik:
FYI — we just hit what looks like this same lifetime bug while dogfooding rocJITsu against a ROCm nightly, and reduced it to a standalone ROCm-SDK-only reproducer:

https://gist.github.com/benvanik/be03fc5f05aa1a4729bc88068f11cdd3

It performs two fully balanced hsa_init/hsa_shut_down lifetimes. The physical-hardware control passes, while rocJITsu crashes during the second hsa_init in EventState::set_event with two AsyncEventsLoop threads present. This appears to exercise the caller-owned signal/event mapping lifetime fix in this PR and may be useful as a tight regression test.

atgutier:
> FYI — we just hit what looks like this same lifetime bug while dogfooding rocJITsu against a ROCm nightly, and reduced it to a standalone ROCm-SDK-only reproducer:
> 
> https://gist.github.com/benvanik/be03fc5f05aa1a4729bc88068f11cdd3
> 
> It performs two fully balanced hsa_init/hsa_shut_down lifetimes. The physical-hardware control passes, while rocJITsu crashes during the second hsa_init in EventState::set_event with two AsyncEventsLoop threads present. This appears to exercise the caller-owned signal/event mapping lifetime fix in this PR and may be useful as a tight regression test.

Thanks for this, will be a big help!

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
```diff
@@ -1213,17 +1496,30 @@ RJ_INTERPOSER_EXPORT int ioctl(int fd, unsigned long request, ...) {
...
-        uint32_t flags;
-        int32_t fd;
-      };
       auto *prime = static_cast<drm_prime_handle *>(arg);
       if (prime->fd < 0) {
         errno = EINVAL;
         return -1;
       }
-      prime->handle = static_cast<uint32_t>(prime->fd) + 1u;
+      prime->handle = InterposerContext::ctx.gem_handle_for_fd(prime->fd);
```
kuhar: We should allocate GEM handles independently of transient dma-buf fd numbers and scope them to each DRM file instance. `PRIME_FD_TO_HANDLE` currently returns `fd + 1`, but ROCr closes the export fd after access setup; Linux may then reuse that number while the previous handle is still live, and `track_gem_flags()` tears down the previous entry and its PTEs. Please mint stable handles, retain only a temporary fd-to-import association, reap remaining handles when the DRM file closes, and add an fd-reuse regression with two concurrently live BOs.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
```diff
@@ -1319,6 +1615,55 @@ RJ_INTERPOSER_EXPORT int ioctl(int fd, unsigned long request, ...) {
...
+      // failure rather than a phantom success so userspace does not record a mapping
+      // that was never installed and later fault on the GPU page table. gem_map/
+      // gem_unmap do the bookkeeping AND the page-table mutation atomically under
+      // fd_mutex_ (so a concurrent GEM_CLOSE cannot leave dangling PTEs); they
+      // return false on no-driver / unknown handle / out-of-bounds / failed mmap.
+      if (!InterposerContext::ctx.driver_is_simulated()) {
+        errno = ENODEV;
+        return -1;
+      }
+      if (va->operation == AMDGPU_VA_OP_MAP || va->operation == AMDGPU_VA_OP_REPLACE) {
```
kuhar: We should either reject `REPLACE` and `CLEAR` until their ownership semantics are modeled, or handle them separately. Treating `REPLACE` as an ordinary MAP leaves the range recorded under the old handle, so closing that handle later unmaps the replacement. Likewise, `gem_unmap()` modifies the page table without first verifying that the supplied handle owns the exact range, and reports success even when no record was erased. Please validate or transfer mapping ownership before changing the PTEs.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/compute_unit.cpp
```diff
@@ -409,6 +392,24 @@ void ComputeUnitCore::issue_instruction(Wavefront *active) {
...
+  // and pc-advance below must not run on the dead slot. The dedicated
+  // onAmdgpuWavefrontHalted hook already fired (with live state) from halt().
+  // s_endpgm is never a memory op, so just reclaim the decoded instruction.
+  //
+  // Note the intentional asymmetry: an s_endpgm that defers to ENDING (pending
+  // memory waits) is NOT halted here, so it DOES fire onAmdgpuAfterExecuteInstruction
+  // below; the immediate-halt case does not. onAmdgpuWavefrontHalted is the
+  // authoritative terminal hook and fires in both cases — consumers should observe
+  // termination there, not via the after-execute hook.
+  if (active->is_halted()) {
```
kuhar, line 423: I think we should add a focused hook-ordering test for this new immediate-halt branch. For a kernel ending in `s_endpgm`, assert that the terminator emits BEFORE_INSTRUCTION but not AFTER_INSTRUCTION, while non-terminators retain matched hooks; `OrderingPlugin` already captures the needed mnemonics. If the deferred ENDING case intentionally still emits AFTER_INSTRUCTION, covering that variant too would pin the distinction and protect the guard that prevents hooks or logging from reading a slot after `halt()` frees its registers.

kuhar (commented):
We should make the configured RCCL cases green before merging, or explicitly gate and document the unsupported configuration. At current head `410384e01c`, `RcclDaemonTest.AllReduce` failed after 100.31 seconds with both ranks reporting HIP `invalid device function` at their first RCCL operation. The same test passed at base `72b967429114` in 6.31 seconds and failed on the prior patch head, so the fix commits have not restored this integration surface.

benvanik:
Follow-up dogfood result: I built the current PR head against ROCm SDK nightly `7.15.0a20260718` and reran four standalone SDK-only reducers.

- **PASS:** complete HSA shutdown followed by reinitialization — https://gist.github.com/benvanik/be03fc5f05aa1a4729bc88068f11cdd3
- **PASS:** delayed publication of an advertised `INVALID` AQL slot — https://gist.github.com/benvanik/84d225c8661a2c65909c4640c8e9b19c
- **PASS:** HSA initialization in an ASAN process; the shipped nightly takes about 104 seconds and returns `OUT_OF_RESOURCES`, while this PR initializes in well under a second — https://gist.github.com/benvanik/5c1216c815ff11f2f78bc3de12a90692
- **Still fails:** queue destruction leaves all four per-queue polling threads alive (`6 -> 10 -> 10`), now tracked separately as #8813 — https://gist.github.com/benvanik/167a5d97ac396d8cbb78cf315b54f647

The shipped July 18 nightly predates this PR and still reproduces the first three failures, so these are source-candidate results pending a containing known-green nightly.

benvanik:
FYI from downstream source qualification against the matching July 18 ROCm SDK: this PR also stabilizes a parallel local-launch teardown case that the existing upstream test exposes.

On current `develop`, `InterposerDupTest.Dup2OntoFreshFdRoutesKfd` passes 20/20 serial runs, but concurrent launch stress produced 3 failures in 160 runs (two exit 139) and 4 exit-139 failures in a later 320-run A/B/A confirmation. In each crash, the child GTest printed `PASSED`; the process then died during post-test lifetime/teardown. Giving every process a unique `ROCJITSU_RUNTIME_DIR` did not eliminate it. Running the same test binary through this PR's release-built launcher/interposer passed 160/160 concurrent runs. This looks covered by the process/VM lifetime work here rather than a distinct issue, and it may be a useful regression stress shape.

One RCCL qualification sharp edge also surfaced. `RcclDaemonTest::run_rccl_rank` disables P2P and SHM and pins `NCCL_SOCKET_IFNAME=lo`, but leaves IB enabled. On a host with software RoCE interfaces, RCCL selected that fabric and failed communicator initialization while reconciling real NIC PCI topology with rocJITsu's synthetic GPU topology. Completing the test's socket-only contract with `NCCL_IB_DISABLE=1` made all five collectives pass serially and in one `ctest -j5` run (with the matching gfx950 RCCL device pack installed):

```diff
-    cmd += " NCCL_P2P_DISABLE=1 NCCL_SHM_DISABLE=1 HSA_NO_SCRATCH_RECLAIM=1"
+    cmd += " NCCL_P2P_DISABLE=1 NCCL_SHM_DISABLE=1 NCCL_IB_DISABLE=1"
+           " HSA_NO_SCRATCH_RECLAIM=1"
```

benvanik:
Follow-up: I configured and built this exact PR head with `BUILD_TESTING=ON` against the matching July 18 SDK, then ran the complete discovered corpus with `ctest -j32`. Result: **1,511 passed, 0 failed, 1 intentionally skipped** (`ExecutionPluginTest.MfmaFastPathReadHookReportsRace`). All five RCCL collectives were included, using the matching gfx950 device pack and an outer `NCCL_IB_DISABLE=1` pending the rank-environment tweak above.

benvanik:
FYI: the current `43874e5` head still leaves the detached local VM engine
running through rocJITsu finalization, and short-process stress can race engine
startup against C++ teardown. I filed #8817 with a tight ROCm-SDK-only
reproducer and independent finalizer-order auditor:

https://github.com/ROCm/rocm-systems/issues/8817

https://gist.github.com/benvanik/af4004a69434955e82129e3415828c64

The obvious unconditional join in rocJITsu's finalizer hangs later HIP/ROCR
teardown because of DSO finalization order. The packet includes the phase-aware
five-file experiment we used instead; applied unchanged to this current head,
it passes the audit, 320/320 concurrent short-process runs, and all 1,539 source
tests with the one intentional skip.

emulation/rocjitsu/docs/vm-design.md
kuhar: We should describe this timing in terms of step() loop iterations rather than executed instructions. A fully stalled CU still advances by the full quantum; please update both this paragraph and the public ExecComputeUnit prose to match that behavior.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/compute_unit.cpp
kuhar: The new TerminatorEmitsBeforeButNotAfterInstruction test covers immediate halt, but the deferred ENDING path still has no ordering coverage. It would be useful to add the outstanding-memory-wait case so the intentional AFTER_INSTRUCTION behavior cannot drift.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
kuhar: The stable handles fix fd-number reuse, but they are still global rather than scoped to a DRM file description. GEM_VA and GEM_CLOSE accept a handle without validating the caller fd, and dup/dup2/fcntl aliases are tracked as independent DRM files, so closing one alias can reap handles while another remains live. Please model the shared file description, validate handle ownership, and reap only when its final fd closes.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
```diff
@@ -787,11 +860,239 @@ class InterposerContext {
...
+    if (it == gem_entries_.end())
+      return false;
+    GemEntry &gem = it->second;
+    const GemMapping range{va_address, map_size};
+    if (std::find(gem.installed_vas.begin(), gem.installed_vas.end(), range) ==
+        gem.installed_vas.end())
+      return false; // This handle does not own the exact range — do not touch PTEs.
+    if (!drv->gem_va_unmap(va_address, map_size))
+      return false;
+    std::erase(gem.installed_vas, range);
```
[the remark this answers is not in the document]
kuhar: This still applies at 43874e5: interposer.cpp uses std::find in the GEM helpers but still has no explicit <algorithm> include. Please add the direct include so this does not depend on transitive headers.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/interposer.cpp
kuhar: The ownership transfer is implemented now, but REPLACE and CLEAR still have no regression coverage. It would be useful to test that replacing A with B leaves B mapped after A closes, and that CLEAR updates the owner bookkeeping so its later close does not unmap the range again.

benvanik:
FYI, I rebuilt current head `7830653` against the matching July 18 SDK with
the same phase-aware lifetime patch from #8817; it still applies unchanged
after the two config-discovery commits.

That exact composition passes the independent finalizer-order audit, 320/320
concurrent short-process runs, and the standalone reinitialization,
delayed-AQL-header, and ASAN-startup reducers. The complete source corpus also
passes 1,539/1,539 tests with the one intentional skip, including all five RCCL
collectives.

The updated standalone packet and source diagnosis are here:
https://gist.github.com/benvanik/af4004a69434955e82129e3415828c64

kuhar (approved):

PR #8673, opened 2026-07-16
feat(rocr): PC Sampling: implement lock-free Multi-producer Single-consumer queue

shwetagkhatri:
## Summary
- Replace mutex-protected pending_flush_count with a lock-free MPSC (multi-producer single-consumer) queue for XCC flush notifications
- Eliminates atomic contention on shared counter
- Consumer drains queue as wakeup signal for aggregated delivery
- Lock-free fast path for producer threads
- Scales better with XCC count (8+ on MI300X)

## Test plan
- [ ] Build verification on MI300X (gfx942)
- [ ] PC sampling functional tests with multiple XCCs
- [ ] Stress test with concurrent flush notifications

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/inc/amd_gpu_agent.h
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/inc/amd_gpu_agent.h
@@ -965,6 +965,46 @@ class GpuAgent : public GpuAgentInt {
   //   - Per-XCC mutex serializes XCC thread vs consumer thread for each XCC's buffer
   //   - Consumer thread aggregates data and delivers callbacks when SUM(pending) >= buffer_size
   //   - Callbacks contain data from whichever XCCs have pending samples at threshold time
+
+  // Lock-free MPSC (multi-producer single-consumer) queue for XCC flush notifications.
+  // Multiple XCC flush threads enqueue their IDs; the consumer drains the queue as a wakeup signal.
+  // Aggregation/delivery currently still scans all XCCs when building the staging buffer.
+
+  struct alignas(64) pcs_mpsc_queue_t {
+    static constexpr uint32_t CAPACITY = 64;
+    static constexpr uint32_t EMPTY_SLOT = UINT32_MAX;
+
+    std::atomic<uint32_t> buffer[CAPACITY];
+    alignas(64) std::atomic<uint32_t> write_pos;  // Written by producers; cache-line isolated from buffer
+    alignas(64) uint32_t read_pos;                // Written by consumer only; isolate from write_pos to avoid false sharing
+
+    pcs_mpsc_queue_t() : write_pos(0), read_pos(0) {
+      for (uint32_t i = 0; i < CAPACITY; ++i) {
+        buffer[i].store(EMPTY_SLOT, std::memory_order_relaxed);
+      }
+    }
+
+    // Enqueue XCC ID - called by multiple producer threads (XCC flush threads)
+    // Returns true on success, false if queue is full (should not happen in practice)
+    bool enqueue(uint32_t xcc_id);
+
+    // Try to dequeue an XCC ID - called by single consumer thread
+    // Returns true and sets xcc_id on success, false if queue is empty
+    bool try_dequeue(uint32_t& xcc_id);
+
+    // Check if queue has data - called by consumer to check before sleeping
+    bool has_data() const;
+
+    // Reset queue to empty state - must be called before starting a new session
+    void reset() {
+      for (uint32_t i = 0; i < CAPACITY; ++i) {
+        buffer[i].store(EMPTY_SLOT, std::memory_order_relaxed);
+      }
+      write_pos.store(0, std::memory_order_relaxed);
+      read_pos = 0;
+    }
+  };
+
   struct alignas(64) per_xcc_pcs_data_t {
     pcs_sampling_data_t* device_data;         // This XCC's device buffer region
     os::Thread thread;                        // Thread handle for this XCC's flush thread
@@ -1009,10 +1049,10 @@ class GpuAgent : public GpuAgentInt {
 
     /* Consumer thread for aggregated callback delivery */
     std::thread consumer_thread;            // Aggregates data and delivers callbacks
-    std::mutex consumer_mutex;              // Protects consumer_cv and pending_flush_count
+    pcs_mpsc_queue_t flush_queue;           // Lock-free queue of XCC IDs with pending data
+    std::mutex consumer_wakeup_mutex;       // Protects consumer_cv (minimal lock for wakeup only)
     std::condition_variable consumer_cv;    // Wakes consumer when XCC threads have new data
     std::atomic<bool> consumer_exit;        // Signal consumer thread to exit
-    std::atomic<uint32_t> pending_flush_count;  // How many XCCs have notified consumer
     std::mutex delivery_mutex;              // Serializes callback delivery (consumer vs Flush)
 
     pcs::PcsRuntime::PcSamplingSession* session;
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
@@ -4143,7 +4143,7 @@ hsa_status_t GpuAgent::PcSamplingStart(pcs::PcsRuntime::PcSamplingSession& sessi
     pcs_data->xcc_data[xcc_id].which_buffer = 0;
   }
   pcs_data->consumer_exit.store(false, std::memory_order_release);
-  pcs_data->pending_flush_count.store(0, std::memory_order_release);
+  pcs_data->flush_queue.reset();  // Clear any stale entries from previous session
 
   struct ThreadData {
     GpuAgent* agent;
@@ -4728,6 +4728,55 @@ hsa_status_t GpuAgent::PcSamplingFlushDeviceBuffersPerXCC_PM4(
   return HSA_STATUS_SUCCESS;
 }
 
+// Lock-free MPSC Queue Implementation
+// This queue allows XCC flush threads to notify the consumer thread which XCCs
+// have pending data, eliminating the need for a mutex-protected counter.
+// The queue uses fetch_add for concurrent writes and compare_exchange for reads.
+
+bool GpuAgent::pcs_mpsc_queue_t::enqueue(uint32_t xcc_id) {
+  // Reserve a slot using atomic fetch_add
+  uint32_t pos = write_pos.fetch_add(1, std::memory_order_acq_rel);
+  uint32_t slot = pos % CAPACITY;
+
+  // Wait for slot to be empty (should be immediate in normal operation)
+  uint32_t expected = EMPTY_SLOT;
+  int attempts = 0;
+  while (!buffer[slot].compare_exchange_weak(expected, xcc_id,
+                                             std::memory_order_release,
+                                             std::memory_order_relaxed)) {
+    if (++attempts > 100) {
+      std::this_thread::yield();
+    }
+    if (attempts > 10000) {
+      // Queue full - consumer is severely behind. Return failure to avoid hang.
+      return false;
+    }
+    expected = EMPTY_SLOT;
+  }
+  return true;
+}
+
+bool GpuAgent::pcs_mpsc_queue_t::try_dequeue(uint32_t& xcc_id) {
+  uint32_t slot = read_pos % CAPACITY;
+
+  // Read the value from the slot
+  uint32_t val = buffer[slot].load(std::memory_order_acquire);
+  if (val == EMPTY_SLOT) {
+    return false;  // Queue is empty
+  }
+
+  // Mark slot as empty and advance read position
+  buffer[slot].store(EMPTY_SLOT, std::memory_order_release);
+  read_pos++;
+  xcc_id = val;
+  return true;
+}
+
+bool GpuAgent::pcs_mpsc_queue_t::has_data() const {
+  uint32_t slot = read_pos % CAPACITY;
+  return buffer[slot].load(std::memory_order_acquire) != EMPTY_SLOT;
+}
+
 void GpuAgent::PcSamplingThreadPerXCC(pcs_data_t& pcs_data, uint32_t xcc_id,
                                       const char* thread_name) {
   try {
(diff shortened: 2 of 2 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -4728,6 +4728,55 @@ hsa_status_t GpuAgent::PcSamplingFlushDeviceBuffersPerXCC_PM4(
...
 }
 
+// Lock-free MPSC Queue Implementation
+// This queue allows XCC flush threads to notify the consumer thread which XCCs
+// have pending data, eliminating the need for a mutex-protected counter.
+// The queue uses fetch_add for concurrent writes and compare_exchange for reads.
+
+bool GpuAgent::pcs_mpsc_queue_t::enqueue(uint32_t xcc_id) {
+  // Reserve a slot using atomic fetch_add
+  uint32_t pos = write_pos.fetch_add(1, std::memory_order_acq_rel);
```
cfreeamd, line 4738: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** `enqueue()` reserves a slot with `write_pos.fetch_add(1)` here, before publishing it via CAS below. If the CAS loop exhausts its 10000 attempts and returns `false`, `write_pos` has already advanced but this slot is never filled. Since `try_dequeue()`/`has_data()` only ever inspect the single slot at `read_pos % CAPACITY` and advance strictly in order, the consumer will eventually reach that unfilled slot, see `EMPTY_SLOT`, stop draining, and never wake again for later items (which land in higher slots). One enqueue failure silently wedges the consumer for the rest of the session. Even guarded as "should not happen," the failure mode is a hang rather than graceful degradation — consider not advancing `write_pos` for a slot that may fail to publish, or making the consumer tolerant of holes.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -4728,6 +4728,55 @@ hsa_status_t GpuAgent::PcSamplingFlushDeviceBuffersPerXCC_PM4(
   return HSA_STATUS_SUCCESS;
 }
 
+// Lock-free MPSC Queue Implementation
+// This queue allows XCC flush threads to notify the consumer thread which XCCs
+// have pending data, eliminating the need for a mutex-protected counter.
+// The queue uses fetch_add for concurrent writes and compare_exchange for reads.
```
cfreeamd, line 4734: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** This comment is reversed: writes (`enqueue`) use `fetch_add` + `compare_exchange`; reads (`try_dequeue`) use plain `load`/`store`, not `compare_exchange`.

cfreeamd (commented):

Reviewed the lock-free MPSC queue changes in amd_gpu_agent.h/.cpp. The memory ordering on the CAS/load pairing and the CAPACITY=64 wraparound alignment look correct. Left inline notes on a liveness concern in enqueue(), a design question about the unused payload, and a small comment fix.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -4866,22 +4925,31 @@ void GpuAgent::PcSamplingConsumerThread(pcs_data_t& pcs_data) {
...
 
       if (pcs_data.consumer_exit.load(std::memory_order_acquire)) {
         break;
       }
 
-      // Try to deliver aggregated samples (threshold check is inside)
+      // Drain the queue to clear notifications
+      uint32_t xcc_id;
+      bool had_data = false;
+      while (pcs_data.flush_queue.try_dequeue(xcc_id)) {
```
cfreeamd, line 4944: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** The dequeued `xcc_id` is discarded here — the queue is drained purely as a wakeup token, and `PcSamplingDeliverAggregatedSamples` scans all XCCs regardless (as the header comment notes). Given the payload isn't consumed, replacing the previous `std::atomic<uint32_t> pending_flush_count` with a 64-slot lock-free ring adds real complexity and the liveness hazard noted in `enqueue()` for no functional gain. Consider keeping a simple atomic flag/counter for the wakeup, or adopting the queue only once delivery actually consumes per-XCC IDs.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -4728,6 +4728,55 @@ hsa_status_t GpuAgent::PcSamplingFlushDeviceBuffersPerXCC_PM4(
...
+  buffer[slot].store(EMPTY_SLOT, std::memory_order_release);
+  read_pos++;
+  xcc_id = val;
+  return true;
+}
+
+bool GpuAgent::pcs_mpsc_queue_t::has_data() const {
+  uint32_t slot = read_pos % CAPACITY;
+  return buffer[slot].load(std::memory_order_acquire) != EMPTY_SLOT;
+}
```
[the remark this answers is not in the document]
vlaindic: I agree that this scenario could happen in theory.

projects/rocr-runtime/runtime/hsa-runtime/core/inc/amd_gpu_agent.h
```diff
@@ -965,6 +965,46 @@ class GpuAgent : public GpuAgentInt {
...
+    // Reset queue to empty state - must be called before starting a new session
+    void reset() {
+      for (uint32_t i = 0; i < CAPACITY; ++i) {
+        buffer[i].store(EMPTY_SLOT, std::memory_order_relaxed);
+      }
+      write_pos.store(0, std::memory_order_relaxed);
+      read_pos = 0;
+    }
+  };
+
```
vlaindic, line 1007: I don't mind leaving it here, but I believe that MPSC lock-free queue could be a type of its own outside of the amd_gpu_agent, as we might decide to reuse it later.

vlaindic (commented):
Ovearll, I don't mind adding this type of queue to the code base, but seems to me like an overkill, as we don't use the XCC_ID to flush particular buffer, and we still iterate over all buffers like in the old approach. 

Also, please consider these AI review comments, some of them might be worth.

# PR #8673 Review Comments
## feat(rocr): PC Sampling: implement lock-free Multi-producer Single-consumer queue

Commit: ab73bc1fd63b1d2edc940220d2b74c9138ca7b72
URL: https://github.com/ROCm/rocm-systems/pull/8673

---

## Comment 1 — CRITICAL
**File:** `projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp`
**Line:** 4738 (the `write_pos.fetch_add` line in `enqueue()`)

```markdown
**[Critical] Enqueue failure leaves a permanent hole — consumer wedged for the rest of the session**

`write_pos.fetch_add(1)` here reserves a slot *before* the CAS loop below publishes the value.
If the CAS loop exhausts its 10 000 attempts and returns `false`, `write_pos` has already advanced
but the slot remains `EMPTY_SLOT` permanently.

`try_dequeue()` / `has_data()` walk slots strictly in `read_pos` order. When the consumer
eventually reaches that unfilled slot it sees `EMPTY_SLOT`, concludes the queue is empty,
and stops draining — even though all later slots may hold valid notifications. One `enqueue`
failure silently wedges the consumer for the remainder of the session (a hang, not graceful
degradation).

Suggested fixes (pick one):

1. **Reserve via slot-CAS instead of `fetch_add`** — don't advance `write_pos` until you know
   the slot is available; only then write `xcc_id` unconditionally.
2. **Make the consumer hole-tolerant** — have `try_dequeue()` skip `EMPTY_SLOT` entries up to
   `write_pos` rather than stopping at the first empty slot.
3. **Drop the ring for a simpler primitive** — see the separate comment on payload usage;
   if the `xcc_id` payload is never consumed, an `std::atomic<uint32_t> pending_count` is
   sufficient and doesn't have this failure mode.
```

---

## Comment 2 — IMPORTANT (was CRITICAL, downgraded per reviewer)
**File:** `projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp`
**Line:** 4775 (the `has_data()` function body)

```markdown
**[Important] `has_data()` races with in-flight producers — consumer can sleep through a notification**

`has_data()` is evaluated inside `consumer_cv.wait()`'s predicate while holding
`consumer_wakeup_mutex`. A producer may have already called `write_pos.fetch_add()` (slot
reserved, `write_pos` advanced) but not yet published the value via CAS. At that moment
`has_data()` reads `EMPTY_SLOT` and returns `false`.

The `condition_variable` predicate contract requires it to be stable from the moment it is
tested until the lock is re-acquired after a wakeup. This queue violates that: the slot
value can flip from `EMPTY_SLOT` to `xcc_id` concurrently with predicate evaluation. The
consumer returns to sleep and misses the notification until the *next* producer fires one.

Under light load (infrequent flushes) this can delay sample delivery noticeably. Under stop/exit
sequences with a single final flush per XCC, the missed wakeup could stall shutdown.

Mitigation: increment a separate `std::atomic<int> pending_count` *before* calling
`notify_one()`, and base the predicate on that counter rather than on the queue's slot state.
This decouples the "something is pending" signal from the CAS publication latency.
```

---

## Comment 3 — IMPORTANT
**File:** `projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp`
**Line:** 4943 (the drain loop: `while (pcs_data.flush_queue.try_dequeue(xcc_id))`)

```markdown
**[Important] Dequeued `xcc_id` is never used — the queue payload carries no information**

The loop drains the queue purely as a wakeup token:

```cpp
uint32_t xcc_id;
bool had_data = false;
while (pcs_data.flush_queue.try_dequeue(xcc_id)) {
    had_data = true;         // xcc_id discarded here
}
PcSamplingDeliverAggregatedSamples(pcs_data, session);  // scans all XCCs unconditionally
```

`PcSamplingDeliverAggregatedSamples` loops `for xcc_id in 0..num_xcc` regardless of which
XCC sent the notification — which is also called out in the header comment:
*"Aggregation/delivery currently still scans all XCCs when building the staging buffer."*

This means a 64-slot CAS-based ring with alignment padding and a liveness hazard (see comment
on `enqueue()`) is doing the work of a boolean flag. The prior `pending_flush_count` atomic was
simpler, correct, and sufficient.

Two clean paths forward:

1. **Keep a simple counter for now** — restore `std::atomic<uint32_t> pending_count` as the
   wakeup signal and open a follow-up to introduce the MPSC queue *once* delivery actually
   consumes per-XCC IDs (selective drain instead of full scan).
2. **Commit to per-XCC delivery** — have the consumer only drain the specific XCCs enqueued,
   eliminating the full scan. That's the real scalability win on MI300X (8+ XCCs), but it's a
   larger change to `PcSamplingDeliverAggregatedSamples`.

As written, the PR description's claim of "scales better with XCC count" is not yet realised.
```

---

## Comment 4 — IMPORTANT
**File:** `projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp`
**Line:** 4803 (first `lock_guard(consumer_wakeup_mutex)` before `notify_one()`)

```markdown
**[Important] Holding `consumer_wakeup_mutex` around `notify_one()` is unnecessary and adds contention on the hot path**

```cpp
{
    std::lock_guard<std::mutex> lock(pcs_data.consumer_wakeup_mutex);
    pcs_data.consumer_cv.notify_one();
}
```

The standard `condition_variable` idiom does *not* require holding the associated mutex when
calling `notify_one()`. The mutex only needs to be held when *modifying the predicate state*,
so that the consumer doesn't miss a transition that happens between predicate evaluation and
`wait()`. Here the predicate state is the queue (updated lock-free), not the mutex — so the
mutex around `notify_one()` buys nothing.

What it costs: every XCC flush thread (up to 8+ on MI300X) serialises through
`consumer_wakeup_mutex` on every notification, turning the "lock-free fast path for producer
threads" stated in the PR description into a mutex round-trip. The same unnecessary lock
appears on line 4834 in the normal flush path.

Fix — remove the `lock_guard` in both places:

```cpp
// Before:
{
    std::lock_guard<std::mutex> lock(pcs_data.consumer_wakeup_mutex);
    pcs_data.consumer_cv.notify_one();
}

// After:
pcs_data.consumer_cv.notify_one();
```

The `consumer_wakeup_mutex` is still needed for the consumer's `unique_lock` inside `wait()`;
removing it from the producer side is safe and is the standard pattern.
```

---

## Comment 5 — IMPORTANT
**File:** `projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp`
**Line:** 4734 (the block comment above the implementation)

```markdown
**[Nit → Important] Implementation comment describes the algorithm backwards**

> *"The queue uses `fetch_add` for concurrent writes and `compare_exchange` for reads."*

This is reversed:
- **Writes** (`enqueue`): `fetch_add` to reserve a slot, then `compare_exchange_weak` to publish.
- **Reads** (`try_dequeue`): plain `load` to check the slot, plain `store` to clear it — no CAS.

Incorrect lock-free algorithm comments are a maintenance hazard: a future reader trying to
understand the memory-ordering guarantees will start from a wrong mental model.

Suggested replacement:

```cpp
// Lock-free MPSC Queue Implementation
// Producers (XCC flush threads) reserve a slot with fetch_add on write_pos, then publish
// the value via compare_exchange_weak (spinning until the slot is EMPTY_SLOT).
// The single consumer reads with a plain load and clears slots with a plain store,
// advancing read_pos strictly in order.
```
```

---

## Comment 6 — IMPORTANT
**File:** `projects/rocr-runtime/runtime/hsa-runtime/core/inc/amd_gpu_agent.h`
**Line:** 6 (copyright header)

```markdown
**[Clarity] Copyright year should include 2026 — substantial new code added in this PR**

The header currently reads:
```
Copyright (c) 2014-2025, Advanced Micro Devices, Inc. All rights reserved.
```

This PR adds ~130 lines of new implementation (the full `pcs_mpsc_queue_t` type and its methods).
For newly added or substantially extended code, the copyright year should be bumped to include
the current year. Please update both `amd_gpu_agent.h` and `amd_gpu_agent.cpp` to:
```
Copyright (c) 2014-2026, Advanced Micro Devices, Inc. All rights reserved.
```
```

---

## Comment 7 — IMPORTANT
**File:** `projects/rocr-runtime/runtime/hsa-runtime/core/inc/amd_gpu_agent.h`
**Line (in PR description, not inline)**

*(Post as a top-level PR comment, not inline)*

```markdown
**[Hygiene] Missing JIRA ID; motivation for the contention claim not substantiated**

The PR has no JIRA ID (`ABC-1234` pattern). The description explains *what* changed but not
*why* `pending_flush_count` was a measured problem. A reviewer cannot assess whether the
complexity trade-off (CAS ring + liveness hazard + missed-wakeup race) is warranted without
knowing:

- The workload / scenario where contention was observed.
- Whether this was a measured bottleneck or a pre-emptive optimisation.

Please add either the JIRA ID or a short paragraph: what you observed, at what scale
(XCC count, flush rate), and what metric improved. If the context is internal-only, a one-liner
on the PR pointing reviewers to the ticket is sufficient.
```

---

## Posting Instructions

```bash
COMMIT=ab73bc1fd63b1d2edc940220d2b74c9138ca7b72
REPO=ROCm/rocm-systems
PR=8673

# Comment 1 — enqueue hole (line 4738, cpp)
gh api repos/$REPO/pulls/$PR/comments \
  -f body='...' \
  -f commit_id=$COMMIT \
  -f path='projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp' \
  -F line=4738 \
  -f side=RIGHT

# Comment 2 — has_data race (line 4775, cpp)
# Comment 3 — payload unused (line 4943, cpp)
# Comment 4 — notify_one under lock (line 4803, cpp)
# Comment 5 — comment reversed (line 4734, cpp)
# Comment 6 — copyright year (line 6, header)

# Comment 7 — PR-level (no file/line)
gh api repos/$REPO/issues/$PR/comments \
  -f body='...'
```

PR #8675, opened 2026-07-16
feat(rccl): CE Allreduce with pipeline (gfx1250)

AbandiGa:
## Motivation

Introduces a pipelined CE-based AllReduce path that enables large-message reductions without being constrained by the 256 MiB temporary buffer. The operation is decomposed into scatter, local reduction, and allgather phases, using double-buffered slots (NCCL_CE_NUM_SLOTS=2) and two streams to overlap Copy Engine transfers with the persistent reduction kernel. To test with messages grater than 256 MiB -R 2 -x 2 need to be used.

## JIRA ID
AICOMRCCL-909

## Submission Checklist


projects/rccl/src/device/ce_reduce.cc
```diff
@@ -26,9 +26,50 @@ THE SOFTWARE.
...
+  uint32_t* doneGen = d_barrierSync + 1;
+  __syncthreads();
+  if (threadIdx.x == 0) {
+    uint32_t prev = __hip_atomic_fetch_add(arrival, 1, __ATOMIC_RELAXED, __HIP_MEMORY_SCOPE_AGENT);
+    if (prev + 1 == (uint32_t)numBlocks) {
+      __hip_atomic_store(arrival, 0, __ATOMIC_RELAXED, __HIP_MEMORY_SCOPE_AGENT);
+      __hip_atomic_store(doneGen, gen, __ATOMIC_RELEASE, __HIP_MEMORY_SCOPE_AGENT);
+    } else {
+      while (__hip_atomic_load(doneGen, __ATOMIC_ACQUIRE, __HIP_MEMORY_SCOPE_AGENT) != gen) {
+        __builtin_amdgcn_s_sleep(1);
```
alex-breslow-amd, line 56: The release and acquire semantics at the system scope are going to introduce hardware cache writebacks and invalidations on gfx950.  If there's a way to get by with relaxed atomics at the device scope, I'd recommend it.  It should help performance.  This is an optimization obviously, not a blocker.

projects/rccl/src/device/ce_reduce.cc
```diff
@@ -77,99 +118,196 @@ template<typename T> struct ReduceOp<T, 3> {
...
+    const size_t currentChunkElems = isTailChunk ? tailChunkElems : baseChunkElems;
+    const size_t nVec = currentChunkElems / W;
+    if (totalSteps > 1) {
+      if (blockIdx.x == 0 && threadIdx.x == 0) {
+          while (true) {
+            int ready = 0;
+            for (int r = 0; r < nRanks; r++) {
+                uint32_t sv = __hip_atomic_load(
+                    (uint32_t*)&signalBuffer[(size_t)slot * nRanks + r],
+                    __ATOMIC_ACQUIRE, __HIP_MEMORY_SCOPE_SYSTEM);
```
alex-breslow-amd, line 165: The acquire is cheaper than release, but it still causes a whole cache invalidation on gfx950 at the system and device scopes.  If there's any way to avoid this, I'd recommend it.  

If you're loading from the system scope, I don't think you need the acquire.  The barrier in old-school RCCL doesn't do an acquire IIRC.

projects/rccl/src/device/ce_reduce.cc
```diff
@@ -77,99 +118,196 @@ template<typename T> struct ReduceOp<T, 3> {
...
-  const size_t tailBase = nVec * W;
-  const size_t tailLen  = chunkElems - tailBase;
-  for (size_t i = tid; i < tailLen; i += stride) {
-    T a = in[tailBase + i];
-    for (int r = 1; r < nRanks; r++) {
-      T v = in[(size_t)r * chunkElems + tailBase + i];
-      a = ReduceOp<T, RedOp>::apply(a, v);
+  (void)hipGetLastError();
+    switch (redOp) {
+      case 0: {
```
alex-breslow-amd, line 260: This is gross, but if we inherited it from NCCL, it's probably safer to keep it I guess.  My preference would be putting the actual enum names.

nileshnegi (approved):

PR #8679, opened 2026-07-16
feat(hrr): [AIRUNTIME-2530] capture and replay driver memcpy APIs

andraghetti:
## Motivation

`hipDrvMemcpy3D`, `hipDrvMemcpy3DAsync` and `hipDrvMemcpy2DUnaligned` were no-op on replay. Their `HIP_MEMCPY3D` / `hip_Memcpy2D` argument is a struct *pointer*, stale at playback, so capture recorded only the pointer and the destination buffer was never reproduced.

## Technical Details

Moves the three APIs from no-op playback to manual capture plus manual playback, mirroring `hipMemcpy3D`. Branching is keyed off `srcMemoryType` / `dstMemoryType`, as these APIs have no `kind` field.

**Capture** (`hip_capture.cpp`, plus regenerated `hrr_api_args.h` and dispatch)

- Inlines the struct into the event (`drv3d_bytes[192]` / `drv2d_bytes[136]`, `static_assert`ed against the real size) and snapshots the H2D source or D2H expected blob.
- Sizes the H2D blob by the pitched host rect footprint, not the `WidthInBytes * Height * Depth` volume. Replay keeps the recorded pitches and offsets and the runtime does not size-validate the host side, so a volume-sized blob is read past its end whenever the rect is not dense. It only ever grows a blob.
- Records only successful calls, as the generated shims did, since any non-success handler return aborts the whole replay.

**Playback** (`hip_playback.cpp`)

- Reconstructs the struct, substitutes the blob for `srcHost`, translates the embedded `srcDevice` / `dstDevice` via the alloc map, and drains H2D with `hrr_sync_after_replayed_h2d`.
- Skips instead of substituting when the blob is missing or shorter than the recorded rect, so pre-fix archives cannot over-read. Host-to-host is an explicit no-op.
- **Declines array-typed rects unconditionally**: if either memory type is `hipMemoryTypeArray`, warn once and return without copying. An array is addressed through `srcArray` / `dstArray` with the matching `srcDevice` / `dstDevice` unset, which no branch below can describe.
- Diagnostics take a `tag` naming the calling API instead of hard-coding `"3D"`.

**Newest commit, default-stream sync for async D2H, both sides.** `hipDrvMemcpy3DAsync(p, nullptr)` is still asynchronous and a recorded default stream translates to `nullptr`. Capture now syncs before snapshotting `dstHost` even on the null stream, dropping the blob if that sync fails, and `replay_memcpy3d_d2h` syncs unconditionally and propagates the error.

**Scope warning:** `replay_memcpy3d_d2h` is shared, so the replay half of that change also affects `hipMemcpy3DAsync` and its `_spt` alias, beyond the driver APIs. A `hipMemcpy3DAsync` D2H check on a recorded default stream previously skipped the sync and could race its own readback.

**Tests**

- `Unit_HRR_DrvMemcpy3D_Direct` was zero-effect self-copies. It is now a real chain (H2D, async D2D, D2D) validated by D2H, plus a pitched two-slice H2D rect whose inter-row padding must survive. New `Unit_HRR_DrvMemcpy2DUnaligned_Direct` and its roundtrip do the same for 2D. Two `hrr.yaml` entries.
- New `hrr_run_exact_roundtrip()`: both roundtrips replay with `HIP_HRR_D2H_EXACT=1` and require a validated D2H blob. Windows keeps `hrr_run_playback`'s best-effort D2H policy.
- Canary `0x2D2D2D2D` becomes `0x5C5C5C5C`. The old value decodes to 9.8e-12 as f32 and passed the validator's `atol=rtol=1e-3` fallback against replay's zero-initialised allocations, so the 2D roundtrip could not fail.
- The `hipMemcpy3DPeer` / `hipMemcpy3DPeerAsync` calls (the only ones in the HRR test tree) move onto a scratch buffer that feeds no D2H check, since both are no-op on playback.

**Known gap, not fixed here.** The D2H expected blob is still sized by the flat volume, because replay reads the device side back flat and never substitutes `dstHost`. Safe, but a pitched *destination* rect is not validated faithfully; tracked separately.

**Merge order.** #8849 also adds a `hrr_run_exact_roundtrip()` with a different body; whichever merges second reconciles the two.

## Issue Tracking

JIRA ID: AIRUNTIME-2530

## Test Plan

Capture each Direct workload, replay under `HIP_HRR_D2H_EXACT=1`, require exit 0 and a validated D2H blob; as a negative control, stub the three playback handlers to `return hipSuccess` and replay the same archives; then run the full HRR unit suite on a gfx950 host.

## Test Result

Verified on a gfx950 host at this revision, with the array-decline and default-stream-sync commits included: the full HRR unit suite passes, 19/19 in the GPU-free pass and 44/44 in the GPU pass, with `Unit_HRR_MemcpyPeerRoundtrip` skipped as it needs a second GPU. `Unit_HRR_DrvMemcpy3DRoundtrip`, `Unit_HRR_DrvMemcpy2DUnalignedRoundtrip` and `Unit_HRR_DrvMemcpyRoundtrip` all pass.

The coverage is not vacuous, and this was demonstrated rather than argued: with the three driver-memcpy playback handlers stubbed out and rebuilt, the same archives give `0 pass, 2 fail` and both new roundtrips fail. Blob sizing is confirmed by artifact, 1856- and 832-byte blobs on disk against 512 and 256 under the old naive volume, and none of the three APIs emits a NOOP warning any more.

The array-decline branch is exercised by patching a recorded descriptor's memory type in a throwaway archive, which shows the branch runs and drops the copy without counting a failure. A real array cannot be created on this device, which reports no image support, so that a *mapped* array reaches the branch remains a source-level property.

## Submission Checklist


projects/clr/hipamd/src/hrr/hip_capture.cpp
```diff
@@ -1193,6 +1193,141 @@ hipError_t capture_hipMemcpy3DAsync_spt(const struct hipMemcpy3DParms* p, hipStr
...
+    if (n > 0) {
+      if (is_async && stream) {
+        hipError_t sync_r = g_real_table.hipStreamSynchronize_fn(stream);
+        if (sync_r != hipSuccess) {
+          LogPrintfWarning("[HRR capture] hipStreamSynchronize failed (%d): D2H drv blob skipped",
+                           sync_r);
+          hrr_cap::writer::write_event_raw(api_id, &a.hdr, sizeof(a));
+          return;
+        }
+      }
```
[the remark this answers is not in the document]
andraghetti: Applied in `8a019fd`. The guard was `if (is_async && stream)`, so an async driver memcpy issued on the default stream (`stream == nullptr`) was never synchronized, and `dstHost` could be snapshotted before the copy landed and recorded as a stale expected output. `hipStreamSynchronize_common()` handles `nullptr` explicitly by syncing the blocking streams and returning success, so the guard bought nothing; the sync is now unconditional on the async path, which is what the `hipMemcpyAsync` and `hipMemcpyWithStream` capture paths already do.

The same guard on the replay side, in `replay_memcpy3d_d2h()` (raised as a suppressed comment in the review body, so it has no thread of its own), is fixed in the same commit, and there the sync status is now propagated instead of discarded so a device failure is not reported as a data mismatch. That function is shared, so the `hipMemcpy3D` / `hipMemcpy3DAsync` readback becomes reliable too.

Left in place: the identical guard in the capture paths for `hipMemcpy3DAsync` and `hipMemcpy2DAsync`. Same bug, but those APIs are outside this PR and a capture-side sync changes shipped behaviour, so they belong in a separate change. Neither change has been re-run on hardware yet.

gandryey (approved):

projects/clr/hipamd/src/hrr/playback/hip_playback.cpp
```diff
@@ -2323,22 +2323,30 @@ hipError_t playback_hipStreamQuery_spt(PlaybackContext& ctx, const uint8_t* pl)
...
+        // dead device from being reported as a data mismatch. Both match the
+        // sibling 2D path in replay_memcpy2d().
+        if (r == hipSuccess) r = hipStreamSynchronize(stream);
     } else {
         r = hipMemcpy(actual.data(), src_live, byte_count, hipMemcpyDeviceToHost);
     }
     if (r != hipSuccess) {
-        fprintf(stderr, "[HRR] hipMemcpy3D D2H: device readback failed: %d (%s)\n",
-                r, hipGetErrorString(r));
+        fprintf(stderr, "[HRR] %s D2H: device readback failed: %d (%s)\n",
```
chrispaquot, line 2349: Tag default value is 3D not hipMemcopy3D. Should you add hipMemcopy%s instead?

danielbolsson: Good catch, and the regression is real: before this PR the line read `hipMemcpy3D D2H: device readback failed`, and with the `"3D"` default the two pre-existing callers, `playback_hipMemcpy3D` and `playback_hipMemcpy3DAsync`, now both print `3D D2H: ...`. That is less than the log said before, it makes the sync and async variants indistinguishable, and it disagrees with the `src ... not mapped` message a few lines up in those same two functions, which prints the full API name.

I went with a slightly different fix than `hipMemcpy%s`, because the prefix cannot be baked into the format string here. `tag` is not always a suffix: the driver handlers pass whole API names through it, `"hipDrvMemcpy3D"` and `"hipDrvMemcpy3DAsync"` from `replay_drvmemcpy3d`, and `"hipDrvMemcpy2DUnaligned"` / `"hipMemcpyParam2D"` / `"hipMemcpyParam2DAsync"` from `replay_drvmemcpy2d`, so `hipMemcpy%s` would render `hipMemcpyhipDrvMemcpy3D`. `tag` is also forwarded to `hrr_d2h_validate()`, which has its own format strings, so prefixing this one `fprintf` would leave the readback-failure message and the validation message disagreeing about the API name.

Instead the default is dropped and both call sites pass their own name, `"hipMemcpy3D"` and `"hipMemcpy3DAsync"`. That restores the old message, separates sync from async, matches what the new driver handlers already do, and removes a silent default that the next caller could inherit by accident. Diagnostics only, and nothing parses these lines: the tests read the D2H summary through `hrr_parse_d2h_summary()`. Landing in the next push to the branch.

PR #8680, opened 2026-07-16
feat(clr): add metadata in HRR capture stage

LeoZhao-BJ:
## Motivation

feat: Add HRR capture metadata to manifests

## Technical Details

 Summary
  - Added HRR metadata collection for capture archives.
  - Metadata is written and embedded in per-process manifest.json.
  - Added Unit_HRR_MetadataManifest coverage in hip-tests.

  Details
  - Captures runtime metadata:
      - hip_runtime_version
      - comgr version via amd_comgr_get_version
  - Captures device metadata from initialized HIP internal state:
      - device count from hip::g_devices
      - selected hipGetDeviceProperties fields via hip::ihipGetDeviceProperties
  - Avoids public HIP API calls during hip_capture_init() to prevent re-entering HIP initialization.
  - Keeps metadata best-effort and separate from HRR event stream semantics.

JIRA ID: AIRUNTIME-2509

## Test Plan

./HrrTest 'Unit_HRR_MetadataManifest' -s

## Test Result

all passed

## Submission Checklist

### DIFF
--- a/projects/clr/hipamd/src/hrr/hip_capture_metadata.cpp
+++ b/projects/clr/hipamd/src/hrr/hip_capture_metadata.cpp
@@ -14,7 +14,11 @@
 #include <string>
 #include <vector>
 
-#include "../hip_internal.hpp"  // hip::g_devices, hip::ihipGetDeviceProperties
+namespace hip {
+class Device;
+extern std::vector<hip::Device*> g_devices;
+extern hipError_t ihipGetDeviceProperties(hipDeviceProp_t* props, hipDevice_t device);
+}  // namespace hip
 
 namespace hrr_cap {
 namespace metadata {
@@ -118,10 +122,8 @@ std::string collect_comgr_json() {
   amd_comgr_get_version(&major, &minor);
   std::ostringstream os;
   os << "{\n"
-     << "    \"available\": true,\n"
      << "    \"major\": " << static_cast<unsigned long long>(major) << ",\n"
-     << "    \"minor\": " << static_cast<unsigned long long>(minor) << ",\n"
-     << "    \"source\": \"amd_comgr_get_version\"\n"
+     << "    \"minor\": " << static_cast<unsigned long long>(minor) << "\n"
      << "  }";
   return os.str();
 }
--- a/projects/clr/hipamd/src/hrr/hip_capture_metadata.cpp
+++ b/projects/clr/hipamd/src/hrr/hip_capture_metadata.cpp
@@ -136,33 +136,52 @@ std::string collect_runtime_json() {
   return os.str();
 }
 
-std::string collect_devices_json() {
-  std::ostringstream os;
-  os << "[";
+struct DeviceMetadata {
+  std::string devices_json;
+  std::string errors_json;
+  int captured_count = 0;
+};
 
-  const int count = static_cast<int>(hip::g_devices.size());
+DeviceMetadata collect_device_metadata() {
+  std::ostringstream devices;
+  std::ostringstream errors;
+  devices << "[";
+  errors << "[";
 
-  for (int device = 0; device < count; ++device) {
-    if (device > 0) os << ",";
-    os << "\n"
-       << "    {\n"
-       << "      \"ordinal\": " << device << ",\n";
+  const int device_count = static_cast<int>(hip::g_devices.size());
+  bool first_device = true;
+  bool first_error = true;
+  int captured_count = 0;
 
+  for (int device = 0; device < device_count; ++device) {
     hipDeviceProp_t prop{};
     const hipError_t prop_err = hip::ihipGetDeviceProperties(&prop, device);
-    if (prop_err == hipSuccess) {
-      append_prop_fields(os, prop);
-    } else {
-      os << "      \"properties_error\": " << quote(hip_error_name(prop_err)) << "\n";
+    if (prop_err != hipSuccess) {
+      if (!first_error) errors << ",";
+      first_error = false;
+      errors << "\n"
+             << "    { \"ordinal\": " << device
+             << ", \"api\": \"ihipGetDeviceProperties\""
+             << ", \"error\": " << quote(hip_error_name(prop_err)) << " }";
+      continue;
     }
 
-    os << "\n"
-       << "    }";
+    if (!first_device) devices << ",";
+    first_device = false;
+    ++captured_count;
+    devices << "\n"
+            << "    {\n"
+            << "      \"ordinal\": " << device << ",\n";
+    append_prop_fields(devices, prop);
+    devices << "\n"
+            << "    }";
   }
 
-  os << "\n"
-     << "  ]";
-  return os.str();
+  devices << "\n"
+          << "  ]";
+  errors << "\n"
+         << "  ]";
+  return {devices.str(), errors.str(), captured_count};
 }
 
 }  // namespace
@@ -175,9 +194,11 @@ std::string Metadata::collect_json() const {
 
   int count = 0;
   count = static_cast<int>(hip::g_devices.size());
+  DeviceMetadata device_metadata = collect_device_metadata();
   os << "  \"device_count\": " << count << ",\n";
-
-  os << "  \"devices\": " << collect_devices_json() << "\n"
+  os << "  \"captured_device_count\": " << device_metadata.captured_count << ",\n";
+  os << "  \"device_metadata_errors\": " << device_metadata.errors_json << ",\n";
+  os << "  \"devices\": " << device_metadata.devices_json << "\n"
      << "}";
   return os.str();
 }
--- a/projects/clr/hipamd/src/hrr/hip_capture_metadata.cpp
+++ b/projects/clr/hipamd/src/hrr/hip_capture_metadata.cpp
@@ -14,11 +14,7 @@
 #include <string>
 #include <vector>
 
-namespace hip {
-class Device;
-extern std::vector<hip::Device*> g_devices;
-extern hipError_t ihipGetDeviceProperties(hipDeviceProp_t* props, hipDevice_t device);
-}  // namespace hip
+#include "../hip_internal.hpp"  // hip::g_devices, hip::ihipGetDeviceProperties
 
 namespace hrr_cap {
 namespace metadata {
--- a/projects/clr/hipamd/src/hrr/hip_capture_metadata.cpp
+++ b/projects/clr/hipamd/src/hrr/hip_capture_metadata.cpp
@@ -10,6 +10,7 @@
 #include "hip/hip_runtime_api.h"
 
 #include <cstdio>
+#include <cstring>
 #include <sstream>
 #include <string>
 #include <vector>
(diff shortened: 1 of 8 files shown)


projects/clr/hipamd/src/hrr/hip_capture_metadata.cpp
```diff
@@ -0,0 +1,194 @@
...
+     << "  }";
+  return os.str();
+}
+
+std::string collect_runtime_json() {
+  std::ostringstream os;
+  os << "{\n";
+
+  const int hip_version = HIP_VERSION;
+  os << "    \"hip_runtime_version\": " << quote(version_string_from_int(hip_version)) << ",\n"
```
akshayv, line 135: These fields currently have the same info. should they be looking at different fields? or should we remove one of them?

projects/clr/hipamd/src/hrr/hip_capture_metadata.cpp
```diff
@@ -0,0 +1,194 @@
...
+     << "      }";
+}
+
+std::string collect_comgr_json() {
+  size_t major = 0;
+  size_t minor = 0;
+  amd_comgr_get_version(&major, &minor);
+  std::ostringstream os;
+  os << "{\n"
+     << "    \"available\": true,\n"
```
akshayv: This doesnt match the comment in the header: // Returns a complete JSON object. The object is always valid JSON; individual
  // sections carry an "available": false marker when optional probes fail.

Should we have a path where the probes fail?

akshayv: Alternatively we can remove the available field

projects/clr/hipamd/src/hrr/hip_capture_metadata.cpp
```diff
@@ -0,0 +1,194 @@
...
+#include <string>
+#include <vector>
+
+#include "../hip_internal.hpp"  // hip::g_devices, hip::ihipGetDeviceProperties
+
+namespace hrr_cap {
+namespace metadata {
+namespace {
+
+std::string json_escape(const char* s) {
```
akshayv, line 28: One thing worth considering is brining in a json library dependecy like nlohmann_json (which is used in a few places in rocm_systems) so we can get rid of some of this json parsing/formatting code.. i'm not sure if this is acceptable to the hip team members.. maybe @gandryey  can comment?

projects/clr/hipamd/src/hrr/hip_capture_writer.cpp
```diff
@@ -481,14 +496,20 @@ static void write_manifest_stdio(const char* output_dir, bool complete) {
...
           "  \"event_count\": %llu,\n"
-          "  \"blob_count\": %llu\n"
-          "}\n",
+          "  \"blob_count\": %llu",
           static_cast<unsigned long long>(g_pid),
           static_cast<unsigned long long>(g_parent_pid),
           complete ? "true" : "false",
           static_cast<unsigned long long>(g_event_count.load()),
           static_cast<unsigned long long>(g_blob_count.load()));
+  if (is_json_object_fragment(g_metadata_json)) {
```
akshayv: Since we are already writing the metadata file right next to this one. do we need this data in manifest.json as well? could we skip this field in manifest.json and restrict to the metadata file?

projects/clr/hipamd/src/hrr/hip_capture_writer.cpp
```diff
@@ -557,27 +601,50 @@ static void update_root_manifest() {
     const std::string name = ent.path().filename().string();
     if (name.rfind("pid-", 0) != 0) continue;
     ProcessManifestEntry entry{};
-    if (read_process_manifest((ent.path() / "manifest.json").string(), &entry))
+    if (read_process_manifest((ent.path() / "manifest.json").string(), &entry)) {
```
akshayv: Personally I'm not sure if it helps much to move this into the root folder.. this function gets called when the cpature is a success and the workload is successfully complete and in most cases that wont happen.. so we cant rely on anything this function does in the playback.. 

i think its good enough if the metadata.json is in the pid-*/ folders, it also gives additional flexibility in case somehow the metadata is different between the processes that are running

akshayv:
Looks good to me! some minor comments. 

We probably need a followup for the playback to use this info, but that can come later.

perhaps @gandryey  can have a look as well? thanks!

projects/clr/hipamd/src/hrr/hip_capture_metadata.cpp
LeoZhao-BJ: here avoid json lib dependency intentionally, if we have more usage on json parsing, yes it is better to link lib.

projects/clr/hipamd/src/hrr/hip_capture_metadata.cpp
LeoZhao-BJ: yes, it is same on AMD platform, but if apply to nv platform, these 2 field could be different. 
if we only target AND plaform, could just keep one, e.g. hip_runtime_version.

projects/clr/hipamd/src/hrr/hip_capture_metadata.cpp
LeoZhao-BJ: initial intention is to use comgr dynamic interface, but it exposes more underneath thing, now it uses direct link to call comgr api. yes, could probably remove this part.

projects/clr/hipamd/src/hrr/hip_capture_metadata.cpp
akshayv: Thanks Leo, just both fields use the exact same call though: version_string_from_int(hip_version).. even for NV, this would give the same results? or maybe i'm wrong?

projects/clr/hipamd/src/hrr/hip_capture_writer.cpp
LeoZhao-BJ: so far, this info is stored in both pid and root manifest, ideally we just need one piece, but not sure if there is case for different device?

projects/clr/hipamd/src/hrr/hip_capture_metadata.cpp
LeoZhao-BJ: could be different on NV platform, see here.
https://rocm.docs.amd.com/projects/HIP/en/latest/reference/hip_runtime_api/modules/initialization_and_version.html

projects/clr/hipamd/src/hrr/hip_capture_writer.cpp
LeoZhao-BJ: yes, duplicated, will remove metadata file, just keep change in manifest.json

projects/clr/hipamd/src/hrr/hip_capture_metadata.cpp
LeoZhao-BJ: removed

projects/clr/hipamd/src/hrr/hip_capture_writer.cpp
akshayv: i think its better in the pid.. most cases i think it wont get copied to the root since this currently used for crashing/failing workloads

LeoZhao-BJ: ok, let me remove root part tomorrow.

projects/clr/hipamd/src/hrr/hip_capture_metadata.cpp
LeoZhao-BJ: remove hip_driver_version, just keep hip_runtime_version

projects/clr/hipamd/src/hrr/hip_capture_writer.cpp
```diff
@@ -481,13 +489,19 @@ static void write_manifest_stdio(const char* output_dir, bool complete) {
...
-          "  \"blob_count\": %llu\n"
-          "}\n",
+          "  \"blob_count\": %llu",
           static_cast<unsigned long long>(g_pid),
           static_cast<unsigned long long>(g_parent_pid),
           complete ? "true" : "false",
           static_cast<unsigned long long>(g_event_count.load()),
           static_cast<unsigned long long>(g_blob_count.load()));
+  const std::string metadata_json = hrr_cap::metadata::collect_json();
+  if (is_json_object_fragment(metadata_json)) {
```
akshayv: So we are now writing in function write_manifest_stdio which is called on a clean shutdown.. there is another function called emergency_finalize which is called when the process crashes due to GPU errors. i think this needs to be written there as well, otherwise in many cases we wont have this metadata. Can we write it there as well?

LeoZhao-BJ: ok, let me add it

LeoZhao-BJ: done.

akshayv (approved):

gandryey:
Summary is below and i will send .md file with more details

A. Silent failures | ⚠️ | device_count and devices array read g_devices.size() twice; dead int count = 0 init
B. Test depth | ⚠️ | Level 2 (shape). json_has_key is a substring match — does not detect field removal or renaming
C. Coverage honesty | ⚠️ | Emergency-finalizer (crash) path completely untested
D. Binary safety | ✅ | N/A
E. Documentation | ✅ | Accurate
F. Cross-cutting | ⚠️ | json_escape builds a heap copy of input to iterate it; spurious cast in collect_comgr_version; lockless read of g_metadata_json in crash handler has no explaining comment
G. Testability | ⚠️ | All tests require GPU subprocess; pure JSON logic not unit-testable

gandryey:
<html>
<body>
<h2 data-line="4" class="code-line" dir="auto" id="summary" style="margin-top: 24px; font-weight: 600; margin-bottom: 16px; line-height: 1.25; font-size: 1.5em; padding-bottom: 0.3em; border-bottom: 1px solid rgba(255, 255, 255, 0.18); border-top-color: rgba(255, 255, 255, 0.18); border-right-color: rgba(255, 255, 255, 0.18); border-left-color: rgba(255, 255, 255, 0.18); color: rgb(212, 212, 212); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe WPC&quot;, &quot;Segoe UI&quot;, system-ui, Ubuntu, &quot;Droid Sans&quot;, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Summary</h2>
Category | Status | Key findings
-- | -- | --
A. Silent failures | ⚠️ | device_count and devices array read g_devices.size() twice; dead int count = 0 init
B. Test depth | ⚠️ | Level 2 (shape). json_has_key is a substring match — does not detect field removal or renaming
C. Coverage honesty | ⚠️ | Emergency-finalizer (crash) path completely untested
D. Binary safety | ✅ | N/A
E. Documentation | ✅ | Accurate
F. Cross-cutting | ⚠️ | json_escape builds a heap copy of input to iterate it; spurious cast in collect_comgr_version; lockless read of g_metadata_json in crash handler has no explaining comment
G. Testability | ⚠️ | All tests require GPU subprocess; pure JSON logic not unit-testable


</body>
</html>

gandryey (approved):

PR #8686, opened 2026-07-16
[rocr-runtime][trap handler] Refactors the GFX9 PC sampling trap handler to mirror the structure already established in trap_handler_gfx12.s

vlaindic:
Description:

  Refactors the GFX9 PC sampling trap handler to mirror the structure already established in trap_handler_gfx12.s.

  Changes:

  - Hoist the buffer entry address calculation (&bufferX[local_entry]) and s_memrealtime timestamp read above the hosttrap/stochastic branch — these are common to both sample types and were previously duplicated in each label.
  - For GFX9.4+, also hoist all common sample field stores (timestamp, EXEC mask, workgroup IDs, wave_in_wg + chiplet, HW_ID) above the branch, so each sample-type label only handles what is unique to it:
    - .fill_sample_hosttrap: stores wave PC at offset 0x00
    - .fill_sample_stochastic: stores snapshot PC at 0x00 and snapshot DATA/DATA1 at 0x24
  - Remove the FILL_SAMPLE_COMMON_GFX94 macro, which had a single call site, and inline its body directly.

  The GFX9.0–9.3 path is unchanged in behaviour; it still stores common fields inline before branching to .fill_sample_hosttrap (stochastic sampling does not exist on those generations).

No functional change.

## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/trap_handler/trap_handler.s
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/trap_handler/trap_handler.s
@@ -206,6 +206,66 @@
   .endif
  .endm
 
+ // ============================================================================
+ // GFX9.4+ PC Sampling Refactored Macros
+ // ============================================================================
+ // These macros unify duplicated code between hosttrap and stochastic paths.
+
+ // CALC_XCC_OFFSET_GFX94 - Calculate per-XCC buffer address for multi-XCC systems
+ // Input:
+ //   ttmp[2:3] = buffer base address
+ //   ttmp4     = per_xcc_size (stride per XCC)
+ //   ttmp5     = XCC_ID
+ // Output:
+ //   ttmp[14:15] = buffer_base + (XCC_ID * per_xcc_size)
+ // Clobbers: ttmp[14:15] (output), uses ttmp4, ttmp5
+ .macro CALC_XCC_OFFSET_GFX94
+  .if (.amdgcn.gfx_generation_minor >= 4)
+    s_mul_i32                             ttmp14, ttmp4, ttmp5            // ttmp14 = lo32(offset)
+    s_mul_hi_u32                          ttmp15, ttmp4, ttmp5            // ttmp15 = hi32(offset)
+    s_add_u32                             ttmp14, ttmp2, ttmp14           // ttmp14:15 = base + offset
+    s_addc_u32                            ttmp15, ttmp3, ttmp15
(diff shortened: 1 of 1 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/trap_handler/trap_handler.s
```diff
@@ -485,8 +394,26 @@ trap_entry:
   //  }
   //}
 
-  // ttmp[14:15] is tma->host_trap_buffers; Available: ttmp[2:3], ttmp[4:5], ttmp7, ttmp13
+  // ttmp[14:15] is tma->host_trap_buffers; ttmp[2:3] is buffer base (for NULL check); Available: ttmp[4:5], ttmp7, ttmp13
 .profile_trap_handlers_gfx9:
+  // Exit immediately if the buffer pointer is NULL (sampling not configured for this trap type).
+  s_cmp_eq_u64                          ttmp[2:3], 0
```
cfreeamd, line 400: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** `ttmp[2:3]` is populated in `trap_entry` via an asynchronous `s_load_dwordx2`. In the original code the gfx9.4+ hosttrap and stochastic arms each did `s_waitcnt lgkmcnt(0)` before testing `s_cmp_eq_u64 ttmp[2:3], 0`. After this refactor the NULL check moved here, but on the gfx9.4+ path the `s_waitcnt lgkmcnt(0)` now sits *after* this comparison (just before the per-XCC multiply). So this `s_cmp_eq_u64` can read `ttmp[2:3]` before the scalar load retires — a racy NULL test that can either falsely `.exit_trap` (dropped sample) or falsely pass into the offset math. The gfx9.0-9.3 arm is fine because its `trap_entry` still waits before branching. Please add an `s_waitcnt lgkmcnt(0)` before this `s_cmp_eq_u64` (or restore it on the gfx9.4+ `trap_entry` arms).

cfreeamd (commented):

Reviewed the GFX9 trap-handler consolidation. The control flow for both single-XCC (gfx9.0-9.3) and multi-XCC (gfx9.4+) paths, the shared common-field store, and the field offsets all look consistent with the pre-refactor code. Left one inline note on a dropped `s_waitcnt` on the gfx9.4+ path. This is a hardware trap path not covered by ordinary CI — worth validating on real gfx94x + a single-XCC gfx9 part (or FFM) before merge.

PR #8687, opened 2026-07-16
test(rccl): add CPU affinity restore regression test

amd-arozanov:
## Motivation

initTransportsRank() must restore the CPU affinity mask it saved before applying the GPU-local one. A regression from 2.29.2 re-applied the temporary comm->cpuAffinity instead of the saved affinitySave, leaking the GPU-local affinity into the caller thread (NCCL issue 2033, fixed upstream in 2.30.3). 

## Technical Details

Add an isolated single-rank gtest that widens the thread affinity, runs ncclCommInitRank (which executes initTransportsRank on the calling thread), and asserts the mask is unchanged afterwards. Register it in the mi300x_mellanox_ib test runner config.

## JIRA ID

AICOMRCCL-1537

## Test Plan

Validated on MI350X (gfx950):

## Test Result

Passes with the current state of develop branch, fails with the version without the fix 

[ RUN      ] CpuAffinity.RestoredAfterInitRank
[       OK ] CpuAffinity.RestoredAfterInitRank (3551 ms)

## Submission Checklist

### DIFF
--- a/projects/rccl/test/CpuAffinityTests.cpp
+++ b/projects/rccl/test/CpuAffinityTests.cpp
@@ -1,0 +1,81 @@
+/*************************************************************************
+ * Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * See LICENSE.txt for license information
+ ************************************************************************/
+
+#include <gtest/gtest.h>
+#include <rccl/rccl.h>
+
+#include <sched.h>
+#include <unistd.h>
+#include <algorithm>
+#include <cerrno>
+#include <cstring>
+
+#include "StandaloneUtils.hpp"
+#include "common/ProcessIsolatedTestRunner.hpp"
+
+namespace RcclUnitTesting
+{
+  /**
+   * \brief Verify the calling thread's CPU affinity is restored after ncclCommInitRank.
+   *
+   * Regression guard for NCCL issue #2033 / AICOMRCCL-1537: initTransportsRank() must
+   * restore the mask it saved before applying the GPU-local one. Single-rank
+   * ncclCommInitRank runs initTransportsRank on the calling thread, so a leak is
+   * observable here. The check only bites when the GPU-local CPU set is a strict
+   * subset of the process mask (multi-NUMA hosts); elsewhere the invariant still holds.
+   * ******************************************************************************************/
+  TEST(CpuAffinity, RestoredAfterInitRank)
+  {
+    RUN_ISOLATED_TEST("CpuAffinity_RestoredAfterInitRank", []()
+    {
+      int numDevices;
+      HIPCALL(hipGetDeviceCount(&numDevices));
+      if (numDevices < 1) {
+        GTEST_SKIP() << "No devices available.";
+      }
+
+      // Widen to the full online CPU set so the GPU-local subset is more likely to be
+      // strictly narrower, then record what the kernel actually granted.
+      long numCpus = sysconf(_SC_NPROCESSORS_ONLN);
+      ASSERT_GT(numCpus, 0);
+
+      cpu_set_t fullMask;
+      CPU_ZERO(&fullMask);
+
+      int maxCpus = static_cast<int>(std::min<long>(numCpus, CPU_SETSIZE));
+      for (int cpu = 0; cpu < maxCpus; ++cpu) {
+        CPU_SET(cpu, &fullMask);
+      }
+
+      if (sched_setaffinity(0, sizeof(fullMask), &fullMask) != 0) {
+        GTEST_SKIP() << "Could not widen CPU affinity, cannot exercise the "
+                     << "GPU-local subset case: " << strerror(errno);
+      }
+
+      cpu_set_t before;
+      CPU_ZERO(&before);
+      ASSERT_EQ(sched_getaffinity(0, sizeof(before), &before), 0)
+          << "sched_getaffinity failed: " << strerror(errno);
+
+      ncclComm_t comm;
+      ncclUniqueId id;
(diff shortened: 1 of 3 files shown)


projects/rccl/test/CpuAffinityTests.cpp
```diff
@@ -0,0 +1,74 @@
...
+      long numCpus = sysconf(_SC_NPROCESSORS_CONF);
+      ASSERT_GT(numCpus, 0);
+      cpu_set_t fullMask;
+      CPU_ZERO(&fullMask);
+      for (int cpu = 0; cpu < static_cast<int>(numCpus); ++cpu) {
+        CPU_SET(cpu, &fullMask);
+      }
+
+      (void)sched_setaffinity(0, sizeof(fullMask), &fullMask);
+
```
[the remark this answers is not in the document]
amd-arozanov: Fixed

projects/rccl/test/CpuAffinityTests.cpp
```diff
@@ -0,0 +1,74 @@
...
+      cpu_set_t after;
+      CPU_ZERO(&after);
+      ASSERT_EQ(sched_getaffinity(0, sizeof(after), &after), 0)
+          << "sched_getaffinity failed: " << strerror(errno);
+
+      NCCLCHECK(ncclCommDestroy(comm));
+
+      ASSERT_EQ(memcmp(&before, &after, sizeof(cpu_set_t)), 0)
+          << "CPU affinity was not restored after ncclCommInitRank "
+          << "(NCCL issue #2033 / AICOMRCCL-1537 regression)";
```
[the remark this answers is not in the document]
amd-arozanov: Fixed

amd-arozanov:
%RCCL_TEST_CONFIG
mi300x_mellanox_ib.json

projects/rccl/test/CpuAffinityTests.cpp
```diff
@@ -0,0 +1,81 @@
...
+namespace RcclUnitTesting
+{
+  /**
+   * \brief Verify the calling thread's CPU affinity is restored after ncclCommInitRank.
+   *
+   * Regression guard for NCCL issue #2033 / AICOMRCCL-1537: initTransportsRank() must
+   * restore the mask it saved before applying the GPU-local one. Single-rank
+   * ncclCommInitRank runs initTransportsRank on the calling thread, so a leak is
+   * observable here. The check only bites when the GPU-local CPU set is a strict
+   * subset of the process mask (multi-NUMA hosts); elsewhere the invariant still holds.
```
nikitaxgusev: Nit: this passes trivially when the GPU-local CPU set equals the full mask (single-NUMA hosts), so a green result does not prove the regression is caught. Consider `GTEST_SKIP`/logging when the mask never narrowed, so PASS is not mistaken for real coverage.

projects/rccl/test/CpuAffinityTests.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+      // Widen to the full online CPU set so the GPU-local subset is more likely to be
+      // strictly narrower, then record what the kernel actually granted.
+      long numCpus = sysconf(_SC_NPROCESSORS_ONLN);
+      ASSERT_GT(numCpus, 0);
+
+      cpu_set_t fullMask;
+      CPU_ZERO(&fullMask);
+
+      int maxCpus = static_cast<int>(std::min<long>(numCpus, CPU_SETSIZE));
```
nikitaxgusev, line 107: Nit: static `cpu_set_t` caps at `CPU_SETSIZE` (1024). The `min(numCpus, CPU_SETSIZE)` guard is correct, but on >1024-CPU hosts the mask is partial; a short comment noting dynamic `CPU_ALLOC` is out of scope would help future readers.

projects/rccl/test/CpuAffinityTests.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+      ncclComm_t comm;
+      ncclUniqueId id;
+      NCCLCHECK(ncclGetUniqueId(&id));
+      HIPCALL(hipSetDevice(0));
+      NCCLCHECK(ncclCommInitRank(&comm, 1, id, 0));
+
+      cpu_set_t after;
+      CPU_ZERO(&after);
+      ASSERT_EQ(sched_getaffinity(0, sizeof(after), &after), 0)
```
nikitaxgusev: Nit: if this `ASSERT_EQ` fails, the function returns before `ncclCommDestroy(comm)`, leaking `comm`. Harmless in an isolated process, but destroying first (or a scope guard) is cleaner.

nikitaxgusev (approved):
LGTM

i-kosarev (approved):
LGTM

projects/rccl/test/CpuAffinityTests.cpp
amd-arozanov: Logged instead of skipped. The test now reads GPU 0's NUMA-local CPU set from sysfs
(the same set ncclTopoGetCpuAffinity uses) and warns when it already covers the whole
process mask; the assertion stays unconditional.

projects/rccl/test/CpuAffinityTests.cpp
amd-arozanov: Added that comment at the clamp. Keeping the static cpu_set_t: a partial mask above
1024 CPUs can only weaken the test, never make it fail.

projects/rccl/test/CpuAffinityTests.cpp
amd-arozanov: Fixed: status and errno go into locals, ncclCommDestroy runs unconditionally, asserts
come after

PR #8689, opened 2026-07-16
fix(rocr): update gran_vgprs calculation of blit shaders for gfx1250

dayatsin-amd:
## Summary

The formula for `compute_pgm_rsrc1.GRANULATED_WORKITEM_VGPR_COUNT` (used to compute `gran_vgprs` in `GpuAgent::AssembleShader`) has changed for gfx1250. It is now:

```
max(0, ceil(vgprs_used / 16) - 1)
```

## JIRA ID

JIRA ID : ROCM-19804

## Test plan

- [ ] Build ROCr and confirm blit-kernel dispatch (copy/fill) works on gfx1250.
- [ ] Verify `granulated_workitem_vgpr_count` in the assembled blit-shader headers matches `max(0, ceil(vgprs_used / 16) - 1)` on gfx1250 and is unchanged on other targets.

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
@@ -436,7 +436,7 @@ void GpuAgent::AssembleShader(const char* func_name, AssembleTarget assemble_tar
     int gran_sgprs = std::max(0, (int(asic_shader->num_sgprs) - 1) / 8);
     int gran_vgprs;
     if (supported_isas()[0]->GetMajorVersion() == 12 && supported_isas()[0]->GetMinorVersion() >= 5) {
-      // gfx1250 (MI450) changed the VGPR granularity: the field is now
+      // gfx1250 changed the VGPR granularity: the field is now
       // max(0, ceil(vgprs_used / 16) - 1). See SWDEV-512636 / SWDEV-510239.
       gran_vgprs = std::max(0, (int(asic_shader->num_vgprs) + 15) / 16 - 1);
     } else {
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
@@ -434,14 +434,14 @@ void GpuAgent::AssembleShader(const char* func_name, AssembleTarget assemble_tar
     amd_kernel_code_t* header = reinterpret_cast<amd_kernel_code_t*>(code_buf);
 
     int gran_sgprs = std::max(0, (int(asic_shader->num_sgprs) - 1) / 8);
-    int gran_vgprs;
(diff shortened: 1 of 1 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -434,7 +434,14 @@ void GpuAgent::AssembleShader(const char* func_name, AssembleTarget assemble_tar
     amd_kernel_code_t* header = reinterpret_cast<amd_kernel_code_t*>(code_buf);
 
     int gran_sgprs = std::max(0, (int(asic_shader->num_sgprs) - 1) / 8);
-    int gran_vgprs = std::max(0, (int(asic_shader->num_vgprs) - 1) / 4);
+    int gran_vgprs;
+    if (supported_isas()[0]->GetMajorVersion() == 12 && supported_isas()[0]->GetMinorVersion() >= 5) {
```

**[Optional]** Every other architecture gate in `AssembleShader` (and throughout this file) uses the `isa_` member directly — including the `ENABLE_WAVEFRONT_SIZE32` / `wavefront_size` logic in this same header-population block — rather than `supported_isas()[0]`. They're functionally identical here (the constructor pushes `isa_` first, so index 0 is `isa_`), but this is the only spot reaching through `supported_isas()[0]`, which adds an implicit index-0 assumption the rest of the code doesn't make. Consider `isa_->GetMajorVersion()` / `isa_->GetMinorVersion()` for consistency.

Optionally, parameterizing the granularity avoids duplicating the whole formula:
```cpp
const int vgpr_gran = (isa_->GetMajorVersion() == 12 && isa_->GetMinorVersion() >= 5) ? 16 : 4;
int gran_vgprs = std::max(0, (asic_shader->num_vgprs + vgpr_gran - 1) / vgpr_gran - 1);
```
Both are non-blocking; the current code is correct.

cfreeamd (commented):

The gfx1250 gran_vgprs formula in amd_gpu_agent.cpp is correct: `(num_vgprs + 15) / 16 - 1` equals `ceil(num_vgprs / 16) - 1`, matching the comment and the 16-VGPR granularity, and the `std::max(0, …)` guard covers the zero case. Non-gfx1250 paths are unchanged. One optional style note left inline.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
dayatsin-amd: Good suggestion — parameterized the granularity so the formula is written once:

```cpp
const int vgpr_gran = (supported_isas()[0]->GetMajorVersion() == 12 &&
                       supported_isas()[0]->GetMinorVersion() >= 5)
                          ? 16
                          : 4;
int gran_vgprs =
    std::max(0, (int(asic_shader->num_vgprs) + vgpr_gran - 1) / vgpr_gran - 1);
```

I kept `supported_isas()[0]` (rather than `isa_`) here for consistency with the adjacent `ENABLE_WAVEFRONT_SIZE32` logic in this same header-population block, which already uses `supported_isas()[0]->GetWavefront()`. Pushed in 8fa4fb8.

cfreeamd (approved):

PR #8692, opened 2026-07-16
fix(rocjitsu): Recover canonicalized RDNA4 indirect calls

bjacob:
Clang emits RDNA4 device-call address builders that sign-extend the high half returned by s_getpc_b64 before adding a signed 48-bit text-relative displacement. The indirect-branch analysis previously invalidated that builder, so it failed to recover the call target and omitted the call edge required for correct CFG and liveness analysis.

Recognize only the exact self-register s_sext_i32_i16 form and preserve the local PC offset. Add a regression test that decodes the compiler sequence and verifies its fixup, callee, continuation, and recovered indirect-call edge.

ISSUE ID : #8701

### DIFF
--- a/emulation/rocjitsu/tests/analysis/liveness_test.cpp
+++ b/emulation/rocjitsu/tests/analysis/liveness_test.cpp
@@ -411,6 +411,110 @@ TEST(CfgAnalysis, RecoveredIndirectBranchEdgeStartsAtConsumerBlock) {
   EXPECT_FALSE(has_predecessor(*fallthrough, consumer));
 }
 
+TEST(CfgAnalysis, RecoversRdna4CanonicalizedGetpcCall) {
+  constexpr uint16_t kPcSreg = 2;
+  constexpr uint16_t kReturnSreg = 30;
+  constexpr uint32_t kLiteralOperand = 255;
+  constexpr uint32_t kTargetDelta = static_cast<uint32_t>(-20);
+
+  // Clang's RDNA4 device-call sequence canonicalizes the high 32-bit getpc
+  // half to a signed 16-bit value before adding the low/high text delta.
+  std::vector<uint32_t> words = {
+      pack_sop1(0x48, 0, kReturnSreg),                         // 0x00: callee return.
+      build_s_nop(0, ROCJITSU_CODE_ARCH_RDNA4),                // 0x04: padding.
+      build_s_nop(0, ROCJITSU_CODE_ARCH_RDNA4),                // 0x08: padding.
+      build_s_nop(0, ROCJITSU_CODE_ARCH_RDNA4),                // 0x0c: padding.
+      pack_sop1(0x47, kPcSreg, 0),                             // 0x10: s_getpc_b64.
+      pack_sop1(0x0f, kPcSreg + 1, kPcSreg + 1),               // 0x14: s_sext_i32_i16.
+      pack_sop2(0, kPcSreg, kPcSreg, kLiteralOperand),         // 0x18: s_add_co_u32.
+      kTargetDelta,                                            // 0x1c: target delta.
+      pack_sop2(4, kPcSreg + 1, kPcSreg + 1, kLiteralOperand), // 0x20: s_add_co_ci_u32.
+      UINT32_MAX,                                              // 0x24: high delta.
+      pack_sop1(0x49, kReturnSreg, kPcSreg),                   // 0x28: s_swappc_b64.
+      build_s_endpgm(ROCJITSU_CODE_ARCH_RDNA4),                // 0x2c: continuation.
+  };
+
+  TestCodeObject co(std::move(words));
+  auto decoder = Decoder::create(ROCJITSU_CODE_ARCH_RDNA4);
+  ASSERT_NE(decoder, nullptr);
+  constexpr std::array<uint64_t, 2> extra_leaders{0, 16};
+  auto blocks = BasicBlock::build(co, *decoder, ROCJITSU_CODE_ARCH_RDNA4, extra_leaders);
+
+  auto *builder = block_starting_at(blocks, 16);
+  auto *caller = block_starting_at(blocks, 40);
+  auto *continuation = block_starting_at(blocks, 44);
+  auto *callee = block_starting_at(blocks, 0);
+  ASSERT_NE(builder, nullptr);
+  ASSERT_NE(caller, nullptr);
+  ASSERT_NE(continuation, nullptr);
+  ASSERT_NE(callee, nullptr);
+  const std::array<std::string_view, 4> expected_mnemonics = {"s_getpc_b64", "s_sext_i32_i16",
+                                                              "s_add_co_u32", "s_add_co_ci_u32"};
+  size_t mnemonic_index = 0;
+  for (const Instruction &inst : builder->instructions()) {
+    ASSERT_LT(mnemonic_index, expected_mnemonics.size());
+    EXPECT_EQ(inst.mnemonic(), expected_mnemonics[mnemonic_index++]);
+  }
+  EXPECT_EQ(mnemonic_index, expected_mnemonics.size());
+  ASSERT_EQ(caller->num_instructions(), 1u);
+  ASSERT_NE(caller->terminator(), nullptr);
+  EXPECT_EQ(caller->terminator()->mnemonic(), "s_swappc_b64");
+  ASSERT_EQ(caller->static_indirect_call_fixups().size(), 1u);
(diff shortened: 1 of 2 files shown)


emulation/rocjitsu/tests/analysis/liveness_test.cpp
```diff
@@ -411,6 +411,62 @@ TEST(CfgAnalysis, RecoveredIndirectBranchEdgeStartsAtConsumerBlock) {
   EXPECT_FALSE(has_predecessor(*fallthrough, consumer));
 }
 
+TEST(CfgAnalysis, RecoversRdna4CanonicalizedGetpcCall) {
```
kuhar, line 414: It would be useful to add rejection cases alongside this positive test for the two register shapes that must not be treated as no-ops: `s_sext_i32_i16 pair_lo, pair_lo` and `s_sext_i32_i16 pair_hi, other_reg`. In both cases, please assert that no fixup is recovered. The new recognizer intentionally relies on `dst == src == pair_hi`, but the current test would still pass if either part of that matcher were accidentally relaxed.

kuhar (commented):
Just one comment

emulation/rocjitsu/tests/analysis/liveness_test.cpp
bjacob: Done. I added negative coverage for both `s_sext_i32_i16 pair_lo, pair_lo` and `s_sext_i32_i16 pair_hi, other_reg`, asserting that neither fixups nor call edges are recovered.

The second case exposed a genuine fail-closed gap in write invalidation, so the transfer now explicitly kills the builder whenever a noncanonical sign extension overwrites either tracked half. The positive and negative RDNA4 CFG tests pass.

emulation/rocjitsu/tests/analysis/liveness_test.cpp
```diff
@@ -411,6 +411,110 @@ TEST(CfgAnalysis, RecoveredIndirectBranchEdgeStartsAtConsumerBlock) {
...
+  constexpr uint16_t kReturnSreg = 30;
+  constexpr uint32_t kLiteralOperand = 255;
+  constexpr uint32_t kTargetDelta = static_cast<uint32_t>(-20);
+
+  struct SignExtensionOperands {
+    uint16_t dst;
+    uint16_t src;
+  };
+  constexpr std::array<SignExtensionOperands, 2> cases{{
+      // The low half is not the sign-extension destination.
```
kuhar: It would help to fix this comment: the initializer below does make the low half the sign-extension destination. Could we say "The low half is not a valid sign-extension destination" so the rejected case stays clear?

kuhar (commented):
Just one new comment

emulation/rocjitsu/tests/analysis/liveness_test.cpp
bjacob: Updated the comment to say that the low half is not a valid sign-extension destination. This now matches the deliberately rejected initializer below it.

kuhar (commented):
Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp:1160`

Nice, targeted recovery of the RDNA4 canonicalization idiom -- this closes a real gap in indirect-call discovery. One forward-looking thought: the recognizer is gated on a single `ctx.arch == ROCJITSU_CODE_ARCH_RDNA4` equality, but this canonicalization comes out of the shared compiler backend, so RDNA3.5 / GFX1250 (and plausibly future RDNA arches) are likely to emit the very same `s_sext_i32_i16 pair_hi, pair_hi` before the signed 48-bit delta. Have you thought about keying this on a capability rather than one enum value?

The rest of this file already models per-arch opcode differences through the `scalar_pc_opcode` / `scalar_sop2_opcode` tables. A parallel predicate (e.g. `arch_canonicalizes_getpc_high_half(arch)` or an entry in an opcode/feature table) would let a future arch opt in with a one-line table addition instead of editing this transfer, and it would give the negative RDNA4 test a natural home to grow into a cross-arch matrix. Not asking to broaden the PR scope now -- RDNA4-only is the right first step -- just suggesting the gate be shaped so widening it later is a table edit rather than a code change. Worth a sentence in the comment / PR description noting the arch check is deliberately narrow for now.

Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp:1164`

The canonical self-register match here re-implements the SOP1 encoding-prefix check plus manual `dst`/`src` bit extraction inline, but this is exactly what the existing `sop1_same_sreg(inst, word, mnemonic, sreg)` helper already expresses (dst == src == sreg). The signed-delta subtract matcher a few lines up already leans on it for `s_abs_i32` (see the `sop1_same_sreg(abs_inst, ..., "s_abs_i32", tmp_sreg)` call). Routing the positive case through the same helper would keep all SOP1 encoding knowledge in one place and make this recognizer read consistently with its neighbors, e.g. sop1_same_sreg(inst, word, "s_sext_i32_i16", static_cast<uint16_t>(pair_lo + 1)).

The fail-closed branch below still needs the raw `dst`, so keeping a small local decode for that path is fine, but the positive form is a clean fit for the shared helper. This is a nice foundation for a family of canonicalizing no-op recognizers, and pushing the encoding details behind helpers is what will let that family grow cleanly.

Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp:1170`

Can the rejection path remain owned by the generic def-invalidation layer? RDNA4 SSextI32I16Sop1 already exposes SDST as dst_operands_[0], and rdna4::Operand::to_register_ref maps OPR_SDST to an SGPR, so falling through to invalidate_written_sgprs kills either noncanonical destination without parsing it here. I rebuilt this file with lines 1170-1173 removed and both new positive/negative CFG tests still pass. Keeping only the positive sop1_same_sreg matcher makes this transfer describe the one semantic exception it owns; otherwise this recognizer duplicates destination decoding and gives this mnemonic a second invalidation policy.

Please split or drop the Harden CDNA DBT dispatch waits against CI contention commit from this PR unless the RDNA4 recovery depends on it. The PR title and description define a narrow indirect-call-analysis change, while this extra commit independently raises four CDNA dispatch waits from 5-10 seconds to 30 seconds. That changes failure-detection behavior for unrelated GPU tests and deserves its own review and validation rather than being coupled to this fix.

Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp:1160`

Missing regression test for the arch gate on this recognizer. The RDNA4 canonicalization branch is gated on ctx.arch == ROCJITSU_CODE_ARCH_RDNA4 in addition to the s_sext_i32_i16 mnemonic check, but nothing pins that gate. I verified it is load-bearing rather than redundant: CDNA4 decodes the identical mnemonics for this whole builder shape -- s_getpc_b64, s_sext_i32_i16, and s_swappc_b64 all exist on CDNA4 (cdna4/sop1.cpp:440/540/567; cdna4/test_encodings.h gives sext=0xBE801700 -> op 0x17, getpc op 0x1c, swappc op 0x1e). So the mnemonic check alone would match on CDNA4; only ctx.arch prevents it. Consequence: if this check is ever widened or turned into a capability predicate, the exact getpc+sext+add+swappc sequence would start being recovered on CDNA -- where the compiler never emits this canonicalization -- and no current test would notice, since none of the CDNA recovery tests contain a sext. Please add a negative regression test that builds this same shape under ROCJITSU_CODE_ARCH_CDNA4 (getpc op 0x1c, sext op 0x17, swappc op 0x1e) and asserts no fixups/call edges are recovered. This is the guard case from the original review ask before it was narrowed to the two register-shape rejections, and it becomes the natural regression test the day the gate turns into a capability table.

Original anchor: `emulation/rocjitsu/tests/analysis/liveness_test.cpp:463`

This positive test never asserts the recovered replaceable range. It locks in source_target_offset and the call edge, but not the fixup's source_recovery_begin_offset / source_recovery_end_offset. That range is the contract binary_translator.cpp:767-769 consumes to relocate the builder, and it is the only externally observable effect of this branch's updated.source_recovery_end_offset write at indirect_branch_discovery.cpp:1165. Worth noting: because the trailing s_add_co_u32 / s_add_co_ci_u32 overwrite source_recovery_end_offset again (indirect_branch_discovery.cpp:764/784/795), line 1165 is currently unobservable in this test -- deleting it would not fail anything. Asserting the range (it should span from the getpc's next instruction through the end of the high add) both documents the intended contract and would catch a regression the day the sext is the last builder instruction before the consumer.

PR #8693, opened 2026-07-16
fix(rocjitsu): Harden AMDGPU text growth against malformed ELF

bjacob:
Native instrumentation grows executable text in code objects supplied by applications. The patcher previously trusted ELF table bounds and alignment values, could request an effectively unbounded padding allocation, and extended every function symbol in the text section through the appended cave.

Validate section and program header tables before copying them, cap composed alignment, preserve the file and virtual alignment residues of moved sections and LOAD segments, and grow only the function symbol contiguous with the old text tail. All failures leave the original image unchanged.

Add regression coverage for transactional rejection of an extreme section alignment; existing text-growth, LOAD-alignment, descriptor, relocation, and symbol-range tests cover the successful paths.

ISSUE ID : #8701

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
@@ -19,11 +19,14 @@ RJ_DIAGNOSTIC_POP
 #include <cstddef>
 #include <cstring>
 #include <limits>
+#include <new>
 #include <numeric>
 #include <optional>
 // Standard library
 #include <span>
+#include <stdexcept>
 #include <unordered_set>
+#include <utility>
 #include <vector>
 
 namespace rocjitsu {
@@ -32,25 +35,34 @@ namespace {
 using KD = rocr::llvm::amdhsa::kernel_descriptor_t;
 namespace kd = rocr::llvm::amdhsa;
 
-[[nodiscard]] std::vector<Elf64_Shdr> read_section_headers(const std::vector<uint8_t> &image,
-                                                           const Elf64_Ehdr &ehdr) {
-  assert(ehdr.e_shentsize == sizeof(Elf64_Shdr) && "unsupported section header size");
-  assert(ehdr.e_shoff + static_cast<uint64_t>(ehdr.e_shnum) * sizeof(Elf64_Shdr) <= image.size() &&
-         "section header table out of bounds");
+[[nodiscard]] std::optional<std::vector<Elf64_Shdr>>
+read_section_headers(const std::vector<uint8_t> &image, const Elf64_Ehdr &ehdr) {
+  if ((ehdr.e_shoff == 0) != (ehdr.e_shnum == 0))
+    return std::nullopt;
+  if (ehdr.e_shoff == 0)
+    return std::vector<Elf64_Shdr>{};
+
+  if (ehdr.e_shentsize != sizeof(Elf64_Shdr) || ehdr.e_shoff > image.size() ||
+      static_cast<uint64_t>(ehdr.e_shnum) >
+          (static_cast<uint64_t>(image.size()) - ehdr.e_shoff) / sizeof(Elf64_Shdr))
+    return std::nullopt;
 
   std::vector<Elf64_Shdr> shdrs(ehdr.e_shnum);
   std::memcpy(shdrs.data(), image.data() + ehdr.e_shoff, shdrs.size() * sizeof(Elf64_Shdr));
   return shdrs;
 }
 
-[[nodiscard]] std::vector<Elf64_Phdr> read_program_headers(const std::vector<uint8_t> &image,
-                                                           const Elf64_Ehdr &ehdr) {
-  if (ehdr.e_phoff == 0 || ehdr.e_phnum == 0)
-    return {};
+[[nodiscard]] std::optional<std::vector<Elf64_Phdr>>
+read_program_headers(const std::vector<uint8_t> &image, const Elf64_Ehdr &ehdr) {
+  if ((ehdr.e_phoff == 0) != (ehdr.e_phnum == 0))
+    return std::nullopt;
+  if (ehdr.e_phoff == 0)
+    return std::vector<Elf64_Phdr>{};
 
-  assert(ehdr.e_phentsize == sizeof(Elf64_Phdr) && "unsupported program header size");
-  assert(ehdr.e_phoff + static_cast<uint64_t>(ehdr.e_phnum) * sizeof(Elf64_Phdr) <= image.size() &&
-         "program header table out of bounds");
+  if (ehdr.e_phentsize != sizeof(Elf64_Phdr) || ehdr.e_phoff > image.size() ||
+      static_cast<uint64_t>(ehdr.e_phnum) >
+          (static_cast<uint64_t>(image.size()) - ehdr.e_phoff) / sizeof(Elf64_Phdr))
+    return std::nullopt;
 
   std::vector<Elf64_Phdr> phdrs(ehdr.e_phnum);
   std::memcpy(phdrs.data(), image.data() + ehdr.e_phoff, phdrs.size() * sizeof(Elf64_Phdr));
@@ -73,17 +85,55 @@ void write_elf_tables(std::vector<uint8_t> &image, const Elf64_Ehdr &ehdr,
   }
 }
 
-void insert_file_bytes(std::vector<uint8_t> &image, Elf64_Ehdr &ehdr,
-                       std::vector<Elf64_Shdr> &shdrs, std::vector<Elf64_Phdr> &phdrs,
-                       uint64_t file_offset, std::span<const uint8_t> bytes,
-                       std::optional<size_t> grown_section_index, bool grow_load_at_segment_end) {
-  assert(file_offset <= image.size() && "ELF insertion offset out of bounds");
+[[nodiscard]] bool can_add_u64(uint64_t value, uint64_t delta) {
+  return value <= std::numeric_limits<uint64_t>::max() - delta;
+}
+
+[[nodiscard]] bool insert_file_bytes(std::vector<uint8_t> &image, Elf64_Ehdr &ehdr,
+                                     std::vector<Elf64_Shdr> &shdrs, std::vector<Elf64_Phdr> &phdrs,
+                                     uint64_t file_offset, std::span<const uint8_t> bytes,
+                                     std::optional<size_t> grown_section_index,
+                                     bool grow_load_at_segment_end) {
+  if (file_offset > image.size() || bytes.size() > image.max_size() - image.size())
+    return false;
   if (bytes.empty())
-    return;
+    return true;
 
   const uint64_t delta = bytes.size();
-  image.insert(image.begin() + static_cast<std::ptrdiff_t>(file_offset), bytes.begin(),
-               bytes.end());
+  if ((ehdr.e_shoff >= file_offset && !can_add_u64(ehdr.e_shoff, delta)) ||
+      (ehdr.e_phoff != 0 && ehdr.e_phoff >= file_offset && !can_add_u64(ehdr.e_phoff, delta)))
+    return false;
+  for (size_t i = 0; i < shdrs.size(); ++i) {
+    if (grown_section_index && i == *grown_section_index)
+      continue;
+    if (shdrs[i].sh_type != SHT_NULL && shdrs[i].sh_offset >= file_offset &&
+        !can_add_u64(shdrs[i].sh_offset, delta))
+      return false;
+  }
+  for (const Elf64_Phdr &phdr : phdrs) {
+    if (!can_add_u64(phdr.p_offset, phdr.p_filesz))
+      return false;
+    const uint64_t old_end = phdr.p_offset + phdr.p_filesz;
+    if (phdr.p_offset >= file_offset) {
+      if (!can_add_u64(phdr.p_offset, delta))
+        return false;
+      continue;
+    }
+    const bool inside_segment = file_offset < old_end;
+    const bool at_segment_end = grow_load_at_segment_end && file_offset == old_end;
+    if (phdr.p_type == PT_LOAD && (inside_segment || at_segment_end) &&
+        (!can_add_u64(phdr.p_filesz, delta) || !can_add_u64(phdr.p_memsz, delta)))
+      return false;
+  }
+
+  try {
+    image.insert(image.begin() + static_cast<std::ptrdiff_t>(file_offset), bytes.begin(),
+                 bytes.end());
+  } catch (const std::bad_alloc &) {
+    return false;
+  } catch (const std::length_error &) {
+    return false;
+  }
 
   if (ehdr.e_shoff >= file_offset)
     ehdr.e_shoff += delta;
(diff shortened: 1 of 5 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
```diff
@@ -32,25 +33,27 @@ namespace {
...
+[[nodiscard]] std::optional<std::vector<Elf64_Phdr>>
+read_program_headers(const std::vector<uint8_t> &image, const Elf64_Ehdr &ehdr) {
   if (ehdr.e_phoff == 0 || ehdr.e_phnum == 0)
-    return {};
+    return std::vector<Elf64_Phdr>{};
 
-  assert(ehdr.e_phentsize == sizeof(Elf64_Phdr) && "unsupported program header size");
-  assert(ehdr.e_phoff + static_cast<uint64_t>(ehdr.e_phnum) * sizeof(Elf64_Phdr) <= image.size() &&
-         "program header table out of bounds");
+  if (ehdr.e_phentsize != sizeof(Elf64_Phdr) || ehdr.e_phoff > image.size() ||
```
kuhar, line 62: We should reject inconsistent program-header metadata instead of treating it as an absent table. With `e_phoff == 0`, `e_phnum == 1`, and a valid `e_phentsize`, `AmdGpuCodeObject` accepts the fixture and `replace_text` succeeds while skipping the program header the ELF claims to contain. Please require both offset and count to indicate absence, and add a regression that also verifies the image is unchanged.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
```diff
@@ -524,25 +566,25 @@ bool CodeObjectPatcher::replace_text(std::span<const uint8_t> new_text) {
       if (!shift_section_vaddr[i])
         continue;
       [[maybe_unused]] const uint64_t old_addr = shdrs[i].sh_addr;
-      shdrs[i].sh_addr += padded_file_delta;
+      shdrs[i].sh_addr += *padded_file_delta;
```
kuhar, line 648: This would be safer if the delta-adjusted ELF fields were preflighted before `insert_file_bytes` mutates `image_`. Setting the following allocated `.rodata` section's `sh_addr` to `UINT64_MAX` and growing text by four bytes currently returns true and emits `sh_addr == 3`. Please reject that input without mutation, and cover the equivalent overflow checks for `p_vaddr`, `p_paddr`, and the other unsigned delta additions.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
```diff
@@ -267,16 +282,20 @@ void grow_text_function_symbols(std::vector<uint8_t> &image, const Elf64_Ehdr &e
...
+      // already reaches the old end of .text. Growing every function through
+      // the cave makes their ranges overlap inter-function alignment padding;
+      // a subsequent instrumentation pass then tries to decode that padding as
+      // instructions. The CFG builder sees all function ranges together, so a
+      // branch from an earlier function can still reach the cave block covered
+      // by the tail function.
       if (symbol.st_size == 0)
         continue;
 
+      if (symbol_text_offset > old_text_size ||
```
kuhar, line 338: I think we should add a regression with two `STT_FUNC` symbols and assert that an interior function remains unchanged while only the function reaching the old `.text` end grows. The current fixtures have one function spanning all of `.text`, so they pass both the old grow-every-function logic and this new filter. A focused two-function probe exercises the intended result (interior size 4; tail size 4 growing to 12), but that negative invariant is not covered by the test suite.

kuhar (commented):
It would help to add focused negative coverage for the new fail-closed paths that currently have no regression. At `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp:38`, malformed section/program-header entry sizes, truncated headers, and out-of-range table offsets or counts should make `replace_text` return false without mutation. At `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp:159`, a shifted `PT_LOAD` with `p_align` above the 2 MiB cap or an overflowing LCM should do the same. The only new negative test exercises the section `sh_addralign` accumulator, so it cannot catch regressions in either the parser checks or the separate LOAD-alignment path.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
```diff
@@ -32,25 +33,27 @@ namespace {
 using KD = rocr::llvm::amdhsa::kernel_descriptor_t;
 namespace kd = rocr::llvm::amdhsa;
 
-[[nodiscard]] std::vector<Elf64_Shdr> read_section_headers(const std::vector<uint8_t> &image,
-                                                           const Elf64_Ehdr &ehdr) {
-  assert(ehdr.e_shentsize == sizeof(Elf64_Shdr) && "unsupported section header size");
-  assert(ehdr.e_shoff + static_cast<uint64_t>(ehdr.e_shnum) * sizeof(Elf64_Shdr) <= image.size() &&
-         "section header table out of bounds");
+[[nodiscard]] std::optional<std::vector<Elf64_Shdr>>
+read_section_headers(const std::vector<uint8_t> &image, const Elf64_Ehdr &ehdr) {
```
atgutier, line 39: I may just be confused, but the PR description says we previously trusted the section header info, but here we were asserting, where as now we return nullopt. Shouldn't we reject these cases instead of returning nullopt?

atgutier (commented):
One question from me and this from the agent reviewer:

MAJOR — must confirm before merge

  The 2 MiB kMaxSupportedElfAlignment cap may silently reject legitimate code objects —
  code_object_patcher.cpp:126. Applied inside checked_lcm_u64, so any shifted LOAD segment or
  allocated section whose p_align/sh_addralign (or the LCM across several) exceeds 2 MiB fails the
  whole patch with a silent return false at the call site (no diagnostic). AMDGPU code objects
  commonly use large-page LOAD alignment — 0x200000 (2 MiB exactly) passes, but anything above is
  rejected. This turns hardening into a potential functional regression on valid input. Crucially, the
   cap isn't even the real anti-DoS guard — the checked_align_up overflow check and the max_size()
  allocation bound (L553) already prevent the unbounded allocation. Fix: either drop the cap and rely
  on those existing guards, or raise/justify it against real large-page AMDGPU alignments (and
  document why). Verify against a real large-page code object before merge.

  MINOR

  - Test coverage is thin for a security PR — the one added negative test covers only the
  section-alignment rejection path. No negative coverage for the other new guards: OOB e_shoff/e_phoff
   / oversized e_shnum/e_phnum, too-small e_shentsize, checked_align_up overflow, or a malformed
  program-header table. Most importantly, nothing asserts a non-tail function symbol is not grown —
  which is the exact behavior this PR narrows (the existing symbol test only checks the tail symbol is
   grown). Add a two-FUNC-symbol test (one ending mid-section, one at the tail) asserting only the
  tail grows, plus OOB/entsize rejection tests each asserting emit() == original.
  - Dead duplicate bounds check — code_object_patcher.cpp:295: symbol_text_offset > old_text_size is
  re-tested after being handled by an identical continue at L282; always false here. Drop it, keeping
  only the real st_size != old_text_size - symbol_text_offset predicate.
  - Hardening stops at replace_text — update_elf_flags (L610) still does an unchecked
  reinterpret_cast<Elf64_Ehdr*> write and text_bytes() (L479) returns an unchecked range, on the same
  untrusted image. Not introduced here (out of scope), but a follow-up should apply the same
  size-checked discipline for consistency with the threat model.

  NIT

  New checked_*/validation helpers lack Doxygen stating their nullopt contract (overflow vs cap vs
  OOB); the write_elf_tables asserts (L68,73) still use the wrap-prone off + size <= image.size() form
   the PR deliberately replaced elsewhere (asserts, post-validation, so not exploitable — but
  inconsistent with the PR's own convention).

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
bjacob: Done. Program-header absence now requires both `e_phoff == 0` and `e_phnum == 0`; inconsistent metadata is rejected. The regression also verifies that `emit()` remains byte-for-byte identical to the original image.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
bjacob: Done. `replace_text()` is now transactional: all changes are made to a temporary image and committed only after every adjustment succeeds. Section addresses, program offsets and sizes, `p_vaddr`, `p_paddr`, symbols, relocations, and descriptor-relative fields use checked arithmetic. The new overflow regressions verify rejection with byte-for-byte unchanged output.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
bjacob: Added two `STT_FUNC` symbols. The interior function remains size 4, while only the function reaching the old `.text` tail grows into the appended cave.

bjacob:
> It would help to add focused negative coverage for the new fail-closed paths that currently have no regression. At `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp:38`, malformed section/program-header entry sizes, truncated headers, and out-of-range table offsets or counts should make `replace_text` return false without mutation. At `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp:159`, a shifted `PT_LOAD` with `p_align` above the 2 MiB cap or an overflowing LCM should do the same. The only new negative test exercises the section `sh_addralign` accumulator, so it cannot catch regressions in either the parser checks or the separate LOAD-alignment path.

Added focused negative coverage for inconsistent, incorrectly sized, truncated, and out-of-range program-header tables, as well as malformed section-header entry sizes. Each patcher regression verifies `replace_text()` returns `false` and `emit()` remains unchanged.

I also added separate regressions for an unrepresentable large alignment and alignment-LCM overflow. Section-table offset/count malformations are rejected earlier by `AmdGpuCodeObject`, before a `CodeObjectPatcher` with a usable `.text` section can be constructed.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
bjacob: Codex suggested this reply, WDYT?

> `nullopt` is the internal rejection signal here. `replace_text()` immediately converts it into `false`; malformed tables are not treated as absent or accepted. I also added regressions that verify rejection leaves the image unchanged.

bjacob:
@atgutier overall review:

Addressed the major and minor findings:

- Removed the arbitrary 2 MiB alignment ceiling. Alignment composition now rejects only arithmetic overflow or an allocation that cannot be represented or performed.
- Added malformed-table, large-alignment, LCM-overflow, unsigned-addition overflow, and unchanged-image regressions.
- Added two-function coverage proving that the interior function does not grow.
- Removed the duplicate `symbol_text_offset > old_text_size` check.

I agree that hardening `text_bytes()` and `update_elf_flags()` is worthwhile follow-up work, but those pre-existing APIs are outside this text-growth change.

kuhar (commented):
The alignment limit should be either caller-owned or part of a documented supported-input contract. At `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp:617`, a small accepted ET_REL image with a following allocated section whose `sh_addralign` is `1 << 27` needs a representable 128 MiB delta, but the `max(image.size(), 64 MiB)` padding ceiling rejects it even though `vector::max_size()` permits it. Alignment composition therefore still rejects more than arithmetic overflow or an unrepresentable allocation, contrary to the current resolution. Please remove the policy cap, or expose and document it at the `CodeObjectPatcher` API boundary.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
kuhar: The overflow regression should also cover the directly reachable guards that remain untested. `RejectsDeltaAdditionOverflowWithoutMutation` currently exercises `sh_addr` and four program-header fields, but not the `st_value` overflow at line 292 or the kernel-descriptor entry-offset underflow at lines 510-513. The existing fixtures can reach these by setting the moved `.rodata` symbol value to `UINT64_MAX` and the moved descriptor entry offset to `INT64_MIN`, respectively. Please add both cases and assert that `replace_text()` returns false while `emit()` remains byte-for-byte unchanged.

bjacob: Added both requested cases. The regression now sets the moved `.rodata` symbol's `st_value` to `UINT64_MAX` and, separately, the moved kernel descriptor's entry offset to `INT64_MIN`. Both cases verify that `replace_text()` returns `false` and `emit()` remains byte-for-byte identical to the original image.

bjacob:
Removed the implementation-owned padding ceiling. Text growth now rejects only arithmetic overflow, a size beyond `vector::max_size()`, or an allocation failure; it no longer rejects an otherwise representable alignment based on an undocumented policy cap.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
```diff
@@ -478,10 +546,20 @@ bool CodeObjectPatcher::replace_text(std::span<const uint8_t> new_text) {
...
-  auto *ehdr = reinterpret_cast<Elf64_Ehdr *>(image_.data());
-  auto header = *ehdr;
-  auto shdrs = read_section_headers(image_, header);
-  auto phdrs = read_program_headers(image_, header);
+  if (image_.size() < sizeof(Elf64_Ehdr))
+    return false;
+  // Build the replacement transactionally. Malformed metadata discovered late
+  // in relocation or descriptor adjustment must not leave a partially shifted
+  // code object behind.
+  std::vector<uint8_t> image = image_;
```
kuhar, line 566: The risk here is that `replace_text` still has neither a finite resource budget nor a complete exception boundary. A small accepted object can make `padded_file_delta` hundreds of GiB through `sh_addralign` or `p_align`; `inserted.resize()` eagerly zeroes that range, so `vector::max_size()` and `bad_alloc` do not protect an overcommitting host from the OOM killer. The initial `image_` copy and the section, program-header, and shift-map vector allocations can also throw outside the current catches, while both production callers expect failure to be reported as `false` and do not catch these exceptions.

We should pass an explicit caller-owned output or padding budget and reject before allocating when it is exceeded. It would also be safer to put the exception-to-`false` boundary around the complete transaction so every allocation failure leaves the code object unchanged.

kuhar (commented):
One comment

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
bjacob: Addressed. `CodeObjectPatcher::replace_text` now requires every caller to provide an explicit `max_file_growth` budget, including alignment padding, rather than owning a hidden global policy.

The complete replacement transaction is now behind a `noexcept` boundary that converts `std::bad_alloc` and `std::length_error` into failure without mutating the original code object. The aligned file delta is checked against the caller’s budget before allocating padding.

I also added a regression with pathological but representable ELF alignment that verifies the operation is rejected without attempting the large allocation and leaves the input byte-identical.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
```diff
@@ -73,17 +85,55 @@ void write_elf_tables(std::vector<uint8_t> &image, const Elf64_Ehdr &ehdr,
...
+    return false;
+  for (size_t i = 0; i < shdrs.size(); ++i) {
+    if (grown_section_index && i == *grown_section_index)
+      continue;
+    if (shdrs[i].sh_type != SHT_NULL && shdrs[i].sh_offset >= file_offset &&
+        !can_add_u64(shdrs[i].sh_offset, delta))
+      return false;
+  }
+  for (const Elf64_Phdr &phdr : phdrs) {
+    if (!can_add_u64(phdr.p_offset, phdr.p_filesz))
```
kuhar, line 114: We should reject program headers whose declared file range is outside the input before using them. In the existing two-LOAD fixture, setting the first `p_filesz`/`p_memsz` to 4096 for an 840-byte image still lets `replace_text()` succeed and emits the impossible segment at 12288 bytes; the check here only rules out integer wrap. Please validate `p_offset + p_filesz <= image.size()` and `p_filesz <= p_memsz`, with a regression that expects failure and byte-identical output.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
```diff
@@ -118,34 +168,42 @@ void insert_file_bytes(std::vector<uint8_t> &image, Elf64_Ehdr &ehdr,
...
   uint64_t alignment = 1;
   // Later LOAD segments remain loader-valid only when their file shift keeps
   // p_offset % p_align congruent with p_vaddr % p_align.
   for (const Elf64_Phdr &phdr : phdrs) {
     if (phdr.p_type != PT_LOAD || phdr.p_align <= 1)
       continue;
-    if (phdr.p_offset >= file_offset)
-      alignment = checked_lcm_u64(alignment, phdr.p_align);
+    if (phdr.p_offset >= file_offset) {
+      const auto next = checked_lcm_u64(alignment, phdr.p_align);
```
kuhar, line 202: We should reject malformed ELF alignments before composing them. The ELF contract permits 0/1 or a positive power of two for `p_align` and `sh_addralign`, but the current LCM path accepts other values: for example, a shifted section with `sh_addralign == 3` can produce a delta divisible by 3 and emit the same invalid header. Please add section- and program-header regressions that require `replace_text()` to fail with byte-identical output.

kuhar (commented):
A few comments

PR #8694, opened 2026-07-16
fix(rocjitsu): Reject AMDGPU s_clause trampoline anchors

bjacob:
An s_clause instruction groups the following memory operations for scheduling. Relocating the clause or any covered instruction into an out-of-line trampoline breaks that grouping and can change program semantics.

Teach the native instrumentor to reject both the s_clause instruction and every instruction in its declared run. Add focused tests for a direct clause anchor and an interior covered instruction.

ISSUE ID : #8701

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
@@ -18,6 +18,7 @@
 #include "rocjitsu/isa/decoder.h"
 #include "rocjitsu/isa/instruction.h"
 
+#include <algorithm>
 #include <array>
 #include <cstring>
 #include <optional>
@@ -55,6 +56,22 @@ constexpr std::string_view kRfePrefix = "s_rfe_";
   if (mnemonic.size() >= kRfePrefix.size() && mnemonic.substr(0, kRfePrefix.size()) == kRfePrefix)
     return true;
   return false;
+}
+
+[[nodiscard]] bool is_s_clause(const Instruction &inst) { return inst.mnemonic() == "s_clause"; }
+
+[[nodiscard]] uint32_t s_clause_following_instruction_count(const Instruction &inst) {
+  if (!is_s_clause(inst) || inst.size() != sizeof(uint32_t) || inst.num_src_operands() != 1)
+    return 0;
+
+  const Operand *clause = inst.src_operand(0);
+  if (clause == nullptr)
+    return 0;
+
+  // The decoded OPR_CLAUSE retains the packed SIMM16 value. SIMM16[5:0]
+  // encodes one less than the number of following instructions; bits 8:11 are
+  // BREAK_SPAN and are not part of the count.
+  return (static_cast<uint32_t>(clause->encoding_value()) & 0x3fu) + 1u;
 }
 
 // Per-site result of Instrumentor::patch's preflight: the chosen trampoline
@@ -179,6 +196,10 @@ bool is_relocatable_anchor(const Instruction &anchor, uint64_t anchor_offset,
   }
   if (is_denylisted_mnemonic(anchor.mnemonic())) {
     report(error_out, "anchor mnemonic is in the PC-relative denylist");
+    return false;
+  }
+  if (is_s_clause(anchor)) {
+    report(error_out, "anchor mnemonic is s_clause");
     return false;
   }
   return true;
@@ -330,10 +351,20 @@ bool Instrumentor::ensure_blocks_built(std::string *error_out) {
   }
   decoder_ = std::move(decoder);
   blocks_ = BasicBlock::build(obj_, *decoder_, arch_);
+  // BasicBlock::build returns blocks in .text order. Keep clause state across
+  // block boundaries because a branch target may split the linear instruction
+  // stream in the middle of a clause.
+  uint32_t clause_remaining = 0;
   for (const auto &block : blocks_) {
     uint64_t cur = block->start_offset();
     for (const Instruction &inst : block->instructions()) {
+      if (clause_remaining > 0) {
+        clause_blocked_offsets_.insert(cur);
+        --clause_remaining;
+      }
       offset_to_inst_.emplace(cur, &inst);
+      if (is_s_clause(inst))
+        clause_remaining = std::max(clause_remaining, s_clause_following_instruction_count(inst));
       cur += static_cast<uint64_t>(inst.size());
     }
   }
@@ -412,6 +443,12 @@ Instrumentor::ResolvedPoints Instrumentor::resolve_points() {
     auto site = validate_anchor(*anchor, pt.anchor_offset, text_bytes, pt, arch_, &perr);
     if (!site) {
       out.errors.push_back(std::move(perr));
+      continue;
+    }
+
+    if (clause_blocked_offsets_.contains(pt.anchor_offset)) {
+      out.errors.emplace_back("anchor is inside an s_clause run, anchor_offset = " +
+                              std::to_string(pt.anchor_offset));
       continue;
     }
 
--- a/emulation/rocjitsu/tests/patch/instrumentor_test.cpp
+++ b/emulation/rocjitsu/tests/patch/instrumentor_test.cpp
@@ -271,6 +271,17 @@ TEST(Validator, RejectsDenylistedMnemonic) {
   }
 }
 
+TEST(Validator, RejectsSClauseAnchor) {
+  static constexpr uint32_t kRaw = 0xBF850001u;
+  TestInstruction anchor("s_clause", 4, /*flags=*/0, std::nullopt, &kRaw);
+  auto text = dummy_text();
+  InstrumentationPoint pt;
+
+  std::string err;
+  EXPECT_FALSE(validate_anchor(anchor, 0, text, pt, ROCJITSU_CODE_ARCH_RDNA4, &err).has_value());
+  EXPECT_NE(err.find("s_clause"), std::string::npos) << "error was: " << err;
+}
+
 // Instruction::size() is a signed int; the `size != 4 && size != 8` check must
 // reject negative sizes (which decoders never emit in practice) rather than let
 // them through to the later cast to unsigned, where they would wrap.
@@ -412,7 +423,9 @@ uint64_t align_up_for_test(uint64_t value, uint64_t alignment) {
 // Build a minimal gfx950 ELF whose .text holds @p words verbatim. Lets a test
 // embed a real multi-word instruction (e.g. an 8-byte SOP1 + literal) rather
 // than only s_nop fillers.
-std::vector<uint8_t> make_gfx950_elf_with_text_words(const std::vector<uint32_t> &words) {
+std::vector<uint8_t>
+make_gfx950_elf_with_text_words(const std::vector<uint32_t> &words,
+                                uint32_t machine_flags = EF_AMDGPU_MACH_AMDGCN_GFX950) {
   const uint64_t text_offset = 0x100;
   const uint64_t text_size = words.size() * sizeof(uint32_t);
   const uint64_t rodata_size = 4;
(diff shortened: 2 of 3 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
```diff
@@ -57,6 +57,17 @@ constexpr std::string_view kRfePrefix = "s_rfe_";
...
 
+[[nodiscard]] bool is_s_clause(const Instruction &inst) { return inst.mnemonic() == "s_clause"; }
+
+[[nodiscard]] uint32_t s_clause_following_instruction_count(const Instruction &inst) {
+  if (!is_s_clause(inst) || inst.raw_encoding() == nullptr || inst.size() != sizeof(uint32_t))
+    return 0;
+
+  uint32_t word = 0;
+  std::memcpy(&word, inst.raw_encoding(), sizeof(word));
+  return (word & 0xffffu) + 1u;
```
kuhar: We should extract the clause length from `SIMM16[5:0]` (`(word & 0x3f) + 1`) rather than the full 16-bit immediate. GFX12 uses bits 8-11 for `BREAK_SPAN`, so a valid `s_clause 515` has four members but this returns 516 and rejects unrelated anchors after the clause. Please add a regression with nonzero break bits that also checks the first instruction after the clause remains anchorable.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
```diff
@@ -332,8 +347,17 @@ bool Instrumentor::ensure_blocks_built(std::string *error_out) {
   blocks_ = BasicBlock::build(obj_, *decoder_, arch_);
   for (const auto &block : blocks_) {
     uint64_t cur = block->start_offset();
+    uint32_t clause_remaining = 0;
```
kuhar: This would be safer if clause coverage were tracked across the linear instruction stream instead of resetting `clause_remaining` for each `BasicBlock`. `BasicBlock::build` splits at recovered branch targets, so a target inside a clause starts a new block and makes the remaining members eligible anchors. Please add a regression with a branch target on a clause member to verify the tail remains blocked.

kuhar (commented):
A few comments

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
```diff
@@ -57,6 +57,17 @@ constexpr std::string_view kRfePrefix = "s_rfe_";
   return false;
 }
 
+[[nodiscard]] bool is_s_clause(const Instruction &inst) { return inst.mnemonic() == "s_clause"; }
+
+[[nodiscard]] uint32_t s_clause_following_instruction_count(const Instruction &inst) {
```
accauble, line 63: Nit: I would love a comment that says this counts the number of instructions that follow a given s_clause instruction (for some reason my first read thought this was a check to see if the given instruction fell in an s_clause block)

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
accauble: Nit: The PR description described what this was for really well. A related comment here could help a code reader.

emulation/rocjitsu/tests/patch/instrumentor_test.cpp
```diff
@@ -698,6 +709,27 @@ TEST(Instrumentor, RejectsOffsetInteriorToMultiWordInstruction) {
       << "error was: " << result.errors.front();
 }
 
+TEST(Instrumentor, RejectsInstructionInsideSClauseRun) {
```
accauble: Could we add a test to make sure that the first instruction after the clause would be accepted?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
accauble: Does it make sense to move this higher up (into BasicBlock for instance) so that DBT can use it too?

atgutier (commented):
Nothing from me. But the agent has comments:

BLOCKER — must fix

  s_clause run-length mask is wrong (0xffff instead of 0x3f) — instrumentor.cpp:66. The decode masks
  the entire 16-bit immediate to get the clause length, but the length lives only in the low 6 bits;
  SIMM16[11:8] is a separate BREAK_SPAN field. Correct count is SIMM16[5:0] + 1. The reviewer
  cross-checked the ISA encoding across every RDNA generation that has the instruction.

  return (word & 0xffffu) + 1u;   // wrong: folds BREAK_SPAN into the count

  The two new tests pass only because they set BREAK_SPAN == 0 (0xBF850001 → count 2 either way). But
  any real clause using clause-breaks — e.g. SIMM16 = 0x0301 (length 2, break-span 3) — returns 0x301
  + 1 = 770, over-blocking 770 following offsets and wrongly rejecting a large run of valid downstream
   anchors. It fails safe w.r.t. clause semantics (over-rejects, never leaves a covered instruction
  relocatable), so it's not a crash — but it's a real functional defect masked by tests that never
  exercise the upper bits. Fix: mask 0x3f (with a named constant), or better, consume the
  already-decoded operand (see MAJOR).

  MAJOR

  Hand-rolled raw-byte decode duplicates instruction decoding — instrumentor.cpp:66. The instruction
  is already fully decoded — SClauseSopp exposes a simm16 operand typed OPR_CLAUSE. Re-deriving the
  field from raw_encoding() with a manual mask is exactly what produced the BLOCKER, and it puts the
  ISA field-width convention in two places. Fixing it properly (read the decoded operand) eliminates
  the BLOCKER at its source. This is the root cause, not just a style issue.

  MINOR

  - Clause run walked per-basic-block — instrumentor.cpp:352: clause_remaining resets to 0 at each
  block boundary, so a clause whose body is split by a branch target (a leader mid-clause) leaves the
  tail anchorable again. Unlikely in well-formed compiler output (clause bodies are straight runs) but
   a real gap for adversarial code and invisible in the tests. Document the assumption or carry the
  count across fallthrough-contiguous blocks.
  - Clause-blocked check runs after probe resolution — instrumentor.cpp:453: the guard is placed after
   validate_anchor/resolve_probe_index (which mutate probe state), recovered only by an all-or-nothing
   clear(). Move it right after validate_anchor to fail fast and keep anchor-legality checks together.
   Also the s_clause-self insert into clause_blocked_offsets_ is dead (already rejected earlier in
  is_relocatable_anchor).
  - is_relocatable_anchor doxygen not updated for the new s_clause rule (instrumentor.h:168); and the
  public predicate rejects the s_clause opcode but not clause-interior instructions (interior
  detection needs stream context, lives only inside Instrumentor). Today's external callers are CDNA2
  (no s_clause) so unaffected, but the predicate's contract is now incomplete for RDNA — document or
  centralize.

 NIT

  The interior test builds a gfx950 (CDNA3) ELF but decodes with RDNA4 — CDNA has no s_clause at all,
  so the machine flag and decoder arch are inconsistent (decode follows arch_=RDNA4, so it still
  works, but it's misleading — use an RDNA4-flagged ELF). Missing high-value boundary tests: the last
  covered instruction (the off-by-one boundary), the first instruction past the run (proves the
  rejection isn't over-broad — exactly what the 0xffff bug breaks), a clause with BREAK_SPAN set
  (would have caught the BLOCKER directly), and a truncated run at end-of-text.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
bjacob: Fixed. Clause length is derived from the decoded `OPR_CLAUSE` operand, using `SIMM16[5:0]` only; `BREAK_SPAN` bits are excluded. The regression uses nonzero `BREAK_SPAN` and verifies both the last covered instruction and the first legal instruction after the clause.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
bjacob: Clause state now follows the linear decoded stream across `BasicBlock` boundaries. A regression creates a branch target on a covered clause member and verifies that the clause tail remains rejected.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
bjacob: Added a comment explaining that the decoded count describes the instructions following the `s_clause`, rather than testing whether the clause instruction itself belongs to a run.

emulation/rocjitsu/tests/patch/instrumentor_test.cpp
bjacob: Added. The boundary regression rejects the last covered instruction and accepts the first instruction immediately after the clause run. It also sets nonzero `BREAK_SPAN` bits so an over-broad count would fail the test.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
bjacob: Added a comment at the traversal explaining that `BasicBlock::build` returns blocks in `.text` order and that clause state must cross block boundaries because a branch target may split the linear instruction stream inside a clause.

bjacob:
@atgutier overall review:

Addressed the blocker and follow-up findings:

- Clause length now comes from the decoded `OPR_CLAUSE` operand and masks only `SIMM16[5:0]`; `BREAK_SPAN` is excluded.
- Clause coverage carries across the linear decoded stream and therefore across `BasicBlock` boundaries.
- Clause-interior rejection runs immediately after anchor validation, before probe resolution.
- The redundant `s_clause` self-entry in `clause_blocked_offsets_` was removed.
- The public `is_relocatable_anchor()` contract now documents that `s_clause` itself is rejected there, while stream-context interior rejection belongs to `Instrumentor`.
- Tests use an RDNA4-flagged ELF and cover nonzero `BREAK_SPAN`, the last covered instruction, the first instruction after the run, a branch-target split, and a truncated run at end of text.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
```diff
@@ -330,10 +350,20 @@ bool Instrumentor::ensure_blocks_built(std::string *error_out) {
...
   for (const auto &block : blocks_) {
     uint64_t cur = block->start_offset();
     for (const Instruction &inst : block->instructions()) {
+      if (clause_remaining > 0) {
+        clause_blocked_offsets_.insert(cur);
+        --clause_remaining;
+      }
       offset_to_inst_.emplace(cur, &inst);
+      if (is_s_clause(inst))
+        clause_remaining = s_clause_following_instruction_count(inst);
```
kuhar: We should preserve the outer coverage when another `s_clause` appears inside an active run. After decrementing the outer remainder, this assignment can replace it with a smaller inner count; for example, `s_clause 3` followed by `s_clause 0` leaves the final two outer members anchorable. Please keep the larger remaining range and add an overlapping-clause regression so malformed objects remain fail-closed.

kuhar (commented):
One new comment

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
bjacob: Fixed this by retaining the larger of the active outer remainder and the newly decoded inner clause count. Added a malformed overlapping-clause regression using `s_clause 3` followed by `s_clause 0`; the final outer member remains rejected while the first instruction after the outer run remains a valid anchor.

kuhar (approved):
LGTM

accauble (approved):

PR #8695, opened 2026-07-16
fix(rocjitsu): Keep DRM layout checks warning-clean with Clang

bjacob:
Recent libdrm headers intentionally use nested anonymous types. Clang diagnoses those declarations under rocJITsu warning settings before the ABI layout assertions can compile.

Suppress only -Wnested-anon-types around the external header include, leaving project sources and every other diagnostic unchanged.

ISSUE ID : #8701

### DIFF
--- a/emulation/rocjitsu/tests/drm_info_layout_test.cpp
+++ b/emulation/rocjitsu/tests/drm_info_layout_test.cpp
@@ -10,7 +10,11 @@
 #include <cstddef>
 #include <cstdint>
 
+#include "rocjitsu/base/rj_compiler.h"
+RJ_DIAGNOSTIC_PUSH
+RJ_DIAGNOSTIC_IGNORE_NESTED_ANON_TYPES
 #include <libdrm/amdgpu_drm.h>
+RJ_DIAGNOSTIC_POP
 
 namespace {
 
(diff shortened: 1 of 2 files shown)


emulation/rocjitsu/tests/drm_info_layout_test.cpp
```diff
@@ -10,7 +10,14 @@
 #include <cstddef>
 #include <cstdint>
 
+#if defined(__clang__)
+#pragma clang diagnostic push
```
kuhar: I think we should use the existing diagnostic helpers in `rocjitsu/base/rj_compiler.h` here. The test suite already wraps third-party headers with `RJ_DIAGNOSTIC_PUSH`/`RJ_DIAGNOSTIC_POP`; adding an `RJ_DIAGNOSTIC_IGNORE_NESTED_ANON_TYPES` mapping for Clang (and an empty definition for unaffected compilers) would keep this compiler-specific suppression in the same shared abstraction.

kuhar (commented):
One comment

emulation/rocjitsu/tests/drm_info_layout_test.cpp
bjacob: Done. Added `RJ_DIAGNOSTIC_IGNORE_NESTED_ANON_TYPES` to `rj_compiler.h`, mapped to the Clang diagnostic and left empty for unaffected compilers.

kuhar (approved):

PR #8702, opened 2026-07-16
refactor(simdojo): add concurrent memory primitives

lialan:
# refactor(simdojo): add concurrent memory primitives

ISSUE ID: #6447

> [!IMPORTANT]
> This is **PR 1 of 6** in the parallel CPU simulation stack.
> It is based on `develop`; the next layer is
> `users/lialan/rocjitsu_parallel_cpu_02_gpu_state`.
>
> Review only `develop...users/lialan/rocjitsu_parallel_cpu_01_simdojo` for
> this layer.

## Stack

| # | Layer | Branch | PR |
|---|---|---|---|
| **1** | **Concurrent Simdojo memory primitives** | `users/lialan/rocjitsu_parallel_cpu_01_simdojo` | **this PR** |
| 2 | Shared Rocjitsu GPU-state concurrency | `users/lialan/rocjitsu_parallel_cpu_02_gpu_state` | #8703 |
| 3 | XCD execution partitioning | `users/lialan/rocjitsu_parallel_cpu_03_xcd` | #8705 |
| 4 | Compositional plugin threading policy | `users/lialan/rocjitsu_parallel_cpu_04_plugins` | #8706 |
| 5 | Functional CU dispatch pool | `users/lialan/rocjitsu_threadpool_cpu` | #6962 |
| 6 | Functional-mode performance work | `users/lialan/rocjitsu_parallel_cpu_06_perf` | TBD |

## Summary

Add generic concurrency primitives to Simdojo without changing Rocjitsu
dispatch policy.

- Protect sparse memory with striped synchronization.
- Add block read and write operations.
- Add generic cache-line access and invalidation helpers.
- Add focused concurrent sparse-memory coverage.

## Why

Later Rocjitsu worker and engine parallelism needs thread-safe shared storage.
Keeping these generic primitives in the bottom layer makes that foundation
reviewable independently of GPU-specific policy.

## Validation

- Built `rocjitsu_tests`.
- Passed `SparseMemoryThreadingTest.ConcurrentDifferentPageWritesArePreserved`.
- Passed the layer-range pre-commit sweep.

### DIFF
--- a/.github/workflows/rocjitsu-corpus-tests.yml
+++ b/.github/workflows/rocjitsu-corpus-tests.yml
@@ -49,24 +49,30 @@ jobs:
       options: --user 0:0
       env:
         DEBIAN_FRONTEND: noninteractive
-    continue-on-error: true
+    continue-on-error: ${{ matrix.allow_failure }}
     strategy:
       fail-fast: false
       matrix:
         include:
           - name: release
+            allow_failure: true
+            test_regex: ''
             build_type: Release
             enable_asan: 0
             enable_ubsan: 0
             enable_tsan: 0
             enable_msan: 0
           - name: asan-ubsan
+            allow_failure: true
+            test_regex: ''
             build_type: RelWithDebInfo
             enable_asan: 1
             enable_ubsan: 1
             enable_tsan: 0
             enable_msan: 0
           - name: asan-ubsan-gcc
+            allow_failure: true
+            test_regex: ''
             build_type: RelWithDebInfo
             enable_asan: 1
             enable_ubsan: 1
@@ -74,6 +80,17 @@ jobs:
             enable_msan: 0
             c_compiler: gcc
             cxx_compiler: g++
+          - name: tsan-sparse-memory
+            allow_failure: false
+            # The simulator-wide stress tests exceed TSan's fixed 128-lock
+            # deadlock-tracker capacity; keep the race gate on this change's
+            # concurrent SparseMemory surface.
+            test_regex: '^SparseMemory(Threading)?Test\.'
+            build_type: RelWithDebInfo
+            enable_asan: 0
+            enable_ubsan: 0
+            enable_tsan: 1
+            enable_msan: 0
 
     steps:
       - name: Install base dependencies
@@ -188,7 +205,12 @@ jobs:
           WORKERS="$(( ($(nproc) + 1) / 2 ))"
           cmake --build "${ROCJITSU_BUILD_DIR}" -j "${WORKERS}"
 
+          CTEST_SELECTION=()
+          if [[ -n "${{ matrix.test_regex }}" ]]; then
+            CTEST_SELECTION=(-R "${{ matrix.test_regex }}")
+          fi
           ctest --test-dir "${ROCJITSU_BUILD_DIR}" \
+            "${CTEST_SELECTION[@]}" \
             -E "${EXCLUDED_TESTS}" \
             --output-on-failure \
             -j "${WORKERS}"
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
@@ -143,6 +143,56 @@ public:
 
   uint8_t *translate_debug(uint64_t addr, uint32_t vmid) const { return translate(addr, vmid); }
 
+  /// @brief Find the contiguous host range containing a VMID-scoped GPU VA.
+  /// @details KFD dispatches use per-process page tables. Kernel-symbol
+  /// resolution needs a daemon-accessible host pointer range so it can scan
+  /// backward from the kernel descriptor to the loaded ELF header.
+  std::pair<uint64_t, uint64_t> find_host_range(uint64_t addr, uint32_t vmid) const {
+    if (vmid == 0) {
+      auto *host = translate(addr, vmid);
+      if (!host)
+        return {0, 0};
+      return {reinterpret_cast<uint64_t>(host), PAGE_SIZE};
+    }
+
+    std::shared_lock vmid_lock(vmid_mutex_);
+    auto vmid_entry = vmid_table_.find(vmid);
+    if (vmid_entry == vmid_table_.end())
+      return {0, 0};
+
+    auto &entry = vmid_entry->second;
+    std::shared_lock page_table_lock(*entry.mutex);
+    const uint64_t page = addr >> PAGE_SHIFT;
+    auto page_entry = entry.page_table->find(page);
+    if (page_entry == entry.page_table->end())
+      return {0, 0};
+
+    uint64_t first_page = page;
+    uint8_t *first_host_page = page_entry->second.host_ptr;
+    while (first_page > 0) {
+      auto previous_page_entry = entry.page_table->find(first_page - 1);
+      if (previous_page_entry == entry.page_table->end() ||
+          previous_page_entry->second.host_ptr + PAGE_SIZE != first_host_page)
+        break;
+      --first_page;
+      first_host_page = previous_page_entry->second.host_ptr;
+    }
+
+    uint64_t last_page = page;
+    uint8_t *last_host_page = page_entry->second.host_ptr;
+    while (true) {
+      auto next_page_entry = entry.page_table->find(last_page + 1);
+      if (next_page_entry == entry.page_table->end() ||
+          next_page_entry->second.host_ptr != last_host_page + PAGE_SIZE)
+        break;
+      ++last_page;
+      last_host_page = next_page_entry->second.host_ptr;
+    }
+
+    const uint64_t range_size = ((last_page - first_page) + 1) << PAGE_SHIFT;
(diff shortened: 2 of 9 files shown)


atgutier (commented):
I didn't spot any issues myself, but the agent reviewer did:

BLOCKER

  read_block/write_block multi-page atomicity is unspecified and, as written, non-atomic —
  sparse_memory.h. The block ops iterate page-by-page, and each chunk independently acquires and
  releases its stripe lock (via read_sparse_chunk/write_sparse_chunk). A single-page block is atomic
  w.r.t. another writer on that page (fine), but a block spanning N pages is N independent critical
  sections — two concurrent overlapping multi-page blocks can interleave (page-1 from writer A, page-2
   from writer B) producing a torn result matching neither. This is the exact primitive the later
  parallel layers build on, so the contract must be nailed down. It may be an acceptable contract
  (per-page atomic, block non-atomic), but it's completely undocumented — the @details says nothing
  about atomicity. Fix: decide and document. Minimum — add "Thread-safe; atomicity is per-page only, a
   multi-page block is not atomic w.r.t. concurrent overlapping block ops." If block-level atomicity
  is required, gather sorted unique stripe indices and std::scoped_lock them before touching any page.

  MAJOR

  1. map_host_pages publishes host_pages_mapped_ = true before the map is populated —
  sparse_memory.h:192-197. The early store(true) happens outside the lock and before the entries
  exist, then the authoritative value is stored again under the lock. Not UB (every reader re-checks
  find under the lock, so an observed-true-but-missing entry correctly falls back to sparse), but the
  early store buys nothing, muddies the memory-ordering reasoning, and invites a future maintainer to
  treat the atomic as a data-ready flag when it's only an advisory fast-path skip hint. Fix: delete
  the early if (size > 0) store(true) line; keep only the post-population store.
  2. Test coverage doesn't match the PR's claims — l2_cache_test.cpp + CMakeLists.txt. Three problems:
   (a) the file is misnamed — l2_cache_test.cpp contains only a SparseMemory threading test, nothing
  about L2 cache; (b) the new cache.h helpers (allocate_with_data, invalidate_all_vmids,
  line_data_for_read) have zero test coverage; (c) the one threading test exercises only the easy case
   the PR body names — every thread writes disjoint pages, so same-page stripe contention (the actual
  hard part) is never stressed, and there's no multi-page block test (which would answer the BLOCKER's
   contract question). Plus RJ_ENABLE_TSAN is OFF by default, so a lost update passes silently unless
  CI explicitly enables it. For a PR whose "entire point" is thread-safety, this is under-tested. Fix:
   rename the file; add a same-page concurrent-writer test, a multi-page block test, and cache.h
  invalidation/allocate tests; confirm CI runs under TSan.

  MINOR

  - Block API uses raw uint8_t*+size_t instead of std::span — diverges from the sibling
  GpuMemory::read_block(std::span<...>) and, because GpuMemory : public SparseMemory, the same-named
  methods with different signatures create a name-hiding footgun (base read_block unreachable without
  qualification). Give SparseMemory's block methods std::span signatures.
  - for_each_page/num_pages acquire all 1024 stripe shared-locks — std::shared_mutex isn't recursive,
  so any future caller holding a stripe lock and then calling these deadlocks. No such nested caller
  today; add a @note (or debug guard) that they must not be called while holding a stripe lock.

emulation/rocjitsu/lib/simdojo/include/simdojo/components/sparse_memory.h
```diff
@@ -267,65 +227,91 @@ class SparseMemory : public Component {
...
+    std::memcpy(&get_page_locked(stripe, addr)[addr & PAGE_MASK], src, size);
+  }
+
+  mutable std::array<PageStripe, NUM_PAGE_STRIPES> page_stripes_;
 
   /// @brief Page-indexed host range map. Key = page_number (gpu_va >> 12),
   /// value = host_ptr for the start of that page. O(1) lookup per access.
   mutable std::shared_mutex host_range_mutex_;
   std::unordered_map<uint64_t, uint8_t *> host_page_map_;
+  std::atomic<bool> host_pages_mapped_{false};
```
newling: My agent has informed me that host mapped paged can possibly be removed altogether. My understanding is that in GpuMemory there is another concept of host memory (behind virtual memory) that is definitely needed, but for SparseMemory, it isn't used anywhere. 

My agent has a POC showing it can be removed https://github.com/ROCm/rocm-systems/pull/8726. I wonder if you need this in an upcoming PR, otherwise maybe we can consider removing it in the interest of simplicity?

newling (commented):
PageStripe idea makes sense to me. This ensures that thread writing to one page doesn't make threads reading/writing on other pages freeze (assuming those pages are in different stripes to the one the first thread is writing to). It'll be interesting to know if this unlocks performance anywhere, but that'll have to wait until we have actual multithreaded reads/writes happening. 

Agent review has a few suggestions, if you could please point your agent at this: https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr8702_final.md

lialan:
@atgutier Documented per-page atomicity, removed the legacy host map, switched to `std::span`, eliminated all-stripe locking, expanded coverage, and added focused TSan CI. Full build and 1,368 tests pass.

lialan:
@newling Removed the legacy host map, added VMID-aware lookup, clarified atomicity, improved hashing, and expanded tests. Full build and focused TSan pass.

emulation/rocjitsu/lib/simdojo/include/simdojo/components/sparse_memory.h
lialan: Removed the unused `SparseMemory` host map and moved host-range lookup to VMID-aware `GpuMemory`, with coverage for contiguous and non-contiguous mappings.

.github/workflows/rocjitsu-corpus-tests.yml
```diff
@@ -49,31 +49,48 @@ jobs:
...
             build_type: RelWithDebInfo
             enable_asan: 1
             enable_ubsan: 1
             enable_tsan: 0
             enable_msan: 0
             c_compiler: gcc
             cxx_compiler: g++
+          - name: tsan
+            allow_failure: false
+            # The simulator-wide stress tests exceed TSan's fixed 128-lock
```
newling, line 85: Alternatively is it possible to reduce the number of threads in the test to range [64, 128)? At a high level it seems to me that this multithreading work would benefit from TSan checks.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
```diff
@@ -143,6 +143,56 @@ class GpuMemory : public simdojo::SparseMemory {
 
   uint8_t *translate_debug(uint64_t addr, uint32_t vmid) const { return translate(addr, vmid); }
 
+  /// @brief Find the contiguous host range containing a VMID-scoped GPU VA.
+  /// @details KFD dispatches use per-process page tables. Kernel-symbol
+  /// resolution needs a daemon-accessible host pointer range so it can scan
+  /// backward from the kernel descriptor to the loaded ELF header.
+  std::pair<uint64_t, uint64_t> find_host_range(uint64_t addr, uint32_t vmid) const {
```
newling, line 150: (can ignore my commentary here) It seems like the precedent is to have all methods inlined in the class header, so this is following the code style here. In pre-agent times I would have put the impl in a separate .cpp, I'm not sure what changed.

newling (commented):
Looks good, but give me a minute to do a final agent review to see if there's anything worth highlighting that I've missed. 

My only nit is regarding the test exclusion.

.github/workflows/rocjitsu-corpus-tests.yml
newling: I misunderstood this. I thought that the test_regex here was an exclusion, but it is an inclusion list. 

I think it'd be better to write it as an exclusion list, so that future tests will be included by default. 

Chatting with my agent he suggested I post this:

 Could this TSan selection be expressed as an exclusion list instead of an inclusion list? The current `-R '^SparseMemory(Threading)?Test\.'` only runs the SparseMemory tests, so the other new tests in this PR are not covered by TSan. If only a few simulator stress tests exceed TSan’s lock-tracker limit, it would
  be better to run the normal rocjitsu test set under TSan and exclude just those known-incompatible tests with `-E`. That way future tests get TSan coverage by default.

  If the incompatible set is large or unstable, the current include-list may be pragmatic, but then the job should be named/commented as a narrow tsan-sparse-memory gate rather than a general TSan job.

newling (approved):
My agent doesn't have anything to report worth raising ([here](https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr8702.md) is its review). 

LGTM (but would like the tsan changes finalized). 

FYI @amd-eochoalo changes to rocjitsu-corpus-tests.yml here it might be useful to see

.github/workflows/rocjitsu-corpus-tests.yml
lialan: Renamed the job to `tsan-sparse-memory` to make the intentionally narrow scope explicit. The full simulator suite is not currently viable under TSan because lock-heavy stress tests exceed its fixed 128-lock tracker.

PR #8703, opened 2026-07-16
fix(rocjitsu): make shared GPU state concurrency-safe

lialan:
# fix(rocjitsu): make shared GPU state concurrency-safe

ISSUE ID: #6447

> [!IMPORTANT]
> This is **PR 2 of 6** in the parallel CPU simulation stack.
> It depends on the concurrent Simdojo primitives in
> `users/lialan/rocjitsu_parallel_cpu_01_simdojo`.
>
> Review only
> `users/lialan/rocjitsu_parallel_cpu_01_simdojo...users/lialan/rocjitsu_parallel_cpu_02_gpu_state`
> for this layer.

## Stack

| # | Layer | Branch | PR |
|---|---|---|---|
| 1 | Concurrent Simdojo memory primitives | `users/lialan/rocjitsu_parallel_cpu_01_simdojo` | #8702 |
| **2** | **Shared Rocjitsu GPU-state concurrency** | `users/lialan/rocjitsu_parallel_cpu_02_gpu_state` | **this PR** |
| 3 | XCD execution partitioning | `users/lialan/rocjitsu_parallel_cpu_03_xcd` | #8705 |
| 4 | Compositional plugin threading policy | `users/lialan/rocjitsu_parallel_cpu_04_plugins` | #8706 |
| 5 | Functional CU dispatch pool | `users/lialan/rocjitsu_threadpool_cpu` | #6962 |
| 6 | Functional-mode performance work | `users/lialan/rocjitsu_parallel_cpu_06_perf` | #8965 |

## Summary

Make GPU state shared by concurrent execution paths safe to access.

- Track KFD page-table generations and invalidate stale translations.
- Make `GpuMemory` translation and memory-type caches concurrency-safe.
- Synchronize shared L2 state across concurrent CPU execution.
- Add GPU-memory and cache concurrency regressions.

## Why

The XCD and CU-dispatch layers above this one execute against shared GPU state.
Their correctness depends on cache, memory, and address-translation state being
safe before any worker policy is enabled.

## Validation

- Built normal and TSan `rocjitsu_tests`.
- Passed focused `GpuMemory`, `GpuMemoryThreading`, `SparseMemory`,
  `SparseMemoryThreading`, `L2Cache`, and `L2CacheThreading` tests.
- Passed the full memory-threading TSan gate: 11 tests.
- Passed the full non-benchmark `rocjitsu_tests` suite: 1,473 passed with one
  pre-existing skip.
- Passed the layer-range pre-commit sweep.

### DIFF
--- a/.github/workflows/rocjitsu-corpus-tests.yml
+++ b/.github/workflows/rocjitsu-corpus-tests.yml
@@ -97,9 +97,9 @@ jobs:
             allow_failure: false
             run_static_asan_launch: true
             # The simulator-wide stress tests exceed TSan's fixed 128-lock
-            # deadlock-tracker capacity; keep the race gate on this change's
-            # concurrent SparseMemory surface.
-            test_regex: '^(SparseMemory(Threading)?Test|GpuMemoryThreadingTest)\.'
+            # deadlock-tracker capacity; keep the race gate focused on the
+            # concurrent memory surfaces hardened by this change.
+            test_regex: '^(SparseMemory(Threading)?Test|GpuMemoryThreadingTest|L2CacheThreadingTest)\.'
             build_type: RelWithDebInfo
             enable_asan: 0
             enable_ubsan: 0
--- a/.github/workflows/rocjitsu-corpus-tests.yml
+++ b/.github/workflows/rocjitsu-corpus-tests.yml
@@ -99,7 +99,7 @@ jobs:
             # The simulator-wide stress tests exceed TSan's fixed 128-lock
             # deadlock-tracker capacity; keep the race gate on this change's
             # concurrent SparseMemory surface.
-            test_regex: '^SparseMemory(Threading)?Test\.'
+            test_regex: '^(SparseMemory(Threading)?Test|GpuMemoryThreadingTest)\.'
             build_type: RelWithDebInfo
             enable_asan: 0
             enable_ubsan: 0
--- a/docs/superpowers/specs/2026-07-24-msc-atomic-alias-coherence-design.md
+++ b/docs/superpowers/specs/2026-07-24-msc-atomic-alias-coherence-design.md
@@ -1,0 +1,94 @@
+# MSC Atomic Alias Coherence Design
+
+## Context
+
+`L2Cache::atomic_rmw` acquires the device-wide atomic boundary before accessing
+backing memory. The boundary flushes every registered L2 and then advances the
+global coherence epoch while holding the L2 maintenance locks.
+
+The direct-memory path therefore observes the flushed value. The linked-port
+path instead reads through `MemorySideCache` (MSC). MSC is write-through and
+keys clean cache lines by `(VMID, GPU virtual address)`. If two such keys alias
+the same host storage, a clean line under one key can remain stale after another
+key publishes dirty L2 data. An atomic RMW through the stale key can consequently
+use the old value.
+
+## Goals
+
+- Make linked-port atomic RMWs observe the authoritative value after all L2
+  dirty data has been published.
+- Ensure stale MSC aliases are not visible after a coherence epoch change.
+- Preserve the existing device-wide atomic serialization and L2 locking model.
+- Avoid scanning or invalidating the entire 128 MiB MSC on every atomic.
+- Add regression coverage for two L2s sharing one MSC and HBM through aliased
+  VMID/virtual-address mappings.
+- Remove the unused `register_cache` helper so there is one L2 registration
+  implementation and one ordering invariant.
+
+## Design
+
+Each MSC cache tag will record the device-coherence epoch at which its line was
+filled or refreshed. MSC already protects tag lookup and replacement with the
+corresponding stripe lock, so the epoch metadata will use that same protection.
+
+On an MSC access:
+
+1. Look up the `(VMID, GPU virtual address)` cache line under its stripe lock.
+2. Compare the tag's recorded epoch with
+   `DeviceCacheCoherence::current_epoch()`.
+3. If the epochs differ, require the old entry to be clean, invalidate it, and
+   refill the line from backing HBM.
+4. Stamp a newly filled or refreshed entry with the current epoch.
+5. Continue the requested read or write.
+
+The atomic boundary's existing order remains unchanged:
+
+1. Acquire the global exclusive coherence lock.
+2. Lock every registered L2 maintenance mutex in sorted address order.
+3. Flush dirty L2 data through MSC to HBM.
+4. Advance the global coherence epoch.
+5. Perform the linked-port RMW while the boundary remains held.
+
+After step 4, any MSC alias retained from an earlier epoch is stale. The RMW
+read therefore invalidates and refills that alias from HBM, where step 3
+published the authoritative value. Other old aliases are invalidated lazily on
+their next access.
+
+MSC is write-through, so an epoch-stale entry is expected to be clean and can
+be discarded without losing data. No MSC registration or new global lock is
+needed.
+
+## Alternatives
+
+### Eagerly flush and invalidate the entire MSC
+
+Registering MSC with the atomic boundary and calling `flush_all()` would be
+simple, but it scans 1,048,576 tags and locks all 256 stripes for every atomic.
+That cost is disproportionate to an individual RMW.
+
+### Bypass MSC only for atomic backing accesses
+
+Sending the RMW directly to HBM would obtain the correct operand, but clean MSC
+aliases could remain stale and be returned after the atomic. A generation or
+invalidation mechanism would still be required, while the backing protocol
+would become more complex.
+
+## Validation
+
+Add a regression with two L2 caches linked to one MSC and HBM controller:
+
+1. Map different `(VMID, GPU virtual address)` pairs to the same host word.
+2. Read through the second alias to retain a clean MSC value of `0`.
+3. Dirty `40` through the first L2 without immediately publishing it.
+4. Increment through the second L2's atomic RMW.
+5. Verify the callback observes `40`, the atomic result and host word are `41`,
+   and subsequent reads through both aliases return `41`.
+
+Run the focused L2/MSC tests in Release and under TSan, followed by the relevant
+non-benchmark test suite.
+
+Measure end-to-end performance against commit `b685943638` using representative
+large kernels from `rocjitsu-test-corpus`, including a large softmax, a large
+pack/unpack case, and a large matrix multiplication. Use the same pinned-CPU,
+alternating-run methodology as the previously published PR benchmark where the
+harness permits it.
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h
@@ -199,6 +199,7 @@ public:
     auto *base = static_cast<uint8_t *>(host_ptr);
     for (size_t off = 0; off < size; off += kPageSize)
       page_table_[(gpu_va + off) >> kPageShift] = {base + off, mtype};
+    page_table_generation_.fetch_add(1, std::memory_order_release);
   }
 
   /// @brief Unmap pages from this process's GPU page table.
@@ -206,10 +207,17 @@ public:
     std::unique_lock lock(page_table_mutex_);
     for (size_t off = 0; off < size; off += kPageSize)
       page_table_.erase((gpu_va + off) >> kPageShift);
+    page_table_generation_.fetch_add(1, std::memory_order_release);
   }
 
   mutable std::shared_mutex page_table_mutex_;
   PageTable page_table_;
+
+  /// @brief Page table version counter, bumped (release) on every map/unmap.
+  /// @details GpuMemory keeps per-thread TLB-like translation caches keyed by
+  ///          this generation; a mismatch on load (acquire) invalidates the
+  ///          cached entry and forces a fresh page-table walk.
+  std::atomic<uint64_t> page_table_generation_{1};
 
   // -- Per-process state --
 
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h
@@ -252,7 +252,7 @@ public:
   }
 
   /// @brief Return the mutation counter used by GpuMemory translation caches.
-  std::atomic<uint64_t> *page_table_generation() { return &page_table_generation_; }
+  const uint64_t *page_table_generation() const { return &page_table_generation_; }
 
   mutable std::shared_mutex page_table_mutex_;
   PageTable page_table_;
@@ -302,15 +302,13 @@ public:
   DebugSession debug_session_;
 
 private:
-  void publish_page_table_mutation_locked() {
-    page_table_generation_.fetch_add(1, std::memory_order_release);
-  }
+  void publish_page_table_mutation_locked() { ++page_table_generation_; }
 
-  /// @brief Page table version counter, bumped (release) on every PTE mutation.
+  /// @brief Page table version counter, bumped on every PTE mutation.
   /// @details GpuMemory keeps per-thread TLB-like translation caches keyed by
-  ///          this generation; a mismatch on load (acquire) invalidates the
-  ///          cached entry and forces a fresh page-table walk.
-  std::atomic<uint64_t> page_table_generation_{1};
+  ///          this generation; all reads and writes occur while holding
+  ///          page_table_mutex_, so the counter itself does not need atomics.
+  uint64_t page_table_generation_{1};
 };
 
 } // namespace rocjitsu
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h
@@ -202,7 +202,7 @@ public:
     // Keep publication in the page-table critical section. Cached readers
     // validate this generation while holding the shared side of the same lock;
     // publishing after unlock would permit a stale-cache hit in between.
-    page_table_generation_.fetch_add(1, std::memory_order_release);
+    publish_page_table_mutation_locked();
   }
 
   /// @brief Unmap pages from this process's GPU page table.
@@ -212,7 +212,43 @@ public:
       page_table_.erase((gpu_va + off) >> kPageShift);
     // See map_pages(): the mutation and generation publication are one
     // page-table critical section by design.
-    page_table_generation_.fetch_add(1, std::memory_order_release);
+    publish_page_table_mutation_locked();
+  }
+
+  /// @brief Replace mapped host pages while preserving their other PTE fields.
+  /// @details The mutation and cache-generation publication occur under one
+  /// page-table critical section. Only entries still pointing at the expected
+  /// old page are changed.
+  void remap_page_host_ptrs(uint64_t gpu_va, void *old_host_ptr, void *new_host_ptr, size_t size) {
+    std::unique_lock lock(page_table_mutex_);
+    auto *old_base = static_cast<uint8_t *>(old_host_ptr);
+    auto *new_base = static_cast<uint8_t *>(new_host_ptr);
+    bool changed = false;
+    for (size_t off = 0; off < size; off += kPageSize) {
+      auto it = page_table_.find((gpu_va + off) >> kPageShift);
+      if (it != page_table_.end() && it->second.host_ptr == old_base + off &&
+          it->second.host_ptr != new_base + off) {
+        it->second.host_ptr = new_base + off;
+        changed = true;
+      }
+    }
+    if (changed)
+      publish_page_table_mutation_locked();
+  }
+
+  /// @brief Update the MTYPE of mapped pages and invalidate cached PTE copies.
+  void set_page_mtype(uint64_t gpu_va, size_t size, amdgpu::Mtype mtype) {
+    std::unique_lock lock(page_table_mutex_);
+    bool changed = false;
+    for (size_t off = 0; off < size; off += kPageSize) {
+      auto it = page_table_.find((gpu_va + off) >> kPageShift);
+      if (it != page_table_.end() && it->second.mtype != mtype) {
+        it->second.mtype = mtype;
+        changed = true;
+      }
+    }
+    if (changed)
+      publish_page_table_mutation_locked();
   }
 
   /// @brief Return the mutation counter used by GpuMemory translation caches.
@@ -266,7 +302,11 @@ public:
   DebugSession debug_session_;
 
 private:
-  /// @brief Page table version counter, bumped (release) on every map/unmap.
+  void publish_page_table_mutation_locked() {
+    page_table_generation_.fetch_add(1, std::memory_order_release);
+  }
+
+  /// @brief Page table version counter, bumped (release) on every PTE mutation.
   /// @details GpuMemory keeps per-thread TLB-like translation caches keyed by
   ///          this generation; a mismatch on load (acquire) invalidates the
   ///          cached entry and forces a fresh page-table walk.
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h
@@ -199,6 +199,9 @@ public:
     auto *base = static_cast<uint8_t *>(host_ptr);
     for (size_t off = 0; off < size; off += kPageSize)
       page_table_[(gpu_va + off) >> kPageShift] = {base + off, mtype};
+    // Keep publication in the page-table critical section. Cached readers
+    // validate this generation while holding the shared side of the same lock;
+    // publishing after unlock would permit a stale-cache hit in between.
     page_table_generation_.fetch_add(1, std::memory_order_release);
   }
 
@@ -207,17 +210,16 @@ public:
     std::unique_lock lock(page_table_mutex_);
     for (size_t off = 0; off < size; off += kPageSize)
       page_table_.erase((gpu_va + off) >> kPageShift);
+    // See map_pages(): the mutation and generation publication are one
+    // page-table critical section by design.
     page_table_generation_.fetch_add(1, std::memory_order_release);
   }
 
+  /// @brief Return the mutation counter used by GpuMemory translation caches.
+  std::atomic<uint64_t> *page_table_generation() { return &page_table_generation_; }
+
   mutable std::shared_mutex page_table_mutex_;
   PageTable page_table_;
-
-  /// @brief Page table version counter, bumped (release) on every map/unmap.
-  /// @details GpuMemory keeps per-thread TLB-like translation caches keyed by
-  ///          this generation; a mismatch on load (acquire) invalidates the
-  ///          cached entry and forces a fresh page-table walk.
-  std::atomic<uint64_t> page_table_generation_{1};
 
   // -- Per-process state --
 
@@ -264,6 +266,11 @@ public:
   DebugSession debug_session_;
 
 private:
+  /// @brief Page table version counter, bumped (release) on every map/unmap.
+  /// @details GpuMemory keeps per-thread TLB-like translation caches keyed by
+  ///          this generation; a mismatch on load (acquire) invalidates the
+  ///          cached entry and forces a fresh page-table walk.
+  std::atomic<uint64_t> page_table_generation_{1};
 };
 
 } // namespace rocjitsu
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
@@ -370,7 +370,8 @@ int SimulatedKfd::open() {
   proc->event_state_.reset();
   for (auto &g : gpus_) {
     if (auto *mem = g.soc ? g.soc->memory() : nullptr) {
-      mem->register_process(pid, &proc->page_table_, &proc->page_table_mutex_);
+      mem->register_process(pid, &proc->page_table_, &proc->page_table_mutex_,
+                            &proc->page_table_generation_);
       if (!daemon_mode_)
         mem->set_passthrough(true);
     }
@@ -431,7 +432,8 @@ uint32_t SimulatedKfd::open_process(pid_t client_pid) {
     proc->event_state_.reset();
     for (auto &g : gpus_) {
       if (auto *mem = g.soc ? g.soc->memory() : nullptr) {
-        mem->register_process(pid, &proc->page_table_, &proc->page_table_mutex_);
+        mem->register_process(pid, &proc->page_table_, &proc->page_table_mutex_,
+                              &proc->page_table_generation_);
         if (client_pid > 0)
           mem->set_process_client_pid(pid, client_pid);
       }
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
@@ -371,7 +371,7 @@ int SimulatedKfd::open() {
   for (auto &g : gpus_) {
     if (auto *mem = g.soc ? g.soc->memory() : nullptr) {
       mem->register_process(pid, &proc->page_table_, &proc->page_table_mutex_,
-                            &proc->page_table_generation_);
+                            proc->page_table_generation());
       if (!daemon_mode_)
         mem->set_passthrough(true);
     }
@@ -433,7 +433,7 @@ uint32_t SimulatedKfd::open_process(pid_t client_pid) {
     for (auto &g : gpus_) {
       if (auto *mem = g.soc ? g.soc->memory() : nullptr) {
         mem->register_process(pid, &proc->page_table_, &proc->page_table_mutex_,
-                              &proc->page_table_generation_);
+                              proc->page_table_generation());
         if (client_pid > 0)
           mem->set_process_client_pid(pid, client_pid);
       }
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
@@ -1769,17 +1769,7 @@ int SimulatedKfd::ipc_export_handle_ioctl(KfdProcess &proc, void *arg) {
                          "true sharing)");
       }
 
-      {
-        std::unique_lock ptlk(proc.page_table_mutex_);
-        auto *old_base = static_cast<uint8_t *>(alloc.host_ptr);
-        auto *new_base = static_cast<uint8_t *>(new_host_ptr);
-        for (size_t off = 0; off < alloc.size; off += KfdProcess::kPageSize) {
-          uint64_t page_num = (alloc.gpu_va + off) >> KfdProcess::kPageShift;
-          auto pt_it = proc.page_table_.find(page_num);
-          if (pt_it != proc.page_table_.end() && pt_it->second.host_ptr == old_base + off)
-            pt_it->second.host_ptr = new_base + off;
-        }
-      }
+      proc.remap_page_host_ptrs(alloc.gpu_va, alloc.host_ptr, new_host_ptr, alloc.size);
 
       if (alloc.host_ptr_owned)
         libc_passthrough().munmap(alloc.host_ptr, alloc.size);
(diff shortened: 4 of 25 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h
```diff
@@ -230,18 +230,26 @@ class KfdProcess {
     auto *base = static_cast<uint8_t *>(host_ptr);
     for (size_t off = 0; off < size; off += kPageSize)
       page_table_[(gpu_va + off) >> kPageShift] = {base + off, mtype};
+    page_table_generation_.fetch_add(1, std::memory_order_release);
```
atgutier: Not sure if it matters too much, but should we move the lock/critical section into its own scope so we don't hold the lock when doing this atomic?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h
```diff
@@ -230,18 +230,26 @@ class KfdProcess {
...
   }
 
   /// @brief Unmap pages from this process's GPU page table.
   void unmap_pages(uint64_t gpu_va, size_t size) {
     std::unique_lock lock(page_table_mutex_);
     for (size_t off = 0; off < size; off += kPageSize)
       page_table_.erase((gpu_va + off) >> kPageShift);
+    page_table_generation_.fetch_add(1, std::memory_order_release);
```
atgutier: Same comment here: do we need to hold the lock when performing this atomic?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
```diff
@@ -36,7 +37,9 @@ namespace amdgpu {
 /// issuing wave through the memory hierarchy.
 class GpuMemory : public simdojo::SparseMemory {
 public:
-  explicit GpuMemory(std::string name) : simdojo::SparseMemory(std::move(name)) {
+  explicit GpuMemory(std::string name)
+      : simdojo::SparseMemory(std::move(name)),
+        instance_id_(next_instance_id_.fetch_add(1, std::memory_order_relaxed)) {
```
atgutier, line 49: Should this be an atomic store of 0?

atgutier (commented):
I have some questions, and the agent has this:

 MAJOR — must fix (this is the "make shared state safe" layer, so these matter now)

  1. translate()→memcpy TOCTOU: a real use-after-free window — gpu_memory.h (read8/16/32/64, write*,
  read_block, write_block). translate() releases both page_table_mutex_ and vmid_mutex_ and returns a
  raw host_ptr; callers then memcpy to/from it with no lock. A concurrent unmap_pages (unique lock,
  erases the entry, bumps the gen, frees/reuses the backing allocation) between the translate and the
  memcpy → torn read / use-after-free / silent corruption. Pre-existing (identical in the base), but
  this PR's explicit charter is to make shared state race-free before PR3 (XCD partitioning) and PR5
  (CU-dispatch pool) turn on workers that hit exactly these block paths while a control thread can
  unmap. Fix: either (a) document + assert a hard invariant that page-table teardown is quiesced (all
  GPU work drained) before any unmap/unregister, or (b) perform the copy inside page_table_mutex_
  (shared) rather than returning a bare pointer.
  2. L1 Vector Cache and LDS got NO synchronization — l1_vector_cache.h/lds.h. The PR summary claims
  "Synchronize L1, L2, and LDS," but only L2 gained a mutex. L1V/LDS are per-CU members, and this PR
  adds unguarded mutable state to L1V (cached_read_line_*, non-atomic read_count_/store_count_) that's
   mutated on every load/store; line_data_for_read also mutates LRU. If PR5 runs multiple
  waves/workgroups on one CU concurrently, every one is a data race. Fix: if L1/LDS stay strictly
  thread-confined to one worker per CU, fix the PR description (it overclaims) and add a @par Thread
  safety note stating they're not thread-safe and require per-CU confinement (mirroring L2's
  documented contract); if PR5 shares a CU across workers, they need the same locking as L2 before
  that lands.
  3. cached_read_line_data_ shortcut has an unenforced residency invariant — l1_vector_cache.h. The
  cached raw pointer into cache storage is validated only by (addr, vmid), not by "still resident in
  that way." A plain LRU eviction of the cached line by a different address in the same set (via
  allocate_with_data, which doesn't go through invalidate) leaves cached_read_line_data_ pointing at a
   way now owned by a different line → stale/aliased read. Single-threaded-correct today but fragile.
  Fix: drop the shortcut (it duplicates the LRU fast path) or re-validate residency / clear it on any
  eviction, not just explicit invalidate.

  MINOR

  - L2 multi-line read/write is per-line locked, not atomic across lines — l2_cache.cpp; same
  un-documented block-atomicity gap as #8702. Acceptable for functional sim, but the "thread-safe"
  Doxygen should say individual lines are atomic while a multi-line op is not a consistent snapshot.
  - translate/pte_mtype duplicate ~35 lines each (shared_lock + fast-path + walk + cache-fill,
  differing only in payload) — factor into a templated cached_walk so the two invalidation paths can't
   drift (drift would be a correctness bug).
  - Concurrency tests exercise mostly disjoint state — the real races this PR fixes aren't stressed:
  no translate-vs-unmap TOCTOU, no contended same-line LRU-mutation, no translation-cache invalidation
   under a concurrent map_pages gen bump (the whole point of the generation protocol — the
  invalidation test is single-threaded). The atomic_rmw same-line serialization tests are good. Add a
  readers-vs-remapper test (assert readers see valid ptr or nullptr, never stale) under TSan (no TSan
  wiring seen for these targets — verify).
  - global_atomic_mutex is a process-wide static array of 2048 mutexes — intended (cross-XCD atomic
  serialization, a test depends on it) but undocumented as a deliberate singleton; add a comment so
  nobody "fixes" it into a member and breaks cross-L2 atomicity.

 NIT

  page_table_generation_ is a public KMD member plumbed into the VM layer by raw atomic* (leaks KMD
  internals — an accessor would be cleaner); C-style UINT64_MAX vs std::numeric_limits; the non-atomic
   L1 counters (fine if per-CU confinement is documented).

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h
lialan: Kept inside the lock intentionally. Publishing after unlock could let readers reuse a stale cached PTE.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h
lialan: Same here. Erase and generation update must remain one critical section to prevent stale cache reuse.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
lialan: `fetch_add` is intentional. Each `GpuMemory` lifetime needs a unique token to reject stale TLS cache entries.

lialan: To be more specific: The atomic is a global counter; each `GpuMemory` gets a unique immutable ID. Storing 0 would reset the counter and  could make a new object match stale TLS cache entries.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/lds.h
```diff
@@ -136,6 +143,11 @@ class Lds : public simdojo::MemoryInterface {
   void vector_load(const uint64_t *addrs, uint64_t lane_mask, uint32_t elem_size,
                    uint32_t num_elems, uint8_t *dst, uint32_t base_offset = 0) {
     uint32_t stride = num_elems * elem_size;
+    uint32_t bulk_base = 0;
+    if (contiguous_full_wave_access(addrs, lane_mask, stride, base_offset, bulk_base)) {
```
kuhar: It would be good to add focused regression coverage for the new LDS bulk memcpy path. A standalone probe exercises full-wave load/store and partial/non-contiguous fallbacks successfully, but the test suite has no direct Lds::vector_load/vector_store coverage. Please cover full-wave load/store plus the stride-zero, partial-mask, non-contiguous, out-of-bounds, and nonzero-base-offset fallbacks so the fast path stays equivalent to the scalar path.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l2_cache.cpp
```diff
@@ -219,9 +244,24 @@ void L2Cache::flush_all(uint32_t vmid) {
       max_addr = line_addr;
   });
   util::Logger::vm("L2 flush: ", dirty_count, " dirty lines [0x", std::hex, min_addr, "-0x",
-                   max_addr, "]", std::dec, " total_writes=", write_count_);
+                   max_addr, "]", std::dec,
+                   " total_writes=", write_count_.load(std::memory_order_relaxed));
   cache_.invalidate_all();
 }
 
+void L2Cache::invalidate_range(uint64_t addr, uint32_t size) {
```
kuhar: It would be useful to add a normal multi-set test for invalidate_range(). The only current case starts at UINT64_MAX and touches one set, so it does not verify SetRangeLocks or that a middle subrange reloads from backing while adjacent cached lines remain intact. Please cover a multi-line subrange, plus the size == 0 early return.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
```diff
@@ -307,7 +307,8 @@ int SimulatedKfd::open() {
   proc->event_state_.reset();
   for (auto &g : gpus_) {
     if (auto *mem = g.soc ? g.soc->memory() : nullptr) {
-      mem->register_process(pid, &proc->page_table_, &proc->page_table_mutex_);
+      mem->register_process(pid, &proc->page_table_, &proc->page_table_mutex_,
```
kuhar, line 373: We should invalidate the new PTE cache when ipc_export_handle_ioctl() mutates existing page-table entries. The host_ptr rewrite at simulated_kfd.cpp:1611 and mtype rewrite at simulated_kfd.cpp:1650 both happen under page_table_mutex_, but neither increments page_table_generation_. A thread can therefore reuse a cached old pointer after the old mapping is unmapped at simulated_kfd.cpp:1616. Please route these updates through a generation-bumping helper or publish one generation increment in each mutation critical section.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l2_cache.h
```diff
@@ -195,21 +204,109 @@ class L2Cache : public simdojo::Component {
...
 
 private:
+  std::mutex &set_mutex(uint64_t addr) const { return set_mutexes_[CacheStore::set_index(addr)]; }
+
+  static std::mutex &global_atomic_mutex(uint64_t addr) {
+    // Deliberately process-wide rather than per-L2: separate XCD-local L2s can
+    // target the same backing address, so device-wide atomics must rendezvous
+    // on the same striped lock even when they originate from different L2s.
+    static std::array<std::mutex, NUM_SETS> mutexes;
+    return mutexes[CacheStore::set_index(addr)];
```
kuhar: We should key cross-L2 atomic arbitration by resolved backing identity, not the GPU-VA set. Two VAs, including VAs in different VMIDs, can map to the same host_ptr while hashing to different global_atomic_mutex() stripes, so concurrent RMWs can update the same storage under different locks. Please serialize at the backing-memory layer or lock by resolved backing address, and add an aliased-VA regression.

kuhar (commented):
The L2 maintenance paths need a TSan-visible locking strategy. Running build-clang-23-tsan/tests/rocjitsu_tests with the L2CacheThreadingTest.ConcurrentFlushAllPreservesDirtyWritebacks filter aborts at TSan’s 128-lock tracker limit because flush_all() tries to acquire all 2,048 set mutexes; the workflow therefore excludes every L2CacheThreadingTest. A single maintenance shared_mutex, shared for ordinary set operations and exclusive for whole-cache maintenance, would preserve exclusion with one tracked lock and let this concurrency code join the TSan gate.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/lds.h
lialan: Added tests for the full-wave `Lds::vector_load`/`vector_store` path and all the fallback cases you called out.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
lialan: Updated both IPC-export PTE mutations to bump the generation under the page-table lock.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l2_cache.cpp
lialan: Added a multi-set `invalidate_range()` test that checks the middle subrange reloads, adjacent lines stay cached, and zero-size is a no-op.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l2_cache.h
lialan: Okay, I moved the atomic locking into `GpuMemory`..

lialan:
> The L2 maintenance paths need a TSan-visible locking strategy. Running build-clang-23-tsan/tests/rocjitsu_tests with the L2CacheThreadingTest.ConcurrentFlushAllPreservesDirtyWritebacks filter aborts at TSan’s 128-lock tracker limit because flush_all() tries to acquire all 2,048 set mutexes; the workflow therefore excludes every L2CacheThreadingTest. A single maintenance shared_mutex, shared for ordinary set operations and exclusive for whole-cache maintenance, would preserve exclusion with one tracked lock and let this concurrency code join the TSan gate.

Addressed those concerns.

.github/workflows/rocjitsu-corpus-tests.yml
```diff
@@ -83,10 +83,9 @@ jobs:
             cxx_compiler: g++
           - name: tsan-sparse-memory
             allow_failure: false
-            # The simulator-wide stress tests exceed TSan's fixed 128-lock
-            # deadlock-tracker capacity; keep the race gate on this change's
-            # concurrent SparseMemory surface.
-            test_regex: '^SparseMemory(Threading)?Test\.'
+            # Keep the race gate focused on the concurrent memory surfaces
+            # hardened by this change.
```
newling: > by this change

Can you make this comment general to the project rather the PR?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l2_cache.cpp
```diff
@@ -76,14 +75,16 @@ void L2Cache::ensure_line(uint64_t addr, uint32_t vmid) {
...
   uint32_t copied = 0;
   if (mtype == Mtype::UC) {
     while (copied < size) {
       const uint64_t ea = addr + copied;
       const uint32_t line_offset = CacheStore::line_offset(ea);
       const uint32_t chunk = std::min(size - copied, LINE_SIZE - line_offset);
 
-      flush_line(ea, vmid);
+      std::lock_guard set_lock(set_mutex(ea));
+      flush_line_locked(ea, vmid);
```
newling, line 188: L2 cache entries appear to be keyed by (VMID, 128-byte-aligned GPU virtual address), rather than by the resolved backing-memory identity. If two different (VMID, VA) pairs map to the same backing line, are they intentionally treated as separate cache entries? If so, how are stale clean copies, invalidation, and dirty writeback handled across those aliases? The PR seems to resolve backing identity for atomics; please clarify whether ordinary L2 accesses intentionally have different aliasing semantics and add or point to relevant coverage.

newling:
Agent review: https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr8703.md. There are 2 [actionable items](https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr8703.md#actionable-items) highlighted. There are suggested tests outlined in the review (I haven't verified these). In the [suggestions section](https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr8703.md#suggestions) there are some things that look like pre-existing issues that this PR is building on top of, and a suggestion that this PR could be 2 PRs.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l2_cache.cpp
lialan: Ordinary L2 aliases intentionally stay separate, while atomics rendezvous by backing identity; I documented the distinction and added coverage for clean staleness, CC refetch, and dirty writeback.

.github/workflows/rocjitsu-corpus-tests.yml
lialan: Yep, updated it to describe the CPU-dispatch race gate generally.

lialan:
> Agent review: https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr8703.md. There are 2 [actionable items](https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr8703.md#actionable-items) highlighted. There are suggested tests outlined in the review (I haven't verified these). In the [suggestions section](https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr8703.md#suggestions) there are some things that look like pre-existing issues that this PR is building on top of, and a suggestion that this PR could be 2 PRs.

both actionable items are addressed. 

i can split out the L1/LDS changes if that is needed.

newling:
> i can split out the L1/LDS changes if that is needed.

Yes please 🙏

lialan:
@newling I split the L1/LDS and hot-path performance changes out of this PR into stacked draft #8965. #8703 is now limited to shared GPU-state concurrency, and I rebased #8705, #8706, and #6962 on top. The focused 48-test set passes and #8703's pre-commit is green. Could you take another look at #8703? The moved performance changes are ready for review in #8965 too.

newling:
@lialan thank you! I'll review https://github.com/ROCm/rocm-systems/pull/8703 and https://github.com/ROCm/rocm-systems/pull/8965 now.

docs/superpowers/specs/2026-07-21-pr8703-memory-invalidation-design.md
```diff
@@ -0,0 +1,53 @@
+# PR 8703 Memory Translation and Invalidation Design
```
newling: This file should presumably not be in the PR

lialan: Yep, removed it from this PR and restacked the downstream branches.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
```diff
@@ -216,91 +262,82 @@ class GpuMemory : public simdojo::SparseMemory {
...
-    }
     if (vmid > 0 && write_client_memory(addr, &val, 8, vmid))
       return;
     SparseMemory::write64(addr, val);
   }
 
 private:
+  static std::mutex &backing_atomic_mutex(uintptr_t key) {
+    static std::array<std::mutex, 4096> mutexes;
+    key >>= 3;
```
newling: Can you document these magic numbers?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
```diff
@@ -36,7 +41,11 @@ namespace amdgpu {
 /// issuing wave through the memory hierarchy.
 class GpuMemory : public simdojo::SparseMemory {
 public:
-  explicit GpuMemory(std::string name) : simdojo::SparseMemory(std::move(name)) {
+  explicit GpuMemory(std::string name)
+      : simdojo::SparseMemory(std::move(name)),
+        // A monotonic lifetime token prevents a thread-local cache from
```
newling: I'd have thought the liftetime of GpuMemory outlasts the lifetime of a compute thread.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h
```diff
@@ -235,15 +235,61 @@ class KfdProcess {
...
+        it->second.mtype = mtype;
+        changed = true;
+      }
+    }
+    if (changed)
+      publish_page_table_mutation_locked();
+  }
+
+  /// @brief Return the mutation counter used by GpuMemory translation caches.
+  std::atomic<uint64_t> *page_table_generation() { return &page_table_generation_; }
```
newling: If this could be `const  std::atomic<uint64_t> *page_table_generation() const` it'd be clearer how is rensponsible for bumping  the generation

newling: Morever does this need to be atomic? It looks like it always mutated when page_table_mutex_ is locked (so only 1 mutator)

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
```diff
@@ -54,19 +63,29 @@ class GpuMemory : public simdojo::SparseMemory {
...
-  void register_process(uint32_t pid, KfdProcess::PageTable *pt, std::shared_mutex *mu) {
+  /// @param generation Optional mutation counter used by translation caches.
+  ///        Omitting it disables the per-thread fast path for this page table.
+  void register_process(uint32_t pid, KfdProcess::PageTable *pt, std::shared_mutex *mu,
+                        std::atomic<uint64_t> *generation = nullptr) {
     util::Logger::cp("VMID_REG pid=", pid, " mem=0x", std::hex, reinterpret_cast<uintptr_t>(this),
                      std::dec, " pt_size=", pt->size());
     std::unique_lock lk(vmid_mutex_);
-    vmid_table_[pid] = {pt, mu};
+    vmid_table_[pid] = {pt, mu, 0, generation};
```
newling:
```suggestion
    vmid_table_[pid] = {pt, mu, /* what is this variable? */ 0, generation};
```

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
```diff
@@ -54,19 +63,29 @@ class GpuMemory : public simdojo::SparseMemory {
...
   /// @brief Unregister a process from the VMID table.
   void unregister_process(uint32_t pid) {
     util::Logger::cp("VMID_UNREG pid=", pid, " mem=0x", std::hex, reinterpret_cast<uintptr_t>(this),
                      std::dec);
     std::unique_lock lk(vmid_mutex_);
-    vmid_table_.erase(pid);
+    auto it = vmid_table_.find(pid);
+    if (it == vmid_table_.end())
+      return;
+    if (it->second.generation)
```
newling: What is the purpose of this page table generation increment? The page table doesn't seem to have changed here

newling (commented):
Thanks for splitting this again. I still haven't managed to review the whole PR, but have left a few comments inline. Nothing big. I'll follow this up again with another agent review. 

I guess it's inevitable that multi-threading a gpu emulator is going to complex, but I'm still missing the big picture of ownership between the emulated parts, and the functional vs cycle-accurate rules of engagement. I'd be tempted to throw some time and tokens at pinning down more exactly what we're emulating about L2: for functional and cycle-accurate modes, what are the expectations?

newling:
Agent presents one actionable item, looks like a bug: https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr8703_followup_2.md

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
lialan: Named and documented the stripe count, granule shift, hash folds, and PID salt.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
lialan: The TLS caches can outlive a `GpuMemory` on long-lived host threads, so I kept the lifetime token and clarified the comment.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h
lialan: Good point — it is now a `const uint64_t *`, with the plain counter protected by `page_table_mutex_`.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
lialan: Changed this to designated fields, so the zero is now explicitly `.client_pid = 0`.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
lialan: Right — this is the VMID binding generation, so I renamed it and only bump it when that binding changes.

lialan:
@newling Fixed.`atomic_rmw` now keeps the VMID/page-table locks and storage classification stable for the full fallback RMW, and the concurrent-map regression passes in regular and TSan runs.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l2_cache.cpp
```diff
@@ -219,9 +251,44 @@ void L2Cache::flush_all(uint32_t vmid) {
...
+  if (size == 0)
+    return;
+  std::shared_lock maintenance_lock(maintenance_mutex_);
+  const uint64_t line_start = CacheStore::line_address(addr);
+  const uint64_t requested_lines =
+      (static_cast<uint64_t>(CacheStore::line_offset(addr)) + size + LINE_SIZE - 1) / LINE_SIZE;
+  const uint64_t addressable_lines =
+      (std::numeric_limits<uint64_t>::max() - line_start) / LINE_SIZE + 1;
+  const uint64_t line_count = std::min(requested_lines, addressable_lines);
+  auto locks = lock_sets_for_range(line_start, line_count);
```
kuhar: This would be safer if wide invalidations did not acquire every set mutex at once. At current head, a TSan probe over 129 consecutive lines aborts at the deadlock detector 128-lock capacity, while the checked-in range tests cover at most three lines. Please use the exclusive maintenance lock for broad ranges or process sets incrementally, and add a TSan case spanning more than 128 sets.

emulation/rocjitsu/tests/amdgpu_vm_test.cpp
```diff
@@ -439,6 +447,377 @@ TEST(GpuMemoryTest, VmidMappedKernelSymbolUsesTranslatedHostPointer) {
...
+
+  KfdProcess process(kVmid);
+  memory.register_process(kVmid, &process.page_table_, &process.page_table_mutex_,
+                          process.page_table_generation());
+  memory.set_process_client_pid(kVmid, getpid());
+
+  // Mirror GpuMemory's private stripe selection to force the mapped and
+  // fallback identities onto different mutexes.
+  auto atomic_stripe = [](uintptr_t key) {
+    constexpr unsigned kGranuleShift = 3;
```
kuhar: I think it would help to stop this regression from reimplementing the private atomic-stripe hash. The test repeats the production granule, fold, stripe-count, and PID-salt values; if the hash changes, it can select keys that collide in the real implementation and silently stop reproducing the old race. Please expose a stable test helper or structure the synchronization so the test does not predict private stripe selection.

kuhar (commented):
Could you post benchmark numbers as you land these patches to make sure we are not accidentally introducing regressions?

Also, see https://github.com/ROCm/rocm-systems/pull/8703#discussion_r3630672756.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l2_cache.h
kuhar: The issue I am worried about is that cross-L2 arbitration still uses process-local host addresses rather than stable backing-object identity. At current head 91216eee30, IPC import maps the same memfd at a distinct MAP_SHARED address; two synchronized atomic_rmw increments through different mappings selected different stripes and left the shared value at 1 instead of 2. The existing CrossL2AtomicRmwAliasedVasIsSerialized test maps both VMIDs to the same backing.data() pointer, so it cannot exercise this path. We should key arbitration by backing object plus offset or serialize above the mappings, hold it across the dirty writeback and RMW, and cover two distinct memfd mappings with a resident dirty line.

lialan:
Benchmark results for the follow-up patch against `91216eee`, using 7 alternating runs pinned to CPUs 0-7 (median ns/op):

| Case | Baseline | Follow-up | Change |
|---|---:|---:|---:|
| Cross-L2 same-address atomic | 284.622 | 327.116 | +14.93% |
| Cross-L2 independent-address atomic | 652.464 | 325.367 | -50.13% |
| Invalidate 3 lines | 114.078 | 115.965 | +1.65% |
| Invalidate 129 lines | 1525.776 | 563.192 | -63.09% |

The independent-address baseline was noisy (400-764 ns/op), so I would treat that result as directional only. A stable `scaling_test` rerun at 1/2/4/8 threads stayed within 3% across vector, tiled, and MFMA workloads. The +14.9% same-address cost is the expected tradeoff from process-wide alias serialization; regular workload scaling showed no material regression. Release was 1,667 passed / 1 expected skip, and the focused TSan run was 38/38.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l2_cache.h
lialan: Fixed with one process-wide L2 lock held across dirty writeback and RMW, plus a distinct memfd-mapping regression.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l2_cache.cpp
lialan: Wide invalidations now use the exclusive maintenance lock, with a 129-line TSan regression.

emulation/rocjitsu/tests/amdgpu_vm_test.cpp
lialan: Reworked the test to verify page-table lock lifetime without duplicating the private hash.

newling:
A new actionable item in a new review: https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr8703_compendium.md#actionable-items-2

It looks like a scalar cache on one XCD and atomic write on another XCD need some way of synchronizing, on first glance the test the agent suggests makes sense to me

Fun times with multi-threading 🥵

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l2_cache.h
lialan: Extended this to a device-wide cache boundary with exact dirty-byte publication and lazy epoch invalidation.

lialan:
@newling Fixed. Atomics now use a device-wide cache boundary, publish exact dirty L2 bytes, and invalidate L1/L2 via epochs. Added cross-XCD scalar/vector and distinct-memfd regressions; Release is 1,682/1 and focused TSan is 36/36.

lialan:
Latest benchmark results for the final coherence patch `3fe4f4a11c` against its parent `b685943638`. The same benchmark helpers were used for both builds, with 7 alternating runs pinned to CPUs 0-7 (median ns/op):

| Case | Parent | PR head | Change |
|---|---:|---:|---:|
| Cross-L2 same-address atomic | 327.067 | 393.816 | +20.41% |
| Cross-L2 independent-address atomic | 330.641 | 394.094 | +19.19% |
| Atomic, 1 hierarchy | 98.483 | 99.789 | +1.33% |
| Atomic, 2 hierarchies | 99.031 | 107.312 | +8.36% |
| Atomic, 4 hierarchies | 97.688 | 111.785 | +14.43% |
| Atomic, 8 hierarchies | 95.964 | 135.485 | +41.18% |
| Scalar L1 hit | 6.688 | 12.707 | +90.00% |
| Vector L1 hit | 8.572 | 14.688 | +71.35% |
| Invalidate 3 lines | 116.937 | 118.555 | +1.38% |
| Invalidate 129 lines | 554.726 | 560.891 | +1.11% |

`scaling_test` was also rerun pinned to CPUs 0-7; each cell is the median of three internal runs. Across vector add, tiled matmul, and MFMA:

- 1 thread: +4.22% to +5.58%
- 2 threads: +2.36% to +5.46%
- 4 threads: -0.50% to +0.42%
- 8 threads: +2.83% to +3.26%

The microbenchmark cost is concentrated in tiny uncontended L1 hits and atomic boundaries with many registered hierarchies. Full simulated workloads stay within 5.6%, and the 4/8-thread cases stay within 3.3%.

Release: 1,798 passed / 1 expected skip. Focused TSan: 57/57.

kuhar:
@lialan Thanks for the table, very useful. Could you also run a few large kernels from the test corpus to show the e2e effect?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/device_cache_coherence.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+#include <algorithm>
+#include <cassert>
+#include <functional>
+
+namespace rocjitsu {
+namespace amdgpu {
+namespace {
+
+template <typename T> void register_cache(std::vector<T *> &caches, T *cache) {
```
kuhar: I think we should remove this unused `register_cache` helper, or route `register_l2_cache()` through a shared sorted-insert helper. The live registration path uses `lower_bound`, while this no-call-site helper appends, leaving two incompatible versions of the ordering invariant.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l2_cache.h
```diff
@@ -123,37 +139,41 @@ class L2Cache : public simdojo::Component {
...
+
+    if (backing_memory_) {
+      backing_read_transactions_.fetch_add(1, std::memory_order_relaxed);
+      backing_write_transactions_.fetch_add(1, std::memory_order_relaxed);
+      backing_memory_->atomic_rmw(addr, size, [&](uint8_t *target) { fn(target, 0); }, vmid);
+    } else {
+      assert((size == sizeof(uint32_t) || size == sizeof(uint64_t)) &&
+             "L2 atomic size must be 4 or 8 bytes");
+      std::array<uint8_t, sizeof(uint64_t)> target{};
+      send_backing(addr, target.data(), size, simdojo::MessageOp::READ, vmid);
```
kuhar, line 174: We should include the memory-side cache in this atomic boundary, or bypass/invalidate it for the backing RMW. On the linked-port path, MSC retains clean lines by `(vmid, GPU VA)`, so one alias can stay stale after another L2 publishes dirty data through a different key. A current-head probe with two L2s sharing one MSC/HBM and aliased VMID/VAs read 0 from the stale alias, published 40 through the other alias, then produced 1 instead of 41 after the atomic increment. The existing alias regressions use `set_backing_memory()` and bypass MSC; could we add the linked-port variant and make its RMW reach authoritative storage?

kuhar (commented):
A few comments

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/device_cache_coherence.cpp
lialan: Removed the unused helper.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l2_cache.h
lialan: Fixed. MSC lines now track the device coherence epoch and refresh stale aliases; added the linked-port regression.

lialan:
@kuhar I reran this against latest `origin/develop`with 7 alternating CPU-pinned runs: softmax +1.04%, pack i8 +0.59%, unpack +0.01%, and GEMM +0.65%. All correctness checks passed, and these deltas are within the noise range.

atgutier (approved):
Nice job!

docs/superpowers/specs/2026-07-24-msc-atomic-alias-coherence-design.md
```diff
@@ -0,0 +1,94 @@
+# MSC Atomic Alias Coherence Design
```
newling, line 1: This file slipped in again!

PR #8705, opened 2026-07-16
feat(rocjitsu): partition execution across XCDs

lialan:
# feat(rocjitsu): partition execution across XCDs

ISSUE ID: #6447

## Summary

Partition simulation topology along complete XCD ownership boundaries while
preserving the existing single-partition behavior.

- Clamp requested engine partitions to the number of XCDs visible to the VM.
- Assign complete XCD subtrees round-robin across single- and multi-GPU VMs.
- Keep topology components outside XCD subtrees on partition zero.
- Keep the generic FM partitioner as explicit opt-in `partition_balanced()`;
  multi-threaded engines must choose a partition policy before `create()`.
- Return `ROCJITSU_STATUS_UNSUPPORTED` when `rj_vm_step()` is used with
  multiple engine partitions; `rj_vm_run()` remains the parallel path.
- Document the threading and public API constraints.

The independent checkpoint restore fix was split into #9370.

## Safety constraints

- A single XCD subtree is never split across engine partitions.
- CUs sharing an XCD-local L2 cache remain on the same engine thread.
- External link owners used by manual topologies are retained once on
  partition zero.

## Validation

- Rebased as one commit onto current `origin/develop`
  (`5d3ca048215e6d15106581827ad16a46ba621fe9`).
- Built `rocjitsu_tests` and `scaling_test` in Release mode.
- Passed 21 focused C API, XCD partitioning, explicit-policy, termination,
  stress, and multithreaded workload tests.
- Completed the full 1–8 thread `scaling_test` sweep.
- Passed the full branch-diff pre-commit sweep.

### DIFF
--- a/emulation/mirage/rocjitsu_sys/src/lib.rs
+++ b/emulation/mirage/rocjitsu_sys/src/lib.rs
@@ -622,8 +622,10 @@ mod tests {
 
     #[test]
     fn status_codes_match_c_api() {
-        const C_STATUS_HEADER: &str =
-            include_str!("../../../rocjitsu/lib/rocjitsu/include/rocjitsu/base/rj_status.h");
+        const C_STATUS_HEADER: &str = include_str!(concat!(
+            env!("CARGO_MANIFEST_DIR"),
+            "/../../rocjitsu/lib/rocjitsu/include/rocjitsu/base/rj_status.h"
+        ));
         let c_statuses: Vec<String> = C_STATUS_HEADER
             .lines()
             .map(str::trim)
--- a/emulation/mirage/rocjitsu_sys/src/lib.rs
+++ b/emulation/mirage/rocjitsu_sys/src/lib.rs
@@ -40,6 +40,8 @@ pub const ROCJITSU_STATUS_OUT_OF_RESOURCES: RjStatus = 3;
 pub const ROCJITSU_STATUS_INVALID_CODE_OBJECT: RjStatus = 4;
 /// A required file could not be opened or read.
 pub const ROCJITSU_STATUS_INVALID_FILE: RjStatus = 5;
+/// The requested operation is not supported by this configuration.
+pub const ROCJITSU_STATUS_UNSUPPORTED: RjStatus = 6;
 
 /// Platform-specific handle type (`rj_handle_t`); an fd on Linux.
 pub type RjHandle = c_int;
@@ -616,6 +618,29 @@ mod tests {
         assert_eq!(RjDaemonStatus::Stopped as i32, 0);
         assert_eq!(RjDaemonStatus::Running as i32, 2);
         assert_eq!(RjDaemonStatus::Error as i32, 4);
+    }
+
+    #[test]
+    fn status_codes_match_c_api() {
+        const C_STATUS_HEADER: &str =
+            include_str!("../../../rocjitsu/lib/rocjitsu/include/rocjitsu/base/rj_status.h");
+        let c_statuses: Vec<String> = C_STATUS_HEADER
+            .lines()
+            .map(str::trim)
+            .filter(|line| line.starts_with("ROCJITSU_STATUS_"))
+            .map(|line| line.trim_end_matches(',').to_owned())
+            .collect();
+        let rust_statuses = vec![
+            format!("ROCJITSU_STATUS_SUCCESS = {ROCJITSU_STATUS_SUCCESS}"),
+            format!("ROCJITSU_STATUS_ERROR = {ROCJITSU_STATUS_ERROR}"),
+            format!("ROCJITSU_STATUS_INVALID_ARGUMENT = {ROCJITSU_STATUS_INVALID_ARGUMENT}"),
+            format!("ROCJITSU_STATUS_OUT_OF_RESOURCES = {ROCJITSU_STATUS_OUT_OF_RESOURCES}"),
+            format!("ROCJITSU_STATUS_INVALID_CODE_OBJECT = {ROCJITSU_STATUS_INVALID_CODE_OBJECT}"),
+            format!("ROCJITSU_STATUS_INVALID_FILE = {ROCJITSU_STATUS_INVALID_FILE}"),
+            format!("ROCJITSU_STATUS_UNSUPPORTED = {ROCJITSU_STATUS_UNSUPPORTED}"),
+        ];
+
+        assert_eq!(c_statuses, rust_statuses);
     }
 
     #[test]
--- a/emulation/rocjitsu/docs/configuration.md
+++ b/emulation/rocjitsu/docs/configuration.md
@@ -55,9 +55,18 @@ The example above is intentionally minimal and single-threaded.
 | Field | Type | Description |
 |---|---|---|
 | `max_ticks` | int | Maximum simulation ticks (0 = unlimited) |
-| `num_threads` | int | Number of PDES engine partitions/worker threads. |
+| `num_threads` | int | Simdojo engine partitions (one per XCD when partitioned) |
 | `exec_mode` | string | `"functional"` or `"cycle"` |
 | `vm.arch` | string | Architecture: `cdna3`, `cdna4`, etc. |
+
+### Simulation threading
+
+`num_threads` controls Simdojo engine partitions and their worker threads.
+The value is clamped to the number of XCDs visible to the VM. With
+`num_threads: 1`, all XCDs stay in one engine partition. With
+`num_threads: 4` on the 8-XCD CDNA4 configs, whole XCD subtrees are assigned
+round-robin to four partitions; with `num_threads: 8`, each XCD gets its own
+partition. A single XCD is never split across partitions.
 
 ### Topology
 
--- a/emulation/rocjitsu/docs/configuration.md
+++ b/emulation/rocjitsu/docs/configuration.md
@@ -68,6 +68,12 @@ The value is clamped to the number of XCDs visible to the VM. With
 round-robin to four partitions; with `num_threads: 8`, each XCD gets its own
 partition. A single XCD is never split across partitions.
 
+For multi-GPU VMs, clamping uses the aggregate XCD count across all SoCs.
+Partition assignment follows one global XCD ordering across the SoCs and is
+deliberately locality-agnostic. For example, two 8-XCD GPUs permit up to 16
+partitions, while `num_threads: 4` assigns XCDs from both GPUs to each
+partition.
+
 ### Topology
 
 Components are defined hierarchically under `topology.root`. Range
--- a/emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/vm/rj_vm.h
+++ b/emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/vm/rj_vm.h
@@ -206,11 +206,15 @@ RJ_API_EXPORT void rj_vm_destroy(rj_vm_t *vm);
 /// @brief Step the entire VM by one tick.
 ///
 /// @details Processes all simulation events at the next timestamp, advancing the
-/// simulation by one tick.
+/// simulation by one tick. Stepping is supported only for single-partition VMs;
+/// use rj_vm_run() when the configuration has multiple engine partitions.
 /// @param[in] vm VM handle.
-/// @param[out] active Non-zero if any wavefront is still executing.
+/// @param[out] active Non-zero if any wavefront is still executing. Written only
+/// on success.
 /// @retval ROCJITSU_STATUS_SUCCESS Step completed successfully.
 /// @retval ROCJITSU_STATUS_INVALID_ARGUMENT @p vm is NULL.
+/// @retval ROCJITSU_STATUS_ERROR The VM has no SoC.
+/// @retval ROCJITSU_STATUS_UNSUPPORTED The VM has multiple engine partitions.
 RJ_API_EXPORT rj_status_t rj_vm_step(rj_vm_t *vm, int *active);
 
 /// @brief Run the VM to completion or until max_ticks is reached.
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/config_loader.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/config_loader.h
@@ -48,12 +48,18 @@ struct TopologyBuildResult {
 ///
 /// Contains the engine configuration and the built component tree. Provides
 /// convenience accessors for the SoC and GPU memory. After loading, wire
-/// the topology into a SimulationEngine:
+/// the topology into a SimulationEngine. Multi-threaded rocjitsu topologies
+/// use the XCD-aware policy from rocjitsu/vm/amdgpu/partitioning.h:
 /// @code
 ///   auto loaded = load_config("config.json", kEmbeddedSchema);
-///   SimulationEngine engine(loaded.engine_config);
+///   auto *soc = loaded.soc();
+///   simdojo::SimulationEngine engine(loaded.engine_config);
 ///   engine.topology().set_root(loaded.take_root());
 ///   loaded.wire_links(engine.topology());
+///   if (loaded.engine_config.num_threads > 1 &&
+///       !amdgpu::partition_topology_by_xcds(
+///           engine.topology(), soc, loaded.engine_config.num_threads))
+///     throw std::invalid_argument("multi-threaded topology requires an XCD");
 ///   engine.create();
 /// @endcode
 struct LoadedConfig {
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/config_loader.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/config_loader.h
@@ -51,13 +51,18 @@ struct TopologyBuildResult {
 /// the topology into a SimulationEngine. Multi-threaded rocjitsu topologies
 /// use the XCD-aware policy from rocjitsu/vm/amdgpu/partitioning.h:
 /// @code
+///   #include "rocjitsu/vm/amdgpu/partitioning.h"
+///
 ///   auto loaded = load_config("config.json", kEmbeddedSchema);
 ///   auto *soc = loaded.soc();
+///   loaded.engine_config.num_threads =
+///       rocjitsu::amdgpu::clamp_xcd_partition_count(
+///           soc, loaded.engine_config.num_threads);
 ///   simdojo::SimulationEngine engine(loaded.engine_config);
 ///   engine.topology().set_root(loaded.take_root());
 ///   loaded.wire_links(engine.topology());
 ///   if (loaded.engine_config.num_threads > 1 &&
-///       !amdgpu::partition_topology_by_xcds(
+///       !rocjitsu::amdgpu::partition_topology_by_xcds(
 ///           engine.topology(), soc, loaded.engine_config.num_threads))
 ///     throw std::invalid_argument("multi-threaded topology requires an XCD");
 ///   engine.create();
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/partitioning.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/partitioning.cpp
@@ -4,6 +4,7 @@
 #include "rocjitsu/vm/amdgpu/partitioning.h"
 
 #include <array>
+#include <cassert>
 #include <unordered_map>
 
 namespace rocjitsu {
@@ -29,11 +30,15 @@ bool partition_topology_by_xcds(simdojo::Topology &topology, std::span<SoC *> so
     return false;
 
   topology.partition_manual(num_partitions, [&](simdojo::Component *component) {
-    for (auto *candidate = component; candidate != nullptr;
-         candidate = static_cast<simdojo::Component *>(candidate->parent())) {
+    for (auto *candidate = component; candidate != nullptr;) {
       auto it = xcd_partitions.find(candidate);
       if (it != xcd_partitions.end())
         return it->second;
+
+      auto *parent = candidate->parent();
+      candidate = dynamic_cast<simdojo::Component *>(parent);
+      assert((parent == nullptr || candidate != nullptr) &&
+             "component parents must also be components");
     }
     return simdojo::PartitionID{0};
   });
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/partitioning.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/partitioning.cpp
@@ -1,0 +1,49 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: MIT
+
+#include "rocjitsu/vm/amdgpu/partitioning.h"
+
+#include <array>
+#include <unordered_map>
+
+namespace rocjitsu {
+namespace amdgpu {
+
+bool partition_topology_by_xcds(simdojo::Topology &topology, std::span<SoC *> socs,
+                                uint32_t num_partitions) {
+  if (num_partitions == 0)
+    return false;
+
+  std::unordered_map<simdojo::Component *, simdojo::PartitionID> xcd_partitions;
+  uint32_t global_xcd_index = 0;
+  for (auto *soc : socs) {
+    if (!soc)
+      continue;
+    for (uint32_t xcd_index = 0; xcd_index < soc->num_xcds(); ++xcd_index) {
+      xcd_partitions[soc->xcd(xcd_index)] = global_xcd_index % num_partitions;
+      ++global_xcd_index;
+    }
+  }
+
+  if (xcd_partitions.empty())
+    return false;
+
+  topology.partition_manual(num_partitions, [&](simdojo::Component *component) {
+    for (auto *candidate = component; candidate != nullptr;
+         candidate = static_cast<simdojo::Component *>(candidate->parent())) {
+      auto it = xcd_partitions.find(candidate);
+      if (it != xcd_partitions.end())
+        return it->second;
+    }
+    return simdojo::PartitionID{0};
+  });
+  return true;
+}
+
+bool partition_topology_by_xcds(simdojo::Topology &topology, SoC *soc, uint32_t num_partitions) {
+  std::array<SoC *, 1> socs = {soc};
+  return partition_topology_by_xcds(topology, std::span<SoC *>(socs), num_partitions);
+}
+
+} // namespace amdgpu
+} // namespace rocjitsu
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/partitioning.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/partitioning.cpp
@@ -13,7 +13,7 @@ namespace amdgpu {
 
 uint32_t clamp_xcd_partition_count(std::span<SoC *> socs, uint32_t requested_partitions) {
   uint32_t num_xcds = 0;
-  for (auto *soc : socs) {
+  for (SoC *soc : socs) {
     if (soc)
       num_xcds += soc->num_xcds();
   }
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/partitioning.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/partitioning.cpp
@@ -3,12 +3,28 @@
 
 #include "rocjitsu/vm/amdgpu/partitioning.h"
 
+#include <algorithm>
 #include <array>
 #include <cassert>
 #include <unordered_map>
 
 namespace rocjitsu {
 namespace amdgpu {
+
+uint32_t clamp_xcd_partition_count(std::span<SoC *> socs, uint32_t requested_partitions) {
+  uint32_t num_xcds = 0;
+  for (auto *soc : socs) {
+    if (soc)
+      num_xcds += soc->num_xcds();
+  }
+
+  return std::clamp(requested_partitions, 1u, std::max(num_xcds, 1u));
+}
+
+uint32_t clamp_xcd_partition_count(SoC *soc, uint32_t requested_partitions) {
+  std::array<SoC *, 1> socs = {soc};
+  return clamp_xcd_partition_count(std::span<SoC *>(socs), requested_partitions);
+}
 
 bool partition_topology_by_xcds(simdojo::Topology &topology, std::span<SoC *> socs,
                                 uint32_t num_partitions) {
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/partitioning.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/partitioning.cpp
@@ -7,6 +7,8 @@
 #include <array>
 #include <cassert>
 #include <unordered_map>
+#include <unordered_set>
+#include <vector>
 
 namespace rocjitsu {
 namespace amdgpu {
@@ -31,13 +33,19 @@ bool partition_topology_by_xcds(simdojo::Topology &topology, std::span<SoC *> so
   if (num_partitions == 0)
     return false;
 
+  const std::vector<simdojo::Component *> components = topology.collect_all_components();
+  const std::unordered_set<simdojo::Component *> topology_components(components.begin(),
+                                                                     components.end());
   std::unordered_map<simdojo::Component *, simdojo::PartitionID> xcd_partitions;
   uint32_t global_xcd_index = 0;
-  for (auto *soc : socs) {
+  for (SoC *soc : socs) {
     if (!soc)
       continue;
     for (uint32_t xcd_index = 0; xcd_index < soc->num_xcds(); ++xcd_index) {
-      xcd_partitions[soc->xcd(xcd_index)] = global_xcd_index % num_partitions;
+      simdojo::Component *xcd = soc->xcd(xcd_index);
+      if (!topology_components.contains(xcd))
+        return false;
+      xcd_partitions[xcd] = global_xcd_index % num_partitions;
       ++global_xcd_index;
     }
   }
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/rj_vm.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/rj_vm.cpp
@@ -18,7 +18,6 @@ RJ_DIAGNOSTIC_IGNORE_PEDANTIC
 #include "linux/uapi/kfd_ioctl.h"
 RJ_DIAGNOSTIC_POP
 
-#include <algorithm>
 #include <cerrno>
 #include <cstring>
 #include <memory>
@@ -43,18 +42,19 @@ rj_status_t create_from_loaded(config::LoadedConfig &loaded, rj_vm_mode_t mode, 
   auto s = std::make_unique<rj_vm_t>();
   s->soc = loaded.soc();
   auto num_xcds = s->soc->num_xcds();
-  uint32_t num_xcds_available = num_xcds;
+  std::vector<SoC *> partition_socs;
+  partition_socs.reserve(loaded.extra_gpu_builds.size() + 1);
+  partition_socs.push_back(s->soc);
   if (loaded.num_gpus > 1) {
     for (auto &eb : loaded.extra_gpu_builds) {
       if (auto *extra_soc = dynamic_cast<SoC *>(eb.root.get()))
-        num_xcds_available += extra_soc->num_xcds();
+        partition_socs.push_back(extra_soc);
     }
   }
   // XCD partitions (config num_threads): run each XCD on its own engine
   // partition/thread so the XCDs execute concurrently across their separate L2s.
-  const uint32_t num_threads_available = std::max(num_xcds_available, 1u);
   const uint32_t num_threads_used =
-      std::clamp<uint32_t>(loaded.engine_config.num_threads, 1u, num_threads_available);
+      amdgpu::clamp_xcd_partition_count(partition_socs, loaded.engine_config.num_threads);
   loaded.engine_config.num_threads = num_threads_used;
 
   bool serve = (mode == RJ_VM_MODE_LOCAL || mode == RJ_VM_MODE_DAEMON);
@@ -66,17 +66,14 @@ rj_status_t create_from_loaded(config::LoadedConfig &loaded, rj_vm_mode_t mode, 
 
   s->engine_config = loaded.engine_config;
   s->engine = std::make_unique<simdojo::SimulationEngine>(loaded.engine_config);
-  std::vector<SoC *> partition_socs;
 
   if (loaded.num_gpus > 1 && !loaded.extra_gpu_builds.empty()) {
     std::vector<std::unique_ptr<SoC>> socs;
     std::vector<uint32_t> gpu_ids;
-    partition_socs.reserve(loaded.extra_gpu_builds.size() + 1);
 
     auto root0 = loaded.take_root();
     root0.release();
     socs.push_back(std::unique_ptr<SoC>(s->soc));
-    partition_socs.push_back(s->soc);
     gpu_ids.push_back(loaded.devices.empty() ? 0 : loaded.devices[0].gpu_id);
 
     for (size_t i = 0; i < loaded.extra_gpu_builds.size(); ++i) {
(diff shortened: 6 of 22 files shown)


atgutier (commented):
## BLOCKER                                                                                                                                                                                   11:38:52 [54/2000]

  [BLOCKER] lib/rocjitsu/src/rocjitsu/vm/soc.cpp:14-36 — soc_dispatch violates XCD partition ownership

  Reasoning: With num_threads > 1, the new partitioner assigns each XCD and its descendants to a separate engine thread. set_soc_dispatch() then registers CUs and SPIs from every foreign XCD directly with the
  primary XCD’s command processor. That primary CP invokes CU scheduling inline from its own partition thread.

  Debug builds abort when the command processor starts. In release builds, the assertion disappears and the primary thread mutates foreign-partition CU state and event queues, creating data races and
  incorrectly ordered events.

  Evidence:

  - rj_vm.cpp:47-51,106-125 preserves multiple engine partitions, enables soc_dispatch, and then partitions by XCD.
  - soc.cpp:21-33 explicitly registers every foreign CU and SPI with the primary CP.
  - command_processor.cpp:406-417 asserts that every CU dispatched by a CP belongs to the same partition.
  - compute_unit.h:675-692 documents that schedule_work() mutates thread-affine state and accesses the CU’s partition context.
  - command_processor.h:152-160 shows that add_compute_unit() also repoints the CU’s CP and idle callback.
  - Functional links execute destination handlers synchronously, so the link layer does not restore destination-thread affinity.
  - The added tests inspect partition IDs but never run a dispatch with both soc_dispatch=true and multiple workers.

  Suggested fix: For this layer, reject or force num_threads=1 whenever soc_dispatch is enabled. A real parallel implementation must retain XCD-local CP/SPI/CU ownership and exchange dispatch and completion
  messages through partition-aware queues. Add a multi-thread dispatch-to-completion regression under TSan.

  Confidence: high.

  [BLOCKER] PR #8705 merge state — the retargeted stack cannot merge into current develop

  Reasoning: The PR body still describes this as layer 3 based on the layer-2 branch, but the PR is now based directly on develop. That makes GitHub include stale dependency commits in the effective PR and
  produces unresolved conflicts with subsequent cache/coherence changes.

  Evidence:

  - Current head: 3442a8a3f7d21572fad6acdcf29e19421f481009
  - Current base: develop at 78e081e9bc6e233d7ded81282de79413bf5fc648
  - GitHub currently reports mergeable=CONFLICTING and mergeStateStatus=DIRTY.
  - The intended layer is only a5c7c5961046e99a97ac9d9978fb35824cfeb586..3442a8a3, covering 14 files and +353/-8; the retargeted PR reports 30 files.
  - The merge tree conflicts in the corpus workflow, l1_scalar_cache.h, l2_cache.{h,cpp}, memory_pipeline.cpp, rj_vm.cpp, l2_cache_test.cpp, and shared_infra_test.cpp.

  Suggested fix: Drop the stale dependency commits and replay only a5c7c596..3442a8a3 on current develop, resolving against the current cache/coherence implementation. The resulting head needs another review
  because conflict resolution touches execution and memory behavior.

  Confidence: high.

  ## MAJOR

  [MAJOR] lib/rocjitsu/src/rocjitsu/vm/rj_vm.cpp:225-234 — multi-partition VMs cannot use public rj_vm_step()

Reasoning: The PR deliberately permits and clamps configurations to multiple XCD partitions, but the public stepping API still calls the single-thread-only SimulationEngine::step(). Debug bu11:38:57 [7/2000]
  immediately. With assertions disabled, step() accesses only contexts_[0], so events on every other XCD partition stop advancing.

  Evidence:

  - rj_vm.cpp:47-51 now preserves num_threads > 1 up to the XCD count.
  - rj_vm.cpp:231 unconditionally calls vm->engine->step().
  - lib/simdojo/src/simulation.cpp:130-132 asserts num_threads == 1.
  - simulation.cpp:143-166 pushes to and processes only contexts_[0].
  - The public header and VM documentation expose rj_vm_step() without a multi-partition restriction.
  - No added test invokes the C API stepping path with multiple partitions.

  Suggested fix: Either implement a coordinated multi-partition epoch step, or make rj_vm_step() return a defined unsupported/configuration error when num_threads > 1 and document that only rj_vm_run()
  supports parallel VMs. Add a C API regression for both debug and release builds.

  Confidence: high.

  ## MINOR

  [MINOR] schemas/simulation_config.fbs:9-15 — unrelated FlatBuffers field was deleted instead of deprecated

  Reasoning: The change removes ComputeUnitConfig.functional_quantum while adding SimulationConfig.soc_dispatch. Although the field currently appears unused, deleting a FlatBuffers table field weakens schema
  evolution: its ordinal can be accidentally reused and existing serialized configurations or checkpoints can be misinterpreted by future schemas.

  Evidence: The layer diff removes functional_quantum:uint32 from the middle of ComputeUnitConfig; it is unrelated to XCD partitioning.

  Suggested fix: Restore the field, preferably as:

  functional_quantum:uint32 (deprecated);

  Keep soc_dispatch appended at the end of SimulationConfig.

  Confidence: high.

lialan:
@atgutier Rejected cross-partition soc_dispatch; made multi-partition `rj_vm_step()` return `ROCJITSU_STATUS_UNSUPPORTED`; retained functional_quantum; and preserved soc_dispatch across checkpoint save/restore.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/rj_vm.cpp
```diff
@@ -98,7 +121,14 @@ rj_status_t create_from_loaded(config::LoadedConfig &loaded, rj_vm_mode_t mode,
     s->engine->topology().set_root(std::move(vm_ptr));
     loaded.wire_links(s->engine->topology());
     s->soc->wire_backing(s->engine->topology());
+    partition_socs.push_back(s->soc);
+    if (loaded.soc_dispatch)
+      s->soc->set_soc_dispatch(true);
   }
+  if (xcd_partitions > 1)
+    amdgpu::partition_topology_by_xcds(
```
newling: Considering the utility of the generic algorithm as no one seems to use, perhaps it should be removed. 

rocjitsu explicitly installs its XCD-aware partitioning policy here, should SimulationEngine::create() stop invoking the generic FM partitioner implicitly? The tests where placement matters already choose explicit mappings. We could retain the generic algorithm temporarily as an opt-in partition_balanced() operation, require multi-partition callers to select a policy explicitly, and remove the algorithm later if no real consumer remains (or remove immediately).

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/rj_vm.cpp
```diff
@@ -39,6 +43,18 @@ rj_status_t create_from_loaded(config::LoadedConfig &loaded, rj_vm_mode_t mode,
...
+    for (auto &eb : loaded.extra_gpu_builds) {
+      if (auto *extra_soc = dynamic_cast<SoC *>(eb.root.get()))
+        partition_xcds += extra_soc->num_xcds();
+    }
+  }
+  // XCD partitions (config num_threads): run each XCD on its own engine
+  // partition/thread so the XCDs execute concurrently across their separate L2s.
+  uint32_t xcd_partitions =
+      std::clamp<uint32_t>(loaded.engine_config.num_threads, 1u, std::max(partition_xcds, 1u));
+  loaded.engine_config.num_threads = xcd_partitions;
```
newling: I'd prefer 2 explicit variable names, rather than clamping num_threads here. 

num_threads_used = max(n_xcds, num_threads_available)?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/checkpoint.cpp
```diff
@@ -196,10 +196,10 @@ LoadedConfig restore_checkpoint(const std::string &path) {
   engine_config.max_ticks = fb_config->max_ticks();
   engine_config.num_threads = fb_config->num_threads();
 
-  // Build a VirtualMachine from the old-format config, then extract SoC as root.
-  auto vm = std::make_unique<VirtualMachine>(vm_config);
-  auto *soc_ptr = vm->soc();
-  auto *mem_ptr = vm->memory();
+  // Rebuild the SoC root expected by LoadedConfig and create_from_loaded().
```
newling: This appears to fix a pre-existing mismatch between restore_checkpoint(), which previously returned a VirtualMachine root, and create_from_loaded(), which expects LoadedConfig::soc() to find a SoC root. That seems like an independent public checkpoint-restoration bug rather than part of XCD partitioning. Could the root-shape repair and C-API round-trip test move to a separate bug-fix PR?

emulation/rocjitsu/docs/configuration.md
```diff
@@ -55,10 +55,32 @@ The example above is intentionally minimal and single-threaded.
...
+
+`num_threads` controls Simdojo engine partitions, not the CU dispatch
+worker pool. The value is clamped to the number of XCDs visible to the
+VM. With `num_threads: 1`, all XCDs stay in one engine partition. With
+`num_threads: 4` on the 8-XCD CDNA4 configs, whole XCD subtrees are
+assigned round-robin to four partitions; with `num_threads: 8`, each
+XCD gets its own partition. A single XCD is never split across
+partitions.
+
+`soc_dispatch` controls which command processor (CP) accepts queues for
```
newling: I don't understand the utility of this flag. It requires num_threads = 1, which suggests it is not something we want enabled to hit peak parallelism (even with the later PRs that add within-XCD parallelism)? I know we discussed this on the original PR https://github.com/ROCm/rocm-systems/pull/6962 but my preference would be to postpone adding it until it's use is clearer. Can we keep the current focused on XCD partitioning please?

newling (commented):
My request is to split this into 3 PRs (can do after all other reviewers comments are addressed) as it seems like there are are least 3 different aspects which would be easier to reason about in isolation. Let me know if I've missed something and they're intertwined, or splitting is not easy.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/rj_vm.cpp
kuhar: We should return `ROCJITSU_STATUS_UNSUPPORTED` when `soc_dispatch` is combined with more than one XCD partition, and cover both public constructors. The incompatibility is currently detected by an exception from `engine->create()`; a local probe returned `INVALID_ARGUMENT` from `rj_vm_create_from_string()` and `INVALID_FILE` from `rj_vm_create()`. An explicit check after computing `xcd_partitions` would give the same configuration error one consistent status.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/partitioning.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include "rocjitsu/vm/amdgpu/partitioning.h"
+
+#include <array>
+#include <unordered_map>
+
+namespace rocjitsu {
+namespace amdgpu {
+
+bool partition_topology_by_xcds(simdojo::Topology &topology, std::span<SoC *> socs,
```
kuhar, line 31: I think we should replace the hand-rolled XCD partitioners in `vector_add_test.cpp`, `matmul_test.cpp`, and `scaling_test.cpp` with `partition_topology_by_xcds()`. The new unit tests only inspect placement, while those existing multi-threaded tests dispatch workloads and check golden results. I verified the vector-add swap is drop-in and the golden-reference test still passes; using the helper there would add behavioural coverage of the shipped implementation while removing duplicate mapping code.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/config_loader.h
```diff
@@ -66,6 +66,7 @@ struct LoadedConfig {
   DbtGuestConfig dbt_guest;             ///< Optional DBT guest-GPU discovery config.
   uint32_t num_gpus = 1;                ///< Number of simulated GPU instances.
   std::vector<KfdDeviceConfig> devices; ///< Per-GPU configs (populated when num_gpus > 1).
+  bool soc_dispatch = false; ///< Consolidate cross-XCD dispatch onto each SoC's primary CP.
```
kuhar: We should make `soc_dispatch` work through the documented C++ loading path, either by applying it while building the SoC or by adding an explicit application step to that API. The example above calls `load_config()`, installs the root, wires links, and creates the engine, but the parsed flag is only applied inside the private `create_from_loaded()` C-API path. As a result, a direct C++ consumer can parse `soc_dispatch: true` and silently create a normal SoC; the checkpoint test already has to call `set_soc_dispatch()` manually.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/rj_vm.cpp
```diff
@@ -39,6 +43,18 @@ rj_status_t create_from_loaded(config::LoadedConfig &loaded, rj_vm_mode_t mode,
...
+  if (loaded.num_gpus > 1) {
+    for (auto &eb : loaded.extra_gpu_builds) {
+      if (auto *extra_soc = dynamic_cast<SoC *>(eb.root.get()))
+        partition_xcds += extra_soc->num_xcds();
+    }
+  }
+  // XCD partitions (config num_threads): run each XCD on its own engine
+  // partition/thread so the XCDs execute concurrently across their separate L2s.
+  uint32_t xcd_partitions =
+      std::clamp<uint32_t>(loaded.engine_config.num_threads, 1u, std::max(partition_xcds, 1u));
```
kuhar: It would be good to add C-API coverage for the documented `num_threads` clamp, including the total XCD count across multiple GPUs. The current tests never request more partitions than visible XCDs. Local probes observed `64 -> 1` on the one-XCD `gfx1100_w7900.json`, `64 -> 8` on `gfx950_cdna4.json`, and successful creation/run with 16 partitions on the two-GPU config. Please pin the lower bound (`0 -> 1`), a single-GPU upper bound, and the aggregated two-GPU upper bound.

kuhar (commented):
A few comments

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/rj_vm.cpp
lialan: Added an explicit post-clamp check returning ROCJITSU_STATUS_UNSUPPORTED, with file and string constructor coverage.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/partitioning.cpp
lialan: Replaced the manual partitioners in vector_add, matmul, and scaling tests with partition_topology_by_xcds().

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/config_loader.h
lialan: Applied soc_dispatch during JSON load and checkpoint restore for every SoC, with direct C++ and checkpoint coverage.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/rj_vm.cpp
lialan: Added C-API coverage for 0 to 1, single-GPU 64 to 8, and two-GPU 64 to 16, including a successful multi-GPU run.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/partitioning.cpp
```diff
@@ -0,0 +1,49 @@
...
+    return false;
+
+  topology.partition_manual(num_partitions, [&](simdojo::Component *component) {
+    for (auto *candidate = component; candidate != nullptr;
+         candidate = static_cast<simdojo::Component *>(candidate->parent())) {
+      auto it = xcd_partitions.find(candidate);
+      if (it != xcd_partitions.end())
+        return it->second;
+    }
+    return simdojo::PartitionID{0};
```
newling, line 67: I found it a little confusing that every component outside an XCD subtree is assigned to partition 0. For example, GpuMemory is called directly and concurrently by all XCD partition workers, with its own internal synchronization, so being in partition 0 does not appear to mean that its work executes on worker 0. Not requesting a change here, but it would be useful to clarify what partition membership guarantees - event-queue ownership, lifecycle membership, thread affinity, etc. That might also reveal whether shared or structural components should be allowed to remain unpartitioned

emulation/rocjitsu/docs/configuration.md
```diff
@@ -55,10 +55,20 @@ The example above is intentionally minimal and single-threaded.
...
 |---|---|---|
 | `max_ticks` | int | Maximum simulation ticks (0 = unlimited) |
-| `num_threads` | int | Number of PDES engine partitions/worker threads. |
+| `num_threads` | int | Simdojo engine partitions (one per XCD when partitioned) |
 | `exec_mode` | string | `"functional"` or `"cycle"` |
 | `vm.arch` | string | Architecture: `cdna3`, `cdna4`, etc. |
 
+### Simulation threading
+
+`num_threads` controls Simdojo engine partitions, not the CU dispatch
```
newling: nit: I guess this comment is a bit premature given it's position in the stack (no CU worker pool yet)

emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/vm/rj_vm.h
```diff
@@ -206,11 +206,14 @@ RJ_API_EXPORT void rj_vm_destroy(rj_vm_t *vm);
...
-/// simulation by one tick.
+/// simulation by one tick. Stepping is supported only for single-partition VMs;
+/// use rj_vm_run() when the configuration has multiple engine partitions.
 /// @param[in] vm VM handle.
-/// @param[out] active Non-zero if any wavefront is still executing.
+/// @param[out] active Non-zero if any wavefront is still executing. Written only
+/// on success.
 /// @retval ROCJITSU_STATUS_SUCCESS Step completed successfully.
 /// @retval ROCJITSU_STATUS_INVALID_ARGUMENT @p vm is NULL.
+/// @retval ROCJITSU_STATUS_UNSUPPORTED The VM has multiple engine partitions.
```
newling, line 217:
```suggestion
/// @retval ROCJITSU_STATUS_UNSUPPORTED The VM has multiple engine partitions, or some other failure?.
```

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/partitioning.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include <array>
+#include <unordered_map>
+
+namespace rocjitsu {
+namespace amdgpu {
+
+bool partition_topology_by_xcds(simdojo::Topology &topology, std::span<SoC *> socs,
+                                uint32_t num_partitions) {
+  if (num_partitions <= 1)
```
newling: Can the function not proceed as normal for num_partitions == 1? All XCDs just get partition 0.

newling (commented):
Thanks for splitting it up, I enjoyed reviewing this one! Overall it's looking good I think. In additional to my minor comments inline, I got an agent review: https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr8705.md that suggests 2 actionable items (neither of which change the overall code design)

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/rj_vm.cpp
lialan: Renamed it to partition_balanced() and made multi-threaded engines require an explicit partition policy.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/rj_vm.cpp
lialan: Split this into num_threads_available and num_threads_used.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/checkpoint.cpp
lialan: Moved the checkpoint root-shape fix and C API regression to #9370.

emulation/rocjitsu/docs/configuration.md
lialan: Removed soc_dispatch; this PR now covers XCD partitioning only.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/partitioning.cpp
lialan: Clarified that partition membership owns scheduled events and link classification; it does not imply thread affinity for direct calls or lifecycle hooks.

emulation/rocjitsu/docs/configuration.md
lialan: Removed the premature CU worker-pool wording.

emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/vm/rj_vm.h
lialan: Documented ERROR for a missing SoC and UNSUPPORTED for multiple partitions.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/partitioning.cpp
lialan: Allowed one partition; all XCDs map to partition 0, with coverage.

newling (commented):
LGTM, but an agent highlights a failure mode in the balanced partitioned: https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr8705_followup.md#1-update-the-xcd-helper-documentation-for-its-one-partition-behavior 

It doesn't seem to me like this is a bug introduced by this PR (the default balanced partitioner predates this PR), but could you have a look anyway please?

For possible future consideration (not this PR):
- num_threads as public variable name, I think it should be made more descriptive of what it actually is
- pin down what the partitioning of components means, consider different partitionings for different uses, consider not requiring full partitions

atgutier (commented):
## MAJOR                                                12:26:21 [71/1651]

  [MAJOR] lib/simdojo/src/simulation.cpp:315-319 — asynchronous exit can
  deadlock multi-XCD execution

  Reasoning: rj_vm_request_exit() may set done_ while workers are leaving a
  completed barrier phase. One worker can observe done_ == true and return
  at line 319 without dropping from the barrier, while another worker
  observes the earlier value and enters the next phase. That worker
  eventually waits at a barrier whose expected count still includes the
  returned worker, hanging rj_vm_run() indefinitely.

  This latent Simdojo defect is materially exposed by this PR: lib/rocjitsu/
  src/rocjitsu/vm/rj_vm.cpp:124-128 now makes multi-partition execution a
  supported C-API path for multi-XCD VMs.

  Evidence:

  - The comment claiming all workers see done_ simultaneously is only valid
    when done_ is set by the barrier completion function. It is false for
    the public asynchronous rj_vm_request_exit() path.

  - Earlier exit checks correctly call arrive_and_drop() at
    simulation.cpp:287-290 and 302-305; the post-barrier exit does not.

  - tests/simdojo_sim_test.cpp:581-600 does not reliably cover the race. Its
    max_ticks=500 escape path can complete before the delayed exit request,
    and the test explicitly accepts normal completion.

  - PR #9441 contains the matching repair for this race.

  Suggested fix: require #9441 before merging, or apply the equivalent post-
  barrier logic:

  if (done_.load(std::memory_order_acquire)) {
    ctx.local_next.store(TICK_MAX, std::memory_order_relaxed);
    barrier_->arrive_and_drop();
    return;
  }

  Also remove the test’s max_ticks escape and require
  ExitReason::EXIT_REQUEST; ideally exercise the exit transition repeatedly
  or add synchronization that forces workers to straddle the phase boundary.

  Confidence: High.

## MINOR                                                12:26:24 [25/1651]

  [MINOR] docs/simdojo.md:272 — documented multi-threaded examples now throw
  during create()

  Reasoning: This PR changes SimulationEngine::create() to require an
  explicit topology partition policy whenever num_threads > 1. Both
  documented four-thread examples still call create() directly, and the
  lifecycle text incorrectly says create() performs partitioning.

  Evidence:

  - lib/simdojo/src/simulation.cpp:29-35 throws when a multi-threaded engine
    has no partitions.

  - docs/simdojo.md:272-289 constructs a four-thread engine without calling
    partition_balanced() or partition_manual().

  - docs/simdojo.md:515-525 repeats the invalid sequence and states that
    create() partitions the topology.

  - The loaded-configuration example in lib/rocjitsu/src/rocjitsu/config/
    config_loader.h:52-58 has the same problem for configurations requesting
    multiple threads.

  Suggested fix: install an explicit partition policy in the generic
  examples before create(), update the lifecycle description to say that
  create() validates an already-selected multi-thread policy, and show the
  XCD-aware policy in the loaded rocjitsu example.

  Confidence: High.

  ## NIT

  [NIT] lib/rocjitsu/src/rocjitsu/vm/amdgpu/partitioning.h:22 — helper
  documentation excludes a supported case

  Reasoning: The documentation says the helper installs a partition only
  when num_partitions > 1, but the implementation accepts one partition,
  assigns all XCDs to partition zero, and returns true.

Evidence:

  - partitioning.cpp:12-40 rejects only zero partitions.
  - tests/partitioning_test.cpp:120-130 explicitly tests successful one-
    partition behavior.

  Suggested fix: document the condition as “num_partitions is nonzero and at
  least one XCD is present,” and clarify that false means zero partitions or
  no XCDs.

  Confidence: High.

 MINOR — partitioning.h:20-27 doc contradicts code at num_partitions == 1

  The header says the manual partition installs only "if num_partitions is
  greater than one", but partitioning.cpp:14 early-returns only for 0. At 1
  it does install — and that differs observably from partition_balanced(1),
  which never calls classify_links(), so external link owners keep
  INVALID_PARTITION_ID and never get lifecycle hooks. Harmless today
  (guarded by if (num_threads_used > 1) at rj_vm.cpp:124), but the new test
  SinglePartitionAssignsAllComponentsToZero locks in behaviour the doc
  denies. Fix the doc; optionally have partition_balanced(1) delegate to
  partition_manual so the paths can't drift.

  MINOR — rj_vm.cpp:124 discards the return value

  Every other caller checks it (ASSERT_TRUE in the tests). If it returned
  false with num_threads_used > 1, partitions stay empty, create() throws,
  and rj_vm_create maps it to ROCJITSU_STATUS_INVALID_FILE — a wrong,
  undiagnosable status. Unreachable today, but the invariant spans 70 lines.
   Mark [[nodiscard]] and check it. (Also: std::span converts from
  std::vector directly; the explicit construction is unnecessary.)

  MINOR — add_queued_link() is new public API with no production caller

  Only caller tree-wide is simdojo_sim_test.cpp:308, and QueuedLink is
  constructed nowhere else. It exists so the new
  EngineRejectsCrossPartitionQueuedLink test can build one, and it
  duplicates add_link's port-validation asserts verbatim. It's also
  unmentioned in the PR description. Either factor the shared validation
  into a private helper and justify it, or drop the method and the test.

  MINOR — simulation.cpp:78-95 asserts promoted to throws, message discarded

  Cross-partition latency/QueuedLink checks moved from assert (no-op under
  NDEBUG) to throw std::invalid_argument. Release builds that previously ran
   now fail VM creation — reasonable — but rj_vm_create catches
  std::exception and discards what(), so users can't tell which link is bad.
   Include endpoint paths in the message and log e.what() before mapping to
  a status.


  NITs

  - partitioning.cpp:33 — static_cast<Component*>(candidate->parent())
  downcasts from Node*. Safe today (all parents are CompositeComponent) and
  copied verbatim from the deleted test helper, but this is production code
  now; prefer a parent_component() accessor or dynamic_cast + assert.
  - simdojo_sim_test.cpp:206 — Topology declared before ProducerComponent
  external, so external dies first and ~Topology() holds dangling pointers.
  Benign but fragile; reverse the declarations.

lialan:
@newling @atgutier addressed both comments.

kuhar (commented):
Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/config_loader.h:59`

We should keep the documented C++ loading recipe aligned with the C API and make the snippet self-contained. At `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/config_loader.h:52-62`, the recipe neither shows the required `rocjitsu/vm/amdgpu/partitioning.h` include nor clamps the requested thread count to the visible XCD count. A focused head probe with the 8-XCD `gfx950_cdna4.json` produced 64 partitions, 56 empty, while the C API clamps the same request to 8. Please centralize the XCD count/clamp in a helper used by both paths and show the required include in the example.

It would be good to assert the actual GPU1 XCD placement in `XcdPartitioningTest.CApiClampsMultiGpuThreadsToTotalXcdCountAndRuns`, not just the clamped count and successful run. A mutation removing `partition_socs.push_back(extra_soc)` left all three C-API partitioning tests green even though every GPU1 XCD fell onto partition 0 and eight workers became idle. Please walk `vm->vm->soc(gpu)->xcd(i)` here and compare each partition ID with the global XCD index; that directly covers the multi-GPU SoC-list assembly in `create_from_loaded()`.

It would be useful to add one single-partition success test for `rj_vm_step()` alongside the new multi-partition rejection test. Changing the guard to reject every nonzero thread count still left all 22 focused XCD/topology tests green, because the suite only exercises invalid and unsupported calls. Please create a one-partition VM and assert `ROCJITSU_STATUS_SUCCESS` plus `active == 0`; this also pins the new contract that `active` is written on success.

Original anchor: `emulation/rocjitsu/lib/simdojo/include/simdojo/sim/topology.h:190`

This would be safer if `partition_manual()` applied the caller's assigner to external link-endpoint owners, or explicitly documented why those owners must live on partition 0. At `emulation/rocjitsu/lib/simdojo/include/simdojo/sim/topology.h:184`, the contract says the callback assigns each component, but a focused repro with an assigner that always returns 1 leaves the external owner on partition 0; with a zero-latency link, that unrequested split makes `SimulationEngine::create()` throw. Please also add an engine-level external-producer test that runs through initialization, startup, and message delivery so the chosen external-owner contract is covered end to end.

I think we should mirror `ROCJITSU_STATUS_UNSUPPORTED = 6` in the manual Rust FFI constants. The public C enum at `emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/base/rj_status.h:32` now includes value 6, while `emulation/mirage/rocjitsu_sys/src/lib.rs:28-42` still stops at `ROCJITSU_STATUS_INVALID_FILE = 5`. Adding the constant keeps Rust consumers able to name and compare the new status; a parity check or generated binding could prevent the two lists from drifting again.

Can we update the two public Simdojo usage contracts affected here? At `emulation/rocjitsu/lib/simdojo/include/simdojo/sim/simulation.h:110`, the example still says `create()` partitions the topology although multi-threaded callers now select a policy explicitly; please show `partition_balanced()` or make the example explicitly single-threaded, and describe `create()` as validation/initialization. At `emulation/rocjitsu/lib/simdojo/include/simdojo/sim/topology.h:184`, `partition_manual()` is described as test-only even though the new production XCD helper calls it; please describe it as the explicit policy entry point for hardware or ownership constraints.

lialan:
@kuhar Addressed all six items: centralized XCD clamping, strengthened multi-GPU and `rj_vm_step()` coverage, applied manual assignment to external owners with an end-to-end test, added Rust status parity, and corrected the SimDojo docs. Release and sanitizer tests pass.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/partitioning.cpp
```diff
@@ -3,13 +3,29 @@
...
 #include <array>
 #include <cassert>
 #include <unordered_map>
 
 namespace rocjitsu {
 namespace amdgpu {
 
+uint32_t clamp_xcd_partition_count(std::span<SoC *> socs, uint32_t requested_partitions) {
+  uint32_t num_xcds = 0;
+  for (auto *soc : socs) {
```
kuhar: No auto here, per the style guide

lialan: Used the explicit `SoC *` type.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/rj_vm.cpp
```diff
@@ -39,6 +42,20 @@ rj_status_t create_from_loaded(config::LoadedConfig &loaded, rj_vm_mode_t mode,
...
+  partition_socs.push_back(s->soc);
+  if (loaded.num_gpus > 1) {
+    for (auto &eb : loaded.extra_gpu_builds) {
+      if (auto *extra_soc = dynamic_cast<SoC *>(eb.root.get()))
+        partition_socs.push_back(extra_soc);
+    }
+  }
+  // XCD partitions (config num_threads): run each XCD on its own engine
+  // partition/thread so the XCDs execute concurrently across their separate L2s.
+  const uint32_t num_threads_used =
```
kuhar, line 57: It would be useful to warn when clamping changes the requested count. create_from_loaded() overwrites loaded.engine_config.num_threads with the usable count, so num_threads: 64 succeeds as 8 partitions on the 8-XCD config with no diagnostic and callers cannot observe the requested value afterward. Please emit a concise warning with the requested and effective counts before the assignment.

emulation/rocjitsu/docs/configuration.md
```diff
@@ -55,10 +55,19 @@ The example above is intentionally minimal and single-threaded.
...
 | `vm.arch` | string | Architecture: `cdna3`, `cdna4`, etc. |
 
+### Simulation threading
+
+`num_threads` controls Simdojo engine partitions and their worker threads.
+The value is clamped to the number of XCDs visible to the VM. With
+`num_threads: 1`, all XCDs stay in one engine partition. With
+`num_threads: 4` on the 8-XCD CDNA4 configs, whole XCD subtrees are assigned
+round-robin to four partitions; with `num_threads: 8`, each XCD gets its own
+partition. A single XCD is never split across partitions.
```
kuhar, line 69: We should document the multi-GPU semantics here. The clamp uses the total XCD count across all SoCs, and the global round-robin can place XCDs from different GPUs in the same partition; for example, two 8-XCD GPUs allow 16 partitions, while num_threads: 4 makes each partition own XCDs from both GPUs. A short note that the aggregate ordering is deliberate and locality-agnostic would make the user-visible mapping unambiguous.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/partitioning.cpp
```diff
@@ -0,0 +1,70 @@
...
+        return it->second;
+
+      auto *parent = candidate->parent();
+      candidate = dynamic_cast<simdojo::Component *>(parent);
+      assert((parent == nullptr || candidate != nullptr) &&
+             "component parents must also be components");
+    }
+    return simdojo::PartitionID{0};
+  });
+  return true;
```
kuhar, line 69: This would be safer if the helper rejected SoCs that are not part of the supplied topology. With only soc0 rooted but both soc0 and soc1 passed, it returns true, creates 16 partitions with eight empty, and leaves every soc1 XCD at INVALID_PARTITION_ID. Please validate that each requested XCD was assigned, or explicitly document the topology-membership precondition and add a negative test.

kuhar (commented):
A few comments

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/rj_vm.cpp
lialan: Added a warning with requested and effective counts.

emulation/rocjitsu/docs/configuration.md
lialan: Documented aggregate, locality-agnostic multi-GPU mapping.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/partitioning.cpp
lialan: Rejects XCDs outside the topology before mutation, with coverage.

lialan:
@newling @kuhar The latest feedback is addressed. Could you please review again?

emulation/mirage/rocjitsu_sys/src/lib.rs
```diff
@@ -618,6 +620,29 @@ mod tests {
         assert_eq!(RjDaemonStatus::Error as i32, 4);
     }
 
+    #[test]
+    fn status_codes_match_c_api() {
+        const C_STATUS_HEADER: &str =
+            include_str!("../../../rocjitsu/lib/rocjitsu/include/rocjitsu/base/rj_status.h");
```
newling: High-level concern about a relative-path that starts ../../../ but I'll let @amd-arosa comment on this from the rust side if she wants

newling (approved):
The agent has raised a concern about an undocumeted lifetime rule here:
https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr8705_followup_2.md#1-state-or-enforce-the-lifetime-of-external-link-endpoint-owners, if could you please address that in the documentation

LGTM, but please allow a moment for other reviewers to have another look

atgutier (commented):
This PR was fine to merge, but I have some post-PR comments that may be worth fixing:

  ## MAJOR

  [MAJOR] lib/rocjitsu/src/rocjitsu/vm/amdgpu/completion_tracker.cpp:199 — Concurrent XCD completions
  can lose a shared HSA-signal decrement

  Reasoning: The XCD partitioning installed at lib/rocjitsu/src/rocjitsu/vm/rj_vm.cpp:123–127 allows
  command processors on different XCDs to retire dispatches concurrently.
  CompletionTracker::fire_signal() decrements a mapped completion signal using an atomic load followed
  by a separate atomic store. Its fallback path uses an ordinary read followed by an ordinary write.

  If two dispatches on different XCDs share a completion signal initialized to two, both workers can
  read two and store one. One decrement is lost, so a host wait or dependent barrier packet waiting
  for zero can hang indefinitely.

  Evidence:

  - completion_tracker.cpp:209–214 performs separate atomic_ref::load() and atomic_ref::store()
    operations.

  - completion_tracker.cpp:216–221 performs the same compound operation using non-atomic
    GpuMemory::read64() and write64().

  - One CompletionTracker exists per command processor, so the per-CP queue locks do not serialize
    different XCDs.

  - gpu_memory.h:160–196 already provides GpuMemory::atomic_rmw() across mapped, client-process, and
    sparse storage.

  Suggested fix: Perform the completion-signal decrement as one GpuMemory::atomic_rmw() operation and
  return the old/new value from that transaction. Do not retain a separate non-atomic fallback. Add a
  regression with two queues assigned to different XCDs, two same-tick dispatch completions, and one
  signal initialized to two; require a final value of zero and successful waiter/dependent-packet
  progress.

  Confidence: high.

[MAJOR] lib/rocjitsu/src/rocjitsu/vm/rj_vm.cpp:123 — Parallel XCD dispatches collide in shared race-
  detector state and can cause use-after-free

  Reasoning: Each XCD owns a separate CommandProcessor, and each processor begins its private dispatch
  counter at one. The VM nevertheless installs the same plugin-group instance on every XCD. The race
  plugin identifies a workgroup only by {dispatch_id, wg_id}.

  Consequently, the first dispatch and first workgroup from two XCDs both use {1,0}. The second
  onAmdgpuWorkgroupDispatched() replaces the first detector in detectors_, destroying it while
  wavefront plugin states from the first XCD retain raw WaveRaceState* pointers into it. Later
  instruction hooks dereference those dangling pointers. Dispatch names, disassembly caches, and
  observed-race records also alias unrelated kernels.

  Evidence:

  - command_processor.h:321 initializes every CP’s next_dispatch_id_ to one.
  - command_processor.cpp:1297 assigns dp.dispatch_id = next_dispatch_id_++.
  - soc.cpp:81–84 and amdgpu/xcd.h:63–67 share one plugin group across every XCD.
  - plugins/race_detector/plugin.cpp:162–167 constructs the key solely from dispatch_id and wg_id.
  - plugin.cpp:230–245 protects the map operation with a mutex but unconditionally replaces the keyed
    unique_ptr, then stores raw pointers into the detector in each wavefront.

  - The mutex prevents simultaneous map corruption; it does not prevent the logical key collision or
    resulting destruction.

  Suggested fix: Give dispatches a globally unique identity across GPU, XCD, and command-processor
  instances, or carry a stable GPU/XCD/CP identifier through every plugin callback and include it in
  all plugin keys. Add an ASan/TSan regression using the race plugin with two queues on different
  XCDs, each running its first dispatch/workgroup zero concurrently, and verify independent detector
  and report state.

  Confidence: high. The CP-local identifier collision was latent, but this PR materially exposes it by
  making those XCD callbacks genuinely concurrent.

@lialan

Issue #6628, opened 2026-05-31
[rocjitsu] Steps towards (tightly coupled) standalone plugins


newling:
# Plugin dlopen loading

This is a proposal to load plugins dynamically at runtime with dlopen, and thereby have plugins that are build entirely outside of rocjitsu. It is definitely open to change. 
## Context

rocjitsu has an execution plugin system that hooks into the GPU/RISC-V emulation loop. Plugins receive callbacks at key points (instruction execution, memory operations, barriers, dispatch lifecycle) and can implement instrumentation like race detection, logging, cycle counting.

Currently, plugins are compiled directly into `librocjitsu.so` as object libraries. The idea of this task is to make each plugin a standalone shared library (`.so`) loaded at runtime via `dlopen`, so that plugins can be developed and built independently, and plugin code is not linked into the rocjitsu binary when not in use.

There might initial loading infrastructure in the `util` header library that may be relevant (@atgutier mentioned this).

## Design choice: tight coupling

This proposal uses **tightly coupled C++ plugins**. Plugins subclass `ExecutionPlugin` (a C++ abstract class with virtual hooks) and use rocjitsu types directly (`amdgpu::Wavefront`, `Instruction`, etc.).

With this design an external plugin author would need to:

1. Clone the rocjitsu source tree (or install its headers)
2. `#include "rocjitsu/plugins/execution_plugin.h"` — this transitively
   includes wavefront, instruction, ISA traits, and other rocjitsu types
3. Subclass `ExecutionPlugin`, override the hooks they need
4. Export a C factory function: `extern "C" ExecutionPlugin *createMyPlugin()`
5. Build their `.so` with rocjitsu's include paths
6. Pass the `.so` path to rocjitsu at runtime

This approach gives plugins full access to rocjitsu's internal types, but provides **no guarantees of support**. If the `ExecutionPlugin` interface  or any type it exposes changes, all external plugins must be recompiled. This is acceptable when plugins ship from the same source tree (our current use case).

If we later need to support external plugins with backward compatibility or versioning, we can introduce a stable C API layer (`rj_vm_plugin_*`) that translates between opaque handles and internal types. This is a separate, larger effort and is not proposed here, although  also a good design, assuming someone is willing to provide versioning and support.

### Directory layout

The main development branch (which I hope to start landing to develop in pieces) has all plugin work inside the plugins directory. Concrete plugins each have their own sub-directory, with the interface in the root directory.
As follows:

```
[...]

PR #8706, opened 2026-07-16
refactor(rocjitsu): simplify execution plugins and loader

lialan:
## Motivation

Define a clear threading contract for rocJitsu execution plugins so parallel CPU simulation can run high-frequency callbacks concurrently without weakening correctness for plugins with shared mutable state. Simplify the loader boundary after removing numeric plugin ABI versioning and profiled execution. This PR is the plugin-policy foundation for #6962.

Execution plugins are repository-owned matching-build components. The host and bundled plugins must be rebuilt and shipped together; independently built or stale plugins are unsupported.

## What changed

- Serialize infrequent lifecycle, dispatch, workgroup, wavefront, and barrier hooks with one re-entrant group lock.
- Run high-frequency instruction, memory-routing, and register hooks concurrently by default.
- Let plugins with shared mutable hot-hook state opt into group serialization through `requires_serial_hot_hooks()`.
- Bypass callback locking for the shared empty group and require groups to be finalized before publication.
- Keep `ExecutionPluginGroup` final so contained plugins fully determine callback policy.
- Remove plugin hook profiling while warning and ignoring the removed top-level `profiled` configuration key.
- Keep the three loader exports—`rocjitsu_plugin_metadata`, `rocjitsu_plugin_create`, and `rocjitsu_plugin_destroy`—without compatibility or version negotiation.
- Reduce `PluginMetadata` to the operational name and configuration-schema fields, rename `plugin_abi.h` to `plugin_exports.h`, and remove stale contact/version metadata, version logging, profiling CI exclusion, and unused sink helpers.
- Preserve plugin-owned destruction and loaded-library lifetime rules.

## Compatibility

The three loader export names, plugin names, configuration schemas, sink behavior, and callback behavior remain unchanged. The metadata layout and header name intentionally remain an in-tree implementation boundary rather than a stable external ABI.

## Issue Tracking

GitHub issue: #6447

## Validation

Run on final #8706 head `9ae6f38fa35a8077d6d5131a6590c7ade479f006`:

- Release builds passed for `rocjitsu_tests`, `rocjitsu_bin`, `rocjitsu_shared`, `rocjitsu_plugin_logging_so`, and `rocjitsu_plugin_race_so`.
- `ExecutionPluginTest.*`, `ExecutionPluginDeathTest.*`, and `PluginLoaderTest.*`: 47 passed; 1 architecture-only MFMA test skipped.
- `LoggingTest.dispatch_logged`, `RaceTest.gfx950_vgpr_waitcnt`, and `RaceTest.gfx950_vgpr_waitcnt_race`: 3 passed.
- Both bundled plugin shared objects export exactly the three required loader symbols.
- Pre-commit passed across the complete PR diff, and `git diff --check` passed.

## Submission Checklist


newling (commented):
https://github.com/ROCm/rocm-systems/pull/8706/changes is showing changes that already merged, not sure if you need to rebase or something

atgutier (commented):
## MAJOR

  [MAJOR] lib/rocjitsu/src/rocjitsu/vm/plugins/execution_plugin.h:56–58 — New virtual method is added
  without bumping the runtime-plugin contract version

  Reasoning: The change inserts requires_serial_execution() into ExecutionPlugin’s virtual-method
  sequence before its lifecycle and execution callbacks. A plugin shared object compiled against the
  previous interface consequently has a different vtable layout. The loader nevertheless accepts it
  because both the previous and current interfaces report plugin ABI version 2.

  On the supported Linux C++ ABI, the new policy query occupies the stale plugin’s former onInit slot.
  Calling it invokes a void function through a bool signature. Subsequent lifecycle and execution
  callbacks are also shifted to incorrect vtable entries, leading to wrong callback ordering,
  incompatible calls, argument corruption, or a crash.

  Evidence:

  - The new virtual is added at execution_plugin.h:56–58.
  - plugin_abi.h:35–43 says plugins must use the same ExecutionPlugin layout and identifies the
    numeric contract as the stale-build guard.

  - plugin_abi.h:57–62 explicitly requires a bump for incompatible ExecutionPlugin interface changes
    but still defines kPluginAbiVersion = 2.

  - plugin_loader.cpp:100–105 accepts any shared object reporting version 2 and then invokes it
    through ExecutionPlugin.

  - The current bad-ABI fixture is compiled against the new class layout and merely reports
    current_version + 1; it cannot expose an old-layout plugin that still reports version 2.

  Suggested fix: Bump kPluginAbiVersion to 3 and rebuild the bundled plugins. Add a fixture compiled
  against a preserved v2 interface that reports version 2 and verify that the v3 host rejects it
  before calling its factory. Retain a normal v3 fixture to verify loading and lifecycle dispatch.

  Confidence: high.

  ## MINOR

  [MINOR] lib/rocjitsu/src/rocjitsu/vm/plugins/execution_plugin_group.h:89–109 — Group subclasses can
  still hide callbacks from the new capability policy

  Reasoning: has_hooks() and requires_serial_execution() are nonvirtual and inspect only the private
  plugins_ vector, but ExecutionPluginGroup remains subclassable and all AMDGPU callback methods
  remain virtual. An empty subclass can override a callback while still reporting has_hooks() ==
  false. A subclass with unsynchronized mutable callback state also reports
  requires_serial_execution() == false.

 The next execution-policy consumer can therefore skip a real callback or run it concurrently,
  recreating exactly the hidden group behavior this refactor is intended to eliminate.

  Evidence:

  - execution_plugin_group.h:89–96 derives both capabilities solely from contained plugins.
  - Lines 109–187 retain virtual AMDGPU dispatch methods.
  - The class has a virtual destructor and is not final.
  - The converted profiler no longer needs inheritance; it correctly represents its behavior as a
    contained ProfiledExecutionPlugin.

  - No remaining in-tree subclass was found, so preserving subclass overrides currently provides no
    demonstrated benefit.

  Suggested fix: Make ExecutionPluginGroup final and its forwarding callbacks nonvirtual, requiring
  all behavior and decorators to be represented as contained ExecutionPlugin objects. If inheritance
  must remain supported, the capability methods must be virtual and every subclass must accurately
  report its hidden behavior.

  A focused regression can define an empty group subclass overriding
  onAmdgpuBeforeExecuteInstruction() and demonstrate that it currently reports no hooks and no serial
  requirement.

  Confidence: high; immediate impact is architectural because no current in-tree subclass remains, but
  this layer is explicitly the policy foundation for the following execution layer.

newling:
I haven't looked at the code changes, but can you give some examples of when plugins would fail to just work with this multithreading? I'm trying to decide if there should be (a) default to allow (b) no default (c) default to don't allow

newling (changes requested):

lialan:
@newling Rebased and the merged #8705 changes are gone.

lialan:
@atgutier Addressed both: ABI v3 rejects a preserved-v2 fixture before creation, and `ExecutionPluginGroup` is final with nonvirtual forwarding.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/plugin_loader.cpp
```diff
@@ -233,12 +233,16 @@ PluginLoader::configure_plugin_group(const std::string &config_json, const std::
...
+  configure_sinks(root, *plugins);
+  load_from_config(config_json, *plugins, plugin_dir);
 
+  if (!profiled)
+    return std::shared_ptr<ExecutionPluginGroup>(std::move(plugins));
+
+  auto group = std::make_shared<ExecutionPluginGroup>();
   configure_sinks(root, *group);
-  load_from_config(config_json, *group, plugin_dir);
+  group->add(std::make_unique<ProfiledExecutionPlugin>(std::move(plugins)));
```
kuhar: We should add one loader test for `{"profiled":true,"plugins":{"good":{}}}` that calls `onInit()` and checks the `good:init` trace. The existing profiled tests all wrap an empty group, so they do not exercise the new decorator forwarding a real loaded plugin or preserving its lifetime and sink setup.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/profiled_execution_plugin.h
```diff
@@ -40,109 +39,108 @@ class ProfiledExecutionPluginGroup : public ExecutionPluginGroup {
...
-    ExecutionPluginGroup::onInit();
+  explicit ProfiledExecutionPlugin(std::unique_ptr<ExecutionPluginGroup> plugins)
+      : ExecutionPlugin("profile"), plugins_(std::move(plugins)), start_time_(Clock::now()) {
+    if (!plugins_)
+      plugins_ = std::make_unique<ExecutionPluginGroup>();
   }
 
+  ExecutionPluginGroup &plugins() { return *plugins_; }
+
+  void onInit() override { plugins_->onInit(); }
```
kuhar: This would be safer if `ProfiledExecutionPlugin` explicitly overrode `requires_serial_execution()` to return `true` and documented the unsynchronized counters and dispatch map. The class currently relies on the base default, so changing that default would silently make the profiler concurrently eligible. Please pin this class-specific policy in a test; if the profiler is ever made parallel-safe, the override should then propagate the wrapped group requirement. Also, can we drop the unused mutable `plugins()` accessor? It lets future callers alter the wrapped set after construction without a current use case.

docs/superpowers/specs/2026-08-01-pr8706-review-fixes-design.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Goal
+
+Address every actionable review comment on PR 8706 while keeping the pull
+request limited to execution-plugin policy. The final branch must exclude the
+already-merged PR 8705 history, reject stale v2 runtime plugins before calling
+their factory, prevent group subclasses from hiding callbacks, and explain the
+conservative threading default with concrete examples.
+
+## Branch shape
```
kuhar: We should remove this file from the product documentation, or trim it to durable technical rationale. `Branch shape`, `Validation`, and `GitHub follow-up` record one-time review, rebase, and push procedure, while the lasting plugin-contract and threading rationale already lives in `emulation/rocjitsu/docs/plugins.md`.

kuhar (commented):
`emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/execution_plugin.h:51`: The `slot_index()` contract needs to account for nested groups. Each group assigns indices from zero, so the profiled path gives the decorator and its first inner plugin slot 0. Nothing fails today because the profiler does not use wavefront state, but another stateful decorator would alias the inner plugin entry in `Wavefront::plugin_states_`. We should either allocate indices across the whole nested plugin tree or document and enforce that decorator plugins cannot use wavefront state.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/plugin_loader.cpp
lialan: Added a lifecycle test that loads good, calls onInit(), and verifies the create and init traces.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/profiled_execution_plugin.h
lialan: Done: the profiler now explicitly requires serial execution, documents its unsynchronized state, tests the policy, and drops the mutable accessor.

docs/superpowers/specs/2026-08-01-pr8706-review-fixes-design.md
lialan: Removed the one-time review design document.

emulation/rocjitsu/docs/plugins.md
```diff
@@ -220,6 +220,30 @@ records their register dependencies when they are issued. A later completion
 updates storage without emitting the same instruction-level write again.
 Synchronization retires the corresponding outstanding operations.
 
+### Dispatch threading
+
+Hook presence and callback policy are derived from the plugins contained by an
+`ExecutionPluginGroup`. An empty group has no hooks. A group requires serial
+command-processor callbacks when any contained plugin requires them.
+
+Plugins conservatively require serial callbacks by default because concurrent
```
newling: I think the plugin hooks can divided into 2 groups, the ones that don't benefit from parallelization and which we should lock at the group level, such as 
wavefront dispatch, workgroup complete, etc.

and those that might benefit:
instruction executed, register read, etc. 

For the first partition (the ones that fire infrequently) we should have a lock at the group level. i.e. we have something like:

    // RAII lock mutex here
    for (auto &p : plugins_)
      p->onAmdgpuWavefrontDispatched(wf);

These become automatically thread safe, irrespective of what the plugin does in its virtual function. 

The ones that might benefit I think we should assume are thread safe, because they are all at the workgroup / wavefront level, which is smaller than the granularity of the multithreading you're aiming for (SE > workgroup > wavefront). 

Does that sound reasonable?

newling (changes requested):

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/execution_plugin_group.h
```diff
@@ -85,95 +90,133 @@ class ExecutionPluginGroup {
   uint32_t num_plugins() const { return static_cast<uint32_t>(plugins_.size()); }
   bool empty() const { return plugins_.empty(); }
 
-  // -- Lifecycle --
-  virtual void onInit() {
-    for (auto &p : plugins_)
-      p->onInit();
+  /// Compute units only need to dispatch hooks when the group owns plugins.
+  bool has_hooks() const { return !empty(); }
```
newling: has_hooks isn't used?

emulation/rocjitsu/tests/plugin_loader_v2_fixture.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include <cstdint>
+#include <cstdio>
+#include <cstdlib>
+#include <span>
+#include <string>
+#include <utility>
+
+namespace rocjitsu {
+
+// Preserved from plugin ABI v2, before requires_serial_execution() was added.
```
newling: hmm, why do we have an ABI for plugins, I thought we decided in https://github.com/ROCm/rocm-systems/issues/6628 to not have an ABI? I don't think we should be versioning, I'm not planning on versioning when I add a new plugin hooks. My concern is this is going to turn into unused bloat @amd-arosa @atgutier

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/profiled_execution_plugin.h
```diff
@@ -40,112 +39,128 @@ class ProfiledExecutionPluginGroup : public ExecutionPluginGroup {
...
-
-  void onInit() override {
-    if (auto *s = build_composite_sink("profile.log"))
-      sink_ = s;
-    ExecutionPluginGroup::onInit();
-  }
+  /// Add a profiling decorator as the sole plugin in @p outer. The outer and
+  /// nested groups share a slot allocator, keeping wavefront-state indices
+  /// unique across arbitrarily nested decorator trees.
+  static ProfiledExecutionPlugin *add_to(ExecutionPluginGroup &outer,
```
newling: This API is hard to understand. I think it is ok to completely drop ProfiledExecutionPlugin in this PR, I can recreate it later from scratch.

newling (changes requested):
Thanks for trying out my design proposal! I am onboard with the overall direction now. 

Wondering what the best names for `dispatch_serialized` and `dispatch_hot` might be. How about `dispatch_with_plugin_lock`  and `dispatch_without_plugin_lock` ? 

One concern I have with this design is that if 1 plugin wants to be serialized, then all the plugins must be serialized, even though the different plugins cannot access each others state. I think this is ok for now though, let's reconsider this later. 

We should document the plugin (i.e. single plugin, not group) methods, documenting which ones need to be thread safe if the safety flag is not set.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/execution_plugin_group.h
lialan: Removed `has_hooks()` and its tests.

lialan:
Addressed the review: renamed the helpers to `dispatch_with_plugin_lock` / `dispatch_with_optional_plugin_lock`, and documented the concurrency contract on every plugin hook.

newling (commented):
This LGTM. I'm going to post in teams if removing plugin ABI is ok before approving...

newling (approved):
LGTM

kuhar (commented):
Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/execution_plugin.h:61`

Can we name this after the policy it now controls, for example `requires_serial_hot_hooks()` (and match the group accessor)? The simulation still runs in parallel; this return value only decides whether high-frequency callbacks take the group mutex, while infrequent callbacks are always serialized. It would also help to state here that the group samples the value once in `add()`, so an implementation must return a stable value from construction onward.

Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/execution_plugin_group.h:237`

We shouldn't use the same non-recursive mutex for serialized hooks that can re-enter high-frequency callbacks. `Wavefront::halt()` enters `onAmdgpuWavefrontHalted()` with `callback_mutex_` held; `HaltSnapshotPlugin` then reads registers through `RegisterAccess`, which calls back into `onAmdgpuReadSgpr()` / `onAmdgpuReadVgprLanes()` and tries to acquire that mutex again when any plugin opts into serialization.

The focused repro hangs until killed with a serial plugin present and exits normally without it. Could we separate the locking domains, or provide and document a non-notifying register-access path for serialized hooks? A regression combining a serial plugin with a register-reading halt hook would cover the current gap.

Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/execution_plugin_group.h:217`

We should skip dispatch before locking when `plugins_` is empty. Every CU, command processor, completion tracker, and hart shares `empty_group()`, so the plugin-free default currently funnels infrequent hooks from all partition threads through one process-wide mutex.

The focused benchmark measured the empty serialized hook at 10.1 ns/call on one thread and 39.7 ns/call on 4-8 threads, while the identical unlocked empty loop scaled down to 0.2 ns/call at 8 threads. An early `plugins_.empty()` return is behavior-preserving, matches the documented empty-group behavior, and should have a concurrency regression test.

Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/plugin_loader.cpp:99`

We shouldn't remove the last stale-plugin guard without another way to reject an old metadata layout. A plugin built against the previous header returns a non-null `PluginMetadata` whose first field is `int abi`; this host interprets that storage as `name` and can pass a garbage pointer into the comparison and warning at lines 104-106 instead of rejecting it safely.

The same-build requirement is only documented, while plugins are still discovered from normal library search paths. Could we add an automatically generated same-build identity (or another guard checked before dereferencing layout-dependent fields) and retain a fixture with the previous metadata layout?

Original anchor: `emulation/rocjitsu/tests/execution_plugin_test.cpp:359`

This would be less flaky if the overlap and non-overlap cases used separate timing bounds. In the repeated TSan run, `HighFrequencyHooksRunConcurrentlyByDefault` once consumed its full 100 ms wait and still passed, so a modest scheduling delay can make the positive test report serialization. The two serialized cases, meanwhile, intentionally pay the timeout in full. Could the positive overlap case use a substantially more generous deadline while the negative cases keep a short bound?

Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/execution_plugin_group.h:77`

It would be good to state the configure-before-publication contract on this class. `add()`, `add_sink()`, and `set_sink_dir()` mutate vectors and the cached `serialize_hot_hooks_` flag without synchronization, while dispatch reads them concurrently. Current callers populate a fresh group before `set_plugin_group()`, so they are safe, but a future caller could reasonably mutate a shared group and introduce a data race. Please document that plugins and sinks must be finalized before the group is handed to the simulation and remain immutable afterward.

We should clean up the remaining documentation after removing profiling and `has_hooks()`:

- `emulation/rocjitsu/docs/plugins.md:113` still lists profiling data as plugin output.
- `emulation/rocjitsu/docs/plugins.md:214` says hook presence is derived, although the group no longer exposes or computes hook presence; only callback policy is derived. The empty-group guarantee also needs the dispatch early-out described separately before it is true operationally.
- `emulation/mirage/rocjitsu_sys/src/lib.rs:381` still advertises a `profiled` config section that the loader now ignores.

Updating these together would keep the public configuration and threading descriptions aligned with this change.

newling:
> Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/execution_plugin.h:61`
> 
> Can we name this after the policy it now controls, for example `requires_serial_hot_hooks()` (and match the group accessor)? The simulation still runs in parallel; this return value only decides whether high-frequency callbacks take the group mutex, while infrequent callbacks are always serialized. It would also help to state here that the group samples the value once in `add()`, so an implementation must return a stable value from construction onward.
> 
> Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/execution_plugin_group.h:237`
> 
> We shouldn't use the same non-recursive mutex for serialized hooks that can re-enter high-frequency callbacks. `Wavefront::halt()` enters `onAmdgpuWavefrontHalted()` with `callback_mutex_` held; `HaltSnapshotPlugin` then reads registers through `RegisterAccess`, which calls back into `onAmdgpuReadSgpr()` / `onAmdgpuReadVgprLanes()` and tries to acquire that mutex again when any plugin opts into serialization.
> 
> The focused repro hangs until killed with a serial plugin present and exits normally without it. Could we separate the locking domains, or provide and document a non-notifying register-access path for serialized hooks? A regression combining a serial plugin with a register-reading halt hook would cover the current gap.
> 
> Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/execution_plugin_group.h:217`
> 
> We should skip dispatch before locking when `plugins_` is empty. Every CU, command processor, completion tracker, and hart shares `empty_group()`, so the plugin-free default currently funnels infrequent hooks from all partition threads through one process-wide mutex.
> 
> The focused benchmark measured the empty serialized hook at 10.1 ns/call on one thread and 39.7 ns/call on 4-8 threads, while the identical unlocked empty loop scaled down to 0.2 ns/call at 8 threads. An early `plugins_.empty()` return is behavior-preserving, matches the documented empty-group behavior, and should have a concurrency regression test.
> 
> Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/plugin_loader.cpp:99`
> 
> We shouldn't remove the last stale-plugin guard without another way to reject an old metadata layout. A plugin built against the previous header returns a non-null `PluginMetadata` whose first field is `int abi`; this host interprets that storage as `name` and can pass a garbage pointer into the comparison and warning at lines 104-106 instead of rejecting it safely.
> 
> The same-build requirement is only documented, while plugins are still discovered from normal library search paths. Could we add an automatically generated same-build identity (or another guard checked before dereferencing layout-dependent fields) and retain a fixture with the previous metadata layout?
> 
> Original anchor: `emulation/rocjitsu/tests/execution_plugin_test.cpp:359`
> 
> This would be less flaky if the overlap and non-overlap cases used separate timing bounds. In the repeated TSan run, `HighFrequencyHooksRunConcurrentlyByDefault` once consumed its full 100 ms wait and still passed, so a modest scheduling delay can make the positive test report serialization. The two serialized cases, meanwhile, intentionally pay the timeout in full. Could the positive overlap case use a substantially more generous deadline while the negative cases keep a short bound?
> 
> Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/execution_plugin_group.h:77`
> 
> It would be good to state the configure-before-publication contract on this class. `add()`, `add_sink()`, and `set_sink_dir()` mutate vectors and the cached `serialize_hot_hooks_` flag without synchronization, while dispatch reads them concurrently. Current callers populate a fresh group before `set_plugin_group()`, so they are safe, but a future caller could reasonably mutate a shared group and introduce a data race. Please document that plugins and sinks must be finalized before the group is handed to the simulation and remain immutable afterward.
> 
> We should clean up the remaining documentation after removing profiling and `has_hooks()`:
> 
> * `emulation/rocjitsu/docs/plugins.md:113` still lists profiling data as plugin output.
> * `emulation/rocjitsu/docs/plugins.md:214` says hook presence is derived, although the group no longer exposes or computes hook presence; only callback policy is derived. The empty-group guarantee also needs the dispatch early-out described separately before it is true operationally.
> * `emulation/mirage/rocjitsu_sys/src/lib.rs:381` still advertises a `profiled` config section that the loader now ignores.
> 
> Updating these together would keep the public configuration and threading descriptions aligned with this change.

There are some good points here. I hadn't considered nested hook calls. One mutex per hook would resolve this (there are no recursive hooks). 

I suppose it's useful to state performance expectations and goals up front, so that reviewers know where to focus. My thinking was that for the case where a single plugin must be serialized, we prioritize code simplicity over performance. At least at this early stage -- only add complexity to speed up the case where all plugins are parallelizable. Not saying that's correct approach, but that was my implicit thought when reviewing

atgutier (commented):
[MAJOR] /tmp/pr8706/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/      17:10:39 [85/1206]
  execution_plugin_group.h:179 — Register observation from a lifecycle callback can deadlock

  Reasoning: onAmdgpuWavefrontHalted() holds callback_mutex_ while invoking plugins. If any plugin in
  the group opts into serialized hot hooks, a register read made by a lifecycle plugin re-enters the
  same group through onAmdgpuReadSgpr() or onAmdgpuReadVgprLanes(). That nested callback tries to
  acquire the same non-recursive mutex on the same thread and blocks permanently.

  Evidence:

  - The lock remains held around plugin code at /tmp/pr8706/emulation/rocjitsu/lib/rocjitsu/src/
    rocjitsu/vm/plugins/execution_plugin_group.h:179, while hot hooks reuse the mutex at /tmp/pr8706/
    emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/execution_plugin_group.h:236.

  - The existing HaltSnapshotPlugin performs supported live-register reads from its halt callback at /
    tmp/pr8706/emulation/rocjitsu/tests/halt_snapshot_plugin.h:107.

  - SGPR reads notify the plugin group at /tmp/pr8706/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/
    amdgpu/compute_unit.h:409; VGPR observations do likewise at /tmp/pr8706/emulation/rocjitsu/lib/
    rocjitsu/src/rocjitsu/vm/amdgpu/register_access.h:1029.

  - The halt callback deliberately runs before registers are freed at /tmp/pr8706/emulation/rocjitsu/
    lib/rocjitsu/src/rocjitsu/vm/amdgpu/wavefront.cpp:15.

  Suggested fix: make group serialization re-entrant for the owning thread while retaining exclusion
  against other threads—either a recursive mutex or an outermost-dispatch owner/depth mechanism.
  Alternatively, provide an explicitly non-notifying snapshot accessor for lifecycle callbacks. Add a
  timeout-bounded regression combining a halt-snapshot plugin with any serial-hot-hook plugin.

  Confidence: high.

 [MAJOR] /tmp/pr8706/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/plugin17:10:42 [54/1206]
  Removing the ABI discriminator turns stale plugins into crashes or wrong-vtable calls

  Reasoning: the previous metadata structure began with int abi; the new structure begins with const
  char *name. When a previously built v2 plugin is discovered, the loader interprets abi == 2 as a
  pointer near address 0x2 and dereferences it during the metadata-name comparison. Even if metadata
  handling happens to survive, the newly inserted requires_serial_execution() virtual shifts the old
  plugin’s hook slots, producing undefined behavior when hooks are called.

  Evidence:

  - The new metadata layout starts with name at /tmp/pr8706/emulation/rocjitsu/lib/rocjitsu/src/
    rocjitsu/vm/plugins/plugin_abi.h:69.

  - The loader reads that layout without a version or magic check at /tmp/pr8706/emulation/rocjitsu/
    lib/rocjitsu/src/rocjitsu/vm/plugins/plugin_loader.cpp:98.

  - The new virtual method is inserted at /tmp/pr8706/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/
    plugins/execution_plugin.h:56.

  - The base revision had a leading ABI field, rejected mismatches before reading later metadata, and
    included a bad-ABI fixture. This PR removes all three protections.

  - Plugins are still found through normal dynamic-linker paths, where an older installed or build-
    tree .so can remain after a partial update.

  Suggested fix: retain a leading contract version and bump it to v3. Check it before accessing any
  pointer fields or invoking the factory. Restore a fixture that uses the preserved v2 layout—not
  merely the current layout with a different integer—and verify rejection occurs before factory
  invocation.

  Confidence: high.

[MAJOR] /tmp/pr8706/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/      17:10:45 [21/1206]
  execution_plugin_group.h:96 — Plugin-free parallel simulations now contend on one global mutex

  Reasoning: every lifecycle, dispatch, workgroup, wavefront, and barrier callback locks
  callback_mutex_ before iterating plugins_, even when the vector is empty. All default components
  share the single process-wide empty_group() instance. Independent Linux simulation partitions
  therefore serialize these events on one mutex despite having no plugin work to perform.

  Evidence:

  - The unconditional locking starts at /tmp/pr8706/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/
    plugins/execution_plugin_group.h:96, with the mutex helper at /tmp/pr8706/emulation/rocjitsu/lib/
    rocjitsu/src/rocjitsu/vm/plugins/execution_plugin_group.h:236.

  - The process-wide singleton is returned at /tmp/pr8706/emulation/rocjitsu/lib/rocjitsu/src/
    rocjitsu/vm/plugins/execution_plugin_group.h:217.

  - These callbacks occur per dispatch, workgroup, wavefront, completion, and barrier—for example /
    tmp/pr8706/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp:1070, /
    tmp/pr8706/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/wavefront.cpp:24, and /tmp/
    pr8706/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/compute_unit.cpp:193.

  - The base implementation performed only an empty loop. The new behavior contradicts the documented
    “near-zero overhead” empty path.

  This is likely to regress Linux multi-partition scaling. The exact magnitude depends on dispatch
  granularity and thread count, but the shared mutex can force cache-line bouncing and eventually
  mutex parking/wakeup under contention.

  Suggested fix: bypass locking and callback construction when plugins_.empty(). Document that group
  membership is frozen before execution, or publish a stable immutable has_plugins flag. Add a
  concurrent empty-group benchmark or scaling regression.

  Confidence: high.

  NIT

  [NIT] /tmp/pr8706/emulation/rocjitsu/tests/execution_plugin_test.cpp:357 — Positive-concurrency test
  depends on a 100 ms scheduler window

  The overlap test expects the second thread to enter within 100 ms. An overloaded or instrumented
  runner can miss that window and report no overlap even when the implementation is correct. A
  deterministic two-party handshake would remove that timing dependency.

  Confidence: medium.

lialan:
> I suppose it's useful to state performance expectations and goals up front, so that reviewers know where to focus. My thinking was that for the case where a single plugin must be serialized, we prioritize code simplicity over performance. At least at this early stage -- only add complexity to speed up the case where all plugins are parallelizable. Not saying that's correct approach, but that was my implicit thought when reviewing

@newling I agreed. I used one recrusive mutex to preserve cross hook exclusion while allowing nested register reads.

docs/superpowers/specs/2026-08-04-pr8706-review-followup-design.md
```diff
@@ -0,0 +1,65 @@
+# PR 8706 Review Follow-up Design
+
```
newling: This file not for merging?

newling (changes requested):
Please remove the superpower doc

docs/superpowers/specs/2026-08-04-pr8706-review-followup-design.md
lialan: Removed.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/plugin_abi.h
```diff
@@ -69,16 +67,15 @@ inline constexpr int kPluginAbiVersion = 2;
 /// `rocjitsu_plugin_destroy` may free it.
 using PluginHandle = void *;
 
+/// @brief Signature of the generated same-build identity accessor.
+using PluginBuildIdentityFn = const char *(*)();
```
newling: What is this for? Our approach to plugins is that if they're not in tree we don't support them and we don't try to avoid sharp edges. It seems to me (and I might be wrong) that this "symbol name of the same-build identity accessor" is going against that idea.

newling (commented):
Question about the change to plugin_abi.h, I was expecting it to be just code removal but it looks like somethings been added, and I'm not sure what it is. Left a comment in line

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/plugin_abi.h
lialan: right, this only guarded stale builds, removed.

kuhar (commented):
Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/execution_plugin.h:63`

We should spell out the default cold-vs-hot concurrency boundary wherever plugin authors act on it. When `requires_serial_hot_hooks()` is false, `dispatch_with_plugin_lock()` excludes other infrequent callbacks, but a high-frequency callback can still overlap it. The current “always serialized” and “Serialized by ExecutionPluginGroup” wording can therefore overstate the guarantee. Could we qualify the header and `docs/plugins.md`, add the hot-hook audit to the “Adding a new plugin” checklist, and cover cold-vs-hot overlap in the concurrency tests?

Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/plugin_loader.cpp:227`

It would be good to warn when a config still contains the removed top-level `profiled` key. This key was previously documented, but the new parser silently ignores it and proceeds without `HOOK_PROFILE` output. Detecting it while the root map is already available would make the behavior change visible instead of silently changing the run.

newling:
I had an agent separate out the 2 orthogonal removals in https://github.com/ROCm/rocm-systems/pull/9731 but it's fine if they're included in this PR rather (up to you). It cut a bit deeper than this PR, there's a summary of what knock-on changes it includes in an agent review of this PR: https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr8706.md. So the only asks are now to please (1) update PR description and (2) finish up removing debris (or feel free to extract a base commit that is 9731 and rebase your actual multithreading-related work on that, all in this PR)

newling:
> Please remove the superpower doc

On develop: https://github.com/ROCm/rocm-systems/tree/develop/docs/superpowers/specs

Your agent is really determined to commit these 😄

emulation/rocjitsu/tests/execution_plugin_test.cpp
```diff
@@ -712,6 +870,150 @@ TEST(ExecutionPluginTest, NoPluginNoCrash) {
...
+      probe_ptr->probe(), std::chrono::milliseconds(20),
+      [&]() { group.onAmdgpuWorkgroupCompleted(0, 0); },
+      [&]() { group.onAmdgpuReadSgpr(nullptr, 0); });
+
+  EXPECT_TRUE(result.first_entered);
+  EXPECT_FALSE(result.overlap_observed);
+  EXPECT_EQ(probe_ptr->probe().max_active(), 1);
+}
+
+TEST(ExecutionPluginTest, EmptyGroupDispatchIsSafeAcrossThreads) {
```
kuhar: It would be useful for this test to distinguish the empty-group fast path from serialized dispatch. Removing the `plugins_.empty()` return would still pass this test, but it would route every default group through the process-wide singleton mutex. Could we add a deterministic test seam (for example, a test-only lock-acquisition counter) that fails when empty dispatch reaches the lock? A timing bound would be much more scheduler-sensitive.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/execution_plugin.h
```diff
@@ -53,64 +53,91 @@ class ExecutionPlugin {
...
   // -- Lifecycle hooks ------------------------------------------------------
 
   /// Called when the emulated driver opens (simulation is ready to accept work).
+  /// Infrequent hook; see the concurrency contract on requires_serial_hot_hooks().
   virtual void onInit() {}
 
   /// Called when the emulated driver closes (simulation is shutting down).
   /// All simulation state is still valid during this callback.
+  /// Infrequent hook; see the concurrency contract on requires_serial_hot_hooks().
   virtual void onShutdown() {}
```
kuhar, line 81: We should document the lifecycle callbacks separately from the hot/infrequent hook policy. The group mutex does not isolate `onInit()` or `onShutdown()` from the default unlocked hot hooks, so a plugin could otherwise infer that teardown is safe while a register or instruction callback is still using its state. Please state the host-side precondition here—initialization completes before simulation callbacks start, and shutdown begins only after they stop—and make sure every caller enforces it; if that ordering is not guaranteed, the lifecycle path needs synchronization beyond this lock.

kuhar (commented):
A few comments

lialan:
@kuhar @newling @atgutier @amd-arosa The latest review feedback has been addressed. Could you please take another look at this PR when you have a chance? Thanks!

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/execution_plugin.h
lialan: addressed!

newling:
@lialan please respond to (1) and (2) that I mentioned in https://github.com/ROCm/rocm-systems/pull/8706#issuecomment-5209191693 here, maybe I'm missing something but I'm not sure that have been addressed? ie

(1) the title is no longer accurate for what this change does
(2) have we removed code that isn't needed, now that ABI versioning is gone?

lialan:
@newling Addressed both issues.
* updated title.
* residuals are removed as well.

newling (approved):
> @newling Addressed both issues.
> 
> * updated title.
> * residuals are removed as well.

Thanks. In that case it's all good from my end, LGTM

PR #8708, opened 2026-07-16
feat(rocr): Suballocate system memory

dayatsin-amd:
## Summary

Extends the ROCr fragment suballocator to **system memory** (both fine- and
coarse-grain), not just VRAM. Previously only local (device) memory could be
suballocated, so every small system allocation (e.g. `hipHostMalloc`, signals,
kernarg, internal runtime structures) required a dedicated kernel BO. Sharing a
2 MB block via the fragment allocator reduces BO count and per-allocation
overhead for small system allocations.

The change is broken into focused commits:

- **Allow suballocator for system memory allocations** — gate suballocation on
  `!NoAddress` instead of `IsLocalMemory()`.
- **Map system memory fragment blocks to all GPUs** — only apply
  `AllocateRestrict` to local backing blocks so system fragments inherit GPU
  mappings from their 2 MB block.
- **Tear down system allocator before destroying agents** — avoids a
  shutdown-time use of the fragment heap after its backing agent is gone.
- **Capture raw region pointer in system allocator closure** — the closure
  captured a `std::shared_ptr<const MemoryRegion>`, forcing `std::function` to
  heap-allocate its capture; tearing down the `BaseShared` copy during
  `Unload()` overran that block (ASan-confirmed heap-buffer-overflow). Capturing
  a raw pointer (the region outlives the closure) makes the closure trivially
  destructible and stored inline.
- **Zero-initialize suballocated system fragments** — recycled fragments hold
  stale data, unlike dedicated BOs backed by fresh (zeroed) kernel pages.
  Callers relying on freshly allocated system memory being zero (e.g. an
  `hipHostMalloc` buffer used as a GPU/host atomic accumulator) otherwise see
  garbage. Fragments are zeroed on allocation to match dedicated-BO semantics;
  local (device) memory is not host-writable here and is left untouched.
- **Add rocrtst for system-memory fragment overlap and coherence** — a new
  `MemoryFragment` rocrtst suite that directly exercises the fine-grain
  system-memory suballocator (see Test plan for the specific cases).
- **Skip the new tests when the suballocator is disabled** — when
  `HSA_DISABLE_FRAGMENT_ALLOCATOR=1` (the same flag the runtime honors),
  allocations are served by dedicated BOs rather than sub-allocated fragments,
  so the overlap / zero-init / coherence checks no longer exercise their
  intended path. The tests skip in that case (emitting the conventional
  `[ SKIPPED ]` marker, since the vendored gtest predates `GTEST_SKIP()`) rather
  than reporting misleading coverage.

## JIRA ID

JIRA ID : AILIROCR-70

## Test plan

Tests run on **gfx950** (e3-2b, CDNA) and **gfx1100** (RX 7900 GRE, RDNA3) with
these changes applied — all passing:

- [x] **Full HIP catch test suite** (4293 tests): **99% pass**. **0 heap
      corruption**, **0 system-scope atomic failures**, **0 cooperative-kernel
      failures**. The only 4 failures are pre-existing flakies that also
      fail/flake on the unmodified baseline lib and pass in isolation:
      `hipStreamAttachMemAsync_Negative` (fails on baseline too),
      `hipExtLaunchKernelGGL_Functional`, `hipStreamAddCallback_StrmSyncTiming`,
      and the `ffsll` device-intrinsic sanity test (unrelated to memory).
- [x] **`atomicCAS_system` / `atomicAdd_system` / `atomicSub_system`**
      (all type variants): pass with the zero-init fix. These rely on
      zero-initialized `hipHostMalloc` buffers and previously failed without it.
- [x] **`hipLaunchCooperativeKernel_Basic` / `...MultiDevice_Basic`**: pass with
      the zero-init fix (also depend on a zeroed `hipHostMalloc` accumulator);
      confirmed deterministically passing in isolation.
- [x] **New `rocrtstFunc.Memory_Fragment_*` suite** (built from source and run
      on gfx950 and gfx1100): all three cases pass against the patched runtime.
  - `Memory_Fragment_Overlap_Test`: allocates many small fragments with unique
    per-buffer/per-offset fingerprints; verifies no clobbering and no
    address-range overlap, including across a free/re-allocate churn phase that
    recycles backing storage. Runs on both fine-grain system memory and
    device/VRAM (coarse-grain).
  - `Memory_Fragment_ZeroInit_Test`: verifies fresh and recycled fragments are
    zero-initialized (system memory only — the runtime intentionally zeroes only
    system fragments, not device fragments).
  - `Memory_Fragment_Coherence_Test`: GPU atomically adds into a sub-allocated
    fragment; confirms the CPU observes the result after a system-scope release.
    Runs on both fine-grain system memory and device/VRAM. Device cases are
    skipped on non-large-BAR systems where the host cannot access VRAM (as seen
    on the gfx1100 box, where the system cases ran and the device cases were
    skipped).
  - Sensitivity check: with the fragment zero-init removed,
    `Memory_Fragment_ZeroInit_Test` fails at the first recycled fragment
    (Overlap/Coherence still pass), confirming the test detects the regression.
  - Disabled-allocator check: with `HSA_DISABLE_FRAGMENT_ALLOCATOR=1` all three
    cases report `[ SKIPPED ]` and do no work; unset/`=0` they run normally.
- [x] **Targeted HSA reproducers**: single- and multi-threaded fine-grain
      suballocation stress — no fragment overlap and correct CPU coherence; a
      recycled-fragment probe confirmed the stale-data (zero-init) root cause.

Baseline comparison: the unmodified lib fails 13 tests (flakies + graph-node
timeouts); with these changes the failing set is a strict subset (4), i.e. no
new regressions introduced.


projects/rocr-runtime/rocrtst/suites/functional/memory_fragment.cc
```diff
@@ -0,0 +1,408 @@
...
+  memset(oldValues, 0, kBufBytes);
+
+  err = hsa_amd_agents_allow_access(1, &gpuAgent, NULL, sysMemory);
+  ASSERT_EQ(err, HSA_STATUS_SUCCESS);
+  err = hsa_amd_agents_allow_access(1, &gpuAgent, NULL, gpuScratch);
+  ASSERT_EQ(err, HSA_STATUS_SUCCESS);
+  err = hsa_amd_agents_allow_access(1, &gpuAgent, NULL, oldValues);
+  ASSERT_EQ(err, HSA_STATUS_SUCCESS);
+  err = hsa_amd_agents_allow_access(1, &gpuAgent, NULL, kernArguments);
+  ASSERT_EQ(err, HSA_STATUS_SUCCESS);
```
cfreeamd, line 431: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** `kernArguments->n` is never assigned here (or anywhere in this function). The `args_t` struct has an `n` field at line 45 of this file; `d` is set to `kValue` but `n` is left at whatever the allocator returns — which for a dedicated kernarg BO is zero-initialized kernel pages, so `n=0`.

If `test_atomic_add` does `if (tid < n) a[tid] += d`, the kernel won't execute any threads, `sysMemory` stays zero, and `ASSERT_EQ(sysMemory[i], kValue)` will fail for all elements. If the kernel uses the grid size without an `n` bounds check, this is fine.

Worth confirming which applies — if `n` is needed, add `kernArguments->n = kNumElements;` after `kernArguments->d = kValue;`.

cfreeamd (commented):

ROCr-scoped files reviewed: `amd_kfd_driver.cpp`, `amd_memory_region.cpp`, `runtime.cpp` (runtime changes) and `memory_fragment.cc/.h`, `common.{cc,h}`, `main.cc`, `test_categories.yaml` (tests). Inline comments on two points in `memory_fragment.cc`; no comments on the runtime files (those changes look correct).

projects/rocr-runtime/rocrtst/suites/functional/memory_fragment.cc
```diff
@@ -0,0 +1,408 @@
...
+constexpr size_t kFragmentBlockSize = 2 * 1024 * 1024;
+
+// Deterministic, per-buffer/per-offset fingerprint. Distinct (id, offset)
+// pairs produce distinct bytes, so if two live fragments ever overlap, the
+// later writer clobbers the earlier one and read-back verification fails.
+inline uint8_t FragByte(uint32_t id, size_t offset) {
+  uint32_t v = id * 2654435761u + static_cast<uint32_t>(offset) * 40503u + 0x9e37u;
+  return static_cast<uint8_t>(v ^ (v >> 8) ^ (v >> 16));
+}
+
```
cfreeamd, line 39: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `kValue` and the `args_t` typedef below it (lines 40–46) are declared at file scope *outside* the anonymous namespace, between `}  // namespace` and the class method definitions. `kFragmentBlockSize` and `FragByte` (lines 26–34) are correctly inside the anonymous namespace above. Moving `kValue` and `args_t` into the same anonymous namespace (or making `args_t` function-local to `FragmentCoherenceTest`, which is the only user) would keep all TU-private definitions consistently scoped.

projects/rocr-runtime/rocrtst/suites/functional/memory_fragment.cc
dayatsin-amd: Good catch — confirmed `test_atomic_add` is grid-size bounded and takes only `(sysMemory, gpuMemory, oldValues, value)`; it never reads a count arg (`oldValues[tid] = atomic_add(&sysMemory[tid], value)` with `tid = get_global_id(0)`). So `n` was dead. Rather than assign it, I dropped the unused `n` field from `args_t` entirely (commit 0b89ed8d15). Coherence test still passes on gfx950 and gfx1100.

projects/rocr-runtime/rocrtst/suites/functional/memory_fragment.cc
dayatsin-amd: Done — moved `kValue` and `args_t` into the anonymous namespace alongside `kFragmentBlockSize`/`FragByte` so all TU-private definitions are consistently scoped (commit 0b89ed8d15).

cfreeamd:

Both prior findings resolved:
- **[Question] `kernArguments->n`**: author confirmed `test_atomic_add` is grid-size bounded and never reads a count arg; dropped the dead `n` field from `args_t` (commit `0b89ed8d`). Verified on gfx950 and gfx1100. Convincing resolution.
- **[Nit] `kValue`/`args_t` scope**: moved into the anonymous namespace alongside `kFragmentBlockSize`/`FragByte` (same commit).

The follow-on commit (`ac160231`) also replaced `fprintf(stdout, ...)` skip notices with `std::cout` and dropped the unused `<cstdio>` include — consistent with the rest of the suite.

The device/VRAM pool coverage added since the first review (both `FragmentOverlapOnPool` and `FragmentCoherenceOnPool` exercised on large-BAR VRAM in addition to fine-grain system memory) looks correct: `PoolAccessibleByAgent` guards ensure the test only runs where the CPU can read VRAM; CPU access is explicitly granted via `hsa_amd_agents_allow_access` before host verification; the lighter workload (`kSizesLight`, 128 buffers) bounds runtime over the slower PCIe BAR path.

No remaining concerns.

cfreeamd (approved):

PR #8710, opened 2026-07-16
fix(rocshmem) Make hsakmt compile on rocm 7.13/14 despite errors in hsakmt-config.cmake 

abouteiller:
## Motivation

rocSHMEM with -DUSE_SDMA=ON will not compile on theRock 7.13 and 7.14 because the hsakmt cmake package imports RHEL specific paths. 

## Technical Details

Workaround: remove the offending paths from the imported target, and make sure we search for hsakmt dependencies in the right place instead.

## Issue Tracking

JIRA ID: AIROCSHMEM-464

## Test Plan

Builds and run SDMA on both rocm-7.13 and rocm-7.14. Still builds on older rocm.

## Test Result

Builds and run with both, installed as deb packages on Ubuntu. Tested still functional on 7.2.1.

## Submission Checklist

### DIFF
--- a/projects/rocshmem/CMakeLists.txt
+++ b/projects/rocshmem/CMakeLists.txt
@@ -314,6 +314,21 @@ if (NOT BUILD_TESTS_ONLY)
 
   if (USE_SDMA)
     find_package(hsakmt REQUIRED)
+    # Workaround for RoCM 7.13/14
+    # hsakmtTargets.cmake contains paths baked in from the ROCm RHEL CI build:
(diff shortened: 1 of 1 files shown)


edgargabriel:
we really need to add an SDMA compilation path to our CI

edgargabriel (approved):
compilation worked with the modified rocJenkins CI run which activated the SDMA path. The tests themselves failed, but it didn't look like its related to this problem, the regular IPC tests failed (without SDMA) with out-of-memory issues, it looked more like GPUs being in a bad state.

PR #8711, opened 2026-07-16
feat(rccl): Share QPs between PAT ReduceScatter and AllGather

thomas-huber:
## Motivation

PAT walks the same binomial neighbors for ReduceScatter and AllGather, but the two collectives used
opposite peer directions on them: ReduceScatter received from `rank-delta` and sent to `rank+delta`,
AllGather did the mirror. Connections are directional, so every neighbor link existed twice, each
copy carrying its own queue pairs and ring buffers. `ncclTransportPatConnect` ran two
`ncclTransportP2pConnect` passes per binomial mask per channel, and a communicator paid for both
sets whether or not it ever issued both collectives.

This is inherited from upstream NCCL, where the two-pass connect has been unchanged since PAT was
added in 2.23.4. That is why the change ships behind an opt-out.

## Technical Details

AllGather is flipped onto ReduceScatter's peer directions so both collectives address one
connection set. Four sites independently spell out that direction and all four have to agree:

| File | Change |
|---|---|
| `src/transport/generic.cc` | one connect pass per mask instead of two |
| `src/device/prims_simple.h` | both PAT modes use `recv = rank-delta`, `send = rank+delta` |
| `src/include/collectives.h` | `PatAGAlgorithm` data-block rank re-indexed to `(rank - s + nranks) % nranks` |
| `src/proxy.cc` | `ncclPatternPatDown` peers match `PatUp` |

If the transport and proxy disagree the run hangs; if only `PatAGAlgorithm` disagrees the payload is
wrong. Both are covered by the unit tests.

**Opt-out.** `RCCL_PAT_SHARED_QPS` (default `1`); `0` restores the original two-pass behavior. The
flag has to reach device code, so it is read in `commAlloc`, carried through
`ncclKernelComm::patSharedQps`, and read on device from `ncclShmem.comm.patSharedQps`.
`PatAGAlgorithm` gained a required `sharedConns` parameter with no default, so a missed call site is
a compile error rather than a silent mismatch. The mirror pass is emitted before the shared pass so
the disabled path reproduces the original connect order exactly, and the `Connected binomial trees`
init message gains a suffix only when sharing is off.

**Scale of the saving.** One pass instead of two does not halve a job's queue pairs: ring and tree
connections are unaffected, and the `delta = nRanks/2` mask is its own mirror and was never
doubled. Whole-communicator reduction is 25% at 8 nodes rising to 36.65% at 64; measured against
PAT's own connections it is a steady 45-47%.

## Issue Tracking

JIRA ID: AICOMRCCL-1538

## Test Plan

1 rank per node with Direct PAT forced (`NCCL_PAT_ENABLE=1`, `NCCL_ALGO=PAT`,
`RCCL_HIERARCHICAL_ALLGATHER=0`), 32KiB to 1GiB.

1. Queue-pair counts from `NCCL_DEBUG_SUBSYS=INIT,NET` logs, compared against a closed-form
   prediction, at 8/16/32/64 nodes and at ten non-power-of-two rank counts.
2. Correctness asserted on every run (`Out of bounds values : 0 OK`), together with a non-zero PAT
   row count so a run that silently fell back to Ring cannot be counted as a pass.
3. Opt-out validated with three arms rather than two: the pre-sharing parent commit, the flag build
   with sharing off, and the same library with sharing on. The load-bearing assertion is that the
   flag-off count equals the parent count exactly.
4. gtest MPI fixtures at 4 and 8 ranks covering both the shared path and the fallback.
5. Everything re-run after rebasing onto current develop.

## Test Result

Queue pairs per communicator, baseline versus shared. The totals are summed over every rank in the
communicator; the per-rank columns are what a single process opens on its own NIC, which is the
number that matters against a device queue-pair budget:

| Nodes | Total baseline | Total shared | Per rank baseline | Per rank shared | Reduction |
|---:|---:|---:|---:|---:|---:|
| 8 | 960 | 720 | 120 | 90 | 25.00% |
| 16 | 2688 | 1872 | 168 | 117 | 30.36% |
| 32 | 6912 | 4560 | 216 | 142 | 34.03% |
| 64 | 16896 | 10704 | 264 | 167 | 36.65% |

Every measured count matches the closed-form prediction. Per-rank figures are averages; the shared
count is not perfectly uniform across ranks, and the busiest rank is the relevant one for a device
limit: 100 at 8 nodes, 124 at 16, 148 at 32 and 172 at 64.

Queue pairs also scale linearly with channel count, so `NCCL_MIN_NCHANNELS` /
`NCCL_MAX_NCHANNELS` and `NCCL_IB_QPS_PER_CONNECTION` move these numbers proportionally. All
figures above are at the Ruby default of 8 channels with one queue pair per connection.

Non-power-of-two rank counts 5, 6, 7, 9, 12, 15, 17, 24 and 31 all reduce, by 12.5% to 40.3%. At 3
ranks the count is unchanged, which is correct rather than a miss: the mask set `{1,2}` is closed
under `d -> nRanks-d`, so the existing dedup in `transport.cc` already collapsed the second pass.
This holds only for 2 and 3 ranks.

With `RCCL_PAT_SHARED_QPS=0` the count reproduced the parent commit exactly at all 13 scales
tested, powers of two and not, and the separate-connections log marker appeared on every rank with
the flag off and never with it on.

Both unit tests pass at 4 and 8 ranks.
`PatSharedConnectionMPITest.AllGatherReusesReduceScatterConnections` snapshots each channel's
connected flags after `ncclCommInitRank` and asserts the set a PAT AllGather adds is exactly the
shared set; `SeparateConnectionsWhenSharingDisabled` asserts the mirrored connections do exist with
the flag off, so the fallback cannot bit-rot unnoticed. `NCCL_PARAM` caches into a process-lifetime
static, so the two cannot run in one invocation; each skips with an actionable message when the
cached value is wrong, and the fallback runs in a second invocation.

AllGather and ReduceScatter were correct at every scale in every arm, with no regression against
the baseline. All counts after the rebase are identical to those before it.

## Submission Checklist

- [x] Validated at 8, 16, 32 and 64 nodes and at ten non-power-of-two rank counts.
- [x] Opt-out provided and proven to restore the original connection set exactly.
- [x] Unit tests cover both the shared path and the fallback.
- [x] Rebased onto current develop and revalidated.

### DIFF
--- a/projects/rccl/src/device/all_gather.h
+++ b/projects/rccl/src/device/all_gather.h
@@ -181,7 +181,8 @@ struct RunWorkColl<ncclFuncAllGather, T, RedOp, NCCL_ALGO_PAT, NCCL_PROTO_SIMPLE
 
     if (tid == nworkers) { // Algo computation thread
       PatAGAlgorithm<T> patAlgo(chunkCount * sizeof(T), NCCL_STEPS, NCCL_PAT_NWORKERS / WARP_SIZE, channelOffset,
-                                channelOffset + channelCount, count, chunkCount, rank, nranks);
+                                channelOffset + channelCount, count, chunkCount, rank, nranks,
+                                ncclShmem.comm.patSharedQps);
       int parallelFactor = shmem->parallelFactor = patAlgo.getParallelFactor();
       (void)parallelFactor;// unused variable - compiler warning
       int step = 0;
--- a/projects/rccl/src/device/prims_simple.h
+++ b/projects/rccl/src/device/prims_simple.h
@@ -739,8 +739,10 @@ public:
     if (tid < 32 && ((1UL << tid) < nranks)) {
       int rank = ncclShmem.comm.rank;
       uint32_t delta = 1 << tid;
+      // When sharing, RS and AG both recv from rank-delta and send to rank+delta; otherwise AG mirrors RS.
+      const bool shared = ncclShmem.comm.patSharedQps != 0;
       // Load recv peer
-      int recvPeer = mode == primsModePatRs ? (rank - delta + nranks) % nranks : (rank + delta) % nranks;
+      int recvPeer = (shared || mode == primsModePatRs) ? (rank - delta + nranks) % nranks : (rank + delta) % nranks;
       struct ncclPatPeer* peer = ((struct ncclPatPeer*)recvPeers) + tid;
       struct ncclConnInfo* conn = peer->conn = channel->peers[recvPeer]->recv + connIndexRecv;
       peer->step = conn->step;
@@ -750,7 +752,7 @@ public:
       peer->accSize = 0;
       peer->connStepSize = conn->stepSize / sizeof(T);
       // Load send peer
-      int sendPeer = mode == primsModePatAg ? (rank - delta + nranks) % nranks : (rank + delta) % nranks;
+      int sendPeer = (!shared && mode == primsModePatAg) ? (rank - delta + nranks) % nranks : (rank + delta) % nranks;
       peer = ((struct ncclPatPeer*)sendPeers) + tid;
       conn = peer->conn = channel->peers[sendPeer]->send + connIndexSend;
       peer->step = conn->step;
--- a/projects/rccl/src/include/collectives.h
+++ b/projects/rccl/src/include/collectives.h
@@ -713,6 +713,7 @@ public:
   }
 };
 
+// When sharedConns is set, AllGather reuses PatRSAlgorithm's peer directions, so its data blocks are indexed rank-s.
 template <typename T>
 class PatAGAlgorithm {
   size_t offset;
@@ -722,6 +723,7 @@ class PatAGAlgorithm {
   int nelem;
   int rank;
   int nranks;
+  int sharedConns;
   int nrPow2;
   int postFreq;
   int lastA;
@@ -806,8 +808,9 @@ class PatAGAlgorithm {
 
 public:
   __device__ __host__ PatAGAlgorithm(int stepSize, int stepDepth, int maxParallelFactor, size_t offset, size_t end,
-                                     size_t count, int chunkCount, int rank, int nranks)
-    : offset(offset), end(end), count(count), chunkCount(chunkCount), rank(rank), nranks(nranks) {
+                                     size_t count, int chunkCount, int rank, int nranks, int sharedConns)
+    : offset(offset), end(end), count(count), chunkCount(chunkCount), rank(rank), nranks(nranks),
+      sharedConns(sharedConns) {
     parallelFactor = maxParallelFactor;
     aggDelta = nrPow2 = (1 << log2Up(nranks));
 
@@ -844,7 +847,8 @@ public:
     } else if (phase == 0) {
       int s = a * aggDelta + as;
       if (s >= nranks) skip = 1;
-      int recvDataRank = (rank + s) % nranks;
+      // C++ % truncates toward zero, so (rank - s + nranks) % nranks is still negative when s > rank + nranks.
+      int recvDataRank = sharedConns ? ((rank - s) % nranks + nranks) % nranks : (rank + s) % nranks;
       ps->outIx = recvDataRank * count + offset;
       ps->sendDim = -1;
       ps->recvDim = 0;
@@ -859,7 +863,7 @@ public:
       if (s >= nranks) skip = 1;
       ps->sendDim = firstBitSet(s, nrPow2);
       s -= (1 << ps->sendDim);
-      int sendDataRank = (rank + nranks + s) % nranks;
+      int sendDataRank = sharedConns ? ((rank - s) % nranks + nranks) % nranks : (rank + nranks + s) % nranks;
       ps->outIx = sendDataRank * count + offset;
       ps->recvDim = s ? firstBitSet(s, nrPow2) : -1;
       ps->sendOffset = ps->recvOffset = (a % postFreq) * nelem;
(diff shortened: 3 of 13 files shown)


srirajpaul (commented):
looks good. should we have CVAR to go back to original behavior in case anything goes wrong? Also were you able to test for non-powers of 2 like 5 ranks?

thomas-huber:
> looks good. should we have CVAR to go back to original behavior in case anything goes wrong? Also were you able to test for non-powers of 2 like 5 ranks?

Hey @srirajpaul, thanks for your review!

For the non-powers of 2, definitely. I tested behavior and performance at 3, 5, 6, 7, 9, 12, 15, 17, 24, and 31 nodes. Everything looked solid, no perf regressions. If you'd like to take a look at the data, please checkout the Jira ticket where i've attached the non-power2 experiment info.

Regarding the CVAR to go back to original behavior... this is a good point. I've just reached out and started a convo internally about this so let's see what our teammates have to say.

thomas-huber:
@srirajpaul Per our teams discussion, I've added an environment variable to allow us to toggle this shared-qp improvement on and off. See `NCCL_PAT_SHARED_QPS`

srirajpaul (approved):
looks good. you named it NCCL_PAT_SHARED_QPS. generally, our own CVARS we name with RCCL, like RCCL_PAT_SHARED_QPS.

thomas-huber:
> looks good. you named it NCCL_PAT_SHARED_QPS. generally, our own CVARS we name with RCCL, like RCCL_PAT_SHARED_QPS.

Hey @srirajpaul I've RCCL-ified the name of the env var. I also noticed that the lazy qp init env var is NCCL_* based. I will address that in a separate follow-up PR.

Re-requesting review.

mustafabar (commented):
I asked Opus 5 to review this, and posted the comments that I agree with
 
 2. All three new tests skip in ordinary CI.
  Every test gates on oneRankPerNode() and min_processes=4, so on a single-node 8-GPU CI runner nothing runs. SeparateConnectionsWhenSharingDisabled
  additionally skips unless RCCL_PAT_SHARED_QPS=0 is in the environment before launch, and nothing in the diff wires a second gtest invocation. As it stands
  the opt-out path — the whole safety argument for the default flip — has no CI coverage and can bit-rot exactly as the comment says it shouldn't. Either
  add the second invocation to the test driver, or the fallback is effectively untested.

3. Missing coverage for grouped RS+AG.
  AlternatingReduceScatterAndAllGatherOnOneCommunicator covers sequential launches. The new risk sharing introduces is two collectives with different step
  counts queuing proxy ops on the same connection and step counter — that is most exposed inside one ncclGroupStart/End, which isn't tested.

4. PatAGAlgorithm phase 0 can now produce a negative index.
  recvDataRank = (rank - s + nranks) % nranks with s >= nranks (the skip = 1 case) goes negative, making outIx a large negative intptr_t. Upstream's (rank +
  s) % nranks was non-negative by construction. patCopy appears to discard skipped steps, so this is probably latent rather than live — but the safe form
  ((rank - s) % nranks + nranks) % nranks costs nothing and removes the dependence on that guarantee.

Minor: the PR description says NCCL_PAT_SHARED_QPS; the code registers RCCL_PAT_SHARED_QPS. Worth fixing in the description before merge so the opt-out people copy/paste actually works.

thomas-huber:
> I asked Opus 5 to review this, and posted the comments that I agree with
> 
> 2. All three new tests skip in ordinary CI.
>    Every test gates on oneRankPerNode() and min_processes=4, so on a single-node 8-GPU CI runner nothing runs. SeparateConnectionsWhenSharingDisabled
>    additionally skips unless RCCL_PAT_SHARED_QPS=0 is in the environment before launch, and nothing in the diff wires a second gtest invocation. As it stands
>    the opt-out path — the whole safety argument for the default flip — has no CI coverage and can bit-rot exactly as the comment says it shouldn't. Either
>    add the second invocation to the test driver, or the fallback is effectively untested.
> 3. Missing coverage for grouped RS+AG.
>    AlternatingReduceScatterAndAllGatherOnOneCommunicator covers sequential launches. The new risk sharing introduces is two collectives with different step
>    counts queuing proxy ops on the same connection and step counter — that is most exposed inside one ncclGroupStart/End, which isn't tested.
> 4. PatAGAlgorithm phase 0 can now produce a negative index.
>    recvDataRank = (rank - s + nranks) % nranks with s >= nranks (the skip = 1 case) goes negative, making outIx a large negative intptr_t. Upstream's (rank +
>    s) % nranks was non-negative by construction. patCopy appears to discard skipped steps, so this is probably latent rather than live — but the safe form
>    ((rank - s) % nranks + nranks) % nranks costs nothing and removes the dependence on that guarantee.
> 
> Minor: the PR description says NCCL_PAT_SHARED_QPS; the code registers RCCL_PAT_SHARED_QPS. Worth fixing in the description before merge so the opt-out people copy/paste actually works.

Hey @mustafabar , thanks so much for the review. I pushed changes to address your comments, let me know what you think.

2. All three new tests skip in ordinary CI.
- PAT only engages at one rank per node so no single-node runner can execute these regardless of GPU count, but I've now registered the suites in test_categories_mpi.yaml with a second RCCL_PAT_SHARED_QPS=0 category so the opt-out gets covered the moment a multi-node job runs them, and I moved the negative-index check into a host-only test that runs in CI today. Do you know the best way to hook rccl-UnitTestsMPI into the existing 4-node lane in therock-test-packages-multi-node.yml, which already allocates exactly our minimum required # of nodes but only runs the heatmap script?

4. Missing coverage for grouped RS+AG.
- Added `Added GroupedReduceScatterAndAllGatherOnOneCommunicator` test which runs both collectives inside one ncclGroupStart/End pair so the proxy ops get queued onto the shared connection.

5. PatAGAlgorithm phase 0 can now produce a negative index.
- I fixed this at all 3 occurrences using the strategy you recommended.

srirajpaul (approved):

nawrinsu (approved):

wenkaidu (approved):

PR #8712, opened 2026-07-16
fix(rocSHMEM): alleviate resource usage

amirsojoodi:
## Motivation

Temporary fix the resource over-usage in rocSHMEM IPC backend.

## Technical Details

Either marking `copy_bulk` as `no_inline` or setting Unrolling factor from 16 to 8 can temporarily silence the CI issues. 

## Issue Tracking

JIRA ID: AIROCSHMEM-455

### DIFF
--- a/projects/rocshmem/src/util.hpp
+++ b/projects/rocshmem/src/util.hpp
@@ -402,7 +402,7 @@ constexpr bool is_blocking(MemcpyKind k) {
 }
 
 template <int ChunkSize, CachePolicy LoadPolicy, CachePolicy StorePolicy, int Unroll>
-__device__ __attribute__((noinline)) void copy_bulk(void* dst, void* src,
+__device__ __noinline__ void copy_bulk(void* dst, void* src,
                                           int n_chunks, int tid, int stride) {
   using Acc = AsmAccess<ChunkSize, LoadPolicy, StorePolicy>;
   using T = typename Acc::type;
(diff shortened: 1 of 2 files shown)


projects/rocshmem/src/util.hpp
```diff
@@ -402,7 +402,7 @@ constexpr bool is_blocking(MemcpyKind k) {
 }
 
 template <int ChunkSize, CachePolicy LoadPolicy, CachePolicy StorePolicy, int Unroll>
-__device__ __forceinline__ void copy_bulk(void* dst, void* src,
+__device__ __attribute__((noinline)) void copy_bulk(void* dst, void* src,
```
omor1: Can we use the HIP macro `__noinline__` instead?

projects/rocshmem/src/util.hpp
```diff
@@ -502,7 +502,7 @@ template <MemcpyKind Kind = MemcpyKind::Put>
   if (size == 0) return;
 
   constexpr int ChunkSize = 16;
-  constexpr int Unroll    = 16;
+  constexpr int Unroll    = 8;
```
omor1, line 505: Usually more of an issue for e.g. arrays, but compilers sometimes have issues promoting function-local `constexpr` variables from the stack to read-only data if they aren't `static constexpr`. It shouldn't be an issue for a plain integer though.

Do we have different unroll or chunk sizes for different functions? If not, it would be nice to have only a single instance of the variable.

Doesn't need to be fixed here, just a comment for something to look at in a future PR.

omor1 (approved):
lgtm, approved on my end if it fixes the issue on gfx942/ROCM 7.0.2/all_backends platform

Comments are not blockers.

projects/rocshmem/src/util.hpp
amirsojoodi: Thank you.
Originally I implemented the memcpy functions with different unrolling and chunkSize factors. Then after some time, I realized that having different values for those is causing occupancy to drop and they stuck there. I will try to collect more data and if they stay the same, I will move them to the top.

PR #8720, opened 2026-07-16
feat(ci): enable KPACK_SPLIT_ARTIFACTS to test multi-arch packaging

geomin12:
Enable the kpack split artifacts flag to see what breaks in non-multi-arch CI. This separates host code from device kernels for better build sharding.

Related: https://github.com/ROCm/TheRock/issues/4769

### DIFF
--- a/.github/workflows/therock-ci-linux.yml
+++ b/.github/workflows/therock-ci-linux.yml
@@ -89,7 +89,7 @@ jobs:
         env:
           amdgpu_families: ${{ env.AMDGPU_FAMILIES }}
           package_version: ${{ inputs.package_version }}
-          extra_cmake_options: "-DTHEROCK_ROCM_SYSTEMS_SOURCE_DIR=../ -DTHEROCK_ENABLE_RCCL=OFF ${{ inputs.cmake_options }}"
+          extra_cmake_options: "-DTHEROCK_ROCM_SYSTEMS_SOURCE_DIR=../ -DTHEROCK_ENABLE_RCCL=OFF -DTHEROCK_FLAG_KPACK_SPLIT_ARTIFACTS=ON ${{ inputs.cmake_options }}"
           BUILD_DIR: build
         run: |
           python TheRock/build_tools/github_actions/build_configure.py
--- a/.github/workflows/therock-ci-windows.yml
+++ b/.github/workflows/therock-ci-windows.yml
@@ -107,7 +107,7 @@ jobs:
           cmake_preset: windows-release
           amdgpu_families: ${{ env.AMDGPU_FAMILIES }}
           package_version: ${{ inputs.package_version }}
-          extra_cmake_options: "-DTHEROCK_ROCM_SYSTEMS_SOURCE_DIR=../ -DTHEROCK_ENABLE_ML_LIBS=OFF ${{ inputs.cmake_options }}"
+          extra_cmake_options: "-DTHEROCK_ROCM_SYSTEMS_SOURCE_DIR=../ -DTHEROCK_ENABLE_ML_LIBS=OFF -DTHEROCK_FLAG_KPACK_SPLIT_ARTIFACTS=ON ${{ inputs.cmake_options }}"
         run: |
           # clear cache before build and after download
           ccache -z


geomin12:
all tests above are expected errors!

.github/workflows/therock-ci-linux.yml
```diff
@@ -89,7 +89,7 @@ jobs:
         env:
           amdgpu_families: ${{ env.AMDGPU_FAMILIES }}
           package_version: ${{ inputs.package_version }}
-          extra_cmake_options: "-DTHEROCK_ROCM_SYSTEMS_SOURCE_DIR=../ -DTHEROCK_ENABLE_RCCL=OFF ${{ inputs.cmake_options }}"
+          extra_cmake_options: "-DTHEROCK_ROCM_SYSTEMS_SOURCE_DIR=../ -DTHEROCK_ENABLE_RCCL=OFF -DTHEROCK_FLAG_KPACK_SPLIT_ARTIFACTS=ON ${{ inputs.cmake_options }}"
           BUILD_DIR: build
         run: |
           python TheRock/build_tools/github_actions/build_configure.py
```
ScottTodd, line 95: Checking the CI run for this: https://github.com/ROCm/rocm-systems/actions/runs/29531359289/job/87770823918?pr=8720

CMake is configured with: `-DTHEROCK_FLAG_KPACK_SPLIT_ARTIFACTS=OFF -DTHEROCK_FLAG_KPACK_SPLIT_ARTIFACTS=ON` and the second `ON` takes precedence:
```
 -- Build flags:
--   * KPACK_SPLIT_ARTIFACTS = ON (-DTHEROCK_FLAG_KPACK_SPLIT_ARTIFACTS=ON)
```

The first flag value is coming from `build_configure.py` itself: https://github.com/ROCm/TheRock/blob/44164a5a330ebd8ae581c1d91f8f4a7251808666/build_tools/github_actions/build_configure.py#L111-L117 . I had planned on actually removing _that_ instead of adding this `extra_cmake_options` layer on top in each of rocm-libraries and rocm-systems. Can we do that instead, or are there other repositories that still need the old behavior of `build_configure.py`?

.github/workflows/therock-ci-linux.yml
ScottTodd, line 1: Comparing CI runs before and after this change to ensure that artifacts are actually being split via kpack:

Before:
* https://github.com/ROCm/rocm-systems/actions/runs/29576353622
* https://therock-ci-artifacts-external.s3.amazonaws.com/ROCm-rocm-systems/29576353622-linux/index.html
  * `blas_lib_gfx94X-dcgpu.tar.zst`

After:
* https://github.com/ROCm/rocm-systems/actions/runs/29531359289?pr=8720
* https://therock-ci-artifacts-external.s3.amazonaws.com/ROCm-rocm-systems/29531359289-linux/index.html
  * `blas_lib_gfx942.tar.zst`

That is what I would expect.

---

Note that the PR run here also ran "Linux MI455 Build" which uploaded `blas_lib_gfx125X-dcgpu.tar.zst`. I'm concerned by that. Both the standard-CI and the MI455-specific CI are building (and uploading!) some common files, so there is potential for build flags to be misaligned across each job and for there to be race conditions and other unexpected issues. If those builds are separate then they should be namespaced and not upload overlapping files to the same cloud storage directory.

We could consider that tangential here, but this is a meaningful flag flip across the jobs - maybe we could flip the kpack flag for that as well?
https://github.com/ROCm/rocm-systems/blob/2ce29e121027dab11d5c1a78cc847d9a655b19cf/.github/workflows/therock-build-linux.yml#L86-L93

(flipping in build_configure.py itself would have the same effect, then we wouldn't need to be as careful)

ScottTodd (commented):
The CI logs here look good I think. I'd prefer to achieve this flag flip via the `build_configure.py` script https://github.com/ROCm/TheRock/blob/44164a5a330ebd8ae581c1d91f8f4a7251808666/build_tools/github_actions/build_configure.py#L111-L117 to roll it out across all repositories and not need to track the extra flag here though.

.github/workflows/therock-ci-linux.yml
geomin12: ah good point! i'm going to land this now: https://github.com/ROCm/TheRock/pull/6694, then all this needs is a hash bump for it to apply

geomin12:
closing PR for https://github.com/ROCm/TheRock/pull/6694

The flag will be applied on hash commit bump after https://github.com/ROCm/TheRock/pull/6694 lands! (i'll check and make sure no suspicious failures related to this)

PR #8721, opened 2026-07-16
fix(rocSHMEM): attach the waitcnts to their respective loads

amirsojoodi:
# Motivation

`test_typed_int_put_get_wave` deterministically returns corrupted GET data (payload shifted by 8 bytes, garbage in leading elements). This bug showed up after ROCm 7.13.

## Root cause

AsmAccess<N>::load() (src/assembly.hpp) issues its async vector-memory load (flat_load_dwordx4/etc.) and the matching s_waitcnt drain as two separate, unrelated asm volatile blocks. LLVM doesn't recognize a raw inline-asm load as a vector-memory op, so in the new compiler the scheduler just schedules some other load isntructions that consumes the loaded value before the wait, causing the waitcnt to hit the empty bucket and return immediately. I confirmed at the disassembly level. Only the raw-asm scalar load path (copy_bulk tail loop, copy_remainder) was affected; the buffered intrinsic-based path was not.

## Alternatives considered

Compiler atomic builtins were considered too. Currently, there is no way to express the existing NonTemporal/FlatCache hints together in the same instruction except for the assembly solution. So, no.

## Fix

Move the `s_waitcnt/s_wait_loadcnt` into the same asm volatile block as the load, for every non-Standard-policy branch across all AsmAccess types and architectures . The pattern is already the same and it was meant to be like that. `store()` is untouched by design: raw `put/get` carries no ordering guarantee; only fence()/quiet()/wait_until() do. Remove the now-redundant standalone wait calls in util.hpp.

## Testing

Disassembly-confirmed fix; isolated primitive reproduced pass; MPI functional test passed. 

## Issue Tracking

JIRA ID: AIROCSHMEM-440

### DIFF
--- a/projects/rocshmem/src/assembly.hpp
+++ b/projects/rocshmem/src/assembly.hpp
@@ -787,31 +787,42 @@ struct AsmAccess<16, LoadPolicy, StorePolicy> {
       type val{};
 #if defined(__gfx942__) || defined(__gfx950__)
       if constexpr (LoadPolicy == CachePolicy::FlatCache) {
-        asm volatile("flat_load_dwordx4 %0, %1" : "=v"(val) : "v"(src) : "memory");
+        asm volatile("flat_load_dwordx4 %0, %1\n"
+                     "s_waitcnt vmcnt(0)" : "=&v"(val) : "v"(src) : "memory");
       } else if constexpr (LoadPolicy == CachePolicy::DeviceScope) {
-        asm volatile("flat_load_dwordx4 %0, %1, sc0" : "=v"(val) : "v"(src) : "memory");
+        asm volatile("flat_load_dwordx4 %0, %1, sc0\n"
+                     "s_waitcnt vmcnt(0)" : "=&v"(val) : "v"(src) : "memory");
       } else if constexpr (LoadPolicy == CachePolicy::NonTemporal) {
-        asm volatile("flat_load_dwordx4 %0, %1, nt" : "=v"(val) : "v"(src) : "memory");
+        asm volatile("flat_load_dwordx4 %0, %1, nt\n"
+                     "s_waitcnt vmcnt(0)" : "=&v"(val) : "v"(src) : "memory");
       } else if constexpr (LoadPolicy == CachePolicy::SystemScope) {
-        asm volatile("flat_load_dwordx4 %0, %1, sc0 sc1" : "=v"(val) : "v"(src) : "memory");
+        asm volatile("flat_load_dwordx4 %0, %1, sc0 sc1\n"
+                     "s_waitcnt vmcnt(0)" : "=&v"(val) : "v"(src) : "memory");
       } else if constexpr (LoadPolicy == CachePolicy::SystemScopeNT) {
-        asm volatile("flat_load_dwordx4 %0, %1, sc0 sc1 nt" : "=v"(val) : "v"(src) : "memory");
+        asm volatile("flat_load_dwordx4 %0, %1, sc0 sc1 nt\n"
+                     "s_waitcnt vmcnt(0)" : "=&v"(val) : "v"(src) : "memory");
       }
 #elif defined(__gfx90a__) || defined(__gfx1100__)
       if constexpr (LoadPolicy == CachePolicy::FlatCache) {
-        asm volatile("flat_load_dwordx4 %0, %1" : "=v"(val) : "v"(src) : "memory");
+        asm volatile("flat_load_dwordx4 %0, %1\n"
+                     "s_waitcnt vmcnt(0)" : "=&v"(val) : "v"(src) : "memory");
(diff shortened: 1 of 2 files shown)


edgargabriel (approved):

projects/rocshmem/src/assembly.hpp
```diff
@@ -787,31 +787,42 @@ struct AsmAccess<16, LoadPolicy, StorePolicy> {
       type val{};
 #if defined(__gfx942__) || defined(__gfx950__)
       if constexpr (LoadPolicy == CachePolicy::FlatCache) {
-        asm volatile("flat_load_dwordx4 %0, %1" : "=v"(val) : "v"(src) : "memory");
+        asm volatile("flat_load_dwordx4 %0, %1\n"
+                     "s_waitcnt vmcnt(0)" : "=&v"(val) : "v"(src) : "memory");
```
omor1, line 791: General note: I think that instead of a full generic `"memory"` clobber (which indicates arbitrary memory read/write), we can do something more targeted:
```C++
asm volatile ("flat_load_dwordx4 %0, %1\n"
              "s_waitcnt vmcnt(0)"
              : "=v" (val)
              :  "v" (src), "m" (*reinterpret_cast<const char (*)[16]>(src)));
// or         :  "v" (src), "m" (*reinterpret_cast<const __uint128_t*>(src)));
```

This informs the compiler that the 16 bytes from dereferencing `src` will be an input operand to the inline assembly, even though the assembly template doesn't directly use the operand. See the discussion in the [GCC documentation](https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html#Clobbers-and-Scratch-Registers-1) for more details.
(The output argument also doesn't need to be marked as an "earlyclobber" via `&`.)

I haven't taken a look at the definition of `store`, but I think the same thing applies there, just with `"=m" (*reinterpret_cast<...>(dst))` as an output operand. Since that changes the number of operands, the template might need to be change `%1` to `%2` as well to ensure that it still refers to the dst pointer input operand—maybe it makes sense to use the "asmSymbolicName" feature? E.g. `flat_load_dwordx4 %[val], %[src]` in the template and `[val] "=v" (val)` in the output operand, etc.

projects/rocshmem/src/assembly.hpp
```diff
@@ -787,31 +787,42 @@ struct AsmAccess<16, LoadPolicy, StorePolicy> {
...
       } else if constexpr (LoadPolicy == CachePolicy::DeviceScope) {
-        asm volatile("flat_load_dwordx4 %0, %1, sc0" : "=v"(val) : "v"(src) : "memory");
+        asm volatile("flat_load_dwordx4 %0, %1, sc0\n"
+                     "s_waitcnt vmcnt(0)" : "=&v"(val) : "v"(src) : "memory");
```
omor1, line 794: Don't these all need to be `s_waitcnt vmcnt(0) lgkmcnt(0)`? If we're using `flat_load` it's because we don't know whether it's coming from global memory or LDS.

If we _do_ know it's global memory, then we can do `global_load` and `s_waitcnt vmcnt(0)`

omor1 (approved):
looks alright, but please take a look at my comments—one is a potential perf improvement and the other is a correctness issue

projects/rocshmem/src/assembly.hpp
amirsojoodi: Not in order:

The reason I used `clobber` for all of the asm blocks was because:
> Using the "memory" clobber effectively forms a read/write memory barrier for the compiler. 

I didn't want the compiler to move my instructions around, and this really helped. For Loads specifically, I should study this further with a follow up work, but all in all I think it is still required.

Regarding the "m" method: I tried it but I get compilation error. It turns out it's a CPU trick and not an AMDGPU thing. See [here](https://github.com/llvm/llvm-project/blob/c4962f5721a36e198234019b846d3788738f766a/llvm/lib/Target/AMDGPU/SIISelLowering.cpp#L19514)

projects/rocshmem/src/assembly.hpp
amirsojoodi: That's correct. I missed it here, but I had them before:

```cpp
if constexpr (LP != CachePolicy::Standard) {
      wait_on_vmem_and_lds(0);
}
```

amirsojoodi: Actually, I thought about this further and I think it doesn't need lgkcnt. I know that I was doing it before, but I think I was doing it wrongly. 
 
The reason we still need flat instructions is that we are passing both global memory and scratch memory addresses to these instructions. That's why my original "flat to global" [PR](https://github.com/ROCm/rocm-systems/pull/5000) failed 
 
So, because we are only passing global and scratch memory addresses to memcpy variants then it is safe to wait only on vmcnt. This may require further discussion and/or analysis. For now, I will merge the PR

PR #8722, opened 2026-07-16
feat(ci): add skip-therockci label support to skip CI

geomin12:
Add support for the `skip-therockci` label to skip TheRock CI on PRs. When this label is applied to a pull request, both Linux and Windows CI jobs will be skipped.

This matches the behavior in rocm-libraries.


lumachad:
Thanks for the PR. Looks good to me, but if I can make a suggestion, could we use the same label as TheRock? That's "ci:skip".

lumachad (approved):
Apart from the ci:skip suggestion, LGTM.

PR #8723, opened 2026-07-16
refactor(rocrtst): remove hwloc dependency from memory async copy tests

dayatsin-amd:
## Summary

- Drop the mandatory **hwloc** build/link dependency from rocrtst by using `HSA_AGENT_INFO_NODE` for NUMA topology in `Memory_Async_Copy` instead of hwloc PCIe lookup.
- Remove vendored hwloc headers/libs and the `hwloc` CMake discovery block.
- Switch the optional NUMA async-copy test (`ENABLE_COPY_NUMA`) to **libnuma** APIs when enabled.
- Add a simple-topology fallback when no GPU shares a CPU NUMA node (restores prior single-node behavior and avoids the old hwloc/OpenCL shutdown hack).

## JIRA ID

JIRA ID : ROCM-1282

## Test plan

- [x] Build `rocrtst64` on **dayatsin-test-01** (gfx1100, `/opt/rocm`) without hwloc
- [x] Build `rocrtst64` on **e3-2b** (gfx950, TheRock toolchain) without hwloc
- [x] Run `rocrtstPerf.Memory_Async_Copy` and `rocrtstPerf.Memory_Async_Copy_On_Engine` on both systems — 2/2 passed on each host


cfreeamd (commented):

Reviewed the ROCr-scoped changes (rocrtst memory_async_copy). The hwloc removal, CMake cleanup, and dropping the old hwloc `hsa_shut_down()`/`hsa_init()` teardown hack in `Close()` all look correct. One correctness concern on the NUMA local/remote classification left inline.

projects/rocr-runtime/rocrtst/suites/performance/memory_async_copy.cc
```diff
@@ -712,38 +695,25 @@ static hsa_status_t GetGPUAgents(hsa_agent_t agent, void* data) {
...
-    printf("gpu nodeset: %s\n", s1);
-    printf("cpu nodeset: %s\n", s2);
-    if (!hwloc_bitmap_isequal(gpu_numa_node->nodeset,
-                                              ptr->cpu_hwl_numa_nodeset())) {
+  if (gpu_numa_node_id != MemoryAsyncCopy::kUnknownNumaNode) {
+    if (ptr->verbosity() >= MemoryAsyncCopy::VERBOSE_STANDARD) {
+      printf("gpu numa node: %u\n", gpu_numa_node_id);
+      printf("cpu numa node: %u\n", ptr->cpu_numa_node_id());
+    }
+    if (gpu_numa_node_id != ptr->cpu_numa_node_id()) {
```
cfreeamd, line 716: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** This NUMA-locality comparison looks broken. Both `gpu_numa_node_id` and `cpu_numa_node_id()` come from `HSA_AGENT_INFO_NODE`, which in ROCr returns the agent's KFD/HSA topology `node_id()` (see `amd_gpu_agent.cpp` / `amd_cpu_agent.cpp`, both carrying the TODO "associate with OS NUMA support"). Every agent has a unique KFD node id (`runtime.cpp`: `agents_by_node_[agent->node_id()]`), so a CPU agent and a GPU agent never share a node id — `gpu_numa_node_id != cpu_numa_node_id()` is therefore always true.

Consequences: the "local GPU" branch is never taken, `gpu_local_agent1_` is never set here, and `FindTopology()` always falls through to the `GetSimpleAgentInfo` fallback, which assigns local1/local2/remote purely by iteration order with no NUMA basis. The benchmark's local-vs-remote (H2D vs H2DRemote) distinction is silently lost. The prior code compared the GPU's PCI-ancestor NUMA node (hwloc) against the CPU's NUMA node, which was a real locality test.

Suggest either deriving true NUMA locality (e.g. libnuma + `/sys/bus/pci/devices/.../numa_node`, or KFD `io_links` proximity), or — if dropping NUMA awareness is intentional — removing the dead local/remote path so the code isn't misleading.

projects/rocr-runtime/rocrtst/config/platform_config.yaml
```diff
@@ -188,7 +188,7 @@ platforms:
     blocked_tests:
       - rocrtstFunc.Memory_Max_Mem
       - rocrtstFunc.Memory_Available  # Uses max & available mem; not accurate
-      - rocrtstPerf.Memory_Async_Copy # Requires hwloc support
+      - rocrtstPerf.Memory_Async_Copy # NUMA topology test; may be skipped on single-node systems
```
cfreeamd, line 191: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** `Memory_Async_Copy` stays in `blocked_tests`; only the comment changed. If the new simple-topology fallback is meant to let this run on single-node systems, is keeping it blocked here intentional, or should it now be unblocked on this platform?

PR #8728, opened 2026-07-16
test(rccl): add test for sparse-asymmetric sendRecv

isaki001:
## Motivation

NCCL 2.29.7 fixed a hang in scheduleP2pTasksToPlan() triggered by repeated sparse P2P patterns (AICOMRCCL-1112). The bug caused p2pEpoch to be incremented at the top of the scheduling loop before any rounds were processed — if the kernel plan budget ran out mid-round, the next plan call would increment the epoch again before resuming, creating an epoch mismatch between paired send/recv ranks and a deadlock.

RCCL was never susceptible to the intermediate bug state (it merged 2.29.2+2.29.3+2.29.7 in one shot), but no regression test existed to guard against future regressions in this code path.

## Technical Details

Adds `test/test_sparse_send_recv.cpp`, a new rccl-UnitTestsMPI regression test, and wires it into the MI300X test runner config.

Test design — star topology with budget pressure:

  - Rank 0 (hub) sends to and receives from every other rank, issuing REPS * (nRanks-1) operations per group call. Leaf ranks each issue REPS operations to rank 0 only.
  - With NCCL_WORK_ARGS_BYTES=512 and NCCL_WORK_FIFO_BYTES=512, the ~7-round plan budget is exhausted mid-group on rank 0, forcing a second kernel plan with an incremented epoch. 
  - Leaf ranks fit entirely within the first plan (epoch=1). This is the epoch mismatch that triggers the original hang.
  - NCCL_MIN_NCHANNELS=4 / NCCL_MAX_NCHANNELS=4 fix the channel count for deterministic plan splitting.
  - 5 outer iterations (ITERS=5) allow corrupted resources->step state to accumulate across group calls.
  - The test skips itself if the required environment variables are not set.

  Asymmetric MPI error handling:

  Because rank 0 and leaf ranks follow structurally different code paths, ASSERT_MPI_EQ (which requires all ranks to call the same check at the same execution point) is not used inside the group body — doing so causes the underlying MPI_Allreduce to hang on asymmetric paths. EXPECT_EQ is used for NCCL calls within the group, with ASSERT_MPI_EQ used only at symmetric synchronization points (ncclGroupEnd, hipStreamSynchronize).

Also includes a compile-time guard for RegistrationMPITests.cpp to handle the unavailability of hipMemLocationTypeHost before ROCm 7.12.

## Issue Tracking

JIRA ID AICOMRCCL-1112

## Test Plan

2-node execution on MI300 via python3 test_runner.py with mi300x_mellanox_ib.json.

## Test Result

Test passes on MI300. 

## Submission Checklist


projects/rccl/test/test_sparse_send_recv.cpp
```diff
@@ -0,0 +1,201 @@
...
+        }
+    } else {
+        // Spoke: send to and receive from rank 0 only.
+        ASSERT_EQ(ncclSuccess,
+                      ncclSend(send_buf, NUM_ELEMS, ncclFloat, 0, comm, stream));
+        ASSERT_EQ(ncclSuccess,
+                      ncclRecv(recv_buf, NUM_ELEMS, ncclFloat, 0, comm, stream));
+    }
+    ASSERT_EQ(ncclSuccess, ncclGroupEnd());
+    ASSERT_EQ(hipSuccess, hipStreamSynchronize(stream));
```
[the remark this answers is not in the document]
isaki001: this has been address and mpi assertions were added

projects/rccl/test/test_sparse_send_recv.cpp
```diff
@@ -0,0 +1,258 @@
...
+// ranks (REPS=4 gives rank 0 60 rounds, enough to exhaust the ~7-round plan budget).
+constexpr int    MIN_RANKS           = 16;
+// Repetitions of the full star pattern within ONE ncclGroupStart/End call.
+// Creates the asymmetric round count: rank 0 gets REPS*(nRanks-1) rounds,
+// each leaf gets REPS rounds. This is what forces the epoch mismatch.
+constexpr int    REPS                = 4;
+// Outer iterations. Corrupted resources->step accumulates across group calls;
+// 5 iterations is enough to make the hang definitive rather than flaky.
+constexpr int    ITERS               = 5;
+constexpr size_t NUM_ELEMS           = 1024;
```
alex-breslow-amd, line 68: Thoughts on making these compile-time tunable with gflags?

alex-breslow-amd (approved):
LGTM

projects/rccl/test/test_sparse_send_recv.cpp
isaki001: I had not considered it. 
I see the value the value in having that flexibility, but I do not anticipate these needing changes often or having different values among runners for different architectures.
Do you think I should make adjustments?

alex-breslow-amd (approved):
Still LGTM

mch (approved):

PedramAlizadeh (approved):

PR #8732, opened 2026-07-16
fix(rocjitsu): Inconsistent SCC read/write

lua1235:
## Motivation
  Correct scalar condition code (SCC) handling across ISA profiles. Several scalar instruction families
  either incorrectly modified SCC or failed to model SCC effects. This was exposed by failing rocRAND tests.

  ## Technical Details
  - Model implicit SCC reads and writes from the XMLs.
  - Preserve SCC for scalar unary operations that do not modify it.
  - Correct SAVEEXEC and WREXEC behavior, including:
      - Current and legacy mnemonic spellings.
      - B32 and B64 EXEC handling.
      - Boolean operand orientation.
      - SCC updates from the resulting EXEC value.
      - Aliased source and destination operands.
  - Reject unsupported or malformed SAVEEXEC/WREXEC semantics.
  - Correct ADDK signed-overflow SCC behavior and scalar maximum tie handling.
  - Centralize SCC effects for inline lowering operations.

  ## Issue Tracking

  N/A

  ## Test Plan
  
ctests, rocjitsu-test-corpus

  ## Test Result

all pass

  ## Submission Checklist


emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -2254,6 +2254,7 @@ def _op_sel_hi_2_expr(self, enc_name: str) -> str:
         {
             'vector_dot',
             'vector_swap',
+            'scalar_addk',
```
kuhar, line 2257: We should include `scalar_mulk` in `_READS_DST_CLASSES` and extend `test_addk_registers_its_read_write_destination_as_a_source` to cover it. `S_MULK_I32` reads and overwrites SDST, but the generated CDNA4 and gfx1250 constructors currently register only SIMM16 in `src_operands_`; RDNA4 is correct only because its XML already marks SDST as input/output. Leaving MULK out makes def-use and probe-clobber analysis profile-dependent for the same read-modify-write operand contract.

emulation/rocjitsu/lib/python/amdisa/tests/test_sema_derive.py
```diff
@@ -2121,18 +2249,82 @@ def test_cselect(self):
...
+                _FakeSem(
+                    'S_AND_NOT1_WREXEC_B64',
+                    'scalar_wrexec',
+                    'and_not1',
+                    'b64',
+                    'nonzero',
+                )
+            )
+        )
+        assert '(old_exec & (~src))' in exec_not_src
```
kuhar: It would be useful to extend `test_wrexec_operand_orientation` with `andn1` and `andn2`, since those are the mnemonics whose mapping changed in this patch. The test currently checks only the `and_not0`/`and_not1` aliases, so swapping the `andn1`/`andn2` branches in `_ScalarWrexec.derive` would still pass this unit test even though the correct results are `old_exec & ~src` and `src & ~old_exec`, respectively. The C++ execution test catches this end to end, but the derivation-layer test should pin the mapping it owns.

kuhar (commented):
A few comments

emulation/rocjitsu/tests/scalar_scc_test.cpp
```diff
@@ -0,0 +1,1047 @@
...
+  }};
+
+  const uint32_t sb = fixture.sgpr_base();
+  for (const auto &test_case : cases) {
+    const auto words = encode_sopk(profile, "s_mulk_i32", /*sdst=*/4, test_case.immediate);
+    auto inst = fixture.decode(words, "s_mulk_i32");
+    ASSERT_NE(inst, nullptr) << profile.name;
+
+    fixture.cu->write_sgpr(sb + 4, test_case.initial);
+    fixture.cu->execute_instruction(inst.get(), *fixture.wf);
```
kuhar: It would be useful to assert SCC preservation in `run_mulk_cases`. The loop checks only SDST, so `MulkUsesOldDestinationAndSignedImmediate` would still pass if a future `S_MULK_I32` path started writing SCC; unlike the ADDK and unary cases, SCC is neither seeded nor consumed here. Please seed both SCC values and verify each survives, or use `expect_scc_consumer`, so this test covers the preservation contract.

kuhar (approved):
LGTM, just one suggestion

Issue #8714, opened 2026-07-16
feat(rocjitsu): rj_daemon_start


amd-arosa:
make tests cleaner, and remove a lot of unsafe code from mirage

PR #8733, opened 2026-07-17
feat(rocjitsu): rj_daemon_start

amd-arosa:
## Motivation

make tests cleaner, and remove a lot of unsafe code from mirage

## Technical Details

Added the public C lifecycle API:
rj_daemon_start()
rj_daemon_stop()
rj_daemon_status()
Opaque rj_daemon_t handle and observable lifecycle states

## Issue Tracking

GitHub issue: https://github.com/ROCm/rocm-systems/issues/8714

## Test Plan

1. Build all configured RocJitsu targets.
2. Run the direct daemon API suite repeatedly to exercise explicit and RAII teardown.
3. Run daemon CLI, interposer, debugger descriptor, HIP, and RCCL integration tests.
4. Run affected Mirage and rocjitsu_sys tests.
5. Run formatting, pre-commit, diagnostics, and whitespace checks.

## Test Result

1. Full RocJitsu build passed.
2. All 8 DaemonApi tests passed across three repetitions.
3. All 22 relevant daemon, CLI, debugger, HIP, interposer, and RCCL integration tests passed.
4. All 20 affected Mirage and rocjitsu_sys tests passed.
5. Pre-commit, formatting, diagnostics, and git diff --check passed.

## Submission Checklist


kuhar (commented):
A few comments

emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/daemon/rj_daemon.h
```diff
@@ -0,0 +1,78 @@
...
+#define ROCJITSU_DAEMON_RJ_DAEMON_H_
+
+#include "rocjitsu/base/rj_compiler.h"
+#include "rocjitsu/base/rj_status.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/// @addtogroup daemon
```
kuhar, line 19: We should expose the new daemon API through the existing umbrella-header layout. The other API families define `<family>/api.h` and include it from `rocjitsu/rocjitsu.h`; without `rocjitsu/daemon/api.h`, umbrella-header consumers cannot see `rj_daemon_*`, and `@addtogroup daemon` has no group definition.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/rj_daemon.cpp
```diff
@@ -0,0 +1,608 @@
...
+constexpr uint32_t kMaxPayloadBytes = 16 * 1024 * 1024;
+
+bool checked_product(size_t lhs, size_t rhs, size_t *product) {
+  if (lhs != 0 && rhs > std::numeric_limits<size_t>::max() / lhs)
+    return false;
+  *product = lhs * rhs;
+  return true;
+}
+
+bool validate_ioctl_payload(uint32_t command, const void *buffer, size_t buffer_size) {
```
kuhar, line 61: It would be useful to add GPU-free RPC tests that reach the new IOCTL validation, rather than only the outer `args_bytes` mismatch. The existing malformed-message test returns before `validate_ioctl_payload`. A small set covering a valid `AMDKFD_IOC_GET_VERSION` call plus a `WAIT_EVENTS` payload whose inline array overflows or is one byte short would exercise both the happy path and the security-relevant bounds checks; malformed cases should also assert that the connection is dropped while the daemon remains `RJ_DAEMON_STATUS_RUNNING`.

emulation/mirage/rocjitsu_sys/src/lib.rs
```diff
@@ -50,6 +50,23 @@ pub struct RjVm {
...
 }
 
+/// Opaque daemon handle (`rj_daemon_t`). Only ever held behind a pointer.
+#[repr(C)]
+pub struct RjDaemon {
+    _private: [u8; 0],
+}
+
+/// Observable daemon lifecycle state (`rj_daemon_status_t`).
+#[repr(i32)]
```
kuhar, line 60: We should keep the raw `rj_daemon_status` FFI return type as `c_int` and convert it to a Rust status type after checking the value. Because this library is loaded dynamically, a newer C library can return an unknown lifecycle state; receiving that directly as a `repr(i32)` enum creates an invalid Rust value before the wrapper can handle it.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/rj_daemon.cpp
```diff
@@ -0,0 +1,608 @@
...
+bool remove_stale_socket(const rj_daemon_t *daemon, const sockaddr_un &address,
+                         socklen_t address_length) {
+  struct stat before {};
+  if (::lstat(daemon->socket_path.c_str(), &before) != 0 || !S_ISSOCK(before.st_mode))
+    return false;
+
+  UniqueFd probe(::socket(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0));
+  if (probe.get() < 0)
+    return false;
+  if (::connect(probe.get(), reinterpret_cast<const sockaddr *>(&address), address_length) == 0)
```
kuhar, line 516: This liveness probe needs to be nonblocking or have a bounded timeout. A blocking AF_UNIX `connect()` can wait indefinitely when a live listener accept queue is full: with a backlog-1 listener and two queued clients, a third connection was still blocked after one second. `EAGAIN`, `EINPROGRESS`, and timeout should preserve the existing endpoint and make `rj_daemon_start()` return an error instead of hanging.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/fd_utils.h
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: MIT
+
+/// @file fd_utils.h
+/// @brief Utilities for owning Linux file descriptors.
+
+#ifndef ROCJITSU_KMD_LINUX_FD_UTILS_H_
```
atgutier: I wonder if this should go in the utils header-only lib and generalized to work with Windows? Or is that not possible?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/rj_daemon.cpp
```diff
@@ -0,0 +1,608 @@
...
+        response.request_id = header.request_id;
+        if (!rpc_send_exact(client_fd, &response, sizeof(response)))
+          goto done;
+        break;
+      }
+
+      case RPC_IOCTL: {
+        if (process_id == 0 || header.payload_bytes > kMaxPayloadBytes ||
+            header.payload_bytes < sizeof(RpcIoctlRequest)) {
+          goto done;
```
atgutier: This is a very C-style Linux-driver specific sort of idiom and generally an anti-pattern in C++. Please get rid of the gotos in favor of modern C++ style handling.

atgutier (commented):
Don't forget to bump the version and changelog for this non-breaking change.

Also have some agent comments:

 BLOCKERS

  1. Use-after-free between rj_daemon_status and rj_daemon_stop — rj_daemon.cpp
  rj_daemon_stop() does teardown(daemon); delete daemon;, but rj_daemon_status() dereferences the same
   handle: daemon->status.load(...). Any caller that races status against stop (or calls status after
  stop) reads freed memory. This defeats the entire purpose of the safe-lifecycle API the PR
  introduces.
  Fix: separate stop (halt threads, close sockets, transition to a terminal state) from destroy (free
  the handle). Either (a) keep the handle alive after stop and require an explicit
  rj_daemon_destroy(), or (b) ref-count the handle so status holds a reference for the duration of the
   load. Option (a) is simpler and matches the C-ABI ownership model already used elsewhere.

  2. rj_daemon_status returns a #[repr(i32)] Rust enum by value across FFI — rocjitsu_sys/src/lib.rs
  type FnDaemonStatus = unsafe extern "C" fn(*const RjDaemon) -> RjDaemonStatus; where RjDaemonStatus
  is #[repr(i32)]. Any C-side return value outside the declared discriminants (0..=4) is instant UB —
  the Rust enum has no representation for it.
  Fix: declare the FFI return as c_int, then map explicitly:
  let raw: c_int = (self.status_fn)(self.handle);
  RjDaemonStatus::from_raw(raw) // -> #[non_exhaustive] enum with Unknown(c_int) catch-all
  Make RjDaemonStatus #[non_exhaustive] with an Unknown(c_int) variant so an unrecognized value is
  representable, not UB.

  ---
  MAJORS

  3. rj_daemon_status_t has no fixed underlying type — rj_daemon.h
  The C enum relies on the compiler's default underlying type, which is not guaranteed to be int-width
   across toolchains. This is the C side of BLOCKER #2. Fix: pin it — typedef enum rj_daemon_status_t
  : int32_t { ... } rj_daemon_status_t; (or the project's C-enum-width idiom) so the ABI width the
  Rust c_int reads is guaranteed.

  4. Readiness handshake can flip RUNNING→ERROR before the ready-write — rj_daemon.cpp
  The status can be observed as RUNNING and then transition to ERROR before the readiness write
  completes, so a client that polls status instead of waiting on the handshake can see a
  stale-optimistic state. Fix: don't publish RUNNING until after the readiness barrier; keep the
  pre-ready state distinct (e.g. STARTING).

  5. Missing [[nodiscard]] on rj_daemon_start / rj_daemon_status — rj_daemon.h
  Both return status the caller must check; dropping the return silently loses failures. Fix: annotate
   both with [[nodiscard]].

  6. Tighten the unsafe impl Send justification — daemon.rs
  The unsafe impl Send needs a comment establishing why the handle is safe to move across threads (the
   C daemon owns its own synchronization; the handle is just an opaque pointer with no
  thread-affinity). Without the justification the invariant can silently break if the C side later
  gains thread-local state.

7. Document the synchronous-until-ready dependency — daemon.rs
  The Rust wrapper depends on rj_daemon_start blocking until the daemon is ready. That contract lives
  only in the C implementation; if it ever becomes async the Rust side races. Fix: document the
  dependency at the FFI boundary so the coupling is explicit.

emulation/mirage/rocjitsu_sys/src/lib.rs
amd-arosa: done

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/rj_daemon.cpp
amd-arosa: fe2457a2732680b0590241ccf2525bf1d65ef75f

emulation/mirage/rocjitsu_sys/src/lib.rs
amd-arosa: ab79610bdfba6417a97f271caa736366137d9757

emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/daemon/rj_daemon.h
amd-arosa: c7249bbfea0aad68409e37e94de572a05838f96f

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/rj_daemon.cpp
amd-arosa: 8ad904adf848cb56a9bde9c7b2669a5cb5df4ec8

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/fd_utils.h
amd-arosa: a8fb45353308aeba6d1c52b7ab4818e5ec83b243

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/rj_daemon.cpp
amd-arosa: 3191bd6

emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/daemon/rj_daemon.h
```diff
@@ -0,0 +1,83 @@
...
+/// @{
+
+/// @brief Opaque handle for a running daemon.
+typedef struct rj_daemon_t rj_daemon_t;
+
+/// @brief Observable daemon lifecycle state.
+typedef enum rj_daemon_status_t : int32_t {
```
kuhar: We should keep this public C header compatible with the C versions supported by the existing API. The fixed-underlying enum syntax at line 26 is only standard in C23; a direct include fails with `cc -std=c11 -pedantic-errors`. Please use a C-compatible representation, such as an `int32_t` typedef plus named constants, while keeping the exported return type fixed-width.

kuhar (commented):
Just one comment

emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/daemon/rj_daemon.h
amd-arosa: e090bd42d1d8478c76d15e776b449f5b9a93e9ed

amd-arosa:
fixing RocjitsuCliDaemon.LaunchesApplicationAfterDaemonIsReady

emulation/mirage/core/src/ctl.rs
```diff
@@ -424,6 +424,14 @@ impl MirageCtl for FileCtl {
                     nix::unistd::Pid::from_raw(pid),
                     nix::sys::signal::Signal::SIGKILL,
                 );
+                // Do not return while the host can still be tearing down
+                // emulator-owned files under the launcher's runtime directory.
+                for _ in 0..50 {
```
kuhar, line 429: This would be safer if the daemon used a child-reaping barrier rather than `process_alive()` here. `spawn_host_for()` drops the direct child handle, and the long-lived `mirage_daemon` path has no `waitpid`/SIGCHLD reaper. An exited host can therefore remain a zombie for which `kill(pid, 0)` still succeeds, so daemon-backed session deletion waits through both polling loops and never establishes the teardown guarantee in the comment. Please retain and reap the child, add a central reaper with waitable completion, or synchronize on an explicit shutdown signal; a daemon API regression should verify prompt deletion with no unreaped host.

kuhar (commented):
A few comments

emulation/rocjitsu/tests/util_unique_handle_test.cpp
```diff
@@ -0,0 +1,105 @@
...
+};
+
+using CountingHandle = util::BasicUniqueHandle<CountingHandleTraits>;
+
+static_assert(!std::is_copy_constructible_v<CountingHandle>);
+static_assert(!std::is_copy_assignable_v<CountingHandle>);
+static_assert(std::is_nothrow_move_constructible_v<CountingHandle>);
+static_assert(std::is_nothrow_move_assignable_v<CountingHandle>);
+
+TEST(UniqueHandle, MovesAndReleasesOwnership) {
```
kuhar, line 40: It would be useful to add a move-assignment test for a live `BasicUniqueHandle`. The current four tests cover move construction, reset, destruction, and native close, but none exercises `operator=(BasicUniqueHandle&&)`, where `reset(other.release())` must close the target handle before adopting the source. Please add a `CountingHandle` case that verifies the replaced handle is closed, the source is empty, and the target owns the incoming handle.

amd-arosa: done 22678d7

emulation/mirage/core/src/ctl.rs
amd-arosa: mirage is designed to have ANY number of daemons running.
0, 1, 2. whatever.
host are independent if that makes sense.
their state is mostly communicated over the filesystem, and checking for pid/cid.
hosts are supposed to be responsible for cleaning up themselves.

session_destroy is designed to always destroy the target, even if it wasn't shutdown properly / isn't responding. this was more important previously when rj ran in main thread of host.

amd-arosa: so I MAY be starting a zombie horde.
will fix this in followup pr #9048

emulation/rocjitsu/tests/hip_vector_add_test.cpp
```diff
@@ -14,6 +14,11 @@
 
 #include <gtest/gtest.h>
 
+// ROCR keeps process-lifetime runtime state allocated after hipDeviceReset().
+// Ignore only allocations whose stack includes the external HSA runtime while
+// retaining LeakSanitizer coverage for rocjitsu and this test executable.
+extern "C" const char *__lsan_default_suppressions() { return "leak:libhsa-runtime64.so\n"; }
```
kuhar: This would be safer if the suppression matched the specific ROCR allocation sites that are intentionally process-lifetime. `leak:libhsa-runtime64.so` suppresses any leak whose allocation stack contains a frame from that library, including leaks in rocjitsu code reached through an HSA callback or interposed path. Because that is the main path exercised by this test, the module-wide rule can hide the regressions this test is supposed to catch. Could we derive narrower symbol-level suppressions from the observed ROCR leak stacks?

kuhar (approved):
LGTM overall, just one new comment + one reply

emulation/mirage/core/src/ctl.rs
kuhar: The risk here is that this wait still cannot observe host exit under `mirage_daemon`: `spawn_host_for()` drops the child handle, so an exited host remains a zombie until its long-lived parent reaps it, and `kill(pid, 0)` stays successful (I reproduced that state locally). Please either include the reaping or waitable-completion fix here, or defer this loop and its teardown guarantee to #9048.

emulation/rocjitsu/tests/hip_vector_add_test.cpp
amd-arosa: okay done

amd-arosa: undone

PR #8735, opened 2026-07-17
fix(clr): Report all kernel timestamps in graph profiling trace

saleelk:
The per-segment kernel names were set via setKernelNamesRef() which overwrites on each segment dispatch, leaving only the last segment's names for ReportActivity. Switch to addKernelNames() (guarded by profilingInfo().enabled_) so all segments' kernel names accumulate and every kernel dispatch gets its timestamp reported in the trace.

Also propagate per-signal queue_index through ProfilingSignal and AccumulateCommand so graph kernels land on correct profiling lanes.

## Issue Tracking

JIRA ID: AIRUNTIM-0

## Submission Checklist


saleelk:
> 
> 
> * ❌ Unit Test
> 

No unit test required

chrispaquot (approved):

PR #8737, opened 2026-07-17
feat(hotswap): share retarget buffers across loads instead of copying

suryajasper:
## Summary

The hotswap retarget cache already stores each rewritten code object as a
`shared_ptr<vector<uint8_t>>`, but it handed every load a **private
`malloc`+`memcpy` copy** through a `unique_ptr` (`OwnedElfBuffer`). In a
multi-GPU job, every agent that loads the same code object paid a full copy of
the retargeted ELF — hundreds of MB for large device libraries such as RCCL —
and the buffer's lifetime was duplicated per executable.

This change makes consumers **share the cache's buffer by reference** instead of
copying it:

- Replaces `OwnedElfBuffer` (`unique_ptr<void, &std::free>`) with
  `ElfBufferRef` (`shared_ptr<vector<uint8_t>>`) throughout the hotswap module.
- On a cache hit, the loader is handed the cache's own buffer (refcount++), not
  a fresh copy. The single retarget output is shared by the in-memory cache and
  every executable that loads it, and is freed once the last holder releases it.
- `RetargetCodeObject` now returns the buffer (`nullptr` on failure) with the
  size carried by the buffer itself, replacing the bool + out-param API.

The only remaining copy is the unavoidable copy-out of COMGR's output across the
COMGR C ABI on a cold miss.

## Why this is safe

The loader consumes the retargeted ELF **read-only**: it copies loadable
segments out into its own allocated memory and applies relocations there, never
mutating the source ELF image (`ExecutableImpl::LoadSegmentsV2` copies via
`Segment::Copy`; relocations write into the loaded segment). `InitAsBuffer`
aliases the bytes rather than editing them. Sharing one immutable buffer across
concurrent loads therefore cannot corrupt a shared cache entry.

The per-executable keepalive (needed because the loader holds a raw pointer into
the buffer for the executable's lifetime) is preserved; its OOM path is reworked
from a raw-pointer leak into a bounded global fallback with a last-resort
reference leak.

Note: there is no single-flight, so concurrent cold misses on the same key may
each run COMGR and race to store (first-wins); the loser still returns its own
valid buffer, so a single shared buffer is only guaranteed once an entry is
committed. This matches prior behavior.

## Relationship to other work

Builds on the in-memory retarget cache from #8274 (merged). Independent of the
two-tier disk cache in #8659.

## Test plan

- [x] New unit test `HotswapRewrite.CachedLoadsShareOneBufferNoCopy`: two
  cache-served loads of the same code object receive the **same** buffer pointer
  (proves zero-copy sharing) and both executables retain a reference to it.
- [x] Adapted existing `HotswapRewrite.*` tests to the new return-by-value API.
- [x] Registered the new test across all rocrtst tiers in `test_categories.yaml`.
- [x] Builds clean and full `hotswap_rewrite` suite passes on MI300X (32/32),
  including the new test.

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/inc/hotswap.hpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/inc/hotswap.hpp
@@ -44,10 +44,11 @@
 #define HSA_RUNTIME_CORE_INC_HOTSWAP_HPP_
 
 #include <cstddef>
-#include <cstdlib>
+#include <cstdint>
 #include <memory>
 #include <optional>
 #include <string>
+#include <vector>
 
 #include "core/inc/amd_hsa_loader.hpp"
 #include "inc/hsa.h"
@@ -57,7 +58,14 @@ namespace hotswap {
 
 struct AgentGfxRevision;
 
-using OwnedElfBuffer = std::unique_ptr<void, decltype(&std::free)>;
+// Retargeted code objects are shared, not uniquely owned: a single retarget
+// result is handed to every agent that loads the same code object (e.g. all
+// GPUs in a multi-GPU job) and is simultaneously held by the in-memory retarget
+// cache, so consumers reference-count one buffer instead of each holding a
+// private copy. The loader consumes the ELF read-only (it copies segments out
+// into its own memory and applies relocations there), so sharing one immutable
+// buffer across concurrent loads is safe.
+using ElfBufferRef = std::shared_ptr<std::vector<uint8_t>>;
 
 struct CodeObjectView {
   const void* data = nullptr;
@@ -110,19 +118,20 @@ struct LoadAgentCodeObjectCallbacks {
 
 std::string GetCodeObjectIsaName(const void* elf_data, size_t elf_size);
 
-bool RetargetCodeObject(const void* elf_data, size_t elf_size,
-                        const char* source_isa, const char* target_isa,
-                        OwnedElfBuffer* out_elf_buffer, size_t* out_elf_size,
-                        bool request_entry_trampolines = false,
-                        bool request_strict_mode = false);
+// Returns the retargeted ELF as a shared buffer, or nullptr on failure. The
+// size is carried by the buffer itself (ElfBufferRef->size()).
+ElfBufferRef RetargetCodeObject(const void* elf_data, size_t elf_size,
+                                const char* source_isa, const char* target_isa,
+                                bool request_entry_trampolines = false,
+                                bool request_strict_mode = false);
 
 RetargetCodeObjectResult TryRetargetCodeObject(
     const CodeObjectView& code_object, hsa_agent_t agent,
-    OwnedElfBuffer* out_elf_buffer, size_t* out_elf_size);
+    ElfBufferRef* out_elf_buffer);
 
 RetargetCodeObjectResult TryRetargetCodeObject(
     amd::hsa::loader::CodeObjectReaderImpl* reader, hsa_agent_t agent,
-    OwnedElfBuffer* out_elf_buffer, size_t* out_elf_size);
+    ElfBufferRef* out_elf_buffer);
 
 hsa_status_t LoadAgentCodeObjectWithHotswap(
     hsa_executable_t executable, hsa_agent_t agent,
@@ -131,7 +140,7 @@ hsa_status_t LoadAgentCodeObjectWithHotswap(
     const LoadAgentCodeObjectCallbacks& callbacks);
 
 void RetainRewrittenElfBuffer(hsa_executable_t executable,
-                              OwnedElfBuffer elf_buffer);
+                              ElfBufferRef elf_buffer);
 void ReleaseRetainedRewrittenElfBuffers(hsa_executable_t executable);
 
 #ifdef ROCR_HOTSWAP_TESTING
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
@@ -70,7 +70,15 @@ namespace hotswap {
 namespace {
 
 std::mutex g_retained_rewritten_elf_buffers_mutex;
-std::unordered_map<uint64_t, std::vector<OwnedElfBuffer>> g_retained_rewritten_elf_buffers;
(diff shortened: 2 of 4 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/inc/hotswap.hpp
```diff
@@ -57,7 +58,14 @@ namespace hotswap {
...
 
-using OwnedElfBuffer = std::unique_ptr<void, decltype(&std::free)>;
+// Retargeted code objects are shared, not uniquely owned: a single retarget
+// result is handed to every agent that loads the same code object (e.g. all
+// GPUs in a multi-GPU job) and is simultaneously held by the in-memory retarget
+// cache, so consumers reference-count one buffer instead of each holding a
+// private copy. The loader consumes the ELF read-only (it copies segments out
+// into its own memory and applies relocations there), so sharing one immutable
+// buffer across concurrent loads is safe.
+using ElfBufferRef = std::shared_ptr<std::vector<uint8_t>>;
```
cfreeamd, line 68: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** This is the load-bearing invariant for the whole change: zero-copy sharing is only correct if the loader treats the input ELF as strictly read-only. Previously every load received a private malloc'd copy, so in-place patching of the ELF would have been harmless; now concurrent loads of the same code object share one immutable buffer. If `load_rewritten_code_object` / `amd_hsa_loader` ever writes into the provided ELF bytes (in-place relocation, header fixups), concurrent multi-GPU loads would race and corrupt each other. The comment asserts the loader copies segments out and relocates in its own memory — please confirm that holds for all load paths before merge, since it's now a correctness requirement rather than just an optimization detail.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -710,14 +723,26 @@ hsa_status_t LoadAgentCodeObjectWithHotswap(hsa_executable_t executable, hsa_age
...
+    // pointer into this buffer for the executable's lifetime, so the reference
+    // must not drop here. Park it in a global fallback that outlives the
+    // executable rather than leaking it irrecoverably; if even that cannot
+    // grow, the reference is intentionally leaked as a last resort so the
+    // loaded code object's ELF pointer stays valid.
+    try {
+      std::scoped_lock lock(g_oom_retained_elf_buffers_mutex);
+      g_oom_retained_elf_buffers.push_back(std::move(elf_buffer));
+    } catch (const std::bad_alloc&) {
+      new ElfBufferRef(std::move(elf_buffer));  // deliberate last-resort leak
```
cfreeamd, line 744: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** `new ElfBufferRef(std::move(elf_buffer))` can itself throw `std::bad_alloc`. `RetainRewrittenElfBuffer` is called from `LoadAgentCodeObjectWithHotswap`, which returns `hsa_status_t` — an escaping exception here would propagate into C-ABI-facing runtime code. The outer try/catch guards the global-vector `push_back`, but this last-resort allocation is unguarded. Use `new (std::nothrow) ElfBufferRef(...)` so the keepalive fallback can never throw.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -607,25 +617,29 @@ RetargetCodeObjectResult TryRetargetCodeObject(const CodeObjectView& code_object
...
   if (retarget_attempted) {
     try {
       std::scoped_lock lock(g_retarget_cache_mutex);
       if (g_retarget_cache.size() < kMaxRetargetCacheEntries) {
         CachedRetargetResult entry;
         entry.succeeded = rewritten;
         if (rewritten) {
-          const auto* data = static_cast<const uint8_t*>((*out_elf_buffer).get());
-          entry.elf_bytes = std::make_shared<std::vector<uint8_t>>(data, data + *out_elf_size);
+          entry.elf_bytes = *out_elf_buffer;
```
cfreeamd, line 642: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** `RetargetCodeObject` now returns nullptr for both deterministic COMGR failures and a `make_shared` `bad_alloc`. Because `retarget_attempted` is true whenever COMGR ran, a transient OOM after a successful rewrite is stored here as `succeeded=false` and treated as a deterministic cached failure on later loads — contradicting the comment that transient allocation failures aren't cached. Pre-existing, but the refactor preserves it. Consider not caching when the failure is allocation-driven, or softening the comment.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -492,38 +498,44 @@ bool RetargetCodeObject(const void* elf_data, size_t elf_size, const char* sourc
...
+    output_buffer = std::make_shared<std::vector<uint8_t>>(output_size);
+  } catch (const std::bad_alloc&) {
     api->release_data(output);
-    return false;
+    return nullptr;
   }
 
   size_t copy_size = output_size;
-  if (api->get_data(output, &copy_size, static_cast<char*>(output_buffer.get())) !=
+  if (api->get_data(output, &copy_size, reinterpret_cast<char*>(output_buffer->data())) !=
```
cfreeamd, line 522: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** The `copy_size` returned by this `get_data` call is discarded; the buffer keeps the size-query `output_size`. If COMGR ever returned fewer bytes on the copy than on the size query, the oversized `->size()` would be handed to the loader and cached. In practice they match (and this matches prior behavior), so low risk — a `copy_size == output_size` assert or a shrink would harden it.

cfreeamd (commented):

Reviewed the ROCr hotswap ownership refactor (unique_ptr copy-per-load → shared_ptr refcounted sharing). The design is clean and the new `CachedLoadsShareOneBufferNoCopy` test plus the `test_categories.yaml` updates look correct. Left inline notes on the loader read-only invariant (the correctness crux of the change), an exception-escape in the last-resort OOM path, and a couple of pre-existing robustness nits.

suryajasper:

PR #8739, opened 2026-07-17
fix(rocr): PC Sampling: fix lost-wakeup race in consumer notification

shwetagkhatri:
Wrap pending_flush_count.fetch_add() and consumer_cv.notify_one() inside the consumer_mutex lock at both the exit path and normal flush path. This ensures the increment and the consumer's wait() predicate check are mutually exclusive, eliminating the lost-wakeup race condition.

The race occurred because:
1. Consumer evaluates pending_flush_count == 0 (under lock)
2. Consumer releases lock, about to enter futex wait
3. Producer increments pending_flush_count - OUTSIDE lock
4. Producer calls notify_one()
5. Consumer enters futex wait - MISSES notification

On MI300X with 8 XCCs flushing simultaneously, all producers could
complete fetch_add + notify_one before the consumer entered wait(),
causing permanent hang since no subsequent traps would fire.

With the fix, either:
- Consumer is already waiting: increment + notify wakes it
- Consumer checks predicate: sees incremented count, doesn't wait

## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
@@ -4317,8 +4317,11 @@ hsa_status_t GpuAgent::PcSamplingStart(pcs::PcsRuntime::PcSamplingSession& sessi
   pcs_data->session->stop();
 
   // Stop consumer thread first
-  pcs_data->consumer_exit.store(true, std::memory_order_release);
-  pcs_data->consumer_cv.notify_one();
+  {
+    std::lock_guard<std::mutex> lock(pcs_data->consumer_mutex);
+    pcs_data->consumer_exit.store(true, std::memory_order_release);
+    pcs_data->consumer_cv.notify_one();
+  }
   if (pcs_data->consumer_thread.joinable()) {
     pcs_data->consumer_thread.join();
   }
@@ -4380,8 +4383,11 @@ hsa_status_t GpuAgent::PcSamplingStop(pcs::PcsRuntime::PcSamplingSession& sessio
   // 2. Consumer may have unprocessed notifications - that's OK, Flush handles it
   // 3. Stopping consumer first avoids wasteful concurrent access (both use same buffers/mutexes)
   // 4. Final flush reads all data from host buffers and delivers via callback
-  pcs_data->consumer_exit.store(true, std::memory_order_release);
-  pcs_data->consumer_cv.notify_one();
+  {
+    std::lock_guard<std::mutex> lock(pcs_data->consumer_mutex);
+    pcs_data->consumer_exit.store(true, std::memory_order_release);
+    pcs_data->consumer_cv.notify_one();
+  }
   if (pcs_data->consumer_thread.joinable()) {
     pcs_data->consumer_thread.join();
   }
@@ -4825,8 +4831,12 @@ void GpuAgent::PcSamplingThreadPerXCC(pcs_data_t& pcs_data, uint32_t xcc_id,
 
       if (val == -1) {
         // Exit signal received - notify consumer and exit
-        pcs_data.pending_flush_count.fetch_add(1, std::memory_order_release);
-        pcs_data.consumer_cv.notify_one();
+        // fetch_add + notify_one must be under same lock to prevent lost-wakeup race
+        {
+          std::lock_guard<std::mutex> lock(pcs_data.consumer_mutex);
+          pcs_data.pending_flush_count.fetch_add(1, std::memory_order_release);
+          pcs_data.consumer_cv.notify_one();
+        }
         break;
       } else if (val != 0) {
         // Spurious wakeup - continue waiting
@@ -4851,8 +4861,15 @@ void GpuAgent::PcSamplingThreadPerXCC(pcs_data_t& pcs_data, uint32_t xcc_id,
       }
 
       // Notify consumer thread that new data is available
-      pcs_data.pending_flush_count.fetch_add(1, std::memory_order_release);
-      pcs_data.consumer_cv.notify_one();
+      // fetch_add + notify_one must be under same lock to prevent lost-wakeup race.
+      // The consumer's wait() predicate checks pending_flush_count under this same lock, so either:
+      // 1. Consumer is waiting: we increment, then notify wakes it up
+      // 2. Consumer checks predicate: it sees our incremented count and doesn't wait
+      {
+        std::lock_guard<std::mutex> lock(pcs_data.consumer_mutex);
+        pcs_data.pending_flush_count.fetch_add(1, std::memory_order_release);
+        pcs_data.consumer_cv.notify_one();
+      }
     }
 
     debug_print("%s (XCC %u)::Exiting\n", thread_name, xcc_id);
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
@@ -4220,8 +4220,8 @@ hsa_status_t GpuAgent::PcSamplingStart(pcs::PcsRuntime::PcSamplingSession& sessi
     HSA::hsa_signal_store_screlease(pcs_data->xcc_data[xcc_id].done_sig1, 1);
     pcs_data->xcc_data[xcc_id].which_buffer = 0;
   }
-  pcs_data->consumer_exit.store(false, std::memory_order_release);
-  pcs_data->pending_flush_count.store(0, std::memory_order_release);
+  pcs_data->consumer_exit.store(false, std::memory_order_relaxed);
+  pcs_data->pending_flush_count = 0;
 
   struct ThreadData {
     GpuAgent* agent;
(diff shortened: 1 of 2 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -4851,8 +4855,15 @@ void GpuAgent::PcSamplingThreadPerXCC(pcs_data_t& pcs_data, uint32_t xcc_id,
...
-      pcs_data.pending_flush_count.fetch_add(1, std::memory_order_release);
-      pcs_data.consumer_cv.notify_one();
+      // fetch_add + notify_one must be under same lock to prevent lost-wakeup race.
+      // The consumer's wait() predicate checks pending_flush_count under this same lock, so either:
+      // 1. Consumer is waiting: we increment, then notify wakes it up
+      // 2. Consumer checks predicate: it sees our incremented count and doesn't wait
+      {
+        std::lock_guard<std::mutex> lock(pcs_data.consumer_mutex);
+        pcs_data.pending_flush_count.fetch_add(1, std::memory_order_release);
+        pcs_data.consumer_cv.notify_one();
```
cfreeamd, line 4883: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** This fix is correct here, but it looks incomplete. The consumer's `wait()` predicate (in `PcSamplingConsumerThread`) wakes on `pending_flush_count > 0` **or** `consumer_exit`. This PR wraps the `pending_flush_count` producer sites in `consumer_mutex`, but the two `consumer_exit` producer sites still do an unlocked `store + notify_one`:

- `PcSamplingStart` failure cleanup (~L4320-4321)
- `PcSamplingStop` (~L4383-4384)

```cpp
pcs_data->consumer_exit.store(true, std::memory_order_release);
pcs_data->consumer_cv.notify_one();   // not under consumer_mutex
```

This has the identical lost-wakeup race. In `PcSamplingStop` the XCC threads have already exited before the exit is signaled, so no further `pending_flush_count` notifications will arrive. If the consumer just drained the last batch (reset the count to 0, delivered samples, looped back) and is between its predicate check and blocking in `wait()`, the unlocked exit notify can be missed → the consumer blocks forever → `consumer_thread.join()` hangs during teardown. Same applies to the error-path cleanup in `PcSamplingStart`.

Suggest applying the same pattern to both exit sites:
```cpp
{
  std::lock_guard<std::mutex> lock(pcs_data->consumer_mutex);
  pcs_data->consumer_exit.store(true, std::memory_order_release);
  pcs_data->consumer_cv.notify_one();
}
```

cfreeamd (commented):

Reviewed the PC Sampling lost-wakeup fix in `amd_gpu_agent.cpp`. The two fixed producer sites in `PcSamplingThreadPerXCC` are correct. Left one inline note: the same lock discipline appears to be needed at the `consumer_exit` notification sites (`PcSamplingStart` cleanup and `PcSamplingStop`), which still notify without holding `consumer_mutex`.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -4851,8 +4855,15 @@ void GpuAgent::PcSamplingThreadPerXCC(pcs_data_t& pcs_data, uint32_t xcc_id,
...
+      }
```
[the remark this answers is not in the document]
shwetagkhatri: Fixed

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
shwetagkhatri: Good catch @cfreeamd ! I have updated the commit to fix both consumer_exit producer sites as well.

vlaindic (commented):
## Review: PC Sampling lost-wakeup fix

The core fix (wrapping `fetch_add` + `notify_one` under `consumer_mutex` in the XCC producer threads) is correct and well-motivated. Four follow-up items below.

---

### 1. IMPORTANT — `consumer_exit` notify without lock: same race left alive

This PR fixes the lost-wakeup race for the `pending_flush_count` path but leaves the identical race alive for the `consumer_exit` predicate. Two call sites set `consumer_exit` and call `notify_one()` outside `consumer_mutex`:

**Startup failure cleanup (~line 4320):**
```cpp
pcs_data->consumer_exit.store(true, std::memory_order_release);
pcs_data->consumer_cv.notify_one();  // ← not under consumer_mutex
```

**`PcSamplingStop` (~line 4383):**
```cpp
pcs_data->consumer_exit.store(true, std::memory_order_release);
pcs_data->consumer_cv.notify_one();  // ← not under consumer_mutex
```

The consumer's `wait()` predicate checks `consumer_exit` under `consumer_mutex`:
```cpp
pcs_data.consumer_cv.wait(lock, [&pcs_data]() {
    return pcs_data.pending_flush_count.load(...) > 0 ||
           pcs_data.consumer_exit.load(...);  // ← predicate variable, same race window
});
```

The lost-wakeup window is identical to the one described in the PR: consumer evaluates `consumer_exit == false`, releases the lock, is about to sleep — `consumer_exit.store(true)` + `notify_one()` fires outside the lock — consumer enters `wait()` and sleeps indefinitely.

Apply the same fix at both call sites:
```cpp
{
    std::lock_guard<std::mutex> lock(pcs_data->consumer_mutex);
    pcs_data->consumer_exit.store(true, std::memory_order_relaxed);
    pcs_data->consumer_cv.notify_one();
}
```

---

### 2. Nit — `memory_order_release` on `fetch_add` inside lock is redundant

At both new lock scopes added by this PR:
```cpp
std::lock_guard<std::mutex> lock(pcs_data.consumer_mutex);
pcs_data.pending_flush_count.fetch_add(1, std::memory_order_release);  // ← redundant ordering
```

The mutex unlock already emits a release fence — `memory_order_release` on the atomic does no additional work here. Using it inside a lock sends a false signal that the atomic ordering is load-bearing, when the synchronization is entirely provided by the mutex. `memory_order_relaxed` is the correct and idiomatic choice inside a lock scope.

Same applies to the `consumer_exit.store()` fix suggested above, and as a follow-up to the pre-existing `memory_order_acquire` on the `load()` calls inside `PcSamplingConsumerThread` (those also sit under the lock).

```cpp
// preferred
pcs_data.pending_flush_count.fetch_add(1, std::memory_order_relaxed);
```

---

### 3. Nit — Exit-path comment less precise than flush-path comment

The flush-path comment (added in this PR) explains both cases clearly:
```cpp
// fetch_add + notify_one must be under same lock to prevent lost-wakeup race.
// The consumer's wait() predicate checks pending_flush_count under this same lock, so either:
// 1. Consumer is waiting: we increment, then notify wakes it up
// 2. Consumer checks predicate: it sees our incremented count and doesn't wait
```

The `val == -1` exit-path comment is shorter and omits the two-case explanation:
```cpp
// fetch_add + notify_one must be under same lock to prevent lost-wakeup race
```

Extend the exit-path comment to match, so the invariant is explained consistently at both sites.

---

### 4. Nit — `pending_flush_count` can be a plain `uint32_t` after this fix

After this PR, every access to `pending_flush_count` is under `consumer_mutex`:

- `fetch_add` at both producer notify sites — now under `lock_guard<consumer_mutex>` (this PR)
- `load` inside the `wait()` predicate — under `unique_lock<consumer_mutex>` (consumer side)
- `store(0)` reset after `wait()` — under `unique_lock<consumer_mutex>` (consumer side)
- `store(0)` at initialization — single-threaded, no race

The mutex serializes all of these, so there is no concurrent unsynchronized access. `std::atomic<uint32_t>` is unnecessary — a plain `uint32_t` is correct and communicates the intent more clearly: *this variable is protected by `consumer_mutex`, not by its own atomicity.*

Note: `consumer_exit` is different — it has one lockless read at the outer `while` loop condition in `PcSamplingConsumerThread` (outside the `unique_lock` scope), so it legitimately needs to remain atomic.

Suggested change in `amd_gpu_agent.h`:
```cpp
// before
std::atomic<uint32_t> pending_flush_count;

// after
uint32_t pending_flush_count;  // protected by consumer_mutex
```

And drop the memory order annotations on all accesses — they become plain reads and writes.

vlaindic (changes requested):
Thanks for the fix, this solves the major issue, but apparently, two more (less frequent) might be missing. Please check the details here: https://github.com/ROCm/rocm-systems/pull/8739#pullrequestreview-4724162132

shwetagkhatri:
> Thanks for the fix, this solves the major issue, but apparently, two more (less frequent) might be missing. Please check the details here: [#8739 (review)](https://github.com/ROCm/rocm-systems/pull/8739#pullrequestreview-4724162132)

Thank you @vlaindic ! I should have searched for all consumer_cv.notify_one() calls and verified that each one was under consumer_mutex.. I just fixed all your review comments. Can you please give it another look? Thanks you.

vlaindic (approved):
LGTM, thanks!

vlaindic (approved):
All four issues from the previous review are resolved:

1. **consumer_exit notify without lock** — both exit sites (start-failure path and PcSamplingStop) are now properly wrapped under consumer_mutex before notify_one(). ✅
2. **memory_order_release inside lock** — replaced with memory_order_relaxed everywhere inside lock scopes; plain reads/writes where atomics were demoted. ✅
3. **Exit-path comment completeness** — the two-case explanation is consistently applied at all notify sites. ✅
4. **pending_flush_count as plain uint32_t** — demoted from std::atomic<uint32_t> to uint32_t in the header; all accesses are now plain reads/writes under consumer_mutex. ✅

LGTM.

cfreeamd (approved):

dayatsin-amd (approved):

PR #8740, opened 2026-07-17
fix(rocr): Ignore HSA_TOOLS_LIB in secure-execution context

ApurvMishra-amd:
## Motivation

`Runtime::LoadTools()` read the user-controlled `HSA_TOOLS_LIB` environment variable and passed each token straight to `dlopen()`, with no check for privileged execution. When ROCr is loaded by a `setuid`/`setgid` or file-capability binary that does not scrub its environment, this is a local privilege-escalation primitive: an attacker can point `HSA_TOOLS_LIB` at an arbitrary .so and have it loaded with elevated privileges.

## Technical Details

`glibc` drops its own dangerous env vars (`LD_PRELOAD`, ...) in secure-execution mode but knows nothing about `HSA_TOOLS_LIB`, so apply the same discipline here: when `geteuid()!=getuid()`, `getegid()!=getgid()`, or `getauxval(AT_SECURE)` is set, ignore `HSA_TOOLS_LIB entirely`. Normal (non-secure) execution is unaffected.

## Issue Tracking

JIRA ID: ROCM-26867

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
@@ -52,6 +52,11 @@
 #include <dlfcn.h>
 #include <amdgpu_drm.h>
 #include <sys/mman.h>
+#include <unistd.h>
+#include <sys/auxv.h>
+#ifndef AT_SECURE
+#define AT_SECURE 23
+#endif
 #endif
 
 #include "core/inc/runtime.h"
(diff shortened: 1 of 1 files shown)


cfreeamd (commented):

Reviewed the `runtime.cpp` change. The secure-context detection and clearing of env-var tool libs looks correct and is properly guarded under `#if defined(__linux__)`, so it doesn't affect the FreeBSD/Windows builds. Clearing `names` fully prevents the `dlopen` path, and the already-loaded-tool discovery path is correctly left untouched. One non-blocking inline note.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
```diff
@@ -2953,10 +2958,28 @@ void Runtime::LoadTools() {
...
+  // escalation primitive. glibc scrubs its own env vars (LD_PRELOAD, ...) in
+  // secure-execution mode but not HSA_TOOLS_LIB, so ignore it ourselves here.
   std::string tool_names = flag_.tools_lib_names();
   std::vector<std::string> names;
   if (tool_names != "") {
     names = parse_tool_names(std::move(tool_names));
+
+#if defined(__linux__)
+    const bool secure =
+        (geteuid() != getuid()) || (getegid() != getgid()) || (getauxval(AT_SECURE) != 0);
```
cfreeamd, line 2973: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** `getauxval(AT_SECURE)` is the canonical secure-execution flag — the kernel sets it for setuid, setgid, and file-capability transitions — so the `geteuid()!=getuid()` / `getegid()!=getgid()` comparisons are logically subsumed by it. Keeping them is harmless (and arguably good, since `getauxval` returns 0 on lookup failure on older glibc, so the euid/egid checks still catch the setuid/setgid cases). No change needed — flagging only so the redundancy reads as intentional defense-in-depth rather than an oversight.

cfreeamd (approved):

PR #8741, opened 2026-07-17
feat(hotswap): bounded two-tier (memory + disk) retarget cache with shared buffers

suryajasper:
## Summary

Extends the hotswap retarget cache into a bounded, two-tier (in-memory + disk)
cache **built on the shared-buffer paradigm** so that repeated code-object loads
— across GPUs within a process and across process launches on a node — skip the
expensive COMGR retarget **without copying the retargeted ELF on a cache hit**.

This is the disk + byte-budget cache from #8659, re-based onto the shared-buffer
retarget change (#8737) so that both cache tiers hand the loader the cache's own
buffer by reference (refcount++) instead of a per-load `malloc`+`memcpy`.

**Proposed as a replacement for #8659** — reviewers may choose to supersede it
with this implementation.

## Relationship to other PRs

- **Stacked on #8737** ("share retarget buffers across loads instead of
  copying"). This branch contains #8737's commit as its base; once #8737 lands,
  this rebases cleanly onto develop.
- **Supersedes #8659** ("bounded two-tier retarget cache"). The disk-tier
  mechanism here is #8659's, unchanged — the only difference is the zero-copy
  (`ElfBufferRef`) adaptation.

## Technical details

The change is two commits:

1. **share retarget buffers** (#8737): `OwnedElfBuffer` (`unique_ptr<void,
   &free>`) → `ElfBufferRef` (`shared_ptr<vector<uint8_t>>`). A cache hit hands
   back the cache's own buffer, refcounted, rather than a private copy.

2. **byte-budget LRU + disk tier** (this PR's addition, from #8659):
   - **In-memory tier:** bounded by an LRU **byte budget** (default 4 GiB,
     `HSA_HOTSWAP_MEM_BUDGET`; `0` = unbounded) instead of a fixed entry count.
     Failure sentinels consume no budget; a lone oversized entry is retained.
   - **Disk tier:** cross-process, immutable, write-once/read-many, no size cap.
     Consulted on an in-memory miss before COMGR; a disk hit is promoted into the
     in-memory tier **and shared with the caller (no copy)**. Entries are
     namespaced by a COMGR-identity salt (resolved lib path + size + mtime +
     format version), published atomically (temp file + `fsync` + `rename`), and
     validated (magic/version/salt/size) on read. Controls:
     `HSA_HOTSWAP_CACHE_DIR`, `HSA_HOTSWAP_DISK_CACHE=0`.
   - **Async writer:** a single background thread owned by the ROCr Runtime
     lifecycle (`HotswapCacheStartup`/`Shutdown` from `Runtime::Load`/`Unload`)
     keeps the one-time cold persist off the load critical path. Each write task
     shares the same buffer, so eviction before the write lands cannot lose the
     bytes.

The cold-path COMGR copy-out (across the COMGR C ABI) is the only remaining copy;
from there the one buffer is shared by the in-memory cache, the disk writer, and
every consuming load.

## Controls

| Env var | Effect |
|---------|--------|
| `HSA_HOTSWAP_MEM_BUDGET` | In-memory byte budget (default 4 GiB; `0` = unbounded) |
| `HSA_HOTSWAP_CACHE_DIR` | Disk cache location (default `$XDG_CACHE_HOME/rocm/hotswap` or `$HOME/.cache/rocm/hotswap`) |
| `HSA_HOTSWAP_DISK_CACHE=0` | Disable the disk tier |

## Test plan

- [x] New `HotswapRewrite.CachedLoadsShareOneBufferNoCopy`: two cache-served
  loads receive the **same** buffer pointer (proves zero-copy sharing).
- [x] `RetargetCacheLru.*` (byte/size accounting, eviction, recency,
  oversized-entry retention, unbounded mode, failure sentinels).
- [x] `RetargetDiskCache.*` (write/read round-trip, salt-mismatch miss,
  missing-entry miss, distinct-key isolation, idempotent overwrite, async writer
  drain-at-shutdown).
- [x] Existing `RetargetCacheServesSecondLoadFromCache` and decision/routing
  tests adapted to the shared-buffer API.
- [x] Builds clean and full `hotswap_rewrite` suite passes on MI300X (45/45).


cfreeamd (commented):
🤖 AI-assisted pre-review (for @cfreeamd) — model: opus

Reviewed the full PR (all files under `projects/rocr-runtime/`). The concurrency core is solid: the `RetargetCache` LRU (intrusive `std::list` + `unordered_map` with a stored `lru_it`), the `DiskWriter` background thread, and the shutdown drain (predicate-guarded `cv_.wait` + inner re-drain loop) are all correct, and the disk tier is genuinely robust — atomic temp-file+rename publish, per-toolchain salt from the COMGR library identity, `fstat`-on-fd, and payload-size validation before allocation. New tests are properly registered in `test_categories.yaml` across all four categories. Findings are Optional/Nit — the most worthwhile is the `ParseMemBudgetEnv` partial-parse handling (suffix/hex env values silently mis-size or disable the bound).

Note (no line anchor): the POSIX-only `RetargetDiskCache.*` tests are `#if !_WIN32`-gated but are listed in `test_categories.yaml` without corresponding `exclude_windows` entries. GTest treats a filter naming a non-existent test as a no-op, so this is likely harmless, but worth confirming the CI runner doesn't fail-hard on missing test names on Windows.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -115,10 +131,216 @@ struct CachedRetargetResult {
...
+  if (value[0] == '-') {
+    return kDefaultRetargetCacheMemBudget;
+  }
+  // Accept a plain byte count. errno/strtoull guard against garbage; on any
+  // parse failure fall back to the default rather than an accidental 0
+  // (unbounded) or overflow. An explicit "0" is honored as unbounded.
+  errno = 0;
+  char* end = nullptr;
+  const unsigned long long parsed = std::strtoull(value.c_str(), &end, 10);
+  if (end == value.c_str() || errno != 0) {
```
cfreeamd, line 334: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** `ParseMemBudgetEnv` silently accepts partial parses. `strtoull(base=10)` stops at the first non-digit, and the guard only rejects the *no-digits* case (`end == value.c_str()`). So:
- `"256K"` / `"1M"` → parses just the number and drops the suffix (a 1000×-smaller budget than intended)
- `"0x100"` → stops at `x`, returns `0`, which this function treats as *unbounded* — the opposite of the user's intent

Since this is an env var (a system boundary) a malformed value should fall back to the default rather than silently mis-size the cache. Suggest checking that the whole string was consumed after `strtoull`:
```cpp
while (*end == ' ' || *end == '\t') ++end;
if (*end != '\0') return kDefaultRetargetCacheMemBudget;
```
(and a `SIZE_MAX` bound before the cast for 32-bit builds).

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -387,6 +627,390 @@ std::optional<RewriteDecision> DecideHotswapRewrite(
...
+    return nullptr;
+  }
+  DiskCacheHeader header;
+  if (std::fread(&header, sizeof(header), 1, f) != 1) {
+    std::fclose(f);
+    return nullptr;
+  }
+  if (std::memcmp(header.magic, kDiskCacheMagic, sizeof(kDiskCacheMagic)) != 0 ||
+      header.format_version != kDiskCacheFormatVersion ||
+      header.comgr_salt != salt || header.payload_size == 0) {
```
cfreeamd, line 790: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** `ReadDiskCache` rejects any header with `payload_size == 0`, but `WriteDiskCache` will happily publish a header-only file for an empty payload — so such an entry can never be read back. It's unreachable in production (`RetargetCodeObject` guards `output_size == 0` and returns `nullptr`, so an empty buffer is never enqueued), but it *is* observable via `DiskCacheWriteForTesting`. Consider an early return for `payload.empty()` in `WriteDiskCache` so the write/read contract is symmetric; the guard here can stay as a cross-check.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -115,10 +131,216 @@ struct CachedRetargetResult {
...
+        *out_bytes = it->second.result.elf_bytes;
+      }
+      return Lookup::kHitSuccess;
+    }
+    return Lookup::kHitFailure;
+  }
+
+  // Inserts (or replaces) `key` with a success entry owning `bytes`, evicting
+  // LRU entries as needed to respect the budget. `bytes` must be non-null.
+  void PutSuccess(uint64_t key, std::shared_ptr<std::vector<uint8_t>> bytes) {
```
cfreeamd, line 183: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `PutSuccess`'s doc says `bytes` must be non-null, but the body accepts null and stores a `succeeded=true` entry with null `elf_bytes`, creating a phantom success that occupies an LRU slot and forces callers to guard `kHitSuccess && cached_bytes`. Making the contract self-enforcing (`if (!bytes) { PutFailure(key); return; }`) removes the caller-side guard requirement.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -748,14 +1438,101 @@ void ForceRetargetCodeObjectFailureForTesting(bool force) {
...
+// Drives the background DiskWriter directly: starts it, enqueues `n` writes of
+// `payload` under `dir` (keys 0..n-1, salt fixed), then Stop() which must drain
+// all pending writes before joining. Returns the number of entries that are
+// readable back afterward (should equal `n` if the drain-at-shutdown path is
+// correct). Exercises the real async writer concurrency logic.
+int DiskWriterDrainRoundTripForTesting(const std::string& dir, int n,
+                                       const std::vector<uint8_t>& payload) {
+#if HOTSWAP_DISK_CACHE_SUPPORTED
+  constexpr uint64_t kSalt = 0xA5A5A5A5A5A5A5A5ULL;
+  DiskWriter& writer = GetDiskWriter();
```
cfreeamd, line 1513: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** `DiskWriterDrainRoundTripForTesting` drives the *global* `GetDiskWriter()` singleton and calls `Stop()` at the end, leaving `running_ = false`. Any later test in the same binary that relies on disk persistence (e.g. an end-to-end load after `HotswapCacheStartup`) will then silently drop all disk writes. Consider restarting the writer before returning, or scoping the drain test to a locally-constructed `DiskWriter` rather than the shared singleton.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -607,33 +1252,51 @@ RetargetCodeObjectResult TryRetargetCodeObject(const CodeObjectView& code_object
...
+    rewritten = (*out_elf_buffer != nullptr);
   }
 
   // Cache the result. Only deterministic COMGR outcomes are cached; transient
   // allocation failures and test-forced failures are not, so a later attempt
-  // with the same code object can still succeed.
+  // with the same code object can still succeed. On success the SAME buffer is
+  // stored, persisted, and returned (no copy) — the cache, the disk writer, and
+  // the caller all share one refcounted object. Concurrent cold misses on the
+  // same key may each run COMGR and race to store (first-wins); the loser still
```
cfreeamd, line 1268: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** Comment says concurrent cold misses "race to store (first-wins)", but `PutSuccess` → `InsertLocked` replaces unconditionally and the disk tier's `rename()` also last-wins. Both buffers are valid (COMGR is deterministic) so correctness is fine — just update the comment to "last-wins (each valid)".

projects/rocr-runtime/rocrtst/suites/functional/hotswap/hotswap_rewrite_test.cc
```diff
@@ -839,4 +854,236 @@ TEST(HotswapRewrite, RuntimeLoadRequiredStrictRewrittenLoadFailureReturnsError)
...
+}
+
+// --- Disk tier round-trip unit tests (POSIX only) ---------------------------
+//
+// Exercise the disk write/read format, salt validation, and atomic publish in
+// isolation via the synchronous testing hooks. No COMGR or device needed.
+
+#if !defined(_WIN32) && !defined(_WIN64)
+namespace {
+int g_next_test_dir_counter = 0;
```
cfreeamd, line 992: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `g_next_test_dir_counter` is a plain `int`. Fine today since GTest runs serially, but if parallel execution is ever enabled two `ScopedTempDir` constructions could race. `std::atomic<int>` would make it future-proof.

suryajasper:

PR #8743, opened 2026-07-17
WarpSpeed Refactor and Gating to SPX

alex-breslow-amd:
## Motivation
Restore performance on single node for MI350 when WarpSpeed enabled.

## Technical Details
**New as of 7/22/2026**
-Address @Kapil-Shyam-Pawar's feedback regarding changes getting picked up by the testrunner.  Thanks for the pointers! :-)

**New as of 7/17/2026**
- Synced with team.  We disable WarpSpeed on non-SPX modes (e.g., DPX, QPX, CPX).  Thanks @mustafabar and @wenkaidu for the suggestion.
- Move WarpSpeed enablement logic into separate functions that we can unit test
- Add unit tests for enablement logic
- Fixes regression in single node performance from PR#8077

**Old as of 7/16/2026**
Drop in performance due to max channels being capped at 32 for large message sizes on single node.  This is one possible fix.  Need to verify that it plays nice with CPX and still prevents the issue mentioned in PR#8077. 

We need to decide if we should just revert to not using WarpSpeed and go back to the original logic.  Will consult with team tomorrow.

Another option is checking which mode we're in e.g., DPX or CPX and having custom max channel counts for that.

## Issue Tracking
JIRA ID: AICOMRCCL-1553

## Test Plan

Ran with this PR on develop and without it.  Verified that performance returns and channel count is again 56 for 1GiB.

## Test Result

See above

## Submission Checklist


wenkaidu:
LGTM. I think SPX should be used as gate for other resource intensive algorithm as well, like direct algorithms.

projects/rccl/src/init.cc
```diff
@@ -1338,6 +1338,13 @@ static ncclResult_t ncclP2pSchedule(struct ncclComm* comm) {
   return ncclSuccess;
 }
 
+#ifdef ENABLE_WARP_SPEED
+static bool willEnableWarpSpeed(struct ncclComm* parent, struct ncclComm* comm, int nNodes) {
+  return rcclParamWarpSpeedForceEnable() > 0 ||
```
Kapil-Shyam-Pawar, line 1344: if force enabled on any mode other than SPX, is there any chance of channels going out of bounds and crashing?
if not, maybe we can add a comment somewhere on how it's handled?

projects/rccl/src/rccl_wrap.cc
```diff
@@ -743,7 +743,7 @@ bool rcclIsAboveWarpSpeedThreshold(struct ncclComm* comm, struct ncclTaskColl* i
 
 bool rcclCanUseWarpSpeedAuto(struct ncclComm* comm, int nNodes) {
   return IsArchMatch(comm->topo->nodes[GPU].nodes[0].gpu.gcn, "gfx950") && (nNodes == 1) &&
-         (rcclParamWarpSpeedAutoMode() != 0);
+         (rcclParamWarpSpeedAutoMode() != 0) && comm-> cuCount > 128; // Only use in SPX mode, 256 CU on gfx950
```
Kapil-Shyam-Pawar, line 746: nit: stray space `comm-> cuCount`?
and would it be better if 128 were a named constant for readability?

projects/rccl/test/RcclWrapTests.cpp
```diff
@@ -1544,4 +1545,353 @@ TEST(Rcclwrap, RcclUseHierarchicalAllGatherTests)
...
+//
+// These exercise the pure decision/tuning helpers that gate WarpSpeed:
+//   - rcclCanUseWarpSpeedAuto     (arch/node/env eligibility)
+//   - rcclGetMaxWarpsPerBlock     (warps-per-block multiplier)
+//   - rcclWarpSpeedComputeNChannels (connect.cc channel math)
+//   - rcclWarpSpeedAdjustChannels   (enqueue.cc per-collective channel adj.)
+// ---------------------------------------------------------------------------
+
+// rcclCanUseWarpSpeedAuto: gfx950 single-node with auto mode on -> eligible.
+TEST(RcclwrapWarpSpeed, CanUseWarpSpeedAuto_Gfx950SingleNode_True)
```
Kapil-Shyam-Pawar: I would suggest renaming `RcclwrapWarpSpeed` -> `Rcclwrap` for these tests or adding `RcclwrapWarpSpeed` to config so that they get picked up by the code cov ci - https://github.com/ROCm/rocm-systems/blob/bba30fedd1cc47eff941b0df42bc91bd574cb379/projects/rccl/tools/scripts/test_runner/configs/mi300x_mellanox_ib.json#L521-L523

Kapil-Shyam-Pawar (approved):

projects/rccl/src/init.cc
alex-breslow-amd: Is possible yes.  I can update this.  I thought it was reasonable to let it crash out, but now I'm less convinced.

alex-breslow-amd: This is a valid concern.  I will do it as a follow-up task.

projects/rccl/src/rccl_wrap.cc
alex-breslow-amd: Probably, will complete as follow up since similar gating should be applied to AllGather.

alex-breslow-amd:
> LGTM. I think SPX should be used as gate for other resource intensive algorithm as well, like direct algorithms.

I'll add a follow up task for this.  I think that makes sense at least on gfx942 and gfx950.

alex-breslow-amd:
Merging in develop because CI failed due to ROCK issue.  Fix and bump have landed: https://github.com/ROCm/TheRock/commit/2e7c190b99caa226a8644eda6eca720b3db102d5 and https://github.com/ROCm/rocm-systems/pull/9056

PR #8744, opened 2026-07-17
refactor(hotswap): use a process-wide exact-content single-flight cache

harsh-amd:
## Summary

Replace the lookup/copy cache from #8274 and the earlier reader-scoped revision
of this PR with a process-wide, exact-content single-flight cache.

- Reuse one rewrite across distinct HSA readers when source bytes and every
  transform field are identical. This is required because CLR creates distinct
  readers for the same code object on different GPUs.
- Detect mutation between supported HSA calls. Reader memory is application
  owned and is not guaranteed immutable, so pointer, reader, and ISA identity
  are insufficient cache keys.
- Use xxHash64 only to select a bucket, followed by exact size and byte
  comparison against an immutable source snapshot. Forced hash collisions
  cannot alias results.
- Coalesce a matching cold miss on the calling threads: one caller runs COMGR
  and matching callers wait. No worker threads are added.
- Weak-intern one immutable source snapshot per distinct live content and share
  it across transforms. Publish one immutable output per distinct live
  transform and retain it through the loaded code object existing shared
  ownership path.
- Do not cache failures. Current waiters observe the same typed failure and a
  later call starts a new generation.
- Assign each code-object reader a monotonic numeric diagnostic ID. The cache
  never retains or compares reader pointers.
- Reject same-thread recursive joins on the same exact flight with
  `kReentrantRequest` instead of allowing a producer to wait on itself.

Concurrent application writes to reader memory during an HSA call remain
unsupported. A supported miss snapshots the bytes before COMGR, so the rewrite
uses one stable source sequence.

## Identity and synchronization

The transform identity contains source ISA, target ISA, entry-trampoline mode,
and strict mode. Reader identity is diagnostic only and is deliberately not
part of cache identity.

The first exact-content/transform caller inserts a flight under its transform
bucket mutex. COMGR runs outside all cache mutexes. Success or failure is
published under the same mutex used by predicate waiters before `notify_all`.
Waiters own their flight generation, so a failed generation can be removed and
retried without changing the result observed by old waiters.

Source candidate collection and publication use per-content buckets. Exact
comparison runs outside the content-bucket mutex and publication rechecks a
generation counter. Different contents, including forced hash collisions, and
different transforms can produce concurrently.

The cache lock order is table mutex before bucket mutex. Normal lookup releases
the table mutex before taking a bucket mutex, source and transform bucket
mutexes are never held together, and no path acquires the table mutex while
holding a bucket mutex. Predicate waits atomically release their bucket mutex.

The lazy process cache uses `std::call_once`; ROCR Unix builds use
`-fno-threadsafe-statics`. Allocation failure falls back to an uncached rewrite
where possible.

Liveness depends on COMGR returning, the producer thread not being
asynchronously terminated, and eventual scheduler progress. A stuck COMGR call
can stall matching waiters without creating a cache lock cycle; this design does
not claim wait-freedom or scheduler fairness.

## Memory, eviction, and file readers

For source size `B` and live rewritten outputs `S_i`, exact reuse over mutable
application memory requires retaining an immutable source identity in the worst
case. This implementation retains one shared `B`-byte snapshot plus one
`S_i`-byte output per distinct live transform, meeting the payload lower bound
`B + sum(S_i)` without per-hit output copies.

Ready entries and source-intern entries are weak. The last output owner releases
the output immediately; the source snapshot is released after the last output
or flight using it. Expired entries are removed on access and empty metadata
buckets are periodically swept. There is no strong payload retention, cache
arena, LRU, TTL, or byte-capacity eviction policy.

File-backed and memory readers use the same in-process content cache. This does
not add a persistent file cache. Disk-cache trust, permissions, toolchain salt,
atomic publication, corruption handling, and invalidation remain separate from
loader lifetime and in-process single-flight correctness.

## Metrics

Normal runtime builds expose and print the following with
`HSA_HOTSWAP_VERBOSE=1`:

- producer calls/failures, ready hits, coalesced results, cross-reader results,
  and reentrant rejections;
- hash and exact-compare byte/time counters, plus wait and lock-hold time;
- cumulative/live/peak source-snapshot and output bytes; and
- content/transform bucket, ready-entry, and in-flight counts.

## Evidence

- Built and linked `hsa-runtime64`.
- Built `hotswap_rewrite`, `hotswap_cache`, and `hotswap_cache_benchmark`.
- Focused rewrite/cache CTests pass and passed 50 consecutive iterations.
- All 13 cache tests pass under GCC ThreadSanitizer, including distinct-reader
  reuse, mutation, forced collision isolation, failure/retry, last-owner
  release, same-thread reentrant rejection, same-content/different-transform
  concurrency, and randomized schedules through 64 threads.
- TLC generated 510,325 states, found 121,729 distinct states, completed depth
  17, and found no invariant violation. The model covers exact key selection,
  forced collisions, one leader per generation, leader exclusion from waiters,
  publication, weak expiry, and failure/retry ordering.
- In the optimized four-caller 1 MiB cold benchmark, content single-flight uses
  one producer, three coalesced results, one 1 MiB source snapshot, and one
  1 MiB output. Its warm batch uses zero producers and allocates/copies no
  output.
- `git diff --check` passes.

Commands, the cost model, synchronization argument, benchmark schema, and TLC
configuration are in
[CACHE_EVIDENCE.md](projects/rocr-runtime/rocrtst/suites/functional/hotswap/CACHE_EVIDENCE.md).

## Remaining merge gate

This PR remains draft until the original four-GPU RCCL initialization path runs
with normal CLR reader creation. The final verbose metrics must show
`producer_calls=1`, `cross_reader_results=3`, `reentrant_rejections=0`, one
rewritten ELF in `produced_output_bytes`, and live payload equal to one source
snapshot plus one rewritten ELF.

Kernel execution, debugger/profiler code-object inspection, coredump behavior,
and repeated executable destruction must also pass on a supported GPU system.
The synthetic benchmark is not a substitute for these hardware checks.

JIRA ID: DCGPUBU-1871

This addresses the copy/lifetime concerns in #8737 while correcting the cache
scope and mutable-reader assumptions that a reader-scoped cache cannot satisfy.

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap_cache.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap_cache.cpp
@@ -52,6 +52,7 @@
 #include <mutex>
 #include <new>
 #include <string>
+#include <thread>
 #include <unordered_map>
 #include <utility>
 #include <vector>
@@ -212,15 +213,18 @@ struct RetargetCacheBucketKeyHash {
 
 struct RetargetReadyEntry {
   std::weak_ptr<const RetargetedElf> elf;
-  const void* producer_reader_identity = nullptr;
+  uint64_t producer_reader_id = 0;
 };
 
 struct RetargetFlight {
-  RetargetFlight(SourceSnapshotRef source_snapshot, const void* reader_identity)
-      : source(std::move(source_snapshot)), producer_reader_identity(reader_identity) {}
+  RetargetFlight(SourceSnapshotRef source_snapshot, uint64_t reader_id)
+      : source(std::move(source_snapshot)),
+        producer_reader_id(reader_id),
+        producer_thread_id(std::this_thread::get_id()) {}
 
   SourceSnapshotRef source;
-  const void* producer_reader_identity = nullptr;
+  uint64_t producer_reader_id = 0;
+  std::thread::id producer_thread_id;
   bool completed = false;
   size_t waiter_count = 0;
   RetargetOperationResult result;
@@ -256,9 +260,8 @@ RetargetOperationResult RunRetargetProducer(const ContentRetargetCache::Producer
   }
 }
 
-bool IsCrossReader(const void* producer_reader_identity, const void* reader_identity) {
-  return producer_reader_identity != nullptr && reader_identity != nullptr &&
-      producer_reader_identity != reader_identity;
+bool IsCrossReader(uint64_t producer_reader_id, uint64_t reader_id) {
+  return producer_reader_id != 0 && reader_id != 0 && producer_reader_id != reader_id;
 }
 
 }  // namespace
@@ -270,6 +273,7 @@ class RetargetCacheMemoryTracker final {
   std::atomic<uint64_t> ready_hits{0};
   std::atomic<uint64_t> cross_reader_results{0};
   std::atomic<uint64_t> coalesced_results{0};
+  std::atomic<uint64_t> reentrant_rejections{0};
   std::atomic<uint64_t> hash_bytes{0};
   std::atomic<uint64_t> hash_nanoseconds{0};
   std::atomic<uint64_t> exact_compare_bytes{0};
@@ -446,7 +450,7 @@ ContentRetargetCache::ContentRetargetCache(ContentHashFunction hash_function)
 
 RetargetOperationResult ContentRetargetCache::GetOrCompute(const void* source_data,
                                                            size_t source_size,
-                                                           const void* reader_identity,
+                                                           uint64_t reader_id,
                                                            const RetargetCacheKey& key,
                                                            const Producer& producer) {
   const auto run_producer = [&](const SourceSnapshotRef& source) {
@@ -546,7 +550,7 @@ RetargetOperationResult ContentRetargetCache::GetOrCompute(const void* source_da
       }
       if (elf->source().get() == source.get()) {
         impl_->tracker->ready_hits.fetch_add(1, std::memory_order_relaxed);
-        if (IsCrossReader(ready->producer_reader_identity, reader_identity)) {
+        if (IsCrossReader(ready->producer_reader_id, reader_id)) {
           impl_->tracker->cross_reader_results.fetch_add(1, std::memory_order_relaxed);
         }
         AddElapsed(&impl_->tracker->lock_hold_nanoseconds, lock_acquired);
@@ -557,6 +561,12 @@ RetargetOperationResult ContentRetargetCache::GetOrCompute(const void* source_da
 
     for (const auto& candidate : bucket->in_flight) {
       if (candidate->source.get() != source.get()) continue;
+      // A producer cannot wait for the flight that only it can complete.
+      if (candidate->producer_thread_id == std::this_thread::get_id()) {
+        impl_->tracker->reentrant_rejections.fetch_add(1, std::memory_order_relaxed);
+        AddElapsed(&impl_->tracker->lock_hold_nanoseconds, lock_acquired);
+        return {{}, RetargetError::kReentrantRequest};
+      }
 
       flight = candidate;
       ++flight->waiter_count;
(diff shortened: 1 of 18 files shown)


harsh-amd:
Evidence added in `808f6974c2`.

### Correctness and synchronization

- Added a dedicated isolated cache test target with deterministic same-key single-flight, different-key parallelism, exact separation of all four transform-key fields, weak ownership, producer exception, `std::bad_alloc`, failure wakeup, and retry cases.
- Added randomized stress for 1, 2, 4, 8, 16, 32, and 64 callers. The final six-test suite passed 50 consecutive CTest runs.
- Built the isolated target with GCC ThreadSanitizer and ran it on WSL2 using non-PIE plus `setarch x86_64 -R`. All six tests passed, including the 64-thread stress case, with no TSAN report.
- Added a TLA+/TLC model covering successful and failed publication, weak-entry expiry, and a retry starting before old waiters wake. TLC completed the state graph with no invariant violation: 74,189 generated states, 18,685 distinct states, depth 18, zero states left.
- Retained integration coverage for optional fallback, required fail-closed behavior, rewritten-load failure, shared loader ownership, and retryable failures.

### Cost evidence

The benchmark is a behavioral model of no cache, the #8274 lookup-only/copy cache, and this PR. It uses five-run median latency and reports producer calls, payload allocations, peak/retained payload bytes, ready hits, and coalesced results.

For 8 concurrent callers, an 8 MiB output, and 2 ms synthetic COMGR time:

| Strategy | Phase | Latency | Producers | Allocations | Peak payload | Retained payload | Hit/coalesced |
|---|---|---:|---:|---:|---:|---:|---:|
| no cache | cold | 4530 us | 8 | 8 | 64 MiB | 64 MiB | 0 |
| #8274 model | cold | 8211 us | 8 | 16 | 128 MiB | 72 MiB | 0 |
| reader single-flight | cold | 3833 us | 1 | 1 | 8 MiB | 8 MiB | 7 coalesced |
| #8274 model | warm | 3617 us | 0 | 8 | 72 MiB | 72 MiB | 8 hits |
| reader single-flight | warm | 1379 us | 0 | 0 | 8 MiB | 8 MiB | 8 hits |

The proof boundary and lower-bound argument are documented in `CACHE_EVIDENCE.md`: for one cold exact key, every correct design needs at least one producer completion and one live `S`-byte result. This path attains those exact counts. The claim is count/asymptotic optimality under that model, not universal constant-time optimality.

### Build validation

- `hsa-runtime64` full build passed.
- `hotswap_rewrite`, `hotswap_cache`, and `hotswap_cache_benchmark` built.
- Focused `hotswap_rewrite` and `hotswap_cache` CTests passed.
- `git diff --check` passed.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap_cache.cpp
```diff
@@ -0,0 +1,747 @@
...
+      tracker.live_source_snapshot_bytes.load(std::memory_order_relaxed),
+      std::memory_order_relaxed);
+  tracker.peak_live_output_bytes.store(tracker.live_output_bytes.load(std::memory_order_relaxed),
+                                       std::memory_order_relaxed);
+}
+#endif
+
+namespace {
+std::once_flag g_process_retarget_cache_once;
+std::unique_ptr<ContentRetargetCache> g_process_retarget_cache;
```
cfreeamd, line 737: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** `g_process_retarget_cache` is a namespace-scope `std::unique_ptr`, so it is destroyed during `__cxa_finalize` at process exit / `dlclose`. If a background loader thread still touches the process cache during teardown (e.g. RCCL/CLR multi-device shutdown), that is a use-after-free window. This is not a regression — the previous `g_retarget_cache` map + mutex had the same exposure — but since this is now explicitly a process-wide singleton, consider intentionally leaking it (raw `new`, never reset, as the runtime does for other process-lifetime state) to avoid static-destruction-order hazards. Non-blocking.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap_cache.cpp
```diff
@@ -0,0 +1,747 @@
+////////////////////////////////////////////////////////////////////////////////
+//
+// The University of Illinois/NCSA
+// Open Source License (NCSA)
+//
+// Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved.
```
cfreeamd, line 6: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** This new runtime file uses the NCSA/UIUC license block, which matches the surrounding ROCr runtime convention. Repo policy for new files leans toward an `SPDX-License-Identifier` header — defer to maintainers on whether SPDX is now required for new runtime sources. Non-blocking.

cfreeamd (commented):

Reviewed the ROCr hotswap retarget-cache refactor. The concurrency core in `hotswap_cache.cpp` looks correct: table→bucket lock order with no reverse edge (sweep uses try_lock, SnapshotMetrics keeps the same order), correct publish/wait happens-before with waiters retaining a `shared_ptr` to the flight, generation-recheck source interning that keeps hash collisions distinct via exact `memcmp` identity, and exception mapping verified by the failure/`bad_alloc` tests. Ownership move from the executable-handle keepalive to `LoadedCodeObjectImpl::elf_owner` (aliasing `shared_ptr`) is a clean improvement. `amd_hsa_loader.cpp`, `hsa.cpp`, `executable.cpp/.hpp`, and `hotswap.cpp` call-site plumbing look correct.

Note: the two auto-checklist flags are false positives — the new tests are standalone CTest executables (like `hotswap_rewrite`), not `rocrtst64` gtest suites, so `test_categories.yaml`/`platform_config.yaml` need no update; and the license header follows the existing NCSA runtime convention. Two non-blocking inline notes below.

suryajasper:

PR #8745, opened 2026-07-17
feat(amdsmi): add feature-flagged WSL/WDDM GPU backend seam

marifamd:
## Motivation

Enable AMD SMI to run under WSL2, where there is no `/sys/class/drm` or
`/dev/kfd` and the GPU is reached through WDDM D3DKMT via `/dev/dxg`, **without**
forking the tool or scattering `if (wsl) ... else ...` through every public
function.

This PR lands the infrastructure for that as an experimental, default-off
feature so it can be merged and iterated on without affecting native builds, CI,
or existing users. The real D3DKMT data path is brought up incrementally behind
the same seam.

## Technical Details

Single-backend seam, one decision point:

- `AMDSmiWslBackend` (new `include/amd_smi/impl/amd_smi_wsl_backend.h`,
  `src/amd_smi/amd_smi_wsl_backend.cc`) — one class the dispatcher delegates to.
  A GPU stays a single device type; only the transport varies. Unsupported
  queries return `AMDSMI_STATUS_NOT_SUPPORTED`.
- `AMDSMI_WSL_INTERCEPT(...)` — a single line placed right after
  `AMDSMI_CHECK_INIT()` in each hooked function; it is the only WSL/native
  branch. Hooked as a demo slice: `amdsmi_get_gpu_asic_info`,
  `amdsmi_get_gpu_board_info`, `amdsmi_get_power_info`, `amdsmi_get_temp_metric`.

Two-layer gate so nothing changes by default:

- **Compile-time:** CMake `ENABLE_WSL_BACKEND` (default **OFF**) drives
  `#cmakedefine AMDSMI_ENABLE_WSL_BACKEND`. When off, the intercept macro expands
  to nothing and the WSL source is not compiled — native TUs are unchanged.

The current backend returns synthetic mock data so the seam runs on a native
host; the production D3DKMT implementation replaces the method bodies with no
change to the dispatcher. See the new how-to page
(`docs/how-to/amdsmi-wsl-mode.md`) for build/run/impact details, including the
impact-on-existing-scripts guidance (no API/CLI/JSON changes; unsupported fields
report `N/A`).

Remaining work is tracked as FUTURE WORK (real D3DKMT port, re-enable
`dxgkrnl` auto-detect once real data replaces the mock, handle→adapter mapping,
extend to the remaining ~42 supported functions, an `ENABLE_WSL_BACKEND=ON` CI
job, Python-level tests).

## Issue Tracking

JIRA ID: ROCM-28065

## Test Plan

- Build with `-DENABLE_WSL_BACKEND=ON -DBUILD_TESTS=ON` and run the new
  `WslBackend*` gtests in `tests/amd_smi_test/functional/wsl_backend_test.cc`.
- Build with the flag OFF (default) and confirm the WSL define/source are absent
  and the native library builds with `-Werror`.

## Test Result

- `WslBackend*` gtests: **8/8 passed** (activation on/off, non-`1` values, mock
  asic/board/power returns, temp capability subset, null → `INVAL`).

## Submission Checklist


projects/amdsmi/src/CMakeLists.txt
```diff
@@ -238,6 +251,8 @@ install(
 # error. Build them, but keep them out of the installed dev headers.
 set(INSTALL_IMPL_HEADERS ${IMPL_HEADERS})
 list(FILTER INSTALL_IMPL_HEADERS EXCLUDE REGEX "_test")
+list(FILTER INSTALL_IMPL_HEADERS EXCLUDE REGEX "amd_smi_wsl_device")
+list(FILTER INSTALL_IMPL_HEADERS EXCLUDE REGEX "amd_smi_wsl_syms")
```
gabrpham, line 336: I imagine we don't want to include the IGPUBackend from amd_smi_gpu_backend.h as part of our public ABI contract yet, so I would think we would want to filter it out here too:
```suggestion
list(FILTER INSTALL_IMPL_HEADERS EXCLUDE REGEX "amd_smi_wsl_device")
list(FILTER INSTALL_IMPL_HEADERS EXCLUDE REGEX "amd_smi_wsl_syms")
list(FILTER INSTALL_IMPL_HEADERS EXCLUDE REGEX "amd_smi_gpu_backend")
```

projects/amdsmi/amdsmi_cli/amdsmi_init.py
```diff
@@ -54,8 +54,17 @@
...
+def check_wsl_dxg():
+    """Returns true if running under WSL2 with an AMD GPU (/dev/dxg present)."""
+    return Path("/dev/dxg").exists()
+
+
 def check_amdgpu_driver():
     """Returns true if amdgpu is found in the list of initialized modules"""
+    # WSL2: /dev/dxg is the WDDM device node used instead of amdgpu
+    if check_wsl_dxg():
+        return True
```
gabrpham, line 99: Does the WSL feature also apply to Linux guests on Hyper-V? If not, then this check would skip checking for the Linux amdgpu there, since /dev/dxg is related to Hyper-V stuff.

gabrpham (changes requested):
Please see comments below. There are a number of other things that might need to be addressed, but they're hidden behind the feature flag for now so it should be fine for the time being and we can bring that back up later.

projects/amdsmi/src/amd_smi/amd_smi.cc
```diff
@@ -2173,16 +2210,37 @@ amdsmi_status_t amdsmi_get_violation_status(amdsmi_processor_handle processor_ha
 
 amdsmi_status_t amdsmi_get_gpu_fan_rpms(amdsmi_processor_handle processor_handle,
                                         uint32_t sensor_ind, int64_t* speed) {
+#ifdef ENABLE_WSL_BACKEND
+  AMDSMI_CHECK_INIT();
+  amd::smi::AMDSmiGPUDevice* gpu_device = nullptr;
+  amdsmi_status_t r = get_gpu_device_from_handle(processor_handle, &gpu_device);
+  if (r != AMDSMI_STATUS_SUCCESS) return r;
```
adam360x: Unless I'm mistaken, I think style wise we've tried to stay away from these single line return

projects/amdsmi/src/amd_smi/amd_smi.cc
```diff
@@ -5264,6 +5427,38 @@ amdsmi_status_t amdsmi_get_gpu_process_list_by_pid(amdsmi_processor_handle* proc
     if (r != AMDSMI_STATUS_SUCCESS) continue;
 
     uint32_t gpu_index = gpu_device->get_gpu_id();
+#ifdef ENABLE_WSL_BACKEND
+    if (auto* b = gpu_device->backend()) {
+      std::vector<amdsmi_proc_info_t> processes;
+      if (b->GetProcessList(&processes) != AMDSMI_STATUS_SUCCESS) continue;
+      for (auto& proc_info : processes) {
+        auto& entry = pid_map[proc_info.pid];
```
adam360x: Don't we already have this exact proc behavior somewhere else in the file or repo? I would think it's the same thing and we could make a helper function to condense code

projects/amdsmi/src/amd_smi/amd_smi_wsl_device.cc
```diff
@@ -0,0 +1,780 @@
...
+
+// librocdxg handle — owned by TryPopulate, valid for the process lifetime.
+static void* g_rocdxg_handle = nullptr;
+
+// Helper: resolve one dlsym symbol; print and return false on failure.
+template <typename T>
+static bool bind_sym(void* handle, const char* name, T& fn) {
+  fn = reinterpret_cast<T>(dlsym(handle, name));
+  if (!fn) {
+    std::cerr << "[WSL] missing symbol: " << name << std::endl;
```
adam360x: Are we not using our regular log statements here and throughout this file?

projects/amdsmi/src/amd_smi/amd_smi_wsl_device.cc
```diff
@@ -0,0 +1,780 @@
...
+      num_xcc_(props.NumXcc),
+      unique_id_(props.UniqueID),
+      local_mem_size_(props.LocalMemSize),
+      bdf_(make_bdf(props)),
+      marketing_name_(marketing_name_from_hsa(props)) {}
+
+amdsmi_status_t WSLGPUBackend::TryPopulate(std::vector<AMDSmiSocket*>& sockets,
+                                           std::set<AMDSmiProcessor*>& processors) {
+  // Not WSL if /dev/dxg is absent.
+  if (access("/dev/dxg", F_OK) != 0) return AMDSMI_STATUS_NOT_SUPPORTED;
```
adam360x: Style choice but perhaps we could make this path a constant somewhere

projects/amdsmi/src/amd_smi/amd_smi_wsl_device.cc
```diff
@@ -0,0 +1,780 @@
...
+  info->vram_max_bandwidth = std::numeric_limits<decltype(info->vram_max_bandwidth)>::max();
+  return AMDSMI_STATUS_SUCCESS;
+}
+
+amdsmi_status_t WSLGPUBackend::GetMemoryTotal(amdsmi_memory_type_t mem_type, uint64_t* total) {
+  if (total == nullptr) return AMDSMI_STATUS_INVAL;
+  if (mem_type != AMDSMI_MEM_TYPE_VRAM && mem_type != AMDSMI_MEM_TYPE_VIS_VRAM)
+    return AMDSMI_STATUS_NOT_SUPPORTED;
+#ifdef AMDSMI_HAS_ROCDXG_SMI
+  amdsmi_status_t r = load_device_info();
```
adam360x, line 425: Do we want to consider static metric caches like in amd-smi here for these things that don't change?

projects/amdsmi/amdsmi_cli/amdsmi_init.py
adam360x: > `/dev/dxg` is created by `dxgkrnl` for **any** WDDM GPU (NVIDIA/Intel/AMD), so this returns `True`
> regardless of vendor even though the helper is named `..._dxg`/documented as "with an AMD GPU". It
> also short-circuits the `/sys/module/amdgpu/initstate == "live"` check, which breaks Hyper-V Linux
> guests that expose `/dev/dxg` while using the native `amdgpu` driver (per @gabrpham above).
>
> Separately: this path ships **unconditionally**, i.e. it's not behind `ENABLE_WSL_BACKEND` like the
> C++ backend, so it changes default CLI GPU-detection on non-WSL hosts and weakens the PR's
> "native builds unchanged / off by default" guarantee. Can we gate this on an explicit WSL signal
> **and** confirm AMD vendor before returning `True`?

projects/amdsmi/src/amd_smi/amd_smi_wsl_device.cc
```diff
@@ -0,0 +1,780 @@
...
+  }
+  return AMDSMI_STATUS_SUCCESS;
+}
+
+// -----------------------------------------------------------------------------
+// IGPUBackend method implementations
+// -----------------------------------------------------------------------------
+
+#ifdef AMDSMI_HAS_ROCDXG_SMI
+amdsmi_status_t WSLGPUBackend::load_device_info() const {
```
adam360x, line 312: > `load_device_info()` is `const` but lazily mutates the `mutable device_info_` / `device_info_loaded_`
> members with no synchronization, and the intercepts in `amd_smi.cc` return **before**
> `SMIGPUDEVICE_MUTEX(gpu_device->get_mutex())` is taken. Two threads calling e.g.
> `amdsmi_get_gpu_asic_info` / `amdsmi_get_gpu_board_info` on the same handle will race on this cache
> (and on the global `g_wsl_syms` / `g_wsl_active`). Please guard the lazy load (per-instance
> `std::mutex` or `std::once_flag`), or take the device mutex on the WSL path too.

projects/amdsmi/docs/how-to/amdsmi-wsl-mode.md
```diff
@@ -0,0 +1,104 @@
...
+
+To support this without maintaining a separate build of the tool, AMD SMI has an
+optional **WSL backend**. The public API, the CLI, and the Python bindings are
+unchanged. When the backend is active, the affected queries are answered from
+the WDDM path instead of DRM/sysfs; everything else behaves as before.
+
+```{note}
+The WSL backend is experimental and gated behind a build flag that is **off by
+default**. A default AMD SMI build and its packages are byte-for-byte the native
+tool. The current backend ships a mock data provider so the path can be
```
adam360x: > This note (and the PR description's "two-layer gate" with `AMDSMI_WSL_MODE=1` + "activation is
> explicit-only") no longer matches the code: `WSLGPUBackend::TryPopulate()` activates on
> `access("/dev/dxg", F_OK) == 0` with no env-var gate, and it calls the real `rocdxg_smi_*` telemetry
> rather than a mock. The `AMDSMI_WSL_INTERCEPT(...)` macro from the description also isn't present
> (the code uses inline `#ifdef` blocks). Could we update the doc/description to the actual
> auto-detect + real-telemetry behavior, or reintroduce the runtime gate if the "never silently
> synthetic" property is still intended?

projects/rocr-runtime/libhsakmt/src/dxg/librocdxg.ver
```diff
@@ -114,6 +114,21 @@ hsaKmtMemHandleFreePreserveMetadata;
 hsaKmtMemoryGetCpuAddr;
 hsaKmtMemoryCpuMap;
 hsaKmtGetAmdGPUDeviceFd;
+rocdxg_smi_get_device_count;
+rocdxg_smi_get_device_info;
+rocdxg_smi_get_bdf_info;
```
adam360x: > These six entries — `rocdxg_smi_get_bdf_info`, `rocdxg_smi_get_asic_info`,
> `rocdxg_smi_get_board_info`, `rocdxg_smi_get_vram_info`, `rocdxg_smi_get_driver_info`,
> `rocdxg_smi_get_vbios_info` — aren't defined in `rocdxg_smi.cpp` (only the aggregate
> `rocdxg_smi_get_device_info` is). They look left over from a per-getter design; harmless with GNU ld
> but misleading. Suggest removing them until they're actually implemented.

projects/amdsmi/amdsmi_cli/subcommands/metric.py
```diff
@@ -128,6 +128,10 @@ def metric_gpu(
                 args.base_board = base_board
             if gpu_board:
                 args.gpu_board = gpu_board
+            if partition:
+                args.partition = partition
+            if not hasattr(args, 'partition'):
```
adam360x: > This reads like a fix for the WSL case (partition unsupported → `args.partition` never set → later
> `AttributeError`), but it runs on every platform and isn't explained. What's the failure it guards
> against, and should it be scoped to that path rather than applied unconditionally here?

projects/amdsmi/src/amd_smi/amd_smi_wsl_device.cc
```diff
@@ -0,0 +1,780 @@
...
+  if (count == 0) return AMDSMI_STATUS_SUCCESS;
+
+  std::vector<rocdxg_smi_process_info_t> rocdxg_processes(count);
+  hstatus = g_wsl_syms.rocdxg_smi_enum_processes(node_id_, &count, rocdxg_processes.data());
+  if (hstatus != HSAKMT_STATUS_SUCCESS) return rocdxg_to_amdsmi_status(hstatus);
+
+  processes->reserve(count);
+  for (uint32_t i = 0; i < count; ++i) {
+    amdsmi_proc_info_t out = {};
+    out.pid = rocdxg_processes[i].process_id;
```
adam360x: > `out.pid` comes from `D3DKMTEnumProcesses`, i.e. **Windows host** PIDs, so `/proc/<pid>/exe` in the
> WSL guest won't resolve them and `name` will be empty for essentially every process. Either resolve
> the name through a WDDM/host mechanism or document that `name` is unavailable on this path.

projects/amdsmi/src/amd_smi/amd_smi_wsl_device.cc
```diff
@@ -0,0 +1,780 @@
...
+  if (!fn) {
+    std::cerr << "[WSL] missing symbol: " << name << std::endl;
+    return false;
+  }
+  return true;
+}
+
+// dlopen librocdxg.so and bind all required symbols into g_wsl_syms.
+static bool load_rocdxg() {
+  if (g_rocdxg_handle) return true;
```
adam360x, line 132: > `librocdxg` also globally exports `amdgpu_device_initialize`, `amdgpu_query_gpu_info`,
> `drmCommandWriteRead`, etc. Loading it `RTLD_GLOBAL` can interpose those into the process. Since every
> symbol is resolved explicitly via `dlsym` into `g_wsl_syms`, `RTLD_LOCAL` is safer:
>
> ```suggestion
>   g_rocdxg_handle = dlopen("librocdxg.so.1", RTLD_NOW | RTLD_LOCAL);
>   if (!g_rocdxg_handle)
>     g_rocdxg_handle = dlopen("librocdxg.so", RTLD_NOW | RTLD_LOCAL);
> ```

adam360x (changes requested):

cfreeamd (commented):

Scope: `projects/rocr-runtime/` only. The new `rocdxg_smi.{h,cpp}` WSL/WDDM SMI shim is well-structured and defensive (availability/null probes, `clear_out()` zeroing, snprintf-bounded string copies, PCIe-gen table bounded by `gen < 6`); the `dxcore_loader`/`thunks.h` optional-export wiring and the `debug.cpp`→`queues.cpp` stub relocations look correct. Inline notes below are cleanup + one ABI-validation question — no blocking correctness issue found on the ROCr side.

projects/rocr-runtime/libhsakmt/include/hsakmt/hsakmt.h
```diff
@@ -930,6 +930,25 @@ hsaKmtGetCoreDeviceInfo(
...
+hsaKmtGetCoreRuntimeInfo(
+    struct kfd_runtime_info *runtime_info // OUT
+    );
+
+/**
+  Get core device info from KFD.
+*/
+HSAKMT_STATUS
+HSAKMTAPI
+hsaKmtGetCoreDeviceInfo(
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** Duplicate declarations. At PR head, `hsaKmtGetCoreRuntimeInfo` is declared twice (line 922 and 938) and `hsaKmtGetCoreDeviceInfo` twice (line 928 and 947). Redeclaration compiles, but this is an unintended duplicate block — the same three symbols are also duplicated in `librocdxg.ver` and relocated in `debug.cpp`/`queues.cpp`, so this looks like a rebase/merge artifact. Drop the re-added block.

projects/rocr-runtime/libhsakmt/src/dxg/librocdxg.ver
```diff
@@ -72,12 +72,15 @@ hsaKmtGetTileConfig;
...
 hsaKmtSetMemoryUserData;
 hsaKmtGetQueueInfo;
+hsaKmtGetKernelQueueId;
 hsaKmtAllocQueueGWS;
 hsaKmtGetKernelQueueId;
 hsaKmtRuntimeEnable;
 hsaKmtRuntimeDisable;
 hsaKmtCheckRuntimeDebugSupport;
 hsaKmtGetRuntimeCapabilities;
+hsaKmtGetCoreRuntimeInfo;
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** Duplicate version-script entries: `hsaKmtGetKernelQueueId` (line 75 duplicates the pre-existing 77), `hsaKmtGetCoreRuntimeInfo` (82 duplicates 85), and `hsaKmtGetCoreDeviceInfo` (83 duplicates 86) each now appear twice. GNU ld warns on duplicate symbols in a version node — remove the three added duplicates.

projects/rocr-runtime/libhsakmt/src/dxg/rocdxg_smi.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+void copy_string(char* dst, const char* src) {
+  if (dst == nullptr) return;
+  std::memset(dst, 0, ROCDXG_SMI_MAX_STRING_LENGTH);
+  if (src != nullptr) {
+    std::snprintf(dst, ROCDXG_SMI_MAX_STRING_LENGTH, "%s", src);
+  }
+}
+
+struct D3DKMT_ENUMPROCESSES {
```
cfreeamd, line 103: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** `struct D3DKMT_ENUMPROCESSES { LUID AdapterLuid; uint64_t Buffer; uint64_t BufferCount; }` is a hand-rolled local definition passed straight to the `D3DKMTEnumProcesses` thunk. Its field order/size/packing must match the dxgkrnl/WSL ABI exactly or the ioctl will misread the buffer. Is this layout validated against the canonical DXGKMT header? A mismatch would silently corrupt process enumeration rather than fail.

projects/rocr-runtime/libhsakmt/src/dxg/rocdxg_smi.cpp
```diff
@@ -0,0 +1,532 @@
...
+      copy_string(info->vbios.part_number, vbios.part_number);
+      copy_string(info->vbios.build_date,  vbios.date);
+    }
+  }
+
+  // Cache
+  {
+    uint32_t idx = 0;
+    if (wdev->GetL1CacheSize() > 0)
+      info->cache.cache[idx++] = {wdev->GetL1CacheSize() / 1024, 1, 0x2, 2, 0};
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** The L1 cache entry is initialized `{GetL1CacheSize()/1024, 1, 0x2, 2, 0}` — `num_cache_instance = 0` — while the L2/L3 entries below use `1`. Zero instances for a populated cache level looks like a copy/paste slip; likely should be `1` (or the true instance count).

projects/rocr-runtime/libhsakmt/include/hsakmt/rocdxg_smi.h
```diff
@@ -0,0 +1,205 @@
+/*
+ * Copyright (c) Advanced Micro Devices, Inc. All rights reserved.
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** New file — use the SPDX license header form rather than the legacy MIT permission block:
```
//
```
(Same applies to the new `rocdxg_smi.cpp`.)

fcui-amd:
AMD SMI currently uses dlopen() to access the DXG device through librocdxg.so, adding another layer of indirection. Could AMD SMI communicate with /dev/dxg directly instead? In addition, librocdxg.so loads libhsa-runtime64.so during initialization, which seems unnecessary for AMD SMI.

On native Linux, AMD SMI accesses devices through sysfs and can operate independently of the rest of the ROCm stack. On WSL, this additional indirection introduces extra runtime dependencies.

Alternatively, if the KFD thunk APIs can provide all required SMI functionality, could AMD SMI use the same thunk-based interface on native Linux as well? This could provide a more consistent architecture across native Linux and WSL.

bill-shuzhou-liu:
> AMD SMI currently uses dlopen() to access the DXG device through librocdxg.so, adding another layer of indirection. Could AMD SMI communicate with /dev/dxg directly instead? In addition, librocdxg.so loads libhsa-runtime64.so during initialization, which seems unnecessary for AMD SMI.
> 
> On native Linux, AMD SMI accesses devices through sysfs and can operate independently of the rest of the ROCm stack. On WSL, this additional indirection introduces extra runtime dependencies.
> 
> Alternatively, if the KFD thunk APIs can provide all required SMI functionality, could AMD SMI use the same thunk-based interface on native Linux as well? This could provide a more consistent architecture across native Linux and WSL.

If AMDSMI were to communicate directly with /dev/dxg, there are essentially two possible approaches.

The first would be to bypass libdxcore.so and issue the required ioctls directly to the device. In practice, this would require reverse engineering the protocol used between libdxcore.so and /dev/dxg. Beyond the engineering effort involved, this is not a viable long-term solution because the protocol is private to Microsoft and could change at any time, potentially breaking compatibility without notice.

The second option would be to use libdxcore.so directly. However, this would reintroduce the Windows SDK dependencies that we are actively trying to avoid, since interaction would need to be implemented through the D3DKMT* APIs.

Even if we removed librocdxg.so from the picture, we would still depend on WKMI (libthunk_proxy.so), which contains private driver interfaces and headers that cannot be open-sourced. This library is already linked with librocdxg.so, and our current approach effectively adds an abstraction layer around these private dependencies.

Yes, from an AMDSMI perspective, the ideal solution would be for the driver to expose the required functionality through sysfs in WSL, similar to how AMD SMI operates on BM Linux. My understanding is that the lack of such an interface is a limitation of the current WSL environment, and the existing design is largely an adaptation to those constraints.

projects/amdsmi/amdsmi_cli/amdsmi_init.py
```diff
@@ -71,8 +71,17 @@
...
 
+def check_wsl_dxg():
+    """Returns true if running under WSL2 with an AMD GPU (/dev/dxg present)."""
+    return Path("/dev/dxg").exists()
+
+
 def check_amdgpu_driver():
     """Returns true if amdgpu is found in the list of initialized modules"""
+    # WSL2: /dev/dxg is the WDDM device node used instead of amdgpu
+    if check_wsl_dxg():
```
adam360x, line 98: > `/dev/dxg` is created by `dxgkrnl` for **any** WDDM GPU (NVIDIA/Intel/AMD), so this returns `True`
> regardless of vendor — the `check_wsl_dxg()` docstring (line 75) even claims "with an AMD GPU" but
> nothing checks the vendor. It also short-circuits the `/sys/module/amdgpu/initstate == "live"` check,
> which breaks Hyper-V Linux guests that expose `/dev/dxg` while using the native `amdgpu` driver.
> Finally, this ships **unconditionally** (not behind `ENABLE_WSL_BACKEND`), so it changes default CLI
> GPU-detection on non-WSL hosts — contrary to the "native builds unchanged / off by default" goal.
> Now that the PR is ready-for-review with a companion enable PR (ROCm/TheRock#6885), can we gate this
> on an explicit WSL signal **and** confirm an AMD adapter before returning `True`?

projects/amdsmi/docs/how-to/amdsmi-wsl-mode.md
adam360x: > The doc is now out of date: the backend calls real `rocdxg_smi_*` telemetry, but line 24 still says
> *"The current backend ships a mock data provider"* and the Limitations bullet (lines 102–103) still
> says *"The shipped backend is a mock … values are placeholders."* Line 81 also references a removed
> runtime gate — *"or never set `=1`"* — which no longer exists (activation is `/dev/dxg` auto-detect).
> Please reconcile this file, `CHANGELOG.md` ("Ships a mock data provider…"), and the PR description
> with the shipped behavior.

projects/amdsmi/amdsmi_cli/subcommands/metric.py
```diff
@@ -128,6 +128,10 @@ def metric_gpu(
                 args.base_board = base_board
             if gpu_board:
                 args.gpu_board = gpu_board
+            if partition:
+                args.partition = partition
+            if not hasattr(args, "partition"):
```
adam360x: > This reads like a fix for the WSL case (partition unsupported → `args.partition` never set → later
> `AttributeError`), but it runs on every platform and isn't explained. What failure does it guard,
> and should it be scoped to that path rather than applied unconditionally here?

projects/rocr-runtime/libhsakmt/src/dxg/librocdxg.ver
```diff
@@ -72,12 +72,15 @@ hsaKmtGetTileConfig;
 hsaKmtQueryPointerInfo;
 hsaKmtSetMemoryUserData;
 hsaKmtGetQueueInfo;
+hsaKmtGetKernelQueueId;
```
adam360x: > Duplicate version-script entries: `hsaKmtGetKernelQueueId` (line 75 duplicates the pre-existing 77),
> `hsaKmtGetCoreRuntimeInfo` (82 duplicates 85), `hsaKmtGetCoreDeviceInfo` (83 duplicates 86). GNU ld
> warns on duplicate symbols in a version node — these look like merge artifacts (the matching
> `hsakmt.h` twin decls were fixed in `17e7719`, but these `.ver` dupes remain). Drop the three added
> lines:
>
> Line 75:
> ```suggestion
> ```
> Lines 82–83:
> ```suggestion
> ```

adam360x (changes requested):

projects/rocr-runtime/libhsakmt/src/dxg/librocdxg.ver
bill-shuzhou-liu: updated

adam360x (approved):

gabrpham (approved):

chrispaquot (approved):

kentrussell (approved):

projects/rocr-runtime/libhsakmt/CMakeLists_wsl.txt
```diff
@@ -195,6 +197,12 @@ install ( TARGETS ${ROCDXG_TARGET} EXPORT ${ROCDXG_TARGET}Targets
 # Install public headers
 #install ( DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include/${ROCDXG_TARGET} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
 #  COMPONENT dev PATTERN "*drm*" EXCLUDE )
+install ( FILES
+  ${CMAKE_CURRENT_SOURCE_DIR}/include/hsakmt/hsakmt.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/include/hsakmt/hsakmttypes.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/include/hsakmt/rocdxg_smi.h
+  DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/hsakmt
+  COMPONENT dev )
```
ScottTodd, line 205: This broke package tests in TheRock: https://github.com/ROCm/TheRock/pull/7005#discussion_r3686421681

With this change these files are provided by _both_ the core-runtime and wsl-rocdxg artifacts when files must be uniquely provided by only a single artifact.

The comments further down in this file also suggest this should not be needed:
```cmake
# Since librocdxg.pc and libhsakmt.pc are installed to the same pkgconfig directory,
# we can directly use libhsakmt's header file path in the includedir.
# This allows librocdxg to reference the same header files as libhsakmt without
# duplicating header installation.
configure_file ( librocdxg.pc.in librocdxg.pc @ONLY )
```

ScottTodd: I'd vote to revert. We're going to block rocm-systems updates in TheRock until this is fixed and will carry patch files for a smaller number of changes that are important enough.

Those artifact structure tests are not running in rocm-systems currently, so we were relying on code organization and review to catch the issue. I'll see what it will take to add test coverage here (the new multi-arch workflows can do this, but they aren't in use here yet)

marifamd: Revert it @dennyiriawan This should not have made it in

dennyiriawan: Agreed, we should have another trial bump to iron out all the issues before merging it

PR #8748, opened 2026-07-17
fix(clr): size single-grid cooperative sync alloc to full MGSyncInfo

neoblizz:
## Motivation

Fix an out-of-bounds write in the single-grid cooperative-groups launch path. `kSGInfoSize` is used to size the allocation that backs a `Device::MGSyncInfo`, but it was set to `sizeof(MGSyncData)` (8 bytes) while the code writes a full `MGSyncInfo` (~48 bytes). The write currently lands inside the shared kernarg pool's free region, so it only survives on allocator slack and can corrupt adjacent kernel arguments (or lose barrier fields across a chunk-boundary flush) depending on allocation order, alignment, and pool size.

## Technical Details

In `projects/clr/rocclr/device/rocm/rocvirtual.cpp`, the `HiddenMultiGridSync` case for `singleGridSync` does:

```cpp
syncInfo = reinterpret_cast<Device::MGSyncInfo*>(allocKernArg(Device::kSGInfoSize, 64));
```

and then populates every `MGSyncInfo` field (mgs, grid_id, num_grids, prev_sum, all_sum, sgs, num_wg). `allocKernArg` -> `ManagedBuffer::Acquire` is a bump allocator over a single pre-allocated pool: it advances the cursor by the requested size only. Passing `kSGInfoSize = sizeof(MGSyncData) = 8` reserves 8 bytes but writes ~48, so the trailing ~40 bytes are written into space the allocator still considers free.

```cpp
// projects/clr/rocclr/device/device.hpp
- static constexpr size_t kSGInfoSize = kMGSyncDataSize;
+ static constexpr size_t kSGInfoSize = sizeof(MGSyncInfo);
```

No struct layout change, so no ABI or device-library coupling.

## Issue Tracking

GitHub issue: https://github.com/ROCm/rocm-systems/issues/9722

## Test Plan

- Build clr and confirm compilation.
- Run a single-device cooperative kernel (`hipLaunchCooperativeKernel` with `this_grid().sync()`) and verify correct execution.

## Test Result

Compiles ok, and runs `this_grid().sync()` ok.

## Submission Checklist


neoblizz:
> 🧪 **Unit Test**	❌ Fail	**Error:** Source/code files changed without an accompanying unit test.
> **Expected:** add at least one test file named like `test_<name>.py` / `test_<name>.cpp` (or `<name>_test.*`).
> **Current:** code file(s) changed: `projects/clr/rocclr/device/device.hpp`; no test file found

A test would look something like this, but the current gtest harness doesn't include those dirs so I'll have to make cmake changes as well for a one-line change:

```cpp
#include <cstddef>
#include <cstdint>

namespace {

// The single-grid path writes a full MGSyncInfo into a kSGInfoSize allocation.
// The allocation must therefore cover the whole struct.
static_assert(kSGInfoSize >= sizeof(MGSyncInfo),
              "kSGInfoSize must cover the whole MGSyncInfo written by the "
              "single-grid cooperative launch path");

// Guard against silently reverting to the old (buggy) 8-byte size.
static_assert(kSGInfoSize != kMGSyncDataSize,
              "kSGInfoSize must not be sizeof(MGSyncData); that under-sizes the "
              "single-grid cooperative sync allocation");

}  // namespace

int main() { return 0; }
```

Instead, we can add a `static_assert` in the code itself.

chrispaquot (approved):

chrispaquot (approved):

sourabhuday (approved):

PR #6240, opened 2026-05-19
[rocpfiler-systems] Unify control logic with different triggers

marantic-amd:
## Motivation

Pause/resume gating was inconsistent across control surfaces:
- **roctx markers** gated at source — visible gaps in the cached Perfetto/RocPD output.
- **`ROCPROFSYS_TRACE_DELAY` / `TRACE_DURATION` / `SAMPLING_DURATION`** only filtered output recording, not production. `TRACE_DELAY` in particular did not halt GPU producers — kernel_dispatch slices leaked into the cached trace for the entire delay window.
- **`ROCPROFSYS_SAMPLING_DELAY`** was a settings entry with no consumer.

This PR introduces a single `core/control/session` to which all triggers (roctx, time-window) attach and to which all subsystems (rocm, sampling, mpi, kokkos, process_sampler, etc.) subscribe. `tool_init` now consults the session state before starting rocprofiler-sdk contexts so the cached output reflects the configured window — TRACE_DELAY produces an honest gap, SAMPLING_DELAY is wired through the same plumbing, and roctx region-filter sessions correctly suppress non-region data.

### New module: `source/lib/core/control/`

A standalone control plane that owns pause/resume state and dispatch:

- **`session`** — the central state machine. Holds a vector of `vote_entry` snapshots (one per attached trigger) and a vector of `subscriber`s. Maintains a per-scope `std::array<std::atomic<bool>, SCOPE_COUNT> m_active` for lock-free `is_active(scope)` reads on the hot path.
- **`trigger`** — abstract interface: name, `event_scope`, `initial_vote`. Concrete triggers (`triggers::roctx`, `triggers::time_window<Clock>`) publish vote changes via `session::publish(*this, new_vote)`.
- **`subscriber`** — value type with `on_pause` / `on_resume` callbacks, a `name`, and a `scopes` list (default `{global}`; subsystems affected by narrower scopes like `sampling_only` list them in addition to global).
- **`vote_entry`** — value-semantic snapshot of a trigger's current vote; the session never holds a pointer to the trigger object.
- **`clocks::steady`** (production) + **`clocks::manual`** (test-only) — both satisfy a duck-typed `clock` contract consumed by `time_window<Clock>`.

This module replaces the deleted `library/rocprofiler-sdk/trace_control.{cpp,hpp}` (212 LOC). It depends on nothing in `library/` or `rocprof-sys/` — clean lower-layer placement.

### Source-side gating

[...]

PR #8749, opened 2026-07-17
feat(rocprofiler-systems): Add basic foundation for unified trace control

marantic-amd:
## Motivation

- At the moment different mechanisms exist inside our codebase which in their own way allow user to control the flow of the tracing/profiling session. This leads to those mechanisms not being coherent and flaky. This PR is first in series that sets a foundation for unified approach in trace/profiler control with intention to resolve surrounding bugs and flakiness, and make a centralized point for this feature.

## Technical Details

- Renames the old `trace_control` class to `control::session`; introduces `control::subscriber` (a declarative pause/resume registration replacing the old monolithic callback-pair fan-out) and the lock-free `is_active()` atomic. Session's mutating API is named for what it does to a trigger or a vote: `register_trigger()`/`unregister_trigger()`, `publish_vote()`, `is_active_excluding_trigger()`.
- Extracts ROCtx's region-filter and user-pause state into `control::triggers::roctx`, decoupled from `session`
- Switches internal vote storage to value-based `vote_entry` snapshots so the session never holds a pointer to a trigger. Trigger and session lifetimes are fully decoupled.
- Adds the `ClockPolicy` concept, `clocks::steady` (production), and the `triggers::time_window<Clock>` template, purely additive, not wired into the runtime within this PR. The test-only virtual clock (`control::testing::mock_clock`) is intended to be used for testing purposes once the system is wired in codebase.

## JIRA ID

JIRA ID :  AIPROFSYST-91

## Test Plan

- `ctest --test-dir build/debug-optimized -R 'control|session|roctx'` (targeted regression net for this PR's touched subsystems).
- Full GTest unit suite (`bin/rocprof-sys-unit-tests`), unfiltered, to catch any regression outside the targeted set.

## Test Result

- Build: clean, whole project, no warnings introduced
- Full CTest suite passed
- All unit tests pass

## Submission Checklist


projects/rocprofiler-systems/source/lib/core/control/tests/mock_clock.hpp
mradosav-amd, line 1: This doesn't make sense to be here. As well, if we are not using it, we should not add it at all. Add unit tests for those classes or remove this file.

projects/rocprofiler-systems/source/lib/core/control/triggers/time_window.hpp
```diff
@@ -0,0 +1,115 @@
...
+    /// Spawn the worker thread that advances the window through delay and
+    /// duration phases. Idempotent: a second call is a no-op. Not safe to
+    /// call concurrently with stop() from a different thread than the one
+    /// serializing start()/stop() calls (guarded via m_lifecycle_mutex).
+    void start()
+    {
+        std::scoped_lock const lk{ m_lifecycle_mutex };
+        if(!has_window()) return;
+        if(m_thread.joinable()) return;
+        m_thread = std::thread{ [this]() { worker(); } };
```
mradosav-amd, line 56: add don't scoped_thread guard from `runtime.hpp` this will cause to sample our own thread.

projects/rocprofiler-systems/source/lib/core/control/triggers/roctx.hpp
```diff
@@ -0,0 +1,60 @@
...
+{
+class session;
+}
+
+namespace rocprofsys::control::triggers
+{
+class roctx : public trigger
+{
+public:
+    roctx(session& sess, std::string_view trace_regions);
```
mradosav-amd: for time_window you are deleting copy/move ctors. here that is missing.

projects/rocprofiler-systems/source/lib/core/control/session.hpp
```diff
@@ -0,0 +1,68 @@
...
+    /// to reflect the initial state. Subscribers default to "running", so
+    /// only the paused-initial case needs to be broadcast.
+    void force_initial_pause();
+
+    [[nodiscard]] bool is_active() const noexcept
+    {
+        return m_active.load(std::memory_order_relaxed);
+    }
+
+    /// True iff every trigger except @p name has voted active or abstain.
```
mradosav-amd:
```suggestion
    /// True if every trigger except @p name has voted active or abstain.
```

projects/rocprofiler-systems/source/lib/core/control/session.hpp
```diff
@@ -0,0 +1,68 @@
...
+    /// Used where a trigger's own write decision must
+    /// also respect other triggers' votes without double-counting its own.
```
mradosav-amd:
```suggestion
    /// also respect other triggers votes without double-counting its own.
```

projects/rocprofiler-systems/source/lib/core/control/session.hpp
```diff
@@ -0,0 +1,68 @@
...
+    [[nodiscard]] bool is_active_excluding_trigger(std::string_view name) const noexcept;
+
+private:
+    std::vector<vote_entry> m_votes;
```
mradosav-amd: maybe it's better to be `std::unordered_map` since we are always searching by name?

projects/rocprofiler-systems/source/lib/core/control/session.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+    if(was_active == now_active) return;
+    if(now_active)
+        notify_resume();
+    else
+        notify_pause();
+}
+
+vote_entry&
+session::find_or_insert_locked(std::string_view name, vote v)
```
mradosav-amd: For me this is misleading name..

projects/rocprofiler-systems/source/lib/core/control/trigger.hpp
mradosav-amd, line 1: trigger and subscriber are kinda part of session. you are not including subscriber when using it, you are just including session.

projects/rocprofiler-systems/source/lib/core/control/session.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+    [[nodiscard]] bool is_active() const noexcept
+    {
+        return m_active.load(std::memory_order_relaxed);
+    }
+
+    /// True iff every trigger except @p name has voted active or abstain.
+    /// Used where a trigger's own write decision must
+    /// also respect other triggers' votes without double-counting its own.
+    [[nodiscard]] bool is_active_excluding_trigger(std::string_view name) const noexcept;
```
mradosav-amd: is this called?

projects/rocprofiler-systems/source/lib/core/control/session.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+void
+session::force_initial_pause()
+{
+    if(is_active()) return;
+    notify_pause();
+}
+
+void
+session::publish_vote(const trigger& trig, vote new_vote)
```
mradosav-amd: Why do we have `publish_vote` when we have `register_trigger` and `unregister_trigger`?

Can we configure trigger upon registration to know how to publish vote?

projects/rocprofiler-systems/source/lib/rocprof-sys/library/rocprofiler-sdk.hpp
```diff
@@ -42,8 +42,8 @@ pause();
 void
 resume();
 
-std::shared_ptr<control::trace_control>
-get_trace_controller();
+std::shared_ptr<control::session>
+get_session();
```
mradosav-amd: Right now, when we are shifting towards clock controlling session. I think it's not good idea that session is owned by rocprofiler-sdk.

projects/rocprofiler-systems/source/lib/core/control/trigger.hpp
```diff
@@ -0,0 +1,40 @@
...
+#include <string_view>
+
+namespace rocprofsys::control
+{
+enum class vote
+{
+    abstain,
+    active,
+    paused
+};
```
mradosav-amd: maybe `action` ?

```suggestion
enum class action
{
    skip,
    trace,
    pause
};
```

projects/rocprofiler-systems/source/lib/core/control/clock.hpp
mradosav-amd, line 1: This should be hidden inside steady clock.

mradosav-amd (commented):
I get the idea. Please address those few stuff.

projects/rocprofiler-systems/source/lib/core/control/session.hpp
marantic-amd: Typo

projects/rocprofiler-systems/source/lib/core/control/session.hpp
marantic-amd: Sure, I think it makes sense to use that here

projects/rocprofiler-systems/source/lib/core/control/trigger.hpp
marantic-amd: This goes both for this and for the comment regarding the `mock_clock`. This is the split of the drafted PR #6240  so I found that this made sense to be here since it will be all wired in within the upcoming PRs. TO me it looked like logically it fits within this PR

projects/rocprofiler-systems/source/lib/core/control/session.cpp
marantic-amd: If I understand correctly, your suggestion is to triggers themself know how to publish their vote toward session? If so, I think it makes sense to be that way, since `trigger` is the origin of the "vote"

projects/rocprofiler-systems/source/lib/core/control/trigger.hpp
marantic-amd: This really depends how you form it. If we look at this as refering to session, then `session active` or `session pause` make sense. If the aim is globally then, I agree, your suggestion makes sense.

projects/rocprofiler-systems/source/lib/core/control/session.cpp
mradosav-amd: Yes. Session should give them way to to publish "vote" upon registration. 
We want triggers to control session. This way we leave session _vulnerable_ ..

projects/rocprofiler-systems/source/lib/core/control/clock.hpp
marantic-amd: The additional clocks that are introduced in the second PR that builds upon this one, which justifies the `clock.hpp` being a separate header

projects/rocprofiler-systems/source/lib/rocprof-sys/library/rocprofiler-sdk/tests/test_roctx_client.cpp
```diff
@@ -102,56 +103,79 @@ TEST_F(roctx_client_test, constructor_creates_controller)
...
 
 TEST_F(roctx_client_test, constructor_without_region_filter)
 {
     using namespace rocprofsys::rocprofiler_sdk;
 
-    const roctx_client_config        config{ true, true, true, false, "" };
-    roctx_client<mock_marker_policy> client(config);
-    EXPECT_NE(client.get_controller(), nullptr);
-    EXPECT_FALSE(client.get_controller()->region_filter_active());
+    const roctx_client_config        config{ .pause_resume_enabled   = true,
```
dgaliffiAMD: It is looking really good with the initializers.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/rocprofiler-sdk.cpp
```diff
@@ -2858,12 +2825,10 @@ flush_counter_tracks_to_zero(rocprofiler_timestamp_t timestamp)
 
 }  // namespace
```
dgaliffiAMD, line 2902: This one was found by review agent:

**Static destruction-order use-after-destruction of the global session**

  This hunk removes the deterministic teardown that used to keep the roctx client safe — old `shutdown()` called `roctx_client->get_controller()->shutdown()` before rocprofiler finalization. Nothing replaces it, which opens a destruction-order hazard:

  - `g_roctx_client` (anon-namespace global in this TU) holds a `control::session&` bound to `g_control_session`, which lives in a **different** translation unit (`library.cpp`).
  - It's never `reset()`, so it gets its real destructor at process exit / `dlclose`, where cross-TU static destruction order is **unspecified**.
  - If `g_control_session` is destroyed first, `~roctx_client → ~unique_ptr<roctx> → roctx::~roctx()` runs `m_session.unregister_trigger(*this)`, locking `m_actions_mutex` and mutating `m_actions` on a **destroyed** `session` → use-after-destruction (UB).

  Sanitizer CI is green because finalize normally runs first and `g_control_session.shutdown()` clears the maps — but `shutdown()` neither destroys the session nor detaches the trigger, so the teardown-order hazard stays latent.

  **Suggested fix** — release the client here, while the session is still alive:

  ```cpp
  void shutdown()
  {
      if(tool_data && tool_data->client_id && tool_data->client_fini)
          tool_data->client_fini(*tool_data->client_id);
      g_roctx_client.reset();   // destroy the trigger while the session is still alive
  }
  ```

  Alternatively, hand out a registration handle from `session` whose destruction is a no-op once the session is gone. Note #9074 adds more session&-holding globals (g_trace_window etc.), but those are cleaned via stop_time_windows(); the roctx client is the one still relying on unspecified order.

projects/rocprofiler-systems/source/lib/core/control/triggers/roctx.cpp
```diff
@@ -0,0 +1,163 @@
+// Copyright (c) Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: MIT
+
+#include "roctx.hpp"
+
+#include "common/delimit.hpp"
+#include "core/control/session.hpp"
+
+#include "logger/debug.hpp"
+#include <spdlog/fmt/ranges.h>
```
dgaliffiAMD, line 11:
```suggestion
#include <spdlog/fmt/fmt.h>
#include <spdlog/fmt/ranges.h>
```
Fixes clang-tidy error: Use of undeclared identifier 'fmt' from line 31-32.

projects/rocprofiler-systems/source/lib/core/control/triggers/roctx.cpp
```diff
@@ -0,0 +1,163 @@
...
+    bool was_empty = false;
+    {
+        std::scoped_lock const lk{ m_mutex };
+        was_empty = m_active_range_ids.empty();
+        m_active_range_ids.insert(range_id);
+    }
+
+    if(was_empty)
+    {
+        m_in_region.store(true, std::memory_order_relaxed);
```
dgaliffiAMD: **`m_in_region` updated outside the lock that guards `m_active_range_ids`**

  **Files:** `core/control/triggers/roctx.cpp` — `on_range_start` (L47–64), `on_range_stop` (L66–92)

  **Problem:** `m_active_range_ids` is mutated under `m_mutex`, but the derived `m_in_region.store(...)` runs *after* the lock is released. The mutex orders the set operations; it does **not** order the stores that follow them. So, the store order can contradict the order the critical sections actually executed, breaking the invariant `m_in_region == !m_active_range_ids.empty()`.

  **Interleaving** (filtered region, starts empty):

  | Step | Thread A `on_range_start(1)` | Thread B `on_range_stop(1)` | set | pending store |
  |------|------------------------------|-----------------------------|-----|---------------|
  | 1 | lock: `was_empty=true`, insert 1, unlock | | `{1}` | A→true |
  | 2 | | lock: erase 1, `now_empty=true`, unlock | `{}` | B→false |
  | 3 | | `m_in_region.store(false)` | `{}` | — |
  | 4 | `m_in_region.store(true)` | | `{}` | — |

  Set ends empty, so `m_in_region` should be `false`, but A's store lands last and it sticks at `true`. `should_write_markers()` then returns `true` outside every target region → **markers written where they should be suppressed**. The mirror interleaving leaves it `false` inside an open region → **markers dropped**.

  This is a correctness desync, not UB — the relaxed atomics are individually race-free; it's the *invariant* between the flag and the set that is unprotected.

  **Secondary:** `refresh_state()` (L157–162) also recomputes and `publish_action()`s to the session outside the lock, so two concurrent calls can publish stale-vs-fresh actions in reverse order, latching the session's `"roctx"` action (which drives the pause/resume fan-out to *all* subscribers) wrong even after the flag is fixed.

  **Fix:** Move `m_in_region.store(...)` inside the `m_mutex` critical section so its write order matches the set-mutation order. To also close the publish-ordering half, compute the action under the same lock and pass it into `publish_action` (or hold `m_mutex` across `refresh_state()`). If holding across `refresh_state`, note the resulting lock order `roctx.m_mutex → session mutexes` and keep it one-directional.

  **Stack note:** #9433 revisits this exact gating formula and adds a "range opened while paused" test — natural home for the in-lock store.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/rocprofiler-sdk.cpp
```diff
@@ -2753,17 +2696,41 @@ tool_init(rocprofiler_client_finalize_t fini_func, void* user_data)
...
+                .value_or(std::string{}),
+            " ,;:\t\n");
+        const auto has_marker_domain =
+            (std::find(_domains.begin(), _domains.end(), "marker_api") !=
+                 _domains.end() ||
+             std::find(_domains.begin(), _domains.end(), "roctx") != _domains.end());
+        const auto roctx_traced_regions = config::get_trace_region();
+        const auto has_trace_regions    = !roctx_traced_regions.empty();
+
+        // Case 1: no marker domain and no trace regions — nothing to do
```
dgaliffiAMD: Nit-pick: Why is there a "Case 1" and not any other cases? Are they in subsequent PRs?

projects/rocprofiler-systems/source/lib/rocprof-sys/library.cpp
```diff
@@ -667,41 +670,31 @@ rocprofsys_init_tooling_hidden(void)
...
+                { &component::mpi_gotcha::pause, &component::mpi_gotcha::resume, "mpi" },
+                { &ucx_t::pause, &ucx_t::resume, "ucx" },
+                { &shmem_t::pause, &shmem_t::resume, "shmem" },
+                { &component::vaapi_gotcha::pause, &component::vaapi_gotcha::resume, "vaapi" },
+                { &::rocprofsys::pthread_gotcha::pause, &::rocprofsys::pthread_gotcha::resume, "pthread" },
+                { &component::numa_gotcha::pause, &component::numa_gotcha::resume, "numa" },
+                { &rocprofsys::kokkosp::pause, &rocprofsys::kokkosp::resume, "kokkos" },
+                { &process_sampler::pause, &process_sampler::resume, "process_sampler" },
+                { &invoke_external_pause_callbacks, &invoke_external_resume_callbacks, "external" },
+            } };
```
dgaliffiAMD:
```suggestion
            const auto subscribers = std::to_array<control::subscriber>({
                { .on_pause = &rocprofiler_sdk::pause, .on_resume = &rocprofiler_sdk::resume, .name = "rocm" },
                { .on_pause = &sampling::pause, .on_resume = &sampling::resume, .name = "sampling" },
                { .on_pause = &component::mpi_gotcha::pause, .on_resume = &component::mpi_gotcha::resume, .name = "mpi" },
                { .on_pause = &ucx_t::pause, .on_resume = &ucx_t::resume, .name = "ucx" },
                { .on_pause = &shmem_t::pause, .on_resume = &shmem_t::resume, .name = "shmem" },
                { .on_pause = &component::vaapi_gotcha::pause, .on_resume = &component::vaapi_gotcha::resume, .name = "vaapi" },
                { .on_pause = &::rocprofsys::pthread_gotcha::pause, .on_resume = &::rocprofsys::pthread_gotcha::resume, .name = "pthread" },
                { .on_pause = &component::numa_gotcha::pause, .on_resume = &component::numa_gotcha::resume, .name = "numa" },
                { .on_pause = &rocprofsys::kokkosp::pause, .on_resume = &rocprofsys::kokkosp::resume, .name = "kokkos" },
                { .on_pause = &process_sampler::pause, .on_resume = &process_sampler::resume, .name = "process_sampler" },
                { .on_pause = &invoke_external_pause_callbacks, .on_resume = &invoke_external_resume_callbacks, .name = "external" },
            });
```
nit-pick:
Using `std::to_array` to avoid the drift of the hardcoded `11`. Using designated initializers to avoid clang-tidy warning.

projects/rocprofiler-systems/source/lib/core/control/clock.hpp
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include <chrono>
+#include <cstdint>
```
dgaliffiAMD:
```suggestion

```

projects/rocprofiler-systems/source/lib/core/control/triggers/time_window.hpp
```diff
@@ -0,0 +1,120 @@
...
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include "core/control/clock.hpp"
+#include "core/control/session.hpp"
+#include "core/control/trigger.hpp"
+#include "core/state.hpp"
+
+#include <atomic>
```
dgaliffiAMD:
```suggestion
```
clang-tidy: unused include of `<atomic>`.

projects/rocprofiler-systems/source/lib/core/control/clock.hpp
dgaliffiAMD:
```suggestion
```
clang-tidy: unused include of `<cstdint>`.

projects/rocprofiler-systems/source/lib/core/control/session.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+#pragma once
+
+#include "subscriber.hpp"
+#include "trigger.hpp"
+
+#include <atomic>
+#include <mutex>
+#include <string>
+#include <string_view>
```
dgaliffiAMD, line 10:
```suggestion
```
clang-tidy: unused include of `<string_view>`.

projects/rocprofiler-systems/source/lib/core/control/trigger.hpp
```diff
@@ -0,0 +1,48 @@
...
+protected:
+    trigger() = default;
+
+    void bind_action_setter(action_setter setter) noexcept
+    {
+        assert(!m_set_action);
+        m_set_action = std::move(setter);
+    }
+
+    void publish_action(action act) const { m_set_action(act); }
```
dgaliffiAMD:
```suggestion
    void publish_action(action act) const 
    { 
        assert(m_set_action);
        m_set_action(act); 
    }
```
Is it worth having an assert or null check here too? Current implementations set `bind_action_setter` right in the constructor, but it will protect future implementations.

projects/rocprofiler-systems/source/lib/rocprof-sys/library.cpp
marantic-amd: This would work with this PR. The second one introduces the scope of the pause (which can be sampling only) but that can be dealt  there

projects/rocprofiler-systems/source/lib/core/control/session.hpp
marantic-amd: The string view is used in PR2 in this stack, but you are right to flag it as unused here.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/rocprofiler-sdk.cpp
marantic-amd: This can be resolved both here or either in https://github.com/ROCm/rocm-systems/pull/9074. I can add the fix here and then I will rebase all consecutive PRs on top of the changes, after I go through all your comments.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/rocprofiler-sdk.cpp
marantic-amd: To be honest, this is the one cleanup I missed. There were some additional checks that eventually I found redundant but the comment was left out of that cleanup. It can  be removed completely, it brings just more noise not clarification 😁

projects/rocprofiler-systems/source/lib/core/control/subscriber.hpp
```diff
@@ -0,0 +1,17 @@
...
+#pragma once
+
+#include <functional>
+#include <string>
+
+namespace rocprofsys::control
+{
+struct subscriber
+{
+    std::function<void()> on_pause;
```
mradosav-amd: add explicit ctor with arguments. and remove default ctor.
with this we enable to create subscriber without initialized std::function

projects/rocprofiler-systems/source/lib/core/control/triggers/roctx.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+    /// Marker-write gate: paused always suppresses writes; otherwise write
+    /// iff no filter is configured or a target region is currently open.
+    [[nodiscard]] bool should_write_markers() const noexcept
+    {
+        return m_should_write.load(std::memory_order_relaxed);
+    }
+
+private:
+    session&                           m_session;
```
mradosav-amd: no need to tight it to session

dgaliffiAMD (approved):

projects/rocprofiler-systems/source/lib/rocprof-sys/library/rocprofiler-sdk.cpp
```diff
@@ -92,41 +93,7 @@ namespace
...
-            config::get_use_timemory(),
-            config::get_perfetto_annotations(),
-            roctx_traced_regions,
-        };
-        g_roctx_client = std::make_shared<roctx_client<>>(roctx_config);
-    }
-
-    return g_roctx_client;
-}
+control::session*               g_session      = nullptr;
```
mradosav-amd:
```suggestion
std::shared_ptr<control::session> g_session      = nullptr;
```

projects/rocprofiler-systems/source/lib/core/control/triggers/roctx.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include <string>
+#include <string_view>
+#include <unordered_set>
+
+namespace rocprofsys::control::triggers
+{
+class roctx
+{
+public:
+    roctx(session& sess, std::string_view trace_regions);
```
mradosav-amd:
```suggestion
    roctx(const std::shared_ptr<session>& sess, std::string_view trace_regions);
```

projects/rocprofiler-systems/source/lib/core/control/triggers/roctx.hpp
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+private:
+    std::set<std::string, std::less<>> m_trace_regions;
+    std::unordered_set<std::uint64_t>  m_active_range_ids;
+    std::atomic<bool>                  m_in_region{ false };
+    std::atomic<bool>                  m_user_paused{ false };
+    std::atomic<bool>                  m_should_write{ true };
+    std::mutex                         m_mutex;
+    session&                           m_session;
```
mradosav-amd:
```suggestion
    std::shared_ptr<session>                           m_session;
```

projects/rocprofiler-systems/source/lib/core/control/triggers/time_window.hpp
```diff
@@ -0,0 +1,115 @@
...
+class time_window
+{
+public:
+    struct config
+    {
+        clock_duration delay{};
+        clock_duration duration{};
+    };
+
+    time_window(session& sess, Clock& clk, config cfg)
```
mradosav-amd:
```suggestion
    time_window(const std::shared_ptr<session>& sess, Clock& clk, config cfg)
```

projects/rocprofiler-systems/source/lib/core/control/triggers/time_window.hpp
```diff
@@ -0,0 +1,115 @@
...
+        std::scoped_lock const lk{ m_lifecycle_mutex };
+        if(!m_thread.joinable()) return;
+        m_clock.interrupt();
+        m_thread.join();
+    }
+
+private:
+    static constexpr std::string_view trigger_name = "time_window";
+
+    session&     m_session;
```
mradosav-amd:
```suggestion
    std::shared_ptr<session>     m_session;
```

mradosav-amd (approved):
just some minor changes. other stuff look ok

kcossett-amd:
I've investigated the `runtime-instrument`/`binary-rewrite` failure. The condition fix (https://github.com/ROCm/rocm-systems/pull/9996/changes#diff-7aa9eb486d6498bd67d85737c680f982343d2304dcb9548367fbaebdd2f067d8R137) in #9996 enabled a previously-dead branch that now calls `closeSymtab` on the target binary before Dyninst opens it for the real rewrite. This is exposing an existing bug (the PR itself is NOT a problem).

Calling that is destructive: `closeSymtab` -> `~Symtab` -> `MappedFile::closeMappedFile` -> `munmap` of the file image, but hte `Elf_X` object survives because `Elf_X::end()` is a no-op upstream. Dyninst's `DwarfHandle::all_dwarf_handles` cache is keyed only on filename, so the reopen returns the cached handle still pointing at the dead `Elf_X`. The crash lands when `_fini` triggers lazy type parsing and gelf_getehdr reads the unmapped image

Commenting out `closeSymtab` is a sound workaround for now (https://github.com/ROCm/rocm-systems/blame/develop/projects/rocprofiler-systems/source/bin/rocprof-sys-instrument/rocprof-sys-instrument.hpp#L152) (it leaks one Symtab, mostly harmless), but the real fix belongs in Dyninst... `createDwarfHandle` must not reuse a cached handle bound to a different `Elf_X`.

I tested the workaround and the problem vanishes. I'll have to get to work on the dyninst side of things.

FYI @marantic-amd

kcossett-amd:
You'll need to rebase one more time to pick up the ASan workaround.

projects/rocprofiler-systems/source/lib/core/constraint.cpp
```diff
@@ -45,14 +45,21 @@ clock_name(std::string _v)
...
-                                ROCPROFSYS_CLOCK_IDENTIFIER(CLOCK_MONOTONIC),
-                                ROCPROFSYS_CLOCK_IDENTIFIER(CLOCK_PROCESS_CPUTIME_ID),
-                                ROCPROFSYS_CLOCK_IDENTIFIER(CLOCK_MONOTONIC_RAW),
-                                ROCPROFSYS_CLOCK_IDENTIFIER(CLOCK_REALTIME_COARSE),
-                                ROCPROFSYS_CLOCK_IDENTIFIER(CLOCK_MONOTONIC_COARSE),
-                                ROCPROFSYS_CLOCK_IDENTIFIER(CLOCK_BOOTTIME) };
+const std::set<clock_identifier>&
+accepted_clock_ids()
+{
+    // NOLINTBEGIN
```
mradosav-amd: what warnings are we suppressing here? put comment

projects/rocprofiler-systems/source/lib/core/constraint.cpp
```diff
@@ -62,7 +69,7 @@ find_clock_identifier(const Tp& _v)
     if constexpr(std::is_integral<Tp>::value)
```
mradosav-amd: can we fetch accepted_clock_ids() at start of function?

projects/rocprofiler-systems/source/lib/core/constraint.cpp
```diff
@@ -284,7 +293,7 @@ spec::operator()(const stages& _stages) const
 const std::set<clock_identifier>&
 get_valid_clock_ids()
 {
-    return accepted_clock_ids;
+    return accepted_clock_ids();
 }
```
mradosav-amd: do we need this function? just wrapping accepted_clock_ids

projects/rocprofiler-systems/source/lib/core/control/clock.hpp
```diff
@@ -0,0 +1,13 @@
...
+
+#pragma once
+
+#include <chrono>
+
+namespace rocprofsys::control
+{
+using clock_duration = std::chrono::nanoseconds;
+using clock_time_point =
+    std::chrono::time_point<std::chrono::steady_clock, clock_duration>;
```
mradosav-amd, line 12: not sure if this is necessary. 
`std::chrono::nanoseconds` and `std::chrono::time_point<std::chrono::steady_clock, std::chrono::nanoseconds>` are clear as well.

projects/rocprofiler-systems/source/lib/core/control/clocks/steady.cpp
```diff
@@ -0,0 +1,38 @@
...
+clock_time_point
+steady::now() const noexcept
+{
+    return std::chrono::time_point_cast<clock_duration>(std::chrono::steady_clock::now());
+}
+
+bool
+steady::sleep_until(clock_time_point deadline)
+{
+    std::unique_lock<std::mutex> clk_lcoks{ m_mutex };
```
mradosav-amd, line 24: Add thread internal state guard. this will cause deadlock.
Please *add it for whole stack* for all mutex usage.

projects/rocprofiler-systems/source/lib/core/control/session.cpp
```diff
@@ -0,0 +1,134 @@
...
+{
+    const std::scoped_lock subs_lk{ m_subscribers_mutex };
+    m_subscribers.push_back(std::move(sub));
+}
+
+void
+session::register_trigger(std::string_view name, action initial)
+{
+    const std::scoped_lock trig_lk{ m_actions_mutex };
+    m_actions[std::string{ name }] = initial;
```
mradosav-amd: should we use insert or emplace

projects/rocprofiler-systems/source/lib/rocprof-sys/library/rocprofiler-sdk.cpp
```diff
@@ -2901,12 +2865,37 @@ flush_counter_tracks_to_zero(rocprofiler_timestamp_t timestamp)
...
+{
+    if(g_roctx_client || !g_session) return;
+
+    const auto _domains = rocprofsys::delimit(
+        config::get_setting_value<std::string>(std::string{ env_vars::ROCM_DOMAINS })
+            .value_or(std::string{}),
+        " ,;:\t\n");
+    const auto has_marker_domain =
+        (std::find(_domains.begin(), _domains.end(), "marker_api") != _domains.end() ||
+         std::find(_domains.begin(), _domains.end(), "roctx") != _domains.end());
```
mradosav-amd: can call core::rocprofiler_sdk::get_buffered_domains (or `sdk/tracing-config.hpp` after refactor) , not to parse string, and to check for actual enum.

if you think this is out of scope for this PR, create Jira ticket, put it here and resolve conversation

projects/rocprofiler-systems/source/lib/rocprof-sys/library/rocprofiler-sdk.cpp
```diff
@@ -2901,12 +2865,37 @@ flush_counter_tracks_to_zero(rocprofiler_timestamp_t timestamp)
 
 }  // namespace
 
-std::shared_ptr<control::trace_control>
-get_trace_controller()
+void
+bind_session(std::shared_ptr<control::session> sess)
```
mradosav-amd: `set_session`? I though that this `bind_` is doing something more than set.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/rocprofiler-sdk.cpp
marantic-amd: I will add a ticket to Jira for this. This stack is already growing as it is. It's minor refactor but would prefer to do it in separate scope.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/rocprofiler-sdk.cpp
marantic-amd: Makes sense, I will rename it.

projects/rocprofiler-systems/source/lib/core/control/clock.hpp
marantic-amd: It's used throughout the code, and made sense to have it here in such way for better readability. At this point it is the mater of taste, and I would prefer to leave it as such, and if we agree on a different approach remove it is some future PR.

projects/rocprofiler-systems/source/lib/core/constraint.cpp
marantic-amd: Missed to add that one. I will update the comments

projects/rocprofiler-systems/source/lib/core/control/triggers/time_window.hpp
```diff
@@ -0,0 +1,137 @@
...
+#include <thread>
+#include <utility>
+
+namespace rocprofsys::control::triggers
+{
+template <typename Clock>
+class time_window
+{
+public:
+    struct config
```
mradosav-amd:
```suggestion
    struct phases
```

mradosav-amd: not sure do we need this type to be public at all or do we need type at all
```
clock_duration delay{};
clock_duration duration{};
```

as members of time_window perfectly explain it.

projects/rocprofiler-systems/source/lib/core/control/triggers/roctx.cpp
```diff
@@ -0,0 +1,184 @@
...
+    const bool now_empty = m_active_range_ids.empty();
+    if(now_empty)
+    {
+        m_in_region.store(false, std::memory_order_relaxed);
+    }
+    return now_empty;
+}
+
+void
+roctx::warn_if_paused_region_ended()
```
mradosav-amd, line 127: no need for this function
```
if(paused)
warn
```

is clear in code as well. no need for function

projects/rocprofiler-systems/source/lib/core/control/triggers/roctx.cpp
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    if(filter_active())
+    {
+        LOG_INFO("roctx trigger: filter active for regions: [{}]",
+                 fmt::join(m_trace_regions, ", "));  // NOLINT
+    }
+
+    m_should_write.store(compute_should_write(), std::memory_order_relaxed);
+    m_session->register_trigger(trigger_name, compute_action());
```
mradosav-amd:
```suggestion
refresh_state();
```

projects/rocprofiler-systems/source/lib/core/control/triggers/roctx.cpp
```diff
@@ -0,0 +1,184 @@
...
+    {
+        const auto delimited =
+            rocprofsys::common::delimit(std::string{ trace_regions }, ",");
+        m_trace_regions.insert(delimited.begin(), delimited.end());
+    }
+
+    if(filter_active())
+    {
+        LOG_INFO("roctx trigger: filter active for regions: [{}]",
+                 fmt::join(m_trace_regions, ", "));  // NOLINT
```
mradosav-amd: specify which NOLINT warning.

projects/rocprofiler-systems/source/lib/core/control/triggers/roctx.cpp
```diff
@@ -0,0 +1,184 @@
...
+roctx::compute_should_write() const noexcept
+{
+    const bool filter    = filter_active();
+    const bool in_region = m_in_region.load(std::memory_order_relaxed);
+    const bool paused    = m_user_paused.load(std::memory_order_relaxed);
+    return !paused && (!filter || in_region);
+}
+
+void
+roctx::refresh_state()
```
mradosav-amd, line 204: this is called on exit of most of functions. we can maybe create raii object. 
just suggestion.

projects/rocprofiler-systems/source/lib/core/control/triggers/roctx.cpp
marantic-amd: The clang tidy sees this conditional branching when outside of helper fiunction and it raises the "complexity" boundary issue within the check.

projects/rocprofiler-systems/source/lib/core/control/session.cpp
```diff
@@ -0,0 +1,134 @@
...
+    m_active.store(resolve_locked(), std::memory_order_relaxed);
```
mradosav-amd: should we notify subscribers upon registering/ unregistering new trigger?

projects/rocprofiler-systems/source/lib/core/control/triggers/time_window.hpp
marantic-amd: I think `config` better describes what is going on here

projects/rocprofiler-systems/source/lib/core/control/triggers/roctx.cpp
marantic-amd: `refresh_state` is calling `m_session->set_action` so it's not a plain pick-and-replace at this case.

projects/rocprofiler-systems/source/lib/core/control/triggers/time_window.hpp
marantic-amd: Making them separate members makes more sense. I will prefer t ogo in that direction

projects/rocprofiler-systems/source/lib/core/constraint.cpp
marantic-amd: Currently `accepted_clock_ids` is in an anonymous namespace inside that file and `get_valid_clocks_ids` is is the only function that exposes that data to the rest of the codebase. It's just encapsulation boundary I would like to keep in this first PR.

projects/rocprofiler-systems/source/lib/core/control/session.cpp
marantic-amd: I used `[]` operator on purpose here. It's intended to overwrite an existing entry. In case of insert or emplace it would be a silent no-op if the key already exists.

projects/rocprofiler-systems/source/lib/rocprof-sys/library.cpp
```diff
@@ -1052,6 +1061,9 @@ rocprofsys_finalize_hidden(void)
     LOG_DEBUG("Shutting down ROCm...");
     rocprofiler_sdk::shutdown();
 
+    LOG_DEBUG("Shutting down control session...");
+    get_control_session()->shutdown();
```
habajpai-amd, line 1182: The control session is shut down after rocprofiler_sdk::shutdown(). Destroying the ROCtx trigger during SDK shutdown can change a paused session to active and call every resume callback. This may resume components during finalization. Please shut down the control session first and add a test that shutdown does not
send resume callbacks.

projects/rocprofiler-systems/source/lib/rocprof-sys/library.cpp
```diff
@@ -939,6 +945,9 @@ rocprofsys_finalize_hidden(void)
         LOG_DEBUG("Shutting down ROCm...");
         rocprofiler_sdk::shutdown();
 
+        LOG_DEBUG("Shutting down control session...");
+        get_control_session()->shutdown();
```
habajpai-amd, line 1064: The control session is shut down after rocprofiler_sdk::shutdown(). Destroying the ROCtx trigger during SDK shutdown can change a paused session to active and call every resume callback. This may resume components during finalization. Please shut down the control session first and add a test that shutdown does not
send resume callbacks.

projects/rocprofiler-systems/source/lib/core/control/session.cpp
```diff
@@ -0,0 +1,134 @@
...
+session::notify_pause()
+{
+    const auto _thread_state_guard = state::thread::scoped(state::thread::Internal);
+    const std::scoped_lock notify_lk{ m_subscribers_mutex };
+    for(const auto& sub : m_subscribers)
+    {
+        LOG_DEBUG("session: pausing subscriber '{}'", sub.name);
+        if(sub.on_pause)
+        {
+            sub.on_pause();
```
habajpai-amd, line 173: Subscriber callbacks are called while session locks are held. If a callback calls subscribe(), shutdown(), or changes a trigger, it can deadlock. Please copy the callbacks while locked and call them after releasing the lock. Also handle callback exceptions so they cannot terminate the process.

projects/rocprofiler-systems/source/lib/core/control/session.cpp
```diff
@@ -0,0 +1,134 @@
...
+void
+session::shutdown()
+{
+    const auto _thread_state_guard = state::thread::scoped(state::thread::Internal);
+    {
+        const std::scoped_lock subs_lk{ m_subscribers_mutex };
+        m_subscribers.clear();
+    }
+    {
+        const std::scoped_lock action_lock{ m_actions_mutex };
```
habajpai-amd, line 37: shutdown() does not lock m_notify_mutex. A trigger update can overlap shutdown and leave the session paused after shutdown has cleared all subscribers. Please use m_notify_mutex here so shutdown waits for an update already in progress.

projects/rocprofiler-systems/source/lib/core/control/clocks/steady.cpp
```diff
@@ -0,0 +1,41 @@
...
+    return !m_cv.wait_until(clk_lcoks, deadline, [this] { return m_interrupted; });
+}
+
+void
+steady::interrupt()
+{
+    const auto _thread_state_guard = state::thread::scoped(state::thread::Internal);
+    {
+        const std::scoped_lock clk_lcoks{ m_mutex };
+        m_interrupted = true;
```
habajpai-amd, line 37: interrupt() never clears m_interrupted. After stop(), start() creates another worker, but that worker exits immediately. Please reset the interrupt state before restarting, or make time_window one-shot.

Also, stop() can try to join its own worker thread if a subscriber calls it from that thread. join() then throws, and noexcept makes the process terminate. Please prevent a self-join.

projects/rocprofiler-systems/source/lib/core/control/triggers/roctx.cpp
```diff
@@ -0,0 +1,209 @@
...
+    if(was_empty)
+    {
+        m_in_region.store(true, std::memory_order_relaxed);
+    }
+}
+
+void
+roctx::on_range_stop(std::uint64_t range_id)
+{
+    const scope_exit refresh{ [this] { refresh_state(); } };
```
habajpai-amd: The scope guard is created before the filter check, so the early return still calls refresh_state() and takes the session locks. This happens for every ROCtx range stop when filtering is disabled. Please create the guard after the early return and refresh only when the state changes.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/rocprofiler-sdk/tests/test_roctx_client.cpp
```diff
@@ -423,116 +459,112 @@ TEST_F(roctx_client_control_test, selective_region_pause_then_region_ends)
...
-    ctrl->handle_range_start(1, "Region 1");
-    EXPECT_TRUE(ctrl->should_write_markers());
+    client->get_trigger().on_range_start(1, "Region 1");
+    EXPECT_TRUE(client->get_trigger().should_write_markers());
 
     ctrl->shutdown();
 
-    EXPECT_FALSE(ctrl->region_filter_active());
-    EXPECT_TRUE(ctrl->should_write_markers());
+    EXPECT_TRUE(client->get_trigger().filter_active());
```
habajpai-amd, line 587: The session is already active before shutdown, and the assertions only check the trigger. This test would pass even if session::shutdown() did nothing. Please start with a paused session, check that shutdown makes it active, and verify that removed subscribers are not called afterward.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/rocprofiler-sdk/tests/test_roctx_client.cpp
```diff
@@ -163,20 +187,31 @@ class roctx_client_control_test : public mock_cleanup_base
...
     {
-        const roctx_config_t config{ true, false, false, false, regions };
-        auto                 client = std::make_unique<roctx_client_t>(config);
+        const roctx_config_t config{ .pause_resume_enabled   = true,
+                                     .use_perfetto           = false,
+                                     .use_timemory           = false,
+                                     .perfetto_annotations   = false,
+                                     .selected_trace_regions = regions };
+
+        auto client = std::make_unique<roctx_client_t>(m_session, config);
```
habajpai-amd, line 211: These helpers create the trigger before adding the subscriber, so the subscriber misses the initial pause. Production subscribes first and then creates the trigger. Please add a test that uses the production order.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/rocprofiler-sdk/tests/test_roctx_client.cpp
```diff
@@ -712,12 +744,24 @@ class roctx_push_pop_region_test : public mock_cleanup_base
...
-        auto                 client = std::make_unique<roctx_client_t>(config);
-        client->get_controller()->register_region_pause_resume_callbacks(
-            [this]() { start_count++; }, [this]() { stop_count++; });
+        const roctx_config_t config{ .pause_resume_enabled   = true,
+                                     .use_perfetto           = false,
+                                     .use_timemory           = false,
+                                     .perfetto_annotations   = false,
+                                     .selected_trace_regions = regions };
+
+        auto client = std::make_unique<roctx_client_t>(m_session, config);
```
habajpai-amd, line 850: These helpers create the trigger before adding the subscriber, so the subscriber misses the initial pause. Production subscribes first and then creates the trigger. Please add a test that uses the production order.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/rocprofiler-sdk/tests/test_roctx_client.cpp
```diff
@@ -189,23 +224,26 @@ class roctx_client_control_test : public mock_cleanup_base
 //   CodeB            => NOT profiled (paused)
 //   roctx_resume     => start callback fires; should_write becomes true
 //   CodeC            => profiled
+//   CodeD            => profiled
+//
+// Without a region filter, should_write_markers() always returns true.
```
habajpai-amd: This says should_write_markers() is always true without a region filter, but the test expects false after on_pause(). Please clarify that an explicit pause still disables marker writes when no region filter is configured.

projects/rocprofiler-systems/source/lib/rocprof-sys/library/rocprofiler-sdk/tests/test_roctx_client.cpp
```diff
@@ -102,56 +103,79 @@ TEST_F(roctx_client_test, constructor_creates_controller)
...
-    const roctx_client_config config{ true, true, true, false, "TestRegion" };
-    const roctx_client<mock_marker_policy> client(config);
-    EXPECT_NE(client.get_controller(), nullptr);
+    const roctx_client_config config{ .pause_resume_enabled   = true,
+                                      .use_perfetto           = true,
+                                      .use_timemory           = true,
+                                      .perfetto_annotations   = false,
+                                      .selected_trace_regions = "TestRegion" };
+    const auto                session = std::make_shared<rocprofsys::control::session>();
+    const roctx_client<mock_marker_policy> client(session, config);
```
habajpai-amd, line 112: This test only constructs the client, and its name still mentions the removed controller. Please rename it and check the expected initial trigger state or the supplied session.

habajpai-amd (dismissed):

projects/rocprofiler-systems/source/lib/core/control/session.cpp
marantic-amd: Fixed in last stacked PR

projects/rocprofiler-systems/source/lib/core/control/session.cpp
marantic-amd: Fixed in last stacked PR

projects/rocprofiler-systems/source/lib/core/control/clocks/steady.cpp
marantic-amd: Fixed in last stacked PR

projects/rocprofiler-systems/source/lib/core/control/triggers/roctx.cpp
marantic-amd: You overstated the cost claim here, to be quite honest. `refresh_state()` already early-returns before any notify when nothing changed so it's two uncontended mutex ops, not any real heavy work

habajpai-amd:
@marantic-amd 
Re-reviewed the PR. The latest update fixes the shutdown order, the ROCtx no-filter path, the weak shutdown test, the subscriber-order tests, the incorrect comment, and the constructor test. I found no new issues in this update.

Two reachable concurrency problems still need to be fixed in this PR:
1. roctx::refresh_state() runs after m_mutex is released. Two application threads can publish updates in the wrong order and leave the session active after the last target range has closed. Please compute and publish one state value in the same protected order.
2. session::shutdown() does not use m_notify_mutex. A trigger update can overlap shutdown and leave the session paused after its state and subscribers were cleared. Subscriber callback re-entry and the time_window restart/self-join issues also remain, but I am fine with those being fixed in the later stacked PRs if those PRs are linked here.

Keeping Request Changes until the two concurrency issues above are fixed in this PR.

marantic-amd:
> @marantic-amd Re-reviewed the PR. The latest update fixes the shutdown order, the ROCtx no-filter path, the weak shutdown test, the subscriber-order tests, the incorrect comment, and the constructor test. I found no new issues in this update.
> 
> Two reachable concurrency problems still need to be fixed in this PR:
> 
> 1. roctx::refresh_state() runs after m_mutex is released. Two application threads can publish updates in the wrong order and leave the session active after the last target range has closed. Please compute and publish one state value in the same protected order.
> 2. session::shutdown() does not use m_notify_mutex. A trigger update can overlap shutdown and leave the session paused after its state and subscribers were cleared. Subscriber callback re-entry and the time_window restart/self-join issues also remain, but I am fine with those being fixed in the later stacked PRs if those PRs are linked here.
> 
> Keeping Request Changes until the two concurrency issues above are fixed in this PR.

@habajpai-amd 
The two requests you have are already answered. I do not see the 
1. This is not a concerning issue in my opinion since in our use cases it's never reached, and the wiring itself in the consecutive PRs will prevent this.
2. As mentioned in the specific comments, this is already resolved in later PR and I will not be pulling changes from the later PRs in stack here, since this should just be foundational work and the later wirings and fixes are where the real issues that surfaced during testing and development are addressed.

PR #8752, opened 2026-07-17
feat(hotswap): add lazy per-kernel presentation mode

martin-luecke:
## Motivation

HotSwap translates AMD GPU machine code from one ISA to another, so a kernel built for one AMD GPU can run on a different one.

The immediate driver is upcoming GPUs, whose code objects often work well before the silicon is widely available. HotSwap lets us develop against that software --frameworks, libraries, and their kernels-- and run it in CI on current hardware, giving early signal on whether it loads, dispatches, and produces correct results instead of waiting for scarce or unavailable silicon.

## Technical details

The translation itself is being added to COMGR (`amd_comgr_hotswap_transpile`) as separate, in-progress work. This PR is the runtime integration that drives it and generates no code itself. It is an early proposal, brought for guidance rather than as a finished design.

Frameworks and libraries pick which code objects to load based on the GPU they see, so with this patch the runtime presents a source ISA that differs from the device's physical execution ISA, and tracks both. The stack loads source-ISA code objects, and ROCclr admits them instead of rejecting them as incompatible. The first time a source-ISA kernel is dispatched, ROCr translates it through COMGR and caches the result for reuse. The dispatch path then rewrites the AQL packet's kernel_object to the translated kernel before it reaches hardware. This is transparent to the application and looks like an ordinary dispatch.

Translation is lazy because a code object may hold many kernels while a workload dispatches only a few. Anything we can't handle safely, e.g. an unsupported translation, is refused. We are currently translating code objects from gfx1250 to gfx942 or gfx950 (e.g. running MI450 kernels on MI350), but the mechanism is not tied to those.

The features in this patch are controlled at runtime by `HSA_HOTSWAP_PRESENT_ISA`. With it unset, the runtime runs exactly as it does today: the new checks all short-circuit, so no kernel is translated and no dispatch is rewritten. The rest is inert too -- `execution_isas()` returns the device's usual ISA and the presented-ISA fields fall back to it. So this PR with the variable should behave like a build without it.

## Discussion

We'd like guidance from the runtime owners on:

- how you see the overall approach of presenting one ISA while executing another, and whether this capability belongs in ROCr/ROCclr or elsewhere
- whether the presented-ISA / execution-ISA split sits in the right place across ROCr and ROCclr
- how you'd want it staged for review across queue/dispatch, loader/kernel-object tracking, and ROCclr device presentation.

JIRA ID: ROCM-28199

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/inc/hotswap_env.hpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/inc/hotswap_env.hpp
@@ -1,0 +1,37 @@
+////////////////////////////////////////////////////////////////////////////////
+//
+// The University of Illinois/NCSA
+// Open Source License (NCSA)
+//
+// Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved.
+//
+////////////////////////////////////////////////////////////////////////////////
+
+#ifndef HSA_RUNTIME_CORE_INC_HOTSWAP_ENV_HPP_
+#define HSA_RUNTIME_CORE_INC_HOTSWAP_ENV_HPP_
+
+#include <algorithm>
+#include <cctype>
+#include <cstdlib>
+#include <string>
+
+namespace rocr {
+namespace hotswap {
+
+inline bool IsEnvFlagEnabled(const char* name) {
+  const char* raw_value = std::getenv(name);
+  if (!raw_value) return false;
+
+  std::string value(raw_value);
+  if (value.empty()) return false;
+
+  std::transform(value.begin(), value.end(), value.begin(),
+                 [](unsigned char c) { return std::tolower(c); });
+  return value != "0" && value != "off" && value != "false" &&
+         value != "no" && value != "n" && value != "f";
+}
+
+}  // namespace hotswap
+}  // namespace rocr
+
+#endif  // HSA_RUNTIME_CORE_INC_HOTSWAP_ENV_HPP_
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/inc/hotswap_env.hpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/inc/hotswap_env.hpp
@@ -32,6 +32,14 @@ inline bool IsEnvFlagEnabled(const char* name) {
   return IsEnvFlagValueEnabled(os::GetEnvVar(env_name));
 }
 
+inline bool IsPresentationModeEnabled() {
+  static const bool Enabled = [] {
+    if (IsEnvFlagEnabled("HSA_HOTSWAP_DISABLE")) return false;
+    return IsEnvFlagEnabled("HSA_HOTSWAP_PRESENT_ISA");
+  }();
+  return Enabled;
+}
+
 }  // namespace hotswap
 }  // namespace rocr
 
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/inc/hotswap_env.hpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/inc/hotswap_env.hpp
@@ -7,6 +7,9 @@
 #ifndef HSA_RUNTIME_CORE_INC_HOTSWAP_ENV_HPP_
 #define HSA_RUNTIME_CORE_INC_HOTSWAP_ENV_HPP_
 
+#ifdef ROCR_HOTSWAP_COMGR_ADAPTER
+#include "core/inc/isa.h"
+#endif
 #include "core/util/os.h"
 
 #include <algorithm>
@@ -39,6 +42,46 @@ inline bool IsPresentationModeEnabled() {
   }();
   return Enabled;
 }
+
+#ifdef ROCR_HOTSWAP_COMGR_ADAPTER
+inline bool ValidateHotSwapPresentationTarget(
+    const std::string& target_env, const std::string& override_env,
+    const core::Isa& execution_isa, std::string& failure) {
+  const std::string& configured_target =
+      target_env.empty() ? override_env : target_env;
+  if (configured_target.empty() || configured_target == "0" ||
+      configured_target == "1") {
+    failure =
+        "HSA_HOTSWAP_TARGET must name the execution ISA when "
+        "HSA_HOTSWAP_PRESENT_ISA is set";
+    return false;
+  }
+
+  std::string configured_name(configured_target);
+  if (configured_name.rfind("amdgcn-amd-amdhsa--", 0) != 0)
+    configured_name = "amdgcn-amd-amdhsa--" + configured_name;
+  const core::Isa* configured_isa = core::IsaRegistry::GetIsa(configured_name);
+  if (!configured_isa) {
+    failure = "configured HotSwap target ISA '" + configured_name +
+              "' is not recognized by ROCr";
+    return false;
+  }
+
+  // Require the same processor; IsCompatible applies target-ID "any"
+  // semantics to omitted features and rejects explicitly mismatched features.
+  if (configured_isa->GetProcessorName() != execution_isa.GetProcessorName() ||
+      !core::Isa::IsCompatible(*configured_isa, execution_isa, 0)) {
+    failure = "configured HotSwap target ISA '" +
+              configured_isa->GetIsaName() +
+              "' does not match physical execution ISA '" +
+              execution_isa.GetIsaName() + "'";
+    return false;
+  }
+
+  failure.clear();
+  return true;
+}
+#endif
 
 }  // namespace hotswap
 }  // namespace rocr
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/inc/hotswap_env.hpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/inc/hotswap_env.hpp
@@ -1,11 +1,8 @@
-////////////////////////////////////////////////////////////////////////////////
-//
-// The University of Illinois/NCSA
-// Open Source License (NCSA)
-//
-// Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved.
-//
-////////////////////////////////////////////////////////////////////////////////
+/*
+ * Copyright (c) Advanced Micro Devices, Inc., or its affiliates.
+ *
+ * SPDX-License-Identifier: MIT
+ */
 
 #ifndef HSA_RUNTIME_CORE_INC_HOTSWAP_ENV_HPP_
 #define HSA_RUNTIME_CORE_INC_HOTSWAP_ENV_HPP_
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/inc/hotswap_env.hpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/inc/hotswap_env.hpp
@@ -10,25 +10,29 @@
 #ifndef HSA_RUNTIME_CORE_INC_HOTSWAP_ENV_HPP_
 #define HSA_RUNTIME_CORE_INC_HOTSWAP_ENV_HPP_
 
+#include "core/util/os.h"
+
 #include <algorithm>
 #include <cctype>
-#include <cstdlib>
 #include <string>
 
 namespace rocr {
 namespace hotswap {
 
-inline bool IsEnvFlagEnabled(const char* name) {
-  const char* raw_value = std::getenv(name);
-  if (!raw_value) return false;
-
-  std::string value(raw_value);
+inline bool IsEnvFlagValueEnabled(std::string value) {
   if (value.empty()) return false;
 
   std::transform(value.begin(), value.end(), value.begin(),
                  [](unsigned char c) { return std::tolower(c); });
   return value != "0" && value != "off" && value != "false" &&
          value != "no" && value != "n" && value != "f";
+}
+
+inline bool IsEnvFlagEnabled(const char* name) {
+  std::string env_name(name);
+  if (!os::IsEnvVarSet(env_name)) return false;
+
+  return IsEnvFlagValueEnabled(os::GetEnvVar(env_name));
 }
 
 }  // namespace hotswap
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
@@ -497,7 +497,7 @@ uint64_t AqlQueue::AddWriteIndexRelease(uint64_t value) {
 
 void AqlQueue::StoreRelaxed(hsa_signal_value_t value) {
 #ifdef ROCR_HOTSWAP_COMGR_ADAPTER
-  if (value >= 0 && rocr::amd::hsa::loader::ShouldCheckHotSwapDispatchKernelObjects()) {
+  if (rocr::amd::hsa::loader::ShouldCheckHotSwapDispatchKernelObjects()) {
     std::lock_guard<std::mutex> lock(hotswap_dispatch_lock_);
     // HotSwap's dispatch intercept relies on seeing every published dispatch
     // before the CP does. It is only valid for producers that ring the doorbell
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
@@ -72,11 +72,25 @@
 #include "core/inc/amd_gpu_pm4.h"
 #include "core/inc/hsa_amd_tool_int.hpp"
 #include "core/inc/amd_core_dump.hpp"
+#include "loader/AMDHSAKernelDescriptor.h"
+#ifdef ROCR_HOTSWAP_COMGR_ADAPTER
+#include "core/runtime/hotswap_aql_patch.h"
+#include "loader/executable.hpp"
+#endif
 
 namespace rocr {
 namespace AMD {
 
 #define SCRATCH_ALT_RATIO 4
+
+#ifdef ROCR_HOTSWAP_COMGR_ADAPTER
+// This file already uses x86 sfence for hardware doorbell ordering. Keep the
+// HotSwap packet-body fence on the same host target set.
+#if !defined(__x86_64__) && !defined(_M_X64)
+#error "ROCR_HOTSWAP_COMGR_ADAPTER packet fencing requires x86 sfence support"
+#endif
+static void HotSwapPacketBodyFence() { _mm_sfence(); }
+#endif
 
 AqlQueue::AqlQueue(core::SharedQueue* shared_queue, GpuAgent* agent, size_t req_size_pkts,
                    HSAuint32 node_id, ScratchInfo& scratch, core::HsaEventCallback callback,
@@ -100,7 +114,8 @@ AqlQueue::AqlQueue(core::SharedQueue* shared_queue, GpuAgent* agent, size_t req_
       exceptionState(0),
       suspended_(false),
       priority_(HSA::HSA_AMD_QUEUE_PRIORITY_NORMAL),
-      exception_signal_(nullptr) {
+      exception_signal_(nullptr),
+      last_doorbell_value_(UINT64_MAX) {
 
   // Queue size is a function of several restrictions.
   const uint32_t min_pkts = ComputeRingBufferMinPkts();
@@ -118,8 +133,8 @@ AqlQueue::AqlQueue(core::SharedQueue* shared_queue, GpuAgent* agent, size_t req_
 
   if (core::Runtime::runtime_singleton_->KfdVersion().supports_metadata_prefetch &&
       metadata_prefetch &&
-      agent_->supported_isas()[0]->GetMajorVersion() == 12
-      && agent_->supported_isas()[0]->GetMinorVersion() >= 5) {
+      agent_->execution_isas()[0]->GetMajorVersion() == 12
+      && agent_->execution_isas()[0]->GetMinorVersion() >= 5) {
     /* First valid version of meta data prefetch - Version is 0.0 */
     dispatch_version_.set_version(0, 0);
     barrier_version_.set_version(0, 0);
@@ -128,6 +143,7 @@ AqlQueue::AqlQueue(core::SharedQueue* shared_queue, GpuAgent* agent, size_t req_
   // Allocate the AQL packet ring buffer.
   AllocRegisteredRingBuffer(queue_size_pkts);
   if (ring_buf_ == nullptr) throw std::bad_alloc();
+  hotswap_patched_generations_.assign(queue_size_pkts, UINT64_MAX);
   MAKE_NAMED_SCOPE_GUARD(RingGuard, [&]() { FreeQueueMemory(); });
 
   // Zero the amd_queue_ structure to clear RPTR/WPTR before queue attach.
@@ -199,7 +215,7 @@ AqlQueue::AqlQueue(core::SharedQueue* shared_queue, GpuAgent* agent, size_t req_
     assert(amd_queue_.private_segment_aperture_base_hi != 0 && "No private region found.");
   }
 
-  if (agent_->supported_isas()[0]->GetMajorVersion() >= 11)
+  if (agent_->execution_isas()[0]->GetMajorVersion() >= 11)
     queue_scratch_.mem_alignment_size = 256;
   else
     queue_scratch_.mem_alignment_size = 1024;
@@ -480,6 +496,51 @@ uint64_t AqlQueue::AddWriteIndexRelease(uint64_t value) {
 }
 
 void AqlQueue::StoreRelaxed(hsa_signal_value_t value) {
+#ifdef ROCR_HOTSWAP_COMGR_ADAPTER
+  if (value >= 0 && rocr::amd::hsa::loader::ShouldCheckHotSwapDispatchKernelObjects()) {
+    std::lock_guard<std::mutex> lock(hotswap_dispatch_lock_);
+    // HotSwap's dispatch intercept relies on seeing every published dispatch
+    // before the CP does. It is only valid for producers that ring the doorbell
+    // for each submitted range; a producer that publishes write_dispatch_id
+    // without ringing this signal cannot use presentation-mode HotSwap safely.
+    core::AqlPacket* packets =
+        reinterpret_cast<core::AqlPacket*>(amd_queue_.hsa_queue.base_address);
+    const uint64_t end = static_cast<uint64_t>(value);
+    const uint64_t queue_size = amd_queue_.hsa_queue.size;
+    uint64_t first = 0;
+    bool has_work = false;
+    const char* failure_reason = nullptr;
+    if (!rocr::AMD::hotswap::lazy::ComputeDoorbellPatchRange(
+            last_doorbell_value_, end, queue_size, first, has_work,
+            failure_reason)) {
+      fprintf(stderr, "hotswap: refusing doorbell patch: %s\n",
+              failure_reason ? failure_reason : "invalid AQL doorbell range");
+      std::abort();
+    }
+
+    for (uint64_t index = first; has_work && index <= end; ++index) {
+      const uint64_t slot = index % queue_size;
+      if (hotswap_patched_generations_[slot] == index) continue;
+      core::AqlPacket& packet = packets[slot];
+      // The producer has already published a valid packet header by the time
+      // the doorbell signal reaches us. Hide the packet while HotSwap resolves
+      // and rewrites the body, then republish the original header with release
+      // ordering before ringing the hardware doorbell below. Lazy translation
+      // is synchronous here: the COMGR/loader path must not submit work to this
+      // queue while its packet is hidden behind this doorbell.
+      if (!rocr::AMD::hotswap::lazy::PatchPublishedKernelDispatchPacket(
+              packet, rocr::amd::hsa::loader::PrepareHotSwapDispatchKernelObject,
+              IsDeviceMemRingBuf() && needsPcieOrdering()
+                  ? HotSwapPacketBodyFence
+                  : nullptr))
+        continue;
+      hotswap_patched_generations_[slot] = index;
+    }
+    if (last_doorbell_value_ == UINT64_MAX || end > last_doorbell_value_)
+      last_doorbell_value_ = end;
+  }
+#endif
+
   if (core::Runtime::runtime_singleton_->thunkLoader()->IsDTIF() ||
         core::Runtime::runtime_singleton_->thunkLoader()->IsDXG()) {
     HSAKMT_CALL(hsaKmtQueueRingDoorbell(queue_id_, value));
@@ -1097,7 +1158,7 @@ void AqlQueue::HandleInsufficientScratch(hsa_signal_value_t& error_code,
 
     // For gfx10+ devices we must attempt to assign the smaller of 256 lanes or 16 groups to each
     // engine.
-    if (agent_->supported_isas()[0]->GetMajorVersion() >= 10 &&
+    if (agent_->execution_isas()[0]->GetMajorVersion() >= 10 &&
         maxGroupsPerEngine < 16 &&
                               lanes_per_group * maxGroupsPerEngine < 256) {
       uint64_t groups_per_interleave = (256 + lanes_per_group - 1) / lanes_per_group;
@@ -1221,7 +1282,7 @@ void AqlQueue::HandleInsufficientScratch(hsa_signal_value_t& error_code,
   if (scratch.large) {
     amd_queue_.queue_properties |= AMD_QUEUE_PROPERTIES_USE_SCRATCH_ONCE;
     // Set system release fence to flush scratch stores with older firmware versions.
-    if ((agent_->supported_isas()[0]->GetMajorVersion() == 8) && (agent_->GetMicrocodeVersion() < 729)) {
+    if ((agent_->execution_isas()[0]->GetMajorVersion() == 8) && (agent_->GetMicrocodeVersion() < 729)) {
       pkt->dispatch.header &=
           ~(((1 << HSA_PACKET_HEADER_WIDTH_SCRELEASE_FENCE_SCOPE) - 1)
             << HSA_PACKET_HEADER_SCRELEASE_FENCE_SCOPE);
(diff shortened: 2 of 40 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
```diff
@@ -480,6 +496,51 @@ uint64_t AqlQueue::AddWriteIndexRelease(uint64_t value) {
 }
 
 void AqlQueue::StoreRelaxed(hsa_signal_value_t value) {
+#ifdef ROCR_HOTSWAP_COMGR_ADAPTER
+  if (value >= 0 && rocr::amd::hsa::loader::ShouldCheckHotSwapDispatchKernelObjects()) {
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** The `value >= 0` sub-condition is redundant — `ShouldCheckHotSwapDispatchKernelObjects()` already gates this path. It's also technically incorrect: `hsa_signal_value_t` is `int64_t` while the doorbell/write index is monotonically increasing `uint64_t`, so if the cumulative value ever exceeded `INT64_MAX` it would read negative and silently skip the intercept (i.e. dispatch untranslated packets). Unreachable in practice, but since the intercept is fail-closed by design, consider dropping the `value >= 0` check entirely so the gate can't be bypassed by a sign artifact.

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
```diff
@@ -247,6 +396,606 @@ static bool CodeObjectIsaIsGfx125Family(const std::string& codeIsa) {
...
+  }
+
+  if (record->Kind != HotSwapKernelKind::LazySource)
+    FatalHotSwapDispatch(*address, record->Name, record->Kind,
+                         record->Failure.empty()
+                             ? "kernel object is not translated"
+                             : record->Failure,
+                         record->SourceKind, record->TargetGfx);
+
+  std::lock_guard<std::mutex> Guard(record->Mutex);
```
cfreeamd, line 1008: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** `record->Mutex` is held across the full `LoadLazyTranslatedKernel` → COMGR transpile call (potentially seconds). Any other queue dispatching the same source kernel blocks here on first use, and every *subsequent* dispatch of an already-translated kernel still re-acquires this mutex just to re-verify+patch. Consider promoting `record->Kind` to a terminal `Translated` state after the first success (see line 891) so later dispatches hit the lock-free fast path instead of serializing on the record mutex.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa.cpp
```diff
@@ -2145,6 +2146,11 @@ hsa_status_t LoadSizedCodeObject(
                               uri, loaded_code_object);
 }
 
+bool HotswapPresentationModeEnabled() {
```
cfreeamd, line 2149: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** `HotswapPresentationModeEnabled()` calls `getenv` on every `hsa_executable_load_agent_code_object`, whereas the dispatch-side gate `ShouldCheckHotSwapDispatchKernelObjects()` caches its env lookups in a static local. Presentation mode is fixed at process start; using the same cache-once pattern here avoids per-load `getenv` overhead and removes any chance of the load-time and dispatch-time gates disagreeing if the environment is mutated mid-run.

projects/rocr-runtime/runtime/hsa-runtime/loader/hotswap_kernel_registry.h
```diff
@@ -0,0 +1,344 @@
...
+    records_.erase(it);
+    return true;
+  }
+
+  std::shared_ptr<HotSwapKernelRecord> Lookup(uint64_t address) const {
+    std::lock_guard<std::mutex> guard(mutex_);
+    auto it = records_.find(address);
+    if (it != records_.end()) return it->second;
+    for (const auto& entry : records_) {
+      if (entry.second->IsRegisteredRocrBlit(address)) return entry.second;
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** The blit-object fallback in `Lookup` is an O(n) linear scan over *all* registered kernels, and it fires on the copy hot path (packets whose address falls in a blit range). Blit registrations are few (roughly one per kernel type per agent), so keeping them in a small dedicated `std::vector` would bound this fallback to a constant scan instead of walking every registered user kernel.

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
```diff
@@ -247,6 +396,606 @@ static bool CodeObjectIsaIsGfx125Family(const std::string& codeIsa) {
...
+
+// Appends one JSON object (the given comma-separated fields) as a line to the
+// file named by HSA_HOTSWAP_PROOF_LOG. Best-effort and inert when unset; used to
+// prove which kernels were translated/dispatched during a run.
+static void AppendHotSwapProofJson(const std::string& jsonFields) {
+  const char* path = std::getenv("HSA_HOTSWAP_PROOF_LOG");
+  if (!path || !path[0]) return;
+  static std::mutex ProofLogMutex;
+  std::lock_guard<std::mutex> Guard(ProofLogMutex);
+  std::ofstream out(path, std::ios::app);
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `AppendHotSwapProofJson` re-reads `HSA_HOTSWAP_PROOF_LOG` via `getenv` and reopens/closes the file on every event. When the proof log is active during a benchmark this open/close/getenv cycle under the mutex can become a bottleneck; caching the path in a static local (and optionally keeping the `ofstream` open) would avoid it.

projects/rocr-runtime/runtime/hsa-runtime/core/inc/hotswap_env.hpp
```diff
@@ -0,0 +1,37 @@
+////////////////////////////////////////////////////////////////////////////////
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** New file is missing the `SPDX-License-Identifier: MIT` line. Add the standard AMD header:
```cpp
//
```

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap_aql_patch.h
```diff
@@ -0,0 +1,135 @@
+////////////////////////////////////////////////////////////////////////////////
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** New file is missing the `SPDX-License-Identifier: MIT` line — the copyright boilerplate is present but the one-line SPDX identifier should be added.

projects/rocr-runtime/runtime/hsa-runtime/loader/hotswap_kernel_registry.h
```diff
@@ -0,0 +1,344 @@
+////////////////////////////////////////////////////////////////////////////////
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** New file is missing the `SPDX-License-Identifier: MIT` line. Add the standard AMD SPDX header.

projects/rocr-runtime/runtime/hsa-runtime/tests/hotswap_runtime_test.cpp
```diff
@@ -0,0 +1,563 @@
+////////////////////////////////////////////////////////////////////////////////
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** New file is missing the `SPDX-License-Identifier: MIT` line. Add the standard AMD SPDX header.

cfreeamd (commented):
🤖 AI-assisted pre-review (for @cfreeamd) — model: opus

Reviewed the `projects/rocr-runtime/` portion of this PR (clr changes read for context only, not commented). The lazy-presentation machinery is well-structured and the fail-closed invariant (abort rather than dispatch a foreign-ISA kernel) is maintained end-to-end; the feature is cleanly gated behind `ROCR_HOTSWAP_COMGR_ADAPTER` + `ShouldCheckHotSwapDispatchKernelObjects()` so the non-hotswap path is untouched. Registry-vs-record lock ordering is consistent and the blit registration is lifetime-symmetric. Findings are all Optional/Nit — no blocking correctness issues found. `amd_blit_kernel.cpp`, `amd_blit_sdma.cpp`, `amd_loader_context.cpp`, `amd_topology.cpp`, `hotswap.cpp`, `agent.h`, `amd_aql_queue.h`, `executable.hpp`, and the `CMakeLists.txt` changes look correct (mechanical `supported_isas()`→`execution_isas()` substitutions plus the symmetric blit register/unregister pair).

projects/rocr-runtime/runtime/hsa-runtime/tests/hotswap_runtime_test.cpp
martin-luecke: Done in [8358003](https://github.com/ROCm/rocm-systems/pull/8752/commits/83580033bddf0e48e349241b9f3d177b0a2bee0a)

projects/rocr-runtime/runtime/hsa-runtime/loader/hotswap_kernel_registry.h
martin-luecke: Done in [8358003](https://github.com/ROCm/rocm-systems/pull/8752/commits/83580033bddf0e48e349241b9f3d177b0a2bee0a)

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap_aql_patch.h
martin-luecke: Done in [8358003](https://github.com/ROCm/rocm-systems/pull/8752/commits/83580033bddf0e48e349241b9f3d177b0a2bee0a)

projects/rocr-runtime/runtime/hsa-runtime/core/inc/hotswap_env.hpp
martin-luecke: Done in [8358003](https://github.com/ROCm/rocm-systems/pull/8752/commits/83580033bddf0e48e349241b9f3d177b0a2bee0a)

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
martin-luecke: Done and centralized in [54bf1bf](https://github.com/ROCm/rocm-systems/pull/8752/commits/54bf1bf0249d7eb98d66466f9873acee8574837a)

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
martin-luecke: Removed in [fe4edb0](https://github.com/ROCm/rocm-systems/pull/8752/commits/fe4edb0f9eec27fcc17b1fbe13ebe364de2a5d7d)

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hsa.cpp
martin-luecke: Addressed in [2daaaed](https://github.com/ROCm/rocm-systems/pull/8752/commits/2daaaed9daa734e4900108cb44b00154ac749867)

projects/rocr-runtime/runtime/hsa-runtime/loader/hotswap_kernel_registry.h
martin-luecke: Addressed in [6d39624](https://github.com/ROCm/rocm-systems/pull/8752/commits/6d39624b1a0fc165978100434a27eb535322b2f4)

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
martin-luecke: Addressed in 33f5d44: successful lazy translations now publish a terminal translated state, and later source dispatches patch directly to the target kernel without re-entering the translation mutex.

cfreeamd:

All six prior findings (3 Optional + 3 Nit) are resolved, and each fix was verified in the current diff rather than taken from the "Done" replies:

- `amd_aql_queue.cpp` `value >= 0` guard removed; the doorbell range now goes through a fail-closed `ComputeDoorbellPatchRange` with a `uint64_t` cast (fe4edb0).
- `executable.cpp` lazy-translation lock: `record->Kind` promoted to a terminal `Translated` state (release/acquire) with a lock-free fast path; the record mutex is only taken on the first untranslated dispatch (33f5d44).
- `hsa.cpp` load-time gate now shares the cached `IsPresentationModeEnabled()` used on the dispatch side, removing per-load `getenv` (54bf1bf).
- Registry blit fallback now scans a dedicated `rocr_blit_records_` list instead of every registered kernel (6d39624).
- Proof log caches its path and keeps a persistent `ofstream` (2daaaed).
- SPDX-MIT headers added to all four new files (83580033).

The reworked `StoreRelaxed` intercept is new code but is fail-closed and gated behind `ROCR_HOTSWAP_COMGR_ADAPTER` (off by default) — no new blocking concern, and the added invariant comment addresses the earlier reentrancy question. Nothing outstanding from the AI pass.

cfreeamd (approved):

atgutier (commented):
I need to review this more deeply, so please do not merge anything until I signoff.

To answer the discussion question: this really doesn't belong in the runtime at all. This can and should be done almost entirely transparently to the runtimes.

E.g., https://github.com/ROCm/rocm-systems/tree/develop/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/hooks

stellaraccident (changes requested):
Thanks for attempting to drill through to solution and get to POC, but can you please coordinate with Tony Gutierrez offline on the supportable way to do this.

atgutier (approved):

stellaraccident (commented):
Thanks for syncing

atgutier (commented):
I worked with an agent to generate this review alongside my own planning on how to do this with less intrusion into rocr. I am not a clr expert and can't really speak to that part. Please work with Saleel or German there on how they'd like this architected in.

There are some immediate issues that should be addressed. If they are addressed I'll approve this conditioned on future refactoring to move most of this into comgr as an API tool that ROCr can load.

Here are the findings in the agent's words:

# PR 8752 Review: Lazy Per-Kernel ISA Presentation

- PR: `ROCm/rocm-systems#8752`
- Title: `feat(hotswap): add lazy per-kernel presentation mode`
- Base: `develop` at `74420f3522ce7509975c44c1eba0889a48804597`
- Head: `2fb7008f9ead6acf9cfd12845660bfc2ec1f9c3d`
- Scope: 39 files, +2,754/-273
- Review method: read-only GitHub full-file inspection because the PR head was not available locally

## BLOCKER

### Multi-producer publication can execute an unpatched source kernel

**Location:** `projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp:498`

**Reasoning:** `HSA_QUEUE_TYPE_MULTI` producers may reserve adjacent queue slots and publish them out of reservation order. If producer B publishes and rings index 11 while producer A's index 10 remains invalid, the scanner advances its high-water mark to 11. When A later publishes index 10, hardware is already armed through index 11 and may execute index 10 before A's lower doorbell store reaches the scanner.

The existing doorbell mutex serializes scanner invocations, but it does not serialize packet reservation and publication. Rescanning at the later doorbell is not sufficient because the hardware may observe the newly valid packet before that doorbell call begins.

**Evidence:**

- `amd_aql_queue.cpp:513-540` scans through the largest observed doorbell and advances `last_doorbell_value_`.
- `core/runtime/hotswap_aql_patch.h:36-39` skips every later `end <= last_doorbell`.
- `loader/hotswap_kernel_registry.h:207-211` accepts only `HSA_QUEUE_TYPE_MULTI`, where concurrent publication is expected.
- `tests/hotswap_runtime_test.cpp:183-192` codifies the lower-doorbell replay skip but does not test legitimate out-of-order publication.

**Suggested fix:** Do not use a doorbell high-water mark as proof that every earlier generation was published and inspected. A general solution must intercept before the packet header becomes valid. For the deadline milestone, constrain the feature to a proven serialized reservation-to-publication-to-doorbell protocol, such as dedicated/single-producer queues or serialization covering every producer sharing the physical queue. Add a two-producer test in which the higher index publishes and rings first.

**Confidence:** High.

### Direct hardware doorbells bypass translation

**Location:** `projects/clr/rocclr/device/rocm/rocvirtual.cpp:1402`

**Reasoning:** The new scanner runs in `AqlQueue::StoreRelaxed`, but CLR can write the hardware MMIO doorbell directly. Normal, graph/batch, and scheduler submissions can therefore send a source-ISA kernel object to hardware without executing the scanner.

The bypass also remains available to non-CLR callers through `HSA_AMD_QUEUE_INFO_DOORBELL_ID`; changing only the known CLR call sites would not close the runtime contract.

**Evidence:**

- ROCr exposes the raw doorbell pointer at `projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp:575-583`.
- CLR obtains it for ordinary queues at `projects/clr/rocclr/device/rocm/rocdevice.cpp:3583-3595`.
- Normal submission bypasses the HSA signal at `rocvirtual.cpp:1402-1407`.
- Graph/batch submission bypasses it at `rocvirtual.cpp:1929-1933`.
- Scheduler submission obtains and uses another raw doorbell at `rocvirtual.cpp:4304-4311` and `rocvirtual.cpp:4351-4357`.
- No presentation-mode check disables these paths.

**Suggested fix:** Refuse `HSA_AMD_QUEUE_INFO_DOORBELL_ID` while presentation mode is active. Also defensively force CLR's ordinary and scheduler doorbell pointers to null in that mode. Add normal, graph/batch, and scheduler tests with direct-doorbell configuration enabled.

**Confidence:** High.

### Per-kernel child executables do not share application-visible global state

**Location:** `projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp:809`

**Reasoning:** Each translated kernel is loaded into a new isolated `ExecutableImpl`. The child does not inherit the parent executable's external definitions, loaded data-segment addresses, writable globals, or runtime-updated variables. A translated kernel can consequently fail relocation or access a duplicate child allocation disconnected from application symbol operations.

Device-enqueue runtime handles are a concrete example: CLR updates the variable in the source executable, while no corresponding update is made to the translated child.

**Evidence:**

- `executable.cpp:809-829` constructs, loads, and freezes a new isolated executable.
- The COMGR call at `executable.cpp:729-762` supplies source bytes, ISA names, and a kernel name, but not the parent executable's relocated symbol addresses.
- `ExecutableImpl::DefineProgramExternalVariable` and `DefineAgentExternalVariable` retain definitions in an individual executable at `executable.cpp:1708-1760`; those definitions are not copied into the child.
- CLR writes `RuntimeHandle` into the source executable at `projects/clr/rocclr/device/rocm/rockernel.cpp:99-144`.

**Suggested fix:** Have COMGR return authoritative per-kernel eligibility and dependency information. Restrict lazy translation to kernels certified as self-contained. Code involving internal or external writable globals, runtime handles, device enqueue, function-pointer tables, unresolved externals, or other shared state must use whole-object translation or be rejected before dispatch.

**Confidence:** High for external variables and runtime handles; high that retained internal globals receive separate storage if included in the per-kernel output.

## MAJOR

### An exception can leave a packet invalid and silently wedge the queue

**Location:** `projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap_aql_patch.h:76`

**Reasoning:** The patch helper publishes an invalid header before invoking translation and restores the header only on normal return. Translation allocates vectors, strings, executables, and logging state, any of which can throw. The public signal-store API catches exceptions and returns `void`; the caller receives no failure while the packet remains invalid and the hardware doorbell is not rung.

**Evidence:**

- Header invalidation and normal-only restoration are at `hotswap_aql_patch.h:76-90`.
- Translation allocates output buffers and an executable at `loader/executable.cpp:794-829`.
- Retained state is appended at `executable.cpp:887-890`.
- Proof-log construction can allocate at `executable.cpp:940-949`.
- `hsa_signal_store_relaxed` catches exceptions at `core/runtime/hsa.cpp:1230-1236`.
- The manual thread-local state change at `executable.cpp:815-819` is not exception-safe either.

**Suggested fix:** Complete translation before entering the short packet-rewrite section and make that section non-throwing. At minimum, catch all exceptions inside the patch helper and enter a deterministic fail-closed path rather than unwinding through the void signal API. Use RAII for the thread-local state change.

**Confidence:** High.

### Translated children and registry entries survive final shutdown

**Location:** `projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp:142`

**Reasoning:** Both the registry and translated-executable collection are heap-allocated process statics. Translated children are appended but never removed. Final runtime shutdown destroys the normal loader and agent resources without clearing this state. A later initialization retains invalid agent handles and kernel addresses, while address reuse can trigger the registry's duplicate-address abort.

**Evidence:**

- The process-static registry and executable vector are created at `executable.cpp:142-159`.
- Translated children are retained at `executable.cpp:886-891`.
- Duplicate registrations abort at `loader/hotswap_kernel_registry.h:251-265`.
- `Runtime::Unload` destroys the normal loader and later releases agent resources at `core/runtime/runtime.cpp:2708-2747` without translated-child or registry finalization.
- `Loader::Destroy` only resets debugger state and explicitly does not delete executables or loaded objects at `executable.cpp:1069-1075`.

**Suggested fix:** Add an explicit finalization operation that destroys translated children and clears registry state while the relevant agent resources remain valid. If safe finalization cannot land under the immediate deadline, make this mode explicitly one-shot and reject reinitialization after any lazy translation instead of reusing stale state. Full cleanup must also be ordered with queue and tool teardown.

**Confidence:** High.

### Executing target code is absent from normal debugger and profiler discovery

**Location:** `projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp:829`

**Reasoning:** The translated child is frozen directly rather than through `AmdHsaCodeLoader::FreezeExecutable`, and it is deliberately kept outside the loader's executable collection. Normal debug-map publication and executable enumeration expose the source object while PCs execute from the hidden target object.

Every child also receives the same logical URI, which is not a stable reopenable identity for the translated ELF.

**Evidence:**

- The child ID is deliberately outside the loader namespace at `executable.cpp:162-165`.
- It is loaded with the constant URI `hotswap-lazy-kernel` at `executable.cpp:817-818`.
- It calls `targetExec->Freeze` directly at `executable.cpp:829`.
- Normal freeze publishes loaded objects into the debugger map at `executable.cpp:1130-1147`.
- Normal executable iteration sees only the loader-owned executable vector.

**Suggested fix:** Route child load, freeze, enumeration, and destruction through an explicit lifecycle that produces normal load/unload notifications. Give each translated object a stable unique identity derived from COMGR's output/cache identity. Until this is qualified, debugger and profiler use should be declared unsupported and incompatible tool configurations should be gated rather than described as transparent.

**Confidence:** High for loader/debug-map invisibility; exact profiler behavior also depends on the installed tool.

### The configured target is not bound to the physical agent ISA

**Location:** `projects/rocr-runtime/runtime/hsa-runtime/loader/hotswap_kernel_registry.h:173`

**Reasoning:** Any explicit target is accepted without comparing it to the physical agent. CLR then uses that environment-selected ISA for hardware-adjacent settings. In a mixed-GPU system, the same global target is applied to every enumerated GPU. Failure may occur only when the translated child is loaded, after CLR has already made hardware decisions using the wrong ISA.

**Evidence:**

- ROCr establishes the physical ISA at `core/runtime/amd_gpu_agent.cpp:197-206`.
- `ResolveHotSwapPresentationTarget` accepts any explicit target without comparing it to `execution_gfx` at `hotswap_kernel_registry.h:173-193`.
- CLR similarly accepts any distinct known target at `projects/clr/rocclr/device/rocm/rocdevice.cpp:117-141`.
- CLR installs that value as `executionIsa` at `rocdevice.cpp:700-721`.

**Suggested fix:** Validate during agent initialization that the selected target exactly matches that agent's physical execution ISA. Define mixed-system behavior now: either present only matching agents and leave other agents unchanged, or fail initialization globally with a clear diagnostic.

**Confidence:** High.

### Translation failure has only a process-abort contract

**Location:** `projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp:899`

**Reasoning:** Unsupported input, COMGR failure, registry mismatch, or resource overflow calls `std::abort`. The queue's registered error callback, completion signal, later packets, and repeat-dispatch behavior are not handled. This is fail-closed for hardware safety, but it is not a general runtime error contract.

**Evidence:**

- Dispatch rejection funnels through `FatalHotSwapDispatch` at `executable.cpp:899-925`.
- Translation and segment failures invoke it at `executable.cpp:958-1028`.
- `AqlQueue` retains an error callback and callback data, but the new failure path does not use them.

**Suggested fix:** Process abort can be accepted for a narrowly scoped deadline milestone if that restriction is explicit and tested end-to-end. Track a proper queue-poison/error-callback contract covering completion signals, packets behind the failure, subsequent doorbells, repeated dispatches, queue destruction, and tool-visible status.

**Confidence:** High.

## MINOR

### Static packages do not propagate their COMGR dependency

**Location:** `projects/rocr-runtime/runtime/hsa-runtime/CMakeLists.txt:266`

**Reasoning:** COMGR is linked privately to the concrete runtime target. Static packaging whole-archives the target and manually reconstructs its public dependencies, but COMGR is absent from that list. Static consumers can therefore receive unresolved `amd_comgr_hotswap_*` symbols or an incomplete packaged dependency graph.

**Evidence:**

- The private COMGR link is at `CMakeLists.txt:266-274`.
- The static interface bypasses the concrete target's link requirements and reconstructs dependencies at `CMakeLists.txt:461-485`.
- The reconstructed list contains hsakmt and platform libraries but not COMGR.

**Suggested fix:** When the adapter is enabled, add COMGR to the static interface and exported package dependency metadata.

**Confidence:** High.

### The new unit test is POSIX-only but not platform-gated

**Location:** `projects/rocr-runtime/runtime/hsa-runtime/CMakeLists.txt:276`

**Reasoning:** Enabling the COMGR adapter defaults the unit test on for every platform, but the test uses `fork`, `waitpid`, and POSIX headers.

**Evidence:**

- The option and test target are defined at `CMakeLists.txt:276-294`.
- POSIX headers appear at `tests/hotswap_runtime_test.cpp:16-17`.
- Multiple death tests invoke `fork` and `waitpid`.

**Suggested fix:** Gate the current test target on UNIX or port the death tests to a cross-platform facility.

**Confidence:** High.

## Immediate Deadline-Safe Requests

The enabled feature should not merge until the following contained corrections are made:

1. Close both dispatch escape paths:
   - out-of-order multi-producer packet publication;
   - direct hardware doorbell writes.
2. Add an authoritative COMGR eligibility result and reject or eagerly translate objects that require shared program state.
3. Make packet invalidation and the translation-mode thread-local state exception-safe.
4. Bind the configured target ISA to the physical agent during initialization.
5. Establish a minimum safe shutdown/reinitialization contract, even if the first milestone is explicitly one-shot.
6. Propagate COMGR correctly in static packages and platform-gate the POSIX test.

For a tightly scoped initial milestone, process abort on translation failure, a homogeneous-system restriction, and lack of debugger/profiler support can be temporarily accepted only if those limitations are explicit, tested, and not described as transparent production behavior.

## COMGR Ownership Changes Worth Making Now

The following are narrow changes that improve the current implementation without requiring the full architectural refactor:

- Remove the duplicated gfx/e_flags table at `loader/executable.cpp:1033-1060`. `SourceMach` is logged and stored but not consumed by translation. COMGR should authoritatively accept or reject source/target/kernel combinations.
- Move the hard-coded `lazy-v2` cache namespace at `loader/hotswap_kernel_registry.h:196-205` into COMGR. Cache format, namespace, and schema evolution belong with the translator.
- Return per-kernel eligibility and dependency information covering globals, undefined symbols, runtime handles, device enqueue, function-pointer data, and other whole-object requirements.
- Expose a stable translator/output/cache identity for naming translated objects and associating tool records.
- If practical, introduce a COMGR translation-session handle so source bytes and parsed state do not need to be duplicated between the loader and every lazy translation operation.

The queue interception, physical-agent validation, AQL packet rewriting, loaded GPU allocation, and runtime finalization should not be moved wholesale into core COMGR. Those operations require runtime state that an offline translation library does not own.

## Required Future API-Table Hook Refactor

The current ROCr guest/presented-ISA implementation should be treated as a deadline-oriented adapter, not the permanent architecture. A named follow-up should require moving presentation and translation orchestration into a COMGR-backed HSA API-table hook component.

The intended ownership should be:

- **Core COMGR:** translation semantics, eligibility and dependency analysis, source/target validation, cache identity, cache format, and translated ELF production.
- **COMGR-backed hook DSO:** presented-ISA policy, HSA API-table wrapping, code-object interception, translation orchestration, source/target mappings, retained translated storage, caching policy, and hook-owned lifecycle state.
- **ROCr:** the authoritative physical device model, ordinary target-ISA loading, and narrowly scoped generic hook/lifetime facilities.
- **CLR:** ordinary bundle selection from the presented view, with only narrowly required selection and direct-doorbell safeguards.

The hook implementation should be a separate DSO rather than part of core COMGR. COMGR is used by compilers and offline clients with no active HSA runtime; core COMGR should not acquire dependencies on HSA API tables, queues, profilers, or shutdown ordering.

### Hook refactor requirements

The tracked refactor must cover:

1. **Hook installation ordering.** Identity, code-object-reader, and load hooks may need installation before profiler/tool registration. Destruction and finalization coordination may need an outer post-registration wrapper.
2. **Complete reader interception.** Cover core memory/file readers and AMD vendor readers obtained through extension-table queries, including version, table-size, field-presence, concurrency, and handle-lifetime rules.
3. **Presented identity outside ROCr agents.** Implement logical ISA presentation through API-table query wrapping rather than permanent guest-ISA fields in ROCr's physical agent objects.
4. **Eager whole-object translation first.** The initial hook-owned path should translate a complete object before it reaches the ordinary ROCr loader. This avoids the current per-kernel global-state split.
5. **A safe lazy publication point.** Lazy translation must not rely on a doorbell-only scan for unrestricted multi-producer queues. It requires an earlier packet-publication hook, a proxy queue, or a fully proven serialized producer protocol.
6. **Complete packet formats.** Support both standard kernel-dispatch packets and AMD extended dispatch packets, rewriting kernel object and resource fields while preserving dependency and cluster metadata.
7. **Direct-doorbell closure.** API-table hooks cannot observe raw MMIO writes. Direct doorbell exposure must remain disabled whenever lazy interception is required.
8. **Translated storage identity.** Provide stable retained file or memfd storage when debuggers reopen code objects by URI. Preserve the source URI separately as provenance.
9. **Debugger and profiler identity.** Publish load/unload records for the target code that actually executes, with correct symbol, range, PC-attribution, and lifetime behavior.
10. **Shutdown rendezvous.** Coordinate queues, tools, translated executables, retained storage, and DSO lifetime before freeing state. Returning from ordinary `hsa_shut_down` is not by itself proof that tools no longer retain code-object records.
11. **Removal of guest state from ROCr.** After migration, remove `presented_isa_`, `execution_isas()`, translation-specific registries, and translation-specific environment policy from ROCr. ROCr should return to a physical-ISA-only model.

### Target data flow

The safest first hook-owned milestone is eager whole-object translation:

```text
application / CLR
       |
       | source code-object load
       v
COMGR-backed HSA hook DSO ----> core COMGR translation/cache
       |
       | target-ISA ELF
       v
ordinary physical ROCr loader
```

In this model, ROCr does not need to maintain a guest ISA, translate code, or create hidden per-kernel executables. Lazy dispatch can be added later after publication ordering, lifetime, and tool rebinding are qualified.

## Tracked Follow-Up TODOs

The following can be deferred if the first milestone is explicitly restricted:

- Implement the COMGR-backed API-table hook DSO and its autoload/registration contract.
- Add whole-object fallback and shared data-symbol handling.
- Replace process abort with a complete queue-error and completion-signal state machine.
- Qualify debugger/profiler load, unload, URI, and PC attribution.
- Add full mixed-GPU behavior rather than a homogeneous-system restriction.
- Add a COMGR translation-session API and eliminate duplicate source-buffer ownership.
- Remove ROCr guest/presented-ISA fields and translation-specific policy after hook migration.
- Qualify standard and extended packets, graphs, scheduler queues, direct-doorbell rejection, concurrent producers, writable globals, device enqueue, shutdown/reinitialization, and tool attribution on real hardware.

## Verdict

The PR is a useful proof of the lazy translation flow and already places machine-code generation in COMGR, but the enabled feature is not safe to merge unchanged. The multi-producer publication race and direct-doorbell bypass can execute source-ISA code on the physical target. The isolated per-kernel child also lacks a valid shared-state contract for common program globals and runtime handles.

The deadline-compatible path is to fix the dispatch blockers, add COMGR-owned eligibility, make the patch path exception-safe, validate the physical target, and establish a minimum lifecycle restriction. The larger API-table hook refactor should be recorded as an explicit architectural requirement, with removal of ROCr guest-ISA state as a completion criterion.

cfreeamd (commented):
🤖 AI-assisted pre-review (for @cfreeamd) — follow-up after `hotswap: bind presentation target to physical agents`

Reviewed the new commit. It makes the physical agent's execution ISA authoritative for the translation target and adds `hotswap::ValidateHotSwapPresentationTarget`, which validates the configured `HSA_HOTSWAP_TARGET`/`ISA_OVERRIDE` against each GPU agent's real ISA at construction (processor-name equality + `Isa::IsCompatible(configured, execution, 0)`), failing init closed on mismatch with a clear diagnostic. Checked the details: the `IsCompatible` argument order (code-object=configured, agent=execution, 0=non-generic) is correct; the new call site is correctly inside the `#ifdef ROCR_HOTSWAP_COMGR_ADAPTER` branch (no build break); CMake adds `isa.cpp` to the unit-test target for the new dependency; and dropping the `"0"` fallback in the non-presentation branch is safe because `EnvGfxName` already normalizes `"0"` to empty. Tests were rewritten to cover invalid-ISA and feature-mismatch cases. All gated behind `ROCR_HOTSWAP_COMGR_ADAPTER` (off by default). No new concerns.

atgutier (changes requested):
I need to review this in depth again, but I think the isa presentation mode stuff is not needed urgently either, correct?

We're limiting the scope of this to A0/B0 and both are gfx1250 ISAs.

martin-luecke:
@atgutier The cross-presentation stuff is exactly the part that is urgent.
I will convert this to a draft PR until I have addressed all of your concerns, so no reviewing effort is wasted on an intermediate state.
I will ping when this is ready for another round of reviews.

atgutier (changes requested):
Please make this a standalone tools lib.

PR #8755, opened 2026-07-17
[CLR] Add -Werror and -Wall while resolving warnings

AidanBeltonS:
## Motivation

Improve the quality of code in the CLR repository by making -Werror and -Wall default. This build removes warnings for CLR built with both GCC and Clang

## Technical Details

  **Trivial fixes (the vast majority):**
  - Unused variables — removed dead locals, or kept the call and dropped the assignment when the RHS had side effects (e.g. getDeviceMemory, event->ready(),
   pinSystemMemory), or marked [[maybe_unused]] for deliberate volatile reads / discard cases.
  - -Wsign-compare — changed loop counters int→size_t/uint32_t, or added static_cast<size_t>(...) on device-id/index comparisons against .size().
  - -Wreorder — reordered constructor initializer lists to match member declaration order (in command.hpp, device.cpp, rocdevice.cpp, rocvirtual.cpp,
  hip_global.cpp, hip_graph_internal.hpp, etc.), including moving base-class init first.
  - -Wunused-function — marked genuinely dead file-local functions [[maybe_unused]].
  - -Wnodiscard — added (void) casts on ignored hipError_t returns.
  - -Wsequence-point UB in rocvirtual.cpp: active_chunk_ = ++active_chunk_ % n → (active_chunk_ + 1) % n.

  **Less-obvious changes:**  
  - -Wparentheses in hip_memory.cpp: fixed a mis-grouped assert(a || b && "msg").
  - -Wswitch in rocdevice.cpp: switched over static_cast<int>(hsa_status) so extended HSA enum cases are accepted; in hip_memory.cpp added the unreachable
  hipWriteBuffer/hipReadBuffer cases.
  - -Wsizeof-pointer-memaccess in hip_device.cpp: replaced strncpy(dst, src, sizeof(src)) with a bounded snprintf.
  - Generated file hip_playback_generated.cpp: fixed the generator (gen_hrr_api_args.py) to emit [[maybe_unused]] for the a payload cast, then applied the
  same to the committed generated file (201 sites).

  **Functions with implicitly ignored returns set to [[no_discard]] and explicity ignored warnings**
  - amd::Memory::getDeviceMemory     
  - roc::Memory::pinSystemMemory 
  - roc::Device::getRocMemory 
  - device::BlitManager::copyBufferToImage
  - device::BlitManager::copyImageToBuffer
  - hip::Event::ready (+EventDD override) 
  - GraphMemAllocNode::Execute

  **Clang -Werror fixes**
  - -Winconsistent-missing-override: added override to ~40 methods in rocvirtual.hpp, command.hpp, and rocblit.hpp (KernelBlitManager).
  - -Wmissing-braces: = {0} → = {} on struct value-inits (8 sites).
  - -Wunused-private-field: [[maybe_unused]] on genuinely-retained members (kept, per your instruction).
  - -Woverloaded-virtual: added using-declarations for hidden base SetParams/initClBinary.
  - -Wmismatched-tags: made Graph/GraphNode/UserObject/ProfilingSignal forward-decls consistently class.
  - -Wunknown-attributes: gated the GCC-only optimize attribute behind defined(__GNUC__) && !defined(__clang__).
  - Unused functions / const-variables / lambda capture: [[maybe_unused]] or removed as appropriate.
  - Real bug fixed — -Wsometimes-uninitialized: node was read uninitialized in the ExtSemaphore/default switch cases (hip_graph.cpp); initialized to
  nullptr.
  - Real bug fixed — -Wtautological-overlap-compare: the graph kernel-node priority check priority < Low || priority > High was always true (enum is
  High=-1..Low=1); operands corrected to < High || > Low.

  **Deleted unused and unreserved functions and members**
  - constexpr auto ComputeTableSize(size_t num_funcs);
  - constexpr char const* OFFLOAD_KIND_HIPV4 = "hipv4";
  - constexpr char const* OFFLOAD_KIND_HCC = "hcc";
  - static inline bool isGenericTarget(const void* image);
  - unsigned int flags_;
  - size_t size_;     //<! Allocated size
  - static constexpr uint32_t kAnn_uint = 3;  // uint64 uint_value
  - static constexpr clk_value_type_t ClkValueMapType[6][6];
  - static constexpr BlockNameIdMapType viBlockIdOrcaToRocr[];
  - static constexpr hsa_barrier_and_packet_t kBarrierAcquirePacket;
  - static constexpr hsa_barrier_and_packet_t kBarrierReleasePacket;
  - address kernArgOffset_ = nullptr;  //!< KernelArg buffer to used when graph capturing is enabled
  - EGLContext eglInternalContext_;
  - EGLContext eglTempContext_;
  - int type_;       //!< Channel type (textures/surfaces only)
  - int norm_;       //!< Normalisation flag (textures/surfaces only)
  - isCompatibleCodeObject
  - maxThreadsPerCU
  - EmitSlice

**Doxygen**
  - Makes doxygen warnings fail build
  - Resolves existing doxygen issues

## Issue Tracking

AIRUNTIME-2507

## Test Plan

N/A

## Test Result

N/A

## Submission Checklist


g-h-c:
`-Werror` is great for developers and all of them should use in the development environments by default.
But for Linux distribution maintainers, `-Werror` could be nightmare, because porting your code to a new system with a different compiler version might fail due to new warnings added to it.
Thusly, how easy is it to disable `-Werror` for the whole of `rocm-systems`?

AidanBeltonS:
> `-Werror` is great for developers and all of them should use in the development environments by default. But for Linux distribution maintainers, `-Werror` could be nightmare, because porting your code to a new system with a different compiler version might fail due to new warnings added to it. Thusly, how easy is it to disable `-Werror` for the whole of `rocm-systems`?

Good suggestion. I have added `CLR_ENABLE_WERROR` and also documented it in the README.md this creates a single point for this to be enabled or disabled

projects/clr/CMakeLists.txt
```diff
@@ -41,6 +41,33 @@ endif()
 option(CLR_BUILD_HIP "Build HIP" OFF)
 option(CLR_BUILD_OCL "Build OCL" OFF)
 
+# Strict warnings default to ON for Linux, OFF elsewhere (MSVC has its own /wd
+# flags). Defined here so the clr_warnings target below is visible to every
+# subtree added via add_subdirectory().
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+    set(_clr_werror_default ON)
```
g-h-c: - The default should be OFF.

- And then in development environments and CI/CD it should be ON.

This is to allow people to use different compilers to build the code with less problems. I know the compiler version is pinned in TheRock's folder `compiler/` but I am not sure we can assume people will use always TheRock.

It is debatable. Some references:

https://flameeyes.blog/2009/02/25/future-proof-your-code-dont-use-werror/

https://embeddedartistry.com/blog/2017/05/22/werror-is-not-your-friend/

AidanBeltonS: So I have had a look at the resources and our build system in this context. Ill re-state the resources arguments.

1. Werror creates a dependency between our library and the specific compiler version and type (gcc, clang, msvc, etc)
2. Werror in the CI doesn't allow the whole run to complete so there can be hidden failures found after the ones addressed are resolved.

The ways they recommend resolving this is:
1. Have -Werror off by default
2. If you can run CI without -Werror then parse the log for warnings and then fail if any are found.

This is a debatable option as you have pointed out. I think we run into two cases
1. I don't expect people to be building clr from source outside of AMD often, as you need some detailed knowledge on how to stitch all the bits of the project together so they build and run correctly. If you do have these skills then setting -DWERROR_ENABLE=off is a trivial task. 
2. The fix of having this default to off and then turned on for the CI runs into a different issue. This would have to be hardcoded into TheRock and difficult to enable or disable, defeating the aim.

I personally think enabling it by default and making it easy to disable is the right way to go for CLR in isolation. The question for me is how do we make it easy to disable/enable in TheRock? As that is closer to what we would expect someone to build from source to use

TLDR; The Rock uses CLR defaults, which can be overridden but currently don't have a good way to expose enabling/disabling it. My suggestion is to have it on by default and then monitor if it is an issue and if it is we wire in a way to turn it on/off in TheRock and adjust the CLR default to off

g-h-c (approved):

projects/hip/docs/doxygen-input/doxy.cfg
```diff
@@ -816,7 +816,7 @@ WARN_NO_PARAMDOC       = NO
 # a warning is encountered.
 # The default value is: NO.
```
g-h-c, line 817: Default value is now YES

AidanBeltonS: That is intended, many of the warnings were from doxygen. So when user facing/documented objects are changed the docs must at least build cleanly

lumachad:
@AidanBeltonS Why is this labelling rocdbgapi and rocr-debug-agent?

AidanBeltonS:
> @AidanBeltonS Why is this labelling rocdbgapi and rocr-debug-agent?

This was an error after a rebase, I have resolved the issue

mangupta (approved):

projects/clr/hipamd/src/hrr/playback/hip_playback_generated.cpp
```diff
@@ -130,30 +130,31 @@ static hipError_t playback_hipArrayGetInfo(PlaybackContext& ctx, const uint8_t*
 }
 
 static hipError_t playback_hipBindTexture(PlaybackContext& ctx, const uint8_t* payload) {
-  const auto* a = reinterpret_cast<const hrr_args_hipBindTexture*>(payload);
+  [[maybe_unused]] const auto* a = reinterpret_cast<const hrr_args_hipBindTexture*>(payload);
```
AlexeySachkov, line 133: For my own education: why is it marked as `maybe_unused`? Is differencing it while passing to `hipBindTexture` is not a good enough use, or `hipBindTexture` here is a ruse of its own which doesn't constitute a use and compiler sees through it?

projects/clr/rocclr/compiler/lib/utils/options.cpp
```diff
@@ -1354,7 +1353,7 @@ static std::string getValidDumpPath(const std::string& path) {
   }
 
   const std::string curPath = path + ".";
-  long pathname_max = pathconf(curPath.c_str(), _PC_PATH_MAX);
+  [[maybe_unused]] long pathname_max = pathconf(curPath.c_str(), _PC_PATH_MAX);
```
AlexeySachkov: Huh, our clang-format configuration allows trailing whitespaces?

projects/clr/rocclr/device/rocm/rocvirtual.cpp
```diff
@@ -1138,9 +1126,8 @@ bool VirtualGPU::processMemObjects(const amd::Kernel& kernel, const_address para
               amd::Image* img = mem->asImage();
 
               // Copy memory from the original image buffer into the backing store image
-              bool result =
-                  blitMgr().copyBufferToImage(*devBuf, *devCpImg, offs, offs, img->getRegion(),
-                                              true, img->getRowPitch(), img->getSlicePitch());
+              (void)blitMgr().copyBufferToImage(*devBuf, *devCpImg, offs, offs, img->getRegion(),
```
AlexeySachkov, line 1132: Not even `assert` that the copy was successful?

projects/clr/rocclr/utils/nontemporal.hpp
```diff
@@ -243,7 +243,14 @@ namespace amd {
 
 // ================================================================================================
 #if IS_LINUX
-__attribute__((optimize("unroll-all-loops"), always_inline)) static inline void nontemporalMemcpy(
+#if defined(__GNUC__) && !defined(__clang__)
+// The 'optimize' attribute is GCC-specific; Clang does not recognize it and warns
```
AlexeySachkov: Why not use [`__has_attribute`](https://gcc.gnu.org/onlinedocs/gcc-12.1.0/cpp/_005f_005fhas_005fattribute.html)? Clang [supports](https://clang.llvm.org/docs/LanguageExtensions.html#has-attribute) it as well

AidanBeltonS: Good shout, done.

projects/clr/rocclr/device/rocm/rocvirtual.cpp
AidanBeltonS: There are many cases where we ignore the results. This change doesn't re-write them as I want to minimize functional changes. I felt it was best to mark the returned type as no discard and then add (void). My hope is that the code smell will cause people to fix the cases, rather than have one very large change which tries to re-structure all the error handling and risk changing behavior. An assert would be possible, but then it hides the smell, and raises the question of why not just fully fix the issue. This is a debatable case, one could argue a few different approaches, I am open to suggestions

projects/clr/rocclr/compiler/lib/utils/options.cpp
AidanBeltonS: Seems so, thanks for the catch. Ill add it to my notes to follow up on, clang-format ignoring it, as its worth catching these cases

projects/clr/hipamd/src/hrr/playback/hip_playback_generated.cpp
AidanBeltonS: In some but not all cases it is used, `hip_playback_generated.cpp` is an automatically generated set of functions from `projects/clr/hipamd/src/hrr/gen_hrr_api_args.py` so some functions do use `a` but others do not.

So I have applied `[[maybe_unused]]` to the script which generates this so it catches both cases where it is used and it is ignored. The alternative is making this not a generated file, but I think that is outside the scope of this change

AlexeySachkov: Yeah, I could have noted that it was generated, makes sense then.

litvaOo (approved):

.github/workflows/therock-ci-linux.yml
```diff
@@ -87,7 +87,7 @@ jobs:
         env:
           amdgpu_families: ${{ env.AMDGPU_FAMILIES }}
           package_version: ${{ inputs.package_version }}
-          extra_cmake_options: "-DTHEROCK_ROCM_SYSTEMS_SOURCE_DIR=../ -DTHEROCK_ENABLE_RCCL=OFF ${{ inputs.cmake_options }}"
+          extra_cmake_options: "-Dhip-tests_CMAKE_ARGS='-DHIP_TESTS_ENABLE_WERROR=ON' -Dhip-clr_CMAKE_ARGS='-DCLR_ENABLE_WERROR=ON' -DTHEROCK_ROCM_SYSTEMS_SOURCE_DIR=../ -DTHEROCK_ENABLE_RCCL=OFF ${{ inputs.cmake_options }}"
```
geomin12: we should make this adjustment in TheRock (https://github.com/ROCm/TheRock/blob/fc11b46e08ebbfbb406bad9d081454d5c7c81aae/core/CMakeLists.txt#L575). this code should be agnostic and not allow these component specific conditionals

few reasons:
- this will be added to every single project build, even if it is not needed. it's wasted CI resources and time on developers.
- these workflows will be decommissioned shortly in favor of multi-arch (EOW 8/27). whatever is "hard-coded" in here will not be in use

geomin12 (changes requested):

.github/workflows/therock-ci-linux.yml
geomin12: ah more helpful, you can add it here: https://github.com/ROCm/rocm-systems/blob/develop/.github/scripts/therock_matrix.py so it is scoped per project

PR #8756, opened 2026-07-17
[rocprofiler-compute] Detect conflicting rocm installations before profiling

ggottipa-amd:
## Motivation

A workload that carries its own ROCm, such as a PyTorch or pip ROCm wheel, can load ROCm libraries that the profiler also loads. Two copies of `libamd_comgr` in one process abort the run, and two copies of a rocprofiler library cannot both register with rocprofiler.
This PR detects the situation before profiling, reports it, and explains a failed run. 

## Technical Details

**Detection** (`src/utils/rocm_stack_check.py`): runs once before each profile run, except one started with `--attach-pid`.
- Profiler's ROCm: the installation holding the rocprofiler-sdk tool, otherwise `ROCM_PATH`, otherwise `/opt/rocm`.
- Workload copies: `LD_LIBRARY_PATH`, `LD_PRELOAD`, the executable's RPATH/RUNPATH with `$ORIGIN`, `$LIB` and `$PLATFORM` expanded, and the site-packages of a Python workload's interpreter.
- Not a conflict: a copy under the profiler's ROCm root, one that resolves to the same file or holds identical contents, and one named in `LD_PRELOAD`. This keeps a single ROCm split across directories, such as the `_rocm_sdk_core` and `_rocm_sdk_devel` halves of a wheel, from being reported.

**Reporting**: a conflict is a warning naming the library and the path each side loads it from, with each path labelled as found on a directory the loader searches or in site-packages, which the loader does not search. Profiling continues.

**Signal handlers**: any second copy, conflicting or not, sets `ROCPROF_SIGNAL_HANDLERS=0` for the run so that an abort ends the process rather than re-entering rocprofv3's handlers. A value already in the environment is kept. The run reports the change.

**Failed run**: where rocprofv3's output reports the conflict itself, through the LLVM or the rocprofiler registry message, that is repeated with the paths the check found. Where it does not, the copies are named as a possible cause only if the run was aborted. Both messages state how to get onto a single ROCm.

**Tests**
  - `torch_trace` and `triton_trace` ctests that load the second ROCm via torch, fail in dual stack environments. 
  -  `tests/test_rocm_stack_check.py` adds 23 tests and `tests/test_utils_profile.py` adds 2 tests. Both are registered as ctests and listed in all four categories in `tests/test_categories.yaml`.
  
## Issue Tracking

JIRA ID: ROCM-25860

## Test Plan

```
ctest -R '^(test_rocm_stack_check|test_utils_profile)$' --output-on-failure
```

## Test Result

The failing scenario from the defect ticket, a torch workload whose wheel ROCm conflicts with the system ROCm, now reports the conflicting libraries and completes instead of hanging.

image

Tests Pass

## Submission Checklist


abchoudh-amd (approved):
LGTM, except a few style comments which we can discuss offline.

vedithal-amd (changes requested):
Will review soon

vedithal-amd:
@ggottipa-amd rebased onto the latest `rocprofiler-compute-develop` and force-pushed to fix the conflicts on this PR.

vedithal-amd (changes requested):
This belongs in amd-comgr, not here.

comgr statically links SPIRV-LLVM-Translator, whose `lib/SPIRV/SPIRVReader.cpp:108` declares a file-scope `cl::opt`:

```cpp
cl::opt<bool> SPIRVEnableStepExpansion(
    "spirv-expand-step", cl::init(true),
    cl::desc("Enable expansion of OpenCL step and smoothstep function"));
```

That constructor registers into the shared `libLLVM`'s process-global option registry (`libamd_comgr.so.3` exports 0 LLVM symbols; `libLLVM.so.23.0git` exports 470 `llvm::cl::` ones), so a second comgr image aborts regardless of load order or who loads it.

Critically, **this PR does not fix `rocprofv3`**. Plain `rocprofv3 --kernel-trace --pmc SQ_WAVES` aborts identically against a divergent stack with no rocprof-compute involved, because its launcher appends the tool libraries to `LD_PRELOAD`. Anything we do here leaves that path broken.

A 1,475-line change in rocprofiler-compute is a hacky and fragile workaround for a defect a layer below us. It depends on guessing which `libamd_comgr` the workload will `dlopen` before the workload runs, via `LD_LIBRARY_PATH` scans, ELF `RPATH` parsing, site-packages globbing, and an import-probe subprocess. Guess wrong and it either does nothing or adds a third candidate image. It also spreads state through the profile path (`profiler_base.py`, both backends, `utils_profile.py`) for a case that should not exist.

The correct fix is in amd-comgr (https://github.com/ROCm/llvm-project/tree/amd-staging/amd/comgr): drop the static `cl::opt`, or skip registration when the name is already in `cl::getRegisteredOptions()`. That fixes rocprofv3, rocprof-compute, and every other comgr consumer at once.

vedithal-amd:
Product guidance instead of runtime inference.

For PyTorch profiling on ROCm, install ROCm and torch together from TheRock so the process has one ROCm stack:

```
pip install --index-url https://rocm.nightlies.amd.com/whl-multi-arch/ \
  "rocm[profiler,devel,libraries,device-gfx942]" "torch[device-gfx942]"
```

TheRock's torch depends on the rocm-sdk packages rather than bundling its own ROCm, so there is no second `libamd_comgr` to collide with. Install both in one command; installing torch afterwards can downgrade ROCm. Docs: https://github.com/ROCm/TheRock/blob/main/RELEASES.md

This is what our test image already uses (`docker/Dockerfile.therock.wheel.pytorch:24-29`), and it is the guidance we should give users instead of doing pre-flight inference of the workload's ROCm stack at profile time.

projects/rocprofiler-compute/src/utils/rocm_stack_check.py
```diff
@@ -0,0 +1,884 @@
...
+                    if _real_path(path) not in candidates.on_search_path
+                ),
+            )
+        )
+    return StackFindings(
+        profiler_libs.get(COMGR_LIB_NAME), tuple(conflicts), tuple(identical_copies)
+    )
+
+
+def _workload_candidates(
```
vedithal-amd, line 480: Predicting which copy the loader will pick before the workload runs is too fragile to build on. With a bash script or `mpirun` as `argv[0]` there is no interpreter and no readable RUNPATH, so this silently degrades to `LD_LIBRARY_PATH` and `LD_PRELOAD` only and misses the venv entirely.

projects/rocprofiler-compute/src/utils/rocm_stack_check.py
```diff
@@ -0,0 +1,884 @@
...
+    file or holds identical contents, belongs to the same stack. A copy named
+    in LD_PRELOAD is excluded, since the workload and the profiler both load
+    it.
+    """
+    return [
+        path
+        for path in _unique_paths(candidates)
+        if not _under_directory(path, rocm_root)
+        and not _same_library(path, profiler_lib)
+        and _real_path(path) not in preloaded
```
vedithal-amd, line 557: Excluding `LD_PRELOAD` copies assumes a matching soname merges the two. I tested it and it does not: a same-soname preloaded duplicate still produced the LLVM abort, and `check_single_rocm_stack` returned no conflicts for it while flagging the benign `_rocm_sdk_core` copies instead.

projects/rocprofiler-compute/src/utils/rocm_stack_check.py
```diff
@@ -0,0 +1,884 @@
...
+            list(findings.conflicts), ROCPROF_SIGNAL_HANDLERS_ENV in env
+        )
+    except Exception:
+        console_debug("stack", f"ROCm stack check failed:\n{traceback.format_exc()}")
+        return StackFindings()
+    console_warning("stack", message)
+    return findings
+
+
+def disable_rocprof_signal_handlers(
```
vedithal-amd, line 199: The hang is a rocprofiler-sdk defect: `rocprofv3_signal` swallows the workload's restore-to-default, so rocprofv3 re-enters its own handler indefinitely. Minimal reproducer attached, about twenty lines of C with no ROCm duplication involved. Please file this with the SDK team and drop the env var change from this PR.

<details>
<summary>Bug report for the rocprofiler-sdk team (click to expand)</summary>

# rocprofv3 never terminates when the workload installs its own SIGABRT handler

## Motivation

Found while profiling a PyTorch workload whose wheel supplies its own ROCm.
Two copies of `libamd_comgr` in one process make LLVM abort with
`Option 'spirv-expand-step' registered more than once`. CPython's SIGABRT
handler restores the default disposition and re-raises, and from there the run
never finishes.

The duplicate ROCm only supplies the abort. The reproducer below shows the
problem does not depend on it: any workload that installs a SIGABRT handler and
re-raises is enough. That covers most Python workloads, since CPython and
abseil both behave this way.

## Impact

The run gets stuck and never exits. It has to be killed by hand, and its
profiling output is lost. Any tool that wraps rocprofv3 and waits on the child
is stuck with it.

## Environment

| | |
|---|---|
| Host | `asrock-1w300-f2-1.mkm.dcgpu` |
| GPU | AMD Instinct MI350X (gfx950) |
| ROCm | nightly `10.0.0a20260730`, installed as wheels into `/rocm-venv` |
| Wheels | `rocm-sdk-core`, `rocm-sdk-devel`, `rocm-sdk-libraries`, `rocm-profiler`, all `10.0.0a20260730` |
| rocprofv3 | 1.3.5, git revision `82882e7dc2510e70ad8bb14962a9201b8b400716` |
| Python | 3.12.13 (`/rocm-venv`) |
| OS | Ubuntu 22.04.5 LTS, kernel 5.15.0-143-generic |
| Compiler | gcc 11.4.0 |

Reproduced with `rocprofv3` invoked directly, no wrapper tools involved.

## Technical details

1. The workload aborts. `rocprofv3_error_signal_handler` runs and finalizes.
2. It finds the workload's handler in `get_chained_signals()` and invokes it.
3. The workload's handler restores the default disposition and re-raises.
4. rocprofv3's handler is entered again, and the cycle repeats without end.

Step 3 is where the cycle closes. `signal` is interposed by `rocprofv3_signal`,
so the workload's `signal(SIGABRT, SIG_DFL)` never reaches libc. The interposer
records `SIG_DFL` as the new chained handler and reinstalls rocprofv3's own
handler in its place, so the workload's attempt to restore the default and die
is swallowed and the following `raise` re-enters rocprofv3.
`rocprofv3_sigaction` does the same for the `sigaction` form.

Setting `ROCPROF_SIGNAL_HANDLERS=0` avoids it, because that flag gates both
interposers, so `get_chained_signals()` stays empty and step 2 never happens.
It is a usable workaround, but it also turns chaining off for workloads that are
not affected, so it is not a fix.

## Reproducer

```c
/* abort_handler.c */
#include <signal.h>
#include <stdio.h>
#include <unistd.h>

static void on_abort(int sig, siginfo_t *info, void *ctx) {
    (void)info; (void)ctx;
    const char msg[] = "workload SIGABRT handler ran\n";
    write(STDERR_FILENO, msg, sizeof(msg) - 1);
    /* What CPython and abseil do: re-raise so the process dies. */
    signal(sig, SIG_DFL);
    raise(sig);
}

int main(void) {
    struct sigaction sa;
    sigemptyset(&sa.sa_mask);
    sa.sa_flags = SA_SIGINFO;
    sa.sa_sigaction = on_abort;
    sigaction(SIGABRT, &sa, NULL);
    fprintf(stderr, "workload: handler installed, aborting\n");
    fflush(stderr);
    abort();
    return 0;
}
```

```bash
gcc -o abort_handler abort_handler.c

# Expected behavior without the profiler: exits 134 immediately.
./abort_handler

# Never terminates. Kill it promptly.
rocprofv3 --kernel-trace -d ./out -- ./abort_handler

# Workaround: exits 134 immediately.
ROCPROF_SIGNAL_HANDLERS=0 rocprofv3 --kernel-trace -d ./out -- ./abort_handler
```

> Kill the second command as soon as you have confirmed it, rather than leaving
> it running.

### Observed output

The chained handler runs once, then the last line repeats without end.

```
[rocprofv3_error_signal_handler] rocprofv3 caught signal 6...
[rocprofv3_error_signal_handler] rocprofv3 will wait for 0 children to exit
[rocprofv3_error_signal_handler] rocprofv3 finalizing after signal 6...
[rocprofv3] output generation :: 0.000361 sec
[rocprofv3] tool finalization :: 0.000503 sec
[rocprofv3_error_signal_handler] rocprofv3 found chained signal handler for 6... executing chained sigaction (SIGINFO)
workload SIGABRT handler ran
[rocprofv3_error_signal_handler] rocprofv3 caught signal 6...
[rocprofv3_error_signal_handler] rocprofv3 caught signal 6...
[rocprofv3_error_signal_handler] rocprofv3 caught signal 6...
... repeats indefinitely
```

In the original PyTorch case the process parked in `futex_wait_queue_me`
instead of spinning, so the same cycle can present as either a livelock or a
deadlock depending on what the workload's handler does.

## Suggested fix

Two options, and they compose.

Add a re-entrancy guard to `rocprofv3_error_signal_handler`: mark the signal as
already being handled before invoking the chained handler, and on re-entry
restore the default disposition and re-raise rather than finalizing again.

Stop the interposers from swallowing a reset to default. When `rocprofv3_signal`
or `rocprofv3_sigaction` is called with `SIG_DFL` or `SIG_IGN`, pass it through
rather than reinstalling rocprofv3's handler. A workload asking for the default
disposition is usually on its way out.

Both keep chaining, which is the useful behavior for workloads that are not
affected.

</details>

projects/rocprofiler-compute/src/utils/utils_profile.py
```diff
@@ -270,6 +277,7 @@ def run_prof(
             stripped = line.strip()
             if stripped:
                 _classify_output_line(stripped)
+        explain_failed_run(output, stack_findings)
```
vedithal-amd, line 280: This is the piece worth keeping. Match the LLVM comgr signature on the captured output in the existing failure branch, emit a short message telling the user to install PyTorch and ROCm together, and link the docs. That covers the real case and removes the need for `rocm_stack_check.py`.

projects/rocprofiler-compute/docs/how-to/profile/mode.rst
```diff
@@ -396,6 +396,62 @@ Raw performance counter data produced by the underlying
    ``--retain-rocpd-output`` are deprecated and will be removed in a future release.
    ``.db`` files will be retained by default and the analyze step will read them directly.
 
+.. _rocm-stack-check:
```
vedithal-amd, line 399: Profile mode reference is the wrong home for this, and the detail level is too deep for end users. It is an installation concern, so it belongs in quickstart Prerequisites. Suggested replacement below, added as step 4 after the Python environment step in `docs/install/quickstart.rst:62`, with everything here removed.

```rst
.. _install-pytorch-with-rocm:

4. If you plan to profile a PyTorch or Triton workload, install PyTorch from
   the same source as ROCm. The `TheRock <https://github.com/ROCm/TheRock>`__
   nightly index publishes both, and its PyTorch packages pull in the matching
   ROCm for you. Replace ``gfx950`` with your GPU target:

   .. code-block:: shell-session

      pip install --index-url https://rocm.nightlies.amd.com/whl-multi-arch/ \
          "torch[device-gfx950]" "torchvision[device-gfx950]"

   See `TheRock releases <https://github.com/ROCm/TheRock/blob/main/RELEASES.md>`__
   for other targets and packages.

   Some PyTorch packages bundle their own copy of ROCm. When you profile one of
   those, your process ends up with two copies of ROCm, which conflict and make
   the run fail or stop responding. Installing both from the same source keeps a
   single ROCm in the process.
```

Two cross-references to add while you are here, rather than repeating the guidance:

- `docs/how-to/profile/mode.rst:959` (Torch trace, Requirements), extend the first bullet: ``Valid PyTorch installation in the profiling environment, installed alongside ROCm (see :ref:`install-pytorch-with-rocm`).``
- `docs/how-to/profile/mode.rst:1191` (Triton trace, Requirements), the same cross-reference for Triton, since this section already defers to Torch trace for everything else.

vedithal-amd (changes requested):
The narrowed scope is the right direction, but the pre-run stack introspection should come out. It reports zero conflicts for the environment that actually aborts, and it cannot see a venv at all when `argv[0]` is a bash or `mpirun` wrapper. Detection belongs after the crash, matching the profiler's own output.

The hang is a rocprofiler-sdk defect rather than something we should work around. I wrote up a minimal reproducer and attached it below; please file it with the SDK team and drop the `ROCPROF_SIGNAL_HANDLERS` change. The glog regex fix is a real bug worth keeping, and worth its own line in the PR description.

PR #8758, opened 2026-07-17
test(rccl): added AsyncSignal test

alexander-sannikov:
Adds HostApiTest.PutSignalEnqueueIsAsync, an MPI regression test asserting that ncclPutSignal host-side enqueue stays asynchronous (returns in microseconds) while the target stream is kept busy by a spin kernel, rather than blocking for the full stream duration. It guards against the implicit CPU-GPU synchronization regression on the putSignal/CE path caused by staging the signal value through a pageable host memcpy. Fixes AICOMRCCL-1517

## Motivation

Fixes AICOMRCCL-1517

## Issue Tracking

AICOMRCCL-1517

## Test Plan

Validated manually.
Test enabled in CI.

## Test Result
Manual test: PASSED

## Submission Checklist

### DIFF
--- a/projects/rccl/test/HostApiMPITests.cpp
+++ b/projects/rccl/test/HostApiMPITests.cpp
@@ -63,10 +63,24 @@
 
 #include <cstdint>
 #include <cstdlib>
+#include <chrono>
 
 using namespace MPITestConstants;
 using namespace RCCLTestGuards;
 using namespace RCCLHostApiHelpers;
+
+// ---------------------------------------------------------------------------
+// busy wait kernel
+// ---------------------------------------------------------------------------
+__global__ void hostApiSpinKernel(unsigned long long cycles, volatile int* sink)
+{
+    unsigned long long start = clock64();
+    int                acc   = 0;
+    while(static_cast<unsigned long long>(clock64() - start) < cycles)
+        acc++;
+    if(sink)
+        *sink = acc;
+}
 
 namespace RcclUnitTesting
 {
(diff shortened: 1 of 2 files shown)


alexander-sannikov:
%RCCL_TEST_CONFIG
mi300x_mellanox_ib.json
SUITE=host_api_tests_default

projects/rccl/test/HostApiMPITests.cpp
```diff
@@ -63,11 +63,25 @@
...
 
+// ---------------------------------------------------------------------------
+// busy wait kernel
+// ---------------------------------------------------------------------------
+__global__ void hostApiSpinKernel(unsigned long long cycles, volatile int* sink)
+{
+    unsigned long long start = clock64();
+    int                acc   = 0;
+    while(static_cast<unsigned long long>(clock64() - start) < cycles)
+        acc++;
```
amd-arozanov, line 80: minor: acc overflows a signed int many times over kSpinMs, which is UB. Make it unsigned int acc to be safe?

projects/rccl/test/HostApiMPITests.cpp
```diff
@@ -250,6 +279,99 @@ TEST_F(HostApiTest, SinglePutRank0ToRank1)
...
+                       spinCycles, static_cast<volatile int*>(sink));
+
+    auto t0 = std::chrono::steady_clock::now();
+    ncclResult_t putRes = ncclSuccess;
+    for(int i = 0; i < kAsyncNumPuts && putRes == ncclSuccess; ++i)
+    {
+        size_t off = static_cast<size_t>(i) * kAsyncPutElems;
+        putRes = ncclPutSignal(
+            static_cast<uint8_t*>(winBuf) + off, kAsyncPutElems, ncclUint8,
+            peer, win, off, kSigIdx, kCtx, kFlags, comm, stream);
```
amd-arozanov, line 356: Source (winBuf + off) and the peer's PUT destination (peerWinOffset = off) overlap at the same offset, so this is a data race in a real scenario. Fine here since it is a timing test and data is never verified, but please add a comment saying that so nobody adds a VerifyBuf later, or split source/recv regions to fix the race.

projects/rccl/test/HostApiMPITests.cpp
alexander-sannikov: We don't care about value here. it is only needed to prevent compiler optimization.

amd-arozanov (approved):

VadimKutovoi (approved):

alexander-sannikov:
%RCCL_TEST_CONFIG
mi300x_mellanox_ib.json
SUITE=host_api_tests_default

PR #8767, opened 2026-07-17
refactor(rocprofiler-systems): Refactoring AI-NIC support detection

dgaliffiAMD:
## Motivation

  In TheRock builds, `ROCPROFSYS_BUILD_AINIC` was always `OFF`, silently disabling AI NIC profiling support even when `ROCPROFSYS_USE_AINIC=ON`. The root cause was in `cmake/Packages.cmake`: the AMD SMI version detection constructed a hardcoded path
  `${ROCM_PATH}/include/amd_smi/amdsmi.h`. TheRock explicitly unsets `ROCM_PATH` for all subproject builds, so the header was never found, version detection returned empty strings, and the `>= 26.3` check failed.

  Beyond fixing detection, this branch replaces the original binary-introspection verification (an `nm` symbol probe) with a robust, runtime, hardware-independent check, so it also works when tests run on a different machine than the one that built the binaries.

## Issue Tracking

- JIRA ID: ROCM-28005

## Technical Details

  **CMake detection (`cmake/Packages.cmake`)**
  - Read the AMD SMI version directly from the `amd_smi_VERSION` variable reported by config-mode `find_package(amd_smi)`, instead of locating and regex-parsing `amdsmi.h`. No dependency on `ROCM_PATH` or any environment variable, and no header-path/regex fragility.
  - When the package reports no version, fall back to `0.0` (AI NIC unsupported) with a visible status message rather than failing silently.
  - Gate `ROCPROFSYS_BUILD_AINIC` on `amd_smi_VERSION VERSION_GREATER_EQUAL 26.3` together with `ROCPROFSYS_USE_AINIC`.

  **Expose compiled-in support as a runtime signal**
  - Promote `ROCPROFSYS_BUILD_AINIC=1` to the shared `rocprofiler-systems-compile-definitions` INTERFACE target so it propagates to core/avail (not just the pmc/amd_smi backends); drop the now-redundant per-target define on the amd_smi backend.
  - Register the `ROCPROFSYS_USE_AINIC` and `ROCPROFSYS_SAMPLING_AINICS` settings only when `ROCPROFSYS_BUILD_AINIC=ON`. Their presence in `rocprof-sys-avail --settings` is therefore a direct, hardware-independent indicator of compile-time support.
  - Add `get_ainic_supported()` and macro-guard `get_sampling_ainics()` for the disabled case (`source/lib/core/config.{hpp,cpp}`).

  **Tests (`tests/pytest/test_ainic.py`)**
  - Merge the compile-time check and the RDMA performance-track check into one `TestAINIC` class; delete `test_ainic_compiled.py`, `test_ainic_perf.py`, and the standalone `check_ainic_support.sh`.
  - The build-artifact check now asserts `rocprof-sys-avail --settings` lists the AI NIC settings (no `nm`, no amd_smi symbol names).
  - Split the `ainic` marker in `conftest.py` into `ainic` (informational label) and `ainic_required` (hardware gate). CTest names: `ainic-settings-present` and `ainic-performance-tracks`.

## Test Plan
  - [x] `ctest -V -R ainic-settings-present` — passes on an AI NIC-enabled build; skips (not fails) on builds against AMD SMI < 26.3 via `amdsmi_min_version("26.3")`.
  - [x] `ctest -V -R ainic-performance-tracks` — skips without AI NIC hardware; excluded from CI via `-LE "network|gpu"` (carries the `network` label).
  - [x] AI NIC-off build (`-DROCPROFSYS_USE_AINIC=OFF`): `rocprof-sys-avail --settings` lists no AI NIC settings; `config.cpp` compiled without the define.
  - [x] TheRock configure log shows `AMD SMI version detected: <major>.<minor>` and `AINIC support enabled (AMD SMI >= 26.3)`; `CMakeCache.txt` has `ROCPROFSYS_BUILD_AINIC:INTERNAL=ON`.

## Test Result

  Verified on a build with AMD SMI 26.5:
  - `ainic-settings-present` — PASSED; `avail --settings` lists `ROCPROFSYS_USE_AINIC` and `ROCPROFSYS_SAMPLING_AINICS`.
  - Throwaway AINIC-off build — `avail --settings` lists no AI NIC settings, confirming the negative case.
  - CMake `amd_smi_VERSION` detection verified working on a TheRock build.

## Known follow-ups (not in this branch)
  
  - `--ai-nics` CLI flag in `argparse.cpp` is still registered unconditionally (surface-sweep finding) — should be gated on `ROCPROFSYS_BUILD_AINIC`.
  - Optional: rename the `network` label to `network-enabled`/`network-required` to parallel the `ainic`/`ainic_required` split.


projects/rocprofiler-systems/tests/pytest/test_ainic_compiled.py
```diff
@@ -2,84 +2,56 @@
 # SPDX-License-Identifier: MIT
 
 """
-Tests that AI NIC support was compiled into librocprof-sys.so.
+Tests that AI NIC support was compiled into the rocprofiler-systems binaries.
```
kcossett-amd: At this point, it sounds more like this set of tests should live in `test_binaries.py`

projects/rocprofiler-systems/tests/pytest/test_ainic_compiled.py
```diff
@@ -2,84 +2,56 @@
...
 from rocprofsys import RocprofsysConfig
 
-# AI-NIC support is only compiled in when AMD SMI >= 26.3. Skip this build-artifact
-# check on builds against older AMD SMI, where AI-NIC is legitimately disabled.
+# AI NIC support is only compiled in when AMD SMI >= 26.3. Skip this build-artifact
+# check on builds against older AMD SMI, where AI NIC is legitimately disabled.
 pytestmark = [
     pytest.mark.rocprof_binary,
     pytest.mark.amdsmi_min_version("26.3"),
 ]
```
kcossett-amd:
```suggestion
pytestmark = [
pytest.mark.ainic_compiled,
pytest.mark.amdsmi_min_version("26.3"),
]
```
Then register it in `conftest.py`. I'd like to keep `rocprof_binary` as the marker that groups the `test_binaries.py` module.

That is if you don't move all these tests to `test_binaries.py`.

projects/rocprofiler-systems/tests/pytest/test_ainic_compiled.py
```diff
@@ -2,84 +2,56 @@
...
-        proves that the NIC collector code was compiled in.
+        These settings are only registered when the binaries are compiled with
+        ROCPROFSYS_BUILD_AINIC=ON, so their presence proves AI NIC support was
+        compiled in. No NIC hardware is required.
         """
         result = subprocess.run(
-            ["nm", "-D", str(librocprof_sys)],
+            [str(rocprof_config.rocprofsys_avail), "--settings"],
             capture_output=True,
             text=True,
```
kcossett-amd: Add a timeout to the subprocess.

projects/rocprofiler-systems/docs/how-to/nic-profiling.rst
```diff
@@ -183,66 +183,34 @@ Verifying AI NIC compile-time support
...
-detection may fail silently and produce a binary without AI NIC support, even
-though AMD SMI is present.
+later and ``ROCPROFSYS_USE_AINIC=ON`` (the default).
 
-To verify whether a given binary has AI NIC support compiled in, inspect the
-dynamic symbol table of ``librocprof-sys.so``. The AMD SMI NIC API symbols
-(such as ``amdsmi_get_nic_rdma_port_statistics``) appear as undefined
-references only when the NIC collector was compiled in.
+The AI NIC settings (such as ``ROCPROFSYS_USE_AINIC``) are only registered when
+the binaries are compiled with ``ROCPROFSYS_BUILD_AINIC=ON``. Their presence in
```
kcossett-amd:
```suggestion
the ``rocprof-sys`` binaries are compiled with ``ROCPROFSYS_BUILD_AINIC=ON``. Their presence in
```

projects/rocprofiler-systems/tests/pytest/test_ainic_compiled.py
dgaliffiAMD: Yeah ... I was torn. I agree with you, but it also makes sense to be grouped with the `ainic` tests. 
I think that I will go with your original suggestion and merge the two "test_ainic" files.

ajanicijamd (approved):

projects/rocprofiler-systems/docs/how-to/nic-profiling.rst
```diff
@@ -183,66 +183,34 @@ Verifying AI NIC compile-time support
...
+command produces no output, the binaries were built without AI NIC support.
 
 **Automated pytest check**
 
 The pytest test suite includes a test that performs the same check:
 
 .. code-block:: shell
 
-   ctest -V -R ainiccompiledin-ainic-dynamic-symbols-present
+   ctest -V -R ainic-settings-present
```
kcossett-amd: Do we need the `-V` here? 
You mention below what to do if the test fails below.

kcossett-amd (approved):
Thanks

projects/rocprofiler-systems/docs/how-to/nic-profiling.rst
dgaliffiAMD: Fair. Honestly, I am not sure the user documentation needs to reference specifics of the automated tests at all. Depending on your method of installation, they may not even be present on your system.

kcossett-amd: Yeah, that and the fact that the suite requires python 3.8, which some may not have.

PR #8769, opened 2026-07-17
[rocjitsu] Report unsupported AQL packet errors

kuhar:
## Motivation

Reject unsupported standard and AMD vendor AQL packets without consuming them, fault the queue through ROCr-compatible error signaling, and carry queue-error writes across daemon RPC. Add local, daemon, legacy-mask, and metadata-marshalling coverage.

## Technical Details

Rocjitsu incorrectly consumed unsupported AQL packets as successful no-ops. In particular, an unknown AMD vendor packet format advanced the queue read index and decremented its completion signal without reporting an error.

Match physical GPU behavior by leaving invalid packets unconsumed and reporting `HSA_STATUS_ERROR_INVALID_PACKET_FORMAT` through the queue error callback

## Issue Tracking

Issue: https://gist.github.com/benvanik/f54edc521e112e5b3c58ced494e410a1

## Test Plan

Unit tests.

## Test Result

OK

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/remote_driver.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/remote_driver.cpp
@@ -7,7 +7,15 @@
 #include "rocjitsu/kmd/linux/remote_driver.h"
 #include "rocjitsu/kmd/linux/kfd_ioctl_utils.h"
 #include "rocjitsu/kmd/linux/rpc.h"
+#include "util/log.h"
 
+#include "rocjitsu/base/rj_compiler.h"
+RJ_DIAGNOSTIC_PUSH
+RJ_DIAGNOSTIC_IGNORE_PEDANTIC
+#include "hsa/amd_hsa_queue.h"
+RJ_DIAGNOSTIC_POP
+
+#include <atomic>
 #include <cassert>
 #include <cerrno>
 #include <chrono>
@@ -35,6 +43,7 @@ namespace {
 
 constexpr bool has_embedded_pointers(unsigned long request) {
   switch (canonical_ioctl_request(request)) {
+  case AMDKFD_IOC_CREATE_QUEUE:
   case AMDKFD_IOC_WAIT_EVENTS:
   case AMDKFD_IOC_MAP_MEMORY_TO_GPU:
   case AMDKFD_IOC_UNMAP_MEMORY_FROM_GPU:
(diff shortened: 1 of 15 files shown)


atgutier (commented):
Agent comments:

 ---
  MAJOR

  1. remote_driver.cpp:653-654 — appended signal-write bytes can corrupt a DBG_TRAP ENABLE
  runtime-info buffer
  Signal writes are appended to every ioctl response and drained per-process, so they piggyback on
  whichever ioctl the client issues next — which can be DBG_TRAP ENABLE. The legacy embedded-pointer
  output-copy sizes its source region from the whole response payload, not the echoed inline region:
  if (has_embedded_pointers(request) && resp->payload_bytes > arg_size) {
    size_t extra = resp->payload_bytes - arg_size;   // = inline_reserved + signal_bytes
    ...
    case KFD_IOC_DBG_TRAP_ENABLE:
      copy_len = std::min(saved_dbg_rinfo_size, std::min(dbg->enable.rinfo_size, extra));
  If the client passed rinfo_size > sizeof(kfd_runtime_info) and any signal write is pending, copy_len
   extends into the appended signal bytes → they get memcpy'd into the debugger's rinfo buffer. Silent
   corruption (no crash — it's the client's own buffer). The APERTURES and GET_DEVICE_SNAPSHOT cases
  happen to be safe because their min bound equals the inline region exactly.
  Fix: derive extra from the request-reserved size, not the response:
  if (has_embedded_pointers(request) && wire_args_size > arg_size) {
    size_t extra = wire_args_size - arg_size;   // excludes appended signal writes
  This fixes ENABLE and makes the whole extra-copy path uniformly immune to the appended region —
  worth a comment pointing back to the rpc.h:89-94 invariant, since every future variable-length
  output path must clamp to the request-reserved inline size.

  ---
  MINOR

  2. simulated_kfd.cpp:1453 — reinterpret_cast read of RpcQueueErrorState is strict-aliasing/lifetime
  UB
  The daemon reads the inlined wire struct via reinterpret_cast<const
  RpcQueueErrorState*>(args->ctx_save_restore_address) and dereferences members, but no such object
  was constructed in that raw payload buffer. Alignment is fine in practice, but it's inconsistent
  with the deliberate std::memcpy discipline the client side uses (remote_driver.cpp:635). Fix:
  RpcQueueErrorState state{}; std::memcpy(&state, reinterpret_cast<const void*>(...), sizeof(state));.

  3. command_processor.cpp:499-509 — non-atomic fallback store drops the release ordering it promises
  signal_kfd_event's fast path uses std::atomic_ref::store(release), but the memory_->write64 fallback
   (misaligned/page-crossing/unmapped-in-view VA) is a plain non-atomic write. Latent because
  amd_signal_t is 64-byte aligned so value (+8) and mailbox (+16) never misalign or page-cross, and
  the daemon path returns early before reaching here. Fix: document that the fallback is only reached
  for daemon/cross-process VAs where ordering is re-established by the RPC drain + signal_interrupt;
  optionally assert((va & 7) == 0) on the local path.

 4. Callback lifetime — remote_queue_error_cb_ captures this (SimulatedKfd); ~SimulatedKfd doesn't
  stop the CP thread or clear the callback. Same pattern the pre-existing interrupt_cb_ already relies
   on, so no new bug class, but it adds a second callback depending on "engine tears down CP threads
  before the driver." The engine's thread-stop ordering lives in simdojo (out of PR scope) and
  couldn't be confirmed. Verify that ordering is ironclad; if not, clear the callbacks / stop the CP
  at the top of ~SimulatedKfd.

  5. command_processor.h:70-75 — six new HwQueue error fields lack @brief, unlike surrounding members.
   error_reason_va carries a non-obvious semantic (it's the value address = amd_signal_t base + 8, not
   the signal base). Fix: one-line /// on at least error_reason_va and faulted.

  6. report_queue_error/signal_kfd_event depend on an undocumented hw_queue_mutex_ precondition.
  queue.faulted is read/written with no internal lock; safe only because both fetch_from_queue callers
   hold hw_queue_mutex_ (recursive). Fix: add /// @pre hw_queue_mutex_ is held. to both declarations
  (command_processor.h:186-190).

  7. rpc.h:84 — enum RpcQueueErrorFlags lacks the @brief that adjacent new wire declarations carry.

  ---
  NIT

  - command_processor.cpp:519 — if event_id == 0 after the fallback read, the payload is written but
  no wake fires and nothing is logged; a lost wakeup would be invisible. Consider a Logger::warn.
  - command_processor.cpp:47-50 — tie the 20/23 exception codes and 32/256 legacy masks to "ROCr
  AqlQueue queue-error codes" in a comment; the coupling to an external ABI is non-obvious.
  - simulated_kfd.h:~113 — take_signal_writes isn't [[nodiscard]] while sibling signal_write_count is;
   discarding it silently loses queued writes.
  - remote_driver.cpp:62 — read_client_value<T> dereferences a raw client VA with no check; acceptable
   per design (client's own address space, can't reach daemon memory) but worth a one-line
  precondition comment.

benvanik:
FYI, I built the current PR head against ROCm SDK nightly `7.15.0a20260718` and reran the standalone reducer in both local and daemon modes. Both pass with the physical contract: `completion=1, read_index=0, queue_error=0x1009` (`HSA_STATUS_ERROR_INVALID_PACKET_FORMAT`).

The shipped July 18 nightly still consumes/completes the packet with no queue error. Updated reducer and result: https://gist.github.com/benvanik/f54edc521e112e5b3c58ced494e410a1

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/remote_driver.cpp
```diff
@@ -558,6 +628,28 @@ int RemoteDriver::send_ioctl(unsigned long request, void *arg) {
...
+      return -EPROTO;
+    const size_t writes_bytes = payload.size() - wire_args_size;
+    if (writes_bytes % sizeof(RpcSignalWrite) != 0)
+      return -EPROTO;
+    for (size_t offset = 0; offset < writes_bytes; offset += sizeof(RpcSignalWrite)) {
+      RpcSignalWrite write{};
+      std::memcpy(&write, payload.data() + wire_args_size + offset, sizeof(write));
+      if (write.address == 0)
+        continue;
+      if ((write.address & (alignof(uint64_t) - 1)) != 0) {
```
newling, line 644: Please review this alignment constraint, it might need relaxing. Entry in agent's review: https://github.com/newling/agent-public-reviews/blob/main/rocm-systems/review_pr8769.md#1-do-not-drop-valid-four-byte-aligned-queue-error-signals (and feel free to check if there's anything else of use in that review)

PR #8770, opened 2026-07-17
feat(rocr): Skip gfx125x entry trampoline when unclaused-VMEM prologue already present

dayatsin-amd:
## Summary

Detect the GFX1250 unclaused-VMEM workaround prologue at kernel entry and skip the entry trampoline when it is already present, plus a minimal (no-workaround) trampoline mode for testing.

### 1. Skip the trampoline when the compiler workaround is already present
Detect the GFX1250 unclaused-VMEM workaround prologue (`global_wb <scope:SCOPE_CU>` + `v_nop`, from llvm PR #208467) at each kernel entry in `ExecutableImpl::InstallTrampolinesGfx125x`. When a kernel's entry already begins with that 16-byte prologue, the trampoline would only duplicate the workaround, so the kernel is filtered out and dispatch goes straight to the real entry. Kernels lacking the prologue still get a trampoline.

- Detection reads the post-relocation host shadow of the code segment (valid pre-Freeze) and compares the first 4 dwords at the kernel entry.
- The prologue is captured in a shared `kGfx1250UnclausedVmemPrologue` constant, reused by `BuildTrampolineGfx1250`.
- Each trampoline injection is logged with the kernel symbol name (gated by `LOADER_ENABLE_LOGGING=1`); `KdFixup` carries the name for this.

### 2. Minimal (no-workaround) trampoline mode for testing
`LOADER_ENABLE_TRAMPOLINE_NO_WA=1` makes the loader emit a minimal stub that loads the real kernel entry address into `s[100:101]` and jumps via `s_set_pc_i64`, omitting the `global_wb` / `v_nop` cache-writeback workaround used by the full trampoline path (`LOADER_ENABLE_TRAMPOLINE=1`).

- `LOADER_ENABLE_TRAMPOLINE=1` — full stub with `global_wb` + `v_nop` + `s_mov` + `s_set_pc`
- `LOADER_ENABLE_TRAMPOLINE_NO_WA=1` — minimal stub with `s_mov` + `s_set_pc` only
- If both are set, the NO_WA variant takes precedence

## JIRA ID

JIRA ID : LCOMPILER-2448

## Test plan

- [x] Build rocr-runtime with the change (gfx1250 TheRock toolchain)
- [x] Kernel compiled WITH the compiler workaround (patched clang): entry has `global_wb`/`v_nop`, and `InstallTrampolinesGfx125x` skips it (no injection log for that kernel)
- [x] Kernel compiled WITHOUT the workaround (original compiler from the 7.14 gfx1250 tarball): trampoline is injected as before, and the run succeeds
- [x] Verified on a gfx1250 system: both binaries run correctly; injection log lists only the kernels that actually receive a trampoline
- [x] `LOADER_ENABLE_TRAMPOLINE_NO_WA=1` emits the minimal stub; `LOADER_ENABLE_TRAMPOLINE=1` emits the full stub; disabled when neither is set

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
@@ -232,13 +232,25 @@ static constexpr size_t kTrampolineStubStride =
 // and readable, which the allocation's zero-fill already guarantees.
 static constexpr size_t kInstPrefUnitBytes = 128;  // GFX11+ CP I$ prefetch line size
 
+// The GFX1250 unclaused-VMEM workaround prologue (llvm PR #208467). The compiler
+// emits these 4 dwords -- global_wb <scope:SCOPE_CU> followed by v_nop -- at every
+// hardware kernel entry so the first VMEM instruction is unclaused. It is exactly
+// the sequence the entry trampoline itself prepends, so when a kernel's entry
+// already begins with it the trampoline would only duplicate the workaround.
+static constexpr uint32_t kGfx1250UnclausedVmemPrologue[4] = {
+    0xEE0B007C,  // global_wb <scope:SCOPE_CU>
+    0x00000000,  // :
+    0x00000000,  // :
+    0x7E000000,  // v_nop
+};
+
 static void BuildTrampolineGfx1250(uint8_t* buf, uint64_t target) {
   auto* w = reinterpret_cast<uint32_t*>(buf);
 
-  w[0] = 0xEE0B007C;  // global_wb <scope:SCOPE_CU>
-  w[1] = 0x00000000;  // :
-  w[2] = 0x00000000;  // :
-  w[3] = 0x7E000000;  // v_nop (padding)
+  w[0] = kGfx1250UnclausedVmemPrologue[0];  // global_wb <scope:SCOPE_CU>
+  w[1] = kGfx1250UnclausedVmemPrologue[1];  // :
+  w[2] = kGfx1250UnclausedVmemPrologue[2];  // :
+  w[3] = kGfx1250UnclausedVmemPrologue[3];  // v_nop (padding)
   w[4] = 0xBEE400FF;  // s_mov_b32 s100, target_lo
   w[5] = static_cast<uint32_t>(target);
   w[6] = 0xBEE500FF;  // s_mov_b32 s101, target_hi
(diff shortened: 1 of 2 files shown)


dayatsin-amd:
My plan is to remove support for  LOADER_ENABLE_TRAMPOLINE_NO_WA later. But I want to keep it for now to avoid having to provide separate builds for testing.

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
```diff
@@ -1557,8 +1594,42 @@ hsa_status_t ExecutableImpl::LoadSegmentV2(const code::Segment *data_segment,
...
 hsa_status_t ExecutableImpl::InstallTrampolinesGfx125x(hsa_agent_t agent) {
-  const size_t n = kd_fixups_.size();
+  // Skip kernels whose entry already carries the compiler-inserted unclaused-VMEM
+  // workaround prologue (llvm PR #208467): a trampoline would only duplicate the
+  // global_wb/v_nop that is already there, so dispatch can go straight to the real
+  // entry. Kernels still needing the workaround keep their trampoline.
+  std::vector<KdFixup> fixups;
+  fixups.reserve(kd_fixups_.size());
+  for (const auto& f : kd_fixups_) {
+    if (KernelEntryHasUnclausedVmemPrologue(context_, f.code_seg, f.kd_vaddr + f.entry_off)) {
```
cfreeamd, line 1625: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** This prologue-skip filter runs in both normal and NO_WA modes. If the current gfx1250 compiler always emits the unclaused-VMEM prologue at kernel entry (llvm PR #208467), then in `LOADER_ENABLE_TRAMPOLINE_NO_WA=1` mode every kernel is filtered out here and `BuildTrampolineGfx1250NoWa` never actually runs. Is NO_WA intended only for code objects that *lack* the compiler prologue (older CO output), i.e. is skipping prologue-bearing kernels in NO_WA mode the deliberate behavior? Just confirming the interaction is intended, not a bug.

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
```diff
@@ -1557,8 +1594,42 @@ hsa_status_t ExecutableImpl::LoadSegmentV2(const code::Segment *data_segment,
...
+      continue;
```
cfreeamd, line 1626: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** For a testing/debugging-only feature, the injection log below (`Loader: injecting gfx125x entry trampoline for kernel ...`) has no counterpart for the kernels skipped here — which is exactly what this PR adds. Consider logging skipped kernels in this filter loop (symmetric with the injection log) so a tester can confirm which kernels took the fast path vs. still needed a trampoline.

cfreeamd (commented):

Reviewed `loader/executable.cpp` and `loader/executable.hpp` (rocr-runtime). The NO_WA stub encodings (`s_mov` s100/s101 + `s_set_pc_i64` `0xBE804864`) match the tested `BuildTrampolineGfx1250` path and the `llvm-mc` verification block; the prologue-detection bounds check and null-check look correct; the new `KdFixup::name` push is guarded by `trampoline_enabled_gfx125x_` so there's no non-test overhead; and the filtered `fixups` set is used consistently for pool sizing, prefetch-guard sizing, and the build loop. Left two non-blocking notes inline (one question, one optional).

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
dayatsin-amd: Skipping the kernels because the fix is already inserted by compiler is the expected code path so my reasoning is to only print when the fix is not inserted by compiler as this would help the user figure out which shader they need to re-compile.

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
dayatsin-amd: The NO_WA is only temporary and for evaluating performance in dedicated runs. It will be removed later

cfreeamd (approved):

PR #8771, opened 2026-07-17
feat(rocprofiler-sdk): Add ppc64le support for /proc/cpuinfo parsing

benrichard-amd:
## Motivation

Part of https://github.com/ROCm/TheRock/issues/5518 for ROCm code portability.

Enable rocprofiler-sdk to run on ppc64le (POWER) platforms. The existing
`/proc/cpuinfo` parser only handles x86 fields, causing all CPU agent
entries to fail `is_valid()` on POWER, which means no CPU agents are
discovered by the profiler.

## Technical Details

- Move `processor` field parsing outside the x86 `#if` block since it
  exists on all architectures.
- Add ppc64le-specific parsing for the `cpu` field, which holds the model
  name on POWER (e.g. "POWER10 (architected), altivec supported").
- Default x86-only fields (`family`, `model`, `physical_id`, `core_id`,
  `apicid`, `vendor_id`) on ppc64le so `is_valid()` passes and CPU agents
  are properly registered. Use `processor` index as a proxy for `apicid`
  to allow CPU agent matching.
- Add `#error` for unsupported architectures to catch missing
  `/proc/cpuinfo` parsing at compile time.

## Test Plan

- Built and ran rocprofiler-sdk unit tests on x86 (gfx90a) to confirm no
  regressions -- same test results with and without changes.
- Built and verified on ppc64le POWER10 system.

## Test Result

All pre-existing test passes/failures are identical with and without this
change. No regressions introduced.

## Submission Checklist


benrichard-amd:
*ping* 

Can someone take a look? Does not change any existing functionality.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/platform/gnulinux/agent.cpp
```diff
@@ -199,6 +199,17 @@ parse_cpu_info()
...
+#elif defined(__powerpc64__) || defined(__PPC64__)
+                // On POWER, "cpu" holds the model name (e.g. "POWER10, altivec supported")
+                else if(itr.find("cpu") == 0)
+                {
+                    info_v.model_name = value;
+                    info_v.model_name =
+                        sdk::parse::strip(std::string{info_v.model_name}, " \t\n\v\f\r");
+                }
+#else
+#    error "Unsupported architecture: /proc/cpuinfo parsing not implemented"
```
jrmadsen: I'm not sure we want to fail to compile here. I'd rather emit a warning (use `ROCP_CI_LOG(WARNING)` for the message so it is a warning in production and a failure in CI)

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/platform/gnulinux/agent.cpp
```diff
@@ -213,6 +224,17 @@ parse_cpu_info()
...
+#if defined(__powerpc64__) || defined(__PPC64__)
+        // Default x86-specific fields that have no equivalent in POWER /proc/cpuinfo.
+        // Use processor index as a proxy for apicid to allow CPU agent matching.
+        if(info_v.family < 0) info_v.family = 0;
+        if(info_v.model < 0) info_v.model = 0;
+        if(info_v.physical_id < 0) info_v.physical_id = 0;
+        if(info_v.core_id < 0) info_v.core_id = 0;
+        if(info_v.apicid < 0) info_v.apicid = info_v.processor;
+        if(info_v.vendor_id.empty()) info_v.vendor_id = "Unknown";
+#endif
```
jrmadsen, line 241: Add an else clause here that allows the CPU to be treated as valid (but emit a warning

jrmadsen (changes requested):

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/platform/gnulinux/agent.cpp
benrichard-amd: Removed.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/platform/gnulinux/agent.cpp
benrichard-amd: Fill in default values and emit warning on unknown CPU.

jrmadsen (approved):

PR #8777, opened 2026-07-17
[Do not review] Experiment with thread-limit

kcossett-amd:
## Motivation

Checking the state of `thread-limit` tests with `TheRock CI`.

## Submission Checklist


kcossett-amd:
Investigation concluded. As a workaround, I'll store `ROCPROFSYS_MAX_THREADS` in `rocprof-sys-avail` and query it at runtime. We'll use that value to run the test.

PR #8784, opened 2026-07-17
[rocjitsu] Fix zero-record gfx1250 buffer bounds

### LINKED ISSUE #8751: rocjitsu (gfx1250): SIGSEGV in L2Cache::send_backing on a GEMM/matmul with large output dim (vector-load path)
## Summary

On an emulated **gfx1250 (MI450X)** device under **rocjitsu** (functional mode, via mirage), a single **matmul / GEMM with a large output dimension** crashes with a native **SIGSEGV in the emulated vector-load cache path**. A plain `torch.matmul` reproduces it — no model, optimizer, or attention involved — so torch training/inference on gfx1250 is currently blocked (embedding/vocab projections and typical hidden/FFN GEMMs cross the threshold).

This is a **different code path** from the recently fixed scalar-memory issue (#8539 / #8652, `s_load` ioffset scaling): that fix is present and works; this crash is on the **vector global-load** path.

## Environment

- rocjitsu functional emulator launched via `mirage run`, built from the current `emulation/mirage` (includes #8652).
- Emulated agent: MI450X (gfx1250); pure software emulation, **no physical GPU**.
- Container torch built for gfx1250 (`torch.cuda.get_arch_list() == ['gfx1250']`).

## Steps to reproduce

1. Build mirage and create a gfx1250 profile:

```bash
# from ROCm/rocm-systems
cd emulation/mirage
./scripts/mirage-docker-build.sh
export MIRAGE_BIN="$PWD/build/manylinux/bin/mirage"

"$MIRAGE_BIN" profile create mi450x \
  --emulator rocjitsu --agent MI450X \
  --num-nodes 1 --gpus-per-node 1 --no-input
```

2. Save `repro.py`:

```python
import torch

d = torch.device("cuda:0")
a = torch.randn(64, 128, device=d)
b = torch.randn(128, 1024, device=d)   # output dim 1024 -> crash; <= ~912 is fine
c = a @ b
torch.cuda.synchronize()
print("OK", tuple(c.shape))
```

3. Run under mirage, using a container image whose torch is built for gfx1250:

```bash
"$MIRAGE_BIN" run --in-process --profile mi450x \
  --image <ROCm image with a gfx1250 torch> \
  --mount "$PWD/repro.py:/repro.py:ro" \
  -- python3 /repro.py
```

**Observed:** `Segmentation fault (core dumped)` (exit 139).
**Expected:** prints `OK (64, 1024)`.

## Output-dimension threshold

Fixing `M=64, K=128` and sweeping the output dim `N` (same result for `torch.matmul`, `F.linear`, and `nn.Linear`):

| N | ≤ 912 | 960 | 1024 | 2048 |
|---|-------|-----|------|------|
| result | OK | crash | crash | crash |

The crash appears once `N` crosses ~912–960, i.e. when the GEMM library selects a different kernel variant. It is deterministic (fails every run in a fresh process).

## Native backtrace

`librocjitsu.so` (not stripped). The fault is on a simulation worker thread servicing a global **vector** load:

```
Thread received signal SIGSEGV, Segmentation fault.
=> mov (%rsi),%rax        # rsi = 0x1000, si_addr = 0x1000  (near-null deref)
#0  rocjitsu::amdgpu::L2Cache::send_backing(...)
#1  rocjitsu::amdgpu::L2Cache::ensure_line(...)
#2  rocjitsu::amdgpu::L2Cache::fetch_line(...)
#3  rocjitsu::amdgpu::L1VectorCache::ensure_line(...)
#4  rocjitsu::amdgpu::L1VectorCache::read_bytes(...)
#5  rocjitsu::amdgpu::L1VectorCache::load(...)
#6  rocjitsu::amdgpu::GlobalMemPipeline::initiate_access(...)
#7  rocjitsu::amdgpu::MemoryPipeline::issue(...)
#8  rocjitsu::amdgpu::ComputeUnitCore::issue_instruction(...)
#9  simdojo::SimulationEngine::worker_loop(...)  ->  rj_vm_run()
```

While filling a cache line for the GEMM's vector load, `send_backing` dereferences a near-null pointer (faulting address `0x1000`) — an unguarded out-of-range access in the emulated memory backing rather than a wild GPU address.

## Pointers

- `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/l2_cache.cpp` — `L2Cache::send_backing` (crash frame).
- `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/addr_calc.cpp` — `flat_global_calculate_addresses` applies `vaddr *= scale` (`scaled_vaddr_factor`) under the same `scale_offset` folding that #8652 addressed for scalar loads; possibly a sibling issue in the vector path (unconfirmed).

## Notes

gfx1250 simulation is listed as Experimental in the rocjitsu support matrix. Filing separately from #8539 since that specific (scalar `s_load`) issue is fixed and verified, while this vector-load GEMM crash remains.

kuhar:
## Motivation

Fix gfx1250 buffer address calculation for zero-sized resource descriptors.

## Technical Details

The gfx1250 MUBUF path treated num_records == 0 as unbounded, allowing accesses through an all-zero optional-buffer descriptor. This could produce near-null addresses and crash workloads using RocJITsu directly

No ISA regeneration required.

## Issue Tracking

Fixes: https://github.com/ROCm/rocm-systems/issues/8751

## Test Plan

Ctest + repro from #8751

## Test Result

OK

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/addr_calc.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/addr_calc.cpp
@@ -238,9 +238,7 @@ void mubuf_calculate_addresses(const VbufferMachineInst &inst, amdgpu::Wavefront
     uint64_t total_offset =
         amdgpu::addr_calc::buffer_total_offset(index, stride, offset_part, soffset_val);
     bool oob;
-    if (oob_raw) {
-      oob = offset_part >= num_records;
-    } else if (stride > 0) {
+    if (!oob_raw && stride > 0) {
       oob = index >= num_records;
     } else {
       oob = offset_part >= num_records;
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/addr_calc.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/addr_calc.cpp
@@ -237,15 +237,13 @@ void mubuf_calculate_addresses(const VbufferMachineInst &inst, amdgpu::Wavefront
     uint32_t offset_part = amdgpu::addr_calc::buffer_offset_part(voffset, ioff);
     uint64_t total_offset =
         amdgpu::addr_calc::buffer_total_offset(index, stride, offset_part, soffset_val);
-    bool oob = false;
-    if (!oob && num_records != 0) {
-      if (oob_raw) {
-        oob = offset_part >= num_records;
-      } else if (stride > 0) {
-        oob = index >= num_records;
-      } else {
-        oob = offset_part >= num_records;
-      }
+    bool oob;
+    if (oob_raw) {
+      oob = offset_part >= num_records;
+    } else if (stride > 0) {
+      oob = index >= num_records;
+    } else {
+      oob = offset_part >= num_records;
     }
     if (oob) {
       d.lane_mask &= ~(1ULL << lane);
--- a/emulation/rocjitsu/tests/shared_infra_test.cpp
+++ b/emulation/rocjitsu/tests/shared_infra_test.cpp
@@ -3410,6 +3410,49 @@ TEST(Gfx1250AddrCalcTest, VbufferZeroNumRecordsMasksAllLanes) {
   EXPECT_EQ(d.per_lane_addr[1], 0ULL);
 }
 
+TEST(Gfx1250AddrCalcTest, VbufferStructuredStrideChecksIndexBounds) {
+  amdgpu::GpuMemory mem("gfx1250_vbuffer_structured_stride_mem");
+  amdgpu::L2Cache l2("gfx1250_vbuffer_structured_stride_l2");
+  amdgpu::ComputeUnitCore::Config cfg{};
+  cfg.arch = ROCJITSU_CODE_ARCH_GFX1250;
+  cfg.num_wf_slots = 1;
+  cfg.sgprs_per_wf = 128;
+  cfg.vgprs_per_wf = 16;
+  cfg.lds_size_kb = 64;
+  auto cu = amdgpu::ComputeUnitCore::create("gfx1250_vbuffer_structured_stride_cu", cfg, &mem, &l2);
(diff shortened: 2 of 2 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/addr_calc.cpp
```diff
@@ -229,37 +221,29 @@ void mubuf_calculate_addresses(const VbufferMachineInst &inst, amdgpu::Wavefront
...
-    if (!oob && num_records != 0) {
-      if (oob_raw) {
-        oob = offset_part >= num_records;
-      } else if (stride > 0) {
-        oob = index >= num_records;
-      } else {
-        oob = offset_part >= num_records;
-      }
+    bool oob;
+    if (oob_raw) {
```
atgutier: if (oob_raw) and the else case do the same thing so we could just have if (stride > 0) {} else {} right?

emulation/rocjitsu/tests/shared_infra_test.cpp
```diff
@@ -3373,6 +3372,44 @@ TEST(RdnaAddrCalcTest, Gfx1250VbufferWrapsOffsetPartBeforeBoundsCheck) {
...
+  cfg.vgprs_per_wf = 16;
+  cfg.lds_size_kb = 64;
+  auto cu = amdgpu::ComputeUnitCore::create("gfx1250_vbuffer_zero_records_cu", cfg, &mem, &l2);
+  ASSERT_NE(cu, nullptr);
+
+  auto *wf = cu->dispatch_wf(0, 0, 128, 16);
+  ASSERT_NE(wf, nullptr);
+  wf->set_exec(0x3ULL);
+
+  // A null optional pointer is represented by an all-zero descriptor. Its
```
atgutier, line 3391: Could we test non-zero descriptors with non-zero strides too?

atgutier (approved):
Approved with some nits.

newling (approved):
LGTM

I think there's an effort to run upstream workloads through rocjitsu, I wonder if that'd have caught this. hipblaslt is running some gfx1250 jobs in CI (ffm), perhaps it'd be good to get them running on rocjitsu too

Issue #8625, opened 2026-07-15
[rocjitsu] Model fieldless MR ISA operands as first-class semantic metadata


accauble:
## Summary/Motivation

MR ISA operands without a `<FieldName>` (VCC, EXEC, SCC, M0, PC, memory pseudo-operands, the FMA-K/`S_SETREG` literal, image `vaddr`, etc.) currently carry no semantics. Their architectural effects are instead modeled by hard-coded code: VCC carry paths call `wf.set_vcc()`, compare-x calls `wf.set_exec()`, scalar ALU calls `wf.write_scc()`, branches write `wf.pc`, and memory paths infer EXEC/M0/scratch/LDS state from instruction fields. This makes liveness, analysis, and cross-ISA sharing blind to real side effects and is a blocker for DBI probe/trampoline spill planning (which must know when EXEC/VCC/SCC/M0 have to be preserved).

The goal is to make fieldless operands first-class semantic metadata, then selectively connect each category to analysis, execution, disassembly, and cross-ISA sharing based on what it actually represents *without* faking special registers as SGPR/VGPR ranges or corrupting ordinary scratch liveness.

## The Plan

I want to break the tasks down to be as reviewable as possible since small changes to amdisa can lead to overwhelming code changes. The plan is to add 
Staged and reversible. Lock the taxonomy and add an explicit architectural-effect path first, land analysis metadata before touching runtime
behavior, then migrate one semantic family at a time with tests pinned before each refactor. 

Tasks:

- [x] **Enable fieldless operands in amdisa**: parser keeps fieldless operands, gives them stable names, constructors build them, and they're mostly inert for now (`OPR_SIMM32` is the one value-bearing exception). (PR #8788)
- [ ] **Architectural-effect API (C++)**: add a way to represent special-reg/memory effects on `InstDefUse`, without touching `RegisterSet` liveness. Shape it for future DBIT spill consumers. (PR #9112)
  - [ ] **Generate operand-level special-reg effects**: Generate an operand-level helper that maps real architectural special operands to special effects while keeping `to_register_ref()` for ordinary registers only.
  - [ ] **Make sure instruction-level side effect tests exist**: Tests need to exist for savexec, VCC carry, CMPX, branches, PC ops, etc. (Analysis tests; not simulation side effect tests)
[...]

PR #8788, opened 2026-07-17
feat(rocjitsu): Redo fieldless operand parsing

accauble:
## Motivation

The amdisa parser did not include fieldless operands (ones that are implicit or hardcoded into the instruction). The generated C++ still accounted for the hidden side effects, (see set_vcc()) but missing operands makes it difficult for rocjitsu analyses to use them (e.g. liveness analysis).

## Technical Details

This redoes PR #8426. The concept is similar -- most of the fieldless operands are inert for now, since changing just one can involve many changes to the C++ code and will quickly become unreviewable. This also keeps `OPR_SIMM32` not inert because there were some "bandaids" to make it work previously, and adding more bandaids on top of that was not worth it. Instead of a frozen set and lots of gates with comments to keep things in sync, this redesigns the "inert" gating to check capability flags that are set by a single policy. As we gradually update the operands (see Issue #8625), we just have to update the policy to reflect their integration.

## Test Result

All old tests pass. New tests were added for the new functionality.

## Submission Checklist


emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -6229,6 +6378,32 @@ def gen_insts(self) -> None:
...
+                        # Patch the 32-bit immediate operand from the encoding
+                        # base's literal_ member so it reads the real value
+                        # through the normal accessor (execute reads the operand,
+                        # not literal_). This must cover BOTH ways the immediate
+                        # is modeled across ISAs: rdna4 et al. emit a fieldless
+                        # `simm32`, while gfx1250 emits a field-BEARING `literal`
+                        # (OPR_SIMM32). _literal_operand_from_expr_stmt returns
+                        # None for any non-literal operand (e.g. the OPR_HWREG
+                        # selector), so it is safe to offer it every operand.
+                        # Fieldless operands still receive there capability
```
atgutier: Typo: there -> their

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -707,6 +706,42 @@ def _semantic_source_operands(
...
 
+    @classmethod
+    def _execute_operand_participates(cls, opnd: Operand) -> bool:
+        """Whether codegen should expose this operand to execute templates.
+
+        Field-bearing operands always participate. A fieldless operand
+        participates only if it reads a real value (its policy ``reads_value``
+        capability), which today is just the literal ``OPR_SIMM32``. Most
+        fieldless operands are hardwired side effects still handled by the
+        semantic emitters (VCC/EXEC/SCC/M0/PC) and stay out of the positional
```
atgutier, line 767: Should these all be unified or should we keep those non-immediate fieldless operands out of the positional list? The PR description states making the PR reviewable as a goal, is that the main reason we don't unify here?

atgutier (commented):
Some from the agent:

MINOR

  1. operand.h — new capability data members are public, undermining the set-once invariant.
  reads_value_, writable_, fieldless_ sit in a public: block, so any holder of an Operand& can flip
  them past apply_fieldless_caps(). Fix: move the raw members to protected: (subclass ctors set
  is_vgpr_; apply_fieldless_caps is a member and retains access); keep only accessors public.

  2. operand.h — apply_fieldless_caps is an unguarded public post-construction mutator. The
  thread-safety argument depends on "called only during construction, never races a reader" (readers
  do lockless bool reads on the CU thread). All current call sites are generated constructors, so it's
   safe today, but the invariant is load-bearing and unenforced. Fix: document the invariant;
  optionally assert construction-only use or route the flag-set through the ctor and drop the setter.

  3. simd_codegen.py:883-896 — SIMD FMA table hard-codes the literal operand name simm32. The
  inst.simm32_ → inst.simm32.encoding_value_ change silently assumes the operand is named simm32;
  gfx1250 names it literal. Harmless today because the sharing preflight keeps gfx1250's FMA
  ISA-local, but a latent trap if gfx1250 is ever pulled into this shared SIMD path (would emit
  inst.simm32 and fail to compile). Fix: assert/document the name assumption, or derive the K
  expression from the resolved literal operand name.

  4. parser.py:27-28 — stale module docstring. Item #7 still documents the removed "FMAMK/FMAAK fall
  back to inst_.simm32" behavior, which this PR eliminates. Fix: update to state FMAMK/FMAAK now
  require the fieldless OPR_SIMM32 operand.

  ---
  NIT

  - cross_isa.py:108 — TODO: Add fieldless operands is misleading; the fieldless exclusion is
  deliberate and load-bearing (keeps rdna FMA shareable while separating gfx1250's field-bearing
  literal). Reword or drop.
  - test_fieldless_operands.py — no direct unit test for FieldlessCaps.__post_init__ raising, the
  DPP/SDWA by-name selection, or the FMA len(src_ops)<3 ValueError (all exercised indirectly / via
  full regen). Cheap to pin, especially the __post_init__ invariant (a cross-language contract with
  the SIMD write gate).
  - instruction.h disassembly — is_fieldless() skip changes golden output (new smoke test covers
  s_and_saveexec_b64); confirm no wider disasm golden corpus regressed in CI.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
accauble: I think you are asking why not add all the fieldless operands to the positional list now, then yeah, the main reason is that there are quite a few and I thought that propagating all at once would cause a horrible amount of C++ churn and would basically be impossible to review. I do plan to get them there eventually, unless someone is against that.

emulation/rocjitsu/lib/python/amdisa/tests/test_literal_codegen.py
```diff
@@ -393,6 +437,30 @@ def test_scalar_mul_u64_generated_execute_reads_full_source_pairs():
...
+    rdna_codegen.config = SimpleNamespace(unshared_execute_keys=frozenset())
+
+    gfx_codegen = object.__new__(CodeGenerator)
+    gfx_codegen.isa_spec = SimpleNamespace(
+        arch_name='gfx1250', profile=Gfx1250Profile()
+    )
+    gfx_codegen.shared_plan = plan
+    gfx_codegen.config = SimpleNamespace(unshared_execute_keys=frozenset())
+
+    assert rdna_codegen._can_share_execute('s_fmaak_f32', enc_name='ENC_SOP2')
```
kuhar, line 527: We should make `_can_share_execute()` match the full `(mnemonic, encoding_name)` key. With this test plan, `_can_share_execute("s_fmaak_f32", enc_name="ENC_SOP1")` returns `True` even though only `ENC_SOP2` is shared, which can select a shared execute body for an ISA-local layout. Please add that wrong-encoding negative assertion here.

kuhar (commented):
Just one comment, I think it looks fine otherwise

accauble:
@atgutier 

> 
> 1. operand.h — new capability data members are public, undermining the set-once invariant.
>    reads_value_, writable_, fieldless_ sit in a public: block...

Fixed.

> 2. operand.h — apply_fieldless_caps is an unguarded public post-construction mutator. The
>    thread-safety argument depends on "called only during construction, never races a reader" (readers
>    do lockless bool reads on the CU thread)...

Documented.

> 3. simd_codegen.py:883-896 — SIMD FMA table hard-codes the literal operand name simm32. The
>    inst.simm32_ → inst.simm32.encoding_value_ change silently assumes the operand is named simm32;
>    gfx1250 names it literal...

Asserted and documented and tested.

> 4. parser.py:27-28 — stale module docstring. Item #7 still documents the removed "FMAMK/FMAAK fall
>    back to inst_.simm32" behavior, which this PR eliminates....

Done.

> * cross_isa.py:108 — TODO: Add fieldless operands is misleading; the fieldless exclusion is
>   deliberate and load-bearing (keeps rdna FMA shareable while separating gfx1250's field-bearing
>   literal). Reword or drop.

Reworded.

> * test_fieldless_operands.py — no direct unit test for FieldlessCaps.**post_init** raising, the
>   DPP/SDWA by-name selection, or the FMA len(src_ops)<3 ValueError...

Added.

> * instruction.h disassembly — is_fieldless() skip changes golden output (new smoke test covers
>   s_and_saveexec_b64); confirm no wider disasm golden corpus regressed in CI.

Confirmed.

emulation/rocjitsu/tests/scalar_scc_test.cpp
```diff
@@ -281,11 +281,23 @@ void expect_wrexec_def_use(const ScalarSccProfile &profile, const WrexecPair &pa
 
     const RegisterRef source{RegClass::SGPR, kSourceSgpr, width};
     const RegisterRef destination{RegClass::SGPR, kDestSgpr, width};
-    ASSERT_EQ(inst->num_src_operands(), 1) << profile.name << " " << mnemonic;
-    ASSERT_EQ(inst->num_dst_operands(), 1) << profile.name << " " << mnemonic;
+    // The implicit EXEC read/write and SCC def are modeled as inert fieldless
+    // operands (to_register_ref() == nullopt): they add to the operand counts
+    // but contribute nothing to def/use at this time. The field-bearing SGPR
+    // source/dest stay at index 0.
+    ASSERT_EQ(inst->num_src_operands(), 2) << profile.name << " " << mnemonic;
```
accauble, line 291: @lua1235 I had to update these tests. I want to make sure you saw that and this doesn't wreak havoc on your end.

emulation/rocjitsu/lib/python/amdisa/tests/test_generator_profile_gates.py
```diff
@@ -2591,7 +2591,10 @@ def test_generated_atomic_def_use_follows_return_control(
     cdna4_flat = (amdgpu_generated_root / 'cdna4' / 'flat.cpp').read_text()
     cdna4_add = cdna4_flat.split('FlatAtomicAddX2Flat::FlatAtomicAddX2Flat')[1]
     cdna4_add = cdna4_add.split('void FlatAtomicAddX2Flat::execute_impl')[0]
-    assert 'num_dst_ = 0;' in cdna4_add
+    # The atomic writes memory with OPR_GPUMEM operand as the baseline
```
accauble, line 2690: @Groverkss Just want to make you aware that I made some minor changes to the tests you just added.

emulation/rocjitsu/tests/decode_smoke_test.cpp
```diff
@@ -383,7 +383,18 @@ TEST(Rdna3DecodeTest, GlobalFlatAllOnesSaddrIsNull) {
   std::unique_ptr<Instruction> inst(decoder->decode(words.data()));
   ASSERT_NE(inst, nullptr);
   EXPECT_EQ(inst->mnemonic(), "global_load_b64");
-  EXPECT_EQ(inst->num_src_operands(), 1);
+  // vaddr + the fieldless OPR_GPUMEM memory operand (saddr is null, so it
+  // is not added); gpumem is counted but currently carries no register.
+  EXPECT_EQ(inst->num_src_operands(), 2);
```
accauble, line 388: @Groverkss Just want to make you aware that I made some minor changes to the tests you just added.

emulation/rocjitsu/lib/python/amdisa/cross_isa.py
```diff
@@ -103,10 +103,20 @@ def _sem_key(sem: InstructionSemantics | None) -> tuple[str, str | None, str | N
...
+    body. As operands are promoted to participate in execute this exclusion
+    must narrow to match ``_execute_operand_participates``: an operand
+    that changes the body has to be in the signature, or differing bodies
+    collide (caught by the body-equality assert in the generator). ``simm32``
+    already participates and is the boundary case.
+    """
     return tuple(
         (op.name, op.operand_type, op.size, op.is_input, op.is_output)
         for op in inst.operands
+        if not op.fieldless
```
kuhar: Please make this signature use the same participation rule as execute generation. `_execute_operand_participates()` intentionally keeps fieldless `OPR_SIMM32` because it changes the generated body, but this filter drops it with the inert side-effect operands. As a result, two otherwise-identical instructions—one with the value-bearing literal and one without it—produce equal structural signatures even though their execute operand lists differ, so the analyzer can admit an incompatible pair into the shared plan and rely on the later body preflight to recover. Filter out only fieldless operands whose policy says they do not read a value, and update the signature test to cover both inert exclusion and `OPR_SIMM32` inclusion.

emulation/rocjitsu/lib/python/amdisa/codegen/execute/simd_codegen.py
```diff
@@ -882,23 +885,43 @@
...
-    'v_fmamk_f16_vop2': ('uint32_t', 'inst.simm32_', _FMA_MULK_F16),
+    'v_fmamk_f16_vop2': ('uint32_t', 'inst.simm32.encoding_value_', _FMA_MULK_F16),
     'v_madmk_f16_vop2': ('uint32_t', 'inst.simm32.encoding_value_', _FMA_MULK_F16),
-    # v_fmaak_f16 (RDNA only): dst = fma(s0, s1, K), K = f16(simm32_). Same f16
-    # FMA functor as v_madak_f16, differing only in the literal field (simm32_ vs
-    # simm32.encoding_value_); the SIMD path is identical to the tested madak_f16.
-    'v_fmaak_f16_vop2': ('uint32_t', 'inst.simm32_', _FMA_ADDK_F16),
+    # v_fmaak_f16 (RDNA only): dst = fma(s0, s1, K), K = f16(simm32).
+    # Same f16 FMA functor as v_madak_f16.
+    'v_fmaak_f16_vop2': ('uint32_t', 'inst.simm32.encoding_value_', _FMA_ADDK_F16),
```
kuhar: It would be useful to cover the two f16 FMA-K forms whose literal read changes here with a scalar-vs-SIMD test. The existing correctness fixture only covers CDNA4, where `v_fmaak_f16` and `v_fmamk_f16` do not exist, so this needs an RDNA-keyed fixture (or an arch-parameterized version) rather than adding RDNA opcodes to `kCases`. That would exercise `inst.simm32.encoding_value_` directly for both changed f16 paths.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/instruction.h
```diff
@@ -265,8 +268,8 @@ class Instruction : public util::IListNode<Instruction, util::IListParent<BasicB
   /// @brief Instruction's source operands (max 6).
   std::array<Operand *, 6> src_operands_{};
   uint8_t num_src_ = 0;
-  /// @brief Instruction's destination operands (max 2).
-  std::array<Operand *, 2> dst_operands_{};
+  /// @brief Instruction's destination operands (max 3).
+  std::array<Operand *, 3> dst_operands_{};
```
kuhar, line 276: This is a great foundation for pulling special-register side effects into the real operand model -- but bumping `dst_operands_` from 2 to 3 (and the generator now pushing fieldless def/use operands into `src_operands_`/`dst_operands_` positionally) quietly puts us at the edge of the fixed-size arrays with no guard.

Evidence from the regen at this head:
- `num_dst_ = 3` is emitted for 194 instructions -- i.e. the size-3 `dst_operands_` is already at *full* capacity (zero headroom) for the save-exec-style ops (sdst + fieldless EXEC def + fieldless SCC def).
- `num_src_ = 6` is emitted for 3 instructions -- `src_operands_` (max 6) is likewise at capacity.
- The generator increments `dst_idx`/`src_idx` for each qualifying operand and emits `num_dst_ = {dst_idx}` / `num_src_ = {src_idx}` with no generation-time bound check.

The PR's own direction ("TODO: Incorporate fieldless operand side effects into execute bodies", and the plan to promote the remaining fieldless special registers) means the *next* fieldless def added to any of those boundary instructions writes past the end of a `std::array` -- silent UB, no compile error, no assert.

Since this is exactly the kind of corner the fieldless work is designed to grow into, could we add a cheap generation-time tripwire (e.g. `assert dst_idx <= 3` / `assert src_idx <= 6` right before emitting `num_dst_`/`num_src_`), or better, derive the array bounds from the max observed across the regen so the C++ capacity and the generator can't drift apart? That keeps the door open for promoting more fieldless operands without a booby-trap.

emulation/rocjitsu/lib/python/amdisa/cross_isa.py
kuhar: Nice that the whole fieldless story keys off one policy table -- but here `_operand_signature` excludes *all* fieldless operands, while `_execute_operand_participates` (in `_generator.py`) includes value-bearing fieldless operands (today `OPR_SIMM32`). So the sharing signature and the execute-visible operand set have *already* diverged: two instructions that differ only in whether they carry a fieldless `simm32` produce identical signatures here and get grouped as shareable, even though their generated bodies differ.

The docstring is admirably honest about this ("`simm32` already participates and is the boundary case") and the body-equality assert in the generator is the backstop -- so this fails loudly at generation rather than emitting wrong code, which is good. My concern is forward-looking: as more fieldless operands get promoted to participate (the effect/display slices you have stubbed out), this exclusion has to be narrowed by hand in lockstep, and the two notions of "participates" will keep drifting.

Could we have `_operand_signature` filter through the same predicate as the generator -- i.e. exclude a fieldless operand only when it does *not* `reads_value` (share `_execute_operand_participates`, or a small shared helper both call)? Then the sharing signature stays a faithful proxy for the execute body by construction, and promoting the next fieldless operand is a one-line policy-table change instead of a "remember to update two exclusion sites" gotcha.

kuhar (commented):
Stepping back: `fieldless_policy.py` as a single source of truth (role / caps / display / effect on one row, with `validate_fieldless_taxonomy` as a parse-time tripwire) is a really clean foundation, and I like that runtime capability and codegen participation both derive from it so they cannot silently disagree. A few forward-looking notes -- none blocking:

- The `effect` field and `FieldlessDisplay.VISIBLE` are exactly the right extension points for the special-register-effect and disassembly slices. Worth a one-line pointer in the PR description sketching that arc so future readers see where `MEMORY_PSEUDO`/`SPECIAL_REGISTER` are heading (e.g. FLAT_SCRATCH's "revisit when it drives execute behavior" note).

- The value-bearing literal (`OPR_SIMM32`, the FMAAK/FMAMK K) now participates in execute but is `HIDDEN` in disassembly, and `instruction.h` carries a matching "TODO: Include explicit fieldless operands". That is fine today (the literal was never printed via the operand path before), but it means the assembler-form literal will need to become `VISIBLE` when you get there -- keeping display keyed off the same policy row (rather than a separate suppression check) will make that a one-liner.

- If you would like, I am happy to review the follow-up that promotes the remaining special registers into execute bodies -- that is where the participation/signature/array-capacity concerns I flagged separately will all converge, so it would be good to land those guards first.

kuhar (commented):
Original anchor: `emulation/rocjitsu/lib/python/amdisa/fieldless_policy.py:155`

We should classify `OPR_FLAT_SCRATCH` as architectural register state, not `MEMORY_PSEUDO`. The RDNA3 ISA manual §3.1 calls FLAT_SCRATCH a per-wave hardware state and “the base address of scratch memory ... used by Flat and Scratch instructions”; the CDNA4 manual likewise places the 64-bit base in NumSGPRs-5/-6. That makes it categorically different from `OPR_DSMEM`/`OPR_GPUMEM`. Because this table is the golden-tested semantic source of truth and the future effect slice consumes `role`, deferring the classification will route effect/liveness handling down the wrong path. The normal-access caps can stay inert until the special-register effect work, but the role should be `SPECIAL_REGISTER` (or a dedicated register-state category) now.

Original anchor: `emulation/rocjitsu/lib/python/amdisa/fieldless_policy.py:158`

We should not model fieldless `OPR_VGPR` as an inert placeholder. RDNA4 ISA §10.1 defines `VADDR0-4` as five 8-bit fields that each select one or a consecutive group of VGPRs; §10.9.2 maps those fields to the BVH ray inputs, and the BVH8/Dual instruction descriptions mark VADDR2/3 as input/output state. At this head the generated image constructors instead create `vaddr(..., 0)` with `(reads_value=false, writable=false, is_vgpr=false)`, so `to_register_ref()` and liveness lose real architectural VGPR dependencies. NSA may require a composite operand rather than one base+width `Operand`, but please represent that explicitly instead of golden-locking the address bundle as non-VGPR/inert.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/instruction.h
accauble: Addressed

emulation/rocjitsu/lib/python/amdisa/cross_isa.py
accauble: I have updated this but since `OPR_SIMM32` is now included in the signature, this makes shared execute changes and legalization table changes. They are limited to only a few instructions so I think it is still reviewable. Lmk if you would prefer I revert the change and make it a follow up PR.

accauble:
> Original anchor: `emulation/rocjitsu/lib/python/amdisa/fieldless_policy.py:155`
> 
> We should classify `OPR_FLAT_SCRATCH` as architectural register state, not `MEMORY_PSEUDO`. The RDNA3 ISA manual §3.1 calls FLAT_SCRATCH a per-wave hardware state and “the base address of scratch memory ... used by Flat and Scratch instructions”; the CDNA4 manual likewise places the 64-bit base in NumSGPRs-5/-6. That makes it categorically different from `OPR_DSMEM`/`OPR_GPUMEM`. Because this table is the golden-tested semantic source of truth and the future effect slice consumes `role`, deferring the classification will route effect/liveness handling down the wrong path. The normal-access caps can stay inert until the special-register effect work, but the role should be `SPECIAL_REGISTER` (or a dedicated register-state category) now.

Alright, that's fair. Done.

> 
> Original anchor: `emulation/rocjitsu/lib/python/amdisa/fieldless_policy.py:158`
> 
> We should not model fieldless `OPR_VGPR` as an inert placeholder. RDNA4 ISA §10.1 defines `VADDR0-4` as five 8-bit fields that each select one or a consecutive group of VGPRs; §10.9.2 maps those fields to the BVH ray inputs, and the BVH8/Dual instruction descriptions mark VADDR2/3 as input/output state. At this head the generated image constructors instead create `vaddr(..., 0)` with `(reads_value=false, writable=false, is_vgpr=false)`, so `to_register_ref()` and liveness lose real architectural VGPR dependencies. NSA may require a composite operand rather than one base+width `Operand`, but please represent that explicitly instead of golden-locking the address bundle as non-VGPR/inert.

I am really hesitant to add functionality in this PR. The reason I was okay enabling `OPR_SIMM32` is because it was already hacked in, and it was actually simpler to enable it than to update the hacks to accommodate its presence. Representing `OPR_VGPR` explicitely would be adding completely new functionality and it's not trivial. There is a plan to enable it eventually (see Issue #8625).

kuhar (approved):
LGTM

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
atgutier: Yep, that is what I was asking. No need to do it all at once. If it's in the plan then that sounds great.

atgutier (approved):
Approved with some agent comments for you to consider, either here or as a follow up:

MINOR

  2. Legalization tables changed semantic decisions at scale — described as "mechanical," only
  structurally tested — code/dbt/generated/legalization_*.h
  The PR frames the ±264–776-line churn per table as a mechanical operand-index consequence, but it's
  a real DBT behavior change: in legalization_cdna3_to_cdna4.h alone, ~1024 entries flip
  Action::Lower→Action::Identity (4699→3675 Lower), and there's a large net Lower reduction across all
   tables. The direction is consistent with the design (now that fieldless OPR_SIMM32 participates in
  the operand signature with canonicalized 32-bit size, literal FMA/MAD-K instructions that previously
   "differed" and got Lower now correctly match as Identity) — and translate_test.cpp adds
  LegalizationFmaK.* cases plus a zero-ILLEGAL/non-empty invariant. But the added tests only pin the
  FMA-K family and the structural invariant, not the specific decision flips. Residual risk: if the
  operand-signature comparator now treats an inert fieldless operand as "matching" on some other
  family, an instruction that genuinely needs lowering could silently pass through unlowered.
  Fix: author to (a) confirm the comparator only flips Lower→Identity when the two ISAs are truly
  encoding-compatible for that opcode, and (b) pin a couple of non-FMA-K flipped outcomes so the guard
   isn't FMA-K-only.

  3. FieldlessCaps docstring misstates the runtime gate — fieldless_policy.py:84-91
  The writable|is_vgpr ⟹ reads_value invariant is correct and necessary, but the stated rationale
  ("SIMD write helpers gate on reads_value") is wrong: the write helper gates on is_writable(); it's
  simd_capable() (required on every operand incl. the destination) that gates on reads_value(). So a
  writable, !reads_value destination would be excluded from the SIMD path entirely. Right invariant,
  wrong explanation — will mislead whoever next decides whether it can be relaxed. Fix the prose (and
  the mirrored test comment).

  4. FMA-K literal read bypasses the RegisterAccess facade — execute_shared.h (v_fmaak/v_fmamk
  f16/f32)
  Reads the literal as inst.simm32.encoding_value_ directly instead of
  RegisterAccess(wf).read_scalar(inst.simm32) like the sibling SOP2 FMA-K executors added in the same
  diff. Not a regression (develop read the raw member too) and value-identical, but inconsistent with
  the facade contract (read-observation plugins won't see it). Route it through read_scalar for
  uniformity.

  5. apply_fieldless_caps construction-only contract enforced only by convention — operand.h:137-142
  Verified safe today (all ~9,500 call sites are in generated constructors; no other mutation path;
  decode+execute share one CU thread so the lockless read is race-free by program order). But nothing
  in C++ prevents a future non-ctor caller from mutating a live operand → data race. Cheap insurance:
  assert(!fieldless_) at the top so a second/late call trips in debug. This becomes load-bearing when
  #8625 lands a writable fieldless operand (see below).

 6. Latent write-path divergence for a future writable fieldless operand — isa_operand_simd_inl.h /
  generated operand.cpp
  write_lane_chunk gates on is_writable(), but resolved_vgpr_offset_for_operand (used within it) gates
   on reads_value(); the scalar fallback write_lane resolves without the reads_value gate. Today
  harmless — the Python __post_init__ enforces writable ⟹ reads_value and no emitted triple violates
  it (only (f,f,f) and (t,f,f) exist). But the SIMD-fast-path vs scalar-fallback would disagree the
  moment a writable, !reads_value operand exists. Encode the invariant as a C++ assert(reads_value_ ||
   !writable_) in apply_fieldless_caps now.

  7. Cross-file capacity literals hand-duplicated — _generator.py:6801-6813 vs instruction.h
  _DST_OPERANDS_CAPACITY=3 / _SRC=6 are manually kept in sync with the std::array<Operand*,N> sizes in
   the hand-written header (the dst array just grew 2→3 for this PR). Grow one without the other and
  the assert either falsely blocks generation or misses overflow. Source from a single shared constant
   or add a regen test that greps the header.

  8. Latent S_SETREG_IMM32_B32 literal double-patch — _generator.py:6939-7001
  Two constructor-body loops can each emit a simm32 = ... fixup for the same operand on an ISA whose
  SOPK carries a literal machine-inst struct. Benign today (last-wins; current regen clean) but not
  enforced mutually exclusive. Make the S_SETREG loop elif-exclusive with the generic literal loop.

  ---
  NIT

  - Policy display/effect columns are declared "single source of truth" but not yet wired to any path
  (documented stubs for #8625) — the framing overstates current reach.
  - instruction.h:244 leaves a TODO suppressing all fieldless operands (incl. the SIMM32 literal) from
   disassembly — confirmed not a regression, fine to ship.
  - operand.h mixes public raw fields (size_bits_, encoding_value_) with the new protected caps flags
  — add a one-line note on why, to preempt review churn.
  - Vestigial src2 fallback in _semantic_source_operands after the synthesis path was removed (dead
  for real generation).

accauble:
@atgutier -- Since most of these were really quick fixes I just did most of them here.
> 
> 2. ...Fix: author to (a) confirm the comparator only flips Lower→Identity when the two ISAs are truly
>    encoding-compatible for that opcode, and (b) pin a couple of non-FMA-K flipped outcomes so the guard
>    isn't FMA-K-only.

Confirmed. The fieldless `simm32` only exists on the FMA-K/MAD-K family, so I can't test the non-FMA-K outcome.

> 3. FieldlessCaps docstring misstates the runtime gate...Fix the prose (and
>    the mirrored test comment).

Done.

> 4. FMA-K literal read bypasses the RegisterAccess facade — execute_shared.h (v_fmaak/v_fmamk
>    f16/f32)... Route it through read_scalar for  uniformity.

Done.

> 5. apply_fieldless_caps construction-only contract enforced only by convention... Cheap insurance:
>    assert(!fieldless_) at the top so a second/late call trips in debug. This becomes load-bearing when
>    #8625 lands a writable fieldless operand (see below).

Done.

> 6. Latent write-path divergence for a future writable fieldless operand ... Encode the invariant as a C++ assert(reads_value_ || !writable_) in apply_fieldless_caps now.

Done. 

> 7. Cross-file capacity literals hand-duplicated...Source from a single shared constant
>    or add a regen test that greps the header.

This is a good point. Unfortunately, sourcing from a single source isn't a clearly easy task (instruction.h is not an auto-generated file) and I dunno about a test that greps the header. If you really want that I can add that in the next PR. Another option which I would float by you is generating a header file that includes constants? We don't actually have that AFAICT so I didn't want to add that here but if you like that design I could add it to the next PR.

> 8. Latent S_SETREG_IMM32_B32 literal double-patch... Make the S_SETREG loop elif-exclusive with the generic literal loop.

Done.

> 
> * Policy display/effect columns are declared "single source of truth" but not yet wired to any path
>   (documented stubs for #8625) — the framing overstates current reach.

Reworded.

> * operand.h mixes public raw fields (size_bits_, encoding_value_) with the new protected caps flags
>   — add a one-line note on why, to preempt review churn.

Done.

PR #8789, opened 2026-07-17
fix(rocprof-systems): install rocJPEG/rocDecode tests only when data available

sputhala-amd:
## Motivation

Building/installing the `jpegdecode` and `videodecode` examples against ROCm distributions that ship rocJPEG/rocDecode (headers + library) but not their sample data - the JPEG images (`share/rocjpeg/images`) or the H26x videos
(`share/rocdecode/video`), e.g. the nightly TheRock multi-arch tarballs - caused the tests to fail due to missing data.

## Technical Details

- `tests/pytest/test_videodecode.py`: Added a `require_video_data` fixture that inspects `rocprofsys_examples_dir/videos` and skips the test when the directory is absent or contains no *.mp4 files.

- `tests/pytest/test_jpegdecode.py`: Added a `require_jpeg_data` fixture that inspects `rocprofsys_examples_dir/images` and skips the test when the directory is absent or empty.

- Runtime data-availability gating: The decode tests now determine at execution time whether the required sample media is present, instead of relying on build-time configuration, so the example binaries are still exercised whenever data exists.

- Skip reason: Both fixtures emit a message identifying the missing data directory and noting the run was "possibly built against a non-test build which doesn't have those files", so skipped runs are clearly attributable to a non-test package rather than a test failure or environment misconfiguration.

## Issue Tracking

Jira ID: AIPROFSYST-587

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-systems/tests/pytest/test_jpegdecode.py
+++ b/projects/rocprofiler-systems/tests/pytest/test_jpegdecode.py
@@ -53,6 +53,23 @@ def get_run_args(rocprof_config) -> list[str]:
     return ["-i", str(rocprof_config.rocprofsys_examples_dir / "images"), "-b", "32"]
 
 
+@pytest.fixture
+def require_jpeg_data(rocprof_config) -> None:
+    """Skip the test at runtime when the sample image data is not available.
+
+    The jpegdecode example is always built when rocJPEG is present, but the
+    sample images are only shipped by test builds of ROCm. When they are missing
+    there is nothing to decode, so skip instead of failing.
+    """
+    images_dir = rocprof_config.rocprofsys_examples_dir / "images"
+    image_files = sorted(images_dir.iterdir()) if images_dir.is_dir() else []
+    if not image_files:
+        pytest.skip(
+            f"No rocJPEG sample images found in {images_dir}; "
+            "possibly built against a non-test build which doesn't have those files."
+        )
+
+
 # =============================================================================
 # JPEG decode tests
 # =============================================================================
@@ -68,7 +85,15 @@ def get_run_args(rocprof_config) -> list[str]:
 )
 @pytest.mark.class_name("jpeg-decode")
 class TestJPEGDecode(RocprofsysTest):
-    def test(self, mode, jpeg_decode_env, jpeg_decode_rules, get_run_args, gpu_info):
+    def test(
+        self,
+        mode,
+        jpeg_decode_env,
+        jpeg_decode_rules,
+        get_run_args,
+        gpu_info,
+        require_jpeg_data,
+    ):
         result = self.run_test(
             mode,
             "jpegdecode",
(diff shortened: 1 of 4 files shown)


projects/rocprofiler-systems/tests/pytest/test_jpegdecode.py
```diff
@@ -53,6 +53,23 @@ def get_run_args(rocprof_config) -> list[str]:
...
+@pytest.fixture
+def require_jpeg_data(rocprof_config) -> None:
+    """Skip the test at runtime when the sample image data is not available.
+
+    The jpegdecode example is always built when rocJPEG is present, but the
+    sample images are only shipped by test builds of ROCm. When they are missing
+    there is nothing to decode, so skip instead of failing.
+    """
+    images_dir = rocprof_config.rocprofsys_examples_dir / "images"
+    image_files = sorted(images_dir.iterdir()) if images_dir.is_dir() else []
```
kcossett-amd: Why are you sorting here?

projects/rocprofiler-systems/tests/pytest/test_videodecode.py
```diff
@@ -51,6 +51,23 @@ def get_run_args(rocprof_config) -> list[str]:
...
+@pytest.fixture
+def require_video_data(rocprof_config) -> None:
+    """Skip the test at runtime when the sample video data is not available.
+
+    The videodecode example is always built when rocDecode is present, but the
+    H26x sample videos are only shipped by test builds of ROCm. When they are
+    missing there is nothing to decode, so skip instead of failing.
+    """
+    videos_dir = rocprof_config.rocprofsys_examples_dir / "videos"
+    video_files = sorted(videos_dir.glob("*.mp4")) if videos_dir.is_dir() else []
```
kcossett-amd: Why are you sorting here?

projects/rocprofiler-systems/tests/pytest/test_jpegdecode.py
kcossett-amd: I see here you accept any pattern whilst in videodecode, it is limited to mp4 files.

sputhala-amd: Absolutely unnecessary sort. Removed.

projects/rocprofiler-systems/tests/pytest/test_videodecode.py
sputhala-amd: Absolutely unnecessary sort. Removed.

projects/rocprofiler-systems/tests/pytest/test_jpegdecode.py
sputhala-amd: Correct, when building the videodecode samples the cmake only copies mp4 files. I vaguely remember that the other formats present in the directory were giving me problems when I first copied the samples from upstream project.
Changed the logic here to look for any files.

sputhala-amd: We leave the responsibility to copy the files to the build step.
At the test runtime we will only check if the data files are available and will run the test with whatever was copied in the data directory.

kcossett-amd (approved):
Thank you

PR #8793, opened 2026-07-17
fix(rocr): Enable profiling on internal blit copy queues

dayatsin-amd:
## Summary

The command processor latches `amd_queue_.queue_properties` (including the `ENABLE_PROFILING` bit) at a queue's **first dispatch** and does not re-read the in-memory value afterwards. When async-copy profiling is toggled *after* an internal blit copy queue has already run copies (e.g. a profiler attaching late), changing the bit has no effect: blit-kernel copy completion timestamps (`amd_signal_t.start_ts`/`end_ts`) silently remain `0` even though `queue_properties` shows the bit set (`0xa`).

This PR fixes it in two complementary ways:

1. **Enable profiling on the internal blit copy queues right after creation** (`QueueBlitOnly` and the shared `QueueUtility`), before their first dispatch, so the common case (profiling enabled before/at first use) reliably latches `ENABLE_PROFILING` on with no runtime overhead.

2. **Re-latch on change via queue remap.** `AqlQueue::SetProfiling` now detects when the `ENABLE_PROFILING` bit actually changes on a queue that has already submitted packets (`write index != 0`) and unmaps/remaps the queue (`Suspend()`/`Resume()`) so the CP FW re-reads the queue properties. This makes enabling **and disabling** (and re-enabling) profiling take effect at any point in a queue's lifetime, not just before the first dispatch.

Together these ensure copy completion timestamps are populated whenever profiling is on, regardless of when it is toggled. They are only surfaced to callers when profiling is actually enabled, since `hsa_amd_profiling_get_async_copy_time` gates on `async_copy_agent()`.

This also addresses the earlier review question about the disable path: `hsa_amd_profiling_async_copy_enable(false)` now genuinely takes effect (the queue is remapped and the CP stops writing timestamps), and a later re-enable works again.

Notes:
- Scope is limited to the two internal blit-copy queues; `QueuePCSampling`, GWS/counted queues, and all user-created queues are unaffected.
- The remap only happens when the bit actually changes on a queue that has already dispatched (`write index != 0`); `write_dispatch_id` is a monotonic 64-bit counter and never wraps back to `0`, so the guard is safe.
- SDMA copies use a separate timestamp mechanism and are unaffected.

## JIRA ID

JIRA ID : ROCM-28243

## Test plan

- [x] Late-enable repro (H2D/D2H via `hsa_amd_memory_async_copy`, `HSA_ENABLE_SDMA=0`, profiling enabled *after* warm-up copies): timestamps were `0` before, valid non-zero after — verified on **gfx1250**.
- [x] Disable-then-re-enable toggle: after packets have been dispatched, disable profiling, run copies, re-enable, and confirm timestamps are non-zero again — verified on **gfx1250**. A/B: fails (all-zero) on stock runtime, passes with this change.
- [x] Regression: profiling enabled *before* first copy still returns valid timestamps (no change).

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
@@ -1585,7 +1585,22 @@ hsa_status_t AqlQueue::GetCUMasking(uint32_t num_cu_mask_count, uint32_t* cu_mas
 }
 
 void AqlQueue::SetProfiling(bool enabled) {
+  bool need_update = false;
+  const bool cur = AMD_HSA_BITS_GET(amd_queue_.queue_properties,
+    AMD_QUEUE_PROPERTIES_ENABLE_PROFILING) != 0;
+
+  if (cur != enabled && LoadWriteIndexRelaxed() != 0) {
+    // If the queue is already enabled/disabled, we need to update the queue properties and we have already submitted packets,
+    // then we need to unmap and remap the queue for CP FW to re-read the queue properties.
+    need_update = true;
+  }
+
   Queue::SetProfiling(enabled);
+
+  if (need_update) {
+    Suspend();
+    Resume();
+  }
 
   if (enabled) agent_->CheckClockTicks();
   return;
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
@@ -908,10 +908,21 @@ void GpuAgent::InitDma() {
     return queue;
   };
 
+  // Enable profiling on the internal blit copy queues right after creation so that
+  // copy completion timestamps are always available regardless of when async-copy
+  // profiling is later turned on (e.g. a profiler attaching after copies have run).
   // Dedicated compute queue for host-to-device blits.
-  queues_[QueueBlitOnly].reset(queue_lambda);
+  queues_[QueueBlitOnly].reset([queue_lambda]() {
+    auto queue = queue_lambda();
+    queue->SetProfiling(true);
+    return queue;
+  });
   // Share utility queue with device-to-host blits.
-  queues_[QueueUtility].reset(queue_lambda);
+  queues_[QueueUtility].reset([queue_lambda]() {
+    auto queue = queue_lambda();
+    queue->SetProfiling(true);
+    return queue;
+  });
 
   // Dedicated compute queue for PC Sampling CP-DMA commands. We need a dedicated queue that runs at
   // highest priority because we do not want the CP-DMA commands to be delayed/blocked due to
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
@@ -908,10 +908,10 @@ void GpuAgent::InitDma() {
     return queue;
   };
 
-  // Enable profiling on the internal blit copy queues right after creation so that
-  // copy completion timestamps are always available regardless of when async-copy
-  // profiling is later turned on (e.g. a profiler attaching after copies have run).
-  // Dedicated compute queue for host-to-device blits.
+  // Enable profiling on the internal blit copy queues right after creation to avoid
+  // having to unmap and remap the queue for CP FW to re-read the queue properties when
+  // profiling is later turned on. If profiling is disabled later on these queues, then
+  // the queue unmap and remap will be triggered.
   queues_[QueueBlitOnly].reset([queue_lambda]() {
     auto queue = queue_lambda();
     queue->SetProfiling(true);


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -900,10 +900,21 @@ void GpuAgent::InitDma() {
...
   };
 
+  // Enable profiling on the internal blit copy queues right after creation so that
+  // copy completion timestamps are always available regardless of when async-copy
+  // profiling is later turned on (e.g. a profiler attaching after copies have run).
   // Dedicated compute queue for host-to-device blits.
-  queues_[QueueBlitOnly].reset(queue_lambda);
+  queues_[QueueBlitOnly].reset([queue_lambda]() {
+    auto queue = queue_lambda();
+    queue->SetProfiling(true);
```
cfreeamd, line 917: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** This enables *queue-level* profiling, but the agent-level `profiling_enabled_` flag stays false at init (only flipped by `hsa_amd_profiling_async_copy_enable(true)`). The DMA copy paths set `out_signal.async_copy_agent(...)` — the tag `TranslateTime` uses for system-domain translation — only `if (profiling_enabled())` (agent-level), and `BlitKernel::SubmitLinearCopyCommand` similarly gates `LogSignalDuration` on `agent_->profiling_enabled()`.

For a copy that runs *before* a profiler enables async-copy profiling, the timestamp is captured but the signal wasn't tagged with `async_copy_agent` at submit time. Does `hsa_amd_profiling_get_async_copy_time` then return a correctly-translated value for that earlier copy, or is the agent-level tag required? If it's required, enabling only queue-level profiling here may not fully achieve the "timestamps always available regardless of when async-copy profiling is later turned on" goal — the agent-level flag / tagging path would also need to be on. Could you confirm the end-to-end path?

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -900,10 +900,21 @@ void GpuAgent::InitDma() {
...
+    return queue;
+  });
   // Share utility queue with device-to-host blits.
-  queues_[QueueUtility].reset(queue_lambda);
+  queues_[QueueUtility].reset([queue_lambda]() {
```
cfreeamd, line 921: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** `QueueUtility` is shared with device-to-host blits and also carries utility PM4 (cache invalidation, `ExecutePM4`). Enabling profiling on it unconditionally adds timestamp-capture cost to all of its dispatches even when no profiler is attached. Presumably negligible, but worth confirming the overhead is acceptable for the always-on internal path. Non-blocking.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -900,10 +900,21 @@ void GpuAgent::InitDma() {
...
+    auto queue = queue_lambda();
+    queue->SetProfiling(true);
```
cfreeamd, line 923: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** `hsa_amd_profiling_async_copy_enable(false)` → `GpuAgent::profiling_enabled(false)` → `EnableDmaProfiling(false)` → `blit->EnableProfiling(false)` → `queue_->SetProfiling(false)` on these same queues. So the init-time always-on state is undone after the first disable cycle, and the "always available" guarantee no longer holds. If the intent is truly always-on for the internal queues, should the disable path skip them (or should the agent default to profiling-enabled instead)?

cfreeamd (commented):

Reviewed the ROCr change in `amd_gpu_agent.cpp` (`GpuAgent::InitDma`). The lambda capture and lazy-init mechanics look correct (queue_lambda captured by value, return type unchanged, profiling applied on first materialization). Left a couple of questions inline about how queue-level profiling interacts with the agent-level `profiling_enabled_` flag and the `EnableDmaProfiling(false)` path.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
dayatsin-amd: That is ok

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
dayatsin-amd: That is ok

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
dayatsin-amd: I updated the disable profiling path to make sure we unmap/remap the queues so that CP can pick up the latest amd_queue flags.

But on initialization, we want to avoid the queue unmap/remap so we still initialize the queues with profiling enabled.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
```diff
@@ -1571,8 +1571,23 @@ hsa_status_t AqlQueue::GetCUMasking(uint32_t num_cu_mask_count, uint32_t* cu_mas
...
+
+  if (cur != enabled && LoadWriteIndexRelaxed() != 0) {
+    // If the queue is already enabled/disabled, we need to update the queue properties and we have already submitted packets,
+    // then we need to unmap and remap the queue for CP FW to re-read the queue properties.
+    need_update = true;
+  }
+
   Queue::SetProfiling(enabled);
 
+  if (need_update) {
```
cfreeamd, line 1600: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** The re-latch here issues two `UpdateQueue` ioctls (`Suspend()`→pct 0, `Resume()`→pct 100) to make CP FW re-read the profiling flag. This only fires on the toggle path (init skips it since `LoadWriteIndexRelaxed() == 0`), so two things worth confirming: (1) is the profiling toggle serialized against DMA submission on this internal blit queue, or could another thread be pushing copies while the queue is briefly unmapped? (2) `Resume()` unconditionally reactivates whenever `suspended_` is set — if the queue happened to already be suspended by another actor (the coredump path is the only other `Suspend()` user) when `SetProfiling` toggles, this would resume a queue that was meant to stay down. Both look practically unreachable for the internal blit queues, but confirming the assumption would be good. Non-blocking.

cfreeamd (commented):

Thanks for the responses and the `623e5a93` remap fix. The three prior items are resolved: the queue-level vs agent-level translation question and the shared-utility-queue overhead were confirmed OK, and the disable/re-enable path is now handled by re-latching via `Suspend()/Resume()` in `AqlQueue::SetProfiling`. The change detection reads the current ENABLE_PROFILING bit before the base call and only remaps on a real state change with packets already submitted, so init stays remap-free — matching the intent. One non-blocking follow-up question left inline on the new remap block.

cfreeamd (approved):

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
dayatsin-amd: Another thread could be pushing copies to that queue while the queue is unmapped and re-mapped. Our architecture can handle this and on queue re-map, CP FW will check whether there are new packets in the queue and process them.

PR #8794, opened 2026-07-17
feat(hotswap): disk-persistent tier on the single-flight retarget cache

suryajasper:
## Summary

Convergence of the in-flight hotswap retarget-cache efforts: this **stacks on
@harsh-amd's single-flight cache (#8744)** and layers on the **cross-process disk
tier** from our #8741, so we get both properties in one design:

- **Single-flight + exact-content keys** (from #8744): concurrent cold loads of the
  same code object coalesce onto one COMGR run; the key hashes to a bucket then does
  an exact byte compare, so hash collisions cannot alias.
- **Cross-process / cross-run persistence** (this PR): a retarget survives process
  exit and is reused on the next launch.

The disk read/write is invoked **inside the single-flight producer**, so the leader
performs the one disk read and the coalesced waiters share its result — disk I/O is
single-flighted too, and runs outside all cache mutexes.

## Relationship to other PRs

- **Stacked on #8744** (@harsh-amd). This branch is based on that branch; its commits
  appear here until #8744 merges, after which this rebases onto develop.
- **Supersedes our #8737 / #8741 / #8659** for the combined design: #8744's in-memory
  single-flight cache replaces our `ElfBufferRef` byte-budget-LRU in-memory tier, and
  this PR keeps only our disk tier on top. #8744 also directly answers the two loudest
  review criticisms of our PRs (no single-flight; 64-bit key collision).

## Technical details

The only new commit adds, on top of #8744:

- **Disk tier** (write-once/read-many, immutable, salt-namespaced, atomic
  temp+`fsync`+`rename`, header-validated on read; no size cap — users clear the dir).
  POSIX only. Controls: `HSA_HOTSWAP_CACHE_DIR`, `HSA_HOTSWAP_DISK_CACHE=0`.
- **Producer bridge**: on a disk hit the producer builds a `RetargetedElf` from the
  persisted bytes **carrying the source snapshot** (required by #8744's
  snapshot-identity check), returning without COMGR; on a cold COMGR success it
  enqueues an async `DiskWriteTask` whose payload is an independent copy.
- **Background writer** owned by the ROCr Runtime lifecycle
  (`HotswapCacheStartup`/`Shutdown` from `Runtime::Load`/`Unload`); Harsh's cache adds
  no runtime wiring, so this is conflict-free.
- **Disk salt** derived from the loaded COMGR library identity (dladdr-resolved path +
  size + mtime + format version), captured in `GetComgrApi`.

The disk primitives are the same code as #8741 (unchanged apart from the buffer-type
naming). The in-memory cache is #8744's, unmodified.

## Controls

| Env var | Effect |
|---------|--------|
| `HSA_HOTSWAP_CACHE_DIR` | Disk cache location (default `$XDG_CACHE_HOME/rocm/hotswap` or `$HOME/.cache/rocm/hotswap`) |
| `HSA_HOTSWAP_DISK_CACHE=0` | Disable the disk tier |

## Test plan

- [x] New `RetargetDiskCache.DiskHitIsSingleFlighted`: a concurrent burst does exactly
  **one** disk read, `producer_calls==1`, `coalesced_results==N-1`, and all results
  share one buffer. Passed 10/10 repeat runs (not flaky).
- [x] Ported `RetargetDiskCache.*` disk unit tests (round-trip, salt-mismatch,
  missing-entry, distinct-key isolation, idempotent overwrite, async drain-at-shutdown).
- [x] Builds clean on MI300X; `hotswap_rewrite` 39/39 and `hotswap_cache` 13/13
  (#8744's single-flight suite unaffected) pass.

Draft: pending #8744 landing and on-device gfx1250 A0 validation.


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -548,120 +943,157 @@ RetargetCodeObjectResult TryRetargetCodeObject(const CodeObjectView& code_object
...
-      // A deterministic COMGR failure is cached. If the rewrite was required,
-      // fail loudly rather than falling back to the unpatched original.
-      if (decision->rewrite_required) {
-        return {RetargetCodeObjectStatus::kRequiredRewriteFailed, true};
+    // Disk hit: build a RetargetedElf from the persisted bytes, WITHOUT COMGR.
+    // It must carry `source_snapshot` so GetOrCompute's snapshot-identity check
+    // accepts it.
+    if (disk_ok) {
+      std::shared_ptr<std::vector<uint8_t>> disk_bytes =
+          ReadDiskCache(DiskCachePath(disk_dir, disk_key, salt), salt);
```

**[Important]** The disk tier trusts a 64-bit FNV key (`ComputeRetargetCacheKey`, encoded in the filename) as the *sole* identity for a hit. `ReadDiskCache` validates only magic/version/`comgr_salt`/`payload_size` — it never re-verifies the payload against the source bytes. This breaks the invariant the in-memory tier deliberately upholds ("neither correctness nor isolation depends on hash strength" — it does an exact `memcmp` via `SourceSnapshot::Equals`). Worse, on a disk hit the producer wraps the disk bytes in a `RetargetedElf` carrying the *current* `source_snapshot`, so `GetOrCompute`'s `elf->source().get() == flight->source.get()` identity guard accepts it unconditionally — the in-memory exact-comparison safety net cannot catch a disk collision. A 64-bit key collision (or a stale entry the salt didn't invalidate) would silently load the wrong retargeted GPU code. Recommend embedding source size + a strong content digest in `DiskCacheHeader` and verifying on read. (Copilot flagged the same collision independently — agreement strengthens it.)

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -389,8 +368,417 @@ std::optional<RewriteDecision> DecideHotswapRewrite(
...
+  }
+  std::string value = os::GetEnvVar("HSA_HOTSWAP_DISK_CACHE");
+  std::transform(value.begin(), value.end(), value.begin(),
+                 [](unsigned char c) { return std::tolower(c); });
+  return value == "0" || value == "off" || value == "false" || value == "no" ||
+         value == "n" || value == "f";
+}
+
+// Resolves the cache root directory, or "" if none is usable.
+std::string GetDiskCacheDir() {
```
cfreeamd, line 600: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** Trust-model question on the disk tier: the cache root is `$HSA_HOTSWAP_CACHE_DIR` (unsanitized override), else `$XDG_CACHE_HOME/rocm/hotswap`, else `$HOME/.cache/rocm/hotswap`, and a hit is accepted after checking only `comgr_salt` — which is derived from *public, non-secret* toolchain metadata (`stat` size/mtime/path of the COMGR lib). Any principal that can write into the cache dir (a world-writable `HSA_HOTSWAP_CACHE_DIR`, a shared `XDG_CACHE_HOME`, a multi-user tmp misconfig) can plant a crafted `<key>.co` with a correct, derivable salt and get arbitrary retargeted code loaded onto the GPU. In the default per-user `$HOME/.cache` this stays within the user's own trust boundary, but the env override + non-secret salt make the boundary worth an explicit decision. `CACHE_EVIDENCE.md` itself says a disk cache "must independently define its trust model" — this is that cache. Suggest documenting the intended boundary and rejecting group/world-writable cache dirs (refuse if `st_mode & (S_IWGRP|S_IWOTH)` or not owned by the euid). Flagging for gating-reviewer sign-off.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -389,8 +368,417 @@ std::optional<RewriteDecision> DecideHotswapRewrite(
...
+
+  std::mutex mutex_;
+  std::condition_variable cv_;
+  std::list<DiskWriteTask> queue_;
+  std::thread thread_;
+  bool running_ = false;
+  bool stopping_ = false;
+};
+
+DiskWriter& GetDiskWriter() {
```

**[Important]** `GetDiskWriter()` returns a function-local `static DiskWriter writer;`. `hsa-runtime/CMakeLists.txt:180` compiles the runtime with `-fno-threadsafe-statics`, so first-time initialization of this static has no compiler guard and concurrent first access is UB. It's *currently masked* because `HotswapCacheStartup()` (Runtime::Load, single-threaded) forces the first call before concurrent loads — but only when the cache is enabled, and the masking is undocumented/fragile. Note `GetProcessRetargetCache()` in `hotswap_cache.cpp` deliberately uses `std::call_once` for exactly this reason; the two singletons should be consistent. Recommend wrapping the `DiskWriter` singleton in `std::call_once` so it doesn't depend on Load()-ordering. (Copilot flagged this independently.)

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -389,8 +368,417 @@ std::optional<RewriteDecision> DecideHotswapRewrite(
...
+  return true;
+}
+
+std::string ToHex(uint64_t v) {
+  char buf[17];
+  std::snprintf(buf, sizeof(buf), "%016llx", static_cast<unsigned long long>(v));
+  return std::string(buf);
+}
+
+std::string DiskCacheSubdir(const std::string& dir, uint64_t salt) {
```

**[Optional]** The `comgr_salt` is used as a *subdirectory name* (`DiskCacheSubdir`), and it changes on every COMGR lib size/mtime/path change (i.e. every ROCm update). Old `<salt>/` subtrees are never swept, so the persistent cache accumulates a full generation of `.co` artifacts per toolchain version indefinitely, with no LRU/TTL/size cap (`CACHE_EVIDENCE.md` acknowledges no eviction). For a persistent on-disk cache this is an operational disk leak. Consider a best-effort sweep of stale salt subdirs on startup, or documenting a cleanup mechanism / size bound.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap_cache.cpp
```diff
@@ -0,0 +1,747 @@
+////////////////////////////////////////////////////////////////////////////////
+//
+// The University of Illinois/NCSA
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** New file uses the legacy NCSA license header. New files should carry SPDX:
```cpp
//
```
Non-blocking.

cfreeamd (commented):

Reviewed the ROCr changes (hotswap two-tier retarget cache). The in-memory single-flight cache (`hotswap_cache.cpp` `ContentRetargetCache::GetOrCompute`), the `DiskWriter` drain/shutdown, `WriteDiskCache`'s atomic temp+fsync+rename publish, and the `CodeObjectMemoryOwner` lifetime rework all look correct. Inline comments focus on the new disk-persistent tier: payload identity/verification, the on-disk trust model, and a function-local static under `-fno-threadsafe-statics`. Two of these overlap with Copilot's independent findings.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -387,8 +366,417 @@ std::optional<RewriteDecision> DecideHotswapRewrite(
...
+  std::list<DiskWriteTask> queue_;
+  std::thread thread_;
+  bool running_ = false;
+  bool stopping_ = false;
+};
+
+DiskWriter& GetDiskWriter() {
+  static DiskWriter writer;
+  return writer;
+}
```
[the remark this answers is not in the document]
suryajasper: Fixed in d5ea4d6b23. `GetDiskWriter()` now uses a `std::call_once` + `unique_ptr<DiskWriter>` singleton at namespace scope, mirroring `GetProcessRetargetCache()` in `hotswap_cache.cpp` so both process singletons are initialized consistently and don't depend on `Runtime::Load()` ordering under `-fno-threadsafe-statics`.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
suryajasper: Fixed in d5ea4d6b23 — wrapped the `DiskWriter` singleton in `std::call_once` + `unique_ptr`, matching the `GetProcessRetargetCache()` pattern you referenced so the two singletons are consistent and the init no longer relies on the (fragile, undocumented) Load()-ordering mask. Good catch on the consistency point.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap_cache.cpp
suryajasper: Fixed in d5ea4d6b23 — switched to the SPDX MIT header matching the recently-added siblings in `core/runtime/` (e.g. `amd_sdma_queue.cpp`).

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
suryajasper: Thanks — we looked at this closely and are choosing not to add on-read content verification here. Reasoning:

**The in-memory tier's guarantee doesn't transfer cheaply to disk.** Its collision-immunity comes from a full `memcmp` (`SourceSnapshot::Equals`) against source bytes that are *already resident*. Replicating that on disk means storing the entire source blob alongside each output and re-reading it on every hit to compare — for RCCL that's ~578MB read back on the fast path, which defeats the point of the disk tier. So a true exact-compare is disproportionate.

**A wider digest would only be a wider hash, not the exact-compare invariant.** Adding a `source_digest` to the header is really just turning the 64-bit identity into ~128-bit. It raises the bar but is still probabilistic — it does *not* give the "correctness independent of hash strength" property, so it wouldn't actually satisfy the stated concern; it'd just move the exponent.

**The residual risk is negligible for this cache.** The key is 64-bit over (source bytes + ISA + flags). Birthday bound for a realistic per-user cache of even ~10⁴–10⁵ distinct objects puts collision probability around 10⁻¹¹–10⁻⁹ — well below ambient hardware error rates. This is a per-user, non-adversarial cache (default `$HOME/.cache`), so there's no adversary manufacturing collisions (see the trust-model thread).

Net: the cost of a real fix (full source re-read) is high, the cheap fix (wider hash) doesn't deliver the invariant, and the risk it addresses is negligible. Documenting the 64-bit-key limitation in `CACHE_EVIDENCE.md` rather than adding verification. Happy to revisit if a concrete collision or a shared/multi-writer deployment model comes up.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -546,120 +941,157 @@ RetargetCodeObjectResult TryRetargetCodeObject(const CodeObjectView& code_object
...
+    const size_t source_size = source_snapshot ? source_snapshot->size() : code_object.size;
+
+    // This producer runs only on the single-flight leader, outside all cache
+    // mutexes, so the disk read/write below is coalesced across waiters and
+    // never serializes the in-memory cache.
+#if HOTSWAP_DISK_CACHE_SUPPORTED
+    const uint64_t disk_key = ComputeRetargetCacheKey(
+        source_data, source_size, decision->source_isa, decision->target_isa,
+        decision->request_entry_trampolines, decision->request_strict_mode);
+    std::string disk_dir;
```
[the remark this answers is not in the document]
suryajasper: Not adding on-read content verification — see the detailed reply on @cfreeamd's parallel comment. Short version: a true exact-compare (the in-memory tier's actual guarantee) would require storing + re-reading the full source blob per hit (~578MB for RCCL), and a lighter `source_digest` would just widen the 64-bit hash to ~128-bit without delivering the collision-immunity invariant. For a per-user, non-adversarial cache the 64-bit-key collision probability is ~10⁻⁹ or below at realistic cache sizes, so we're documenting the limitation rather than paying for verification.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
suryajasper: This is intentional and we're leaving it as-is. The cache is per-user by design — default root is `$HOME/.cache/rocm/hotswap`, inside the user's own trust boundary. A user writing entries into their own cache directory isn't a threat we're defending against (they can already run whatever code they like in their own process). We're deliberately not obscuring the cache from its owner.

The env override (`HSA_HOTSWAP_CACHE_DIR`) is opt-in; if someone points it at a world-writable/shared location they've stepped outside the default per-user model by choice, same as any cache-dir override. Rejecting group/world-writable dirs would break legitimate shared-cache setups, so we'd rather not impose that policy. Will document the intended boundary (per-user, non-adversarial) in `CACHE_EVIDENCE.md` alongside the key-strength note.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
suryajasper: Acknowledged as a known limitation — not addressing it in this PR. `CACHE_EVIDENCE.md` already documents that the cache has no eviction, and this matches the other disk-cache PRs landed over the past week (none added LRU/TTL/sweep). The per-salt accumulation is bounded in practice: a new salt subtree only appears on a COMGR lib change (ROCm update), and each holds only the retargeted objects actually loaded on that toolchain. If disk footprint becomes an operational concern we can add a best-effort stale-salt sweep as a follow-up, but it's out of scope here.

cfreeamd:

Re-reviewed after commits d5ea4d6b (thread-safe singleton + SPDX) and 6e9f9c60 (trust-model/key-strength/eviction docs). All prior [Important] items are now resolved or addressed with a documented, reasoned decision:

- **[Important] Thread-unsafe `GetDiskWriter()` singleton** — Fixed. Now a namespace-scope `std::unique_ptr<DiskWriter>` guarded by `std::call_once` (`g_disk_writer_once`), mirroring `GetProcessRetargetCache()`, with a comment noting `-fno-threadsafe-statics`. Correct and consistent.
- **[Important] 64-bit disk key as sole identity** — Rebuttal accepted. The point that a true exact-compare would require re-reading the full source blob per hit (hundreds of MB), and that a wider digest only shifts the collision exponent without delivering the in-memory tier's exact-compare invariant, is sound; the ~10⁻⁹ birthday bound for a per-user non-adversarial cache is below ambient hardware error rates. Documented in `CACHE_EVIDENCE.md`. (Optional, non-blocking: storing `source_size` in `DiskCacheHeader` and comparing it on read is nearly free and would reject the common different-sized-source collision — worth considering as cheap insurance, but not required.)
- **[Important] Disk-cache trust model** — Rebuttal accepted as a documented decision: per-user, non-adversarial boundary (`$HOME/.cache`), `HSA_HOTSWAP_CACHE_DIR` opt-in override is the user's own responsibility. This is a security-relevant surface (executable GPU code loaded from disk); the decision is now explicit and documented per the `CACHE_EVIDENCE.md` requirement, so this is the gating reviewers' (kentrussell/dayatsin-amd) explicit sign-off to give rather than a code defect.
- **[Optional] Unbounded disk growth / no eviction** — Acknowledged and documented as known limitation with a possible follow-up sweep. Fine.
- **[Nit] SPDX header** — Fixed on `hotswap_cache.cpp`.

No remaining code-level concerns from the AI pre-review. The single-flight design, lifetime rework, and disk atomic-publish path remain correct.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -389,8 +368,425 @@ std::optional<RewriteDecision> DecideHotswapRewrite(
...
+    running_ = false;
+    stopping_ = false;  // ready for a clean restart on a later Start()
+  }
+
+  // Enqueues a write. If the writer isn't running (creation failed, not
+  // started, or already stopping), silently drops the task — persistence is
+  // best-effort.
+  void Enqueue(DiskWriteTask task) {
+    std::scoped_lock lock(mutex_);
+    if (!running_ || stopping_) {
```
shwetagkhatri: Recommendation: Is this code path reached when we are OOM? In that case, we could add a metric for dropped disk write tasks.

dayatsin-amd (approved):

shwetagkhatri (approved):
Recommendation: We could add HSA_HOTSWAP_DISK_CACHE_MAX_SIZE in future. 
As we could hit unbounded disk usage in very long running running systems with many toolchain versions

Adding some recommendations here. Otherwise LGTM.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
suryajasper: The no-eviction behavior is a known limitation (documented in CACHE_EVIDENCE.md — the disk tier accumulates one generation of artifacts per toolchain salt with no LRU/TTL/size cap), and the OOM path at line 710 is the best-effort drop where a push_back fails under memory pressure — currently only logged under HSA_HOTSWAP_VERBOSE. A dropped write is just a future cache miss, never a correctness issue, so rather than expand this PR we'll open a follow-up to add HSA_HOTSWAP_DISK_CACHE_MAX_SIZE (with a best-effort sweep of stale salt subtrees) alongside a dropped_disk_writes metric.

atgutier (commented):
My immediate high-level questions are:

- Why are we introducing new env vars instead of only relying on solely XDG standard (I see they are used as fallbacks) and fallbacks?

- Why is the caching layer being pushed into the platform runtime and not the translation layer?

Overall, I think much of the design and additions into the platform and language runtimes needs to be rethought. I think a single hook into the translation layer would suffice for this sort of work.

projects/rocr-runtime/rocrtst/suites/functional/hotswap/CACHE_EVIDENCE.md
```diff
@@ -0,0 +1,303 @@
+# HotSwap content cache evidence
```
stellaraccident: This AI document is unlikely to stay current as the codebase evolves. Please find another place for this kind of task-specific documentation, rewrite/simplify it or drop it.

stellaraccident (commented):
Please coordinate with Tony Gutierrez on constraints for getting this landed. I'm becoming concerned by the amount of non-local changes being made across the runtimes (rocr/clr) that are hard coding concepts that I thought were to be handled by the plugin.

atgutier (commented):
I worked with an agent on a detailed review, along with my own planning on how to achieve this with minimal intrusion into ROCr, and here are our initial findings with immediately actionable items and findings that can be marked as todos. Once the immediately actionable findings are addressed, I will approve conditioned on moving this feature into comgr in a near-term future refactor:

• The in-process single-flight cache and generic loader ownership work are reasonable deadline-compatible foundations, but the new default-on disk tier has two merge-blocking correctness/lifecycle problems and several contained fixes that should land now.

  Reviewed: ROCm/rocm-systems#8794, head ec73d9e69f0b33d53d2b1e544979cb055062d98f

  ## Blockers

  ### 1. The disk-cache salt can identify a different COMGR than the one
  that generated the cached code

  Evidence:

  - hotswap.cpp:248-297 loads COMGR once and retains function pointers and
    its pathname.

  - hotswap.cpp:466-485 later recomputes the salt by calling stat() on that
    pathname.

  - hotswap.cpp:978-981 obtains that salt for cache operations.

  If ROCm is upgraded while a process is running, the mapped COMGR remains
  the old inode while the pathname can refer to the replacement. Old COMGR
  output can then be written under the new COMGR salt and consumed by a
  later process actually using the new COMGR.

  Path, file size, and second-resolution modification time also do not fully
  identify translation semantics.

  Ask for now:

  - Preferably add a small COMGR API returning an opaque, stable translator/
    cache-schema fingerprint covering the rewrite implementation and
    semantic options.

  - Capture it once with the loaded COMGR API, rather than recomputing
    identity from its path.

 - If that API cannot meet the deadline, identify the mapped image using
    its ELF build ID or equivalent immutable identity and make the disk tier
    opt-in/default-off until the authoritative COMGR fingerprint exists.

  This is the most valuable low-disruption piece to move toward COMGR
  immediately.

  ### 2. Final hsa_shut_down can synchronously drain arbitrary disk work
  while holding ROCr’s bootstrap lock

  Evidence:

  - runtime.cpp:149-175 holds bootstrap_lock() through Runtime::Unload().
  - runtime.cpp:2681-2684 invokes HotswapCacheShutdown() during unload.
  - hotswap.cpp:686-703 stops and joins the writer.
  - hotswap.cpp:723-745 drains every queued task before exiting.
  - hotswap.cpp:571-632 performs fflush, fsync, and rename for each task.

  The queue has no bound. A process can therefore enter final shutdown with
  many large pending writes and block all serialized ROCr initialization/
  shutdown while waiting on slow, full, remote, FUSE-backed, or otherwise
  stalled storage.

  Ask for now:

  - Bound queued tasks and total queued bytes.
  - Treat persistence as best effort: discard queued-but-not-started writes
    during shutdown or otherwise establish a genuinely bounded shutdown
    contract.

  - Do not require a complete disk backlog drain while holding the process-
    global bootstrap lock.

  Moving the complete writer into COMGR immediately would be disruptive;
  bounding it is the deadline-safe correction.

  ## Major findings

  ### 3. The writer queue retains unbounded copies of translated ELFs

  Evidence:

 - hotswap.cpp:649-654 gives each queued task ownership of a complete
    payload.

  - hotswap.cpp:705-719 uses an unbounded task list.
  - hotswap.cpp:1018-1028 copies every cold translated result into another
    vector for persistence.

  - CACHE_EVIDENCE.md:127-139 omits these pending copies from its memory
    bound.

  A workload producing objects faster than disk fsync can consume arbitrary
  additional resident memory. This matters especially because the PR
  discusses code objects hundreds of megabytes in size.

  Ask for now:

  - Reserve queue capacity before making the payload copy.
  - Apply both byte and task limits.
  - Drop persistence when the limit is reached.
  - Add queued, peak, and dropped-byte metrics.
  - Correct the memory claim in CACHE_EVIDENCE.md.

  ### 4. Persistent lookup identity and payload integrity are probabilistic

  Evidence:

  - hotswap.cpp:374-395 reduces the source and options to a 64-bit FNV key
    used as the filename.

  - hotswap.cpp:408-421 does not store a full source/transform digest in the
    header.

  - hotswap.cpp:524-566 validates header fields and payload length, but not
    an output checksum.

  - By contrast, the in-memory cache treats its hash as a bucket and
    performs an exact comparison in hotswap_cache.cpp:469-478.

  A 64-bit collision can return unrelated GPU executable code. Same-length
  corruption can also pass the disk read checks. The per-user, non-
  adversarial assumption lowers the likelihood but does not remove the
  wrong-code failure mode.

  Ask for now:

 - Use a collision-resistant digest over a canonical tuple containing
    source bytes, source and target ISA, all semantic rewrite options,
    translator/schema identity, and output-provenance version.

  - Store and validate that digest in the header.
  - Add an output integrity digest.

  This does not require rereading the source from disk; the implementation
  already scans it to calculate the current key.

  ### 5. The disk-writer thread starts globally, even when no translation
  will be persisted

  Evidence:

  - runtime.cpp:2674 calls cache startup during every Runtime::Load().
  - hotswap.cpp:672-683 creates the writer thread.
  - No gfx1250 eligibility check or first-write condition precedes startup.
  - HSA_HOTSWAP_DISABLE=1 does not provide a complete global lifecycle
    bypass.

  Every ordinary ROCr process can acquire a disk-cache thread despite never
  using hotswap. This unnecessarily embeds COMGR-specific cache lifecycle
  into the platform runtime.

  Ask for now:

  - Start the writer lazily on the first accepted persistence task.
  - Make start-and-enqueue atomic under the writer state lock.
  - Keep shutdown idempotent and no-op if the writer was never constructed.
  - At minimum, honor the global hotswap-disable state.

  ### 6. Disabled mode can construct the writer during shutdown

  Evidence:

  - hotswap.cpp:774-780 skips writer construction during startup when the
    cache is disabled.

  - hotswap.cpp:783-787 nevertheless calls GetDiskWriter().Stop()
    unconditionally.

 - hotswap.cpp:761-766 allocates the singleton from GetDiskWriter().

  Thus, a configuration documented as disabled can allocate the writer
  object during teardown. Allocation failure can also complicate an already
  partially committed final shutdown.

  Ask for now:

  - Add a no-throw “get if constructed” path.
  - Make startup and shutdown symmetric.
  - Verify disabled mode creates neither a thread nor writer state.

  ### 7. The claimed disk-hit single-flight test does not exercise the
  production path

  hotswap_rewrite_test.cc:968-1043 manually constructs a single-flight
  producer and calls test disk helpers with caller-selected keys and salts.
  It does not test:

  - Production key calculation.
  - Real COMGR identity/salt generation.
  - Environment selection.
  - TryRetargetCodeObject.
  - A cold production translation followed by a later disk hit.
  - Runtime startup and shutdown.

  Ask for now:

  - Add one injectable production-path test that performs a cold write,
    releases in-memory payload ownership, then proves a subsequent request
    obtains the result from disk without running COMGR.

  - Add tests for queue saturation, disabled/no-use startup, shutdown with a
    backlog, and mapped-COMGR/path replacement.

  ## What should move toward COMGR now

The most useful immediate boundary improvements
  are:

  1. Add a COMGR-owned opaque translator/cache-schema fingerprint. ROCr
     cannot authoritatively infer translation identity from library
     filesystem metadata.

  2. Put disk lookup/store behind one internal adapter whose compilation and
     naming explicitly identify it as COMGR-mode functionality.

  3. Keep all translation-specific options behind that provider boundary.
     Avoid spreading persistent-cache concepts into readers, executable
     objects, or additional ROCr lifecycle surfaces.

  4. Preserve CodeObjectMemoryOwner. It is generic loader plumbing that both
     current COMGR rewriting and later rocjitsu-owned translated storage can
     use.

  5. Avoid creating a public HSA cache ABI. The current internal interfaces
     leave room for the planned ownership change.

  If the first item is too much for this deadline, the safe compromise is to
  keep the in-process cache enabled and ship the persistent tier as opt-in
  until the identity contract lands.

  ## Appropriate tracked TODOs

  These are architectural cleanups:

  - Move persistent cache key canonicalization, format, schema version,
    translator identity, storage policy, eviction, writer thread, and writer
    shutdown into COMGR.

  - Later move COMGR’s in-memory single-flight cache into COMGR once it can
    return shared immutable output ownership without copying for each ROCr
    load.

  - Reduce ROCr to generic input viewing, translated-byte lifetime
    ownership, and loading.

  - Make loader failure rollback transactional. The new owner is attached
    before all segment, symbol, and relocation work succeeds, so an already-
    existing partial-load retention behavior can now pin much larger
    rewritten buffers.

  ## Bottom line

  The deadline-friendly path is:

  - Keep the generic memory-owner seam.
  - Keep the exact-content in-process single-flight cache.
  - Fix the COMGR identity, queue bound, shutdown behavior, lazy startup,
    and persistent integrity now.

  - If stable COMGR identity cannot land promptly, make disk persistence
    opt-in.

harsh-amd:
Implemented the requested disk-cache hardening in one commit:
`a609805ede0602c7d9dc946c44bcb1628ca31054`.

- COMGR identity now comes from the optional, dynamically resolved
  `amd_comgr_get_cache_identifier` proposed in
  [ROCm/llvm-project#3551](https://github.com/ROCm/llvm-project/pull/3551).
  ROCr captures it once from the loaded COMGR. With an older COMGR, HotSwap
  and the in-process cache continue to work while persistence is skipped.
- Disk persistence remains default-on with no required HotSwap environment
  variable. The custom path variable is removed; storage follows
  `$XDG_CACHE_HOME/rocm/hotswap`, then `$HOME/.cache/rocm/hotswap`.
  `HSA_HOTSWAP_DISK_CACHE=0` remains only as a kill switch.
- Persistent identity is BLAKE2b-256 over the canonical source/ISA/options/
  schema/COMGR tuple. The header stores and verifies the input identity,
  COMGR identity, exact size, and a separate output integrity digest.
- The writer starts lazily, aliases the existing immutable output instead of
  copying it, and caps outstanding work at 16 tasks / 1 GiB including the
  active write. Admission drops and filesystem failures have task/byte
  metrics.
- Shutdown drops queued work while under the ROCr bootstrap lock, then joins
  at most the active write after releasing that lock. The amended commit also
  serializes concurrent shutdown waiters found during self-review.
- Added production cold-write/fresh-cache disk-hit coverage, forced-COMGR-
  failure proof, independent task/byte saturation tests, corruption rejection,
  default/disable and no-use lifecycle tests, concurrent shutdown coverage,
  and known BLAKE2 vectors.

Validation on `mi400-2`:

- final `hsa-runtime64` build: pass
- focused `hotswap` CTests: 3/3 pass
- TSAN: cache 13/13 and rewrite/disk 46/46 pass
- TLC: 510,325 states generated, 121,729 distinct, complete depth 18, no
  invariant violation
- installed old COMGR with the new symbol confirmed absent: production-path
  compatibility test passes
- five interleaved baseline/updated benchmark pairs: `single-flight+disk`
  median workload ratio 1.0012 cold (+0.12%, noise) and 0.9938 warm
  (0.62% faster), so no material regression

`CACHE_EVIDENCE.md` was moved to the requested Confluence child page:
[HotSwap Cache Evidence](https://amd.atlassian.net/wiki/spaces/MLSE/pages/1813539100/HotSwap+Cache+Evidence).

@atgutier, please re-review when the new CI run is ready.

atgutier (changes requested):
Please make this a standalone tools lib.

PR #8798, opened 2026-07-18
feat(rocjitsu): observe VGPR instruction writes

newling:
## Summary

Add an execution-plugin hook for instruction-visible VGPR writes and route write observation through `RegisterAccess`, mirroring the existing VGPR read hook.

This is the plugin-plumbing portion split from #7890. It does not add or change race-detector behavior.

The hook reports physical register, lane mask, and byte mask. Packed 16-bit operands report the selected source half, the preserved destination half read during merging, and the written destination half separately. Memory-pipeline completion writes remain raw VM storage operations and do not fire the instruction-write hook.

## Testing

- Built `rocjitsu_tests`
- `RegisterAccessTest` and `ExecutionPluginTest`: 19 passed, 1 skipped
- AMD ISA generator profile tests: 102 passed, 1 skipped
- Pre-commit checks passed

### DIFF
--- a/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
+++ b/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
@@ -7261,17 +7261,25 @@ class CodeGenerator:
                             _dpp_old_dst_type = (
                                 'uint64_t' if _uses_64bit_dpp_dst else 'uint32_t'
                             )
-                            _dpp_dst_read_method = (
-                                'read_lane64' if _uses_64bit_dpp_dst else 'read_lane'
+                            _dpp_dst_storage_read_method = (
+                                'read_operand_storage64'
+                                if _uses_64bit_dpp_dst
+                                else 'read_operand_storage'
                             )
-                            _dpp_dst_write_method = (
-                                'write_lane64' if _uses_64bit_dpp_dst else 'write_lane'
+                            _dpp_dst_storage_write_method = (
+                                'write_operand_storage64'
+                                if _uses_64bit_dpp_dst
+                                else 'write_operand_storage'
                             )
-                            _dpp_physical_dst_read_method = (
-                                'read_vgpr64' if _uses_64bit_dpp_dst else 'read_vgpr'
+                            _dpp_physical_dst_storage_read_method = (
+                                'read_vgpr_storage64'
+                                if _uses_64bit_dpp_dst
+                                else 'read_vgpr_storage'
                             )
-                            _dpp_physical_dst_write_method = (
-                                'write_vgpr64' if _uses_64bit_dpp_dst else 'write_vgpr'
+                            _dpp_physical_dst_storage_write_method = (
+                                'write_vgpr_storage64'
+                                if _uses_64bit_dpp_dst
+                                else 'write_vgpr_storage'
                             )
                             _dpp_preamble = ''
 
@@ -7326,14 +7334,15 @@ class CodeGenerator:
                                 if _has_sdwa_encoding:
                                     _old_dst_read = (
                                         f'inst_.src0 == amdgpu::SRC_DPP\n'
-                                        f'            ? amdgpu::RegisterAccess(wf).{_dpp_dst_read_method}({_dst_name}, ln)\n'
-                                        f'            : amdgpu::RegisterAccess(wf.cu()).{_dpp_physical_dst_read_method}(vb + {_dst_reg_expr}, ln)'
+                                        f'            ? amdgpu::RegisterAccess(wf).{_dpp_dst_storage_read_method}({_dst_name}, ln)\n'
+                                        f'            : amdgpu::RegisterAccess(wf.cu()).{_dpp_physical_dst_storage_read_method}(vb + {_dst_reg_expr}, ln)'
                                         if _uses_logical_dpp_dst
-                                        else f'amdgpu::RegisterAccess(wf.cu()).{_dpp_physical_dst_read_method}(vb + {_dst_reg_expr}, ln)'
+                                        else f'amdgpu::RegisterAccess(wf.cu()).{_dpp_physical_dst_storage_read_method}(vb + {_dst_reg_expr}, ln)'
                                     )
                                     _dpp_preamble += (
                                         f'  {_dpp_old_dst_type} sdwa_old_dst_[64] = {{}};\n'
-                                        '  if (sdwa_dst_sel_ != amdgpu::sdwa::DWORD ||\n'
+                                        '  if ((sdwa_dst_sel_ != amdgpu::sdwa::DWORD &&\n'
+                                        '       sdwa_dst_unused_ == amdgpu::sdwa::UNUSED_PRESERVE) ||\n'
                                         '      inst_.src0 == amdgpu::SRC_DPP) {\n'
                                         '    uint32_t vb = wf.vgpr_alloc().base;\n'
                                         '    uint64_t ex = wf.exec();\n'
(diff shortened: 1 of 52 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/register_access.h
```diff
@@ -657,44 +702,50 @@ class RegisterAccess {
...
 
   [[nodiscard]] OperandWriteView write_operand(const Operand &op, uint64_t lane_mask) const {
-    // Kept at acquisition time so future write-observation plugins can validate
-    // the instruction's destination lane set before writable storage is exposed.
-    (void)lane_mask;
     Wavefront &wf = mutable_wavefront();
-    return OperandWriteView(op, wf, op.simd_vgpr_storage_mut(wf));
+    VgprStorage *storage = op.simd_vgpr_storage_mut(wf);
+    if (storage)
+      op.simd_notify_write_mut(wf, lane_mask, rocjitsu::ExecutionPlugin::kFullByteMask);
```
kuhar, line 753: We should apply the architectural DPP write mask before notifying plugins. `write_operand(vdst, exec)` currently reports every EXEC lane, then the cleanup at `_generator.py:7197-7206` restores excluded lanes through another observed write. A plugin therefore sees preserved lanes as destination definitions even though the hook contract says `lane_mask` is the lanes written by the instruction. Please notify only `exec & dpp_write_mask` (or delay notification until that mask is known) and add a partial row/bank-mask case.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/register_access.h
```diff
@@ -733,8 +784,9 @@ class RegisterAccess {
     return read_vgpr_region(physical_reg, 2, uint64_t{1} << lane, byte_mask).lane64(0, lane);
   }
 
-  void write_vgpr(uint32_t physical_reg, uint32_t lane, uint32_t value) const {
-    write_vgpr_region(physical_reg, 1, uint64_t{1} << lane).set_lane(0, lane, value);
+  void write_vgpr(uint32_t physical_reg, uint32_t lane, uint32_t value,
+                  uint8_t byte_mask = 0xF) const {
+    write_vgpr_region(physical_reg, 1, uint64_t{1} << lane, byte_mask).set_lane(0, lane, value);
```
kuhar, line 834: We should report only the SDWA bytes that are architecturally written. For a byte/word `dst_sel` with `dst_unused:PRESERVE`, the semantic write is observed as full-dword and the postamble at `_generator.py:7129-7137` performs another observed full-dword merge. That makes `byte_mask` 0xF even though the unselected bytes are preserved. Please propagate the selected byte mask and keep the internal preservation merge unobserved, with a preserve-mode test.

kuhar (commented):
Could we round out the new write-observation tests with the contract cases that are still missing? The current suite passes, but it only inspects a 32-bit positive write and native-store rejection. Representative gaps are: `emulation/rocjitsu/tests/register_access_test.cpp:287`, assert that a 64-bit destination reports both physical registers; `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/register_access.h:179` and `:101`, exercise `store_narrow` and the end-of-wave window guard; and `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/memory_pipeline.cpp:166`, assert that a completion remains unobserved as documented. These can live in focused tests without duplicating every accessor combination.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/memory_pipeline.cpp
```diff
@@ -41,6 +41,10 @@ uint32_t extend_scalar_load(const uint8_t *bytes, uint32_t elem_size, bool sign_
   return value;
 }
 
+uint32_t read_vgpr_storage(const ComputeUnitCore &cu, uint32_t reg, uint32_t lane) {
```
kuhar: It would help to document that `read_vgpr_storage` deliberately bypasses observation. The old D16 merge lambda explained that preserved destination bytes are storage state, not an instruction-visible source; moving the read into this shared helper dropped that rationale, making it easy to replace with `RegisterAccess::read_vgpr` and introduce spurious plugin reads. A one-line comment above the helper should preserve the distinction.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/register_access.h
```diff
@@ -657,44 +702,50 @@ class RegisterAccess {
   }
 
   [[nodiscard]] OperandWriteView write_operand(const Operand &op, uint64_t lane_mask) const {
```
kuhar, line 749: We should update the `RegisterAccess` overview to cover the new write contract. It currently documents only read notifications, while write views and regions now notify at acquisition and later stores must remain within `observed_lane_mask`. This is central behavior for callers, so please state the write-observation timing and mask invariant alongside the existing read rules.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -9298,10 +9299,12 @@ def gen_operand(self) -> None:
                     uint32_t off = packed->reg + (wf.vgpr_msb_for_role(vgpr_msb_role()) << 8);
                     uint32_t voff = wf.gpr_idx_en() ? amdgpu::apply_gpr_idx(wf, off, true) : off;
                     uint32_t idx = wf.vgpr_alloc().base + voff;
-                    uint32_t old = amdgpu::RegisterAccess(wf.cu()).read_vgpr(idx, lane);
+                    uint8_t write_byte_mask = packed->shift ? rocjitsu::ExecutionPlugin::kHighHalfByteMask : rocjitsu::ExecutionPlugin::kLowHalfByteMask;
+                    uint8_t preserved_byte_mask = packed->shift ? rocjitsu::ExecutionPlugin::kLowHalfByteMask : rocjitsu::ExecutionPlugin::kHighHalfByteMask;
+                    uint32_t old = amdgpu::RegisterAccess(wf.cu()).read_vgpr(idx, lane, preserved_byte_mask);
```
kuhar: We should read the preserved half through raw storage during a packed-16 destination merge. `_generator.py:9304` uses `RegisterAccess::read_vgpr`, so each destination write emits a source-read hook for bytes the instruction does not read; `Packed16ReadsAndWritesObserveSelectedByteHalves` currently asserts that artifact. This is inconsistent with the unobserved D16 merge in `memory_pipeline.cpp`. Please use the VM/raw storage accessor and update the test to expect no preserved-half read.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/register_access.h
newling: I agree with this, rocjitsu currently implements DPP destinations by first writing every active EXEC lane, then restoring the lanes excluded by the DPP mask.

I now use raw storage for the old-destination snapshot and restoration. This prevents those internal save-and-restore operations from firing extra plugin callbacks. However, the main write callback still reports too many lanes: it reports all EXEC lanes instead of only EXEC & dpp_write_mask.

Fixing that is difficult in the current execution structure. The generic RegisterAccess code fires the write callback using EXEC, while the DPP mask is known by a different, encoding-specific layer. Making the callback exact requires passing the DPP mask into the generic write path or separating plugin reporting from the emulator’s storage operations. I intend to address that in a follow-up rather than adding more DPP-specific state to Wavefront.

The new characterization test records the current over-reporting and the exact result we ultimately want. I also found that this problem already exists for reads: DPP source staging can report more source lanes than the instruction actually uses. Thanks for pointing this out; I was not aware of the read-side problem before this review.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/register_access.h
newling: This has the same underlying issue as DPP: rocjitsu first performs a full-dword write, then restores preserved bytes.

The snapshot and merge now use raw storage, preventing extra callbacks from that bookkeeping. However, the main callback still reports byte_mask = 0xF instead of only the bytes selected by dst_sel. I intend to fix this with DPP in the follow-up execution/effects refactor. The new characterization test records the current over-reporting and the exact byte masks we ultimately want. SDWA source reads also have similar pre-existing over-reporting.

kuhar (approved):
LGTM

newling:
@kuhar I'm going to land this, as it frees up my actual race detector plugin development. But I have an improved design for the foundations in the works that should solve the dpp/sdwa issues (while leaving the plugin API in this PR unchanged). 

Thanks for the review!

newling:
> @kuhar I'm going to land this, as it frees up my actual race detector plugin development. But I have an improved design for the foundations in the works that should solve the dpp/sdwa issues (while leaving the plugin API in this PR unchanged).

Improved design follow-up: https://github.com/ROCm/rocm-systems/pull/9250

PR #8805, opened 2026-07-18
fix(amdsmi): route UMA carveout through fwupd on UEFI-HII platforms

yalmusaf:
## Motivation

On HP Strix Halo / UEFI-HII platforms (e.g. ZBook Ultra G1a, Z2 Mini G1a) the amdgpu `/sys/class/drm/<card>/device/uma/carveout` sysfs node is absent, so `amd-smi static --mem-carveout` printed `N/A` and `amd-smi set --mem-carveout <index>` reported "not supported" — even though the "Dedicated Graphics Memory" carveout knob genuinely exists and is writable through the UEFI-HII / fwupd BIOS-settings interface (the same interface `fwupdmgr get-bios-settings` uses). This adds an fwupd fallback so the carveout can be read and set on those platforms.

## Technical Details

- New CLI adapter `amdsmi_cli/fwupd_bios.py` that reads/writes the "Dedicated Graphics Memory" enumeration via `fwupdmgr --json` (`get-bios-settings` / `set-bios-setting`). It shells out with list-argv (no shell), gates on fwupd version (read >= 1.8.4, write >= 2.1.1), matches the setting by name + enumeration type, and honors `AMDSMI_DRY_RUN`.
- `static.py` / `set_value.py`: when the C getter/setter returns `NOT_SUPPORTED` (amdgpu node absent), fall back to fwupd instead of reporting N/A. The amdgpu-node-present path is unchanged.
- Privilege is handled by fwupd/PolicyKit, so the fwupd write path does not pre-require root (no explicit `sudo`).
- Three-state messaging: amdgpu node present (unchanged) / absent-but-fwupd (options listed, JSON tagged `source: "fwupd"`) / neither (`N/A (no UMA carveout interface: ...)`).
- Implemented entirely in the Python CLI; the C library, C API, ctypes wrapper, and Rust/Go bindings are untouched.
- Docs (`docs/how-to/amdsmi-cli-tool.md`, `docs/reference/amdsmi-py-api.md`) and `CHANGELOG.md` updated.

## Issue Tracking

JIRA ID: ROCM-26247

## Test Plan

New mock-based unit test `tests/python/unit/gpu/test_cli_mem_carveout_fwupd.py` (fwupd `subprocess` / `shutil.which` mocked, fake `amdsmi` module injected) so it runs with no GPU, no root, and no compiled library. Covers: option parsing + 0-based indexing, current-index resolution, exact `set-bios-setting` argv (index -> value string), `AMDSMI_DRY_RUN` suppresses the write, fwupd version gating (read/write), absent-`fwupdmgr` graceful degradation, same-name/wrong-type rejection, and the static human / JSON (`source: fwupd`) / neither-interface messaging.

```
python3 -m pytest tests/python/unit/gpu/test_cli_mem_carveout_fwupd.py -q
```

## Test Result

`14 passed` locally. Note: end-to-end validation of the fwupd write path requires an HP UEFI-HII unit with fwupd >= 2.1.1 (Ubuntu 26.04+) and a reboot; that hardware path is not exercised by the mocked unit tests.

## Submission Checklist


superm1:
Something worth considering - should `amd-smi` just use fwupd for everything BIOS setting related?  Rather than hitting the sysfs directly on the non-HP systems.

projects/amdsmi/amdsmi_cli/fwupd_bios.py
```diff
@@ -0,0 +1,171 @@
...
+from typing import List, Optional
+
+# Reading BIOS settings works from fwupd 1.8.4; writing the AMD carveout setting
+# needs 2.1.1 (Ubuntu 26.04+).
+_MIN_READ_VERSION = (1, 8, 4)
+_MIN_WRITE_VERSION = (2, 1, 1)
+
+# The knob is matched case-insensitively by name and by enumeration type, rather
+# than a single hard-coded string, so vendor naming variants still resolve.
+_CARVEOUT_NAME_CANDIDATES = ("dedicated graphics memory",)
```
superm1: The semantics might not be identical across vendors.  I would instead suggest using the appstream ID to discover and decide how to treat it.

AMD's attribute
```
com.amd-gpu.uma_carveout
```

HP's 
```
com.hp-bioscfg.Dedicated_Graphics_Memory
```

projects/amdsmi/amdsmi_cli/fwupd_bios.py
```diff
@@ -0,0 +1,171 @@
...
+
+def _run_fwupdmgr(args: List[str]) -> Optional[subprocess.CompletedProcess]:
+    exe = shutil.which("fwupdmgr")
+    if exe is None:
+        return None
+    return _run([exe, *args])
+
+
+def _fwupd_version() -> Optional[tuple]:
+    result = _run_fwupdmgr(["--version"])
```
superm1: you should be using `fwupdmgr --version --json` if you're going to wrap the tool

superm1: Here's some sample output.

```
╰─❯ fwupdmgr --version --json
{
  "Versions" : [
    {
      "Type" : "compile",
      "AppstreamId" : "info.libusb",
      "Version" : "1.0.27"
    },
    {
      "Type" : "compile",
      "AppstreamId" : "com.hughsie.libxmlb",
      "Version" : "0.3.24"
    },
    {
      "Type" : "compile",
      "AppstreamId" : "com.hughsie.libjcat",
      "Version" : "0.2.3"
    },
    {
      "Type" : "compile",
      "AppstreamId" : "org.freedesktop.fwupd",
      "Version" : "2.0.20"
    },
    {
      "Type" : "runtime",
      "AppstreamId" : "org.freedesktop.fwupd-efi",
      "Version" : "1.4"
    },
    {
      "Type" : "runtime",
      "AppstreamId" : "com.hughsie.libxmlb",
      "Version" : "0.3.24"
    },
    {
      "Type" : "runtime",
      "AppstreamId" : "com.hughsie.libjcat",
      "Version" : "0.2.3"
    },
    {
      "Type" : "runtime",
      "AppstreamId" : "org.kernel",
      "Version" : "6.17.0-1028-oem"
    },
    {
      "Type" : "runtime",
      "AppstreamId" : "org.freedesktop.fwupd",
      "Version" : "2.0.20"
    }
  ]
}
```

projects/amdsmi/amdsmi_cli/fwupd_bios.py
superm1: Here is the json for one with AMD's
```
{
  "BiosSettings": [
    {
      "Name": "Dedicated Video Memory",
      "Description": "GPU unified memory architecture carveout size for system memory",
      "Filename": "/sys/devices/pci0000:00/0000:00:08.1/0000:c2:00.0/uma",
      "BiosSettingId": "com.amd-gpu.uma_carveout",
      "BiosSettingCurrentValue": "Minimum (512 MB)",
      "BiosSettingFilename": "carveout",
      "BiosSettingReadOnly": false,
      "BiosSettingType": 1,
      "BiosSettingPossibleValues": [
        "Minimum (512 MB)",
        "(1 GB)",
        "(2 GB)",
        "(4 GB)",
        "Medium (8 GB)",
        "High (16 GB)"
      ]
    }
  ]
}
```

adam360x: Done in 567b446 — `_find_carveout` now matches by the AppStream `BiosSettingId`, preferring AMD's `com.amd-gpu.uma_carveout` and falling back to HP's `com.hp-bioscfg.Dedicated_Graphics_Memory`, instead of the display name. Verified on a real HP ZBook Ultra G1a: it matches the HP id and correctly ignores similarly-named decoys such as `Memory Size`.

projects/amdsmi/amdsmi_cli/fwupd_bios.py
adam360x: Done in 567b446 — `_fwupd_version` now runs `fwupdmgr --version --json` and reads the `org.freedesktop.fwupd` runtime entry from `Versions[]` (falling back to the compile-time entry), and treats a non-zero return code as unavailable. Verified against fwupd 2.1.1 on an HP ZBook Ultra G1a.

projects/amdsmi/amdsmi_cli/subcommands/static.py
```diff
@@ -1377,6 +1327,74 @@ def static_gpu(
...
+            uma_info = amdsmi_interface.amdsmi_get_gpu_uma_carveout_info(args.gpu)
+            logging.debug(f"UMA carveout info: {uma_info}")
+            self._render_mem_carveout(
+                static_dict, uma_info.get("options", []), uma_info.get("current_index", -1)
+            )
+        except amdsmi_exception.AmdSmiLibraryException as e:
+            not_supported = (
+                e.get_error_code() == amdsmi_interface.amdsmi_wrapper.AMDSMI_STATUS_NOT_SUPPORTED
+            )
+            fwupd_setting = fwupd_bios.get_carveout_setting() if not_supported else None
```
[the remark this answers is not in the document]
adam360x: Good catch — fixed in e91402911cc.

The fwupd fallback in `_store_mem_carveout` is now gated behind an APU check: `if not_supported and self.helpers.is_device_apu(args.gpu)`. The new `AMDSMIHelpers.is_device_apu()` reads the ASIC `AMDGPU_IDS_FLAGS_FUSION` flag (the same fusion flag already used in `amdsmi_helpers.py`). On a hybrid APU+dGPU system the discrete GPU no longer surfaces the APU's system-wide fwupd carveout — it just reports the sysfs `NOT_SUPPORTED` result as before. Added regression test `test_static_dgpu_skips_fwupd`.

projects/amdsmi/amdsmi_cli/subcommands/set_value.py
```diff
@@ -1837,6 +1768,147 @@ def format_fan_error(message, include_driver_note=False):
...
+            self.logger.print_output()
+            self.helpers.prompt_reboot()
+
+        except amdsmi_exception.AmdSmiLibraryException as e:
+            if e.get_error_code() == amdsmi_interface.amdsmi_wrapper.AMDSMI_STATUS_NO_PERM:
+                raise PermissionError("Command requires elevation") from e
+            if e.get_error_code() == amdsmi_interface.amdsmi_wrapper.AMDSMI_STATUS_NOT_SUPPORTED:
+                # The kernel node is absent; the knob may still be exposed by the
+                # platform BIOS through fwupd (HP UEFI-HII).
+                self._set_mem_carveout_fwupd(args)
```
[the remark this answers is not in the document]
adam360x: Fixed in e91402911cc.

The fwupd set path in `_set_mem_carveout` now only runs when the target is the integrated GPU — gated by `self.helpers.is_device_apu(args.gpu)` (ASIC `AMDGPU_IDS_FLAGS_FUSION`). A selected dGPU that returns `NOT_SUPPORTED` no longer enters the fwupd path (which could otherwise change the APU's BIOS allocation); it instead reports "Not supported: UMA carveout applies to the integrated GPU (APU) only." Added regression test `test_set_dgpu_skips_fwupd`.

projects/amdsmi/amdsmi_cli/subcommands/set_value.py
```diff
@@ -1837,6 +1768,156 @@ def format_fan_error(message, include_driver_note=False):
             self.logger.clear_multiple_devices_output()
             return
 
+    def _set_mem_carveout(self, args):
+        """Set the VRAM carveout, falling back to fwupd on UEFI-HII platforms.
```
superm1, line 1772: I thought you were going to flip this so that fwupd is the default and then the attribute directly is the fallback

projects/amdsmi/amdsmi_cli/subcommands/set_value.py
```diff
@@ -1837,6 +1768,156 @@ def format_fan_error(message, include_driver_note=False):
...
+        (NOT_SUPPORTED) the platform BIOS may still expose the knob through
+        fwupd (HP UEFI-HII), so route the write there instead of refusing.
+        """
+        # Validate single GPU (VRAM is per-GPU)
+        if isinstance(args.gpu, list) and len(args.gpu) > 1:
+            raise ValueError(
+                "VRAM carveout can only be set for a single GPU. Please specify --gpu <id>"
+            )
+
+        try:
```
superm1, line 1784: fwupd will only expose this on an APU, I guess this argument only really matters for the fallback path that means

projects/amdsmi/amdsmi_cli/fwupd_bios.py
superm1, line 1: So if you were to build with gobject introspection in https://github.com/ROCm/TheRock/pull/7278 then you actually could just use a VERY simple gobject introspection python script for a client here.

Here's a sample of what I mean by a simple client:

https://github.com/fwupd/fwupd/blob/main/contrib/firmware_packager/simple_client.py

superm1 (commented):
IMO it is unfortunate to have both a python and a C implementation in tree, can we just have the C implementation and the python one uses that?

projects/amdsmi/amdsmi_cli/subcommands/set_value.py
adam360x: Done — consolidated onto the C library in 52f82f1b88. `amdsmi_get_gpu_uma_carveout_info()` / `amdsmi_set_gpu_uma_carveout()` now try fwupd first and fall back to the amdgpu sysfs attribute, so fwupd is the default and the raw attribute is the fallback. The Python duplicate is gone and the CLI just calls the C API — this line is part of `set_value.py`, which is now reverted to `develop`.

Verified on an HP ZBook Ultra G1a (Strix Halo APU): the getter returns the seven **Dedicated Graphics Memory** options through libfwupd with the current value marked.

projects/amdsmi/amdsmi_cli/subcommands/set_value.py
adam360x: Right. The fwupd path is now gated to the integrated (FUSION) GPU inside the C library, so on a hybrid APU+dGPU system only the APU consults the platform-wide BIOS setting and a dGPU handle goes straight to the sysfs fallback. Confirmed on hardware that the APU reports `AMDGPU_IDS_FLAGS_FUSION` (`asic.flags = 0x19`). This thread was on the now-reverted `set_value.py`; the gating lives in the C library (`amd_smi.cc`).

projects/amdsmi/amdsmi_cli/fwupd_bios.py
adam360x: Took this a step further and dropped the Python fwupd client entirely. Since TheRock#7278 (now approved) vendors libfwupd, the C library links it directly (`-DAMDSMI_ENABLE_FWUPD=ON`) and the CLI just calls the existing C API — so there's no second implementation to maintain, GI script or otherwise. Default builds without libfwupd compile a small stub that returns `NOT_SUPPORTED`, leaving the sysfs path unchanged. `fwupd_bios.py` is removed in 52f82f1b88.

adam360x:
Thanks @superm1 — addressed the "two implementations" concern by consolidating onto a single **C** implementation and removing the Python one (52f82f1b88):

- Deleted `fwupd_bios.py` and its CLI plumbing; `static.py` / `set_value.py` / `amdsmi_helpers.py` are reverted to `develop`, so the CLI is unchanged from upstream and simply calls the existing `amdsmi_get_gpu_uma_carveout_info()` / `amdsmi_set_gpu_uma_carveout()` C API.
- The fwupd logic now lives in the C library (`src/amd_smi/fwupd_carveout.cc`) behind `-DAMDSMI_ENABLE_FWUPD=ON`, linking the libfwupd vendored by TheRock#7278. It tries fwupd first and falls back to the amdgpu sysfs node, gated to the integrated (FUSION) GPU.
- Default builds (no libfwupd) compile a stub returning `NOT_SUPPORTED`; added a hardware-free C++ unit test for that contract.

The net PR diff is now **C-side only** — no CLI/Python changes.

**Validated end-to-end on an HP ZBook Ultra G1a (Strix Halo APU):**
- GET returns the 7 `Dedicated Graphics Memory` options via libfwupd (`512 MB` … `96 GB`), current value marked.
- APU gate confirmed: `asic.flags = 0x19` (FUSION set).
- SET resolves the setting/value correctly, validated via dry-run and an idempotent set-to-current no-op.

One hardware note: writing a *different* value on this particular unit is rejected by its firmware with `Invalid argument` — reproducible with `fwupdmgr set-bios-setting` and a direct sysfs write as well, so it's a platform/BIOS limitation on this box rather than the library. The read path and setting discovery work correctly through libfwupd.

superm1 (approved):

adam360x:
Closing, handled in https://github.com/ROCm/rocm-systems/pull/11066/changes/83838cee8132d0364ee242bcdc78c5234cf96da7

PR #8808, opened 2026-07-18
fix(clr): internal linkage for bounded builtin index accessors

michaelselehov:
Add a __HIP_DEVICE_BUILTIN_INTERNAL macro that marks the accessor with internal_linkage, making it an exact definition IPSCCP can track. Applied only to the block-size bounded accessors (threadIdx, blockDim); blockIdx and gridDim keep the existing macro. NFC for HIP semantics.

## Motivation

GenericBandwidth (and other kernels whose addressing is strided by the block
size) regressed on gfx950 / MI350 under full device LTO.

Root cause is in how the HIP built-in index accessors are emitted. `threadIdx.*`
and `blockDim.*` resolve through small `linkonce_odr` accessor functions. Because
`linkonce_odr` is not an *exact* definition, LLVM's IPSCCP refuses to track their
return values and therefore cannot attach the `range()` return attribute that the
underlying `__ockl_get_local_*` / workgroup-size builtins already carry. With the
range lost, strided GEPs derived from these indices drop `inbounds`, which blocks
address-offset folding in the backend -- extra `v_lshl_add_u64` address math.

## Technical Details

Add a `__HIP_DEVICE_BUILTIN_INTERNAL` variant of the accessor macro that marks the
generated `__get_*` wrapper with `__attribute__((internal_linkage))`. Internal
linkage makes the accessor an exact, non-interposable definition in every TU, so
IPSCCP can propagate the `range(0, 1024)` inferred from the workitem/workgroup
builtins.

Only the accessors that are actually bounded by the block size use the new macro:

- `__hip_builtin_threadIdx_t` (x/y/z)
- `__hip_builtin_blockDim_t` (x/y/z)

`blockIdx` and `gridDim` are not block-size bounded and keep the existing macro.

This is NFC for HIP semantics: the accessor still forwards to the same builtin;
only the linkage of the tiny wrapper changes, restoring the range / `inbounds`
information the optimizer needs.

## Issue Tracking

JIRA ID: SWHIPSDK-47

## Test Plan

No new unit test: this is a device-codegen / performance fix, not a functional or
API change, and cannot be meaningfully exercised by a hip-tests unit test.
Validated on hardware with HIPPerf GenericBandwidth under full device LTO:

- MI300 and MI350 (gfx950).
- Codegen marker: presence of the redundant `v_lshl_add_u64` address arithmetic
  in the generated ISA (proxy for lost `inbounds` / offset folding).
- Throughput: HIPPerf GenericBandwidth bandwidth vs the pre-regression baseline.

## Test Result

- Offset folding restored: the redundant `v_lshl_add_u64` address arithmetic
  disappears from the generated ISA, matching the pre-regression codegen.
- HIPPerf GenericBandwidth recovered to the pre-regression baseline, closing a
  regression of up to ~11% on MI300 and up to ~20% on MI350 (gfx950).
- No codegen / behavioral change on the older (pre-regression) toolchain:
  rebuild clean, ISA unchanged, hip-tests pass, perf stable.

### DIFF
--- a/projects/clr/hipamd/include/hip/amd_detail/amd_hip_runtime.h
+++ b/projects/clr/hipamd/include/hip/amd_detail/amd_hip_runtime.h
@@ -268,10 +268,24 @@ __DEVICE__ unsigned int __hip_get_grid_dim_z() { return __ockl_get_num_groups(2)
   __declspec(property(get = __get_##DIMENSION)) unsigned int DIMENSION;                            \
   __DEVICE__ unsigned int __get_##DIMENSION(void) { return FUNCTION; }
 
+// Give the (block-size bounded) built-in index accessors internal linkage so
+// that IPSCCP can attach the range() return attribute inferred from the
+// underlying __ockl_get_local_* / workgroup-size builtins. As linkonce_odr the
+// accessors are not exact definitions, so IPSCCP refuses to track their return
+// values; under full device LTO the missing range lets strided GEPs lose
+// inbounds, which blocks offset folding (extra v_lshl_add_u64, higher VGPR
+// pressure, lower occupancy). Only dimensions bounded by the block size
+// (threadIdx, blockDim) use this variant.
+#define __HIP_DEVICE_BUILTIN_INTERNAL(DIMENSION, FUNCTION)                                         \
+  __declspec(property(get = __get_##DIMENSION)) unsigned int DIMENSION;                            \
+  __attribute__((internal_linkage)) __DEVICE__ unsigned int __get_##DIMENSION(void) {              \
+    return FUNCTION;                                                                               \
+  }
+
 struct __hip_builtin_threadIdx_t {
-  __HIP_DEVICE_BUILTIN(x, __hip_get_thread_idx_x());
-  __HIP_DEVICE_BUILTIN(y, __hip_get_thread_idx_y());
-  __HIP_DEVICE_BUILTIN(z, __hip_get_thread_idx_z());
+  __HIP_DEVICE_BUILTIN_INTERNAL(x, __hip_get_thread_idx_x());
+  __HIP_DEVICE_BUILTIN_INTERNAL(y, __hip_get_thread_idx_y());
+  __HIP_DEVICE_BUILTIN_INTERNAL(z, __hip_get_thread_idx_z());
 #ifdef __cplusplus
   __device__ operator dim3() const { return dim3(x, y, z); }
 #endif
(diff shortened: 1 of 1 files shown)


michaelselehov:
This PR intentionally ships without a new unit test. The change is a device-side
codegen / performance fix, not a functional or API change: it only gives the
block-size-bounded builtin index accessors (`threadIdx.*`, `blockDim.*`) internal
linkage so that IPSCCP can attach the `range()` return attribute and preserve
`inbounds` on the derived address computation. There is no observable behavioral
difference for a hip-tests runtime test to assert -- the accessors still forward
to the same builtins and return the same values.

It was validated on real hardware instead: HIPPerf GenericBandwidth under full
device LTO on MI300 and MI350 (gfx950). The fix removes the redundant
`v_lshl_add_u64` address arithmetic from the generated ISA (matching the
pre-regression codegen) and restores GenericBandwidth throughput to the
pre-regression baseline; the older toolchain shows no codegen or behavioral
change.

Removing the `Not ready to Review` label so clr-reviewers can pick it up.
@ROCm/clr-reviewers -- please take a look. Tracking: JIRA ID SWHIPSDK-47.

projects/clr/hipamd/include/hip/amd_detail/amd_hip_runtime.h
```diff
@@ -268,10 +268,24 @@ __DEVICE__ unsigned int __hip_get_grid_dim_z() { return __ockl_get_num_groups(2)
...
+// accessors are not exact definitions, so IPSCCP refuses to track their return
+// values; under full device LTO the missing range lets strided GEPs lose
+// inbounds, which blocks offset folding (extra v_lshl_add_u64, higher VGPR
+// pressure, lower occupancy). Only dimensions bounded by the block size
+// (threadIdx, blockDim) use this variant.
+#define __HIP_DEVICE_BUILTIN_INTERNAL(DIMENSION, FUNCTION)                                         \
+  __declspec(property(get = __get_##DIMENSION)) unsigned int DIMENSION;                            \
+  __attribute__((internal_linkage)) __DEVICE__ unsigned int __get_##DIMENSION(void) {              \
+    return FUNCTION;                                                                               \
+  }
```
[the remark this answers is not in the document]
michaelselehov: Resolved.

projects/clr/hipamd/include/hip/amd_detail/amd_hip_runtime.h
```diff
@@ -268,10 +268,30 @@ __DEVICE__ unsigned int __hip_get_grid_dim_z() { return __ockl_get_num_groups(2)
...
+// that IPSCCP can attach the range() return attribute inferred from the
+// underlying __ockl_get_local_* / workgroup-size builtins. As linkonce_odr the
+// accessors are not exact definitions, so IPSCCP refuses to track their return
+// values; under full device LTO the missing range lets strided GEPs lose
+// inbounds, which blocks offset folding (extra v_lshl_add_u64, higher VGPR
+// pressure, lower occupancy). Only dimensions bounded by the block size
+// (threadIdx, blockDim) use this variant. Guard the attribute with
+// __has_attribute so that compilers without internal_linkage fall back to the
+// existing accessor definition instead of emitting -Wattributes.
+#if defined(__has_attribute) && __has_attribute(internal_linkage)
```
AlexeySachkov: This is not a portable use of `__has_attribute`, see [gcc docs](https://gcc.gnu.org/onlinedocs/gcc-12.1.0/cpp/_005f_005fhas_005fattribute.html)

michaelselehov: Thanks. Switched to the nested idiom.

mangupta (approved):

AlexeySachkov (approved):

PR #8624, opened 2026-07-15
refactor(amdsmi): deterministic CLI exit codes + generic -g all regression guard

charis-poag-amd:
# refactor(amdsmi): deterministic CLI exit codes + generic `-g all` regression guard

## Summary
Makes `amd-smi` process exit codes **deterministic and self-consistent**. The CLI now records per-device failures and decides a single exit code once, after every device has been attempted — instead of aborting a multi-device loop on the first error or letting an uncaught exception collapse to `1`. Adds generic `-g all` set/reset regression guards so every GPU option is exercised for both the failure and success paths.

**Refs:** JIRA `ROCM-25289` · intended to replace #6862 · minimal reference implementation for the #862 fixes.

## Motivation
Scripts/CI consume `amd-smi`'s exit code (the motivation behind `ROCM-25289`).   
Previously:
- Several `amd-smi`'s CLI's set & reset displayed SMI library errors, but would exit 0. This caused scripts to be unaware of any issues during the request.
- Many distinct failures **aliased onto `1`/`2`**, so callers couldn't distinguish import failure vs. bad command vs. driver-not-loaded.
- An **uncaught library error exited `1`** — colliding with `AMDSMI_STATUS_INVAL` (also `1`).
- A single device failure in a `-g all` loop could **abort early**, leaving remaining GPUs unattempted.
- The printed `code` field disagreed with `$?` (printed `-7`, exited `7`).

## Two exit-code domains (important)
Exit codes come from **two deliberately non-overlapping sources**:

1. **Library failures** — surfaced as the **real `amdsmi_status_t` value**, folded to a byte (`abs(status) & 0xFF`). Real statuses are `0`–`56` (e.g. `INVAL=1`, `NOT_SUPPORTED=2`, `NO_PERM=10`, `SETTING_UNAVAILABLE=55`) and pass through unchanged. The two 32-bit sentinels fold into the top of the byte range: `AMDSMI_STATUS_MAP_ERROR (0xFFFFFFFE) → 254`, `AMDSMI_STATUS_UNKNOWN_ERROR (0xFFFFFFFF) → 255`.
2. **CLI-invented codes** — parse-time / CLI-only conditions live in the reserved **`192`–`208` band** and are **never** produced by the library.

[...]

PR #8819, opened 2026-07-18
fix(amdsmi): emit valid JSON for ras --cper --json with no entries

yalmusaf:
## Motivation

QA (DCGPUSDV-4010) runs `sudo amd-smi ras --cper --json` on multi-GPU MI300/MI325/MI350 systems and feeds stdout to `json.loads()`. When there are no CPER entries — the common case — the command printed nothing, so `json.loads` failed with `Expecting value: line 1 column 1 (char 0)`. This makes `amd-smi ras --cper --json` always emit exactly one valid JSON document.

## Technical Details

- `amdsmi_cli/subcommands/ras.py`: `--cper --json` output now goes through the standard logger path for both cases — `self.logger.multiple_device_output = all_json_rows; self.logger.print_output(multiple_device_enabled=True, emit_empty=True)` — instead of a bespoke `json.dumps`. With no entries (and not `--follow`) this emits a single `[]`; with entries it emits one aggregated array across all GPUs; `--follow` does not spam `[]`.
- `amdsmi_cli/amdsmi_logger.py`: adds an **opt-in** `emit_empty` flag (default `False`) to `print_output` / `_print_json_output`. The only behavioral change is the stdout guard `if json_output:` -> `if json_output or emit_empty:`, so an empty list renders as `[]` instead of nothing. Default-off keeps every existing `multiple_device_enabled=True` caller byte-for-byte unchanged; only the ras `--cper` path opts in.
- The partition warning (`WARNING: CPER files are only available on primary partitions`) is now guarded by `not is_json_format()`, so it no longer corrupts `--json` output.

Python-CLI-only; the C library, C API, ctypes wrapper, and Rust/Go bindings are untouched.

## Issue Tracking

JIRA ID: DCGPUSDV-4010

## Test Plan

New mock-based unit test `tests/python/unit/gpu/test_cli_ras_cper_json.py` (real `AMDSMILogger` in JSON mode, driver boundary faked; no GPU/root/library). Covers: no entries -> stdout is exactly `[]` and `json.loads` succeeds; two GPUs each with an entry -> a single `json.loads` list; non-primary partition + `--json` -> no `WARNING` text in stdout.

```
python3 -m pytest tests/python/unit/gpu/test_cli_ras_cper_json.py -q
```

## Test Result

`3 passed` (proven RED against pre-fix `ras.py`: the no-entries and partition-warning assertions fail). Existing `tests/python/cli/test_ras.py`: 1 passed, 1 skipped (no regression).

## Submission Checklist


marifamd (changes requested):
Move the prints to the logger or remove all logger usage from the subcommand. Let's not mix and match.
Look at https://github.com/ROCm/rocm-systems/pull/8624 for the ras.py changes

marifamd (approved):

gabrpham (approved):

PR #8820, opened 2026-07-19
feat(clr): isolate single-grid cooperative counter on its own cache line

neoblizz:
## Motivation

The single-grid cooperative counter (`MGSyncInfo::sgs`) shares a cache line with the read-only fields of `MGSyncInfo`. Its device-scope atomic traffic evicts that line across XCDs on every workgroup arrival, so the grid barrier's cost scales with workgroup count. Isolating the counter on its own cache line removes the false sharing.

## Technical Details

`MGSyncInfo::sgs` becomes a pointer to a counter the runtime allocates on its own 128B cache line (`kSGSyncLineSize`), instead of an embedded `MGSyncData`. The device library reads it through the pointer. `static_assert`s pin the `MGSyncInfo` offset/size so it stays in sync with `struct mg_info` in device-libs `cg.cl`.

 ### Dependencies and merge order:
  - **Depends on #8748 (kSGInfoSize fix) - must merge first; this PR is stacked on it.
  - **ABI-lockstep with https://github.com/ROCm/llvm-project/pull/3464** (device-libs `cg.cl`) - both must merge together. Merging either side alone changes the shared `mg_info`/`MGSyncInfo` contract and breaks cooperative-groups grid sync.
  - Order: #8748 -> then this PR + https://github.com/ROCm/llvm-project/pull/3464together.

## Test Plan

- Build clr + device-libs; run a cooperative `this_grid().sync()` kernel.
- Cross-workgroup producer/consumer correctness at full occupancy.

## Test Result

Validated on MI300X (gfx942) and MI355X (gfx950): correctness PASS (500k barrier episodes); grid-barrier latency ~2.2x (gfx942) / ~2.66x (gfx950) vs baseline (upstream clr and rocm-systems).

## Submission Checklist


chrispaquot (approved):

projects/clr/rocclr/device/rocm/rocvirtual.cpp
```diff
@@ -4656,8 +4656,13 @@ bool VirtualGPU::submitKernelInternal(const amd::NDRangeContainer& sizes, const
           syncInfo->mgs = nullptr;
         }
         if (multiGridSync || singleGridSync) {
-          // Update sync data address.
-          syncInfo->sgs = {0};
+          // Give the single-grid counter its own cache line so its device-scope
+          // atomic traffic does not evict the read-only fields in MGSyncInfo.
+          // Reserve the whole line so the next kernarg allocation cannot share it.
+          auto* sgs = reinterpret_cast<Device::MGSyncData*>(
```
b-sumner: We can do this by adding padding to the original struct.  A separate new allocation is unnecessary.

b-sumner (changes requested):

projects/clr/rocclr/device/device.hpp
```diff
@@ -1767,6 +1770,14 @@ class Device : public RuntimeObject {
   static constexpr size_t kMGSyncDataSize = sizeof(MGSyncData);
   static constexpr size_t kMGInfoSizePerDevice = kMGSyncDataSize + sizeof(MGSyncInfo);
   static constexpr size_t kSGInfoSize = sizeof(MGSyncInfo);
+  // The single-grid counter is placed on its own cache line so its device-scope
+  // atomic traffic does not evict the read-only fields sharing MGSyncInfo.
+  static constexpr size_t kSGSyncLineSize = 128;
```
b-sumner, line 1775: I think we need 256 to cover all GPUs.

neoblizz: Done. Bumped to 256 so the counter is isolated on the largest cache line across supported GPUs. Used an alignment attribute on sgs (alignas(256) in device.hpp, __attribute__((aligned(256))) in cg.cl) instead of manual padding, so sgs lands at offset 256 and the struct is 512B.

b-sumner (approved):
LGTM

mangupta (commented):

**[BLOCKING]** MGSyncInfo layout change breaks the host/device-libs `mg_info` contract with no companion change or version guard

`MGSyncInfo` is a layout contract shared with `struct mg_info` in the device library (ockl `cg.cl`), which is built and shipped from a different component. This change both reorders the fields (`sgs` moved from before `num_wg` to after it) and inserts 256-byte alignment padding, moving `sgs` to offset 256 and growing `sizeof(MGSyncInfo)` from tens of bytes to 512. The matching device-libs change is not in this PR, and nothing at build or load time detects a runtime/device-libs mismatch.

**Suggested fix:** Do not merge the runtime half standalone. Either (a) land the ockl `cg.cl` change first and gate this on a minimum device-libs/ROCm version, or (b) carry an explicit compatibility signal the runtime can check (e.g. a layout/version field the device library validates, or a build-time dependency on the updated device-libs package) so a stale device library fails loudly instead of hanging. Reference the required device-libs commit in the PR description and add a `projects/clr/CHANGELOG.md` entry, since a mixed-version install is release-visible and silently breaks cooperative launches.



**[IMPORTANT]** Only the single-grid allocation honors the new 256-byte `MGSyncInfo` alignment; the multi-grid placement stride is not a multiple of it

`alignas(256)` raises `alignof(MGSyncInfo)` to 256 for every `MGSyncInfo` object in the runtime, but only the `singleGridSync` allocation was updated to request that alignment. The multi-grid path, which uses the same `syncInfo` variable and the same type, still obtains its storage from an unmodified site, and the per-device stride it indexes with — `kMGInfoSizePerDevice = kMGSyncDataSize + sizeof(MGSyncInfo)` = 8 + 512 = 520 — is not a multiple of 256. Any `MGSyncInfo` placed at that stride is under-aligned relative to its declared alignment, so the padding buys nothing there (the counter can again share a line with the read-only fields) and the object is accessed at an alignment its type says it does not have.

**Suggested fix:** Apply the same alignment to every site that allocates or places an `MGSyncInfo`: pass `Device::kSGSyncLineSize` to the multi-grid `allocKernArg` as well, and round `kMGInfoSizePerDevice` up to a multiple of `alignof(MGSyncInfo)` (e.g. `amd::alignUp(kMGSyncDataSize + sizeof(MGSyncInfo), kSGSyncLineSize)`) so the strided slots stay aligned. Add a `static_assert(kMGInfoSizePerDevice % alignof(MGSyncInfo) == 0)` so a future size change cannot silently reintroduce the skew.

neoblizz:
> 🤖 AI-assisted pre-review (for @mangupta)
> 
> **[BLOCKING]** MGSyncInfo layout change breaks the host/device-libs `mg_info` contract with no companion change or version guard
> 
> `MGSyncInfo` is a layout contract shared with `struct mg_info` in the device library (ockl `cg.cl`), which is built and shipped from a different component. This change both reorders the fields (`sgs` moved from before `num_wg` to after it) and inserts 256-byte alignment padding, moving `sgs` to offset 256 and growing `sizeof(MGSyncInfo)` from tens of bytes to 512. The matching device-libs change is not in this PR, and nothing at build or load time detects a runtime/device-libs mismatch.
> 
> **Suggested fix:** Do not merge the runtime half standalone. Either (a) land the ockl `cg.cl` change first and gate this on a minimum device-libs/ROCm version, or (b) carry an explicit compatibility signal the runtime can check (e.g. a layout/version field the device library validates, or a build-time dependency on the updated device-libs package) so a stale device library fails loudly instead of hanging. Reference the required device-libs commit in the PR description and add a `projects/clr/CHANGELOG.md` entry, since a mixed-version install is release-visible and silently breaks cooperative launches.

That is why we have this PR; https://github.com/ROCm/llvm-project/pull/3464 also approved and ready to go. They both roughly need to be merged at the same time, looking for guidance on how to make that happen. And the description of the PR also mentions this dependency. 


> **[IMPORTANT]** Only the single-grid allocation honors the new 256-byte `MGSyncInfo` alignment; the multi-grid placement stride is not a multiple of it
> 
> `alignas(256)` raises `alignof(MGSyncInfo)` to 256 for every `MGSyncInfo` object in the runtime, but only the `singleGridSync` allocation was updated to request that alignment. The multi-grid path, which uses the same `syncInfo` variable and the same type, still obtains its storage from an unmodified site, and the per-device stride it indexes with — `kMGInfoSizePerDevice = kMGSyncDataSize + sizeof(MGSyncInfo)` = 8 + 512 = 520 — is not a multiple of 256. Any `MGSyncInfo` placed at that stride is under-aligned relative to its declared alignment, so the padding buys nothing there (the counter can again share a line with the read-only fields) and the object is accessed at an alignment its type says it does not have.
> 
> **Suggested fix:** Apply the same alignment to every site that allocates or places an `MGSyncInfo`: pass `Device::kSGSyncLineSize` to the multi-grid `allocKernArg` as well, and round `kMGInfoSizePerDevice` up to a multiple of `alignof(MGSyncInfo)` (e.g. `amd::alignUp(kMGSyncDataSize + sizeof(MGSyncInfo), kSGSyncLineSize)`) so the strided slots stay aligned. Add a `static_assert(kMGInfoSizePerDevice % alignof(MGSyncInfo) == 0)` so a future size change cannot silently reintroduce the skew.

`multi-grid` is not part of this work, I believe that focuses on multi-GPU stuff.

PR #8821, opened 2026-07-19
[rocjitsu] Rearm recycled doorbell slots

kuhar:
## Motivation

Fix a race in simulated KFD doorbell-slot reuse that could stall the first dispatch on a newly created HSA queue.
This manifested in CDNA4-CDNA3 dbt on simulated target flakes under parallel ctest runs.

## Technical Details

When a queue was destroyed, its doorbell offset was recycled while the slot retained the previous queue’s final value—typically 0. A new HwQueue starts with last_doorbell = ~0, so the command processor could observe and consume a phantom ~0 -> 0 transition before packet submission. The real first submission also rings the doorbell with write index 0, which then appeared unchanged and was never dispatched.

## Issue Tracking

N/A

## Test Plan

ctest with -j16 in a loop

## Test Result

No flakes

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
@@ -16,6 +16,8 @@ RJ_DIAGNOSTIC_POP
 #include "util/log.h"
 
 #include <algorithm>
+#include <atomic>
+#include <cassert>
 #include <cerrno>
 #include <cstdint>
 #include <cstdlib>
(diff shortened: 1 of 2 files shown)


amd-eochoalo (approved):

kuhar:
Closing this as superseded by #8672. That PR landed the recycled-doorbell rearm in SimulatedKfd::create_queue_ioctl, preventing a reused slot from losing the replacement queue’s first write-index-0 dispatch. After rebasing onto develop, this PR has no runtime changes left—only the focused ReusedDoorbellSlotIsRearmedForFirstPacket regression test. The original Triton dynamic-matmul path also passed 320/320 local stress runs at 16-way concurrency using develop’s runtime. If we still want the focused regression test, it can be carried separately.

PR #8822, opened 2026-07-19
fix(rocr): restore optional A0 HotSwap fallback

harsh-amd:
## Motivation

ROCm/rocm-systems#8387 made gfx1250 B0-to-A0 HotSwap rewrites mandatory. COMGR still rejects several valid production code objects, so those pre-existing rewrite failures changed from loading the original object to HSA_STATUS_ERROR_INVALID_CODE_OBJECT. The Jul-17 quick suite consequently regressed in hiptensor, rocPRIM, rocThrust, hipSPARSE, rocSPARSE, rocSOLVER, and hipBLASLt.

A controlled A/B used the same Jul-18 compiler, libraries, tests, GPU, and ROCr revision, changing only the A0 rewrite-required policy. The failing COMGR return codes remained identical, but restoring optional fallback made the affected tests execute successfully.

## Technical Details

- Keep legacy gfx1250 B0-to-A0 retargeting best-effort when COMGR cannot emit a rewrite.
- Fall back to the original code object if an optional A0 rewrite fails or if its rewritten object fails to load.
- Preserve fail-closed behavior for opt-in non-A0 strict-mode rewrites.
- Update decision and runtime-load tests for both A0 failure paths.

This restores the behavior before commit 8b9a6bf3ce. Mandatory A0 enforcement can be reconsidered after COMGR supports the production code-object population or exposes a result that distinguishes a proven-safe no-rewrite case from an unsupported required rewrite.

## Issue Tracking

JIRA ID: ROCM-28260

## Test Plan

- Build libhsa-runtime64 from current develop with this change.
- Build and run hotswap_gfx_query and hotswap_rewrite.
- Run the original affected quick-suite matrix on mi400 GPU 2 with AMD_COMGR_HOTSWAP_ENTRY_TRAMPOLINES=1.
- Verify the exact production cases that previously terminated on COMGR rewrite failure.

## Test Result

- ROCr HotSwap tests: 2/2 passed.
- hiptensor: 108/108 CTest entries passed.
- rocPRIM: 172/172 CTest entries passed.
- rocThrust: 336/336 CTest entries passed.
- hipSPARSE: 4/4 CTest entries passed.
- rocSPARSE: three shorter CTest entries passed; the native quick executable passed 18,124/18,124 tests in 318.9 seconds. Its existing CTest entry has a 300-second timeout.
- rocSOLVER: ffm-quick passed; the native quick executable passed 6,188/6,188 tests in 331.3 seconds. Its existing CTest entry has a 300-second timeout.
- hipBLASLt: 1/1 quick entry passed, with zero out-of-range register diagnostics.
- No GPU fault or hang occurred.

The rocSPARSE and rocSOLVER timeout budgets are separate harness issues exposed because the formerly crashing cases now complete.

## Submission Checklist


shwetagkhatri (approved):
LGTM

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
```diff
@@ -359,7 +359,9 @@ std::optional<RewriteDecision> DecideHotswapRewrite(
         WithGfx1250SteppingFeature(source_isa, Gfx1250Stepping::kB0);
     decision.target_isa =
         WithGfx1250SteppingFeature(target_isa, Gfx1250Stepping::kA0);
-    decision.rewrite_required = true;
+    // Preserve the legacy A0 fallback when COMGR cannot rewrite a code object.
+    // Required behavior remains opt-in through strict mode for non-A0 targets.
+    decision.rewrite_required = false;
```
cfreeamd, line 364: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** This branch unconditionally sets `rewrite_required = false` and returns *before* strict mode is evaluated, so there is no longer any runtime opt-in — not even `HSA_HOTSWAP_STRICT_MODE=1` — to restore the fail-closed mandatory-A0 enforcement that 8b9a6bf3ce added. The comment ("Required behavior remains opt-in through strict mode for non-A0 targets") is accurate but easy to misread as implying strict mode still gates the A0 path.

A consequence: a code object that genuinely needs the A0 patches but that COMGR fails to rewrite will now silently load the un-retargeted B0 object on A0 hardware rather than erroring. The PR description acknowledges this tradeoff (COMGR can't yet distinguish "no rewrite needed" from "rewrite needed but unsupported"), so this is a deliberate, A/B-validated choice — non-blocking.

Worth considering: gate this path on strict mode (`decision.rewrite_required = options.strict_mode_enabled;`) so fail-closed A0 stays available as an explicit opt-in, or tighten the comment to state A0 is intentionally always best-effort for now.

cfreeamd (commented):

Reviewed the ROCr changes. The one-line policy flip in `hotswap.cpp` (`rewrite_required = false` on the gfx1250 B0→A0 branch) is internally consistent with the `RetargetCodeObjectStatus` / `LoadAgentCodeObjectWithHotswap` fallback wiring, and the two renamed tests in `hotswap_rewrite_test.cc` correctly reflect the new best-effort A0 fallback (kOriginal / kRewritten→kOriginal load paths).

Note: the two test changes are renames within the existing `HotswapRewrite` suite — no new `TEST(...)` entries and the suite name is unchanged — so no `test_categories.yaml` / `platform_config.yaml` update is required here.

One optional design note left inline.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap.cpp
harsh-amd: Thanks for the careful read. One clarification on the premise, though: HSA_HOTSWAP_STRICT_MODE was never an opt-in for this branch. Strict mode gates the B0→B0 rewrite path in the second block (request_strict_mode → rewrite_required = request_strict_mode, source and target both pinned to B0 via hotswap_rewrite_with_options()). The branch you’re commenting on is the separate B0→A0 retarget on COMGR’s legacy path — it selects the A0 instruction patches through the ISA feature strings and has always run without consulting strict mode.

So gating the A0 branch on options.strict_mode_enabled would give strict mode a second, unrelated meaning (it would start governing both B0→B0 and B0→A0), which we’d rather not do. The two mechanisms are intentionally orthogonal. This change only affects whether an A0 retarget is best-effort vs. mandatory, and — per the PR description — best-effort is the deliberate choice while COMGR can’t distinguish “no rewrite needed” from “rewrite needed but unsupported.”

Fair point that the inline comment (“Required behavior remains opt-in through strict mode for non-A0 targets”) reads as if strict mode gates the A0 path — it’s meant to say the opposite (strict mode is for the non-A0, B0→B0 path). Noting that for a future wording pass. Keeping the logic as-is.

cfreeamd (approved):

rahulc-gh:
Looks like the windows xFails (since the test pipeline got cancelled, its not treated as xfail ) which is what causing the CI summary to be red . 
 
Talked to Erman will override and land it in rocm-sys

PR #8826, opened 2026-07-19
fix(rocjitsu): initialize gfx1250 clustered launch identity

### LINKED ISSUE #8824: [rocjitsu] gfx1250 cluster rank and workgroup X are initialized incorrectly
A matching ROCm nightly compiler plus the public extended AQL packet launches two gfx1250 workgroups in one `(2,1,1)` cluster. A lane-0 atomic appends each workgroup's compiler-visible `{cluster rank, global workgroup X}` without using either identity to address its own output.

rocJITsu reports:

```text
records=2
record[0] cluster_rank=0 workgroup_x=0
record[1] cluster_rank=0 workgroup_x=2
cluster_rank_mask=0x1 expected=0x3
workgroup_x_mask=0x5 expected=0x3
```

The required unordered records are `{0,0}` and `{1,1}`. Both records being present rules out a missing workgroup or output alias.

The current source mechanisms match the result:

- `gfx1250/sopk.cpp` returns constant zero for `HW_REG_IB_STS2`; matching Clang reads bits `[24:21]` for `__builtin_amdgcn_cluster_workgroup_flat_id()`.
- The command processor writes the full `grid_wg_id_x` to TTMP9 and does not initialize the low TTMP6 field; matching Clang reconstructs X with `s_lshl1_add_u32 ttmp9, (ttmp6 & 15)`, yielding 2 for workgroup 1.

This blocks compiler-authored cluster kernels even though rocJITsu's extended-packet and cluster-load machinery runs. Participant masks, LDS slices, source tiles, and output coordinates can all silently use the wrong identity. The global-X symptom also reaches ordinary compiler-generated multi-workgroup launches.

The complete SDK-only packet includes HIP source, raw public-HSA extended-packet submission, exact output, disassembly guard, source diagnosis, and regression suggestions:

https://gist.github.com/benvanik/5f11b54b134814cd8e698187c0914f13

I scanned the currently open rocJITsu PRs touching launch/dispatch/cluster state before filing. None changes either mechanism; #8672 still documents TTMP9 as the full grid workgroup X, and #8799 is an NFC ownership change for the packet ABI.

benvanik:
## Motivation

Compiler-generated gfx1250 clustered kernels reconstruct global workgroup
coordinates from cluster coordinates in TTMP7/TTMP9 and within-cluster
coordinates in TTMP6. They read the flat within-cluster rank from IB_STS2.
rocJITsu instead populated TTMP7/TTMP9 with ordinary workgroup coordinates,
left TTMP6 unset, and returned zero for IB_STS2. A 2x1x1 cluster consequently
reported `{rank=0, workgroup_x=2}` for its second workgroup instead of
`{rank=1, workgroup_x=1}`.

## Technical Details

- Pack the local x/y/z coordinates and maximum flat local ID into TTMP6, and
  place the cluster x/y/z coordinates in TTMP9/TTMP7.
- Return `Wavefront::cluster_rank()` through the gfx1250 IB_STS2 hardware
  register implementation and keep the generated ISA source in sync.
- Exercise the actual `s_getreg_b32` path over a 4x4x4 workgroup grid formed
  from 2x2x2 clusters, checking every TTMP field and all ranks 0 through 7.

## Issue Tracking

Fixes #8824.

## Test Plan

- Run the complete rocJITsu CTest suite.
- Run the public standalone ROCm SDK reproducer against the resulting shared
  library.

## Test Result

- All 1,558 registered rocJITsu tests passed; one race-hook test was
  intentionally skipped by its fixture.
- The standalone reproducer reports ranks 0/1 and workgroup X coordinates 0/1
  for the two workgroups in a 2x1x1 cluster.

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/sopk.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/sopk.cpp
@@ -29,6 +29,8 @@ constexpr uint32_t HW_REG_GPR_ALLOC = 6;
 constexpr uint32_t HW_REG_VGPR_ALLOC = 7;
 constexpr uint32_t HW_REG_WAVE_SCHED_MODE = 26;
 constexpr uint32_t HW_REG_IB_STS2 = 28;
+constexpr uint32_t HW_REG_IB_STS2_WG_IN_CLUSTER_SHIFT = 21;
+constexpr uint32_t HW_REG_IB_STS2_WG_IN_CLUSTER_MASK = 0xFu;
 
 [[maybe_unused]] uint32_t insert_hwreg_field(uint32_t reg_val, uint32_t src, uint32_t offset,
                                              uint32_t mask) {
@@ -59,7 +61,8 @@ constexpr uint32_t HW_REG_IB_STS2 = 28;
     reg_val = wf.wave_sched_mode_raw();
     return true;
   case HW_REG_IB_STS2:
-    reg_val = 0;
+    reg_val = (wf.cluster_rank() & HW_REG_IB_STS2_WG_IN_CLUSTER_MASK)
+              << HW_REG_IB_STS2_WG_IN_CLUSTER_SHIFT;
     return true;
   default:
     return false;
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
@@ -44,6 +44,17 @@ namespace {
 constexpr uint32_t kMaxClusterWorkgroups = kClusterMulticastMaskBits;
 static_assert(kMaxClusterWorkgroups <= kClusterMulticastMaskBits);
 
+// GFX12 launch-state scalar selectors used by compiler-generated workgroup
+// and cluster identity sequences.
+constexpr uint32_t kGfx12Ttmp6 = 114;
+constexpr uint32_t kGfx12Ttmp7 = 115;
+constexpr uint32_t kGfx12Ttmp9 = 117;
+
+constexpr uint32_t kGfx12Ttmp6ClusterLocalXShift = 0;
+constexpr uint32_t kGfx12Ttmp6ClusterLocalYShift = 4;
+constexpr uint32_t kGfx12Ttmp6ClusterLocalZShift = 8;
+constexpr uint32_t kGfx12Ttmp6ClusterMaxFlatIdShift = 24;
+
 struct PlannedWorkgroup {
   uint32_t local_wg_id = 0;
   uint32_t global_wg_id = 0;
@@ -285,16 +296,30 @@ void CommandProcessor::init_wavefront_regs(ComputeUnitCore *cu, Wavefront *wf,
   }
 
   if (cu->arch() == ROCJITSU_CODE_ARCH_GFX1250 || cu->arch() == ROCJITSU_CODE_ARCH_RDNA4) {
-    constexpr uint32_t ttmp7 = 115;
-    constexpr uint32_t ttmp9 = 117;
     // The simulator aliases TTMP scalar selectors into the wavefront SGPR
     // block, so the block must include slots through TTMP9.
-    if (cu->config().sgprs_per_wf <= ttmp9) {
+    if (cu->config().sgprs_per_wf <= kGfx12Ttmp9) {
       throw std::runtime_error("RDNA4/gfx1250 TTMP launch payload requires at least 118 SGPR "
                                "slots per wavefront");
     }
-    cu->write_sgpr(sbase + ttmp7, ((wg_id_z & 0xFFFFu) << 16) | (wg_id_y & 0xFFFFu));
-    cu->write_sgpr(sbase + ttmp9, grid_wg_id_x);
+
+    const uint32_t cluster_size_x = nonzero_or_one(pkt.cluster_size_x);
+    const uint32_t cluster_size_y = nonzero_or_one(pkt.cluster_size_y);
+    const uint32_t cluster_size_z = nonzero_or_one(pkt.cluster_size_z);
+    const uint32_t cluster_local_x = grid_wg_id_x % cluster_size_x;
+    const uint32_t cluster_local_y = wg_id_y % cluster_size_y;
+    const uint32_t cluster_local_z = wg_id_z % cluster_size_z;
+    const uint32_t cluster_max_flat_id = cluster_size_x * cluster_size_y * cluster_size_z - 1;
+
+    const uint32_t ttmp6 = (cluster_local_x << kGfx12Ttmp6ClusterLocalXShift) |
+                           (cluster_local_y << kGfx12Ttmp6ClusterLocalYShift) |
+                           (cluster_local_z << kGfx12Ttmp6ClusterLocalZShift) |
+                           (cluster_max_flat_id << kGfx12Ttmp6ClusterMaxFlatIdShift);
+    const uint32_t ttmp7 = ((wg_id_z / cluster_size_z) << 16) | (wg_id_y / cluster_size_y);
+    const uint32_t ttmp9 = grid_wg_id_x / cluster_size_x;
+    cu->write_sgpr(sbase + kGfx12Ttmp6, ttmp6);
+    cu->write_sgpr(sbase + kGfx12Ttmp7, ttmp7);
+    cu->write_sgpr(sbase + kGfx12Ttmp9, ttmp9);
   }
 
   // Workitem IDs per AMDHSA ABI. The SPI decomposes the flat thread index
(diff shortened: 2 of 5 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
```diff
@@ -285,16 +296,30 @@ void CommandProcessor::init_wavefront_regs(ComputeUnitCore *cu, Wavefront *wf,
...
+    const uint32_t cluster_local_x = grid_wg_id_x % cluster_size_x;
+    const uint32_t cluster_local_y = wg_id_y % cluster_size_y;
+    const uint32_t cluster_local_z = wg_id_z % cluster_size_z;
+    const uint32_t cluster_max_flat_id = cluster_size_x * cluster_size_y * cluster_size_z - 1;
+
+    const uint32_t ttmp6 = (cluster_local_x << kGfx12Ttmp6ClusterLocalXShift) |
+                           (cluster_local_y << kGfx12Ttmp6ClusterLocalYShift) |
+                           (cluster_local_z << kGfx12Ttmp6ClusterLocalZShift) |
+                           (cluster_max_flat_id << kGfx12Ttmp6ClusterMaxFlatIdShift);
+    const uint32_t ttmp7 = ((wg_id_z / cluster_size_z) << 16) | (wg_id_y / cluster_size_y);
```
kuhar, line 318: We should preserve the 16-bit TTMP7 field boundaries when packing the cluster-grid Y/Z IDs. The old code masked both halves, but the new Y quotient is ORed into the full word; for example, Y `0x10000` with Z zero produces `0x00010000`, which decodes as Z=1. Please mask both quotients with `0xFFFFu` before combining them.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
```diff
@@ -285,16 +296,30 @@ void CommandProcessor::init_wavefront_regs(ComputeUnitCore *cu, Wavefront *wf,
...
+
+    const uint32_t cluster_size_x = nonzero_or_one(pkt.cluster_size_x);
+    const uint32_t cluster_size_y = nonzero_or_one(pkt.cluster_size_y);
+    const uint32_t cluster_size_z = nonzero_or_one(pkt.cluster_size_z);
+    const uint32_t cluster_local_x = grid_wg_id_x % cluster_size_x;
+    const uint32_t cluster_local_y = wg_id_y % cluster_size_y;
+    const uint32_t cluster_local_z = wg_id_z % cluster_size_z;
+    const uint32_t cluster_max_flat_id = cluster_size_x * cluster_size_y * cluster_size_z - 1;
+
+    const uint32_t ttmp6 = (cluster_local_x << kGfx12Ttmp6ClusterLocalXShift) |
```
kuhar, line 314: We should populate the per-axis cluster maxima in TTMP6 bits 12, 16, and 20. LLVM reads those fields as `CLUSTER_WORKGROUP_MAX_ID_{X,Y,Z}` and computes each global workgroup ID as `cluster_id * (max + 1) + cluster_local_id`. With a 2x2x2 launch the current `0x07000xyz` payload leaves every maximum at zero, so compiler-generated workgroup IDs collapse clusters onto one another. Please encode `cluster_size_{x,y,z} - 1` (adding `0x00111000` for this test case) and cover the compiler-generated workgroup-ID sequence.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/sopk.cpp
```diff
@@ -59,7 +61,8 @@ constexpr uint32_t HW_REG_IB_STS2 = 28;
     reg_val = wf.wave_sched_mode_raw();
     return true;
   case HW_REG_IB_STS2:
-    reg_val = 0;
+    reg_val = (wf.cluster_rank() & HW_REG_IB_STS2_WG_IN_CLUSTER_MASK)
```
kuhar, line 64: We also need to model `IB_STS2.CLUSTER_ID` (bits 6-9), not just `WG_IN_CLUSTER`. When cluster dimensions are unknown, LLVM reads that field and treats zero as a non-clustered launch, selecting TTMP7/TTMP9 directly. Because `s_getreg` currently returns zero there, variable-dimension clustered kernels report cluster coordinates instead of global workgroup coordinates. Please set a nonzero architectural cluster ID for clustered waves and add an end-to-end workgroup-ID test without fixed `amdgpu-cluster-dims`.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
```diff
@@ -285,16 +296,30 @@ void CommandProcessor::init_wavefront_regs(ComputeUnitCore *cu, Wavefront *wf,
...
       throw std::runtime_error("RDNA4/gfx1250 TTMP launch payload requires at least 118 SGPR "
                                "slots per wavefront");
     }
-    cu->write_sgpr(sbase + ttmp7, ((wg_id_z & 0xFFFFu) << 16) | (wg_id_y & 0xFFFFu));
-    cu->write_sgpr(sbase + ttmp9, grid_wg_id_x);
+
+    const uint32_t cluster_size_x = nonzero_or_one(pkt.cluster_size_x);
+    const uint32_t cluster_size_y = nonzero_or_one(pkt.cluster_size_y);
+    const uint32_t cluster_size_z = nonzero_or_one(pkt.cluster_size_z);
+    const uint32_t cluster_local_x = grid_wg_id_x % cluster_size_x;
```
kuhar, line 309: This would be safer if TTMP6 and `IB_STS2.WG_IN_CLUSTER` derived their cluster-local identity from the same coordinate. TTMP6 decomposes offset-inclusive `global_wg_id`, while the rank uses `cluster_rank_for_local_wg(local_wg_id)` and ignores `workgroup_id_offset`; the values can disagree whenever that offset is not cluster-aligned. Please either share the decomposition or reject unaligned clustered offsets, and cover a nonzero-offset case.

kuhar (commented):
A few comments

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/gfx1250/sopk.cpp
kuhar: This also has to be a generator-side fix, specifically around rocm-systems/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py:2853.

benvanik:
As discussed, you can take this over. I have stopped work on #8826 and
preserved the local prototype and validation evidence as a technical handoff.

I locally prototyped all five review follow-ups and preserved the diff as
reference material:
https://gist.github.com/benvanik/d49b8533dde92188a7e56acb6889498c

The useful findings are:

- TTMP6 needs the per-axis `(cluster_size - 1)` fields in bits 12, 16, and 20,
  in addition to the flat maximum in bits 24-27.
- Both TTMP7 cluster-grid quotients need independent 16-bit masks.
- Dynamic-dimension LLVM 23 output reads `IB_STS2[9:6]`, compares it with
  zero, and selects between raw TTMP7/TTMP9 coordinates and coordinates
  reconstructed from TTMP6. Setting a nonzero `CLUSTER_ID` for clustered waves
  makes that path work; `WG_IN_CLUSTER` remains in bits 21-24.
- TTMP decomposition and `WG_IN_CLUSTER` must use the same offset-inclusive
  flat workgroup ID. The prototype permits a nonzero offset only when it is a
  multiple of the rank sequence period: `size_x` for X-only clusters,
  `grid_x * size_y` when Y participates, or
  `grid_x * grid_y * size_z` when Z participates. It rejects other clustered
  offsets rather than publishing contradictory identities.
- A cluster axis greater than 16 cannot fit the four-bit TTMP maximum field;
  the prototype fails that launch explicitly.

The main regression embeds the exact scalar identity sequence emitted by
LLVM 23 without `amdgpu-cluster-dims`. It launches a 4x4x4 grid with 2x2x2
clusters and aligned nonzero workgroup offset 32, then checks TTMP6/7/9,
`CLUSTER_ID`, flat rank, and reconstructed global X/Y/Z for all 64 workgroups.
Separate negative tests cover an axis of 17 and a misaligned offset.

Local evidence for the prototype:

- generator tests: 117 passed, 1 skipped;
- focused cluster ABI tests: 6/6 under ASAN+UBSAN;
- all gfx1250 tests: 175/175 under ASAN+UBSAN;
- all built in-process core tests: 1,450/1,450 under ASAN+UBSAN;
- the three Release buffer-async DBT cases that failed old CI passed serially,
  then passed 30/30 across ten concurrent rounds against the July 18 SDK.

The failed existing Release job used SDK `7.14.0a20260624`; all three tests
started together at the tail of the suite and hit the same dispatch-completion
failure. That looks separate from this gfx1250 launch work, but a fresh run of
the replacement PR will give the clean answer.

The gist includes the generated `sopk.cpp` only to show the intended behavior.
The generator source of truth remains authoritative.

kuhar:
Closing as superseded by https://github.com/ROCm/rocm-systems/pull/8831

PR #8828, opened 2026-07-19
[rocjitsu] Fix remaining decoder metadata issues

kuhar:
## Motivation

Port the remaining decoder and def/use fixes discovered while validating waitcheck kernels, without bringing the waitcheck analyzer itself onto develop. Discovered by running waitcheck on pytorch kernels.

## Technical Details

- use the semantic GFX12 NULL SADDR value when deriving flat/global VADDR width
- model RDNA4/gfx1250 VOP3 compare, cndmask, and carry masks as wave32 SGPRs
- make atomic source/destination metadata follow payload and return-control semantics
- decode CDNA4 F8/F6/F4 MFMA suffixes both standalone and behind VOP3PX2 prefixes
- retain gfx1250 high-bank VGPRs in ISA-independent RegisterSet analysis
- regenerate all ISA sources

## Issue Tracking

N/A

## Test Plan

- run-all-isa-generation-build.sh
- amdisa Python tests
- focused decoder and liveness CTests
- full Clang and GCC CTest suites excluding Rocblas and Rccl

## Test Result

OK

## Submission Checklist


atgutier (commented):
MAJOR — latent inconsistency this PR half-fixes (worth finishing while here)

  _generator.py:2229, :6342, :6352 — the operand-suppression logic still uses the raw literal 0x7F to decide whether to add saddr as an operand:
  'if (inst_.saddr == 0x7F) return;'        # :2229
  '  if (inst_.saddr != 0x7F) {'            # :6342 (SCRATCH)
  '} else if (inst_.seg == 2 && inst_.saddr != 0x7F) {'  # :6352 (GLOBAL)
  After this PR, VADDR-width and execution both key off 124, but operand-suppression still keys off 127. They now disagree: an encoding with saddr==124 (the
   real NULL) is treated as no-SADDR by execution and the width helper, yet the suppression path adds a spurious saddr src operand — corrupting def/use
  metadata (the exact class of bug this PR targets). Because the PR only fixed one of the two sentinel sites, the subsystem is left internally inconsistent.

  - Evidence: execution has_saddr = != 124; width helper (post-PR) = == 124; suppression = != 127. Three sites, two different sentinels.
  - Suggested fix: replace the 0x7F literals in the suppression paths with the same OPR_SREG_NULL predicate, and add a decode/liveness test asserting
  saddr==124 produces no saddr operand and a 64-bit VADDR. Note: this is pre-existing, not a regression introduced here — but it's the same metadata
  subsystem the PR is fixing, so finishing it now is the right call.

  MINOR

  1. _generator.py:6303 — inst_sem.elem_size * 8 for the cmpswap vdata_return view has no None guard, while every sibling atomic path uses sem.elem_size or
  4. elem_size is int | None; a cmpswap classified with None would raise TypeError at generation time. Fix: (inst_sem.elem_size or 4) * 8.
  2. _generator.py:~2349 — dead entries 'vector_cmpx'/'vector_cmpx_class' in the _operand_size_override tuple: the override only fires on OPR_SREG
  destinations, but cmpx always writes OPR_EXEC, so those members can never match. Harmless but misleading; drop them or add a comment that cmpx writes EXEC
   (handled separately).
  3. shared/simd_glue.h (read_wave_mask_scalar, ~L243) — new hand-maintained helper is missing the required @brief/@details Doxygen. Add a brief noting it
  reads one SGPR for Wave32, the SGPR pair for Wave64.

  NIT

  - shared/execute_shared.h — carry-out writes are open-coded (if (wf.wf_size() <= 32) write_scalar else write_scalar64) while the SIMD path uses the
  write_wave_mask_scalar helper. Functionally identical; using the helper would remove the duplicated wave-size branch.
  - CDNA4 MFMA F8/F6/F4 source operands (cdna4/vop3p.cpp) are declared at a fixed 256-bit (8-reg) width regardless of F8/F6/F4 packing. This over-declares
  src uses for F6/F4 — conservative for liveness (safe) and pre-existing (matches develop's prefixed sibling), but the codegen could tighten it if the
  format selector is available at decode time.

kuhar:
@atgutier thanks, should be addressed now

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -2282,6 +2283,19 @@ def _encoding_implicit_uses_impl(
...
         return ''
 
+    def _saddr_null_expr(self) -> str:
+        """Return the generated selector spelling for a disabled SADDR.
+
+        RDNA operand tables expose the architectural selector as
+        OPR_SREG_NULL. Older CDNA tables omit that enum entry even though
+        their FLAT encoding still reserves the all-ones SADDR selector.
+        """
+        has_sreg_null = any(
```
lua1235: This cannot be selected based on whether OPR_SREG_NULL appears anywhere in the ISA operand tables. RDNA3 ENC_FLAT (flat_load/store) still recognizes all-ones 7-bit SADDR value (0x7f) for NULL. 
Please make this encoding/profile-specific

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/register_set.h
```diff
@@ -28,8 +29,7 @@ namespace rocjitsu {
 /// number of SGPR/VGPR indices addressable within a wavefront register file.
 inline constexpr size_t REGISTER_SET_MAX_SGPRS =
     std::max<size_t>(amdgpu::CdnaIsaBase::MAX_SGPRS_PER_WF, amdgpu::RdnaIsaBase::MAX_SGPRS_PER_WF);
-inline constexpr size_t REGISTER_SET_MAX_VGPRS =
-    std::max<size_t>(amdgpu::CdnaIsaBase::MAX_VGPRS_PER_WF, amdgpu::RdnaIsaBase::MAX_VGPRS_PER_WF);
+inline constexpr size_t REGISTER_SET_MAX_VGPRS = MAX_SUPPORTED_ADDRESSABLE_VGPRS_PER_WF;
```
lua1235, line 32: This constant is also used as the allocation ceiling by LivenessAnalysis::find_free_run(). Raising it to MAX_SUPPORTED_ADDRESSABLE_VGPRS_PER_WF = 1024 (gfx1250 limit) fallows that function to return v256+ to RDNA4/CDNA3 lowering paths that narrow the result to uint8_t; for example, cdna4_to_rdna4.cpp can turn v256 into v0 before both emission. Please separate the 1024-register analysis storage capacity from the destination-specific allocation limit.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -2340,6 +2354,44 @@ def _dst_is_also_source(self, inst: Instruction) -> bool:
...
+            and enc_name.upper() == 'ENC_VOP3'
+            and not opnd.is_output
+            and opnd.name == 'src2'
+            and opnd.operand_type.upper() == 'OPR_SREG'
+            and self.isa_spec.profile.vop3_cndmask_selector_size_bits is not None
+        ):
+            return str(self.isa_spec.profile.vop3_cndmask_selector_size_bits)
+        if (
+            sem is not None
+            and sem.semantic_class == 'vector_add_co'
```
lua1235: I believe that `vector_div_scale` and `vector_mad_64_32` also need to be accounted for in this function.

lua1235 (commented):
Some comments

lua1235 (approved):

PR #8831, opened 2026-07-19
[roctjisu] Fix cluster launch identity

### LINKED ISSUE #8824: [rocjitsu] gfx1250 cluster rank and workgroup X are initialized incorrectly

kuhar:
## Motivation

Compiler-generated gfx1250 clustered kernels reconstruct global workgroup
coordinates from cluster coordinates in TTMP7/TTMP9 and within-cluster
coordinates in TTMP6. They read the flat within-cluster rank from IB_STS2.
rocJITsu instead populated TTMP7/TTMP9 with ordinary workgroup coordinates,
left TTMP6 unset, and returned zero for IB_STS2. A 2x1x1 cluster consequently
reported `{rank=0, workgroup_x=2}` for its second workgroup instead of
`{rank=1, workgroup_x=1}`.

This is a fixup to https://github.com/ROCm/rocm-systems/pull/8826 by Ben with the review comments resolved.

## Technical Details

- Pack the local x/y/z coordinates and maximum flat local ID into TTMP6, and
  place the cluster x/y/z coordinates in TTMP9/TTMP7.
- Return `Wavefront::cluster_rank()` through the gfx1250 IB_STS2 hardware
  register implementation and keep the generated ISA source in sync.
- Exercise the actual `s_getreg_b32` path over a 4x4x4 workgroup grid formed
  from 2x2x2 clusters, checking every TTMP field and all ranks 0 through 7.

## Issue Tracking

Fixes #8824.

## Test Plan

- Run the complete rocJITsu CTest suite.
- Run the public standalone ROCm SDK reproducer against the resulting shared
  library.

## Test Result

- All 1,558 registered rocJITsu tests passed; one race-hook test was
  intentionally skipped by its fixture.
- The standalone reproducer reports ranks 0/1 and workgroup X coordinates 0/1
  for the two workgroups in a 2x1x1 cluster.

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
@@ -42,6 +42,8 @@ namespace {
 // The supported cluster size must fit the M0 multicast mask captured at issue time.
 constexpr uint32_t kMaxClusterWorkgroups = kClusterMulticastMaskBits;
 static_assert(kMaxClusterWorkgroups <= kClusterMulticastMaskBits);
+static_assert(kMaxClusterWorkgroups <= 16,
+              "TTMP6 cluster max and max-flat-ID fields are 4 bits wide");
 
 // GFX12 launch-state scalar selectors used by compiler-generated workgroup
 // and cluster identity sequences.
@@ -49,10 +51,17 @@ constexpr uint32_t kGfx12Ttmp6 = 114;
 constexpr uint32_t kGfx12Ttmp7 = 115;
 constexpr uint32_t kGfx12Ttmp9 = 117;
 
+// LLVM's gfx1250 architected-SGPR ABI maps TTMP6 as seven 4-bit fields:
+// cluster-local XYZ, cluster-max XYZ, and max-flat-ID from low to high bits.
+// TTMP7 holds 16-bit cluster-grid Y/Z IDs, while TTMP9 holds cluster-grid X.
 constexpr uint32_t kGfx12Ttmp6ClusterLocalXShift = 0;
 constexpr uint32_t kGfx12Ttmp6ClusterLocalYShift = 4;
 constexpr uint32_t kGfx12Ttmp6ClusterLocalZShift = 8;
+constexpr uint32_t kGfx12Ttmp6ClusterMaxXShift = 12;
+constexpr uint32_t kGfx12Ttmp6ClusterMaxYShift = 16;
+constexpr uint32_t kGfx12Ttmp6ClusterMaxZShift = 20;
 constexpr uint32_t kGfx12Ttmp6ClusterMaxFlatIdShift = 24;
+constexpr uint32_t kGfx12Ttmp7ClusterGridDimensionMask = 0xFFFFu;
 
 struct PlannedWorkgroup {
   uint32_t local_wg_id = 0;
@@ -86,6 +95,7 @@ void validate_cluster_shape(const DispatchEntry &dp) {
     return;
   auto cluster_size =
       static_cast<uint64_t>(dp.cluster_size_x) * dp.cluster_size_y * dp.cluster_size_z;
+  // This also keeps every TTMP6 cluster dimension/max field within 4 bits.
   if (cluster_size == 0 || cluster_size > kMaxClusterWorkgroups) {
     throw std::runtime_error(
         std::format("unsupported workgroup cluster size {}x{}x{} ({} workgroups)",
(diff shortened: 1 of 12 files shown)


benvanik:
Confirmed from downstream dogfooding: the current PR head looks good and
satisfies #8824 for us.

We built this PR against the public ROCm SDK nightly and ran both the standalone
SDK identity reducer and our production path (`Loom -> gfx1250 HSACO -> IREE
AMDGPU HAL -> extended AQL -> rocJITsu`). Cluster ranks and compiler-visible
workgroup IDs are now correct. More importantly, the real ID4 short-QKV shape
(`BF16[128,4608] x scaled-FP8-E4M3[4608,13824] -> BF16[128,13824]`) launched
108 workgroups as 54 `(1,2,1)` clusters, executed 144 K32 iterations per
workgroup, and produced all 1,769,472 BF16 outputs exactly. Seven production-HAL
cases pass, including a nonuniform K64 host-reference case, a BF16
scale-rounding discriminator, and ordinary fallback controls.

The compiler results are neat too:

- Cluster multicast halves duplicated LHS staging from 121.500 MiB to 60.750
  MiB and reduces modeled global payload from 185.678 MiB to 124.928 MiB
  (32.718% before cache effects).
- The generated gfx1250 K32 schedule uses 289 dynamic global loads and 2,304
  WMMA issues per workitem, versus 865 and 4,608 for the gfx1100 K16 schedule,
  at the same 25% modeled occupancy and with zero spills.

The full source, bound HSACO, disassembly, compiler/resource report, seven-case
results, and cross-target A/B are here:
https://gist.github.com/benvanik/7d00c5de52e3383b55aaa4f1043f85c2

No gfx1250 hardware was present, so this is functional pre-silicon evidence
rather than a silicon performance claim. From our side, however, the
ABI/identity behavior fixed here is now doing useful nontrivial work end to end.
Thanks!

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
```diff
@@ -285,18 +313,51 @@ void CommandProcessor::init_wavefront_regs(ComputeUnitCore *cu, Wavefront *wf,
...
-    cu->write_sgpr(sbase + ttmp9, grid_wg_id_x);
+
+    // RDNA4 does not use the gfx1250 cluster-launch TTMP ABI. Extended
+    // dispatches still use generic cluster placement, but retain the ordinary
+    // workgroup coordinates in TTMP6/7/9.
+    uint32_t ttmp6 = 0;
+    uint32_t ttmp7 = ((wg_id_z & kGfx12Ttmp7ClusterGridDimensionMask) << 16) |
+                     (wg_id_y & kGfx12Ttmp7ClusterGridDimensionMask);
+    uint32_t ttmp9 = grid_wg_id_x;
+    if (cu->arch() == ROCJITSU_CODE_ARCH_GFX1250) {
```
amd-eochoalo: I think this should be driven from isa_properties, but I can change that later.

amd-eochoalo (approved):

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
kuhar: Good point, will fix

PR #8840, opened 2026-07-20
feat(hrr): [AIRUNTIME-2531] replay per-thread-default-stream memset APIs

andraghetti:
## Motivation

The per-thread-default-stream memset APIs `hipMemset_spt`, `hipMemsetAsync_spt`, `hipMemset2D_spt` and `hipMemset2DAsync_spt` were no-op on replay (`NOOP_PLAYBACK`), so their destination buffer was never reproduced during playback. They are the `_spt` siblings of the already-faithful `hipMemset*` handlers.

## Technical Details

Replay-only fix. Capture already records every field these calls need and a memset carries no host blob, so `hip_capture_generated.cpp` and `hrr_api_args.h` are unchanged.

- Removes the four APIs from `NOOP_PLAYBACK_APIS` in `gen_hrr_api_args.py` and regenerates. The handler now calls the real API with `translate_ptr` on the destination and `translate_stream` on the async variants, identical in shape to the proven non-`_spt` handlers.
- It calls the real `_spt` entry points rather than aliasing to the legacy ones. In `hip_memory.cpp` the `_spt` variants submit on `getPerThreadDefaultStream()`, while `hipMemset` / `hipMemset2D` submit on the legacy null stream, which carries different implicit-synchronisation semantics. Aliasing would replay the work on a different stream than the recording used.

Tests, in `hrr.yaml`, `hrr_spt_workload_test.cc` and `hrr_roundtrip_test.cc`:

- `Unit_HRR_MemsetSpt_Direct`: each variant fills its own buffer with its own byte pattern and each buffer gets its own D2H, so every API has an independent oracle. One shared buffer would leave only the last writer observable.
- `Unit_HRR_MemsetSptRoundtrip`: replays under `HIP_HRR_D2H_EXACT=1` and requires a clean exit plus four validated D2H buffers. The patterns (`0x41` / `0x42` / `0x44` / `0x4C`) are also out of tolerance in every encoding the default validator tries, so the workload stays falsifiable without exact mode.

Coverage note: `Unit_HRR_MemsetExtra_Direct` already called all four `_spt` memsets but freed those buffers before its only D2H, so that coverage was capture-only. This PR upgrades it to replay coverage rather than adding first coverage.

**Merge order.** #8849 and #9287 also edit `hrr_workload_test.cc` around `Unit_HRR_MemsetVariants_Direct`, and #8849 adds its roundtrip at the same anchor in `hrr_roundtrip_test.cc`. Textual conflict, not semantic.

## Rework

`_spt` is not an API surface applications call by name: it is what the ordinary memset entry points resolve to when a translation unit is built with `-fgpu-default-stream=per-thread`. The workload therefore moves into its own translation unit compiled with that option and calls the ordinary names, which is also why it cannot stay in `hrr_workload_test.cc`: the option is set per source, and the other workloads must keep resolving to the plain entry points. The async cases now pass a null stream, since `PER_THREAD_DEFAULT_STREAM` substitutes only a null or legacy stream and the previously created stream made those calls identical to the plain variants. `hipMemcpy` is un-redirected inside that unit because the generated `hipMemcpy_spt` shim records no data blob, and the roundtrip now asserts the recorded API ids so that losing either the compile option or the un-redirect fails the test rather than quietly weakening it.

## Issue Tracking

JIRA ID: AIRUNTIME-2531

## Test Plan

Build `libamdhip64`, `hrr-playback` and the `HrrTest` Catch2 target on a gfx950 host, then run `Unit_HRR_MemsetSpt_Direct`, `Unit_HRR_MemsetSptRoundtrip`, and the rest of the HRR unit suite as a regression pass.

## Test Result

Verified on a gfx950 host at this revision: the full HRR unit suite passes, GPU-free pass and GPU pass both clean, 44 round-trip tests green in 42.5 s. `Unit_HRR_MemsetSptRoundtrip` passes, and `Unit_HRR_MemcpyPeerRoundtrip` is skipped as it needs a second GPU.

Also checked statically: the 2-D geometry covers each validated buffer exactly (`pitch * height == 1024`, the allocation size), so no untouched padding reaches a validated blob. Against a port of `hrr_d2h_validate()`, all four patterns put an all-zero replay buffer out of tolerance in every candidate encoding, where the pre-rework `0x2A` pattern passed in default mode.

## Submission Checklist


projects/hip-tests/catch/unit/hrr/hrr_workload_test.cc
```diff
@@ -936,6 +936,87 @@ TEST_CASE("Unit_HRR_MemsetVariants_Direct", "[.][hrr-direct]") {
...
+  int* d_set          = nullptr;
+  int* d_set_async    = nullptr;
+  int* d_set2d        = nullptr;
+  int* d_set2d_async  = nullptr;
+  HIP_CHECK(hipMalloc(&d_set, SZ));
+  HIP_CHECK(hipMalloc(&d_set_async, SZ));
+  HIP_CHECK(hipMalloc(&d_set2d, SZ));
+  HIP_CHECK(hipMalloc(&d_set2d_async, SZ));
+
+  HIP_CHECK(hipMemset_spt(d_set, 0x41, SZ));
```
gandryey: I don't think we should have _spt explicit calls in the test

gandryey (changes requested):

projects/hip-tests/catch/unit/hrr/hrr_workload_test.cc
andraghetti: Covered by 455c72ac: the workload moved into its own translation unit, `hrr_spt_workload_test.cc`, which calls the ordinary `hipMemset*` names and lets `-fgpu-default-stream=per-thread` on that one source select the `_spt` entry points. It needs to be a separate translation unit because the option is per source. `hrr_workload_test.cc` is no longer touched by this PR.

We had written it with explicit calls because that is the existing convention: `hrr_workload_test.cc` on `develop` already contains 28 explicit `_spt` calls. What do you think about those?

gandryey (approved):

PR #8844, opened 2026-07-20
fix(rocprofiler-sdk): validate Perfetto buffer size limits

srihari-uttanur:
## Motivation

The --perfetto-buffer-size argument accepted a unrestricted python integer which was later converted to C++ uint32_t causing overflow/underflows for out of bound integer values like -5,0,4194304 which caused hang in the profiler.

This PR adds proper error handling when such arguments are being passed.

Perfetto represents BufferConfig::size_kb as uint32_t, but the tracing service converts the configured value to bytes and requires the resulting buffer size to fit in uint32_t. This limits the accepted range to:

    1 <= size_kb <= floor(UINT32_MAX / 1024) = 4194303

## Technical Details

Added consistent validation across the supported configuration paths:

- validate --perfetto-buffer-size during argparse processing
- validate values loaded from YAML and JSON after argument merging
- validate parsed ROCPROF_PERFETTO_BUFFER_SIZE_KB values in C++
- guard both the rocprofv3 and rocpd Perfetto emitters
- explicitly narrow to uint32_t only after successful validation

## Issue Tracking

JIRA ID: AIPROFSDK-992

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-sdk/source/bin/rocprofv3.py
+++ b/projects/rocprofiler-sdk/source/bin/rocprofv3.py
@@ -42,6 +42,16 @@ CONST_VERSION_INFO = {
     "rocm_version": "@rocm_version_FULL_VERSION@",
 }
 
+# Perfetto's TraceConfig BufferConfig.size_kb field is a uint32_t, and the
+# tracing service allocates size_kb * 1024 bytes and rejects the config when
+# that byte count does not fit in a uint32_t. So although the option is named
+# in "KB", the value is treated as KiB (multiplied by 1024). The usable range
+# is 1 KiB up to floor((2^32 - 1) / 1024) = 4,194,303 KiB. These bounds mirror
+# the C++ limits (defaults::perfetto_buffer_size_{min,max}_kb in
+# source/lib/output/output_config.hpp) so both validation paths agree.
+PERFETTO_BUFFER_SIZE_KB_MIN = 1
+PERFETTO_BUFFER_SIZE_KB_MAX = ((1 << 32) - 1) // 1024
+
 
 class dotdict(dict):
     """dot.notation access to dictionary attributes"""
@@ -122,6 +132,24 @@ def strtobool(val):
     else:
         val_type = type(val).__name__
         raise ValueError(f"invalid truth value {val} (type={val_type})")
+
+
+def perfetto_buffer_size_kb(val):
+    if isinstance(val, bool) or not isinstance(val, (int, str)):
+        raise argparse.ArgumentTypeError(f"invalid integer value: {val}")
+
+    try:
+        value = int(val)
+    except (TypeError, ValueError):
+        raise argparse.ArgumentTypeError(f"invalid integer value: {val}")
+
+    if not PERFETTO_BUFFER_SIZE_KB_MIN <= value <= PERFETTO_BUFFER_SIZE_KB_MAX:
+        raise argparse.ArgumentTypeError(
+            f"must be between {PERFETTO_BUFFER_SIZE_KB_MIN} and "
+            f"{PERFETTO_BUFFER_SIZE_KB_MAX} KB"
+        )
+
+    return value
 
 
 def get_mpi_rank_and_size(custom_rank_env=None, custom_size_env=None):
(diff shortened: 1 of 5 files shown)


projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/att_no_intercept_quick_scan.cpp
```diff
@@ -362,10 +362,10 @@ backend_code_object_load(agent_state_t&
 {
     if(data.storage_type == ROCPROFILER_CODE_OBJECT_STORAGE_TYPE_MEMORY)
     {
-        record_code_object_symbols(state,
-                                   data.code_object_id,
-                                   reinterpret_cast<const void*>(data.memory_base),
-                                   data.memory_size);
+        const void* memory_base = nullptr;
```
bgopesh: Dont understand this change. Seems unreated

projects/rocprofiler-sdk/source/lib/output/generatePerfetto.cpp
```diff
@@ -101,8 +101,13 @@ write_perfetto(
     auto shmem_size_hint = ocfg.perfetto_shmem_size_hint;
     auto buffer_size_kb  = ocfg.perfetto_buffer_size;
 
+    LOG_IF(FATAL, !defaults::is_valid_perfetto_buffer_size(buffer_size_kb))
```
bgopesh: This check is triplicated.  Extract a single validate_perfetto_buffer_size(size_t) helper; call once at the point of assignment. Since perfetto_buffer_size is only assigned in parse_env(), the two downstream checks are redundant

projects/rocprofiler-sdk/source/bin/rocprofv3.py
```diff
@@ -42,6 +42,9 @@
     "rocm_version": "@rocm_version_FULL_VERSION@",
 }
 
+PERFETTO_BUFFER_SIZE_KB_MIN = 1
+PERFETTO_BUFFER_SIZE_KB_MAX = ((1 << 32) - 1) // 1024
```
bgopesh, line 53: please explain this value in a comment

bgopesh (changes requested):

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/att_no_intercept_quick_scan.cpp
srihari-uttanur: Yes this is not related to the PR but it fixes a warning caused in the thread trace, I didn't add the details about this.

projects/rocprofiler-sdk/source/lib/output/generatePerfetto.cpp
yhuiYH: Well, the rocpd side can also set --perfetto_buffer_size, so it's not only assigned in parse_env().  I see that on the rocpd side, we end up hitting this check in projects/rocprofiler-sdk/source/lib/python/rocpd/source/perfetto.cpp.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk-tool/att_no_intercept_quick_scan.cpp
srihari-uttanur: removed this change, as its been fixed after rebase.

projects/rocprofiler-sdk/source/bin/rocprofv3.py
srihari-uttanur: Done

bgopesh (approved):

projects/rocprofiler-sdk/source/lib/output/generatePerfetto.cpp
srihari-uttanur: We currently have 2 checks, one in output_config::parse_env() and another for rocpd in PerfettoSession ctor. As the rocpd path sets it directly through the pybind binding def_readwrite("perfetto_buffer_size", …)

projects/rocprofiler-sdk/source/lib/output/output_config.hpp
```diff
@@ -49,6 +51,23 @@ namespace defaults
 {
 constexpr auto perfetto_buffer_size_kb     = (1 * common::units::GiB) / common::units::KiB;
 constexpr auto perfetto_shmem_size_hint_kb = 64;
+constexpr auto perfetto_buffer_size_min_kb = size_t{1};
+constexpr auto perfetto_buffer_size_max_kb =
+    static_cast<size_t>(std::numeric_limits<uint32_t>::max()) / common::units::KiB;
+
+constexpr bool
```
SrirakshaNag: Why do we need this separate function here? Also, look like it is not used outside of the output_config.hpp, if so, why do we have it in .hpp file?

SrirakshaNag (changes requested):

projects/rocprofiler-sdk/source/lib/output/output_config.hpp
srihari-uttanur: I have moved it to .cpp

SrirakshaNag (approved):
Looks good to me

PR #8850, opened 2026-07-20
[rocprofiler-compute] Analyze multi-process PC sampling results

abchoudh-amd:
## Motivation

Enable analyze mode to combine PID-prefixed PC sampling results from multi-process workloads into one report. Process-scoped database identities prevent equal kernel, dispatch, and code-object IDs from different processes from being conflated.

## Technical Details

- Load PID-prefixed result files in numeric process order
- Require unique PID metadata for every loaded record
- Aggregate equivalent PC samples 
- Preserve process-scoped database identity and ISA attribution
- Keep public CSV and database-view columns unchanged

## Issue Tracking

JIRA ID: AIPROFCOMP-702

## Test Plan

- [x] Run pytest tests/test_pc_sampling_analysis.py
- [x] Run pytest tests/test_analysis_db.py
- [x] Run pytest tests/test_analysis_orm.py

## Test Result

- [x] pytest tests/test_pc_sampling_analysis.py passes
- [x] pytest tests/test_analysis_db.py passes
- [x] pytest tests/test_analysis_orm.py passes

## Submission Checklist


projects/rocprofiler-compute/src/utils/file_io.py
```diff
@@ -376,3 +411,70 @@ def is_single_panel_config(
...
+            candidate_path
+            for candidate_path in direct_child_files
+            if candidate_path.name == legacy_results_filename
+        ),
+        None,
+    )
+    return (legacy_result_file,) if legacy_result_file is not None else ()
+
+
+def _validate_pc_sampling_process_ids(
```
xuchen-amd, line 463: this helper function is ran twice: `load_pc_sampling_results()` and `process_pc_sampling_kernel_traces()`, is the redundant call needed?

projects/rocprofiler-compute/src/utils/parser.py
```diff
@@ -533,15 +533,103 @@ def load_pc_sampling_data(
...
 
+def _load_pc_sampling_data_from_records(
+    method: str,
+    tool_data_records: list[dict[str, Any]],
+    sorting_type: str,
+    kernel_name: Optional[str] = None,
+    num_rows: Optional[int] = None,
+) -> pd.DataFrame:
+    process_frames = [
+        load_pc_sampling_data_per_kernel(
```
xuchen-amd: `load_pc_sampling_data_per_kernel` is called for each process, which does: load raw samples → aggregate → enrich → sort → convert offset to hex → truncate → project columns. Then the wrapper throws away that sorting and formatting — it re-aggregates across processes, re-sorts, and re-truncates.

projects/rocprofiler-compute/src/utils/file_io.py
```diff
@@ -376,3 +411,70 @@ def is_single_panel_config(
         console_warning(
             "Found multiple panel config sets but incomplete for all archs."
         )
+
+
+def _select_pc_sampling_result_files(
```
xuchen-amd: this function does two things: finds PID-prefixed files AND falls back to the legacy file. let's split the functionalities.

projects/rocprofiler-compute/src/utils/analysis_orm.py
```diff
@@ -510,6 +512,116 @@ def _json_sanitize(value: object) -> object:
             return None
         return value
 
+    @staticmethod
+    def _build_pc_sampling_view_select() -> Select[Any]:
```
xuchen-amd: are there opportunities to break down this function?

xuchen-amd (changes requested):
thanks for working on this! left some comments.

projects/rocprofiler-compute/docs/data/analyze/analysis_data_dump_schema.mmd
cfallows-amd, line 1: If we are adding to the schema, we need to regenerate the image. Ask agent to generate it, make sure it is same quality/size image and is still in black and white.

projects/rocprofiler-compute/src/utils/analysis_orm.py
```diff
@@ -121,14 +122,15 @@ class KernelRooflineData(Base):
 
 class Dispatch(Base):
     __tablename__ = f"{PREFIX}dispatch"
-    # dispatch_id is unique within a kernel.
-    __table_args__ = (UniqueConstraint("kernel_uuid", "dispatch_id"),)
+    # dispatch_id is unique within a kernel and process.
+    __table_args__ = (UniqueConstraint("kernel_uuid", "pid", "dispatch_id"),)
```
cfallows-amd: This changes the expected columns- should we add something to warn users (maybe this can be in changelog?) that older schemas generated will not work any longer?

cfallows-amd: Ctest passed so I assume this is not an issue with any generated workloads from the tests folder?

projects/rocprofiler-compute/src/utils/parser.py
```diff
@@ -485,20 +485,20 @@ def load_pc_sampling_data(
...
     kernel (one ``-k``). The output schema is identical either way. Callers
-    pass the already-parsed *tool_data*.
+    pass the already-parsed *tool_data_records*.
     """
-    if not file_prefix or file_prefix.lower() == "none" or tool_data is None:
+    if not file_prefix or file_prefix.lower() == "none" or not tool_data_records:
         return pd.DataFrame()
 
-    pc_sampling_method = detect_pc_sampling_method(tool_data)
+    pc_sampling_method = detect_pc_sampling_method(tool_data_records[0])
```
cfallows-amd: Optional nip-pick suggestion: add a comment above this line to specify that there is an assumption all entries of tool_data
_records was profiled with the same pc sampling method, hence using first entry.
Just adds clarity of what's happening here, since we don't offer any sort of mismatched sampling methods.

projects/rocprofiler-compute/src/utils/file_io.py
```diff
@@ -214,25 +217,54 @@ def build_agent_to_gpu_map_from_json(
...
+
+
 @demarcate
-def load_pc_sampling_results(workload_path: str) -> Optional[dict[str, Any]]:
+def load_pc_sampling_results(workload_path: str) -> list[dict[str, Any]]:
     """
-    Parse ``ps_file_results.json`` and return its ``rocprofiler-sdk-tool[0]``
-    record. Returns ``None`` if the file is absent or fails to parse (a
-    warning is logged in the latter case).
+    Discover and parse every PC sampling ``rocprofiler-sdk-tool[0]`` record.
```
cfallows-amd: This description isn't accurate anymore, can you transfer some of this into a description for _parse_pc_sampling_result_file

projects/rocprofiler-compute/src/utils/file_io.py
cfallows-amd: And please add descriptions to specify functionality of each one. Thanks

projects/rocprofiler-compute/src/utils/file_io.py
cfallows-amd: Also a one-liner to explain why it is index 0 for clarity about the SDK output.

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
```diff
@@ -458,37 +476,27 @@ def add_code_object_isa(
...
         # Reuse an existing code object instead of creating a duplicate.
         existing_code_objects = {
             (store.pid, store.code_object_id): store
             for store in workload_obj.code_object_stores
         }
 
         for pid, disassemblies in load_code_object_disassemblies(workload_path).items():
-            if pid != surviving_pid:
+            tool_data = tool_data_by_pid.get(pid)
+            if tool_data is None:
```
cfallows-amd, line 505: Is it possible this could be [], and needs to be a "not" check? Trying to trace if this could possibly be populated with a value of empty list, but not sure. The is None only checks if key DNE.

projects/rocprofiler-compute/src/utils/parser.py
```diff
@@ -533,15 +533,103 @@ def load_pc_sampling_data(
...
+    return aggregated_df[df.columns]
+
+
+def _sort_pc_sampling_display_rows(
+    df: pd.DataFrame,
+    sorting_type: str,
+) -> pd.DataFrame:
+    if sorting_type == "count":
+        return df.sort_values(by=["count"], ascending=False)
+    if sorting_type != "offset":
```
cfallows-amd: This case where the pc_sampling_sorting_type should already be caught by our argparser, where we already reserve the choices to "offset" or "count".  Do we need this block?

projects/rocprofiler-compute/tests/test_analysis_db.py
cfallows-amd, line 1: Please add brief descriptions for the tests added that do not have them already- what is being tested and what to expect as the outcome.

projects/rocprofiler-compute/tests/test_analysis_orm.py
cfallows-amd, line 1: Descriptions for test cases, with expected outcomes for each.

projects/rocprofiler-compute/src/utils/analysis_orm.py
abchoudh-amd: Done.

projects/rocprofiler-compute/src/utils/analysis_orm.py
abchoudh-amd: This would affect the consumer of analysis database, e.g., Optiq. We have to communicate this to them.

Added it to ChangeLog.

projects/rocprofiler-compute/src/rocprof_compute_analyze/analysis_db.py
abchoudh-amd: It can produce an empty `dict`. In that case, there would be no ISA ingestion.

projects/rocprofiler-compute/src/utils/parser.py
abchoudh-amd: Refactored and removed the check.

projects/rocprofiler-compute/src/utils/parser.py
abchoudh-amd: Fixed it.

projects/rocprofiler-compute/src/utils/file_io.py
abchoudh-amd: After refactoring, this has become redundant. I have removed it.

xuchen-amd (approved):
thanks for addressing my comments, I still have some questions that I will communicate offline. approving for now.

vedithal-amd (changes requested):
reviewing

vedithal-amd:
Rebased this branch onto the rebased `users/abchoudh/pcs-profile`.

- Guard in `src/utils/parser.py` collapsed to `if not tool_data_records:`.
- Test call sites lost the prefix positional. Kept the base's deletion of `test_load_pc_sampling_data_empty_prefix` / `_none_prefix` rather than letting the stack resurrect them.
- `CHANGELOG.md`: merged your PC sampling entries into develop's existing 3.9.0 `### Changed` / `### Removed` sections.

Two spots git merged cleanly but broke, fixed in `3fd9874f1b`:

- 15 calls in `tests/test_pc_sampling_analysis.py` still passed `"ps_file"` as the removed positional, so it landed in `sorting_type`.
- `src/utils/parser.py:418` interpolated `file_prefix` in the conflicting-methods error, a `NameError`. Reworded to "found across result files".

335 unit tests pass, ruff clean, config hash and template checks pass. GPU and analyze-workload tests not run.

projects/rocprofiler-compute/CHANGELOG.md
```diff
@@ -7,18 +7,24 @@ Full documentation for ROCm Compute Profiler is available at [https://rocm.docs.
...
 ### Added
 
+* Added multi-process PC sampling across profile and analyze modes.
+  * Profile mode writes one PID-prefixed `<pid>_ps_file_results.json` file per process, and analyze mode combines those files into a single analysis.
+
 ### Changed
 
 * ML API tracing options (--torch-trace/--triton-trace/--ml-api-trace) are no longer allowed with PC-sampling-only profiling; the run now fails with an error telling the user to drop the ML API tracing flag or add a counter block, since without counters there is nothing to correlate the markers against.
 
 * Deprecated the `--join-type` profile mode option; it no longer has any effect and will be removed in a future release.
```
vedithal-amd, line 20: Remove whitespace

projects/rocprofiler-compute/CHANGELOG.md
```diff
@@ -7,18 +7,24 @@ Full documentation for ROCm Compute Profiler is available at [https://rocm.docs.
...
 
+
 ### Removed
 
 * Removed the CSV profile output backend and the `--format-rocprof-output` profile mode option. Profiling now always uses the `rocpd` output format, which was already the default.
 
 * Removed analyze support for workloads produced by the CSV profile backend. Such workloads are now rejected with an error telling you to re-profile with a current release.
 
+* Removed analyze support for unprefixed or PID-less PC sampling result files; affected workloads must be re-profiled.
```
vedithal-amd: This is implicit and not required to be handled

projects/rocprofiler-compute/CHANGELOG.md
```diff
@@ -7,18 +7,24 @@ Full documentation for ROCm Compute Profiler is available at [https://rocm.docs.
 
 ### Added
 
+* Added multi-process PC sampling across profile and analyze modes.
+  * Profile mode writes one PID-prefixed `<pid>_ps_file_results.json` file per process, and analyze mode combines those files into a single analysis.
+
```
vedithal-amd, line 14: CHANGELOG should not explain the internal mechanics like %pid%_ps_file_results.json those are not useful for end user quickly reading the changes.

Just a one liner in the Resolved section saying the following should be enough:

Fixed PC sampling profiling and analysis to cover every process in a multi-process workload, previously only one process's samples were reported.

vedithal-amd (changes requested):

projects/rocprofiler-compute/CHANGELOG.md
vedithal-amd: will fix

projects/rocprofiler-compute/CHANGELOG.md
vedithal-amd: will fix

projects/rocprofiler-compute/CHANGELOG.md
vedithal-amd: will fix

vedithal-amd:
- Renumbered PC sampling dispatch ids across processes. Ids restart per process, so two processes emitted duplicate ids and the dispatch unique constraint rejected them. Follows the counter path's `GroupIdAssigner`.
- Restored the `kernel_name` unique constraint and removed the per-PID kernel rows. `pc_sampling.csv` was referencing `kernel_uuid`s absent from `kernel.csv`, so PC sampling data was unreachable for counter-backed runs.
- Dropped `pid` from the counter dispatch path; profile mode folds PID into `Dispatch_ID` and strips the column, so it was always `None`.
- Added a `pid` column to the PC sampling table and the view, and put `pid` and `code_object_id` back into row identity. Rows from different processes are no longer summed together.
- Reverted the view to the pre-PR statement and deleted the cross-process merge helpers this makes dead.
- Reset the display index so the table reads 0..n-1.
- Kept the changelog entry under `Added`, covering the PID-prefixed profile output and the per-process analyze report.
- Verified on a two-process workload: both pids reported, per-kernel sample ratios within 3.5% of each other, and no sampling row with an unresolvable `kernel_uuid`.

vedithal-amd (approved):
Verified locally on a two-process PC sampling profile, in both CLI and database output: dispatch ids are unique across processes, every `pc_sampling` `kernel_uuid` resolves in `kernel.csv`, and rows stay separated by pid. Unit suites and ruff clean.

PR #8852, opened 2026-07-20
[Do not review] Find out why gfx950 workflow rocminfo fails

kcossett-amd:
## Motivation

I've noticed some tests being skipped in GFX950 workflow due to no valid GPU. That should be impossible and means rocminfo fails. I had it so that if rocminfo fails, it would assume no GPU.

## Submission Checklist


kcossett-amd:
It is no longer failing...

PR #8853, opened 2026-07-20
fix(rocr): ROCM-27679 Swap ops alignment checks update

pghoshamd:
## Motivation

When a hip application tries swap operations using incorrectly aligned addresses in memory_async_batch_copy API, rocr would send back an expected error code. 
Hip layers plans to add a fallback logic to catch this negative case later on but it is recommended to note this alignment requirement around the API documentation.  

## Submission Checklist


kentrussell (approved):

projects/rocr-runtime/runtime/hsa-runtime/inc/hsa_ext_amd.h
```diff
@@ -2392,8 +2392,9 @@ typedef struct hsa_amd_memory_copy_op_s {
  * Each operation is self-describing via its @c type field. A BROADCAST operation
  * is a single op that copies one source to multiple destinations via @c dst_list
  * and @c num_entries. A SWAP operation exchanges two buffers using @c src_size and
- * @c dst_size.
- *
+ * @c dst_size. SWAP operations require addresses to be 64-byte aligned for gfx94/gfx95X
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** Trailing whitespace: this line ends with a trailing space after `gfx95X`, and the blank comment line just below was changed from ` *` to ` * ` (trailing space added). Please strip the trailing whitespace on both — some CI whitespace linters flag it.

**[Nit]** `gfx94/gfx95X` mixes a bare two-digit form with an `X`-suffixed form. Recommend a consistent style — either `gfx94x/gfx95x` or the explicit targets (e.g. `gfx942/gfx950`) — so it's unambiguous which ASICs the 64-byte rule covers.

**[Question]** Can you confirm 64-byte (gfx94/gfx95X) and 32-byte (gfx1250) are the correct and complete alignment constraints? This comment becomes the public contract for callers building `hsa_amd_memory_copy_op_s` arrays, so it'd be good to be sure no other supported family needs a different value.

cfreeamd (commented):

Docs-only change to the `hsa_amd_memory_copy` Doxygen block in `hsa_ext_amd.h`. No code/behavior change. A couple of whitespace/notation nits and one clarifying question on the alignment values are noted inline.

dayatsin-amd:
This PR documents the swap alignment requirement (64-byte for gfx94/gfx95X, 32-byte for gfx1250), but the runtime enforcement in `amd_blit_sdma.cpp` currently has two gaps that mean a misaligned swap is not always rejected with `HSA_STATUS_ERROR_INVALID_ARGUMENT`:

**1. The check always uses 32-byte alignment, even on gfx94/gfx95X.**
The only validation is in `BlitSdma::SubmitBodies`:

```cpp
if (is_swap) {
  constexpr size_t kAlign = SDMA_PKT_COPY_LINEAR_SWAP_WAITSIGNAL_GFX1250::kAlignment_; // == 32
  for (size_t i = 0; i < num_entries; ++i) {
    const uint32_t d = indices[i];
    if ((reinterpret_cast<uintptr_t>(dst_list[d]) & (kAlign - 1)) != 0 ||
        (reinterpret_cast<uintptr_t>(src_list[d]) & (kAlign - 1)) != 0)
      return HSA_STATUS_ERROR_INVALID_ARGUMENT;
  }
}
```

`kAlign` is hardcoded to the gfx1250 value (32), but `SubmitBodies` is shared by the classic gfx94/gfx95X path too. The classic packet `SDMA_PKT_COPY_LINEAR_SWAP::kAlignment_` is 64. So on gfx94/gfx95X, an address that is 32-byte- but not 64-byte-aligned passes validation while violating the requirement this PR documents. Correct per-arch alignment (64 vs 32) is only asserted in `BuildSwapCopyCommand`, and asserts are compiled out in release builds.

**2. The gfx125+ fused coordinator path skips the check entirely.**
In `DmaCopyFanOutOp`, only non-coordinator engine groups go through `SubmitBodies`. The coordinator group's entries are submitted via `SubmitFusedCoordinator`, which has no alignment validation. For small batches where all entries land in the coordinator group, no alignment error is returned.

Suggested fix:
- Select `kAlign` per-arch in `SubmitBodies` (64 for classic gfx94/gfx95X, 32 for gfx1250).
- Add the same check to `SubmitFusedCoordinator`, or hoist the validation up into `DmaCopyFanOutOp` before dispatch so both paths are covered.

projects/rocr-runtime/runtime/hsa-runtime/inc/hsa_ext_amd.h
pghoshamd: Addressed the first two nits in the latest commit. As for the last question, there is no other alignment constraints apart from the ones added to the API documentation, i.e., 64-byte (gfx94x/gfx95x) and 32-byte (gfx1250).

dayatsin-amd (approved):

cfreeamd (approved):

PR #8854, opened 2026-07-20
[rocjitsu]: RAII-own dispatch resources for queue-first teardown

atgutier:
## Motivation

Issue https://github.com/ROCm/rocm-systems/issues/8818 (recycled doorbell can lose or falsely complete replacement queue work) had two halves: a runtime doorbell bug and a test-lifetime weakness that masked its symptoms. The runtime fix landed in https://github.com/ROCm/rocm-systems/pull/8672. This change addresses the test-lifetime half.

In the CDNA4->CDNA3 dispatch tests, a lost or timed-out dispatch did not surface as a clean test failure. It appeared as a confusing exit-time segmentation fault, because the affected helpers only released their queue and packet-referenced
resources on the success path. When a dispatch failed, the queue was left alive with a pending simulator dispatch that still referenced freed packet operands, and hsa_shut_down() then tore into that live state. That misdirection cost real debugging time while diagnosing https://github.com/ROCm/rocm-systems/issues/8818, so the tests must fail cleanly and deterministically regardless of dispatch outcome.

Three helpers in cdna4_to_cdna3_dispatch_test.cpp — run_dynamic_copy_loop, run_buffer_async_triton_matmul, and run_flash_attention_triton — created a code-object reader, executable, device/kernarg pool allocations, a queue, and a
completion signal, then released them via manual cleanup at the function tail, AFTER the ASSERT_EQ on the completion-signal wait (and after the result-verifying assertions). Any early return — a dispatch timeout, a signal-wait mismatch, or a
verification failure — skipped that cleanup entirely, leaking the queue and leaving a live dispatch for hsa_shut_down() to trip over. Only run_triton_matmul already used an RAII owner (HsaMatmulResources).

## Technical Details

- Generalize HsaMatmulResources into a single reusable HsaDispatchResources. It replaces the fixed a/b/c buffer members with a device_allocs vector plus an alloc_device(pool, bytes, out) helper that allocates a global-pool block, records it for destruction, and returns the hsa_status_t so a failed allocation reports the actual OOM/invalid-pool error rather than only a null pointer. It covers helpers with two, three, or four device buffers uniformly.
- Its destructor destroys the queue FIRST, stopping queue execution before releasing anything an in-flight AQL packet can reference, then the completion signal, kernarg, recorded device allocations, executable, and reader. Because
  it is a stack RAII owner, this ordering runs on every exit path — success, assertion, or timeout.
- Mark HsaDispatchResources non-copyable and non-movable (deleted copy/move constructors and assignment): it is the sole owner of raw HSA handles and allocation pointers, so an accidental copy would double-free/destroy in the dtor. All call sites use it as a local, so deleting copy and move is sufficient.
- Convert all four helpers to HsaDispatchResources and delete their manual tail-cleanup blocks. Callers keep local typed pointers (a_dev, q_dev, etc.) into the recorded allocations for packet setup and verification; ownership lives
  in the resources object.

## Issue Tracking

#8818

## Test Plan

Run all CI tests

## Test Result

All CI tests: PASSED

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/tests/dbt/cdna4_to_cdna3_dispatch_test.cpp
+++ b/emulation/rocjitsu/tests/dbt/cdna4_to_cdna3_dispatch_test.cpp
@@ -1067,7 +1067,7 @@ void run_flash_attention_triton(const std::vector<uint8_t> &elf_bytes, const Dis
 
   if (observed)
     *observed = std::move(o_host);
-  // Resources released by HsaDispatchResources::~ (queue first).
+  // Resources released by HsaDispatchResources destructor (queue first).
 }
 
 } // namespace
--- a/emulation/rocjitsu/tests/dbt/cdna4_to_cdna3_dispatch_test.cpp
+++ b/emulation/rocjitsu/tests/dbt/cdna4_to_cdna3_dispatch_test.cpp
@@ -914,7 +914,7 @@ void run_buffer_async_triton_matmul(const std::vector<uint8_t> &elf_bytes,
 
   if (observed)
     *observed = std::move(c_host);
-  // Resources released by HsaDispatchResources::~ (queue first).
+  // Resources released by HsaDispatchResources destructor (queue first).
 }
 
 void run_flash_attention_triton(const std::vector<uint8_t> &elf_bytes, const DispatchTarget &target,
--- a/emulation/rocjitsu/tests/dbt/cdna4_to_cdna3_dispatch_test.cpp
+++ b/emulation/rocjitsu/tests/dbt/cdna4_to_cdna3_dispatch_test.cpp
@@ -568,7 +568,7 @@ void run_dynamic_copy_loop(const std::vector<uint8_t> &elf_bytes, const Dispatch
     }
     EXPECT_EQ(mismatches, 0u) << mismatches << " mismatches for N=" << n;
   }
-  // Resources released by HsaDispatchResources::~ (queue first).
+  // Resources released by HsaDispatchResources destructor (queue first).
 }
 
 void translate_triton_fixture(const char *name, uint32_t mach, std::vector<uint8_t> &elf_bytes,
--- a/emulation/rocjitsu/tests/dbt/cdna4_to_cdna3_dispatch_test.cpp
+++ b/emulation/rocjitsu/tests/dbt/cdna4_to_cdna3_dispatch_test.cpp
@@ -362,20 +362,51 @@ struct HsaShutdownGuard {
   ~HsaShutdownGuard() { hsa_shut_down(); }
 };
 
-struct HsaMatmulResources {
+/// @brief RAII owner for one raw-HSA dispatch's resources.
+/// @details Every dispatch helper builds the same object graph (code-object
+/// reader, executable, device/kernarg pool allocations, queue, completion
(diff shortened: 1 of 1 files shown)


emulation/rocjitsu/tests/dbt/cdna4_to_cdna3_dispatch_test.cpp
```diff
@@ -362,20 +362,37 @@ struct HsaShutdownGuard {
...
+/// @details Every dispatch helper builds the same object graph (code-object
+/// reader, executable, device/kernarg pool allocations, queue, completion
+/// signal) and must tear it down in a fixed order even when a mid-helper
+/// ASSERT/timeout returns early. Manual tail cleanup is skipped on any early
+/// return, which leaks the queue and — worse — leaves a live simulator dispatch
+/// referencing freed packet operands when hsa_shut_down() later runs, turning a
+/// lost/timed-out dispatch into a confusing exit-time crash instead of a clean
+/// test failure. This owner destroys the queue FIRST (stopping execution) before
+/// releasing anything an in-flight AQL packet can reference.
+struct HsaDispatchResources {
```
kuhar, line 375: We should make `HsaDispatchResources` non-copyable, and either non-movable or explicitly movable. Its implicit copy constructor duplicates the HSA handles and allocation pointers, so an accidental copy would double-destroy or free them; the current call sites only use this owner as a local, so deleting copy and move is sufficient here.

emulation/rocjitsu/tests/dbt/cdna4_to_cdna3_dispatch_test.cpp
```diff
@@ -362,20 +362,37 @@ struct HsaShutdownGuard {
...
   bool reader_created = false;
   bool executable_created = false;
   bool signal_created = false;
 
-  ~HsaMatmulResources() {
+  /// Allocate a global-pool block and record it for destruction. Returns the
+  /// pointer (nullptr on failure, so callers still ASSERT on it).
+  void *alloc_device(hsa_amd_memory_pool_t pool, size_t bytes) {
+    void *ptr = nullptr;
+    if (hsa_amd_memory_pool_allocate(pool, bytes, 0, &ptr) == HSA_STATUS_SUCCESS && ptr)
```
kuhar: It would be useful to preserve the allocation status in `alloc_device()`. Returning only `nullptr` turns a failed `hsa_amd_memory_pool_allocate` into `ASSERT_NE(ptr, nullptr)`, whereas the previous code reported the actual `hsa_status_t`; returning the status or exposing it to the caller would keep OOM and invalid-pool failures diagnosable.

kuhar (commented):
Do we still need [#8821](https://github.com/ROCm/rocm-systems/pull/8821), or does #8854 fully cover the intended fix? The runtime behavior from #8821 is needed because #8854 only improves cleanup after a dispatch stalls; it does not prevent the recycled-doorbell race. However, the equivalent doorbell-rearm behavior appears to already be present in the base of #8854 through #8672.

If #8672 fully subsumes the runtime change from #8821, could we close #8821 and keep only #8854? The focused `ReusedDoorbellSlotIsRearmedForFirstPacket` regression test from #8821 does not appear to have landed with #8672, so it may still be worth carrying that test over before closing #8821.

kuhar (approved):

PR #8859, opened 2026-07-20
refactor(rocprofiler-systems): Remove Timemory readlink()

lloginov-amd:
## Motivation

This PR is part of a PR series to remove the dependency on Timemory filepath utilities. This particular PR replaces Timemory readlink() function with in-tree implementation based on std::filesystem::read_symlink().

## Technical Details

  - replaced the in-tree path::readlink() and the external tim::filepath::readlink() with a new path::read_symlink() built on std::filesystem::read_symlink
  - removed the now-dead path::is_link() helper, which was only used internally by the old readlink() implementation
  - updated the one production call site in rocprof-sys-dl/dl.cpp and removed the #include <timemory/utility/filepath.hpp>
  - replaced the old Readlink_* and IsLink_* unit tests with ReadSymlink_* tests, adding a new broken-symlink case
  - Behavior is unchanged: returns the raw symlink target on success, returns the original input on any filesystem error (including non-symlink paths and nonexistent paths).
  - Added 3 checks for symlink() value in unit tests to fix compilation warnings
  - Removed unused get_exe_realpath() from internal_libs.cpp

## Issue Tracking

JIRA ID : AIPROFSYST-528

## Test Plan

Existing CTest suite and new tests should pass

## Test Result

Existing and new tests pass

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-systems/source/lib/common/tests/test_path.cpp
+++ b/projects/rocprofiler-systems/source/lib/common/tests/test_path.cpp
@@ -135,23 +135,29 @@ TEST_F(PathTest, IsLink_SymbolicLink)
 
 TEST_F(PathTest, IsLink_NonexistentPath) { EXPECT_FALSE(is_link("/nonexistent/path")); }
 
-TEST_F(PathTest, Readlink_SymbolicLink)
+TEST_F(PathTest, ReadSymlink_SymbolicLink)
 {
-    std::string target    = create_file("readlink_target.txt");
-    std::string link_path = create_symlink(target, "readlink_link");
-    EXPECT_EQ(readlink(link_path), target);
+    std::string target    = create_file("read_symlink_target.txt");
+    std::string link_path = create_symlink(target, "read_symlink_link");
+    EXPECT_EQ(read_symlink(link_path), target);
 }
 
-TEST_F(PathTest, Readlink_NotALink)
+TEST_F(PathTest, ReadSymlink_NotALink)
 {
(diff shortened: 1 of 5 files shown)


projects/rocprofiler-systems/source/lib/common/tests/test_path.cpp
mradosav-amd: I know it's not related to this PR, but can you please add check for this line, since we are ignoring return value from `symlink` and we get compiler warning. Just add actual result check.

mradosav-amd (approved):

projects/rocprofiler-systems/source/lib/common/tests/test_path.cpp
lloginov-amd: Fixed, thanks! 2 usages in test_path.cpp and 1 usage in test_discover_llvm_libdir.cpp

dgaliffiAMD (approved):
Please update CHANGELOGS before merging. 

PR Hygiene: Please update the "technical details" in the description. It looks like it was copied into "motivation" instead.

lloginov-amd:
> Please update CHANGELOGS before merging.
> 
> PR Hygiene: Please update the "technical details" in the description. It looks like it was copied into "motivation" instead.

@dgaliffiAMD Updated "technical details". As for changelog - this PR just changes internal implementation and does not affect users in any way, so I would prefer to keep it out of changelog

lloginov-amd:
PR bot: times out but checks successful

PR #8861, opened 2026-07-20
[rocjitsu]: Retire idle CP doorbell pollers after the last queue

### LINKED ISSUE #8813: [rocjitsu] idle command-processor doorbell pollers survive the last HSA queue
# rocJITsu keeps idle command-processor pollers alive after the last HSA queue

> [!NOTE]
> [ROCm/rocm-systems#8672](https://github.com/ROCm/rocm-systems/pull/8672)
> does **not** fix this issue at the head tested on July 18. Its release build
> retains the same idle polling workers as the shipped nightly.

## Environment

- ROCm SDK and rocJITsu baseline: `7.15.0a20260718`
- Physical control: gfx1100 on the same SDK
- rocJITsu configuration: `gfx942_cdna3_kmd.json`

## Problem

rocJITsu starts one `CommandProcessor::doorbell_poll_loop` worker for each
command processor that receives a host-accessible HSA queue. Destroying the
last such queue removes its queue state but leaves the worker running until VM
shutdown. An idle worker repeatedly scans an empty queue set and sleeps for
only 100 microseconds before trying again.

The leak is bounded by the number of command processors, not proportional to
the total number of queues created. `SoC::assign_queue_cp()` distributes queues
round-robin across XCD command processors, so a small number of queues can look
like one worker per queue. This reducer uses 32 queues and three complete
create/destroy rounds: the shipped build reaches a stable plateau after the
first round, and every idle worker remains at that plateau.

## Expected behavior

Once a command processor owns no host-accessible queues, its doorbell monitor
should exit without keeping an idle polling resource alive. Queue destruction
must not synchronously join a monitor that may be inside engine or event code;
the monitor can observe the empty queue set and exit itself. A subsequent queue
registration can join the already-exited `std::jthread` before starting a new
monitor.

## Reproducer

Build `repro.c` using only the ROCm SDK and the Linux C library:

```bash
export ROCM_PATH=/path/to/rocm-7.15.0a20260718
"$ROCM_PATH/bin/amdclang" -std=c11 -Wall -Wextra -Werror \
  -I"$ROCM_PATH/include" repro.c \
  -L"$ROCM_PATH/lib" -Wl,-rpath,"$ROCM_PATH/lib" \
  -lhsa-runtime64 -o repro
```

Physical hardware:

```bash
LD_LIBRARY_PATH="$ROCM_PATH/lib:$ROCM_PATH/lib/llvm/lib" ./repro
```

rocJITsu:

```bash
runtime_dir="$(mktemp -d)"
printf '%s\n' \
  "$ROCM_PATH/share/rocjitsu/configs/gfx942_cdna3_kmd.json" \
  >"$runtime_dir/config_path"
ROCJITSU_RUNTIME_DIR="$runtime_dir" \
LD_PRELOAD="$ROCM_PATH/lib/librocjitsu.so" \
LD_LIBRARY_PATH="$ROCM_PATH/lib:$ROCM_PATH/lib/llvm/lib" \
  ./repro
rm -rf "$runtime_dir"
```

## Observed results

Physical gfx1100:

```text
baseline threads: 3
round 1 threads with 32 live queues: 3
round 1 threads after destroying all queues: 3
round 2 threads with 32 live queues: 3
round 2 threads after destroying all queues: 3
round 3 threads with 32 live queues: 3
round 3 threads after destroying all queues: 3
queue thread count returned to baseline after every round
```

Shipped rocJITsu gfx942 configuration:

```text
baseline threads: 6
round 1 threads with 32 live queues: 12
round 1 threads after destroying all queues: 12
round 2 threads with 32 live queues: 12
round 2 threads after destroying all queues: 12
round 3 threads with 32 live queues: 12
round 3 threads after destroying all queues: 12
```

GDB identifies the retained workers as:

```text
rocjitsu::amdgpu::CommandProcessor::doorbell_poll_loop(std::stop_token)
```

## Candidate fix validation

A local source fix makes an empty monitor exit asynchronously and serializes
reaping/restarting the `std::jthread` separately from the queue-state lock. The
same reducer then reports:

```text
baseline threads: 5
round 1 threads with 32 live queues: 12
round 1 threads after destroying all queues: 5
round 2 threads with 32 live queues: 12
round 2 threads after destroying all queues: 5
round 3 threads with 32 live queues: 12
round 3 threads after destroying all queues: 5
queue thread count returned to baseline after every round
```

A separate queue-progress reducer also passes create, destroy, reuse, blocked
barrier release, and completion-wakeup paths with this fix. The shipped build
passes that progress reducer too; the evidence here is an idle worker lifetime
and wakeup problem, not an established stale-queue correctness failure.

## Impact

Long-lived processes that exercise queues across several command processors
retain a bounded set of otherwise unnecessary workers. Each empty worker wakes
roughly every 100 microseconds for the remainder of the VM lifetime, adding
host scheduling and power overhead even when no HSA queues exist.

atgutier:
## Motivation

Each CommandProcessor starts one doorbell_poll_loop thread when it receives its
first host-accessible (KFD) queue, but nothing stopped that thread when the last
such queue was destroyed. unregister_queue deliberately does not join the monitor
-- the join can deadlock a poller that is mid-iteration inside engine or event
code holding other locks -- and no other path retired it before VM shutdown. Every
CP that ever serviced a queue therefore kept an idle poller alive for the rest of
the process, each waking roughly every 100us to scan an empty queue set.

The leak is bounded by command-processor count, not queue count (SoC round-robins
queues across XCD CPs), so it plateaus rather than growing without limit, but
long-lived processes that touch several CPs carry a fixed set of otherwise-idle
workers that add host scheduling and power overhead when no HSA queues exist.
Reported as issue https://github.com/ROCm/rocm-systems/issues/8813.

## Technical details

Give the monitor an asynchronous self-exit and serialize its reap/restart on a
mutex distinct from the queue-state lock, so queue destruction never has to join a
running thread:

- Add doorbell_thread_mutex_ (guarding doorbell_running_ and doorbell_thread_),
  deliberately never nested under hw_queue_mutex_, so monitor lifecycle and
  queue-state operations cannot deadlock.
- doorbell_poll_loop() now checks at the top of each iteration whether the CP owns
  any host-accessible queue and has no pending invalid/stall retry; if not, it
  clears doorbell_running_ and returns. It acquires doorbell_thread_mutex_ with a
  TRY-lock, never a blocking lock: stop_doorbell_monitor() holds that mutex while
  it joins this very thread, so a blocking acquire would deadlock. A failed try
  just defers the exit check to the next 100us iteration.
- Add ensure_doorbell_monitor(), called from register_queue() AFTER hw_queue_mutex_
  is released. Under doorbell_thread_mutex_ it no-ops when a monitor is already
  running, otherwise reaps a previously self-exited jthread (join is immediate
  since it already returned) and launches a fresh one. This is how a CP whose
  monitor retired gets polling back when a new queue lands on it.
- ensure_doorbell_monitor() constructs the jthread BEFORE setting doorbell_running_,
  so a thread-creation failure (std::system_error from the jthread constructor)
  leaves the flag false and a later call retries, rather than latching the flag
  true with no running monitor and no-oping forever.
- stop_doorbell_monitor() now also takes doorbell_thread_mutex_ and clears
  doorbell_running_, keeping the destructor/teardown path consistent with the new
  lifecycle flag.

Correctness of the self-exit vs. register race rests on two invariants: the queue
is added to hw_queues_ before ensure_doorbell_monitor() runs, and the monitor
evaluates has_kfd_queues() under hw_queue_mutex_ while holding
doorbell_thread_mutex_. Every interleaving either keeps a live monitor (it now
sees the new queue) or leaves doorbell_running_ false for ensure_doorbell_monitor()
to reap and restart -- so whenever a host-accessible queue exists, a monitor is
running or the registering call starts one.

## Issue Tracking

#8813

## Test Plan

Run CI tests
Run issue #8813 reproducer

## Test Result

- Standalone reproducer (32 queues x 3 create/destroy rounds, counting /proc/self/task): pre-fix plateaus at 12 threads and never recedes; with the fix the count returns to baseline after every round while still climbing to 12 with live queues, confirming the monitor restarts each round.
- Issue #8813 doorbell reproducer still passes (queue destroy/recreate dispatches correctly with the monitor now churning): 80/80.
- Queue-lifecycle suites (SimulatedKfd/KfdIoctl/Daemon/Interposer/GuestKfd/ Cdna4ToCdna3): 105/105.
- Full suite: 1582/1582.

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
@@ -465,19 +465,15 @@ void CommandProcessor::register_queue(HwQueue queue) {
       engine()->primary_release();
       is_primary_ = false;
     }
-    // Start the doorbell poll thread for KFD (host-accessible) queues under the
-    // same lock that guards doorbell_thread_'s companion state. Internal test
-    // queues inject doorbell events directly via schedule_event_now(). Starting the
-    // jthread while holding hw_queue_mutex_ is safe: the new thread's first action
-    // (scan_doorbells) takes hw_queue_mutex_, so it simply blocks until this scope
-    // releases. Holding the lock also removes a data race on doorbell_thread_ if a
-    // second register_queue ever runs concurrently.
-    if (start_poll && !doorbell_thread_.joinable()) {
-      util::Logger::cp(
-          [&](auto &os) { os << std::format("{}: STARTING doorbell thread", name()); });
-      doorbell_thread_ = std::jthread([this](std::stop_token stop) { doorbell_poll_loop(stop); });
-    }
   }
+  // Start (or restart) the doorbell poll thread for KFD (host-accessible) queues
+  // AFTER releasing hw_queue_mutex_. ensure_doorbell_monitor() serializes on its
+  // own doorbell_thread_mutex_ and may join a monitor that self-exited when the
+  // previous last queue was destroyed; joining while holding hw_queue_mutex_ could
+  // deadlock against that exiting monitor's final scan. Internal test queues inject
+  // doorbell events directly via schedule_event_now() and need no monitor.
+  if (start_poll)
+    ensure_doorbell_monitor();
 }
 
 void CommandProcessor::unregister_queue(uint32_t queue_id, uint32_t process_id) {
@@ -515,11 +511,39 @@ void CommandProcessor::set_doorbell_base(uint32_t process_id, void *base) {
   }
 }
 
-void CommandProcessor::stop_doorbell_monitor() {
+void CommandProcessor::ensure_doorbell_monitor() {
+  std::lock_guard<std::mutex> lock(doorbell_thread_mutex_);
+  // A monitor is already servicing this CP's queues — nothing to do. (The
+  // register→ring window is fine: the live monitor scans every registered queue
+  // each pass, so it will pick up the queue this call just added.)
+  if (doorbell_running_)
+    return;
+  // No monitor is running. If a previous one self-exited (it saw the last
+  // host-accessible queue destroyed) its jthread is joinable-but-finished; join it
+  // here before launching a fresh one so the handle does not leak. request_stop()
+  // is harmless on an already-returned thread and makes the join immediate.
   if (doorbell_thread_.joinable()) {
     doorbell_thread_.request_stop();
     doorbell_thread_.join();
   }
+  util::Logger::cp([&](auto &os) { os << std::format("{}: STARTING doorbell thread", name()); });
+  // Construct the thread BEFORE setting doorbell_running_: if the jthread
+  // constructor throws (std::system_error on thread-creation failure) the flag
+  // must stay false so a later ensure_doorbell_monitor() retries instead of
+  // no-oping forever. We still hold doorbell_thread_mutex_, and the loop's
+  // self-exit path only clears the flag under a TRY-lock of that mutex, so the
+  // new thread cannot observe or clear doorbell_running_ until we release here.
+  doorbell_thread_ = std::jthread([this](std::stop_token stop) { doorbell_poll_loop(stop); });
+  doorbell_running_ = true;
+}
+
+void CommandProcessor::stop_doorbell_monitor() {
+  std::lock_guard<std::mutex> lock(doorbell_thread_mutex_);
+  if (doorbell_thread_.joinable()) {
+    doorbell_thread_.request_stop();
+    doorbell_thread_.join();
+  }
+  doorbell_running_ = false;
 }
 
 uint64_t CommandProcessor::read_gpu_u64(uint64_t va, uint32_t vmid) const {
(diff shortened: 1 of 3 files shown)


kuhar (commented):
We should make the stale teardown path revalidate lifecycle state before stopping the monitor. At `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp:476`, `ensure_doorbell_monitor()` returns when the old monitor is running, but `handle_doorbell()` can already have computed `should_release` under `hw_queue_mutex_` and be waiting to call `stop_doorbell_monitor()` after unlocking. If a KFD queue registers in that window, registration observes the old monitor and returns, then stale teardown stops it and can release the primary again, leaving the new queue without a poller. Please revalidate after acquiring the lifecycle lock, or use a desired-state/generation scheme so old teardown work cannot invalidate a successful registration.

I think we should add targeted coverage for the doorbell-monitor lifecycle. The existing KFD/queue tests pass without observing monitor state, so they would not fail if self-retirement or restart regressed. Please cover that removing the last host-accessible queue retires the monitor within a bounded window, a later host-accessible queue restarts it and receives a doorbell, and a non-host-accessible test queue does not start it. The registration-versus-stale-teardown interleaving at `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp:476` also needs a regression test.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
```diff
@@ -515,11 +511,39 @@ void CommandProcessor::set_doorbell_base(uint32_t process_id, void *base) {
   }
 }
 
+void CommandProcessor::ensure_doorbell_monitor() {
```
lialan, line 514: CI says this needs a regression test.

kuhar (commented):
We should add focused regression coverage for the new doorbell-monitor lifecycle. The existing queue suites exercise host-accessible queue creation and destruction, but they do not verify that the monitor retires after the last queue disappears or that a later queue registration reaps and restarts it; the PR policy check also reports that no test file changed. emulation/rocjitsu/tests/gfx1250_sim_test.cpp:229 already has HostSdmaQueueForTest for this path. Please cover destroy-last-then-register-and-ring-again, plus the register-versus-self-exit edge that could otherwise leave a live queue unmonitored. If retirement needs an observable, a small test-only state or start counter would be more deterministic than sleeping against the 100 us poll cadence.

Otherwise LGTM.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
lialan: Add a gtest that keeps one CP alive, removes its last host queue, waits for retirement, then repeatedly re-registers/submits and verifies progress. Include immediate queue churn and removal with a pending retry.

lialan (approved):
I am not seeing any correctness defects. except the missing regression test coverage.


LGTM.

amd-arosa (approved):
good read, and excellent docs. lgtm

kuhar (approved):

PR #8865, opened 2026-07-20
[hip-tests] Get hip-tests building with -Werror

AidanBeltonS:
## Motivation

Improve the quality of code in the hip-test repository by making -Werror and -Wall default. This build has warnings as errors for hip-tests built for Linux

## Technical Details

  - New ENABLE_WERROR option (catch/CMakeLists.txt): defaults ON on Linux (UNIX), OFF elsewhere. When enabled it adds -Wall -Werror within the existing
  AMD-platform flag block (leaving the nvidia/nvcc path untouched). Removed the -Wno-unused-parameter suppression so unused params are now caught;
  -Wno-deprecated is retained. Disable with -DENABLE_WERROR=OFF.
  - README (catch/README.md): documented the new flag alongside ENABLE_SPIRV.

  Warning fixes (so -Werror passes)

  - C++17 standard mismatch — unit/synchronization/CMakeLists.txt forced -std=c++14, which clashed with C++17 inline variables in hip_test_common.hh (57
  errors). Changed to -std=c++17.
  - VLAs → std::vector — hipGetProcAddressMemoryApis.cc (hipStream_t stream[Ns]) and hipMemcpyAsync_old.cc (std::thread T[NUM_THREADS]), where the sizes are
   runtime values.
  - Unused parameters — dropped names on override/base-case/dispatch signatures (hip_test_listener.cc, thread_block_tile.cc); [[maybe_unused]] on the
  image-support-gated GL surface kernels (hipGLImageDataVerification.cc).
  - Ignored [[nodiscard]] hipError_t — wrapped real calls in HIP_CHECK; explicit (void) only for intentional error-drain/reset idioms (hrr_workload_test.cc,
   hipGLImageDataVerification.cc).
  - -Wpsabi — per-file -Wno-psabi for the OCP fp4/6/8 tests (unit/deviceLib/CMakeLists.txt); the AVX wide-vector return type is inherent, not a defect.
  - Uninitialized-on-default-path (real latent bugs) — zero/nullptr-init expected_value (hipStreamValue.cc) and src/dst
  (hipGraphExecMemcpyNodeSetParams.cc).
  - Macro redefinition — removed the redundant local HIPRTC_CHECK in hrr_workload_test.cc (framework already provides it).

  Dead-code cleanup (uncovered while fixing -Wunused-function)

  - Removed test_hipMallocFromPoolAsync_MThread + its worker thread_Test1 (hipMallocFromPoolAsync.cc) — the old non-thread-safe MThread helper, orphaned
  when Unit_hipMallocFromPoolAsync_MThread_{Default,Max}Thresh moved to the thread-safe runMThreadReleaseThresholdTest.
  - Collapsed the identical checkGpuEnablePal/testEnvironmentVariable stubs in hipSetEnv_helper.cc into a direct initializeHIP() call in main.

## Issue Tracking

JIRA ID : AIRUNTIME-2507

## Test Plan

Updates existing testing to improve code quality

## Test Result

N/A-->

## Submission Checklist

### DIFF
--- a/projects/hip-tests/catch/multiproc/childMalloc.cc
+++ b/projects/hip-tests/catch/multiproc/childMalloc.cc
@@ -50,13 +50,13 @@ bool testMallocFromChild() {
       testResult = false;
 
     if (A_d != nullptr) {
-      hipFree(A_d);
+      HIP_CHECK(hipFree(A_d));
     }
 
     // send the value on the write-descriptor:
     write(fd[1], &testResult, sizeof(testResult));
 
-    hipFree(A_d);
+    HIP_CHECK(hipFree(A_d));
     // close the write descriptor:
     close(fd[1]);
     exit(0);
--- a/projects/hip-tests/catch/multiproc/childMalloc.cc
+++ b/projects/hip-tests/catch/multiproc/childMalloc.cc
@@ -49,14 +49,12 @@ bool testMallocFromChild() {
     else
       testResult = false;
 
-    if (A_d != nullptr) {
-      HIP_CHECK(hipFree(A_d));
-    }
-
     // send the value on the write-descriptor:
     write(fd[1], &testResult, sizeof(testResult));
 
-    HIP_CHECK(hipFree(A_d));
+    if (A_d != nullptr) {
+      HIP_CHECK(hipFree(A_d));
+    }
     // close the write descriptor:
     close(fd[1]);
     exit(0);
--- a/projects/hip-tests/catch/unit/graph/hipGraphExecMemcpyNodeSetParams.cc
+++ b/projects/hip-tests/catch/unit/graph/hipGraphExecMemcpyNodeSetParams.cc
@@ -215,7 +215,7 @@ HIP_TEST_CASE(Unit_hipGraphExecMemcpyNodeSetParams_Negative_Parameters) {
  *    - HIP_VERSION >= 5.2
  */
 HIP_TEST_CASE(Unit_hipGraphExecMemcpyNodeSetParams_Negative_Changing_Memcpy_Direction) {
-  int *host, *dev, *src, *dst;
+  int *host, *dev, *src = nullptr, *dst = nullptr;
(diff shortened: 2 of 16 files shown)


mangupta (approved):

projects/hip-tests/catch/multiproc/childMalloc.cc
```diff
@@ -49,14 +49,12 @@ bool testMallocFromChild() {
...
-    if (A_d != nullptr) {
-      hipFree(A_d);
-    }
-
     // send the value on the write-descriptor:
     write(fd[1], &testResult, sizeof(testResult));
 
-    hipFree(A_d);
+    if (A_d != nullptr) {
+      HIP_CHECK(hipFree(A_d));
```
AlexeySachkov, line 56: BTW, `hipFree` handles null pointers just fine, so no check necessary

projects/hip-tests/catch/unit/graph/hipGraphExecMemcpyNodeSetParams.cc
```diff
@@ -215,7 +215,7 @@ HIP_TEST_CASE(Unit_hipGraphExecMemcpyNodeSetParams_Negative_Parameters) {
  *    - HIP_VERSION >= 5.2
  */
 HIP_TEST_CASE(Unit_hipGraphExecMemcpyNodeSetParams_Negative_Changing_Memcpy_Direction) {
-  int *host, *dev, *src, *dst;
+  int *host, *dev, *src = nullptr, *dst = nullptr;
```
AlexeySachkov, line 218: Presumably, this could also be fixed by putting `FAIL()` into `default:` switch case so that compiler sees that we won't go further when those are not initialized.

amd-chiranjeevi:
@mangupta  @AidanBeltonS 
Reverting this changes , as it blocking the rocm-systems bump
more details here: https://github.com/ROCm/rocm-systems/issues/9185

PR #8869, opened 2026-07-20
fix(rocDecode): Runtime environment variable cleanup

jeffqjiangNew:
## Motivation

This PR resolves the libva load issue with the sample apps and removes the environment variable workaround.

## Technical Details

- Symptom
Running a sample app linked to FFmpeg libs on an installed TheRock stack can result in a runtime libva error.
- Root cause
rocDeocde core library links to the vendor libva which comes with the TheRock stack. FFmpeg links to the system libva. These two libva copies can have different versions and difference function symbols. Since the sample app links to FFmpeg libs before rocDecode lib, the system libva enters the global scope first and gets loaded. It then tries to open a VA driver version which is not in the TheRock stack.
- The workaround
Currently we set both LD_PRELOAD and LIBVA_DRIVERS_PATH env vars to force loading the vendor libva copy.
- The fix
In each sample's CMakeLists.txt, force librocdecode first in DT_NEEDED via --push-state,--no-as-needed ${rocdecode_LIBRARY} --pop-state.

## Issue Tracking

AICV-192

## Test Plan

Remove both LD_PRELOAD and LIBVA_DRIVERS_PATH env vars and run a FFmpeg linked sample app. Run full CTest including FFmpeg demuxer use case tests.

## Test Result

The sample app runs fine. CTest passes.

## Submission Checklist

### DIFF
--- a/projects/rocdecode/docs/how-to/using-rocDecode-bitstream.rst
+++ b/projects/rocdecode/docs/how-to/using-rocDecode-bitstream.rst
@@ -58,7 +58,7 @@ The ``videodecoderaw.cpp`` example also demonstrates how to use the bitstream re
 
     .. code:: shell
 
-        export ROCM_PATH=<path to your ROCm installation>
+        export ROCM_PATH=path_to_rocm_installation
 
 .. |videodecode| replace:: ``videodecode.cpp``
 .. _videodecode: https://github.com/ROCm/rocm-systems/tree/develop/projects/rocdecode/samples/videoDecode/videodecode.cpp
--- a/projects/rocdecode/docs/how-to/using-rocDecode-bitstream.rst
+++ b/projects/rocdecode/docs/how-to/using-rocDecode-bitstream.rst
@@ -54,13 +54,11 @@ The ``videodecoderaw.cpp`` example also demonstrates how to use the bitstream re
 
 .. note:: 
 
-    To run the sample, you'll need to set the ``ROCM_PATH`` environment variable to point to the location of your ROCm installation, then set ``LD_PRELOAD`` and ``LIBVA_DRIVERS_PATH`` to point to the ROCm systems libraries and drivers:
+    To run the sample, you'll need to set the ``ROCM_PATH`` environment variable to point to the location of your ROCm installation:
 
     .. code:: shell
 
-        export LD_PRELOAD=$ROCM_PATH/lib/rocm_sysdeps/lib/librocm_sysdeps_va.so.2:$ROCM_PATH/lib/rocm_sysdeps/lib/librocm_sysdeps_va-drm.so.2
-
-        export LIBVA_DRIVERS_PATH=$ROCM_PATH/lib/rocm_sysdeps/lib/
+        export ROCM_PATH=<path to your ROCm installation>
 
 .. |videodecode| replace:: ``videodecode.cpp``
 .. _videodecode: https://github.com/ROCm/rocm-systems/tree/develop/projects/rocdecode/samples/videoDecode/videodecode.cpp
--- a/projects/rocdecode/samples/videoDecode/CMakeLists.txt
+++ b/projects/rocdecode/samples/videoDecode/CMakeLists.txt
@@ -151,7 +151,7 @@ if(HIP_FOUND AND FFMPEG_FOUND AND rocdecode_FOUND AND Threads_FOUND AND rocprofi
     # vendored VA driver. It must be injected as a raw path because CMake
     # topologically reorders normal link entries, placing FFmpeg ahead otherwise.
     if(EXISTS "${ROCM_PATH}/lib/rocm_sysdeps/lib")
-      set(LINK_LIBRARY_LIST -Wl,--no-as-needed ${rocdecode_LIBRARY} -Wl,--as-needed ${LINK_LIBRARY_LIST})
+      set(LINK_LIBRARY_LIST -Wl,--push-state,--no-as-needed ${rocdecode_LIBRARY} -Wl,--pop-state ${LINK_LIBRARY_LIST})
     endif()
     target_link_libraries(${PROJECT_NAME} ${LINK_LIBRARY_LIST})
 else()
--- a/projects/rocdecode/samples/videoDecode/CMakeLists.txt
+++ b/projects/rocdecode/samples/videoDecode/CMakeLists.txt
@@ -144,14 +144,14 @@ if(HIP_FOUND AND FFMPEG_FOUND AND rocdecode_FOUND AND Threads_FOUND AND rocprofi
       target_compile_definitions(${PROJECT_NAME} PUBLIC USE_AVCODEC_GREATER_THAN_58_134=1)
     endif()
     # On TheRock installs libva is vendored under lib/rocm_sysdeps, but FFmpeg still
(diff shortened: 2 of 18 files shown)


AryanSalmanpour (changes requested):
@jeffqjiangNew 9 rocdecode CTESTs are failing with the current PR.

https://github.com/ROCm/rocm-systems/actions/runs/29769363209/job/88449819570?pr=8869

```
The following tests FAILED:
	  7 - video_decode-HEVC (Failed)
	  8 - video_decode-AVC (Failed)
	  9 - video_decode-AV1 (Failed)
	 10 - video_decode-VP9 (Failed)
	 11 - video_decodePerf-HEVC (Failed)
	 12 - video_decodeBatch (Failed)
	 13 - video_decodeRGB-HEVC (Failed)
	 14 - video_decodeMem-HEVC (Failed)
	 15 - video_decodeRGB-Resize (Failed)
```

projects/rocdecode/test/CMakeLists.txt
```diff
@@ -150,24 +150,14 @@ endif()
 set(USING_THE_ROCK OFF)
 if(EXISTS "${ROCM_PATH}/lib/rocm_sysdeps/lib")
   set(USING_THE_ROCK ON)
-  set(THEROCK_SYSDEPS_LIB "${ROCM_PATH}/lib/rocm_sysdeps/lib")
```
AryanSalmanpour, line 153: Please don't remove these codes. This causes failures on CI.

AryanSalmanpour (changes requested):

projects/rocdecode/test/CMakeLists.txt
jeffqjiangNew: Yes the change in this file resulted in the old libva load issue. Change reverted. CI now passed.

projects/rocdecode/docs/how-to/using-rocDecode-bitstream.rst
```diff
@@ -54,13 +54,11 @@ The ``videodecoderaw.cpp`` example also demonstrates how to use the bitstream re
 
 .. note:: 
 
-    To run the sample, you'll need to set the ``ROCM_PATH`` environment variable to point to the location of your ROCm installation, then set ``LD_PRELOAD`` and ``LIBVA_DRIVERS_PATH`` to point to the ROCm systems libraries and drivers:
```
AryanSalmanpour, line 57: Please do not remove or modify the LD_PRELOAD and LIBVA_DRIVERS_PATH sections in the document. These settings are important for end users and should remain as documented.

AryanSalmanpour (changes requested):

projects/rocdecode/docs/how-to/using-rocDecode-bitstream.rst
jeffqjiangNew: Removing LD_PRELOAD and LIBVA_DRIVERS_PATH is the sole purpose of this PR: to make user's life easier. rocDecode should not ask the user to do extra env var setup when possible. Lets discuss offline.

AryanSalmanpour: @jeffqjiangNew The previous CI result indicated that both environments are required to pass. Without them, the CI fails. This demonstrates that the change made in this PR does not adequately resolve the issue of removing these environments.

projects/rocdecode/samples/videoDecode/CMakeLists.txt
```diff
@@ -143,6 +143,16 @@ if(HIP_FOUND AND FFMPEG_FOUND AND rocdecode_FOUND AND Threads_FOUND AND rocprofi
...
+    # On TheRock installs libva is vendored under lib/rocm_sysdeps, but FFmpeg still
+    # links the system libva, so both load at runtime. Force the vendored libva to be
+    # an early direct dependency (--no-as-needed keeps it despite no direct symbol
+    # references) so its symbols win and it loads the matching VA driver.
+    if(EXISTS "${ROCM_PATH}/lib/rocm_sysdeps/lib")
+      find_package(Libva QUIET)
+      if(Libva_FOUND)
+        set(LINK_LIBRARY_LIST -Wl,--no-as-needed Libva::va Libva::va_drm -Wl,--as-needed ${LINK_LIBRARY_LIST})
+      endif()
+    endif()
```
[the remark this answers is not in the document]
jeffqjiangNew: find_library(NAMES va) looks for a file literally named libva.so. The sysdeps dir is the first HINT, so whether it wins depends entirely on whether a libva.so linker symlink exists there. libva.so should always exist in sysdeps dir. If there is no libva.so, VAAPI path won't be functional.

projects/rocdecode/docs/how-to/using-rocDecode-bitstream.rst
jeffqjiangNew: @AryanSalmanpour The CTest builds the sample app on the fly. That's why we need to add the two env vars in the test cmake file. If we were to use the pre-built sample app with the fix in CTest, CI will pass. These are different build and use processes.

AryanSalmanpour: @jeffqjiangNew That's exactly why we should not remove the documentation for these environment variables. End users typically build the samples themselves. As a result, these environment variables remain relevant to their workflow and should be documented as-is.

jeffqjiangNew: @AryanSalmanpour The purpose of this PR is to remove the env var setup from the user while making the sample app functional.

projects/rocdecode/docs/how-to/using-rocDecode-bitstream.rst
```diff
@@ -54,13 +54,11 @@ The ``videodecoderaw.cpp`` example also demonstrates how to use the bitstream re
...
 
-    To run the sample, you'll need to set the ``ROCM_PATH`` environment variable to point to the location of your ROCm installation, then set ``LD_PRELOAD`` and ``LIBVA_DRIVERS_PATH`` to point to the ROCm systems libraries and drivers:
+    To run the sample, you'll need to set the ``ROCM_PATH`` environment variable to point to the location of your ROCm installation:
 
     .. code:: shell
 
-        export LD_PRELOAD=$ROCM_PATH/lib/rocm_sysdeps/lib/librocm_sysdeps_va.so.2:$ROCM_PATH/lib/rocm_sysdeps/lib/librocm_sysdeps_va-drm.so.2
-
-        export LIBVA_DRIVERS_PATH=$ROCM_PATH/lib/rocm_sysdeps/lib/
+        export ROCM_PATH=<path to your ROCm installation>
```
spolifroni-amd:
```suggestion
        export ROCM_PATH=path_to_rocm_installation
```

All instances should be changed to this. However, I think maybe this should be added to the install docs instead so that it gets set right away upon installation. Then we won't need the example here and we can replace this with "ensure that ROCM_PATH has been set" or something like that

spolifroni-amd (changes requested):
The rocm_path lines need to be updated. But it would actually be better to make setting ROCM_PATH part of the installation steps that way it's set once and we're done.

AryanSalmanpour (approved):

spolifroni-amd (approved):

Issue #8879, opened 2026-07-20
[rocjitsu] DBI Roadmap


accauble:
## Summary

We are planning to make rocjitsu an instrumentation platform. The trampoline path can now redirect selected GPU instructions, but there is still tons more work to be done to get to a point where we can call probes, preserve state, write host-visible records, and build tools such as a basic-block counter. This issue tracks that remaining work so implementation PRs stay reviewable and design decisions are explicit.

## What Already Works

We have implemented two very basic steps that serve as a good implementation starting point (IMO). Future PRs should be able to be smaller now that the DBI workflow is there. The two basic steps were "instrument with a nop" (aka an inlined nop; PR #6742) and "instrument with a call to an empty probe" (PR #7534). 

Capabilities
- [x] Patch given `.text` offsets with jumps to trampolines
- [x] Generate a trampoline that adds either a `nop` instruction or a call to a probe that just returns back via `s30:s31`.
  - [x] Relocate the original instruction and branch back
  - [x] If a probe, save `scc`, load probe target address into dead register pair , use `s_swappc` with `s30:s31`, and restore `scc`
- [x] Find the instrumentation probe (resolve the symbol), verify that the probe body conforms to current rocjitsu requirements (right now it cannot take arguments and must return through `s30:s31`), and copy it into our `.text` section

Major constraints
- Requires kernels to have 32 SGPRs available (since we return through `s30:s31`)
- The instrumentation is always placed after the original instruction
- No API yet
- Requires dead registers for saving/restoring
- "All-or-nothing" instrumentation: if one instrumentation site fails, then it doesn't get instrumented.

## Driving Tool

An EXEC-weighted FLOP counter, with a basic block counter as a convenient intermediate step. It helps force the framework's remaining gaps into a schedule, including whole-kernel instrumentation, per-dispatch buffers, concurrent drain, and relocation reach. This particular design will instrument "regions", which are basic blocks broken up by EXEC-modifying instructions. The device just needs to pass the EXEC value (along with other metadata back to the host), and the host aggregates the counts across the regions. A separate static analysis will handle all the static information.

 ### Design Plan

I expect to do the following in this order (ish).

- [ ] Pass arguments to probes (PR #11332)
- [ ] Per-dispatch buffer delivery via kerarg payload
- [ ] Concurrent drain protocol
- [ ] Probe writes record with site, ID, mask
- [ ] Pass EXEC as an argument
- [ ] Producer memory contract
[...]

PR #8888, opened 2026-07-20
feat(rocjitsu): Register Spilling for CDNA3-4/RDNA4

theSK2005:
## Motivation

DBI for arbitrary probes requires being able to spill registers that are live at the anchor if the probe would clobber them. This PR enables spilling for VGPRs, SGPRs, EXEC, VCC, and M0.
VGPRs are spilled directly into scratch. SGPRs are spilled into scratch using v_writelane/v_readlane into a bridge VGPR. EXEC, VCC, and M0 are stored into a dead SGPR.

Relies on SpillManager changes made in #8406 and moves DBT's visit_kernel_descriptors functions to a shared location for kernel discovery. Moves instruction_builders into a new builders folder.

Defers:
- Multi-kernel and multi-.text code objects
- RDNA2-3.5 (AmdGpuCodeObject doesn't support, trivial to add register spilling once support is added there)
- CDNA1-2 (ABI convention is to use MUBUF instructions instead of FLAT to store, requires a lot of additional overhead)
- AccVGPR
- FLAT_SCRATCH-clobbering probes
- EXEC/VCC spilling only when clobbered (relies on 100% accurate def/use tracking, can't guarantee now)

## Test Plan
- Unit tests for spill plans
- Static tests that patch and decode the ELF file (similar to existing nop probe tests)
- End-to-end simulator/behavior tests that execute a patched kernel
 
## Test Result
All tests pass

## Submission Checklist


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp
```diff
@@ -74,6 +74,47 @@ void append_word(std::vector<uint8_t> &dst, uint32_t w) {
...
+  std::vector<uint32_t> epilogue;
+};
+
+[[nodiscard]] SpillBracket build_spill_bracket(const std::vector<SpillSlot> &vgpr_spills,
+                                               const std::vector<SgprSpillSlot> &sgpr_spills,
+                                               uint16_t bridge_vgpr, rj_code_arch_t arch) {
+  constexpr uint16_t kUniformLane = 0; // An SGPR is uniform; one lane suffices.
+  SpillBracket bracket;
+
+  // VGPRs: batch stores, then batch loads + a single wait before the readlanes.
```
accauble: We cannot spill VGPRs without making `EXEC` all 1s (and thus also saving and restoring `EXEC`). I personally don't mind postponing VGPR spilling to a follow-up PR, especially since your `EXEC` PR hasn't landed yet. If you choose to postpone, just make sure it's logged in the DBI issue (https://github.com/ROCm/rocm-systems/issues/8879) and that the plans refuse to instrument an instruction that has a VGPR conflict.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/amdgpu_elf.h
```diff
@@ -129,6 +129,26 @@ inline constexpr rj_code_arch_t arch_for_elf_mach(uint32_t mach) {
...
+    return 16;
+  case ROCJITSU_CODE_ARCH_CDNA3:
+  case ROCJITSU_CODE_ARCH_CDNA4:
+  case ROCJITSU_CODE_ARCH_RDNA1:
+  case ROCJITSU_CODE_ARCH_RDNA2:
+  case ROCJITSU_CODE_ARCH_RDNA3:
+  case ROCJITSU_CODE_ARCH_RDNA3_5:
+  case ROCJITSU_CODE_ARCH_RDNA4:
+    return 8;
+  default:
```
accauble: Let's fail on default so that we know to update this when adding a new ISA.

emulation/rocjitsu/tests/dbi/dbi_spill_sim_test.cpp
```diff
@@ -0,0 +1,340 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: MIT
+
+/// @file dbi_spill_sim_test.cpp
```
accauble, line 4: This description is stale

emulation/rocjitsu/tests/dbi/dbi_spill_sim_test.cpp
```diff
@@ -0,0 +1,340 @@
...
+  for (uint32_t lane = 0; lane < kWaveSize; ++lane)
+    EXPECT_EQ(v3[lane], kSentinel)
+        << "lane " << lane << ": v2 was not restored after the probe clobbered it";
+}
+
+// Negative control: prove the *restore* is what preserves the value. Overwrite
+// the epilogue scratch_load (the restore) with s_nop, so v2 stays clobbered (0)
+// and the copy into v3 reads 0. Then run the intact patched text to confirm the
+// zero result is caused by the missing restore, not unrelated corruption.
+TEST_F(DbiSpillSimFixture, MissingRestoreLeavesVgprClobbered) {
```
accauble: Perfect.

emulation/rocjitsu/tests/patch/dbi_test_fixtures.h
```diff
@@ -0,0 +1,329 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: MIT
+
+/// @file gfx950_test_fixtures.h
+/// @brief Self-contained, header-only builders for minimal gfx950 (CDNA4) ELFs
+///        used by DBI probe/spill tests: a single-kernel target ELF (with a
+///        discoverable `.kd` descriptor) and a probe ELF exporting one function
+///        symbol, plus small readback helpers.
+///
+/// These mirror the in-file helpers in tests/patch/instrumentor_test.cpp (which
```
accauble: Since you've already creating a header that can be shared and it copies from `tests/patch/instrumentor_test.cpp`, please finish the refactor and have it use the new header. This will require cleaning this file up a bit more though. For example, the constants should probably reference gfx950 if they are ISA-specific.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instruction_builder.h
```diff
@@ -695,4 +696,116 @@ build_s_nop(uint16_t cycles = 0, rj_code_arch_t arch = ROCJITSU_CODE_ARCH_RDNA4)
...
+}
+
+/// @brief VOP3 src encoding for a non-negative inline integer (lane selector).
+[[nodiscard]] inline constexpr uint16_t vop3_inline_uint(uint16_t value) {
+  return static_cast<uint16_t>(kScalarPositiveInlineBase + value);
+}
+
+/// @brief Encode v_writelane_b32: bridge SGPR @p sgpr_src into lane of @p vgpr_dst.
+[[nodiscard]] inline std::array<uint32_t, 2>
+build_v_writelane_b32(uint16_t vgpr_dst, uint16_t sgpr_src, uint16_t lane, rj_code_arch_t arch) {
```
accauble: Is there a reason you are only implementing rdna4 and cdna4 to these builders?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
```diff
@@ -90,6 +92,21 @@ std::string reg_name(RegisterRef ref) {
   return std::string(prefix) + std::to_string(ref.index);
 }
 
+// Largest positive byte offset encodable in the scratch store/load offset field,
+// per arch. 0 means the arch has no scratch spill emitter (spilling unsupported).
+// The fields are signed, so the cap is the positive half: CDNA4 writes the 12-bit
+// FLAT offset (pad_12 left 0), RDNA4 has a signed 24-bit VSCRATCH ioffset.
+uint32_t max_scratch_offset_bytes(rj_code_arch_t arch) {
```
accauble, line 118: Please add for other supported ISAs as well.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
```diff
@@ -274,6 +291,131 @@ bool check_spill_policy(const RegisterSet &spill_set, SpillPolicy policy, std::s
   return true;
 }
 
+bool plan_vgpr_spills(const RegisterSet &spill_set, SpillManager &spills, rj_code_arch_t arch,
+                      std::vector<SpillSlot> &out, std::string *error_out) {
+  out.clear();
+
+  // VGPR-first slice: SGPRs/AccVGPRs need a bridge VGPR + writelane/readlane that
```
accauble: I don't understand this comment

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
```diff
@@ -575,14 +725,51 @@ InstrumentedCodeObjectDebug Instrumentor::patch_with_debug_summaries() {
...
+      // fail closed inside the planners.
       const RegisterSet clobbers =
           compute_instrumentation_clobbers(*summary, plan.builder_clobbers);
       const RegisterSet spill = compute_spill_set(live, clobbers);
-      if (!check_spill_policy(spill, SpillPolicy::NoSpillsSupported, &err)) {
-        result.errors.push_back(std::move(err));
-        continue;
+      if (!spill.none()) {
+        // Single-kernel assumption: spilling needs exactly one kernel descriptor
+        // with non-zero fixed scratch to grow (see §9 of the spill plan).
```
accauble: Remove ref to your spill plan

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.h
accauble: Stale

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.h
```diff
@@ -249,8 +250,9 @@ validate_anchor(const Instruction &anchor, uint64_t anchor_offset,
 /// now so the contract has a name and a later spill-capable slice can add a
 /// value without a rename.
 enum class SpillPolicy {
-  NoSpillsSupported, ///< Any non-empty spill set is a hard failure.
-  // TODO: SpillsSupported once a bootstrap save/restore sequence exists.
+  NoSpillsSupported,   ///< Any non-empty spill set is a hard failure.
+  VgprSpillsSupported, ///< VGPRs spill to scratch; SGPRs/AccVGPRs still fail closed.
```
accauble: Are these used somewhere?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.h
```diff
@@ -269,6 +271,30 @@ enum class SpillPolicy {
 [[nodiscard]] bool check_spill_policy(const RegisterSet &spill_set, SpillPolicy policy,
```
accauble: Either update this function or if you don't have use for it, then you can remove it.

accauble (changes requested):
Good start. Could you clarify what you planned to accomplish with this PR and what you see as deferred? What should users be able to expect from it? The AccVGPRs is clearly deferred, as is multi-kernel instrumentation. Is there anything else you can think of?

How do you see spilling to dead registers fitting in here?

Additionally, I don't mind keeping non-RDNA4/CDNA4 *untested*, but DBI should be pretty architecture agnostic. There are only a few places where you had to add ISA-specific stuff, so let's update those for all supported ISAs.

Couple tests that the agent couldn't find and that I think would be good to have:
* A check that a multi-kernel instrumentation fails
* You have separate VGPR and SGPR spilling tests. I'd be interested to one that does both and make sure those play well together.
* I'd also make sure spilling multiple SGPRs plays nicely too.

Please update the dbi design documentation too.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp
theSK2005: SGPR spills use the same scratch load dword instruction as used in VGPR spills that masks on EXEC so I think I'll have to support EXEC spilling anyways. I'll look into this.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.h
theSK2005: I'll remove it, I don't use it

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.h
theSK2005: No, they aren't

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
```diff
@@ -865,6 +865,18 @@ bool CodeObjectPatcher::patch_kernel_descriptor(uint64_t file_offset,
   return true;
 }
 
+bool CodeObjectPatcher::set_private_segment_fixed_size(uint64_t descriptor_file_offset,
```
Groverkss, line 1320: Do we not already have this? What is DBT doing for this?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instruction_builder.h
```diff
@@ -695,4 +743,114 @@ build_s_nop(uint16_t cycles = 0, rj_code_arch_t arch = ROCJITSU_CODE_ARCH_RDNA4)
   return build_sopc_encoding(arch, sopc_op_cmp_lg_u32(arch), ssrc0, ssrc1);
 }
 
+// DBI register-spilling primitives (CDNA4 and RDNA4 only). These are the
```
Groverkss: I'm not sure if we should be implement architecture specific filling details here. DBT uses architecture specific files which do this. It might be worth looking at that and maybe sharing the building stuff there. I would expect that there should be a single api with arch specific implementations for these things. instruction_builder.h/.cpp was supposed to be for simple things.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/amdgpu_code_object.cpp
```diff
@@ -307,33 +307,87 @@ namespace {
...
+// file offset (for descriptor writeback).
+struct LocatedKernelDescriptor {
+  rocr::llvm::amdhsa::kernel_descriptor_t desc;
+  uint64_t file_offset;
+};
+
+// Find the allocated section whose address range covers @p kd_vaddr and decode
+// the descriptor out of that section's own bytes (no ELF re-walk). Returns
+// nullopt when no section covers the descriptor or it does not fully fit.
+[[nodiscard]] std::optional<LocatedKernelDescriptor>
```
Groverkss: What does DBT do for this? We should reuse code between DBT/DBI please. I'm sure DBT knows how to locate kernel descriptors.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/amdgpu_code_object.h
```diff
@@ -19,6 +19,16 @@
 
 namespace rocjitsu {
 
+/// @brief One kernel descriptor discovered in a code object, with the coordinates
```
Groverkss: I doubt we need all this. DBT gets all this information already, we shouldn't duplicate it and reuse it.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/amdgpu_elf.h
```diff
@@ -129,6 +129,26 @@ inline constexpr rj_code_arch_t arch_for_elf_mach(uint32_t mach) {
   }
 }
 
+/// @brief VGPR-count encoding granule for @p arch. Descriptor VGPR count =
+///        (granulated_workitem_vgpr_count + 1) * granule. Matches LLVM's
+///        AMDGPUBaseInfo::getVGPREncodingGranule().
+inline constexpr uint32_t vgpr_encoding_granule(rj_code_arch_t arch) {
```
Groverkss: This shouldn't like in amdgpu_elf right?

Groverkss (changes requested):

Groverkss:
Please try to reuse as much as possible between DBT/DBI. We don't want to have two seperate implementations of these things.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instruction_builder.h
theSK2005: What is the reasoning for separating builders by architecture? If the issue is that this shouldn't be in instruction_builder.h/.cpp, I can move the new builders to a separate file to be shared between DBT and DBI. Is [this](https://github.com/ROCm/rocm-systems/blob/develop/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/semantic/cdna3_scratch.cpp) the equivalent file for DBT?

theSK2005:
@Groverkss I think that KdTranslation and KernelDescriptorTranslator are too bulky for my use here. I'll create a shared version of the kernel descriptor structs/functions I have here that can be called by both DBT and DBI.

theSK2005:
@Groverkss I'll put the DBT/DBI sharing stuff in a separate stacked PR, so that those issues don't block this PR. 

Edit: That PR has now been added #9046. Leaving DBT/DBI reuse refactoring comments open here until it is confirmed that the other PR properly addresses them.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
theSK2005: Currently, we only support CDNA4 and RDNA4, when I'll add other archs here as they are supported

kuhar:
It doesn't build for me:
```
../rocm-systems/emulation/rocjitsu/tests/patch/instrumentor_test.cpp:2369:19: error: loop variable ‘<structured bindings>’ creates a copy from type ‘const std::pair<short unsigned int, unsigned int>’ [-Werror=range-loop-construct]
 2369 |   for (const auto [sgpr, off] :
      |                   ^~~~~~~~~~~
../rocm-systems/emulation/rocjitsu/tests/patch/instrumentor_test.cpp:2369:19: note: use reference type to prevent copying
 2369 |   for (const auto [sgpr, off] :
      |                   ^~~~~~~~~~~
      |                   &
cc1plus: all warnings being treated as errors
```

theSK2005:
@kuhar Can you try building again? I made a commit that should fix that.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
theSK2005: This is a narrower descriptor setter, DBT has a full descriptor rewrite.

kuhar (commented):
Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp:749`

We should either model implicit EXEC/VCC definitions or reject probe bodies that contain them. build_probe_clobber_summary() scans only explicit special-state operands, and the source notes that v_cmp and v_cmpx implicit writes are invisible. Those probes therefore leave touches_vcc or touches_exec false, so the trampoline does not preserve that state before the relocated instruction. Please add compare-based coverage for the chosen fail-closed or preservation path.

Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp:237`

This would be safer if every temporary search were capped by kernel_sgpr_count, or the descriptor were grown consistently. plan_probe_call() scans all REGISTER_SET_ALLOCATABLE_SGPRS; with a 32-SGPR kernel and s0-s29 unavailable, it can select s32:s33 even though the object-level check establishes only that s30:s31 exists. Please thread the actual allocation through the planner and fail closed when no in-range temporary remains.

Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp:102`

We should add an RDNA4 store-counter wait before the first spill reload, and make the static patching test assert that ordering. `build_spill_bracket()` emits scratch stores in the prologue but only `s_wait_loadcnt 0` in the epilogue; RDNA4 tracks those stores on STORECNT, so that wait does not order the reloads. The new `Rdna4SpillsLiveClobberedVgpr` test checks only the load wait, and the functional simulator completes memory synchronously, so its passing round-trip cannot catch the missing store fence.

Original anchor: `emulation/rocjitsu/tests/patch/instrumentor_test.cpp:2150`

I think we should rename this test and tighten the diagnostic assertion to the guard it actually reaches. In the current `Cdna4ScratchOffsetPastFieldFailsClosed`, `private_bytes=0x1000` creates `SpillManager(4096, 4099)`, so the first 4-byte slot fails the per-lane scratch limit (`4100 > 4099`) before the 12-bit offset check. `find("scratch")` matches both diagnostics. Either assert `"per-lane scratch limit"` and rename the test accordingly, or drop this integration test and rely on `PlanVgprSpillsFailsWhenOffsetExceedsField` for the offset branch.

Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp:325`

We should restore the anchor EXEC mask before calling the probe, then widen it again only for spill reloads. At a spilling site the prologue sets EXEC=-1, and s_swappc_b64 runs before any restore, so a per-lane probe also executes in lanes that were inactive at the anchor. If that behavior is intentionally deferred, please fail closed for affected probes or document and track the limitation explicitly.

The shared DBI test support would be easier to maintain at the common test root. `tests/patch/dbi_test_fixtures.h` is now included by `tests/code`, `tests/dbi`, and `tests/patch`, so please move or rename it to something like `tests/dbi_test_util.h`, update its stale lines 12-16 (the instrumentor test no longer owns a duplicate copy), and make `dbi_spill_sim_test.cpp` include `<array>` and `<string_view>` directly instead of relying on the helper transitively.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.h
kuhar: We should update the remaining roadmap text too. The header still lists probe-call bodies and EXEC policy management as future work at lines 28-30, and lines 239-245 still say builder planning and dead-register selection land later, although this patch implements those paths.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
```diff
@@ -107,35 +108,54 @@ std::string reg_name(RegisterRef ref) {
...
+  case ROCJITSU_CODE_ARCH_CDNA3:
+  case ROCJITSU_CODE_ARCH_CDNA4:
+  case ROCJITSU_CODE_ARCH_RDNA1:
+  case ROCJITSU_CODE_ARCH_RDNA2:
+  case ROCJITSU_CODE_ARCH_RDNA3:
+  case ROCJITSU_CODE_ARCH_RDNA3_5:
+  case ROCJITSU_CODE_ARCH_RDNA4:
+    return 8;
+  default:
+    return 4;
```
accauble: Let's fail on default so that we know to update this when adding a new ISA.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.h
accauble: Marked resolved but kuhar's comments not addressed AFAICT

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
```diff
@@ -549,6 +677,15 @@ InstrumentedCodeObjectDebug Instrumentor::patch_with_debug_summaries() {
   // by the calling convention, so the kernel must allocate up to it.
   const std::optional<uint32_t> kernel_sgpr_count = obj_.min_kernel_sgpr_count(arch_);
 
+  // Register spilling targets a single-kernel code object for now: any site that
```
accauble, line 778: You're calling `min_kernel_sgpr_count` which does the kernel descriptor scan and then you immediately do the scan here again. Is there a cleaner way to handle this?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/amdgpu_code_object.cpp
```diff
@@ -287,26 +313,21 @@ namespace {
...
-      if (base == 0 || kd_vaddr < base)
-        continue;
-      const uint64_t off = kd_vaddr - base;
-      if (off + sizeof(KD) > section->size())
-        continue;
-      KD desc;
-      std::memcpy(&desc, section->data() + off, sizeof(desc));
-      const uint32_t granulated = AMDHSA_BITS_GET(
-          desc.compute_pgm_rsrc1, kd::COMPUTE_PGM_RSRC1_GRANULATED_WAVEFRONT_SGPR_COUNT);
+  // Discover descriptors through the shared scanner (the single DBT/DBI discovery
```
accauble: Comment is not entirely accurate since it actually means you are walking this twice in the DBI path (see comment in instrumentor).

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/builders/instruction_builder.h
```diff
@@ -19,6 +19,7 @@
```
accauble, line 32: Could we include a comment on the design here? What is this for and what is it NOT for?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/builders/instruction_builder.h
```diff
@@ -68,6 +69,14 @@ inline constexpr uint32_t kSopcEncodingPrefix = 0x17E;
 inline constexpr uint32_t kSopkEncodingPrefix = 0xB;
 inline constexpr uint16_t kScalarPositiveInlineBase = 128;
 inline constexpr uint16_t kDelayAluSaluDep1 = 9;
+// Scalar-operand codes for special registers, used as the ssrc/sdst of a plain
+// s_mov to save/restore them across a probe call. VCC_LO and EXEC_LO are stable
+// across every AMDGPU generation; M0 moved (see scalar_operand_m0).
```
accauble: Not sure what the M0 ref is doing here

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/builders/instruction_builder.h
```diff
@@ -68,6 +69,14 @@ inline constexpr uint32_t kSopcEncodingPrefix = 0x17E;
...
+inline constexpr uint16_t kScalarOperandVccLo = 106;
```
accauble: These three constants should not be defined this way. They should come from operand_types.h

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/builders/instruction_builder.h
```diff
@@ -297,6 +306,30 @@ inline constexpr uint16_t kDelayAluSaluDep1 = 9;
   return static_cast<uint16_t>(kScalarPositiveInlineBase + value);
 }
 
+/// @brief Scalar-operand code for M0 on @p arch.
+///
+/// M0 is operand 124 on gfx9 / gfx10.x (CDNA1-4, RDNA1/2) but was moved to 125
+/// on gfx11+ (RDNA3/3.5/4, gfx1250), where 124 became NULL. Callers that encode
+/// an s_mov touching M0 must use the per-arch code.
+[[nodiscard]] inline constexpr uint16_t scalar_operand_m0(rj_code_arch_t arch) {
+  switch (arch) {
```
accauble, line 330: Same here, these should come from operand_types

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.h
```diff
@@ -33,6 +33,30 @@ struct InlineAsmItem {
   std::vector<uint32_t> words;
 };
 
+/// @brief One saved register: a VGPR live at the anchor and clobbered by
+///        instrumentation, with its stable per-lane byte offset in the DBI
+///        spill zone (assigned by SpillManager).
+struct SpillSlot {
```
accauble, line 46: You have two structs that look exactly the same here. Do they need o be distinguished as such?

accauble (changes requested):
Overall, I think this is looking a lot better. I have mostly minor/nit comments and questions I think.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.h
```diff
@@ -33,6 +33,30 @@ struct InlineAsmItem {
...
+///        with its stable per-lane byte offset in the DBI spill zone (assigned by
+///        SpillManager). @ref cls selects the emit path in build_spill_bracket: a
+///        VGPR spills straight to scratch; an SGPR is bridged through a VGPR lane
+///        (writelane/readlane) because SGPRs cannot reach scratch directly.
+///
+/// VGPR and SGPR slots are held in separate vectors (TrampolinePlan::vgpr_spills /
+/// sgpr_spills), so @ref cls is redundant with the owning vector but records the
+/// register file explicitly for the emitter and future classes (AccVGPR).
+struct SpillSlot {
+  RegClass cls = RegClass::VGPR; ///< Register file: VGPR (direct) or SGPR (bridged).
```
accauble: Do you see this getting used?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/builders/instruction_builder.h
```diff
@@ -66,8 +89,24 @@ inline constexpr uint32_t kSopcEncodingPrefix = 0x17E;
...
+// from the operand table and checked in the same test as the other inline codes.
+inline constexpr uint16_t kScalarPositiveInlineBase = cdna4::OPR_SRC_POS_INT_MIN;
 inline constexpr uint16_t kDelayAluSaluDep1 = 9;
+// Scalar-operand codes for special registers, used as the ssrc/sdst of a plain
+// s_mov to save/restore them across a probe call. VCC_LO and EXEC_LO are stable
+// across every AMDGPU generation, so we take them from a representative arch's
+// generated operand table; the generation-stability check lives in
+// InstructionBuilder.ScalarOperandCodesMatchGeneratedTables. M0 moved between
+// generations and is resolved per-arch (see scalar_operand_m0).
+inline constexpr uint16_t kScalarOperandVccLo = cdna4::OPR_SDST_VCC_LO;
```
accauble: This is not a cdna4-only document and while these three constants may be the same across supported ISAs, there is no guarantee that they will be. I would thus prefer they not be global constants like this

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp
```diff
@@ -215,8 +322,41 @@ std::optional<TrampolineBytes> TrampolineBuilder::emit_probe_call(const Trampoli
...
+
+  // Full-mask the spill store so a probe that widens EXEC cannot leave inactive-
+  // lane copies unsaved. EXEC was just saved and is restored after the loads.
+  if (full_mask_exec)
+    env.push_back(build_s_mov_b64(kScalarOperandExecLo, kScalarInlineNegOne, plan.arch));
+
+  // Spill saves: store each live+clobbered register before the call.
+  env.insert(env.end(), spill.prologue.begin(), spill.prologue.end());
+
+  // Restore the anchor EXEC before the call so the probe runs under the anchor mask,
```
accauble, line 424: I understand why you did it this way and I would not call this a blocker (at some point we will have to include a policy for handling EXEC). However, I am of the opinion that the default policy should actually be to keep the EXEC register -1s.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp
```diff
@@ -280,18 +303,128 @@ RegisterSet compute_spill_set(const RegisterSet &live_at_anchor,
...
+  if (spill_set.none())
+    return true;
+
+  const uint32_t max_offset = max_scratch_offset_bytes(arch);
+  if (max_offset == 0) {
+    report(error_out, "probe-call spilling not supported for target architecture");
+    return false;
+  }
+
+  // Bridge = lowest VGPR neither live at the anchor nor already a spilled VGPR,
```
accauble: Why can't the bridge be a spilled vgpr? Wouldn't that even be a better choice, since you spill VGPRs first....so that register has already been saved.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp
```diff
@@ -74,6 +82,54 @@ void append_word(std::vector<uint8_t> &dst, uint32_t w) {
...
+  std::vector<uint32_t> epilogue;
+};
+
+[[nodiscard]] SpillBracket build_spill_bracket(const std::vector<SpillSlot> &vgpr_spills,
+                                               const std::vector<SpillSlot> &sgpr_spills,
+                                               uint16_t bridge_vgpr, rj_code_arch_t arch) {
+  constexpr uint16_t kUniformLane = 0; // An SGPR is uniform; one lane suffices.
+  SpillBracket bracket;
+
+  // VGPRs: batch stores, then batch loads + a single wait before the readlanes.
```
accauble: I wonder if you need to make sure that all loads have completed before handling the spills

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp
```diff
@@ -215,8 +322,41 @@ std::optional<TrampolineBytes> TrampolineBuilder::emit_probe_call(const Trampoli
...
+
   std::vector<uint32_t> env;
 
+  // The site spills iff there is anything to spill; this drives the EXEC full-mask
+  // toggles only. EXEC save/restore is decided by special_state_saves membership.
+  const bool full_mask_exec = !plan.vgpr_spills.empty() || !plan.sgpr_spills.empty();
+
+  // SGPR pair holding the saved anchor EXEC (populated by the save loop below).
+  // Reused to restore the anchor mask before the call; always present when spilling.
+  uint16_t exec_temp = 0;
```
accauble, line 393: This initialization seems dangerous. Perhaps we should verify that it gets set?

accauble (changes requested):
Some non-blocking comments. The one blocking comment I have is that the `docs/dbi-design.md` document needs to get updated.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.h
theSK2005: I feel like it's a nice thing to have so that a SpillSlot doesn't require the context of the vector it's in to know what RegClass it represents

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp
theSK2005: talked and resolved

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp
theSK2005: Yes, I think you're right about this

kuhar (commented):
This would be safer if probe calls failed closed when descriptor discovery cannot provide `kernel_sgpr_count`. The new planner cap handles a known count, but `instrumentor.cpp:766` leaves the device-wide default when the optional is empty, so a descriptorless object can again choose temporary SGPRs beyond the kernel allocation. `AmdGpuCodeObjectSgpr.NoKernelDescriptorReturnsNullopt` and descriptorless `InstrumentorProbePatch.EmitsValidElfWithProbeMetadata` cover both halves of the gap. Please require a known count for probe calls, or otherwise derive and update the actual allocation.

Location: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp:766`.

We should add an RDNA4 store-counter wait before the first spill reload, and make the static patching test assert that ordering. `build_spill_bracket()` emits scratch stores in the prologue but only `s_wait_loadcnt 0` in the epilogue; RDNA4 tracks those stores on STORECNT, so that wait does not order the reloads. The new `Rdna4SpillsLiveClobberedVgpr` test checks only the load wait, and the functional simulator completes memory synchronously, so its passing round-trip cannot catch the missing store fence.

Location: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp:102`.

I think we should rename this test and tighten the diagnostic assertion to the guard it actually reaches. In the current `Cdna4ScratchOffsetPastFieldFailsClosed`, `private_bytes=0x1000` creates `SpillManager(4096, 4099)`, so the first 4-byte slot fails the per-lane scratch limit (`4100 > 4099`) before the 12-bit offset check. `find("scratch")` matches both diagnostics. Either assert `"per-lane scratch limit"` and rename the test accordingly, or drop this integration test and rely on `PlanVgprSpillsFailsWhenOffsetExceedsField` for the offset branch.

Location: `emulation/rocjitsu/tests/patch/instrumentor_test.cpp:2150`.

We should restore the anchor EXEC mask before calling the probe, then widen it again only for spill reloads. At a spilling site the prologue sets `EXEC=-1`, and `s_swappc_b64` runs before any restore, so a per-lane probe also executes in lanes that were inactive at the anchor. If that behavior is intentionally deferred, please fail closed for affected probes or document and track the limitation explicitly.

Location: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp:325`.

The shared DBI test support would be easier to maintain at the common test root. `tests/patch/dbi_test_fixtures.h` is now included by `tests/code`, `tests/dbi`, and `tests/patch`, so please move or rename it to something like `tests/dbi_test_util.h`, update its stale lines 12-16 (the instrumentor test no longer owns a duplicate copy), and make `dbi_spill_sim_test.cpp` include `<array>` and `<string_view>` directly instead of relying on the helper transitively.

Location: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/builders/spill_builders.h:7`

Please update the file-level summary to match the public helpers below. It currently says this header provides “the load-completion wait,” but `build_wait_stores_complete()` is now equally important, especially for RDNA4's split counters. Naming both store- and load-completion waits here will keep the advertised scope aligned with the implementation.

Location: `emulation/rocjitsu/tests/dbi/dbi_spill_sim_test.cpp:80`

Please bring the local architecture descriptions in line with the three-architecture test matrix. Include CDNA3 in this `DbiSim` class comment and add a CDNA3 example to the constructor comment at line 86. The SGPR section heading at line 316 also still says “CDNA4 and RDNA4,” although `DbiCdna3SgprSpillSimFixture` is defined below. These stale lists make the newly added CDNA3 coverage easy to miss.

Location: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/kernel_descriptor_scan.h:7`

Please use `#pragma once` for this new header. That is the documented header convention in `emulation/rocjitsu/docs/style.md`, and the other new headers in this change (`spill_builders.h` and `dbi_test_util.h`) already follow it.

The rocjitsu style guide explicitly asks us to avoid decorative comment lines (`docs/style.md:65`). Please replace the separator blocks introduced by this change in `tests/dbi_test_util.h`, `tests/dbi/dbi_spill_sim_test.cpp`, and the new spill sections in `tests/patch/instrumentor_test.cpp` with ordinary section comments or Doxygen headings. The section names already carry the useful information without the repeated `====` lines.

It would be good to update the special-state documentation now that the decoder exposes named VCC/EXEC destinations for `v_cmp` and `v_cmpx`. The old invisible/deferred rationale still appears in `probe_clobber.h:31-35`, `instrumentor.cpp:755-759`, `trampoline_builder.h:165-172`, and the related compare-coverage comments. Keeping unconditional preservation is fine, but these comments should describe it as conservative policy rather than a decoder limitation.

Location: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/probe_clobber.h:31`.

It would be useful to exercise the two new multi-register paths in the simulator. `Cdna4SpillsLiveClobberedVgprAndSgpr` and `Cdna4SpillsMultipleLiveClobberedSgprs` only search for emitted words, so they would still pass if the combined restores were misordered. The existing simulator fixtures cover one spilled register at a time; please add one VGPR-plus-SGPR round trip and one two-SGPR round trip through the shared bridge, asserting every restored value.

Location: `emulation/rocjitsu/tests/patch/instrumentor_test.cpp:2689` and `:2745`.

We should update `emulation/rocjitsu/docs/dbi-design.md` to match the feature set now in this PR. On current head `1ba4b466d9`, lines 7, 79, and 293 still say non-empty spill sets fail closed, although the implementation now saves and restores VGPRs, SGPRs, and AccVGPRs. The probe-call envelope at lines 269-285 also omits the full-mask store/load brackets, anchor EXEC restoration for the call, and special-state preservation, while the testing section omits `tests/dbi/dbi_spill_sim_test.cpp`. Please make these sections match the current behavior and leave only unsupported boundaries as future work.

This would be safer if `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/kernel_descriptor_scan.cpp:42-43` rejected a symbol name with no NUL terminator inside its linked string table. When `strnlen()` returns the full remaining size and those final bytes end in `.kd`, the following `strcmp()` continues past the table boundary looking for a terminator. Since this scanner accepts an arbitrary image span, please reject `len == strtab_size - sym.st_name` before the suffix check, or use a bounded suffix comparison, and add a truncated-string-table test.

accauble:
> We should restore the anchor EXEC mask before calling the probe, then widen it again only for spill reloads.

Unless I have misunderstood something, this version of the code does do this. This is something that the user will be able to configure later on (with an EXEC policy) since most probes likely won't need the value of the EXEC register and restoring would be a waste of time. We could also examine the probe to see if it is used before it is written, and decide that way.

theSK2005:
> > We should restore the anchor EXEC mask before calling the probe, then widen it again only for spill reloads.
> 
> Unless I have misunderstood something, this version of the code does do this. This is something that the user will be able to configure later on (with an EXEC policy) since most probes likely won't need the value of the EXEC register and restoring would be a waste of time. We could also examine the probe to see if it is used before it is written, and decide that way.

Yes, this is what the code currently does.

theSK2005:
@kuhar for this comment, "It would be useful to exercise the two new multi-register paths in the simulator. Cdna4SpillsLiveClobberedVgprAndSgpr and Cdna4SpillsMultipleLiveClobberedSgprs only search for emitted words, so they would still pass if the combined restores were misordered. The existing simulator fixtures cover one spilled register at a time; please add one VGPR-plus-SGPR round trip and one two-SGPR round trip through the shared bridge, asserting every restored value."
I have sim test for VGPR, SGPR, and AGPR round trip in the AccVGPR PR #9396 so I'll just add the two-SGPR round trip test.

kuhar (commented):
This scanner should use overflow-safe extent checks throughout. At `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/kernel_descriptor_scan.cpp:73`, `e_shoff + e_shnum * sizeof(Elf64_Shdr)` can wrap before `image.data() + e_shoff`; the section, string-table, and descriptor ranges at lines 91, 100, and 117-120 have the same unchecked-addition pattern. Because this API accepts an arbitrary byte span, a crafted ELF can drive out-of-bounds pointer formation or reads. Please validate every range as `offset <= size && length <= size - offset` before forming a pointer, and add hostile-header tests for wrapping offsets.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp
kuhar: The issue I am worried about is the same ordering at probe return. At the current `trampoline_builder.cpp:425-438`, `emit_probe_call()` goes directly from `s_swappc_b64` to the spill fills, while `verify_probe_callable()` accepts load-producing bodies. On RDNA4, the fills’ `s_wait_loadcnt 0` does not drain a probe `s_load_*` on KMCNT, so that load can complete after `v_readlane_b32` and overwrite the restored SGPR. We should emit `build_wait_all_loads_complete(arch)` after the call and before the first fill, with a static ordering test.

theSK2005: You're right, I'll add this

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp
```diff
@@ -128,6 +128,16 @@ struct SpillBracket {
     bracket.prologue.insert(bracket.prologue.end(), store.begin(), store.end());
   }
 
+  // Drain the probe's in-flight loads before any fill: a probe body may issue an
+  // unwaited load into a register we are about to restore, and on RDNA4 the fills'
+  // s_wait_loadcnt misses a scalar s_load (KMCNT), so that load could retire after
+  // the readlane and clobber the restored value. Emitted after the call, before the
+  // first fill; mirrors the prologue drain.
+  if (has_vgpr || has_sgpr) {
+    const auto drain = build_wait_all_loads_complete(arch);
```
kuhar: Please spell out the type here, per the coding style, since it's not obvious from the immediate context.

theSK2005: I noticed that Allyson's nop probe PR also used auto types for these build variables. Should I adjust those and all of the variables in test files too?

accauble: I'd say don't make extra changes that are not in scope of this PR.

kuhar (approved):
We should bound a descriptor symbol by its owning section before reading it. In emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/kernel_descriptor_scan.cpp:128-138, delta is checked only against the whole image, so a 64-byte .kd symbol can extend past sh_size into an adjacent section and still be returned and later mutated. Please require range_in_bounds(delta, sizeof(KD), shdr[sec_idx].sh_size) and add a crossing-section regression test.

LGTM % the above

accauble (changes requested):
Please have your agent update the dbi design doc with your register spilling decisions. --> `docs/dbi-design.md`

atgutier (commented):
## MAJOR

### RDNA4's "all loads complete" sequence omits sampling and BVH loads

Location: `lib/rocjitsu/src/rocjitsu/code/builders/spill_builders.h:177`

Reasoning: `build_wait_all_loads_complete()` emits only `s_wait_loadcnt_dscnt 0` and `s_wait_kmcnt 0` for RDNA4. RDNA4 has separate `SAMPLECNT` and `BVHCNT` counters. A pre-anchor sampling/BVH result can therefore arrive after its destination was spilled, or a probe result can arrive after the host value was restored, corrupting architectural state.

Evidence: RDNA4 defines and implements separate `s_wait_samplecnt` and `s_wait_bvhcnt` instructions, but neither is included in this supposedly complete drain.

Suggested fix: Include `s_wait_samplecnt 0` and `s_wait_bvhcnt 0`, then test the exact instruction order and encodings.

Confidence: High.

### CDNA bridge selection can allocate inside the AccVGPR window

Primary locations:

- `lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp:388`
- `lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp:823`

Reasoning: SGPR spills use an ordinary VGPR as a writelane/readlane bridge. The scan bounds this search using the descriptor's full unified VGPR allocation. On CDNA, that allocation can contain an ordinary-VGPR prefix followed by an AccVGPR window selected by `ACCUM_OFFSET`. The scan can therefore select a `vN` that aliases accumulator state.

Evidence: `instrumentor.cpp:823` decodes the full allocation and passes it as `kernel_vgpr_count`. The project's canonical descriptor model explicitly separates ordinary VGPR count, unified allocation, and AccVGPR base in:

- `lib/rocjitsu/src/rocjitsu/code/dbt/kernel_descriptor_translator.h:124`
- `lib/rocjitsu/src/rocjitsu/code/dbt/kernel_descriptor_translator.cpp:595`

For example, with AccVGPR base 16 and `v0:v15` live, the current scan selects `v16`.

Suggested fix: Decode `ACCUM_OFFSET` and cap ordinary bridge selection at the AccVGPR base. Reuse an already-spilled valid ordinary VGPR or fail closed if none exists.

Confidence: High.

### RDNA4 Wave64 descriptors are decoded using the Wave32 VGPR granule

Primary locations:

- `lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp:132`
- `lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp:823`

Reasoning: The new helper always uses an eight-VGPR descriptor granule for RDNA4. RDNA Wave64 descriptors use a four-VGPR granule. This overestimates the kernel's allocated ordinary VGPRs and can make the SGPR bridge scan choose an unallocated VGPR without growing the descriptor.

Evidence: The canonical implementation uses eight for Wave32 and four for Wave64 in `lib/rocjitsu/src/rocjitsu/code/dbt/kernel_descriptor_translator.cpp:338`, deriving wave size from the descriptor at `kernel_descriptor_translator.cpp:195`. With a Wave64 field value of zero, only `v0:v3` are declared; this code treats `v0:v7` as allocated and may select `v4`.

Suggested fix: Share the existing wave-size-aware descriptor decoder. Add paired RDNA4 Wave32 and Wave64 tests using a zero granulated-count field.

Confidence: High.

### Probe-call load drains occur too late and disappear entirely at no-spill sites

Primary locations:

- `lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp:104`
- `lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp:134`
- `lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp:393`

Reasoning: The pre-call drain is conditional on having spills, and special EXEC/VCC/M0 state is written into dead-register temporaries before that drain. Liveness does not model a previously issued asynchronous load still targeting an otherwise-dead register. Such a load can retire over the newly saved state. No-spill calls receive no boundary drain at all. Similarly, a probe can return with an outstanding result that escapes the call boundary and overwrites later host execution.

Evidence: Both drains are inside `if (has_vgpr || has_sgpr)`. Temp selection relies on anchor liveness and probe clobbers, neither of which protects against pre-anchor asynchronous writes. Probe validation permits loads and does not require a terminal wait. The underlying no-spill exposure partly predates this PR, but the PR introduces persistent special-state temporaries and a synchronization design that still writes them before draining.

Suggested fix: Emit an architecture-complete load drain at the very start of every probe-call envelope, before any special-state, target, or link temporary is written. Emit the matching drain immediately after probe return, before restoration or relocated host code. Remove the redundant spill-only drains and update envelope word accounting.

Confidence: High.

## MINOR

### Malformed descriptor values can trigger signed-overflow undefined behavior

Location: `lib/rocjitsu/src/rocjitsu/code/kernel_descriptor_scan.cpp:143`

Reasoning: Arbitrary unsigned `st_value` is converted to `int64_t` and then added to the signed entry offset without range checking. Extreme malformed ELF values can overflow the signed addition.

Suggested fix: Implement checked unsigned addition or subtraction according to the sign of `kernel_code_entry_byte_offset`, rejecting values that are not representable. Add positive- and negative-extreme malformed-image tests.

Confidence: High. This affects malformed input only.

### Descriptor symbols are bounded by the whole image, not their owning section

Primary locations:

- `lib/rocjitsu/src/rocjitsu/code/kernel_descriptor_scan.cpp:128`
- `lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp:1130`

Reasoning: The scanner verifies that the computed descriptor fits in the ELF image but does not require `delta + sizeof(KD) <= sh_size`. A crafted symbol can point through the end of its declared section into subsequent file data and still be accepted. DBI can later overwrite those unrelated bytes while updating private scratch size.

Suggested fix: Require a file-backed owning section and validate the complete 64-byte descriptor against both the section and image bounds.

Confidence: High. This affects malformed input only.

theSK2005:
> ## MAJOR
> ### RDNA4's "all loads complete" sequence omits sampling and BVH loads
> Location: `lib/rocjitsu/src/rocjitsu/code/builders/spill_builders.h:177`
> 
> Reasoning: `build_wait_all_loads_complete()` emits only `s_wait_loadcnt_dscnt 0` and `s_wait_kmcnt 0` for RDNA4. RDNA4 has separate `SAMPLECNT` and `BVHCNT` counters. A pre-anchor sampling/BVH result can therefore arrive after its destination was spilled, or a probe result can arrive after the host value was restored, corrupting architectural state.
> 
> Evidence: RDNA4 defines and implements separate `s_wait_samplecnt` and `s_wait_bvhcnt` instructions, but neither is included in this supposedly complete drain.
> 
> Suggested fix: Include `s_wait_samplecnt 0` and `s_wait_bvhcnt 0`, then test the exact instruction order and encodings.
> 
> Confidence: High.
> 
> ### CDNA bridge selection can allocate inside the AccVGPR window
> Primary locations:
> 
> * `lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp:388`
> * `lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp:823`
> 
> Reasoning: SGPR spills use an ordinary VGPR as a writelane/readlane bridge. The scan bounds this search using the descriptor's full unified VGPR allocation. On CDNA, that allocation can contain an ordinary-VGPR prefix followed by an AccVGPR window selected by `ACCUM_OFFSET`. The scan can therefore select a `vN` that aliases accumulator state.
> 
> Evidence: `instrumentor.cpp:823` decodes the full allocation and passes it as `kernel_vgpr_count`. The project's canonical descriptor model explicitly separates ordinary VGPR count, unified allocation, and AccVGPR base in:
> 
> * `lib/rocjitsu/src/rocjitsu/code/dbt/kernel_descriptor_translator.h:124`
> * `lib/rocjitsu/src/rocjitsu/code/dbt/kernel_descriptor_translator.cpp:595`
> 
> For example, with AccVGPR base 16 and `v0:v15` live, the current scan selects `v16`.
> 
> Suggested fix: Decode `ACCUM_OFFSET` and cap ordinary bridge selection at the AccVGPR base. Reuse an already-spilled valid ordinary VGPR or fail closed if none exists.
> 
> Confidence: High.
> 
> ### RDNA4 Wave64 descriptors are decoded using the Wave32 VGPR granule
> Primary locations:
> 
> * `lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp:132`
> * `lib/rocjitsu/src/rocjitsu/code/patch/instrumentor.cpp:823`
> 
> Reasoning: The new helper always uses an eight-VGPR descriptor granule for RDNA4. RDNA Wave64 descriptors use a four-VGPR granule. This overestimates the kernel's allocated ordinary VGPRs and can make the SGPR bridge scan choose an unallocated VGPR without growing the descriptor.
> 
> Evidence: The canonical implementation uses eight for Wave32 and four for Wave64 in `lib/rocjitsu/src/rocjitsu/code/dbt/kernel_descriptor_translator.cpp:338`, deriving wave size from the descriptor at `kernel_descriptor_translator.cpp:195`. With a Wave64 field value of zero, only `v0:v3` are declared; this code treats `v0:v7` as allocated and may select `v4`.
> 
> Suggested fix: Share the existing wave-size-aware descriptor decoder. Add paired RDNA4 Wave32 and Wave64 tests using a zero granulated-count field.
> 
> Confidence: High.
> 
> ### Probe-call load drains occur too late and disappear entirely at no-spill sites
> Primary locations:
> 
> * `lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp:104`
> * `lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp:134`
> * `lib/rocjitsu/src/rocjitsu/code/patch/trampoline_builder.cpp:393`
> 
> Reasoning: The pre-call drain is conditional on having spills, and special EXEC/VCC/M0 state is written into dead-register temporaries before that drain. Liveness does not model a previously issued asynchronous load still targeting an otherwise-dead register. Such a load can retire over the newly saved state. No-spill calls receive no boundary drain at all. Similarly, a probe can return with an outstanding result that escapes the call boundary and overwrites later host execution.
> 
> Evidence: Both drains are inside `if (has_vgpr || has_sgpr)`. Temp selection relies on anchor liveness and probe clobbers, neither of which protects against pre-anchor asynchronous writes. Probe validation permits loads and does not require a terminal wait. The underlying no-spill exposure partly predates this PR, but the PR introduces persistent special-state temporaries and a synchronization design that still writes them before draining.
> 
> Suggested fix: Emit an architecture-complete load drain at the very start of every probe-call envelope, before any special-state, target, or link temporary is written. Emit the matching drain immediately after probe return, before restoration or relocated host code. Remove the redundant spill-only drains and update envelope word accounting.
> 
> Confidence: High.
> 
> ## MINOR
> ### Malformed descriptor values can trigger signed-overflow undefined behavior
> Location: `lib/rocjitsu/src/rocjitsu/code/kernel_descriptor_scan.cpp:143`
> 
> Reasoning: Arbitrary unsigned `st_value` is converted to `int64_t` and then added to the signed entry offset without range checking. Extreme malformed ELF values can overflow the signed addition.
> 
> Suggested fix: Implement checked unsigned addition or subtraction according to the sign of `kernel_code_entry_byte_offset`, rejecting values that are not representable. Add positive- and negative-extreme malformed-image tests.
> 
> Confidence: High. This affects malformed input only.
> 
> ### Descriptor symbols are bounded by the whole image, not their owning section
> Primary locations:
> 
> * `lib/rocjitsu/src/rocjitsu/code/kernel_descriptor_scan.cpp:128`
> * `lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp:1130`
> 
> Reasoning: The scanner verifies that the computed descriptor fits in the ELF image but does not require `delta + sizeof(KD) <= sh_size`. A crafted symbol can point through the end of its declared section into subsequent file data and still be accepted. DBI can later overwrite those unrelated bytes while updating private scratch size.
> 
> Suggested fix: Require a file-backed owning section and validate the complete 64-byte descriptor against both the section and image bounds.
> 
> Confidence: High. This affects malformed input only.

"CDNA bridge selection can allocate inside the AccVGPR window" this issue is fixed in the followup PR #9396, currently working on fixing the rest.

accauble (approved):
Docs updated in PR #9396

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/code_object_patcher.cpp
theSK2005: I unified DBI and DBT's kernel descriptor read/write, still keeping this function as a narrower rewrite rather than a full rewrite

Groverkss (approved):

PR #8890, opened 2026-07-20
feat(amdsmi): map `amd-smi --version` to the `version` subcommand

marifamd:
## Motivation
`amd-smi --version` previously raised `AmdSmiInvalidSubcommandException` ("AMD-SMI Command '--version' is invalid"). Users expect `--version` to work like the near-universal CLI convention, so this maps it to the existing `version` subcommand.

## Technical Details
- In `amdsmi_cli.py`, a leading `--version` flag (case-insensitive, consistent with the existing `--rocm-smi` handling) is rewritten to the `version` subcommand before argument parsing. Routing through the real `version` subparser means it inherits full option support (`--gpu_version`, `--cpu_version`, `--nic_version`) with no bespoke handler.
- No other command is affected; `metric --version`, invalid subcommands, and `--help` behave exactly as before.

## Verification
Verified against an installed build on an MI-series host:

| Case | Result |
| ---- | ------ |
| `amd-smi --version` | full version output (was error before) |
| `amd-smi --version --gpu_version` / `--json` | modifiers work through the alias |
| `amd-smi --VERSION` | case-insensitive |
| `amd-smi version` (control) | unchanged |
| `amd-smi --bogus` | still InvalidSubcommandException |
| `amd-smi metric --version` | still InvalidParameterException (unchanged) |
| `amd-smi --help` | unchanged |

## Unit Tests
`tests/python/cli/test_version.py` adds `amd-smi --version` and `amd-smi --version --gpu_version` to the existing version CLI test.

## Documentation
- `docs/how-to/amdsmi-cli-tool.md`: note that `--version` is an alias for the `version` command.
- `CHANGELOG.md`: added entry under Added.

## JIRA ID
N/A


gabrpham (approved):

projects/amdsmi/tests/python/cli/test_version.py
```diff
@@ -17,6 +17,9 @@ def test_command(self):
             ("amd-smi version", self.PASS),
             ("amd-smi version --cpu_version", self.PASS),
             ("amd-smi version --gpu_version", self.PASS),
+            # `--version` is an alias for the `version` subcommand.
+            ("amd-smi --version", self.PASS),
+            ("amd-smi --version --gpu_version", self.PASS),
```
marifamd, line 22: --gpu_version might be a bug

PR #8790, opened 2026-07-17
refactor(rocprofiler-sdk): migrate thread trace off the per-queue callback registry

amd-vkale:
## Motivation

This PR isolates the thread-trace slice of the callback-removal effort (reference PRs #8730 / #8586) into a small, single-service change, per review guidance to land callback removal one service at a time.

### Underlying Problem

Dispatch thread trace (ATT) currently registers a `queue_callbacks_t` on every HSA queue. That hides “is this service active?” inside a per-queue map, makes enqueue and completion the same walk, and serializes every GPU while any dispatch ATT context is active.

This PR:

1. **Stops registering** dispatch ATT with that map. The write interceptor and async signal handler call explicit free functions instead (`thread_trace::write_hook`, `signal_completion_hook`, `is_any_active`).
2. **Scopes dispatch ATT to configured GPU agents** so disjoint agent sets can be active together; serialization is refcounted per agent and overlapping contexts conflict via `intersects()`.

Device thread trace is unchanged in its hook wiring; only dispatch ATT migrates here. The per-queue callback registry remains for counters, PC sampling, and SPM until those services land in their own PRs.

### Notes on Bigger Picture

Sibling migrations: counter collection (#8891), SPM (#8887), PC sampling (#8895).

Kernel replay (#7960, #8622) benefits from callback removal but this PR is independent and behavior-preserving on its own.

## Technical Details

**Both hooks iterate active contexts** (not registered). `post_kernel_call` self-filters by finding `THREAD_TRACE_CLIENT_ID`-tagged packets in `inst_pkt` via `dynamic_cast`, so routing differs from counter/SPM provenance via `packet_return_map`.

```mermaid
flowchart LR
  WriteInterceptor --> WriteHook["thread_trace::write_hook: active + collects_on"]
  AsyncSignalHandler --> CompletionHook["signal_completion_hook: active"]
  StartContext --> Serialize["enable_serialization(configured agents)"]
  StopContext --> Unserialize["disable_serialization(configured agents)"]
```

Permalinks pin to `d39785f`.

### Hooks and call sites

- Declarations: [queue_hooks.hpp](https://github.com/ROCm/rocm-systems/blob/d39785f/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/thread_trace/queue_hooks.hpp)
[...]

PR #8586, opened 2026-07-14
Remove Callbacks from rocProfiler-SDK

amd-vkale:
## Motivation

Remove unneeded callbacks to ensure unnecessary profiling is not done for tools. 

This is useful for PRs #7960 and #7966. 

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #7960, opened 2026-06-29
feat(rocprofiler-sdk): kernel replay service (callback tracing)

amd-vkale:
## Motivation

This PR adds an experimental **kernel replay service** to ROCprofiler-SDK: a **public callback-tracing API and its implementation**. It is **self-contained** — a custom ROCprofiler-SDK tool can configure `ROCPROFILER_CALLBACK_TRACING_KERNEL_REPLAY` and collect multiple hardware-counter groups within a **single application run** without waiting for the `rocprofv3` CLI.

The `rocprofv3` flag, tool wiring, and end-to-end CLI tests live in **[#10439](https://github.com/ROCm/rocm-systems/pull/10439)** (stacked on this PR). They are **not** required for this PR to merge.

Hardware exposes a limited number of counter registers per block, so a counter set larger than that cannot be collected in a single execution of a kernel. The established workaround is to partition the counters into groups and re-run the entire application once per group, which multiplies process launch, runtime initialization and all host-side work by the number of groups, and which yields counters drawn from that many distinct executions — so any derived metric combining counters across groups is computed from samples that were never simultaneously true. Kernel replay instead re-executes the individual dispatch once per group within a single run, restoring the device memory it reads between passes so every pass sees identical inputs.

The goals of this design are to:

1. expose replay as its own `KERNEL_REPLAY` callback-tracing domain so it composes with counter collection, kernel tracing and thread trace instead of displacing them;
2. ask the tool for a pass count per dispatch so that the pass count can be the counter-group count and dispatches that need nothing opt out at no cost;
3. support early exit and open-ended replay; and
4. give tools a loop-scoped, agent-scoped way to enable or disable a service for a single pass, which is what makes collecting a different counter group on each pass expressible at all.

### Stack

| PR | Base | Scope | Can land alone? |
|---|---|---|---|
| **#7960 (this PR)** | `develop` | ROCprofiler-SDK public API + implementation, conceptual/API docs, unit tests | **Yes** — custom tools work; `rocprofv3` is unchanged |
| [#10439](https://github.com/ROCm/rocm-systems/pull/10439) | this PR | `rocprofv3 --kernel-replay-beta-enabled`, `rocprofiler-sdk-tool` wiring, integration tests | After this PR |
| [#10780](https://github.com/ROCm/rocm-systems/pull/10780) | this PR | Public ABI gate, perf-helper unit tests, benchmark configuration, conceptual docs | Merged into this PR |
[...]

PR #8891, opened 2026-07-20
refactor(rocprofiler-sdk): migrate counter collection off the per-queue callback registry

amd-vkale:
## Motivation

The purpose of this PR is to isolate the counter-collection slice of the callback-removal effort (reference PRs #8730 / #8586) into a small, single-service change, per review guidance to land callback removal one service at a time. 

### Underlying Problem 

Dispatch counter collection currently registers a `queue_callbacks_t` on every HSA queue. That hides “is this service active?” inside a per-queue map, makes enqueue and completion the same walk, and serializes every GPU while any counter context is active.

This PR:

1. **Stops registering** counter collection with that map. The write interceptor and async signal handler call explicit free functions instead.
2. **Scopes collection to GPU agents** via `rocprofiler_dispatch_counting_service_set_agents`, so disjoint agent sets can be active together.

Public dispatch callback/record APIs are unchanged except for `set_agents`. Device counting is untouched. The per-queue callback map remains for unmigrated services.

### Notes on Bigger Picture 

Note the associated focused effort of the PR #8790 for migrating thread trace off the per-queue callback registry. 

The kernel replay feature PR #7960 is greatly enhanced by this PR. However, note this is independent of the kernel replay feature and can be used for other features in ROCprofiler-SDK.

## Technical Details

**Enter vs exit use different context sets.** Enter iterates **active** contexts; exit iterates **registered** contexts so in-flight dispatches still complete after stop.

```mermaid
flowchart LR
  WriteInterceptor --> EnterHook["kernel_dispatch_phase_enter_hook: active"]
  AsyncSignalHandler --> ExitHook["kernel_dispatch_phase_exit_hook: registered"]
  ContextStop --> Drain["enabled=false, drain, unserialize, callback_thread_stop, clear slot"]
```

Architecture: [queue_callback_removal_architecture.md](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/docs/conceptual/queue_hooks/queue_callback_removal_architecture.md). Permalinks pin to `70bef64`.

### Hooks and call sites

- Declarations: [queue_hooks.hpp](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/queue_hooks.hpp)
- Enter (`collects_on`, `COUNTERS_CLIENT_ID`): [queue_hooks.cpp#L45-L84](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/queue_hooks.cpp#L45-L84)
- Exit (provenance via `packet_return_map`): [queue_hooks.cpp#L86-L107](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/queue_hooks.cpp#L86-L107)
- `is_any_active`: [queue_hooks.cpp#L109-L113](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/queue_hooks.cpp#L109-L113)
- `no_real_consumers` gate: [queue.cpp#L318-L328](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue.cpp#L318-L328)
- Enter call site: [queue.cpp#L633-L644](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue.cpp#L633-L644)
- Exit call site: [queue.cpp#L172-L179](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue.cpp#L172-L179)
- Batching disabled while active: [queue.cpp#L768-L781](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue.cpp#L768-L781)
- `start_context` (no `add_callback`): [core.cpp#L148-L172](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/core.cpp#L148-L172)
- Stop order (`enabled` → drain → unserialize → callback thread): [core.cpp#L174-L209](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/core.cpp#L174-L209)
- Stop services **before** active-slot CAS: [context.cpp#L377-L405](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/context/context.cpp#L377-L405)
- Producer tags: [client_ids.hpp#L33-L56](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_hooks/client_ids.hpp#L33-L56)

### Per-agent scoping

- API: [dispatch_counting_service.h#L148-L185](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/include/rocprofiler-sdk/dispatch_counting_service.h#L148-L185)
- `collects_on` / `intersects`: [context.hpp#L82-L110](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/context/context.hpp#L82-L110)
- Conflict uses `intersects`: [context.cpp#L305-L314](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/context/context.cpp#L305-L314)
- `set_agents` locked while active: [core.cpp#L211-L248](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/core.cpp#L211-L248)
- Serialization refcount: [queue_controller.hpp#L97-L175](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_controller.hpp#L97-L175)

## Reviewer Guide

1. Confirm enter = `get_active_contexts`, exit = `get_registered_contexts` ([queue_hooks.cpp](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/queue_hooks.cpp)).
2. Counters-only runs still enter `WriteInterceptor` ([queue.cpp#L318-L328](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue.cpp#L318-L328)).
3. `start_context` does **not** call `add_callback` ([core.cpp#L148-L172](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/core.cpp#L148-L172)).
4. `context::stop_context` stops counters **before** the active-slot CAS ([context.cpp#L377-L405](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/context/context.cpp#L377-L405)).
5. `counters::stop_context` drains before teardown ([core.cpp#L174-L209](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/core.cpp#L174-L209)).
6. `collects_on` filters **before** `queue_cb` ([queue_hooks.cpp#L62-L66](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/queue_hooks.cpp#L62-L66)).
7. In-flight regression: [queue_hooks_test.cpp#L194-L326](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/tests/queue_hooks_test.cpp#L194-L326) — map drops by one per completion **and** every dispatch reaches the record callback after stop.
8. Per-agent: [per_agent_scoping_test.cpp](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/tests/per_agent_scoping_test.cpp), integration [per_agent_scoping_app.cpp](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/tests/counter-collection-per-agent/per_agent_scoping_app.cpp).

**Answer:** In-flight record after stop? GPU-1-only context leave GPU-0 alone? `packet_return_map` empty with every record delivered?

**Out of scope:** deleting `Queue::_callbacks`; migrating other services; device-counting.

## Issue Tracking
JIRA ticket: 
AIPROFSDK-1017

## Test Plan

- [is_any_active](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/tests/queue_hooks_test.cpp#L172-L175); [stop-while-in-flight](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/tests/queue_hooks_test.cpp#L194-L326)
- Per-agent unit + multi-GPU integration (links above)
- Start/stop tests assert `enabled`, not `iterate_callbacks` ([core.cpp](https://github.com/ROCm/rocm-systems/blob/70bef64/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/tests/core.cpp))

## Test Result

Intended coverage listed above. Treat the Checks tab as source of truth for CI.

## Submission Checklist


projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/queue_hooks.cpp
```diff
@@ -0,0 +1,99 @@
+// MIT License
+//
+// Copyright (c) 2025 Advanced Micro Devices, Inc. All rights reserved.
```
vlaindic:
```suggestion
// Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
```

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/queue_hooks.cpp
```diff
@@ -0,0 +1,99 @@
...
+active_counter_contexts_filter()
+{
+    return [](const context::context* ctx) -> bool {
+        return ctx && ctx->dispatch_counter_collection != nullptr;
+    };
+}
+}  // namespace
+
+void
+write_hook(const hsa::Queue&                                        queue,
```
vlaindic: Could we think of a better name? `kernel_dispatch_hook` or something similar. Or `kernel_dispatch_phase_enter_hook` or something. Write might be a bit misleading as it's not clear that this is about write interceptor hook.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_hooks/client_ids.hpp
```diff
@@ -0,0 +1,43 @@
...
+namespace queue_hooks
+{
+// Tags identifying the producer subsystem of each inst_pkt_t entry.
+// Only distinctness matters; values are fixed for test-order stability.
+// Services are migrated off the per-queue callback registry one at a time; the
+// unused ids are reserved so the tag values stay stable across those PRs.
+constexpr int64_t COUNTERS_CLIENT_ID     = 1;
+constexpr int64_t THREAD_TRACE_CLIENT_ID = 2;
+constexpr int64_t PC_SAMPLING_CLIENT_ID  = 3;
+constexpr int64_t SPM_CLIENT_ID          = 4;
```
vlaindic: Any reason for not using enum? Becuase we need `int64_t`?

vlaindic (commented):
Overall, I'm fine with the PR, added couple optional comments. 

Also, please check the AI generated comments. 

# PR #8891 Review — `refactor(rocprofiler-sdk): migrate counter collection off the per-queue callback registry`

> **Status:** "Not ready to Review" label is set — this review is early feedback.

---

## Findings

---

### ⚠️ IMPORTANT [hygiene]: Test Plan and Test Result sections are empty

**Issues Detected**
- Both `## Test Plan` and `## Test Result` contain only the template placeholder comments — no actual content.

**Suggested Actions**
- Fill in the Test Plan: at minimum, state which tests were run (counter-tests, rocprofv3 counter integration tests) and on which hardware.
- Fill in the Test Result: paste pass/fail summary. Even "counter-tests: all pass on MI300X" is sufficient.

**Risk/Impact Assessment**
- Medium — reviewers cannot verify that the change was exercised before merging.

---

### ⚠️ IMPORTANT [hygiene]: JIRA ID is indirect and exempt condition requires validation

**Issues Detected**
- The PR says "This PR does not have a JIRA ticket but is motivated by a ticket like that of kernel replay: AIPROFSDK-68." The JIRA ID is a reference to a *related* ticket, not the ticket for this work.
- The description does provide a clear, standalone justification (problem, change, validation path via cross-linked PRs), so the exemption *partially* applies — but the "motivated by a ticket like" phrasing is weaker than needed.

**Suggested Actions**
- Either create a JIRA ticket for this refactor and link it directly, or explicitly state "No JIRA ticket; standalone PR — see Technical Details for full justification." Don't use "like AIPROFSDK-68" phrasing, which implies an indirect relationship.

**Risk/Impact Assessment**
- Low — traceability concern only.

---

### ⚠️ IMPORTANT [clarity]: Copyright year is 2025 on all newly added files — should be 2026

**Issues Detected**
All five newly added files carry `Copyright (c) 2025`:
- `counters/queue_hooks.cpp`
- `counters/queue_hooks.hpp`
- `counters/tests/queue_hooks_test.cpp`
- `hsa/queue_hooks/client_ids.hpp`
- (CMakeLists.txt has no copyright header — that's fine)

**Suggested Actions**
- Update copyright header in `counters/queue_hooks.cpp` from 2025 to 2026 — file is newly added in this PR.
- Update copyright header in `counters/queue_hooks.hpp` from 2025 to 2026 — file is newly added in this PR.
- Update copyright header in `counters/tests/queue_hooks_test.cpp` from 2025 to 2026 — file is newly added in this PR.
- Update copyright header in `hsa/queue_hooks/client_ids.hpp` from 2025 to 2026 — file is newly added in this PR.

**Risk/Impact Assessment**
- Low — policy/compliance issue only.

---

### ⚠️ IMPORTANT [generic+clarity]: Attach-mode in-flight dispatch race is deleted without explanation

**Issues Detected**
The old `stop_context` contained an explicit guard: when `registration::is_attached()`, it removed the per-queue callback so new dispatches during the detach drain could pass through without incrementing `_active_kernels`. That block is deleted in this PR with only the comment:
> "No per-queue callback to remove; counters::write_hook no-ops once dispatch_counter_collection is disabled above."

The comment asserts the new path is safe, but the assertion is not obviously correct:

- `stop_context` sets `enabled = false` under a write lock.
- `write_hook` calls `queue_cb`, which reads `enabled` under a read lock — so the no-op is real for new dispatches.
- But `get_active_contexts()` is a separate snapshot: a context can still appear in `get_active_contexts()` for some window after `enabled` is set to false (the context is only removed from the active set when `rocprofiler_stop_context` fully deactivates it).
- During that window, `is_any_active()` returns `true`, `should_batch_packets` stays `false`, and `write_hook` is called — but `queue_cb` correctly short-circuits. This is safe.

**However:** the PR description does not explain this reasoning, and the inline comment does not mention the window. A reviewer reading this will correctly ask "what about in-flight dispatches during attach-mode detach?" and have no answer in the code.

**Suggested Actions**
- Add a comment at the deletion site in `stop_context` explaining the two-phase safety: `enabled` is set to false first (making `queue_cb` no-op for new dispatches), and the context will be removed from `get_active_contexts()` when the stop completes, at which point `is_any_active()` returns false and `should_batch_packets` can return to true.
- Add a sentence to the PR description's "Technical Details" section covering the attach-mode case and why the old `is_attached()` block is no longer needed.

**Risk/Impact Assessment**
- Medium — correctness argument is sound based on the code, but the missing explanation creates review friction and risks a future author misunderstanding the invariant.

---

### ⚠️ IMPORTANT [generic]: `write_hook` and `signal_completion_hook` in `queue_hooks.cpp` lack function docstrings

**Issues Detected**
The header `queue_hooks.hpp` has terse one-line comments above `write_hook` and `signal_completion_hook`, but the `.cpp` implementations have no docstrings at all. The functions carry non-trivial contracts:
- `write_hook` mutates `inst_pkt` (appends) and OR-folds `is_serialized` — callers must understand these are out-params, not inputs.
- `signal_completion_hook` iterates all *active* contexts (not the one that produced the packet) — the reason for this is subtle.

**Suggested Actions**
- Add a brief comment to `write_hook` in `queue_hooks.hpp` noting that `inst_pkt` and `is_serialized` are output params (appended/OR-folded respectively), and that the function is called once per kernel dispatch from `WriteInterceptor`, not once per active context.
- Add a comment to `signal_completion_hook` explaining that it re-queries active contexts rather than taking a captured context pointer because the completion fires asynchronously and contexts may have started/stopped since the dispatch.

**Risk/Impact Assessment**
- Low — maintainability concern, not correctness.

---

### 💡 SUGGESTION [generic]: `queue_hooks_test.cpp` — single negative-path test is thin for a new public hook

**Issues Detected**
`queue_hooks_test.cpp` tests only `is_any_active()` returning false with no active context. The two primary hooks (`write_hook`, `signal_completion_hook`) have no unit-level tests beyond what the existing `core.cpp` integration tests indirectly exercise.

**Suggested Actions**
- This is acknowledged as a host-only, GPU-free unit test file, which limits what can be tested without standing up HSA. At minimum, extend `queue_hooks_test.cpp` with a test that activates a mock context (the existing `core.cpp` test infrastructure already does this) and asserts `is_any_active()` returns true. This validates the full active/inactive cycle without GPU hardware.
- Consider extending `counters/tests/core.cpp:start_stop_buffered_ctx` to assert that `counters::is_any_active()` is true after start and false after stop — symmetrical to the old `iterate_callbacks` check that was removed.

**Risk/Impact Assessment**
- Low — the existing integration tests cover real behavior; this is about defense-in-depth at the unit level.

---

### 💡 SUGGESTION [hygiene]: `client_ids.hpp` "unused ids reserved" comment could name the follow-up PRs

**Issues Detected**
`client_ids.hpp` says:
> "Services are migrated off the per-queue callback registry one at a time; the unused ids are reserved so the tag values stay stable across those PRs."

The constants `THREAD_TRACE_CLIENT_ID`, `PC_SAMPLING_CLIENT_ID`, `SPM_CLIENT_ID` are not used by any code in this PR.

**Suggested Actions**
- Add references to the follow-up PRs (PR #8790 for thread trace is already mentioned in the description) as comments next to each unused constant, or at least note the planned migration order. This prevents a future developer from wondering if these are dead code.

**Risk/Impact Assessment**
- Low — cosmetic/maintainability only.

---

## Summary

| Severity | Count | Tags |
|----------|-------|------|
| ⚠️ IMPORTANT | 4 | hygiene (2), clarity (1), generic+clarity (1) |
| 💡 SUGGESTION | 2 | generic (1), hygiene (1) |

**Overall Risk: Medium** — The refactor logic is sound, the primary behaviors (enabled guard, no_real_consumers gate, should_batch disable) are correctly wired. The main concern is the missing test plan, the unexplained attach-mode safety argument, and stale copyright years.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/queue_hooks.cpp
amd-vkale: I'd agree the name isn't great - I actually think kernel_dispatch_phase_enter_hook would be reasonably appropriate here. I will look into changing it to that.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_hooks/client_ids.hpp
amd-vkale: I don't think we'd ever need an int64_t for a client id - I'd say it's a reasonably safe change to use `enum`.

bwelton (changes requested):
**Correctness / lifetime: in-flight counter completions are dropped when a context is stopped while dispatches are still executing**

Thanks for splitting the counter-collection migration out — the enqueue path and the `get_notifiers`/`should_batch`/`queue_id` cleanup look clean. One blocking issue:

Completion routing changed from *provenance-based* to *liveness-based*. Previously the completion callback lived in the queue-controller's per-queue map and fired for an in-flight dispatch regardless of whether the context was still active, and the old `stop_context` relied on this ("keep callbacks registered to avoid dropping counter data for in-flight dispatches"). The new `signal_completion_hook` (`counters/queue_hooks.cpp:83-89`) instead re-queries `get_active_contexts(...)` at completion time, but `context::stop_context` nulls the active slot (`context/context.cpp:371`) *before* calling `counters::stop_context` (`context/context.cpp:380`).

So a kernel dispatched while the context was active but completing after `rocprofiler_stop_context()` returns finds zero active counter contexts, and `completed_cb` never runs. Result:
- counter records for those in-flight kernels are silently lost, and
- the `packet_return_map` entry (erased only in `completed_cb`, `dispatch_handlers.cpp:162`) is leaked, with possible misassociation if the packet address is later reused.

`counters::stop_context` (`core.cpp:172-191`) doesn't drain in-flight GPU work, so nothing compensates. Stop-while-in-flight is an ordinary lifecycle/finalize pattern, so this isn't an edge case.

One possible fix: route completion by packet provenance instead of current activeness. `completed_cb` already resolves ownership via pointer lookup in `packet_return_map` and self-filters, so iterating registered (rather than only active) counter contexts in `signal_completion_hook` would keep completions reachable after stop — provided the callback thread and `counter_callback_info` objects are kept alive until in-flight dispatches drain (`stop_context` currently calls `callback_thread_stop()` unconditionally at `core.cpp:190`).

Test coverage note: the new `queue_hooks_test.cpp` only checks `is_any_active() == false` with no contexts, and the trimmed `tests/core.cpp` start/stop tests just inspect the `enabled` flag — nothing would catch this. Please add a test that starts a counter context, enqueues kernels, stops the context while dispatches are in flight, and asserts every kernel produces a record and `packet_return_map` ends up empty.

amd-vkale:
> **Correctness / lifetime: in-flight counter completions are dropped when a context is stopped while dispatches are still executing**
> 
> Thanks for splitting the counter-collection migration out — the enqueue path and the `get_notifiers`/`should_batch`/`queue_id` cleanup look clean. One blocking issue:
> 
> Completion routing changed from _provenance-based_ to _liveness-based_. Previously the completion callback lived in the queue-controller's per-queue map and fired for an in-flight dispatch regardless of whether the context was still active, and the old `stop_context` relied on this ("keep callbacks registered to avoid dropping counter data for in-flight dispatches"). The new `signal_completion_hook` (`counters/queue_hooks.cpp:83-89`) instead re-queries `get_active_contexts(...)` at completion time, but `context::stop_context` nulls the active slot (`context/context.cpp:371`) _before_ calling `counters::stop_context` (`context/context.cpp:380`).
> 
> So a kernel dispatched while the context was active but completing after `rocprofiler_stop_context()` returns finds zero active counter contexts, and `completed_cb` never runs. Result:
> 
> * counter records for those in-flight kernels are silently lost, and
> * the `packet_return_map` entry (erased only in `completed_cb`, `dispatch_handlers.cpp:162`) is leaked, with possible misassociation if the packet address is later reused.
> 
> `counters::stop_context` (`core.cpp:172-191`) doesn't drain in-flight GPU work, so nothing compensates. Stop-while-in-flight is an ordinary lifecycle/finalize pattern, so this isn't an edge case.
> 
> One possible fix: route completion by packet provenance instead of current activeness. `completed_cb` already resolves ownership via pointer lookup in `packet_return_map` and self-filters, so iterating registered (rather than only active) counter contexts in `signal_completion_hook` would keep completions reachable after stop — provided the callback thread and `counter_callback_info` objects are kept alive until in-flight dispatches drain (`stop_context` currently calls `callback_thread_stop()` unconditionally at `core.cpp:190`).
> 
> Test coverage note: the new `queue_hooks_test.cpp` only checks `is_any_active() == false` with no contexts, and the trimmed `tests/core.cpp` start/stop tests just inspect the `enabled` flag — nothing would catch this. Please add a test that starts a counter context, enqueues kernels, stops the context while dispatches are in flight, and asserts every kernel produces a record and `packet_return_map` ends up empty.

@bwelton Thanks for this - I am prioritizing fixing this gap right now.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/hsa/queue_hooks/client_ids.hpp
amd-vkale: See above for reference for where this fix is in the stacked PR and the specifics of the fix.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/queue_hooks.cpp
amd-vkale: See above for reference for where this fix is in the stacked PR and the specifics of the fix.

The main point is that the new names I chose were  kernel_dispatch_phase_enter_hook  and kernel_dispatch_phase_exit_hook.

amd-vkale:
**Mergeability (local review + GPU unit tests, 2026-08-18)**

#8891 as it stands is still `CHANGES_REQUESTED` for the in-flight completion drop. That fix is implemented and tested on stacked #9586 (see the comment there). Plan: fast-forward #9586 into this branch so the blocking review can be re-checked on one PR.

#9868 (per-agent scoping) is **not** being folded. It does not compile on current head and is a feature, not this review-fix. See that PR.

amd-vkale:
**Ready for review as a whole.** Stacked review-fixes (#9586) and per-agent dispatch scoping (#9868, including the mutex compile fix) are now in this branch (`70bef64ca0`).

Per-agent scoping was directed by jrmadsen and is folded here rather than waiting on a second approval cycle. bwelton’s earlier `CHANGES_REQUESTED` on the in-flight completion path is addressed by the #9586 commits already in this tip. Please re-review the combined diff against `develop`.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/queue_hooks.cpp
```diff
@@ -0,0 +1,126 @@
...
+        }
+    }
+    if(!has_counter_packets) return;
+
+    // Route by packet provenance, not current activeness: completed_cb self-filters via
+    // packet_return_map, so in-flight dispatches still complete after stop_context removes the
+    // context from the active list. This is what guarantees a completion that arrives is delivered;
+    // the drain in stop_context is what bounds when completions arrive. Kernel replay needs the
+    // same property, since each pass completes separately.
+    auto contexts = context::get_registered_contexts(counter_contexts_filter());
```
bgopesh, line 110: get_registered_contexts() is now walked unsynchronized from the HSA async signal handler on every dispatch completion.

- Problem: kernel_dispatch_phase_exit_hook calls context::get_registered_contexts(counter_contexts_filter()), which iterates *get_registered_contexts_impl() — a stable_vector<std::optional<context>> — while holding no lock. register_context can emplace_back / construct into that array concurrently. Reading itr.has_value() while another thread is running optional::emplace() on the same slot is a data race, not a benign one.
- Evidence this is real, not theoretical: this PR adds if(!itr.has_value()) continue; guards at four sites in context.cpp (:135, :443, :461, :497). Those guards only matter if slots can be observed in a not-yet-engaged state — which is precisely the window that is unsafe to read concurrently. The guards make the common case survive; they do not make the read atomic. I want to flag that this is circumstantial: I'm inferring intent from the guards rather than from a crash report.
- Impact: The old code reached completion callbacks through Queue::_callbacks, a per-queue Synchronized map. This PR replaces that with an unsynchronized walk of process-global state, executed from AsyncSignalHandler for every dispatch completion. Torn read → dereferencing a half-constructed context → segfault or silent garbage in the completion path.
- Fix: take get_contexts_mutex() for the read, or — better, since this is an async handler where blocking is forbidden (Rule 4.2) — publish a snapshot: maintain a lock-free array of registered counter-collection contexts updated at register/deregister, and have the exit hook read that. If the array is genuinely append-only-and-never-reallocating and this is already safe, please say so in a comment at the call site; today nothing states the invariant.

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/core.cpp
```diff
@@ -220,28 +185,68 @@ stop_context(const context::context* ctx)
...
+        // and the counter_callback_info objects stay alive until in-flight dispatches drain, so the
+        // drain is what makes that condition hold literally rather than by argument.
+        //
+        // context::stop_context calls this function while the context is still in the active list,
+        // which is what keeps the service visible for the duration of the drain: the enter hook
+        // still reaches queue_cb, and queue_cb's disabled path returns serialize=true so the
+        // serialized->unserialized transition stays coordinated until disable_serialization() runs
+        // below. That ordering is why no separate "draining" flag is needed -- but it only works
+        // because the drain happens here, before the slot is cleared.
+        hsa::queue_controller_sync();
```
bgopesh, line 199: context::stop_context now holds the global contexts mutex across an unbounded GPU drain — context/context.cpp:396 → counters/core.cpp:199

- Problem: context::stop_context opens with auto _lk = std::unique_lock<std::mutex>{get_contexts_mutex()};. The reordering moves counters::stop_context() above the CAS — and counters::stop_context now begins with a newly added hsa::queue_controller_sync(), which iterates every queue and calls Queue::sync(), waiting on the GPU. The drain is new in this PR; the mutex hold is pre-existing. Their combination is not.
- Impact: Every rocprofiler_start_context / rocprofiler_stop_context in the process blocks behind one context's GPU drain. A hung or long-running kernel on any queue stalls all context lifecycle operations, including on unrelated agents — which directly undercuts the per-agent-independence this PR is built to deliver. set_dispatch_agents (core.cpp:228) takes the same mutex and inherits the stall.
- Fix: drain outside the lock. Mark the service disabled and drop the lock before queue_controller_sync(), then reacquire for the CAS; or move the drain into a phase that runs after _lk is released. If the drain genuinely must be inside the lock, that needs to be stated and justified in the comment — the current comment explains the ordering thoroughly but never addresses the lock scope.

bgopesh (changes requested):

projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/queue_hooks.hpp
```diff
@@ -0,0 +1,68 @@
...
+signal_completion_hook(const hsa::Queue&                           queue,
+                       const hsa::rocprofiler_packet&              kernel_packet,
+                       std::shared_ptr<hsa::queue_info_session_t>& session,
+                       hsa::packet_data_t&                         packet,
+                       hsa::inst_pkt_t&                            inst_pkt,
+                       kernel_dispatch::profiling_time             dispatch_time);
+
+// True if any context currently has dispatch counter collection active.
+bool
+is_any_active();
```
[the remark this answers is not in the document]
amd-vkale: @copilot The specific premise here does not hold: there is no "enabled flag" on the service to consult. [`dispatch_counter_collection_service`](https://github.com/ROCm/rocm-systems/blob/users/vkale/remove-callbacks-counters/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/context/context.hpp#L82) carries `callbacks` and the agent set, and nothing else — `git grep enabled` over `counters/*.hpp` on this branch returns nothing. And `get_active_contexts` already restricts to started contexts, so "active context with a non-null `dispatch_counter_collection`" is exactly what "has dispatch counter collection active" means. The comment and the implementation agree.

There is a real precision gap underneath it, though, and it is worth writing down: `is_any_active()` is a **process-wide** question, not an agent-scoped one. It returns true whenever any started context has the service configured, even if that context collects only on an agent other than the one being dispatched to. The agent narrowing happens later and deliberately, at [`queue_hooks.cpp` L66](https://github.com/ROCm/rocm-systems/blob/users/vkale/remove-callbacks-counters/projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/queue_hooks.cpp#L66) (`collects_on(agent_id)`), because `queue_cb` returns `serialize=true` even on its disabled path, so filtering has to precede the call rather than happen inside it.

That matters for exactly the callers you flagged: a GPU-1-only context makes `is_any_active()` true on a GPU-0 queue, so the write interceptor is forced on and batching disabled process-wide even though no counters will be collected there. That is intentional — the decision is made before the agent is known — but the one-line comment does not say so.

Proposed clarification, no behavior change:

```cpp
// True if any started context has dispatch counter collection configured. Process-wide and not
// agent-scoped: a context that collects only on another agent still counts here, because the
// interceptor/batching decision is made before the dispatch's agent is known. Agent narrowing
// happens on the dispatch path via collects_on().
```

Happy to land that if it reads right to reviewers — flagging rather than pushing it, since it is comment-only and this PR is mid-review.

amd-vkale:
@copilot resolve the merge conflicts in this pull request

amd-vkale:
Status for reviewers: the counter collection migration, together with all the review fixes that were previously split into the stacked PR #11869, is now consolidated.

**Where to review the code: #11970.** It sits on `users/vkale/remove-callbacks-counters--with-cursor-fixes`, a branch in `ROCm/rocm-systems` rather than in my fork, because CodeQL does not run on pull requests from a fork. #11970 targets `develop`.

This PR (#8891) is now stacked on #11970 and holds only a marker commit with no file changes, so its "Files changed" tab will look empty. It stays open to keep this review history attached to the work.

#11869 has been closed as superseded — every commit that was on it is an ancestor of the consolidated branch, so the lifecycle, race, scope and hook tag fixes are all present. Two further fixes were made on top of it: the completion hook test fixture now uses a null session so the test does not require a live HSA runtime, and a wording fix in `per_agent_scoping_app.cpp`.

PR #8882, opened 2026-07-20
[ci] Fix therock_configure_ci_test.py and run it continuously

ScottTodd:
## Motivation

The tests in `.github/scripts/tests/therock_configure_ci_test.py` are not running as part of any workflows and were recently broken. This fixes them and starts running them as part of `therock-ci.yml`

Follow-up to:
* https://github.com/ROCm/rocm-systems/pull/6954#discussion_r3617222497
* https://github.com/ROCm/rocm-systems/pull/7723#discussion_r3617390551

Getting these unit tests passing will make it easier to apply adjustments to the filters that run CI and to roll out review policy changes / enforcement as part of https://github.com/ROCm/rocm-systems/issues/7842.

## Technical Details

I considered adding a new dedicated unit_tests job but there is only a single test file right now, so it seemed simpler to extend the existing setup job which always runs and is already part of a required check. We could later add a more dedicated home for such quick unit tests that operate on just the repository sources themselves without needing any subproject source builds.

## Test Plan

* `pytest .github\scripts\tests\therock_configure_ci_test.py` newly passes
* new workflow step runs and passes:
    ```
    Run python -m pytest .github/scripts/tests/therock_configure_ci_test.py
      
    ============================= test session starts ==============================
    platform linux -- Python 3.12.13, pytest-9.1.1, pluggy-1.6.0
    rootdir: /home/runner/work/rocm-systems/rocm-systems
    collected 28 items
    
    .github/scripts/tests/therock_configure_ci_test.py ..................... [ 75%]
    .......                                                                  [100%]
    
    ============================== 28 passed in 0.61s ==============================
    ```

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/rocm-systems/blob/develop/CONTRIBUTING.md.

Issue #7842, opened 2026-06-25
PR Bot Enablement


samathew121:
### Motivation

Therock-pr-bot tool to add policy based checks on PRs.
right now we are adding label "not ready for review" for PRs that does not have issue/jira linked and PRs that has zero unit test to it. We have other checks at place around forbidden files, pre-commit failure, branch name convention. and more checks marked as soon to enabled.

### Technical Details

The checks are purely based on config file named policy.yml based PR BOT. this is not a llm/AI based bot. This is using code to deterministically provide checks.

### Test Plan

Currently tested in therock-infra repo and on this PR, further plan to extend for systems and libraries repo.

### Test Result

<img width="458" height="326" alt="Image" src="https://github.com/user-attachments/assets/a56b713b-2735-4b66-b466-6a6c967649ba" />

PR #8893, opened 2026-07-20
[ci] Extend/dedup path filters in therock_configure_ci.py

ScottTodd:
## Motivation

The `SKIPPABLE_PATH_PATTERNS` list that controls which paths can be modified without triggering the full "TheRock CI" workflows had a few overlapping patterns like `*.gitignore` and `projects/*/.gitignore` and was missing patterns for some newly added directories like `tools/systems_pr_bot/*` (see https://github.com/ROCm/rocm-systems/issues/7842).

## Technical Details

I (well, with some nudging by Codex) decided to also restructure the `test_is_path_skippable` test case to make new subtests easier to add.

## Test Plan

New + existing unit tests (now running/passing thanks to https://github.com/ROCm/rocm-systems/pull/8882)

## Submission Checklist

### DIFF
--- a/.github/scripts/tests/therock_configure_ci_test.py
+++ b/.github/scripts/tests/therock_configure_ci_test.py
@@ -187,14 +187,15 @@ class ConfigureCITest(unittest.TestCase):
         # None input
         self.assertFalse(therock_configure_ci.check_for_non_skippable_path(None))
 
-    @patch("subprocess.run")
-    def test_docs_only_change_returns_empty_list(self, mock_run):
+    @patch("therock_configure_ci.get_modified_paths")
+    def test_docs_only_change_returns_empty_list(self, mock_get_modified):
         args = {"is_pull_request": True, "base_ref": "HEAD^"}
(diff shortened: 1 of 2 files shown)


ScottTodd:
I'm trying out stacked PRs here FYI: https://github.github.com/gh-stack/. Not sure how that will handle it if I merge the base PR first. I might be able to merge the stack together using squash and merge 🤔

geomin12 (approved):
oh thanks! this will help, but sooooooon to multi-arch so we don't need to deal with these files

geomin12:
> I'm trying out stacked PRs here FYI: https://github.github.com/gh-stack/. Not sure how that will handle it if I merge the base PR first. I might be able to merge the stack together using squash and merge 🤔

woah this is cool

ScottTodd:
> but sooooooon to multi-arch so we don't need to deal with these files

is there a different file that will be used to determine when multi-arch CI should run for changes in rocm-systems?

PR #8896, opened 2026-07-20
build(rocr): Fix debug builds with clang

lancesix:
## Motivation

When creating debuild build with clang, we have build failures introduced by 99a0be1e16 "rocr/coredump: Extend list of required memory ranges for lightweight cores":

    .../libamdhsacode/lnx/amd_core_dump.cpp:315:46: warning: flag '#' results in undefined
          behavior with 'p' conversion specifier [-Wformat]
      315 |       debug_print("Added aql_queue_t range: %#p - %#p (size: %zu)\n",
          |                                             ~^~
    .../core/util/utils.h:161:21: note: expanded from macro 'debug_print'
      161 |     fprintf(stderr, fmt, ##__VA_ARGS__);                                                           \
          |                     ^~~
    .../libamdhsacode/lnx/amd_core_dump.cpp:315:52: warning: flag '#' results in undefined
          behavior with 'p' conversion specifier [-Wformat]
      315 |       debug_print("Added aql_queue_t range: %#p - %#p (size: %zu)\n",
          |                                                   ~^~
    .../core/util/utils.h:161:21: note: expanded from macro 'debug_print'
      161 |     fprintf(stderr, fmt, ##__VA_ARGS__);                                                           \
          |                     ^~~
    .../libamdhsacode/lnx/amd_core_dump.cpp:325:21: error: arithmetic on a pointer to void
      324 |                   static_cast<void*>(aql_queue->amd_queue_.hsa_queue.base_address)
          |                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      325 |                     + aql_queue->amd_queue_.hsa_queue.size * sizeof(hsa_kernel_dispatch_packet_t),
          |                     ^
    .../core/util/utils.h:161:28: note: expanded from macro 'debug_print'
      161 |     fprintf(stderr, fmt, ##__VA_ARGS__);                                                           \
          |                            ^~~~~~~~~~~
    .../libamdhsacode/lnx/amd_core_dump.cpp:339:65: error: arithmetic on a pointer to void
      339 |                   static_cast<void*>(queue_info.SaveAreaHeader) + queue_info.SaveAreaSizeInBytes,
          |                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^
    .../core/util/utils.h:161:28: note: expanded from macro 'debug_print'
[...]

PR #8898, opened 2026-07-20
rocr/coredump: fix void* pointer arithmetic in debug_print

lamb-j:
## Problem

`#7502` (`rocr: Fix coredump generation to pipes, and to file on lustrefs`, merged to `develop`) introduced arithmetic directly on a `void*` in `amd_core_dump.cpp`. This is a GNU extension that GCC accepts but **Clang rejects as a hard error**:

```
amd_core_dump.cpp:325:21: error: arithmetic on a pointer to void
amd_core_dump.cpp:339:65: error: arithmetic on a pointer to void
```

This breaks any Clang build of rocr-runtime. It surfaced in the SPIRV-LLVM-Translator CI, which builds rocr-runtime with the AMD Clang it just built (`ROCm/SPIRV-LLVM-Translator` Linux CI, which checks out `rocm-systems` `develop`).

## Fix

Both sites compute an end pointer (`base + size`) passed to a `%#p` `debug_print` argument. Compute the end address via `reinterpret_cast<uint64_t>` — the same idiom already used by the adjacent `filter.add_range(...)` calls and the scratch-range print just above — and cast the result back to `void*` for the `%#p` specifier. No behavior change.

## Notes

- Only diagnostic `debug_print` output is affected; runtime behavior is unchanged.
- These were the only two `void*`-arithmetic sites in the file.

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/lnx/amd_core_dump.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/lnx/amd_core_dump.cpp
@@ -321,8 +321,9 @@ static hsa_status_t build_lightweight_coredump_ranges(MemoryRegionFilter& filter
       // Same goes for the ring buffer.
       debug_print("Added ring buffer range: %#p - %#p (size: %zu)\n",
                   aql_queue->amd_queue_.hsa_queue.base_address,
-                  static_cast<void*>(aql_queue->amd_queue_.hsa_queue.base_address)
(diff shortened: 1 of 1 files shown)


lamb-j:
Closed in favor of #8896 (by @lancesix), which fixes the same `void*` arithmetic errors and additionally clears the `%#p` `-Wformat` UB warnings by standardizing the range prints on `%#" PRIx64` + `uint64_t`. That's the more complete fix and is owned by the author of the original change. Note: this regression is currently breaking Clang builds of rocr-runtime (e.g. the ROCm/SPIRV-LLVM-Translator Linux CI), so landing #8896 would unblock it.

PR #8903, opened 2026-07-21
fix(rocprof-trace-decoder): handle and table lifetime after main

ApoKalipse-V:
## Motivation

Rocprof-trace-decoder does not handle well functions called both before/after main() due to lifetime issues of static objects.
This PR addresses it.

JIRA ID : AIPROFSDK-1001

## Submission Checklist

### DIFF
--- a/projects/rocprof-trace-decoder/source/gfx10/gfx10wave.cpp
+++ b/projects/rocprof-trace-decoder/source/gfx10/gfx10wave.cpp
@@ -31,6 +31,9 @@
 #include "mi400/mi400wave.h"
 #include "segment.hpp"
 
+#include <cstddef>
+#include <new>
+
 #define INST_JUMP_TYPE 5
 #define INST_TRAP_TYPE 6
 #define INST_RFE_TYPE  117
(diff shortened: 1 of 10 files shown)


projects/rocprof-trace-decoder/source/gfx10/gfx10wave.cpp
```diff
@@ -129,7 +132,11 @@ enum EINST
     einst_final
 };
 
-static std::unordered_map<int, mapped_inst_t> table_map_to_common_type{
+// clang-format off
+using InstructionTable = std::unordered_map<int, mapped_inst_t>;
```
meserve-amd: Minor: naming here, should be `instruction_table_t` or similar. Generally we name types with _t or _type at the end, and use lowercase

projects/rocprof-trace-decoder/source/gfx11/gfx11wave.cpp
```diff
@@ -26,6 +26,9 @@
 #include <utility>
 #include <vector>
 
+#include <cstddef>
+#include <new>
```
meserve-amd, line 28: Nit: should be sorted in alphabetical order with other STL level includes. Here and in other files

projects/rocprof-trace-decoder/source/handle.hpp
```diff
@@ -132,8 +135,10 @@ class HandleData
 
     static std::unordered_map<uint64_t, std::shared_ptr<HandleData>>& get_map()
     {
-        static std::unordered_map<uint64_t, std::shared_ptr<HandleData>> map;
-        return map;
+        using Map = std::unordered_map<uint64_t, std::shared_ptr<HandleData>>;
```
meserve-amd: Minor: name this map type something unique

meserve-amd (approved):
Looks good and I like the test. Some minor comments on convention and naming.

PR #8905, opened 2026-07-21
fix(rocprofiler-systems): Add missing label_excludes for full test category

### LINKED ISSUE #8904: [rocprofiler-systems] thread-limit tests run in TheRock when they should not
In https://github.com/ROCm/TheRock/pull/6700, when running with `test_filter:full`, the `thread-limit` set of tests ran, which should be impossible. They should be excluded.

Investigation shows that the `full` label in `test_categories.yaml` (https://github.com/ROCm/rocm-systems/blob/develop/projects/rocprofiler-systems/tests/test_categories.yaml#L99) is missing the `label_excludes`. 

This needs to be fixed ASAP.

kcossett-amd:
## Motivation

Fixes #8904 

When running TheRock CI (on TheRock) under `test_filter:full`, `thread-limit` tests ran. This should not be the case under any circumstance as they will fail.

Investigation shows that the `full` label was missing `common_therock_label_excludes`.

## Technical Details

Add `common_therock_label_excludes` to the `full` test category.
Note that this category is meant for TheRock CI.

## Test Plan

N/A (local `TheRock CI` runs `standard` test type, which has the label. IIRC, only weeklies run `full`).
This must be fixed before then just incase.

## Test Result

N/A (see section above)

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-systems/tests/test_categories.yaml
+++ b/projects/rocprofiler-systems/tests/test_categories.yaml
@@ -101,6 +101,7 @@ test_categories:
     regex_includes:
       - ".*"
     regex_excludes: *_common_therock_regex_excludes
+    label_excludes: *_common_therock_labels_exclude
     added_supplementary_labels:
       - "full"
       - "all"


sputhala-amd (approved):

dgaliffiAMD (approved):

kcossett-amd:
This PR will add back the lost functionality of `--repeat until-pass:3` that we originally had in TheRock: https://github.com/ROCm/TheRock/pull/6733

PR #8911, opened 2026-07-21
feat(hip): enable stream capture for hipStreamWaitValue/WriteValue/Ba…

anugodavar:
…tchMemOp

- Add STREAM_CAPTURE hooks to hipStreamWaitValue32/64, hipStreamWriteValue32/64, and hipStreamBatchMemOp so they record BatchMemOp graph nodes during capture instead of executing immediately.
- Declare and implement capture handlers in hip_graph_capture.hpp/hip_graph.cpp.
- Fix hipGraphBatchMemOpNode to deep-copy paramArray to prevent dangling pointer on graph replay.
- Fix batchMemOps() globalWorkSize from count to 1 so ops execute sequentially in array order per CUDA spec.
- Inline sequential for-loop in __amd_rocclr_batchMemOp in blitcl.cpp to be self-contained until the fix lands in upstream llvm-project amdblit.cl.

## Issue Tracking

JIRA ID: AIRUNTIME-2528

## Submission Checklist

### DIFF
--- a/projects/clr/hipamd/src/hip_graph.cpp
+++ b/projects/clr/hipamd/src/hip_graph.cpp
@@ -3788,4 +3788,110 @@ hipError_t hipGraphExecBatchMemOpNodeSetParams(hipGraphExec_t hGraphExec, hipGra
   }
   HIP_RETURN(reinterpret_cast<hip::hipGraphBatchMemOpNode*>(clonedNode)->SetParams(nodeParams));
 }
+
+hipError_t capturehipStreamBatchMemOp(hipStream_t& stream, unsigned int& count,
+                                      hipStreamBatchMemOpParams*& paramArray,
+                                      unsigned int& flags) {
+  ClPrint(amd::LOG_DETAIL_DEBUG, amd::LOG_API,
+          "[hipGraph] Current capture node StreamBatchMemOp on stream : %p", stream);
+  if (!hip::isValid(stream)) {
+    return hipErrorContextIsDestroyed;
+  }
+  if (paramArray == nullptr || count == 0 || count > 256 || flags != 0) {
+    return hipErrorInvalidValue;
+  }
+  hip::Stream* s = reinterpret_cast<hip::Stream*>(stream);
+
+  hipBatchMemOpNodeParams nodeParams{};
+  nodeParams.ctx = reinterpret_cast<hipCtx_t>(hip::getCurrentDevice());
+  nodeParams.count = count;
+  nodeParams.paramArray = paramArray;
+  nodeParams.flags = flags;
+
+  // hipGraphBatchMemOpNode deep-copies paramArray internally.
+  hip::GraphNode* node = new hip::hipGraphBatchMemOpNode(&nodeParams);
+
+  hipError_t status = ihipGraphAddNode(node, s->GetCaptureGraph(),
+                                       s->GetLastCapturedNodes().data(),
+                                       s->GetLastCapturedNodes().size());
+  if (status != hipSuccess) {
+    return status;
+  }
+  s->SetLastCapturedNode(node);
+  return hipSuccess;
+}
+
+// Stream-capture helpers for individual stream memory operations.
+// Each is captured as a single-element BatchMemOp node, matching CUDA behaviour
+// where cuStreamWaitValue32/64 and cuStreamWriteValue32/64 produce
+// CU_GRAPH_NODE_TYPE_BATCH_MEM_OP nodes when captured.
+static hipError_t captureStreamMemOp(hipStream_t stream, hipStreamBatchMemOpParams& op) {
+  if (!hip::isValid(stream)) {
+    return hipErrorContextIsDestroyed;
+  }
+  hip::Stream* s = reinterpret_cast<hip::Stream*>(stream);
+  hipBatchMemOpNodeParams nodeParams{};
+  nodeParams.ctx = reinterpret_cast<hipCtx_t>(hip::getCurrentDevice());
+  nodeParams.count = 1;
+  nodeParams.paramArray = &op;
+  nodeParams.flags = 0;
+
+  hip::GraphNode* node = new hip::hipGraphBatchMemOpNode(&nodeParams);
+  hipError_t status = ihipGraphAddNode(node, s->GetCaptureGraph(),
+                                       s->GetLastCapturedNodes().data(),
+                                       s->GetLastCapturedNodes().size());
+  if (status != hipSuccess) {
+    return status;
+  }
+  s->SetLastCapturedNode(node);
+  return hipSuccess;
+}
+
+hipError_t capturehipStreamWaitValue32(hipStream_t& stream, void*& ptr, uint32_t& value,
+                                       unsigned int& flags, uint32_t& /*mask*/) {
+  ClPrint(amd::LOG_DETAIL_DEBUG, amd::LOG_API,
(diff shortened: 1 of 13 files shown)


projects/clr/hipamd/src/hip_graph.cpp
```diff
@@ -3788,4 +3788,110 @@ hipError_t hipGraphExecBatchMemOpNodeSetParams(hipGraphExec_t hGraphExec, hipGra
...
+  s->SetLastCapturedNode(node);
+  return hipSuccess;
+}
+
+hipError_t capturehipStreamWaitValue32(hipStream_t& stream, void*& ptr, uint32_t& value,
+                                       unsigned int& flags, uint32_t& /*mask*/) {
+  ClPrint(amd::LOG_DETAIL_DEBUG, amd::LOG_API,
+          "[hipGraph] Current capture node StreamWaitValue32 on stream : %p", stream);
+  hipStreamBatchMemOpParams op{};
+  op.operation = hipStreamMemOpWaitValue32;
```
saleelk, line 3858: Where did the mask go?

projects/hip-tests/catch/config/configs/unit/stream.yaml
```diff
@@ -415,6 +415,18 @@ stream:
   Unit_hipStreamBatchMemOp_Negative_Tests:
     <<: *level_2
     tags: [memory]
+  Unit_hipStreamWaitWriteValue32_Capture:
```
jaydeeppatel1111, line 426: Original issue is related hipStreamMemOps hang and can be reproduced without graph involved so it is good to add test without graph use. something that just to test ordering of batch mem ops.

Issue is reproduced without Graph involved. hipStreamBatchMemOp, __amd_batchMemOp runs the batch ops in parallel and that does not garuntee the program ops order. Specifically in this case, 
	Op1 : wave0 - GPU1 updates GPU0's flag to 1
	Op2 : wave1 - GPU0 waits on its flag == 1
	Op3 : wave2 - GPU0 resets its flag = 0.

projects/clr/rocclr/device/blitcl.cpp
```diff
@@ -192,8 +192,35 @@ const char* BlitLinearSourceCode = BLIT_KERNELS(
       __amd_copyBufferRectAligned(src, dst, srcRect, dstRect, size);
     }
 
+    // TODO: Once the sequential for-loop fix lands in llvm-project/amd/device-libs/opencl/src/misc/amdblit.cl
+    // (replacing get_global_id(0) with a for loop), revert this back to:
+    //   __amd_batchMemOp(params, count);
     __kernel void __amd_rocclr_batchMemOp(__global void* params, uint count) {
```
jaydeeppatel1111, line 230: I think, This will break blit kernel compilation.

projects/clr/rocclr/device/blitcl.cpp
```diff
@@ -192,8 +192,35 @@ const char* BlitLinearSourceCode = BLIT_KERNELS(
...
-      __amd_batchMemOp(params, count);
+      __global BatchMemOpParams* p = (__global BatchMemOpParams*)params;
+      for (uint i = 0; i < count; i++) {
+        switch (p[i].operation) {
+          case STREAM_WAIT_VALUE_32:
+            __amd_streamOpsWait((__global atomic_uint*)p[i].waitValue.address, NULL,
+                                (uint)p[i].waitValue.value, (uint)p[i].waitValue.flags,
+                                (ulong)~0UL);
+            break;
+          case STREAM_WRITE_VALUE_32:
```
saleelk: where are these defined?

projects/hip-tests/catch/config/configs/unit/stream.yaml
anugodavar: added test to check the non-graph functionality.

projects/clr/rocclr/device/blitcl.cpp
anugodavar: updated

projects/clr/rocclr/device/blitcl.cpp
anugodavar: added enums for these in the same file. We should removed them when changes are in upstream llvm

projects/clr/hipamd/src/hip_graph.cpp
anugodavar: - CUDA: mask is in the hipStreamWaitValue32 signature for API compatibility but is always ignored — not in hardware, not in BatchMemOp
- AMD non-capture: mask IS used — passed to the AQL barrier-value packet which applies it in hardware
- AMD capture path: mask is dropped — matches CUDA since hipStreamBatchMemOpParams (which mirrors CUstreamBatchMemOpParams) has no mask field

saleelk (approved):

projects/hip-tests/catch/unit/stream/hipStreamBatchMemOp.cc
```diff
@@ -91,6 +106,174 @@ HIP_TEST_CASE(Unit_hipStreamBatchMemOp_Negative_Tests) {
...
+    }
+  }
+
+<<<<<<< HEAD
+  for (int d = 0; d < deviceCount; d++) {
+=======
+  // Pop contexts in reverse creation order — hipCtxCreate pushes onto the per-thread
+  // context stack, so the last-created context is on top and must be popped first.
+  for (int d = deviceCount - 1; d >= 0; d--) {
+>>>>>>> ae0d968372 (fix(hip-tests): address Copilot review comments for BatchMemOp tests)
```
eble-amd: incomplete merge

anugodavar:
unit tests are added

projects/clr/hipamd/src/hip_graph.cpp
```diff
@@ -3788,4 +3788,116 @@ hipError_t hipGraphExecBatchMemOpNodeSetParams(hipGraphExec_t hGraphExec, hipGra
   }
   HIP_RETURN(reinterpret_cast<hip::hipGraphBatchMemOpNode*>(clonedNode)->SetParams(nodeParams));
 }
+
+hipError_t capturehipStreamBatchMemOp(hipStream_t& stream, unsigned int& count,
+                                      hipStreamBatchMemOpParams*& paramArray,
+                                      unsigned int& flags) {
```
AlexeySachkov, line 3794: Why do we need to accept everything by reference here? `hipStream_t` is an opaque pointer, plain integers take less space is passed directly, etc.

projects/clr/hipamd/src/hip_graph.cpp
```diff
@@ -3788,4 +3788,116 @@ hipError_t hipGraphExecBatchMemOpNodeSetParams(hipGraphExec_t hGraphExec, hipGra
...
+  if (!hip::isValid(stream)) {
+    return hipErrorContextIsDestroyed;
+  }
+  if (paramArray == nullptr || count == 0 || count > 256 || flags != 0) {
+    return hipErrorInvalidValue;
+  }
+  hip::Stream* s = reinterpret_cast<hip::Stream*>(stream);
+
+  hipBatchMemOpNodeParams nodeParams{};
+  nodeParams.ctx = reinterpret_cast<hipCtx_t>(hip::getCurrentDevice());
```
AlexeySachkov, line 3806: Is it a valid operation? `hip::getCurrentDevice()` returns `Device *` and I do not see why it would be interchangeable with context. `Device` class does have `asContext()` method though

PR #8912, opened 2026-07-21
fix(rocr): load thunk libraries from the runtime directory

fcui-amd:
## Motivation

# rocminfo
WSL environment detected.
Cannot load librocdxg.so, failed:librocdxg.so: cannot open shared object file: No such file or directory
GetExportAddress failed: /opt/venv/lib/python3.14/site-packages/_rocm_sdk_devel/bin/../lib/libhsa-runtime64.so.1: undefined symbol: hsaKmtOpenKFD
hsa_init Failed, possibly no supported GPU devices

## Technical Details

Add a fallback path for loading thunk libraries when the normal platform library search fails.
The fallback locates the directory containing the currently loaded ROCR runtime library and attempts to load the thunk library from the same directory.

## Test Plan

rocminfo pass without specified LD_LIBRARY_PATH setting

## Test Result

rocminfo pass

## Submission Checklist


kentrussell (commented):
Thunk part is fine. Will let David/Chris give the approval for the ROCr side

projects/rocr-runtime/libhsakmt/src/libhsakmt.ver
```diff
@@ -69,11 +69,14 @@ hsaKmtQueryPointerInfo;
...
 hsaKmtGetQueueInfo;
 hsaKmtAllocQueueGWS;
+hsaKmtGetKernelQueueId;
 hsaKmtRuntimeEnable;
 hsaKmtRuntimeDisable;
 hsaKmtCheckRuntimeDebugSupport;
 hsaKmtGetRuntimeCapabilities;
 hsaKmtDebugTrapIoctl;
+hsaKmtGetCoreRuntimeInfo;
+hsaKmtGetCoreDeviceInfo;
```
cfreeamd, line 79: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** These three exports (`hsaKmtGetKernelQueueId`, `hsaKmtGetCoreRuntimeInfo`, `hsaKmtGetCoreDeviceInfo`, mirrored in `librocdxg.ver`) come from the separate "export missing libhsakmt symbols" commit and are unrelated to the adjacent-path loader change. Two things worth confirming: (a) are all three actually **defined** in libhsakmt? A `.ver` entry for an undefined symbol is a silent no-op. (b) Is widening the public export surface intended as part of this PR, or would it be cleaner to split it out? The ABI change and the loader change have independent risk/review profiles.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/thunk_loader.cpp
```diff
@@ -88,8 +96,18 @@ namespace core {
...
+          thunk_handle = rocr::os::LoadLib(relative_path.c_str());
+        }
+        if (thunk_handle == nullptr) {
+          const char* error = rocr::os::DlError();
+          fprintf(stderr, "Cannot load %s, failed:%s\n", library_name.c_str(),
+            error == nullptr ? "unknown error" : error);
+        }
+      }
+      if (thunk_handle != nullptr) {
         debug_print("Load %s successully!\n", library_name.c_str());
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** When the library is loaded via the adjacent-path fallback, this success message still prints `library_name` (the original name), so the log won't reveal that the fallback path was actually used. Consider logging the path that succeeded to aid diagnosis. (Also a pre-existing typo here while the line is nearby: "successully" → "successfully".)

cfreeamd (commented):

The adjacent-path loader fallback in `thunk_loader.cpp` and the new `os::GetAdjacentLibraryPath` implementations (Linux `dladdr`, Windows `GetModuleHandleExA`/`GetModuleFileNameA`) look correct — self-locating via the runtime's own address, with sane truncation handling on Windows and a null-guard on `DlError()`. Two inline notes below (one Question on the bundled `.ver` exports, one Nit on the success log). Noting the PR is labeled "Not ready to Review," so this is early feedback.

projects/rocr-runtime/libhsakmt/src/libhsakmt.ver
fcui-amd: Sorry, this commit belongs to another PR and was included here by mistake.

dayatsin-amd (approved):

cfreeamd (commented):

Both prior items are resolved:

- **[Question] bundled `.ver` ABI additions** — resolved. Confirmed by the author reply ("belongs to another PR and was included here by mistake") and the commit was dropped; `libhsakmt.ver` / `librocdxg.ver` are no longer in the diff. The PR is now a single commit scoped to the adjacent-path loader.
- **[Nit] success log + typo** — resolved. The success `debug_print` now logs `loaded_path` (so the log reveals when the adjacent fallback path was used), and the `successully` typo is corrected to `successfully`.

No new concerns from the changes. The loader fallback and both `os::GetAdjacentLibraryPath` implementations remain correct.

cfreeamd (approved):

cfreeamd (approved):

PR #8916, opened 2026-07-21
feat(clr): any-order overlap for oversubscribed shared HW queues.

jaydeeppatel1111:
## Motivation
When the number of active HIP streams exceeds the HW queue pool
(`GPU_MAX_HW_QUEUES`, default 4), extra streams are recycled onto shared
physical HSA queues. Today every AQL kernel dispatch carries the barrier bit
(`HSA_PACKET_HEADER_BARRIER`), which forces the command processor to drain all
prior packets on that queue before starting the next one. On a shared queue this
serializes *independent* streams against each other, even though only
*intra-stream* order needs to be preserved. On oversubscribed workloads
(common on MI450-class parts) this leaves the GPU underutilized.
This change lets independent streams that happen to share a pooled HW queue
overlap, while strictly preserving ordering within each stream.

## Technical Details
Gated entirely behind a new opt-in flag (default off):
  `DEBUG_HIP_SHARED_QUEUE_ANYORDER` (release uint, 0=off/1=on).
  Opt-in/unsafe: the app must guarantee the sharing streams are independent
  (no cross-stream data hazard).
Per-packet decision (replaces the earlier first-kernel-only approach). For each
kernel dispatch on a non-dedicated, non-graph stream we decide *per packet*
whether the barrier bit can be cleared, tracking state per (stream, HW queue):
- Per-stream (`VirtualGPU`): `last_dispatch_index_` / `last_dispatch_queue_`
  (this stream's previous dispatch slot and the ring it refers to),
  `sharedAnyOrderDispatched_` (ever dispatched on a shared ring),
  `sharedAnyOrderActive_` (this packet is a user-kernel dispatch to evaluate),
  `sharedAnyOrderEligible_` (independent enough to clear).
- Per-HW-queue (`Device`): `shared_queue_barrier_index_` — newest slot on that
  ring whose barrier bit was *kept* — guarded by a dedicated monitor lock.
`Device::SharedQueueAnyOrderDecision(queue, my_slot, my_prev_slot, eligible, treat_as_first)`:
  - First-ever dispatch by this stream on this ring → clear (no in-stream
    predecessor to fence against).
  - Known predecessor slot → clear only if a kept barrier landed strictly
    between the stream's previous slot and this one
    (`prev < largest_kept_barrier < current`), i.e. the predecessor is already
    fenced on the ring.
  - Predecessor exists but slot unknown → keep the bit (preserve ordering).
  - When the bit is kept, this dispatch becomes the ring's newest barrier.
Eligibility (all required to clear): ring is actually shared (`refCount > 1`),
no event wait, not cooperative, not a dedicated queue, not in graph capture,
and not the explicit `hipExt*AnyOrderLaunch` path (which clears unconditionally).
Correctness safeguards:
  - Every dispatch — user kernels *and* internal fills/copies — updates the
    stream's predecessor slot, so a kernel following an internal memset/copy is
    never misread as "first".
  - Explicit barrier/fence packets always land with the bit set and are
    recorded via `SharedQueueRecordBarrier()`.
  - `SharedQueueResetBarrier()` drops stale per-queue state when a fresh HW
    queue is created at a previously-used key.
  - `last_dispatch_index_` is preserved across release/re-acquire of the *same*
    ring and reset only when binding to a *different* ring.
  - Acquire/release fence scopes are unchanged; clearing the barrier bit does
    not alter memory-visibility scope.

## Issue Tracking
JIRA ID : AIRUNTIME-74

## Test Plan
New unit tests (`catch/unit/stream/hipSharedQueueAnyOrderOverlap.cc`), each
passing with the flag both OFF and ON, and under forced single-ring
(`GPU_MAX_HW_QUEUES=1`):
  - `_Independence` — per-stream isolation under heavy oversubscription (32 streams).
  - `_EventDependencyHonored` — explicit cross-stream event deps not broken.
  - `_IntraStreamOrderAcrossActivation` — sole-tenant → shared transition.
  - `_InternalPredecessorOrdering` — kernel after internal memset/copy.
  - `_PriorityStreamsIndependence` — hi/lo priority pool isolation.
  - `_NullStreamOrdering` — null/blocking stream correctness.
  - `_GraphUnaffected` — graph dependencies preserved (feature inert in capture).
New perf scenarios (`catch/performance/scenarios/stream/hipPerfSharedQueueAnyOrderOverlap.cc`):
  - `Performance_hipPerfSharedQueueAnyOrderOverlap` — independent-kernel overlap swept 2/4/8/16/32 streams.
  - `Performance_hipPerfSharedQueueAnyOrderOverlap_EventDependency` — overlap with producer→event→consumer pairs.
Tests registered in the config leveling YAMLs (`config/configs/unit/stream.yaml`,
`config/configs/performance.yaml`). Full `Unit_` sweep (4508 tests) run with the
flag OFF and ON to check for regressions.

## Test Result
Environment: gfx1250, DKMS 2374403, ROCk 169, CLR + ROCR-Runtime built from src.
Correctness:
  - All 7 unit tests pass with flag OFF and ON (default pool and `GPU_MAX_HW_QUEUES=1`).
  - Single-ring 32-stream stress (`_Independence` x10 each): OFF 10/10, ON 10/10, no GPU wedge.
  - Full sweep: OFF 11 fail / 4508, ON 11 fail / 4508 — **identical** failure set →
    0 flag-ON regressions. (The 11 are pre-existing device-library/numeric
    baseline failures — `bf16_basic`, `__h*_Accuracy` compare intrinsics,
    `ext_ocp_fp8_scaled_decode_host`, `gcc_half_types` — that fail with the flag
    off as well, unrelated to this change.)
Performance (independent kernels, µs/iter, lower is better):
```
  | streams | OFF     | ON      | speedup |
  |---------|---------|---------|---------|
  | 2       |  33,030 |  33,810 | ~1.0x   |
  | 4       |  34,880 |  36,443 | ~1.0x   |
  | 8       | 106,916 |  36,979 | 2.9x    |
  | 16      | 139,967 |  52,905 | 2.6x    |
  | 32      | 293,737 | 145,269 | 2.0x    |
```
  Feature yields 2–2.9x overlap on oversubscribed independent work and is
  correctly neutral otherwise: ≤ pool size is unchanged, and the
  event-dependency scenario is within noise ON vs OFF (dependency-bound
  dispatches are ineligible to clear the bit — no win, no regression).

## Submission Checklist


jaydeeppatel1111:
Closing as superseded by #9895 (merged to develop).

PR #8920, opened 2026-07-21
clr/hip-tests: fix half/bfloat16 __hne NaN comparison

iassiour:
## Motivation
 `__hne` (half/bfloat16 ordered not-equal) returns the wrong result for NaN inputs on all architectures: it uses C's `!=`, which is the *unordered* not-equal (NaN→true), when the ordered predicate must  
  return false. This makes `__hne` identical to `__hneu` — they emit the same machine instruction (`v_cmp_neq`). CUDA's `__hne` uses `setp.ne` (ordered, NaN→false).

## Technical Details
  - **fp16** (`amd_hip_fp16.h`): `__hne`/`__hne2` reimplemented as `(x < y) || (x > y)`, which folds to a single `v_cmp_lg` (the native ordered ne) — zero cost, just the correct opcode. The unordered
  bool2 reductions (`__hbequ2`/`__hbneu2`/`__hbleu2`/`__hbgeu2`/`__hbltu2`/`__hbgtu2`) were aliasing their ordered counterparts; fixed to reduce the unordered vector intrinsics instead.                   
  - **bf16** (`amd_hip_bf16.h`): same `__hne` fix.                                                                                                                                       
  - **Tests**: re-enable the 10 fp16 comparison accuracy tests previously disabled under EXSWHTEC-356. Add NaN-comparison coverage to the existing bf16 `Unit_bf16_basic` test.    

## Issue Tracking
JIRA ID: ROCM-28334

## Test Plan
  - Ran all fp16 comparison accuracy tests (`__hne`, `__hne2`, `__hbne2`, `__hbequ2`, `__hbgeu2`, `__hbgtu2`, `__hbleu2`, `__hbltu2`, `__hisinf2`, `__hisnan2`): all pass.
  - Ran device test with `(1,1)`, `(1,2)`, `(NaN,1)`, `(1,NaN)`, `(NaN,NaN)` for both fp16 and bf16 `__hne`/`__hneu`: verified `__hne` now differs from `__hneu` on NaN (returns 0 vs 1).                   
  - bf16 NaN-comparison section of `Unit_bf16_basic`: passes with fix, fails without (all 6 NaN pairs).                                                                                  
  - Verified no regression on sibling comparisons (`__heq`, `__hequ`, `__hbneu2`, etc.). 
## Test Result
All comparison tests pass. The fix is a single-opcode swap (`v_cmp_neq` → `v_cmp_lg`), verified on gfx942 — same instruction count, zero performance cost.

## Submission Checklist

### DIFF
--- a/projects/hip-tests/catch/unit/deviceLib/bfloat16.cc
+++ b/projects/hip-tests/catch/unit/deviceLib/bfloat16.cc
@@ -155,10 +155,20 @@ __global__ void bf16_compare(float* val, unsigned* res, size_t size) {
   }
 }
 
+// Order of predicates written by bf16_compare_nan(); the host uses the same
+// indices to label failures and look up the expected result.
+enum Bf16NanPredicate {
+  kHeq, kHne, kHlt, kHle, kHgt, kHge,        // ordered scalar   -> false for NaN
+  kHequ, kHneu, kHltu, kHleu, kHgtu, kHgeu,  // unordered scalar -> true for NaN
+  kHbeq2, kHbne2,                            // ordered bool2    -> false for NaN
+  kBf16NanPredicateCount
+};
+
 // Test NaN comparison semantics: for inputs where at least one operand is NaN the
 // ordered predicates (no 'u' suffix) must return false and the unordered predicates
-// ('u' suffix) must return true. Returns 1 only if every relation has the expected
-// result. Each a[i]/b[i] pair must have at least one NaN.
+// ('u' suffix) must return true. Each thread writes one slot per predicate so the
+// host can identify exactly which intrinsic misbehaved. Each a[i]/b[i] pair must
+// have at least one NaN.
 __global__ void bf16_compare_nan(const float* a, const float* b, unsigned* res, size_t size) {
   auto i = threadIdx.x;
   if (i < size) {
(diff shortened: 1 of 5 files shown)


b-sumner (approved):
If __hne is meant to be an ordered comparison, then this is a good fix.

iassiour:
> If __hne is meant to be an ordered comparison, then this is a good fix.

Yes I think __hne is meant to  be an ordered comparison as there is already a __hneu variant (for unordered)  where NaN inputs generate true results. That is consistent with https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/group__CUDA__MATH____HALF__COMPARISON.html#group__cuda__math____half__comparison_1ga29c75ff5a19b9e4808b97d9f55259468

projects/hip-tests/catch/unit/deviceLib/bfloat16.cc
```diff
@@ -155,6 +155,31 @@ __global__ void bf16_compare(float* val, unsigned* res, size_t size) {
...
+__global__ void bf16_compare_nan(const float* a, const float* b, unsigned* res, size_t size) {
+  auto i = threadIdx.x;
+  if (i < size) {
+    const __hip_bfloat16 x = __float2bfloat16(a[i]);
+    const __hip_bfloat16 y = __float2bfloat16(b[i]);
+    const __hip_bfloat162 x2{x, x};
+    const __hip_bfloat162 y2{y, y};
+
+    // Ordered scalar predicates -> false for NaN.
+    const bool ordered = !__heq(x, y) && !__hne(x, y) && !__hlt(x, y) && !__hle(x, y) &&
```
cjatin: minor nit: its fine, but when these tests do fail due to some reason, its not fun debugging which condition went wrong.

cjatin (approved):

projects/hip-tests/catch/unit/deviceLib/bfloat16.cc
iassiour: Good point. I changed it to check each condition separately.

gandryey (approved):

PR #8921, opened 2026-07-21
fix(rocprof-systems): Fixes to pytest path resolution

sputhala-amd:
## Motivation

The pytest harness could silently resolve rocprofiler-systems binaries and `example/target` executables from the system PATH (e.g. `/opt/rocm/bin`) when they were missing from the build/install tree. That let a test exercise a different,
already-installed build instead of the one under test, producing misleading pass/fail results.

## Technical Details

- `_find_executable` - no longer falls back to `shutil.which`; it searches only the provided build/install `bin` paths.
- `get_target_executable` no longer falls back to PATH in the install and the build branch. Both branches now resolve only from the in-tree `examples/bin` directories.
- Both continue to honor the `ROCPROFSYS_BUILD_DIR` (build tree) and `ROCPROFSYS_INSTALL_DIR` (install prefix) overrides, which drive the searched directories.
- Clearer "not found" errors and the `get_target_executable` docstring now documents both search orders and states PATH is never consulted.
- Note: an example/target absent from the tree now skips (`FileNotFoundError` -> `pytest.skip`) instead of profiling a `PATH` copy.

## Submission Checklist


sputhala-amd:
This branch has stale commits from a different branch. Closing this PR and creating a new one with the same changes: https://github.com/ROCm/rocm-systems/pull/8942

PR #8923, opened 2026-07-21
clr: Fix missing else-branches in is_invocable guards

maarquitos14:
## Motivation

Some `__builtin_amdgcn_*` calls with `__builtin_amdgcn_is_invocable` guards present two classes of bugs by not handling the case where `is_invocable` returns false at runtime:

1. Existing fallback made unreachable. In files like amd_hip_fp16.h and amd_hip_bf16.h, a CAS-loop fallback existed under `#else` for when `__has_builtin` is false. The builtin call is wrapped inside an `if (is_invocable)` but left the return statement outside the if, making the `#else CAS fallback` dead code. On SPIR-V targets, `__has_builtin` is true (the compiler knows the builtin) but `is_invocable` might return false at runtime (the hardware doesn't support it), so the function returns the original value without performing the atomic, resulting in a silent no-op.
2. No fallback added. In files like amd_hip_cooperative_groups_memcpy.h and amd_warp_functions.h, some builtin calls are wrapped with `if (is_invocable)` without adding an else branch. When `is_invocable` is false, the operation is silently skipped. As an example, for the memcpy case, this means pointer arithmetic advances without any data actually being copied.

## Technical Details

There are different types of fixes depending on the affected builtin:

- When a CAS-loop fallback is available (amd_hip_fp16.h, amd_hip_bf16.h, amd_hip_unsafe_atomics.h, amd_hip_atomic.h): restructured the preprocessor/runtime flow so the CAS loop is reachable when `is_invocable` is false.
- When a fence fallback is available (amd_device_functions.h, amd_warp_sync_functions.h, hip_cooperative_groups_helper.h): added `__atomic_thread_fence` as the else-branch for `__builtin_amdgcn_fence guards`.
- When a software clamp fallback mechanism is available (amd_hip_fp8.h): added min/max software saturation as the else-branch for fmed3f guards.
- When there is no software equivalent (amd_warp_functions.h, amd_hip_cooperative_groups.h, amd_hip_cooperative_groups_memcpy.h, amd_hip_fp4.h, amd_hip_fp6.h): added `__builtin_trap()` to abort loudly rather than silently producing wrong results (e.g. no op).

In general, the philosophy is use a fallback when possible, otherwise produce an error rather than a silent no-op/wrong result.

## Issue Tracking
N/A

## Test Plan
Build and HeCBench/f16atomic-hip test, which was the test that uncovered the issue for `unsafeAtomicAdd`.

## Test Result
HeCBench/f16atomic-hip test calls `unsafeAtomicAdd` which was silently no-op'ing, resulting in wrong validation. With this patch, the result validates correctly.

## Submission Checklist

### DIFF
--- a/projects/clr/hipamd/include/hip/amd_detail/amd_device_functions.h
+++ b/projects/clr/hipamd/include/hip/amd_detail/amd_device_functions.h
@@ -652,7 +652,10 @@ __device__ inline __attribute__((always_inline)) long long int clock() { return 
 
 // hip.amdgcn.bc - named sync
 __device__ inline void __named_sync() {
-  if (__builtin_amdgcn_is_invocable(__builtin_amdgcn_s_barrier)) __builtin_amdgcn_s_barrier();
+  if (__builtin_amdgcn_is_invocable(__builtin_amdgcn_s_barrier))
+    __builtin_amdgcn_s_barrier();
+  else
+    __builtin_trap();
 }
 
 #endif  // __HIP_DEVICE_COMPILE__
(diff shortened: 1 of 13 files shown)


projects/clr/hipamd/include/hip/amd_detail/amd_device_functions.h
```diff
@@ -683,16 +686,22 @@ __device__ inline __hip_uint64_t __lanemask_eq() {
 __device__ inline static void __threadfence() {
   if (__builtin_amdgcn_is_invocable(__builtin_amdgcn_fence))
     __builtin_amdgcn_fence(__ATOMIC_SEQ_CST, "agent");
+  else
+    __atomic_thread_fence(__ATOMIC_SEQ_CST);
 }
 
 __device__ inline static void __threadfence_block() {
   if (__builtin_amdgcn_is_invocable(__builtin_amdgcn_fence))
```
arsenm, line 694: These guards should not be used on all of these unconditionally available builtins

projects/clr/hipamd/include/hip/amd_detail/amd_hip_atomic.h
```diff
@@ -599,13 +599,27 @@ __device__ inline double atomicMax_system(double* addr, double val) {
 __device__ inline unsigned int atomicInc(unsigned int* address, unsigned int val) {
   if (__builtin_amdgcn_is_invocable(__builtin_amdgcn_atomic_inc32))
     return __builtin_amdgcn_atomic_inc32(address, val, __ATOMIC_RELAXED, "agent");
```
arsenm, line 601: This builtin should not be used, this function should be reimplemented with the atomic statement attributes and the scoped atomic builtins

PR #8925, opened 2026-07-21
fix(amdsmi): write enabled/disabled to ptl_enable sysfs node

### LINKED ISSUE #8924: amd-smi set --ptl-status reports success but state stays Disabled (writes "1"/"0" instead of "enabled"/"disabled")
## Summary

`sudo amd-smi set --ptl-status 1` reports *"Successfully set PTL state to Enabled"* for every GPU, but a subsequent read still reports `Disabled`. No reboot changes this — the write simply never takes effect.

## Root cause

The `device/ptl/ptl_enable` sysfs node accepts (and reports on read) the strings `"enabled"`/`"disabled"`. `rsmi_set_gpu_ptl_state()` routes through `Device::writeDevInfo(DevInfoTypes, uint64_t)`, which wrote `std::to_string(val)` — i.e. `"1"`/`"0"`. The driver's `store` handler doesn't recognize `"1"`, silently ignores it, and returns success for the `write()`. As a result:

1. The `write()` syscall succeeds, so no error is generated.
2. `rsmi_set_gpu_ptl_state()` returns `SUCCESS`, so the CLI prints "Successfully set…".
3. The state never changes, so the next read still shows `Disabled`.

The read path (`rsmi_get_gpu_ptl_state()`) already compares against `"enabled"`/`"disabled"`, and the rdc PTL guard reads/writes those same strings — so the write path was the sole inconsistency.

## Reproduction (on a driver that exposes `ptl/ptl_enable`)

```
sudo amd-smi set --ptl-status 1     # -> "Successfully set PTL state to Enabled"
amd-smi static --gpu 0 | grep PTL   # -> PTL_STATE: Disabled
```

Direct sysfs confirmation:
```
P=/sys/class/drm/card0/device/ptl/ptl_enable
echo 1       | sudo tee $P; cat $P   # stays "disabled"
echo enabled | sudo tee $P; cat $P   # -> "enabled"
```

## Fix

Write `"enabled"`/`"disabled"` from `writeDevInfo()` for `kDevPtlStatus`, and pass `returnWriteErr=true` so a genuinely rejected write surfaces the real errno instead of a generic status. The CLI interface (`--ptl-status 0|1`) is unchanged.

## Affected hardware

Any GPU exposing the `ptl/ptl_enable` sysfs node (e.g. MI300X/MI308X on PTL-enabled driver builds).

adam360x:
## Motivation

`sudo amd-smi set --ptl-status 1` reports *"Successfully set PTL state to Enabled"* on every GPU, but the state still reads `Disabled` afterwards, with no error and no reboot fixing it. The write was silently a no-op.

## Technical Details

The `device/ptl/ptl_enable` sysfs node accepts — and reports on read — the strings `"enabled"`/`"disabled"`. `rsmi_set_gpu_ptl_state()` routes through `Device::writeDevInfo(DevInfoTypes, uint64_t)`, which wrote `std::to_string(val)`, i.e. `"1"`/`"0"`. The driver's `store` handler doesn't recognize `"1"`, silently ignores it, and returns success for the `write()`, so:

1. the `write()` syscall succeeds → no error is generated;
2. `rsmi_set_gpu_ptl_state()` returns `SUCCESS` → the CLI prints "Successfully set…";
3. the state never changes → the next read still shows `Disabled`.

The read path (`rsmi_get_gpu_ptl_state()`) already compares against `"enabled"`/`"disabled"`, and the in-tree rdc PTL guard reads/writes those same strings, so the write path was the sole inconsistency.

This change writes `"enabled"`/`"disabled"` for `kDevPtlStatus`, and passes `returnWriteErr=true` so a genuinely rejected write surfaces the real errno instead of being masked as a generic status. The CLI interface (`--ptl-status 0|1`) is unchanged.

## Issue Tracking

Closes #8924

## Test Plan

- `clang-format --Werror` on the changed file.
- `-fsyntax-only` compile of the translation unit using `compile_commands.json`.
- Direct sysfs check on a PTL-capable box:
  ```
  P=/sys/class/drm/card0/device/ptl/ptl_enable
  echo 1       | sudo tee $P; cat $P   # old behavior -> stays "disabled"
  echo enabled | sudo tee $P; cat $P   # new behavior -> "enabled"
  ```

## Test Result

- clang-format: clean.
- Syntax-only compile: exit 0 (only pre-existing, unrelated warnings).
- Existing CLI tests exercise the `--ptl-status 0|1` interface, which is unchanged.

## Submission Checklist

### DIFF
--- a/projects/amdsmi/CHANGELOG.md
+++ b/projects/amdsmi/CHANGELOG.md
@@ -115,6 +115,9 @@ Full documentation for amd_smi_lib is available at [https://rocm.docs.amd.com/pr
 
 ### Resolved Issues
 
+- **Fixed `amd-smi set --ptl-status` silently failing to change PTL state**.  
+  - The set path wrote `"1"`/`"0"` to the `ptl/ptl_enable` sysfs node, which only accepts `"enabled"`/`"disabled"`; the driver ignored the numeric write while the API still reported success. The state now changes as expected, and a rejected write returns a real error instead of a generic success.
+
 - **Fixed `amd-smi set --power-cap` rejecting the minimum allowed value**.  
   - The lower bound is now inclusive, so setting the power cap to the exact minimum of the reported range (e.g. `210` when the range is 210-300W) succeeds instead of failing validation, matching the inclusive range shown in the error message.
 
--- a/projects/amdsmi/CHANGELOG.md
+++ b/projects/amdsmi/CHANGELOG.md
@@ -8,6 +8,9 @@ Full documentation for amd_smi_lib is available at [https://rocm.docs.amd.com/pr
 
 ### Resolved Issues
 
+- **Fixed `amd-smi set --ptl-status` silently failing to change PTL state**.  
+  - The set path wrote `"1"`/`"0"` to the `ptl/ptl_enable` sysfs node, which only accepts `"enabled"`/`"disabled"`; the driver ignored the numeric write while the API still reported success. The state now changes as expected, and a rejected write returns a real error instead of a generic success.
+
 - **Fixed `amd-smi process` hiding compute processes owned by other users**.  
   - A caller without permission to read another process's `/proc/<pid>/fd` was misdetected as running in a separate PID namespace, which caused the whole compute-process list to come back empty. Such processes are now listed with a redacted (`N/A`) name instead of being dropped.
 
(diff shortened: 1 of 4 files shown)


marifamd (approved):

yalmusaf (approved):

gabrpham (approved):
Looks good to me. Approved.

PR #8932, opened 2026-07-21
[hip-tests] Standardize test level discovery

amilanov-amd:
## Motivation

Currently, it is not strictly clear with which level hip-tests will be executed with, due to a lot of ways they can be actually executed. 

## Technical Details

Standardize to three possible ways:
1. Catch2 command-line tag filter, e.g. ./test "[level_2]"
2. HIP_TEST_LEVEL environment variable, e.g. HIP_TEST_LEVEL=level_2
3. Hardcoded default (kDefaultLevel)

Remove all other invisible, non-deterministic, fallbacks.

Define level_2 as default level, using its parameters for test execution.

Define a strategy to determine level when multiple are passed - highest order wins (used for test parameters).

## Issue Tracking

-

## Test Plan

Ensure build passes. All tests pass

## Test Result

Local build passed, local tests pass

## Submission Checklist

### DIFF
--- a/projects/hip-tests/catch/hipTestMain/hip_test_listener.cc
+++ b/projects/hip-tests/catch/hipTestMain/hip_test_listener.cc
@@ -22,12 +22,12 @@ THE SOFTWARE.
 
 #include <catch2/reporters/catch_reporter_event_listener.hpp>
 #include <catch2/reporters/catch_reporter_registrars.hpp>
-#include <catch2/catch_test_case_info.hpp>
 #include <catch2/interfaces/catch_interfaces_config.hpp>
 #include <hip_test_params.hh>
 #include <hip_test_context.hh>
 #include <regex>
 #include <set>
+#include <sstream>
 #include <string>
 #include <cstdlib>
 #include <cstdio>
@@ -56,10 +56,18 @@ THE SOFTWARE.
  * misconfiguration and aborts the run rather than silently testing with the
  * wrong parameters.
  *
- * Both the command-line filter and HIP_TEST_LEVEL accept multiple levels using
- * the Catch2 tag format (e.g. "[level_1],[level_2]"); when more than one is
- * given the highest level wins. HIP_TEST_LEVEL additionally accepts the bare
- * form without brackets (e.g. "level_2" or "level_1,level_2").
+ * Both sources are parsed with the same rules, following the Catch2 test spec
+ * grammar: a comma separated list of filters that are OR'd, each filter a
+ * sequence of AND'd patterns that may be negated with '~'. A pattern counts
+ * only if it reads exactly "level_N", so a malformed tag ("[level_2x]") is
+ * ignored, while an exclusion ("~[level_2]") removes that level from the set
+ * that can still run. Of the levels that remain, the highest wins:
+ *   "[level_1],[level_2]"    -> level_2
+ *   "~[level_3]~[level_4]"   -> level_2 (highest functional level left)
+ *   "[level_0]~[level_0]"    -> no level (nothing left to run)
+ * The single exception between the two sources is that HIP_TEST_LEVEL also
+ * accepts the bare form without brackets ("level_2", "level_1,level_2",
+ * "~level_4").
  */
 class HipTestParameterListener : public Catch::EventListenerBase {
 public:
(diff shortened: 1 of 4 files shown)


projects/hip-tests/catch/include/hip_test_params.hh
```diff
@@ -134,11 +134,15 @@ private:
...
-     * @brief Fallback parameters (if no level specified)
+     * @brief Fallback parameters (if no level specified) - mirror level_2.
      */
-    std::vector<size_t> defaultMemorySizes;
-    std::vector<int> defaultBlockSizes;
-    int defaultIterations = 1000;
-    int defaultWarmups = 100;
-    size_t defaultMaxMemory = 2147483648; // 2GB
+    // 64, 256, 1K, 4K, 16K, 64K, 256K, 1M, 10M, 50M, 100M, 500M, 1G, 2G
+    const std::vector<size_t> defaultMemorySizes = {64,        256,       1024,       4096,
```
cjatin: if these are const and we just use them to look up, can we use `constexpr std::array` for these values?

cjatin (approved):

projects/hip-tests/catch/hipTestMain/hip_test_listener.cc
```diff
@@ -22,156 +22,230 @@ THE SOFTWARE.
...
  */
 class HipTestParameterListener : public Catch::EventListenerBase {
 public:
     using Catch::EventListenerBase::EventListenerBase;
-    
+
 private:
+    /// Levels understood by the test suite (must match definitions.yaml).
+    static constexpr const char* kSupportedLevels[] = {"level_0", "level_1", "level_2",
+                                                       "level_3", "level_4"};
```
mangupta, line 79: 🤖 AI-assisted pre-review (for @mangupta)

**[IMPORTANT]** Level validation uses a hardcoded list instead of the levels actually generated into the parameter header

The resolved level is validated against the hand-maintained `kSupportedLevels` array (`level_0`..`level_4`) rather than against the levels `TestParameterStore::initialize()` actually loaded from the build-generated parameter header. The two can disagree in both directions, and each direction produces a bad outcome: a level that is in `kSupportedLevels` but absent from the generated header passes validation, `loadLevelConfig()` sets a level with no entries, and every accessor silently falls through to the hardcoded fallback parameters — exactly the "silently testing with the wrong parameters" outcome the new validation is documented to prevent; conversely a level that is added to the YAML config and generated into the header but not added to `kSupportedLevels` (e.g. `level_5`) makes `./test "[level_5]"` or `HIP_TEST_LEVEL=level_5` call `std::exit(EXIT_FAILURE)` and kill the whole run even though its parameters exist.

**Suggested fix:** Validate against the levels the store actually loaded (e.g. expose a `hasLevel(const std::string&)`/`levels()` accessor on `TestParameterStore` backed by `allParams`) and use that both for the fatal check and as the seed set for the exclusion-only branch of `collectLevels()`. If a hardcoded list must stay, have `parse_config.py` emit the supported-level list into the generated header so the C++ side cannot drift, and include the set of known levels in the abort message.

projects/hip-tests/catch/hipTestMain/hip_test_listener.cc
```diff
@@ -22,156 +22,230 @@ THE SOFTWARE.
...
+
+    /// Level used when neither the command line nor HIP_TEST_LEVEL specify one.
+    static constexpr const char* kDefaultLevel = "level_2";
```
mangupta, line 82: 🤖 AI-assisted pre-review (for @mangupta)

**[IMPORTANT]** Unconfigured runs escalate from level_0-derived parameters to level_2, and per-test-case level tags are no longer honored

With `testCaseStarting()` removed, a test's own `[level_N]` tag no longer selects its parameters; a single level is resolved once per process. Under ctest — where, per this file's own comment, each test is launched by name rather than by a `[level_X]` filter — the command-line source yields nothing, so any runner that does not export `HIP_TEST_LEVEL` now falls to `kDefaultLevel = "level_2"`. Tests written and tagged as `[level_0]` smoke tests therefore execute with comprehensive level_2 parameters (14 memory sizes up to 2 GB, 10 block sizes, 8 GB max memory), which is a large, silent increase in runtime and device-memory pressure and can OOM on small-memory runners that previously passed.

**Suggested fix:** Either default `kDefaultLevel` to the cheapest level (`level_0`) so an unconfigured run stays smoke-sized, or land the ctest wiring together with this change (set `HIP_TEST_LEVEL`, or pass `[level_N]`, in the test registration properties) so CI selects the level explicitly. Whichever is chosen, log the effective level and the reason at run start (already partly done) and call out the default change in the PR description so downstream runners can react.

mangupta (commented):

**[IMPORTANT]** New level-resolution grammar and fatal misconfiguration path have no test coverage and are not testable as written

This change replaces a one-line regex extraction with a multi-rule parser (comma-split OR filters, AND'd patterns, `~` negation, exclusion-only filters expanding to the full supported set, bare-tag mode only for the environment variable, strict `level_N` matching, highest-wins reduction) plus a new fatal `std::exit(EXIT_FAILURE)` path, and no test exercises any of it. The helpers are private statics inside the `.cc` with no header, so they cannot be unit-tested without a refactor, and a silent mis-resolution changes the parameters of every parameterized test in the suite rather than failing loudly.

**Suggested fix:** Move `collectLevels()`/`highestLevel()`/`isSupportedLevel()` into a small internal header (or an anonymous-namespace-free helper compiled into the test main) and add a Catch2 test case covering the four scenarios above, asserting the resolved level string for each input rather than only the final run parameters. Alternatively, if the fatal path cannot be unit-tested in-process, cover it with a ctest `WILL_FAIL`/expected-exit-code test that runs the binary with `HIP_TEST_LEVEL=level_9`.

projects/hip-tests/catch/hipTestMain/hip_test_listener.cc
amilanov-amd: Unnecessary overengineering, since levels will rarely be changed.

PR #8935, opened 2026-07-21
fix(clr): restore per-stream queue_id in graph profiling (7c23d18 reg…

anugodavar:
…ression)

PR #8400 (7c23d18e7b) removed KernelTimestamp.queue_index when fixing the kernelNamesOwner UAF, causing all graph kernel dispatches to be reported on the launch stream queue_id instead of the actual parallel stream they ran on. Profiling traces showed 1 stream for paths4/full4 graphs regardless of their topology.

Restore:
- KernelTimestamp struct with start/end/queue_index fields
- queue_index parameter in addTimestamps() (UINT32_MAX = unknown)
- signal->queue_index_ passed through addTimestamps() in rocvirtual.cpp
- Per-packet queue_id assignment in ReportActivity so each kernel lands on its actual internal parallel stream

Verified with GraphBench paths4/full4 — now shows 4 distinct stream queue_ids with correct event distribution.

## Submission Checklist


saleelk (approved):

anugodavar:
profiling related bug so no unit test

PR #8941, opened 2026-07-21
feat(rocr): Use global_prefetch_b8 for gfx125x entry-trampoline unclaused-VMEM workaround

dayatsin-amd:
## Summary

Align the gfx125x kernel-entry trampoline with the compiler's updated GFX1250 unclaused-VMEM workaround (llvm-project PR #210874), which now emits `global_prefetch_b8 v0, [s0, s1] scope:SCOPE_SE th:TH_LOAD_RT` + `v_nop` at each hardware kernel entry instead of `global_wb scope:SCOPE_CU` + `v_nop`.

- The kernel-entry trampoline stub (`BuildTrampolineGfx1250`) now leads with `global_prefetch_b8` (scope:SCOPE_SE) in place of `global_wb`.
- The prologue detection that skips a redundant trampoline (`kGfx1250UnclausedVmemPrologue`) now matches the `global_prefetch_b8` + `v_nop` encoding (`0xEE174000, 0x00040000, 0x00000000, 0x7E000000`), verified with `llvm-mc --mcpu=gfx1250 --show-encoding` (`TH_LOAD_RT` is the default TH=0; `scope:SCOPE_SE` sets bit `0x04`).
- Remove the experimental `LOADER_ENABLE_TRAMPOLINE_NO_WA` env var and its minimal no-workaround stub. `LOADER_ENABLE_TRAMPOLINE=1` is the only knob (disabled by default; for testing).

## JIRA ID

JIRA ID : LCOMPILER-2448

## Test plan

- [x] Build `libhsa-runtime64` against a gfx1250 TheRock toolchain built with llvm-project PR #210874
- [x] Compiler emits `global_prefetch_b8 v0, s[0:1] scope:SCOPE_SE` + `v_nop` at kernel entry
- [x] Kernel compiled WITH the workaround (patched clang) + `LOADER_ENABLE_TRAMPOLINE=1`: entry prologue is detected and the trampoline is skipped for that kernel; the app runs
- [x] Kernel compiled WITHOUT the workaround (stock compiler) + `LOADER_ENABLE_TRAMPOLINE=1`: trampoline is injected; the app runs
- [x] Verified on a gfx1250 system

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
@@ -119,7 +119,6 @@ public:
   const amd::options::PrefixOption* Substitute() const { return &substitute; }
 
   bool TrampolineEnabled() const { return trampoline_enabled_; }
-  bool TrampolineNoWaEnabled() const { return trampoline_no_wa_enabled_; }
 
   bool ParseOptions(const std::string& options);
   void Reset();
@@ -141,7 +140,6 @@ private:
   amd::options::PrefixOption substitute;
   amd::options::OptionParser option_parser;
   bool trampoline_enabled_ = false;
-  bool trampoline_no_wa_enabled_ = false;
 };
 
 LoaderOptions::LoaderOptions(std::ostream& error) :
@@ -162,17 +160,11 @@ LoaderOptions::LoaderOptions(std::ostream& error) :
   option_parser.AddOption(&dump_dir);
   option_parser.AddOption(&substitute);
 
-  // LOADER_ENABLE_TRAMPOLINE=1: enable gfx125x kernel-entry trampolines.
-  // LOADER_ENABLE_TRAMPOLINE_NO_WA=1: enable trampolines without the global_wb
-  // cache-writeback workaround (s_mov + s_set_pc only). Trampolines are disabled
-  // by default; these env vars are for testing only.
+  // LOADER_ENABLE_TRAMPOLINE=1: enable gfx125x kernel-entry trampolines (disabled
+  // by default; for testing only).
   const char* enable_trampoline = getenv("LOADER_ENABLE_TRAMPOLINE");
   if (enable_trampoline && std::strcmp(enable_trampoline, "1") == 0) {
     trampoline_enabled_ = true;
-  }
-  const char* enable_trampoline_no_wa = getenv("LOADER_ENABLE_TRAMPOLINE_NO_WA");
-  if (enable_trampoline_no_wa && std::strcmp(enable_trampoline_no_wa, "1") == 0) {
-    trampoline_no_wa_enabled_ = true;
   }
 }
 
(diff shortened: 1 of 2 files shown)


dayatsin-amd:
This is the compiler PR: https://github.com/llvm/llvm-project/pull/210874

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
```diff
@@ -232,23 +225,26 @@ static constexpr size_t kTrampolineStubStride =
...
+// th:TH_LOAD_RT followed by v_nop -- at every hardware kernel entry so the first
+// VMEM instruction is unclaused. It is exactly the sequence the entry trampoline
+// prepends, so when a kernel's entry already begins with it the trampoline would
+// only duplicate the workaround. gfx1250 encoding verified with
+// llvm-mc --mcpu=gfx1250 --show-encoding (TH_LOAD_RT is the default TH=0;
+// SCOPE_SE sets bit 0x04 in the third encoding byte).
 static constexpr uint32_t kGfx1250UnclausedVmemPrologue[4] = {
-    0xEE0B007C,  // global_wb <scope:SCOPE_CU>
-    0x00000000,  // :
+    0xEE174000,  // global_prefetch_b8 v0, [s0, s1] ...
```
cfreeamd, line 237: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** The old `global_wb scope:SCOPE_CU` was operand-free, so its encoding was fixed for every kernel. `global_prefetch_b8 v0, [s0, s1]` encodes a VADDR (v0) and SADDR (s0,s1), and `KernelEntryHasUnclausedVmemPrologue` does an exact 4-dword compare against these hardcoded operands. If the compiler ever emits its entry prefetch with different register operands, the compare silently returns `false` and a (redundant, but harmless) trampoline gets installed — weakening the skip optimization.

The correctness-critical direction is safe (an accidental exact 4-dword match is effectively impossible), so this is non-blocking. But is the compiler guaranteed to always emit `global_prefetch_b8 v0, [s0, s1]` with these exact operands at hardware kernel entry across all gfx125x kernels? If the operands can vary, a one-line comment noting the detection is best-effort would set the right expectation.

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
```diff
@@ -1637,9 +1617,9 @@ static bool KernelEntryHasUnclausedVmemPrologue(Context* context, Segment* code_
 
 hsa_status_t ExecutableImpl::InstallTrampolinesGfx125x(hsa_agent_t agent) {
   // Skip kernels whose entry already carries the compiler-inserted unclaused-VMEM
-  // workaround prologue (llvm PR #208467): a trampoline would only duplicate the
-  // global_wb/v_nop that is already there, so dispatch can go straight to the real
-  // entry. Kernels still needing the workaround keep their trampoline.
+  // workaround prologue: a trampoline would only duplicate the
```
cfreeamd, line 1620: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** The commit message cites the driving compiler change (llvm-project PR #210874) and the prior code comments cited the earlier workaround's PR (#208467), but the updated in-code comments dropped the PR reference. Consider adding `(llvm-project PR #210874)` here (or in the `kGfx1250UnclausedVmemPrologue` comment block) to preserve the traceability that was there before.

cfreeamd (commented):

Focused review of the two `loader/executable.cpp`/`.hpp` changes. The switch from `global_wb scope:SCOPE_CU` to `global_prefetch_b8 v0, [s0, s1] scope:SCOPE_SE` looks correct and internally consistent: generation (`BuildTrampolineGfx1250`) and detection (`KernelEntryHasUnclausedVmemPrologue`) both key off the single `kGfx1250UnclausedVmemPrologue` constant so they can't drift, the stub size is unchanged (3 dwords + v_nop), and the `LOADER_ENABLE_TRAMPOLINE_NO_WA` / `BuildTrampolineGfx1250NoWa` removal is complete (no dangling refs anywhere in the repo). Two non-blocking inline notes below. I could not independently run llvm-mc to verify the raw encoding bytes, so those are taken on your stated verification.

cfreeamd (approved):
Leave the nit and question to your discretion.

projects/rocr-runtime/runtime/hsa-runtime/loader/executable.cpp
dayatsin-amd: Yes, this is ok. Compiler is injecting exact same sequence

PR #8942, opened 2026-07-21
fix(rocprofiler-systems): Fix path resolution for pytest

sputhala-amd:
## Motivation

The pytest harness could silently resolve rocprofiler-systems binaries and `example/target` executables from the system PATH (e.g. `/opt/rocm/bin`) when they were missing from the build/install tree. That let a test exercise a different,
already-installed build instead of the one under test, producing misleading pass/fail results.

## Technical Details

- `_find_executable` - no longer falls back to `shutil.which`; it searches only the provided build/install `bin` paths.
- `get_target_executable` no longer falls back to PATH in the install and the build branch. Both branches now resolve only from the in-tree `examples/bin` directories.
- Both continue to honor the `ROCPROFSYS_BUILD_DIR` (build tree) and `ROCPROFSYS_INSTALL_DIR` (install prefix) overrides, which drive the searched directories.
- Clearer "not found" errors and the `get_target_executable` docstring now documents both search orders and states PATH is never consulted.
- Note: an example/target absent from the tree now skips (`FileNotFoundError` -> `pytest.skip`) instead of profiling a `PATH` copy.

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-systems/tests/pytest/rocprofsys/config.py
+++ b/projects/rocprofiler-systems/tests/pytest/rocprofsys/config.py
@@ -169,15 +169,19 @@ class RocprofsysConfig:
     def get_target_executable(self, name: str) -> Path:
         """Get path to a test target executable.
 
-        When is_installed is True, searches in the following order:
-        1. rocprofsys_build_dir/name (build directory layout)
-        2. rocprofsys_examples_dir/name/name (build directory layout)
-        3. PATH lookup
+        The executable is always resolved from the build/install tree; the
+        system PATH is never consulted, so the suite cannot silently pick up an
+        unrelated copy (e.g. one under /opt/rocm/bin).
 
-        When is_installed is False, searches in the following order:
+        When is_installed is True, searches in the following order:
         1. rocprofsys_examples_dir/name
         2. rocprofsys_bin_dir/name
-        3. PATH lookup
+
+        When is_installed is False, searches in the following order:
+        1. rocprofsys_examples_dir/examples/python/name (python scripts)
+        2. rocprofsys_examples_dir/name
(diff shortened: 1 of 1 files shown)


projects/rocprofiler-systems/tests/pytest/rocprofsys/config.py
```diff
@@ -229,34 +227,17 @@ def get_target_executable(self, name: str) -> Path:
...
-            if exe:
-                return Path(exe)
-
             raise FileNotFoundError(
                 f"Target executable '{name}' not found. Searched in:\n"
                 f"  - {self.rocprofsys_examples_dir}/examples/python/{name}\n"
                 f"  - {self.rocprofsys_examples_dir}/{name}\n"
                 f"  - {self.rocprofsys_examples_dir}/examples/code-coverage/{name}\n"
-                f"  - {self.rocprofsys_examples_dir}/examples/rccl/{name}\n"
-                f"  - {self.rocprofsys_examples_dir}/examples/{name}/{name}\n"
```
kcossett-amd, line 257: Was this removed because it was unused (being the `examples/<name>/<name>`)

sputhala-amd: Yes. I dont think we build anything in that path?

kcossett-amd: IIRC I had added that as a fallback of some sort, which is likely no longer needed now that examples binaries should match `{self.rocprofsys_examples_dir}/{name}` except for `code-coverage` and `python`.

kcossett-amd (approved):
Thank you.

PR #8945, opened 2026-07-21
feat(RCCL): upstream gfx1250 features (rccl/gfx1250 → develop)

mustafabar:
## Motivation

Upstream the staged gfx1250 (MI450) feature set from `rccl/gfx1250` into `develop`.

## Technical Details

Brings the gfx1250 work developed on `rccl/gfx1250` into `develop`, including DDA kernels + DDA fabric transport, LL/LL128 DDA collectives, CE collectives, larger unroll-factor support (default unroll 32 on gfx1250 with a fail-fast guard for ungenerated factors), and CUMEM/MNNVL auto-enable.

## Issue Tracking

JIRA ID: AICOMRCCL-1563, AICOMRCCL-1271, AICOMRCCL-1402, AICOMRCCL-1140, AICOMRCCL-1136, AICOMRCCL-1129, AICOMRCCL-1095

## Test Plan
1. We are conducting frequent sweeps of the application sensitive combos that impact MI450. Those include [graph/eager mode] x [symmetric, unregistered buffers] x [default/zero CU policy]
2. We are scoping out the automation of these runs through CI unit tests, while CI infra is made available to run UTs as part of the CI

## Submission Checklist


projects/rccl/src/include/algorithms/all_gather/all_gather_dda_fabric_ll128.h
```diff
@@ -0,0 +1,151 @@
...
+      if (lane < kDdaLL128DataElems) {
+        const size_t e = base + (size_t)lane;
+        const uint64_t v = (e < nWords) ? sw[e] : 0ull;
+        ddaLL128StoreWord(&dst[ln].w[lane], v);
+      }
+      // Unfenced: the payload-store instruction above precedes this flag store
+      // in warp program order; gfx1250 preserves the visibility order.
+      if (lane == kDdaLL128FlagElem) {
+        ddaLL128StoreWord(&dst[ln].w[kDdaLL128FlagElem], (uint64_t)flag);
+      }
```
garzaran, line 122: This does not seem correct. The LL128 protocol is based on the premise that we build a single line of 128B that also contains the flag and we store that line atomically, so that when the receiver side sees the flag, the data is already there. 

In this code, we do not have an atomic store. Rather, it relies on the fact that the same thread is performing two stores (to different addresses) and because the store of the flag comes after the store of the data, the expectation is that these will be delivered in order. In practice, the hardware cannot guarantee that and this can create issues. 

My recommendation is that we revert this PR and we add support for LL128 at a later time, using a similar approach as what the ring algo in RCCL is using

alex-breslow-amd: Repasting here what I suggested in the meeting when this finally lands:
1) Use the cooperative atomics builtins: https://clang.llvm.org/docs/AMDGPUBuiltinReference.html#builtin-amdgcn-cooperative-atomic-load-16x8b.  The final field that is a char* is a syncscope.  I believe the values for device and system are respectively "agent" and "" <-- empty.  These haven't been backported to MI350 as far as I know so their use has to be gated on gfx1250.
2) Use 16-byte per lane stores and loads (better performance).  You can move twice as much data per instruction.

projects/rccl/src/ce_coll.cc
```diff
@@ -122,8 +130,31 @@ ncclResult_t ncclCeInit(struct ncclComm* comm) {
...
+  {
+    size_t maxChunkBytes  = NCCL_CE_AR_MAX_MSG_BYTES / comm->nRanks;
+    size_t ceARTmpBufSize = alignUp((comm->nRanks + 1) * maxChunkBytes, 16);
+    uint8_t* ceARTmpBuf   = nullptr;
+    ncclWindow_vidmem* arWinDev;
+    ncclWindow_vidmem* arWinDevHost;
+
+    NCCLCHECKGOTO(ncclMemAlloc((void**)&ceARTmpBuf, ceARTmpBufSize), ret, fail_ar);
+    NCCLCHECKGOTO(ncclDevrWindowRegisterInGroup(comm, ceARTmpBuf, ceARTmpBufSize,
+                                                NCCL_WIN_COLL_SYMMETRIC, &arWinDev), ret, fail_ar);
```
garzaran: @AbandiGa , I wonder if we can re-use the buffers allocated for DDA for the copy engines, rather than allocating new bufers for the CE. I am afraid that the new optimizations are adding a lot of memory, and that can be problematic as app run at scale (or as the scaleup domain increases). 
I think it is fine to merge as it is, but we should create a ticket to review this.

projects/rccl/src/ce_coll.cc
```diff
@@ -752,6 +795,172 @@ ncclResult_t ncclCeGather(struct ncclComm* comm, struct ncclCeCollArgs* args, cu
...
+  // the bypass call site in collectives.cc.
+  if (recvWin == nullptr) {
+    NCCLCHECKGOTO(ncclDevrFindWindow(comm, recvbuff, &recvWin), ret, fail);
+  }
+  fastPath = (recvWin != nullptr) &&
+             (recvWin->winFlags & NCCL_WIN_COLL_SYMMETRIC);
+  collArgs.recvWin = recvWin;
+
+  NCCLCHECKGOTO(ncclCeInitBatchOpsParams(&batchOpsParams, comm->nRanks), ret, fail);
+
```
garzaran, line 868: @AbandiGa I assume this synchronization is needed when the allgather from the previous allreduce wrote to the temp buffer. In that case, you want to make sure that the data has been copied out to the user buffer before starting the next allreduce. Is that the correct understanding?

projects/rccl/src/rccl_wrap.cc
```diff
@@ -570,6 +571,36 @@ bool rcclUseAllGatherDirect(struct ncclComm* comm, size_t& msgSize) {
...
+  }
+
+  // Requires single-node symmetric memory support with CTA_POLICY_ZERO (CE mode).
+  if (!comm->symmetricSupport) return false;
+  if (comm->config.CTAPolicy != NCCL_CTA_POLICY_ZERO) return false;
+  if (comm->nNodes != 1) return false;
+
+  // count must divide evenly so every rank owns an equal shard.
+  if (count == 0 || count % (size_t)comm->nRanks != 0) return false;
+
```
garzaran, line 614: Are we planning to support the case where the input buffer is not a multiple of the number of ranks? I am not sure how often this occurs in the applications, but the size of the allreduce is independent of the #ranks.

projects/rccl/docs/api-reference/env-variables.rst
```diff
@@ -46,9 +46,11 @@ in the following table.
...
           release the physical GPU memory of a suspended communicator. See
           :ref:`suspend-resume` for the full prerequisites.
-      - | ``0``: Disabled (default).
-        | ``1``: Enabled.
-        | ``-2``: Auto-detect; enable when the platform supports VMM.
+      - | ``0``: Disabled.
+        | ``1``: Enabled on any architecture.
+        | ``-2``: Auto-detect (default); enable when the platform supports VMM.
+          Auto-detect is limited to gfx1250, the only architecture where the VMM
+          path is validated. Use ``1`` to force it on elsewhere.
```
garzaran, line 54: This is a bit outside of this PR. While it might be true that we need CUMEM to support  ncclCommSuspend and ncclCommResume, I find it misleading, as the main use case is not this one. 
How about something like this: 
Enables cuMem virtual memory management (VMM) for device-side memory allocations API, used for direct intra-node GPU-to-GPU communication to avoid additional copies to staging buffers. It is also required for "ncclCommSuspend" xxx

garzaran (commented):
I went through the changes for the copy engines. Allreduce and case where symmetric memory is not supported.

projects/rccl/src/include/comm.h
```diff
@@ -634,6 +642,14 @@ struct ncclComm {
   // True when ddaScratch is VMM (cuMem) backed (fabric path); selects the
   // matching deallocator at teardown.
   bool ddaScratchIsVmm;
+  // Device-resident per-block epoch cells for the LL-protocol DDA collectives,
+  uint32_t* ddaLLEpochDev;
+  int ddaLLEpochLen;
+  // Dedicated small epoch array for the LL AllReduce tier (kDdaFabricLLArMaxBlocks
+  // cells, seeded to a disjoint high flag namespace). Keeps that latency-bound
+  // tier's per-launch epoch reset cheap without the shared 256-wide array's cost.
+  uint32_t* ddaLLArEpochDev;
```
mustafabar: LL AR uses its own epoch counter (ddaLLArEpochDev) while AG/A2A/RS-LL share ddaLLEpochDev, but all four compute the same bank offset into the same peer scratch. Double buffering only separates launches within one flag sequence, so an AR can land at distance 1 on the bank a just-finished AG used and overwrite lines a slower peer is still polling which can cause that peer hangs.  kDdaLLArEpochSeed prevents a false flag match, not a clobber. Suggest folding AR onto ddaLLEpochDev as a hang safety measure. Same issue between LL and LL128: they share the counter but use different slot strides, so both LL banks fall inside LL128's bank 0. This is ok while RCCL_DDA_LL128=0, but needs a common bank stride before enabling. CC: @srirajpaul / @nileshnegi

isaki001:
I reviewed collectives.cc, paths.cc, and connect.cc. LGTM

projects/rccl/src/rccl_wrap.cc
AbandiGa: Yes, we can support this, but it will likely require some code changes. I can address it in a follow-up PR after both this PR and the CE AR pipeline (#8936) are merged.

projects/rccl/src/ce_coll.cc
AbandiGa: ncclCeInitBatchOpsParams is not synchronization, it's needed to allocate the srcs/dsts/sizes arrays used later for batched memcpy ops.

projects/rccl/src/include/comm.h
srirajpaul: It was using the same ddaLLEpochDev. @nileshnegi changed it as in the LL128 PR. not sure whether it was to get more performance in LL or something else. @nileshnegi do you remember why it was done?

projects/rccl/src/include/rccl_float8.h
nileshnegi, line 63: is this one not better suited for gfx1250?

garzaran (approved):
I just have one comment about a change in the documentation about an env variable. This can be addressed later. I will approve the PR. 
I reviewed the part that affects the copy engines.

projects/rccl/src/include/algorithms/all_reduce/all_reduce_dda_fabric_ll128.h
```diff
@@ -0,0 +1,134 @@
...
+                                        int selfRank, int nRanksRt,
+                                        uint32_t* __restrict__ epochDev, // per-block LL epoch cells (shared AG+AR)
+                                        int epochLen, // number of cells in epochDev
+                                        size_t slotStrideLines) { // per-call lines/slot (from host)
+
+  const int nRanks = NRANKS_CT ? NRANKS_CT : nRanksRt;
+  const size_t bytes = count * sizeof(T);
+  const size_t nWords = bytes >> 3; // 8B payload words
+  const size_t numLines = ddaLL128NumLines(nWords); // 128B lines this size
+  const size_t slot = slotStrideLines; // lines per slot (per-call)
```
srirajpaul, line 66: Is LL128 enabled?
the slot size is fixed constant in most places like the LL. Here it seems to be based on data size which looks like can cause issues if data is given is reverse size order. like if the first collective is on large data then it will occupy few MB of scratch. now if we call second collective with small data may be 1KB, it will go to the next bank, but here the bank is calculated based on data size (or slot size) which may be only few KBs and so it may overwrite the scratch of previous collective that spans few MB. if LL128 is not enabled then we can ignore it for now.

projects/rccl/src/dda_all_reduce_fabric_ll128.cu
```diff
@@ -0,0 +1,175 @@
...
+
+using meta::comms::kDdaLL128ArMaxBytes;
+using meta::comms::kDdaLL128DataElems;
+using meta::comms::kDdaLL128Lanes;
+using meta::comms::LLLine128;
+
+// Per-call slot stride in 128B lines for a message of `numLines` lines. The
+// stride matches the message exactly (compact layout) so small all-reduces keep
+// their scratch slots close together for good L2/TLB locality.
+static inline size_t ddaLL128ArSlotLines(size_t numLines) {
```
srirajpaul, line 40: added a question based on slot size. this look like an optimization but may cause overwriting of scratch between two collectives.

projects/rccl/src/include/rccl_float8.h
mustafabar: @KawtharShafie found out the ISA used with gfx942 are portable to gfx1250. There's support for pk8 fp16 arithmetic for gfx1250, but needs further testing and will have a follow-up PR

projects/rccl/src/rccl_wrap.cc
```diff
@@ -913,9 +972,16 @@ ncclResult_t commSetUnrollFactor(struct ncclComm* comm) {
...
     if (comm->unroll < NCCL_UNROLL_1 || comm->unroll >= NCCL_NUM_UNROLLS) {
-      WARN("Invalid RCCL_UNROLL_FACTOR %d specified. Valid values are 0 to 2 corresponding to unroll factors of 1, 2, "
-           "and 4 respectively.",
-           comm->unroll);
+      WARN("Invalid RCCL_UNROLL_FACTOR %d specified. Valid values are 0 to %d corresponding to unroll factors of 1, 2, "
+           "4, 8, 16, and 32 respectively.",
+           comm->unroll, NCCL_NUM_UNROLLS - 1);
+      return ncclInvalidArgument;
+    }
+    if (!ncclDevFuncUnrollGenerated[comm->unroll]) {
```
PedramAlizadeh, line 980: ncclDevFuncUnrollGenerated[] is build-wide (generate.py returns all_unrolls for any non-local-arch-only build), not per-arch. However, unroll 8/16/32 kernel tables are only compiled under defined(__gfx1250__). This check alone lets RCCL_UNROLL_FACTOR=3/4/5 pass on non-gfx1250 archs in a standard multi-arch build and then crash on a null function-table dispatch. Gating also on arch can resolve the issue:
```suggestion
    if (!ncclDevFuncUnrollGenerated[comm->unroll] ||
        ((comm->unroll == NCCL_UNROLL_8 || comm->unroll == NCCL_UNROLL_16 || comm->unroll == NCCL_UNROLL_32) &&
         !IsArchMatch(comm->archName, "gfx1250"))) {
```

projects/rccl/src/rccl_wrap.cc
```diff
@@ -926,8 +992,29 @@ ncclResult_t commSetUnrollFactor(struct ncclComm* comm) {
     else comm->unroll = NCCL_UNROLL_2;
   } else if (IsArchMatch(comm->archName, "gfx908") || ((IsArchMatch(comm->archName, "gfx942") && comm->cuCount > 80)))
     comm->unroll = NCCL_UNROLL_2;
+  else if (IsArchMatch(comm->archName, "gfx1250")) comm->unroll = NCCL_UNROLL_32;
   else comm->unroll = NCCL_UNROLL_4;
 
+  // Guard against a default that wasn't built for this arch (e.g. the generation
+  // matrix was narrowed). Fall back to any generated unroll rather than segfault.
+  if (!ncclDevFuncUnrollGenerated[comm->unroll]) {
```
PedramAlizadeh, line 1000: Same build-wide-vs-per-arch gap applies to the default-path fallback trigger:
```suggestion
  if (!ncclDevFuncUnrollGenerated[comm->unroll] ||
      ((comm->unroll == NCCL_UNROLL_8 || comm->unroll == NCCL_UNROLL_16 || comm->unroll == NCCL_UNROLL_32) &&
       !IsArchMatch(comm->archName, "gfx1250"))) {
```

projects/rccl/src/rccl_wrap.cc
```diff
@@ -926,8 +992,29 @@ ncclResult_t commSetUnrollFactor(struct ncclComm* comm) {
...
+    int fallback = -1;
+    for (int u = NCCL_NUM_UNROLLS - 1; u >= NCCL_UNROLL_1; u--) {
+      if (ncclDevFuncUnrollGenerated[u]) {
```
PedramAlizadeh, line 1003: The fallback-selection loop itself needs the same guard, otherwise it can fall back to an arch-invalid unroll factor too:
```suggestion
      if (ncclDevFuncUnrollGenerated[u] &&
          !((u == NCCL_UNROLL_8 || u == NCCL_UNROLL_16 || u == NCCL_UNROLL_32) &&
            !IsArchMatch(comm->archName, "gfx1250"))) {
```

projects/rccl/src/enqueue.cc
```diff
@@ -3619,17 +3672,82 @@ static ncclResult_t taskAppend(struct ncclComm* comm, struct ncclInfo* info) {
...
+      }
+      // enable ce allreduce for allreduce with size >= 4MB and <= 256MB
+      // Size gate for CE AllReduce: ceARTmpBuf is sized for at most
+      // NCCL_CE_AR_MAX_MSG_BYTES total bytes.
+      bool ceAllReduceFits = true;
+      ncclSymRegType_t winRegType;
+      NCCLCHECK(ncclGetSymRegType(sendWin, recvWin, &winRegType));
+      bool ceAvailable = !ceCapturing && ncclCeAvailable(comm, info->coll, info->op, info->datatype, winRegType);
+      if (info->coll == ncclFuncAllReduce) {
+        if (!ceArGraphAllowed || !rcclUseCeAllReduce(comm, info->count, info->datatype, info->op)) {
```
PedramAlizadeh, line 3717: ncclAllReduceWithBias (collectives.cc:687-693) tags its ncclInfo with coll = ncclFuncAllReduce and a non-null acc, then calls ncclEnqueueCheck/taskAppend directly. It never goes through ncclAllReduce_impl. This CE-eligibility check has no awareness of info->acc: rcclUseCeAllReduce(), ncclCeAllReduce(), and the ncclFuncAllReduce case in ncclLaunchCeColl() never read or apply a bias. Any bias AllReduce that also satisfies CE's normal eligibility (op=sum, single-node, symmetric support, 4 MiB–256 MiB (CE is default-on via RCCL_CE_ALLREDUCE=1)) will silently return a result with the bias term dropped, with no error. Excluding CE whenever a bias is present resolves the issue: 
```suggestion
        if (!ceArGraphAllowed || info->acc != nullptr || !rcclUseCeAllReduce(comm, info->count, info->datatype, info->op)) {
```

PedramAlizadeh (approved):

PR #8946, opened 2026-07-21
More ConSan/gfx1250 progress.

bjacob:


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/consan/consan_moi_inline_atomic.inc
```diff
@@ -464,53 +462,162 @@ void try_apply_inline_atomic_ordering_patch(std::span<const uint8_t> bytes,
       return std::nullopt;
     }
     words.push_back(*compare_succeeded);
+    words.push_back(
+        build_v_mov_b32_e32(static_cast<uint16_t>(*options.scratch_vgpr + 3u), kRdna4VccLo, arch));
+    words.push_back(
+        build_v_mov_b32_e32(static_cast<uint16_t>(*options.scratch_vgpr + 4u), kRdna4VccHi, arch));
+  }
+  const auto mbcnt_lo = instrumentation::build_v_mbcnt_lo_u32_b32(
```
kuhar, line 475: We should elect the publisher from the live EXEC mask here. `v_mbcnt` with `-1` gives absolute lane IDs, so if a divergent atomic runs with lane 0 inactive, no active lane compares equal to zero and `s_and_saveexec_b64` clears EXEC. The fetch-add and record stores are then skipped, silently dropping the sanitizer event. The new barrier and inline-shadow publishers already snapshot EXEC and use its lo/hi halves as the `mbcnt` mask; this path should do the same.

kuhar (commented):
A few comments

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/consan/consan_moi.cpp
```diff
@@ -540,12 +559,14 @@ inventory_consan_moi_auto_report(const ConSanResult &result, const ConSanOptions
...
     // kernel descriptor's fixed-LDS field (for example, target-managed tensor
-    // transfers). Such a descriptor cannot host a descriptor-grown local
-    // mirror, so provision the external full-aperture fallback just as we do
-    // for group-FLAT accesses.
+    // transfers or dynamically sized LDS). Such a descriptor cannot host a
+    // descriptor-grown local mirror. A nonzero field is not sufficient proof:
+    // some dynamic-LDS objects retain a one-byte placeholder even though a
+    // decoded access itself spans more bytes. Provision the external
+    // full-aperture fallback just as we do for group-FLAT accesses.
     const bool descriptor_opaque_lds = result.arch_name == "gfx1250" &&
```
kuhar, line 583: Report sizing should carry the kernel `has_dynamic_lds` signal into this decision, matching placement. A gfx1250 kernel with hidden dynamic LDS and a nonzero fixed allocation can take the external-shadow path while `inline_lds_bytes < selected_native_lds_extent` remains false, for example for a store at offset zero. The auto plan then sizes only the fixed prefix even though runtime dynamic offsets index the external table. Please provision the full gfx1250 workgroup aperture (320 KiB), or reject when that bound cannot fit; the current conservative fallback covers only 64 KiB.

kuhar (approved):

Issue #8939, opened 2026-07-21
[rocjitsu][rocr] Add support for rocjitsu code API for Hotswap DBT


atgutier:
### Background

Currently rocr hotswap for gfx1250 A0 machines relies on comgr backend. Add a HSA_HOTSWAP_ENABLE env var with two options: 1 (comgr) / 2 (rocjitsu) . Preserve existing behavior by keeping 1 (comgr) as default. 

### Feature Description

This feature will add support for rocjitsu's code library and DBT feature (< 5MB) as a hotswap backend which can be enabled by using the HSA_HOTSWAP_ENABLE=2 env variable.

### ROCm Component

rocr-runtime, rocjitsu

Groverkss:
rocr-runtime pr to get rocr to pick up rocjitsu hooks as a backend: https://github.com/ROCm/rocm-systems/pull/9262 
TheRock pr to ship the rocjitsu hooks shared library: https://github.com/ROCm/TheRock/pull/6872

PR #8947, opened 2026-07-21
[rocjitsu] gfx1250 B0 -> A0 translation

Groverkss:
## Motivation

Because A0 and B0 **share one ELF machine ID and the `gfx1250` target triple**, a compiler-produced code object is byte-identical for both steppings — the binary carries no signal that distinguishes them. To run B0-targeted code on A0 silicon, rocjitsu needs a *same-ISA* DBT path that rewrites the affected instructions in place while leaving everything else untouched.

This PR adds that path: a gfx1250 B0 → A0 legalization/expansion stage in the DBT pipeline, the supporting analyses it needs (VGPR-MSB bank tracking, indirect-branch/lane-stash recovery, relocation function tables), and the plumbing to select the silicon revision out of band (since it can't be inferred from the code object). Every workaround whose hardware behavior is not yet proven **fails closed** rather than silently emitting a form that could miscompile.

### Instruction legalization + semantic expansion (`code/dbt/{legalization,semantic}/gfx1250_b0_to_a0.*`)

A two-stage design: a **classifier** (`requires_errata_expansion`) decides which mnemonics need attention, and a **semantic rule table** performs the rewrite. Handled families:

- **Cluster loads** — kept as cluster loads but wrapped to execute with `M0 = 0` (save M0 → set M0=0 → load → restore M0). gfx1250 encodes `M0 = 125`, `NULL = 124` (inverse of
- **Fail-closed** (no proven lowering yet): bare low-precision K=128 FP8/BF8 WMMA, standalone 32x16 FP4 WMMA, and the deferred barrier/sleep family. These reject the whole kernel rather than emit a form the legalizer itself would refuse.

### gfx1250 VGPR-MSB analysis (`analysis/gfx1250_vgpr_msb.*`)

gfx1250 vector operands encode only the low 8 bits of a VGPR index; `s_set_vgpr_msb` supplies the high 2 bits per role. A dedicated analysis proves the per-role bank at each program point so expansions that must name a physical VGPR (or reject when the bank is unknown) are correct across control flow.

### Indirect-branch / lane-stash recovery (`analysis/indirect_branch_discovery.*`)

Recovers statically-built `setpc`/`swappc` targets (including the RCCL idiom that stashes a getpc-built PC into VGPR lanes via `writelane`/`readlane`) so relocation can rewrite the builder in place. Recovery is **block-local and fails closed**: state resets at every basic-block *leader* (branch targets included), so a stash reached on only one path — or jumped over — cannot seed a consumer it never dynamically reaches. Wide (≥64-bit) writes and MSB-bank changes invalidate stashed lanes; incomplete facts propagate monotonically and disqualify a consumer from a direct transfer window.

### Relocation function tables (`code/relocation_function_table.*`)

Support for rewriting getpc-relative address-builder ranges once final target offsets are known, enabling correct relocation of recovered indirect edges.

### Revision plumbing (config-driven, not ABI-driven)

- `simulation_config.fbs`: new `DbtSiliconRevision` enum + `guest_revision`/`host_revision` on `DbtGuestConfig` (appended, default `unspecified`, backward-compatible).
- `DbtGuestConfig` + HSA hook map the config revision into `BinaryTranslatorOptions`; a same-target gfx1250 load now translates when both revisions are set and fails closed otherwise.
- `rj_code.h` keeps `rj_code_dbt_options_t` at its original two fields; the C entry point rejects same-arch gfx1250.

### amdisa codegen

Parser/generator support for implied-literal encodings (extension DWORDs appended to a parent form), regenerated into the gfx1250 encoding tables.

## Issue Tracking

#8939

## Test Plan

Run CI tests.

- **Unit**: `translate_test.cpp` (per-family expansion + fail-closed coverage, operand-pinned assertions for M0/byte-select lane-mask/ADDTID formula), `liveness_test.cpp` (VGPR-MSB banks, CFG-local lane recovery incl. a branch-into-consumer diamond regression), `relocation_function_table_test.cpp`, `config_test.cpp` (revision round-trip), `rj_dbt_translate_smoke_test.cpp` (CLI flags).
- **HIP dispatch**: `relocation_function_table_hip_test.cpp` + device kernel.
- **Full Release serial suite: 1902/1902 pass** (1 intentionally skipped).
- **ASan/UBSan clean** — this PR also fixes the enum-load UB that failed the `asan-ubsan` CI job (removed with the C-API enum).

## Test Result

All passed

## Submission Checklist


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/binary_translator.cpp
```diff
@@ -1082,8 +1189,11 @@ TranslatedCodeObject BinaryTranslator::translate(const AmdGpuCodeObject &obj) {
       translation.configure_skipped_stub();
     }
 
-    std::string message = "skipped kernel " + kernel_label(*scope.translation) +
-                          " after translation error: " + std::move(failure.message);
+    std::string message =
```
Groverkss, line 1205: Note that in our hsa hooks we record which kernels we skipped and emit an error at runtime if they were dispatched. hotswap doesn't hook into running so we can't do that.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp
```diff
@@ -1470,6 +1603,145 @@ void add_recovered_successors(const std::vector<IndirectCallFixup> &recovered,
...
+  // every recorded lane, and any SGPR write invalidates a reconstructed half.
+  if (ctx.arch != ROCJITSU_CODE_ARCH_GFX1250)
+    return;
+
+  BlockState builders;
+  std::unordered_map<VectorLaneSlot, StashedPcHalf, VectorLaneSlotHash> slots;
+  std::array<std::optional<StashedPcHalf>, REGISTER_SET_MAX_SGPRS> read_halves;
+  std::set<uint16_t> active_read_halves;
+
+  for (size_t index = 0; index < ctx.insts.size(); ++index) {
```
kuhar, line 1751: We should not recover these VGPR-lane call targets with a linear scan. `slots` survives every `is_block_terminator()` and is cleared only by `s_endpgm`, so a lane value written on a non-taken branch or in a helper ending with `s_set_pc_i64` can be reused by later `v_readlane_b32` instructions. That can create an `IndirectCallFixup` for a value that never reaches the call at runtime. Please make the lane facts CFG-aware, or reject recovery when control flow intervenes.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/gfx1250_b0_to_a0.cpp
```diff
@@ -0,0 +1,1343 @@
...
+    return ExpandResult::failed(
+        "gfx1250 cluster load could not allocate a dead SGPR for M0 preservation");
+  }
+
+  std::vector<uint32_t> words;
+  words.reserve(6);
+  append_words(words,
+               gfx1250::build_sop1(gfx1250::kSMovB32Sop1,
+                                   {.ssrc0 = kSgprNull, .sdst = static_cast<uint8_t>(*scratch)}));
+  append_words(words, gfx1250::build_sop2(gfx1250::kSPackHhB32B16Sop2,
```
kuhar: Please make the A0-specific raw operand encoding explicit here. The generated,
revision-neutral gfx1250 tables describe 124 as NULL and 125 as M0, but this
errata sequence intentionally uses 124 as M0 for A0; the ADDTID lowering
does the same at line 955. Please use a name such as `kA0M0 = 124` for those
operands, keep `kSgprNull = 124` for the VGLOBAL SADDR sentinel, and add a short
comment citing the A0 behavior. The numeric alias may be correct across these
fields/steppings, but reusing a constant named NULL makes the workaround look
like a bug and invites an incorrect change to the intended/B0 encoding 125.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/semantic/gfx1250_b0_to_a0.cpp
```diff
@@ -0,0 +1,1343 @@
...
+  append_words(
+      words, gfx1250::build_vop3(gfx1250::kVMbcntHiU32B32Vop3, {.vdst = static_cast<uint8_t>(temp),
+                                                                .src0 = 193,
+                                                                .src1 = gfx1250_vgpr_src(temp)}));
+  append_words(words,
+               gfx1250::build_vop3(gfx1250::kVLshlrevB32Vop3, {.vdst = static_cast<uint8_t>(temp),
+                                                               .src0 = gfx1250_inline_u32(2),
+                                                               .src1 = gfx1250_vgpr_src(temp)}));
+  append_words(words,
+               gfx1250::build_vop3(gfx1250::kVAddNcU32Vop3, {.vdst = static_cast<uint8_t>(temp),
```
kuhar, line 817: Maybe document the A0 erratum formula that this lowering implements and pin it
in the test. The revision-neutral gfx1250 simulator models the documented ADDTID
formula as `lane * (((M0 >> 16) & 0x1ff) * 4) + offset`; that is not evidence
that this A0 workaround's `(M0 + lane*4) & 0xfffff` sequence is wrong. A
translate-then-execute test against that model would test the intended/B0
contract, not A0 silicon. A short reference comment plus exact emitted
operand/word assertions would make the stepping-specific contract explicit.

kuhar (commented):
A few comments

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/binary_translator.cpp
```diff
@@ -87,6 +89,32 @@ LegalizationLookupFn select_legalization(rj_code_arch_t guest, rj_code_arch_t ho
...
+/// `__builtin_unreachable()`. Current clang lowers those empty gfx1250 bodies to
+/// one 64-bit `s_setreg_imm32_b32 hwreg(HW_REG_WAVE_MODE, 25, 1), 1`
+/// instruction, followed by zero alignment padding. WAVE_MODE[25] is
+/// REPLAY_MODE; the instruction is ordinary prologue state rather than a CFG
+/// terminator, but the source function has no defined fallthrough execution.
+///
+/// Keep the match deliberately exact. Other reachable fallthrough into opaque
+/// text remains a hard error, including a longer block which merely ends with
+/// this instruction.
+[[nodiscard]] bool is_gfx1250_compiler_unreachable_stub(const BasicBlock &block,
```
Groverkss, line 105: I've thought about this earlier and Ideally we should just assume if something is undecodable, it cannot be reached, but safe for now. I'm thinking about kernels which could take different paths based on isa (if someone writes something like that)

emulation/rocjitsu/tests/analysis/liveness_test.cpp
```diff
@@ -664,6 +716,395 @@ TEST(CfgAnalysis, RecoversSignedDeltaTemplateConsumers) {
...
+      0xA980FE00u,
+      52u,
+      0u,                  // 0x04: s_add_nc_u64 ..., lit64(52).
+      set_dst_bank_one[0], // 0x10: s_set_vgpr_msb (DST bank 1, SRC0 bank 0).
+      0xD761002Cu,
+      0x02010000u, // 0x14: v_writelane_b32 v44, s0, 0 (physical v300 under DST bank 1).
+      0xD761002Cu,
+      0x02010201u, // 0x1c: v_writelane_b32 v44, s1, 1.
+      0xD7600000u,
+      0x0201012Cu, // 0x24: v_readlane_b32 s0, v44, 0 (physical v44 under SRC0 bank 0).
```
Groverkss, line 1046: nit: probably should use builders instead of direct words. I extracted these from some tests so probably left them like that.

emulation/rocjitsu/lib/rocjitsu/include/rocjitsu/code/rj_code.h
```diff
@@ -363,9 +363,36 @@ RJ_API_EXPORT const rj_code_inst_t *rj_code_inst_next(const rj_code_inst_t *inst
...
+/// @note This layout is NOT ABI-stable while DBT is in development: fields have
+/// been and may again be added. Callers must recompile against this header; a
+/// binary built against an older layout must not be mixed with a newer library.
 typedef struct rj_code_dbt_options_t {
   rj_code_arch_t guest_arch;
   rj_code_arch_t host_arch;
+  /// @brief Input silicon revision. Required when guest_arch is gfx1250; leave
+  /// UNSPECIFIED otherwise. A translation of a gfx1250 object with an unspecified
+  /// revision fails closed rather than guessing a direction.
+  rj_code_revision_t input_revision;
```
kuhar: We should preserve the existing C ABI instead of growing `rj_code_dbt_options_t` in place. A caller compiled against `develop` passes only the two architecture fields, while `rj_code_translate()` now unconditionally reads two more enum fields through the same exported symbol; upgrading `librocjitsu.so` therefore causes an out-of-bounds read before validation. A source comment requiring recompilation cannot protect already-built clients. Please add a size/versioned options entry point or a new revision-aware translation symbol, and keep the old entry point on its original two-field contract, rejecting same-architecture gfx1250 there.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/legalization/gfx1250_b0_to_a0.cpp
```diff
@@ -0,0 +1,223 @@
...
+  // placement constraints.
+  if (mnemonic == "s_clause")
+    return true;
+
+  for (std::string_view exact : kExactErrataMnemonics) {
+    if (mnemonic == exact)
+      return true;
+  }
+
+  // Every cluster-load form needs either demotion to a global load or an M0
```
kuhar: It would be good to update these classifier comments for the round-three behavior. Cluster loads are now kept and run with M0 cleared, while K=128 FP8/BF8 and standalone 32x16 FP4 now fail closed; lines 80-81 and 91-94 still describe demotion and K=64 splitting. The Doxygen above `expand_gfx1250_k128_wmma` in `semantic/gfx1250_b0_to_a0.cpp:966` likewise describes the removed two-instruction lowering.

kuhar (commented):
A few comments and one reply here: https://github.com/ROCm/rocm-systems/pull/8947#discussion_r3631638222

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp
kuhar: The round-three change still does not make this recovery block-local. Resetting after `is_block_terminator()` misses direct branch targets whose lexically preceding instruction falls through: a conditional branch can jump directly to the `v_readlane_b32`/`s_swappc_b64` block while the lexical fallthrough path contains the stash, and the scan will still reuse that skipped stash. Please reset the builders, lane slots, reconstructed halves, and VGPR-MSB state at every actual block leader, and add a branch-into-consumer/diamond regression.

kuhar (approved):

Groverkss:
I know it's my PR but Tony's changes LGTM, I left some comments earlier.

PR #8949, opened 2026-07-21
rocclr: flush image read to system scope for pinned host dst

sourabhuday:
##Motivation
Kernel dispatches use agent-scope release fences by default
(`AMD_OPT_FLUSH=1`). A blocking `clEnqueueReadImage` into a user host pointer
pins that buffer and fills it with a blit kernel (`copyImageToBuffer`), but the
agent-scope fence only flushes to device scope. So the read can return before
the data is actually visible to the CPU.

This breaks the common pattern of reading an image, modifying the buffer on the
CPU, and writing it back (with no flush in between): the write races with the
still-draining read and the image ends up keeping its old contents.

It shows up as `ocltst OCLCreateImage` failures 

## Fix
Promote the read's copy dispatch to system scope in `readImage`, matching what
`readBuffer` already does.

## Test Plan
- Read/modify/write repros now return correct data at all sizes (previously
  wrong for <=256x256).
- `ocltst OCLCreateImage` passes 5/5 (was 3/5).
- Forcing `AMD_OPT_FLUSH=0` upstream gives the same correct behavior.

## Issue Tracking

ROCM-26081

## Submission Checklist

### DIFF
--- a/projects/clr/rocclr/device/rocm/rocblit.cpp
+++ b/projects/clr/rocclr/device/rocm/rocblit.cpp
@@ -1935,6 +1935,10 @@ bool KernelBlitManager::readImage(device::Memory& srcMemory, void* dstHost,
     // Get device memory for this virtual device
     Memory* dstMemory = dev().getRocMemory(amdMemory);
 
+    // Make the copy system-scoped so the pinned host buffer is CPU-visible
+    // when the blocking read returns, like readBuffer does.
+    gpu().addSystemScope();
+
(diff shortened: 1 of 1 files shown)


AlexXAmd (changes requested):
this PR has performance hit. 
 
Only apply this for OCL? (And also fix the readBuffer too, if readBuffer has this issue.)
Check HIP need this or not? HIP might use another different kind of mechanism to make system scope valid.

PR #8952, opened 2026-07-21
doc(rocjitsu): style guide

### LINKED ISSUE #8951: doc(rocjitsu): styleguide

amd-arosa:
## Motivation

to write a styleguide

## Technical Details

adds a styleguide

## Issue Tracking

closes #8951

## Test Plan

read it

## Test Result

looks good

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/docs/style.md
+++ b/emulation/rocjitsu/docs/style.md
@@ -115,6 +115,18 @@ namespace rocjitsu::amdgpu {
 #include <cstdio>  // not <stdio.h>
 ```
 
+- Use `auto` only when the type is obvious from the immediate context (e.g. the
+  RHS spells out the type, such as a constructor call or a cast). Do not use
+  `auto` when the type can't be determined by looking at the line itself.
+
+```cpp
+auto x = foo(); // BAD, type not obvious
+
+auto x = std::make_unique<Foo>(); // OK, can tell from the RHS
+
+Foo foo = static_cast<Foo>(bar); // OK but could use auto instead to avoid repetitiveness
+```
+
 ## Logging
 
 - Never use `fprintf`, `printf`, or `std::cerr`. Use `Logger` from `util/log.h`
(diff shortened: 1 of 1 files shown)


emulation/rocjitsu/docs/style.md
```diff
@@ -0,0 +1,146 @@
...
+consistency over cleverness.
+
+## Types (`struct` vs `class`)
+
+- Use a `struct` **only** for plain-old-data (POD) types where every member is
+  public.
+- Use a `class` whenever you need constructors/destructors or any other non-POD
+  behavior.
+- Within a `class`, use a single `public:`, `protected:`, and `private:` section
+  (only the ones you need), and always in that order.
```
kuhar, line 13: Can we also use `typename` for template arguments instead of class? It helps with greppability.

emulation/rocjitsu/docs/style.md
```diff
@@ -0,0 +1,146 @@
...
+} // namespace rocjitsu::amdgpu
+```
+
+## Language, STL & Preprocessor
+
+- Always target C++20 and use the STL; prefer modern features such as concepts,
+  ranges, and `std::format`.
+- If you need performance and the STL is a poor fit, build custom data structures
+  in `util`.
+- Never use `#define`, C conventions, or macros unless absolutely necessary.
```
kuhar, line 113: this seems a bit too strong given how much code that interacts with kfd / ioctl and other interfaces we have

kuhar (approved):
I'd also add something about the use of `auto` -- the type should be obvious based on the immediate context, e.g.:
1. `auto x = foo(); // BAD, type not obvious`
2. `auto x = std::make_unique<Foo>(); // OK, can tell from the RHS`
3. `Foo foo = static_cast<Foo>(bar); // OK but could use auto instead to avoid repetitveneess`

accauble:
Thanks @amd-arosa! :-)

emulation/rocjitsu/docs/style.md
atgutier: I think it's fine as written, because in those cases it is necessary.

atgutier (approved):
This is a good starting point for a style guide. LGTM.

Just add the text on auto usage and mandating typename per Jakub's ask before you merge.

PR #5425, opened 2026-04-24
[RCCL] use 256 threads per block on gfx950

wenkaidu:
## Motivation

Reduce register spills on gfx950

## Technical Details

Reduce threads per block from 512 to 256 on gfx950

## Test Plan

Check register spills

## Test Result

Scratch Size [bytes per lane]

| kernels| 512 threads | 256 threads |
| -------- | -------- | -------- |
| unroll 1  | 520  | 352  |
| unroll 2  | 544  | 352  |

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #4795, opened 2026-04-07
[RCCL] Parallel build cleaned up v1

doru1004:
Parallel build with clean-up v1

Can be turned off by passing in --no-device-linker to install.sh

PR #8956, opened 2026-07-21
perf(rccl): reduce LL128 register spills and scratch on gfx942/gfx950

lfmeadow:
Ports the behavior-neutral subset of #8451 onto the `rccl/drop_2026-04` line.

### What
- **`-DNDEBUG` on the device Release build** (`DeviceLinker.cmake`): device asserts otherwise pin live registers around every `__assert_fail`, inflating the private (scratch) segment on register-heavy reduce kernels.
- **`reduceCopyPacksWithBias` live-range fix** (`common_kernel.h`): fuse each pack's load+reduce so only one `tmp` pack is live (instead of `tmp[Unroll]`) and defer the bias/accumulator load to its point of use; advance `accPtr` a full hunk to compensate. Arithmetic is unchanged.
- **Inline `runRing`/`runTree*`/`run` on gfx942/gfx950** (all_reduce, all_gather, broadcast, reduce, reduce_scatter, alltoall_pivot, sendrecv): drop the `noinline` attribute. Codegen-only.
- **`extract_metadata.cmake`**: single `llvm-objdump --offloading` call with env-based `ROCM_PATH` resolution.
- Minor supporting includes (`rccl_common.h`, `net.cc`) and a clarifying comment in `prims_ll128.h`.

### Not included
No fp8 packed arithmetic and no fp8 unit tests (both dropped from #8451). `ipc_init.cu`'s debug include is N/A on this base.

### Verification
Clean gfx950 build passes with the srock toolchain and inside the `rocm:7.0.2.1-22` container (Clang 20); `librccl.so` embeds a gfx950 bundle and the unit-test binaries link.

_Stacked on the 3 commits already between `rccl/drop_2026-04` and the requested base `f2af4e818e` (#5425 256 threads/block, channel-cap removal, #4795 parallel-build cleanup)._


lfmeadow:
This was an attempt to try to get close to April drop. Ended up with #8966 instead.

PR #8960, opened 2026-07-21
test(RCCL) Single proc mem reg + Symmetric memory tests 

PJAvinash:
## Motivation

This PR adds changes to testing framework to test symmetric memory for all collectives. 

## Technical Details
This PR adds end-to-end RCCL unit-test coverage for symmetric-memory
registration across eight collective families. It extends the common TestBed
infrastructure to:

- Allocate collective buffers with `ncclMemAlloc`.
- Collectively register and deregister symmetric windows.
- Correctly handle in-place, out-of-place, graph, and non-graph execution.
- Exercise irregular, unaligned, boundary, and large buffer sizes.
- Reliably run multi-GPU and rank-threaded tests in isolated child processes.

The new coverage exposed several runtime and test-framework problems involving
multi-GPU VMM access, registration cleanup after graph execution , RMA copy-engine cleanup, and FP8 result validation.

This PR fixes those issues as part of making the symmetric-memory tests reliable.

Compared with the `develop` merge base, the PR changes 29 RCCL files with
approximately 1,776 insertions and 451 deletions.

The existing unit-test framework primarily exercised ordinary HIP allocations
and optional `ncclCommRegister` buffer registration. It could not reliably test
the complete single-process symmetric-window lifecycle:

1. Allocate VMM-backed buffers with `ncclMemAlloc`.
2. Register the allocations using `ncclCommWindowRegister`.
3. Execute collectives with and without HIP graph capture.
4. Validate in-place rank-dependent buffer views.
5. Deregister windows before freeing their underlying allocations.
6. Repeat the sequence across many sizes, datatypes, ranks, and processes
   without leaking streams, graph objects, registrations, or child processes.

### Symmetric-memory collective coverage

Ten new GTests cover the following collective families:

- AllGather
- AllReduce
- AllToAll
- Broadcast
- Gather
- Reduce
- ReduceScatter
- Scatter

The tests explicitly enable:

- `NCCL_SINGLE_PROC_MEM_REG_ENABLE=1`
- `NCCL_CUMEM_ENABLE=1`
- `MEM_ALLOC_SYMMETRIC_WIN`

They disable the communicator pool so each test child inherits its intended
environment, and they restore all modified environment variables after the test.

The sweeps cover:

- In-place and out-of-place operation where supported.
- HIP graph and normal execution.
- Integer, FP64, FP32, FP16, BF16, FP8 E4M3, and FP8 E5M2 datatypes as applicable.
- Small and irregular counts such as `1`, `3`, `7`, `4314`, and `5003`.
- Boundary counts such as `1048575` and `1048576`.
- A large `10000001`-element AllReduce case.
- Non-zero roots for Gather and Reduce.

Managed memory is intentionally excluded because it is incompatible with the
symmetric VMM allocation mode used by these tests.

### TestBed memory and registration model

The TestBed now distinguishes three allocation modes:

- `MEM_ALLOC_HIP`
- `MEM_ALLOC_MANAGED`
- `MEM_ALLOC_SYMMETRIC_WIN`

Allocation and registration are separate operations. The parent first sends all
allocation commands, then sends registration commands to every participating
child before waiting for acknowledgements. This ordering allows collective
symmetric-window registration to make progress across ranks and processes.

The child registration path:

- Uses `ncclCommWindowRegister(..., NCCL_WIN_COLL_SYMMETRIC)` for symmetric memory.
- Uses separate input and output windows for out-of-place collectives.
- Registers the complete backing allocation for in-place collectives.
- Handles Scatter and ReduceScatter rank-dependent output offsets correctly.
- Uses separate input and output `ncclCommRegister` handles for legacy
  user-registered buffers.
- Validates deferred symmetric-window handles after `ncclGroupEnd`.
- Deregisters every window or registration handle before freeing memory.

### In-place buffer and allocation correctness

The framework now recomputes in-place input and output views whenever a sweep
changes the current element count while reusing the largest allocation.

This fixes rank-dependent offsets for:

- Scatter and ReduceScatter output views.
- Gather and AllGather input views.
- Other in-place collectives where input and output share the same base pointer.

Additional fixes ensure:

- AllGather frees the correct owning allocation.
- Custom reduction scalars are released even when the next test disables scalar mode.
- Device-side expected buffers use an allocation compatible with symmetric buffers.
- Data preparation and device-side validation select the correct GPU first.

### Child-process and IPC reliability

Fork-fresh workers now use `fork` followed by `execl("/proc/self/exe", "--child", ...)`.
This gives each worker a clean process image without inherited HIP/HSA runtime
state, avoiding hangs and crashes caused by using inherited accelerator state
after a bare fork.

The pipe protocol now:

- Reads and writes complete messages.
- Retries interrupted operations.
- Detects EOF and no-progress writes.
- Validates child command IDs.
- Validates inherited pipe descriptors after `exec`.
- Reports child signals and exit codes correctly.
- Closes descriptors in an order that avoids premature EOF and orphaned workers.

The isolated GPU probe also returns the HIP runtime version without initializing
HIP in the main test process. Symmetric-memory tests skip when the runtime is
older than `71260540`.

### Rank-threaded execution

NCCL group state is thread-local. Rank-threaded execution now keeps one rank on
one OpenMP thread for the complete group:

1. `ncclGroupStart`
2. Submit every collective for that rank
3. `ncclGroupEnd`

Serial execution continues to use one group spanning all selected local ranks.
The test fails explicitly if rank threading is requested without OpenMP support
or if OpenMP cannot provide the requested number of rank threads.

### HIP graph lifecycle

Graph state is tracked by actual local-rank and stream indices rather than only
the number of ranks selected for one execution.

The framework now:

- Resets graph-enabled state before capture.
- Validates graph dimensions before launch.
- Launches only successfully instantiated graph executables.
- Synchronizes streams and devices before destroying graph objects.
- Destroys graph executables before graph definitions.
- Clears handles to prevent repeated destruction.

#### Multi-GPU VMM allocation

`ncclMemAlloc` now retains each peer GPU's primary context before applying VMM
access permissions and releases it afterward. If context retention fails, the
partially created mapping, virtual address range, and allocation handle are
cleaned up before returning the error.

This makes cross-GPU VMM access setup reliable for single-process symmetric
allocations.

#### Registration and graph-callback ordering

Explicit communicator registration and deregistration now validate the
communicator and non-blockingly drain completed callbacks before accessing the
registration cache.

This prevents a virtual address reused by VMM after graph destruction from being
associated with stale registration metadata. Internal graph registration keeps
its existing throttled callback-drain behavior.

#### AllToAll user-buffer eligibility

The AllToAll path now:

- Warns when send and receive ranges overlap.
- Requires valid local input and output registrations before enabling the
  user-buffer path outside graph capture.
- Continues to allow the graph-required path while capture is active.

This prevents stale or graph-only registration-cache entries from incorrectly
enabling an unregistered-buffer path. Overlap remains a warning rather than a
hard error.

#### RMA copy-engine teardown

Communicator destruction now always invokes `ncclRmaCeFinalize`.

Finalization is safe when:

- The communicator pointer is null.
- RMA CE initialization did not complete.
- The context array is null.
- Individual context entries are null.

This releases pending initialization tasks, streams, events, signal windows,
device allocations, host allocations, and context storage across both complete
and partially initialized states.

### Test impact

- Existing tests retain `MEM_ALLOC_HIP` as the default allocation mode.
- Only the new tests opt into symmetric-window allocation.
- Symmetric plus managed-memory combinations are skipped.
- Tests using process-scoped environment changes no longer affect later tests.
- OpenMP-enabled test binaries link through `OpenMP::OpenMP_CXX` and receive an
  appropriate build RPATH for shared OpenMP libraries.
- The additional test matrix increases execution time, especially at high GPU counts.

### Runtime impact

This is not a test-only PR. It modifies five RCCL runtime files.

- Primary-context retain/release adds small allocation-time overhead per peer GPU.
- Callback polling adds a small non-blocking cost to explicit registration and
  deregistration.
- RMA CE cleanup runs during communicator destruction and is a low-cost no-op
  when RMA CE was never initialized.
- AllToAll applications with only graph-registration cache entries will not use
  the user-buffer path outside graph capture unless the buffers are also
  explicitly registered.

There are no public API or ABI changes and no new collective algorithms.

## Validation

The final symmetric-memory suite was run for every GPU count from 1 through 8.

- 10 tests from 8 test suites passed for each GPU count.
- 596 generated cases were executed for each GPU count.
- 4,768 generated cases were executed in total.
- No validation failures were reported.
- No child-process or pipe failures were reported.

FP8-specific regression coverage additionally confirmed:

- focused valid FP8 collective cases passed.
- Forced host and device reference-generation paths each passed 120 symmetric
  reduction cases.
- Standalone E4M3 and E5M2 tests accepted both modeled accumulation results and
  rejected missing-rank examples (`6` versus `8` and `4` versus `8`).
- `UT_DEVICE_DATA_FAULT=1` corrupted both references and produced the expected
  one-element mismatch per rank for both FP8 formats, confirming that the
  alternative reference cannot mask corruption.

No modeled valid result was rejected, and no injected FP8 corruption was
accepted in this matrix.

Kernel dispatch tracing also confirmed that:

- The ordinary FP8 AllReduce path dispatched a traditional `RING/SIMPLE` kernel.
- The symmetric FP8 AllReduce path dispatched `AllReduce_RSxLD_AGxST`.
- The observed FP8 output difference was therefore caused by different
  intermediate accumulation precision, not by the symmetric test accidentally
  using the same traditional kernel on the tested hardware.

### RCCL runtime

- **`projects/rccl/src/allocator.cc`**  
  Retains each peer GPU primary context while applying VMM access and fully
  rolls back the allocation on failure, improving multi-GPU `ncclMemAlloc`
  reliability.

- **`projects/rccl/src/enqueue.cc`**  
  Warns about overlapping AllToAll buffers and requires valid local
  registrations before enabling the non-graph user-buffer path.

- **`projects/rccl/src/init.cc`**  
  Invokes RMA CE finalization during communicator destruction so RMA resources
  are not retained across communicator lifetimes.

- **`projects/rccl/src/register/register.cc`**  
  Drains completed callbacks before explicit register and deregister operations,
  preventing stale graph registration state after VMM address reuse.

- **`projects/rccl/src/rma/rma_ce.cc`**  
  Makes RMA CE finalization null-safe and capable of cleaning partially
  initialized state.

### Collective tests

- **`projects/rccl/test/AllGatherTests.cpp`**  
  Adds symmetric AllGather coverage across datatypes, alignment boundaries,
  graph modes, and in-place/out-of-place operation.

- **`projects/rccl/test/AllReduceTests.cpp`**  
  Adds symmetric AllReduce coverage from FP64 through FP8, including explicitly
  unaligned and large element counts.

- **`projects/rccl/test/AllToAllTests.cpp`**  
  Adds separate graph and non-graph symmetric AllToAll sweeps using valid
  out-of-place buffers and irregular sizes.

- **`projects/rccl/test/BroadcastTests.cpp`**  
  Adds symmetric Broadcast coverage for multiple datatypes, graph modes, and
  in-place/out-of-place operation.

- **`projects/rccl/test/GatherTests.cpp`**  
  Adds symmetric Gather coverage with a non-zero root and graph/in-place combinations.

- **`projects/rccl/test/ReduceScatterTests.cpp`**  
  Adds broad symmetric ReduceScatter coverage including FP8, graph execution,
  in-place offsets, and size boundaries.

- **`projects/rccl/test/ReduceTests.cpp`**  
  Adds rooted symmetric Reduce coverage across floating-point types, including
  both FP8 formats.

- **`projects/rccl/test/ScatterTests.cpp`**  
  Adds separate graph and non-graph symmetric Scatter tests using
  registration-friendly sizes.

- **`projects/rccl/test/CMakeLists.txt`**  
  Links tests through the imported OpenMP target and adds shared OpenMP library
  directories to the build RPATH.

### Test framework

- **`projects/rccl/test/common/CollectiveArgs.hpp`**  
  Defines allocation modes, separate input/output registration and symmetric
  window handles, and CPU/GPU alternative FP8 expected buffers.

- **`projects/rccl/test/common/CollectiveArgs.cpp`**  
  Recomputes in-place buffer views for each size, corrects allocation ownership,
  scalar cleanup, symmetric expected-data allocation, and GPU selection. It
  allocates, selects, validates, and frees the alternative FP8 references.

- **`projects/rccl/test/common/DeviceDataOps.hpp`**  
  Generates stepwise-FP8 and FP32-accumulation references on the GPU and requires
  an exact match against either modeled result.

- **`projects/rccl/test/common/EnvVars.hpp`**  
  Exposes the HIP runtime version collected by the isolated GPU probe.

- **`projects/rccl/test/common/EnvVars.cpp`**  
  Adds HIP runtime-version serialization to the fork-and-exec GPU probe.

- **`projects/rccl/test/common/PipeUtils.hpp`**  
  Adds complete, interruption-safe pipe read and write helpers.

- **`projects/rccl/test/common/PrepDataFuncs.cpp`**  
  Selects and restores the correct GPU around rank-specific data preparation,
  builds both FP8 references on host and device paths, and corrupts both
  references for the device-data negative control.

- **`projects/rccl/test/common/PtrUnion.cpp`**  
  Routes host and device FP8 validation through the dual-reference oracle and
  reports both expected values in mismatch diagnostics.

- **`projects/rccl/test/common/PtrUnion.hpp`**  
  Extends validation and expected-data APIs with the optional alternative FP8
  reference.

- **`projects/rccl/test/common/ScopedEnvVar.hpp`**  
  Adds scoped environment-variable updates that restore the original process state.

- **`projects/rccl/test/common/TestBed.hpp`**  
  Extends TestBed APIs with allocation-mode selection and split allocation and
  registration operations while preserving HIP allocation defaults.

- **`projects/rccl/test/common/TestBed.cpp`**  
  Implements parent-side symmetric registration orchestration, fork-and-exec
  workers, robust IPC acknowledgements, and deterministic child teardown.

- **`projects/rccl/test/common/TestBedChild.hpp`**  
  Adds the registration command, allocation-mode state, and internal
  deregistration/deallocation interfaces.

- **`projects/rccl/test/common/TestBedChild.cpp`**  
  Implements symmetric registration and deregistration and hardens communicator,
  OpenMP group, graph, stream, timeout, validation, and child-command lifecycles.

- **`projects/rccl/test/common/main.cpp`**  
  Adds the early post-exec child-worker entry point and validates inherited pipe
  descriptors before entering the command loop.

- **`projects/rccl/test/StandaloneTests.cpp`**  
  Adds focused host and GPU tests proving that both modeled FP8 outcomes are
  accepted, missing-rank values are rejected, and corruption must affect both
  references before a valid alternative is no longer available.

- **`projects/rccl/test/host/fakes/nccl_stubs.cc`**  
  Adds the RMA CE finalization stub required by host-only communicator cleanup tests.

## Known limitation

The symmetric tests currently confirm that `ncclCommWindowRegister` returned a
non-null handle. A non-null handle does not by itself prove that the symmetric
backend is available.

When `comm->symmetricSupport == 0`, RCCL can route window registration through
the non-symmetric IPC fallback and still return a valid handle. The collective
can then run through the ordinary ring path, causing the test to report success
without exercising a symmetric kernel. But this is also intentional, as the production use case has a fallback mechanism.

To guarantee the intended coverage on every test system, the test should query:

```cpp
ncclCommQueryProperties(comm, &properties);
```

and can skip the symmetric-memory test when:

```cpp
properties.deviceApiSupport == 0
```

On the hardware used for the validation above, kernel tracing confirmed that
the symmetric AllReduce kernel was dispatched.

## Issue Tracking
AICOMRCCL-1169
AICOMRCCL-1533

## Test Plan
- We need HIP_VERSION > 71260540 for cuMem

## Test Result
- All the collectives pass


projects/rccl/src/enqueue.cc
```diff
@@ -3943,12 +3966,22 @@ static ncclResult_t taskAppend(struct ncclComm* comm, struct ncclInfo* info) {
...
           NCCLCHECK(ncclRegFind(comm, info->recvbuff, comm->nRanks * info->count * ncclTypeSize(info->datatype),
                                 &recvReg));
-          allowUB = captured || (sendReg != NULL && recvReg != NULL);
+          if (sendReg) NCCLCHECK(ncclRegLocalIsValid(sendReg, &sendLocalValid));
+          if (recvReg) NCCLCHECK(ncclRegLocalIsValid(recvReg, &recvLocalValid));   
+
+          /**
+           * To do : using allowUB = (captured || (sendLocalValid && recvLocalValid)); for alltoall, 
+           * with symmetric memory, in Graphmode, with sequence of buffer sizes where max is not aligned 
+           * to 128 bytes, results in data validation errors. setting this to false, until it is resolved.
```
alex-breslow-amd, line 3982: What's the timeline on getting this added?  Will this be a follow-on PR or part of this one?

projects/rccl/src/enqueue.cc
```diff
@@ -3823,6 +3837,15 @@ static ncclResult_t taskAppend(struct ncclComm* comm, struct ncclInfo* info) {
       ncclDevrFindWindow(comm, info->recvbuff, &recvWin);
 
       bool hasSysmemSegment = ncclDevrWindowHasSysmemSegment(sendWin) || ncclDevrWindowHasSysmemSegment(recvWin);
+      size_t totalBytes = comm->nRanks * info->count * ncclTypeSize(info->datatype);
+      // Check if sendbuff and recvbuff overlap
+      bool isOverlapping = rcclBuffersOverlap(info->sendbuff, info->recvbuff, totalBytes);
+      if (isOverlapping && info->coll == ncclFuncAlltoAll) {
+        WARN("AllToAll: Overlapping/In-Place buffers detected [%p, %p) vs [%p, %p). "
```
alex-breslow-amd, line 3844: Isn't this going to cause rccl-tests to have a lot of log-spew for AllToAll because it does benchmark in-place even though it's invalid?

projects/rccl/src/enqueue.cc
```diff
@@ -1271,6 +1271,9 @@ static ncclResult_t addP2pToPlan(struct ncclComm* comm, struct ncclKernelPlan* p
   bool netRegistered[2] = {false, false};
   bool ipcRegistered[2] = {false, false};
 
+  // Check Pointer & Length Alignment (Direct DMA requires at least 16-byte alignment)
```
atulkulk, line 1274: Why are we restricting to 16-byte alignment? I think this assumption is incorrect. AFAIK, we should not have this gating.

projects/rccl/src/enqueue.cc
```diff
@@ -1338,7 +1341,7 @@ static ncclResult_t addP2pToPlan(struct ncclComm* comm, struct ncclKernelPlan* p
       chunkSize[dir] = (chunkSize[dir] / comm->ll128DataElems) * comm->ll128LineElems;
     }
 
-    if (p2pTasks[dir] && p2pTasks[dir]->allowUB) {
+    if (p2pTasks[dir] && p2pTasks[dir]->allowUB && isAligned ) {
```
atulkulk, line 1344: `isAligned` will gate multi-process intra-node and multi-node cases. This will be a regression.

projects/rccl/src/enqueue.cc
```diff
@@ -1347,31 +1350,31 @@ static ncclResult_t addP2pToPlan(struct ncclComm* comm, struct ncclKernelPlan* p
...
         struct ncclConnector* conn =
           dir ? &channelPeers[peerRank]->send[connIndex[dir]] : &channelPeers[peerRank]->recv[connIndex[dir]];
         void* regAddr = NULL;
         if (conn->conn.flags & (NCCL_P2P_WRITE | NCCL_P2P_READ)) {
           // We require users registering buffers on both sides
           NCCLCHECKGOTO(ncclRegisterP2pIpcBuffer(comm, addrs[dir], bytes[dir], peerRank, &regFlag, &regAddr,
-                                                 &plan->cleanupQueue),
-                        ret, cleanup);
+                                               &plan->cleanupQueue),ret, cleanup);
           if (regFlag) {
```
atulkulk, line 1378: nit: Are these newly standardized `clang-format` changes? Probably, better to have a separate PR for it if the changes are not relevant.

projects/rccl/src/enqueue.cc
```diff
@@ -3943,12 +3966,22 @@ static ncclResult_t taskAppend(struct ncclComm* comm, struct ncclInfo* info) {
...
-          allowUB = captured || (sendReg != NULL && recvReg != NULL);
+          if (sendReg) NCCLCHECK(ncclRegLocalIsValid(sendReg, &sendLocalValid));
+          if (recvReg) NCCLCHECK(ncclRegLocalIsValid(recvReg, &recvLocalValid));   
+
+          /**
+           * To do : using allowUB = (captured || (sendLocalValid && recvLocalValid)); for alltoall, 
+           * with symmetric memory, in Graphmode, with sequence of buffer sizes where max is not aligned 
+           * to 128 bytes, results in data validation errors. setting this to false, until it is resolved.
+           */
+          allowUB = false;
```
atulkulk, line 3984: Regression! For normal multi-rank/multi-node cases. Also, user should be warned that UB is turned off if they turned it on explicitly via env var.

atulkulk (changes requested):
In general, do we have a use-case/requirement to support UBR within a single process? Why not just use staging copies for it?

projects/rccl/test/AllGatherTests.cpp
```diff
@@ -122,6 +123,36 @@ namespace RcclUnitTesting
...
+    std::vector<ncclFunc_t>     const funcTypes       = {ncclCollAllGather};
+    std::vector<ncclDataType_t> const dataTypes       = {ncclUint64,ncclUint32,ncclBfloat16,ncclUint8};
+    std::vector<ncclRedOp_t>    const redOps          = {ncclSum};
+    std::vector<int>            const roots           = {0};
+    std::vector<int>            const numElements     = {1,3,7,4314,5003,1048575,1048576};
+    std::vector<bool>           const inPlaceList     = {true,false};
+    std::vector<bool>           const managedMemList  = {false};
+    std::vector<bool>           const useHipGraphList = {true,false};
+
+    setenv("NCCL_SINGLE_PROC_MEM_REG_ENABLE", "1", 1);
```
atulkulk, line 146: Note: If you set this here, it will be applied to the entire `rccl-UnitTests` binary. Use ProcessIsolated framework if you need to use env vars for the tests.

projects/rccl/src/enqueue.cc
PJAvinash: Performance regression or correctness ?

projects/rccl/test/AllGatherTests.cpp
PJAvinash: UT_COMM_POOL=0 before TestBed ( parent ) + ScopedEnvVar resolves this issue. 

When UT_COMM_POOL=0, we take fork() + execl() path for child creation which inherits ScopedEnvVar values for the new child process.

projects/rccl/src/enqueue.cc
PJAvinash: This issue is solved with small bug fix, I'll remove the code comment in a cleanup commit.

projects/rccl/src/enqueue.cc
PJAvinash: This issue is fixed now. updating the comment.

projects/rccl/src/enqueue.cc
PJAvinash: Verified with test cases, It is not needed. removing.

projects/rccl/test/AllToAllTests.cpp
```diff
@@ -112,4 +113,66 @@ namespace RcclUnitTesting
...
+                   << hipRuntimeVer << ") is lower than 71260540";
+    }
+    TestBed testBed;
+
+    // Configuration
+    std::vector<ncclFunc_t>     const funcTypes       = {ncclCollAlltoAll};
+    std::vector<ncclDataType_t> const dataTypes       = {ncclUint8,ncclBfloat16,ncclUint32,ncclUint64};
+    std::vector<ncclRedOp_t>    const redOps          = {ncclSum};
+    std::vector<int>            const roots           = {0};
+    std::vector<int>            const numElements     = {1,4096,5003,5012};
```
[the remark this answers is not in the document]
PJAvinash: The issue is resolved. Goal of this test is to detect the failure path. we are not disabling it.

projects/rccl/src/enqueue.cc
PJAvinash: Removed

PJAvinash:
@argus re-review

projects/rccl/test/common/TestBedChild.cpp
```diff
@@ -1053,4 +1343,200 @@ namespace RcclUnitTesting
...
+    CHILD_NCCL_CALL(ncclGroupEnd(), "ncclGroupEnd RegisterMem");
+    // Ensure GPU memory mapping / TLB invalidations complete on selected ranks.
+    for (int localRank : localRanks)
+    {
+      CHECK_HIP(hipSetDevice(this->deviceIds[localRank]));
+      CHECK_HIP(hipDeviceSynchronize());
+    }
+
+    ErrCode status = TEST_SUCCESS;
+    if (this->memAllocType == MEM_ALLOC_SYMMETRIC_WIN)
```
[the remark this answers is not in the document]
PJAvinash: Good idea, but that requires a child command that calls ncclCommQueryProperties() for every active communicator and returns whether all report deviceApiSupport, and in In RunSimpleSweep(), immediately after InitComms() if MEM_ALLOC_SYMMETRIC_WIN and support is false, destroy communicators and GTEST_SKIP(). This would correctly indicate the tests are skipped but the testcase fails to mimic the IPC fallback path in actual case. Hence having this block of code as is not a problem!?

PJAvinash:
@argus re-review full

PJAvinash:
RCCL version : 2.30.7-Unknown 
HIP version  : 7.16.26370-0000000
ROCm version : 10.2.0.0-9999-1091c911
Hostname     : banff-ccs-aus-p22-05.cs-aus.dcgpu
Librccl path : /NCCL_SYNC_VAL/temp/rocm-systems/projects/rccl/build/release/librccl.so.1
[       OK ] Scatter.SingleProcMemRegDisabledSmoke (25735 ms)
[----------] 3 tests from Scatter (90156 ms total)

[----------] Global test environment tear-down
[==========] 18 tests from 8 test suites ran. (558378 ms total)
[  PASSED  ] 18 tests.
[ INFO     ] Total executed cases: 5008
[ TIMING   ] TEST SUITE          : TEST NAME           :       TIME ms (STATUS)
[ TIMING   ] AllGather           : SingleProcMemReg    :      49.25 sec (PASS)
[ TIMING   ] AllGather           : SingleProcMemRegDisabledSmoke:      23.45 sec (PASS)
[ TIMING   ] AllGather           : TOTAL               :      72.70 sec (PASS)
[ TIMING   ] AllReduce           : SingleProcMemReg    :      48.92 sec (PASS)
[ TIMING   ] AllReduce           : SingleProcMemRegDisabledSmoke:      24.22 sec (PASS)
[ TIMING   ] AllReduce           : TOTAL               :      73.14 sec (PASS)
[ TIMING   ] AlltoAll            : SingleProcMemReg    :      28.97 sec (PASS)
[ TIMING   ] AlltoAll            : SingleProcMemRegGraph:      27.55 sec (PASS)
[ TIMING   ] AlltoAll            : SingleProcMemRegDisabledSmoke:      24.52 sec (PASS)
[ TIMING   ] AlltoAll            : TOTAL               :      81.05 sec (PASS)
[ TIMING   ] Broadcast           : SingleProcMemReg    :      30.20 sec (PASS)
[ TIMING   ] Broadcast           : SingleProcMemRegDisabledSmoke:      22.90 sec (PASS)
[ TIMING   ] Broadcast           : TOTAL               :      53.10 sec (PASS)
[ TIMING   ] Gather              : SingleProcMemReg    :      35.73 sec (PASS)
[ TIMING   ] Gather              : SingleProcMemRegDisabledSmoke:      25.97 sec (PASS)
[ TIMING   ] Gather              : TOTAL               :      61.70 sec (PASS)
[ TIMING   ] ReduceScatter       : SingleProcMemReg    :      44.43 sec (PASS)
[ TIMING   ] ReduceScatter       : SingleProcMemRegDisabledSmoke:      22.58 sec (PASS)
[ TIMING   ] ReduceScatter       : TOTAL               :      67.01 sec (PASS)
[ TIMING   ] Reduce              : SingleProcMemReg    :      36.94 sec (PASS)
[ TIMING   ] Reduce              : SingleProcMemRegDisabledSmoke:      22.58 sec (PASS)
[ TIMING   ] Reduce              : TOTAL               :      59.52 sec (PASS)
[ TIMING   ] Scatter             : SingleProcMemReg    :      32.94 sec (PASS)
[ TIMING   ] Scatter             : SingleProcMemRegGraph:      31.48 sec (PASS)
[ TIMING   ] Scatter             : SingleProcMemRegDisabledSmoke:      25.73 sec (PASS)
[ TIMING   ] Scatter             : TOTAL               :      90.16 sec (PASS)
[ TIMING   ] Total time:       9.31 minutes 

Ran the test sweep on gfx942 for 1 to 4 GPUs. It runs fine, Investigating https://github.com/ROCm/rocm-systems/actions/runs/35139178319/job/104967718174?pr=8960.

PJAvinash:
> RCCL version : 2.30.7-Unknown HIP version : 7.16.26370-0000000 ROCm version : 10.2.0.0-9999-1091c911 Hostname : banff-ccs-aus-p22-05.cs-aus.dcgpu Librccl path : /NCCL_SYNC_VAL/temp/rocm-systems/projects/rccl/build/release/librccl.so.1 [ OK ] Scatter.SingleProcMemRegDisabledSmoke (25735 ms) [----------] 3 tests from Scatter (90156 ms total)
> 
> [----------] Global test environment tear-down [==========] 18 tests from 8 test suites ran. (558378 ms total) [ PASSED ] 18 tests. [ INFO ] Total executed cases: 5008 [ TIMING ] TEST SUITE : TEST NAME : TIME ms (STATUS) [ TIMING ] AllGather : SingleProcMemReg : 49.25 sec (PASS) [ TIMING ] AllGather : SingleProcMemRegDisabledSmoke: 23.45 sec (PASS) [ TIMING ] AllGather : TOTAL : 72.70 sec (PASS) [ TIMING ] AllReduce : SingleProcMemReg : 48.92 sec (PASS) [ TIMING ] AllReduce : SingleProcMemRegDisabledSmoke: 24.22 sec (PASS) [ TIMING ] AllReduce : TOTAL : 73.14 sec (PASS) [ TIMING ] AlltoAll : SingleProcMemReg : 28.97 sec (PASS) [ TIMING ] AlltoAll : SingleProcMemRegGraph: 27.55 sec (PASS) [ TIMING ] AlltoAll : SingleProcMemRegDisabledSmoke: 24.52 sec (PASS) [ TIMING ] AlltoAll : TOTAL : 81.05 sec (PASS) [ TIMING ] Broadcast : SingleProcMemReg : 30.20 sec (PASS) [ TIMING ] Broadcast : SingleProcMemRegDisabledSmoke: 22.90 sec (PASS) [ TIMING ] Broadcast : TOTAL : 53.10 sec (PASS) [ TIMING ] Gather : SingleProcMemReg : 35.73 sec (PASS) [ TIMING ] Gather : SingleProcMemRegDisabledSmoke: 25.97 sec (PASS) [ TIMING ] Gather : TOTAL : 61.70 sec (PASS) [ TIMING ] ReduceScatter : SingleProcMemReg : 44.43 sec (PASS) [ TIMING ] ReduceScatter : SingleProcMemRegDisabledSmoke: 22.58 sec (PASS) [ TIMING ] ReduceScatter : TOTAL : 67.01 sec (PASS) [ TIMING ] Reduce : SingleProcMemReg : 36.94 sec (PASS) [ TIMING ] Reduce : SingleProcMemRegDisabledSmoke: 22.58 sec (PASS) [ TIMING ] Reduce : TOTAL : 59.52 sec (PASS) [ TIMING ] Scatter : SingleProcMemReg : 32.94 sec (PASS) [ TIMING ] Scatter : SingleProcMemRegGraph: 31.48 sec (PASS) [ TIMING ] Scatter : SingleProcMemRegDisabledSmoke: 25.73 sec (PASS) [ TIMING ] Scatter : TOTAL : 90.16 sec (PASS) [ TIMING ] Total time: 9.31 minutes
> 
> Ran the test sweep on gfx942 for 1 to 4 GPUs. It runs fine, Investigating https://github.com/ROCm/rocm-systems/actions/runs/35139178319/job/104967718174?pr=8960.

Issue resolved in [3d4b84d](https://github.com/ROCm/rocm-systems/pull/8960/commits/3d4b84dcff0fd80c8b7d93feb9c59973c6928da8)

projects/rccl/src/enqueue.cc
PJAvinash: Undoing these changes.

projects/rccl/src/enqueue/enqueue.cc
```diff
@@ -3823,6 +3837,15 @@ static ncclResult_t taskAppend(struct ncclComm* comm, struct ncclInfo* info) {
       ncclDevrFindWindow(comm, info->recvbuff, &recvWin);
 
       bool hasSysmemSegment = ncclDevrWindowHasSysmemSegment(sendWin) || ncclDevrWindowHasSysmemSegment(recvWin);
+      size_t totalBytes = comm->nRanks * info->count * ncclTypeSize(info->datatype);
+      // Check if sendbuff and recvbuff overlap
+      bool isOverlapping = rcclBuffersOverlap(info->sendbuff, info->recvbuff, totalBytes);
+      if (isOverlapping && info->coll == ncclFuncAlltoAll) {
+        WARN("AllToAll: Overlapping/In-Place buffers detected [%p, %p) vs [%p, %p). "
```
alex-breslow-amd: I think you can probably set an environment variable in rccl-tests that says whether you're testing in-place all-to-all to turn off this logging or just turn it off if rccl-tests is running this.  Even simpler.

projects/rccl/src/allocator.cc
```diff
@@ -81,10 +81,26 @@ ncclResult_t ncclMemAlloc_impl(void** ptr, size_t size) {
...
+        hipCtx_t ctx;
+        hipError_t err = hipDevicePrimaryCtxRetain(&ctx, i);
+        if (err == hipSuccess) {
+          accessDesc.location.type = CU_MEM_LOCATION_TYPE_DEVICE;
+          accessDesc.location.id = i;
+          accessDesc.flags = CU_MEM_ACCESS_FLAGS_PROT_READWRITE;
+          hipError_t const accessErr =
+            cuMemSetAccess((CUdeviceptr)*ptr, handleSize, &accessDesc, 1);
+          CUDACHECKIGNORE(hipDevicePrimaryCtxRelease(i));
+          CUCHECK(accessErr);
```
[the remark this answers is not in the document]
corey-derochie-amd: I agree this would be ideal.

projects/rccl/src/allocator.cc
```diff
@@ -81,10 +81,26 @@ ncclResult_t ncclMemAlloc_impl(void** ptr, size_t size) {
...
+        } else {
+          WARN("hipDevicePrimaryCtxRetain failed for GPU %d with error %d: %s", i, err, hipGetErrorString(err));
+          (void)cuMemUnmap((CUdeviceptr)*ptr, handleSize);
+          (void)cuMemAddressFree((CUdeviceptr)*ptr, handleSize);
+          (void)cuMemRelease(handle);
+          *ptr = NULL;
+          ret = rcclCudaErrorHandler(err);
+          goto exit;
+        }
```
corey-derochie-amd: Whenever I see a large deviation from what is probably functional NCCL code, I have to wonder why it doesn't work for RCCL. In this case I'm guessing it's because HIP handles context management different than CUDA, is that what's happening here?

PJAvinash: Initially when I was testing symmetric memory in a container with latest HIP, rocm environments, I did see some consistent driver failures with a simple standalone program. Checking the dmesg logs pushed the suspicion towards calling cuMemSetAccess(), without uninitialized context. This is not reproducible normally. I had to run 1000+ iterations (in multithreaded 1 stream per thread case) to reach those errors. I refactored this part and the errors resolved.

projects/rccl/src/allocator.cc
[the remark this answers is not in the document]
PJAvinash: Will add the cleanup block right away.

projects/rccl/src/tuning/sym_model.cc
```diff
@@ -381,6 +381,12 @@ ncclResult_t ncclTuningSymkModelSim(struct ncclTuningInput_t* const inputs, stru
 
   tuning->timeUs = kTime * (1.0f + smPenalty * kBlocks);
   tuning->nChannels = kBlocks;
+#if defined(__HIP_PLATFORM_AMD__) || defined(__HIPCC__)
+  // ncclSymkMaxThreads is the kernel's LL slot/iteration width. Convert it
+  // using the runtime wave size instead of inheriting CUDA's 16-warp launch.
+  tuning->nWarps = std::max(1, ncclSymkMaxThreads / inputs->comm->WarpSize);
```
mch: What motivated this change? Is there a performance impact? It's not discussed in the description as far as I can tell.

projects/rccl/src/rma/rma_ce.cc
```diff
@@ -106,7 +106,18 @@ ncclResult_t ncclRmaCeInit(struct ncclComm* comm) {
 }
 
 ncclResult_t ncclRmaCeFinalize(struct ncclComm* comm) {
+  if (!comm) {
+    return ncclSuccess;
+  }
+
   ncclResult_t ret = ncclSuccess;
+  auto recordCudaError = [&ret](cudaError_t const cudaResult) {
```
atulkulk, line 114: `recordCudaError` is only used for the two CUDA destroys in `ncclRmaCeFinalize`. The actual leak fix is not aborting teardown on the first failure (`CUDACHECKGOTO` → `fail:`), so the remaining contexts / windows still get freed.

We already have `CUDACHECKIGNORE` for that, and that is what `ncclCeFinalize` uses for stream/event destroy. The extra bit the lambda does (push the CUDA error into `ret` via `rcclCudaErrorHandler`) is not needed to stop the leak, and I don’t think anything else in the PR uses it.

It also doesn’t help the tests — `ncclRmaCeFinalize` is stubbed to success in the host unit tests, and the SingleProcMemReg / destroy paths just expect `ncclCommDestroy` to return success. Nothing is injecting a failing `cudaStreamDestroy`/`cudaEventDestroy` here.

Can we just use `CUDACHECKIGNORE` for the stream/event destroy and keep `NCCLCHECKIGNORE` for the NCCL frees? Same leak fix, no new helper.

atulkulk (changes requested):

projects/rccl/src/init.cc
```diff
@@ -25,6 +25,7 @@
 #include "diagnostics.h"
 #include "enqueue.h"
 #include "graph.h"
+#include "rccl_graph_gen.h"
```
atulkulk, line 28: Duplicate. Already on line 69.

projects/rccl/src/group.cc
```diff
@@ -802,7 +802,11 @@ static void ncclGroupSymmetricJobFree(void* _job) {
 static ncclResult_t ncclPrepareTasksAndCollPreconnect(
   struct ncclComm* comm, ncclSimInfo_t* simInfo,
   struct ncclIntruQueue<struct ncclAsyncJob, &ncclAsyncJob::next>* asyncCollJobs) {
-  if (ncclParamSingleProcMemRegEnable()) {
+  // Symmetric task selection performs a bootstrap consensus in ncclPrepareTasks.
+  // Prepare grouped local ranks concurrently so the first rank cannot block
+  // before the other ranks in this process enter the same consensus.
+  bool const symmetricPrepareUsesBootstrap = comm->symmetricSupport && !comm->p2pCrossClique;
+  if (ncclParamSingleProcMemRegEnable() || symmetricPrepareUsesBootstrap) {
```
atulkulk, line 809: The concurrent prepare+preconnect job used to be gated on `NCCL_SINGLE_PROC_MEM_REG_ENABLE` (default 0). Now it also runs for `symmetricSupport && !p2pCrossClique`. That is almost every symmetric comm — `p2pCrossClique` is MNNVL cross-clique, not intra-node P2P — including 1-rank-per-process MPI.

The deadlock this is solving is only when one process owns multiple ranks of the same comm: `groupLaunch` calls `ncclPrepareTasks` inline per comm, and `ncclMakeSymmetricTaskList` does a bootstrap AllGather, so rank 0 waits for rank 1 before rank 1 has entered prepare. 1-rank-per-process does not have that hang.

Can we gate on `intraRanks > 1 && symmetricSupport && !p2pCrossClique` instead? That is the hang (several ranks of this comm in one process). `NCCL_SINGLE_PROC_MEM_REG_ENABLE` can stay the feature flag for single-proc mem reg; it does not need to be in this `if`. `intraRanks` already is the intra-process rank count.

atulkulk (changes requested):

projects/rccl/src/enqueue/enqueue.cc
atulkulk, line 1: This looks like a purely formatting change not related to the PR's intent. Pls remove this. Formatting changes can be a separate PR in itself.

projects/rccl/src/allocator.cc
```diff
@@ -84,11 +84,18 @@ ncclResult_t ncclMemAlloc_impl(void** ptr, size_t size) {
...
         accessDesc.location.id = i;
         accessDesc.flags = CU_MEM_ACCESS_FLAGS_PROT_READWRITE;
-        CUCHECK(cuMemSetAccess((CUdeviceptr)*ptr, handleSize, &accessDesc, 1));
+        CUCHECKGOTO(cuMemSetAccess((CUdeviceptr)*ptr, handleSize, &accessDesc, 1), ret, vmm_fail);
       }
       if (0 == p2p && i != cudaDev) INFO(NCCL_ALLOC, "P2P not supported between GPU%d and GPU%d", cudaDev, i);
     }
     goto exit;
+
+vmm_fail:
```
atulkulk, line 99: `vmm_fail` only runs for `cuMemSetAccess`. Create / reserve / map still `CUCHECK` and return, so a failed map still leaks the handle and VA.

Can we `CUCHECKGOTO` those three as well, with the same `handleCreated` / `addressReserved` / `mapped` flags as `ncclCuMemAlloc`? Unmap only if mapped, address-free only if reserved, release only if the handle was created. That is the rollback we already have on the internal allocator.

projects/rccl/src/init.cc
```diff
@@ -4047,8 +4049,14 @@ static ncclResult_t commDestroySync(struct ncclAsyncJob* job_) {
     }
   }
 
-  if ((ret = ncclProxyStop(comm)) != ncclSuccess) {
-    INFO(NCCL_DESTROY | NCCL_PROXY, "commDestroySync: comm %p (rank = %d) proxy stop error %d", comm, comm->rank, ret);
+  // Finalization is null-safe and also releases partially initialized state.
+  NCCLCHECKIGNORE(ncclRmaCeFinalize(comm), ret);
```
atulkulk: We need `ncclRmaCeFinalize` - develop never calls it, so the CE stream/event and signal windows leak on destroy. Window deregister needs `devr` still alive, so this has to run before `ncclDevrFinalize`.

`commDestroySync` is not that place. Revoke sets `finalizeCalled`, `commReclaim` skips DestroySync, and `commFree` still runs. Partial init also goes through `commFree`.

Can we call it from `commFree` next to `ncclCeFinalize` and leave DestroySync as it was?

projects/rccl/src/tuning/sym_model.cc
atulkulk: The `1` is not a real floor. LL needs `nWarps = ncclSymkMaxThreads / WarpSize` so we launch 256 threads. One warp is 32 or 64 threads into a 256-thread protocol, and GIN reduce-scatter does `nWarps - 2` workers, so 1 is invalid there too.

`std::max` is not needed either. `WarpSize` is 32 or 64 from `hipDeviceGetAttribute` in `commAlloc` (that call failing fails the comm). So this is 256/64=4 or 256/32=8. Integer division is 0 only if `WarpSize > 256`. Divide-by-zero is `WarpSize == 0`, and `max` does not save that — the divide runs first.

Can we just use `ncclSymkMaxThreads / inputs->comm->WarpSize`?

atulkulk (changes requested):

PJAvinash:
This PR is split into smaller PRs.

PR #8962, opened 2026-07-21
test(clr): fix OCLEmptyDeviceEnqueue read-back and drop invalid sub-test

chinmaydd:
## Summary

- Remove the deadbranch sub-test. It dispatched a kernel containing get_default_queue()/enqueue_kernel codegen with no on-device queue created. The OpenCL spec requires the application to create the default device queue when a kernel containing get_default_queue is enqueued; without it the parent kernel's execution is not guaranteed and fails on gfx12 (it only worked on gfx11 by luck).
- Read results with clEnqueueReadBuffer after clFinish instead of reading a buffer mapped before the dispatch. Using a mapped buffer in a kernel is undefined and not coherent on all ASICs, which produced spurious 0xeeeeeeee read-backs.

## Motivation

The deadbranch sub-test did not reflect the BurnIn scenario the original fix targets. BurnIn kernels carry the -O0 device-enqueue ABI but never call enqueue_kernel and run without an on-device queue - covered by the plain sub-test - while real device enqueue (with a queue) is covered by the real-enqueue sub-test. The removed sub-test asserted behavior the spec does not guarantee, so it failed on gfx1201 while the runtime (and BurnIn) were correct.

## Issue Tracking

JIRA ID : SWDEV-598021

## Test Plan

This test passed on gfx11 but failed on gfx12 during the downstream compute PSDB runs. Removing the test here ensures that the changes synced across both repos.

## Test Result

OCLTests should pass.

## Submission Checklist


chinmaydd:
ping on this @rakesroy

rakesroy (approved):

chinmaydd:
Thanks Rakesh. Could someone from @clr-reviewers approve too ?

chrispaquot (approved):

PR #8963, opened 2026-07-21
ci(rocprofiler-compute): Update Docker images on CI workflows

jbonnell-amd:
## Motivation

The `rocprofiler-compute-tarball.yml` and `rocprofiler-compute-rhel8.yml` workflows were failing due to their respective Docker images being removed from Dockerhub.  Since these dockers were originally meant for rocprofiler-systems, we want to publish and maintain our own images to use instead.

## Technical Details

- Updated `.github/workflows/rocprofiler-compute-ghcr.yml`
  - Adds a new "base" option, which uses `Dockerfile.*.ci` but without downloading TheRock tarball
  - Enables workflow_dispatch input to specify which Dockers to update (helpful for manual testing)
- Updated `.github/workflows/rocprofiler-compute-rhel-8.yml` and renamed it to `.github/workflows/rocprofiler-compute-rhel.yml`
  - Original name implied only RHEL8 being tested, but we also use RHEL9 so this keeps it more aligned with the actual workflow
  - Uses new "base" images hosted on GHCR
  - Made some minor changes to accommodate new base images
- Updated `.github/workflows/rocprofiler-compute-tarball.yml`
  - Use new "base" images, updated spacing to make it more readable
- Updated `projects/rocprofiler-compute/docker/containers-ci.yml`
  - Added an explicit RHEL 9.4 option
  - The default RHEL9 tag uses RHEL 9.8, which does not have ROCm repositories set up for it

## Issue Tracking

JIRA ID : AIPROFCOMP-711

## Test Plan

- Manually uploaded new "base" images to GHCR
- Confirmed new images are used in workflows and that they pass as intended

## Test Result

- No regressions with changed workflow files

## Submission Checklist


xuchen-amd (approved):
thanks a lot Jason!

dgaliffiAMD (approved):

PR #8965, opened 2026-07-21
perf(rocjitsu): optimize functional execution

lialan:
## Summary

Optimize RocJITsu functional execution now that #6962 and #10074 are merged.

- Batch GPU memory reads and cache each CU instruction-fetch window.
- Vectorize common 32-bit VGPR writeback and contiguous LDS accesses.
- Reduce L1 vector-cache tag scans and page-type lookups.
- Make the shared dispatch-pool scaling harness allocate disjoint buffers, validate outputs, and report true three-run medians.

## Correctness and safety

- Mask bulk VGPR writes to the runtime wave size, including safe wave32, wave64, partial-mask, and zero-mask paths.
- Preserve fallback paths for masked, non-contiguous, sub-dword, atomic, and out-of-bounds accesses.
- Retry instruction fetches after execution exceptions without retaining stale bytes, and invalidate scalar launch data once per dispatch.
- Keep LDS allocation and growth serialized while allowing sibling workers to access stable, disjoint workgroup regions.

## Validation

- Rebased onto `develop` at `3f6958e26a`.
- Release Ninja build with amdclang++: `rocjitsu_tests` and `scaling_test` passed.
- Focused VGPR, LDS, L1, and instruction-cache regressions: 33/33 passed.
- Full `rocjitsu_tests`: 3,946 passed; 1 disabled.
- Correctness-checked scaling smoke, three-run medians on one host:
  - 1 thread: vector add 119.984 ms, tiled matmul 877.126 ms, MFMA matmul 1470.4 ms.
  - 8 threads: vector add 41.4247 ms, tiled matmul 211.065 ms, MFMA matmul 360.819 ms.
- These timing rows are smoke measurements, not a controlled before/after performance comparison.
- Repository pre-commit hooks passed.

## PR stack

1. #6962 — merged
2. #10074 — merged
3. #8965 — this PR

ISSUE ID: #6447

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
@@ -30,6 +30,7 @@ RJ_DIAGNOSTIC_POP
 #include <format>
 #include <limits>
 #include <optional>
+#include <span>
 #include <stdexcept>
 #include <string>
 #include <sys/mman.h>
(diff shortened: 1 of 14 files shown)


lialan:
Rebased this performance follow-up onto #6962's final head, which is now based on current `origin/develop` (`d66f4e63`). During validation I removed the decoded-instruction object cache because decoder allocations use thread-local lifetime hooks that are incompatible with retaining objects across functional workers; the 16-byte instruction-fetch cache remains. I also preserved cross-page command-processor reads and current cache-coherence behavior.

Validation: all 2,799 `rocjitsu_tests` completed (2,798 passed, 1 expected skip), the two recovered regressions passed 10 repeated iterations each, and repository pre-commit checks passed.

PR #6323, opened 2026-05-20
[rocSHMEM] Refactor IPC memcpy primitives with typed cache policies and pipelined unrolling

amirsojoodi:
## Motivation

The current IPC device-side memcpy implementation uses a scalar load-store loop with no cache policy control and no instruction-level pipelining. This wastes a significant performance opportunity. 

The copy loop should be restructured to issue a batch of loads before any stores, allowing the hardware to pipeline memory operations and hide latency. For this to happen, a lot of changes had to be done. 

As asm instructions are basically black box to the compiler, it cannot automatically insert proper waitcnts in the pipeline. So, either we should use the standard C++ functions, builtin functions, or manual setup. I considered this thoroughly to utilize builtins, but because builtins do not support every little details of all the modern architectures, I decided to keep the assembly functions in the code base for now and implement the pipeline manually with the correct waitcnts. 

Importantly, waitcount instructions should be inserted to provide a window such that no load or store can compete with each other. 

Buffer instructions were initially considered for this pipeline implementation. However, they cannot read from LDS. So, I tried to work around the flat instructions instead.

## Technical Details

Important notes 

1. The wait calls implemented in `assembly.hpp` are used by `memcpy_*` variants before and after flat instructions. 
2. Flat instructions read, or write, one piece of data into, or out of, VGPRs. These instructions use a flat address that includes video memory, system memory, LDS memory, and scratch (private) memory.
3. Importantly, flat instructions cannot return data directly to LDS, but only to VGPRS.
4. Internally, FLAT instructions are executed as both an LDS and a Buffer instruction, so both counters vmcnt and lgkmcnt is affected. So, both should be checked for completion if we want to support both memory types.
5. (CDNA4) There is no way to determine whether a flat instruction uses only LDS or TA.
6. **Caution regarding timing:** "(from CDNA4 ISA) Since the data for a FLAT load can come from either LDS or the texture cache, and because these units have different latencies, there is a potential race condition with respect to the VM_CNT and LGKM_CNT counters. Because of this, the only sensible S_WAITCNT value to use after FLAT instructions is zero." For this matter we should specifically mention in documentation, that using a mix of texture memory or LDS in flat instructions is unsafe. Only one of the memory types can be used at a time by the same wave. (not a mix in the same wave)

## JIRA ID

AIROCSHMEM-416
AIROCSHMEM-415

## Test Result

Functional tests, `wgput`, `waveput`, and `wgget` pass with performance improvement.
[...]

PR #8970, opened 2026-07-21
perf(rccl): port #8451 LL128/scratch reduction improvements to June drop (behavior-neutral)

lfmeadow:
## Summary

Ports the **behavior-neutral** subset of #8451 onto the June drop (`rccl/drop_2026-06-25`, tip `df2d9d40` = #6323). Codegen/build-only: reduces LL128 register spills and scratch-segment usage on gfx942/gfx950 without changing any numerical results. All fp8 arithmetic changes and unit tests from #8451 are intentionally excluded.

The base is the drop-branch tip, so the diff is exactly the single port commit.

## Changes

- **`cmake/DeviceLinker.cmake`** — build device code with `-DNDEBUG` in Release, matching the host Release build. Device asserts otherwise pin live registers around every `__assert_fail`, inflating the private (scratch) segment on register-heavy reduce kernels.
- **`src/device/common_kernel.h`** — in `reduceCopyPacksWithBias`, fuse each pack's load+reduce so only one temporary pack is live (instead of `tmp[Unroll]`), and defer the bias/acc load to its point of use; advance `accPtr` by the full hunk to compensate. (Uses this base's `applyPreOp(redFn, ...)` convention, matching #8451 upstream.)
- **`src/device/{all_reduce,all_gather,broadcast,reduce,reduce_scatter,alltoall_pivot,sendrecv}.h`** — drop the `noinline` attribute on `runRing`/`runTree*`/`run` for gfx942/gfx950 so they inline (codegen-only; no behavior change).
- **`src/device/prims_ll128.h`** — comment only: document why the send FIFO stays indexed by `sendStep[wid]` (NCCL parity; no functional change).
- **`cmake/scripts/extract_metadata.cmake`** — replace the multi-step metadata extraction with a single `llvm-objdump --offloading` call.
- **`src/include/rccl_common.h`, `src/transport/net.cc`** — supporting includes.

## Not included (vs #8451)

- All fp8/bf8 packed-arithmetic helper changes (reverted per Meta's overflow-semantics feedback) and the associated fp8 reduce unit tests.
- The `src/ipc_init.cu` debug include — not applicable at this base.

## Test plan

- [x] Clean gfx950 build in the ROCm 7.0.2.1-22 container (device linker enabled; 0 errors).
- [ ] RCCL unit tests on gfx950.
- [ ] (optional) Compare device scratch/VGPR from `*.resources.json` before/after to confirm the scratch reduction.

### DIFF
--- a/projects/rccl/cmake/DeviceLinker.cmake
+++ b/projects/rccl/cmake/DeviceLinker.cmake
@@ -64,7 +64,10 @@ message(STATUS "Device Linker: GPU targets = ${DL_GPU_TARGETS}")
 if(CMAKE_BUILD_TYPE MATCHES "Debug")
   set(DL_OPT_FLAGS -O1 -g)
 else()
-  set(DL_OPT_FLAGS -O3)
+  # -DNDEBUG matches the host Release build: device asserts otherwise force the
+  # compiler to preserve live registers around every __assert_fail call, which
+  # inflates the private (scratch) segment on register-heavy reduce kernels.
+  set(DL_OPT_FLAGS -O3 -DNDEBUG)
 endif()
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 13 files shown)


lfmeadow:
Gentle ping — any interest in merging this, or should I close it?

PR #8973, opened 2026-07-21
rocr: Replace assert-only error checks with proper error handling

shwetagkhatri:
In release builds (NDEBUG defined), assert() expands to nothing. Multiple code paths used assert() as the ONLY check for critical operations, causing crashes with no diagnostic information.

## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
@@ -625,7 +625,11 @@ void AqlQueue::AllocRegisteredRingBuffer(uint32_t queue_size_pkts) {
         core::MemoryRegion::AllocateExecutable);
   }
 
-  assert(ring_buf_ != NULL && "AQL queue memory allocation failure");
+  if (ring_buf_ == nullptr) {
+    assert(false && "AQL queue memory allocation failure");
+    throw AMD::hsa_exception(HSA_STATUS_ERROR_OUT_OF_RESOURCES,
+                              "AQL queue ring buffer allocation failed.");
+  }
   // Fill the ring buffer with invalid packet headers.
   // Leave packet content uninitialized to help track errors.
   for (uint32_t pkt_id = 0; pkt_id < queue_size_pkts; ++pkt_id)
@@ -1682,7 +1686,10 @@ void AqlQueue::ExecutePM4(uint32_t* cmd_data, size_t cmd_size_b, hsa_fence_scope
 
     if (!in_signal) {
       err = hsa_signal_create(1, 0, NULL, &local_signal);
-      assert(err == HSA_STATUS_SUCCESS);
+      if (err != HSA_STATUS_SUCCESS) {
+        assert(false && "hsa_signal_create failed in ExecutePM4");
+        throw AMD::hsa_exception(err, "Failed to create signal for PM4 execution.");
+      }
     }
 
     constexpr uint32_t AMD_AQL_FORMAT_PM4_IB = 0x1;
@@ -1726,13 +1733,16 @@ void AqlQueue::ExecutePM4(uint32_t* cmd_data, size_t cmd_size_b, hsa_fence_scope
 
     if (in_signal) hsa_signal_store_screlease(*in_signal, 0);
   } else if (!in_signal) {
-    // On gfx9 and newer, if in_signal is not provided, we block and wait for own signal
-    hsa_signal_value_t ret;
-    ret = hsa_signal_wait_scacquire(local_signal, HSA_SIGNAL_CONDITION_LT, 1, (uint64_t)-1,
-                                    HSA_WAIT_STATE_ACTIVE);
+    // On gfx9 and newer, if in_signal is not provided, we block and wait for own signal.
+    // Use do/while loop since hsa_signal_wait_scacquire may return spuriously.
+    // Break condition matches wait condition: HSA_SIGNAL_CONDITION_LT with compare value 1.
+    do {
+      hsa_signal_value_t ret = hsa_signal_wait_scacquire(
+          local_signal, HSA_SIGNAL_CONDITION_LT, 1, (uint64_t)-1, HSA_WAIT_STATE_ACTIVE);
+      if (ret < 1) break;
+    } while (true);
     err = hsa_signal_destroy(local_signal);
-    assert(ret == 0 && err == HSA_STATUS_SUCCESS);
-    (void)ret;
+    assert(err == HSA_STATUS_SUCCESS && "hsa_signal_destroy failed");
   }
   (void)err;
 }
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_blit_sdma.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_blit_sdma.cpp
@@ -123,8 +123,8 @@ uint32_t BlitSdma<useGCR, scopeFields>::gcr_command_size() {
 
 template <bool useGCR, bool scopeFields>
 BlitSdma<useGCR, scopeFields>::BlitSdma()
-    : agent_(NULL),
-      queue_start_addr_(NULL),
+    : agent_(nullptr),
+      queue_start_addr_(nullptr),
       bytes_queued_(0),
       parity_(false),
       cached_reserve_index_(0),
@@ -155,7 +155,7 @@ template <bool useGCR, bool scopeFields> BlitSdma<useGCR, scopeFields>::~BlitSdm
 template <bool useGCR, bool scopeFields>
 hsa_status_t BlitSdma<useGCR, scopeFields>::Initialize(const core::Agent& agent, bool use_xgmi,
                                           size_t linear_copy_size_override, int rec_eng) {
-  if (queue_start_addr_ != NULL) {
+  if (queue_start_addr_ != nullptr) {
     // Already initialized.
     return HSA_STATUS_SUCCESS;
   }
@@ -221,7 +221,7 @@ hsa_status_t BlitSdma<useGCR, scopeFields>::Initialize(const core::Agent& agent,
   queue_start_addr_ =
       (char*)agent_->system_allocator()(kQueueSize, 0x1000, core::MemoryRegion::AllocateExecutable);
 
-  if (queue_start_addr_ == NULL) {
+  if (queue_start_addr_ == nullptr) {
     return HSA_STATUS_ERROR_OUT_OF_RESOURCES;
   }
   MAKE_NAMED_SCOPE_GUARD(cleanupOnException, [&]() { Destroy(); };);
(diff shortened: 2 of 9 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/svm_profiler.cpp
```diff
@@ -153,12 +153,18 @@ void SvmProfileControl::PollSmi() {
   for (int i = 0; i < core::Runtime::runtime_singleton_->gpu_agents().size(); i++) {
     auto gpu_agent = core::Runtime::runtime_singleton_->gpu_agents()[i];
     auto err = gpu_agent->driver().OpenSMI(gpu_agent->node_id(), &files[i + 1].fd);
-    assert(err == HSA_STATUS_SUCCESS);
+    if (err != HSA_STATUS_SUCCESS) {
+      assert(false && "OpenSMI failed");
+      throw AMD::hsa_exception(err, "SVM profiler: OpenSMI failed.");
+    }
```
cfreeamd, line 168: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** These `throw`s escape `PollSmi` before the `smiGuard` scope guard (which closes the opened SMI fds) is established later in the function, so every fd opened in an earlier loop iteration leaks. And since `PollSmi` runs on the SMI polling thread (`PollSmiRun`), an `hsa_exception` that unwinds past the thread entry with no handler will call `std::terminate` and abort the process.

This is a regression: in release builds the prior `assert` was a no-op and the loop simply continued. Recommend logging, closing the fds opened so far, and `return`ing instead of throwing (or restructuring so the guard covers the partially-opened state). Same applies to the write-failure throw just below (~line 167).

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_blit_sdma.cpp
```diff
@@ -2087,7 +2089,10 @@ template <bool useGCR, bool scopeFields> bool BlitSdma<useGCR, scopeFields>::Can
 template <bool useGCR, bool scopeFields>
 void BlitSdma<useGCR, scopeFields>::BuildFenceCommand(char* fence_command_addr, uint32_t* fence,
                                          uint32_t fence_value) {
-  assert(fence_command_addr != NULL);
+  if (fence_command_addr == nullptr) {
+    assert(false && "fence_command_addr is NULL");
+    return;
+  }
```
cfreeamd, line 2095: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** Converting `assert(fence_command_addr != NULL)` to a silent early `return` means that if the pointer ever is null, the caller proceeds to submit a ring slot containing an uninitialized/garbage packet — a hang or corruption that's much harder to diagnose than an abort. This is an unrecoverable "can't happen" invariant on the submit path, so fail-fast (keep the abort / throw) is preferable to a silent no-op. Same pattern in `BuildHdpFlushCommand` (~2986) and `BuildGCRCommand` (~2995).

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
```diff
@@ -1718,13 +1725,15 @@ void AqlQueue::ExecutePM4(uint32_t* cmd_data, size_t cmd_size_b, hsa_fence_scope
...
-    hsa_signal_value_t ret;
-    ret = hsa_signal_wait_scacquire(local_signal, HSA_SIGNAL_CONDITION_LT, 1, (uint64_t)-1,
-                                    HSA_WAIT_STATE_ACTIVE);
+    // On gfx9 and newer, if in_signal is not provided, we block and wait for own signal.
+    // Use do/while loop since hsa_signal_wait_scacquire may return spuriously.
+    do {
+      hsa_signal_value_t ret = hsa_signal_wait_scacquire(
+          local_signal, HSA_SIGNAL_CONDITION_LT, 1, (uint64_t)-1, HSA_WAIT_STATE_ACTIVE);
+      if (ret == 0) break;
+    } while (true);
```
cfreeamd, line 1743: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]/[Nit]** The new do/while breaks only on `ret == 0`, but the wait condition is `HSA_SIGNAL_CONDITION_LT` with compare value 1. If the signal value ever goes negative, `ret == 0` is never satisfied and the loop spins forever — `ret < 1` matches the wait condition and is safe. (Defensive: the signal is created at 1 and decremented once to 0, but the break should match the wait condition.)

Also, this spurious-wakeup do/while rewrite is a functional change beyond "replace assert-only error checks" — is it intentionally bundled into this PR?

cfreeamd (commented):

The assert→graceful-handling pattern is sound and consistently applied. The public-API conversions in hsa.cpp (hsa_queue_create / hsa_soft_queue_create returning a status) and the image/* files (blit_kernel.cpp, image_manager.cpp, image_runtime.cpp returning hsa_status_t) look correct and match the existing early-return style in those functions. Inline notes below focus on the error-path behavior changes in svm_profiler.cpp, amd_blit_sdma.cpp, and amd_aql_queue.cpp.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/svm_profiler.cpp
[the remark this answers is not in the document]
shwetagkhatri: Fixed.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/svm_profiler.cpp
```diff
@@ -153,12 +153,18 @@ void SvmProfileControl::PollSmi() {
...
     files[i + 1].events = POLLIN;
     files[i + 1].revents = 0;
     // Enable collecting masked events.
     auto wrote = write(files[i + 1].fd, &events, sizeof(events));
-    assert(wrote == sizeof(events));
+    if (wrote != sizeof(events)) {
+      assert(false && "write to SMI fd failed");
+      throw AMD::hsa_exception(HSA_STATUS_ERROR, "SVM profiler: write to SMI fd failed.");
+    }
```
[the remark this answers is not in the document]
shwetagkhatri: Fixed.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_blit_sdma.cpp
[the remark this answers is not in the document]
shwetagkhatri: Fixed.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_blit_sdma.cpp
```diff
@@ -2975,13 +2980,19 @@ template <bool useGCR, bool scopeFields> void BlitSdma<useGCR, scopeFields>::Bui
...
   SDMA_PKT_POLL_REGMEM* addr = reinterpret_cast<SDMA_PKT_POLL_REGMEM*>(cmd_addr);
   memcpy(addr, &hdp_flush_cmd, flush_command_size_);
 }
 
 template <bool useGCR, bool scopeFields> void BlitSdma<useGCR, scopeFields>::BuildGCRCommand(char* cmd_addr, bool invalidate) {
-  assert(cmd_addr != NULL);
+  if (cmd_addr == nullptr) {
+    assert(false && "cmd_addr is NULL in BuildGCRCommand");
+    return;
+  }
```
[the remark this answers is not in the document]
shwetagkhatri: Fixed.

shwetagkhatri:
Hi @cfreeamd ,
I have resolved each of your review comments. Please let me know if you see any more issues. Thank you so much for prompt code review help!

projects/rocr-runtime/runtime/hsa-runtime/image/blit_kernel.cpp
```diff
@@ -292,7 +295,10 @@ hsa_status_t BlitKernel::CopyBufferToImage(
   };
 
   KernelArgs* args = (KernelArgs*)Allocate(dst_image_view->component, sizeof(KernelArgs));
-  assert(args != NULL);
+  if (args == nullptr) {
+    assert(false && "Failed to allocate kernel args");
+    return HSA_STATUS_ERROR_OUT_OF_RESOURCES;
+  }
```
cfreeamd, line 301: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** This new `return HSA_STATUS_ERROR_OUT_OF_RESOURCES` skips the end-of-function cleanup `if (&dst_image != dst_image_view) Image::Destroy(dst_image_view);`. When `ConvertImage` created a temporary view (non-trivial channel type/order) and `Allocate` then fails, that view leaks. Same pattern in `CopyImageToBuffer` (~line 415, leaks `src_image_view`) and `CopyImage` (~line 541, can leak both views). Error-path + kernarg-OOM only, so very low probability, but it runs counter to the PR's "proper error handling" goal — consider routing these through a `goto cleanup` / destroy-views-then-return path. (Note: this predates your change — the original asserts had the same gap — so purely optional.)

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/svm_profiler.cpp
```diff
@@ -150,15 +150,32 @@ void SvmProfileControl::PollSmi() {
...
+
   for (int i = 0; i < core::Runtime::runtime_singleton_->gpu_agents().size(); i++) {
     auto gpu_agent = core::Runtime::runtime_singleton_->gpu_agents()[i];
     auto err = gpu_agent->driver().OpenSMI(gpu_agent->node_id(), &files[i + 1].fd);
-    assert(err == HSA_STATUS_SUCCESS);
+    if (err != HSA_STATUS_SUCCESS) {
+      assert(false && "OpenSMI failed");
+      cleanup_fds();
+      return;  // Log and return instead of throwing to avoid process termination
+    }
```
cfreeamd, line 168: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** The comment says "Log and return instead of throwing" but nothing is actually logged here — under `NDEBUG` the `assert(false && ...)` is compiled out, so a release build closes the fds and returns silently with no diagnostic. Either add a real log call (fprintf/debug_print) or reword the comment to "cleanup and return". Same on the write-failure path just below (~line 178).

cfreeamd (commented):

Thanks for the quick turnaround. All three prior findings are resolved and verified:

- **svm_profiler.cpp `PollSmi`** — throws replaced with `opened_fds`-tracked `cleanup_fds()` + `return`; fd accounting is correct on both failure paths and the polling thread no longer risks `std::terminate`. This was the one genuine regression — good fix.
- **amd_blit_sdma.cpp `Build{Fence,HdpFlush,GCR}Command`** — now `abort()` with a diagnostic on null, the fail-fast that was recommended.
- **amd_aql_queue.cpp `ExecutePM4`** — the do/while breaks on `ret < 1`, matching the `LT 1` wait condition; negative-value spin is gone.

Two minor, non-blocking items remain (both flagged by the Copilot bot after your "resolved" note, so likely not yet seen) — inline comments below.

cfreeamd (approved):

PR #8975, opened 2026-07-21
fix(rocprofiler-systems): check AI NIC track presence, not non-zero values

ajanicijamd:
## Motivation

The AI NIC performance test (`test_performance_tracks`) was failing because no RDMA traffic flows during the profiling window. The Perfetto trace validation was checking that `SUM(counter.value) > 0` for each AI NIC counter track, which is too strict: the tracks are correctly written to the trace even when no RDMA traffic occurs.

JIRA ID: ROCM-28094

## Technical Details

Add a `--counter-names-presence-only` flag to `validate-perfetto-proto.py`. When set, the counter-names check verifies only that matching counter tracks **exist** in the trace (`COUNT(DISTINCT track_id) > 0`), without requiring non-zero values. The existing value-sum check is kept in the `else` branch, so all other tests that use `--counter-names` are unaffected.

The flag is threaded through the call chain:
- `validators.py`: `counter_names_presence_only` parameter added to `validate_perfetto_trace()`
- `conftest.py`: same parameter added to `_assert_perfetto()`
- `test_ainic.py`: `counter_names_presence_only=True` passed in `test_performance_tracks`

The ROCpd (`.db`) validation already performs presence-only checks (`rocpd_string_*` and `rocpd_info_pmc_*` table queries), so no change is needed there.

**Simplified `test_performance_tracks` fixture**

Removed the `ainic_download_url_2` fixture and the second `wget` download from `test_performance_tracks`. The two-URL approach was originally needed to extend the test duration and increase the chance of capturing non-zero RDMA metric values. Since the Perfetto validation now only checks for the presence of AI NIC counter tracks (not non-zero values), a single download is sufficient to exercise the NIC sampling path. The fixture was also renamed from `ainic_download_url_1` to `ainic_download_url` to reflect that it is now the only one.

## Test Plan

```
ctest -R ainic
```

## Test Results

Verified on a system with an AMD AI NIC (ionic/Pensando, AMD SMI 26.5):

```
ctest -R ainic
Test project /code/projects/rocprofiler-systems/build/debug
    Start   1: rocprofiler-systems-pytest-config
1/6 Test   #1: rocprofiler-systems-pytest-config ...   Passed    1.49 sec
    Start   2: ainic-settings-present
2/6 Test   #2: ainic-settings-present ..............   Passed    0.32 sec
    Start   3: ainic-sampling-ainics-help-run
3/6 Test   #3: ainic-sampling-ainics-help-run ......   Passed    0.32 sec
    Start   4: ainic-sampling-ainics-help-sample
4/6 Test   #4: ainic-sampling-ainics-help-sample ...   Passed    0.33 sec
    Start   5: ainic-performance-tracks
5/6 Test   #5: ainic-performance-tracks ............   Passed    3.94 sec
    Start 621: rocprofiler-systems-test-cleanup
6/6 Test #621: rocprofiler-systems-test-cleanup ....   Passed    0.07 sec
```

## Submission Checklist


projects/rocprofiler-systems/tests/pytest/rocprofsys/validators.py
kcossett-amd, line 280: Please add `counter_names_presence_only` to the args description list here

projects/rocprofiler-systems/tests/validate-perfetto-proto.py
```diff
@@ -213,6 +213,15 @@ def validate_perfetto_by_label(
         action="store_true",
         help="Verify each counter track has paired start/end entries (even count, last value is 0)",
     )
+    parser.add_argument(
+        "--counter-names-presence-only",
+        action="store_true",
+        help=(
+            "When set, --counter-names checks only that matching counter tracks exist "
+            "in the trace; it does NOT require their values to be non-zero. "
+            "Used for AI NIC tests where RDMA counters may legitimately be 0."
```
kcossett-amd:
```suggestion
```

I'd suggest you move this comment elsewhere, closer to the `ainic` test. Later down the line, we may reuse this flag for a different feature.

You can move it into the `assert_perfetto(...)` call that sets this to `true` in `test_ainic.py` (here https://github.com/ROCm/rocm-systems/pull/8975/changes#diff-d8c36515e5b9926ea296efa5c34d97aa85a425a73d53a095c85e2fbaaee0f264R198)

projects/rocprofiler-systems/tests/pytest/test_ainic.py
```diff
@@ -94,17 +94,11 @@ def ainic_perf_env() -> dict[str, str]:
 
 
 @pytest.fixture
-def ainic_download_url_1() -> str:
+def ainic_download_url() -> str:
     """Download URL for the first file to download."""
     return "https://github.com/ROCm/rocprofiler-systems/releases/download/rocm-6.4.1/rocprofiler-systems-1.0.1-ubuntu-22.04-ROCm-60400-PAPI-OMPT-Python3.sh"
 
 
-@pytest.fixture
```
kcossett-amd, line 102: I see you've reduced the `ainic_download_url`'s from 2 to 1. Is there a reason why? (I couldn't find it in your PR desc).

ajanicijamd: Of course: having two counters made sense for the NIC test, because we needed the test to run
longer so that we can collect the metrics. For AI NIC, we are only checking if the counter tracks
exist in the output and don't care if they are all 0.

projects/rocprofiler-systems/source/lib/core/trace_cache/rocpd_helpers.hpp
```diff
@@ -25,7 +25,8 @@ make_agent_uid(const agent& agnt)
         {
             case agent_type::GPU: return "GPU";
             case agent_type::CPU: return "CPU";
-            case agent_type::NIC: return "NIC";
+            case agent_type::NIC:
+                return std::nullopt;  // profiler-hub does not support NIC; treat as NULL
             default: return std::nullopt;
```
sputhala-amd, line 29: So rocpd schema support NIC as agent type and profiler-hub does not?
I would prefer we fix profiler-hub if possible then.

projects/rocprofiler-systems/tests/pytest/test_ainic.py
ajanicijamd: I documented it in the PR description.

projects/rocprofiler-systems/tests/validate-perfetto-proto.py
ajanicijamd: Done.

projects/rocprofiler-systems/tests/pytest/conftest.py
```diff
@@ -2502,6 +2503,7 @@ def _assert_perfetto(
                 depths=depths,
                 label_substrings=label_substrings,
                 counter_names=counter_names,
+                counter_names_presence_only=counter_names_presence_only,
```
sputhala-amd, line 2563: What is the point of a `counter_names_presence_only` test? 
The actual test is only when we get values for those counters isnt it?

On a real hardware system if there is a genuine bug in counter collection this test would silently pass.

sputhala-amd (commented):
A real test should check the presence of the required hardware possibly with an amd-smi command or have the test drive deterministic RDMA (a small loopback / `ib_send_bw` / `perftest` between two local ports and possibly a skip on CI. If the hardware is not present then lets skip this test.

ajanicijamd:
> A real test should check the presence of the required hardware possibly with an amd-smi command or have the test drive deterministic RDMA (a small loopback / `ib_send_bw` / `perftest` between two local ports and possibly a skip on CI. If the hardware is not present then lets skip this test.

We do check the presence of AI NIC.
- In capabilities.py, property ai_nic_devices runs amd-smi static and returns the list of NETDEV names: [capabilities.py](https://github.com/ROCm/rocm-systems/blob/d8c0684b8d4056aa800ae96ee810bb976de13e11/projects/rocprofiler-systems/tests/pytest/rocprofsys/capabilities.py#L170)
- In conftest.py, we skip the test if AI NIC is unavailable: [conftest.py](https://github.com/ROCm/rocm-systems/blob/d8c0684b8d4056aa800ae96ee810bb976de13e11/projects/rocprofiler-systems/tests/pytest/conftest.py#L523)

sputhala-amd:
> > A real test should check the presence of the required hardware possibly with an amd-smi command or have the test drive deterministic RDMA (a small loopback / `ib_send_bw` / `perftest` between two local ports and possibly a skip on CI. If the hardware is not present then lets skip this test.
> 
> We do check the presence of AI NIC.
> 
> * In capabilities.py, property ai_nic_devices runs amd-smi static and returns the list of NETDEV names: [capabilities.py](https://github.com/ROCm/rocm-systems/blob/d8c0684b8d4056aa800ae96ee810bb976de13e11/projects/rocprofiler-systems/tests/pytest/rocprofsys/capabilities.py#L170)
> * In conftest.py, we skip the test if AI NIC is unavailable: [conftest.py](https://github.com/ROCm/rocm-systems/blob/d8c0684b8d4056aa800ae96ee810bb976de13e11/projects/rocprofiler-systems/tests/pytest/conftest.py#L523)

Then in the CI these tests should have been skipped ?

projects/rocprofiler-systems/tests/pytest/conftest.py
ajanicijamd: The reason for only checking for counter presence is that it is much easier to do.
- This kind of test (presence only) does have a point: it validates that the code is getting the list of AI NIC cards present on the system and creating the tracks in the output files.
- It would of course be much better to have actual network traffic in the test, but this is harder to do without requiring the tester to set up a system with two AI NIC cards.
- If we could use loopback for network transfer during the test, that would be ideal. We have to investigate if loopback is doable with RDMA and Pensando cards in particular.

ajanicijamd:
> > > A real test should check the presence of the required hardware possibly with an amd-smi command or have the test drive deterministic RDMA (a small loopback / `ib_send_bw` / `perftest` between two local ports and possibly a skip on CI. If the hardware is not present then lets skip this test.
> > 
> > 
> > We do check the presence of AI NIC.
> > 
> > * In capabilities.py, property ai_nic_devices runs amd-smi static and returns the list of NETDEV names: [capabilities.py](https://github.com/ROCm/rocm-systems/blob/d8c0684b8d4056aa800ae96ee810bb976de13e11/projects/rocprofiler-systems/tests/pytest/rocprofsys/capabilities.py#L170)
> > * In conftest.py, we skip the test if AI NIC is unavailable: [conftest.py](https://github.com/ROCm/rocm-systems/blob/d8c0684b8d4056aa800ae96ee810bb976de13e11/projects/rocprofiler-systems/tests/pytest/conftest.py#L523)
> 
> Then in the CI these tests should have been skipped ?

Yes, they are.

projects/rocprofiler-systems/tests/pytest/conftest.py
sputhala-amd: Are these tests meant to run as part of manual automation tests on a working environment ? If so, isn't it better to test real traffic? Or should we have the tracks presence / track existence as a proxy for "a supported AI NIC card was detected and selected + the collector was configured" and then add one more subtest which will check if the metrics are collected where it is expected?

Also the description says we are proposing this fix for CI failures, but if the tests are skipped on CI then probably this fix is proposed for another failure ?

ajanicijamd: You are right, the description was incorrect. In CI this test is skipped, because we currently have no CI runners with an AI NIC card. The test was failing when run locally on a system with the hardware installed. I have updated the description.

I agree we should add a subtest that checks for AI NIC traffic, but the plan was to do that in a separate PR.

projects/rocprofiler-systems/source/lib/core/trace_cache/rocpd_helpers.hpp
ajanicijamd: Agreed, but that will have to be another PR.

projects/rocprofiler-systems/tests/pytest/rocprofsys/validators.py
ajanicijamd: Done.

projects/rocprofiler-systems/source/lib/core/trace_cache/rocpd_helpers.hpp
dgaliffiAMD: Yes. I separate ticket has been filed and they are actively working on a fix. We should hold off merging this until that is fixed.

ajanicijamd: In view of that, I have reverted this change.

ajanicijamd: profiler-hub has been updated to support agent type "NIC".

PR #8977, opened 2026-07-21
kfdtest: Add KFDNegative tests for queue reset

amd-alysaliu:
Add ComputeQueueInvalidOpcode and SDMAQueueInvalidOpcode tests for queue reset.

## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDNegativeTest.cpp
+++ b/projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDNegativeTest.cpp
@@ -280,3 +280,217 @@ TEST_F(KFDNegativeTest, BasicSDMAReset) {
 
     TEST_END
 }
+
+/**
+ * Test compute queue reset with invalid opcode
+ *
+ * This test validates that an invalid PM4 opcode triggers queue reset.
+ */
+TEST_F(KFDNegativeTest, ComputeQueueInvalidOpcode) {
+    TEST_START(TESTPROFILE_RUNALL);
+
+    int defaultGPUNode = m_NodeInfo.HsaDefaultGPUNode();
+    ASSERT_GE(defaultGPUNode, 0) << "failed to get default GPU Node";
+
+    const HsaNodeProperties *nodeProps = m_NodeInfo.GetNodeProperties(defaultGPUNode);
+    bool perQueueResetSupported = nodeProps->Capability.ui32.PerQueueResetSupported;
+
+    if (perQueueResetSupported) {
+        int pipefd[2];
+        pipe(pipefd);
+
+        pid_t childPid = fork();
+
+        if (childPid == 0) {
+            // Child process: submit bad packet with invalid opcode
+            KFDBaseComponentTest::TearDown();
+            KFDBaseComponentTest::SetUp();
+
+            LOG() << "Child ==> Wait on parent to set reset event" << std::endl;
+            char buf;
+            read(pipefd[0], &buf, 1);
+
+            PM4Queue queue;
+            ASSERT_SUCCESS(queue.Create(defaultGPUNode));
+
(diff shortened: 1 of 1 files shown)


cfreeamd (commented):

The two new negative tests (`ComputeQueueInvalidOpcode`, `SDMAQueueInvalidOpcode`) look correct: capability-gated, and the parent correctly verifies the per-queue reset is isolated (no reset event leaks to the parent, healthy queue still completes). Comments below are hygiene nits on the compute-test fork path.

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDNegativeTest.cpp
```diff
@@ -280,3 +280,217 @@ TEST_F(KFDNegativeTest, BasicSDMAReset) {
...
+
+    int defaultGPUNode = m_NodeInfo.HsaDefaultGPUNode();
+    ASSERT_GE(defaultGPUNode, 0) << "failed to get default GPU Node";
+
+    const HsaNodeProperties *nodeProps = m_NodeInfo.GetNodeProperties(defaultGPUNode);
+    bool perQueueResetSupported = nodeProps->Capability.ui32.PerQueueResetSupported;
+
+    if (perQueueResetSupported) {
+        int pipefd[2];
+        pipe(pipefd);
```
cfreeamd, line 300: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `ComputeQueueInvalidOpcode` never closes either end of `pipefd` (in the child or the parent), and `pipe()`'s return value is unchecked. The `SDMAQueueInvalidOpcode` test below closes its unused ends correctly (`close(pipe1[1])`, `close(pipe2[0])`, etc.) — worth matching that here for consistency and to avoid leaking fds into the forked child.

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDNegativeTest.cpp
```diff
@@ -280,3 +280,217 @@ TEST_F(KFDNegativeTest, BasicSDMAReset) {
...
+
+        pid_t childPid = fork();
```
cfreeamd, line 302: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `fork()`'s return isn't checked for -1. On fork failure `childPid == -1` falls into the parent branch and `waitpid(-1, ...)` would wait on an arbitrary child rather than reporting the failure. Low risk in a test, but a guard (`ASSERT_GE(childPid, 0)`) would make the failure explicit.

projects/rocr-runtime/libhsakmt/tests/kfdtest/src/KFDNegativeTest.cpp
```diff
@@ -280,3 +280,217 @@ TEST_F(KFDNegativeTest, BasicSDMAReset) {
...
+            // Child process: submit bad packet with invalid opcode
+            KFDBaseComponentTest::TearDown();
+            KFDBaseComponentTest::SetUp();
+
+            LOG() << "Child ==> Wait on parent to set reset event" << std::endl;
+            char buf;
+            read(pipefd[0], &buf, 1);
+
+            PM4Queue queue;
+            ASSERT_SUCCESS(queue.Create(defaultGPUNode));
```
cfreeamd, line 314: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** `ASSERT_*` inside the forked child only `return`s from the test function — it does not terminate the child process. On failure the child drops out of the test body and re-enters the gtest runtime (duplicate test output, child never calls `exit`). Prefer `_exit()` on child-side failure paths. This matches an existing kfdtest fork idiom, so non-blocking — flagging in case you want to harden it.

PR #8985, opened 2026-07-21
fix(rocjitsu): Retain owners for borrowed C API handles

### LINKED ISSUE #8983: [rocjitsu] Retain owners for borrowed C API handles
## Problem

Two public C API child handles borrow pointers into storage owned by parent handles without retaining those parents:

- `rj_code_object_t` borrows an `AmdGpuCodeObject` owned by `rj_code_executable_t`.
- `rj_code_basic_block_t` borrows a `BasicBlock` owned by `rj_code_basic_block_list_t`.

The API documents the returned child handles as independently owned. A caller that destroys a parent before finishing with its child can therefore dereference freed storage.

## Expected behavior

Each borrowed child handle retains its owning parent and releases it when the child handle is freed. Parent and child handles can then be destroyed in any API-permitted order.

## Test coverage

Add regression coverage that destroys the executable before using its code-object handle and destroys the basic-block list before querying its block handle. Exercise this for every currently supported AMDGPU target.

### LINKED ISSUE #8984: [rocjitsu] Test C API child handles after parent destruction
## Problem

The C API target tests do not verify the documented independent ownership of handles returned by `rj_code_executable_get_code_object()` and `rj_code_basic_block_list_get()`. As a result, parent/child lifetime regressions can pass normal decode and analysis tests.

## Proposed coverage

Extend the supported-target C API test path to:

1. Destroy the executable before decoding through its returned code-object handle.
2. Destroy the basic-block list before querying its returned block handle.
3. Run the sequence for gfx90a, gfx942, gfx950, gfx1200, gfx1201, and gfx1250.

## Related issue

Implementation fix: #8983.

amd-arosa:
## Summary

- retain an executable while a borrowed code-object handle references its storage
- retain a basic-block list while a borrowed block handle references its storage
- test parent-before-child destruction for every supported AMDGPU target

## Testing

- clean Release build of `rocjitsu_tests` and `rocjitsu_shared`
- `GfxCodeObjectTargets.*`: 13/13 passed
- repository pre-commit hooks passed for both commits

Fixes #8983.
Fixes #8984.

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/rj_code.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/rj_code.cpp
@@ -120,6 +120,8 @@ rj_status_t rj_code_executable_get_code_object(const rj_code_executable_t *exec,
 
   *obj = new rj_code_object_t{};
   (*obj)->co = co;
+  (*obj)->parent_exec = const_cast<rj_code_executable_t *>(exec);
+  (*obj)->parent_exec->retain();
   (*obj)->retain();
   return ROCJITSU_STATUS_SUCCESS;
 }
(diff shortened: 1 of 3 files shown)


amd-arosa:
i feel this pr is ugly. and not my choice how to do this. im just curious what the right way is? going to try to make an altentive using std:shared_ptr. or maybe like llvm ScopeExit.h?

kuhar (approved):
We should document both new ownership guarantees in the public C API declarations: a handle returned by `rj_code_executable_get_code_object()` keeps the executable backing storage alive, and one returned by `rj_code_basic_block_list_get()` keeps the list alive. The current `@details` clauses only describe the child handle refcount, so callers cannot tell that destroying the parent first is supported.

LGTM

PR #8986, opened 2026-07-21
fix(rocprofiler-sdk): Address review feedback on hsa_amd_queue_create…

saleelk:
… interception

- Handle partial batch success: process all non-null descs[i].queue entries and return the original status instead of bailing early.
- Emit ROCP_WARNING when device-memory queue falls back to system-memory InterceptQueue; document priority/CU-mask loss as known limitation.
- Skip QueueState registration for non-compute queues (SDMA) whose packet format is incompatible with AQL interposition.
- Document late-attach as unsupported for descriptor-based queue creation.
- Revert host_functions test relaxation in 4 validate.py files.

## Issue Tracking

JIRA ID : AIRUNTIME-0

## Submission Checklist


saleelk:
> 
> 
> * ❌ Unit Test
> 

This is rather a comment and refactor, no test needed

bwelton (approved):

PR #8987, opened 2026-07-21
feat(ci): Using external repo CI configuration script in multi-arch CI 

geomin12:
After https://github.com/ROCm/TheRock/pull/6660 landed, we are now using the external configure ci script to maintain in one central location. We are incorporating this script in multi-arch CI!

Working here: https://github.com/ROCm/rocm-systems/actions/runs/29869549929/job/88766078180

ISSUE ID: https://github.com/ROCm/TheRock/issues/3343

### DIFF
--- a/.github/workflows/therock-multi-arch-ci.yml
+++ b/.github/workflows/therock-multi-arch-ci.yml
@@ -85,7 +85,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           repository: ROCm/TheRock
-          ref: users/geomin12/external-repo-change
+          ref: main
           path: _therock
           sparse-checkout: build_tools/github_actions
           sparse-checkout-cone-mode: true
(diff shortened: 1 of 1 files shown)


tony-davis (approved):
Approving this. Same clean consolidation as the rocm-libraries side; moving to the shared `configure_external_repo_ci.py` and dropping the local script plus `pydantic` is the right call, and your linked run is green with the `Configure CI` step actually running. The `base...head` three-dot compare keeps the old merge-base behavior, so PR diffs are unchanged.

A few non-blocking reservations (and I want this in so you can keep iterating and testing):

One, `push` events will now run everything. The step only passes the `pull_request.*` SHAs, which are empty on push, so the script falls through to run-all. It fails safe (over-tests, never under-tests), so no risk; just confirming that's intended. If you want push to stay selective, `--base-sha ${{ github.event.before }}` on push would get you there. Your call.

Two, the `Configure CI` step runs `gh api` without a visible `GH_TOKEN`/`GITHUB_TOKEN` in scope (only a downstream job has one). It worked in your run, so something's providing it, but unauthenticated compare API is rate-limited (60/hr/IP across hosted runners), so it could get flaky. Adding `env: GH_TOKEN: ${{ github.token }}` to that step is cheap insurance.

Three, minor: checking out TheRock at `ref: main` couples this to a moving branch, so a breaking change to the script could break the workflow without warning. Worth pinning to a tag/SHA down the road.

Approving so you're unblocked.

PR #8988, opened 2026-07-21
fix(rocjitsu): Use shared ownership for C API child backing storage

### LINKED ISSUE #8983: [rocjitsu] Retain owners for borrowed C API handles

### LINKED ISSUE #8984: [rocjitsu] Test C API child handles after parent destruction

amd-arosa:
## Summary
- keep executable and decoded-block backing storage alive with `std::shared_ptr`
- use aliasing `std::shared_ptr` instances for borrowed code-object and basic-block handles
- remove raw parent-handle pointers and manual parent `retain()`/`release()` coupling
- preserve the opaque public C ABI and existing handle refcount semantics
- test child handles after destroying their parents for every supported target

This is a modern C++ alternative to #8985.

## Validation
- clean amdclang++ Release build of `rocjitsu_shared` and `rocjitsu_tests`
- `GfxCodeObjectTargets.*`: 13/13 passed
- clean GCC 13 ASan+UBSan build
- sanitizer `GfxCodeObjectTargets.*`: 13/13 passed, including leak detection
- repository pre-commit hooks passed

Fixes #8983
Fixes #8984

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/rj_code.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/rj_code.cpp
@@ -73,7 +73,7 @@ rj_status_t rj_code_executable_create(const char *path, rj_code_executable_t **e
   if (!path || !exec)
     return ROCJITSU_STATUS_INVALID_ARGUMENT;
 
-  auto executable = std::make_unique<Executable>(path);
+  auto executable = std::make_shared<Executable>(path);
   if (!executable->is_valid())
     return ROCJITSU_STATUS_INVALID_CODE_OBJECT;
 
@@ -114,12 +114,12 @@ rj_status_t rj_code_executable_get_code_object(const rj_code_executable_t *exec,
   if (!exec || !exec->exec || !obj)
     return ROCJITSU_STATUS_INVALID_ARGUMENT;
 
-  auto *co = const_cast<Executable *>(exec->exec.get())->code_object(target, index);
+  auto *co = exec->exec->code_object(target, index);
   if (!co)
     return ROCJITSU_STATUS_INVALID_ARGUMENT;
 
   *obj = new rj_code_object_t{};
-  (*obj)->co = co;
+  (*obj)->co = std::shared_ptr<AmdGpuCodeObject>(exec->exec, co);
   (*obj)->retain();
   return ROCJITSU_STATUS_SUCCESS;
 }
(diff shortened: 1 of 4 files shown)


amd-arosa:
how much slower is make_shared vs make_unique ?

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/rj_code.cpp
```diff
@@ -208,7 +208,8 @@ rj_status_t rj_code_basic_block_list_create(rj_code_object_t *obj, rj_code_targe
     return ROCJITSU_STATUS_INVALID_ARGUMENT;
 
   auto owned = std::make_unique<rj_code_basic_block_list_t>();
-  owned->blocks = BasicBlock::build(*obj->co, *decoder, arch);
+  owned->blocks = std::make_shared<rj_code_basic_block_list_t::Storage>(
```
kuhar, line 211: This would be safer if the shared owner behind each block also retained `obj->co`. `BasicBlock::build()` decodes from code-object section storage, and VOPD leaves `Instruction::raw_encoding_` pointing at that input buffer (`rdna4/vopd.cpp:217`, similarly `gfx1250/vopd.cpp:319`). Right now the aliasing block pointer owns only the decoded-block vector, so destroying the code object leaves that pointer dangling. Could we make `Storage` an aggregate containing both `obj->co` and the block vector, then add an ASAN test that destroys `obj` and the list before examining a VOPD-backed block?

kuhar (commented):
Could we keep the backing storage uniquely owned and have borrowed child C handles retain/release their owning parent handles, as in #8985, rather than introducing `shared_ptr` here? The public API already exposes explicit `RefCounted` retain/release/destroy semantics, so `shared_ptr` creates a second, hidden reference count for the same lifetime relationship. The ownership graph is hierarchical, so parent retention keeps one lifetime model aligned with the C API. If the concern is the manual cleanup boilerplate in #8985, a small internal RAII wrapper around the intrusive retain/release operations could remove that boilerplate without introducing shared ownership.

PR #8990, opened 2026-07-21
fix(ci): rewrite nightly workflow to match rocm-libraries pattern

geomin12:
- Replace workflow_dispatch trigger approach with direct TheRock workflow calls
- Add gfx125X (MI455) with custom family_overrides for test runner and limited tests
- Lock TheRock ref once in setup and reuse across all downstream jobs
- Delete therock-ci-nightly.yml (replaced by therock-multi-arch-ci-nightly.yml)

This fixes the issue where test_filter:comprehensive label was incorrectly parsed as a component filter, causing all tests except sanity to be skipped.


.github/workflows/therock-multi-arch-ci-nightly.yml
```diff
@@ -1,37 +1,175 @@
...
+  group: ${{ github.workflow }}-${{ github.event.number || github.sha }}
+  cancel-in-progress: true
+
 jobs:
-  multi-arch-ci-nightly-trigger:
-    runs-on: ubuntu-latest
+  setup:
+    uses: ROCm/TheRock/.github/workflows/setup_multi_arch.yml@main
+    with:
+      build_variant: "release"
```
tony-davis, line 55: Two behaviors ride on `release_type`, which is defaulting to `ci` here. First, comprehensive tests only kick in on the `schedule` trigger (priority 4 in `_determine_test_type`); a manual `workflow_dispatch` of this same file falls all the way through to `quick`, so the dispatch button won't reproduce the nightly it's attached to. Second, `ci` produces dev-style package versions and uses the ci artifact bucket and ccache namespace, not nightly. Setting `release_type: "nightly"` fixes both at once (priority 3 returns `comprehensive` on any trigger, and versioning/bucket become nightly). The same gap exists in the rocm-libraries nightly, so worth mirroring there.

```suggestion
      build_variant: "release"
      # nightly release_type makes tests comprehensive on any trigger (schedule or
      # manual dispatch), and gives artifacts nightly versioning/bucket rather than ci/dev.
      release_type: "nightly"
```

.github/workflows/therock-multi-arch-ci-nightly.yml
```diff
@@ -1,37 +1,175 @@
...
 
 permissions:
-  actions: write
   contents: read
 
+concurrency:
+  # A PR number if a pull request and otherwise the commit hash. This cancels
+  # queued and in-progress runs for the same PR (presubmit) or commit
+  # (postsubmit). The workflow name is prepended to avoid conflicts between
+  # different workflows.
```
tony-davis: Small thing: this comment is inherited from the PR/presubmit workflows, but this file only triggers on `schedule` and `workflow_dispatch`, never on a PR. So `github.event.number` is always empty and this effectively groups by commit. Might as well have the comment say that.

```suggestion
  # This workflow only runs on schedule / workflow_dispatch (never on PRs), so
  # github.event.number is always empty and this effectively groups by commit.
  # The workflow name is prepended to avoid collisions with other workflows.
```

.github/workflows/therock-multi-arch-ci-nightly.yml
```diff
@@ -1,37 +1,175 @@
...
+      # concrete SHA (needs.setup.outputs.ref) for every downstream job.
+      ref: ${{ inputs.therock_ref_override || 'main' }}
+      # TEMPORARY: MI455 bringup - pass family_overrides to configure test runner and limited tests
+      # TODO(geomin12): Remove family_overrides once MI455 is in therock-ci-config
+      external_repo: >-
+        {
+          "repository": "${{ github.repository }}",
+          "ref": "${{ github.sha }}",
+          "family_overrides": {
+            "gfx125x": {
```
tony-davis, line 75: FYI, non-blocking, and really a note for the TheRock side: this override takes effect only because `gfx125x` already exists in `amdgpu_family_matrix.py`, so it merges into that entry. If someone later tries to introduce a brand-new family purely through `external_repo`, it won't work. `select_targets` validates requested families against the un-overridden matrix before `_apply_external_family_overrides` runs, so it would `ValueError` first. Worth a one-line comment on the TheRock helper so the add-new-family branch isn't mistaken for a working path. Nothing to change in this PR.

tony-davis (changes requested):
Thanks Geo! This is a clean port of the rocm-libraries nightly, and the MI455 override is wired up correctly. I traced it through `setup_multi_arch.yml` and `configure_multi_arch_ci.py`, and since `gfx125x` already exists in `amdgpu_family_matrix.py` the override merges into that entry as intended (test runner, limited test set, and the `gfx1250` fetch target all land). The ref-locking also does what the comments promise. My requests are about test scope and artifact typing, plus one coverage question. Most of these are shared with the rocm-libraries nightly, so if we fix them here we should probably mirror them there too.

- (1 + 2) `release_type` is left at its default (`ci`), and two things ride on that. Comprehensive tests only trigger on the `schedule` event, so a manual `workflow_dispatch` of this file (which the `therock_ref_override` input invites) falls through to `quick`. And `ci` gives dev-style versioning plus the ci bucket/ccache namespace rather than nightly. Passing `release_type: "nightly"` fixes both. See the inline suggestion. If these nightlies are pure coverage runs nobody consumes, the versioning half is arguable, but I'd still set it so the dispatch button reproduces the real nightly.
- (3) The deleted `therock-ci-nightly.yml` had a dedicated `therock-rccl-ci-linux` job with its own cmake (`THEROCK_ENABLE_COMM_LIBS`, `THEROCK_ENABLE_MPI`, ...). I can't tell from the diff whether RCCL tests still run under the multi-arch comprehensive path for gfx94X/gfx950. Its schedule was already commented out, so this may already be dead nightly coverage, but can you confirm RCCL is either covered by the multi-arch flow or is a conscious drop during bringup?
- (4) Minor, and really a TheRock-side note: the `family_overrides` mechanism only works here because `gfx125x` is already in the matrix. Details inline.
- (5) The `concurrency` comment is copy-pasted from a PR workflow and talks about PR numbers, but this file never runs on PRs. Cosmetic, inline suggestion.

None of this breaks the happy path; the scheduled nightly should run comprehensive across gfx94X/gfx950/gfx125X plus gfx1151 as intended.

tony-davis (approved):
Approving. The scheduled path is right, and replacing a seven-second trigger job that reported green no matter what with a workflow whose status actually means something is a real win. A few things below, but none of them should hold this up.

First, withdrawing my RCCL question. I went and checked instead of making you do it. The old `therock-ci-nightly.yml` already had its `schedule:` commented out, so it only ever ran on manual dispatch, and `therock-ci.yml` still carries the same `therock-rccl-ci-linux` job over gfx94X-dcgpu and gfx950-dcgpu plus the same `projects` dispatch input. Nothing is lost. Sorry for making you carry that one.

I'll also take the `release_type` revert. If nobody consumes these artifacts then dev versioning and the ci bucket are fine, and I shouldn't have bundled versioning with test scope in one comment. (Out of curiosity, who gave that feedback? Would be good to have it on the thread so nobody re-litigates it in a month.) The one residual is that a manual dispatch of this file falls through to `quick` while the cron correctly gets comprehensive. I checked the obvious workaround and it doesn't work: passing `test_filter:comprehensive` through `linux_test_labels` sets the right test type, but `configure_multi_arch_ci.py` copies the raw labels straight into the downstream `test_labels` output, which is exactly the bug you're fixing here. So a comment saying "manual dispatch runs quick" is probably enough. Not worth another round trip on its own.

Since the bot is still failing the description check (needs a JIRA or ISSUE reference), you have to push again regardless, so here are two one-liners I'd love to ride along with it.

Drop `secrets: inherit` from both build/test jobs. I chased this one down properly: nothing in the Linux or Windows multi-arch build/test chain reads a named secret, and every `secrets.X` in TheRock lives in unrelated bot and benchmark workflows. So it isn't buying us anything today, and it does hand every rocm-systems secret to a workflow definition floating on `ROCm/TheRock@main`. My worry is drift, not today.

Match the concurrency group to `therock-multi-arch-ci-asan-nightly.yml`, which uses just `${{ github.workflow }}`. Right now we include `github.sha`, which makes the behavior depend on whether develop moved overnight: same tip and the runs cancel each other, moved tip and two multi-hour nightlies overlap. That seems bad for the single MI455 runner either way, and the ASAN nightly already picked a consistent answer.

Tiny one while you're in there: the TEMPORARY MI455 header mentions "Container options with ulimit for RCCL compatibility," but there aren't any container options in the JSON, and it names a `gfx125x_family_overrides` that doesn't exist anywhere.

Last thing, and it's a question rather than a review comment. That ASAN nightly is a useful mirror for this change. It quietly `startup_failure`'d with zero jobs on Aug 8 and Aug 9, which lines up with a TheRock main change landing and then getting reverted. Nobody noticed. Under the old design this workflow was green by construction so that cost us nothing; after your change the status is real, which is the whole point, but it only helps if someone is looking. Is that you for now, or should we get an alert wired up? Happy to take that one off your plate if you want.

PR #8994, opened 2026-07-21
[rocjitsu] checked private-spill arithmetic and multi-producer db rewriting

atgutier:
## Motivation

Two fail-open correctness bugs surfaced in re-review of the DBT/virtual-LDS path.

First, the per-lane private segment is initialized from the guest descriptor's private size, which can be near UINT32_MAX. The semantic spill reservations and the descriptor private/group addends extended it with unchecked uint32_t math, so a kernel with private_segment_fixed_size = 0xfffffff8 aligned up to 16 wraps to 0. The virtual-LDS backing-pointer spill then lands at offset 0 and the generated save/restore instructions overwrite the guest's low private-memory region while translation still reports success -- silent scratch corruption.

Second, the non-intercept doorbell fallback rewrote the full published packet range only for HSA_QUEUE_TYPE_SINGLE. A multi-producer queue can reserve and publish several consecutive packets and ring once with the final packet id. If an earlier packet needs virtual-LDS rewriting and the doorbell names a later one, only the named packet was rewritten and next_packet_id was advanced past the earlier one -- so the oversized packet reached the command processor unrevised and the background scanner (clamped to next_packet_id) skipped it too. This path is used whenever intercept queue creation/registration is unavailable.

## Technical Details

Private-spill / descriptor-addend overflow (translation_rule.h, kernel_descriptor_translator.cpp, cdna4_to_cdna3_virtual_lds.cpp, binary_translator.cpp):

- reserve_persistent_semantic_spill_dwords / reserve_semantic_spill_dwords now return std::optional<uint32_t>, computing the aligned base and one-past-end with checked 64-bit math via a shared semantic_spill_reservation() helper; anything exceeding the 32-bit private-size field yields nullopt.
- The failure propagates: assign_virtual_lds_spill_offsets returns optional and its three lowering call sites return ExpandResult::failed; the persistent-spill site in binary_translator.cpp raises a ResourceLimit kernel failure (skip or leave-unchanged). The descriptor private/group addends use util::checked_add and append_descriptor_error on overflow.

Multi-producer doorbell rewriting (rj_hsa_dbt_hooks.cpp):

- Rewrite the whole newly-published range [next_packet_id, doorbell_id] for every queue type when the span fits the ring, instead of only for single-producer queues. note_packet_ready advances the contiguous frontier only across in-order ready packets, so an out-of-order producer's unready hole is never skipped; a doorbell jump larger than the ring still falls back to rewriting the named packet and lets the scanner cover the gap.

## Issue Tracking

#8939

## Test Plan

Run CI tests.
Boundary-overflow unit tests for the spill reservations. Multi-producer fallback doorbell unit test (verified to fail pre-fix).

## Test Result

- Semantic-scratch overflow tests (persistent + transient reservations at 0xfffffff8 fail closed) and descriptor/DBT suites: 112/112.
- HSA hook suite, including the new HSA_QUEUE_TYPE_MULTI two-packet doorbell test (packet 0 rewritten to the virtual descriptor): 46/46.
- Full suite: 1804/1804.

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/hooks/rj_hsa_dbt_hooks.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/hooks/rj_hsa_dbt_hooks.cpp
@@ -53,6 +53,7 @@
 #include <mutex>
 #include <new>
 #include <optional>
+#include <set>
 #include <shared_mutex>
 #include <signal.h>
 #include <span>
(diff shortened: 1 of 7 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/hooks/rj_hsa_dbt_hooks.cpp
```diff
@@ -2286,14 +2286,24 @@ class VirtualLdsDispatchQueueRegistry {
...
+    // consecutive packets and then ring once with the final packet ID. Rewriting
+    // only the doorbell-named packet would let an earlier still-unrewritten
+    // predecessor (e.g. a virtual-LDS dispatch at next_packet_id) reach the CP
+    // unrevised, and advancing next_packet_id past it would make the scanner skip
+    // it too. So rewrite the whole newly-published range [next_packet_id,
+    // packet_id]. note_packet_ready only advances the contiguous frontier across
+    // packets that were actually ready in order, so an out-of-order producer that
+    // leaves an unready hole does not advance the cursor past it -- the scanner
+    // still covers the hole later. The range must fit the ring to bound the scan.
+    if (packet_id >= state.next_packet_id && packet_id - state.next_packet_id < state.queue->size) {
```
kuhar, line 2324: This seems like a potential blocker: the frontier can be left permanently stale, which reintroduces exactly the missed-rewrite (silent oversized-launch) bug this PR sets out to kill. Verified on a size-4 multi-producer queue:

- next_packet_id = 0. Producer rings `packet_id = 3` while packet 0 is still INVALID. This range rewrites 1..3 but `note_packet_ready` never advances the cursor (each observed id != next_packet_id == 0), so next_packet_id stays 0.
- Packet 0 later rings. Range is `[0, 1)`, so the cursor advances only to 1 -- it does NOT catch up across 1..3, which were already ready and doorbelled.
- After 0..3 retire, a producer publishes 4 and 5 and rings once with 5. Now `packet_id - next_packet_id == 5 - 1 == 4 == queue->size`, so the range guard on line 2299 is false and the fallback rewrites ONLY packet 5, then sets next_packet_id = 6. The background scanner is clamped to `scan_begin >= next_packet_id` (line 2874), so packet 4 is never revisited. If packet 4 needs virtual-LDS rewriting it reaches the CP unrevised -> oversized (host-faulting) launch.

Root cause: `note_packet_ready` only steps the cursor when `packet_id == next_packet_id` in a single call; it never re-scans forward across successors observed ready in a prior doorbell, so closing an earlier hole strands the frontier below the already-ready suffix.

Fix direction: when a hole closes, advance the contiguous cursor across all already-observed-ready successors (track per-slot readiness / a published frontier), or keep a separate highest-doorbelled frontier so a later range's lower bound cannot be made stale by an out-of-order ring. Please add a regression covering the hole-close-then-next-batch sequence above -- it's the case the current `MultiProducerDoorbell...` test does not reach.

kuhar (commented):
It would be useful to add one propagation-level overflow regression. The new `SemanticSpillFrame` cases only check that the reservation helper returns `nullopt`; they do not exercise the new `ResourceLimit` failure in `binary_translator.cpp:1207` or an `ExpandResult::failed` path in `cdna4_to_cdna3_virtual_lds.cpp`. One near-`UINT32_MAX` virtual-LDS translation that asserts a fail-closed result would cover the observable behavior without testing every call site.

Groverkss (approved):
lgtm except @kuhar comment

PR #9003, opened 2026-07-21
fix(rocjitsu): Fix local memory fallback for VMID misses

RattataKing:
## Motivation

Rocjitsu may reserve the virtual-address range for a simulated VRAM allocation using a `PROT_NONE` CPU mapping.
In local mode, if a VMID page-table lookup misses, rocjitsu may incorrectly cast the GPU virtual address to a CPU pointer, and causing a seg fault.

## Technical Details

- Create permanent host backing for each VRAM allocation.
- Resolve GPU VA to their owned backing in the local model.
- Prevent inaccessible VRAM allocations from falling back to CPU pointers.
- Preserve backing data across CPU and GPU map/unmap operations.
- Share the permanent backing during IPC export and import.
- Reject spanning `munmap()` ranges that overlap event or doorbell pages.

## Issue Tracking

GitHub issue: https://github.com/ROCm/rocm-systems/issues/9933

## Test Plan

Run CI and RocBLAS SGEMM test from the test corpus.

## Test Result

Passed

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h
@@ -26,6 +26,7 @@
 #include <shared_mutex>
 #include <unordered_map>
 #include <unordered_set>
+#include <utility>
 #include <vector>
 
 #include <sys/types.h> // pid_t
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h
@@ -16,10 +16,13 @@
 #include "rocjitsu/vm/amdgpu/mtype.h"
 #include "util/unique_handle.h"
 
+#include <algorithm>
 #include <atomic>
 #include <cassert>
 #include <cstdint>
+#include <map>
 #include <mutex>
+#include <optional>
 #include <shared_mutex>
 #include <unordered_map>
 #include <unordered_set>
@@ -79,20 +82,64 @@ public:
   struct GpuAllocation {
     uint64_t gpu_va = 0;
     uint64_t size = 0;
+    uint64_t backing_size = 0;
     void *host_ptr = nullptr;
     uint32_t flags = 0;
+    amdgpu::Mtype backing_mtype = amdgpu::Mtype::RW;
     uint64_t handle = 0;
     int memfd = -1;
     uint32_t gpu_id = 0;
     bool user_va = false;
     bool imported = false;
     int dmabuf_fd = -1;
+    bool permanent_backing = false;
+    bool mapped_to_gpu = false;
+    bool backing_authoritative = false;
+    // Page-aligned allocation-relative intervals already consumed from an
+    // existing MAP_FIXED client reservation. Keys are inclusive starts and
+    // values are exclusive ends.
+    std::map<uint64_t, uint64_t> imported_client_extents;
     // True when the driver created host_ptr (mmap it itself) and must munmap it
     // on teardown. False for caller-owned pages (e.g. reused MAP_FIXED pages
     // from the thunk) that the driver must never unmap, since unmapping them
     // races with the owning process still accessing the memory.
     bool host_ptr_owned = false;
   };
+
+  /// @brief A stable lookup result for allocation-owned GPU backing.
+  /// @details An inaccessible result identifies a managed address that must not
+  /// fall through to identity-pointer translation. The guard keeps accessible
+  /// backing alive until the caller finishes the memory operation.
+  struct ResolvedBacking {
+    uint8_t *address = nullptr;
+    uint8_t *page_base = nullptr;
+    uint8_t *range_base = nullptr;
+    uint64_t range_size = 0;
+    amdgpu::Mtype mtype = amdgpu::Mtype::RW;
+    bool gpu_accessible = false;
+    std::shared_lock<std::shared_mutex> guard;
+  };
+
+  [[nodiscard]] static bool is_valid_gpu_range(uint64_t gpu_va, uint64_t size) {
+    return size != 0 && gpu_va <= UINT64_MAX - (size - 1);
+  }
+
+  /// @brief Whether an allocation already covers any page in a proposed range.
+  /// @details The caller must hold alloc_mutex_. GPU mappings are page-granular,
+  /// so byte-disjoint ranges sharing a page are treated as overlapping.
+  [[nodiscard]] bool overlaps_allocation_locked(uint64_t gpu_va, uint64_t size) const {
+    assert(is_valid_gpu_range(gpu_va, size));
+    const uint64_t first_page = gpu_va >> kPageShift;
+    const uint64_t end_page = ((gpu_va + size - 1) >> kPageShift) + 1;
+    for (const auto &[handle, alloc] : allocations_) {
+      (void)handle;
+      const uint64_t alloc_first_page = alloc.gpu_va >> kPageShift;
+      const uint64_t alloc_end_page = ((alloc.gpu_va + alloc.size - 1) >> kPageShift) + 1;
+      if (first_page < alloc_end_page && alloc_first_page < end_page)
+        return true;
+    }
+    return false;
+  }
 
   /// @brief Memory policy descriptor.
   struct MemoryPolicy {
@@ -254,6 +301,41 @@ public:
   /// @brief Return the mutation counter used by GpuMemory translation caches.
   const uint64_t *page_table_generation() const { return &page_table_generation_; }
 
+  /// @brief Resolve a managed GPU range to allocation-owned host backing.
+  /// @details Unmapped allocations still return a result with
+  /// gpu_accessible=false, preventing unsafe identity-pointer fallback. The
+  /// returned guard prevents teardown while accessible backing is in use.
+  [[nodiscard]] std::optional<ResolvedBacking> resolve_backing(uint64_t gpu_va, size_t size) const {
+    if (size == 0 || gpu_va > UINT64_MAX - (static_cast<uint64_t>(size) - 1))
+      return std::nullopt;
+
+    std::shared_lock lock(backing_ranges_mutex_);
+    auto it = std::upper_bound(
+        backing_ranges_.begin(), backing_ranges_.end(), gpu_va,
+        [](uint64_t addr, const BackingRange &range) { return addr < range.first; });
+    if (it == backing_ranges_.begin())
+      return std::nullopt;
+    --it;
+
+    const uint64_t offset = gpu_va - it->first;
+    if (offset >= it->size || static_cast<uint64_t>(size) > it->size - offset)
+      return std::nullopt;
+
+    const uintptr_t address_value = reinterpret_cast<uintptr_t>(it->host_base) + offset;
+    return ResolvedBacking{reinterpret_cast<uint8_t *>(address_value),
+                           reinterpret_cast<uint8_t *>(address_value - (gpu_va & (kPageSize - 1))),
+                           it->host_base,
+                           it->size,
+                           it->mtype,
+                           it->gpu_accessible,
+                           std::move(lock)};
+  }
+
+  /// @brief Republish allocation-backed address ranges after a state change.
+  /// @details The caller must hold alloc_mutex_. Acquiring the unique range lock
+  /// waits for in-flight GPU accesses before backing can be unmapped.
+  void refresh_backing_ranges_locked() { publish_backing_ranges_locked(); }
+
   mutable std::shared_mutex page_table_mutex_;
   PageTable page_table_;
 
@@ -271,6 +353,9 @@ public:
   std::mutex op_mutex_;
   mutable std::mutex alloc_mutex_;
   std::unordered_map<uint64_t, GpuAllocation> allocations_;
+  // Page-aligned CPU alias intervals. Keys are inclusive host addresses and
+  // values are exclusive ends.
+  std::map<uintptr_t, uintptr_t> cpu_mappings_;
   uint64_t next_handle_ = 1;
   uint64_t next_gpu_va_;
 
@@ -309,6 +394,42 @@ private:
   ///          this generation; all reads and writes occur while holding
   ///          page_table_mutex_, so the counter itself does not need atomics.
   uint64_t page_table_generation_{1};
+
+  struct BackingRange {
+    uint64_t first = 0;
+    uint64_t size = 0;
+    uint8_t *host_base = nullptr;
+    amdgpu::Mtype mtype = amdgpu::Mtype::RW;
+    bool gpu_accessible = false;
+  };
+
+  void publish_backing_ranges_locked() {
+    std::vector<BackingRange> ranges;
+    ranges.reserve(allocations_.size());
+    for (const auto &[handle, alloc] : allocations_) {
+      (void)handle;
+      if (!alloc.permanent_backing || !alloc.host_ptr)
+        continue;
+      assert(is_valid_gpu_range(alloc.gpu_va, alloc.size));
+      ranges.push_back({alloc.gpu_va, alloc.size, static_cast<uint8_t *>(alloc.host_ptr),
+                        alloc.backing_mtype, alloc.mapped_to_gpu});
+    }
+
+    std::sort(ranges.begin(), ranges.end(), [](const BackingRange &lhs, const BackingRange &rhs) {
+      return lhs.first < rhs.first;
+    });
+    for (size_t i = 1; i < ranges.size(); ++i) {
+      assert(ranges[i - 1].first <= ranges[i].first);
+      assert(ranges[i - 1].size <= ranges[i].first - ranges[i - 1].first &&
+             "overlapping GPU allocation backings");
+    }
+
+    std::unique_lock lock(backing_ranges_mutex_);
+    backing_ranges_ = std::move(ranges);
+  }
+
+  mutable std::shared_mutex backing_ranges_mutex_;
+  std::vector<BackingRange> backing_ranges_;
 };
 
 } // namespace rocjitsu
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/remote_driver.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/remote_driver.cpp
@@ -8,13 +8,16 @@
 #include "rocjitsu/kmd/linux/kfd_ioctl_utils.h"
 #include "rocjitsu/kmd/linux/rpc.h"
 
+#include <algorithm>
 #include <cassert>
 #include <cerrno>
 #include <chrono>
 #include <cstring>
 #include <fcntl.h>
+#include <iterator>
 #include <limits>
 #include <linux/mman.h>
+#include <map>
 #ifndef MADV_POPULATE_WRITE
 #define MADV_POPULATE_WRITE 23
 #endif
@@ -29,10 +32,59 @@
 #include <sys/un.h>
 #include <thread>
 #include <unistd.h>
+#include <utility>
 #include <vector>
 
 namespace rocjitsu {
 namespace {
+
+template <typename Address>
+void merge_interval(std::map<Address, Address> &intervals, Address begin, Address end) {
+  assert(begin < end);
+
+  typename std::map<Address, Address>::iterator it = intervals.lower_bound(begin);
+  if (it != intervals.begin()) {
+    typename std::map<Address, Address>::iterator previous = std::prev(it);
+    if (previous->second >= begin) {
+      begin = previous->first;
+      end = std::max(end, previous->second);
+      it = intervals.erase(previous);
+    }
+  }
+
+  while (it != intervals.end() && it->first <= end) {
+    end = std::max(end, it->second);
+    it = intervals.erase(it);
+  }
+  intervals.emplace(begin, end);
+}
+
+std::vector<std::pair<uint64_t, uint64_t>>
+find_uncovered_extents(const std::map<uint64_t, uint64_t> &covered, uint64_t begin, uint64_t end) {
+  std::vector<std::pair<uint64_t, uint64_t>> uncovered;
+  uint64_t cursor = begin;
+  std::map<uint64_t, uint64_t>::const_iterator it = covered.upper_bound(begin);
+  if (it != covered.begin()) {
+    --it;
+  }
+
+  for (; it != covered.end() && cursor < end; ++it) {
+    if (it->second <= cursor) {
+      continue;
+    }
+    if (it->first >= end) {
+      break;
+    }
+    if (it->first > cursor) {
+      uncovered.emplace_back(cursor, std::min(it->first, end));
+    }
+    cursor = std::max(cursor, it->second);
+  }
+  if (cursor < end) {
+    uncovered.emplace_back(cursor, end);
+  }
+  return uncovered;
+}
 
 constexpr bool has_embedded_pointers(unsigned long request) {
   switch (canonical_ioctl_request(request)) {
@@ -128,7 +180,7 @@ RemoteDriver::MemfdLookup RemoteDriver::find_memfd_for_addr(void *addr, size_t l
       // Return a DUP of the memfd, not the stored fd, so the caller owns a
       // descriptor whose lifetime is independent of this RemoteDriver. Without
       // this, a concurrent last-close -> RemoteDriver::close() (which closes the
-      // stored handle_memfds_ fds) could close the fd between this return and the
+      // stored allocation fds) could close the fd between this return and the
       // caller's ftruncate/fallocate/mmap, operating on a closed/reused fd. The
       // dup is taken under rpc_mutex_, the same lock close() holds, so the stored
       // fd is guaranteed still open here. The caller MUST close the returned fd.
@@ -156,9 +208,11 @@ RemoteDriver::RemoteDriver(int sock_fd) : sock_(sock_fd) {
 }
 
 RemoteDriver::~RemoteDriver() {
-  for (auto &[handle, fd] : handle_memfds_) {
-    if (fd >= 0)
-      syscall(SYS_close, fd);
+  for (const auto &[handle, allocation] : remote_allocations_) {
+    (void)handle;
+    if (allocation.memfd >= 0) {
+      syscall(SYS_close, allocation.memfd);
+    }
   }
   if (sock_ >= 0)
     syscall(SYS_close, sock_);
@@ -267,11 +321,13 @@ int RemoteDriver::close() {
       sock_ = -1;
     }
 
-    for (auto &[handle, fd] : handle_memfds_) {
-      if (fd >= 0)
-        syscall(SYS_close, fd);
+    for (const auto &[handle, allocation] : remote_allocations_) {
+      (void)handle;
+      if (allocation.memfd >= 0) {
+        syscall(SYS_close, allocation.memfd);
+      }
     }
-    handle_memfds_.clear();
+    remote_allocations_.clear();
     alloc_ranges_.clear();
     // Deliberately do NOT clear the handshake metadata (topology_path_,
     // drm_path_, gpu_info_, has_gpu_info_). It is written once during open() and
@@ -613,10 +669,15 @@ int RemoteDriver::send_ioctl(unsigned long request, void *arg) {
     }
   }
 
-  auto register_allocation = [&](uint64_t handle, uint64_t va_addr, uint64_t size, int memfd) {
-    handle_memfds_[handle] = memfd;
-    if (va_addr != 0 && size > 0)
+  auto register_allocation = [&](uint64_t handle, uint64_t va_addr, uint64_t size, int memfd,
+                                 bool backing_authoritative) {
+    RemoteAllocation allocation;
+    allocation.memfd = memfd;
+    allocation.backing_authoritative = backing_authoritative;
+    remote_allocations_[handle] = std::move(allocation);
+    if (va_addr != 0 && size > 0) {
       alloc_ranges_.push_back({va_addr, size, memfd});
+    }
   };
 
   auto promote_userptr = [&](uint64_t va_addr, uint64_t size, int memfd) {
@@ -658,7 +719,7 @@ int RemoteDriver::send_ioctl(unsigned long request, void *arg) {
     auto *alloc_args = static_cast<kfd_ioctl_alloc_memory_of_gpu_args *>(arg);
     if (num_fds > 0 && received_fds[0] >= 0) {
       register_allocation(alloc_args->handle, alloc_args->va_addr, alloc_args->size,
-                          received_fds[0]);
+                          received_fds[0], false);
 
       if (alloc_args->flags & KFD_IOC_ALLOC_MEM_FLAGS_USERPTR)
         promote_userptr(alloc_args->va_addr, alloc_args->size, received_fds[0]);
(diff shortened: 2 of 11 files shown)


atgutier (commented):
Here are some agent reviews that I think are relevant:

## MAJOR

  [MAJOR, merge-blocking] lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h:320 — Local GPU-only memory
  now performs a failing syscall for every access

  Reasoning: Process-scoped page-table misses no longer use local identity passthrough. Combined with
  simulated_kfd.cpp:374, which registers the local process PID, every such load or store first calls
  process_vm_readv() or process_vm_writev().

  For the PROT_NONE GPUVA reservations this PR intends to support, those calls predictably fail with
  EFAULT. The implementation then emits an unconditional, mutex-protected warning at
  gpu_memory.h:351/366 before finally reaching sparse memory. Hot kernels therefore incur a failed
  kernel syscall and warning construction for every unresolved memory transaction.

  Evidence:

  - gpu_memory.h:320-334 removes process-VMID raw passthrough.
  - gpu_memory.h:343-370 probes every miss with process_vm_*.
  - simulated_kfd.cpp:374 activates those probes for local mode.
  - The baseline that produced merge base bec9696b5cc4 passed gfx942 hip_matmul_matvec.m256_n1_k1024
    in 1.55 seconds.

  - At PR head 6abb4a8ad103, that exact workload timed out at 15 seconds, then timed out again during
    an isolated 30-second retry. The head’s test (release) check consequently failed.

  - The new test at tests/simulated_kfd_test.cpp:97 only reads and writes an accessible stack value.
    Both syscalls succeed, so it never exercises the intended PROT_NONE → sparse memory path or its
    hot-loop cost.

  Suggested fix: represent GPU-only/sparse reservation ranges explicitly and send those addresses
  directly to sparse memory. Host/userptr mappings should continue to install normal page-table
  entries. If generic client probing remains necessary, cache the page/range classification and
  invalidate it during map, unmap, and process teardown. Expected EFAULT probes must not generate an
  unconditional warning per access.

  Add a regression covering a real PROT_NONE reservation with repeated accesses, then rerun both the
  target gfx1201 rocBLAS case and the failing gfx942 matvec case.

  Confidence: high.

 MAJOR (confirm before merge)                                                      10:16:18 [12/1899]

  Pointer-returning callers lost the local-mode identity fallback with no client-memory replacement —
  gpu_memory.h:86 (resolve_host_ptr) → command_processor.cpp:1793 (resolve_sdma_ptr),
  completion_tracker.cpp (fire_queue_idle_signal), find_host_range

  The deleted trailing branch was the only thing making an unmapped-but-host-valid GPUVA
  dereferenceable via the pointer APIs in local mode. The value accessors have a client-memory
  fallback; the pointer path does not — it now returns nullptr where it used to return an identity
  pointer. I confirmed resolve_sdma_ptr still assumes the removed mapping — its own doc says "In local
   mode, the GPU VA IS the host VA (identity mapping), so we cast directly" — and on nullptr it
  returns nullptr, which every SDMA consumer treats as "not ready" and re-arms via
  stop_and_retry_current_packet(). If any SDMA operand or host-allocated signal VA is
  host-valid-but-not-KFD-mapped, the queue now stalls forever instead of completing;
  fire_queue_idle_signal similarly stops firing (unlike fire_signal, it has no fallback).

  This ranges from harmless to a hang depending on whether local-mode SDMA operands and signals are
  always KFD-mapped (PT hit). hipMalloc'd copy buffers likely are; host-allocated signals and userptr
  are the risk.
  Fix: author to confirm those VAs are always mapped in local mode, or extend the client-memory
  fallback to the pointer path (mirror the value accessors — copy the operand in/out via
  process_vm_readv/writev when resolve_host_ptr is null and vmid>0), or PROT-probe the identity
  pointer before returning it.

  MINOR

  New test doesn't lock in the fix — simulated_kfd_test.cpp:97-119
  The test reads/writes a committed host stack variable through the process VMID. Under the OLD code
  that address (well under 0x800000000000) hit the deleted identity passthrough and also returned the
  value — so the test passes on both old and new code and catches no regression. It also never
  exercises the actual crash scenario (a PROT_NONE reservation).
  Fix: add a case that mmaps a PROT_NONE region and asserts read32/write32 through the process VMID
  doesn't crash and returns benign backing (that's the regression the old code failed). Ideally on the
   gfx1201 config from the issue, not gfx950.

  NIT

  - set_process_client_pid and resolve_host_ptr lack the Doxygen @brief/@details the surrounding
  methods have.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
```diff
@@ -343,6 +358,53 @@ class GpuMemory : public simdojo::SparseMemory {
...
+    pid_t pid = client_pid_for_vmid(vmid);
+    if (pid <= 0)
+      return nullptr;
+    uint8_t byte = 0;
+    iovec local{&byte, 1};
+    iovec remote{reinterpret_cast<void *>(addr), 1};
+    ssize_t rc = process_vm_readv(pid, &local, 1, &remote, 1, 0);
+    if (rc != 1)
+      return nullptr;
+    return reinterpret_cast<uint8_t *>(addr & ~PAGE_MASK);
```
kuhar: We should restrict this identity fallback to the simulator process address space. In daemon mode, open_process(client_pid) records a PID from another process, and a successful process_vm_readv only proves that the address is readable there; resolve_sdma_ptr and the completion paths then dereference the returned VA in the daemon. A cross-process probe at this head made resolve_host_ptr return the child mapping even though that address was unmapped in the parent. Please require an explicit local-mode condition such as pid == getpid(), keep daemon misses null or use the copy-based APIs, and add a daemon regression with a client-only mapping.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h
```diff
@@ -291,7 +329,39 @@ class KfdProcess {
...
+    sparse_reservation_pages_.clear();
+    for (const auto &[handle, alloc] : allocations_) {
+      (void)handle;
+      if (alloc.size == 0 || alloc.gpu_va > UINT64_MAX - (alloc.size - 1))
+        continue;
+      const uint64_t first_page = alloc.gpu_va >> kPageShift;
+      const uint64_t last_page = (alloc.gpu_va + alloc.size - 1) >> kPageShift;
+      uint64_t page = first_page;
+      while (true) {
+        auto &state = sparse_reservation_pages_[page];
```
kuhar: We should store sparse reservations as ranges instead of inserting one hash-map entry per 4 KiB page. Running this production method with the checked-in gfx1201 local-memory size of 32,061,259,776 bytes created 7,827,456 entries and used 361,800 KiB maximum RSS just to answer one four-byte query; the rebuild also runs under alloc_mutex_ after allocation invalidation. Please use an interval or boundary representation whose storage scales with allocation count, and add a large-range regression that does not materialize every page.

emulation/rocjitsu/tests/simulated_kfd_test.cpp
```diff
@@ -94,6 +97,155 @@ TEST_F(SimulatedKfdTest, OpenAndClose) {
...
+  uint32_t map_gpu = alloc.gpu_id;
+  kfd_ioctl_map_memory_to_gpu_args map{};
+  map.handle = alloc.handle;
+  map.device_ids_array_ptr = reinterpret_cast<uint64_t>(&map_gpu);
+  map.n_devices = 1;
+  ASSERT_EQ(t.driver()->ioctl(process_id, AMDKFD_IOC_MAP_MEMORY_TO_GPU, &map), 0);
+  EXPECT_EQ(map.n_success, 1u);
+  EXPECT_TRUE(process->contains_sparse_reservation(alloc.va_addr, alloc.size));
+
+  for (uint32_t i = 0; i < 2048; ++i) {
```
kuhar: It would be useful to make this test observe the skipped client probe. The value round-trip is identical if the is_sparse_reservation guard is removed: process_vm_readv/process_vm_writev fail on PROT_NONE, then the same sparse fallback supplies the values. All 10 SimulatedKfdTest cases pass at this head, but this regression would still pass if the GpuMemory classifier wiring disappeared. Please capture the warning or expose a probe counter and assert that the reservation loop performs no client-memory probes.

kuhar (commented):
A few comments

atgutier (commented):
Agent's found more blockers/majors:

## BLOCKER

 [BLOCKER]

lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h:99-102,393-405 — daemon
  return a foreign-process address as a local pointer

  Reasoning: resolve_client_identity_page() uses process_vm_readv() to confirm that an address is
  readable in the client process, then returns that same numeric address for direct dereference in the
  daemon. Successful cross-process reading does not mean the daemon has the address mapped
  identically. This can segfault or access an unrelated daemon mapping.

  Evidence:

  - Remote clients are registered through open_process(client_pid) at simulated_kfd.cpp:410-449.
  - The remote probe occurs at gpu_memory.h:399-404, followed by the unsafe pointer construction at
    line 405.

  - Consumers require a genuinely local pointer and immediately dereference it:
      - command_processor.cpp:1928-1936 uses atomic_ref.
      - command_processor.cpp:1987-1992 performs an atomic load.
      - command_processor.cpp:2024-2044 uses memcpy and atomics.
      - completion_tracker.cpp:88-132,170-238 reads and writes through returned pointers.

  - The new test at simulated_kfd_test.cpp:100-121 covers only same-process local mode.

  Suggested fix: Restrict identity-pointer fallback to a VMID explicitly known to share the
  simulator’s address space, such as client_pid == getpid() plus local mode. For daemon clients,
  return only page-table-backed daemon pointers; otherwise use process_vm_readv/writev copy operations
  or return nullptr. Add a separate-process daemon regression test.

  Confidence: high.

  ## MAJOR

  [MAJOR] lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h:251-275,337-364 — sparse reservations
  expand into one hash entry per page

  Reasoning: The first classification after any allocation change rebuilds an unordered_map entry for
  every page of every allocation. That defeats sparse backing for the large PROT_NONE GPUVA
  reservations this PR is intended to support.

  Evidence:

  - contains_sparse_reservation() initiates the rebuild at lines 259-262.
  - rebuild_sparse_reservation_index_locked() walks every page and creates a hash node at lines 343-
    355.

  - Allocation size is accepted directly at simulated_kfd.cpp:1293-1299.
  - The simulator advertises 64 GiB at simulated_kfd.cpp:1215. A reservation of that size creates
    16,777,216 hash entries, potentially consuming hundreds of MiB and holding alloc_mutex_ for a
    prolonged rebuild.
 - Every allocation mutation invalidates the index, forcing another full rebuild on the next query.
  - The new regression test exercises only one 4 KiB page at simulated_kfd_test.cpp:139-165.

  Suggested fix: Represent sparse and host-backed coverage as intervals/endpoints with coverage
  counts, or scan the normally small allocation set on demand. Host-backed overlap precedence can be
  preserved without per-page storage. Add a large logical-reservation test demonstrating that
  auxiliary storage remains proportional to allocation ranges, not page count.

  Confidence: high.

atgutier (commented):
Thanks for iterating on this. Had some gcc compilation errors and found a few other issues.
                                                                                                                                                                                             
• ## BLOCKER

  [BLOCKER] tests/simulated_kfd_test.cpp:86-101,176 — the new daemon regression test does not compile under GCC with -Werror

  Reasoning: EXPECT_EQ expands to control flow, so using it as the unbraced body of if (done_fd >= 0) triggers -Wdangling-else. GCC’s fortified libc also rejects the ignored write() and read() return values
  despite the (void) casts. The test target therefore cannot build.

  Evidence:

  - ChildProcessGuard::finish() fails at line 98 with -Werror=dangling-else.
  - Lines 91 and 176 fail with -Werror=unused-result in GCC ASan/UBSan.
  - test (asan-ubsan-gcc) failed during compilation.
  - Both TheRock Linux package builds failed on the same source file; their “custom container implementation failed” annotation is secondary, not infrastructure noise.

  Suggested fix: add braces around the conditional containing EXPECT_EQ, and store/check the read() and write() results, including EINTR handling where appropriate.

  Confidence: high.

  ## MAJOR

  [MAJOR] lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp:435-449 — documented local open_process(0) misses the PID required by the new fallback

  Reasoning: rj_vm_device_open(vm, 0, ...) is explicitly documented as the local-mode API and forwards zero to open_process(). That function only records and publishes a PID when client_pid > 0.

  At this head, GpuMemory::translate() no longer identity-maps process VMIDs. Both the new identity-pointer path and the process_vm_* copy paths reject a zero PID. Consequently, a valid local host pointer not
  present in that process’s page table resolves to nullptr, while scalar/block accesses incorrectly use sparse backing instead of host memory.

  Evidence:

  - rj_vm.h:241-249 documents zero for local mode.
  - rj_vm.cpp:320-330 forwards zero directly.
  - simulated_kfd.cpp:435-449 leaves both KfdProcess::client_pid_ and GpuMemory::VmidEntry::client_pid at zero.
  - gpu_memory.h:349-363,406-453 removes the old process-VMID identity fallback and requires a positive PID.
  - The added local test uses driver.open(), whose separate path correctly registers getpid(), so it misses the public rj_vm_device_open(..., 0, ...) path.

  Suggested fix: normalize zero to getpid() when open_process() is used in non-daemon mode, and publish that effective PID to both objects. Alternatively, make local device-open adopt the bootstrap local
  process with balanced references. Add an API-level test covering host read, write, and resolve_host_ptr() through the returned process ID.

  Confidence: high.

  [MAJOR] lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h:139-169,246-329 — sparse backing is not isolated by VMID

  Reasoning: the new classifier intentionally routes valid PROT_NONE GPU allocations directly to inherited SparseMemory, but those fallback calls pass only the GPU virtual address. The base storage also keys
  pages solely by address. Independent VMIDs using the same GPUVA therefore read and write the same bytes.

  Sparse pages are not retired when an allocation or process is removed, so the backing also persists across process/allocation lifetimes.

  Evidence:
 - gpu_memory.h:372-403 makes classified sparse reservations skip client-memory probing.
  - The fallback calls at gpu_memory.h:139-169,246-329 discard vmid.
  - lib/simdojo/include/simdojo/components/sparse_memory.h:208-234 indexes pages only by virtual page address.
  - Non-daemon open_process(0) calls create distinct KfdProcess objects.
  - Every KfdProcess starts default allocation at 0x1000000000.
  - Free and close remove allocation/classifier state but never remove corresponding sparse pages.

  The flat fallback predates this PR for failed client probes, but this change materially exposes it by making that storage the intended successful backing for registered local GPU-only allocations.

  Suggested fix: give sparse allocations explicit process/VMID-scoped backing and reclaim it during allocation/process teardown, while keeping VMID-zero simulator memory separate. Add two simultaneous process
  IDs with identical sparse GPUVAs and verify that their values remain isolated.

  Confidence: high.

kuhar (commented):
Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h:156`

We should update GpuMemoryTest.RegisteredVmidPassthroughMissRespectsUserSpaceLimit for the new registered-VMID contract before landing. At 8632816264f7, resolve_host_ptr(0x4000, 7) returns nullptr at this path while the test still expects the identity address 0x4000; the focused CTest fails, and the same sole failure appeared in the default, GCC 13, Clang ASan/UBSan, and Clang TSan trees (18/19 cases passed in each). If registered process VMIDs are intentionally no longer raw passthrough, please update that expectation while retaining explicit VMID-0 and same-process client-PID coverage.

Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h:237`

We should avoid calling the sparse classifier while atomic_rmw holds the page-table shared lock. The classifier enters KfdProcess::contains_sparse_reservation and takes alloc_mutex_, while map/free paths hold alloc_mutex_ before map_pages/unmap_pages take page_table_mutex_; that creates an AB-BA lock inversion. A deterministic reproducer at 8632816264f7 hung under a 2-second timeout. Please classify before taking the page-table lock, as the scalar paths do, or use a lock-free snapshot, and add a concurrent sparse-atomic versus map/free regression.

emulation/rocjitsu/tests/simulated_kfd_test.cpp
kuhar: It would be useful to extend this regression to cover the pointer-returning path too: add EXPECT_EQ(memory->resolve_host_ptr(alloc.va_addr, process_id), nullptr) inside the reset/count window. The current loop only exercises read32/write32; without this call, the sparse guard in resolve_client_identity_page can regress while the value round-trip still passes. Together with the zero-probe assertion, the pointer check locks in the shortcut used before SDMA or completion code can receive a PROT_NONE address.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
```diff
@@ -101,18 +109,55 @@ class GpuMemory : public simdojo::SparseMemory {
...
+      auto it = vmid_table_.find(pid);
+      if (it == vmid_table_.end())
+        return;
+      sparse_memory = it->second.sparse_memory;
+      classifier = it->second.sparse_reservation_classifier;
+    }
+    if (!sparse_memory)
+      return;
+    sparse_memory->erase_pages_if(addr, size, [&](uint64_t page_addr) {
+      return !classifier || !classifier(page_addr, PAGE_SIZE);
```
kuhar: We should either reject overlapping GPU VA allocations or make reclaim test sparse ownership independently of host precedence. `assign_gpu_va_locked()` accepts requested VAs without checking `allocations_`, while `discard_process_sparse_range()` erases whenever the effective classifier is false. With two sparse handles plus one host-backed handle on the same page, freeing either sparse handle deletes the shared sparse page even though another sparse owner remains; removing the host mapping later exposes empty backing. A sparse+sparse+host free-order test would lock this down.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
```diff
@@ -1061,6 +1106,7 @@ void *SimulatedKfd::dispatch_mmap(KfdProcess &proc, void *addr, size_t length, i
 
   alloc.host_ptr = host_ptr;
```
kuhar, line 1303: This would be safer if an allocation kept one backing across `mmap`/`munmap`. At this line a previously sparse allocation becomes host-backed, but existing bytes are not copied from the process sparse store; `dispatch_munmap` later flips it back without copying host writes. GPU data can therefore disappear at `mmap` and stale sparse data can reappear after `munmap`. Please migrate or reconcile the bytes, or reject this lifecycle explicitly, and add a round-trip regression.

kuhar (commented):
A few comments

emulation/rocjitsu/tests/simulated_kfd_test.cpp
kuhar: I think we should also verify the opposite classifier branch in `tests/simulated_kfd_test.cpp:597`. The current `EXPECT_FALSE` checks only metadata; please write through the process VMID, assert that the host buffer changes, and assert `client_memory_probe_count() > 0`. That would show a userptr reaches client memory instead of process-scoped sparse backing.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
RattataKing: Changed to reject overlapping allocation

emulation/rocjitsu/tests/simulated_kfd_test.cpp
```diff
@@ -94,6 +100,490 @@ TEST_F(SimulatedKfdTest, OpenAndClose) {
...
+  {
+    std::lock_guard<std::mutex> lock(process->alloc_mutex_);
+    EXPECT_EQ(process->cpu_mappings_.size(), 1u);
+    EXPECT_EQ(process->gpu(0).doorbell_page, doorbell_page);
+  }
+
+  process->event_state_.notify_closing();
+  EXPECT_EQ(t.driver()->munmap(doorbell_page, kPageSize), 0);
+  process->event_state_.reset();
+  if (spanning_result != 0)
```
kuhar: We should brace both cleanup conditionals here and at `tests/simulated_kfd_test.cpp:329`. `ninja -C build-gcc-13 rocjitsu_tests` fails at both lines because `EXPECT_EQ` expands to conditional control flow and GCC 13 reports `-Werror=dangling-else`. Clang accepts the code, but the checked-in GCC configuration cannot build this test target until both bodies are braced.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/remote_driver.cpp
```diff
@@ -717,48 +726,55 @@ void *RemoteDriver::mmap(void *addr, size_t length, int prot, int flags, off_t o
...
+      if (memfd >= 0)
+        syscall(SYS_close, memfd);
+      errno = EINVAL;
+      return MAP_FAILED;
+    }
 
     // Pre-copy committed pages (code objects) from the existing anonymous
     // reservation into the memfd before MAP_FIXED replaces them. Uses a temp
     // mapping outside the GPUVM range for the copy target.
     if ((flags & MAP_FIXED) && addr != nullptr) {
```
kuhar: We should preserve GPU-authoritative backing on the remote mmap path. The local path stops importing client pages after `MAP_MEMORY`, but `RemoteDriver::mmap` still copies every resident `MAP_FIXED` source page into the memfd. At this head, a repro with `0x22222222` in the backing and stale `0x11111111` in the reservation maps as `0x11111111`, overwriting the simulated GPU result. Please propagate the first-import/authority state to `RemoteDriver`, or make the daemon own the import before the backing becomes authoritative, and add the equivalent daemon lifecycle regression.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h
```diff
@@ -309,6 +395,42 @@ class KfdProcess {
...
+
+  void publish_backing_ranges_locked() {
+    std::vector<BackingRange> ranges;
+    ranges.reserve(allocations_.size());
+    for (const auto &[handle, alloc] : allocations_) {
+      (void)handle;
+      if (!alloc.permanent_backing || !alloc.host_ptr)
+        continue;
+      assert(is_valid_gpu_range(alloc.gpu_va, alloc.size));
+      ranges.push_back({alloc.gpu_va, alloc.size, static_cast<uint8_t *>(alloc.host_ptr),
```
kuhar, line 415: We should publish the page-rounded backing range here. `create_lazy_vram_backing()` and CPU mmap cover `backing_size`, but this vector covers only `alloc.size`; for a non-page-multiple VRAM allocation, the tail is classified as unmanaged and local `with_host_ptr()` identity-dereferences the client reservation. At this head, a 4097-byte PROT_NONE allocation accepts an in-range write and then faults with SIGSEGV on a tail-page write; with a writable reservation, the tail write lands in client memory instead of the backing. Page-granular overlap checks already reserve the whole backing page, so using `backing_size` is safe. Please update the overlap assertion to the same units and add a non-page-multiple regression.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
```diff
@@ -607,6 +776,10 @@ int SimulatedKfd::close(uint32_t process_id) {
       }
     }
     proc.allocations_.clear();
+    proc.refresh_backing_ranges_locked();
+    for (const auto &[begin, end] : proc.cpu_mappings_)
+      libc_passthrough().munmap(reinterpret_cast<void *>(begin), end - begin);
```
kuhar, line 781: This would be safer if close preserved caller-owned `MAP_FIXED` reservations. The permanent-backing mmap path records those mappings in `cpu_mappings_`, and close unmaps every interval there even though `GpuAllocation::host_ptr_owned` documents that the driver must not unmap thunk-owned pages. A current-head repro maps VRAM over a PROT_NONE reservation, closes KFD, and then `mincore` returns ENOMEM because the VA became a hole. Please track ownership for these intervals and reclaim only driver-created mappings; a close-then-`mincore` regression would lock in the contract.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
```diff
@@ -1270,28 +1567,49 @@ int SimulatedKfd::alloc_memory_ioctl(KfdProcess &proc, void *arg) {
...
-  uint64_t va = args->va_addr;
-  if (va == 0) {
-    va = proc.next_gpu_va_;
-    proc.next_gpu_va_ += (args->size + 0xFFF) & ~0xFFFULL;
-  }
+  std::optional<uint64_t> aligned_size = checked_page_aligned_size(args->size);
+  if (!aligned_size)
+    return -EINVAL;
+
+  std::optional<uint64_t> assigned_va = assign_gpu_va_locked(proc, args->va_addr, *aligned_size);
```
kuhar, line 1574: The overlap check should not reject byte-disjoint userptr registrations that share one host page. `alloc_memory_ioctl()` page-rounds every requested size before `assign_gpu_va_locked()`, so two 64-byte USERPTR buffers 128 bytes apart collide. A current-head repro returns 0 for the first registration and -EINVAL for the second. ROCr can issue separate small host registrations from the same page, so please scope the new page-exclusivity rule to allocation kinds that require exclusive GPU backing, while retaining overlap rejection for permanent-backing VRAM, and add coverage for both cases.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/kmd/linux/simulated_kfd.cpp
```diff
@@ -107,6 +112,147 @@ template <typename... Args> int safe_fcntl(int fd, int cmd, Args... args) {
...
+    return std::nullopt;
+
+  uint64_t gpu_va = proc.next_gpu_va_;
+  while (gpu_va <= UINT64_MAX - *aligned_size) {
+    const uint64_t first_page = gpu_va >> KfdProcess::kPageShift;
+    const uint64_t end_page = ((gpu_va + *aligned_size - 1) >> KfdProcess::kPageShift) + 1;
+    uint64_t next_page = first_page;
+    for (const auto &[handle, alloc] : proc.allocations_) {
+      (void)handle;
+      assert(KfdProcess::is_valid_gpu_range(alloc.gpu_va, alloc.size));
```
kuhar, line 185: We should reject invalid imported ranges before they enter `allocations_` and route requested IPC VAs through the same assignment check. `import_dmabuf_ioctl()` accepts a zero-length memfd and records `{gpu_va=0,size=0}`; the next auto-VA allocation then underflows `alloc.size - 1` in this scan and returns -EINVAL. The same direct-insertion pattern lets scratch and IPC imports bypass overlap validation, and IPC still advances `next_gpu_va_` manually. Please centralize range validation for these insertion paths, reject size zero at import, and add a zero-size dmabuf followed by auto-VA regression plus an overlapping IPC-import case.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/gpu_memory.h
```diff
@@ -478,67 +574,69 @@ class GpuMemory : public simdojo::SparseMemory {
         cache.pte = it->second;
     }
 
-    return fn(cache.found ? &cache.pte : nullptr);
+    std::optional<KfdProcess::ResolvedBacking> backing;
+    if (!cache.found && entry.backing_resolver)
+      backing = entry.backing_resolver(addr, size);
```
kuhar, line 579: Can we cache backing-range resolution with the same generation-keyed mechanism used for PTEs? Permanent-backing VRAM has no PTE, so every load and store currently pays a `std::function` call, `weak_ptr::lock()`, shared lock, and binary search. A current-head RelWithDebInfo microbenchmark measured repeated `read32` at 54.4 ns through this resolver versus 32.4 ns through the cached page-table path, a 1.68x difference. Please cache the resolved range with explicit lifetime invalidation, or otherwise avoid this per-access synchronization on the simulator hot path.

kuhar (commented):
A few comments

atgutier (commented):
A few more comments:

 [MAJOR] lib/rocjitsu/src/rocjitsu/kmd/linux/kfd_process.h:415 - Publish
  the page-owned backing extent, not the requested byte size

  Reasoning: The new permanent-backing design creates a page-rounded memfd
  and stores that extent in alloc.backing_size, but
  publish_backing_ranges_locked() publishes only alloc.size. For a PAGE_SIZE
  + 1 VRAM allocation, most of the second page is consequently classified as
  unmanaged.

  When an access reaches that portion of the page, resolve_backing() returns
  nullopt at lines 321–323. GpuMemory::with_host_ptr() interprets nullopt as
  an unmanaged address and may use identity-pointer or client-memory
  fallback at gpu_memory.h:611–617. In local mode this can dereference a
  synthetic GPU VA and crash. In daemon mode it can access unrelated client
  or sparse memory. This violates the central invariant introduced by this
  PR: an address owned by a managed allocation must never fall through as
  unmanaged.

  The daemon has the same extent mismatch. remote_driver.cpp:721–722
  registers alloc_args->size, while find_memfd_for_addr() requires the
  entire mapping to fit that range at lines 175–180. A page-rounded
  anonymous MAP_FIXED can therefore miss the shared memfd and replace the
  GPUVM range with anonymous memory.

  Evidence:

  - simulated_kfd.cpp:1570–1574 computes the page-aligned allocation size.
  - simulated_kfd.cpp:1597–1604 creates the rounded backing and records
    backing_size.

  - Allocation collision checks are already page-granular.
  - kfd_process.h:415 nevertheless publishes alloc.size.
  - No added test uses a non-page-aligned VRAM allocation.

  Suggested fix:

  - Retain separate logical and page-owned sizes.
  - Publish backing_size through backing_ranges_ for permanent allocations.
  - Register the memfd’s actual size in the daemon’s alloc_ranges_.
  - If accesses beyond the logical byte size should fault or be dropped,
    represent that as a managed fault—not an unmanaged miss.

  - Add local and daemon tests for sizes 1 and PAGE_SIZE + 1. Verify that
    accesses throughout the final page never select identity, client-memory,
    or sparse fallback.

 [MINOR] lib/rocjitsu/src/rocjitsu/kmd/linux/events.cpp:65 — Compare
  normalized munmap intervals for event and doorbell pages

  Reasoning: dispatch_munmap() correctly rounds a positive length to
  complete pages at simulated_kfd.cpp:1330–1337, but the special-page paths
  compare the original byte length:

  - Doorbells require gs.doorbell_page_size == length at
    simulated_kfd.cpp:1347.

  - Event pages require size == page_size at events.cpp:65.

  Therefore, munmap(page, 1) is rejected with EPERM, although Linux treats
  it as unmapping exactly that one page. Ordinary permanent mappings already
  normalize this case. The PR’s own permanent-mapping test expects
  munmap(mapping, 1) to work, but the event and doorbell tests cover only
  full-page release and spanning rejection.

  Suggested fix:

  - Compare the page-aligned start and rounded end against the recorded
    special-page interval.

  - Pass the normalized length to event-page classification.
  - Preserve rejection for requests whose normalized interval truly
    partially overlaps or spans beyond the protected page.

  - Add munmap(page, 1) and munmap(page, PAGE_SIZE - 1) tests for both event
    and doorbell mappings.

  Confidence: High.

PR #9019, opened 2026-07-22
fix(clr): do not allow clr to read out of bounds in some cases

cjatin:
## Motivation

Add checks so that we do not get seg faults when reading modules.

## Technical Details

Some elf can have incorrect sizes, we basically adjust for it.

## JIRA ID

JIRA ID : ROCM-26177

## Test Plan

Tested locally with incorrect elf.

## Test Result

The error is no longer seen.

## Submission Checklist

https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

Co-authored-by: Claude <noreply@anthropic.com>


projects/clr/rocclr/elf/elf.cpp
```diff
@@ -800,45 +800,74 @@ bool Elf::dumpImage(std::istream& is, char** buff, size_t* len) const {
...
-          total_size += shdr[i].sh_size;
-        }
-      }
+  // Only 64-bit AMDGPU code objects are supported.
+  if (eclass != ELFCLASS64 || buf_size < sizeof(Elf64_Ehdr)) {
+    return 0;
+  }
+
+  // end = off + sz, returns false on overflow.
+  auto checked_end = [](uint64_t off, uint64_t sz, uint64_t& end) -> bool {
```
AlexeySachkov, line 815: I would suggest to name the lambda `checked_add` or something like this, which would simplify transition to the standard [`chk_add`](https://cppreference.com/cpp/numeric/ckd_add) once C++26 is adopted (maybe even align on the signature?)

projects/clr/rocclr/elf/elf.cpp
```diff
@@ -800,45 +800,74 @@ bool Elf::dumpImage(std::istream& is, char** buff, size_t* len) const {
...
+  // Validate and bound the section-header table before any dereference.
+  if (ehdr->e_shnum != 0) {
+    if (ehdr->e_shentsize != sizeof(Elf64_Shdr) || ehdr->e_shoff >= buf_size) {
+      return 0;
     }
-  } else if (eclass == ELFCLASS64) {
-    auto ehdr = static_cast<const Elf64_Ehdr*>(emi);
-    auto shdr = reinterpret_cast<const Elf64_Shdr*>(static_cast<const char*>(emi) + ehdr->e_shoff);
+    uint64_t shtab = static_cast<uint64_t>(ehdr->e_shnum) * sizeof(Elf64_Shdr);
+    if (shtab > buf_size - ehdr->e_shoff) {
```
AlexeySachkov, line 834:
```suggestion
    if (shtab + ehdr->e_shoff > buf_size ) {
```

I think that it is safe to say that `shtab + ehdr->e_shoff` won't overflow `uint64_t`:
- `ehdr->e_shnum` is 16-bit value
- `sizeof(Elf64_Shdr) == 40` which we will round up to 64 for simplified calculations, i.e. 2^6
- `ehdr->eshoff` is 32-bit value
- 2^16 * 2^6 + 2^32 <= 2^34 (in fact 2^33 should be enough)

At the same time, `buf_size` could be smaller than `ehdr->eshoff` meaning that in existing form, that subtraction on the right may underflow leading to `shtab > number bigger than shtab == false` and going further.
Of course, we have a check at the very end that `total_size` never goes past `buf_size`, it I think that the suggested check is easier to read and will catch more negative cases earlier.

_Perhaps_ the same applies to a similar `phtab` and `ehdr->e_phoff` check below, but I haven't checked that.

cjatin: Thats true, but we check for `buf_size` in line 830. Which means the math in this line should be safe.
Also I think size of 64bit elf's `Shdr` is 64 and not 40.

projects/clr/rocclr/elf/elf.cpp
cjatin: If its just name, I rather keep it. clr's codebase is far from `C++26`.

projects/clr/rocclr/os/os_win32.cpp
```diff
@@ -675,7 +675,20 @@ bool Os::MemoryMapFileTruncated(const char* fname, const void** mmap_ptr, size_t
...
+    MEMORY_BASIC_INFORMATION mbi;
+    const DWORD readable = PAGE_READONLY | PAGE_READWRITE | PAGE_EXECUTE_READ |
+                           PAGE_EXECUTE_READWRITE | PAGE_WRITECOPY | PAGE_EXECUTE_WRITECOPY;
+    if (VirtualQuery(image, &mbi, sizeof(mbi)) != 0 && mbi.State == MEM_COMMIT &&
+        (mbi.Protect & readable) != 0) {
+      uintptr_t region_end = reinterpret_cast<uintptr_t>(mbi.BaseAddress) + mbi.RegionSize;
+      *region_bound_ptr = static_cast<size_t>(region_end - reinterpret_cast<uintptr_t>(image));
+    }
+  }
+
```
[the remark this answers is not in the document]
cjatin: done

projects/hip-tests/catch/unit/oob/oob_module.cc
```diff
@@ -0,0 +1,230 @@
...
+constexpr char kElfMagic[4] = {0x7f, 'E', 'L', 'F'};
+
+Bytes ReadFile(std::string_view path) {
+  std::ifstream f(std::string(path), std::ios::binary | std::ios::ate);
+  REQUIRE(f.good());
+  auto size = f.tellg();
+  f.seekg(0);
+  Bytes buf(static_cast<size_t>(size));
+  f.read(buf.data(), size);
+  REQUIRE(f.good());
```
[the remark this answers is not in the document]
cjatin: done

projects/clr/rocclr/os/os_posix.cpp
```diff
@@ -876,7 +876,7 @@ bool Os::GetFileHandle(const char* fname, FileDesc* fd_ptr, size_t* sz_ptr) {
 }
 
 bool amd::Os::FindFileNameFromAddress(const void* image, std::string* fname_ptr,
-                                      size_t* foffset_ptr) {
+                                      size_t* foffset_ptr, size_t* region_bound_ptr) {
   // Get the list of mapped file list
   bool ret_value = false;
   std::ifstream proc_maps;
```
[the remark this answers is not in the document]
cjatin: done

projects/clr/rocclr/elf/elf.cpp
AlexeySachkov: > Thats true, but we check for buf_size in line 830

Oh, I missed that. Then indeed there is no underflow. I still think that a rewritten `if` would be easier to read, but that's nitpicking

mangupta (approved):

PR #9020, opened 2026-07-22
fix(rocr): Fix vendor PM4 command buffer overflow handling

fcui-amd:
## Motivation

The previous fix for a heap buffer overflow in `VendorSpecificAqlToPm4` introduced a regression in a PyTorch Transformers text-generation workload using Llama 3.1 8B Instruct in FP16 mode.
This PR reverts that fix and replaces it with a safer PM4 command buffer size check.

pid:8989 tid:0x7ffe07fff6c0 [KernelDispatchAqlToPm4] queue 0x7fe400660000 kernel dispatch head=b02 setup=3 wx=80 wy=1 wz=1 gx=1a00 gy=1 gz=1 ps=0 gs=0 ko=7ff18724cac0 ka=0x7fe401206580 cs=0
pid:8989 tid:0x7ffe07fff6c0 [KernelDispatchAqlToPm4] kernel object property=408 entry=0x7ff187491500 lds=0+0
pid:8989 tid:0x7ffe07fff6c0 [KernelDispatchAqlToPm4] PM4 command buffer overflow in KernelDispatch: cumulative 592 bytes exceeds frame limit 560 bytes
pid:8989 tid:0x7ffe07fff6c0 [AqlToPm4Thread] process compute queue fail status = 00001008
pid:8989 tid:0x7ffe07fff6c0 [AqlToPm4Thread] aql to pm4 thread 0x7fe400660000 exit

## Technical Details

- Revert the previous `VendorSpecificAqlToPm4` heap buffer overflow fix.
- Calculate the required PM4 command buffer frame size before writing commands.
- Skip vendor packet processing when the required size exceeds `cmdbuf_aql_frame_size`.

## Test Plan

Perform PyTorch Transformers text-generation workload with Llama 3.1 8B Instruct, FP16 precision, and batch size 1.
The workload exercises Hugging Face `model.generate()` on ROCm using default text prompts and reports standard inference latency and throughput metrics.

## Test Result

test pass.

## Submission Checklist


projects/rocr-runtime/libhsakmt/src/dxg/wddm/queue.cpp
```diff
@@ -899,31 +877,42 @@ hsa_status_t ComputeQueue::VendorSpecificAqlToPm4(char* cpu, amd_aql_pm4_ib* pac
...
+
+      if (EnableProfiling()) required_size += 2 * sizeof(PM4MEC_COPY_DATA);
+      if (platform_atomic_support_) required_size += sizeof(AtomicTemplate);
+    }
+  }
+
+  if (required_size > cmdbuf_aql_frame_size) {
+    pr_err("PM4 command buffer overflow in VendorSpecific: required %zu bytes, limit %u bytes\n",
+           required_size, cmdbuf_aql_frame_size);
+    process_packet = false;
```
cfreeamd, line 903: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** On overflow this sets `process_packet = false` instead of returning `HSA_STATUS_ERROR_OUT_OF_RESOURCES` (the prior behavior). Control then falls into the existing `else` branch, which for a packet with a completion signal calls `hsakmt_hsa_signal_store_screlease(packet->completion_signal, 0)` — so an oversized vendor PM4 IB is silently skipped on the GPU while the client still observes the signal as complete. There's no hang (it reuses the existing `vendor_packet_process`-disabled skip path), but the client sees success for work that never ran on the GPU. Is drop-and-signal the intended contract for an oversized IB, rather than surfacing an error / queue-error callback? If deliberate, a one-line comment here stating that intent would help future readers.

projects/rocr-runtime/libhsakmt/src/dxg/wddm/queue.cpp
```diff
@@ -899,31 +877,42 @@ hsa_status_t ComputeQueue::VendorSpecificAqlToPm4(char* cpu, amd_aql_pm4_ib* pac
...
+      required_size += sizeof(BarrierTemplate);
+      required_size += device->Major() == 9 ? sizeof(gfx9::AcquireMemTemplate)
+                                            : sizeof(gfx10::AcquireMemTemplate);
+
+      if (EnableProfiling()) required_size += 2 * sizeof(PM4MEC_COPY_DATA);
+      if (platform_atomic_support_) required_size += sizeof(AtomicTemplate);
+    }
+  }
+
+  if (required_size > cmdbuf_aql_frame_size) {
```
cfreeamd, line 897: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `VendorSpecific` no longer has a post-build overflow check, so correctness now rests entirely on `required_size` being an exact upper bound. I verified it currently is (each `Build*` returns exactly `sizeof(<template>)`), but that only holds as long as the `required_size` accounting stays in lockstep with the `Build*` calls in the process block. Consider a cheap debug guard after building (e.g. `assert((i - ib_size) <= cmdbuf_aql_frame_size)`) to catch future drift if a new `Build*` call is added without a matching `required_size` term. Non-blocking.

cfreeamd (commented):

Reviewed the WDDM AQL→PM4 overflow rework in `projects/rocr-runtime/libhsakmt/src/dxg/wddm/queue.cpp`. The core fix looks correct: the `required_size` pre-check now guards the vendor `memcpy` before the copy (the real overflow), and I verified `required_size` is an exact upper bound — every `cmd_util.Build*` helper returns exactly `sizeof(<template>)`, and each term (pm4 bytes, BarrierTemplate, gfx9/gfx10 AcquireMemTemplate per major, PM4MEC_COPY_DATA×2 under profiling, AtomicTemplate, and the trailing ring_rptr write) matches the emitted commands. Removing the old cumulative `i > frame_size` check is also fine given the per-frame-slot buffer model (`ib_start_addr`/`ib_size` reset per submit round). A couple of non-blocking notes inline.

projects/rocr-runtime/libhsakmt/src/dxg/wddm/queue.cpp
fcui-amd: Updated

projects/rocr-runtime/libhsakmt/src/dxg/wddm/queue.cpp
fcui-amd: Updated

dayatsin-amd (approved):

PR #9025, opened 2026-07-22
docs(rocprofiler-compute): document multi-process PC sampling

abchoudh-amd:
## Motivation

Document multi-process PC sampling behavior across profile and analyze modes.

## Technical Details

- Add profile PID-prefixed output example
- Explain unified analyze-mode aggregation
- Document sorting and legacy workload compatibility

## JIRA ID

JIRA ID: AIPROFCOMP-743

## Test Plan

- [x] Build docs

## Test Result

- [x] Docs build

## Submission Checklist


projects/rocprofiler-compute/docs/how-to/pc_sampling.rst
```diff
@@ -36,6 +36,23 @@ For using profiling options for PC sampling the configuration needed are:
...
+Profile multi-process workloads
+-------------------------------
+
+The same profile command supports applications that create multiple processes.
+ROCm Compute Profiler writes a separate PC sampling result and code-object
+information file for each process. The process ID (PID) prefix associates each
+result with the code objects loaded by that process.
+For example, a workload with PIDs 4312 and 4313 produces files like these:
+
+.. code-block:: shell-session
```
vedithal-amd: The output of profiling is an internal contract and users only ever should use the output of analysis.

Does it really make sense to document internal contract in public documentation and increase the maintenance overhead of syncing this piece up whenever we change profiling output structure (which will be more frequent due to profiler hub integration plans)?

Please provide some context on why we need to document this

projects/rocprofiler-compute/docs/how-to/pc_sampling.rst
```diff
@@ -132,6 +149,57 @@ Sorting a single kernel by sample ``count`` instead of ``offset``:
...
+     - 40
+     - ``WAITCNT: 20, ALU_DEPENDENCY: 20``
+
+Sorting and ``--pc-sampling-rows`` are applied after all processes are
+combined. For example, ``--pc-sampling-rows 10`` selects the ten rows with the
+highest unified counts when the sorting type is ``count``.
+
+.. note::
+
+   Analyze mode remains compatible with workloads profiled with previous versions of ROCm Compute Profiler.
```
vedithal-amd: This is not true - we keep making non backward compatible changes in output structure of profile mode and analyze mode adjusts for it.

Also, this line seems out of place with the context and seems unrelated.

Recommend to drop this line and not make such claims

vedithal-amd (changes requested):

vedithal-amd:
Rebased onto the latest `users/abchoudh/pcs-common`. The diff now contains documentation changes only.

vedithal-amd:
Pushed doc corrections on top of the existing commits.

- Rewrote the analyze section around per-process rows and the `pid` column. Rows are not merged across processes, so the aggregation example table was removed.
- Added the `pid` column to the sample output tables.
- Removed the profile output file listing.
- Removed the backward-compatibility note, since older workloads are not discovered by the current loader.

vedithal-amd (approved):

projects/rocprofiler-compute/docs/how-to/pc_sampling.rst
vedithal-amd: Removed the file listing. The section now only states that multi-process workloads are supported.

projects/rocprofiler-compute/docs/how-to/pc_sampling.rst
vedithal-amd: Dropped the note.

PR #9030, opened 2026-07-22
fix(rocjitsu): Implement fp16-ovfl MODE flag and refactor HWREGs

lua1235:
## Motivation

The MODE flag fp16-ovfl, which selects finite overflow behavior for f32->f8 and *->f16 instructions, was not represented in rocjitsu. Also, several HWREG IDs mapped to the wrong status register, and S_GETREG/SETREG_B32 paths were similarly inconsistent. 

## Technical Details

- Refactored generated SOPK S_GETREG_B32/S_SETREG_B32 paths to use shared architecture-aware HWREG helpers that decode {id, offset, size} and read/write backed wave state such as MODE and STATUS. 
- Corrected HWREG numbering for modern archs: MODE is ID 1 and STATUS is ID 2, fixing the old path that treated ID 1 as STATUS.
- Made known HWREG IDs explicit in per-architecture tables, with unsupported/read-only/privileged registers classified instead of falling through generated switches.
- Implemented MODE-backed helpers for the fp16-ovfl flag and added corresponding paths to instructions.

## Issue Tracking

N/A

## Test Plan

Added e2e tests for overflow behavior before/after setting fp16_ovfl, added unit tests.

## Test Result

Tests pass

## Submission Checklist


kuhar (commented):
Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/wavefront.h:252`

We should include MODE in checkpoint save/restore now that it affects conversion results. `WavefrontState` serializes STATUS but not `mode_raw()`, and restore dispatches a fresh wavefront without calling `set_mode_raw()`. A checkpoint taken with FP16_OVFL set therefore resumes with the bit clear. Please serialize MODE and add a round-trip test that preserves FP16_OVFL.

Original anchor: `emulation/rocjitsu/lib/util/include/util/data_types.h:453`

It would be good to keep the FNUZ overflow decision in the same rounding path as the base conversion. The four `*_fnuz_{rne,sr}_overflows` helpers duplicate the exponent/mantissa rounding in the converters, so a later rounding fix can silently make MODE.FP16_OVFL classification diverge. Could we share an overflow flag/helper, or at least add boundary tests for the RNE transition (for example, 247.9 versus the 248.0 tie for FP8)?

kuhar:
@lua1235 can you rebase?

kuhar (commented):
A few comments

emulation/rocjitsu/lib/util/include/util/data_types.h
```diff
@@ -993,11 +1097,31 @@ inline uint16_t f32_to_f16_sr(float val, uint32_t seed) {
       f32_to_binary_float_sr(val, seed, 5, 10, 15, 15, -14, 0x7FFFu, 0x7C00u));
 }
 
+/// @brief Convert F32 to FP16 SR with MODE.FP16_OVFL overflow handling.
+/// @details When `fp16_ovfl` is set, finite overflow that rounded to INF is
+/// clamped to signed MAX_FP16; true infinities remain infinities.
+inline uint16_t f32_to_f16_sr_mode(float val, uint32_t seed, bool fp16_ovfl) {
```
kuhar, line 1103: It would be good to add direct coverage for the `fp16_ovfl=true` branch of `f32_to_f16_sr_mode`. `Fp16OvflMode` covers RNE F16 and SR BF16, while `Cdna4CvtSrTest.SingleResultF16Bf16ConsumesSeedAndFp16OvflMode` exercises F16 SR only with the bit clear, so the existing tests would not catch a regression in the finite-overflow clamp for `v_cvt_sr_f16_f32`. A small helper test plus one instruction-level overflow assertion would close the gap.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/command_processor.cpp
```diff
@@ -180,6 +180,61 @@ bool sgpr_count_is_descriptor_encoded(rj_code_arch_t arch, uint32_t sgpr_gran) {
...
+  case ROCJITSU_CODE_ARCH_CDNA1:
+  case ROCJITSU_CODE_ARCH_CDNA2:
+  case ROCJITSU_CODE_ARCH_CDNA3:
+  case ROCJITSU_CODE_ARCH_CDNA4:
+  case ROCJITSU_CODE_ARCH_RDNA1:
+  case ROCJITSU_CODE_ARCH_RDNA2:
+  case ROCJITSU_CODE_ARCH_RDNA3:
+  case ROCJITSU_CODE_ARCH_RDNA3_5:
+    return true;
+  default:
```
kuhar: Can we make the descriptor-to-MODE architecture classification explicit for RDNA4 and gfx1250? `AqlDispatchTest.InitializesModeFromComputePgmRsrc1` establishes that both intentionally omit DX10_CLAMP, IEEE_MODE, and DEBUG_MODE, but the helpers currently reach that result through `default: false`. Enumerating those architectures—and ideally avoiding a catch-all so a new enum must be classified—would keep the implementation aligned with the comment above it.

kuhar (commented):
Original anchor: `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/config/checkpoint.cpp:127`

We should preserve `WAVE_SCHED_MODE` in checkpoints alongside `MODE`. On gfx1250, HWREG 26 is user-writable and `write_raw_hwreg` stores it in the wavefront, but save serializes only `mode_raw()` and restore leaves this register at zero. A shader that writes and rereads HWREG 26 across a checkpoint therefore observes different state. Could we add the raw field to the schema/save/restore path and cover it in the checkpoint round-trip test?

emulation/rocjitsu/tests/config_test.cpp
```diff
@@ -931,6 +931,65 @@ TEST(CheckpointTest, SaveAndRestoreWave32ExecScratch) {
...
+  auto *wf = cu->dispatch_wf(0, 0, cu->config().sgprs_per_wf, cu->config().vgprs_per_wf);
+  ASSERT_NE(wf, nullptr);
+  constexpr uint32_t kStatus = 0xA5A55A5Au;
+  constexpr uint32_t kWaveSchedMode = 0x5A5AA5A5u;
+  wf->set_status_raw(kStatus);
+  wf->set_mode_raw(amdgpu::Wavefront::FP16_OVFL_BIT);
+  wf->set_wave_sched_mode_raw(kWaveSchedMode);
+  ASSERT_TRUE(wf->fp16_ovfl());
+
+  const char *path = "/tmp/rocjitsu_test_checkpoint_hwreg.bin";
```
kuhar: This is a cross-build test race. Please replace the fixed checkpoint filename with a unique per-invocation temporary path and scoped cleanup. This gtest is registered as an individual CTest case, so separate build trees or CI jobs can execute it concurrently. save_checkpoint truncates the shared file, and each process later removes it; in 25 paired runs from the default and GCC build trees, 2 of 50 invocations failed. The neighboring checkpoint tests have the same existing bug, so a small shared temp-file helper should fix the race consistently.

kuhar (commented):
I found a new bug introduced by the most recent fixes

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/hwreg.cpp
```diff
@@ -0,0 +1,494 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc.
```
amd-eochoalo, line 1: Awesome work here @lua1235

amd-eochoalo: I would prefer if this was auto-generated from the python profiles, but I don't mind using C++ to fix the current status. The only other thing I would add is an /NPI comment stating that new architectures need to define their own hardware register table.

emulation/rocjitsu/lib/python/amdisa/tests/test_profile_properties.py
```diff
@@ -303,6 +303,11 @@ def test_field_renames_vop3p(self):
     def test_field_renames_other_enc_empty(self):
         assert self.p.field_renames('ENC_VOP2') == {}
 
+    def test_hwreg_ids(self):
+        assert self.p.use_hwreg_helpers
```
amd-eochoalo: Just a small inconsistency (or maybe area for improvement)? Why are these registers indices defined in python whereas the rest are defined in C++?

lua1235: Good catch! These are from my changes before I decided to fully refactor HWREGS

amd-eochoalo (approved):
Please wait for Jakub's approval too. Just my small comment about:

* duplicating the hardware register indeces in the python profile
* Add a comment about NPI in the hwreg file (that can be found with the script under scripts/find-npi-task.sh)

kuhar (approved):

kuhar:
@lua1235 you need to rebase again

PR #9031, opened 2026-07-22
fix(rocr): Bounds-check ELF loader section and symbol indices

dayatsin-amd:
## Summary

Hardens the ROCr ELF code-object loader against out-of-bounds heap reads driven by attacker-controlled section/symbol indices when parsing untrusted code objects (reported in SWSPLAT-45972).

`GElfImage::section()` indexed the `sections` vector with unvalidated `sh_link`, `sh_info`, and `st_shndx` fields, and `GElfSymbolTable::symbol()` indexed the `symbols` vector with the unvalidated relocation `r_info` symbol index. A crafted `.hsaco` could supply an out-of-range index, causing `std::vector::operator[]` to read past the vector's backing array and later dereference a garbage pointer (deterministic SIGSEGV / DoS, with conditional info-disclosure).

This complements the already-merged loader hardening in #6994 (segment loading, note parsing, relocation offsets), which did not cover these index-based vectors.

Changes in `libamdhsacode/amd_elf_image.cpp`:
- `section()`: return `nullptr` for out-of-range indices instead of reading past the vector.
- `getStringTable()`: reject out-of-range `sh_link`.
- `getSymtab()` / `getDynsym()`: null-check the `section()` result before `->type()`.
- `GElfSymbol::section()`: reject reserved indices (`>= SHN_LORESERVE`) so a crafted `st_shndx` cannot bypass the `SHN_UNDEF` check.
- `GElfSymbolTable::symbol()`: bounds-check the `r_info` symbol index.
- `GElfSymbolTable::pullData()` / `GElfRelocationSection::pullData()`: reject the code object when the referenced string/symbol/target sections are out of range. Note `sh_info == 0` (the null section) is preserved as legitimate for dynamic relocation sections, so the index is range-checked rather than the returned pointer.

Changes in `loader/executable.cpp`:
- Null-check `GetSection()` / `section()` / `symbol()` results in `LoadDefinitionSymbol`, `SymbolAddress`, `ApplyStaticRelocation`, and `ApplyDynamicRelocation`, returning `HSA_STATUS_ERROR_INVALID_CODE_OBJECT` for malformed input.

## JIRA ID

JIRA ID : SWSPLAT-45972

## Test plan

Verified on a gfx1100 (Radeon RX 7900) system, building `libhsa-runtime64` against a current TheRock toolchain (ROCm 7.14.0a20260612, amdclang++). Compared this branch ("FIXED") against the unmodified `develop` build ("BASELINE").

- [x] Build cleanly against the toolchain — no errors, no new warnings on the changed files.
- [x] Load well-formed code objects and run existing kernels; confirm no regression.
  - HIP `vadd` (1,048,576 elements) launched via the loader: **correct results**.
  - `hsa_executable_load_agent_code_object` on valid gfx1100 code objects (with and without dynamic relocations): **`HSA_STATUS_SUCCESS`** (identical to baseline). This caught and fixed a regression where a dynamic-relocation section (`sh_info == 0`) was initially over-rejected.
- [x] Load crafted code objects with out-of-range `sh_link` / `sh_info` / `st_shndx` / `r_info` symbol index; confirm clean rejection instead of a crash.

| Crafted code object | BASELINE (`develop`) | FIXED (this PR) |
|---|---|---|
| valid object (control) | `HSA_STATUS_SUCCESS` | `HSA_STATUS_SUCCESS` |
| `sh_link` OOB in `.symtab` | **SIGSEGV** | `INVALID_CODE_OBJECT` |
| `sh_link` OOB in `.dynsym` | **SIGSEGV** | `INVALID_CODE_OBJECT` |
| `sh_info` OOB in `.rela.dyn` | **SIGSEGV** | `INVALID_CODE_OBJECT` |
| `sh_link` OOB in `.rela.dyn` | **SIGSEGV** | `INVALID_CODE_OBJECT` |
| `st_shndx` OOB in symbol entry | **SIGSEGV** | `INVALID_CODE_OBJECT` |
| `r_info` symbol index OOB in relocation | **SIGSEGV** | `INVALID_CODE_OBJECT` |

All six attacker-controlled index vectors crash the unpatched loader and are cleanly rejected with `HSA_STATUS_ERROR_INVALID_CODE_OBJECT` after this change.

- [x] HIP Catch2 module / code-object suite (gfx1100) — no regression vs. baseline `libhsa-runtime64`.
  - `unit/module/ModuleTest`: **60 passed / 2 failed / 5 skipped — identical on FIXED and BASELINE**. The 2 failures (`hipModuleGetGlobal.cc:38`, "invalid argument") reproduce byte-for-byte on the unmodified `develop` build, so they are pre-existing/environmental (single-GPU box) and not introduced by this change.
  - `unit/module/module`: **all 8 test cases pass (1,000,397 assertions)** on both FIXED and BASELINE.

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_elf_image.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_elf_image.cpp
@@ -742,14 +742,16 @@ namespace elf {
       }
       GElfSymbolTable* getSymtab(uint16_t index) override
       {
-        if (section(index)->type() == SHT_SYMTAB)
-          return static_cast<GElfSymbolTable*>(section(index));
+        GElfSection* s = section(index);
+        if (s && s->type() == SHT_SYMTAB)
+          return static_cast<GElfSymbolTable*>(s);
         return nullptr;
       }
       GElfSymbolTable* getDynsym(uint16_t index) override
       {
-        if (section(index)->type() == SHT_DYNSYM)
-          return static_cast<GElfSymbolTable*>(section(index));
+        GElfSection* s = section(index);
+        if (s && s->type() == SHT_DYNSYM)
+          return static_cast<GElfSymbolTable*>(s);
         return nullptr;
       }
 
@@ -772,7 +774,12 @@ namespace elf {
       GElfSegment* segment(size_t i) override { return segments[i].get(); }
       Segment* segmentByVAddr(uint64_t vaddr) override;
       size_t sectionCount() override { return sections.size(); }
-      GElfSection* section(size_t i) override { return sections[i].get(); }
+      // Bounds-check attacker-controlled section indices (sh_link, sh_info,
+      // st_shndx). Out-of-range indices return nullptr instead of reading heap
+      // memory past the vector's internal array.
+      GElfSection* section(size_t i) override {
+        return i < sections.size() ? sections[i].get() : nullptr;
+      }
       Section* sectionByVAddr(uint64_t vaddr) override;
       uint16_t machine() const;
       uint16_t etype() const;
@@ -1132,10 +1139,15 @@ namespace elf {
 
     Section* GElfSymbol::section()
     {
-      if (Sym()->st_shndx != SHN_UNDEF) {
-        return symtab->elf->section(Sym()->st_shndx);
+      uint16_t shndx = Sym()->st_shndx;
+      // Reserved indices (SHN_LORESERVE..SHN_HIRESERVE, e.g. SHN_ABS,
+      // SHN_COMMON) do not refer to a real section, and any index >= the
+      // section count is out of bounds. section() itself bounds-checks, but
+      // guard here so a crafted st_shndx does not bypass the SHN_UNDEF check.
+      if (shndx == SHN_UNDEF || shndx >= SHN_LORESERVE) {
+        return 0;
       }
-      return 0;
+      return symtab->elf->section(shndx);
     }
 
     bool GElfSymbol::push(const std::string& name, uint64_t value, uint64_t size, unsigned char type, unsigned char binding, uint16_t shndx, unsigned char other)
@@ -1170,7 +1182,11 @@ namespace elf {
 
     bool GElfSymbolTable::pullData()
     {
+      // sh_link identifies this symbol table's string table; a crafted value
+      // out of section-table range yields nullptr. Reject the code object
+      // rather than storing a garbage string-table pointer.
       strtab = elf->getStringTable(hdr.sh_link);
+      if (!strtab) { return false; }
       for (size_t i = 0; i < data0.size() / sizeof(GElf_Sym); ++i) {
         symbols.push_back(std::unique_ptr<GElfSymbol>(new GElfSymbol(this, data0, i * sizeof(GElf_Sym))));
       }
@@ -1204,7 +1220,10 @@ namespace elf {
 
     Symbol* GElfSymbolTable::symbol(size_t i)
     {
-      return symbols[i].get();
+      // i derives from the attacker-controlled relocation r_info symbol index
+      // (GELF_R_SYM); reject out-of-range values instead of reading past the
+      // symbols vector.
+      return i < symbols.size() ? symbols[i].get() : nullptr;
     }
 
     GElfNoteSection::GElfNoteSection(GElfImage* elf)
@@ -1319,8 +1338,20 @@ namespace elf {
 
     bool GElfRelocationSection::pullData()
     {
+      // sh_info/sh_link identify the target section and referenced symbol
+      // table and are attacker-controlled. sh_info == 0 (the SHN_UNDEF/null
+      // section) is legitimate: it marks a dynamic relocation section, and
+      // section() then returns nullptr, which the loader uses to route dynamic
+      // relocations. Only an out-of-range sh_info is malformed, so validate the
+      // index rather than the returned pointer.
+      if (hdr.sh_info >= elf->sectionCount()) { return false; }
       section = elf->section(hdr.sh_info);
+      // sh_link must reference the associated symbol table;
+      // getReferencedSymbolTable() returns nullptr for an out-of-range index
+      // (section() is bounds-checked). Reject rather than storing a garbage
+      // pointer that later gets virtual-dispatched during relocation.
       symtab = elf->getReferencedSymbolTable(hdr.sh_link);
+      if (!symtab) { return false; }
       Elf_Scn *lScn = elf_getscn(elf->e, ndxscn);
       assert(lScn);
       Elf_Data *lData = elf_getdata(lScn, nullptr);
@@ -1651,6 +1682,9 @@ namespace elf {
 
     GElfStringTable* GElfImage::getStringTable(uint16_t index)
     {
+      // index derives from the attacker-controlled sh_link field; reject
+      // out-of-range values instead of reading past the sections vector.
+      if (index >= sections.size()) { return nullptr; }
       return static_cast<GElfStringTable*>(sections[index].get());
     }
 
--- a/projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_elf_image.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_elf_image.cpp
@@ -738,7 +738,15 @@ namespace elf {
       GElfStringTable* strtab() override;
       GElfSymbolTable* getReferencedSymbolTable(uint16_t index)
       {
-        return static_cast<GElfSymbolTable*>(section(index));
+        // index derives from an attacker-controlled sh_link. section() already
+        // bounds-checks and returns nullptr for an out-of-range index, but a
+        // crafted value may reference an in-range wrong-type section. Verify it
+        // is actually a symbol table so the static_cast cannot produce a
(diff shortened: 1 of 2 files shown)


projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_elf_image.cpp
```diff
@@ -1319,8 +1338,20 @@ namespace elf {
...
     bool GElfRelocationSection::pullData()
     {
+      // sh_info/sh_link identify the target section and referenced symbol
+      // table and are attacker-controlled. sh_info == 0 (the SHN_UNDEF/null
+      // section) is legitimate: it marks a dynamic relocation section, and
+      // section() then returns nullptr, which the loader uses to route dynamic
+      // relocations. Only an out-of-range sh_info is malformed, so validate the
+      // index rather than the returned pointer.
+      if (hdr.sh_info >= elf->sectionCount()) { return false; }
       section = elf->section(hdr.sh_info);
```
cfreeamd, line 1356: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** The comment above states that for the legitimate `sh_info == 0` case, `section()` "returns nullptr, which the loader uses to route dynamic relocations." But `section(0)` is in bounds (`0 < sections.size()`), so it returns `sections[0].get()` — the populated `SHT_NULL` reserved section object, not `nullptr` — unless index 0 is deliberately stored as a null `unique_ptr`. If so, the `section` member ends up pointing at the null section rather than being null, and the stated dynamic-relocation routing rationale may not hold. The guard itself (`sh_info >= sectionCount()`) is fine; just want to confirm `section(0)` behavior and that dynamic relocations are still routed as intended.

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_elf_image.cpp
```diff
@@ -1651,6 +1682,9 @@ namespace elf {
 
     GElfStringTable* GElfImage::getStringTable(uint16_t index)
     {
+      // index derives from the attacker-controlled sh_link field; reject
+      // out-of-range values instead of reading past the sections vector.
+      if (index >= sections.size()) { return nullptr; }
       return static_cast<GElfStringTable*>(sections[index].get());
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** After the new `index >= sections.size()` bounds check, this still does `static_cast<GElfStringTable*>(sections[index].get())` without verifying the target is actually a string table (`SHT_STRTAB`). A crafted `sh_link` pointing at an in-range but wrong-type section (e.g. `SHT_PROGBITS`) yields a mis-typed pointer later used as a string table — a residual type-confusion in the same function this PR hardens. Consider also checking `sections[index]->type() == SHT_STRTAB` and returning nullptr otherwise; callers already handle nullptr after this PR.

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_elf_image.cpp
```diff
@@ -772,7 +774,12 @@ namespace elf {
       GElfSegment* segment(size_t i) override { return segments[i].get(); }
       Segment* segmentByVAddr(uint64_t vaddr) override;
       size_t sectionCount() override { return sections.size(); }
-      GElfSection* section(size_t i) override { return sections[i].get(); }
+      // Bounds-check attacker-controlled section indices (sh_link, sh_info,
+      // st_shndx). Out-of-range indices return nullptr instead of reading heap
+      // memory past the vector's internal array.
+      GElfSection* section(size_t i) override {
+        return i < sections.size() ? sections[i].get() : nullptr;
```
cfreeamd, line 789: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** `section()` can now return `nullptr`, changing its contract. The diff updates the call sites that need it (getSymtab/getDynsym, getStringTable, the executable.cpp accessors). Worth a quick grep to confirm no other caller of `section()` / `getReferencedSymbolTable()` dereferences the result without a null check now that the contract changed.

cfreeamd (commented):

Reviewed the ELF-loader bounds-checking hardening. `executable.cpp` null guards (LoadDefinitionSymbol, SymbolAddress, ApplyStaticRelocation, ApplyDynamicRelocation) and the `section()`/`symbol()`/`getStringTable()` bounds checks in `amd_elf_image.cpp` look correct and fail closed. A few non-blocking notes inline.

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_elf_image.cpp
dayatsin-amd: Confirmed: `section(0)` returns `nullptr`, so the rationale holds. During read, `pullElf()` stores the `SHT_NULL` section as an empty `std::unique_ptr`:

```cpp
} else if (shdr.sh_type == SHT_NULL) {
  section = 0;
  sections.push_back(std::unique_ptr<GElfSection>());
}
```

So `sections[0].get()` is `nullptr` and `section(0)` returns `nullptr` (there is no populated `SHT_NULL` object at index 0). A legitimate dynamic relocation section with `sh_info == 0` therefore gets `section == nullptr`, which is what the loader uses to route dynamic relocations. Only an out-of-range `sh_info` is rejected. No change needed.

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_elf_image.cpp
dayatsin-amd: Good catch — fixed in 3df30b65cb. `getStringTable()` now rejects an in-range but wrong-type section:

```cpp
if (index >= sections.size()) { return nullptr; }
GElfSection* s = sections[index].get();
if (!s || s->type() != SHT_STRTAB) { return nullptr; }
return static_cast<GElfStringTable*>(s);
```

The `!s` guard also covers section index 0 (`SHT_NULL`), which is stored as a null `unique_ptr`. Callers already handle `nullptr`.

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/amd_elf_image.cpp
dayatsin-amd: Grepped all callers. Findings:

- **`executable.cpp`** (`GetSection()`/`section()`/`symbol()`): all updated in this PR to null-check before dereferencing.
- **`getReferencedSymbolTable()`**: this was the one raw `static_cast<GElfSymbolTable*>(section(index))` with no type check. Hardened in 3df30b65cb to require `SHT_SYMTAB`/`SHT_DYNSYM` (same type-confusion class as the string-table comment), so its only caller (`GElfRelocationSection::pullData`) still gets `nullptr` for both out-of-range and wrong-type `sh_link`.
- **`amd_hsa_code.cpp`** (`elfsym->section()->flags()` etc. in `Symbol::IsAgent/Segment/Alignment/IsConst`): these dereference without a null check, but that is pre-existing and not a contract change from this PR — `GElfSymbol::section()` already returned `0` for `SHN_UNDEF` before this PR, and `section(0)` (SHT_NULL) already returned `nullptr`. This PR only changed the *out-of-range* path (previously OOB read) to also return `nullptr`, and `GElfSymbol::section()` filters `>= SHN_LORESERVE`/out-of-range to `0` before reaching those accessors. I've left those out of scope to keep this PR focused on the reported issue.

cfreeamd (commented):

All three prior findings are resolved in commit 3df30b65cb:

- **[Question] `section(0)`** — confirmed via `pullElf()` that an `SHT_NULL` section is stored as an empty `unique_ptr`, so `section(0)` genuinely returns `nullptr`; the routing rationale holds. Concern dropped.
- **[Optional] `getStringTable()` type confusion** — fixed: now rejects an in-range wrong-type section (`!s || s->type() != SHT_STRTAB`).
- **[Optional] audit remaining `section()` callers** — `getReferencedSymbolTable()` had the same raw `static_cast` and is now hardened to require `SHT_SYMTAB`/`SHT_DYNSYM`; other callers were already null-guarded.

The follow-up commit adds real type-confusion hardening beyond the original scope. One non-blocking, out-of-scope note for a possible separate follow-up: the `amd_hsa_code.cpp` symbol accessors (`Symbol::IsAgent`/`IsSegment`/`GetAlignment`/`IsConst`) still dereference `elfsym->section()` without a null check — a crafted out-of-range `st_shndx` would now yield a deterministic null-deref (DoS only, not the prior exploitable OOB + garbage virtual dispatch), so this PR is a strict improvement there. Looks good to me.

cfreeamd (approved):

shwetagkhatri (approved):

amd-justchen:
﻿Will merge given code owner approval.
The TheRock CI failures are only known infra issues, unrelated to the rocr ELF loader bounds-check change:
- `Windows (gfx1151) / Build Windows Packages`: known missing win_flex/winflexbison toolchain in amd-mesa (meson), fixed forward by #10951
- `Linux hip-tests (gfx94X) shard 2 of 4`: 1 of 1254 tests, the known-flaky Unit_HRR_ZeroInitRoundtrip roundtrip test, unrelated to a loader bounds-check
- `Linux rocgdb-gpu (gfx94X)`: expected-fail (xfail) lane
- `Linux MI455 Test sanity`: container sanity check cancelled, zero signal

PR #9033, opened 2026-07-22
fix(amdsmi): correctness and coverage fixes for the Python packaging rework

marifamd:
## Description

Correctness and coverage follow-ups to the AMD SMI Python packaging rework (relanded in #9340). This branch is the packaging-hardening layer on top of the SONAME conflict test + build harness (#7912).

Rebased onto `develop` after #7912 and #9340 merged; the now-upstream shebang and self-locate-loader commits were dropped, and the surviving relocatable-loader change was reconciled with develop's version (keeps develop's `len(parents) > 3` guard and adds a `try/except OSError` fall-through so a present-but-unloadable relocatable library does not shadow the system SONAME).

### Loader / wrapper
- Fall through to the system library when the relocatable TheRock library exists but fails to load (missing deps).
- Dedupe the wrapper `_libraries` init (the ctypesgen preamble already initializes it).
- Read and write the staged wrapper as UTF-8 in `disable_system_fallback.py` so the wheel flag flip is deterministic regardless of build locale; make the staging idempotent so a rebuild that reuses the flipped copy succeeds.

### Packaging correctness
- Stop the RPM from co-owning system Python directories.
- Restrict package `preun`/`prerm` cleanup to a full erase (not upgrades).
- Quote the install prefix when writing `ld.so.conf.d` on RPM.
- Pin permissions on the installed Python module; make the operational log group-writable (not world-writable) and stop the logger from restoring world-writable permissions.
- Only strip release libraries when a strip tool is available; fail the build when the system `python3` cannot be detected.

### Wheel build
- Build one wheel per platform (not per interpreter), tag it `manylinux_2_28`, and verify the bundled `.so`.
- Use a clean PyPI version and expose the commit as metadata.
- Reject regenerating the wrapper against the wheel library; skip the system site-packages install for wheel builds.

### Tests
- Loader-contract tests in `test_abi_compat.py`: preferred bundled `.so`, wheel refuses system fallback when the bundle is missing, the relocatable OSError fall-through, and an ambiguous-wrapper hard failure. Assert the idempotent no-op in the fallback double-run guard.
- Guard against drift between the two installed module copies; install the dual-copy runner so its guard tests run in CI; run the wheel install-location check without a GPU.

### Verification
- `cmake -DBUILD_TESTS=ON` + `make`: 0 errors, `libamd_smi.so.27` + `amdsmitst` built.
- `test_abi_compat.py` (10) and `test_dual_copy_guard.py` (4): all pass.

## JIRA ID

ROCM-3941


.github/workflows/amdsmi-manylinux-build.yml
```diff
@@ -94,6 +94,12 @@ jobs:
             exit 1
           fi
 
+      - name: Verify system/wheel SONAMEs are distinct
+        run: |
+          ${{matrix.target.python-bin}} \
+            "$GITHUB_WORKSPACE/projects/amdsmi/tests/run_amdsmi_pkg_conflict_test.py" \
+            --build-root /tmp/amdsmi-build
```
geomin12: kinda confused on where this /tmp build comes from

geomin12 (commented):
lgtm! but this is majority of amd-smi code changes, i would appreciate an amd-smi team member(s) to review and approve

.github/workflows/amdsmi-manylinux-build.yml
marifamd: it's the build dir for build_wheel.py, do you have a better alternative/convention? I can add a mention in the packaging.md if that would suffice.

dmitrii-galantsev (approved):

gabrpham (approved):

PR #9035, opened 2026-07-22
test(amdsmi): validate the SLES RPM interpreter dependency in CI

marifamd:
## Description

Validate the SLES RPM interpreter dependency in CI. Stacks on #9034 (it invokes `run_amdsmi_cpack_path_test.py`, which #9034 adds).

- Add a SLES packaging job that builds the RPM and asserts the interpreter dependency resolves to a real SLES package (`pythonXY`, not the RHEL `python(abi) = 3.X` or bare `python3` forms).
- Give the SLES job a working toolchain and dependency check; match only the SLES `pythonXY` token in the dependency check; sparse-checkout the SLES packaging job.
- Bound the job with an explicit `timeout-minutes` so a hung container fails fast instead of inheriting the six-hour default.

## JIRA ID

ROCM-3941

### DIFF
--- a/.github/workflows/amdsmi-python-versions.yml
+++ b/.github/workflows/amdsmi-python-versions.yml
@@ -1,0 +1,86 @@
+# ---------------------------------------------------------------------------
+# AMD SMI Python-version loader tests.
+#
+# The amdsmi loader and wrapper must behave identically across every CPython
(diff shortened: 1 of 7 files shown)


gabrpham (approved):

dmitrii-galantsev (approved):

ammallya:
Flaky github fetch sources error, build and test for smi are good

PR #9037, opened 2026-07-22
ConSan improvements (gfx1250 and RDNA4)

bjacob:


kuhar (commented):
Please run the repository's ClangFormat style over the PR-owned C/C++ edits. `clang-format-18 --dry-run --Werror` currently reports changes in `waitcheck.cpp`, `consan_moi_barrier.inc`, `consan_moi_core_types.h.inc`, `consan_moi_pipeline.inc`, `consan_moi_sampled_access.inc`, `consan_sync_analysis.inc`, `consan_validation.inc`, `rj_hsa_dbi_hooks.cpp`, `analysis_test.cpp`, `consan_test_support.h`, and `moi_common_test.cpp`.

The new full-aperture cases at `moi_common_test.cpp:1262` and `moi_common_test.cpp:1284` cannot satisfy the report planner as written. On the current head, `Gfx1250AutoReportUsesExternalApertureForDescriptorOpaqueLds` and `Gfx1250AutoReportCoversFullApertureForDynamicLds` both fail because the 320 KiB aperture is multiplied by 256 dispatch banks: the plan requires 503,324,672 bytes, above its 134,217,728-byte ceiling, and returns `InsufficientReportCapacity`. Please either make the full-aperture layout fit the planner's frozen ceiling or change this path and its tests to exercise the intended fail-closed rejection; the current assertions leave the ConSan unit suite red.

`emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/consan/consan_moi_placement.inc:2754`: Please do not treat a missing hardware dispatch ID as a usable zero-generation Record/Replay record. The first-light slots are retained for the lifetime of the loaded code object, so one site can be claimed in dispatch A and another site in dispatch B. For this fallback both records leave `generation == 0`; replay then substitutes the report header generation (`consan_moi_model.cpp:1205`) and places them in the same workgroup state, allowing a conflict to be reported between accesses that never belonged to the same dispatch. That defeats the dispatch-generation isolation this change adds and can create false sanitizer diagnostics for exactly the full-pressure gfx1250 owners admitted here. Please either preserve a real per-dispatch identity by another mechanism, or classify/exclude these owners as explicit undercoverage instead of replaying zero as a common generation.

kuhar (approved):

PR #9044, opened 2026-07-22
[rocprofiler-systems] Fix thread-limit tests and fix silent CTest generation failure

kcossett-amd:
## Motivation

`ROCPROFSYS_MAX_THREADS` is a compile time constant that is the core of the `thread-limit` tests. For this test to work on TheRock, we need a way to get the compile time value at runtime and run the test with it. The idea here is to store it within `rocprof-sys-avail`, then read it and run the test against that value.

## Technical Details

**Regarding thread-limit**
 - Store `ROCPROFSYS_MAX_THREADS` in `rocprof-sys-avail`. Accessed using `rocprof-sys-avail --max-threads`.
 - Make` test_categories.yaml` copied to build folder during build time, not configure time.
 - Gate per thread finalization logs on `ROCPROFSYS_VERBOSE>0` or in causal mode.
   - This gets rid of some of the many logs that would be seen during this test. I will be leaving the messages that say:
   `[SIG27] Sampler for thread <N> will be triggered 250.0x per second of CPU-time (every 4.000e+00 milliseconds)... `

**Regarding silent failure**:
 - CTest generation step has been silently failing on TheRock regarding thread-limit test as the `pytest.exit(0)` would mask any error during generation. To fix this, we detect any collection errors and report them, failing the build. 
 - Moved `get_thread_limit()` into the `test` definition itself. Evaluating it during build (during CTesttestfile.cmake generation step) would cause a failure in TheRock CI.

## Issue Tracking

Jira ID: AIPROFSYST-523
Jira ID: AIPROFSYST-656
Jira ID: ROCM-29285

## Test Plan

Examining TheRock CI workflow to see that the test ran and passed.

## Test Result

Ongoing.

Test passes in TheRock CI: https://github.com/ROCm/rocm-systems/actions/runs/30031596021/job/89312443869?pr=9044
```
        Start 546: thread-limit-half-sampling
420/498 Test #546: thread-limit-half-sampling ......................................................................................................................................................   Passed    4.60 sec
        Start 547: thread-limit-half-binary-rewrite
421/498 Test #547: thread-limit-half-binary-rewrite ................................................................................................................................................   Passed    5.49 sec
        Start 548: thread-limit-half-runtime-instrument
422/498 Test #548: thread-limit-half-runtime-instrument ............................................................................................................................................   Passed    7.47 sec
        Start 549: thread-limit-half-sys-run
423/498 Test #549: thread-limit-half-sys-run .......................................................................................................................................................   Passed    4.63 sec
        Start 550: thread-limit-at-limit-sampling
424/498 Test #550: thread-limit-at-limit-sampling ..................................................................................................................................................   Passed    7.57 sec
        Start 551: thread-limit-at-limit-binary-rewrite
425/498 Test #551: thread-limit-at-limit-binary-rewrite ............................................................................................................................................   Passed    8.40 sec
        Start 552: thread-limit-at-limit-runtime-instrument
426/498 Test #552: thread-limit-at-limit-runtime-instrument ........................................................................................................................................   Passed   11.47 sec
        Start 553: thread-limit-at-limit-sys-run
427/498 Test #553: thread-limit-at-limit-sys-run ...................................................................................................................................................   Passed    7.76 sec
        Start 554: thread-limit-exhausted-sampling
428/498 Test #554: thread-limit-exhausted-sampling .................................................................................................................................................   Passed    8.72 sec
        Start 555: thread-limit-exhausted-binary-rewrite
429/498 Test #555: thread-limit-exhausted-binary-rewrite ...........................................................................................................................................   Passed    9.85 sec
        Start 556: thread-limit-exhausted-runtime-instrument
430/498 Test #556: thread-limit-exhausted-runtime-instrument .......................................................................................................................................   Passed   12.59 sec
        Start 557: thread-limit-exhausted-sys-run
431/498 Test #557: thread-limit-exhausted-sys-run ..................................................................................................................................................   Passed    8.47 sec
```

## Submission Checklist

### DIFF
--- a/projects/rocprofiler-systems/source/bin/rocprof-sys-avail/avail.cpp
+++ b/projects/rocprofiler-systems/source/bin/rocprof-sys-avail/avail.cpp
@@ -234,6 +234,17 @@ main(int argc, char** argv)
                       "Write the available hardware counters")
         .max_count(1);
 
+    parser
+        .add_argument({ "--max-threads" },
+                      "Print the compile-time limit on the total number of threads that "
+                      "can be profiled in a single process over its lifetime and exit. "
+                      "Thread slots are counted cumulatively and are not reused when a "
+                      "thread exits")
+        .count(0)
+        .action([](parser_t&) {
+            std::cout << "ROCPROFSYS_MAX_THREADS=" << ROCPROFSYS_MAX_THREADS << "\n";
+        });
+
     parser.add_argument({ "-a", "--all" }, "Print all available info")
         .max_count(1)
         .action([&](parser_t& p) {
@@ -676,7 +687,8 @@ main(int argc, char** argv)
     }
 
     if(parser.exists("list-categories") || parser.exists("list-keys") ||
-       parser.exists("list-operations") || parser.exists("list-domains"))
+       parser.exists("list-operations") || parser.exists("list-domains") ||
+       parser.exists("max-threads"))
         return EXIT_SUCCESS;
 
     std::string _pos_regex{};
--- a/projects/rocprofiler-systems/source/bin/rocprof-sys-avail/avail.cpp
+++ b/projects/rocprofiler-systems/source/bin/rocprof-sys-avail/avail.cpp
@@ -242,7 +242,10 @@ main(int argc, char** argv)
                       "thread exits")
         .count(0)
         .action([](parser_t&) {
-            std::cout << "ROCPROFSYS_MAX_THREADS=" << ROCPROFSYS_MAX_THREADS << "\n";
+            // NOTE: test_thread_limit.py depends on the wording here to capture
+            // the compile-time value. Any wording change must be reflected in that file.
+            std::cout << "Compile-time limit on the total number of threads: "
+                      << ROCPROFSYS_MAX_THREADS << "\n";
         });
 
     parser.add_argument({ "-a", "--all" }, "Print all available info")
--- a/projects/rocprofiler-systems/source/bin/rocprof-sys-avail/avail.cpp
+++ b/projects/rocprofiler-systems/source/bin/rocprof-sys-avail/avail.cpp
@@ -242,9 +242,11 @@ main(int argc, char** argv)
                       "thread exits")
         .count(0)
         .action([](parser_t&) {
-            // NOTE: test_thread_limit.py depends on the wording here to capture
-            // the compile-time value. Any wording change must be reflected in that file.
-            std::cout << "Compile-time limit on the total number of threads: "
+            // NOTE: capabilities.py max_threads method depends on the wording here to
+            // capture the compile-time value. Any wording change must be reflected in
+            // that file.
+            std::cout << "Compile-time limit on the total number of threads "
+                         "(ROCPROFSYS_MAX_THREADS): "
                       << ROCPROFSYS_MAX_THREADS << "\n";
         });
 
--- a/projects/rocprofiler-systems/source/lib/rocprof-sys/library.cpp
+++ b/projects/rocprofiler-systems/source/lib/rocprof-sys/library.cpp
@@ -1120,7 +1120,7 @@ rocprofsys_finalize_hidden(void)
     // if they are still running (e.g. thread-pool still alive), the
     // thread-specific data will be wrong if try to stop them from
     // the main thread.
-    auto _thr_verbose = (config::get_use_causal() || config::get_verbose() > 0) ? 1 : 0;
+    const bool _thr_verbose = config::get_use_causal() || config::get_verbose() > 0;
     if(thread_data<thread_bundle_t>::get())
     {
         for(auto& itr : *thread_data<thread_bundle_t>::get())
@@ -1131,7 +1131,7 @@ rocprofsys_finalize_hidden(void)
                 std::string _msg = itr->as_string();
                 auto        _pos = _msg.find(">>>  ");
                 if(_pos != std::string::npos) _msg = _msg.substr(_pos + 5);
-                if(_thr_verbose > 0)
+                if(_thr_verbose)
                 {
                     LOG_INFO("{}", _msg);
                 }
--- a/projects/rocprofiler-systems/source/lib/rocprof-sys/library.cpp
+++ b/projects/rocprofiler-systems/source/lib/rocprof-sys/library.cpp
@@ -1120,7 +1120,7 @@ rocprofsys_finalize_hidden(void)
     // if they are still running (e.g. thread-pool still alive), the
     // thread-specific data will be wrong if try to stop them from
     // the main thread.
-    auto _thr_verbose = (config::get_use_causal()) ? 1 : 0;
+    auto _thr_verbose = (config::get_use_causal() || config::get_verbose() > 0) ? 1 : 0;
     if(thread_data<thread_bundle_t>::get())
     {
         for(auto& itr : *thread_data<thread_bundle_t>::get())
@@ -1131,7 +1131,7 @@ rocprofsys_finalize_hidden(void)
                 std::string _msg = itr->as_string();
                 auto        _pos = _msg.find(">>>  ");
                 if(_pos != std::string::npos) _msg = _msg.substr(_pos + 5);
-                if(_thr_verbose >= 0)
+                if(_thr_verbose > 0)
                 {
                     LOG_INFO("{}", _msg);
                 }
--- a/projects/rocprofiler-systems/tests/pytest/conftest.py
+++ b/projects/rocprofiler-systems/tests/pytest/conftest.py
@@ -612,10 +612,8 @@ def pytest_collection_modifyitems(config, items) -> None:
                 )
 
 
-# Modules that fail to collect (import errors, parametrize-time crashes, ...)
-# are recorded here so CTest generate mode can fail the build loudly instead of
-# silently omitting their tests from the generated suite.
-_collection_errors: list = []
+# Modules that fail to collect are recorded here so CTest generate mode can fail
+_collection_errors: list[pytest.CollectReport] = []
 
 
 def pytest_collectreport(report) -> None:
@@ -624,7 +622,7 @@ def pytest_collectreport(report) -> None:
         _collection_errors.append(report)
 
 
-def pytest_collection_finish(session):
+def pytest_collection_finish(session) -> None:
     if session.config.getoption("--ctest-mode", default="off") == "generate":
         if _collection_errors:
             failures = "\n".join(f"  - {r.nodeid}" for r in _collection_errors)
--- a/projects/rocprofiler-systems/tests/pytest/conftest.py
+++ b/projects/rocprofiler-systems/tests/pytest/conftest.py
@@ -612,8 +612,26 @@ def pytest_collection_modifyitems(config, items) -> None:
                 )
 
 
+# Modules that fail to collect (import errors, parametrize-time crashes, ...)
+# are recorded here so CTest generate mode can fail the build loudly instead of
+# silently omitting their tests from the generated suite.
+_collection_errors: list = []
+
+
+def pytest_collectreport(report) -> None:
+    """Record collection failures for CTest generate mode's abort check."""
+    if report.failed:
+        _collection_errors.append(report)
+
+
 def pytest_collection_finish(session):
     if session.config.getoption("--ctest-mode", default="off") == "generate":
+        if _collection_errors:
+            failures = "\n".join(f"  - {r.nodeid}" for r in _collection_errors)
+            pytest.exit(
+                f"CTestTestfile.cmake generation failed due to:\n{failures}",
+                returncode=1,
+            )
         raw_path = session.config.getoption("--ctest-output-path", default=None)
         output_path = Path(raw_path) if raw_path else None
         _ctest_generate_tests(session.items, output_path)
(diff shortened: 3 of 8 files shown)


projects/rocprofiler-systems/source/bin/rocprof-sys-avail/avail.cpp
```diff
@@ -234,6 +234,17 @@ main(int argc, char** argv)
...
 
+    parser
+        .add_argument({ "--max-threads" },
+                      "Print the compile-time limit on the total number of threads that "
+                      "can be profiled in a single process over its lifetime and exit. "
+                      "Thread slots are counted cumulatively and are not reused when a "
+                      "thread exits")
+        .count(0)
+        .action([](parser_t&) {
+            std::cout << "ROCPROFSYS_MAX_THREADS=" << ROCPROFSYS_MAX_THREADS << "\n";
```
mradosav-amd:
```suggestion
            std::cout << "Compile-time limit on the total number of threads: " << ROCPROFSYS_MAX_THREADS << "\n";
```

Maybe like this is better for user experience.

kcossett-amd: Good point!

projects/rocprofiler-systems/source/bin/rocprof-sys-avail/avail.cpp
```diff
@@ -234,6 +234,20 @@ main(int argc, char** argv)
...
 
+    parser
+        .add_argument({ "--max-threads" },
+                      "Print the compile-time limit on the total number of threads that "
+                      "can be profiled in a single process over its lifetime and exit. "
+                      "Thread slots are counted cumulatively and are not reused when a "
+                      "thread exits")
+        .count(0)
+        .action([](parser_t&) {
+            // NOTE: test_thread_limit.py depends on the wording here to capture
```
dgaliffiAMD: **Comment points at the wrong file.** The "NOTE" says "`test_thread_limit.py` depends on the wording," but the regex actually lives in `capabilities.py` (`max_threads`). Redirect the pointer.

projects/rocprofiler-systems/tests/pytest/rocprofsys/capabilities.py
```diff
@@ -390,6 +390,34 @@ def papi_availability(self) -> bool:
         except (subprocess.SubprocessError, OSError, subprocess.TimeoutExpired):
             return False
 
+    @persistent_cached_property
+    def max_threads(self) -> int:
+        """Compile-time maximum number of threads that can be profiled at once.
```
dgaliffiAMD: **Docstring says "at once" (concurrency), source says cumulative-over-lifetime** — This describes max_threads as a concurrency cap, but `avail.cpp:238-242` defines it as a lifetime/cumulative limit (slots not reused). Reword to match.

projects/rocprofiler-systems/source/bin/rocprof-sys-avail/avail.cpp
```diff
@@ -234,6 +234,20 @@ main(int argc, char** argv)
...
+            // the compile-time value. Any wording change must be reflected in that file.
+            std::cout << "Compile-time limit on the total number of threads: "
+                      << ROCPROFSYS_MAX_THREADS << "\n";
```
dgaliffiAMD:
```suggestion
            std::cout << "Compile-time limit on the total number of threads (ROCPROFSYS_MAX_THREADS): "
                      << ROCPROFSYS_MAX_THREADS << "\n";
```
I like your change. It reads better. However, I would also include a way to anchor it to the actual CMAKE flag, like the suggestion above.

projects/rocprofiler-systems/source/lib/rocprof-sys/library.cpp
```diff
@@ -1118,7 +1118,7 @@ rocprofsys_finalize_hidden(void)
     // if they are still running (e.g. thread-pool still alive), the
     // thread-specific data will be wrong if try to stop them from
     // the main thread.
-    auto _thr_verbose = (config::get_use_causal()) ? 1 : 0;
+    auto _thr_verbose = (config::get_use_causal() || config::get_verbose() > 0) ? 1 : 0;
```
dgaliffiAMD: It seems that this is now just a `bool`. Why not change your variable to make it explicit and little easier to read.

```suggestion
    const bool _thr_verbose = config::get_use_causal() || config::get_verbose() > 0;
```

Then, changing the condition to:

```
if(_thr_verbose)
```

projects/rocprofiler-systems/tests/pytest/conftest.py
```diff
@@ -610,8 +610,26 @@ def base_modifications(item: pytest.Item) -> None:
...
+_collection_errors: list = []
+
+
+def pytest_collectreport(report) -> None:
+    """Record collection failures for CTest generate mode's abort check."""
+    if report.failed:
+        _collection_errors.append(report)
+
+
 def pytest_collection_finish(session):
```
dgaliffiAMD:
```suggestion
def pytest_collection_finish(session) -> None:
```

projects/rocprofiler-systems/tests/pytest/conftest.py
```diff
@@ -610,8 +610,26 @@ def base_modifications(item: pytest.Item) -> None:
                 )
 
 
+# Modules that fail to collect (import errors, parametrize-time crashes, ...)
+# are recorded here so CTest generate mode can fail the build loudly instead of
+# silently omitting their tests from the generated suite.
+_collection_errors: list = []
```
dgaliffiAMD:
```suggestion
_collection_errors: list[pytest.CollectReport] = []
```
Nit-Pick: Add annotations for list type.

projects/rocprofiler-systems/tests/pytest/test_thread_limit.py
```diff
@@ -104,22 +83,20 @@ def get_thread_limit_warning_regex(thread_limit: int) -> str:
...
 class TestThreadLimit(RocprofsysTest):
     BINARY_REWRITE_ARGS = ["-e", "-v", "2", "-i", "1024", "--label", "return", "args"]
     RUNTIME_INSTRUMENT_ARGS = ["-e", "-v", "1", "-i", "1024", "--label", "return", "args"]
 
-    def test(self, mode, thread_count, thread_limit_env):
+    def test(self, mode, thread_ratio, thread_limit_env):
+        thread_limit = get_thread_limit()
+        if thread_ratio == "half":
+            thread_count = thread_limit // 2
+        elif thread_ratio == "at":
```
dgaliffiAMD: Pre-existing, but since you're modifying it already -- "at_limit" or "exact" may read a little better than simply "at".

projects/rocprofiler-systems/tests/pytest/test_thread_limit.py
```diff
@@ -104,22 +83,20 @@ def get_thread_limit_warning_regex(thread_limit: int) -> str:
...
-@pytest.mark.parametrize(
-    "thread_count",
-    [
-        get_thread_limit() // 2,
-        get_thread_limit(),
-        get_thread_limit() * 2,
-    ],
-    ids=["half", "at", "double"],
-)
-@pytest.mark.timeout(480)
```
dgaliffiAMD, line 116: Just checking ... did you mean to remove the timeout?
It was `@pytest.mark.timeout(480)`.

projects/rocprofiler-systems/tests/pytest/test_thread_limit.py
```diff
@@ -104,22 +83,20 @@ def get_thread_limit_warning_regex(thread_limit: int) -> str:
...
+@pytest.mark.parametrize("thread_ratio", ["half", "at", "double"])
 @pytest.mark.class_name("thread-limit")
 class TestThreadLimit(RocprofsysTest):
     BINARY_REWRITE_ARGS = ["-e", "-v", "2", "-i", "1024", "--label", "return", "args"]
     RUNTIME_INSTRUMENT_ARGS = ["-e", "-v", "1", "-i", "1024", "--label", "return", "args"]
 
-    def test(self, mode, thread_count, thread_limit_env):
+    def test(self, mode, thread_ratio, thread_limit_env):
+        thread_limit = get_thread_limit()
+        if thread_ratio == "half":
```
dgaliffiAMD: This is also pre-existing, but it would be easier to read if each case's parameters were defined together, rather than in separate helper methods. 

| label | count | pass_value | fail_value |
| --- | --- | --- | --- |
| "half" | (limit / 2) | (count - 1) | (count +1) |
| "at" | limit | ((limit - 1) - INTERNAL_THREAD_OFFSET) | (limit +1) |

dgaliffiAMD: Maybe something like this - https://github.com/ROCm/rocm-systems/commit/d30ec23d3100a7527f945312554e0962f0420194

projects/rocprofiler-systems/tests/pytest/test_thread_limit.py
kcossett-amd: Yes, the longest test takes ~50-60 seconds. The default limit of 300 is more than enough.

Edit: Test takes longer but still nowhere near 300.

projects/rocprofiler-systems/source/bin/rocprof-sys-avail/avail.cpp
kcossett-amd: I had left it out as it was a compile-time constant. A user knowing about the name of the var just to be able to do nothing with it when using our tool from TheRock didn't seem helpful. 
But I can see the value of it for people who built our tool... I'll implement it.

projects/rocprofiler-systems/source/bin/rocprof-sys-avail/avail.cpp
kcossett-amd: Ah good catch, when I first started working on this, the code lived in `test_thread_limit.py`, I forgot to update the comment when that changed.

projects/rocprofiler-systems/tests/pytest/test_thread_limit.py
kcossett-amd: I see, use a class to enforce type checking in the matrix. I should enforce that.

I'm going to use this but modify it. I'll drop the load class, that is basically the above_limit but with warning detection.

projects/rocprofiler-systems/tests/pytest/test_thread_limit.py
kcossett-amd: I'll use at_limit.

projects/rocprofiler-systems/tests/pytest/test_thread_limit.py
anujshuk-amd: Did we check with 4196, 8192 threads limit ?

projects/rocprofiler-systems/tests/pytest/conftest.py
```diff
@@ -1599,6 +1617,7 @@ def _subrow(label: str, value) -> str:
         "-" * 70,
         "System Capabilities:",
         _row("Detected num procs:", cap.num_procs),
+        _row("Max threads:", cap.max_threads),
```
anujshuk-amd: All the other fields here (num_procs, cap_sys_admin, perf_event_paranoid, etc.) catch their own errors and fall back to a safe default if something goes wrong. But max_threads does not. When it raises on failure by design. So if rocprof-sys-avail --max-threads ever fails this line will blow up and crash the whole --show-config-only header instead of just showing "Not found" for this one row like everything else does. Do you think we should adress or will it work ?

projects/rocprofiler-systems/tests/pytest/test_thread_limit.py
kcossett-amd: At a `[thread-limit] Threads: 16384`, we note:
```
    Start 520: thread-limit-exhausted-sampling
2/6 Test #520: thread-limit-exhausted-sampling .............   Passed  140.89 sec
    Start 521: thread-limit-exhausted-binary-rewrite
3/6 Test #521: thread-limit-exhausted-binary-rewrite .......   Passed  141.47 sec
    Start 522: thread-limit-exhausted-runtime-instrument
4/6 Test #522: thread-limit-exhausted-runtime-instrument ...   Passed  187.02 sec
    Start 523: thread-limit-exhausted-sys-run
5/6 Test #523: thread-limit-exhausted-sys-run ..............   Passed  140.65 sec
```

In reality, it would by default only reach 4096 as limit is 2048. The x8 test was removed.

projects/rocprofiler-systems/tests/pytest/conftest.py
kcossett-amd: Ah good catch. I'll have it return 0, so header reports "Not found", then I'll have the test skipped, because running the test with unknown thread_limit may result in failure. I'd rather not take that risk.

projects/rocprofiler-systems/tests/pytest/test_thread_limit.py
anujshuk-amd: Could you please share the reason for removing it?

We originally kept this test to generate sufficient load and verify that the system does not hang under stress. It was added after we observed hanging issues when the workload increased, so it serves as an important regression test for that scenario.

dgaliffiAMD (approved):

projects/rocprofiler-systems/tests/pytest/test_thread_limit.py
kcossett-amd: After discussion, I'll increase the `exhausted` thread multiplier from `x2` to `x4`.

dgaliffiAMD:
`rocprofiler-sdk` failures are unrelated to this change - https://github.com/ROCm/rocm-systems/actions/runs/31392612408/job/93529856336?pr=9044#step:11:152

PR #7793, opened 2026-06-25
libhsakmt: AMD Infinity Storage - IO size should be truncated to MAX_…

cfreeamd:
…RW_COUNT

AIS IO signal transfer maximum number limit to 2G -- MAX_RW_COUNT  (0x7ffff000)

Jira ROCM-1356

PR #9045, opened 2026-07-22
[libhsakmt] Remove erroneous reduction in transfer size in hsaKmtAisReadWriteFile.

jordan-turbofish:
hsakmt_fmm_get_handle verifies the size passed into the function fits in the object.  It returns the offset from the beginning of the object of the address passed in.  It doesn't make sense to compare the transfer size to this offset.

This is resulting in IO sizes being improperly being reduced in the ioctl call.

This behaviour was introduced in https://github.com/ROCm/rocm-systems/pull/7793

## Issue Tracking

JIRA ID: ROCM-1356

Should fix hash mismatch error described in:
JIRA ID: ROCM-27889

Also fixes ROCM-28565

## Test Plan

Test hipFile against a local TheRock build.

## Test Result

Local build with this change fixed reduced IO sizes we see monitoring AIS in KFD.

## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/libhsakmt/src/ais.c
+++ b/projects/rocr-runtime/libhsakmt/src/ais.c
@@ -63,8 +63,6 @@ HSAKMT_STATUS HSAKMTAPI hsaKmtAisReadWriteFile(void *MemoryAddress,
 		return HSAKMT_STATUS_INVALID_PARAMETER;
 	}
 
-	transfer_size = MIN(transfer_size, size_offset);
-
 	args.in.handle = handle;
 	args.in.fd = fd;
 	args.in.file_offset = file_offset;


cfreeamd (commented):

Reviewed `libhsakmt/src/ais.c`. The removed line `transfer_size = MIN(transfer_size, size_offset)` looks correct to drop: `hsakmt_fmm_get_handle` uses `size_offset` as in/out — on success it overwrites it with the *offset* of the address within the buffer object (`*size_offset = offset`), not a remaining size. So the old MIN clamped the transfer size to the in-BO offset (0 bytes for an address at a BO's start). The requested size is already validated to fit via the callee's `offset + *size_offset > object->size` check, so `args.in.size = transfer_size` and `args.in.handle_offset = size_offset` are both correct after the change.

(Note: PR is labeled "Not ready to Review", so this is just an early pass.)

gaoikawa:
FYI @cfreeamd @kentrussell @dayatsin-amd 

Hoping we can merge this before the next release deadline since this fixes a regression in unreleased code.  Thanks!

cfreeamd (approved):

PR #9046, opened 2026-07-22
refactor(rocjitsu) Share kernel-descriptor discovery between DBT and DBI

theSK2005:
Follow-up to PR #8888 to route DBI descriptor discovery through DBT's existing walk instead of a second copy.

## Motivation
DBI and DBT have overlap for searching kernel descriptors in ELF code objects. This is used to modify their contents. This PR separates out DBT's kernel descriptor walk from the rest of DBT's kernel descriptor functions/structs so that it can be shared with DBI without additional overhead.

## Test Plan
- Added a shared test to pin contract
- Adjusted existing DBI tests that used now refactored code

## Test Result
All tests pass

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/amdgpu_code_object.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/amdgpu_code_object.cpp
@@ -353,41 +353,4 @@ std::optional<uint32_t> AmdGpuCodeObject::min_kernel_sgpr_count(rj_code_arch_t a
   return min_count;
 }
 
-std::vector<KernelDescriptorInfo> AmdGpuCodeObject::kernel_descriptors() const {
-  std::vector<KernelDescriptorInfo> out;
-  if (text_sections().empty())
-    return out;
-  const Section *text = text_sections().front();
-  const uint64_t text_vaddr = text->vaddr();
-  const uint64_t text_end = text_vaddr + text->size();
-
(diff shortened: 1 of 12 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/kernel_descriptor_scan.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include <cstdint>
+#include <span>
+#include <string>
+#include <vector>
+
+namespace rocjitsu {
+
+/// @brief One AMDHSA kernel descriptor located in an ELF image.
+struct ScannedKernelDescriptor {
```
Groverkss: Why change it's name? I don't think there is a good reason to do this.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/amdgpu_code_object.cpp
```diff
@@ -353,41 +353,4 @@ std::optional<uint32_t> AmdGpuCodeObject::min_kernel_sgpr_count(rj_code_arch_t a
   return min_count;
 }
 
-std::vector<KernelDescriptorInfo> AmdGpuCodeObject::kernel_descriptors() const {
```
Groverkss, line 356: I'm confused why this isn't enough?

theSK2005:
merged changes into parent PR, move refactoring questions there

PR #9053, opened 2026-07-22
fix(rocr): Gate HSA_MODEL_LIB trusted-path restriction to AT_SECURE (ROCM-26861)

dayatsin-amd:
## Summary

Follow-up to the SEC-00591 hardening in #7791. That fix read the model-mode
environment variables via `secure_getenv()` and *always* restricted
`HSA_MODEL_LIB` to `HSAKMT_INSTALL_LIBDIR`. That also disabled the
developer/emulation model mode entirely and blocked arbitrary local paths even
for normal, unprivileged processes where `HSA_MODEL_LIB` grants no additional
privilege (whoever sets it already has the process's full rights).

This change enforces the trusted-path restriction based on the process's
privilege state instead of unconditionally:

- Detect secure-execution mode via `getauxval(AT_SECURE)` instead of
  `secure_getenv()`; fall back to "assume elevated" when the auxv query is
  unavailable so the check fails closed.
- In normal (non-`AT_SECURE`) processes, `HSA_MODEL_LIB` may point to any path
  (developer/emulation use). The value is routed through a single bounded,
  validating resolver (`model_resolve_lib_path`) so the raw `getenv()` string
  never reaches `dlopen()` directly.
- Model mode stays disabled under `AT_SECURE`: both `HSA_MODEL_TOPOLOGY` (a
  directory the process reads/parses) and `HSA_MODEL_LIB` (a library it
  `dlopen()`s) are attacker-controllable across a privilege boundary. The
  `HSAKMT_INSTALL_LIBDIR` trusted-path check is retained as defense-in-depth.
- Build: `check_function_exists(getauxval)` replaces the `secure_getenv` probe
  for both the shared and static-drm targets.

Net effect: the SEC-00591 arbitrary-`dlopen`-under-elevated-privileges vector
stays closed, while unprivileged developer/emulation use of `HSA_MODEL_LIB`
with an arbitrary path is restored.

## JIRA ID

JIRA ID : ROCM-26861

## Test plan

- [x] `libhsakmt` (static) builds cleanly with TheRock `amdclang` under the full strict warning set + ASan; `getauxval` detected (`-DHAVE_GETAUXVAL=1`), `hsakmtmodel.c` compiles with no new warnings
- [x] Verified the compiled object references `getauxval` (live `AT_SECURE` path, not the fallback)
- [x] Model-mode smoke test: unprivileged process (`getauxval(AT_SECURE)=0`) with `HSA_MODEL_LIB` pointing to an arbitrary path (outside `/opt/rocm/lib`) loads the FFM `.so` — the stub's `get_hsakmt_model_functions()`/`create_memfd()` fire and `hsakmt_use_model=1`
- [x] Confirm model mode is disabled (no `dlopen`) when launched under an `AT_SECURE` wrapper: a setgid binary (`egid != rgid` ⇒ `getauxval(AT_SECURE)=1`) with a valid `/opt/rocm/lib` `HSA_MODEL_LIB` still results in `hsakmt_use_model=0` and no `dlopen`

Runtime test notes: the two runtime cases were executed on a Linux test host
(kernel 6.8, glibc `getauxval`) against the exact shipped bytes of
`model_init_env_vars()` and its helpers, driven by a minimal fake-FFM `.so`
implementing the v1.1 `hsakmt_model_functions` interface. This validated the
`HSA_MODEL_LIB`/`AT_SECURE` gating logic directly rather than through a full
`libhsa-runtime64` load path.


projects/rocr-runtime/libhsakmt/CMakeLists.txt
```diff
@@ -293,10 +293,10 @@ else()
       target_compile_definitions(${HSAKMT_TARGET} PRIVATE -DHAVE_MEMFD_CREATE=1)
     endif()
 
-    set(CMAKE_REQUIRED_INCLUDES stdlib.h)
-    check_function_exists(secure_getenv HAVE_SECURE_GETENV)
-    if(HAVE_SECURE_GETENV)
-      target_compile_definitions(${HSAKMT_TARGET} PRIVATE -DHAVE_SECURE_GETENV=1)
+    set(CMAKE_REQUIRED_INCLUDES sys/auxv.h)
+    check_function_exists(getauxval HAVE_GETAUXVAL)
```
cfreeamd: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `check_function_exists()` performs a link-only test and does not consult `CMAKE_REQUIRED_INCLUDES` (that variable is honored by `check_symbol_exists`, not `check_function_exists`), so setting it to `sys/auxv.h` here is inert — same as the old `stdlib.h`. Detection still works because `getauxval` is a libc symbol, so this is harmless. If the intent is to verify the declaration is visible, `check_symbol_exists(getauxval "sys/auxv.h" HAVE_GETAUXVAL)` would be the accurate form. Non-blocking.

projects/rocr-runtime/libhsakmt/src/hsakmtmodel.c
```diff
@@ -50,12 +53,19 @@
...
-#ifdef HAVE_SECURE_GETENV
-	return secure_getenv(name);
+#if defined(HAVE_GETAUXVAL) && defined(AT_SECURE)
+	return getauxval(AT_SECURE) != 0;
 #else
-	return getenv(name);
+	/* If we cannot query the secure-execution state, err on the side of
+	 * caution and behave as if elevated so the trusted-path restriction
+	 * on HSA_MODEL_LIB is always enforced. */
+	return true;
```
cfreeamd, line 71: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** The fail-safe default is the right call, but its effect is broader than the comment states. When `HAVE_GETAUXVAL`/`AT_SECURE` are unavailable this returns `true`, which in `model_init_env_vars()` makes `hsakmt_model_topology && !at_secure` false — so model mode is disabled *entirely*, not merely restricted to the trusted lib path. In practice `getauxval` is universal on modern glibc/musl so this is largely theoretical, but a developer wondering why model mode silently won't activate on an exotic target would be helped if the comment noted that model mode is fully gated off in this fallback (not just the HSA_MODEL_LIB path restriction).

projects/rocr-runtime/libhsakmt/src/hsakmtmodel.c
```diff
@@ -97,6 +107,38 @@ static bool model_lib_path_allowed(const char *libname, char *resolved,
...
+ * Returns true when `resolved` holds a usable path, false when the value must
+ * be rejected.
+ */
+static bool model_resolve_lib_path(const char *libname, bool at_secure,
+				   char *resolved, size_t resolved_size)
+{
+	if (!libname || !*libname)
+		return false;
+
+	if (model_lib_path_allowed(libname, resolved, resolved_size))
```
cfreeamd, line 135: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** The header now documents that in unprivileged processes `HSA_MODEL_LIB` "may be any path", but the non-secure relaxation only takes effect for values that `model_lib_path_allowed()` rejects — i.e. absolute paths outside `HSAKMT_INSTALL_LIBDIR`, or paths containing "..". A *bare relative* name (e.g. `libffm.so`) still passes `model_lib_path_allowed()` and is silently rewritten to `HSAKMT_INSTALL_LIBDIR/libffm.so`, so it never reaches the pass-through branch and never gets normal dlopen search-path semantics. Is that intended? If so, a word in the doc would help; if a developer is expected to point at a bare soname resolved via `LD_LIBRARY_PATH`, the relative case would need handling too. Non-blocking.

cfreeamd (commented):

Reviewed the AT_SECURE gating change across the three files. Traced the security-relevant paths in `hsakmtmodel.c`: model mode is gated off under AT_SECURE, `HSA_MODEL_LIB` is routed through `model_resolve_lib_path()` before every `dlopen()`, and the getauxval-unavailable case fails safe to "elevated". Confirmed `hsakmt_model_topology` is only consumed via `get_topology_dir()` guarded by `hsakmt_use_model`, so the raw env value isn't reachable under AT_SECURE. Inline notes are all non-blocking (CMake detection form, scope of the fallback comment, and relative-path handling vs. the "any path" doc).

projects/rocr-runtime/libhsakmt/CMakeLists.txt
dayatsin-amd: Good catch — fixed in cf58c073ec. Switched to `check_symbol_exists(getauxval "sys/auxv.h" HAVE_GETAUXVAL)` (and added `include(CheckSymbolExists)`), which honors `CMAKE_REQUIRED_INCLUDES` and additionally verifies the declaration / `AT_SECURE` are visible via the header, rather than the link-only `check_function_exists`.

projects/rocr-runtime/libhsakmt/src/hsakmtmodel.c
dayatsin-amd: Agreed — clarified in cf58c073ec. The fallback comment now notes that returning `true` here disables model mode *entirely* (since `model_init_env_vars()` gates on `!at_secure`), not merely the `HSA_MODEL_LIB` path restriction, and that `getauxval` is universal on modern glibc/musl so this path is largely theoretical.

projects/rocr-runtime/libhsakmt/src/hsakmtmodel.c
dayatsin-amd: Intended, and now documented in cf58c073ec. A relative name/path is always anchored under `HSAKMT_INSTALL_LIBDIR` (e.g. `libfoo.so` -> `<HSAKMT_INSTALL_LIBDIR>/libfoo.so`) and is deliberately *not* resolved via `LD_LIBRARY_PATH` — this preserves the original #7791 trusted-relative behavior. The non-secure relaxation only applies to *absolute* paths outside the trusted dir (the developer/emulation case). The header doc now spells out all three cases (relative-anchored, absolute-in-trusted-dir, other-absolute-only-when-unprivileged) plus the `..` rejection, so "any path" is no longer ambiguous.

dayatsin-amd:
Thanks for the review @cfreeamd. Addressed all three notes in cf58c073ec (doc/comment + CMake detection only — no behavior change):

- **[Nit] CMake detection:** switched to `check_symbol_exists(getauxval "sys/auxv.h" HAVE_GETAUXVAL)` so `CMAKE_REQUIRED_INCLUDES` is honored and the declaration/`AT_SECURE` visibility is verified.
- **[Optional] Fallback comment:** expanded to note that the `getauxval`-unavailable case disables model mode entirely (via the `!at_secure` gate), not just the path restriction.
- **[Question] Relative paths:** confirmed intended and documented — relative names anchor under `HSAKMT_INSTALL_LIBDIR` (no `LD_LIBRARY_PATH` search); the "any path" relaxation applies to absolute paths outside the trusted dir in non-`AT_SECURE` processes only.

Re-verified with a local `amdclang -fsyntax-only`. Note the therock-pr-bot Unit Test policy failure is expected: this is a security/dev-mode env-var change with no separate unit-test target, and I validated it via a runtime harness on a test host (see the test plan) rather than adding a `test_*` file.

timhuang-amd:
Hi @dayatsin-amd  — we’re hitting an FFM failure when running ffm model,
model: HSA_MODEL_LIB must be under /__w/rockrel/rockrel/build/core/ROCR-Runtime/stage/lib (got …/....so), 

This PR looks like the fix for it. If you can merge this when convenient, it would help us a lot—we have to pick up a newer commit that still includes the failing behavior until this lands. Thanks.

dayatsin-amd:
> Hi @dayatsin-amd — we’re hitting an FFM failure when running ffm model, model: HSA_MODEL_LIB must be under /__w/rockrel/rockrel/build/core/ROCR-Runtime/stage/lib (got …/....so),
> 
> This PR looks like the fix for it. If you can merge this when convenient, it would help us a lot—we have to pick up a newer commit that still includes the failing behavior until this lands. Thanks.

Yes, this PR is meant to fix that issue.

cfreeamd:
🤖 AI-assisted pre-review (for @cfreeamd) — follow-up after author update (cf58c073)

All three prior notes are resolved; the delta is docs/comment + CMake detection only, with no behavior change. Verified against the diff:

- **[Nit] CMake probe** — now `check_symbol_exists(getauxval "sys/auxv.h" HAVE_GETAUXVAL)` with `include(CheckSymbolExists)`; both the shared and static-DRM targets gate on `HAVE_GETAUXVAL`. This honors `CMAKE_REQUIRED_INCLUDES` and verifies the declaration / `AT_SECURE` visibility, as suggested.
- **[Optional] fallback comment** — now notes that the getauxval-unavailable path disables model mode entirely (via the `!at_secure` gate), not just the lib-path restriction.
- **[Question] relative HSA_MODEL_LIB** — header now documents the full resolution policy (relative anchored under `HSAKMT_INSTALL_LIBDIR`, no `LD_LIBRARY_PATH` search; absolute-under-trusted as-is; other absolute only when not AT_SECURE; `..` always rejected). Confirmed intended, preserving #7791 behavior.

No new concerns from the update.

cfreeamd (approved):

PR #194, opened 2025-08-08
pc sampling unit tests

ywang103-amd:

PR #9055, opened 2026-07-22
feat(rccl): integrate MADEngine workloads into CI [AICOMRCCL-1332]

prasanna-amd:
## Summary

Add Llama-3.1-70B training workload (via MADEngine) to the RCCL CI pipeline on the Ruby cluster. A Docker overlay image swaps the CI-built `librccl.so` into a `rocm/primus` base container, enabling end-to-end validation of RCCL builds against real training workloads.

JIRA ID: AICOMRCCL-1332

### Changes

- **`test_madengine.py`** — orchestration: overlay build, manifest generation, SLURM dispatch, live log metric extraction, JSONL perf datastore with per-precision rolling regression detection
- **`therock-rccl-test-madengine.yml`** — new workflow for MADEngine workloads
- **`therock-rccl-ci.yml`** — schedule weekly → nightly, `madengine_nodes` input
- **`therock-rccl-ci-linux.yml`** — wire in MADEngine job for gfx950
- **`rccl_ci_utils.py`**, **`test_pytorch_c10d.py`**, **`test_jax_collective.py`** — CI scripts moved to `projects/rccl/ci/scripts/`

### Review feedback addressed

- Clones pinned: madengine at `ec4de0b58c49`, MAD at `688828bd9d4a` on `mad-rccl` branch
- HF token removed from manifest JSON (passed via `MAD_SECRETS_HFTOKEN` env only)
- `check_regression()` now filters by precision (BF16/FP8 no longer share a rolling mean)
- `notify_webhook` → `teams_webhook` naming mismatch fixed in caller
- PyTorch/JAX CI script fixes extracted to PR #9449

### Validated results (single-node, 8x MI325X)

| Precision | TFLOP/s/GPU | Tokens/s/GPU | Iterations |
|-----------|-------------|--------------|------------|
| BF16 | 773.9 | 1722.1 | 50/50 |
| FP8 | 1249.4 | 2780.4 | 50/50 |

## Test plan

- [ ] Trigger `workflow_dispatch` to validate MADEngine workflow on Ruby
- [ ] Validate single-node BF16/FP8 training completes 50/50 iterations
- [ ] Confirm JSONL perf results written to `/apps/rccl-ci/perf/madengine_results.jsonl`
- [ ] Multi-node (2N) validation (follow-up)

### DIFF
--- a/.github/workflows/therock-rccl-ci-linux.yml
+++ b/.github/workflows/therock-rccl-ci-linux.yml
@@ -28,6 +28,10 @@ on:
       notify_webhook:
         type: string
         default: ''
+      madengine_nodes:
+        description: 'Number of SLURM nodes for MADEngine workloads'
+        type: string
+        default: '2'
 
 permissions:
   contents: read
@@ -208,3 +212,18 @@ jobs:
       artifact_run_id: ${{ inputs.artifact_run_id || github.run_id }}
       notify_email: ${{ inputs.notify_email }}
       notify_webhook: ${{ inputs.notify_webhook }}
+
+  therock-test-madengine:
+    name: "Test MADEngine workloads (scheduled)"
+    if: ${{ !cancelled() && inputs.amdgpu_families == 'gfx950-dcgpu' && (inputs.event_name == 'schedule' || inputs.event_name == 'workflow_dispatch') }}
+    needs: [therock-build-linux]
+    uses: ./.github/workflows/therock-rccl-test-madengine.yml
+    with:
+      amdgpu_families: ${{ inputs.amdgpu_families }}
+      artifact_group: ${{ inputs.artifact_group }}
+      test_runs_on: ruby-linux-slurm-scale-runner
+      artifact_run_id: ${{ inputs.artifact_run_id || github.run_id }}
+      workload: llama-3.1-70b-training
+      nodes: ${{ inputs.madengine_nodes || '2' }}
+      notify_email: ${{ inputs.notify_email }}
+      teams_webhook: ${{ inputs.notify_webhook }}
--- a/.github/workflows/therock-rccl-ci-linux.yml
+++ b/.github/workflows/therock-rccl-ci-linux.yml
@@ -217,7 +217,12 @@ jobs:
     name: "Test MADEngine workloads (scheduled)"
     if: ${{ !cancelled() && inputs.amdgpu_families == 'gfx950-dcgpu' && (inputs.event_name == 'schedule' || inputs.event_name == 'workflow_dispatch') }}
     needs: [therock-build-linux]
+    permissions:
+      contents: read
+      id-token: write
+      packages: write
     uses: ./.github/workflows/therock-rccl-test-madengine.yml
+    secrets: inherit
     with:
       amdgpu_families: ${{ inputs.amdgpu_families }}
       artifact_group: ${{ inputs.artifact_group }}
--- a/projects/rccl/ci/scripts/test_madengine.py
+++ b/projects/rccl/ci/scripts/test_madengine.py
@@ -21,10 +21,12 @@ from __future__ import annotations
 
 import argparse
 import csv
+import hashlib
 import json
 import logging
 import os
 import re
+import shutil
 import subprocess
 import sys
 import tempfile
@@ -63,8 +65,8 @@ WORKLOAD_CONFIGS = {
         ],
         "base_image": "rocm/primus:v26.4",
         "gpu_target": "gfx950",
-        "precision": "bf16",
         "metric_key": "tokens_per_second_per_gpu",
+        "multiple_results": "perf_primus-megatron-Llama-3.1-70B.csv",
         "reference_values": {
             "2N": 1685,
             "4N": 1600,
@@ -75,7 +77,10 @@ WORKLOAD_CONFIGS = {
         "slurm_partition": "meta64",
         "gpus_per_node": 8,
         "time_limit": "03:00:00",
-        "container_mounts": ["/shared:/shared"],
+        "docker_mounts": {"/dev/infiniband": "/dev/infiniband"},
+        "docker_run_options": "--privileged --group-add render --shm-size 64G "
+            "--device=/dev/infiniband --cap-add IPC_LOCK "
+            "--ulimit memlock=-1 -v /sys:/sys:ro -v /run/udev:/run/udev:ro",
     },
     "llama-4-scout-training": {
         "type": "training",
@@ -86,8 +91,8 @@ WORKLOAD_CONFIGS = {
         ],
         "base_image": "rocm/primus:v26.4",
         "gpu_target": "gfx950",
-        "precision": "bf16",
         "metric_key": "tokens_per_second_per_gpu",
+        "multiple_results": "perf_primus-megatron-Llama-3.1-70B.csv",
         "reference_values": {
             "2N": 2734,
             "4N": 2337,
@@ -95,7 +100,10 @@ WORKLOAD_CONFIGS = {
         "slurm_partition": "meta64",
         "gpus_per_node": 8,
         "time_limit": "02:00:00",
-        "container_mounts": ["/shared:/shared"],
+        "docker_mounts": {"/dev/infiniband": "/dev/infiniband"},
+        "docker_run_options": "--privileged --group-add render --shm-size 64G "
+            "--device=/dev/infiniband --cap-add IPC_LOCK "
+            "--ulimit memlock=-1 -v /sys:/sys:ro -v /run/udev:/run/udev:ro",
     },
 }
 
@@ -349,20 +357,27 @@ def patch_madengine_for_cluster(
             log.info("run_orchestrator.py already patched or yum string not found")
 
 
-def get_rccl_commit() -> str:
-    """Get the current RCCL commit hash from git or environment."""
+def get_rccl_commit(rccl_lib: Path | None = None) -> str:
+    """Derive a unique identifier for the RCCL build.
+
+    Checks, in order: RCCL_COMMIT_HASH env, GITHUB_RUN_ID env, sha256 of
+    the librccl.so binary.  Does NOT fall back to ``git rev-parse HEAD``
+    because in CI the checkout is TheRock (not RCCL), which would produce
+    a constant tag and cause stale cache hits on persistent runners.
+    """
     commit = os.environ.get("RCCL_COMMIT_HASH", "")
     if commit:
-        return commit
+        return commit[:12]
 
-    try:
-        result = subprocess.run(
-            ["git", "rev-parse", "HEAD"],
-            capture_output=True, text=True, check=True,
-        )
-        return result.stdout.strip()[:12]
-    except (subprocess.CalledProcessError, FileNotFoundError):
-        return "unknown"
+    run_id = os.environ.get("GITHUB_RUN_ID", "")
+    if run_id:
+        return f"run{run_id}"
+
+    if rccl_lib and rccl_lib.exists():
+        h = hashlib.sha256(rccl_lib.read_bytes()).hexdigest()
+        return h[:12]
+
+    return "unknown"
 
 
 def _rccl_uses_kpack(rccl_lib: Path) -> bool:
@@ -396,7 +411,7 @@ def build_rccl_overlay_image(
     SLURM compute nodes can pull it automatically.  Returns the final
     image tag (registry-qualified if pushed).
     """
-    rccl_commit = get_rccl_commit()
+    rccl_commit = get_rccl_commit(rccl_lib)
     tag = f"{base_image}-rccl-{gpu_target}-{rccl_commit}"
 
     result = subprocess.run(
@@ -482,7 +497,8 @@ ENV NCCL_DEBUG=WARN
         log.info("Overlay image built: %s", tag)
 
     if registry:
-        push_tag = f"{registry}/rccl-ci:{base_image.replace('/', '-')}-{rccl_commit}"
+        safe_base = base_image.replace("/", "-").replace(":", "-")
+        push_tag = f"{registry}/rccl-ci:{safe_base}-{rccl_commit}"
         log.info("Tagging overlay for registry: %s -> %s", tag, push_tag)
         subprocess.run(["docker", "tag", tag, push_tag], check=True)
         log.info("Pushing overlay image to registry: %s", push_tag)
@@ -501,27 +517,25 @@ def generate_manifest(
     nodes: int,
     work_dir: Path,
     nodelist: str = "",
+    registry: str = "",
 ) -> Path:
-    """Generate a madengine manifest.json for the workload."""
+    """Generate a madengine manifest.json for the workload.
+
+    Structure follows the reference template from the mad-rccl branch:
+    deployment config under ``deployment_config``, env vars inside both
+    ``context.docker_env_vars`` and ``deployment_config.env_vars``, mounts
+    in ``context.docker_mounts``.
+    """
     gpus_per_node = workload_config["gpus_per_node"]
-    total_gpus = nodes * gpus_per_node
 
     nccl_env = dict(cluster_config.get("nccl_env", {}))
     if nodes == 1:
         nccl_env.pop("NCCL_NET", None)
-        # Single-node doesn't need multi-NIC; keep only the first interface
-        # to avoid Gloo failures on nodes that lack secondary NICs.
         ifname = nccl_env.get("NCCL_SOCKET_IFNAME", "")
         if "," in ifname:
             nccl_env["NCCL_SOCKET_IFNAME"] = ifname.split(",")[0]
-    env_vars = {
-        **nccl_env,
-        "NCCL_DEBUG": "WARN",
-    }
 
-    if cluster_config.get("mount_host_ib_libs") and nodes > 1:
-        if "NCCL_SOCKET_IFNAME" in nccl_env:
-            env_vars["GLOO_SOCKET_IFNAME"] = nccl_env["NCCL_SOCKET_IFNAME"]
+    socket_ifname = nccl_env.get("NCCL_SOCKET_IFNAME", "")
 
     # HF_TOKEN is passed via MAD_SECRETS_HFTOKEN in the process environment
     # (set in the workflow). Do NOT write it into the manifest — the manifest
@@ -531,44 +545,62 @@ def generate_manifest(
     scripts_dir = work_dir / "scripts" / "primus_megatron-lm"
     if not scripts_dir.is_dir():
         scripts_dir = work_dir / "scripts" / "primus" / "megatron-lm"
-    run_sh = scripts_dir / "run.sh"
-    if run_sh.exists():
-        run_sh_text = run_sh.read_text()
-        if model_repo not in run_sh_text:
-            for alias in workload_config.get("model_repo_aliases", []):
-                if alias in run_sh_text:
-                    log.info("Model repo %s not in run.sh, using alias %s",
-                             model_repo, alias)
-                    model_repo = alias
-                    break
+
     image_key = "overlay"
+    gpu_indices = ",".join(str(i) for i in range(gpus_per_node))
+    render_ds = [128 + i for i in range(gpus_per_node)]
+
+    docker_env_vars = {
+        **nccl_env,
+        "NCCL_DEBUG": "WARN",
+        "NCCL_IB_DISABLE": "0",
+        "NCCL_TIMEOUT": "900",
+        "IBV_SHOW_WARNINGS": "1",
+    }
+    if socket_ifname:
+        docker_env_vars["GLOO_SOCKET_IFNAME"] = socket_ifname
+
+    docker_mounts = dict(workload_config.get("docker_mounts", {}))
+    docker_run_opts = workload_config.get("docker_run_options", "")
+
+    slurm_config = {
+        "partition": cluster_config.get("slurm_partition", workload_config["slurm_partition"]),
+        "qos": cluster_config.get("slurm_qos", ""),
+        "nodes": nodes,
+        "gpus_per_node": gpus_per_node,
+        "time": workload_config["time_limit"],
+        "output_dir": "./slurm_output",
+        "exclusive": True,
+        "enable_node_check": False,
+        "network_interface": socket_ifname,
+        **({"nodelist": nodelist} if nodelist else {}),
+    }
 
     manifest = {
-        "model_repo": model_repo,
         "built_images": {
             image_key: {
                 "docker_image": overlay_image,
-                "local_image": True,
-                "base_docker": overlay_image,
+                "local_image": not bool(registry),
+                "registry_image": overlay_image if registry else None,
+                "registry": registry or None,
+                "base_docker": workload_config["base_image"],
                 "build_status": "SKIPPED",
-                "build_time": 0,
+                "build_duration": 0,
+                "gpu_vendor": "AMD",
             },
         },
         "built_models": {
             image_key: {
                 "name": model_repo,
-                "tags": ["pyt", "pretrain", "training"],
+                "tags": workload_config.get("tags", ["pyt", "pretrain", "training"]),
                 "dockerfile": "N/A (overlay image)",
                 "scripts": f"scripts/{scripts_dir.name}/run.sh",
-                "n_gpus": str(total_gpus),
+                "n_gpus": "-1",
                 "owner": "",
-                "training_precision": workload_config.get("precision", ""),
+                "training_precision": "",
+                "multiple_results": workload_config.get("multiple_results", ""),
                 "args": f"--model_repo {model_repo}",
-                "additional_docker_run_options": (
-                    "--device=/dev/infiniband"
-                    if cluster_config.get("mount_host_ib_libs") and nodes > 1
-                    else ""
-                ),
+                "additional_docker_run_options": docker_run_opts,
                 "data": "",
                 "cred": "",
                 "timeout": None,
@@ -577,40 +609,33 @@ def generate_manifest(
         "context": {
             "gpu_vendor": "AMD",
             "guest_os": "UBUNTU",
-            "docker_gpus": ",".join(str(i) for i in range(gpus_per_node)),
-            "gpu_renderDs": [128 + i for i in range(gpus_per_node)],
-            "docker_env_vars": {
-                "MAD_SYSTEM_GPU_ARCHITECTURE": cluster_config["gpu_target"],
-                "MAD_SYSTEM_NGPUS": str(gpus_per_node),
-                "MAD_SYSTEM_HIP_VERSION": "6.4",
-                "MAD_SYSTEM_GPU_PRODUCT_NAME": cluster_config["gpu_target"],
-                **nccl_env,
-                "NCCL_IB_DISABLE": "0",
-                "NCCL_SOCKET_IFNAME": nccl_env.get("NCCL_SOCKET_IFNAME", "eth0"),
-                "MAD_MULTI_NODE_RUNNER": "",
-            },
-            "docker_mounts": {},
-            "docker_build_arg": {
-                "MAD_SYSTEM_GPU_ARCHITECTURE": cluster_config["gpu_target"],
-            },
+            "docker_gpus": gpu_indices,
+            "gpu_renderDs": render_ds,
+            "docker_env_vars": docker_env_vars,
+            "docker_mounts": docker_mounts,
+            "docker_build_arg": {},
         },
-        "slurm": {
-            "partition": cluster_config.get("slurm_partition", workload_config["slurm_partition"]),
-            "qos": cluster_config.get("slurm_qos", ""),
-            "network_interface": nccl_env.get("NCCL_SOCKET_IFNAME", ""),
-            "nodes": nodes,
-            "ntasks_per_node": gpus_per_node,
-            "gpus_per_node": gpus_per_node,
-            "time": workload_config["time_limit"],
-            "exclusive": True,
-            "enable_node_check": False,
-            **({"nodelist": nodelist} if nodelist else {}),
+        "deployment_config": {
+            "target": "slurm",
+            "slurm": slurm_config,
+            "distributed": {
+                "launcher": "primus",
+                "backend": "nccl",
+                "port": 29500,
+                "nnodes": nodes,
+                "nproc_per_node": gpus_per_node,
+            },
+            "env_vars": {
+                **docker_env_vars,
+                "TORCH_NCCL_ASYNC_ERROR_HANDLING": "1",
+                "TORCH_NCCL_HIGH_PRIORITY": "1",
+                "OMP_NUM_THREADS": "8",
+                "MIOPEN_FIND_MODE": "1",
+            },
+            "debug": False,
+            "docker_gpus": gpu_indices,
         },
-        "env": env_vars,
     }
-
-    if workload_config.get("container_mounts"):
-        manifest["container_mounts"] = workload_config["container_mounts"]
 
     manifest_path = work_dir / "manifest.json"
     manifest_path.write_text(json.dumps(manifest, indent=2))
@@ -623,57 +648,20 @@ def run_madengine(
     output_csv: Path,
     work_dir: Path,
     timeout_minutes: int = 120,
-    gpu_target: str = "gfx950",
-    gpus_per_node: int = 8,
-    nodes: int = 2,
-    slurm_partition: str = "meta64",
-    slurm_time: str = "02:00:00",
-    nccl_env: dict | None = None,
-    network_interface: str = "",
-    nodelist: str = "",
 ) -> int:
-    """Run madengine with the given manifest and return the exit code."""
-    render_ds = [128 + i for i in range(gpus_per_node)]
-    gpu_indices = ",".join(str(i) for i in range(gpus_per_node))
-    env_vars = {
-        "MAD_SYSTEM_GPU_ARCHITECTURE": gpu_target,
-        "MAD_SYSTEM_NGPUS": str(gpus_per_node),
-        "MAD_SYSTEM_HIP_VERSION": "6.4",
-        "MAD_SYSTEM_GPU_PRODUCT_NAME": gpu_target,
-    }
-    if nccl_env:
-        env_vars.update(nccl_env)
-    env_vars["NCCL_IB_DISABLE"] = "0"
-    env_vars["MAD_MULTI_NODE_RUNNER"] = ""
-    additional_context = json.dumps({
-        "gpu_vendor": "AMD",
-        "docker_gpus": gpu_indices,
-        "gpu_renderDs": render_ds,
-        "docker_env_vars": env_vars,
-        "slurm": {
-            "partition": slurm_partition,
-            "network_interface": network_interface,
-            "nodes": nodes,
-            "gpus_per_node": gpus_per_node,
-            "time": slurm_time,
-            "exclusive": True,
-            "enable_node_check": False,
-            **({"nodelist": nodelist} if nodelist else {}),
-        },
-        "distributed": {
-            "launcher": "primus",
-            "nnodes": nodes,
-            "nproc_per_node": gpus_per_node,
-            "backend": "nccl",
-        },
-    })
+    """Run madengine with the given manifest and return the exit code.
+
+    The manifest already contains deployment_config with slurm, distributed,
+    and env_vars sections.  madengine merges deployment_config into
+    additional_context automatically (run_orchestrator.py:225-234), so we
+    do not need to duplicate those here.
+    """
     cmd = [
         "madengine", "run",
         "-m", str(manifest_path),
         "-o", str(output_csv),
         "--live-output",
         "--verbose",
-        "--additional-context", additional_context,
     ]
 
     log.info("Running: %s", " ".join(cmd))
@@ -709,22 +697,41 @@ def run_madengine(
         return 124
 
 
-def parse_perf_csv(csv_path: Path, metric_key: str) -> dict:
-    """Parse madengine perf.csv and extract the key metric."""
-    if not csv_path.exists():
-        log.error("perf.csv not found at %s", csv_path)
-        return {}
+def parse_perf_results(work_dir: Path) -> list[dict]:
+    """Parse madengine performance results.
 
-    log.info("Parsing perf.csv: %s", csv_path)
-    results = {}
-    with open(csv_path) as f:
-        reader = csv.DictReader(f)
-        for row in reader:
-            log.info("perf.csv row: %s", dict(row))
-            for key, value in row.items():
-                results[key] = value
+    Prefers ``perf_entry_super.json`` (31 fixed columns, per-precision rows
+    with ``multi_results``).  Falls back to ``perf.csv`` (variable-width,
+    long-format).  Returns a list of result dicts — one per row.
+    """
+    super_json = work_dir / "perf_entry_super.json"
+    if super_json.exists():
+        try:
+            entries = json.loads(super_json.read_text())
+            if entries:
+                log.info("Parsed %d result(s) from perf_entry_super.json", len(entries))
+                for e in entries:
+                    log.info("  model=%s perf=%s metric=%s status=%s precision=%s",
+                             e.get("model"), e.get("performance"),
+                             e.get("metric"), e.get("status"),
+                             e.get("training_precision"))
+                return entries
+        except (json.JSONDecodeError, TypeError) as exc:
+            log.warning("Could not parse %s: %s", super_json, exc)
 
-    return results
+    csv_path = work_dir / "perf.csv"
+    if csv_path.exists():
+        rows = []
+        with open(csv_path) as f:
+            for row in csv.DictReader(f):
+                log.info("perf.csv row: %s", dict(row))
+                rows.append(dict(row))
+        if rows:
+            log.info("Parsed %d row(s) from perf.csv", len(rows))
+            return rows
+
+    log.warning("No perf results found in %s", work_dir)
+    return []
 
 
 _ANSI_RE = re.compile(r"\x1b\[[0-9;]*m")
@@ -1023,7 +1030,6 @@ def main() -> None:
 
     workload_config = WORKLOAD_CONFIGS[args.workload]
     cluster_config = CLUSTER_CONFIGS[args.cluster]
-    rccl_commit = get_rccl_commit()
     scale = f"{args.nodes}N/{args.nodes * workload_config['gpus_per_node']}GPU"
 
     results_dir = args.results_dir or Path(cluster_config["results_base"])
@@ -1034,6 +1040,8 @@ def main() -> None:
     # Step 1: Find RCCL library
     rccl_lib = find_rccl_library(args.artifact_dir)
     log.info("RCCL library: %s", rccl_lib)
+    rccl_commit = get_rccl_commit(rccl_lib)
+    log.info("RCCL commit/tag: %s", rccl_commit)
 
     # Step 2: Install madengine
     madengine_dir = install_madengine(work_dir)
@@ -1082,79 +1090,69 @@ def main() -> None:
         args.nodes,
         work_dir,
         nodelist=nodelist,
+        registry=args.registry,
     )
 
     # Step 5: Run the workload
     output_csv = work_dir / "perf.csv"
-    nccl_env = dict(cluster_config.get("nccl_env", {}))
-    if args.nodes == 1:
-        nccl_env.pop("NCCL_NET", None)
-        ifname = nccl_env.get("NCCL_SOCKET_IFNAME", "")
-        if "," in ifname:
-            nccl_env["NCCL_SOCKET_IFNAME"] = ifname.split(",")[0]
     exit_code = run_madengine(
         manifest_path, output_csv, work_dir, args.timeout_minutes,
-        gpu_target=cluster_config["gpu_target"],
-        gpus_per_node=workload_config["gpus_per_node"],
-        nodes=args.nodes,
-        slurm_partition=cluster_config["slurm_partition"],
-        slurm_time=workload_config["time_limit"],
-        nccl_env=nccl_env,
-        network_interface=nccl_env.get("NCCL_SOCKET_IFNAME", ""),
-        nodelist=nodelist,
     )
 
-    # Step 6: Parse results — try perf.csv first, fall back to live log
-    metric_value = None
-    metric_key = workload_config["metric_key"]
-    if output_csv.exists():
-        perf_data = parse_perf_csv(output_csv, metric_key)
-        if metric_key in perf_data and perf_data[metric_key]:
-            try:
-                metric_value = float(perf_data[metric_key])
-                log.info("Metric %s = %.1f (from perf.csv)", metric_key, metric_value)
-            except (ValueError, TypeError):
-                log.warning("Could not parse metric %s: %s", metric_key, perf_data.get(metric_key))
-
-        run_id = os.environ.get("GITHUB_RUN_ID", "local")
-        run_artifacts = results_dir / "runs" / run_id
-        try:
-            run_artifacts.mkdir(parents=True, exist_ok=True)
-            subprocess.run(["cp", str(output_csv), str(run_artifacts / "perf.csv")], check=True)
-        except OSError as exc:
-            log.warning("Could not save run artifacts to %s: %s", run_artifacts, exc)
-    else:
-        log.warning("No perf.csv produced — checking live logs for metrics")
-
-    # Fall back to live log parsing when perf.csv has no metrics.
-    # madengine's perf collector can miss metrics when primus defers
-    # collection to the login node, but the live log always has them.
+    # Step 6: Parse results — structured output first, live log fallback
+    perf_results = parse_perf_results(work_dir)
     live_log_runs = parse_live_log_metrics(work_dir)
+
+    # Save run artifacts
+    run_id = os.environ.get("GITHUB_RUN_ID", "local")
+    run_artifacts = results_dir / "runs" / run_id
+    try:
+        run_artifacts.mkdir(parents=True, exist_ok=True)
+        for f in ["perf.csv", "perf_entry_super.csv", "perf_entry_super.json"]:
+            src = work_dir / f
+            if src.exists():
+                shutil.copy2(str(src), str(run_artifacts / f))
+    except OSError as exc:
+        log.warning("Could not save run artifacts to %s: %s", run_artifacts, exc)
+
+    # Extract metric from structured results (perf_entry_super has
+    # performance/metric in long-format: metric name is a value in the
+    # ``metric`` column, not a column name).
+    metric_value = None
+    if perf_results:
+        for row in perf_results:
+            perf_val = row.get("performance", "")
+            if perf_val:
+                try:
+                    metric_value = float(perf_val)
+                    log.info("Metric %s = %.1f (from structured results)",
+                             row.get("metric", "?"), metric_value)
+                except (ValueError, TypeError):
+                    pass
+
+    # Fall back to live log parsing when structured results have no metrics.
     if metric_value is None and live_log_runs:
         best_run = live_log_runs[-1]
-        if metric_key == "tokens_per_second_per_gpu" and "tokens_per_second_per_gpu" in best_run:
-            metric_value = best_run["tokens_per_second_per_gpu"]
-            log.info("Metric %s = %.1f (from live log)", metric_key, metric_value)
-        elif "tflops_avg" in best_run:
-            metric_value = best_run["tflops_avg"]
-            log.info("Metric tflops_avg = %.1f (from live log, key %s not found)", metric_value, metric_key)
+        metric_value = best_run.get("tokens_per_second_per_gpu") or best_run.get("tflops_avg")
+        if metric_value is not None:
+            log.info("Metric = %.1f (from live log fallback)", metric_value)
 
     # Override exit_code if training actually completed successfully.
     # madengine can report failure (exit code 3) when its perf collector
     # can't parse the output format, even though training ran to completion.
     if exit_code != 0 and live_log_runs:
-        all_completed = all(r["completed"] for r in live_log_runs)
-        if all_completed:
+        all_completed = all(r.get("completed", False) for r in live_log_runs)
+        if all_completed and metric_value is not None:
             log.info(
                 "Overriding madengine exit code %d → 0: all %d training run(s) "
-                "completed successfully (live log verification)",
+                "completed with metrics (live log verification)",
                 exit_code, len(live_log_runs),
             )
             for r in live_log_runs:
                 log.info(
                     "  %s: %d/%d iters, %.1f TFLOP/s, %.1f tokens/s/GPU",
-                    r["log_file"], r["iter"], r["total"],
-                    r["tflops_avg"], r["tokens_per_second_per_gpu"],
+                    r.get("log_file", "?"), r.get("iter", 0), r.get("total", 0),
+                    r.get("tflops_avg", 0), r.get("tokens_per_second_per_gpu", 0),
                 )
             exit_code = 0
 
@@ -1201,7 +1199,7 @@ def main() -> None:
                 args.workload,
                 scale,
                 run.get("tokens_per_second_per_gpu"),
-                status,
+                "pass" if run.get("completed", False) else "fail",
                 rccl_commit,
                 extra=extra,
                 precision=run.get("precision"),
--- a/projects/rccl/ci/scripts/test_madengine.py
+++ b/projects/rccl/ci/scripts/test_madengine.py
@@ -17,6 +17,7 @@ Usage from GitHub Actions (on ruby-linux-slurm-scale-runner):
       --nodes 2 \
       --results-dir /apps/rccl-ci/perf
 """
+from __future__ import annotations
 
 import argparse
 import csv
@@ -49,7 +50,7 @@ REGRESSION_THRESHOLD_INFERENCE = 0.05  # 5%
 MADENGINE_REPO = "https://github.com/ROCm/madengine.git"
 MADENGINE_REF = "ec4de0b58c49f05d89dd33e38cc3e81e0fb3d992"
 MAD_REPO = "https://github.com/ROCm/MAD.git"
-MAD_REF = "688828bd9d4a"
+MAD_REF = "688828bd9d4ad5be4196bc6161cabb6d7048ee65"
 MAD_BRANCH = "mad-rccl"
 
 WORKLOAD_CONFIGS = {
@@ -146,14 +147,25 @@ def install_madengine(work_dir: Path) -> Path:
             ["git", "clone", "--depth=1", "--branch", MAD_BRANCH, MAD_REPO, str(mad_dir)],
             check=True,
         )
-        subprocess.run(
-            ["git", "-C", str(mad_dir), "fetch", "--depth=1", "origin", MAD_REF],
-            check=True,
-        )
-        subprocess.run(
-            ["git", "-C", str(mad_dir), "checkout", MAD_REF],
-            check=True,
-        )
+        # Verify the clone landed on the expected SHA. The --branch clone
+        # gives us the branch tip; fetch+checkout is only needed if the
+        # pinned ref differs from the tip (e.g. after the branch moves).
+        head = subprocess.run(
+            ["git", "-C", str(mad_dir), "rev-parse", "HEAD"],
+            capture_output=True, text=True, check=True,
+        ).stdout.strip()
+        if not head.startswith(MAD_REF[:12]):
+            log.info("MAD HEAD %s != pinned %s, fetching...", head[:12], MAD_REF[:12])
+            subprocess.run(
+                ["git", "-C", str(mad_dir), "fetch", "--depth=1", "origin", MAD_REF],
+                check=True,
+            )
+            subprocess.run(
+                ["git", "-C", str(mad_dir), "checkout", MAD_REF],
+                check=True,
+            )
+        else:
+            log.info("MAD HEAD matches pinned ref: %s", head[:12])
 
     log.info("Installing madengine...")
     subprocess.run(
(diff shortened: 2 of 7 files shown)


mkuznet1 (commented):
Nice to see MADEngine wired into RCCL CI — this overlaps heavily with what we're
doing for Primus/SGLang multi-node in MAD, so a few notes to keep us on one path.

1) Image distribution to compute nodes

build_rccl_overlay_image() re-implements docker save -> shared staging ->
per-node `srun docker load`. madengine already does exactly this natively when
the manifest has local_image: true — it saves the image to the MAD_DOCKER_BUILDS
shared dir, workers load the same tar, and it detects a stale image reused under
the same tag via the mad.build_fingerprint label:

https://github.com/ROCm/madengine/blob/ec4de0b58c49f05d89dd33e38cc3e81e0fb3d992/src/madengine/execution/container_runner.py#L2278-L2416

Could you try setting MAD_DOCKER_BUILDS to your shared staging dir and dropping
the custom distribution? If it works on Ruby we'd be on a single mechanism, and
you also get the staleness check for free.

2) Manifest shape + bootstrap interface

The generated manifest puts slurm / env / model_repo / container_mounts at the
top level, but madengine reads deployment_config.{target,slurm,distributed,env_vars}
(see cli/validators.py:154 and primus_backend.py:36). Those top-level blocks are
inert today — the values that actually take effect are the ones duplicated in
--additional-context. Worth moving them into deployment_config so the manifest is
the single source of truth.

Related: nothing sets NCCL_SOCKET_IFNAME / slurm.network_interface. job.sh.j2
only exports it when slurm.network_interface is present, so NCCL bootstrap iface
is left to auto-detection. On Ruby the data-plane ifaces are benic*p1 /31
point-to-point links, so the bootstrap iface should be pinned to fenic0.

We have a static manifest validator that catches both of these (dead config,
missing iface, nodes vs nnodes, leftover placeholders) plus per-cluster env
templates, in https://github.com/ROCm/MAD-private/pull/390 :
  .claude/skills/mad-slurm-multinode/scripts/validate_manifest.sh
  .claude/skills/mad-slurm-multinode/assets/mad.env/mad.env.thor2-bnxt.template
  .claude/skills/mad-slurm-multinode/references/cluster-types.md
Feel free to reuse them.

3) Two things that will break the nightly

- args passes the card name as --model_repo
  ("primus_pyt_megatron_lm_train_llama-3.1-70b_scaleout"), but
  scripts/primus/megatron-lm/run.sh matches base names only
  ("primus_pyt_megatron_lm_train_llama-3.1-70b") with no _scaleout branch and no
  fallback, so $model ends up empty. The model card already carries the correct
  args — dropping the override should fix it.

- Heads-up on a rename: https://github.com/ROCm/MAD-private/pull/388 renames
  those cards from *_scaleout to *_overlay, which will break the hardcoded
  model_repo in WORKLOAD_CONFIGS once it lands. Let's coordinate the timing.

prasanna-amd:
@mkuznet1 Thanks for the thorough review — great catches across the board. Addressed all three in commit 4c3b573:

**1) Image distribution** — Great call on `MAD_DOCKER_BUILDS`. We now set it to `work_dir/docker_builds` so madengine's native tar-based distribution handles multi-node overlay image propagation. Registry push kept as optional fallback. The manifest already had `local_image: true` so this was a straightforward wire-up.

**2) Manifest shape / NIC config** — Added `slurm.network_interface` (set from `NCCL_SOCKET_IFNAME`) to both the manifest and `--additional-context`, so the SLURM template properly exports `NCCL_SOCKET_IFNAME` and `GLOO_SOCKET_IFNAME` via its `{% if network_interface %}` block. Also completed the Ruby NIC config: added `NCCL_IB_TC=104` and `NCCL_IB_QPS_PER_CONNECTION=4`, and fixed `NCCL_SOCKET_IFNAME` to include both `fenic0,enp49s0f0np0` (both needed for bootstrap on Ruby). Will look into the manifest validator and cluster env templates you linked — those would be a nice improvement for a follow-up.

**3) Model naming** — Our `WORKLOAD_CONFIGS` currently uses the plain base names (e.g. `primus_pyt_megatron_lm_train_llama-3.1-70b`) which match `run.sh` today. To handle the upcoming `_scaleout` → `_overlay` rename in MAD-private #388, we added `model_repo_aliases` — `generate_manifest()` now checks `run.sh` for the primary name, and if not found, falls back to `_overlay` then `_scaleout` variants. This way the nightly won't break regardless of which side of the rename lands first.

mkuznet1 (changes requested):
Thanks for pushing this through — 30 fix commits against a cluster madengine had never seen is a
genuinely useful stress test, and nearly everything you had to work around is a real madengine gap
rather than anything wrong on your side.

One request up front, and it is the main point of this review: **please don't merge this as-is.**
madengine is about to get a lot of active development, so anything that lives here as a patch
against madengine's source will rot — quietly, on a nightly job that no PR check exercises. I'd
like as much of this as possible to move into madengine itself, and only the genuinely CI-specific
parts to stay here.

### Before the fix list: run against `mad-rccl`, not `develop`

**`mad-rccl` is the MAD branch this CI should be tracking**, once
[ROCm/MAD#194](https://github.com/ROCm/MAD/pull/194) lands on it. That branch exists to carry the
RCCL-facing benchmarks; it currently sits at develop's tip (`f7e2560`, i.e. after Primus v26.5) and
[#194](https://github.com/ROCm/MAD/pull/194) is open against it. It is prepared in parallel with your work and validated on three cluster
archetypes, and a large part of what this PR builds by hand is already in it. Concretely:

- **The overlay Dockerfiles — please use these rather than rebuilding the mechanism.**
  `docker/primus_megatron_train_rccl_overlay.ubuntu.amd.Dockerfile` builds RCCL from source over
  `rocm/primus:v26.5`, takes repo/branch/commit as build args and verifies at build time that
  every `librccl` on disk matches the built SHA, with optional rdma-core for Broadcom Thor2.
  `docker/sglang_disagg_inference_full_overlay.ubuntu.amd.Dockerfile` does the same for
  RCCL + MoRI + NIXL/Mooncake in one build. That is the same job as the `readelf` kpack sniffing,
  the hardcoded ROCm-SDK venv paths and the in-place `librccl` overwrite at `:385-439` — except it
  is a build-time overlay with a verified SHA rather than a post-hoc file swap, and madengine can
  drive it directly through `context.docker_build_arg`. Swapping a library into a finished image
  and hoping the loader picks it up is the part I would most like to see gone; you already found
  one edge of it when `LD_PRELOAD` had to be reverted for the LLVM symbol clash.
- **The Megatron-LM harness and its metric parsers**, `scripts/primus_megatron-lm/`, writing
  madengine's perf-CSV schema directly. That is what the live-log regex at `:681-739` substitutes
  for.
- **Real model cards** — `primus_pyt_megatron_lm_train_llama-3.1-{8b,70b,405b}_overlay` in
  `models.json`. Your workload is `primus_pyt_megatron_lm_train_llama-3.1-70b`, which is the same
  thing; with a card you would not need to fabricate `built_images`, `built_models` and `context`
  at all.
- **Cluster templates and a bring-up skill** for CX7/Mellanox-RoCE, AMD-AINIC/Pollara and
  Broadcom-Thor2-RoCE: `mad.env` plus manifest templates with the transport variables already
  worked out per archetype.

And the one that bears directly on the exit-code discussion below: #194's last commit fixes a
Primus v26.5 field rename that made both metric parsers match nothing. The symptom is exact —
training completes, SLURM reports COMPLETED, every per-node CSV holds only a header, and `perf.csv`
comes out empty. If you are on a v26.5 base, that is very plausibly the real reason madengine
reports failure on a successful run.

So the concrete ask: point the MAD clone at `mad-rccl` (pinned to a SHA or tag, per the question
below) instead of public `develop`, and build the competitor image from the overlay Dockerfile
there instead of assembling one in the CI script. Could we sync before you invest more in the
local versions? It would delete a large part of this diff rather than relocate it.

### Moving the fixes into madengine

Two ways to do it, whichever is less work for you:

- hand them to me and I'll land them in madengine, or
- open the PRs against madengine yourself and add @mkuznet1, @i-kosarev and @coketaste as reviewers.

I realise this makes the current task bigger rather than smaller. I still think it is the right
call: this CI integration is a long-lived project, and the alternative is maintaining a patch set
against a moving target.

**What I'd like to move upstream.** Already written on our side, so close to free — we hit the
same wall on the same cluster, independently of this PR:

- `--gpus-per-node` emitted unconditionally (`job.sh.j2:9`) — becomes a manifest flag honoured in
  both the template and the script builder, rather than a line deletion
- the 5 s `madengine --version` validation timeout (`slurm.py:244`), too tight for an NFS-backed
  interpreter — raised
- `yum info rocm-libs` hanging on a GPG prompt (`run_orchestrator.py:770`) — capped with a
  `timeout` for every package manager rather than swapped for `rpm -qi`, since the underlying
  problem is a tty-less prompt rather than `yum` specifically

Small, not yet written:

- `$HOME/.local/bin` missing from the sbatch PATH (`job.sh.j2:39`)
- the shared-FS probe not matching `nfs4` (`job.sh.j2:144`)
- `MAD_DOCKER_BUILDS` being discoverable only by reading the source

Needs a little design, but still ours:

- running a pre-built image without having to fabricate `built_images`, `built_models` and a
  `context` block for a run that never builds anything
- a headless submit-node mode, so you don't have to invent `gpu_renderDs` and a HIP version to get
  past the GPU-context guards. Those invented values are the right call given what madengine
  offers today, but they'll be wrong the day the cluster changes and nothing will say so
- a Primus/Megatron perf parser emitting one row per precision, which is what would let the
  live-log scraping go away
- separating "the workload failed" from "metric collection failed", which is what forces the
  exit-code override

**What should stay here:** the JSONL perf store and the rolling regression gate. madengine has no
perf-history or regression concept in its design and I don't intend to add one — that policy
belongs to the consumer, which is where you already have it.

### On the patching helper specifically

`patch_madengine_for_cluster()` (lines 166-314) rewrites madengine's own
`templates/slurm/job.sh.j2`, `deployment/slurm.py` and `orchestration/run_orchestrator.py` by
string match, after a `git clone --depth=1` of madengine's default branch with no pin. Two risks
compound: the clone is unpinned, and every patch fails open — on a miss it logs "already patched
or not found" and continues, so an upstream rename becomes a silent no-op that surfaces days later
as a confusing SLURM error. Given the amount of change coming to madengine, that is close to a
guarantee rather than a risk.

**Could you pin the madengine clone to a SHA in the meantime?** That alone turns the fail-open
patches from a latent surprise into a deterministic one. The pre-warm at `:633-642` also becomes
unnecessary once the probe timeout is raised.

**The same goes for the MAD clone** (`:132-137`), and I think it removes the need for the
`run.sh` grep at `:488-499`. `--model_repo` is a MAD-side argument — `run.sh` parses it at line 33
and the accepted spellings are a MAD contract; madengine never reads it, it just forwards whatever
`args` says. So checking whether your name is a substring of the script, and falling back through
an alias list if not, is really a hedge against MAD moving under you. Pin MAD to a tag or SHA and
the name is simply known. I'd rather not solve this by having madengine resolve names against
`models.json` — the manifest carrying the name explicitly is the design we want, not something to
paper over.

**On the per-node venv (lines 224-287) — is the Python split actually unavoidable?** A `venv` is
only a thin overlay on the interpreter that created it: it points at the head node's
`python3.10` and its compiled extensions are tagged `cpython-310`, so a 3.9 compute node cannot
use it. That is exactly what you hit, and the block works around it by creating a second venv per
node from that node's own `python3` and reinstalling madengine into it.

A conda env doesn't have that property — it ships its own interpreter, so if the env lives on the
shared FS every node sees the same Python and the same `site-packages`, and there is nothing to
bootstrap per node. That is what we use for multi-node runs and it holds across three different
cluster archetypes:

```bash
conda env list | grep -q '^madenv ' || conda create -y -n madenv python=3.12
conda activate madenv
pip install -e ./madengine
```

with the env on the shared work FS and conda sourced in the job script. If that fits your runner,
the whole per-node install disappears — and with it a `pip install` on the critical path of every
node — without needing anything from madengine at all. Worth a try before we treat it as an
upstream gap.

### Questions before merge — I may be wrong on these

Both of the following come from reading the workflows rather than from running them, and you have
clearly been running this against a real cluster, so treat them as questions.

**1. Does the scheduled job actually dispatch?** `therock-rccl-ci-linux.yml:229` passes
`notify_webhook`, while `therock-rccl-test-madengine.yml` declares `teams_webhook` (lines 23 and
50) and reads `inputs.teams_webhook` at 156-157. The sibling pytorch/jax jobs at lines 188, 201
and 214 pass `notify_webhook` because their callees declare it. Reading it cold, that looks like
an undeclared `workflow_call` input, which would fail validation before the job starts — and since
the job is gated on `schedule`/`workflow_dispatch` plus `gfx950-dcgpu`, no PR check would tell us.
If you've already seen it dispatch, ignore this.

**2. Two questions on the HF token.** First, is it reaching the job at all?
`therock-rccl-test-madengine.yml:147` reads `${{ secrets.HF_TOKEN || '' }}`, but I can't find a
`secrets:` block or `secrets: inherit` on the call in `therock-rccl-ci-linux.yml`, which would
make it empty on the scheduled path and fail gated Llama-3.1 access at runtime rather than at
startup.

Second, and more important: can you confirm the token doesn't land anywhere durable? It currently
goes into `env_vars` (`test_madengine.py:486`) → into the manifest (`:563`) → written to disk at
`:570`, and `therock-rccl-test-madengine.yml:178` uploads `manifest.json` with
`retention-days: 30`, which is what CodeQL 768 is pointing at (769, the logging one, was fixed by
`5631aaae`). We had a leak of this kind in madengine and added redaction for `MAD_SECRETS_*` in
printed commands (`fcc4905`), but that only covers what madengine itself prints — it won't help
with a manifest written by the CI script and uploaded as an artifact. Passing the token through
the process environment only, and keeping it out of the manifest, would close that off; madengine
picks up `MAD_SECRETS_*` from the environment, so the manifest doesn't need to carry it.

**3. Which MAD are you actually cloning, and is the script copy doing anything?** `:135` clones
`https://github.com/ROCm/MAD.git` at `--depth=1` with no branch, so that is public MAD's default
branch — `develop`, unpinned, and it moved as recently as yesterday. Two things there I can't
reconcile:

`install_madengine()` copies `MAD/scripts/primus` into the work dir (`:147-152`), but that path no
longer exists on `develop`: [#187](https://github.com/ROCm/MAD/pull/187) (Primus v26.5, merged
2026-08-01) deleted `scripts/primus/megatron-lm/` in favour of the `scripts/Primus` submodule —
which is empty after a non-recursive `--depth=1` clone — and a convention-based
`scripts/primus_train/`. The copy is guarded by `if scripts_src.is_dir()` with no `else`, so on a
clean checkout it is skipped in silence, and then `run_sh.exists()` at `:491` is false, which makes
the alias fallback underneath it unreachable.

Meanwhile the manifest declares `"scripts": "scripts/primus/megatron-lm/run.sh"` (`:518`), and
`--work-dir` points at a persistent path on the cluster. My guess is that tree was populated once,
by hand or by an older MAD layout, and has been there ever since — which works right up until the
runner is rebuilt.

The fix is the one above: clone `mad-rccl` at a pinned SHA, where the Megatron-LM harness lives as
`scripts/primus_megatron-lm/` and is maintained against v26.5. And whichever ref you end up on,
the copy should fail loudly when its source directory is missing rather than skip — a silent
no-op here is what let this survive a breaking upstream restructure without a single error line.

### Worth re-checking

**4. Does `check_regression()` separate precisions?** `5b1af729` started writing one record per
precision run, but the filter at `test_madengine.py:764-784` looks like it keys on workload, scale
and status only. If so, BF16 and FP8 share one rolling 5-run mean and the −2% gate fires on a
change in precision mix rather than on a regression. Could you double-check whether `precision`
needs to be in the filter key?

**5. The exit-code override, and the empty perf underneath it.**
`test_madengine.py:1066-1083` forces exit 0 whenever every training run reached its final iteration.
I understand exactly why it is there, but I think the empty perf it compensates for has a specific,
fixable cause: **the model entry at `:513-524` doesn't declare `multiple_results`.**

That field is how madengine learns where the numbers are. When it is set, madengine passes the
filename into the container as `MAD_OUTPUT_CSV`, the model script writes that CSV, and madengine
turns each row into a perf.csv row — which is also how you get one row per precision without
scraping anything. When it is absent, madengine falls back to grepping the log for a literal
`performance: <value> <metric>` line, and Primus never prints one. So `performance` comes out empty
and the run is marked failed even though training succeeded. The producer for that CSV is the
Megatron-LM report script, which is the other half of question 3 — it isn't on `develop` any more,
and it is in `mad-rccl`.

Setting `multiple_results` and running against `mad-rccl` should remove the need for both the
override and the live-log regex. Separately, madengine genuinely cannot today distinguish "the
workload failed" from "metrics were empty", and that is on our side to fix; as written the override
will also swallow a real node failure that coexists with a completed training loop. Could you narrow
it to the specific empty-perf condition and log loudly whenever it triggers, until we do?

### One note

Thanks for switching to `MAD_DOCKER_BUILDS` and for completing the NIC config with
`slurm.network_interface` — both are the right way to use madengine, and the fact that neither is
discoverable without reading the source is on us to fix.

mkuznet1:
Found the root cause of the empty perf / FAILURE-on-success here, and it's a madengine legacy
trap rather than anything you did wrong.

The `built_models` entry in the generated manifest (`test_madengine.py:532-551`) doesn't declare
`multiple_results`. That field tells madengine the workload reports through a CSV; without it,
madengine silently falls back to grepping the log for a `performance: <value> <metric>` line,
which Primus never prints. Result: empty metric, FAILURE, no error anywhere. That single field is
upstream of the live-log scraping and the exit-code override.

Quickest unblock: copy the `built_models` entry from the MAD model card in `mad-rccl` — it already
carries the right `multiple_results` and `scripts` path — then freeze that manifest as the CI
template.

We're fixing the silent-degradation half in madengine (AICOMNET-366); will follow up here.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+                "args": f"--model_repo {model_repo}",
+                "additional_docker_run_options": (
+                    "--device=/dev/infiniband"
+                    if cluster_config.get("mount_host_ib_libs") and nodes > 1
+                    else ""
+                ),
+                "data": "",
+                "cred": "",
+                "timeout": None,
+            },
```
mkuznet1, line 607: built_models re-declares the MAD card by hand and drops multiple_results (the pinned card has perf_primus-megatron-Llama-3.1-70B.csv). Without it madengine falls back to the performance: <n> <metric> log regex, finds nothing, and writes an empty row with FAILURE — root cause of the live-log scraper (739) and the exit-code override (1145). Load the card from models.json; override only the image.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+            "qos": cluster_config.get("slurm_qos", ""),
+            "network_interface": nccl_env.get("NCCL_SOCKET_IFNAME", ""),
+            "nodes": nodes,
+            "ntasks_per_node": gpus_per_node,
+            "gpus_per_node": gpus_per_node,
+            "time": workload_config["time_limit"],
+            "exclusive": True,
+            "enable_node_check": False,
+            **({"nodelist": nodelist} if nodelist else {}),
+        },
```
mkuznet1: madengine reads deployment config from manifest["deployment_config"] (run_orchestrator.py:225,232) and from --additional-context; a top-level slurm key is ignored. So qos and ntasks_per_node never apply — and the block you pass at 653-662 has no qos, so the job runs without --qos=vip_prio. Nest this under deployment_config, or drop it and keep one source.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+        if "," in ifname:
+            nccl_env["NCCL_SOCKET_IFNAME"] = ifname.split(",")[0]
+    env_vars = {
+        **nccl_env,
+        "NCCL_DEBUG": "WARN",
+    }
+
+    if cluster_config.get("mount_host_ib_libs") and nodes > 1:
+        if "NCCL_SOCKET_IFNAME" in nccl_env:
+            env_vars["GLOO_SOCKET_IFNAME"] = nccl_env["NCCL_SOCKET_IFNAME"]
```
mkuznet1: manifest["env"] is not read by madengine (no reference in the source), so all of env_vars is dead — including GLOO_SOCKET_IFNAME, which is set only here (522-524) and never added to context.docker_env_vars (582-591). For nodes>1 Gloo is left to autodetect on a host with 8 bnxt_re NICs plus fenic0. Move it into docker_env_vars.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+            "time": workload_config["time_limit"],
+            "exclusive": True,
+            "enable_node_check": False,
+            **({"nodelist": nodelist} if nodelist else {}),
+        },
+        "env": env_vars,
+    }
+
+    if workload_config.get("container_mounts"):
+        manifest["container_mounts"] = workload_config["container_mounts"]
```
mkuznet1: container_mounts is not a madengine manifest key either — nothing in the source references it, so /shared:/shared is never mounted into the container. Use context.docker_mounts (currently {} at L592) or additional_docker_run_options.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+                "build_time": 0,
+            },
+        },
+        "built_models": {
+            image_key: {
+                "name": model_repo,
+                "tags": ["pyt", "pretrain", "training"],
+                "dockerfile": "N/A (overlay image)",
+                "scripts": f"scripts/{scripts_dir.name}/run.sh",
+                "n_gpus": str(total_gpus),
```
mkuznet1: n_gpus = str(total_gpus) is 16 on a 2-node run. n_gpus is a per-node field; the MAD card sets -1 deliberately so madengine resolves it per node. Passing the cluster-wide total here is at best redundant.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+                "owner": "",
+                "training_precision": workload_config.get("precision", ""),
```
mkuznet1: training_precision is filled from workload_config["precision"] ("bf16"), but one run of this card emits both BF16 and FP8 rows — which is why the MAD card leaves it empty. Declaring a single precision for a two-precision run is part of what forces the per-precision reconstruction from logs later.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+            "gpu_renderDs": [128 + i for i in range(gpus_per_node)],
+            "docker_env_vars": {
+                "MAD_SYSTEM_GPU_ARCHITECTURE": cluster_config["gpu_target"],
+                "MAD_SYSTEM_NGPUS": str(gpus_per_node),
+                "MAD_SYSTEM_HIP_VERSION": "6.4",
+                "MAD_SYSTEM_GPU_PRODUCT_NAME": cluster_config["gpu_target"],
+                **nccl_env,
+                "NCCL_IB_DISABLE": "0",
+                "NCCL_SOCKET_IFNAME": nccl_env.get("NCCL_SOCKET_IFNAME", "eth0"),
+                "MAD_MULTI_NODE_RUNNER": "",
```
mkuznet1: MAD_MULTI_NODE_RUNNER: "" is exported in both places while distributed.launcher is primus (663-668). The primus launcher deliberately emits no such variable (slurm.py::_generate_primus_command), so this is noise — and a hazard if anything downstream tests presence rather than value.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+    env_vars = {
+        "MAD_SYSTEM_GPU_ARCHITECTURE": gpu_target,
+        "MAD_SYSTEM_NGPUS": str(gpus_per_node),
+        "MAD_SYSTEM_HIP_VERSION": "6.4",
+        "MAD_SYSTEM_GPU_PRODUCT_NAME": gpu_target,
+    }
+    if nccl_env:
+        env_vars.update(nccl_env)
+    env_vars["NCCL_IB_DISABLE"] = "0"
+    env_vars["MAD_MULTI_NODE_RUNNER"] = ""
```
mkuznet1: Same on L590
MAD_MULTI_NODE_RUNNER: "" is exported in both places while distributed.launcher is primus (663-668). The primus launcher deliberately emits no such variable (slurm.py::_generate_primus_command), so this is noise — and a hazard if anything downstream tests presence rather than value.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+    run_sh = scripts_dir / "run.sh"
+    if run_sh.exists():
+        run_sh_text = run_sh.read_text()
+        if model_repo not in run_sh_text:
+            for alias in workload_config.get("model_repo_aliases", []):
+                if alias in run_sh_text:
+                    log.info("Model repo %s not in run.sh, using alias %s",
+                             model_repo, alias)
+                    model_repo = alias
+                    break
```
mkuznet1: The alias fallback picks a model repo by substring-matching run.sh. A rename, or any unrelated mention of the name in that script, silently switches the run to a different card. Resolve the name against models.json instead.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+        network_interface=nccl_env.get("NCCL_SOCKET_IFNAME", ""),
+        nodelist=nodelist,
+    )
+
+    # Step 6: Parse results — try perf.csv first, fall back to live log
+    metric_value = None
+    metric_key = workload_config["metric_key"]
+    if output_csv.exists():
+        perf_data = parse_perf_csv(output_csv, metric_key)
+        if metric_key in perf_data and perf_data[metric_key]:
```
mkuznet1: metric_key is looked up as a column name, but the perf CSV is long-format — model,performance,metric,mode,precision,... — where the metric name is a value in metric (e.g. tok_per_s_per_gpu_avg). No column tokens_per_second_per_gpu exists, so this cannot fire even after multiple_results is fixed. Select the row by metric and read performance.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+    log.info("Parsing perf.csv: %s", csv_path)
+    results = {}
+    with open(csv_path) as f:
+        reader = csv.DictReader(f)
+        for row in reader:
+            log.info("perf.csv row: %s", dict(row))
+            for key, value in row.items():
+                results[key] = value
+
+    return results
```
mkuznet1: results[key] = value across all rows collapses the table to its last row. A real BF16+FP8 run has 8 rows; this keeps only TFLOPS_per_gpu_avg/FP8 and discards the metric you actually want. Return the rows and filter by metric + precision.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+                        "total": int(m.group("total")),
+                        "tflops_avg": float(m.group("tflops_avg")),
+                        "tokens_per_second_per_gpu": float(m.group("tps_avg")),
+                    })
+
+        if current and current.get("iter"):
+            current["completed"] = current["iter"] == current["total"]
+            runs.append(current)
+
+    return runs
```
mkuznet1, line 805: This makes the RCCL repo the third place that must know Primus's iteration-line format (Primus prints it, MAD's benchmark_report.sh parses it, now this). With multiple_results set, perf_super already carries the same per-precision rows. Suggest dropping this once the manifest is fixed; if it stays, demote it to a diagnostic that never feeds the metric or the status.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+                "completed successfully (live log verification)",
+                exit_code, len(live_log_runs),
+            )
+            for r in live_log_runs:
+                log.info(
+                    "  %s: %d/%d iters, %.1f TFLOP/s, %.1f tokens/s/GPU",
+                    r["log_file"], r["iter"], r["total"],
+                    r["tflops_avg"], r["tokens_per_second_per_gpu"],
+                )
+            exit_code = 0
```
mkuznet1: Highest-risk item for CI: any non-zero madengine exit becomes 0 whenever every parsed run reached its last iteration. completed is iter == total over whatever text reached the head-node live log. On 2N a lost node, an OOM on a non-zero rank, or a truncated log still satisfies it and the job goes green — the failure class this workload exists to catch. Gate on "metric not collected", not exit_code != 0.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+    # collection to the login node, but the live log always has them.
+    live_log_runs = parse_live_log_metrics(work_dir)
+    if metric_value is None and live_log_runs:
+        best_run = live_log_runs[-1]
+        if metric_key == "tokens_per_second_per_gpu" and "tokens_per_second_per_gpu" in best_run:
+            metric_value = best_run["tokens_per_second_per_gpu"]
+            log.info("Metric %s = %.1f (from live log)", metric_key, metric_value)
+        elif "tflops_avg" in best_run:
+            metric_value = best_run["tflops_avg"]
+            log.info("Metric tflops_avg = %.1f (from live log, key %s not found)", metric_value, metric_key)
```
mkuznet1: The fallback takes live_log_runs[-1] — the last precision in the log — and reports it unlabelled as the metric in madengine_metric and the summary headline. On a BF16+FP8 run that is the FP8 number. The per-precision loop at 1166 already does this correctly; this scalar should be dropped or carry its precision.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+                args.workload,
+                scale,
+                run.get("tokens_per_second_per_gpu"),
+                status,
+                rccl_commit,
+                extra=extra,
+                precision=run.get("precision"),
+                tflops=run.get("tflops_avg"),
+                tokens_per_sec=run.get("tokens_per_second_per_gpu"),
+            )
```
mkuznet1: One job-wide status is stamped onto every per-precision record. If BF16 passes and FP8 dies, both rows get the same verdict; check_regression then filters on status == "pass" (825) and either poisons the FP8 baseline or throws away a good BF16 point. Status must be per-run, like the records — perf_super carries it per row.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+                metric_value = float(perf_data[metric_key])
+                log.info("Metric %s = %.1f (from perf.csv)", metric_key, metric_value)
+            except (ValueError, TypeError):
+                log.warning("Could not parse metric %s: %s", metric_key, perf_data.get(metric_key))
+
+        run_id = os.environ.get("GITHUB_RUN_ID", "local")
+        run_artifacts = results_dir / "runs" / run_id
+        try:
+            run_artifacts.mkdir(parents=True, exist_ok=True)
+            subprocess.run(["cp", str(output_csv), str(run_artifacts / "perf.csv")], check=True)
```
mkuznet1: subprocess.run(..., check=True) raises CalledProcessError, which is not an OSError — so the handler below catches the mkdir failure but not the cp failure it is aimed at. Use shutil.copy2, or catch subprocess.CalledProcessError too.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+                else:
+                    log.info(msg)
+        regression_msg = "; ".join(regression_msgs) if regression_msgs else "N/A"
+        if is_regression:
+            exit_code = max(exit_code, 1)
+    elif metric_value is not None:
+        is_regression, regression_msg = check_regression(
+            results_dir, args.workload, scale, metric_value,
+            workload_config["type"],
+        )
```
mkuznet1: The fallback path calls check_regression without precision, so it compares only against records whose stored precision is null — a baseline disjoint from the live-log path. Worth normalising the key or stating the intent.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+    metric_value = None
+    metric_key = workload_config["metric_key"]
+    if output_csv.exists():
+        perf_data = parse_perf_csv(output_csv, metric_key)
+        if metric_key in perf_data and perf_data[metric_key]:
+            try:
+                metric_value = float(perf_data[metric_key])
+                log.info("Metric %s = %.1f (from perf.csv)", metric_key, metric_value)
+            except (ValueError, TypeError):
+                log.warning("Could not parse metric %s: %s", metric_key, perf_data.get(metric_key))
```
mkuznet1: Read perf_super.csv/perf_super.json rather than the flat perf.csv. The flat file has no stable schema — its columns come from whatever the model emitted, so width varies per card (38 columns on our 2N Primus run). perf_super is constant-width (31 columns) with model-specific dimensions folded into multi_results as JSON, one object per row. It is only populated when multiple_results is set (see 557-575).

mkuznet1: From perf_super you get straight from madengine what this file rebuilds by hand: multi_results.precision for per-precision keying (instead of scraping the Running: header at 739), per-row status, scale via nnodes/gpus_per_node/n_gpus, madengine's own relative_change, and provenance — docker_image, git_commit, build_number (the SLURM job id) — for the datastore.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+
+        run_id = os.environ.get("GITHUB_RUN_ID", "local")
+        run_artifacts = results_dir / "runs" / run_id
+        try:
+            run_artifacts.mkdir(parents=True, exist_ok=True)
+            subprocess.run(["cp", str(output_csv), str(run_artifacts / "perf.csv")], check=True)
+        except OSError as exc:
+            log.warning("Could not save run artifacts to %s: %s", run_artifacts, exc)
```
mkuznet1: For a single run read perf_entry_super.csv/.json — same 31 columns as perf_super.* but only this run's rows, so no build_number filtering is needed; perf_super.* is the cumulative file. Both pairs are written under fixed names into the run cwd (work_dir, per cwd=work_dir at 701) — -o names only the flat CSV. Copy them into results_dir/runs/<run_id>/ next to perf.csv.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+            '|| echo rocm-libs not installed as RPM", canFail=True))',
+        )
+        if patched != content:
+            run_orch.write_text(patched)
+            log.info(
+                "Patched run_orchestrator.py: replaced 'yum info' with 'rpm -qi' "
+                "to avoid interactive GPG prompt hang"
+            )
+        else:
+            log.info("run_orchestrator.py already patched or yum string not found")
```
mkuznet1, line 357: I'd like this function gone — patching installed madengine by string replacement fails soft (each branch warns and continues), so a madengine bump silently un-patches you. We're upstreaming most of it now (PATH, nfs4, --gpus-per-node, the yum hang) under AICOMNET-366; the per-node install is a genuine madengine gap we'll take as a feature request. Keep it isolated so it can be dropped in one commit.

mkuznet1 (changes requested):
Thanks for taking up the previous round — pinning both clones and moving MAD onto the mad-rccl branch is exactly what we were after. It also means the overlay card and its multiple_results are now available to you directly, instead of having to be reconstructed.

Please model generate_manifest on the manifests we use for manual multinode runs — same MAD branch you pin: https://github.com/ROCm/MAD/tree/mad-rccl/.claude/skills/mad-slurm-multinode/assets/manifests (primus_llama-3.1-70b) — slurm/distributed/env_vars under deployment_config, mounts in context.docker_mounts, multiple_results on the card. Root-level keys are ignored.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+        result = subprocess.run(
+            ["readelf", "-S", str(rccl_lib)],
+            capture_output=True, text=True, timeout=10,
+        )
+        return ".rocm_kpack_ref" in result.stdout
+    except (subprocess.TimeoutExpired, FileNotFoundError):
+        return False
+
+
+def build_rccl_overlay_image(
```
i-kosarev, line 401: 1. :400 + :403 — the tag identifies the wrong commit, so the image freezes. get_rccl_commit() reads RCCL_COMMIT_HASH, which nothing in this PR ever sets, then falls back to git  rev-parse HEAD with no cwd — i.e. $GITHUB_WORKSPACE, which the first checkout populates with ROCm/TheRock pinned at 2e7c190b. The tag is therefore a constant, and the docker image inspect short-circuit at :403 turns that into a cache hit on a persistent runner: after the first nightly, every run reuses the first night's librccl and reports green. The same wrong SHA is recorded as provenance in the JSONL store.

2. :485 — the push tag is not a valid Docker reference, on the scheduled path. rocm/primus:v26.4 → rocm-primus:v26.4, giving ghcr.io/…/rccl-ci:rocm-primus:v26.4-<sha>; a tag cannot contain :, so docker tag fails under check=True. The workflow passes --registry ghcr.io/rocm/rocm-systems, so this is not latent. It also drops gpu_target, which the local tag carries — gfx942 and gfx950 would overwrite each other.

3. :469 — the non-kpack fallback is unreachable and the swap can be a silent no-op. readlink -f exits 0 for a non-existent final component, so when /opt/rocm/lib/librccl.so is absent the || find … never runs and cp creates a stray file instead of overwriting the versioned library the loader maps. On a primus:v26.4 base ROCm lives under _rocm_sdk_libraries/lib and _rocm_sdk_devel/lib, so /opt/rocm/lib/librccl.so.1 may not exist at all and the run quietly measures stock RCCL. No pipefail, no existence check.

4. Nothing verifies the swap. verify_rccl_override() is used only by the pytorch/jax scripts and only asserts a file exists on disk. Combined with 1 and 3, every failure mode above is silent.

5. :446-448, :417. The SDK paths hardcode python3.12 and skip torch/lib. The staging copy skips when the destination exists, so a reused --work-dir ships the previous run's library under a new tag — safe in CI, where WORK_DIR carries github.run_id, but not for the manual runs the validation table came from.

Rather than hardening this, I would recommend to take docker/primus_megatron_train_rccl_overlay.ubuntu.amd.Dockerfile from the mad-rccl branch you already pin: it takes repo/branch/commit and BUILD_GPU_TARGETS as build args, madengine drives it through context.docker_build_arg, and it asserts at build time that every librccl on disk carries the built SHA. That replaces the readelf kpack sniffing, the hardcoded venv paths and the in-place overwrite with a build-time overlay whose provenance is checked rather than assumed — swapping a library into a finished image and hoping the loader picks it up is the part worth deleting. One gap to close in the move: that Dockerfile builds RCCL from source, so if CI must inject the prebuilt TheRock artifact instead, add that as an input to the same Dockerfile so the SHA gate still applies to it.

i-kosarev (changes requested):

prasanna-amd:
@mkuznet1 @i-kosarev Thanks for the detailed review — addressed the critical items in commit a3089bc:

**Fixed in this commit:**

1. **Manifest structure rewrite** — `deployment_config.slurm`, `deployment_config.distributed`, `deployment_config.env_vars` now match the reference template from `mad-rccl`. Root-level `slurm`, `env`, `container_mounts` keys removed. Mounts moved to `context.docker_mounts`, env vars to `context.docker_env_vars`.

2. **`multiple_results` added** — `perf_primus-megatron-Llama-3.1-70B.csv` on the built_models card. Validated on Ruby: madengine now produces both BF16 and FP8 runs, `perf_entry_super.json` populated with 8 result rows (4 metrics × 2 precisions). This was the root cause of empty perf files.

3. **`n_gpus` → `"-1"`**, `training_precision` → `""` — matches MAD card behavior (madengine resolves per-node, card emits both BF16+FP8).

4. **`MAD_MULTI_NODE_RUNNER`** — removed from both env locations.

5. **`GLOO_SOCKET_IFNAME`** — moved into `context.docker_env_vars` and `deployment_config.env_vars` (was in dead `manifest["env"]`).

6. **Commit provenance** — `get_rccl_commit()` no longer falls back to `git rev-parse HEAD` (which returned TheRock's constant commit). Uses `RCCL_COMMIT_HASH` → `GITHUB_RUN_ID` → sha256 of librccl.so.

7. **Push tag format** — replaces both `/` and `:` to avoid invalid Docker reference.

8. **Results parsing** — reads `perf_entry_super.json` first (structured 31-column format), falls back to `perf.csv`. Uses `shutil.copy2` instead of `subprocess.run` for file copy (fixes OSError vs CalledProcessError mismatch).

9. **Exit code override** — now requires `metric_value is not None` (missing metrics keep the failure).

10. **Workflow YAML** — added `secrets: inherit` + `packages: write` to madengine job. Added `perf_entry_super.*` files to upload artifact step.

**Deferred to follow-up (not blocking merge):**

- **`patch_madengine()` removal** — keeping isolated pending upstream fixes (AICOMNET-366). Ready to drop in one commit once upstream lands.
- **Model alias via `models.json`** — currently uses `run.sh` substring matching. Will migrate once we have a stable models.json schema to depend on.
- **Live-log scraper removal** — kept as fallback safety net. `perf_entry_super` is now the primary path; live-log only fires if structured data is empty.
- **Per-precision status** — still stamps job-wide status across all rows. Will use `perf_entry_super` per-row status in follow-up.
- **Regression check precision key** — fallback path compares against null-precision baseline. Will normalize once we have enough historical data with precision keys.

Validated on Ruby single-node (1N/8GPU): BF16 754.2 TFLOP/s/GPU, FP8 1248.7 TFLOP/s/GPU. All perf files populated correctly.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1247 @@
...
+                regression_msgs.append(f"[{run_prec}] {msg}")
+                if reg:
+                    is_regression = True
+                    log.warning(msg)
+                else:
+                    log.info(msg)
+        regression_msg = "; ".join(regression_msgs) if regression_msgs else "N/A"
+        if is_regression:
+            exit_code = max(exit_code, 1)
+    elif metric_value is not None:
```
mkuznet1: The structured path is an `elif`: whenever the live log parses, regression compares live-log tokens/s and `perf_entry_super` is never consulted. The scraper is primary here, not the fallback. `perf_entry_super.json` already carries per-precision rows in `multi_results`, which also removes the need to scrape the `Running:` header at :774 for the precision key.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1247 @@
...
+                args.workload,
+                scale,
+                run.get("tokens_per_second_per_gpu"),
+                "pass" if run.get("completed", False) else "fail",
+                rccl_commit,
+                extra=extra,
+                precision=run.get("precision"),
+                tflops=run.get("tflops_avg"),
+                tokens_per_sec=run.get("tokens_per_second_per_gpu"),
+            )
```
mkuznet1: Same for the datastore: every record here comes from the log scraper, including the per-run status added in this commit. `perf_entry_super.json` carries `status` per row, so this loop can be driven from it, with the live-log branch kept only for when structured output is missing.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1247 @@
...
+    if perf_results:
+        for row in perf_results:
+            perf_val = row.get("performance", "")
+            if perf_val:
+                try:
+                    metric_value = float(perf_val)
+                    log.info("Metric %s = %.1f (from structured results)",
+                             row.get("metric", "?"), metric_value)
+                except (ValueError, TypeError):
+                    pass
```
mkuznet1: `metric_value` is overwritten on every row, so this keeps the last of 8 rows (4 metrics x 2 precisions) — TFLOP/s or tokens/s, whichever lands last. Nothing filters on `metric == metric_key`, and `metric_key` (:68) is now read nowhere. Same last-row collapse as the old `parse_perf_csv`, moved into the caller. Filter on `row["metric"]`, key the row by `multi_results.precision`.

mkuznet1 (changes requested):
Manifest rewrite matches what madengine actually reads (run_orchestrator.py:225-234) — good turnaround.

Acceptance here is multi-node: Llama-3.1-70B on 2 nodes minimum, green, with metrics in perf_entry_super. The description still lists 2N as follow-up; single-node Ruby does not exercise the path this CI exists for.

perf_entry_super is read but unused — metric, regression and datastore still come from the log scraper (inline).

i-kosarev:
> @mkuznet1 @i-kosarev Thanks for the detailed review — addressed the critical items in commit [a3089bc](https://github.com/ROCm/rocm-systems/commit/a3089bc30f355d0c9f6a4a19dff0da3ddf104adf):
> 
> **Fixed in this commit:**
> 
> 1. **Manifest structure rewrite** — `deployment_config.slurm`, `deployment_config.distributed`, `deployment_config.env_vars` now match the reference template from `mad-rccl`. Root-level `slurm`, `env`, `container_mounts` keys removed. Mounts moved to `context.docker_mounts`, env vars to `context.docker_env_vars`.
> 2. **`multiple_results` added** — `perf_primus-megatron-Llama-3.1-70B.csv` on the built_models card. Validated on Ruby: madengine now produces both BF16 and FP8 runs, `perf_entry_super.json` populated with 8 result rows (4 metrics × 2 precisions). This was the root cause of empty perf files.
> 3. **`n_gpus` → `"-1"`**, `training_precision` → `""` — matches MAD card behavior (madengine resolves per-node, card emits both BF16+FP8).
> 4. **`MAD_MULTI_NODE_RUNNER`** — removed from both env locations.
> 5. **`GLOO_SOCKET_IFNAME`** — moved into `context.docker_env_vars` and `deployment_config.env_vars` (was in dead `manifest["env"]`).
> 6. **Commit provenance** — `get_rccl_commit()` no longer falls back to `git rev-parse HEAD` (which returned TheRock's constant commit). Uses `RCCL_COMMIT_HASH` → `GITHUB_RUN_ID` → sha256 of librccl.so.
> 7. **Push tag format** — replaces both `/` and `:` to avoid invalid Docker reference.
> 8. **Results parsing** — reads `perf_entry_super.json` first (structured 31-column format), falls back to `perf.csv`. Uses `shutil.copy2` instead of `subprocess.run` for file copy (fixes OSError vs CalledProcessError mismatch).
> 9. **Exit code override** — now requires `metric_value is not None` (missing metrics keep the failure).
> 10. **Workflow YAML** — added `secrets: inherit` + `packages: write` to madengine job. Added `perf_entry_super.*` files to upload artifact step.
> 
> **Deferred to follow-up (not blocking merge):**
> 
> * **`patch_madengine()` removal** — keeping isolated pending upstream fixes (AICOMNET-366). Ready to drop in one commit once upstream lands.
> * **Model alias via `models.json`** — currently uses `run.sh` substring matching. Will migrate once we have a stable models.json schema to depend on.
> * **Live-log scraper removal** — kept as fallback safety net. `perf_entry_super` is now the primary path; live-log only fires if structured data is empty.
> * **Per-precision status** — still stamps job-wide status across all rows. Will use `perf_entry_super` per-row status in follow-up.
> * **Regression check precision key** — fallback path compares against null-precision baseline. Will normalize once we have enough historical data with precision keys.
> 
> Validated on Ruby single-node (1N/8GPU): BF16 754.2 TFLOP/s/GPU, FP8 1248.7 TFLOP/s/GPU. All perf files populated correctly.

I see at least one critical thing that's left, it is in build_rccl_overlay_image() — non-kpack fallback can silently keep stock RCCL (lines 484-486 of test_madengine.py) 

```
  RCCL_REAL=$(readlink -f /opt/rocm/lib/librccl.so 2>/dev/null || \
              find /opt/rocm*/lib -name 'librccl.so.*.*' -not -type l | head -1)
  cp /tmp/rccl_ci/librccl.so "$RCCL_REAL"
```
readlink -f on a non-existent final path component still exits 0 and prints the resolved (non-existent) path — the || fallback to find never fires. If /opt/rocm/lib/librccl.so isn't present (which happens on rocm/primus-style images, where the runtime lives under _rocm_sdk_libraries/lib / _rocm_sdk_devel/lib instead of the classic /opt/rocm/lib layout), cp just creates a new, unlinked librccl.so file. The actual versioned library the loader resolves (librccl.so.1) is never touched, the build reports success, and the job runs against stock RCCL instead of the target commit — with no error anywhere.

This isn't hypothetical: we hit exactly this failure mode building the Primus overlay on the same base-image family, which is why primus_megatron_train_rccl_overlay.ubuntu.amd.Dockerfile (mad-rccl) has an explicit [ -e /opt/rocm/lib/librccl.so.1 ] check plus a full sweep over every non-symlink librccl.so* on disk instead of resolving a single guessed path.

Given that + the fact that nothing after the cp verifies the swap landed (no readelf/SHA check against the built library, no ldd on the actual runtime lib the process loads), this function can produce a green nightly that measured nothing new. Minimum fix: set -o pipefail, explicit [ -e "$RCCL_REAL" ] || exit 1 before the cp, and a build-time gate that confirms the copied .so's hash/symbol matches the CI-built artifact.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Fixed in a3089bc. Added `multiple_results` field to built_models card, pointing to `perf_primus-megatron-Llama-3.1-70B.csv`. Validated in Run 13 — `perf_entry_super.json` is now populated with per-precision rows.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Fixed in a3089bc. Restructured manifest: slurm config now under `deployment_config.slurm`, removed the redundant top-level `slurm` key and the duplicate `additional_context` block in `run_madengine()`.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Fixed in a3089bc. Moved `GLOO_SOCKET_IFNAME` into `context.docker_env_vars`. Top-level `env` block removed.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Fixed in a3089bc. Mounts now in `context.docker_mounts`. Top-level `container_mounts` removed.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Fixed in a3089bc. Set `n_gpus: "-1"` so madengine resolves per-node.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Fixed in a3089bc. Set `training_precision: ""` so the card runs both BF16 and FP8.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Fixed in a3089bc. Removed `MAD_MULTI_NODE_RUNNER` from both locations.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Fixed in a3089bc (same as above — removed from both places).

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Agreed on the fragility. Keeping the alias fallback for now since models.json isn't available at manifest generation time (it's inside the MAD repo scripts dir, not a standalone file we can query). Will revisit when we migrate to the mad-rccl overlay Dockerfile.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Fixed in a3089bc. Switched to `parse_perf_results()` which reads `perf_entry_super.json` — long-format rows with metric name as a value in the `metric` column, performance value in `performance`.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Fixed in a3089bc. `parse_perf_results()` now returns all rows from `perf_entry_super.json` instead of collapsing to the last row.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Agree — once `perf_entry_super.json` is proven reliable across a few CI runs, the live-log parser should be demoted to diagnostic-only (no metric/status influence). Keeping it as fallback for now until we have confidence in the structured output path.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Agreed this is the highest-risk item. In a3089bc the exit-code override now also requires `metric_value is not None` (not just iteration completion). For 2N with a lost node or OOM on non-zero rank, the structured results from `perf_entry_super.json` will carry per-row status — gating on that instead of live-log completion is the right next step.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Fixed in a3089bc. The per-precision loop now handles metric reporting correctly. The scalar fallback is only used when structured results have no metrics — will carry precision label in follow-up.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Fixed in a3089bc. Status is now per-run: when live_log_runs are available, each JSONL record gets `"pass" if run.completed else "fail"` independently.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Fixed in a3089bc. Switched to `shutil.copy2()` instead of `subprocess.run(["cp", ...])`. Exception handler now catches `OSError` which covers both mkdir and copy failures.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Acknowledged. The fallback path (no live_log_runs) is intentionally a coarser check — it runs when structured results exist but live logs don't. In that case precision is unknown, so it compares against the null-precision baseline. Will normalize once perf_entry_super is the sole source.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Fixed in a3089bc. `parse_perf_results()` now reads `perf_entry_super.json` first (31 fixed columns), falls back to `perf.csv` only if the super file is missing. Also copying `perf_entry_super.*` files to run artifacts dir.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Fixed in a3089bc. Now reading `perf_entry_super.json` (single-run rows). Copying both `perf_entry_super.csv` and `perf_entry_super.json` to `results_dir/runs/<run_id>/`. Also added to the workflow artifact upload step.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Agreed — keeping it isolated in a single function so it can be dropped in one commit once the upstream patches land (AICOMNET-366). The per-node install gap is real and appreciated as a feature request.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Thanks for the thorough review. Addressed items 1, 2, 4 in a3089bc; items 3 and 5 acknowledged:

1. **Provenance/cache hit** — Fixed. `get_rccl_commit()` no longer falls back to `git rev-parse HEAD`. Resolution order is now: `RCCL_COMMIT_HASH` env → `GITHUB_RUN_ID` → sha256 of librccl.so. Each CI run produces a unique tag.

2. **Push tag double-colon** — Fixed. `base_image.replace("/", "-").replace(":", "-")` produces a valid Docker reference.

3. **Non-kpack fallback** — Acknowledged. On primus:v26.4 the kpack path is always taken (TheRock builds with kpack), so the non-kpack branch is effectively dead code for this base image. Will remove or gate it properly in the follow-up when we migrate to the mad-rccl overlay Dockerfile.

4. **No swap verification** — Partially addressed. The sha256-based tag means a cache hit now implies the correct library. Full build-time verification (asserting the SHA of every librccl on disk) is exactly what the mad-rccl Dockerfile provides — agreed that's the right end state.

5. **Hardcoded python3.12 / stale staging** — Acknowledged. CI is safe (WORK_DIR has github.run_id). For manual runs, `--skip-overlay-build` bypasses this entirely.

Agreed on the recommendation to migrate to `primus_megatron_train_rccl_overlay.ubuntu.amd.Dockerfile` from mad-rccl. The gap you noted (it builds from source, CI needs to inject prebuilt TheRock artifacts) is the main blocker — will coordinate on adding that input to the Dockerfile as a follow-up.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Fixed in 1b9e135e36. Built a single `precision_results` list that filters on `row["metric"] == metric_key` instead of overwriting `metric_value` on every row. Each matching row becomes a separate entry with its own precision, so BF16 and FP8 values are preserved independently.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Fixed in 1b9e135e36. `perf_entry_super.json` is now the primary data source. The code builds `precision_results` from structured data first; live-log scraping only fires as a fallback when `not precision_results`. Regression and datastore both consume the same `precision_results` list.

projects/rccl/ci/scripts/test_madengine.py
prasanna-amd: Fixed in 1b9e135e36. Datastore append loop now iterates over `precision_results` (which comes from perf_entry_super.json when available). Per-row status from the structured output is preserved — no more job-wide stamp.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+        current = None
+        with open(log_path) as f:
+            for raw_line in f:
+                line = _ANSI_RE.sub("", raw_line)
+
+                hdr = _RUN_HEADER_RE.search(line)
+                if hdr:
+                    if current and current.get("iter"):
+                        current["completed"] = current["iter"] == current["total"]
+                        runs.append(current)
```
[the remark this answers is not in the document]
prasanna-amd: Fixed in 1b9e135e36. Header-only runs are now preserved with `completed=False`. A run with zero iterations no longer passes the `all(completed)` check.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+            workload_config["type"],
+        )
+        if is_regression:
+            log.warning(regression_msg)
+            exit_code = max(exit_code, 1)
+        else:
+            log.info(regression_msg)
+
+    # Step 8: Append result to datastore (one record per precision run)
+    status = "pass" if exit_code == 0 else "fail"
```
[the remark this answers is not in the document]
prasanna-amd: Fixed in 1b9e135e36. When `precision_results` is empty (no parseable metric from either structured or live-log source), the datastore now explicitly records `status="fail"`.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+    commit = os.environ.get("RCCL_COMMIT_HASH", "")
+    if commit:
+        return commit
+
+    try:
+        result = subprocess.run(
+            ["git", "rev-parse", "HEAD"],
+            capture_output=True, text=True, check=True,
+        )
+        return result.stdout.strip()[:12]
```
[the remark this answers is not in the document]
prasanna-amd: Fixed in a3089bc. `get_rccl_commit()` no longer falls back to `git rev-parse HEAD`. Resolution order: `RCCL_COMMIT_HASH` env → `GITHUB_RUN_ID` → sha256 of librccl.so.

projects/rccl/ci/scripts/test_madengine.py
```diff
@@ -0,0 +1,1249 @@
...
+        log.info("Building overlay image: %s", tag)
+        subprocess.run(
+            ["docker", "build", "--network=none", "-t", tag,
+             "-f", str(dockerfile), str(work_dir)],
+            check=True,
+        )
+        log.info("Overlay image built: %s", tag)
+
+    if registry:
+        push_tag = f"{registry}/rccl-ci:{base_image.replace('/', '-')}-{rccl_commit}"
```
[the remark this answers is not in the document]
prasanna-amd: Fixed in a3089bc. Push tag now replaces both `/` and `:` in `base_image`, so tags like `rocm/primus:v26.4` produce valid Docker references.

.github/workflows/therock-rccl-ci-linux.yml
```diff
@@ -208,3 +212,18 @@ jobs:
       artifact_run_id: ${{ inputs.artifact_run_id || github.run_id }}
       notify_email: ${{ inputs.notify_email }}
       notify_webhook: ${{ inputs.notify_webhook }}
+
+  therock-test-madengine:
+    name: "Test MADEngine workloads (scheduled)"
+    if: ${{ !cancelled() && inputs.amdgpu_families == 'gfx950-dcgpu' && (inputs.event_name == 'schedule' || inputs.event_name == 'workflow_dispatch') }}
+    needs: [therock-build-linux]
+    uses: ./.github/workflows/therock-rccl-test-madengine.yml
```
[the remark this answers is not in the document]
prasanna-amd: Fixed in a3089bc. Added `secrets: inherit` to the nested workflow call so `HF_TOKEN` propagates.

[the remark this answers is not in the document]
prasanna-amd: Fixed in a3089bc. Added `packages: write` at both the job level and the top-level caller workflow.

i-kosarev:
Work transferred to #10396

PR #9056, opened 2026-07-22
ci: bump TheRock ref to 2e7c190b99c (2026-07-22)

araravik-psd:
Description:
  Bumps the pinned TheRock ref from `22004a6bc8a` (2026-07-20) to `2e7c190b99c` (2026-07-22).

  ## What's in this bump

  Includes the fix from ROCm/TheRock#6776 which adds `openmpi` to rocprofiler-sdk's
  `RUNTIME_DEPS` in the TheRock profiler CMakeLists. This resolves the rocprofiler-sdk
  configure failure caused by `find_package(MPI)` being intercepted by TheRock's dep
  provider after openmpi was added as a super-project component.

  ## Why

  Unblocks RCCL CI which has been failing since the previous TheRock bump to `22004a6`
  introduced openmpi as a super-project component without a corresponding dep declaration
  in rocprofiler-sdk.

  ## Follow-up

  A separate PR in rocprofiler-sdk will guard the `find_package(MPI)` call at the source
  level as the longer-term fix, keeping MPI out of standard builds entirely.

### DIFF
--- a/.github/workflows/therock-build-linux.yml
+++ b/.github/workflows/therock-build-linux.yml
@@ -50,7 +50,7 @@ jobs:
         with:
           repository: "ROCm/TheRock"
           path: "TheRock"
-          ref: 22004a6bc8a28393f8b772b7f6e2a3040f11cd8d # 2026-07-20
+          ref: 2e7c190b99caa226a8644eda6eca720b3db102d5 # 2026-07-22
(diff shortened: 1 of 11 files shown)


geomin12 (approved):
lgtm IF ci is passing

PR #9057, opened 2026-07-22
feat(rocr): Changes to allow FreeBSD compilation

dayatsin-amd:
## Summary

This PR cherry-picks the FreeBSD-portability header/source changes from #8430 and
fixes the Linux build breakage that was failing that PR's CI.

The last commit on #8430 (`Update hsakmttypes.h to also include freebsd include`)
misspelled `defined` as `defind` in the `#elif` guard:

```c
#elif defined(__linux__) || defind(__FreeBSD__)
```

The preprocessor treated `defind(__FreeBSD__)` as a function-like use of an
undefined macro, emitting `missing binary operator before token "("` and skipping
the entire Linux/FreeBSD typedef block. That left `HSAKMTAPI` and the
`HSAuintXX`/`HSAintXX` types undefined, cascading into hundreds of errors across
`libhsakmt` and breaking the Linux packages build on gfx94X, gfx950 and gfx125X.

This branch restores the correct spelling so both `__linux__` and `__FreeBSD__`
select the typedef block, unblocking the Linux build while preserving the FreeBSD
support work.

## JIRA ID

ISSUE ID: #9999999

Cherry-picked from #8430 (Sponsored By: The FreeBSD Foundation).

## Test plan

- [ ] `hsakmttypes.h` now preprocesses cleanly on Linux (verified `-fsyntax-only`)
- [ ] Linux Packages build passes CI (gfx94X / gfx950 / gfx125X)

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
@@ -682,18 +682,19 @@ void AqlQueue::CloseRingBufferFD(const char* ring_buf_shm_path, int fd) const {
 
 int AqlQueue::CreateRingBufferFD(const char* ring_buf_shm_path,
                                  uint32_t ring_buf_phys_size_bytes) const {
-#ifdef __linux__
+#if defined(__linux__) || defined(__FreeBSD__)
   int fd;
-#ifdef HAVE_MEMFD_CREATE
-  fd = syscall(__NR_memfd_create, ring_buf_shm_path, 0);
 
-  if (fd == -1) return -1;
+  fd = memfd_create(ring_buf_shm_path, 0);
 
-  if (ftruncate(fd, ring_buf_phys_size_bytes) == -1) {
-    CloseRingBufferFD(ring_buf_shm_path, fd);
+  if (fd != -1) {
+    if (ftruncate(fd, ring_buf_phys_size_bytes) == 0) {
+      return fd;
+    }
+    close(fd);
     return -1;
   }
-#else
+
   fd = shm_open(ring_buf_shm_path, O_CREAT | O_RDWR | O_EXCL, S_IRUSR | S_IWUSR);
 
   if (fd == -1) return -1;
@@ -702,10 +703,10 @@ int AqlQueue::CreateRingBufferFD(const char* ring_buf_shm_path,
     CloseRingBufferFD(ring_buf_shm_path, fd);
     return -1;
   }
-#endif
+
   return fd;
 #else
-  assert(false && "Function only needed on Linux.");
+  assert(false && "Function only needed on Linux and FreeBSD.");
   return -1;
 #endif
 }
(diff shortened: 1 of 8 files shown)


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
```diff
@@ -682,18 +682,19 @@ void AqlQueue::CloseRingBufferFD(const char* ring_buf_shm_path, int fd) const {
...
 int AqlQueue::CreateRingBufferFD(const char* ring_buf_shm_path,
                                  uint32_t ring_buf_phys_size_bytes) const {
-#ifdef __linux__
+#if defined(__linux__) || defined(__FreeBSD__)
   int fd;
-#ifdef HAVE_MEMFD_CREATE
-  fd = syscall(__NR_memfd_create, ring_buf_shm_path, 0);
 
-  if (fd == -1) return -1;
+  fd = memfd_create(ring_buf_shm_path, 0);
```
cfreeamd, line 688: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** This replaces the previous `HAVE_MEMFD_CREATE`-gated raw `syscall(__NR_memfd_create, ...)` with the glibc `memfd_create()` library wrapper. The wrapper is declared in `<sys/mman.h>` only when `_GNU_SOURCE` is defined and only exists in glibc >= 2.27, whereas the raw syscall needed just the syscall number. In practice this is fine — g++ implicitly defines `_GNU_SOURCE` for C++ TUs, `<sys/mman.h>` is already included, ROCm's min toolchain is well past 2.27, and CI is green — so the portability assumption is just now implicit rather than CMake-verified. If any supported build config uses a stricter/older libc this would surface as an implicit-declaration hard error. Also note the (now dead) `HAVE_MEMFD_CREATE` `CHECK_SYMBOL_EXISTS` in `runtime/hsa-runtime/CMakeLists.txt` is no longer referenced and could be removed in a follow-up.

cfreeamd (commented):

Clean supersede of #8430 — the three prior review findings (dead-code `#ifdef` nesting in amd_aql_queue.cpp, `thr_self(0)` null-deref UB in amd_core_dump.cpp, redundant `<stdatomic.h>` in amd_blit_kernel.cpp) are all resolved, and the `defind` typo that broke #8430's HEAD is fixed in hsakmttypes.h. amd_topology.cpp, memory.h, amdxdna_accel.h, and the amd_kfd_driver.cpp include restructure look correct (verified the dropped `<fcntl.h>` is unused in that file). One non-blocking [Question] inline about the memfd_create wrapper's implicit glibc/`_GNU_SOURCE` assumption.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
dayatsin-amd: Yes, we decided to blindly assume glibc has the support for memfd_create() to keep code simple.
https://github.com/ROCm/rocm-systems/pull/8430#discussion_r3589001580

cfreeamd (approved):

PR #9067, opened 2026-07-23
fix(rocr): add missing FARPROC typedef for Windows SDK >= 10.0.28000.0

fcui-amd:
## Motivation

cherry pick pr https://github.com/ROCm/librocdxg/pull/54
This is to fix building librocdxg against Windows SDK 10.0.28000.0 failure.

## Technical Details

add typedef int (*FARPROC)(); (guarded with #ifndef FARPROC) to wddm_types.h.

## Test Plan

build rocdxg with WIN SDK 10.0.28000.0

## Test Result

build pass

## Submission Checklist


projects/rocr-runtime/libhsakmt/include/impl/wddm/wddm_types.h
```diff
@@ -82,6 +82,11 @@ typedef const int16_t *PCWSTR;
 #define DECLARE_HANDLE(name) struct name##__{int unused;}; typedef struct name##__ *name
 #define C_ASSERT(e) typedef char __C_ASSERT__[(e)?1:-1]
 
+/* FARPROC is used by newer Windows SDK (>=10.0.28000.0) d3dkmthk.h */
+#ifndef FARPROC
+typedef int (*FARPROC)();
```
cfreeamd, line 87: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `FARPROC` is introduced here as a typedef, not a macro, so the surrounding `#ifndef FARPROC` guard is always true and effectively does nothing. This file already establishes the idiom for making such guards work — see lines 78-80, where each guarded type is both typedef'd and self-#defined (`#define ULONG ULONG`, etc.). To match that convention and make the guard meaningful (so another header providing `FARPROC` under its own `#ifndef` won't produce a *conflicting* redefinition), add the companion define:

```c
#ifndef FARPROC
typedef int (*FARPROC)();
#define FARPROC FARPROC
#endif
```

Low severity: `#pragma once` prevents self-double-inclusion and C11/C++ allow an identical typedef to repeat, so this only bites if a second definition differs. One-line fix, follows the file's own pattern.

**[Nit]** The real Windows declaration is `INT_PTR (WINAPI *FARPROC)(void)`; this shim uses `int (*)()` (int return vs pointer-width, unspecified args vs `(void)`). Fine for a compile-only shim that's never called through — flagging only for awareness.

cfreeamd (commented):

Reviewed the ROCr change in `wddm_types.h`. The FARPROC shim addition is correctly placed and unblocks the newer Windows SDK (>=10.0.28000.0) build. One inline nit about matching the file's existing typedef+`#define` guard idiom.

projects/rocr-runtime/libhsakmt/include/impl/wddm/wddm_types.h
dayatsin-amd: I agree with this suggestion

With this change, patch LGTM

dayatsin-amd (approved):

PR #9074, opened 2026-07-23
feat(rocprofiler-systems): Integrate time-windows into control session

marantic-amd:
## Motivation

Part 2  of unifying control logic with different triggers. 
- PR 1 introduced `control::session`/`subscriber`/`trigger` and an inert `time_window` trigger template.
- This PR wires that machinery into the runtime for the first time, replacing the old polling-thread-based `TRACE_DELAY`/`TRACE_DURATION`/`SAMPLING_DURATION` scheduling with `time_window` triggers, and adds a `scope` dimension so a subsystem (sampling) can be paused independently of the rest of the tool. 
- Also adds a CPU-time clock option (`ROCPROFSYS_TRACE_PERIOD_CLOCK_ID=cputime`) for trace windows.

## Technical Details

- Deletes `constraint::stages`, `spec::operator()`, and the detached polling thread they drove;   `TRACE_DELAY`/`TRACE_DURATION` are now scheduled by a `time_window<clocks::steady>` (or `clocks::posix`) trigger registered with the session.
- Adds `control::scope` (`global`/`sampling_only`) to `trigger`/`subscriber`/`vote_entry`/ `session`; a subscriber's `scopes` field selects which trigger scopes it responds to.  `SAMPLING_DURATION` is wired through a second, `sampling_only`-scoped `time_window` instance,  replacing the old dedicated `omni.samp.dur` worker thread.
- Adds `clocks::posix` (a `clock_gettime`/`clock_nanosleep`-backed clock satisfying the same duck-typed interface as `clocks::steady`) and `constraint::get_trace_period_clock_id()`, which maps `ROCPROFSYS_TRACE_PERIOD_CLOCK_ID` (`realtime`/`cputime`) to the POSIX clock id used.
- Simplifies `constraint::spec`/`constraint.hpp`. Removes `clock_identifier`, `get_valid_clock_ids()`, and the per-spec clock field (the clock is now a single global setting, not per-window); `get_trace_specs()`'s `TRACE_PERIODS` parsing logic is extracted into
 testable pure functions (`parse_trace_period_entry`/`parse_trace_periods`).

## Issue Tracking

JIRA ID: AIPROFSYST-91

## Test Plan

- Full project build stacked on PR 1's tip.
- Full GTest unit suite (`bin/rocprof-sys-unit-tests`), unfiltered, plus a targeted filter for this PR's touched subsystems  (`--gtest_filter="posix_clock_test.*:session_scope_test.*:constraint_test.*:roctx_marker_gating_test.*"`).

## Test Result

- Build: clean, whole project, no warnings introduced.
- Full GTest suite: 1191/1191 passed at the final state.
- Targeted filter: 18/18 passed.

## Submission Checklist


projects/rocprofiler-systems/CHANGELOG.md
```diff
@@ -26,6 +26,14 @@ Full documentation for ROCm Systems Profiler is available at [https://rocm.docs.
...
 - `ROCPROFSYS_BUILD_TESTING` no longer implies `ROCPROFSYS_BUILD_EXAMPLES`.
 
+### Fixed
+
+- `ROCPROFSYS_TRACE_DELAY`/`ROCPROFSYS_TRACE_DURATION` now actually gate GPU context
+  startup, producing a real gap in cached GPU/RocPD data. Previously they only
+  suppressed downstream category emission. For GPU-only tracing with no marker
+  domain or trace region configured, the configured delay had no effect at all on
+  when GPU data collection actually began.
+
```
dgaliffiAMD: Reminder that this should be in the ROCm 10.1 section now.

projects/rocprofiler-systems/source/lib/core/constraint.cpp
```diff
@@ -4,343 +4,89 @@
...
-#include "logger/debug.hpp"
+#include <spdlog/fmt/fmt.h>
 
-#include <spdlog/fmt/ranges.h>
-
-#include <chrono>
 #include <cstdint>
-#include <ratio>
+#include <ctime>
+#include <stdexcept>
```
dgaliffiAMD:
```suggestion
#include <iterator>
#include <stdexcept>
```
clang-tidy: `make_move_iterator` is used, but `<iterator>` is not directly included.

However, could you for-go the `make_move_iterator` call and just use the `insert`? I know that it introduces a copy, but I don't think this is on the hot path, and it is just 3 primitives being copied. 

```
const auto _periods = parse_trace_periods(_periods_v, _delay_v, _duration_v);
        _v.insert(_v.end(), _periods.begin(), _periods.end());
```

projects/rocprofiler-systems/source/lib/core/constraint.hpp
```diff
@@ -4,101 +4,54 @@
...
+[[nodiscard]] std::vector<spec>
+parse_trace_periods(std::string_view periods, double default_delay,
+                    double default_duration);
 
-const std::set<clock_identifier>&
-get_valid_clock_ids();
+[[nodiscard]] clockid_t
+parse_trace_period_clock_id(std::string_view clock_id_str);
 
 std::vector<spec>
```
dgaliffiAMD:
```suggestion
[[nodiscard]] std::vector<spec>
```
We could add a `[[nodiscard]]` to this function too.

projects/rocprofiler-systems/source/lib/core/control/clocks/posix.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+#ifdef __linux__
+            const auto      next_ns = now().time_since_epoch().count() + this_chunk_ns;
+            struct timespec ts      = { static_cast<time_t>(next_ns / 1'000'000'000LL),
+                                        static_cast<long>(next_ns % 1'000'000'000LL) };
+            if(const auto rc = clock_nanosleep(m_clock_id, TIMER_ABSTIME, &ts, nullptr);
+               rc != 0 && rc != EINTR)
+            {
+                LOG_WARNING("clock_nanosleep failed for clock id {}: {}", m_clock_id,
+                            std::strerror(rc));
```
dgaliffiAMD:
```suggestion
                            std::system_category().message(rc));
```

We could use `std::system_category().message()` for additional thread safety.

projects/rocprofiler-systems/source/lib/core/control/clocks/posix.hpp
```diff
@@ -0,0 +1,100 @@
...
+    posix(posix&&)                 = delete;
+    posix& operator=(posix&&)      = delete;
+
+    [[nodiscard]] clock_time_point now() const noexcept
+    {
+        struct timespec ts = {};
+        if(clock_gettime(m_clock_id, &ts) != 0)
+        {
+            LOG_WARNING("clock_gettime failed for clock id {}: {}", m_clock_id,
+                        std::strerror(errno));
```
dgaliffiAMD:
```suggestion
                        std::system_category().message(errno));
```
We could use std::system_category().message() for additional thread safety.

projects/rocprofiler-systems/source/lib/core/control/clocks/posix.hpp
```diff
@@ -0,0 +1,100 @@
...
+        }
+        const auto ns =
+            static_cast<std::int64_t>(ts.tv_sec) * 1'000'000'000LL + ts.tv_nsec;
```
dgaliffiAMD: clang-tidy: 
We could avoid the "magic number" with a constant and a couple of helpers: 

```
static constexpr std::int64_t ns_per_sec = 1'000'000'000;

static std::int64_t to_ns(const struct timespec& ts) noexcept
{
    return (static_cast<std::int64_t>(ts.tv_sec) * ns_per_sec) + ts.tv_nsec;
}

static struct timespec to_timespec(std::int64_t ns) noexcept
{
    struct timespec ts = {};
    ts.tv_sec          = static_cast<time_t>(ns / ns_per_sec);
    ts.tv_nsec         = static_cast<long>(ns % ns_per_sec);
    return ts;
}
```

projects/rocprofiler-systems/source/lib/rocprof-sys/library/sampling.cpp
dgaliffiAMD:
```suggestion
```
clang-tidy: These includes are no longer used: `<chrono>` and `<condition_variable>`.

projects/rocprofiler-systems/source/lib/core/control/clocks/posix.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+namespace rocprofsys::control::clocks
+{
+/// Clock backed by a POSIX clock (clock_gettime / clock_nanosleep).
+class posix
+{
+public:
+    explicit posix(clockid_t clock_id = CLOCK_REALTIME) noexcept
+    : m_clock_id{ clock_id }
+    {}
```
dgaliffiAMD: Should we have some protection here to ensure `clock_id` is valid? 
Right now, it looks fine. I only see it being usind for `CLOCK_REALIME` (the default) and `CLOCK_PROCESS_CPUTIME_ID` (in `library.cpp`).
However, if one were to create a `posix` with an invalid `clockid_t`, the `now()` would return `0` and `sleep_until` would be in a tight busy loop until `interrupt()` is called and it outputs a warning log with each iteration.

projects/rocprofiler-systems/source/lib/core/control/clocks/posix.hpp
```diff
@@ -0,0 +1,100 @@
...
+        const auto ns =
+            static_cast<std::int64_t>(ts.tv_sec) * 1'000'000'000LL + ts.tv_nsec;
+        return clock_time_point{ clock_duration{ ns } };
+    }
+
+    [[nodiscard]] bool sleep_until(clock_time_point deadline)
+    {
+        static constexpr std::int64_t chunk_ns = 1'000'000;  // 1 ms
+
+        while(now() < deadline)
```
dgaliffiAMD: nit-pick: we are calling `now()` 3 times per iteration. 
We can probably just read it once at the top of each iteration.
For correctness: each call will drift slightly. For example, `remaining_ns` uses a different "now" than `next_ns` does.
For efficiency: we can save 2 of the 3 `clock_gettime` system calls.

projects/rocprofiler-systems/source/lib/core/control/clocks/posix.hpp
```diff
@@ -0,0 +1,100 @@
...
+#include <cstdint>
+#include <cstring>
+#include <ctime>
+#include <mutex>
+#include <thread>
+
+namespace rocprofsys::control::clocks
+{
+/// Clock backed by a POSIX clock (clock_gettime / clock_nanosleep).
+class posix
```
dgaliffiAMD, line 21: nit-pick: Maybe `posix_clock` would be a clearer class name.

projects/rocprofiler-systems/source/lib/rocprof-sys/library.cpp
```diff
@@ -1270,6 +1367,7 @@ rocprofsys_finalize_hidden(void)
         _output_registry.print_summary();
     }
 
+    stop_time_windows();
```
dgaliffiAMD: **Finalize ordering: time-window workers outlive the subsystems they notify**

  In `rocprofsys_finalize_hidden`, `stop_time_windows()` ran at the very end — after `rocprofiler_sdk::shutdown()`, `g_control_session.shutdown()`, etc. But a live `time_window` worker publishes actions via `session::set_action(...)`, which calls subscriber callbacks
  (`rocprofiler_sdk`, `sampling`, categories). So a window elapsing mid-finalize could fire a pause/resume into an already-torn-down subsystem. Benign today, but fragile.

  **Fix:** moved `stop_time_windows()` to the top of finalize (right after the "Finalizing..." log, before any teardown). It interrupts and joins the workers, so none can publish after that point. Idempotent, and the child-process `quick_exit` path is unaffected.

projects/rocprofiler-systems/source/lib/core/constraint.cpp
marantic-amd: I could, this is not on hot path, as you stated, so the copy won't be an expensive tradeoff in this case

projects/rocprofiler-systems/CHANGELOG.md
marantic-amd: Will move it there, I was focused on code changes and rebasing that  Iforgot to update the changelog.

projects/rocprofiler-systems/source/lib/core/control/clocks/posix.hpp
marantic-amd: I agree, I think that once the #7742 is merged we will have more unified way of actually handling stuff like this. But for now, sure, `constexpr` is still better that magic number, that is for sure

projects/rocprofiler-systems/source/lib/core/control/clocks/posix.hpp
marantic-amd: Yes, that could be a case. I will look into it, and see if it is a real issue. Anyway, I agree it makes sense to add that small check.

projects/rocprofiler-systems/source/lib/core/control/clocks/posix.hpp
marantic-amd: I will need to check the actual drift, but it makes sense

projects/rocprofiler-systems/source/lib/rocprof-sys/library.cpp
marantic-amd: I will need to check this with an exact ordering, because that has been the painpoint of the whole implementation to get the ordering right, since we are working with actually controlling the flow of the profiler execution. I will double check, once more to make sure everything is correctly ordered

projects/rocprofiler-systems/source/lib/core/control/clocks/posix.hpp
marantic-amd: Maybe, but we will eb duplicating names I think. Since the class itself is already in the namespace `rocprofsys::control::clocks` it's clear what `posix` refers to.

dgaliffiAMD:
It looks like the test failures are coming from the passing regex used for these tests. We'll need to update those tests to use another way to verify that the time-window and duration are correct.
See:  https://github.com/ROCm/rocm-systems/blob/6e9e95c500918202210861510bb7f379a7f86ba0/projects/rocprofiler-systems/tests/pytest/test_openmp.py#L166-L172.

marantic-amd:
> It looks like the test failures are coming from the passing regex used for these tests. We'll need to update those tests to use another way to verify that the time-window and duration are correct. See:
> 
> https://github.com/ROCm/rocm-systems/blob/6e9e95c500918202210861510bb7f379a7f86ba0/projects/rocprofiler-systems/tests/pytest/test_openmp.py#L166-L172
> 
> .

Yes, I already updated those, but during the rebase it seems that i have dropped those changes by mistake.

marantic-amd:
> > It looks like the test failures are coming from the passing regex used for these tests. We'll need to update those tests to use another way to verify that the time-window and duration are correct. See:
> > https://github.com/ROCm/rocm-systems/blob/6e9e95c500918202210861510bb7f379a7f86ba0/projects/rocprofiler-systems/tests/pytest/test_openmp.py#L166-L172
> > 
> > .
> 
> Yes, I already updated those, but during the rebase it seems that i have dropped those changes by mistake. On the original PR I dropped the regex checks and actually checked against produced traces. I will replicate that logic here, since then we will actually be testing the sampling duration is properly set

projects/rocprofiler-systems/source/lib/core/constraint.hpp
```diff
@@ -4,101 +4,54 @@
...
+parse_trace_period_clock_id(std::string_view clock_id_str);
 
-std::vector<spec>
+[[nodiscard]] std::vector<spec>
 get_trace_specs();
 
-stages
-get_trace_stages();
+[[nodiscard]] clockid_t
+get_trace_period_clock_id();
```
mradosav-amd: make class of of this. even with static methods, we can mock class, we can't mock free functions.

As well, I see that `get_trace_specs` is only needed to be public method. Other methods are there, used internally by `get_trace_specs`. In that way we can encapsulate.

We can make something like:

```cpp

class trace_config_user_input_getter
{
double get_trace_delay() const { return config::get_trace_delay(); }
/// .... 
};

template <typename ConfigGetter>
class trace_config
{
public: 
[[nodiscard]] static std::vector<spec> get_trace_specs();

private:
/// ....
};

```


so we can use it like:


```cpp
// in production:
const auto spec = trace_config<trace_config_user_input_getter>::get_trace_specs();

// in unit test:
const auto spec = trace_config<MOCKED_trace_config_user_input_getter>::get_trace_specs();

```

projects/rocprofiler-systems/source/lib/core/constraint.hpp
```diff
@@ -4,101 +4,54 @@
...
-    bool operator==(int _rhs) const;
-    bool operator==(std::string _rhs) const;
-
-    friend std::ostream& operator<<(std::ostream& _os, const clock_identifier& _v)
-    {
-        return (_os << _v.as_string());
-    }
-};
-
 struct spec
```
mradosav-amd: maybe to find better name, `spec` is short, don't mean anything.

projects/rocprofiler-systems/source/lib/core/constraint.cpp
```diff
@@ -4,351 +4,88 @@
...
+                    env_vars::TRACE_PERIOD_CLOCK_ID, clock_id_str));
 }
 
 std::vector<spec>
 get_trace_specs()
 {
     auto _v = std::vector<constraint::spec>{};
 
+    const auto _delay_v =
+        config::get_setting_value<double>(std::string{ env_vars::TRACE_DELAY })
```
mradosav-amd: let's not read it via `config::get_setting_value` , add method in config `get_trace_delay`, `get_trace_duration`, `get_trace_periods` to have uniform way of reading config. It will be easier for us to refactor it later.

projects/rocprofiler-systems/source/lib/core/control/session.hpp
```diff
@@ -45,34 +57,49 @@ class session
...
-    std::unordered_map<std::string, action> m_actions;
-    std::vector<subscriber>                 m_subscribers;
-    std::atomic<bool>                       m_active{ true };
+    static constexpr std::size_t scope_count = static_cast<std::size_t>(scope::count_);
+
+    using scoped_actions = std::unordered_map<std::string, action>;
+
+    std::array<scoped_actions, scope_count>    m_actions;
+    std::vector<subscriber>                    m_subscribers;
+    std::array<std::atomic<bool>, scope_count> m_active{};
```
mradosav-amd: find better name than `m_active` when reading at code, it's not telling me for what is used for.
and we can use different data structure for this

projects/rocprofiler-systems/source/lib/core/control/session.hpp
```diff
@@ -22,17 +25,26 @@ enum class action
     pause
 };
 
+/// A trigger's blast radius: which subscribers its actions can reach.
+enum class scope : std::size_t
+{
+    global = 0,
+    sampling_only,
```
mradosav-amd:
```suggestion
    sampling, 
```

no need for _only

projects/rocprofiler-systems/source/lib/core/control/session.hpp
```diff
@@ -22,17 +25,26 @@ enum class action
...
+    count_,  // sentinel: number of scopes
+};
+
 struct subscriber
 {
     std::function<void()> on_pause;
     std::function<void()> on_resume;
     std::string           name;
+    std::vector<scope>    scopes = { scope::global };
```
mradosav-amd: why do we need vector of scopes?

projects/rocprofiler-systems/source/lib/core/constraint.hpp
marantic-amd: I looked at the pattern used in PR #9637 and will try to follow it up here also. makes sense to do so, it will be clearer and will be unified with the approach in mentioned PR

projects/rocprofiler-systems/source/lib/core/constraint.hpp
marantic-amd: I agree, `spec` is legacy name, I honestly didn't bother to change. I think that `trace_period` makes sense, since that is exactly what it represent.

projects/rocprofiler-systems/source/lib/core/control/session.hpp
marantic-amd: Perhaps `m_scope_tracing` would be more appropriate naming for this one, I agree that the current name can be vague? Regarding the underlying data structure, I think that this is at the moment the "cheapest" option, paired with `memory_order_relaxed`, or am I missing something?

projects/rocprofiler-systems/source/lib/core/control/session.hpp
marantic-amd: to be quite honest, we don't. Looking at the end consumers of this, maybe the better option is to use a `std::bitset`

projects/rocprofiler-systems/source/lib/core/constraint.hpp
```diff
@@ -4,101 +4,154 @@
...
-    void operator()(const stages&) const;
-
-    double           delay    = 0.0;
-    double           duration = 0.0;
-    std::uint64_t    count    = 0;
-    std::uint64_t    repeat   = 1;
-    clock_identifier clock_id = {};
-};
+    const auto   _parts = rocprofsys::delimit(std::string{ entry }, ":");
+    trace_period _s{ default_delay, default_duration, 1 };
```
mradosav-amd, line 93: better naming than _s

projects/rocprofiler-systems/source/lib/core/constraint.hpp
```diff
@@ -4,101 +4,154 @@
...
-
-    void operator()(const stages&) const;
-
-    double           delay    = 0.0;
-    double           duration = 0.0;
-    std::uint64_t    count    = 0;
-    std::uint64_t    repeat   = 1;
-    clock_identifier clock_id = {};
-};
+    const auto   _parts = rocprofsys::delimit(std::string{ entry }, ":");
```
mradosav-amd, line 92: better naming than _parts, maybe parsed_entry or delimited_entry

projects/rocprofiler-systems/source/lib/core/constraint.hpp
```diff
@@ -4,101 +4,154 @@
...
+    trace_period _s{ default_delay, default_duration, 1 };
+    if(!_parts.empty()) _s.delay = utility::convert<double>(_parts.at(0));
+    if(_parts.size() > 1) _s.duration = utility::convert<double>(_parts.at(1));
+    if(_parts.size() > 2) _s.repeat = utility::convert<std::uint64_t>(_parts.at(2));
```
mradosav-amd, line 96: can we do something like

```cpp

constexpr std::uint64_t default_repeat = 1;

auto delay    = default_delay;
auto duration = default_duration;
auto repeat   = default_repeat;

if (parts.size() > 3) {
    // Log warning: extra fields are ignored.
}

switch (std::min<std::size_t>(parts.size(), 3)) {
case 3:
    repeat = utility::convert<std::uint64_t>(parts[2]);
    [[fallthrough]];

case 2:
    duration = utility::convert<double>(parts[1]);
    [[fallthrough]];

case 1:
    delay = utility::convert<double>(parts[0]);
    [[fallthrough]];

case 0:
    break;
}

return {delay, duration, repeat};
```

projects/rocprofiler-systems/source/lib/core/constraint.hpp
```diff
@@ -4,101 +4,154 @@
...
+    return _s;
+}
+
+template <typename Externals>
+    requires concepts::trace_config_externals<Externals>
+std::vector<trace_period>
+trace_config<Externals>::parse_periods(std::string_view periods, double default_delay,
+                                       double default_duration)
+{
+    auto _v = std::vector<trace_period>{};
```
mradosav-amd, line 106: we can reserve vector, and do std::transform

projects/rocprofiler-systems/source/lib/core/constraint.hpp
```diff
@@ -4,101 +4,154 @@
...
+
+    return _v;
+}
+
+template <typename Externals>
+    requires concepts::trace_config_externals<Externals>
+clockid_t
+trace_config<Externals>::get_trace_period_clock_id()
+{
+    return parse_period_clock_id(Externals::get_trace_period_clock());
```
mradosav-amd, line 154: no need for `parse_period_clock_id` since it's not exposed as public method

projects/rocprofiler-systems/source/lib/core/constraint.hpp
```diff
@@ -4,101 +4,154 @@
...
+    if(clock_id_str == "realtime") return CLOCK_REALTIME;
+    throw std::runtime_error(
+        fmt::format("Unknown {}: '{}'. Valid choices: realtime, cputime",
+                    env_vars::TRACE_PERIOD_CLOCK_ID, clock_id_str));
+}
+
+template <typename Externals>
+    requires concepts::trace_config_externals<Externals>
+std::vector<trace_period>
+trace_config<Externals>::get_trace_specs()
```
mradosav-amd, line 127: so, we provide trace_delay, trace_duration and we have trace_periods. why do we respect both?

I find it weird to enable both at same time, ether use trace_delay and trace_duration OR use trace_periods. 

If we have:
`trace_delay 5` and `trace_duration 10`

and if we have `trace_period 0:5:1`

so we got 0-5;5-10. Ok we will be paused from 0 to 10, but I find it wierd.

projects/rocprofiler-systems/source/lib/core/control/clocks/posix.hpp
```diff
@@ -0,0 +1,123 @@
...
+            const struct timespec ts = to_timespec(next_ns);
+            if(const auto rc = clock_nanosleep(m_clock_id, TIMER_ABSTIME, &ts, nullptr);
+               rc != 0 && rc != EINTR)
+            {
+                LOG_WARNING("clock_nanosleep failed for clock id {}: {}", m_clock_id,
+                            std::system_category().message(rc));
+            }
+#else
+            std::this_thread::sleep_for(std::chrono::nanoseconds{ this_chunk_ns });
+#endif
```
mradosav-amd, line 83: will `std::this_thread::sleep_for` work for linux as well?
as well, this is posix, do we really need `#ifdef __linux__`

projects/rocprofiler-systems/source/lib/core/control/clocks/posix.hpp
```diff
@@ -0,0 +1,123 @@
...
+    {
+        struct timespec ts = {};
+        ts.tv_sec          = static_cast<time_t>(ns / nsec_per_sec);
+        ts.tv_nsec         = static_cast<long>(ns % nsec_per_sec);
+        return ts;
+    }
+
+    clockid_t  m_clock_id;
+    bool       m_usable = true;
+    std::mutex m_mutex;
```
mradosav-amd, line 120: for any usage of mutex, mark it as thread::state::internal. this will make deadlock when running with `rocprof-sys-run -I all`

projects/rocprofiler-systems/source/lib/core/constraint.hpp
```diff
@@ -4,101 +4,154 @@
...
+template <typename Externals>
+    requires concepts::trace_config_externals<Externals>
+clockid_t
+trace_config<Externals>::parse_period_clock_id(std::string_view clock_id_str)
+{
+    if(clock_id_str == "cputime") return CLOCK_PROCESS_CPUTIME_ID;
+    if(clock_id_str == "realtime") return CLOCK_REALTIME;
+    throw std::runtime_error(
+        fmt::format("Unknown {}: '{}'. Valid choices: realtime, cputime",
+                    env_vars::TRACE_PERIOD_CLOCK_ID, clock_id_str));
```
mradosav-amd, line 121:
```suggestion
    clockid_t result;
    if(clock_id_str == "cputime") 
    {
        result = CLOCK_PROCESS_CPUTIME_ID; 
    }
    else if(clock_id_str == "realtime") 
    {
        result = CLOCK_REALTIME; 
    }
    else {
        throw std::runtime_error(
        fmt::format("Unknown {}: '{}'. Valid choices: realtime, cputime",
                    env_vars::TRACE_PERIOD_CLOCK_ID, clock_id_str));
    }        
             
    return result;    
```

maybe easier to follow flow. single return point with result. and if we don't match, we throw.

projects/rocprofiler-systems/source/lib/core/constraint.hpp
```diff
@@ -4,101 +4,154 @@
...
-    clock_identifier();
-    clock_identifier(std::string_view, int);
+/// Translates the trace-window settings supplied by @p Externals into the
+/// value types the control plane consumes.
+template <typename Externals>
+    requires concepts::trace_config_externals<Externals>
+class trace_config
+{
+public:
+    [[nodiscard]] static std::vector<trace_period> get_trace_specs();
```
mradosav-amd, line 70: After reviewing this file, I find that we should move this whole implementation to `config.hpp`.
From config, we are just reading plain env variable values, instead of actually getting config.

Other comments are still applicable even if we move it there.

projects/rocprofiler-systems/source/lib/core/constraint.hpp
```diff
@@ -4,101 +4,154 @@
...
-    double           delay    = 0.0;
-    double           duration = 0.0;
-    std::uint64_t    count    = 0;
-    std::uint64_t    repeat   = 1;
-    clock_identifier clock_id = {};
-};
+    const auto   _parts = rocprofsys::delimit(std::string{ entry }, ":");
+    trace_period _s{ default_delay, default_duration, 1 };
+    if(!_parts.empty()) _s.delay = utility::convert<double>(_parts.at(0));
+    if(_parts.size() > 1) _s.duration = utility::convert<double>(_parts.at(1));
```
mradosav-amd, line 95: duration and delay can be negative numbers?

projects/rocprofiler-systems/source/lib/core/constraint.hpp
marantic-amd: Maybe. I was aiming for early return. But this is not on hot-path so, Maybe a readability improvement has precedence.

projects/rocprofiler-systems/source/lib/core/constraint.hpp
marantic-amd: Discussed offline. This will be clarified better in code, and made more clear.

projects/rocprofiler-systems/source/lib/core/control/clocks/posix.hpp
marantic-amd: Discussed offline. The only reason was minor beneficial gain, but it's mostly irrelevant in our use case

projects/rocprofiler-systems/source/lib/core/constraint.hpp
marantic-amd: YOu are right, it's missing the basic check. I will add it to the PR

projects/rocprofiler-systems/source/lib/core/constraint.hpp
marantic-amd: I think it maybe make more sense to have it in it's own file, something like `trace_period_config.hpp` which will extend the `config` namespace instead of putting all of this in the `config.hpp`, since that header is already huge as it is. This way we make the clear connection, while over-cluttering the existing code

projects/rocprofiler-systems/source/lib/core/control/session.cpp
```diff
@@ -36,80 +52,124 @@ session::subscribe(subscriber sub)
...
 void
 session::force_initial_pause()
 {
-    if(is_active())
+    const std::scoped_lock subs_lk{ m_subscribers_mutex };
+    for(const auto& sub : m_subscribers)
     {
-        return;
+        const bool any_paused_for_sub =
+            sub.scopes.any_of([this](scope listened) { return !is_active(listened); });
```
mradosav-amd: `any_of([](scope s) { return !is_active(s); })` is exactly `!all_of([](scope s) { return is_active(s); })` 
with that we can simplify scope_set?

PR #9078, opened 2026-07-23
fix(workflows): resolve phantom skips, add Debian system-deps

marantic-amd:
## Motivation

- The CI refactor left non-existing workflows shown as skipped due to the resolution logic in the build matrix. This PR resolves that and adds a `system-deps` run for Debian workflows.
- Within this PR is included also:
    -  the fix of broken CDash link
    - Missing build log on successful build

## Technical Details

- `rhel-compiler-check` relocated from the shared `build-group.yml` (where it appeared as a confusing phantom `${{ matrix.gcc-toolset }}`-named skip under Ubuntu 22.04/24.04/Debian) into the orchestrator `build.yml` as its own unconditional top-level job.
-  Debian system-deps coverage added with two new `ci-build-matrix.json` entries (`g++` and `amdclang++`, ROCm 7.2, matching the Ubuntu/RHEL system-deps pattern), carrying over Debian primary's specific cmake flags.
- Fix the broken CDash link generator in `run-ci.py` so that now the link points to the correct build.
- Include the missing build log on successful build in GitHub Actions report.
- Add build log into corresponding CDash entry.

## Issue Tracking

JIRA ID: AIPROFSYST-391
JIRA ID: AIPROFSYST-667
JIRA ID: AIPROFSYST-668

## Submission Checklist


projects/rocprofiler-systems/.github/ci-build-matrix.json
```diff
@@ -512,6 +512,38 @@
...
       "group": "ubuntu-24.04"
     },
+    {
+      "name": "Debian 12 \u2022 g++ \u2022 system deps",
+      "image": "dgaliffiamd/rocprofiler-systems:ci-rocm-7.2-debian-12",
+      "distro": "debian-system-deps",
+      "compiler": "g++",
+      "cc": "gcc",
+      "use_clang_omp": "OFF",
+      "cdash_suffix": "debian-system-deps-g++",
```
dgaliffiAMD:
```suggestion
      "cdash_suffix": "debian-12-system-deps-g++",
```

Adding os-version, `12`, to the label. This is just a nit-pick, right now, but if/when we add a new version of debian, then it will be more seemless.

projects/rocprofiler-systems/.github/ci-build-matrix.json
```diff
@@ -512,6 +512,38 @@
...
+      "group": "debian"
+    },
+    {
+      "name": "Debian 12 \u2022 amdclang++ \u2022 system deps",
+      "image": "dgaliffiamd/rocprofiler-systems:ci-rocm-7.2-debian-12",
+      "distro": "debian-system-deps",
+      "compiler": "amdclang++",
+      "cc": "amdclang",
+      "use_clang_omp": "ON",
+      "cdash_suffix": "debian-system-deps-amdclang++",
```
dgaliffiAMD:
```suggestion
      "cdash_suffix": "debian-12-system-deps-amdclang++",
```
Adding os-version, `12`, to the label. This is just a nit-pick, right now, but if/when we add a new version of debian, then it will be more seemless.

projects/rocprofiler-systems/scripts/run-ci.py
```diff
@@ -662,27 +671,63 @@ def _read_build_xml_failures(args):
...
+
+
+def _cdash_link_url(name: str) -> str:
+    query = urllib.parse.urlencode(
+        {
+            "project": "rocprofiler-systems",
+            "filtercount": "1",
+            "showfilters": "1",
+            "field1": "buildname",
+            "compare1": "65",
```
dgaliffiAMD:
```suggestion
            "compare1": "65", # CDash filter op "65" == "starts with"
```
nit-pick: maybe a comment explaining what "65" is. "starts with", I think? This was not introduced by your PR, but since you're making a nice new function, we might as well make it less mystical.

dgaliffiAMD (approved):

dgaliffiAMD:
This should be safe to merge once the workflows finish. It is a CI-only change.

jbonnell-amd (approved):

dgaliffiAMD:
Test failures are unrelated. This is a local CI change only and should be safe to merge now.

PR #9079, opened 2026-07-23
ci(rocjitsu): enable msan coverage

lialan:
## Motivation

  RocJitsu supports MemorySanitizer through `RJ_ENABLE_MSAN`, but the configuration
  is not currently exercised by CI.

  Add blocking compile-and-link coverage while keeping runtime tests disabled until
  the required dependencies can be fully MSan-instrumented.

  ## Technical Details

  - Add a blocking Clang `msan-auto` CI configuration.
  - Build and link `early_mmap_driver` with MSan using the supported static
  runtime.
  - Verify that `RJ_SANITIZER_RUNTIME=SHARED` is rejected for MSan.
  - Keep existing sanitizer configurations and runtime tests unchanged.
  - Build the host `flatc` tool without MSan instrumentation because it executes
  against uninstrumented host libraries during schema generation.

  This is a follow-up to #8589.

  ## Issue Tracking

  GitHub issue: https://github.com/ROCm/rocm-systems/issues/7891

  ## Test Plan

  - Configure RocJitsu with Clang and `RJ_ENABLE_MSAN=1`.
  - Build and run `early_mmap_driver`.
  - Generate the FlatBuffers schemas using the uninstrumented host `flatc`.
  - Confirm that shared MSan configuration fails with the expected diagnostic.
  - Run pre-commit over the complete branch diff.

  ## Test Result

  - Clang 14 MSan configure, compile, link, and execution passed.
  - FlatBuffers schema generation passed.
  - Shared MSan rejection check passed.
  - Full `origin/develop..HEAD` pre-commit sweep passed.

  Runtime CTests remain disabled for this configuration because MSan requires fully
  instrumented dependencies.

  ## Submission Checklist

### DIFF
--- a/.github/workflows/rocjitsu-corpus-tests.yml
+++ b/.github/workflows/rocjitsu-corpus-tests.yml
@@ -62,8 +62,10 @@ jobs:
             enable_asan: 0
             enable_ubsan: 0
             enable_tsan: 0
+            enable_msan: 0
             enable_expensive_checks: 0
             sanitizer_runtime: STATIC
+            run_tests: 1
           - name: asan-ubsan
             allow_failure: true
             test_regex: ''
@@ -71,8 +73,10 @@ jobs:
             enable_asan: 1
             enable_ubsan: 1
             enable_tsan: 0
+            enable_msan: 0
             enable_expensive_checks: 1
             sanitizer_runtime: SHARED
+            run_tests: 1
           - name: tsan
             allow_failure: true
             test_regex: ''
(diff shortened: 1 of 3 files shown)


kuhar (changes requested):
Discussed this offline: I don't think we have enough evidence or intuition to grow the test matrix with msan and take on the additional complexity / maintenance burden. If we identify another build type we want to test in the future, we may then bundle msan with it.

It's also good signal that we tried enabling msan and it didn't find any issues.

PR #9081, opened 2026-07-23
test(rocprofiler-systems): Add general CLI flag regression tests

anujshuk-amd:
## Motivation
Added end-to-end regression tests for the CLI path shared by rocprof-sys-run and rocprof-sys-sample. Help text tests don't prove flags work; these run real workloads, check metadata.json and outputs, and pin behaviors that broke or shifted during tool_runner unification (env replace, profile conflicts, --output-format, -c/-o, trace/profile/host/device flags, -w/-d, --periods, and -D). Runs through rocprof-sys-run only; malformed -c validation lives in test_config.py.

## Technical Details

Adds test_tool_runner.py (tool_runner marker registered in conftest.py) with end-to-end tests through rocprof-sys-run, which covers the shared tool_runner path used by rocprof-sys-sample too. Tests run parallel-overhead or transpose, check resolved settings in metadata.json, and verify outputs (Perfetto, rocpd,sampling profile) where it matters. Coverage includes preset env replace, --profile/--flat-profile conflicts, --output-format backend selection, -c/-o paths, -T/-L/-P/-F/-H/-D flag mapping, -w/-d, --periods, and -D GPU sampling. Adds test_config.py::test_cli_flag_rejects_invalid so malformed configs passed via -c are rejected alongside the existing env-var config tests.

## Issue Tracking

JIRA ID : AIROCVAL-110

## Test Plan

Ran newly written tests and old tests.

## Test Result
1. **To run : cd to build then runctest -R 'config-cli-flag-rejects-invalid|tool-run'
ctest -R 'config-cli-flag-rejects-invalid|tool-run'
Test project /home/rocm/AnujWork/AIPROSYST485/rocm-systems/projects/rocprofiler-systems/build_deadlock
      Start   1: rocprofiler-systems-pytest-config
 1/23 Test   1: rocprofiler-systems-pytest-config ...........................................................   Passed    1.35 sec
      Start 168: config-cli-flag-rejects-invalid
 2/23 Test 168: config-cli-flag-rejects-invalid .............................................................   Passed    1.77 sec
      Start 466: tool-runner-replace-env-preset-overrides-shell-value-run
 3/23 Test 466: tool-runner-replace-env-preset-overrides-shell-value-run ....................................   Passed    3.30 sec
      Start 467: tool-runner-profile-conflict-profile-and-flat-profile-rejected-run
 4/23 Test 467: tool-runner-profile-conflict-profile-and-flat-profile-rejected-run ..........................   Passed    0.40 sec
      Start 468: tool-runner-output-format-proto-rocpd-enables-both-run
 5/23 Test 468: tool-runner-output-format-proto-rocpd-enables-both-run ......................................   Passed    7.89 sec
      Start 469: tool-runner-output-format-rocpd-only-disables-perfetto-run
 6/23 Test 469: tool-runner-output-format-rocpd-only-disables-perfetto-run ..................................   Passed    4.66 sec
      Start 470: tool-runner-output-format-conflicts-with-old-style-flags-run-trace
 7/23 Test 470: tool-runner-output-format-conflicts-with-old-style-flags-run-trace ..........................   Passed    0.40 sec
      Start 471: tool-runner-output-format-conflicts-with-old-style-flags-run-profile
 8/23 Test 471: tool-runner-output-format-conflicts-with-old-style-flags-run-profile ........................   Passed    0.41 sec
      Start 472: tool-runner-output-format-conflicts-with-old-style-flags-run-flat-profile
 9/23 Test 472: tool-runner-output-format-conflicts-with-old-style-flags-run-flat-profile ...................   Passed    0.40 sec
      Start 473: tool-runner-output-format-conflicts-with-old-style-flags-run-profile-format
10/23 Test 473: tool-runner-output-format-conflicts-with-old-style-flags-run-profile-format .................   Passed    0.40 sec
      Start 474: tool-runner-config-output-values-reach-env-and-metadata-file-run
11/23 Test 474: tool-runner-config-output-values-reach-env-and-metadata-file-run ............................   Passed    5.77 sec
      Start 475: tool-runner-cli-flag-env-mapping-sets-expected-settings-and-artifacts-trace-run
12/23 Test 475: tool-runner-cli-flag-env-mapping-sets-expected-settings-and-artifacts-trace-run .............   Passed    7.69 sec
      Start 476: tool-runner-cli-flag-env-mapping-sets-expected-settings-and-artifacts-profile-run
13/23 Test 476: tool-runner-cli-flag-env-mapping-sets-expected-settings-and-artifacts-profile-run ...........   Passed    6.32 sec
      Start 477: tool-runner-cli-flag-env-mapping-sets-expected-settings-and-artifacts-flat-profile-run
14/23 Test 477: tool-runner-cli-flag-env-mapping-sets-expected-settings-and-artifacts-flat-profile-run ......   Passed    6.12 sec
      Start 478: tool-runner-cli-flag-env-mapping-sets-expected-settings-and-artifacts-host-only-run
15/23 Test 478: tool-runner-cli-flag-env-mapping-sets-expected-settings-and-artifacts-host-only-run .........   Passed    6.25 sec
      Start 479: tool-runner-cli-flag-env-mapping-sets-expected-settings-and-artifacts-device-only-run
16/23 Test 479: tool-runner-cli-flag-env-mapping-sets-expected-settings-and-artifacts-device-only-run .......   Passed    6.30 sec
      Start 480: tool-runner-cli-flag-env-mapping-sets-expected-settings-and-artifacts-host-and-device-run
17/23 Test 480: tool-runner-cli-flag-env-mapping-sets-expected-settings-and-artifacts-host-and-device-run ...   Passed    6.29 sec
      Start 481: tool-runner-wait-duration-wait-sets-delay-envs-run
18/23 Test 481: tool-runner-wait-duration-wait-sets-delay-envs-run ..........................................   Passed    4.14 sec
      Start 482: tool-runner-wait-duration-duration-sets-duration-envs-run
19/23 Test 482: tool-runner-wait-duration-duration-sets-duration-envs-run ...................................   Passed    5.73 sec
      Start 483: tool-runner-trace-periods-single-period-sets-env-run
20/23 Test 483: tool-runner-trace-periods-single-period-sets-env-run ........................................   Passed    5.76 sec
      Start 484: tool-runner-trace-periods-repeated-periods-are-space-joined-run
21/23 Test 484: tool-runner-trace-periods-repeated-periods-are-space-joined-run .............................   Passed    5.73 sec
      Start 485: tool-runner-device-sampling-records-gpu-counters-run
22/23 Test 485: tool-runner-device-sampling-records-gpu-counters-run ........................................   Passed    6.51 sec
      Start 557: rocprofiler-systems-test-cleanup
23/23 Test 557: rocprofiler-systems-test-cleanup ............................................................   Passed    0.17 sec

100% tests passed, 0 tests failed out of 23

Label Time Summary:
all               =  92.24 sec*proc (21 tests)
cleanup           =   0.17 sec*proc (1 test)
comprehensive     =  92.24 sec*proc (21 tests)
extended          =  92.24 sec*proc (21 tests)
full              =  92.24 sec*proc (21 tests)
global            =   1.51 sec*proc (2 tests)
gpu               =  31.65 sec*proc (5 tests)
nightly           =  92.24 sec*proc (21 tests)
pr                =  92.24 sec*proc (21 tests)
precheckin        =  92.24 sec*proc (21 tests)
prerequisite      =   1.35 sec*proc (1 test)
rocprof_config    =   1.77 sec*proc (1 test)
standard          =  92.24 sec*proc (21 tests)
sys_run           =  90.48 sec*proc (20 tests)
tool_runner       =  90.48 sec*proc (20 tests)

Total Test time (real) =  93.79 sec
## Submission Checklist


projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -147,3 +152,209 @@ def test_conflicts_with_legacy_flags(self, target, legacy_args):
...
+# =============================================================================
+# Config file + output path/prefix
+# =============================================================================
+
+
+@pytest.mark.timeout(30)
+@pytest.mark.class_name("tool-runner-config-output")
+@pytest.mark.parametrize("target", TARGETS)
+class TestConfigOutput(RocprofsysTest):
+    """-c/--config and -o/--output actually load from and write to the given paths."""
```
marantic-amd, line 285: This test doesn't actually validate what is stated here as far as I see it. And I think that the base environment already sets up the output directory which setting it again here seem redundant

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -147,3 +152,209 @@ def test_conflicts_with_legacy_flags(self, target, legacy_args):
...
+
+    def test(self, target, test_output_dir):
```
marantic-amd: Just a bare bone `test` name actually breaks the pattern we established when defining the test cases in out codebase

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -147,3 +152,209 @@ def test_conflicts_with_legacy_flags(self, target, legacy_args):
...
+
+
+@pytest.mark.timeout(30)
+@pytest.mark.class_name("tool-runner-trace-profile-flags")
+@pytest.mark.parametrize("target", TARGETS)
+class TestTraceProfileFlags(RocprofsysTest):
+    """-T/-L/-P/-F individually set their corresponding env vars."""
+
+    @pytest.mark.parametrize("flag_args, pass_regex", TRACE_PROFILE_FLAG_CASES)
+    def test(self, target, flag_args, pass_regex):
```
marantic-amd: Same as above. Why bare `test` naming?

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -147,3 +152,209 @@ def test_conflicts_with_legacy_flags(self, target, legacy_args):
...
+@pytest.mark.timeout(30)
+@pytest.mark.class_name("tool-runner-host-device-flags")
+@pytest.mark.parametrize("target", TARGETS)
+class TestHostDeviceFlags(RocprofsysTest):
+    """-H and -D cross-set process-sampling env vars based on argparse.cpp's
+    host/device interaction logic (each reads the other's current state).
+    """
+
+    @pytest.mark.parametrize("flag_args, pass_regex", HOST_DEVICE_FLAG_CASES)
+    def test(self, target, flag_args, pass_regex):
```
marantic-amd: Same as with previous namings

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -147,3 +152,209 @@ def test_conflicts_with_legacy_flags(self, target, legacy_args):
...
+        ],
+        id="host_and_device",
+    ),
+]
+
+
+@pytest.mark.timeout(30)
+@pytest.mark.class_name("tool-runner-host-device-flags")
+@pytest.mark.parametrize("target", TARGETS)
+class TestHostDeviceFlags(RocprofsysTest):
```
marantic-amd: Looks like this class is almost, if not fully, byte-identical to the `TestTraceProfileFlags`

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -147,3 +152,209 @@ def test_conflicts_with_legacy_flags(self, target, legacy_args):
...
+            run_args=["--periods", "0:2", "-v", "2", "--", "ls"],
+            fail_on_not_found=True,
+        )
+        self.assert_regex(result, pass_regex=[r"ROCPROFSYS_TRACE_PERIODS=0:2"])
+
+    def test_repeated_periods_are_space_joined(self, target):
+        result = self.run_test(
+            "baseline",
+            target=target,
+            run_args=["--periods", "0:2", "--periods", "3:2", "-v", "2", "--", "ls"],
```
marantic-amd: `"-v", "2", "--", "ls"` arguments are duplicated 6 times here. TO me it makes sense to extract them to some common pre-defined variable

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
marantic-amd, line 285: If we are already adding the test, we could also add a check when given the mellformed config

projects/rocprofiler-systems/tests/pytest/test_selective_regions.py
```diff
@@ -152,21 +155,40 @@ def test_no_filter(self, mode, marker_style, selective_region_env):
             pass_regex=["Region1", "Region2", "Region3"],
         )
 
-    def test_region_1_filter(self, mode, marker_style, selective_region_env):
+    @pytest.mark.parametrize("region_source", ["env", "cli"])
```
marantic-amd: Seems to me that the "cli" flag is duplicating the whole test suite, instead of just maybe running one test to prove the point that the flag is properly handled, instead of duplicating all tests

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: Good catch. fixed : the test now runs with -v 2 and asserts the echoed ROCPROFSYS_CONFIG_FILE/OUTPUT_PATH/OUTPUT_PREFIX env vars (plus the resolved values in metadata.json), and -o now points at a custom_output subdir instead of test_output_dir.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: Fixed. Renamed to a descriptive name (test_values_reach_env_and_metadata_file / test_sets_expected_settings_and_artifacts) so it follows the naming convention used elsewhere in the file.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: renamed.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: Good point. Merged TestTraceProfileFlags and TestHostDeviceFlags into one TestCliFlagEnvMapping class parametrized over CLI_FLAG_ENV_CASES, so the duplicated structure is gone.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: Taken in to comon var.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: Done. Added test_malformed_file_is_rejected, which feeds a config with a bad ROCPROFSYS_TRACE_DURATION through -c and asserts it's rejected with the expected "invalid value" error on both binaries.

projects/rocprofiler-systems/tests/pytest/test_selective_regions.py
anujshuk-amd: Fixed. Reverted test_region_1_filter back to env-only and added a separate minimal TestSelectiveRegionCliFlag class with just one case per binary (not the full mode/marker_style matrix) to prove --selected-regions is wired up

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -108,12 +154,19 @@ def test_proto_rocpd_enables_both(self, target):
             ],
         )
 
+        settings = _resolved_settings(result)
```
kcossett-amd, line 206: `get_base_environment()` is auto injected into the test during `run_test` unless `no_base_env=True` is specified. 
That means that some of these checks will always be true
(e.g. `ROCPROFSYS_TRACE`)

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -147,3 +210,288 @@ def test_conflicts_with_legacy_flags(self, target, legacy_args):
...
+        None,
+        id="host_and_device",
+    ),
+]
+
+
+@pytest.mark.timeout(30)
+@pytest.mark.class_name("tool-runner-cli-flag-env-mapping")
+@pytest.mark.parametrize("target", TARGETS)
+class TestCliFlagEnvMapping(RocprofsysTest):
```
kcossett-amd, line 387: Again, recall `get_base_environment()` auto injects env vars unless `no_base_env=True`.
Please check the rest of your code.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -19,6 +26,35 @@
     pytest.param("rocprof-sys-sample", marks=pytest.mark.sampling, id="sample"),
 ]
 
+# Shared tail for the CLI-flag tests below: "-v 2" so the tool echoes its
+# resolved env, and "-- ls" as a minimal child process to instrument.
+VERBOSE_LS_ARGS = ["-v", "2", "--", "ls"]
```
kcossett-amd: All your tests pass if I remove `-v 2`...

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -147,3 +210,288 @@ def test_conflicts_with_legacy_flags(self, target, legacy_args):
...
+    "ls" doesn't work here — the runtime never gets far enough into init to
+    validate the config and reject a bad setting (same reason test_config.py's
+    config_target fixture avoids it). Falls back to "ls" if parallel-overhead
+    isn't built, at which point the malformed-config test just won't catch
+    anything meaningful.
+    """
+    try:
+        return str(rocprof_config.get_target_executable("parallel-overhead"))
+    except FileNotFoundError:
+        return shutil.which("ls") or "ls"
```
kcossett-amd: You claim `ls` doesn't work here, then use it as a fallback...?

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -147,3 +210,288 @@ def test_conflicts_with_legacy_flags(self, target, legacy_args):
...
+@pytest.mark.class_name("tool-runner-config-output")
+@pytest.mark.parametrize("target", TARGETS)
+class TestConfigOutput(RocprofsysTest):
+    """-c/--config and -o/--output actually load from and write to the given paths."""
+
+    def test_values_reach_env_and_metadata_file(self, target, test_output_dir):
+        config_dir = test_output_dir / "config"
+        config_dir.mkdir(parents=True, exist_ok=True)
+        empty_cfg = config_dir / "empty.cfg"
+        empty_cfg.write_text("#\n# empty config file\n#\n")
```
kcossett-amd: `create_config_file` fixture exists.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -147,3 +210,288 @@ def test_conflicts_with_legacy_flags(self, target, legacy_args):
             result,
             pass_regex=[r"--output-format.*conflicts.*"],
         )
+
+
+# =============================================================================
+# Config file + output path/prefix
```
kcossett-amd, line 277: `rocprof-sys-run` and `rocprof-sys-sample` share the same backend. In fact, IIRC `rocprof-sys-run -S` is requivalent to `rocprof-sys-sample`. So I don't see the need to test both binaries.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
kcossett-amd, line 1: Many of the new tests here just test the `CLI` arg implies `env` var is set. These should be unit tests and not pytests.

projects/rocprofiler-systems/tests/pytest/test_selective_regions.py
```diff
@@ -223,6 +227,49 @@ def test_region_2_and_3_filter(self, mode, marker_style, selective_region_env):
...
 
 
+# =============================================================================
+# Test Class: Selective Region via --selected-regions CLI flag
+# =============================================================================
+
+
+@pytest.mark.parametrize("mode", ["sys_run", "sampling"])
+@pytest.mark.class_name("selective-region-cli-flag")
+class TestSelectiveRegionCliFlag(RocprofsysTest):
```
kcossett-amd: This should be a unit test (not pytest) testing `CLI` => `env` var.

kcossett-amd (changes requested):

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
rpathani: I am agree on it with the condition if they share the same code path of cli. if that is tested well we can limit the  coverage only with one of binary. @kcossett-amd  and @anujshuk-amd  let me know on it as i havent investigated the cli code path for both the binary.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
rpathani: I agree that unit test coverage is important for smoke testing and quick regression validation. However, the CLI path is equally important, as it directly impacts the user experience. To ensure comprehensive coverage, we should have strong unit tests for validating individual components, along with CLI tests to verify end-to-end integration and user workflows. This combination provides better confidence in both functionality and overall user experience. Also we need to see we are not duplicating the tests scenarios across testsuite.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: Good catch. you are right, -v 2 wasn't gating anything. I checked, and the runner prints the ROCPROFSYS: KEY=value env lines unconditionally, so the assert_regex checks passed with or without -v 2

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
kcossett-amd: All a CLI arg does under the hood is set the equivalent env var (or potentially a set of them) to true. If we already have a test that runs via env var, having another one that does the same thing but with CLI is redundant.
Full E2E tests are more expensive than a unit test, and when the difference is sometimes just CLI -> env var, then a unit test would work better. 

However, I am not opposed to switching all tests to use CLI instead of env vars where applicable. I know we are changing the `examples` `README.md` launch commands to use CLI as opposed to `-e`/env vars.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -2,45 +2,120 @@
...
 ]
 
+# Args for parallel-overhead: <nfib> <nthreads> <nitr>. Big enough to run for
+# about a second and get sampled a bunch, small enough to keep tests quick.
+WORKLOAD_ARGS = ["18", "2", "100000"]
+
+# Args for transpose (the GPU workload): <matrix-size> <iterations> <num-inputs>.
+# Enough kernel launches and runtime for the -D device sampler to poll GPU
+# metrics at least once, while still finishing quickly.
+GPU_WORKLOAD_ARGS = ["4", "300", "50"]
```
kcossett-amd:
```suggestion
# <matrix-size> <iterations> <num-inputs>
# Enough kernel launches and runtime for the -D device sampler to poll GPU
# metrics at least once, while still finishing quickly.
TRANSPOSE_ARGS= ["4", "300", "50"]
```
Make var name self-explanatory

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -2,45 +2,120 @@
...
+    except FileNotFoundError:
+        pytest.skip("parallel-overhead example not built")
+
+
+@pytest.fixture
+def gpu_workload(rocprof_config) -> str:
+    """HIP workload that dispatches real GPU kernels.
+
+    parallel-overhead is CPU-only and can't exercise -D, so the GPU test uses
+    transpose instead.
```
kcossett-amd:
```suggestion
```
Unneeded comment.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -2,45 +2,120 @@
...
+def instrumentable_child(rocprof_config) -> str:
+    """CPU workload with real, instrumentable functions. Used by most tests here."""
+    try:
+        return str(rocprof_config.get_target_executable("parallel-overhead"))
+    except FileNotFoundError:
+        pytest.skip("parallel-overhead example not built")
+
+
+@pytest.fixture
+def gpu_workload(rocprof_config) -> str:
```
kcossett-amd, line 51: Please keep a similar naming convention. `instrumentable_child`'s description implies you are mainly using this for cpu work, so have it match `gpu_workload` naming scheme and call it `cpu_workload`.

Or, change both of these to reflect the binary being used. Scrolling up to see what binary is being used is somewhat cumbersome.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -2,45 +2,120 @@
...
+pytestmark = [pytest.mark.tool_runner]
 
 TARGETS = [
     pytest.param("rocprof-sys-run", marks=pytest.mark.sys_run, id="run"),
     pytest.param("rocprof-sys-sample", marks=pytest.mark.sampling, id="sample"),
 ]
 
+# Args for parallel-overhead: <nfib> <nthreads> <nitr>. Big enough to run for
+# about a second and get sampled a bunch, small enough to keep tests quick.
+WORKLOAD_ARGS = ["18", "2", "100000"]
```
kcossett-amd:
```suggestion
# <nfib> <nthreads> <nitr>. Big enough to run for
# about a second and get sampled a bunch, small enough to keep tests quick.
PARALLEL_OVERHEAD_ARGS = ["18", "2", "100000"]
```
Make var name self-explanatory

projects/rocprofiler-systems/tests/pytest/test_selective_regions.py
dgaliffiAMD: I would recommend removing the "selective regions" change from this Pull Request entirely. 
I am not against additional coverage, but it seems to be "out-of-scope" for this PR. The description aligns with this assessment: "Separately, ..." (from the second paragraph of the "Motivations" section.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -146,4 +258,348 @@ def test_conflicts_with_legacy_flags(self, target, legacy_args):
...
+                f"produce a metadata file under {output_dir}"
+            )
+        self.assert_file_exists(metadata_file, description="Sample metadata output")
+
+        settings = _parse_metadata_settings(metadata_file)
+        assert settings["ROCPROFSYS_CONFIG_FILE"] == str(empty_cfg)
+        assert settings["ROCPROFSYS_OUTPUT_PATH"] == str(output_dir)
+        assert settings["ROCPROFSYS_OUTPUT_PREFIX"] == output_prefix
+
+    def test_malformed_file_is_rejected(
```
dgaliffiAMD: This test might be better off in `test_config.py`. We already have a file dedicated to config file validation.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -146,4 +258,348 @@ def test_conflicts_with_legacy_flags(self, target, legacy_args):
...
+
+# =============================================================================
+# Trace/profile and host/device flags
+# -T/-L/-P/-F and -H/-D each fan one flag out to a few env vars, so they share
+# one parametrized class instead of several near-identical ones.
+# =============================================================================
+
+CLI_FLAG_ENV_CASES = [
+    pytest.param(["-T"], {"ROCPROFSYS_TRACE": True}, "trace", id="trace"),
+    pytest.param(["-L"], {"ROCPROFSYS_TRACE_LEGACY": True}, "trace", id="trace_legacy"),
```
dgaliffiAMD: I wouldn't add the "TRACE_LEGACY" validate. We have plans to deprecate it.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: Taken

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: adressed.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: you are right, with the base env pre-injecting ROCPROFSYS_TRACE/ROCPROFSYS_PROFILE=ON, those =true checks were vacuous. Fixed by seeding the opposite value first, so the assertion can only pass if the flag actually flips it.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: so there's no ls left in the file. They reject the args before the workload runs, so it doesn't change what's tested, but it drops the contradiction.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: Agreed. Since both share the arg-parsing backend, I dropped the second binary and these now run through rocprof-sys-run only (about half the cases).

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: removed.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: added.

projects/rocprofiler-systems/tests/pytest/test_selective_regions.py
anujshuk-amd: Agreed, that was out of scope. Reverted test_selective_regions.py back to develop so this PR no longer touches it.  I will bring the --selected-regions coverage separately.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: Moved it to test_config.py as test_config_cli_flag_rejects_invalid, reusing the existing config fixtures. Kept only the part that adds value there. That -c reaches the same validation as the env var. And dropped it from test_tool_runner.py. Verified it passes.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: Switched to the create_config_file fixture instead of writing the file by hand. Used in both the config-output test and the malformed-config test now.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: As Rohit suggested and as per my understading lets have it. Most of these also assert the resolved metadata.json value and the produced artifact (trace/profile with the workload fib frames), which is real E2E coverage a unit test wouldn't give. Happy to follow your suggestion and standardize on CLI and move the pure CLI env checks to unit tests. Already reduced to one binary

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -108,12 +196,30 @@ def test_proto_rocpd_enables_both(self, target):
...
+        assert settings["ROCPROFSYS_TRACE"] is True
+        assert settings["ROCPROFSYS_USE_ROCPD"] is True
+        assert settings["ROCPROFSYS_PROFILE"] is False
+        assert result.perfetto_file is not None, "expected a perfetto trace file"
+        assert result.rocpd_files, "expected at least one rocpd database file"
+        self.assert_perfetto(
+            result,
+            label_substrings=["fib"],
+            subtest_name="proto trace records workload fib frames",
+        )
```
kcossett-amd, line 216: Fails on my system. The problem is that `parallel-overhead` finishes running before sampling even begins (0.44 seconds). Recall that there is a default sampling delay of 0.5 seconds. 

When it then tries to do the `fib` assert, it won't find it in perfetto.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -135,15 +251,335 @@ def test_rocpd_only_disables_perfetto(self, target):
...
+    """Each flag sets its documented env var(s), verified via the stdout echo
+    and metadata.json, and produces the artifact it implies.
+    """
+
+    BASE_INJECTED = ("ROCPROFSYS_TRACE", "ROCPROFSYS_PROFILE")
+
+    @pytest.mark.parametrize(
+        "flag_args, expected_settings, artifact_kind", CLI_FLAG_ENV_CASES
+    )
+    def test_sets_expected_settings_and_artifacts(
```
kcossett-amd, line 397: All variants of this test fails on my system. The problem is that `parallel-overhead` finishes running before sampling even begins (0.44 seconds). Recall that there is a default sampling delay of 0.5 seconds. 

When it then tries to do the `fib` assert, it won't find it in perfetto.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -135,15 +251,335 @@ def test_rocpd_only_disables_perfetto(self, target):
...
+            result,
+            pass_regex=[
+                f"{key}={str(value).lower()}" for key, value in expected_settings.items()
+            ],
+        )
+
+        settings = _resolved_settings(result)
+        for key, expected in expected_settings.items():
+            if key == "ROCPROFSYS_USE_AMD_SMI":
+                continue
```
kcossett-amd: What is the purpose of this?

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -2,45 +2,118 @@
...
 ]
 
+# <nfib> <nthreads> <nitr>. Big enough to run for about a second and get
+# sampled a bunch, small enough to keep tests quick.
+PARALLEL_OVERHEAD_ARGS = ["18", "2", "100000"]
+
+# <matrix-size> <iterations> <num-inputs>. Enough kernel launches and runtime
+# for the -D device sampler to poll GPU metrics at least once, while still
+# finishing quickly.
+TRANSPOSE_ARGS = ["4", "300", "50"]
```
kcossett-amd: Description in incorrect. Per transpose's README.md:
```
| Position | Description | Default |
| ---------- | ------------- | --------- |
| 1 | Number of CPU threads (each gets a HIP stream) | 2 |
| 2 | Number of kernel iterations | 500 |
| 3 | Synchronize every N iterations | 10 |
```

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -135,15 +251,335 @@ def test_rocpd_only_disables_perfetto(self, target):
...
+                label_substrings=["fib"],
+                subtest_name=f"{flag_args}: trace records workload fib frames",
+            )
+        elif artifact_kind == "profile":
+            assert result.timemory_files, (
+                f"{flag_args} should produce timemory profile output under "
+                f"{result.output_dir}"
+            )
+            # Same check for the profile: fib() must be a recorded sample entry.
+            assert _profile_has_label(result, "fib(long)"), (
```
kcossett-amd: Why `fib(long)` and not `fib`?

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -2,45 +2,118 @@
...
+    """HIP workload that dispatches real GPU kernels."""
+    try:
+        return str(rocprof_config.get_target_executable("transpose"))
+    except FileNotFoundError:
+        pytest.skip("transpose example not built")
+
+
+def workload_args(workload: str) -> list[str]:
+    """Trailing "-- <workload>" so the flags under test act on real work."""
+    return ["--", workload, *PARALLEL_OVERHEAD_ARGS]
```
kcossett-amd: `workload_args` makes it sound like a generic function that could accept either `cpu_workload` or `gpu_workload`. But then you force append `PARALLEL_OVERHEAD_ARGS`. 

This is very confusing.

projects/rocprofiler-systems/tests/pytest/test_config.py
```diff
@@ -252,6 +252,32 @@ def test_variable_boolean_value_is_not_rejected(
...
+            skip_filter=True,
+        )
+        env = MINIMAL_RUNTIME_ENV.copy()
+        result = self.run_test(
+            "sys_run",
+            target=config_target,
+            env=env,
+            sys_run_args=["-c", str(config_file)],
+            fail_on_pass=True,
+        )
```
kcossett-amd:
```suggestion
        result = self.run_test(
            "sys_run",
            target=config_target,
            env=MINIMAL_RUNTIME_ENV,
            sys_run_args=["-c", str(config_file)],
            fail_on_pass=True,
        )
```
No need to copy `MINIMAL_RUNTIME_ENV` into `env` if you are not mutating it any further.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
```diff
@@ -135,15 +251,335 @@ def test_rocpd_only_disables_perfetto(self, target):
...
+            self.assert_perfetto(
+                result,
+                label_substrings=["fib"],
+                subtest_name=f"{flag_args}: trace records workload fib frames",
+            )
+        elif artifact_kind == "profile":
+            assert result.timemory_files, (
+                f"{flag_args} should produce timemory profile output under "
+                f"{result.output_dir}"
+            )
```
kcossett-amd: This is redundant. `metadata.json` and `functions.json` are aways generated and so this check will always pass as `timemory_files` is a `glob` of whatever is in the dir.

You should instead name the file you expect.

kcossett-amd (changes requested):

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: chnaged.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: Chnging to :sampling at 0.1 seconds instead of the default 0.5 seconds, giving Perfetto time to capture fib frames before parallel-overhead exits.

We also keep the workload sufficiently long.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: It had skipped the metadata.json assertion for the AMD-SMI setting.

To make that safe:

-D and -H -D cases are marked pytest.mark.gpu, so they run only where AMD-SMI device sampling can resolve.
-H remains a CPU-safe case and verifies that ROCPROFSYS_USE_AMD_SMI resolves to False.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: corrected.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: fib(long) is the workload’s full demangled C++ signature. I changed the assertion to match the stable fib. Good catch.

projects/rocprofiler-systems/tests/pytest/test_config.py
anujshuk-amd: taken.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: yes true. Changed.

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
kcossett-amd: I don't understand? What is "this"? Is it some change you made or something that is already there?

kcossett-amd: Resolving, tests pass.

kcossett-amd (approved):
Thank you for addressing my comments. Approving on my side. Please wait for someone else to approve as well.

dgaliffiAMD (approved):

projects/rocprofiler-systems/tests/pytest/test_tool_runner.py
anujshuk-amd: Removed.

dgaliffiAMD (changes requested):
Your Pull Request now contains changes to the `systems_pr_bot` scripts. Must have been unintentionally included in a rebase. Please fix before merging.

anujshuk-amd:
> Your Pull Request now contains changes to the `systems_pr_bot` scripts. Must have been unintentionally included in a rebase. Please fix before merging.

@dgaliffiAMD  Please take a look I have adressed this issue.

dgaliffiAMD (approved):

PR #9091, opened 2026-07-23
[AICOMRCCL-1577] test: bound NetIB connection handshake and cap net_ib_transfer timeouts

atulkulk:
## Motivation

Two RCCL test-harness issues on the MI300X + Mellanox IB setup caused hangs and spurious failures. NetIB MPI transport tests could hang indefinitely when a peer's IB fabric was unhealthy; this was hit in practice when one node's IB interfaces were stuck in the INIT port state (no Subnet Manager had moved them to ACTIVE), so the handshake never completed. The ports were later restored to ACTIVE, but the harness must fail fast regardless of fabric health. Separately, Bootstrap_Combined failed comm-init on every rank because its OOB-net interface was misconfigured.

## Technical Details

- `NetIbMPITestBase.hpp`: added a `kConnectTimeoutMs` (30s) watchdog to the `SetupConnection()` handshake loops with throttled polling, so a non-completing handshake fails deterministically instead of spinning forever.
- `mi300x_mellanox_ib.json` (net_ib_transfer): added a 300s suite-level default plus per-test 300s timeouts to the transfer tests that lacked one.
- `mi300x_mellanox_ib.json` (Bootstrap_Combined): set `NCCL_OOB_NET_IFNAME` to the IB HCA `mlx5_0`. With `NCCL_OOB_NET_ENABLE=1` the OOB device is selected from the NET plugin (IB) device list via `matchIfList(props.name, ...)`, so the value must name an HCA, not an ethernet interface (`eth0`/`eth1` matched no device).

## Issue Tracking

JIRA ID: AICOMRCCL-1577

## Test Plan

Ran the NET IB suites (Initialization, Device Properties, Connection Setup, Memory Registration, Data Transfer) and the Bootstrap/OOB Net rccl-tests suite via test_runner with `rccl-UnitTestsMPI` on an MI300X node (8 GPUs, 6 active Mellanox HCAs), after the affected node's IB ports were restored from INIT to ACTIVE.

## Test Result

NET IB core suites: 8 passed, 0 failed, 0 timeouts, 4 topology-gated skips; previously hanging Connection Setup / Data Transfer tests now complete in 9-15s each. Bootstrap_Combined: PASSED (146s, 0 out-of-bounds values).

## Submission Checklist

### DIFF
--- a/projects/rccl/test/transport/NetIbMPI/NetIbMPITestBase.hpp
+++ b/projects/rccl/test/transport/NetIbMPI/NetIbMPITestBase.hpp
@@ -355,7 +355,10 @@ protected:
             int attempts = 0;
             while (!done) {
                 ncclResult_t result = ConnectToRemote(dev, &pair.handle, &pair.sendComm);
-                if (result == ncclSuccess && pair.sendComm != nullptr) {
+                if (result != ncclSuccess) {
+                    return result;
+                }
+                if (pair.sendComm != nullptr) {
                     done = 1;
                     break;
                 }
--- a/projects/rccl/test/transport/NetIbMPI/NetIbMPITestBase.hpp
+++ b/projects/rccl/test/transport/NetIbMPI/NetIbMPITestBase.hpp
@@ -168,6 +168,7 @@ protected:
     // Timing constants
     static constexpr int kDefaultTimeoutMs = 5000;
     static constexpr int kLargeTransferTimeoutMs = 30000;
+    static constexpr int kConnectTimeoutMs = 30000;  // Handshake watchdog (see SetupConnection)
     static constexpr int kPollIntervalUs = 10000;  // 10ms
     static constexpr int kPollIntervalMs = 10;
     static constexpr int kMaxRetryAttempts = 1000;  // For NULL request handling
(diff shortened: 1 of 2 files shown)


projects/rccl/test/transport/NetIbMPI/NetIbMPITestBase.hpp
```diff
@@ -327,23 +331,38 @@ class NetIbMPITest : public MPITestBase {
...
 
             // Connect to peer
             int done = 0;
+            int attempts = 0;
             while (!done) {
                 ncclResult_t result = ConnectToRemote(dev, &pair.handle, &pair.sendComm);
                 if (result == ncclSuccess && pair.sendComm != nullptr) {
                     done = 1;
+                    break;
                 }
```
Kapil-Shyam-Pawar, line 364: return actual failure result here as well?

Kapil-Shyam-Pawar (approved):

projects/rccl/test/transport/NetIbMPI/NetIbMPITestBase.hpp
```diff
@@ -327,23 +331,35 @@ class NetIbMPITest : public MPITestBase {
 
             // Accept connection
             int done = 0;
+            int attempts = 0;
             while (!done) {
                 ncclResult_t result = AcceptConnection(pair.listenComm, &pair.recvComm);
                 if (result == ncclSuccess && pair.recvComm != nullptr) {
                     done = 1;
+                    break;
                 }
```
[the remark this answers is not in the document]
atulkulk: @copilot  Apply the fix to other locations as well.

projects/rccl/test/transport/NetIbMPI/NetIbMPITestBase.hpp
atulkulk: @copilot Apply the fix to other locations as well.

PR #9094, opened 2026-07-23
[rocjitsu] Limit indirect dataflow to pending pairs

kuhar:
## Motivation

Reduce translation time and memory for DeepSeek V4 Flash gfx1250 B0 kernels.

## Technical Details

Propagate SGPR-pair facts only for unresolved cross-block consumers and reject out-of-range selectors.

## Issue Tracking

N/A

## Test Plan

A/B translate and byte-compare all 128 unique DeepSeek V4 Flash code objects with and without pending-pair slicing.

## Test Result

DeepSeek V4 Flash: 326 s to 211 s (1.55x speedup); peak RSS 31 GiB to 20 GiB (36% lower). Outputs were byte-identical.

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp
@@ -1447,10 +1447,11 @@ void scan_block(AnalysisContext &ctx, size_t block_index, std::vector<AnalysisBl
         // source pair unless the instruction also defines it. Real kernels can
         // build one callee address once and issue multiple swappc calls through
         // that same pair on the fallthrough path.
-      } else if (!state.pair_dirty(*consumer_pair)) {
+      } else if (*consumer_pair < kMaxTrackedSgprPair && !state.pair_dirty(*consumer_pair)) {
         // The block did not touch the pair, so any useful fact must come from
         // predecessor blocks. Defer classification until the block-entry
-        // lattice is available.
+        // lattice is available. Out-of-range selectors cannot name a tracked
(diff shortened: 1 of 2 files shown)


atgutier (approved):
Only a nit that should probably be addressed:

 [NIT] lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp:1418-1422 — comment
  overstates which special-register selectors are rejected

  Reasoning: The new condition rejects selectors at or above kMaxTrackedSgprPair, but it does not
  reject every special-register selector as the comment implies.

  Evidence:

  - REGISTER_SET_MAX_SGPRS is 106, making kMaxTrackedSgprPair 105.
  - CDNA encodes flat_scratch at selectors 102/103 and xnack_mask at 104/105.
  - Consequently, pair-low selectors 102 and 104 still pass the < 105 condition and participate in
    pending-pair dataflow.

  - The new test at tests/analysis/liveness_test.cpp:514-529 covers only VCC selector 106, which is
    outside the numeric tracking bound.

  - This is not a functional regression: these lower special selectors were tracked before the PR, and
    the new relevance slice preserves that behavior.

  Suggested fix: Reword the comment and test to say “out-of-range selector,” which accurately
  describes the implementation. If the intent is to exclude every non-GPR selector, use an
  architecture-specific selector predicate and add flat-scratch/XNACK cases.

  Confidence: high.

  The pending-pair optimization itself is sound. Every valid cross-block consumer is collected before
  dataflow, transfers are independent per SGPR pair, and the relevant set is rebuilt after each
  recovered-leader/successor graph reconstruction. SET, KILL, PASS, missing-path incompleteness,
  target saturation, loop propagation, and fixed-point behavior remain unchanged for relevant pairs.
  Transfer keys are independently bounded, so relevant_pairs[pair_lo] is safe.

PR #9096, opened 2026-07-23
hotswap: scale the dispatch extent for modulo-replicated kernels

ftynse:
Draft, stacked on #8752 (lazy per-kernel presentation mode). The first nine commits here belong to that PR; this one adds only the final commit, `hotswap: scale the dispatch extent for modulo-replicated kernels`. Please review after #8752 lands (or treat #8752 as the base).

## What

A kernel raised under the comgr `ScaledModuloReplicationProjection` expects to be launched with a wider block along the wave-carrying x dimension: the projection packs each source wavefront together with its replicas into one target wave, so the block needs W_t/W_s times as many threads in x (a factor of two for a wave32 source on a wave64 target). COMGR already reports that factor per kernel in the transpile result, but the lazy launch path ignored it, so a marked kernel ran with an unscaled block and produced wrong results.

This carries the factor from the transpile result into the kernel record and applies it when the kernel is dispatched. The standard dispatch packet also scales `grid_size_x`, whose extent is counted in workitems, so the block count is unchanged; the ext cluster packet keeps its block count in the cluster fields and needs only the block extent scaled. Kernels that were not modulo-replicated report a factor of one and are left alone.

## Dependency

Needs a COMGR that exposes `AMD_COMGR_HOTSWAP_TRANSPILE_RESULT_SCALED_DISPATCH_FACTOR` (the llvm-project `ScaledModuloReplicationProjection` work).

## Validation

Built comgr + rocr and ran on an MI300X (gfx1250->gfx942 transpile, cold cache): the transpiled Qwen3 RMSNorm reduce is numerically correct at odd row counts (rel ~2.3e-3) with no aperture fault at rows=21 (which faults without the fix), and the dispatch trace shows factor-2 scaling on only the reduce kernels. Added runtime unit tests for the standard- and ext-packet scaling paths.

### DIFF
--- a/projects/clr/rocclr/device/device.hpp
+++ b/projects/clr/rocclr/device/device.hpp
@@ -2288,6 +2288,14 @@ class Device : public RuntimeObject {
     return *isa_;
   }
 
+  //! Returns the ISA used for hardware-adjacent ROCr decisions.
+  const Isa& executionIsa() const {
+    assert(executionIsa_);
+    return *executionIsa_;
+  }
+
+  void setExecutionIsa(const Isa& isa) { executionIsa_ = &isa; }
+
   //! Return a non-zero uint64_t value that uniquely identifies the device.
   //! This can be used when a scalar value handle to the device is require.
   static uint64_t toHandle(const Device* device) {
@@ -2494,7 +2502,8 @@ class Device : public RuntimeObject {
   std::unordered_map<amd::CommandQueue*, bool> activeQueues;  //!< The set of active queues
   uint8_t group_mem_carveout_hint_{0}; //!< LDS carveout percentage (0 = no preference)
  private:
-  const Isa* isa_;  //!< Device isa
+  const Isa* isa_;           //!< Presented/code-selection device isa
+  const Isa* executionIsa_;  //!< Hardware-adjacent device isa
   bool IsTypeMatching(cl_device_type type, bool offlineDevices);
 
 #if defined(WITH_HSA_DEVICE)
--- a/projects/clr/rocclr/platform/command.cpp
+++ b/projects/clr/rocclr/platform/command.cpp
@@ -247,7 +247,7 @@ bool Event::awaitCompletion() {
     ClPrint(LOG_DETAIL_DEBUG, LOG_WAIT, "Waiting for event %p to complete, current status %d",
             this, status());
     auto* queue = command().queue();
-    if ((queue != nullptr) && queue->vdev()->ActiveWait()) {
+    if ((queue != nullptr) && queue->vdev() != nullptr && queue->vdev()->ActiveWait()) {
       while (status() > CL_COMPLETE) {
         amd::Os::yield();
       }
@@ -372,6 +372,11 @@ void Command::enqueue() {
   // update will occur later after flush() with a wait
   if (AMD_DIRECT_DISPATCH) {
     setStatus(CL_QUEUED);
+    if (queue_->vdev() == nullptr) {
+      LogError("Command enqueue failed: command queue has no virtual device");
+      setStatus(CL_INVALID_OPERATION);
+      return;
+    }
 
     // Notify all commands about the waiter. Barrier will be sent in order to obtain
     // HSA signal for a wait on the current queue
--- a/projects/clr/rocclr/platform/commandqueue.cpp
+++ b/projects/clr/rocclr/platform/commandqueue.cpp
@@ -160,6 +160,9 @@ void HostQueue::finish(bool cpu_wait) {
   Command* command = nullptr;
   size_t minBatchSize = 0;
 
+  auto* virtualDevice = vdev();
+  guarantee(virtualDevice != nullptr, "HostQueue has no virtual device");
+
   if (IS_HIP) {
     minBatchSize = DEBUG_CLR_BATCH_CPU_SYNC_SIZE;
 
@@ -176,7 +179,7 @@ void HostQueue::finish(bool cpu_wait) {
 
     // Force blocking wait if requested. That allows to avoid a build up of unreleased CPU commands
     if ((DEBUG_HIP_BLOCK_SYNC > 0) &&
-        (vdev()->QueuedAsyncHandlers().load() > DEBUG_HIP_BLOCK_SYNC)) {
+        (virtualDevice->QueuedAsyncHandlers().load() > DEBUG_HIP_BLOCK_SYNC)) {
       cpu_wait = true;
     }
   } else {
@@ -188,10 +191,10 @@ void HostQueue::finish(bool cpu_wait) {
   ClPrint(LOG_DETAIL_DEBUG, LOG_CMD,
           "finish() called with batch size: %zu, cpu_wait: %d, "
           "fence dirty: %d",
-          batchSize, cpu_wait, vdev()->isFenceDirty());
+          batchSize, cpu_wait, virtualDevice->isFenceDirty());
 
   // Force marker if the batch wasn't sent for CPU update or fence is dirty
-  if (nullptr == command || (batchSize != 0)|| vdev()->isFenceDirty()) {
+  if (nullptr == command || (batchSize != 0)|| virtualDevice->isFenceDirty()) {
     if (nullptr != command) {
       command->release();
     }
@@ -356,9 +359,11 @@ bool HostQueue::isEmpty() {
 
 Command* HostQueue::getLastQueuedCommand(bool retain) {
   if (AMD_DIRECT_DISPATCH) {
+    auto* virtualDevice = vdev();
+    guarantee(virtualDevice != nullptr, "HostQueue has no virtual device");
     // The batch update must be lock protected to avoid a race condition
     // when multiple threads submit/flush/update the batch at the same time
-    std::scoped_lock sl(vdev()->execution());
+    std::scoped_lock sl(virtualDevice->execution());
     // Since the lastCmdLock_ is acquired, it is safe to read and retain the lastEnqueueCommand.
     // It is guaranteed that the pointer will not change.
     if (retain && lastEnqueueCommand_ != nullptr) {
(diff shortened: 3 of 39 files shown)


projects/clr/rocclr/platform/command.cpp
```diff
@@ -247,7 +247,7 @@ bool Event::awaitCompletion() {
     ClPrint(LOG_DETAIL_DEBUG, LOG_WAIT, "Waiting for event %p to complete, current status %d",
             this, status());
     auto* queue = command().queue();
-    if ((queue != nullptr) && queue->vdev()->ActiveWait()) {
+    if ((queue != nullptr) && queue->vdev() != nullptr && queue->vdev()->ActiveWait()) {
```
gandryey, line 250: Could we avoid any cosmetic modifications unrelated to the original change? It should be impossible to have queue without vdev(). Even if it's occurred I would suggest to fail queue creation rather than extra checks of vdev everywhere

projects/clr/rocclr/platform/commandqueue.cpp
```diff
@@ -160,6 +160,9 @@ void HostQueue::finish(bool cpu_wait) {
   Command* command = nullptr;
   size_t minBatchSize = 0;
 
+  auto* virtualDevice = vdev();
```
gandryey, line 163: Avoid any cosmetic modifications for the this big functional change. It complicates the review process and makes incorrect impression about the actual functionality scope.

projects/clr/rocclr/device/device.hpp
```diff
@@ -2494,7 +2502,8 @@ class Device : public RuntimeObject {
   std::unordered_map<amd::CommandQueue*, bool> activeQueues;  //!< The set of active queues
   uint8_t group_mem_carveout_hint_{0}; //!< LDS carveout percentage (0 = no preference)
  private:
-  const Isa* isa_;  //!< Device isa
+  const Isa* isa_;           //!< Presented/code-selection device isa
+  const Isa* executionIsa_;  //!< Hardware-adjacent device isa
```
gandryey, line 2506: Do we really need 2 versions of Isa objects? Is it possible to avoid this complexity? Where do you still use the original isa_ object? I would prefer if we avoid massive renames of isa() to executionIsa() everywhere in the code and keep isa() as it is.

gandryey (changes requested):

cfreeamd (commented):

The API rename (`supported_isas()[0]` → `execution_isa()`/`execution_isas()[0]`) is consistent across `agent.h`, `amd_gpu_agent.h/.cpp`, `amd_blit_kernel.cpp`, `amd_blit_sdma.cpp`, `amd_topology.cpp`, `amd_aql_queue.cpp`, and `amd_loader_context.cpp`; new files carry correct SPDX MIT headers; and the CMakeLists.txt gating options are `OFF` by default.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap_aql_patch.h
```diff
@@ -0,0 +1,120 @@
...
+    fprintf(stderr, "hotswap: refusing kernel dispatch: packet patch failed\n");
+    std::abort();
+  }
+
+  // A kernel raised under the comgr ScaledModuloReplicationProjection runs with
+  // its x extent scaled so each target wave hosts one source wave in its low
+  // lanes and replicas of those lanes above. Only such kernels report a factor
+  // above 1.
+  if (scaled_dispatch_factor > 1) {
+    *workgroup_size_x = static_cast<uint16_t>(*workgroup_size_x *
```
cfreeamd, line 104: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** `static_cast<uint16_t>(*workgroup_size_x * scaled_dispatch_factor)` can silently truncate. The product of `workgroup_size_x` (uint16_t, max 1024 per HSA spec) and `scaled_dispatch_factor` could exceed 65535 in a misconfigured kernel, and the cast wraps silently rather than fail-closing like the surrounding `std::abort()` paths. Consider asserting `*workgroup_size_x * scaled_dispatch_factor <= 65535` (and ideally also `<= workgroup_max_dim_x`) before the cast, or aborting with a message if the result exceeds the hardware limit.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap_aql_patch.h
```diff
@@ -0,0 +1,120 @@
...
+                                              scaled_dispatch_factor);
+    if (grid_size_x) *grid_size_x *= scaled_dispatch_factor;
```
cfreeamd, line 106: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** `*grid_size_x *= scaled_dispatch_factor` on the uint32_t grid field has no overflow guard. A large 1D grid (>2^31 workitems) times a factor of 2 wraps and silently under-dispatches rather than failing. The rest of this function fail-closes with `std::abort()` on any inconsistency — an overflow check here (`*grid_size_x > UINT32_MAX / scaled_dispatch_factor`) would keep that contract.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/hotswap_aql_patch.h
```diff
@@ -0,0 +1,120 @@
...
+                      __ATOMIC_ACQUIRE);
+  const hsa_packet_type_t type =
+      static_cast<hsa_packet_type_t>((header >> HSA_PACKET_HEADER_TYPE) & 0xff);
+  uint64_t* kernel_object = nullptr;
+  uint32_t* private_segment_size = nullptr;
+  uint32_t* group_segment_size = nullptr;
+  uint16_t* workgroup_size_x = nullptr;
+  // Set only for the standard dispatch packet, whose grid extent is in
+  // workitems: it scales alongside the block so the block count stays constant.
+  // The ext (cluster) packet expresses its grid through cluster_count/size, so
```
cfreeamd, line 66: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** The comment asserts that for the ext (cluster) packet, `grid_size_x` staying null is correct because the effective global grid is expressed through `cluster_count`/`cluster_size` rather than `grid_size_x`. Can you confirm this holds for all ext packet variants we intend to support (i.e., that `cluster_count` is NOT in workitems and therefore must not be scaled)? A reference to the ext packet spec field semantics (or a link/comment to where they're defined) would make this assumption auditable by future readers.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
```diff
@@ -480,6 +496,51 @@ uint64_t AqlQueue::AddWriteIndexRelease(uint64_t value) {
...
+            last_doorbell_value_, end, queue_size, first, has_work,
+            failure_reason)) {
+      fprintf(stderr, "hotswap: refusing doorbell patch: %s\n",
+              failure_reason ? failure_reason : "invalid AQL doorbell range");
+      std::abort();
+    }
+
+    for (uint64_t index = first; has_work && index <= end; ++index) {
+      const uint64_t slot = index % queue_size;
+      if (hotswap_patched_generations_[slot] == index) continue;
```
cfreeamd, line 523: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** `hotswap_patched_generations_[slot]` is indexed by `slot = index % queue_size` where `queue_size = amd_queue_.hsa_queue.size`. `hotswap_patched_generations_` is sized with `queue_size_pkts` in the constructor. Please confirm `amd_queue_.hsa_queue.size` is always equal to `queue_size_pkts` (in packets) at this point — if they ever differ (e.g., `hsa_queue.size` were in bytes or rounded differently), the modulo index could run past the vector. An `assert(queue_size == hotswap_patched_generations_.size())` inside the presentation-mode branch would make this invariant explicit.

projects/rocr-runtime/runtime/hsa-runtime/core/inc/amd_gpu_agent.h
```diff
@@ -762,6 +783,12 @@ class GpuAgent : public GpuAgentInt {
   // @brief Enumeration index
   uint32_t enum_index_;
 
+ private:
+  const core::Isa* isa_;
```
cfreeamd, line 787: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** Moving `isa_`, `presented_isa_`, and `presented_isas_` into a new `private:` block (splitting the class into private/protected/... sections) is the right encapsulation, since the whole PR depends on callers going through `execution_isa()`/`execution_isas()` rather than accessing `isa_` directly. A short comment explaining the access pattern ("all execution-ISA reads must go through execution_isa()/execution_isas() so presentation mode can intercept them") would make the reason for privatization clear to future readers.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_gpu_agent.cpp
```diff
@@ -187,36 +194,69 @@ GpuAgent::GpuAgent(HSAuint32 node, const HsaNodeProperties& node_props, bool xna
...
+  assert(isa_ != nullptr && "ISA registry inconsistency.");
 
-  supported_isas_.push_back(isa);
+  supported_isas_.push_back(isa_);
   if (!supported_isas_[0]->GetIsaGeneric().empty()) {
     supported_isas_.push_back(core::IsaRegistry::GetIsa(supported_isas_[0]->GetIsaGeneric()));
   }
 
-  if (supported_isas_[0]->GetMajorVersion() == 12 && supported_isas_[0]->GetMinorVersion() >= 5) {
+  const char *PresentIsa = std::getenv("HSA_HOTSWAP_PRESENT_ISA");
```
cfreeamd, line 215: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `std::getenv("HSA_HOTSWAP_PRESENT_ISA")` is called directly here while the rest of this file routes env access through `rocr::hotswap::IsEnvFlagEnabled` / `os::GetEnvVar`. The direct `getenv` is functionally fine (it only reads the ISA name string; the enable decision just above already goes through `IsEnvFlagEnabled`), but routing it through `os::GetEnvVar` would keep env handling consistent and testable.

PR #9098, opened 2026-07-23
Add hipExtMemcpyBatchAsync with asymmetric swap

iassiour:
## Motivation

`hipMemcpyBatchAsync` selects per-copy operations (swap, indirect, PreferCE) through `hipMemcpyAttributes.flags`, but that struct is ABI-frozen (CUDA-compatible, no reserved space). This PR adds `hipExtMemcpyBatchAsync`, an extended batch-copy entry point with a forward-compatible attributes struct and room to grow (asymmetric swap sizes, a compute-engine copy preference, and reserved per-entry wait/signal).

## Technical Details

**New public API (`hip_runtime_api.h`)**

```c
hipError_t hipExtMemcpyBatchAsync(void** dsts, void** srcs,
                                  size_t* sizes, size_t* sizesDst,
                                  hipExtMemcpyWait* waits,
                                  hipExtMemcpySignal* signals,
                                  size_t count,
                                  hipExtMemcpyAttributes* attrs, size_t* attrsIdxs,
                                  size_t numAttrs,
                                  hipStream_t stream);
```

Per-copy operation selection (linear, swap, indirect src/dst, PreferCE, PreferCU) is via `attrs[i].std.flags` — the same `hipMemcpyFlags` used by `hipMemcpyBatchAsync`, range-mapped by `attrsIdxs`. There is **no** per-entry `ops[]` array.

**New types (`driver_types.h`)**

- **`hipExtMemcpyAttributes`** (56 bytes) — embeds `hipMemcpyAttributes std` (the CUDA-compatible prefix; op selection via `std.flags`) plus a `uint32_t reserved[8]` tail for forward-compat. Do not cast a `hipMemcpyAttributes*` to it.
- **`hipMemcpyFlagExtPreferCU`** — route linear copies onto the shader (compute engine) copy path instead of the copy engine (SDMA). Linear copies only; mutually exclusive with `hipMemcpyFlagExtPreferCE`.
- **`sizesDst`** — per-entry destination (B-side) swap size; `NULL` means symmetric.
- **`hipExtMemcpyWait` / `hipExtMemcpySignal`** (with `hipExtMemcpyWaitOp` / `hipExtMemcpySignalOp`) — reserved for future per-entry GPU-side wait/signal; **must be NULL** for now.
- **No `failIdx`** — the Ext API omits `failIdx` (mirroring CUDA's C base batch-memcpy signature); it remains available on the legacy `hipMemcpyBatchAsync`.

**Behavior / validation**

- Swap, indirect (src/dst), PreferCE, and PreferCU are all driven from `attrs[i].std.flags`. Validation returns `hipErrorInvalidValue` for unknown flag bits, swap combined with indirect, `PreferCE` combined with `PreferCU`, and `PreferCU` combined with swap/indirect (the shader copy path is linear-only).
- Swap requires at least one device operand; host↔host swap returns `hipErrorNotSupported`.
- `hipMemcpyBatchAsync` keeps its signature and converts its `hipMemcpyAttributes` into `hipExtMemcpyAttributes` by value, sharing the batch path.

**Asymmetric swap path**

- `command.hpp` / `device.hpp` carry the per-entry destination size; `rocblit.cpp` emits single-entry ROCr ops with separate `src_size`/`dst_size` for asymmetric batches (symmetric batches use the multi-entry op); ROCr rejects native asymmetric swap so the copy decomposes CLR-side.

**Tracing / ABI plumbing**

- `hip_api_trace` / `hip_prof_str.h` — new dispatch-table slot for `hipExtMemcpyBatchAsync` (`HIP_API_ID` 487, table step version 34) plus callback-args data.
- rocprofiler-sdk — runtime-api id, api-args, ABI entry, ostream, and enum-string support.
- `hip_hcc.map.in` / `amdhip.def.in` — export the new symbol.

## Issue Tracking

JIRA ID: AIRUNTIME-15

## Test Plan

Unit + contract tests for `hipExtMemcpyBatchAsync`: linear D2D, symmetric/asymmetric swap (including multi-attribute and host↔device), indirect src/dst/both, `PreferCU` shader-path routing, plus negatives for reserved wait/signal, unequal `sizesDst`, and attribute validation (unknown flags, swap+indirect, PreferCE+PreferCU, PreferCU+swap/indirect). Legacy `hipMemcpyBatchAsync` regression is retained.

## Test Result

Built CLR + ROCr standalone and ran on MI350X (gfx950): the full batch unit suite passes (indirect cases skip where SDMA indirect is unavailable on the device), `PreferCU` was verified to dispatch the shader copy-batch kernel via runtime logging, and the `mem_batch_copy` contract tests pass.

## Submission Checklist


projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_blit_sdma.cpp
```diff
@@ -1345,14 +1345,21 @@ hsa_status_t BlitSdma<useGCR, scopeFields>::SubmitBodies(
...
   const bool is_indirect = (indirect_src || indirect_dst);
 
+  // B-side size for an entry (asymmetric swap). Defaults to the A side, i.e.
+  // size_list[d], which keeps symmetric swaps and non-swap ops unchanged.
+  auto b_size = [&](uint32_t d) -> size_t {
+    return (is_swap && dst_size_list != nullptr) ? dst_size_list[d] : size_list[d];
+  };
+
   if (is_indirect && !indirect_copy_supported_)
     return HSA_STATUS_ERROR_INVALID_ARGUMENT;
```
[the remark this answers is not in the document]
iassiour: addressed

projects/hip/include/hip/hip_runtime_api.h
```diff
@@ -6149,6 +6149,39 @@ hipError_t hipMemcpyBatchAsync(void** dsts, void** srcs, size_t* sizes, size_t c
...
+ * @brief Perform Batch of 1D copies with extended operation support.
+ *
+ * Extended version of hipMemcpyBatchAsync with per-entry operation flags
+ * and asymmetric swap sizes. GPU-side wait/signal parameters are reserved
+ * for future use and must be NULL.
+ *
+ * @param [in] dsts        - Array of destination pointers.
+ * @param [in] srcs        - Array of source pointers.
+ * @param [in] sizesA      - Array of A-side copy sizes in bytes.
+ * @param [in] sizesB      - Array of B-side sizes for swap (NULL = symmetric).
```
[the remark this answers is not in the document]
iassiour: Done

projects/clr/hipamd/src/hip_memory.cpp
```diff
@@ -3242,70 +3360,35 @@ hipError_t hipMemcpyBatchAsync(void** dsts, void** srcs, size_t* sizes, size_t c
...
+                                  hipExtMemcpySignal* signals,
+                                  hipExtMemcpyOp* ops,
+                                  size_t count,
+                                  hipMemcpyAttributes* attrs, size_t* attrsIdxs, size_t numAttrs,
+                                  size_t* failIdx, hipStream_t stream) {
+  HIP_INIT_API(hipExtMemcpyBatchAsync, dsts, srcs, sizesA, sizesB, waits, signals, ops,
+               count, attrs, attrsIdxs, numAttrs, failIdx, stream);
+  if (!hip::isValid(stream)) {
+    HIP_RETURN(hipErrorInvalidResourceHandle);
   }
```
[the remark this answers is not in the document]
iassiour: Done

projects/clr/hipamd/src/hip_memory.cpp
```diff
@@ -3121,45 +3174,110 @@ hipError_t ihipMemcpyBatch(void** dsts, void** srcs, size_t* sizes, size_t count
...
+      if (ef & hipExtMemcpyOpSwap) {
+        metadata.copyOpType_ = amd::CopyMetadata::kCopyOpSwap;
+      } else if ((ef & hipExtMemcpyOpIndirectSrc) && (ef & hipExtMemcpyOpIndirectDst)) {
+        metadata.copyOpType_ = amd::CopyMetadata::kCopyOpIndirectSrcDst;
+      } else if (ef & hipExtMemcpyOpIndirectSrc) {
+        metadata.copyOpType_ = amd::CopyMetadata::kCopyOpIndirectSrc;
+      } else if (ef & hipExtMemcpyOpIndirectDst) {
+        metadata.copyOpType_ = amd::CopyMetadata::kCopyOpIndirectDst;
+      }
+    }
```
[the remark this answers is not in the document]
iassiour: Done

projects/clr/hipamd/src/hip_memory.cpp
```diff
@@ -3014,9 +3009,50 @@ static inline unsigned int getBatchCopyFlags(hipMemcpyAttributes* attrs, size_t*
...
+hipError_t ihipMemcpyBatch(void** dsts, void** srcs, size_t* sizes, size_t* sizesB,
+                           hipExtMemcpyWait* waits, hipExtMemcpySignal* signals,
+                           hipExtMemcpyOp* ops, size_t count,
                            hipMemcpyAttributes* attrs, size_t* attrsIdxs, size_t numAttrs,
-                           hip::Stream& stream, bool isAsync) {
+                           size_t* failIdx, hip::Stream& stream, bool isAsync) {
+  // Common input validation for both hipMemcpyBatchAsync and hipExtMemcpyBatchAsync.
+  if (dsts == nullptr || srcs == nullptr || sizes == nullptr || count == 0) {
+    return hipErrorInvalidValue;
+  }
```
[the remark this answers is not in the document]
iassiour: Fixed — hoisted the `*failIdx = SIZE_MAX` initialization to the top of `ihipMemcpyBatch`, so the early-return validation failures (null `dsts`/`srcs`/`sizes`, `count==0`, bad `attrs`/`attrsIdxs`) now leave `failIdx` at SIZE_MAX, matching the documented contract.

projects/rocr-runtime/runtime/hsa-runtime/core/inc/amd_blit_sdma.h
```diff
@@ -183,6 +183,10 @@ class BlitSdmaBase : public core::Blit {
   /// @param body_signal  When non-null (!platform_atomic_support_ classic path),
   ///   the body fences this signal to 0 on completion instead of atomic-
   ///   decrementing out_signal.  Caller allocates one per engine group.
+  /// @param dst_size_list  Optional per-entry B-side sizes for asymmetric
+  ///   LINEAR_SWAP (size_list is the A side). When null, or when
+  ///   dst_size_list[d] == size_list[d], the swap is symmetric. Asymmetric
+  ///   entries are rejected with HSA_STATUS_ERROR_INVALID_ARGUMENT.
```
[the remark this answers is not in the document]
iassiour: Fixed — the `dst_size_list` doc now states asymmetric entries are rejected with `HSA_STATUS_ERROR_INVALID_ARGUMENT`, consistent with `amd_gpu_agent.h`.

projects/clr/hipamd/src/hip_memory.cpp
```diff
@@ -3049,10 +3085,12 @@ hipError_t ihipMemcpyBatch(void** dsts, void** srcs, size_t* sizes, size_t count
...
-    const size_t actualSrcSize =
-        (copyFlags & hipMemcpyFlagExtOpIndirectSrc) ? sizeof(void*) : sizes[i];
-    const size_t actualDstSize =
-        (copyFlags & hipMemcpyFlagExtOpIndirectDst) ? sizeof(void*) : sizes[i];
+    const bool indirectSrc = (copyFlags & hipMemcpyFlagExtOpIndirectSrc) ||
+        (ops != nullptr && (ops[i] & hipExtMemcpyOpIndirectSrc));
+    const bool indirectDst = (copyFlags & hipMemcpyFlagExtOpIndirectDst) ||
+        (ops != nullptr && (ops[i] & hipExtMemcpyOpIndirectDst));
+    const size_t actualSrcSize = indirectSrc ? sizeof(void*) : sizes[i];
+    const size_t actualDstSize = indirectDst ? sizeof(void*) : sizes[i];
```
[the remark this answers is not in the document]
iassiour: Fixed — `indirectSrc`/`indirectDst` in the validation loop are now derived ops-authoritatively (`ops != nullptr ? (ops[i] & ...) : (copyFlags & ...)`), so an `ops[i]=hipExtMemcpyOpDefault` entry validates the full `sizes[i]` rather than `sizeof(void*)`. Applied at all three sites.

iassiour:
Also addressed the two low-confidence suggestions from the Copilot review (swap- and indirect-support gates): the `attrs[].flags` scan in each gate is now guarded on `ops == nullptr`, so an `ops[i]=hipExtMemcpyOpDefault` entry is no longer rejected by a swap/indirect flag that `attrs` may also carry. This keeps the gates consistent with the `ops[]`-authoritative op-type resolution used elsewhere. Added `Unit_hipExtMemcpyBatchAsync_OpsDefault_IgnoresAttrIndirectGate` to cover it.

projects/hip-tests/catch/unit/memory/hipMemcpyBatchAsync.cc
```diff
@@ -206,39 +206,39 @@ HIP_TEST_CASE(Unit_hipMemcpyBatchAsync_Negative) {
 
   SECTION("Null destination array") {
     HIP_CHECK_ERROR(hipMemcpyBatchAsync(nullptr, src_ptrs.data(), sizes.data(), kCount, nullptr,
-                                        attrs_idxs, 0, nullptr, stream_guard.stream()),
+                                        attrs_idxs, 0, stream_guard.stream()),
```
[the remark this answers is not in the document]
iassiour: done

projects/hip/include/hip/hip_runtime_api.h
```diff
@@ -6190,6 +6190,41 @@ hipError_t hipMemcpyBatchAsync(void** dsts, void** srcs, size_t* sizes, size_t c
...
+ * @param [in] stream      - Stream to execute on.
+ *
+ * @returns #hipSuccess, #hipErrorInvalidValue, #hipErrorNotSupported,
+ *          #hipErrorInvalidResourceHandle
+ */
+hipError_t hipExtMemcpyBatchAsync(void** dsts, void** srcs,
+                                  size_t* sizes, size_t* sizesDst,
+                                  hipExtMemcpyWait* waits,
+                                  hipExtMemcpySignal* signals,
+                                  size_t count,
```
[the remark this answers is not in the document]
iassiour: PR description updated to match the implemented API: no `ops` parameter (op selection is via `attrs[i].std.flags`, same `hipMemcpyFlags` as `hipMemcpyBatchAsync`), and `sizes`/`sizesDst` in place of `sizesA`/`sizesB`. There is no `hipExtMemcpyOp` type. Resolving.

projects/clr/hipamd/src/hip_memory.cpp
```diff
@@ -3148,23 +3232,52 @@ hipError_t ihipMemcpyBatch(void** dsts, void** srcs, size_t* sizes, size_t count
...
+          // Tail copy A[sizeB..sizeA-1] -> B[sizeB..sizeA-1]. In swap semantics
+          // dsts[i]=A (dstMemory) and srcs[i]=B (srcMemory), so the tail source
+          // is A (dstMemories) and the tail dest is B (srcMemories). Inherit the
+          // swap metadata's engine preference, only overriding the op type.
+          amd::CopyMetadata tailMeta = metadata;
+          tailMeta.copyOpType_ = amd::CopyMetadata::kCopyOpLinear;
+          copy_ops_by_device[device_id].emplace_back(dstMemories[i], srcMemories[i],
+                                                     dstOffsets[i] + sizeB, srcOffsets[i] + sizeB,
+                                                     sizes[i] - sizeB, tailMeta);
+        } else {
```
mangupta: 🤖 AI-assisted pre-review (for @mangupta)

**[BLOCKING]** Asymmetric-swap fallback decomposition is racy: the tail copy reads A after the swap may already have overwritten it, and both ops land in the same unordered batch

For the non-native asymmetric swap path, the code emits two `BatchCopyOp`s into the same `copy_ops_by_device[device_id]` vector: (1) a symmetric swap of the first `sizeB` bytes between A and B, and (2) a linear copy of `A[sizeB..sizeA-1] -> B[sizeB..sizeA-1]`. Nothing sequences (2) after (1), and nothing prevents them from running concurrently on different SDMA engines within the same batch submission. Worse, the two ops touch overlapping/adjacent regions of the same two buffers with no dependency, so the result is only correct if the batch happens to execute them in emission order on a single engine.

**Suggested fix:** Make the ordering explicit rather than relying on emission order: either (a) submit the head swap and the tail linear copy as two separately-ordered batch submissions on the stream (so stream ordering sequences them), or (b) restrict the decomposed pair to a single engine and document/enforce that `hsaCopyBatch`/`rocrCopyBufferBatch` preserves intra-group order for the decomposed pair. At minimum, add a comment-backed invariant plus a test that exercises a multi-entry batch mixing an asymmetric swap with other ops, which the current tests (single-entry, or two entries each asymmetric) do not cover.

projects/clr/hipamd/src/hip_memory.cpp
```diff
@@ -3148,23 +3232,52 @@ hipError_t ihipMemcpyBatch(void** dsts, void** srcs, size_t* sizes, size_t count
...
+    // For an asymmetric swap, sizesDst carries the B-side (destination) length; HW
+    // requires count_a >= count_b (sizes[i] >= sizeB).
+    size_t sizeB = 0;
+    if (sizesDst != nullptr && metadata.copyOpType_ == amd::CopyMetadata::kCopyOpSwap) {
+      sizeB = sizesDst[i];
+      if (sizeB == 0 || sizeB > sizes[i]) {
+        if (failIdx != nullptr) *failIdx = i;
+        return hipErrorInvalidValue;
+      }
+    }
```
mangupta: 🤖 AI-assisted pre-review (for @mangupta)

**[IMPORTANT]** sizesDst[] is silently ignored for non-swap entries instead of being rejected

`sizeB` is only read and validated when `metadata.copyOpType_ == amd::CopyMetadata::kCopyOpSwap`. For any entry in the batch whose attribute range does not set `hipMemcpyFlagExtOpSwap`, a caller-supplied `sizesDst[i]` — including a zero, a garbage value, or a value larger than the allocation — is read never, validated never, and discarded. A caller that passes a `sizesDst` array intending a per-entry destination length for a linear copy gets a silent full-`sizes[i]` copy with no diagnostic.

**Suggested fix:** Either reject a non-zero `sizesDst[i]` on a non-swap entry with `hipErrorInvalidValue` (setting `failIdx = i` for the legacy path), or explicitly document in the `hip_runtime_api.h` doxygen that `sizesDst[i]` is ignored for non-swap entries. Rejecting is preferable and matches the strictness already applied to unknown flag bits and non-zero `reserved[]`; add a negative test covering `sizesDst` set on a default (non-swap) entry.

projects/clr/rocclr/device/rocm/rocblit.cpp
```diff
@@ -864,19 +865,47 @@ bool DmaBlitManager::rocrCopyBufferBatch(const std::vector<hsa_amd_memory_copy_o
...
+          swap.dst = swapPending.dsts[i];
+          swap.dst_agent = swapPending.dst_agents[i];
+          swap.src_size = swapPending.sizes[i];
+          swap.dst_size = swapPending.dst_sizes[i];
+          swap.num_entries = 0;
+          finalOps.push_back(swap);
+        }
+      } else {
+        // Symmetric: batch into multi-entry op
+        multiStore.push_back(std::move(swapPending));
```
mangupta: 🤖 AI-assisted pre-review (for @mangupta)

**[IMPORTANT]** rocrCopyBufferBatch's new asymmetric-swap scalar-op path is unreachable-or-failing dead code given the ROCr side unconditionally rejects asymmetric swaps

`rocrCopyBufferBatch` now detects `has_asymmetric` and emits per-entry scalar swap ops (`num_entries = 0`) carrying distinct `src_size`/`dst_size`. But on the ROCr side, `BlitSdma::SubmitBodies` rejects any asymmetric entry with `HSA_STATUS_ERROR_INVALID_ARGUMENT` whenever `dst_size_list != nullptr && !NativeAsymmetricSwapSupported()`, and `NativeAsymmetricSwapSupported()` is hardcoded `return false`. So if an asymmetric swap ever reaches this branch, the batch fails with an opaque HSA error rather than producing a copy — while the CLR layer above believes it already decomposed all asymmetric swaps and will not have produced one.

**Suggested fix:** Tie the two capability predicates to one source of truth: derive `Settings::sdma_asymmetric_swap_supported_` from the ROCr-reported capability (or a shared ISA check) rather than an independent literal `false`, so flipping native support cannot leave the CLR decomposition and the ROCr rejection out of step. Alternatively, drop the `has_asymmetric` branch in `rocrCopyBufferBatch` for now (leaving the pre-existing assert or an explicit early error) until native support actually lands, so the PR does not ship an untestable path.

projects/clr/hipamd/src/hip_table_interface.cpp
```diff
@@ -3059,6 +3059,19 @@ HIP_PUBLIC_API hipError_t hipMemcpyBatchAsync(void** dsts, void** srcs, size_t*
...
+                                  hipExtMemcpySignal* signals,
+                                  size_t count,
+                                  hipExtMemcpyAttributes* attrs, size_t* attrsIdxs, size_t numAttrs,
+                                  hipStream_t stream) {
+  TRY;
+  return hip::GetHipDispatchTable()->hipExtMemcpyBatchAsync_fn(
+      dsts, srcs, sizes, sizesDst, waits, signals,
+      count, attrs, attrsIdxs, numAttrs, stream);
+  CATCH;
+}
```
mangupta: 🤖 AI-assisted pre-review (for @mangupta)

**[IMPORTANT]** hipExtMemcpyBatchAsync table thunk is missing the HIP_PUBLIC_API visibility attribute used by every neighboring entry point

The new `hipExtMemcpyBatchAsync` definition in `hip_table_interface.cpp` is declared as a bare `hipError_t hipExtMemcpyBatchAsync(...)`, while the immediately adjacent `hipMemcpyBatchAsync` and `hipMemcpy3DBatchAsync` definitions both carry `HIP_PUBLIC_API`. Without that attribute the symbol may not receive the default-visibility/export annotation its neighbors get, so the exported symbol can fail to materialize (or materialize with the wrong visibility) even though it is listed in `amdhip.def.in` and `hip_hcc.map.in`.

**Suggested fix:** Prefix the definition with `HIP_PUBLIC_API` to match `hipMemcpyBatchAsync` and `hipMemcpy3DBatchAsync` in the same file, so the exported symbol's visibility matches the `amdhip.def.in` / `hip_hcc.map.in` entries added alongside it.

projects/hip/include/hip/driver_types.h
```diff
@@ -458,14 +458,66 @@ typedef struct hipMemLocation {
...
+ */
+typedef enum hipExtMemcpySignalOp {
+  hipExtMemcpySignalNone  = 0,  ///< No signal; addr must be NULL
+  hipExtMemcpySignalWrite = 1,  ///< *addr  = data
+  hipExtMemcpySignalAdd   = 2,  ///< *addr += data
+  hipExtMemcpySignalSub   = 3,  ///< *addr -= data
+} hipExtMemcpySignalOp;
+
+/**
+ * Per-entry signal-after-copy for hipExtMemcpyBatchAsync. Reserved for future use.
```
mangupta: 🤖 AI-assisted pre-review (for @mangupta)

**[IMPORTANT]** hipExtMemcpyWait / hipExtMemcpySignal document a fixed 64-byte ABI but carry no static_assert to enforce it

Both new public structs carry a trailing comment asserting `// 64 bytes; zero-initialize. Future fields carve out of the reserved space so the size/ABI stays fixed.` — that is the entire stated justification for the padding commit (`pad reserved wait/signal structs to 64 bytes`). But nothing in the build enforces the size. `hipExtMemcpyAttributes` gets `static_assert(sizeof(hipExtMemcpyAttributes) == 56, ...)` in `hip_memory.cpp`; the wait and signal structs get no equivalent, so a future field addition or a platform with different `void*`/alignment rules can silently change the size that the comment promises is fixed.

**Suggested fix:** Add `static_assert(sizeof(hipExtMemcpyWait) == 64, ...)` and `static_assert(sizeof(hipExtMemcpySignal) == 64, ...)` alongside the existing `hipExtMemcpyAttributes` assert in `hip_memory.cpp`, so all three reserved-tail structs are locked by the same mechanism.

projects/clr/hipamd/src/hip_memory.cpp
```diff
@@ -3247,70 +3360,53 @@ hipError_t hipMemcpyBatchAsync(void** dsts, void** srcs, size_t* sizes, size_t c
...
-  // Validate individual sizes and find first error
-  for (size_t i = 0; i < count; ++i) {
-    if (sizes[i] == 0) {
-      if (failIdx != nullptr) *failIdx = i;
-      HIP_RETURN(hipErrorInvalidValue);
+      extAttrs[i].srcAccessOrder = attrs[i].srcAccessOrder;
+      extAttrs[i].srcLocHint = attrs[i].srcLocHint;
+      extAttrs[i].dstLocHint = attrs[i].dstLocHint;
+      extAttrs[i].flags = attrs[i].flags;
     }
```
mangupta: 🤖 AI-assisted pre-review (for @mangupta)

**[IMPORTANT]** Legacy hipMemcpyBatchAsync wrapper converts attrs only when attrs != nullptr, so a numAttrs>0 with attrs==nullptr loses the pointer before validation

The legacy wrapper guards the conversion on `if (attrs != nullptr && numAttrs > 0)`. When `attrs == nullptr` but `numAttrs > 0`, `extAttrsPtr` stays `nullptr` and `numAttrs` is forwarded unchanged to `ihipMemcpyBatch`, which then evaluates `if (numAttrs > 0) { if (attrs == nullptr || attrsIdxs == nullptr) return hipErrorInvalidValue; }`. That still returns the right code today, but only by coincidence of two separate nullptr conditions lining up. More importantly the conversion loop silently changes what the callee sees: the callee can no longer distinguish "caller passed no attrs" from "caller passed attrs that the wrapper declined to convert", and any future relaxation of the callee's nullptr check would turn this into a forwarded-but-empty attribute set.

**Suggested fix:** Validate `attrs`/`attrsIdxs` non-null and `numAttrs <= count` in the wrapper before the conversion loop (or hoist the whole attribute-shape validation into a small helper called by both entry points), so `attrs[i]` is never read for an index the callee would have rejected.

mangupta (commented):

projects/clr/hipamd/src/hip_memory.cpp
iassiour: I don't believe there's a data hazard. The decomposition splits exactly at sizeB: the head op is a symmetric swap of [0, sizeB) of A and B, and the tail op copies A[sizeB, sizeA) → B[sizeB, sizeA). The two ops touch complementary, non-overlapping byte ranges — the tail reads the part of A the head-swap never writes, and writes the part of B the head-swap never writes. So there's no RAW/WAW between them and the result is identical regardless of execution order or concurrency; the "tail reads A after the swap overwrote it" case can't occur because the swap only touches [0, sizeB)

projects/clr/hipamd/src/hip_memory.cpp
iassiour: Fixed. A non-zero sizesDst[i] on a non-swap entry now returns hipErrorInvalidValue (setting failIdx = i on the legacy path), matching the strictness already applied to unknown flag bits and non-zero reserved[]. Added Unit_hipExtMemcpyBatchAsync_SizesDst_NonSwap_Negative covering it.

projects/clr/hipamd/src/hip_table_interface.cpp
iassiour: Added HIP_PUBLIC_API to the hipExtMemcpyBatchAsync thunk to match hipMemcpyBatchAsync / hipMemcpy3DBatchAsync and the amdhip.def.in / hip_hcc.map.in entries added alongside it.

projects/hip/include/hip/driver_types.h
iassiour: Added static_assert(sizeof(hipExtMemcpyWait) == 64) and static_assert(sizeof(hipExtMemcpySignal) == 64) next to the existing hipExtMemcpyAttributes assert in hip_memory.cpp, so all three reserved-tail
structs are locked by the same mechanism.

PR #9099, opened 2026-07-23
[rocjitsu] Make ISA targets statically pluggable

kuhar:
## Motivation

Make ISA targets statically pluggable and simplify integration of downstream NPI targets without adding them to central target switches.

## Technical Details

- Add a constexpr descriptor to each ISA that owns its names, public and ELF bindings, decoder factory, and execution capability.
- Compose one validated immutable registry per final image from its selected provider headers, and use it for decoder creation and code-object lookup.
- Add canonical-name decoder creation with recoverable unavailable-target errors.
- Remove fine-grained gfx1250 execution registration markers and replace them with one generated immutable callback table while preserving model-only decoding.

## Issue Tracking

N/A

## Test Plan

Run ctest and pytest.

## Test Result

Passed.

## Submission Checklist


amd-arosa (approved):
I sent you two patches to align with the style guide.
There might be something funny happening with `create_decoder_for_target`, but it was pre-existing.
but other than that LGTM.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp
```diff
@@ -466,6 +466,7 @@ class BlockState {
     return base;
   };
 
+  // \NPI new ISA family: classify its scalar PC instruction encodings here.
```
amd-eochoalo, line 469: Thanks! I'm thinking there should be a way to generate these from isa_properties too.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/patch/instruction_builder.cpp
```diff
@@ -26,6 +26,7 @@ enum class ScalarSop2Op {
 /// architecture-specific opcode mapping lives with the instruction packers
 /// instead of the static recovery matcher.
 [[nodiscard]] std::optional<uint8_t> scalar_sop2_opcode(rj_code_arch_t arch, ScalarSop2Op op) {
+  // \NPI new ISA family: classify its scalar SOP2 opcode mapping here.
```
amd-eochoalo, line 29: Same

amd-eochoalo (approved):
I think this is good. Not sure if it would be interesting / good to link kTargetDescriptor to isa_properties somehow in the future (but probably not worth while on this PR at the moment).

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp
kuhar: I don't want to overcomplicate this, debugging these code generators is not fun. We can keep iterating on this.

amd-eochoalo: These ones should be fairly simple to do. And it would consolidate all architecture-based branching or special values into the python profiles. I'll work on a PR once this is merged.

atgutier (approved):
Approved with a minors from the agent:

 ## MINOR

  [MINOR] emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/isa/arch/amdgpu/
  gfx1250/CMakeLists.txt:71 — Full gfx1250 provider drops its object
  libraries’ link requirements

  Reasoning: rocjitsu_isa_gfx1250 exposes the model and execution objects
  through $<TARGET_OBJECTS:...>, but it does not link or otherwise propagate
  the underlying object-library targets. Those targets receive util and
  simdojo_headers requirements through rj_add_object_library().

  This differs from normal object-library providers: the registry helper both
  includes their objects and links the provider target so its usage
  requirements propagate. Because the full gfx1250 provider is an
  INTERFACE_LIBRARY, the registry helper cannot discover the underlying model
  and execution targets. Current broad in-tree consumers independently link
  the missing dependencies, masking the composition gap. A narrow or
  downstream consumer can therefore depend on undocumented transitive setup
  or fail to link as these object libraries acquire additional non-header-
  only dependencies.

  Evidence:

  - gfx1250/CMakeLists.txt:35-64 creates the model and execution object
    libraries through rj_add_object_library().

  - gfx1250/CMakeLists.txt:71-77 forwards only their object files.
  - cmake/rj_add_object_library.cmake:27-30 attaches common dependencies to
    each object library.

  - cmake/rj_add_object_library.cmake:101-104 extracts objects only when the
    registered provider is itself an object library.

  - cmake/rj_add_object_library.cmake:134-139 republishes the full interface
    provider, whose interface currently contains no link dependencies.

  - docs/isa-target-providers.md:66-87 presents the provider/registry
    combination as the consumer-facing composition boundary.

  Suggested fix: retain the explicit $<TARGET_OBJECTS:...> entries, but
  propagate both underlying targets’ usage requirements:

 target_link_libraries(
      rocjitsu_isa_gfx1250
      INTERFACE
          rocjitsu_isa_gfx1250_model
          rocjitsu_isa_gfx1250_exec
  )

  Alternatively, extend the provider helper to record and forward
  implementation dependencies. Add a narrow final-image link test that
  consumes the full gfx1250 registry without also inheriting the simulator’s
  broad dependency list.

  Confidence: medium-high. The usage-requirement propagation gap is definite;
  existing production targets currently mask it, so I did not reproduce an
  in-tree link failure.

PR #9100, opened 2026-07-23
fix(clr): honor NoCU for P2P copies

saleelk:
- Preserve the SDMA requirement through peer-copy submission and prevent compute fallback when NoCU is requested.

## Issue Tracking

JIRA ID: AIRUNTIME-0

## Test Result

No test needed

## Submission Checklist

### DIFF
--- a/projects/clr/rocclr/device/rocm/rocblit.cpp
+++ b/projects/clr/rocclr/device/rocm/rocblit.cpp
@@ -490,8 +490,9 @@ inline bool DmaBlitManager::rocrCopyBuffer(address dst, hsa_agent_t& dstAgent, c
   uint32_t copyMask = 0;
   bool kUseRegularCopyApi = false;
   constexpr size_t kRetainCountThreshold = 8;
-  bool forceSDMA =
+  const bool requireSDMA =
       (copyMetadata.copyEnginePreference_ == amd::CopyMetadata::CopyEnginePreference::SDMA);
+  bool forceSDMA = requireSDMA;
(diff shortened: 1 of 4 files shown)


projects/clr/rocclr/device/rocm/rocblit.cpp
```diff
@@ -490,8 +490,9 @@ inline bool DmaBlitManager::rocrCopyBuffer(address dst, hsa_agent_t& dstAgent, c
   uint32_t copyMask = 0;
   bool kUseRegularCopyApi = false;
   constexpr size_t kRetainCountThreshold = 8;
-  bool forceSDMA =
+  const bool requireSDMA =
       (copyMetadata.copyEnginePreference_ == amd::CopyMetadata::CopyEnginePreference::SDMA);
```
AlexeySachkov, line 494: `copyEnginePreference_` seems to be less of a preference and more of a requirement now :)

saleelk: Right, a better choice of var name would suffice, it was initially a hint but later used as a mechanism to convey hipDevicetoDeviceNoCU flag, and that ruined it.

saleelk: Ideally there shoudlnt be a separate path for P2P at all. Memory needs to be treated as uniform. and One needs to make useof the Object class to determine the behavior.. But a larger refactor is to be slated for another day...

gandryey (approved):

chrispaquot (approved):

PR #9105, opened 2026-07-23
fix(rocprofiler-sdk): cmake flags for ATT

ApoKalipse-V:
## Motivation

A cmake flag was introduced to add/remove quick_scan support for rocprofv3 tool. This flag requires users to manually disable quick_scan in older rocm versions to avoid configure/compiler errors.
This PR sets the default flag value depending on rocm and rocprof-trace-decoder version. The option is still available to forcibly enable/disable quick_scan as needed.

Additionally, the DISPATCH / EVENTs flag was removed as it's no longer needed as feature flag.

JIRA ID : AIPROFSDK-1002

## Submission Checklist


projects/rocprofiler-sdk/cmake/rocprofiler_config_interfaces.cmake
```diff
@@ -134,6 +134,14 @@ endif()
 target_link_libraries(rocprofiler-sdk-hip INTERFACE hip::host)
 rocprofiler_config_nolink_target(rocprofiler-sdk-hip-nolink hip::host)
 
+# ----------------------------------------------------------------------------------------#
+#
+# Advanced thread trace decoder
+#
+# ----------------------------------------------------------------------------------------#
+
+find_package(rocprof-trace-decoder 0.2.1 QUIET CONFIG)
```
jrmadsen: I'd remove the QUIET so you can see when it was not found

jrmadsen (changes requested):

projects/rocprofiler-sdk/cmake/rocprofiler_config_interfaces.cmake
ApoKalipse-V:
```suggestion
find_package(rocprof-trace-decoder 0.2.1 CONFIG)
```

jrmadsen (approved):

bgopesh (approved):

lumachad:
I'm seeing the following in https://github.com/ROCm/rocm-systems/actions/runs/30863484107/job/91948959785?pr=9105. I'm not sure if that's something we should worry about or not.

```
CMake Warning at cmake/rocprofiler_config_interfaces.cmake:143 (find_package):
  Could not find a configuration file for package "rocprof-trace-decoder"
  that is compatible with requested version "0.2.1".

  The following configuration files were considered but not accepted:

    /opt/rocm/lib/cmake/rocprof-trace-decoder/rocprof-trace-decoder-config.cmake, version: 0.2.0

Call Stack (most recent call first):
  CMakeLists.txt:109 (include)


-- Found PkgConfig: /usr/bin/pkg-config (found version "0.29.2") 
```

Since I'm seeing the PR touching 0.2.1, I thought I'd mention.

ApoKalipse-V:
@lumachad , the code coverage workflow does not build the trace decoder currently, which is why you see that failure. It shouldn't be fatal - the side effect is disabling att quick scan and it's tests. This is what that explicit variable DISABLE_ATT_QUICK_SCAN used to do.
Once the docker updates, it'll work like the other workflows (or we can add the decoder to code coverage)

jbonnell-amd (approved):
Approved for `.github/workflows` changes

lumachad:
It seems @jbonnell-amd approved it already. So I'll skip it.

PR #9106, opened 2026-07-23
ci(rocprofiler-compute): fix ROCm version mismatch in sanitizer workflow

xuchen-amd:
## Motivation

The rocprofiler-compute sanitizer CI (asan/tsan/ubsan) was failing at the
"Locate clang sanitizer runtime" pre-flight check before any tests ran.

## Technical Details

The "Install latest nightly ROCm" step extracted the tarball's actual version
into `tarball_rocm_version` but then ignored it, installing to
`${ROCM_PATH}-${ROCM_VERSION}` (i.e. `/opt/rocm-7.0.0`) instead. The nightly
tarball is ROCm 7.15.0, so `clang -print-resource-dir` resolved to a path
under 7.15.0 that did not exist under the 7.0.0 directory, causing the
sanitizer runtime directory check to fail with exit code 1.

Fix: use `tarball_rocm_version` for both the directory name and symlink target.

## Issue Tracking

JIRA ID: N/A

## Test Plan

Re-run the asan/tsan/ubsan CI jobs on this branch.

## Test Result

Pending CI.

## Submission Checklist

### DIFF
--- a/.github/workflows/rocprofiler-compute-sanitizers.yml
+++ b/.github/workflows/rocprofiler-compute-sanitizers.yml
@@ -127,19 +127,12 @@ jobs:
       - name: Output TheRock manifest
         run: cat "${ROCM_PATH}/share/therock/therock_manifest.json"
 
-      - name: Inspect clang resource dir
-        shell: bash
-        run: |
-          resource_dir="$("${ROCM_PATH}/llvm/bin/clang" -print-resource-dir)"
-          echo "resource_dir=${resource_dir}"
-          find "${resource_dir}/lib" -maxdepth 2 | sort || true
-
       - name: Locate clang sanitizer runtime
         id: sanitizer_runtime
         shell: bash
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/rocprofiler-compute-sanitizers.yml
jbonnell-amd: We actually removed the amdgpu install from `rocprofiler-compute-continuous-integration.yml`, we could probably do the same here as well (see that workflow file, we still needed a few package dependencies)

jbonnell-amd: Could remove this as well if removing the amdgpu install

xuchen-amd: revmoved

xuchen-amd: removed

jbonnell-amd (approved):

PR #9107, opened 2026-07-23
fix(rocr-runtime) Fix hardcoded 4KB page size for IPC signal and memory sharing

benrichard-amd:
## Motivation

Part of https://github.com/ROCm/TheRock/issues/5518

The ROCR runtime hardcodes `4096` as the page size in three locations. On architectures with larger page sizes, such as ppc64le (POWER) where the system page size is 64KB (65536), this causes IPC signal creation and IPC memory sharing to fail:

1. `hsa_amd_ipc_signal_create()` returns `HSA_STATUS_ERROR_INVALID_ARGUMENT` because `PageAllocator` allocates IPC signal memory in 4KB chunks, but the thunk layer expects page-aligned allocations.

2. `hsa_amd_ipc_memory_attach()` maps to the wrong offset for fragmented allocations because the fragment offset is stored in units of pages but decoded by multiplying by the hardcoded value 4096 instead of the actual page size, causing data validation failures in the child process.

These failures cause the `rocrtst` IPC test to fail on ppc64le, along with any application relying on cross-process GPU memory or signal sharing.

## Technical Details

Replace hardcoded `4096` with `os::PageSize()` in three locations:

- **`core/common/shared.h`**: `PageAllocator::alloc()` uses `os::PageSize()` for allocation size and alignment instead of hardcoded 4096. Added `#include "core/util/os.h"`.

- **`core/runtime/ipc_signal.cpp`**: `IPCSignal::CreateHandle()` passes `AlignUp(sizeof(SharedSignal), os::PageSize())` as the IPC region size instead of hardcoded 4096. Added `#include "core/util/os.h"`.

- **`core/runtime/runtime.cpp`**: `Runtime::IPCAttach()` decodes the fragment offset as `(handle[6] & 0x1FF) * os::PageSize()` instead of `(handle[6] & 0x1FF) * 4096`.

`os::PageSize()` calls `sysconf(_SC_PAGESIZE)` once and caches the result. On x86 this returns 4096 so the change is a no-op. On ppc64le it returns 65536, fixing the IPC failures.

## Test Plan

- Run `rocrtst` quick suite (16 tests) on x86 (MI210) to verify no regressions
- Run `rocrtst` quick suite (16 tests) on ppc64le (MI210) to verify the fix
- Run standalone IPC signal creation test (`hsa_amd_ipc_signal_create`) on both architectures
- Run standalone IPC memory sharing test (fork, parent fills GPU buffer, child reads via IPC attach) on ppc64le

## Test Result

**x86 (8x MI210)**: 16/16 rocrtst quick suite tests passed, including IPC test.

**ppc64le (2x MI210)**: 16/16 rocrtst quick suite tests passed, including IPC test. Before this fix, the IPC test failed with `HSA_STATUS_ERROR_INVALID_ARGUMENT` (4097) on signal creation and data validation errors on memory sharing.

## Submission Checklist

- [*] Look over the contributing guidelines at https://github.com/ROCm/rocm-systems/blob/develop/CONTRIBUTING.md.


cfreeamd (commented):

Reviewed the ROCr changes (replace hardcoded 4096 with os::PageSize() for IPC signal + memory sharing). The `PageAllocator::alloc` change in `shared.h` looks correct and self-consistent, and `os::PageSize()` is a cached sysconf so the added calls are cheap. Inline comments focus on the two IPC changes, each of which updates only one half of a paired operation (fragment-offset encode/decode, and signal create/attach size) while leaving the matching half hardcoded to 4096 — no-ops on 4KB systems but broken on the 64KB-page systems this PR targets.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
```diff
@@ -1780,7 +1780,7 @@ hsa_status_t Runtime::IPCAttach(const hsa_amd_ipc_memory_t* handle, size_t len,
 
   if ((importHandle.handle[6] & 0x80000000) != 0) {
     isFragment = true;
-    fragOffset = (importHandle.handle[6] & 0x1FF) * 4096;
+    fragOffset = (importHandle.handle[6] & 0x1FF) * os::PageSize();
```
cfreeamd, line 1793: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** This decodes the packed fragment offset with `* os::PageSize()`, but the matching **encode** side in `IPCCreate` is unchanged and still divides by a hardcoded `static const size_t pageSize = 4096;`:

```cpp
// IPCCreate (not in this diff):
static const size_t pageSize = 4096;
uint32_t fragOffset = (ptr - block.base) / pageSize;   // packs offset/4096
handle->handle[6] |= 0x80000000 | fragOffset;
```

On a system where `os::PageSize() != 4096` (e.g. 64KB) the exporter packs `offset/4096` into the 9-bit field while this importer multiplies by 65536 → fragment offset **16× too large**, so `importAddress + fragOffset` (and `len = Min(len, importSize - fragOffset)`) lands outside the intended fragment (wrong address / OOB). Both processes run the same runtime, so this is a pure internal encode/decode disagreement. Fix: use the same unit on both sides — either change the `IPCCreate` `pageSize` constant to `os::PageSize()` (it also feeds `assert(dmabufOffset/pageSize == fragOffset ...)`), or keep the fragment packing on a fixed 4KB granularity on both sides. Changing only the decode is the bug.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/ipc_signal.cpp
```diff
@@ -72,7 +73,7 @@ void IPCSignal::CreateHandle(Signal* signal, hsa_amd_ipc_signal_t* ipc_handle) {
   if (!signal->isIPC())
     throw AMD::hsa_exception(HSA_STATUS_ERROR_INVALID_ARGUMENT, "Signal must be IPC enabled.");
   SharedSignal* shared = SharedSignal::Convert(Convert(signal));
-  hsa_status_t err = Runtime::runtime_singleton_->IPCCreate(shared, 4096, ipc_handle);
+  hsa_status_t err = Runtime::runtime_singleton_->IPCCreate(shared, AlignUp(sizeof(SharedSignal), os::PageSize()), ipc_handle);
```
cfreeamd, line 76: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Important]** The create side now sizes the region as `AlignUp(sizeof(SharedSignal), os::PageSize())`, but the paired **attach** side in `ipc_signal.h` was not updated and still passes a fixed `4096`:

```cpp
// ipc_signal.h (not in this diff):
explicit SharedMemorySignal(const hsa_amd_ipc_memory_t* handle) : signal_(handle, 4096) {}
//   → IPCAttach(handle, /*len=*/4096, ...)
```

On a 64KB-page system the exporter creates a 64KB region while the importer passes `len = 4096` into `IPCAttach` (which clamps via `len = Min(len, importSize - fragOffset)`). For consistency the attach `len` should mirror the create size — `AlignUp(sizeof(SharedSignal), os::PageSize())` — so both halves agree. Leaving one at `4096` reintroduces the hardcode this PR is removing.

benrichard-amd: Fixed. Updated SharedMemorySignal constructor in ipc_signal.h to use AlignUp(sizeof(SharedSignal), os::PageSize()) instead of hardcoded 4096.

  All 4 files now consistently use os::PageSize(). Tested on both x86 (MI210 x8, 16/16 rocrtst pass) and ppc64le (MI210 x2, 16/16
  rocrtst pass).

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp
benrichard-amd: The upstream encode side in IPCCreate already uses os::PageSize() (line 1457: const size_t pageSize = os::PageSize() and line 1475: fragOffset = (ptr - block.base)/pageSize), so both sides are now consistent. The only hardcoded 4096 was on the decode side at line 1783, which this patch fixes.

projects/rocr-runtime/runtime/hsa-runtime/core/inc/ipc_signal.h
```diff
@@ -69,7 +69,7 @@ class SharedMemory {
 /// @brief Container for ipc signal abi block.
 class SharedMemorySignal {
  public:
-  explicit SharedMemorySignal(const hsa_amd_ipc_memory_t* handle) : signal_(handle, 4096) {
+  explicit SharedMemorySignal(const hsa_amd_ipc_memory_t* handle) : signal_(handle, AlignUp(sizeof(SharedSignal), os::PageSize())) {
```
cfreeamd, line 74: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** This constructor now uses `AlignUp`, `os::PageSize()`, and `sizeof(SharedSignal)`, but `ipc_signal.h` doesn't directly include the headers that declare them (`core/util/utils.h` for `AlignUp`, `core/util/os.h` for `os::PageSize`). It compiles today via transitive includes, but adding the explicit `#include`s here would make the header self-contained and robust against future include-graph changes. Non-blocking.

cfreeamd (commented):

Both prior [Important] items are resolved:

1. **Signal create/attach size asymmetry** — fixed: the `SharedMemorySignal` ctor in `ipc_signal.h` now uses `AlignUp(sizeof(SharedSignal), os::PageSize())`, matching the create side. Both halves agree.
2. **Fragment-offset encode/decode mismatch** — this was a false positive in my original review. You're correct that `IPCCreate` in current `develop` already uses `os::PageSize()` for the encode side (runtime.cpp ~L1457/L1475); the only hardcoded `4096` was the decode side, which this PR fixes. My first pass compared against a stale local checkout. Apologies for the noise. Encode/decode are consistent.

Testing on ppc64le (64KB pages) — the actual target platform — with rocrtst passing is exactly the validation this change needed. Looks good.

projects/rocr-runtime/runtime/hsa-runtime/core/inc/ipc_signal.h
benrichard-amd: Fixed

cfreeamd (commented):
🤖 AI-assisted pre-review (for @cfreeamd) — follow-up

The [Optional] include suggestion is now addressed: ipc_signal.h explicitly includes core/util/os.h and core/util/utils.h for the newly-used os::PageSize() and AlignUp(). All prior AI pre-review findings (both [Important] and the [Optional]) are now resolved. No remaining concerns.

cfreeamd (approved):

dayatsin-amd (approved):

PR #9108, opened 2026-07-23
ci(rocjitsu): Enable TSan on remaining tests

### LINKED ISSUE #8922: [rocjitsu] Enable skipped TSan tests on CI
The following TSan regex is used to exclude when running unit tests:

```
            "MatmulStressTest\.TiledAllCU"
            "MatmulStressTest\.MfmaAllCUs"
            "VectorAddStressTest\.AllCUsGoldenReference"
            "Gfx1250SimulationTest\.*"
            "Gfx1250ExecutionTest\.*"
            "InterposerDupTest.SerializedReopenUnderContentionStaysRoutable"
```

These were added after running locally with TSan lead to messages of TSan finding deadlocks and the programming hanging.

lialan:
Closes #8922

## Summary

- Consolidate rocjitsu TSan CI into one blocking shared-runtime `tsan` job.
- Expand shared TSan coverage from the broad exclusions on the `develop` branch to the main rocjitsu test suite.
- Keep the 24 expensive MFMA/WMMA exact checks disabled under TSan to control runtime.
- Retain only two reproduced TSan-specific exclusions: one load-sensitive daemon timing test and one shared-runtime nested-process test.
- Replace the fairness-unspecified `std::shared_mutex` used for cache access serialization with a writer-preference access gate.
- Once `MemorySideCache::flush_all()` is waiting, block new cache accesses so active accesses drain and the flush makes progress while retaining per-stripe concurrency for ordinary reads and writes.
- Cover concurrent cache reads, writes, and flushes, including a deterministic writer-preference test and flush progress under continuous reader traffic.

## Coverage

Compared with the `develop` TSan policy on this head, the shared TSan job newly covers 111 previously excluded tests:

- 60 `Gfx1250SimulationTest` cases.
- 44 `Gfx1250ExecutionTest` cases.
- Four matmul stress cases.
- Two vector-add stress cases.
- One interposer contention case.

The PR also adds `WriterPreferredAccessGateTest.WaitingWriterBlocksLateSharedEntrant`, bringing the consolidated shared TSan selection to 1,859 tests. Removing the static sparse-memory job does not reduce unique test-case coverage because its SparseMemory cases are already included in the shared suite.

The remaining differences from the Clang ASan/UBSan selection are:

- 24 expensive MFMA/WMMA exact checks, which passed separately.
- `DaemonApi.FullListenerQueueDoesNotBlockOrRemoveSocket`, whose five-second assertion is load-sensitive at TSan CI parallelism.
- `RocjitsuCliDaemon.LaunchesApplicationAfterDaemonIsReady`, whose nested HIP process receives `EPERM` from `process_vm_readv` under the shared TSan runtime and times out.

## Validation

- The shared TSan configuration passed 1,858 tests with zero failures before the deterministic gate test was added.
- Focused release and shared-TSan runs passed all three cache/gate regressions.
- The shared-TSan continuous-reader progress and deterministic gate tests each passed 5/5 repeated runs.
- The continuous-reader test took 6.37-6.48 seconds under shared TSan, validating the 60-second starvation watchdog margin.
- The previous head also passed release, Clang ASan/UBSan, GCC ASan/UBSan, pre-commit, and TheRock CI.
- The 24 expensive MFMA/WMMA exact tests passed separately.
- Pre-commit, formatting, and diff checks passed.

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/memory_side_cache.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/memory_side_cache.h
@@ -12,6 +12,7 @@
 #include <cstdint>
 #include <memory>
 #include <mutex>
+#include <shared_mutex>
 #include <string>
 #include <utility>
 
@@ -109,7 +110,7 @@ private:
   }
 
   mutable std::array<std::mutex, STRIPE_COUNT> stripes_;
-  mutable std::mutex flush_mutex_; ///< Exclusive lock for flush_all (must not race with stripes).
+  mutable std::shared_mutex access_mutex_;
   CacheStore cache_;
   simdojo::Port *req_ = nullptr;
   std::vector<simdojo::Port *> cpl_ports_;
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/memory_side_cache.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/memory_side_cache.h
@@ -19,6 +19,30 @@
 namespace rocjitsu {
 namespace amdgpu {
 
+class WriterPreferredAccessGateTestAccess;
+
+/// @brief Shared access gate that gives queued exclusive owners priority.
+///
+/// @details This type is internal to the memory-side cache implementation but
+/// is kept separate so its writer-preference invariant can be tested directly.
+class WriterPreferredAccessGate {
+public:
+  void lock_shared();
+  bool try_lock_shared();
+  void unlock_shared();
+  void lock();
+  void unlock();
+
+private:
+  friend class WriterPreferredAccessGateTestAccess;
+
+  std::mutex mutex_;
+  std::condition_variable cv_;
+  uint32_t active_readers_ = 0;
+  uint32_t waiting_writers_ = 0;
+  bool writer_active_ = false;
+};
+
 /// @brief Memory-side cache component sitting between L2 and HBM on each IOD.
 ///
 /// @details Write-through, write-allocate cache, and no mtype awareness. All traffic from
@@ -101,21 +125,6 @@ public:
   }
 
 private:
-  class WriterPreferredAccessGate {
-  public:
-    void lock_shared();
-    void unlock_shared();
-    void lock();
-    void unlock();
-
-  private:
-    std::mutex mutex_;
-    std::condition_variable cv_;
-    uint32_t active_readers_ = 0;
-    uint32_t waiting_writers_ = 0;
-    bool writer_active_ = false;
-  };
-
   void ensure_line(uint64_t addr, uint32_t vmid);
 
   /// @brief Send a read or write request to the backing store via the req port.
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/memory_side_cache.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/memory_side_cache.h
@@ -9,10 +9,10 @@
 #include "simdojo/sim/message.h"
 
 #include <array>
+#include <condition_variable>
 #include <cstdint>
 #include <memory>
 #include <mutex>
-#include <shared_mutex>
 #include <string>
 #include <utility>
 
@@ -32,10 +32,13 @@ namespace amdgpu {
 /// through the requester port (req), which is connected to the HBM controller via a link.
 ///
 /// @par Thread safety
-/// All public methods are thread-safe. Striped locking (by cache set index)
-/// serializes access because multiple XCDs assigned to the same IOD may share
-/// this MSC from different partition threads. Stripes allow concurrent access
-/// to different cache sets, eliminating contention for non-overlapping addresses.
+/// @c read(), @c write(), and @c flush_all() are thread-safe. Reads and writes
+/// acquire the writer-preference access gate in shared mode before acquiring a
+/// stripe mutex by cache set index. A waiting flush blocks new shared entrants,
+/// then acquires the gate exclusively after active accesses finish. This
+/// guarantees flush progress while preserving concurrent access to different
+/// cache sets. The required lock order is access gate before stripe; no path
+/// may acquire them in the opposite order.
 class MemorySideCache : public simdojo::Component {
 public:
   static constexpr uint32_t LINE_SIZE_BITS = 7; // 128 bytes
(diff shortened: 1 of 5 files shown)


lialan:
let's see how much extra time it need to enable broadened tsan tests. I excluded expensive tests.

emulation/rocjitsu/tests/gfx1250_sim_test.cpp
```diff
@@ -5107,4 +5111,90 @@ TEST(Gfx1250SimulationTest, BufferStoreUsesM0Soffset) {
...
+  while (completed_reads.load(std::memory_order_relaxed) < reader_count)
+    std::this_thread::yield();
+
+  std::atomic<bool> flush_completed = false;
+  std::thread flusher([&] {
+    msc->flush_all();
+    flush_completed.store(true, std::memory_order_release);
+  });
+
+  const auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds(5);
```
kuhar: This would be safer if the starvation watchdog had a much wider margin, such as 60 seconds. In three isolated runs of the TSan binary, this test took 4.38s, 3.84s, and 4.75s, leaving almost no headroom against the current 5s deadline before CI-level parallel load is added. This job is now blocking, and the workflow already excludes another test specifically because its 5s assertion is load-sensitive under TSan. A generous watchdog still catches a true starvation regression without making machine speed the pass/fail signal.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/memory_side_cache.h
```diff
@@ -97,6 +101,21 @@ class MemorySideCache : public simdojo::Component {
   }
 
 private:
+  class WriterPreferredAccessGate {
```
kuhar: I think it would help to test writer preference directly with a controlled handshake. The current integration coverage does not isolate that invariant: `MemorySideCacheFlushMakesProgressUnderContinuousReads` only checks whether a flush finishes before a wall-clock deadline, while `MemorySideCacheFlushSerializesConcurrentAccess` only bounds a read value and checks backing memory after a final single-threaded flush. Could we expose or extract this gate enough to assert deterministically that a waiting writer blocks a later shared entrant and proceeds as soon as the last active reader exits?

kuhar (commented):
A few comments

emulation/rocjitsu/tests/gfx1250_sim_test.cpp
```diff
@@ -5107,4 +5111,90 @@ TEST(Gfx1250SimulationTest, BufferStoreUsesM0Soffset) {
...
+        uint32_t value = 0;
+        msc->read(output_addr, reinterpret_cast<uint8_t *>(&value), sizeof(value));
+        completed_reads.fetch_add(1, std::memory_order_relaxed);
+      }
+    });
+  }
+
+  start.arrive_and_wait();
+  while (completed_reads.load(std::memory_order_relaxed) < reader_count)
+    std::this_thread::yield();
```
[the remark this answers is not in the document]
kuhar: One concrete way to tighten this is to increment a `readers_started` counter exactly once per reader after its first completed `read()`, then wait for `readers_started == reader_count` before starting the flusher. That preserves the intended all-reader contention without relying on an aggregate loop count.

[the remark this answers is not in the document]
lialan: Each reader now increments `readers_started` exactly once after its first completed read, and the flusher starts only after all four readers have reached that point.

emulation/rocjitsu/tests/gfx1250_sim_test.cpp
lialan: made it 60 seconds.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/amdgpu/memory_side_cache.h
lialan: I extracted `WriterPreferredAccessGate` as an internal testable type and added `WriterPreferredAccessGateTest.WaitingWriterBlocksLateSharedEntrant`: it holds an active reader, waits until a writer is queued, verifies a later shared try-lock is rejected, releases the reader, and verifies the writer acquires. Release and shared TSan both pass.

kuhar:
@lialan I see some CI failure on existing tsan tests here: https://github.com/ROCm/rocm-systems/actions/runs/30156792184/job/89675918045?pr=9257#step:8:4828

lialan:
> @lialan I see some CI failure on existing tsan tests here: https://github.com/ROCm/rocm-systems/actions/runs/30156792184/job/89675918045?pr=9257#step:8:4828

Fixed the issue in this PR.

atgutier (approved):
Approved with comment and assuming you address Jakub's comments.

## MINOR

### [MINOR] `emulation/rocjitsu/tests/hip_vector_add_test.cpp:27` — The suppression disables all TSan interceptor tracking originating from the HSA runtime

**Reasoning:** `called_from_lib:libhsa-runtime64.so` is not scoped to the reproduced
async-event-pool false positive. TSan registers the matched module's entire executable code range
with its library-ignore machinery, then ignores intercepted memory operations whose caller PC is
anywhere in that range. In the newly re-enabled nested HIP test, this can also hide a genuine
cross-boundary race between an HSA-runtime `memcpy`, allocator operation, or other interceptor and
instrumented rocjitsu or test code accessing the same storage. Instrumented accesses wholly inside
rocjitsu remain visible, but the module-wide rule removes more cross-boundary evidence than the
single false positive requires.

**Evidence:** Lines 22-28 describe one race in the runtime's async-event pool but install a
module-wide `called_from_lib` suppression. The final workflow's only blocking TSan configuration
uses the shared runtime at `.github/workflows/rocjitsu-corpus-tests.yml:76-85`, and the final
exclusion list no longer excludes `RocjitsuCliDaemon.LaunchesApplicationAfterDaemonIsReady`.
That test starts `hip_vector_add_test` as a nested process, so this broad default suppression is
what permits the newly covered child to complete without TSan's failure exit code. TSan's
`called_from_lib` implementation resolves the template to the library's executable address range
and ignores every interceptor call whose caller falls within that range; it does not match only
the reported async-event-pool stack.

**Suggested fix:** Replace the module-wide rule with a `race:` suppression matching the narrowest
stable symbol in the reproduced report, such as the specific async-event-pool/GetAllEvents frame.
Run the nested test with `print_suppressions=1` and verify that only the intended report matches.
Using an instrumented runtime, or adding an explicit sanitizer synchronization annotation for the
external lock, would preserve still more coverage.

**Confidence:** Medium-high. The module-wide suppression semantics and resulting coverage loss
are verified. The exact narrow symbol template should be selected from the full reproduced TSan
trace, which is not included in the PR artifacts.

PR #9110, opened 2026-07-23
fix(amdsmi): isolate the tests package scriptlets and require the distro install matrix

marifamd:
## Description

Isolate the component maintainer scripts and tighten the ABI/wheel build gates. Stacks on #9035.

**Scriptlet ownership**

- **`fix: stop the tests package from running the main package scriptlets`** — the `amd-smi-lib-tests` package inherited the main package's post/preun scriptlets, so removing the tests package could run the main package's `ld.so.conf.d` cleanup and corrupt the main install.
- **`fix: give every component its own maintainer scripts`** — scoping only the tests package left the same hole open for `asan`. An unprefixed `CPACK_RPM_<KIND>_SCRIPT_FILE` is copied into every component RPM, and a Debian component without its own control-extra inherits the generic one. The scripts manage absolute paths shared by every variant (`/etc/ld.so.conf.d`, `/var/log/amd_smi_lib`, `/etc/logrotate.d`), so an inheriting component clobbers the main package even under a different prefix. Name the owning component group for every script instead; anything not named ships nothing, so a component added later cannot inherit by accident.
- **`fix: resolve the tests cleanup path at install time`** — the tests pre-uninstall built its path from the configure-time prefix instead of `$RPM_INSTALL_PREFIX0`, so a relocated install cleaned the wrong tree.

**Build gates**

- **`test: fail the ABI gate when the binding scan matches nothing`** — the ABI compatibility gate silently passed when the ctypes binding scan matched zero symbols (e.g. a generator rename); it now fails loudly.
- **`build: only require Python3 for the wheel build`** — scope the `find_package(Python3)` requirement to `BUILD_PYTHON_WHEEL=ON` so the default (non-wheel) build does not fail on a missing Python3.
- **`fix: don't fall back to an unset Python3_SITELIB`** — with `find_package(Python3)` no longer running for a plain C build, the module-destination fallback to `Python3_SITELIB` would install to the prefix root. Fall back only when it holds a value, and fail loudly otherwise.

## JIRA ID

ROCM-3941


gabrpham (approved):

dmitrii-galantsev (approved):

ammallya:
Flaky github fetch sources error, build and test for smi are good

PR #9112, opened 2026-07-23
feat(rocjitsu): Introduce SpecialRegisterSet and special uses/defs

accauble:
## Motivation

Fieldless special operands (VCC, EXEC, SCC, M0, and PC) were decoded but inert so analysis had no way to see if an instruction used or defined them. This PR changes that by adding "special register" def/use. It does not change any runtime execution. 

## Technical Details

* I made the decision to give the special registers their own `SpecialRegisterSet` and `special_{uses,defs}`, rather than adding them to the current ordinary register ones for backwards compatibility and to give us more control over them (e.g. allocating scratch for liveness)
* Adds a `to_special_reg_class` which is like the `to_register_ref` function but for special registers. Returns a `RegClass` instead of a `RegisterRef` since the special registers do not have an index/width associated with them (internally).
* `FLAT_SCRATCH` is classified as a special register internally but isn't quite like the others since it doesn't really have an special effect. There is still a place to handle it in Issue #8625.

Note: I don't feel like too much was introduced by this PR, but the line count would beg to differ. I could split this into two PRs: the first would introduce the data structures and the second would populate/test them.

## Issue Tracking

This is the next step in Issue #8625.

## Test Plan
 
* Added a tests to verify that representative instructions have the correct special defs/uses

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
+++ b/emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
@@ -9445,6 +9445,31 @@ inline void unpack_6bit(const uint32_t dwords[6], uint8_t vals[32]) {{
             f'}}'
         )
 
+        # to_special_reg_class(): maps a special operand (VCC/EXEC/SDST_EXEC/
+        # SSRC_SPECIAL_SCC/M0/PC) to its special RegClass so InstDefUse can
+        # record it in special_defs/special_uses. Driven by shared fieldless
+        # operand policy table's effect column.
+        special_ref_cases = []
+        for opnd_type in self.isa_spec.operand_types:
+            effect = fieldless_policy(opnd_type).effect
+            if effect is not None and effect.special_reg is not None:
+                special_ref_cases.append(
+                    f'case OperandType::{opnd_type}: '
+                    f'return RegClass::{effect.special_reg.name};'
+                )
+        special_ref_cases.sort()
+        special_ref_body = '\n'.join(special_ref_cases)
+        special_ref_impl = (
+            f'std::optional<RegClass> Operand::to_special_reg_class() const {{\n'
+            f'switch (opr_type_) {{\n'
+            f'{special_ref_body}\n'
+            f'default:\n'
+            f'  break;\n'
+            f'}}\n'
+            f'return std::nullopt;\n'
+            f'}}'
+        )
+
         operand_ctor_decl = (
             '  Operand(int size_bits, OperandType opr_type, int encoding_value,\n'
             '          bool packed_16bit_source = false, bool packed_16bit_dst = false);\n'
@@ -9569,6 +9594,7 @@ inline void unpack_6bit(const uint32_t dwords[6], uint8_t vals[32]) {{
                 '  std::string name() const override;\n'
                 f'{literal64_decl}'
                 '  std::optional<RegisterRef> to_register_ref() const override;\n'
+                '  std::optional<RegClass> to_special_reg_class() const override;\n'
                 f'{execution_backend_public_decl}'
                 f'{execution_decls}'
                 'private:\n'
@@ -9644,6 +9670,7 @@ inline void unpack_6bit(const uint32_t dwords[6], uint8_t vals[32]) {{
                 cgen.Line(literal64_impl),
                 cgen.Line(name_impl),
                 cgen.Line(ref_impl),
+                cgen.Line(special_ref_impl),
             ]
         )
 
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/def_use_chain.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/def_use_chain.h
@@ -9,6 +9,15 @@
 /// registers, source operands use registers. Operand::to_register_ref()
 /// determines which register class and index are involved. Instruction
 /// subclasses may also report hidden register effects through implicit hooks.
+///
+/// Ordinary register-file effects (SGPR/VGPR/AccVGPR) live in `defs`/`uses`
+/// and drive scratch-allocation liveness. Architectural special-register
+/// effects (EXEC/VCC/SCC/M0/PC/...) and memory effects are exposed on separate
+/// members (`special_defs`/`special_uses`, `memory`) so consumers can query
+/// them without those effects ever entering ordinary liveness. Those separate
+/// members are currently always empty and will be populated once generated
+/// operands report special/memory effects; the ordinary `defs`/`uses` behavior
+/// is unchanged.
 
 #pragma once
 
@@ -19,6 +28,28 @@ namespace rocjitsu {
 class Instruction;
 class Gfx1250VgprMsbAnalysis;
 
+/// @brief Memory-effect summary for one decoded instruction.
+///
+/// @details A coarse description of what memory an instruction touches. Not
+/// currently used. Present now so `InstDefUse` has a stable shape for DBT/DBI
+/// consumers. All flags default to false (no memory effect).
+///
+/// TODO: Update when OPR_GPUMEM, OPR_DSMEM, and OPR_FLAT_SCRATCH become
+/// avaialable.
+struct MemoryEffects {
+  bool reads = false;   ///< Reads memory.
+  bool writes = false;  ///< Writes memory.
+  bool atomic = false;  ///< Performs an atomic read-modify-write (also sets reads+writes).
+  bool global = false;  ///< Touches global/generic memory (OPR_GPUMEM).
+  bool lds = false;     ///< Touches LDS/GDS (OPR_DSMEM).
+  bool scratch = false; ///< Touches scratch/flat-scratch (OPR_FLAT_SCRATCH).
+
+  /// @brief True if the instruction has any memory effect.
+  [[nodiscard]] bool any() const { return reads || writes || atomic || global || lds || scratch; }
+
+  friend bool operator==(const MemoryEffects &, const MemoryEffects &) = default;
+};
+
 /// @brief Registers read and written by one decoded instruction.
 class InstDefUse {
 public:
@@ -26,8 +57,11 @@ public:
   /// @param inst Decoded instruction whose operands have stable lifetimes.
   InstDefUse(const Instruction &inst, const Gfx1250VgprMsbAnalysis *vgpr_msb = nullptr);
 
-  RegisterSet defs;                        ///< Registers overwritten by the instruction.
-  RegisterSet uses;                        ///< Registers read before the instruction writes defs.
+  RegisterSet defs;                ///< Ordinary registers overwritten by the instruction.
+  RegisterSet uses;                ///< Ordinary registers read before the instruction writes defs.
+  SpecialRegisterSet special_defs; ///< Special registers (EXEC/VCC/SCC/M0/PC/...) written.
+  SpecialRegisterSet special_uses; ///< Special registers read.
+  MemoryEffects memory;            ///< Memory-effect summary.
   bool has_exec_masked_vector_def = false; ///< True if any vector def is predicated by EXEC.
   bool has_predicated_def = false;         ///< True if defs preserve old values on some paths.
 };
--- a/emulation/rocjitsu/tests/analysis/liveness_test.cpp
+++ b/emulation/rocjitsu/tests/analysis/liveness_test.cpp
@@ -579,6 +579,50 @@ TEST(InstDefUseSpecialEffects, SpecialEffectsDoNotEnterScratchLiveness) {
   EXPECT_TRUE(du.memory.any());
 }
 
+TEST(SpecialEffectAnalysis, SaveexecReportsExecAndSccFromDecodedOperands) {
+  // Real decoded instruction: s_and_saveexec_b64 s[0:1], s[0:1] (CDNA4). Its
+  // fieldless special operands drive the InstDefUse special sets by direction:
+  //   dst EXEC + dst SCC  -> special_defs
+  //   src EXEC (old exec) -> special_uses
+  // while the ordinary sdst/ssrc0 SGPRs stay in the ordinary defs/uses that feed
+  // scratch allocation. Special effects never enter ordinary RegisterSet liveness.
+  const uint32_t words[] = {0xBE802000u, 0x00000000u};
+  auto decoder = Decoder::create(ROCJITSU_CODE_ARCH_CDNA4);
+  ASSERT_NE(decoder, nullptr);
+  std::unique_ptr<Instruction> inst(decoder->decode(words));
+  ASSERT_NE(inst, nullptr);
+  ASSERT_EQ(inst->mnemonic(), "s_and_saveexec_b64");
+
+  InstDefUse du(*inst);
+
+  EXPECT_TRUE(du.special_defs.contains(RegClass::EXEC));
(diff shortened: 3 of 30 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/def_use_chain.h
```diff
@@ -19,15 +28,40 @@ namespace rocjitsu {
...
 
+/// @brief Memory-effect summary for one decoded instruction.
+///
+/// @details A coarse description of what memory an instruction touches. Not
+/// currently used. Present now so `InstDefUse` has a stable shape for DBT/DBI
+/// consumers. All flags default to false (no memory effect).
+///
+/// TODO: Update when OPR_GPUMEM, OPR_DSMEM, and OPR_FLAT_SCRATCH become
+/// avaialable.
+struct MemoryEffects {
```
kuhar, line 39: The issue I am worried about is that `MemoryEffects{}` currently means both “no memory access” and “not modeled.” A decoded buffer load therefore reports `memory.any() == false`, and the API docs call that “no memory effect.” This would be safer if the unmodeled state were explicit (for example, optional or tri-state), or if `memory` stayed out of `InstDefUse` until it is populated.

emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py
```diff
@@ -9528,6 +9528,31 @@ def gen_operand(self) -> None:
             f'}}'
         )
 
+        # to_special_reg_class(): maps a special operand (VCC/EXEC/SDST_EXEC/
+        # SSRC_SPECIAL_SCC/M0/PC) to its special RegClass so InstDefUse can
+        # record it in special_defs/special_uses. Driven by shared fieldless
+        # operand policy table's effect column.
+        special_ref_cases = []
+        for opnd_type in self.isa_spec.operand_types:
+            effect = fieldless_policy(opnd_type).effect
```
kuhar, line 9454: Should we either include encoded special-register selectors or make the `InstDefUse` contract explicitly fieldless-only? `Operand::to_special_reg_class()` documents the fieldless boundary, but `InstDefUse::special_defs` and `special_uses` read as a complete special-register summary. An explicit `s_mov_b32 exec_lo, 0` decodes through `OPR_SDST` and leaves `special_defs` empty, so a consumer can otherwise miss real EXEC, VCC, M0, or FLAT_SCRATCH effects.

emulation/rocjitsu/tests/analysis/liveness_test.cpp
```diff
@@ -401,6 +403,429 @@ TEST(RegisterSetAnalysis, Cdna4WritelaneDestinationIsUseAndDef) {
...
+  ASSERT_EQ(inst->mnemonic(), "v_add_co_ci_u32_e32");
+
+  InstDefUse du(*inst);
+  EXPECT_EQ(du.special_defs, special_regs({RegClass::VCC}));
+  EXPECT_EQ(du.special_uses, special_regs({RegClass::VCC}));
+}
+
+// PC family: getpc reads PC (writes an ordinary SGPR pair), setpc writes PC
+// (reads an ordinary SGPR pair), s_call does both.
+TEST(SpecialEffectAnalysis, GetpcReadsPc) {
```
kuhar, line 767: It would be useful to add one decode-driven M0 test alongside the EXEC, VCC, SCC, and PC cases. The policy and generator tests pin `OPR_SDST_M0` to `RegClass::M0`, but no decoded instruction exercises that path into `InstDefUse`; `s_movrels_b32` can cover `special_uses`, and `s_set_gpr_idx_idx` can cover defs.

kuhar (commented):
We should update the nearby docs to match the implementation:

- `fieldless_policy.py:43` says the RegClass mapping is not modeled, and `fieldless_policy.py:79` calls FLAT_SCRATCH a memory pseudo-operand, while the effect column and table now model that mapping and classify FLAT_SCRATCH as `SPECIAL_REGISTER`.
- `def_use_chain.h:18` says the special sets are always empty, although only memory remains unpopulated. There is also `avaialable` at line 38.
- `register_set.h:9` says special classes have no consumer, but `SpecialRegisterSet` and `InstDefUse` now consume them.

atgutier (commented):
[MAJOR] lib/python/amdisa/codegen/_generator.py:9348-9359,9388-9407,9531-9554 — special def/use omits encoded special-register selectors                                                     13:09:16 [28/1917]

  Reasoning: InstDefUse::special_defs and special_uses look like complete architectural summaries, but the implementation only recognizes dedicated fieldless operand types. Special registers encoded through
  ordinary scalar selector fields produce neither an ordinary RegisterRef nor a special-register effect.

  A concrete failure is s_mov_b32 exec_lo, 0:

  - cdna3/sop1.cpp:23-30 represents the destination as OPR_SDST.
  - cdna3/operand.cpp:82-106 shows that OPR_SDST can name EXEC, VCC, M0, FLAT_SCRATCH and TTMP registers.
  - cdna3/operand.cpp:802-808 converts only the SGPR range to RegisterRef.
  - cdna3/operand.cpp:1024-1039 maps only dedicated fieldless operand types and has no OPR_SDST or OPR_SSRC handling.
  - analysis/def_use_chain.cpp:78-96 has no fallback, so this instruction reports neither defs nor special_defs[EXEC].

  This also affects encoded VCC, M0, FLAT_SCRATCH, TTMP, EXEC sources, and scalar condition selectors. Moreover, fieldless_policy.py:225-226 classifies fieldless FLAT_SCRATCH as special but supplies no
  FieldlessEffect, so even that form is never returned by to_special_reg_class().

  The current representation cannot express precise selector effects: register_set.h:182-190 treats every special class as a singleton, while the ISA distinguishes TTMP indices and the low/high halves of EXEC,
  VCC, and FLAT_SCRATCH. Class-coarse “may touch” information can be useful, but it is unsafe as an exact liveness kill.

  Existing decode-driven tests cover dedicated fieldless operands only. tests/patch/probe_clobber_test.cpp:31-37 contains the exact encoded EXEC/M0/VCC/FLAT_SCRATCH cases, but compensates with name-based
  recognition rather than exercising special_defs.

  Suggested fix: either:

  - Generate selector-value-aware special references carrying class, index/subregister and width, with decoded tests for EXEC, VCC, M0, FLAT_SCRATCH and distinct TTMPs; or
  - Define this API explicitly as class-coarse “may touch” information, add mappings for all encoded selectors, and ensure consumers never use it as a precise kill set.

  Confidence: high.

  [MAJOR] lib/rocjitsu/src/rocjitsu/analysis/def_use_chain.h:31-50,64 — MemoryEffects{} conflates “no access” with “not modeled”

  Reasoning: the new public member presents an all-false value as “no memory effect,” but InstDefUse never populates it. Consumers therefore cannot distinguish a proven memory-free instruction from an
  unmodeled load, store, or atomic.

  Evidence:

  - analysis/def_use_chain.cpp:75-99 never writes memory.
  - tests/analysis/liveness_test.cpp:811-826 decodes a real buffer_load_b32 and explicitly expects du.memory.any() to be false.
  - def_use_chain.h:42 documents atomic operations as also setting reads and writes, but the public booleans do not enforce this invariant.
  - liveness_test.cpp:543-548 sets only atomic=true, leaving reads and writes false.

  There is no current production consumer, so this does not yet cause a runtime miscompile. It nevertheless establishes an unsafe foundational contract for future DBT/DBI passes.

  Suggested fix: remove memory until implemented, or represent knowledge explicitly using an Unknown/tri-state or std::optional<MemoryEffects>. When populated, use construction APIs that enforce atomic ⇒
  reads+writes and test known load, store, LDS/scratch and atomic effects.

  Confidence: high.


  [NIT] Several exact-head comments contradict the implementation:

  - def_use_chain.h:18-20 says special sets are always empty.
  - register_set.h:7-10 says special classes have no consumer.
  - fieldless_policy.py:40-43 says the C++ RegClass mapping is not modeled.
  - fieldless_policy.py:77-79 calls FLAT_SCRATCH a memory pseudo-operand, while :225-226 classifies it as SPECIAL_REGISTER.
  - def_use_chain.h:37-38 contains avaialable.

  These should be updated together with an explicit fieldless/coarse-effects contract.

  Confidence: high.

atgutier (commented):
A few more agent findings, but no blockers:

## MAJOR

### [MAJOR] `emulation/rocjitsu/lib/python/amdisa/codegen/_generator.py:9448` — Special def/use omits encoded selectors and cannot precisely represent their subregisters

**Reasoning:** `to_special_reg_class()` is generated exclusively from the fieldless-operand policy. Special registers selected through ordinary scalar operand fields consequently produce neither an ordinary `RegisterRef` nor a special-register effect. Additionally, `SpecialRegisterSet` collapses each class to one bit, so merely adding selector cases would conflate distinct TTMPs and the low/high halves of EXEC, VCC, and FLAT_SCRATCH. Such a class-wide def cannot safely serve as an exact liveness kill.

**Evidence:**

- `gfx1250/sop1.cpp:13-20` represents `s_mov_b32` using `OPR_SDST` and `OPR_SSRC`.
- `gfx1250/operand_types.h:72-96` shows that `OPR_SDST` can select TTMP0–15, VCC_LO/HI, M0, and EXEC_LO/HI. Corresponding source selectors occur at lines 362-386.
- `gfx1250/operand.cpp:882-888` maps only the ordinary SGPR portion of `OPR_SDST`.
- The new helper at `gfx1250/operand.cpp:1039-1056` has no `OPR_SDST` or `OPR_SSRC` handling.
- `def_use_chain.cpp:78-96` drops the operand after both conversions return `nullopt`.
- Fieldless FLAT_SCRATCH is also incomplete: `fieldless_policy.py:225-226` classifies it as `SPECIAL_REGISTER` but assigns no `FieldlessEffect`.
- RDNA3.5 flat loads place that operand in their source list at `rdna3_5/flat.cpp:26-49`, but `rdna3_5/operand.cpp:721-739` returns no corresponding special class.
- `register_set.h:182-190` calls all special registers singletons with no meaningful index or width, despite the selector distinctions above.
- The decode tests at `tests/analysis/liveness_test.cpp:593-810` exercise only dedicated fieldless operands.
- `tests/patch/probe_clobber_test.cpp:31-37` contains the exact encoded EXEC, M0, VCC, and FLAT_SCRATCH cases, but explicitly compensates through operand-name recognition instead of `special_defs`.

**Suggested fix:** Introduce a `SpecialRegisterRef` carrying class, subregister/index, and width, with corresponding range-aware set storage. Generate mappings for generic encoded selectors and fieldless forms, including FLAT_SCRATCH. Add decoded tests for EXEC/VCC halves, distinct TTMPs, M0, and encoded and fieldless FLAT_SCRATCH. If the intended result is only class-coarse “may touch” metadata, name and document it accordingly and prevent it from being interpreted as a kill set.

**Confidence:** High. The earlier encoded-selector, fieldless-FLAT_SCRATCH, and precision findings remain unresolved.

### [MAJOR] `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/def_use_chain.cpp:89` — Ordinary EXEC-gated vector instructions omit their implicit EXEC use

**Reasoning:** `special_uses` is documented as the set of special registers read, but extraction only examines source operands. Ordinary vector instructions depend on EXEC for lane gating without carrying a fieldless EXEC operand, so they incorrectly report no EXEC use. A future special-register liveness or DBI spill consumer could treat EXEC as dead across such an instruction.

**Evidence:**

- `cdna3/vop2.cpp:155-195` constructs `VAddF32Vop2` with only `vdst`, `src0`, and `vsrc1`; there is no EXEC source operand.
- Its execution helper reads `wf.exec()` and gates every lane at `shared/execute_shared.h:2999-3009`.
- `def_use_chain.cpp:89-99` scans only source operands and then calls `inst.implicit_uses(uses)`.
- `instruction.h:216-223` makes that hook accept only `RegisterSet&`, so it cannot report special-register dependencies.
- `def_use_chain.cpp:66-70` already recognizes EXEC predication indirectly through `has_exec_masked_vector_def`, but that flag does not populate `special_uses`.
- That flag also cannot cover EXEC-gated operations without a vector destination. For example, `gfx1250/vbuffer.cpp:242-261` gives `buffer_store_b32` no EXEC operand, while its implementation reads EXEC and suppresses inactive lanes at `gfx1250/vbuffer_exec.cpp:196-215`.
- `def_use_chain.h:13-16,63` documents architectural special-register effects and “Special registers read” without limiting the result to fieldless operands.

**Suggested fix:** Add instruction-level special-effect metadata or special-register implicit-use/def hooks, then mark every operation whose semantics consume EXEC—including vector memory operations without vector defs. Alternatively, rename and explicitly document these members as fieldless-operand effects and prohibit treating them as complete architectural liveness.

**Confidence:** High. The narrow fieldless-only behavior appears intentional relative to the PR description, but it contradicts the broader public C++ API contract. There is no production consumer yet, so this does not currently cause a runtime miscompile.

### [MAJOR] `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/def_use_chain.h:31` — `MemoryEffects{}` conflates “no memory access” with “not modeled”

**Reasoning:** The newly exposed member presents an all-false value as no memory effect, but `InstDefUse` never populates it. Consumers therefore cannot distinguish an instruction proven not to access memory from an unmodeled load, store, or atomic.

**Evidence:**

- `def_use_chain.cpp:75-99` never writes `memory`.
- `gfx1250/vbuffer.cpp:110-129` represents `buffer_load_b32` with GPUMEM as a source and sets `MEMORY_OP`.
- `gfx1250/vbuffer.cpp:242-261` represents a store with GPUMEM as a destination.
- `gfx1250/vbuffer.cpp:506-529` represents an atomic with GPUMEM as both source and destination.
- Nevertheless, `tests/analysis/liveness_test.cpp:815-827` explicitly expects a decoded `buffer_load_b32` to report `memory.any() == false`.
- `def_use_chain.h:42` says `atomic` also sets `reads` and `writes`, but the public booleans do not enforce that invariant. The test at `liveness_test.cpp:544-550` sets only `atomic=true`.

**Suggested fix:** Remove or defer the member until it is implemented, or represent knowledge explicitly with an unknown state such as `std::optional<MemoryEffects>` or a tri-state summary. When populated, derive direction and address space from operand/semantic metadata, enforce `atomic => reads && writes` through construction APIs, and test global, LDS, scratch, load, store, and atomic cases.

**Confidence:** High. There is no production consumer yet, but the earlier unsafe-contract finding remains unresolved.

## MINOR

None.

## NIT

### [NIT] `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/def_use_chain.h:18` — Nearby documentation still describes the pre-feature state

**Reasoning:** Exact-head comments contradict the implementation and obscure the already-ambiguous completeness contract.

**Evidence:**

- `def_use_chain.h:18-20` says the special sets are always empty, while `def_use_chain.cpp:84-96` now populates them.
- `fieldless_policy.py:40-43` says the C++ `RegClass` mapping is not modeled.
- `fieldless_policy.py:77-79` calls FLAT_SCRATCH a memory pseudo-operand, while lines 225-226 classify it as a special register.
- `def_use_chain.h:37-38` contains the typo `avaialable`.

**Suggested fix:** Update these comments together with an explicit contract for fieldless versus encoded and implicit effects, class-level versus subregister precision, and unknown memory effects.

**Confidence:** High.

accauble:
Creating a new PR #9613 which takes this PR and PR #9230.

PR #9120, opened 2026-07-23
[rocjitsu] Fixes to B0 -> A0 for DSV4

Groverkss:


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/legalization/gfx1250_b0_to_a0.cpp
```diff
@@ -106,21 +104,17 @@ inline constexpr std::array<std::string_view, 18> kExactErrataMnemonics = {
...
     return true;
 
-  // Additional low-precision WMMA/SWMMAC forms are not yet supported on this
-  // target and are classified so translation fails closed rather than copying
-  // them through unchanged (see the not-yet-supported note above). These have no
-  // semantic rule yet:
-  //   * the bare K=128 F8F6F4 WMMA,
-  //   * the K=64 FP8/BF8 WMMA family, and
-  //   * the FP8/BF8 SWMMAC family (the integer SWMMAC is handled below).
+  // K=64 FP8/BF8 WMMA and K=128 F8F6F4 are present on A0 and retain their
```
kuhar: We should update the classification comments at `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/legalization/gfx1250_b0_to_a0.cpp:90-103` along with the new semantic rules. They still say that K=128 FP8/BF8 and Scale/Scale16 have no A0 lowering and must fail closed, although this patch implements those paths. Leaving that contract stale makes the supported/unsupported split in this legalizer misleading.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/semantic/gfx1250_b0_to_a0.cpp
```diff
@@ -395,16 +404,110 @@ ExpandResult expand_gfx1250_wmma_scale_src2(const Instruction &inst, uint32_t, u
...
+    return static_cast<uint16_t>(bank * 256u + reg);
+  };
+  const auto overlaps = [](uint16_t lhs, uint16_t lhs_count, uint16_t rhs,
+                           uint16_t rhs_count) {
+    return lhs < static_cast<uint32_t>(rhs) + rhs_count &&
+           rhs < static_cast<uint32_t>(lhs) + lhs_count;
+  };
+  const uint16_t first_dst = physical(*dst_bank, matrix.vdst);
+  const uint16_t upper_a = physical(*src0_bank, static_cast<uint16_t>(src0 + kHalfDwords));
+  if (overlaps(first_dst, kHalfDwords, upper_a, kHalfDwords) ||
```
kuhar, line 460: This would be safer if the regular-Scale M=32 split rejected or preserved a shared B range that overlaps the first destination half. At `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/semantic/gfx1250_b0_to_a0.cpp:450`, the guard covers upper A and upper C only. The first replacement writes D[0:7], then the second reuses the original B, so D/B aliasing silently feeds overwritten B values to the upper half. Please add the same physical-range protection used by Scale16 and a focused alias test.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/semantic/gfx1250_b0_to_a0.cpp
```diff
@@ -622,34 +765,68 @@ ExpandResult expand_gfx1250_wmma_scale16(const Instruction &inst, uint32_t, uint
...
+    // Splitting changes the A/D ranges, so the original reuse promises no
+    // longer apply. Select the A-scale lanes corresponding to this M half.
+    pass[0] &= ~((uint32_t{1} << 13) | (uint32_t{1} << 14));
+    set_word_field(pass[0], m_half, 11, 1);
     set_word_field(pass[1], gfx1250_vgpr_src(pass_scale_a), 0, 9);
     set_word_field(pass[1], gfx1250_vgpr_src(pass_scale_b), 9, 9);
     set_word_field(pass[1], gfx1250_vgpr_src(0), 18, 9);
     set_word_field(pass[3], gfx1250_vgpr_src(masked_a), 0, 9);
+    if (is_m32) {
+      set_word_field(pass[2], static_cast<uint32_t>(matrix.vdst + m_half * 8u), 0, 8);
```
kuhar: We should handle or fail closed when the Scale16 M=32 upper D/C half crosses a VGPR bank. At `emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/semantic/gfx1250_b0_to_a0.cpp:790`, `matrix.vdst + 8` and `matrix_c + 8` are encoded in low register fields while `first_wmma_mode` and `accumulate_wmma_mode` retain the original DST/SRC2 banks. Operands starting at v248 therefore wrap to v0 in the same bank and target the wrong physical registers. Please mirror the regular-Scale bank transition and bank-3 rejection logic here, with boundary tests.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/semantic/gfx1250_b0_to_a0.cpp
```diff
@@ -963,25 +1140,121 @@ ExpandResult expand_gfx1250_cvt_f32_fp8_e5m3(const Instruction &inst, uint32_t,
...
+    second_mode = static_cast<uint8_t>((*src0_bank + (src0_crosses_bank ? 1u : 0u)) |
+                                       ((*src1_bank + (src1_crosses_bank ? 1u : 0u)) << 2) |
+                                       (*dst_bank << 4) | (*dst_bank << 6));
+  }
+
+  set_word_field(second[1], kVgprOperandEncoding + ((src0_low + kHalfInputDwords) & 0xffu), 0, 9);
+  set_word_field(second[1], kVgprOperandEncoding + ((src1_low + kHalfInputDwords) & 0xffu), 9, 9);
+  const uint16_t vdst_as_src2 =
+      static_cast<uint16_t>(kVgprOperandEncoding + (first[0] & 0xffu));
+  set_word_field(second[1], vdst_as_src2, 18, 9);
```
kuhar: Please keep the four f16-output K=128 forms fail-closed instead of feeding the first K=64 result back as C. The gfx1250 executor accumulates one K=128 f16-output WMMA in f32 and packs once at the end (`mma_exec.h:2817-2839`); this rewrite packs after K=64, rereads that f16 as C, and packs again. The extra rounding changes results: with C=1 and each half contributing 2^-11, one final f16 round gives 1.0009765625 while the chained sequence gives 1.0. Split the f16 cases from f32 and fail closed until there is a single-rounding A0 lowering, with a rounding-sensitive regression.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/semantic/gfx1250_b0_to_a0.cpp
```diff
@@ -963,25 +1140,121 @@ ExpandResult expand_gfx1250_cvt_f32_fp8_e5m3(const Instruction &inst, uint32_t,
   return ExpandResult::success(std::move(words));
 }
 
-/// @brief Lower a B0-only K=128 FP8/BF8 WMMA for A0.
+/// @brief Split a B0-only K=128 FP8/BF8 WMMA into two A0 K=64 WMMAs.
```
kuhar: For the f32 cases, please prefer the native A0 regular-Scale K=128 form instead of decomposing into two K=64 WMMAs. Encode `v_wmma_scale_f32_16x16x128_f8f6f4` with neutral E8M0 scale words `0x7f7f7f7f`, select FP8/BF8 through `opsel`/`opsel_hi`, and preserve the original C/D. This uses one matrix operation and removes the split-only destination/input alias rejection and VGPR-MSB transition path. It needs one dead SGPR for the neutral scale; fail closed or use a proven f32 split fallback when none is available. Please cover all four format pairs and modifier preservation with raw-word assertions.

kuhar (commented):
We should update the three stale fail-closed tests before merging. At `emulation/rocjitsu/tests/dbt/translate_test.cpp:8459`, `:8683`, and `:8986`, the current Clang and GCC-13 binaries all fail because the new lowerings now succeed; the focused run is 0/3 in both builds.

Please replace those old assertions with operand-level regression checks for the new contracts: the K=128-to-two-K=64 split and D-as-C chaining; byte-for-byte copy-through for K=64 FP8/BF8 and K=128 F8F6F4 while keeping SWMMAC fail-closed; and the four-pass Scale16 M/K split with sliced C/D ranges. It would also be useful to cover the regular-Scale M=32 path and the new overlap, bank-overflow, and even-alignment failure guards so the byte-exact WMMA transformations are pinned.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/semantic/gfx1250_b0_to_a0.cpp
```diff
@@ -349,62 +349,135 @@ ExpandResult expand_gfx1250_tensor_load_to_lds(const Instruction &inst, uint32_t
...
+    return static_cast<uint16_t>(bank * 256u + reg);
+  };
+  const auto overlaps = [](uint16_t lhs, uint16_t lhs_count, uint16_t rhs, uint16_t rhs_count) {
+    return lhs < static_cast<uint32_t>(rhs) + rhs_count &&
+           rhs < static_cast<uint32_t>(lhs) + lhs_count;
+  };
+  const uint16_t first_dst = physical(*dst_bank, matrix.vdst);
+  const uint16_t upper_a = physical(*src0_bank, static_cast<uint16_t>(src0 + kHalfDwords));
+  const uint16_t shared_b = physical(*src1_bank, src1);
+  if (overlaps(first_dst, kHalfDwords, upper_a, kHalfDwords) ||
```
kuhar, line 460: This would be safer if the M=32 split also rejected D[0:7] overlap with scale.src0 or scale.src1. Each emitted prefix rereads both scale VGPRs, so the first WMMA can overwrite a scale value needed by the second even though the original instruction reads it before writing D. Please add physical-range checks for both low-bank scale sources and focused alias regressions.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp
```diff
@@ -1904,36 +1905,106 @@ void recover_vector_lane_stashed_pcs(AnalysisContext &ctx,
...
-        // store the physical index (low + bank*256), so compare the low 8 bits.
-        std::erase_if(slots, [&](const auto &item) {
+        // Operand metadata does not expose a role for every implicit/wide def.
+        // Conservatively invalidate every physical bank sharing this selector.
+        std::erase_if(state.slots, [&](const auto &item) {
           return (item.first.vgpr & 0xffu) == (ref.index & 0xffu);
         });
       });
+
+      if (!builders.active_pairs().empty())
```
kuhar, line 1954: We should treat a direct scalar call as a kill boundary for propagated lane-slot state. build_analysis_blocks gives the call both callee and fallthrough edges, but this scan carries state.slots unchanged into the continuation; a getpc/writelane stash can therefore be recovered after the call even when the callee clobbers that VGPR. The scalar PC-builder path already kills carried facts at calls (lines 1469-1481). Please invalidate the lane slots too and add a call/clobber/continuation regression.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/semantic/gfx1250_b0_to_a0.cpp
```diff
@@ -349,62 +349,135 @@ ExpandResult expand_gfx1250_tensor_load_to_lds(const Instruction &inst, uint32_t
...
+       overlaps(first_dst, kHalfDwords,
+                physical(*src2_bank, static_cast<uint16_t>(src2 + kHalfDwords)), kHalfDwords))) {
+    return ExpandResult::failed(
+        "gfx1250 regular-Scale 32x16 lower destination overlaps an input needed by the upper half");
+  }
+
+  const bool dst_crosses = static_cast<uint32_t>(matrix.vdst) + kHalfDwords > 0xffu;
+  const bool src0_crosses = static_cast<uint32_t>(src0) + kHalfDwords > 0xffu;
+  const bool src2_crosses = src2_is_vgpr && static_cast<uint32_t>(src2) + kHalfDwords > 0xffu;
+  if ((dst_crosses && *dst_bank == 3) || (src0_crosses && *src0_bank == 3) ||
```
kuhar: It would be good to add boundary coverage for the regular-Scale M=32 bank logic. The current positive split stays below v256, and the shared-B overlap test exits before exercising either the VGPR-MSB transition or the bank-3 rejection. Please add the same positive crossing and fail-closed overflow cases already used for the sibling Scale16 path.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/semantic/gfx1250_b0_to_a0.cpp
```diff
@@ -965,23 +1131,80 @@ ExpandResult expand_gfx1250_cvt_f32_fp8_e5m3(const Instruction &inst, uint32_t,
...
+
+  gfx1250::Vop3pMachineInst source{};
+  std::memcpy(&source, inst.raw_encoding(), sizeof(source));
+  constexpr uint16_t kVgprEncoding = 256;
+  if (source.src0 < kVgprEncoding || source.src1 < kVgprEncoding) {
+    return ExpandResult::failed("gfx1250 K=128 WMMA matrix operands are not ordinary VGPR ranges");
+  }
+
+  const std::optional<uint16_t> scale_sgpr = liveness.find_free_sgpr(&inst);
+  if (!scale_sgpr || *scale_sgpr > 105) {
```
kuhar: I think we should add a fail-closed regression for neutral-scale SGPR allocation. The new f32 K=128 lowering requires an ordinary SGPR and rejects either no allocation or a result above s105, while the current positive test always gets s0. Please force the allocatable SGPRs live and verify the allocation diagnostic with the object left unchanged; that also locks in the special-encoding boundary.

kuhar (commented):
A few comments

atgutier (commented):
From the agent:

 ## MAJOR

  [MAJOR] emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/semantic/
  gfx1250_b0_to_a0.cpp:427 — M=32 regular-Scale overlap checks use the wrong
  banks for scale operands

  Reasoning: The two-pass M=32 lowering must preserve both scale operands
  until the second prefix reads them. The new overlap guard maps scale_src0
  and scale_src1 through the matrix SRC0/SRC1 VGPR-MSB banks. Those scale
  operands have no VGPR-MSB role and therefore resolve in bank zero.

  For example, with destination bank 0, matrix SRC0 bank 1, destination v96,
  and scale source v96, the guard compares D against physical v352 and
  permits the lowering. The first pass overwrites physical v96, and the
  second prefix rereads that corrupted scale, producing incorrect matrix
  results.

  Evidence:

  - gfx1250_b0_to_a0.cpp:425-431 checks scale sources using
    physical(*src0_bank, scale_src0) and physical(*src1_bank, scale_src1).

  - isa/arch/amdgpu/gfx1250/vop3p.cpp:6191-6194 constructs the scale sources
    as OPR_SRC_SIMPLE.

  - vop3p.cpp:6210-6213 assigns VGPR-MSB roles only to the matrix destination
    and matrix sources.

  - vm/amdgpu/wavefront.h:131-145 resolves VgprMsbRole::None as bank zero.
  - isa/arch/amdgpu/gfx1250/vop3p_exec.cpp:5182-5188 rereads both scale
    operands for each emitted operation.

  - The tests at tests/dbt/translate_test.cpp:9222-9255 cover overlap only
    under the default zero-bank mode.

  Suggested fix: compare scale sources in their actual unbanked namespace:

  physical(0, scale_src0)
  physical(0, scale_src1)

  Also revisit the rejection at lines 439-443: the claim that matrix A and
  scale A share the SRC0 bank selector appears to be similarly incorrect and
  unnecessarily rejects valid bank-crossing cases. Add nonzero matrix-bank
  tests with bank-zero scale/D overlap.

[MAJOR] emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/ana07:27:02 [103/1969]
  indirect_branch_discovery.cpp:1889 — Lane-stash state survives an indirect
  s_swappc call

  Reasoning: The new cross-block analysis clears lane slots at direct scalar
  calls but not at indirect swap-PC calls. Because indirect calls retain a
  fallthrough continuation, a pre-call lane stash flows into the continuation
  even though the callee may overwrite or repurpose that VGPR.

  A later readlane/swappc can consequently be assigned a stale static target.
  The translator can then replace that genuinely dynamic transfer with a
  direct transfer window.

  Evidence:

  - indirect_branch_discovery.cpp:1834-1840 clears state.slots only for
    direct facts.call_sdst.

  - indirect_branch_discovery.cpp:1889-1898 processes the current swappc
    without clearing its lane slots afterward.

  - indirect_branch_discovery.cpp:1935 publishes the unchanged state.
  - indirect_branch_discovery.cpp:1152-1158 gives indirect calls a
    fallthrough successor.

  - code/dbt/binary_translator.cpp:1512-1517 turns a single recovered target
    into a transfer window, installed at lines 1664-1674.

  - The added Gfx1250DirectCallKillsCarriedLaneStash test covers only direct
    calls.

  Suggested fix: after resolving the current swappc, clear state.slots
  whenever facts.swappc_sdst is present, before publishing the block exit. If
  MODE preservation is not guaranteed by the call contract, invalidate
  vgpr_msb_imm across both direct and indirect calls as well. Add an
  indirect-call/callee-clobber/continuation regression parallel to the
  direct-call test.

  Confidence: high.

  [MAJOR] emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/
  basic_block.cpp:26 — The new trap-2 terminator rule is missing from
  indirect-target discovery

 Reasoning: The PR makes ROCr assertion/abort trap 2 a no-successor
  terminator in the final BasicBlock CFG. The temporary CFG used to prove
  indirect targets still treats the same instruction as ordinary fallthrough.

  A get-PC fact from before the abort can therefore flow into post-trap
  bytes. If those bytes are another kernel/function entry or independently
  reachable block, discovery can attach a target justified solely by the
  impossible trap fallthrough. That recovered target can later be converted
  into a direct transfer.

  Evidence:

  - basic_block.cpp:26-47 recognizes trap ID 2 in both final-CFG terminator
    predicates.

  - indirect_branch_discovery.cpp:670-696 uses only instruction flags for the
    corresponding temporary-CFG predicates.

  - indirect_branch_discovery.cpp:682-684 consequently does not split at trap
    2.

  - indirect_branch_discovery.cpp:742-744 and :1152-1158 preserve fallthrough
    across it.

  - Decoded s_trap does not carry PROGRAM_TERMINATOR; the special behavior
    exists only in the newly added basic_block.cpp helper.

  Suggested fix: centralize the ROCr trap-termination predicate and use it
  for leader and successor construction in both CFG implementations. Test a
  PC builder before trap 2 followed by a post-trap setpc block independently
  entered through extra_leaders or a direct branch; no fact or fixup should
  cross the abort.

  Confidence: medium-high. The CFG disagreement is definite; the harmful case
  requires independently reachable post-trap code.

  [MAJOR] emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/analysis/
  indirect_branch_discovery.cpp:1562 — Entry blocks with loop backedges are
  never seeded as reachable

 Reasoning: The new reachability lattice seeds only blocks with no
  predecessors. Text block zero is nevertheless an external entry even when a
  loop backedge targets it. If block zero and its loop body form an SCC,
  every block initially has an unreachable predecessor and none ever becomes
  reachable.

  A valid builder in the entry block therefore cannot reach a cross-block
  setpc or swappc. Recovery is missed and an otherwise supported kernel can
  fail closed. The same issue affects the vector-lane fixed point.

  Evidence:

  - indirect_branch_discovery.cpp:1562-1566 requires either no predecessors
    or an already-reachable predecessor.

  - Lines 1575-1592 ignore all predecessors while the SCC remains unseeded.
  - Lines 1981-1989 use the same no-predecessor condition before giving block
    zero its architectural bank-zero state.

  - DominatedPcBuilderRemainsCompleteAcrossCallLoopBackedge has a separate
    predecessorless builder block before its loop; it does not exercise a
    backedge into text entry.

  Suggested fix: model external-entry edges explicitly. Block zero must be
  seeded regardless of CFG predecessors, and its external entry state must be
  joined with backedges. For the scalar lattice, a root with incoming
  backedge facts needs an unconstrained external-entry contribution; for lane
  state, the external empty slot state must participate in the meet. Ideally
  pass all known kernel/function entry offsets into both fixed points. Add a
  two-block loop whose backedge targets block zero and whose consumer depends
  on a builder created at entry.

  Confidence: high.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/basic_block.cpp
```diff
@@ -234,8 +255,15 @@ BasicBlock::build(const CodeObject &co, Decoder &decoder, rj_code_arch_t arch,
...
         const bool decode_gap =
             i >= decoded.size() ? next_offset < section_end : decoded[i]->src_loc() != next_offset;
-        if (decode_gap && !terminates)
-          current->falls_through_to_undecodable_text_ = true;
+        if (decode_gap && !terminates) {
+          const bool reaches_gfx1250_zero =
+              arch == ROCJITSU_CODE_ARCH_GFX1250 && next_offset < section_end &&
+              inst_data[next_offset / sizeof(uint32_t)] == 0;
+          if (reaches_gfx1250_zero)
+            current->has_terminator_ = true;
```
kuhar, line 273: We should keep generic reachable zero words fail-closed. The decoder now skips `0x00000000` and marks the preceding block as terminated, but no terminating instruction is emitted. As a result, `{s_nop, 0, s_endpgm}` translates successfully even though the first block can execute past its last emitted `s_nop` into relocated bytes. Please restrict the exception to a proven compiler-stub or alignment boundary and keep ordinary reachable decode gaps rejected; the regression should inspect the translated words/control flow rather than only `result.ok()`.

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/dbt/semantic/gfx1250_b0_to_a0.cpp
```diff
@@ -571,20 +676,108 @@ ExpandResult expand_gfx1250_wmma_scale16(const Instruction &inst, uint32_t, uint
...
+
+  SemanticScratchAllocator allocator(
+      inst, liveness, context,
+      SemanticScratchPolicy{.max_vgprs = 256,
+                            .max_spill_dword_offset = kGfx1250ScratchMaxDwordOffset});
+  SemanticScratchRequest matrix_request;
+  matrix_request.count = static_cast<uint16_t>(matrix_a_width + 5u);
+  matrix_request.alignment = 2;
+  matrix_request.forbidden = architectural_vgprs;
+  matrix_request.allow_spill = true;
```
kuhar, line 708: It would be useful to exercise the new spill-backed Scale16 masked-A path before merging. This change enables `allow_spill` and adds VSCRATCH save/restore plus `s_waitcnt` bracketing, but the current Scale16 tests leave enough low VGPRs free that `lease.spilled` is never true. Please add a high-VGPR-pressure regression modeled on `Cdna4ToCdna3Bitop3UsesSpillBackedScratchWhenVgprsAreFull` that forces a borrowed range and checks the scratch store/load, waits, hazard padding, and restoration. A focused fail-closed case should also verify that an allocation or preservation failure leaves `elf_bytes` unchanged.

kuhar (commented):
Two comments

emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/code/basic_block.cpp
kuhar: @Groverkss said this is as intended

atgutier (commented):
No blockers so this is approved with comments from the agent. Please make sure to add a detailed commit message.

 ## MAJOR                                                                                                                                                                                     

  [MAJOR] lib/rocjitsu/src/rocjitsu/code/dbt/semantic/gfx1250_b0_to_a0.cpp:881-886 — final Scale16 WMMA can have its temporary input overwritten by the next source VALU

  Reasoning: The Scale16 lowering allocates a temporary masked_a VGPR range. An unspilled range only needs to be dead before the source WMMA, so the allocator can select a VGPR that the immediately following
  source instruction defines but never subsequently uses.

  The generated replacement protects masked_a between its internal WMMA passes. It also pads after the final WMMA when spill-restoration loads will overwrite the range. It emits no final padding for an
  unspilled range. A following VALU definition can therefore co-execute inside the final WMMA’s one- or three-slot input-read window and corrupt the matrix input. This directly affects the newly enabled four-
  pass M=32 Scale16 path, which failed closed before this PR.

  Evidence:

  - gfx1250_b0_to_a0.cpp:658-661 defines one protected VALU slot for FP4/FP4 and three for the other formats.
  - :869-879 emits the required padding between generated passes.
  - :881-886 emits final padding only when matrix_scratch.lease->spilled.
  - semantic_scratch.cpp:61-73 selects unspilled scratch using find_free_run().
  - liveness.cpp:281-296 checks liveness only before the source instruction. A dead destination of the next VALU is therefore eligible.
  - Existing tests end with s_endpgm or exercise the spilled restoration case; none places a dead VALU definition of the selected unspilled scratch immediately after Scale16.

  Suggested fix: Emit wmma_valu_hazard_slots V_NOPs after the final generated WMMA unconditionally, before restoration/mode cleanup and before returning to the original instruction stream. Add an unspilled
  M=32 FP4 case followed by a dead write to the selected scratch range, plus a non-FP4 case verifying all three slots.

  Confidence: high.

  [MAJOR] lib/rocjitsu/src/rocjitsu/analysis/indirect_branch_discovery.cpp:1795 — new cross-block lane-stash analysis ignores external entry edges

  Reasoning: The public discovery API receives all known kernel entry offsets through extra_leaders, but the new cross-block vector-lane analysis drops that list. It treats only block zero and predecessorless
  blocks as externally entered.

  If a nonzero kernel entry is also reachable by a direct edge from another entry, the analysis considers only the internal predecessor state. A PC target stashed by the first entry can therefore be propagated
  into a readlane/swappc in the second entry. When the second kernel is launched externally, those VGPR lanes are unconstrained, yet DBT may have replaced its dynamic transfer with the stale recovered direct
  target.

  This is a regression from the previous block-local analysis for branch-separated blocks: that implementation cleared lane state at every direct branch leader.

  Evidence:

  - discover_indirect_branch_edges() accepts extra_leaders at :2150-2159, but calls recover_vector_lane_stashed_pcs(ctx, recovered) without them.
  - Lane recovery builds its graph with build_analysis_blocks(ctx, {}) at :1795.
  - :1974-2015 adds external state only for block zero or blocks with no predecessors.
  - basic_block.cpp:210-211 passes known kernel entries into discovery.
  - tests/analysis/liveness_test.cpp:437-449 already demonstrates that an extra kernel-entry leader may still have a predecessor.
  - binary_translator.cpp:1460-1490 converts a complete single-target recovery into a direct transfer window.

  Suggested fix: Mark section entry and every extra_leader as an explicit external-entry edge. Meet an unconstrained external state into both the SGPR-pair and lane-stash analyses even when the block has
  ordinary predecessors. For lane state, the external contribution must clear stashed slots and initialize VGPR-bank state according to the entry contract. Add a two-entry test where entry A branches into
  entry B and carries a stash, while B is also independently launchable; B’s consumer must remain incomplete or unrecovered.

  Confidence: high.

 [MAJOR] lib/python/amdisa/codegen/_generator.py:1996-2004 — shared DPP generator fix was regenerated only for gfx1250

  Reasoning: The generator now gives every DPP-capable VOP encoding owned raw storage and counts the extension DWORD for 64-bit base encodings. That logic is target-independent, but only gfx1250’s generated
  encodings.{h,cpp} was committed. Other DPP-capable generated targets no longer match the generator.

  For example, the checked-in RDNA4 decoder still treats a 12-byte VOP3+DPP instruction as its eight-byte base instruction. This is an existing runtime defect in that generated target, but this PR makes the
  shared generator authoritative for the fix while leaving its generated tree inconsistent.

  Evidence:

  - _generator.py:1996-2004 computes DPP ownership and explicit size without a gfx1250 architecture gate.
  - :2137-2149 adds the extension size and copies all instruction words into owned storage.
  - isa/arch/amdgpu/rdna4/machine_insts.h:339-378 defines three-word VOP3 DPP16/DPP8 layouts.
  - RDNA4 encodings.cpp:192-201 still omits the DPP extension from Vop3::size_.
  - RDNA4 encodings.h:517-538 has no corresponding raw_words_ storage.
  - The added generator regression exercises only the gfx1250 fixture.
  - CONTRIBUTING.md:94-98 and docs/codegen.md:153-157 require multi-ISA generation and committing all affected outputs.

  Suggested fix: Run the documented multi-ISA regeneration and commit every changed generated encoding file. If this is intentionally gfx1250-only, architecture-gate the generator behavior instead. Add a
  parameterized generator/runtime test for every DPP-capable profile that verifies VOP3 DPP size and preservation of the extension word.

  Confidence: high.

kuhar (approved):

Issue #9130, opened 2026-07-23
rocprofiler-sdk: firmware-version-based feature enablement


bwelton:
### Summary

The firmware restrictions mechanism in rocprofiler-sdk (`source/lib/rocprofiler-sdk/counters/firmware_restrictions.*`, data in `share/rocprofiler-sdk/config.yaml`) currently only enforces **hard minimum-firmware-version floors** at startup (warning-only). There is no way for individual services to enable/disable a specific capability based on whether the installed firmware supports it.

A first building block has been added (see the associated PR): schema v2 introduces an optional per-entry `feature:` field and an internal query API:

```cpp
std::optional<bool>
rocprofiler::counters::agent_supports_feature(const rocprofiler_agent_t* agent,
                                              std::string_view feature);
```

- `nullopt` — no matching entry (feature undeclared / arch mismatch / invalid agent)
- `false`  — a matching feature floor exists and the agent is **below** it
- `true`   — matching feature floor(s) exist and are **all** met

Feature-scoped entries are excluded from the startup hard gate — they are pure capability metadata that services query at runtime and decide, per service, whether the capability is required.

### What this issue tracks

Wire up **actual feature enablement** on top of the query layer:

- [ ] Add consumers that call `agent_supports_feature()` to gate functionality. First candidate: PC sampling host-trap mode in `source/lib/rocprofiler-sdk/pc_sampling/…` (e.g. `is_pc_sampling_method_supported()`), rejecting only an explicit `false` and preserving current behavior on `nullopt`.
- [ ] Populate `config.yaml` with the real firmware feature floors per architecture/firmware type (CP/MEC/SDMA) for the capabilities we actually gate.
- [ ] Decide policy for `nullopt` (undeclared) at each call site (fail-open vs. fail-closed) and document it.
- [ ] Consider caching/observability: the query reads the parsed config once (cached); confirm this is sufficient as more call sites are added.
- [ ] Evaluate whether a public C API is needed so external tools can query firmware feature support (currently internal C++ only).
- [ ] Broaden firmware-type coverage if features depend on firmware beyond CP/MEC/SDMA.

### Context / references

- Query layer + schema v2: see linked PR.
- Semantics decision: feature entries are metadata only; services own the "is this feature required for me" decision.

PR #1159, opened 2025-09-29
[rocprofiler-sdk] Firmware detection in sdk along with checking fw before enabling pc-sampling

sauverma93:
## Motivation

This PR is a re-work of #196 

This PR does the following:

Parses the firmware info for different engines and micro-controllers that are listed in sysfs path. This is done in read_topology()
Reads the info into the new API struct rocprofiler_agent_fw_info_t
Stores the rocprofiler_agent_fw_info_t inside the API struct rocprofiler_agent_v0_t
The pc-sampling service functions - rocprofiler_query_pc_sampling_agent_configurations() and rocprofiler_configure_pc_sampling_service() both functions check firmware as part of their logic
This PR contains API struct updates:

- updated rocprofiler_agent_v0_t
- added a new struct rocprofiler_agent_fw_info_t
- added a new error message enum 'ROCPROFILER_STATUS_FIRMWARE_INCOMPATIBLE' in rocprofiler_status_t

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #9131, opened 2026-07-23
feat(rocprofiler-sdk): add firmware feature-support query layer

bwelton:
## Summary

Extend the rocprofiler-sdk firmware restrictions mechanism so a minimum-firmware-version floor can be scoped to a named capability instead of only being a global hard requirement. This adds the query layer that lets services ask, at runtime, whether the installed firmware supports a specific feature on a given agent.

## Changes

- Add an optional per-entry `feature` field to the `firmware_restrictions` schema (bumped `fw-restriction-schema-version` to `2`). Entries with a `feature` are capability metadata and are **excluded** from the startup hard gate; entries without one remain hard floors enforced at startup (unchanged behavior).
- Add `agent_supports_feature(const rocprofiler_agent_t*, std::string_view)` returning `std::optional<bool>`: `nullopt` = no matching entry (undeclared / arch mismatch / invalid agent), `false` = a matching floor exists and the agent is below it, `true` = all matching floors met. Services decide for themselves whether a `nullopt`/`false` result is fatal.
- Factor the core matching into a pure `evaluate_feature_support()` so the tri-state logic is unit-testable without on-disk config or a live agent; share the arch-match and firmware-type-select helpers with the existing startup gate; cache the parsed installed restrictions.
- Add an illustrative `feature` entry and schema-version docs; strip a pre-existing trailing-whitespace line in the edited config block.
- Add deterministic unit tests: `feature` parsing, startup-gate exclusion, schema-version range, invalid-agent handling, and a table-driven evaluator test covering undeclared / arch-mismatch / below-floor / all-met / empty-arch / unknown-firmware-type / empty-feature cases.

This is a foundational query layer; wiring specific consumers (e.g. PC sampling host-trap) is tracked separately in the linked issue.

## Testing

Built and ran on a gfx942 (MI300) host inside the `therock-lttng-gfx942` ROCm 7.14 container (amdclang++ 23.0.0git, cmake 3.28):

- `cmake -B build -S . -G Ninja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_PREFIX_PATH=/opt/rocm -DROCPROFILER_BUILD_TESTS=ON`
- `cmake --build build --target counter-tests -j 32` — compiles and links cleanly, no warnings/errors on the changed files.
- `./bin/counter-tests --gtest_filter='FirmwareRestrictions.*'` — **21/21 tests pass**, including the 5 new cases (`FeatureFieldParsing`, `FeatureEntryExcludedFromStartupGate`, `SchemaVersionRange`, `AgentSupportsFeatureInvalidAgent`, `EvaluateFeatureSupport`).

`clang-format` and `check-yaml` pre-commit hooks pass on the changed files.

## Tracking

Issue #9130

### DIFF
--- a/projects/rocprofiler-sdk/source/share/rocprofiler-sdk/config.yaml
+++ b/projects/rocprofiler-sdk/source/share/rocprofiler-sdk/config.yaml
@@ -15194,7 +15194,7 @@ rocprofiler-sdk:
             - gfx1151
             - gfx1152
           expression: reduce(TCP_TCP_TA_REQ_STALL,sum)
-  fw-restriction-schema-version: 1
+  fw-restriction-schema-version: 2
   firmware_restrictions:
     - affected_architectures:
         - gfx940
--- a/projects/rocprofiler-sdk/source/docs/api-reference/counter_collection_services.rst
+++ b/projects/rocprofiler-sdk/source/docs/api-reference/counter_collection_services.rst
@@ -333,7 +333,7 @@ Firmware restrictions are defined alongside counter definitions in the ``config.
               expression: 400*reduce(SQ_WAIT_INST_LDS,sum)/reduce(SQ_WAVES,sum)/reduce(GRBM_GUI_ACTIVE,max)
 
       # Required: Firmware restrictions schema version
-      fw_restriction_schema_version: 1
+      fw-restriction-schema-version: 2
 
       # List of firmware restrictions
       firmware_restrictions:
@@ -355,9 +355,18 @@ Firmware restrictions are defined alongside counter definitions in the ``config.
             - "gfx1100"
             - "gfx1101"
 
+        # Example: per-feature capability floor (queried at runtime, not a hard
+        # startup requirement)
+        - firmware_type: CP
+          feature: pc_sampling_host_trap
+          min_version: 210
+          reason: "PC sampling host-trap mode requires CP firmware version 210 or newer on MI300 devices"
+          affected_architectures:
+            - "gfx942"
+
 **Schema elements:**
 
-- ``fw_restriction_schema_version`` (required): Integer specifying the schema version (currently 1).
+- ``fw-restriction-schema-version`` (required): Integer specifying the schema version (currently 2; the ``feature`` field was added in version 2).
 - ``firmware_restrictions``: Array of restriction objects consisting of the following fields:
 
   - ``firmware_type`` (required): Type of firmware being restricted. Supported types include:
@@ -368,6 +377,7 @@ Firmware restrictions are defined alongside counter definitions in the ``config.
   - ``min_version`` (required): Integer specifying the minimum firmware version.
   - ``reason`` (optional): Human-readable explanation for the restriction.
   - ``affected_architectures`` (optional): Array of GPU architecture names, such as "gfx940" and "gfx942" liable to meet this restriction. If empty, the restriction applies to all the architectures.
+  - ``feature`` (optional): String naming a firmware-gated feature. When present, the entry is treated as a *per-feature capability floor* rather than a hard startup requirement: it is **not** enforced during the startup firmware validation and does not prevent ROCprofiler-SDK from initializing. Instead, services query it at runtime (via the internal feature-support API) to decide whether functionality requiring that firmware feature can be enabled on a given agent. When ``feature`` is absent (or empty), the entry is a hard floor as before and is enforced at startup.
 
 Counter definitions file location
 ++++++++++++++++++++++++++++++++++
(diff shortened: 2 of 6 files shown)


projects/rocprofiler-sdk/source/share/rocprofiler-sdk/config.yaml
```diff
@@ -15204,3 +15204,9 @@ rocprofiler-sdk:
       min_version: 186
       reason: CP firmware below version 186 can cause device resets and VM Page Faults on MI3XX 
         devices.
+    - affected_architectures:
+        - gfx942
+      firmware_type: CP
+      feature: pc_sampling_host_trap
+      min_version: 210
+      reason: PC sampling host-trap mode requires CP firmware version 210 or newer on MI300 devices.
```
vlaindic: I guess this doesn't need to be part of this PR, it's only an example of the service that uses the FW feature restrictions, correct?

Also, we don't currently use this to verify whether host-trap is supported, correct?

I don't mind having pc_sampling mentioned here, but would like if @sauverma93 could ensure this and verify against his current work inside the #1159

vlaindic (approved):
Overall, I'm fine with the PR, so approving. Please respond to one question inlined and consider AI generated review comments. Thanks!

# Review comments for PR #9131

https://github.com/ROCm/rocm-systems/pull/9131

Copy/paste each block below as a separate PR review comment.

---

## Comment 1 — IMPORTANT: stale license header year

**File:** `projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/tests/firmware_restrictions_test.cpp`

Update copyright header in `firmware_restrictions_test.cpp` from 2025 to
2026 — this file is renamed from `firmware_restrictions.cpp` and
substantially rewritten in this PR (97 new lines, 5 new test cases), so
it should carry the current year.

---

## Comment 2 — SUGGESTION: clarify the `nullopt` caller contract

**File:** `projects/rocprofiler-sdk/source/lib/rocprofiler-sdk/counters/firmware_restrictions.hpp`

`agent_supports_feature()` / `evaluate_feature_support()` document the
tri-state *return* semantics well, but it's worth spelling out the
*caller* contract explicitly in the header comment: `std::nullopt` means
"the query could not be evaluated" (undeclared feature, arch mismatch, or
invalid agent) — it does not mean "feature unsupported." Only `false`
means an applicable floor was evaluated and not met. Suggest adding a
line such as:

> Note: callers must not treat `std::nullopt` as "feature unsupported" —
> it means the query could not be evaluated. Only `false` means an
> evaluated floor was not met.

This matters because this PR intentionally ships `agent_supports_feature()`
as a dormant query API with no wired consumer yet (per the PR description),
so future callers wiring it up (e.g. PC sampling host-trap enablement)
will rely on this doc comment to get the fatality decision right.

itrowbri (approved):

projects/rocprofiler-sdk/source/share/rocprofiler-sdk/config.yaml
bwelton: You're correct on both counts:

1. **It's only an illustrative example** — it demonstrates the new `feature` schema field, not a live restriction. The functional part of this PR is the schema + query layer (`agent_supports_feature()` / `evaluate_feature_support()`); this entry just shows how a service would declare a per-feature firmware floor.

2. **It is not used to verify host-trap support yet.** `agent_supports_feature()` ships as a dormant query API with no wired consumer in this PR — nothing currently calls it to gate host-trap PC sampling. Wiring consumers (host-trap enablement being the first candidate) is tracked in #9130.

I'm happy to drop the `pc_sampling_host_trap` entry entirely and use a neutral placeholder feature name to avoid implying host-trap is already gated, if that's preferable. @sauverma93 — could you confirm this doesn't conflict with your work in #1159? If the naming/min_version there differs, I'll align or remove it.

bwelton: Update: I've **removed the `pc_sampling_host_trap` example entry from the distributed `config.yaml`**. On reflection it shouldn't ship in the runtime data file — the `min_version: 210` was a placeholder, nothing consumes it yet, and shipping fabricated restriction data to users is wrong.

The shipped `config.yaml` now contains only the real, pre-existing CP/186 hard floor. The `feature` field is still fully demonstrated in the docs (`counter_collection_services.rst`) and exercised by the unit tests via inline YAML, which are the appropriate places for illustrative usage. Schema version stays at `2` since the parser supports the `feature` field regardless of whether any shipped entry uses it.

Real per-feature floors (with verified firmware versions) can be added when a consumer is wired up — tracked in #9130.

amd-hsivasun:
@bwelton This PR has caused conflicts on the internal side. Please resolve the conflicts here https://github.com/AMD-ROCm-Internal/rocm-systems/pull/1411

PR #9153, opened 2026-07-23
perf(rocjitsu): avoid repeated race trace disassembly

newling:
## Motivation

The race detector caches disassembly by instruction PC for use in race reports. It does this repeatedly, even when revisiting the same program count (PC). This overhead was visible in a emulated hipBLASLt workloads from ROCm/rocm-libraries#9363,. 

## Technical Details

Check the cache under an already existing mutex before formatting the instruction. The first disassembly for a PC is still retained and later observations do not replace it. A regression test verifies that a second instruction object at the same PC is not disassembled.

## Issue Tracking

JIRA ID: AIHPBLAS-3605
^ hipblast uses rocjitsu for race detection in CI

## Test Plan

- One unit test, and verify a speed-up locally on the workload in question 

## Test Result

- Race-detector overhead fell from approximately 19–27% to approximately 9–10%.

## Submission Checklist

### DIFF
--- a/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/race_detector/plugin.h
+++ b/emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/race_detector/plugin.h
@@ -72,11 +72,6 @@ struct DisasmCache {
     entries_.emplace(pc, inst.disassemble());
   }
 
-  void record(uint64_t pc, std::string disasm) {
-    std::lock_guard<std::mutex> lock(mutex_);
-    entries_.try_emplace(pc, std::move(disasm));
-  }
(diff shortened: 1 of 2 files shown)


emulation/rocjitsu/lib/rocjitsu/src/rocjitsu/vm/plugins/race_detector/plugin.h
amd-eochoalo: Is this method no longer used?

newling: It was only used in a test, I've removed it now

amd-eochoalo (approved):

PR #9166, opened 2026-07-23
rocr: Fix coredump generation for multi-XCC targets

lancesix:
## Motivation

ROCgdb testsuite revealed that core dumps produced by ROCr were invalid for multi XCC targets.

## Technical Details

Testing revealed 2 issues:
- The CWSR area size reported in the kfd_queue_snapshot_entry reported the number of bytes used by the wave region of the CWSR buffer, just for XCC0.  The debugger however needs to know the full size of reserved per XCC so it can deduce the base of each XCC save region.
- The filter for lightweight core dump was only marking save area region for XCC0 as required.  This was unlikely to be an issue in practice as the allocation covers all the XCCs so the filter should keep it completely, but this still needs fixing.
- The "array_count" field of kfd_dbg_device_info_entry contained the total number of arrays instead of the number of arrays per XCC.

This patch fixes both of those issues.

This patch raises questions regarding the fields contained in HsaQueueInfo.  All the CWSR related fields in this struct only contain XCC0 status, which is probably not sufficient for milti-XCC systems. This patch does not aim at addressing all of the uses of this structure, instead it adds the necessary information required to make CWSR work correctly.

Tested on gfx1031 and gfx942 with ROCgdb's testsuite.

## Issue Tracking

JIRA ID: AIROCGDB-637

## Test Plan

Validate using ROCgdb's testsuite locally, as CI is not able to validate coredump creation.  Need validation on all major GPU families.

## Test Result

Test pass on gfx1031 and gfx942 so far.

## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/libhsakmt/include/hsakmt/hsakmttypes.h
+++ b/projects/rocr-runtime/libhsakmt/include/hsakmt/hsakmttypes.h
@@ -770,6 +770,7 @@ typedef struct
 	HSAuint64 ControlStackUsedInBytes; // Must be 4-Byte aligned
 	HsaUserContextSaveAreaHeader *SaveAreaHeader;
 	HSAuint64 Reserved2;		// runtime/system CU assignment
+	HSAuint64 SaveAreaAllocSize;	// Size of the full save area, per XCC
 } HsaQueueInfo;
 
 typedef struct _HsaQueueResource
--- a/projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/lnx/amd_core_dump.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/lnx/amd_core_dump.cpp
@@ -345,10 +345,11 @@ static hsa_status_t build_lightweight_coredump_ranges(MemoryRegionFilter& filter
 
       debug_print("Added CWSR area range: %#" PRIx64 " - %#" PRIx64 " (size: %zu)\n",
                   reinterpret_cast<uint64_t>(queue_info.SaveAreaHeader),
-                  reinterpret_cast<uint64_t>(queue_info.SaveAreaHeader) + queue_info.SaveAreaSizeInBytes,
-                  queue_info.SaveAreaSizeInBytes);
+                  reinterpret_cast<uint64_t>(queue_info.SaveAreaHeader)
+                    + (queue_info.SaveAreaAllocSize * gpu_agent->properties().NumXcc),
+                  queue_info.SaveAreaAllocSize * gpu_agent->properties().NumXcc);
       filter.add_range(reinterpret_cast<uint64_t>(queue_info.SaveAreaHeader),
-                       queue_info.SaveAreaSizeInBytes);
+                       queue_info.SaveAreaAllocSize * gpu_agent->properties().NumXcc);
     }
   }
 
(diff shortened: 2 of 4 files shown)


projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/lnx/amd_core_dump.cpp
```diff
@@ -425,7 +426,7 @@ static bool GetCoreQueueInfo(AMD::AqlQueue* queue, kfd_queue_snapshot_entry& ent
     return false;
   }
   entry.ctx_save_restore_address = (uint64_t)queue_info.SaveAreaHeader;
-  entry.ctx_save_restore_area_size = (uint32_t)queue_info.SaveAreaSizeInBytes;
+  entry.ctx_save_restore_area_size = (uint32_t)queue_info.SaveAreaAllocSize;
```
cfreeamd, line 429: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Question]** `entry.ctx_save_restore_area_size` is set to the per-XCC `SaveAreaAllocSize`, while the memory range added to the dump filter just above uses `SaveAreaAllocSize * NumXcc`. `kfd_queue_snapshot_entry` has no `num_xcc` field (it's in `kfd_dbg_device_info_entry`), so this only works if the consumer (rocgdb) reconstructs the full per-queue size as `ctx_save_restore_area_size × device num_xcc`. That's consistent with the `array_count` change in `debug.c` (also now per-XCC), so this is likely correct-by-design — just confirming the consumer scales per-XCC values by `num_xcc`. If rocgdb instead expects the full span in this field, it would be under-reported by a factor of NumXcc on multi-XCC parts.

projects/rocr-runtime/libhsakmt/include/hsakmt/hsakmttypes.h
```diff
@@ -770,6 +770,7 @@ typedef struct
 	HSAuint64 ControlStackUsedInBytes; // Must be 4-Byte aligned
 	HsaUserContextSaveAreaHeader *SaveAreaHeader;
 	HSAuint64 Reserved2;		// runtime/system CU assignment
+	HSAuint64 SaveAreaAllocSize;	// Size of the full save area, per XCC
```
cfreeamd, line 773: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** `SaveAreaAllocSize` is appended right after the unused `HSAuint64 Reserved2`; repurposing the reserved slot would have avoided growing `HsaQueueInfo` (not an ABI risk here since libhsakmt is statically linked into libhsa-runtime64, but worth noting for how the struct evolves).

Also consider clarifying the comment to contrast the new field with the existing `SaveAreaSizeInBytes` — allocated (per XCC) vs used (per XCC) — since three call sites still read `SaveAreaSizeInBytes` and the distinction is easy to misuse.

cfreeamd (commented):

The multi-XCC save-area range math in `build_lightweight_coredump_ranges` (`SaveAreaAllocSize * NumXcc` matching the contiguous per-XCC CWSR layout) and the `queues.c` / `hsakmttypes.h` plumbing look internally consistent and correct. The `debug.c` `/ NumXcc` is safe (topology normalizes NumXcc to >=1). Left one question on the per-XCC snapshot-entry size vs the full `× NumXcc` dump range, plus two nits.

projects/rocr-runtime/runtime/hsa-runtime/libamdhsacode/lnx/amd_core_dump.cpp
lancesix: The consumer (ROCdbgapi) needs those values to be per-XCC, and does the scaling when appropriate.

lumachad:
Validated on gfx942 and gfx1201 on top of https://github.com/ROCm/ROCgdb/pull/160. gdb.rocm/runtime-core.exp is clean.

cfreeamd (approved):

amd-bfilipov:
Validated on gfx90a on top of https://github.com/ROCm/ROCgdb/pull/160.

kentrussell (approved):
Seems good to me

PR #9167, opened 2026-07-24
fix(clr): load KPACK code objects per device

### LINKED ISSUE #6696: [Issue]: pytorch gpu segfault when using multiple gpus
### Problem Description

When pytorch application is using more than one GPU for the calculations on same application, then the usage of second GPU will cause the segfault. I have tested this with the 2 different combinations of GPUs.

1) gfx1100 and gfx1030 (AMD Radeon Pro W7800 48GB and AMD Radeon Pro W7800 48GB on shark39 lab machine)
2) gfx1100 and gfx90a (Radeon Pro W7900 and Instinct MI210 on own pc)

I have not yet tested this in computer which have 2 similar GPUs like 2 x RX 9070 on same PC.

Application that I use for testing looks all available GPUs and then does math-calculation which each of them in a loop. Calculation always works on the first GPU, but then when it is tried to perform on the second one it will segfault.

If the visibility of the GPUs is restricted so that the application only see either the first or second gpu, before starting the test it will work. (By using "export HIP_VISIBLE_DEVICES=0" or "export HIP_VISIBLE_DEVICES=1"

Problem only happens on pytorch from the TheRock builds, pytorch from ROCm 7.2.4 works.

### Operating System

Ubuntu 24.04

### CPU

AMD Ryzen Threadripper PRO 7975WX 32-Cores

### GPU

gfx90a, gfx1100, gfx1030 (AMD MI210,  Radeon PRO W7800, Radeon RX 6950 XT)

### ROCm Version

rocm 7.14, rocm nightly from 20260717

### ROCm Component

_No response_

### Steps to Reproduce

Test application pytorch_multi_gpu_test.py:

```
import torch

print("Pytorch version: " + torch.__version__)
print("ROCM HIP version: " + torch.version.hip)
if torch.cuda.is_available():
    dev_cnt=torch.cuda.device_count()
    print("CUDA device count: " + str(dev_cnt))
    for ii in range(dev_cnt):
        try:
            cuda_dev_id = "cuda:" + str(ii)
            cuda_dev_name = torch.cuda.get_device_name(ii)
            print(f"{cuda_dev_id}: {cuda_dev_name}")
            print(f"    running test: torch.zeros(1, device=\"{cuda_dev_id}\")")
            torch.zeros(1, device=cuda_dev_id)
            print("Test ok")
        except Exception as e:
            print(f"An error occurred with device: {cuda_dev_name}")
            print(f"{e}")
```

Failure on gfx1100 and gfx1030 combination with therock nightly build:

```
lamikr@Shark39:~$ python3 -m venv .venv_nightly_20260716 && source .venv_nightly_20260716/bin/activate
(.venv_nightly_20260717) lamikr@Shark39:~$ pip install --index-url https://rocm.nightlies.amd.com/whl-multi-arch torch[device-all] torchvision[device-all] torchaudio rocm[devel,device-all]

(.venv_nightly_20260717) lamikr@Shark39:~$ python pytorch_multi_gpu_test.py 
Pytorch version: 2.12.0+rocm7.15.0a20260717
ROCM HIP version: 7.15.0
CUDA device count: 2
cuda:0: AMD Radeon Pro W7800 48GB
    running test: torch.zeros(1, device="cuda:0")
Test ok
cuda:1: AMD Radeon RX 6950 XT
    running test: torch.zeros(1, device="cuda:1")
Segmentation fault (core dumped)
```

Same works on same nihgtly build if only one GPU is visible at a time:

```
(.venv_nightly_20260717) lamikr@Shark39:~$ export HIP_VISIBLE_DEVICES=0
(.venv_nightly_20260717) lamikr@Shark39:~$ python pytorch_multi_gpu_test.py 
Pytorch version: 2.12.0+rocm7.15.0a20260717
ROCM HIP version: 7.15.0
CUDA device count: 1
cuda:0: AMD Radeon Pro W7800 48GB
    running test: torch.zeros(1, device="cuda:0")
Test ok

(.venv_nightly_20260717) lamikr@Shark39:~$ export HIP_VISIBLE_DEVICES=1
(.venv_nightly_20260717) lamikr@Shark39:~$ python pytorch_multi_gpu_test.py 
Pytorch version: 2.12.0+rocm7.15.0a20260717
ROCM HIP version: 7.15.0
CUDA device count: 1
cuda:0: AMD Radeon RX 6950 XT
    running test: torch.zeros(1, device="cuda:0")
Test ok
```

Test will also pass with both GPUs in use with therock 7.2.4 build.

```

lamikr@Shark39:~$ python3 -m venv .venv_rocm_724 && source .venv_rocm_724/bin/activate
(.venv_rocm_724) lamikr@Shark39:~$ wget -c https://repo.radeon.com/rocm/manylinux/rocm-rel-7.2.4/torch-2.10.0%2Brocm7.2.4.lw.git3d3aa833-cp312-cp312-linux_x86_64.whl
(.venv_rocm_724) lamikr@Shark39:~$  wget -c https://repo.radeon.com/rocm/manylinux/rocm-rel-7.2.4/triton-3.6.0%2Brocm7.2.4.git4ed88892-cp312-cp312-linux_x86_64.whl
(.venv_rocm_724) lamikr@Shark39:~$ pip install torch-2.10.0+rocm7.2.4.lw.git3d3aa833-cp312-cp312-linux_x86_64.whl triton-3.6.0+rocm7.2.4.git4ed88892-cp312-cp312-linux_x86_64.whl

(.venv_rocm_724) lamikr@Shark39:~$ python pytorch_multi_gpu_test.py 
Pytorch version: 2.10.0+rocm7.2.4.git3d3aa833
ROCM HIP version: 7.2.53211
CUDA device count: 2
cuda:0: AMD Radeon PRO W7800 48GB
    running test: torch.zeros(1, device="cuda:0")
Test ok
cuda:1: AMD Radeon RX 6950 XT
    running test: torch.zeros(1, device="cuda:1")
Test ok
```

<details>

```
Paste output here
```

</details>

### Additional Information

_No response_

lamikr:
KPACK previously selected one kpack code object belonging for the first device in a
combined architecture priority list and registered it for every visible device. 
On heterogeneous GPU systems, this assigns a code object belonging to first ISA device type
also to all other incompatible devices types on the system, causing a segmentation fault during kernel launch.

Pytorch application segfaulted if user tried to use cuda:1 device instead of default cuda:0 in systems which
have two or more different types of GPUs installed and visible. (Not hidden with HIP_VISIBLE_DEVICES)

Fix is to get a list of all ISA-device types on the system and try to load the device type specific kpack
object for each of device type and then register them for all GPUs that are same type.

There is however also a possibility that user has installed the rocm support only for some of the
supported device types which will cause that code object is found only for some of the device types.
User could for example have gfx90a and gfx1100 and has installed rocm device support only for gfx1100 by using
command: 

$ pip install --index-url https://rocm.nightlies.amd.com/whl-multi-arch torch[device-gfx1100] numpy
instead of using
$ pip install --index-url https://rocm.nightlies.amd.com/whl-multi-arch torch[device-gfx90a,device-gfx1100] numpy
or
$ pip install --index-url https://rocm.nightlies.amd.com/whl-multi-arch torch[device-all] numpy

Therefore the updated code will not return error if device type specific code is not found but instead will skip that
device type and try to load the device object for next device type.  Reason for this is that it is possible that user wants to use only a certain type of devices in it's applications and has by purpose uninstalled the kpack files for other types of GPUs. 

Instead of segfaulting, Pytorch application will after this patch "throw torch.AcceleratorError: CUDA error: invalid kernel file" if the device type specific kpack files are not installed/available/loaded for the GPU associated to cuda:x index that application try to use. 

Tools used:
        GPT-5.6 Sol Medium

Fixes: https://github.com/ROCm/TheRock/issues/6696

## Motivation

This fixes pytorch crash on systems which have more than 1 different GPUs
and pytorch application try to use other than default cuda/cuda:0 gpu instead
of making that GPU to be a first one with HIP_VISIBLE_DEVICES env variable.

## Technical Details

The problem was that old code assumed that all GPU devices in system
where same type and only loaded the GPU arch-type specific kpack code object
for the first GPU in system. If some of the GPUs on system where different type than
the first visible GPU, then the code object load/execution on that GPU failed causing
a segfault.

If all of the GPUs on system where similar (2 x gfx1201 for example) then there were not problem.
But if there were different types of GPUs like gfx90a and gfx1100 on same system
and application wanted to execute the code by using the CUDA:1 device,
then that caused segfault because there were only the device type code object loaded for CUDA:0 gpu.

Problem required that the ROCM and Pytorch binaries were converted to KPACK format. 

## Issue Tracking

https://github.com/ROCm/TheRock/issues/6696

## Test Plan

I have tested this issue manually 3 systems.
- system with 2 x gfx1251 (Worked also without fix because both GPUs were on same type)
- system with 1 x gfx90a + 1 x gfx1100 (Segfaults without the fix, works with the fix)
- system with 1 x gfx1100 + 1 x gfx1030 (Segfaults without the fix, works with the fix)
- system with only 1 GPU

To be able to land CI test code, we need to have dedicated machine with two different types of GPUs to run a following at least a following code:

```
import torch

print("PyTorch version:", torch.__version__, flush=True)
print("ROCm HIP version:", torch.version.hip, flush=True)
print(f"cuda:1: {torch.cuda.get_device_name(1)}", flush=True)
print(torch.zeros(1, device=f"cuda:1"), flush=True)

```

## Test Result

Tests described in https://github.com/ROCm/TheRock/issues/6696 now works without segfault.
More extensive test sequence which involves installing and uninstalling the device type specific kpack files for some GPU architectures is documented in comment section of this PR.

## Submission Checklist

### DIFF
--- a/projects/clr/hipamd/src/hip_fatbin.cpp
+++ b/projects/clr/hipamd/src/hip_fatbin.cpp
@@ -758,60 +758,68 @@ hipError_t FatBinaryInfo::ExtractKpackBinary(const std::vector<hip::Device*>& de
     return hipErrorInvalidValue;
   }
 
-  // Build architecture priority list from devices
-  // For each device, add native ISA first, then generic fallback.
-  // Reservation in the list is pessimistically assuming there is a generic fallback for each
-  // device.
-  std::vector<std::string> arch_list;
-  arch_list.reserve(devices.size() * 2);
+  // Add KPACK load objects for each device.
+  // For each device type, the KPACK load returns one code object
+  // selected from the architecture priority list by using devices
+  // isa-name or generic name.
+  // Code object needs to be loaded for each device in system separately because
+  // there can be different types of GPUs on the same system and code-object
+  // is GPU type specific. (gfx1201, gfx90a)
   for (auto device : devices) {
+    std::vector<std::string> arch_list;
+    // Architecture names
+    // 1) exact devices ISA name, examples:
+    //  - amdgcn-amd-amdhsa--gfx1100
+    //  - amdgcn-amd-amdhsa--gfx90a:sramecc+:xnack-
+    // 2) generic fallback name, examples:
+    //  - amdgcn-amd-amdhsa--gfx11-generic
+    //  - can also be empty string for some arch like gfx90a
+    arch_list.reserve(2);
     std::string device_name = device->devices()[0]->isa().isaName();
     arch_list.push_back(device_name);
 
-    // Add generic fallback
+    // Add generic fallback arch-name
     auto generic_name = TargetToGeneric(device_name);
     if (!generic_name.empty()) {
       arch_list.push_back(generic_name);
     }
-  }
 
-  // Convert to C-style array for kpack API
-  std::vector<const char*> arch_ptrs;
-  arch_ptrs.reserve(arch_list.size());
-  for (const auto& arch : arch_list) {
-    arch_ptrs.push_back(arch.c_str());
-  }
+    // Convert arch-list to C-style array for kpack API
+    std::vector<const char*> arch_ptrs;
+    arch_ptrs.reserve(arch_list.size());
+    for (const auto& arch : arch_list) {
+      arch_ptrs.push_back(arch.c_str());
+    }
 
-  // Load code object from kpack archive
-  void* code_object = nullptr;
-  size_t code_object_size = 0;
+    // Load device type specific code object from kpack archive
+    void* code_object = nullptr;
+    size_t code_object_size = 0;
 
-  // binary_path is used to resolve relative paths to kpack archives.
-  // bundle_index identifies which code object to load for multi-TU binaries.
-  // The kernel_name (used for TOC lookup) is embedded in the HIPK metadata.
-  kpack_error_t err =
-      kpack_load_code_object(getHipKpackCache(), params.metadata, fname_.c_str(),
-                             static_cast<uint32_t>(params.bundle_index),
-                             arch_ptrs.data(), arch_ptrs.size(), &code_object, &code_object_size);
+    // Binary_path is used to resolve relative paths to kpack archives.
+    // Bundle_index identifies which code object to load for multi-TU binaries.
+    // The kernel_name (used for TOC lookup) is embedded in the HIPK metadata.
+    kpack_error_t err =
+        kpack_load_code_object(getHipKpackCache(), params.metadata, fname_.c_str(),
+                               static_cast<uint32_t>(params.bundle_index), arch_ptrs.data(),
+                               arch_ptrs.size(), &code_object, &code_object_size);
 
-  if (err != KPACK_SUCCESS) {
-    LogPrintfError("kpack_load_code_object failed with error: %d", err);
-    return hipErrorInvalidImage;
-  }
+    if (err != KPACK_SUCCESS) {
+      LogPrintfError("kpack_load_code_object failed with error: %d; isa name: %s", err,
+                     device_name.c_str());
+      return hipErrorInvalidImage;
+    }
 
-  // Add code object to all devices. The kpack buffer isn't backed by a file
-  // on disk, so no fd is passed.
-  for (auto device : devices) {
+    // The kpack buffer isn't backed by a file on disk, so no fd is passed.
     hipError_t hip_err =
         AddDevProgram(device, code_object, code_object_size, amd::Os::FDescInit());
     if (hip_err != hipSuccess) {
       kpack_free_code_object(code_object);
       return hip_err;
     }
-  }
 
-  // Track allocation for cleanup in destructor
-  code_obj_allocations_.insert(code_object);
+    // Track allocation for cleanup in destructor
+    code_obj_allocations_.insert(code_object);
+  }
 
   return hipSuccess;
 #endif
--- a/projects/clr/hipamd/src/hip_fatbin.cpp
+++ b/projects/clr/hipamd/src/hip_fatbin.cpp
@@ -758,22 +758,24 @@ hipError_t FatBinaryInfo::ExtractKpackBinary(const std::vector<hip::Device*>& de
     return hipErrorInvalidValue;
   }
 
-  // Load KPACK code objects per device.
-  // For each device, kpack selects a code object from an architecture
-  // priority list (device ISA name first, then a generic fallback).
-  // This must be done per device because heterogeneous systems can have
-  // different GPU ISAs (e.g., gfx1201 vs gfx90a).
+  // Load one KPACK code object per unique device ISA. Devices with the
+  // same ISA can share the extracted buffer, while heterogeneous devices
+  // require separate architecture selection.
+  std::unordered_map<std::string, std::vector<hip::Device*>> devices_by_isa;
   for (auto device : devices) {
+    devices_by_isa[device->devices()[0]->isa().isaName()].push_back(device);
+  }
+
+  for (const auto& [device_name, matching_devices] : devices_by_isa) {
     std::vector<std::string> arch_list;
     // Architecture names
-    // 1) exact devices ISA name, examples:
+    // 1) exact device ISA name, examples:
     //  - amdgcn-amd-amdhsa--gfx1100
     //  - amdgcn-amd-amdhsa--gfx90a:sramecc+:xnack-
     // 2) generic fallback name, examples:
     //  - amdgcn-amd-amdhsa--gfx11-generic
     //  - can also be empty string for some arch like gfx90a
     arch_list.reserve(2);
-    std::string device_name = device->devices()[0]->isa().isaName();
     arch_list.push_back(device_name);
 
     // Add generic fallback arch-name
@@ -807,12 +809,19 @@ hipError_t FatBinaryInfo::ExtractKpackBinary(const std::vector<hip::Device*>& de
       return hipErrorInvalidImage;
     }
 
-    // The kpack buffer isn't backed by a file on disk, so no fd is passed.
-    hipError_t hip_err =
-        AddDevProgram(device, code_object, code_object_size, amd::Os::FDescInit());
-    if (hip_err != hipSuccess) {
-      kpack_free_code_object(code_object);
-      return hip_err;
+    // Add device type specific kpack code object buffer for each similar type of device.
+    // The kpack buffer is shared by devices with the same ISA and is not
+    // backed by a file on disk, so no fd is passed.
+    for (auto device : matching_devices) {
+      hipError_t hip_err =
+          AddDevProgram(device, code_object, code_object_size, amd::Os::FDescInit());
+      if (hip_err != hipSuccess) {
+        LogPrintfError(
+            "AddDevProgram failed with error: %d; isa name: %s; device id: %d", hip_err,
+            device_name.c_str(), device->deviceId());
+        kpack_free_code_object(code_object);
+        return hip_err;
+      }
     }
 
     // Track allocation for cleanup in destructor
--- a/projects/clr/hipamd/src/hip_fatbin.cpp
+++ b/projects/clr/hipamd/src/hip_fatbin.cpp
@@ -758,13 +758,11 @@ hipError_t FatBinaryInfo::ExtractKpackBinary(const std::vector<hip::Device*>& de
     return hipErrorInvalidValue;
   }
 
-  // Add KPACK load objects for each device.
-  // For each device type, the KPACK load returns one code object
-  // selected from the architecture priority list by using devices
-  // isa-name or generic name.
-  // Code object needs to be loaded for each device in system separately because
-  // there can be different types of GPUs on the same system and code-object
-  // is GPU type specific. (gfx1201, gfx90a)
+  // Load KPACK code objects per device.
+  // For each device, kpack selects a code object from an architecture
+  // priority list (device ISA name first, then a generic fallback).
+  // This must be done per device because heterogeneous systems can have
+  // different GPU ISAs (e.g., gfx1201 vs gfx90a).
   for (auto device : devices) {
     std::vector<std::string> arch_list;
     // Architecture names
--- a/projects/clr/hipamd/src/hip_fatbin.cpp
+++ b/projects/clr/hipamd/src/hip_fatbin.cpp
@@ -804,9 +804,9 @@ hipError_t FatBinaryInfo::ExtractKpackBinary(const std::vector<hip::Device*>& de
                                arch_ptrs.size(), &code_object, &code_object_size);
 
     if (err != KPACK_SUCCESS) {
-      LogPrintfError("kpack_load_code_object failed with error: %d; isa name: %s", err,
-                     device_name.c_str());
-      return hipErrorInvalidImage;
+      LogPrintfWarning("Could not load kpack object for %s, err: %d",
+                     device_name.c_str(), err);
+      continue;
     }
 
     // Add device type specific kpack code object buffer for each similar type of device.
@@ -817,8 +817,8 @@ hipError_t FatBinaryInfo::ExtractKpackBinary(const std::vector<hip::Device*>& de
           AddDevProgram(device, code_object, code_object_size, amd::Os::FDescInit());
       if (hip_err != hipSuccess) {
         LogPrintfError(
-            "AddDevProgram failed with error: %d; isa name: %s; device id: %d", hip_err,
-            device_name.c_str(), device->deviceId());
+            "Could not add kpack object for %s, device id: %d, err: %d",
+            device_name.c_str(), device->deviceId(), hip_err);
         kpack_free_code_object(code_object);
         return hip_err;
       }
--- a/projects/clr/hipamd/src/hip_fatbin.cpp
+++ b/projects/clr/hipamd/src/hip_fatbin.cpp
@@ -766,6 +766,25 @@ hipError_t FatBinaryInfo::ExtractKpackBinary(const std::vector<hip::Device*>& de
     devices_by_isa[device->devices()[0]->isa().isaName()].push_back(device);
   }
 
+  std::vector<int> registered_device_ids;
+  registered_device_ids.reserve(devices.size());
+  std::vector<void*> loaded_code_objects;
+  loaded_code_objects.reserve(devices_by_isa.size());
+
+  // Device and kpack code object cleanup method for error cases
+  auto rollback_kpack_state = [&]() {
+    for (int device_id : registered_device_ids) {
+      if (dev_programs_[device_id] != nullptr) {
+        dev_programs_[device_id]->release();
+        dev_programs_[device_id] = nullptr;
+      }
+    }
+    for (void* code_object : loaded_code_objects) {
+      code_obj_allocations_.erase(code_object);
+      kpack_free_code_object(code_object);
+    }
+  };
+
   for (const auto& [device_name, matching_devices] : devices_by_isa) {
     std::vector<std::string> arch_list;
     // Architecture names
@@ -803,29 +822,48 @@ hipError_t FatBinaryInfo::ExtractKpackBinary(const std::vector<hip::Device*>& de
                                static_cast<uint32_t>(params.bundle_index), arch_ptrs.data(),
                                arch_ptrs.size(), &code_object, &code_object_size);
 
-    if (err != KPACK_SUCCESS) {
-      LogPrintfWarning("Could not load kpack object for %s, err: %d",
-                     device_name.c_str(), err);
+    if (err == KPACK_ERROR_ARCHIVE_NOT_FOUND || err == KPACK_ERROR_ARCH_NOT_FOUND) {
+      LogPrintfWarning(
+          "Could not load device type specific kpack object for ISA %s, err: %d, host binary: %s",
+          device_name.c_str(), err, params.binary_path.c_str());
       continue;
     }
+    if (err != KPACK_SUCCESS) {
+      LogPrintfError(
+          "Failed to load device type specific kpack object for ISA %s, err: %d, "
+          "host binary: %s",
+          device_name.c_str(), err, params.binary_path.c_str());
+      rollback_kpack_state();
+      return hipErrorInvalidImage;
+    }
+
+    loaded_code_objects.push_back(code_object);
+    code_obj_allocations_.insert(code_object);
 
     // Add device type specific kpack code object buffer for each similar type of device.
     // The kpack buffer is shared by devices with the same ISA and is not
     // backed by a file on disk, so no fd is passed.
     for (auto device : matching_devices) {
+      registered_device_ids.push_back(device->deviceId());
       hipError_t hip_err =
           AddDevProgram(device, code_object, code_object_size, amd::Os::FDescInit());
       if (hip_err != hipSuccess) {
         LogPrintfError(
-            "Could not add kpack object for %s, device id: %d, err: %d",
+            "Could not add device type specific kpack object for %s, device id: %d, err: %d",
             device_name.c_str(), device->deviceId(), hip_err);
-        kpack_free_code_object(code_object);
+        rollback_kpack_state();
         return hip_err;
       }
     }
+  }
 
-    // Track allocation for cleanup in destructor
-    code_obj_allocations_.insert(code_object);
+  if (loaded_code_objects.empty()) {
+    // Return an error if no device-specific kpack code object was found for any device.
+    LogPrintfError(
+        "Could not find device type specific kpack code objects for any available device from "
+        "binary: %s",
+        params.binary_path.c_str());
+    return hipErrorInvalidKernelFile;
   }
 
   return hipSuccess;


projects/clr/hipamd/src/hip_fatbin.cpp
```diff
@@ -711,60 +711,68 @@ hipError_t FatBinaryInfo::ExtractKpackBinary(const std::vector<hip::Device*>& de
...
+  // Add KPACK load objects for each device.
+  // For each device type, the KPACK load returns one code object
+  // selected from the architecture priority list by using devices
+  // isa-name or generic name.
+  // Code object needs to be loaded for each device in system separately because
+  // there can be different types of GPUs on the same system and code-object
+  // is GPU type specific. (gfx1201, gfx90a)
   for (auto device : devices) {
+    std::vector<std::string> arch_list;
+    // Architecture names
```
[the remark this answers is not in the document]
lamikr: Addressed on latest commit.

projects/clr/hipamd/src/hip_fatbin.cpp
```diff
@@ -711,60 +711,68 @@ hipError_t FatBinaryInfo::ExtractKpackBinary(const std::vector<hip::Device*>& de
...
+    kpack_error_t err =
+        kpack_load_code_object(getHipKpackCache(), params.metadata, fname_.c_str(),
+                               static_cast<uint32_t>(params.bundle_index), arch_ptrs.data(),
+                               arch_ptrs.size(), &code_object, &code_object_size);
+
+    if (err != KPACK_SUCCESS) {
+      LogPrintfError("kpack_load_code_object failed with error: %d; isa name: %s", err,
+                     device_name.c_str());
+      return hipErrorInvalidImage;
+    }
```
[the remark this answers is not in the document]
lamikr: Good finding, addressed on latest commit.

lamikr:
The Rock Multi-Arch Release test build with rocm-systems version with patches included passed: 
https://github.com/ROCm/TheRock/actions/runs/30290419155

projects/clr/hipamd/src/hip_fatbin.cpp
```diff
@@ -711,60 +711,75 @@ hipError_t FatBinaryInfo::ExtractKpackBinary(const std::vector<hip::Device*>& de
...
+    // The kernel_name (used for TOC lookup) is embedded in the HIPK metadata.
+    kpack_error_t err =
+        kpack_load_code_object(getHipKpackCache(), params.metadata, fname_.c_str(),
+                               static_cast<uint32_t>(params.bundle_index), arch_ptrs.data(),
+                               arch_ptrs.size(), &code_object, &code_object_size);
+
+    if (err != KPACK_SUCCESS) {
+      LogPrintfError("kpack_load_code_object failed with error: %d; isa name: %s", err,
+                     device_name.c_str());
+      return hipErrorInvalidImage;
```
lucbruni-amd, line 837: Is it possible to `continue` here (leaving this device unregistered) while other supported devices are iterated over?

lamikr: Good catch, I need to check this by doing a rocm build where the support has been build only for one device type to verify the behaviour.

In current code in nightly builds if the rocm libraries have support both for the cuda:0 and cuda:1 devices, the cuda:0 usage would work but the usage of cuda:1 device would always cause a crash. So patch fixes a following problem:

1) rocm libraries have build/installed support for both gpu device types
- machine has gfx1100 (cuda:0/default) and gfx90a (cuda:1) GPUs 
- application will try to use cuda:1" device
--> This will cause a crash on current 7.14 and nigtly builds.

But I will double check a following scenario

2) rocm libraries have build/installed support only for one cuda:0 or cuda:1 device type, not for both
- machine has gfx1100 (cuda:0/default) and gfx90a (cuda:1) GPUs 
- rocm sdk libraries has gfx1100 support build in/installed but not the gfx90a support
- application try to use only the cuda:0 device type, not cuda:1. (This works with current versions)
- I am not sure whether this leads now a new problem where the missing gfx90a/cuda:1 device code, would prevent also the usage of cuda:0. In that case, it would propably better just to printout warning for missing device code and continue.

lamikr: Ok, verified that the continue is needed instead of returning error if any of the code-objects are missing.

```
rocminfo | grep gfx
  Name:                    gfx1100
  Name:                    gfx90a

```

1) Original crash behavior with cuda:1 even when kpack code was installed for both gpus

```
$ pip list installed | grep torch
amd-torch-device-gfx1100 2.12.0
amd-torch-device-gfx90a  2.12.0
torch                    2.12.0
torchvision              0.27.1 

$ python pytorch_cuda_0_test.py 
PyTorch version: 2.12.0
ROCm HIP version: 7.15.0
Device count: 2
cuda:0: AMD Radeon Pro W7900 Dual Slot
tensor([0.], device='cuda:0')

$ python pytorch_cuda_1_test.py 
PyTorch version: 2.12.0
ROCm HIP version: 7.15.0
Device count: 2
cuda:1: AMD Instinct MI210
Segmentation fault (core dumped)
```

2) First fix attempt with returrning error instead of continue when some object code was not installed

```
$ pip list installed | grep torch
amd-torch-device-gfx1100 2.12.0
amd-torch-device-gfx90a  2.12.0
torch                    2.12.0
torchvision              0.27.1 

$ python pytorch_cuda_0_test.py 
PyTorch version: 2.12.0
ROCm HIP version: 7.15.0
Device count: 2
cuda:0: AMD Radeon Pro W7900 Dual Slot
tensor([0.], device='cuda:0')

$ pip uninstall amd-torch-device-gfx90a
$ python pytorch_cuda_0_test.py 
PyTorch version: 2.12.0
ROCm HIP version: 7.15.0
Device count: 2
cuda:0: AMD Radeon Pro W7900 Dual Slot
Traceback (most recent call last):
  File "/home/butler/bugs/therock_6696/pytorch_cuda_0_test.py", line 7, in <module>
    print(torch.zeros(1, device=f"cuda:{i}"), flush=True)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
torch.AcceleratorError: CUDA error: device kernel image is invalid
Search for `hipErrorInvalidImage' in https://rocm.docs.amd.com/projects/HIP/en/latest/index.html for more information.
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing AMD_SERIALIZE_KERNEL=3
Device-side assertion tracking was not enabled by user

```

3) With latest fix all scenarios have works as well as possible
    (Last test failed as expected in final test after kpack code object was unistalled for gfx90a/cuda:1
     but cuda:0 test still passed)

```
$ pip list installed | grep torch
amd-torch-device-gfx1100 2.12.0
amd-torch-device-gfx90a  2.12.0
torch                    2.12.0
torchvision              0.27.1 

$ python pytorch_cuda_0_test.py 
PyTorch version: 2.12.0
ROCm HIP version: 7.15.0
Device count: 2
cuda:0: AMD Radeon Pro W7900 Dual Slot
tensor([0.], device='cuda:0')

$ python pytorch_cuda_1_test.py 
PyTorch version: 2.12.0
ROCm HIP version: 7.15.0
Device count: 2
cuda:1: AMD Instinct MI210
tensor([0.], device='cuda:1')

$ pip uninstall amd-torch-device-gfx90a

$ python pytorch_cuda_0_test.py 
PyTorch version: 2.12.0
ROCm HIP version: 7.15.0
Device count: 2
cuda:0: AMD Radeon Pro W7900 Dual Slot
tensor([0.], device='cuda:0')

$ python pytorch_cuda_1_test.py 
PyTorch version: 2.12.0
ROCm HIP version: 7.15.0
Device count: 2
cuda:1: AMD Instinct MI210
Traceback (most recent call last):
  File "/home/butler/bugs/therock_6696/pytorch_cuda_1_test.py", line 7, in <module>
    print(torch.zeros(1, device=f"cuda:{i}"), flush=True)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
torch.AcceleratorError: CUDA error: invalid kernel file
Search for `hipErrorInvalidKernelFile' in https://rocm.docs.amd.com/projects/HIP/en/latest/index.html for more information.
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing AMD_SERIALIZE_KERNEL=3
Device-side assertion tracking was not enabled by user.

```

lucbruni-amd: What if no devices load? Is the behaviour expected/the same as previous? If we want to preserve behaviour we can track a `success_count` and return `hipErrorInvalidImage` when that is `0`.

lamikr:
Everything should now work as expected with the latest fix. Here is my test steps:

1) Test applications for cuda:0 and cuda:1

pytorch_cuda_0_test.py 

```
import torch

print("PyTorch version:", torch.__version__, flush=True)
print("ROCm HIP version:", torch.version.hip, flush=True)
print(f"cuda:0: {torch.cuda.get_device_name(0)}", flush=True)
print(torch.zeros(1, device=f"cuda:0"), flush=True)

```

and pytorch_cuda_1_test.py 

```
import torch

print("PyTorch version:", torch.__version__, flush=True)
print("ROCm HIP version:", torch.version.hip, flush=True)
print(f"cuda:1: {torch.cuda.get_device_name(1)}", flush=True)
print(torch.zeros(1, device=f"cuda:1"), flush=True)

```
2) GPU Devices available on current system

```
$ rocminfo | grep gfx
  Name:                    gfx1100                            
      Name:                    amdgcn-amd-amdhsa--gfx1100         
      Name:                    amdgcn-amd-amdhsa--gfx11-generic   
  Name:                    gfx90a                             
      Name:                    amdgcn-amd-amdhsa--gfx90a:sramecc+:xnack-

```

3) Install pytorch nightly to first show the bug

```
$ python3 -m venv .venv_nightly && source .venv_nightly/bin/activate
$ pip install --index-url https://rocm.nightlies.amd.com/whl-multi-arch torch[device-gfx90a,device-gfx1100] numpy

```


4) List torch and toch-devices with support installed

```
$ pip list installed | grep torch
amd-torch-device-gfx1100 2.13.0+rocm7.15.0a20260728
amd-torch-device-gfx110x 2.13.0+rocm7.15.0a20260728
amd-torch-device-gfx90a  2.13.0+rocm7.15.0a20260728
torch                    2.13.0+rocm7.15.0a20260728

```

5) Test with nightly build to see the crash on cuda1

cuda:0 works like expected

```
$ python pytorch_cuda_0_test.py 
PyTorch version: 2.13.0+rocm7.15.0a20260728
ROCm HIP version: 7.15.0
cuda:0: AMD Radeon Pro W7900 Dual Slot
tensor([0.], device='cuda:0')
```

cuda::1 crashes as described in https://github.com/ROCm/TheRock/issues/6696

```
$ python pytorch_cuda_1_test.py 
PyTorch version: 2.13.0+rocm7.15.0a20260728
ROCm HIP version: 7.15.0
cuda:1: AMD Instinct MI210
Segmentation fault (core dumped)
```

6) Update the libamdhip64.so.7 with the fixed version from this PR (I did manual build to get patched version)

```
$ cp fix_pr_9167/libamdhip64.so.7 .venv_nightly2/lib/python3.12/site-packages/_rocm_sdk_core/lib/libamdhip64.so.7
```

7) Test again and now both the cuda_0 and cuda_1 tests works

cuda:0 works like earlier without fix

```
$python pytorch_cuda_0_test.py
PyTorch version: 2.13.0+rocm7.15.0a20260728
ROCm HIP version: 7.15.0
cuda:0: AMD Radeon Pro W7900 Dual Slot
tensor([0.], device='cuda:0')
```

cuda:1 also works now with the fixed library as expected

```
$ python pytorch_cuda_1_test.py 
PyTorch version: 2.13.0+rocm7.15.0a20260728
ROCm HIP version: 7.15.0
cuda:1: AMD Instinct MI210
tensor([0.], device='cuda:1')
```

8) Additional test with only the gfx1100 device libraries installed to check that gfx1100 will still work
    and gfx90a will now fail due to missing gfx90a device libraries

Uninstall gfx90a support and list installed torch packages

```
$ pip uninstall amd-torch-device-gfx90a
$ pip list installed | grep torch
amd-torch-device-gfx1100 2.13.0+rocm7.15.0a20260728
amd-torch-device-gfx110x 2.13.0+rocm7.15.0a20260728
torch                    2.13.0+rocm7.15.0a20260728

```

cuda:0 test for gfx1100 to see it works

```
$ python pytorch_cuda_0_test.py 
PyTorch version: 2.13.0+rocm7.15.0a20260728
ROCm HIP version: 7.15.0
cuda:0: AMD Radeon Pro W7900 Dual Slot
tensor([0.], device='cuda:0')
```

cuda:1 test behaves now as expected (fails because device libs for gfx90a are not installed)
 
```
$ python pytorch_cuda_1_test.py 
PyTorch version: 2.13.0+rocm7.15.0a20260728
ROCm HIP version: 7.15.0
cuda:1: AMD Instinct MI210
Traceback (most recent call last):
  File "/home/butler/bugs/therock_6696/pytorch_cuda_1_test.py", line 6, in <module>
    print(torch.zeros(1, device=f"cuda:1"), flush=True)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
torch.AcceleratorError: CUDA error: invalid kernel file
Search for `hipErrorInvalidKernelFile' in https://rocm.docs.amd.com/projects/HIP/en/latest/index.html for more information.
For more detailed error information, run with CUDA_LOG_FILE=stderr
Device-side assertion tracking was not enabled by user.

```

9) install device-gfx90a back and uninstall device-gfx1100 so that test can only work with gfx90a and fail with gfx1100 due to missing gfx1100 device libraries

Install gfx90a support back

```
$ pip install --index-url https://rocm.nightlies.amd.com/whl-multi-arch torch[device-gfx90a,device-gfx1100] numpy
```

uinnstall gfx1100 support

```
$ pip uninstall amd-torch-device-gfx1100 amd-torch-device-gfx110x
```

List installed torch apps and device support

```
$pip list installed | grep torch
amd-torch-device-gfx90a 2.13.0+rocm7.15.0a20260728
torch                   2.13.0+rocm7.15.0a20260728

```

cuda:0 test behaves now as expected (failing due to missing gfx1100 device libraries)

```
$ python pytorch_cuda_0_test.py 
PyTorch version: 2.13.0+rocm7.15.0a20260728
ROCm HIP version: 7.15.0
cuda:0: AMD Radeon Pro W7900 Dual Slot
Traceback (most recent call last):
  File "/home/butler/bugs/therock_6696/pytorch_cuda_0_test.py", line 6, in <module>
    print(torch.zeros(1, device=f"cuda:0"), flush=True)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
torch.AcceleratorError: CUDA error: invalid kernel file
Search for `hipErrorInvalidKernelFile' in https://rocm.docs.amd.com/projects/HIP/en/latest/index.html for more information.
For more detailed error information, run with CUDA_LOG_FILE=stderr
Device-side assertion tracking was not enabled by user.

```

cuda:1 test will now work as expected because only gfx90a device libraries are installed

```
$ python pytorch_cuda_1_test.py 
PyTorch version: 2.13.0+rocm7.15.0a20260728
ROCm HIP version: 7.15.0
cuda:1: AMD Instinct MI210
tensor([0.], device='cuda:1')
```

marbre:
@benrichard-amd, @BrianHarrisonAMD as you contributed to kpack, would you be able to help reviewing?

benrichard-amd (approved):
LGTM

projects/clr/hipamd/src/hip_fatbin.cpp
```diff
@@ -758,60 +758,75 @@ hipError_t FatBinaryInfo::ExtractKpackBinary(const std::vector<hip::Device*>& de
...
+    kpack_error_t err =
+        kpack_load_code_object(getHipKpackCache(), params.metadata, fname_.c_str(),
+                               static_cast<uint32_t>(params.bundle_index), arch_ptrs.data(),
+                               arch_ptrs.size(), &code_object, &code_object_size);
+
+    if (err != KPACK_SUCCESS) {
+      LogPrintfWarning("Could not load kpack object for %s, err: %d",
+                     device_name.c_str(), err);
+      continue;
+    }
```
BrianHarrisonAMD: If this fails, and continues, I think we would get a success if everything fails.
I think we should track, and return fail if they all fail.

BrianHarrisonAMD: Also, we only check for != KPACK_SUCCESS, which will swallow the other type of errors.
We very likely want to scope this error check more carefully.

KPACK_ERROR_OUT_OF_MEMORY, KPACK_ERROR_INVALID_FORMAT, KPACK_ERROR_MSGPACK_PARSE_FAILED, KPACK_ERROR_DECOMPRESSION_FAILED, or KPACK_ERROR_IO_ERROR will now get handled the same as arch not found errors. These likely should fail loudly, instead of being treated as if they were unsupported.

If we separate out the error classes, and continue only on missing, then we should consider swapping the acceptable skip cases over to info instead of warnings. If it's an expected case, warning is a bit strong.

projects/clr/hipamd/src/hip_fatbin.cpp
BrianHarrisonAMD, line 1: I know testing this E2E in CI will be difficult, or impossible, since it requires a specific machine config, but is it possible to add some tests similar to the ones in [test_loader_api.cpp](https://github.com/ROCm/rocm-systems/blob/develop/shared/kpack/runtime/tests/test_loader_api.cpp)?

We could also extract pieces of the internal logic into free functions that are easier to test using fake data.

projects/clr/hipamd/src/hip_fatbin.cpp
```diff
@@ -758,60 +758,75 @@ hipError_t FatBinaryInfo::ExtractKpackBinary(const std::vector<hip::Device*>& de
...
+    for (auto device : matching_devices) {
+      hipError_t hip_err =
+          AddDevProgram(device, code_object, code_object_size, amd::Os::FDescInit());
+      if (hip_err != hipSuccess) {
+        LogPrintfError(
+            "Could not add kpack object for %s, device id: %d, err: %d",
+            device_name.c_str(), device->deviceId(), hip_err);
+        kpack_free_code_object(code_object);
+        return hip_err;
+      }
```
BrianHarrisonAMD, line 856: This was pre-existing before the changes here, but I asked AI about how this handles a partial success, and it flagged an issue:

```
 A successful AddDevProgram leaves three non-owning aliases to code_object:
     - amd::Program::binary_[&rootDev] (program.cpp:132), whose ~Program delete[]
       is gated on make_copy — false here (program.cpp:53-54)
     - ClBinary::origBinary_ via saveOrigBinary (devprogram.cpp:1509,
       device.hpp:1117-1120) — the comment on that line says outright that it isn't
       owned by ClBinary
     - ClBinary::binary_ via setBinary with allocated=false for unencrypted ELF, so
       BinaryAllocated is never set and release() won't delete it
       (device.cpp:1857-1872, 1751-1757)

   Nobody owns the buffer, so freeing it while an earlier device in the group has
   already succeeded leaves those aliases dangling. It isn't a double-free, but it
   is reachable on a normal path: Program::build re-reads the stored binary and
   re-enters addDeviceProgram with it whenever getDeviceProgram returns NULL
   (program.cpp:436-448), and FatBinaryInfo::BuildProgram goes straight there
   (hip_fatbin.cpp:880-886).
```

It recommended swapping to this instead:

```C++
       code_obj_allocations_.insert(code_object);   // hoist above the loop
       
        for (auto device : matching_devices) {
         hipError_t hip_err = AddDevProgram(device, code_object, code_object_size,
                                            amd::Os::FDescInit());
         if (hip_err != hipSuccess) {
           LogPrintfError(...);
           return hip_err;                          // drop the manual free
         }
       }

```

---

I don't think this change has to land with this PR, but thought I would flag it.

projects/clr/hipamd/src/hip_fatbin.cpp
lamikr: I rebased the old changes on top of the latest rocm-systems develop branch and added a new patch
that adds now more fine granular error checking. 

I also add a device and kpack code object memory release mechanism for cases they were partially loaded and we return an error. I tested this code-path by causing by purposely the error to happen.

Situations where the device type specific kpack object fails to load for some of the devices but succeeds for other device(s) in case that the kpack object was not installed for that device type are only causing a warning in this code. Later if that device is tried to use pytorch will handle that case correctly by throwing an exception.

Other errors than KPACK_ERROR_ARCHIVE_NOT_FOUND or KPACK_ERROR_ARCH_NOT_FOUND when trying to extract kpack binary will now cause error to be logged, memory cleaned if needed and error to be returned.

In case where device type specific kpack object is not installed for ANY of the devices, we will log error and return error.

Errors and warnings in this code requires using the AMD_LOG_LEVEL=1 (errors shown) or AMD_LOG_LEVEL=2 (also warnings shown) to be visible.

I have rerun all test case scenarios listed earlier and verified that they works now ok and checked also the generated warning and error messages.

projects/clr/hipamd/src/hip_fatbin.cpp
lamikr: I actually fixed this one also by adding a device and kobject device cleaning mechanism in case we return error.

projects/clr/hipamd/src/hip_fatbin.cpp
lamikr: We are planning to add tests to catch these errors but it will require a test that is dedicated to machines having 2 or more different types of gpus available and also checking that first the code-objects are available for all tested devices and tests are passing and then we need to uninstall those code objects and test again that test will fail properly.

lamikr:
Here are logs for additional tests run by using different AMD_LOG_LEVELS and combinations of installed torch device codes.

Installed torch and device support in the beginning:

```
$ pip list installed | grep torch
amd-torch-device-gfx1100 2.12.0
amd-torch-device-gfx90a  2.12.0
torch                    2.12.0
torchvision              0.27.1
```
Check that tests pass for both devices and now warning or error logs are generated

```
$ export AMD_LOG_LEVEL=2

$ python pytorch_cuda_0_test.py 
PyTorch version: 2.12.0
ROCm HIP version: 7.15.26305
cuda:0: AMD Radeon Pro W7900 Dual Slot
tensor([0.], device='cuda:0')

$ python pytorch_cuda_1_test.py 
PyTorch version: 2.12.0
ROCm HIP version: 7.15.26305
cuda:1: AMD Instinct MI210
tensor([0.], device='cuda:1')
```

Then uninstall gfx1100 support and check output on different logging levels when testing gfx1100.
Test should now fail because gfx1100 support is uninstalled.

```
$ pip uninstall amd-torch-device-gfx1100

$ export AMD_LOG_LEVEL=1

$ python pytorch_cuda_0_test.py 
PyTorch version: 2.12.0
ROCm HIP version: 7.15.26305
cuda:0: AMD Radeon Pro W7900 Dual Slot
Traceback (most recent call last):
  File "/home/butler/bugs/therock_6696/pytorch_cuda_0_test.py", line 6, in <module>
    print(torch.zeros(1, device=f"cuda:0"), flush=True)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
torch.AcceleratorError: CUDA error: invalid kernel file
Search for `hipErrorInvalidKernelFile' in https://rocm.docs.amd.com/projects/HIP/en/latest/index.html for more information.
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing AMD_SERIALIZE_KERNEL=3
Device-side assertion tracking was not enabled by user.

$ export AMD_LOG_LEVEL=2
$ python pytorch_cuda_0_test.py 
PyTorch version: 2.12.0
ROCm HIP version: 7.15.26305
cuda:0: AMD Radeon Pro W7900 Dual Slot
:2:hip_fatbin.cpp           :828 : 55533394099 us:  Could not load device type specific kpack object for ISA amdgcn-amd-amdhsa--gfx1100, err: 13, host binary: /home/butler/bugs/therock_6696/rockbuilder/.venv/lib/python3.12/site-packages/torch/lib/libtorch_hip.so
Traceback (most recent call last):
  File "/home/butler/bugs/therock_6696/pytorch_cuda_0_test.py", line 6, in <module>
    print(torch.zeros(1, device=f"cuda:0"), flush=True)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
torch.AcceleratorError: CUDA error: invalid kernel file
Search for `hipErrorInvalidKernelFile' in https://rocm.docs.amd.com/projects/HIP/en/latest/index.html for more information.
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing AMD_SERIALIZE_KERNEL=3
Device-side assertion tracking was not enabled by user.

```

Then do test for gfx90a with different log levels. Test should now pass because gfx90a support is installed but if log level is set to warning, it should show warning about failures to unload code for gfx1100 that is also available on the system.

```
$ export AMD_LOG_LEVEL=1
$ python pytorch_cuda_1_test.py 
PyTorch version: 2.12.0
ROCm HIP version: 7.15.26305
cuda:1: AMD Instinct MI210
tensor([0.], device='cuda:1')

$ export AMD_LOG_LEVEL=2

$ python pytorch_cuda_1_test.py 
PyTorch version: 2.12.0
ROCm HIP version: 7.15.26305
cuda:1: AMD Instinct MI210
:2:hip_fatbin.cpp           :828 : 53004876456 us:  Could not load device type specific kpack object for ISA amdgcn-amd-amdhsa--gfx1100, err: 13, host binary: /home/butler/bugs/therock_6696/rockbuilder/.venv/lib/python3.12/site-packages/torch/lib/libtorch_hip.so
:2:hip_fatbin.cpp           :828 : 53004906886 us:  Could not load device type specific kpack object for ISA amdgcn-amd-amdhsa--gfx1100, err: 13, host binary: /home/butler/bugs/therock_6696/rockbuilder/.venv/lib/python3.12/site-packages/torch/lib/libtorch_hip.so
:2:hip_fatbin.cpp           :828 : 53004917021 us:  Could not load device type specific kpack object for ISA amdgcn-amd-amdhsa--gfx1100, err: 13, host binary: /home/butler/bugs/therock_6696/rockbuilder/.venv/lib/python3.12/site-packages/torch/lib/libtorch_hip.so
:2:hip_fatbin.cpp           :828 : 53004931712 us:  Could not load device type specific kpack object for ISA amdgcn-amd-amdhsa--gfx1100, err: 13, host binary: /home/butler/bugs/therock_6696/rockbuilder/.venv/lib/python3.12/site-packages/torch/lib/libtorch_hip.so
:2:hip_fatbin.cpp           :828 : 53004939923 us:  Could not load device type specific kpack object for ISA amdgcn-amd-amdhsa--gfx1100, err: 13, host binary: /home/butler/bugs/therock_6696/rockbuilder/.venv/lib/python3.12/site-packages/torch/lib/libtorch_hip.so
:2:hip_fatbin.cpp           :828 : 53004949049 us:  Could not load device type specific kpack object for ISA amdgcn-amd-amdhsa--gfx1100, err: 13, host binary: /home/butler/bugs/therock_6696/rockbuilder/.venv/lib/python3.12/site-packages/torch/lib/libtorch_hip.so
tensor([0.], device='cuda:1')

```

Finally uninstall device code also for gfx90a so that none of the devices have it. Now error should be different.
First without not having amd tracing.

```
$ export AMD_LOG_LEVEL=0
$ python pytorch_cuda_1_test.py 
PyTorch version: 2.12.0
ROCm HIP version: 7.15.26305
cuda:1: AMD Instinct MI210
Traceback (most recent call last):
  File "/home/butler/bugs/therock_6696/pytorch_cuda_1_test.py", line 6, in <module>
    print(torch.zeros(1, device=f"cuda:1"), flush=True)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
torch.AcceleratorError: CUDA error: invalid kernel file
Search for `hipErrorInvalidKernelFile' in https://rocm.docs.amd.com/projects/HIP/en/latest/index.html for more information.
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing AMD_SERIALIZE_KERNEL=3
Device-side assertion tracking was not enabled by user.
```

Then with only trace errors added:

```
$ export AMD_LOG_LEVEL=1
$ python pytorch_cuda_1_test.py 
PyTorch version: 2.12.0
ROCm HIP version: 7.15.26305
cuda:1: AMD Instinct MI210
:1:hip_fatbin.cpp           :865 : 56158373905 us:  Could not find device type specific kpack code objects for any available device from binary: /home/butler/bugs/therock_6696/rockbuilder/.venv/lib/python3.12/site-packages/torch/lib/libtorch_hip.so
Traceback (most recent call last):
  File "/home/butler/bugs/therock_6696/pytorch_cuda_1_test.py", line 6, in <module>
    print(torch.zeros(1, device=f"cuda:1"), flush=True)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
torch.AcceleratorError: CUDA error: invalid kernel file
Search for `hipErrorInvalidKernelFile' in https://rocm.docs.amd.com/projects/HIP/en/latest/index.html for more information.
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing AMD_SERIALIZE_KERNEL=3
Device-side assertion tracking was not enabled by user.
```

And finally the warning traces also made visible.

```
$ export AMD_LOG_LEVEL=2
$ python pytorch_cuda_1_test.py
PyTorch version: 2.12.0
ROCm HIP version: 7.15.26305
cuda:1: AMD Instinct MI210
:2:hip_fatbin.cpp           :828 : 56341569019 us:  Could not load device type specific kpack object for ISA amdgcn-amd-amdhsa--gfx90a:sramecc+:xnack-, err: 13, host binary: /home/butler/bugs/therock_6696/rockbuilder/.venv/lib/python3.12/site-packages/torch/lib/libtorch_hip.so
:2:hip_fatbin.cpp           :828 : 56341569603 us:  Could not load device type specific kpack object for ISA amdgcn-amd-amdhsa--gfx1100, err: 13, host binary: /home/butler/bugs/therock_6696/rockbuilder/.venv/lib/python3.12/site-packages/torch/lib/libtorch_hip.so
:1:hip_fatbin.cpp           :865 : 56341569608 us:  Could not find device type specific kpack code objects for any available device from binary: /home/butler/bugs/therock_6696/rockbuilder/.venv/lib/python3.12/site-packages/torch/lib/libtorch_hip.so
Traceback (most recent call last):
  File "/home/butler/bugs/therock_6696/pytorch_cuda_1_test.py", line 6, in <module>
    print(torch.zeros(1, device=f"cuda:1"), flush=True)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
torch.AcceleratorError: CUDA error: invalid kernel file
Search for `hipErrorInvalidKernelFile' in https://rocm.docs.amd.com/projects/HIP/en/latest/index.html for more information.
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing AMD_SERIALIZE_KERNEL=3
Device-side assertion tracking was not enabled by user
```

Then device code object install again and successful test execution as expected. All trace levels are set and no warnings or errors are shown.

```
$ pip install rockbuilder/packages/wheels/torch/kpack-split/amd_torch_device_gfx90a-2.12.0-cp312-cp312-linux_x86_64.whl  rockbuilder/packages/wheels/torch/kpack-split/amd_torch_device_gfx1100-2.12.0-cp312-cp312-linux_x86_64.whl

$ export AMD_LOG_LEVEL=2

$ python pytorch_cuda_0_test.py 
PyTorch version: 2.12.0
ROCm HIP version: 7.15.26305
cuda:0: AMD Radeon Pro W7900 Dual Slot
tensor([0.], device='cuda:0')

$ python pytorch_cuda_1_test.py 
PyTorch version: 2.12.0
ROCm HIP version: 7.15.26305
cuda:1: AMD Instinct MI210
tensor([0.], device='cuda:1')
```

lamikr:
On Linux all tests passed.
On Windows 4 hip-tests are always failing. Same tests will fail on every run and I am seeing same failures also on other PRs like PR 9443 for rocm-systems. 

Failed tests are not directly related to kpack device object loading, so seems to be a non-related problem. Code has also been reviewed and I have got permissions to merge it so it would be available for therock 10.1 release.

gpt5.6 summary between two runs of ci-tests:

Failed tests happens on windows with test-sets:
 "Test hip-tests (shard 1,2,3,4 of 4)". 

Below is the gpt5.6 summary for failed tests between runs 3 and 5:

Persistent failures:

Unit_hipMemcpyBatchAsync_H2D_Functional
Unit_hipMemcpyBatchAsync_H2D_Pageable_Stream_SourceAccess
Unit_hipMemcpyBatchAsync_D2H_Functional
Unit_hipMemcpyDeviceToDeviceNoCU_NoCU_MulStrm

These reproduced with the same failure pattern: copy calls succeeded, but destination data was incorrect. These tests use memory copies without device kernels, so they do not directly exercise KPACK code-object loading.

projects/clr/hipamd/src/hip_fatbin.cpp
lamikr: That case is also handled in latest changes. If the kpack device objects are not available for any of the devices, we will now return error.

We are also separating the kpack archive missing case from other error codes that kpack load can return.

PR #9168, opened 2026-07-24
feat(rocr): add engine-type and SDMA engine-id queue info attributes

saleelk:
Extend hsa_amd_queue_get_info with HSA_AMD_QUEUE_INFO_ENGINE_TYPE and HSA_AMD_QUEUE_INFO_SDMA_ENGINE_ID so callers can query a queue's engine class and its resolved hardware SDMA engine id. Implement the queries for SDMA/AQL/AIE queues, forward them through intercept queues, and expose the entry point via CLR's dynamic ROCr interface.

## Issue Tracking
JIRA ID: AIRUNTIME-2549

## Submission Checklist

### DIFF
--- a/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
+++ b/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
@@ -562,6 +562,11 @@ hsa_status_t AqlQueue::GetInfo(hsa_queue_info_attribute_t attribute, void* value
     case HSA_AMD_QUEUE_INFO_VM_FAULT_REASON:
       *reinterpret_cast<uint32_t*>(value) = vm_fault_reason_;
       break;
+    case HSA_AMD_QUEUE_INFO_ENGINE_TYPE:
+      *static_cast<hsa_amd_queue_engine_t*>(value) = HSA_AMD_QUEUE_ENGINE_COMPUTE;
+      break;
+    case HSA_AMD_QUEUE_INFO_SDMA_ENGINE_ID:
+      return HSA_STATUS_ERROR_INVALID_ARGUMENT;
     default:
       return HSA_STATUS_ERROR_INVALID_ARGUMENT;
   }
--- a/projects/rocr-runtime/runtime/hsa-runtime/inc/hsa_ext_amd.h
+++ b/projects/rocr-runtime/runtime/hsa-runtime/inc/hsa_ext_amd.h
@@ -81,9 +81,10 @@
  * - 1.27 - hsa_amd_queue_signal_external_semaphore, hsa_amd_queue_wait_external_semaphore
  * - 1.28 - hsa_amd_agent_info_t: HSA_AMD_AGENT_INFO_HOST_ALLOC_DMABUF_SUPPORTED
  * - 1.29 - hsa_amd_image_create_v2, hsa_amd_interop_map_buffer_with_size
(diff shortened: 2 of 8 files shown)


projects/rocr-runtime/runtime/hsa-runtime/inc/hsa_ext_amd.h
```diff
@@ -4925,6 +4926,18 @@ typedef enum {
    * is true.  The type of this attribute is uint32_t.
    */
   HSA_AMD_QUEUE_INFO_VM_FAULT_REASON,
+  /*
+   * Hardware engine type targeted by this queue.
+   * The type of this attribute is hsa_amd_queue_engine_t.
+   */
+  HSA_AMD_QUEUE_INFO_ENGINE_TYPE,
```
cfreeamd, line 4933: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Optional]** This adds public ABI but there's no rocrtst coverage for the new attributes. A small test that creates a compute queue (expect `ENGINE_TYPE=COMPUTE`, `SDMA_ENGINE_ID`→`INVALID_ARGUMENT`) and an SDMA queue (expect `ENGINE_TYPE=SDMA` and a resolved, non-`ANY` engine id) would lock in the documented contract and guard against regressions. Non-blocking.

projects/rocr-runtime/runtime/hsa-runtime/core/runtime/amd_aql_queue.cpp
```diff
@@ -562,6 +562,11 @@ hsa_status_t AqlQueue::GetInfo(hsa_queue_info_attribute_t attribute, void* value
     case HSA_AMD_QUEUE_INFO_VM_FAULT_REASON:
       *reinterpret_cast<uint32_t*>(value) = vm_fault_reason_;
       break;
+    case HSA_AMD_QUEUE_INFO_ENGINE_TYPE:
```
cfreeamd, line 565: 🤖 AI-assisted pre-review (for @cfreeamd)

**[Nit]** Minor return-style inconsistency: `SdmaQueue::GetInfo` uses an explicit `return HSA_STATUS_SUCCESS;` per case, while here `ENGINE_TYPE` uses `break;` (falling through to the trailing success return) but `SDMA_ENGINE_ID` uses `return HSA_STATUS_ERROR_INVALID_ARGUMENT;`. Behavior is identical and correct — purely stylistic.

cfreeamd (commented):

Clean, ABI-safe addition. Verified: the two new selectors are appended at the end of `hsa_queue_info_attribute_t` (no value shift), the interface minor is bumped with a changelog line, `SdmaQueue::Initialize()` resolves `sdma_engine_id_` to a concrete non-ANY engine before the handle is returned (matching the documented contract), all concrete queue types are handled (Aql→COMPUTE, Aie→AIE, Sdma→SDMA; non-SDMA queues reject SDMA_ENGINE_ID), `HostQueue::GetInfo` remains an intentional unimplemented stub, and the `hsa_amd_queue_create` doc change is accurate (SDMA creation is implemented; only AIE is not). Two non-blocking notes inline.

cfreeamd (approved):

projects/rocr-runtime/runtime/hsa-runtime/inc/hsa_ext_amd.h
saleelk: There will be comprehensive test once I plug in more stuff and would exercise this path

gandryey (approved):

PR #9174, opened 2026-07-24
fix(rocprofiler-systems): Scope rocSHMEM's NUMA dependency to the rocshmem example

sputhala-amd:
## Motivation

A NUMA-enabled rocSHMEM's installed CMake config (`rocshmem-config.cmake`) calls `find_dependency(NUMA)` and references the `numa::numa` imported target in `roc::rocshmem`'s link interface. As a result, a standalone `find_package(rocshmem)` needs `libnuma` to be resolvable, or it fails to define `roc::rocshmem` and the example is skipped (or, for `REQUIRED` consumers, configuration aborts).

This was previously handled with a project-level `cmake/Modules/FindNUMA.cmake`. That placement is misleading: rocprofiler-systems itself does **not** depend on libnuma. NUMA is pulled in *only* transitively by the optional `rocshmem` example, and only for NUMA-enabled (GDA) rocSHMEM builds. 

Putting a `FindNUMA` module at the parent level implies the profiler has a libnuma dependency and adds it to the top-level module path for a concern that a single example owns.

This fix is not needed on builds that includes the upstream changes in rocshmem in https://github.com/ROCm/rocm-systems/pull/9583.

## Technical Details

- Removed `cmake/Modules/FindNUMA.cmake` — the parent build has no NUMA dependency, so it should not carry a NUMA finder.
- Moved the resolution into `examples/rocshmem/CMakeLists.txt`: before `find_package(rocshmem)`, sets NUMA_DIR to the NUMA config directory so rocSHMEM's internal `find_dependency(NUMA)` can resolve. This keeps the NUMA concern local to the one example that actually needs it.
- The change is a no-op on installs without that layout (the `EXISTS` guard), so non-TheRock configurations are unaffected. The example already loads rocSHMEM with `QUIET` and returns gracefully when it is unavailable.

## Issue Tracking

JIRA ID: AIPROFSYST-479

## Test Plan

TheRock bundles `libnuma` as a "sysdep" and ships a CMake **config package** (`numa-config.cmake`) under `<rocm>/lib/rocm_sysdeps/lib/cmake/NUMA`, which defines the `numa::numa` target (and `NUMA_INCLUDE_DIRS`/`NUMA_LIBRARIES` variables). Because that config lives in a nested prefix that isn't on the default search path, we simply add it to `CMAKE_PREFIX_PATH` — no bespoke `FindNUMA` module is required.

https://github.com/ROCm/TheRock/blob/21578c69b3f582ed13052c564e8ca59c7febeb9d/docs/development/dependencies.md?plain=1#L181

## Submission Checklist


dgaliffiAMD (approved):

dgaliffiAMD (approved):

projects/rocprofiler-systems/examples/rocshmem/CMakeLists.txt
```diff
@@ -24,11 +24,16 @@ if(NOT hip_FOUND)
...
+# A NUMA-enabled rocSHMEM's installed cmake config calls find_dependency(NUMA)
+# to resolve the numa::numa target its link interface references. TheRock ships
+# that resolver as numa-config.cmake under the nested rocm_sysdeps prefix
+# (<rocm>/lib/rocm_sysdeps/lib/cmake/NUMA), which is not on the default search
+# path. Add it so find_package(rocshmem) can satisfy find_dependency(NUMA).
+foreach(_rocm_root ${ROCmVersion_DIR} ${ROCM_PATH} /opt/rocm)
+    if(EXISTS "${_rocm_root}/lib/rocm_sysdeps/lib/cmake/NUMA")
+        list(PREPEND CMAKE_PREFIX_PATH "${_rocm_root}/lib/rocm_sysdeps")
+    endif()
+endforeach()
```
jrmadsen: Prepending `CMAKE_PREFIX_PATH` affects ALL subsequent package finds and (silently) overrides the user's ability to direct where packages are searched for first. This can lead to a lot of confusion and frustration on the users end when they are trying to figure out why their CMAKE_PREFIX_PATH (either in env or via cmdline) is not being respected since this modification won't show up in CMakeCache.txt.

jrmadsen (dismissed):

projects/rocprofiler-systems/examples/rocshmem/CMakeLists.txt
sputhala-amd: Thanks @jrmadsen for your review. I have made the required changes. However upstream changes in rocshmem in https://github.com/ROCm/rocm-systems/pull/9583 makes this change redundant.

sputhala-amd: @jrmadsen the upstream rochshmem changes did not make it to Rocm 10.0 builds, so its better to find the numa packages as a fallback if not already resolved by rocshem.

Please re-review.

projects/rocprofiler-systems/examples/rocshmem/CMakeLists.txt
```diff
@@ -24,18 +24,29 @@ if(NOT hip_FOUND)
...
+if(NOT TARGET roc::rocshmem AND NOT NUMA_DIR)
+    foreach(_rocm_root ${ROCmVersion_DIR} ${ROCM_PATH} /opt/rocm)
+        if(EXISTS "${_rocm_root}/lib/rocm_sysdeps/lib/cmake/NUMA/numa-config.cmake")
+            set(NUMA_DIR "${_rocm_root}/lib/rocm_sysdeps/lib/cmake/NUMA")
+            find_package(rocshmem HINTS ${ROCmVersion_DIR} ${ROCM_PATH} /opt/rocm QUIET)
+            break()
+        endif()
+    endforeach()
+endif()
+
```
ajanicijamd: The code in the fallback block has a flaw in the (admittedly unlikely) case that NUMA_DIR is set, but roc::rocshmem is not defined. In that case, we completely skip the block and `find_package(rocshmem...)` is never called, so `roc::rocshmem` target is never set.

The correct structure should separate those two concerns:

```suggestion

if(NOT TARGET roc::rocshmem)
    # Discover NUMA location only if user didn't provide one
    if(NOT NUMA_DIR)
        foreach(_rocm_root ${ROCmVersion_DIR} ${ROCM_PATH} /opt/rocm)
            if(EXISTS "${_rocm_root}/lib/rocm_sysdeps/lib/cmake/NUMA/numa-config.cmake")
                set(NUMA_DIR "${_rocm_root}/lib/rocm_sysdeps/lib/cmake/NUMA")
                break()
            endif()
        endforeach()
    endif()
    # Retry with whatever NUMA_DIR is now (user-provided or sysdeps-discovered)
    if(NUMA_DIR)
        find_package(rocshmem HINTS ${ROCmVersion_DIR} ${ROCM_PATH} /opt/rocm QUIET)
    endif()
endif()
```

sputhala-amd: Before this block on line 30 we try to `find_package(rocshmem)` and the current block is a fallback is that call fails.
There are multiple scenarios as below:
- If the user supplies a valid `NUMA_DIR` , line 31 succeeds on its own even on pre-#9583 builds, because rocSHMEM's bare `find_dependency(NUMA)` resolves NUMA through the user's `NUMA_DIR`. The fallback is never needed in that case. Valid `NUMA_DIR` -> line 31 succeeds -> `roc::rocshmem` defined -> fallback correctly skipped.
- `NUMA_DIR` is set and `roc::rocshmem` fails due to a different reason, lets say due to missing `rocprofiler-register`, in that case the fallback is not useful anyways, so skipping is good.
- The only case then remains `NUMA_DIR` is set by the user but its wrong path. Then `roc::rocshmem` will fail to be located and the fallback can actually be helpful to set the `NUMA_DIR` correctly.

I think the suggestion you provided is harmless and more defensive against a wrong path specified by the user as  `NUMA_DIR`.

ajanicijamd: This is a really unlikely scenario, but if the user specifies a wrong value for `NUMA_DIR`, cmake would just issue a warning that sounds something like `rocshmem-test skipped: rocSHMEM (roc::rocshmem) not found`. That could potentially be confusing to the user; but this is, as we agree, a very unlikely scenario, so I leave it to you to decide if it's worth changing the code for this.

sputhala-amd: On case 3, I tested this and found that when a user passes a bogus `NUMA_DIR`, the first `find_package(rocshmem)`'s internal `find_package(NUMA)` fails and rewrites `NUMA_DIR` to `NUMA_DIR-NOTFOUND`. And in this case also the fallback will be executed and hence even the 3rd case doesnt apply.

However I think your suggested code is more clear and more readable. Hence I changed the code block as per your suggestion.

projects/rocprofiler-systems/examples/rocshmem/CMakeLists.txt
```diff
@@ -24,18 +24,33 @@ if(NOT hip_FOUND)
...
 # builds we attempt a find_package.
 if(NOT TARGET roc::rocshmem)
     find_package(rocshmem HINTS ${ROCmVersion_DIR} ${ROCM_PATH} /opt/rocm QUIET)
 endif()
+
+# Fallback for older rocSHMEM packages whose installed cmake config calls a bare
+# find_dependency(NUMA) and cannot locate TheRock's sysdeps numa-config.cmake on
+# its own. Only reached when the first attempt failed; newer rocSHMEM
+# (rocm-systems#9583) resolves NUMA internally, so the first attempt already
+# succeeds and this whole block is skipped, deferring entirely to rocSHMEM.
```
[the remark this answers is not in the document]
sputhala-amd: Modified the PR description to match the code logic.

projects/rocprofiler-systems/examples/rocshmem/CMakeLists.txt
sputhala-amd: @jrmadsen - The merging for this PR is blocked by your review. I know you must be busy but checking one last time :-).

dgaliffiAMD (approved):

dgaliffiAMD:
`rocprofiler-compute` test failures are unrelated.

PR #4322, opened 2026-03-22
Add CPU unit tests for graph Rings and Trees

Praneeth-Vallem:
## Summary
- Add host-side unit tests for `ncclGetRings` (ring construction and validation) and `ncclGetTrees` (binary tree structure verification)
- Tests run on CPU without requiring a GPU, extending CI coverage for graph topology code paths

## Test plan
- [ ] Verify tests compile and pass via `cmake -DBUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Debug` + `ctest`
- [ ] Confirm no GPU required for these tests

PR #9177, opened 2026-07-24
Add CPU unit tests for graph Rings and Trees

Praneeth-Vallem:
## Summary
- Add host-side unit tests for ring construction/validation (`ncclBuildRings`) and binary/double-binary tree structure (`ncclGetBtree`, `ncclGetDtree`).
- Tests run on CPU without requiring a GPU, extending CI coverage for graph topology code paths.
- They build as part of `rccl-UnitTestsFixturesDebug` (Debug build, `ROCM_VERSION >= 60400`).

## History
This supersedes #4322, which was closed. That branch was rebased onto the current `develop` (its head is no longer a descendant of the closed PR's tip, so GitHub cannot reopen it), and the review comments from #4322 have been addressed.

## Changes since #4322
- Rebased onto latest `develop`; resolved the additive conflict in `test/CMakeLists.txt`.
- Clarified why `"trees.h"` and `"graph/rings.h"` use different include paths (`src/include/trees.h` vs `src/graph/rings.h`).
- Replaced VLAs in the tree consistency tests with `std::vector`.
- Added a `mirror()` helper and expressed the `ncclGetDtree` expectations directly in terms of it, replacing dense inline derivation comments.
- Extended `FourRanks_AllRanksValid` beyond bounds checks: both trees must have exactly one root and list every non-root rank as a child of its parent.
- Added btree parent/child consistency coverage for non-power-of-2 rank counts (5 and 6).
- Documented the ring-major `[ring * nranks + rank]` layout expected by `ncclBuildRings`, and that rejecting a broken ring is its documented contract (`ncclInternalError`), not undefined behavior.
- Added large-scale btree consistency tests (~1000 ranks: 1000, and 1023/1024/1025 around the power-of-2 boundary), with a spanning check that every rank is reachable from the root exactly once.

## Test plan
- [ ] Verify tests compile and pass via `cmake -DBUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Debug` + `ctest`
- [ ] Confirm no GPU required for these tests

### DIFF
--- a/projects/rccl/test/CMakeLists.txt
+++ b/projects/rccl/test/CMakeLists.txt
@@ -352,6 +352,8 @@ if(BUILD_TESTS)
       TransportTests.cpp
       TimeoutTests.cpp
       mem_manager/MemManagerTests.cpp
+      graph/RingsTests.cpp
+      graph/TreesTests.cpp
       graph/SearchTests.cpp
       graph/XmlTests.cpp
       graph/NetDevsPolicyP2pNetTests.cpp
--- a/projects/rccl/test/graph/RingsTests.cpp
+++ b/projects/rccl/test/graph/RingsTests.cpp
@@ -1,0 +1,244 @@
+/*************************************************************************
+ * Copyright (c) 2025 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * See LICENSE.txt for license information
+ ************************************************************************/
+
+// ncclBuildRings: given prev/next arrays from rank's perspective, produce the
+// ring traversal order starting from `rank`.
+//
+// rings[r*nranks + i] = the i-th rank visited in ring r, starting from `rank`.
+
+#include "nccl.h"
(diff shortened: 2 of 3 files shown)


Praneeth-Vallem:
Supersedes #4322 (closed; could not be reopened after the branch was rebased onto `develop`).

Addressed the review comments from #4322:
- Clarified the differing `trees.h` / `graph/rings.h` include paths.
- Replaced VLAs with `std::vector`.
- Added a `mirror()` helper and expressed `ncclGetDtree` expectations in terms of it.
- Extended `FourRanks_AllRanksValid` with full parent/child + single-root structural checks on both trees.
- Added non-power-of-2 coverage (5, 6) and large-scale coverage (~1000 ranks, plus the 1023/1024/1025 power-of-2 boundary) with a root-spanning check.
- Documented the ring-major layout and the broken-ring `ncclInternalError` contract.

